PR #1383, opened 2015-09-05
Introduce rules and sets

### LINKED ISSUE #866: Change way of setting fixers that should be run
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/805#issuecomment-66501096
> Just my two cents on the current situation. I think fixers shouldn't be in a level, they should just have a name, and then you would have groups of fixers with names too, and potentially groups of groups (I'd suggest some way of distinguishing them to avoid name clashes), and one of these groups would be the default group.
> 
> It would also make it a lot easier to say in the config: ->fixers(['@psr2', 'fixer-x']) if I want only psr-2 without psr-1 fixers (that may not make sense but you get my point). You could also have a @psr group that includes ['@psr0', '@psr1', '@psr2'] for example. And different themed groups maybe like @alignthings could include align_double_arrow and align_equals and any new alignment-related added in the future. Fixers could be in multiple groups too this way.
> 
> What I don't like right now is I can't trust the symfony/default level to not add new fixers I don't like, so I have to list all fixers I want to use explicitly, but then if new fixers are added I may miss them. So if I could "trust" the @psr group because I know for sure that any new fixer there will be ok with me, that would already shorten the list greatly :)
> 
> (by @Seldaek)

### LINKED ISSUE #1305: Changing The Header Comment Doesn't Flush The Cache

keradus:
Hi everyone!
Few changes that I have been working on lately...

I have tried to keep commits very clean to show what has changed step by step.
Below some description... Each point should match commit.
Please ignore merge commits, I will get rid of them later.

This PR may be big, so it may be worth to read changes commit by commit instead of all at once.
Please split discussion into two parts - first do we like changes described below and implemented, and then, if we agreed about changes, I will fill the docs and review of the code itself may begin.

1. **Introduce FixerFactory**
It's a class that will register build in fixers, allow to register custom fixers, sort them and attach config to them (see ConfigAwareInterface).
In future (see next changes in this PR) it will also allow to configure fixers.

2. **Replace level and fixers from configuration in favor of rules, remove getLevel from FixerInterface**
Closes #866 (ping @Seldaek).
Instead of setting level and fixers start setting rules. Many rules can be grouped in sets.
Eg to take all psr2 fixers, remove braces fixer and add strict fixer we can use:
CLI args:
`--rules=@PSR2,-braces,strict`
config file:
    ```php
    ->setRules([
        '@PSR2' => true,
        'braces' => false,
        'strict' => true,
    ])
    ```
One set can include other set (like `@Symfony` is `@PSR2` + extra fixers, `@PSR2` is `@PSR1` + other fixers).
When setting the rules set expand to it's definition and adding/removing fixers is doing step by step (so one may exclude set and then add one fixer from it).
For example, consider the following:
    ```
    @PSR1 = encoding,short_tag
    @PSR2 = @PSR1,braces,elseif
    @Symfony = @PSR2,return

    // then, one uses:
    --rules=strict,@Symfony,-@PSR2,@PSR1,-encoding
    // which expanded step by step into:
    --rules=strict,@PSR2,return,-@PSR2,@PSR1,-encoding
    --rules=strict,return,@PSR1,-encoding
    --rules=strict,return,encoding,short_tag,-encoding
    --rules=strict,return,short_tag
    ```
Also, remove getLevel method from FixerInterface.

3. **Add FixerInterface::configure method**
For now we have only way to turn fixer on or off, without possibility to configure how it should fix the code. This problem manifested first time with HeaderCommentFixer, when we have no way to configure the header itself!
Now, the fixers may have configuration (eg PhpUnitStrictFixer) and we have a way to change default configuration.
Setting the configuration for Fixers when using CLI tool is only possible in .php_cs file, not via CLI args.
In .php_cs file configuration is passed as part of rules definition:
    ```php
    ->setRules([
        '@PSR2' => true,
        'braces' => false,
        'strict' => true,
        'php_unit_strict' => [ 'assertEquals', 'assertNotEquals' ],
    ])
    ```
When a rule is false then fixer is off, when rule is truly then rule is on - especially when true it is on with default configuration and when it's array it is on and array is passed to configure method.
Also, remove Symfony\CS\Fixer\Contrib\HeaderCommentFixer::setHeader method

4. **Cache now detects configuration change not only by tool version and names of fixers, but also their configuration**
Closes #1305. (ping @GrahamCampbell)

5. **Disallow to set non-existing rules**

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -128,38 +128,34 @@ The ``--format`` option can be used to set the output format of the results; ``t
 
 The ``--verbose`` option will show the applied fixers. When using the ``txt`` format it will also displays progress notifications.
 
-The ``--level`` option limits the fixers to apply on the
+The ``--rules`` option limits the rules to apply on the
 project:
 
 .. code-block:: bash
 
-    php php-cs-fixer.phar fix /path/to/project --level=psr1
-    php php-cs-fixer.phar fix /path/to/project --level=psr2
-    php php-cs-fixer.phar fix /path/to/project --level=symfony
+    php php-cs-fixer.phar fix /path/to/project --rules=@PSR2
 
-By default, all PSR fixers are run. The "contrib
-level" fixers cannot be enabled via this option; you should instead set them
-manually by their name via the ``--fixers`` option.
+By default, all PSR fixers are run.
 
-The ``--fixers`` option lets you choose the exact fixers to
+The ``--rules`` option lets you choose the exact fixers to
 apply (the fixer names must be separated by a comma):
 
 .. code-block:: bash
 
-    php php-cs-fixer.phar fix /path/to/dir --fixers=linefeed,short_tag,indentation
+    php php-cs-fixer.phar fix /path/to/dir --rules=linefeed,short_tag,indentation
 
 You can also blacklist the fixers you don't want by placing a dash in front of the fixer name, if this is more convenient,
 using ``-name_of_fixer``:
 
 .. code-block:: bash
 
-    php php-cs-fixer.phar fix /path/to/dir --level=psr2 --fixers=-short_tag,-indentation
+    php php-cs-fixer.phar fix /path/to/dir --rules=-short_tag,-indentation
 
 When using combinations of exact and blacklist fixers, applying exact fixers along with above blacklisted results:
 
 .. code-block:: bash
 
-    php php-cs-fixer.phar fix /path/to/dir --level=psr2 --fixers=return,-short_tag
+    php php-cs-fixer.phar fix /path/to/project --rules=@Symfony,-@PSR1,-return,strict
 
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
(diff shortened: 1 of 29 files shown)


Symfony/CS/AbstractFixer.php
```diff
@@ -21,22 +21,9 @@
...
-        );
-
-        $level = current(explode('\\', substr(get_called_class(), strlen(__NAMESPACE__.'\\Fixer\\'))));
-
-        if (!isset($map[$level])) {
-            throw new \LogicException(sprintf('Can not determine Fixer level: "%s".', $level));
-        }
-
-        return $map[$level];
+        return;
```
GrahamCampbell: ?

keradus: > This PR may be big, so it may be worth to read changes commit by commit instead of all at once.
> Please split discussion into two parts - first do we like changes described below and implemented, and then, if we agreed about changes, I will fill the docs and review of the code itself may begin.

;)

GrahamCampbell: right, of course :)

Seldaek:
I think it makes sense, thanks for all the work!

keradus:
Thank you @Seldaek , it's great to see your opinion here again ! ;)

Any other thoughts guys ?

ping @GrahamCampbell @gharlan @ceeram @SpacePossum @stof @sstok @ccirello @localheinz (and everyone willing!)

GrahamCampbell:
Looks cool, but quite large to review. I'll have to block out like an hour at the weekend to take a proper look at this. ;)

keradus:
That is why I ask to talk about idea itself (event only the part described in PR descr) and after that take a look at code review itself. Thanks ;)

gharlan:
:+1: to pr description (did not look at code).

only one thing: Why introducing a new term `rules`? Why not `fixers`?

```
--fixers=@Symfony,-encoding

->setFixers([
    '@Symfony' => true,
    'encoding' => false,
]);
```

keradus:
> only one thing: Why introducing a new term rules? Why not fixers?

About Config. On 1.x line Config class has property `fixers` and both accessor and mutator. But what are that fixer?
When used before `FixCommand` (eg inside `.php_cs`), the `fixers` are *array of names*.
When used in/after `FixCommand` (eg in `Fixer`), the `fixers` are *array of instances*.

This is extra bad and I had to split it. So we have a Fixer, that is a class that fixes sth, and we have a rule, which is a string value that tells us should we use Fixer or not.

stof:
> ``Fixer::getLevelAsString`` is no longer static
> In non-tests code method is currently run in instance context and there is no way to mock static method.

why is this even described in the PR when the PR removes the concept of level entirely ?

> In .php_cs file configuration is passed as part of rules definition:

this makes it harder to handle the configuration IMO: it forces to add it explicitly instead of having it in groups. IMO, fixer configuration should be separate from the list of rules (and given that the list of rules can be configured in the CLI but config cannot, this makes even more sense).

What about something like this ?

```php
->configureFixer('php_unit_strict', [ 'assertEquals', 'assertNotEquals' ])
```

> Current behavior is very confusing and has so few user scenarios. The whole behavior changed when one add dash to fixers list.

WTF. I did not even know that the current behavior was so weird.


@keradus do you plan a way to define groups easily, so that ZF could define a ZF group for instance in some way for reusability by ZF modules, etc ?

keradus:
> why is this even described in the PR when the PR removes the concept of level entirely ?

I have separated concepts in separate commits, and this move was needed for some of next commit, and if we decided that some change is not good then it would be easier to pick revert only that atomic part.

---------------------------

> this makes it harder to handle the configuration IMO: it forces to add it explicitly instead of having it in groups. IMO, fixer configuration should be separate from the list of rules (and given that the list of rules can be configured in the CLI but config cannot, this makes even more sense).

I have taken inspiration here from ESLint. Splitting it would made sets harder - then they should have both list and configuration, and resolving nested sets will not be fun.
Splitting it won't make things easier for CLI - you will not be able to set fixer configuration as well. And when one will want to add that possibility to CLI - he will do it at same cost whatever variant will be choosen.
Also, putting it all togeter will make a whole fixers configuration _one_, simple data structure like JSON etc, allowing it to move it to separate file (eg, even in this PR one can simply `$config->setRules(loadJSONFile())`). This is great benefit IMO.

---------------------------

> @keradus do you plan a way to define groups easily, so that ZF could define a ZF group for instance in some way for reusability by ZF modules, etc ?

This PR does not focus on that part. It will just allow to have some sets. But yeah, I was thinking about it a bit.
The main point here I can see is that set must be published somewhere, allowing people to use it.
For now it may be published inside repo of PHP CS Fixer (like `@Symfony`). If one have scenario to register it outside, and when one want to use the tool he will need not only to download a tool, but also download a set definition and inject set into tool and then use it - I'm open for that, but that looks so similar to run `->setRules(...)` in `.php_cs`.
One way or another this PR will open a way for such functionality.

ccirello:
LGTM as it add flexibility both in configuration and call time.

keradus:
rebased on master

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -188,97 +184,17 @@ protected function doTestIntegration($testFileName, $testTitle, $fixers, $input,
      */
     protected function getFixersFromConfig($fileName, $config)
```
keradus: @SpacePossum, take a look at it now!

README.rst
```diff
@@ -666,30 +662,16 @@ The example below will add two contrib fixers to the default list of PSR2-level
...
-    <?php
-
-    $finder = Symfony\CS\Finder\DefaultFinder::create()
-        ->in(__DIR__)
-    ;
-
-    return Symfony\CS\Config\Config::create()
-        ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
-        ->fixers(array('trailing_spaces', 'encoding'))
+        ->rules(array(
```
gharlan: `setRules`

keradus: thx!

Symfony/CS/RulesSet.php
```diff
@@ -0,0 +1,192 @@
...
+        } while ($hasSet);
+
+        // filter out all rules that are off
+        $tmpRules = $rules;
+        $rules = array();
+        foreach ($tmpRules as $name => $value) {
+            if ($value) {
+                $rules[$name] = $value;
+            }
+        }
```
gharlan: `$rules = array_filter($tmpRules);` should be the same.

Symfony/CS/Tests/RulesSetTest.php
```diff
@@ -0,0 +1,156 @@
...
+            ),
+            $rulesSet->getRules()
+        );
+    }
+
+    private function assertSameRules(array $expected, array $actual, $message = '')
+    {
+        $this->assertSame(
+            ksort($expected),
+            ksort($actual),
```
gharlan: `ksort` has reference parameter and returns boolean.

Symfony/CS/RulesSet.php
```diff
@@ -0,0 +1,192 @@
...
+
+    private $set;
+    private $rules;
+
+    public static function create(array $set = array())
+    {
+        return new static($set);
+    }
+
+    public function __construct(array $set = array())
```
gharlan: `__construct` should be the first method (symfony standard).

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: all my code... it is gone...   looking good : D

Symfony/CS/Tests/Console/ConfigurationResolverTest.php
```diff
@@ -609,4 +410,63 @@ public function testResolveCacheFileWithConfigAndOption()
...
+            ),
+            $this->resolver->getRules()
+        );
+    }
+
+    private function assertSameRules(array $expected, array $actual, $message = '')
+    {
+        $this->assertSame(
+            ksort($expected),
+            ksort($actual),
```
gharlan: same (`ksort` reference parameter)

gharlan:
What do you think about using sf `OptionResolver` for fixer configurations?
(Similar to sf form types.)

```php
public function configureOptions(OptionsResolver $resolver)
{
    $resolver->setDefaults(array(
        // ,,,
    ));
    $resolver->setRequired(...);
    // etc.
}
```

http://symfony.com/doc/current/components/options_resolver.html

Symfony/CS/Tests/Console/ConfigurationResolverTest.php
keradus: done

Symfony/CS/RulesSet.php
keradus: done

Symfony/CS/Tests/RulesSetTest.php
keradus: done

Symfony/CS/RulesSet.php
keradus: done

keradus:
Then I think we have a consensus about idea itself.
Let us all take a deep look into the code :D

Symfony/CS/Console/ConfigurationResolver.php
```diff
@@ -467,6 +421,16 @@ private function resolveProgress()
     }
 
     /**
+     * Resolve rules.
+     */
+    private function resolveRules()
+    {
+        $rulesSet = new RulesSet($this->parseRules());
+
+        $this->rulesSet = $rulesSet;
```
gharlan: hmm, could be simplified ;)

keradus: thx, done ;)

keradus:
> What do you think about using sf OptionResolver for fixer configurations?
(Similar to sf form types.)

I would not change `FixerInterface::configure` method to get `OptionResolver` on input, but if we want we may use `OptionResolver` inside (hide it from user).
But we have so few and so simple configure cases, do we really need to resolve them by external class?

keradus:
updated to recent changes on repo (like extracting to changes into separate PRs)

keradus:
Diff has only 128KB,
looking good for me ;P

Symfony/CS/AbstractFixer.php
keradus: Most of our fixers don't need configuration - thus, in our abstract fixer, this method is empty to not force most of fixers implement it empty themselves

keradus:
anyone? :)

keradus:
Thank you for review @gharlan !

Then, I guess ready to merge

Symfony/CS/Tests/RulesSetTest.php
```diff
@@ -0,0 +1,164 @@
...
+        foreach (RulesSet::create()->getSetDefinitionNames() as $setName) {
+            $cases = array_merge($cases, RulesSet::create(array($setName => true))->getRules());
+        }
+
+        return array_map(function ($item) { return array($item); }, array_keys($cases));
+    }
+
+    /**
+     * @expectedException        \UnexpectedValueException
+     * @expectedExceptionMessage Set "@foo" does not exist.
```
GrahamCampbell: Can we do this with the fixers too?

keradus: That would be next BC break. Now when you use strict,fooo the fooo will be ignored silently and strict used.
I was thinking about it, wasn't sure - so just have a thought to move it for separate PR, when this one will be merge.

Symfony/CS/Config/Config.php
```diff
@@ -92,19 +89,7 @@ public function getFinder()
...
-        return $this;
-    }
-
-    public function getLevel()
-    {
-        return $this->level;
-    }
-
-    public function fixers($fixers)
+    public function fixers(array $fixers)
```
SpacePossum: could we add a PHPDoc on this what types are expected inside the array? I always get confused with the configuration between FixerInterface instances and names (strings).

keradus: Because on 1.x line this method can get both... names and instances... your confusion made the point.

TBD soon ;)

Symfony/CS/Tests/RulesSetTest.php
GrahamCampbell: Sure. I don't like the idea of silently ignoring, but perhaps exceptions are too harsh. We really need some kind of warning system.

Symfony/CS/FileCacheManager.php
```diff
@@ -33,18 +33,15 @@
 {
     private $cacheFile;
     private $isEnabled;
-    private $fixers;
+    private $rules;
     private $newHashes = array();
     private $oldHashes = array();
 
-    public function __construct($isEnabled, $cacheFile, array $fixers)
+    public function __construct($isEnabled, $cacheFile, $rules)
```
SpacePossum: what type is `$rules` here?

Symfony/CS/Tests/RulesSetTest.php
keradus: Wrong input is wrong input ;)

Symfony/CS/RulesSet.php
```diff
@@ -0,0 +1,228 @@
...
+namespace Symfony\CS;
+
+/**
+ * Set of rules to be used by fixer.
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+final class RulesSet implements RulesSetInterface
```
SpacePossum: it would be nice to provide an abstract base class as end point for others to create RulesSet.. you thinking while typing ;)

keradus: That is why I introduced the interface. Is it really needed to have abstract too? :(

Symfony/CS/Tests/RulesSetTest.php
SpacePossum: GIGO ;)

SpacePossum:
lots of work, looks great :) 
:+1: 

(especially because it will make things more easier when dealing with the configuration system)

Symfony/CS/Config/Config.php
keradus: done

Symfony/CS/FileCacheManager.php
keradus: done

Symfony/CS/Config/Config.php
```diff
@@ -92,19 +89,14 @@ public function getFinder()
...
-    public function getLevel()
-    {
-        return $this->level;
-    }
-
-    public function fixers($fixers)
+    /**
+     * Set fixers.
+     *
+     * @param $fixers FixerInterface[]
```
gharlan: wrong order

Symfony/CS/ConfigInterface.php
```diff
@@ -128,4 +121,24 @@ public function getCacheFile();
...
+     * Keys of array are names of fixers/sets, values are true/false.
+     *
+     * @return array
+     */
+    public function getRules();
+
+    /**
+     * Set rules.
+     *
+     * Keys of array are names of fixers/sets, values are true/false.
```
gharlan: values can be configuration arrays

Symfony/CS/RulesSet.php
GrahamCampbell: no real gain to abstracting it imo

Symfony/CS/ConfigInterface.php
keradus: Nice catch!

Symfony/CS/FileCacheManager.php
```diff
@@ -33,18 +33,24 @@
...
     private $newHashes = array();
     private $oldHashes = array();
 
-    public function __construct($isEnabled, $cacheFile, array $fixers)
+    /**
+     * Create instance.
+     *
+     * @param $isEnabled bool   is cache enabled
+     * @param cacheFile  string cache file
+     * @param $rules     array  array defining rules, format like one for ConfigInterface::setRules
```
gharlan: wrong order of types and variable names. missing one `$`.

Symfony/CS/Config/Config.php
SpacePossum: great, thank so much :)

Symfony/CS/RulesSet.php
SpacePossum: I see, never mind, you are both right, I'll come back when I've a valid use case

Symfony/CS/FixerFactory.php
```diff
@@ -0,0 +1,171 @@
...
+    private $fixers = array();
+
+    /**
+     * Create instance.
+     *
+     * @return FixerFactory
+     */
+    public static function create()
+    {
+        return new static();
```
gharlan: I would prefer `self` here (class if final).

Symfony/CS/FileCacheManager.php
keradus: done

Symfony/CS/Config/Config.php
keradus: done

Symfony/CS/FixerInterface.php
```diff
@@ -18,11 +18,17 @@
...
-    const PSR1_LEVEL = 3;
-    const PSR2_LEVEL = 7;
-    const SYMFONY_LEVEL = 15;
-    const CONTRIB_LEVEL = 32;
+    /**
+     * Set configuration.
+     *
+     * Some fixers may have no configuration, then - simply pass null.
+     * Other ones may have configuration that will change behavior of fixer,
+     * eg `php_unit_strict` fixer allow to configure which methods should be fixer.
```
gharlan: "allows" .. "fixed"

Symfony/CS/FixerFactory.php
keradus: Class is final, no difference, could be

Symfony/CS/Tests/RulesSetTest.php
keradus: added (last commit)

Symfony/CS/Tests/RulesSetTest.php
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public function provideAllRulesFromSets()
+    {
+        $cases = array();
+        foreach (RulesSet::create()->getSetDefinitionNames() as $setName) {
+            $cases = array_merge($cases, RulesSet::create(array($setName => true))->getRules());
+        }
+
+        return array_map(function ($item) { return array($item); }, array_keys($cases));
```
gharlan: cs

keradus: ?

Symfony/CS/FixerInterface.php
keradus: thx, done

Symfony/CS/FixerFactory.php
keradus: done

Symfony/CS/ConfigInterface.php
keradus: done

Symfony/CS/Tests/RulesSetTest.php
gharlan: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#6-closures

keradus: argh, so far thought that one-line lambdas are acceptable, thx

keradus: done

gharlan: No, but you could write it this way: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide-meta.md#31---multi-line-arguments-09082013

Symfony/CS/ConfigInterface.php
```diff
@@ -128,4 +121,27 @@ public function getCacheFile();
...
+     * Set rules.
+     *
+     * Keys of array are names of fixers or sets.
+     * Value for set must be bool (turn it on or off).
+     * Value for fixer may be bool (turn it on or off) or array of configuration
+     * (turn it on and contains configuration for FixerInterface::configure method).
+     *
+     * @param array $rules
+     *
+     * @return ConfigInterface The same instance
```
gharlan: `@return $this`

keradus: done

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -25,10 +25,11 @@
 use Symfony\CS\Error\Error;
 use Symfony\CS\Error\ErrorsManager;
 use Symfony\CS\Fixer;
+use Symfony\CS\FixerFactory;
 use Symfony\CS\FixerInterface;
 use Symfony\CS\Linter\Linter;
 use Symfony\CS\Linter\UnavailableLinterException;
-use Symfony\CS\Utils;
+use Symfony\CS\RulesSet;
```
GrahamCampbell: Is RuleSet not a better name?

keradus: It may be a good point, maybe even [Ruleset](http://www.yourdictionary.com/ruleset) ?

GrahamCampbell: Yeh, actually, that'd be better since "ruleset" is a word. :)

GrahamCampbell: Though, I think .net uses RuleSet rather than Ruleset in it's naming,=.

Symfony/CS/Config/Config.php
```diff
@@ -92,19 +89,14 @@ public function getFinder()
...
-
-    public function fixers($fixers)
+    /**
+     * Set fixers.
+     *
+     * @param FixerInterface[] $fixers
+     *
+     * @return $this
+     */
+    public function fixers(array $fixers)
```
gharlan: `fixers()` vs. `setRules()`.
We should use either `setFixers()` or `rules()`.

Symfony/CS/FileCacheManager.php
```diff
@@ -33,18 +33,24 @@
...
 
-    public function __construct($isEnabled, $cacheFile, array $fixers)
+    /**
+     * Create instance.
+     *
+     * @param bool   $isEnabled is cache enabled
+     * @param string $cacheFile cache file
+     * @param array  $rules     array defining rules, format like one for ConfigInterface::setRules
+     *
+     * @return $this
```
gharlan: A constructor does not have a return value.

Symfony/CS/FixerInterface.php
```diff
@@ -18,11 +18,17 @@
...
-    const CONTRIB_LEVEL = 32;
+    /**
+     * Set configuration.
+     *
+     * Some fixers may have no configuration, then - simply pass null.
+     * Other ones may have configuration that will change behavior of fixer,
+     * eg `php_unit_strict` fixer allows to configure which methods should be fixed.
+     * Finally, some fixers need configuration to work, eg `header_comment`.
+     *
+     * @param array $configuration configuration depends on Fixer
```
gharlan: can be `null`

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -607,10 +568,35 @@ function (FixerInterface $a, FixerInterface $b) {
             }
         }
 
+        $rulesSets = array();
+        foreach (RulesSet::create()->getSetDefinitionNames() as $setName) {
```
gharlan: Creating an empty ruleset to get all set names looks "wrong" to me.

Symfony/CS/Config/Config.php
keradus: There is bigger problem with config, not related to this PR.
First, we should use consistent names.
Second, we use Config to allow user configure the tool, then resolve it, write back to config the resolved configuration and use it inside other classes.
Thus, we should have separate Config for user configuration and Config with resolved configuration. But that should be done as separate PR.

Symfony/CS/Console/Command/FixCommand.php
keradus: Will gladly merge any "better" solution.

Symfony/CS/Config/Config.php
gharlan: ok

Symfony/CS/Console/Command/FixCommand.php
gharlan: Why not a static function here?

keradus: Have a static first, then I try to test them, mock them, and remembered why I hate statics.

gharlan: I hate avoiding statics by using weird workarounds ;)

If you do not want a static method here, you should move this part to another (maybe new) class in my opinion.

keradus: I see the current solution as not a pain, for sure won't have time to change in some way.
Moving to separate class that will only be an array in fact without *any* logic is pointless to me.

stof: moving the definition of rule sets to a separate class will probably happen in the future to provide extension point in them to support shared rule sets (for instance a ZF one or a Drupal one). But for now, it indeed looks useless to split it to a separate class (and could even hurt us later by having an api for sets which would not suit the extensible system).

gharlan: Accepted ;)

But it looks wrong to me to have a method for getting all ruleset names in the interface for _one_ ruleset (`RulesSetInterface`). And IMO it can hurt us later, too.

keradus: RuleSetInterface must know about sets to resolve them ;P

Symfony/CS/Console/Command/FixCommand.php
keradus: Done

Symfony/CS/FixerInterface.php
keradus: done

Symfony/CS/FileCacheManager.php
keradus: done

keradus:
Review issues were fixed.

gharlan:
Looks very good. Thanks!

keradus:
Thank you all for discussion and review.

SpacePossum:
very cool to see this in :+1:   thanks for all the work : D

Soullivaneuh:
@keradus How to remove all fixers by default with the new system?

Soullivaneuh:
Maybe just with `$config->fixers(array())`?

keradus:
Do not touch fixers at all, use only rules.

PR #1375, opened 2015-09-03
[2.x] Add integration tests using fixtures files.

SpacePossum:
There are good unit tests for fixers (and other parts of the code),
but there is no easy way to make integration tests.
  
This PR adds integration tests to the PHP-CS-Fixer code base.
It aims to makes it easy to tests (whole) PHP code (files) using multiple fixers.

Inspired by the fixtures in Twig.

Limitations:
* it doesn't test/do linting (by design)
* file based fixers won't work (like PSR0 and possibly PSR4 if that fixer is made)

A `.test` has the following structure:
```
--TEST--
My test title
--CONFIG--
level=none|psr0|psr1|psr2|symfony
fixers=name1,name2 (additonal fixers)
--fixers=name3,name4 (fixers not to use)
--INPUT--
some code
--EXPECT--
expected code after fixing
```

config items `fixers` and `--fixers` are optional
`--EXPECT--` is optional, if omitted the input is expected to stay the same.

Examples can be seen here:
* https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1375/files#diff-e283518231c001f9337e0093560cb730R1
* https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1375/files#diff-acdfc6b1dc6d83c7a9ae915af8acb5a1R1

PR #1382, opened 2015-09-05
Added PhpdocTypesFixer

GrahamCampbell:
Adding thing fixer required me to abstract the scalar fixer.

---

Replaces #1373.

PR #1384, opened 2015-09-05
Add intergration tests

SpacePossum:
A 1.11 version of;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1375
Biggest diff is the caused by the diff. ways the errormanager is connected to the fixer in the branches.

### DIFF
--- a/Symfony/CS/Tests/AbstractIntegrationTest.php
+++ b/Symfony/CS/Tests/AbstractIntegrationTest.php
@@ -1,0 +1,274 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests;
+
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\CS\ErrorsManager;
+use Symfony\CS\FileCacheManager;
+use Symfony\CS\Fixer;
+use Symfony\CS\FixerInterface;
+
+/**
+ * Integration test base class.
+ *
+ * This test searches for '.test' fixture files in the given directory.
+ * Each fixture file will be parsed and tested against the expected result.
+ *
+ * Fixture files have the following format:
+ *
+ * --TEST--
+ * Example test description
+ * --CONFIG--
+ * level=symfony|none|psr0|psr1|psr2|symfony
+ * fixers=fixer1,fixer2,...*
+ * --fixers=fixer3,fixer4,...**
+ * --INPUT--
+ * Code to fix
+ * --EXPECT--
+ * Expected code after fixing***
+ *
+ *   * Additional fixers may be omitted.
+ *  ** Black listed filters may be omitted.
+ * *** When the expected block is omitted the input is expected not to
+ *     be changed by the fixers.
+ *
+ * @author SpacePossum
+ *
+ * @internal
+ */
+abstract class AbstractIntegrationTest extends \PHPUnit_Framework_TestCase
+{
+    private static $builtInFixers;
+
+    public static function setUpBeforeClass()
+    {
+        $tmpFile = static::getTempFile();
+        if (!is_file($tmpFile)) {
+            $dir = dirname($tmpFile);
+            if (!is_dir($dir)) {
+                $fs = new Filesystem();
+                $fs->mkdir($dir, 0766);
+            }
+        }
+    }
+
+    public static function tearDownAfterClass()
+    {
+        unlink(static::getTempFile());
+    }
+
+    /**
+     * @dataProvider getTests
+     *
+     * @see doTestIntegration()
+     */
+    public function testIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
+    {
+        $this->doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected);
+    }
+
+    /**
+     * Creates test data by parsing '.test' files.
+     *
+     * @return array
+     */
+    public function getTests()
+    {
+        $fixturesDir = realpath($this->getFixturesDir());
+        if (!is_dir($fixturesDir)) {
+            throw new \UnexpectedValueException(sprintf('Given fixture dir "%s" is not a directory.', $fixturesDir));
+        }
+
+        $tests = array();
+
+        foreach (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($fixturesDir), \RecursiveIteratorIterator::LEAVES_ONLY) as $file) {
+            if (!preg_match('/\.test$/', $file)) {
+                continue;
+            }
+
+            $test = file_get_contents($file->getRealpath());
+            $fileName = $file->getFileName();
+            if (!preg_match('/--TEST--[\n](.*?)\s--CONFIG--[\n](.*?)\s--INPUT--[\n](.*?[\n]*)(?:[\n]--EXPECT--\s(.*)|$)/s', $test, $match)) {
+                throw new \InvalidArgumentException(sprintf('Test format invalid for "%s".', $fileName));
+            }
+
+            $tests[] = array($fileName, $match[1], $this->getFixersFromConfig($fileName, $match[2]), $match[3], isset($match[4]) ? $match[4] : null);
+        }
+
+        return $tests;
+    }
+
+    /**
+     * Returns the full path to directory which contains the tests.
+     *
+     * @return string
+     */
+    protected static function getFixturesDir()
+    {
+        throw new \BadMethodCallException('Method "getFixturesDir" must be overridden by the extending class.');
+    }
+
+    /**
+     * Returns the full path to the temporary file where the test will write to.
+     *
+     * @return string
+     */
+    protected static function getTempFile()
+    {
+        throw new \BadMethodCallException('Method "getTempFile" must be overridden by the extending class.');
+    }
+
+    /**
+     * Applies the given fixers on the input and checks the result.
+     *
+     * It will write the input to a temp file. The file will be fixed by a Fixer instance
+     * configured with the given fixers. The result is compared with the expected output.
+     * It checks if no errors were reported during the fixing.
+     *
+     * @param string           $testFileName Filename
+     * @param string           $testTitle    Test title
+     * @param FixerInterface[] $fixers       Fixers to use
+     * @param string           $input        Code to fix
+     * @param string|null      $expected     Expected result or null if the input is expected not to change
+     */
+    protected function doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
+    {
+        $errorsManager = new ErrorsManager();
+        $fixer = new Fixer();
+        $fixer->setErrorsManager($errorsManager);
+
+        $tmpFile = static::getTempFile();
+        if (false === @file_put_contents($tmpFile, $input)) {
+            throw new IOException(sprintf('Failed to write to tmp. file "%s".', $tmpFile));
+        }
+
+        $changed = $fixer->fixFile(new \SplFileInfo($tmpFile), $fixers, false, true, new FileCacheManager(false, null, $fixers));
+        $this->assertTrue($errorsManager->isEmpty(), 'Errors reported during fixing.');
+
+        if (null === $expected) {
+            $this->assertEmpty($changed, sprintf("Expected no changes made to test \"%s\" in \"%s\".\nFixers applied:\n\"%s\".\nDiff.:\n\"%s\".", $testTitle, $testFileName, $changed === null ? '[None]' : implode(',', $changed['appliedFixers']), $changed === null ? '[None]' : $changed['diff']));
+
+            return;
+        }
+
+        $this->assertNotEmpty($changed, sprintf('Expected changes made to test "%s" in "%s".', $testTitle, $testFileName));
+        $this->assertSame($expected, file_get_contents($tmpFile), sprintf('Expected changes do not match result, for "%s" in "%s".', $testTitle, $testFileName));
+
+        // run the test again with the `expected` part, this should always stay the same
+        $this->testIntegration($testFileName, $testTitle.' "--EXPECT-- part run"', $fixers, $expected);
+    }
+
+    /**
+     * Parses the '--CONFIG--' block of a '.test' file and determines what fixers should be used.
+     *
+     * @param string $fileName
+     * @param string $config
+     *
+     * @return FixerInterface[]
+     */
+    protected function getFixersFromConfig($fileName, $config)
+    {
+        static $levelMap = array(
+            'none' => FixerInterface::NONE_LEVEL,
+            'psr1' => FixerInterface::PSR1_LEVEL,
+            'psr2' => FixerInterface::PSR2_LEVEL,
+            'symfony' => FixerInterface::SYMFONY_LEVEL,
+        );
+
+        $lines = explode("\n", $config);
+        if (count($lines) < 1) {
+            throw new \InvalidArgumentException(sprintf('No configuration options found in "%s".', $fileName));
+        }
+
+        $config = array('level' => null, 'fixers' => array(), '--fixers' => array());
+
+        foreach ($lines as $line) {
+            $labelValuePair = explode('=', $line);
+            if (2 !== count($labelValuePair)) {
+                throw new \InvalidArgumentException(sprintf('Invalid configuration line "%s" in "%s".', $line, $fileName));
+            }
+
+            $label = strtolower(trim($labelValuePair[0]));
+            $value = trim($labelValuePair[1]);
+
+            switch ($label) {
+                case 'level' :
+                    if (!array_key_exists($value, $levelMap)) {
+                        throw new \InvalidArgumentException(sprintf('Unknown level "%s" set in configuration in "%s", expected any of "%s".', $value, $fileName, implode(', ', array_keys($levelMap))));
+                    }
+
+                    if (null !== $config['level']) {
+                        throw new \InvalidArgumentException(sprintf('Cannot use multiple levels in configuration in "%s".', $fileName));
+                    }
+
+                    $config['level'] = $value;
+                    break;
+                case 'fixers' :
+                case '--fixers' :
+                    $fixers = explode(',', $value);
+                    foreach ($fixers as $fixer) {
+                        $config[$label][] = strtolower(trim($fixer));
+                    }
+
+                    break;
+                default :
+                    throw new \InvalidArgumentException(sprintf('Unknown configuration item "%s" in "%s".', $label, $fileName));
+            }
+        }
+
+        if (null === $config['level']) {
+            throw new \InvalidArgumentException(sprintf('Level not set in configuration "%s".', $fileName));
+        }
+
+        if (null === self::$builtInFixers) {
+            $fixer = new Fixer();
+            $fixer->registerBuiltInFixers();
+            self::$builtInFixers = $fixer->getFixers();
+        }
+
+        $fixers = array();
+        for ($i = count(self::$builtInFixers) - 1; $i >= 0; --$i) {
+            $fixer = self::$builtInFixers[$i];
+            $fixerName = $fixer->getName();
+            if ('psr0' === $fixer->getName()) {
+                // File based fixer won't work
+                continue;
+            }
+
+            if ($fixer->getLevel() !== ($fixer->getLevel() & $levelMap[$config['level']])) {
+                if (false !== $key = array_search($fixerName, $config['fixers'], true)) {
+                    $fixers[] = $fixer;
+                    unset($config['fixers'][$key]);
+                }
+                continue;
+            }
+
+            if (false !== $key = array_search($fixerName, $config['--fixers'], true)) {
+                unset($config['--fixers'][$key]);
+                continue;
+            }
+
+            if (in_array($fixerName, $config['fixers'], true)) {
+                throw new \InvalidArgumentException(sprintf('Additional fixer "%s" configured, but is already part of the level.', $fixerName));
+            }
+
+            $fixers[] = $fixer;
+        }
+
+        if (!empty($config['fixers']) || !empty($config['--fixers'])) {
+            throw new \InvalidArgumentException(sprintf('Unknown fixers in configuration "%s".', implode(',', empty($config['fixers']) ? $config['--fixers'] : $config['fixers'])));
+        }
+
+        return $fixers;
+    }
+}
--- a/Symfony/CS/Tests/AbstractIntegrationTest.php
+++ b/Symfony/CS/Tests/AbstractIntegrationTest.php
@@ -215,8 +215,7 @@ abstract class AbstractIntegrationTest extends \PHPUnit_Framework_TestCase
                     break;
                 case 'fixers' :
                 case '--fixers' :
-                    $fixers = explode(',', $value);
-                    foreach ($fixers as $fixer) {
+                    foreach (explode(',', $value) as $fixer) {
                         $config[$label][] = strtolower(trim($fixer));
                     }
 
--- a/Symfony/CS/Tests/AbstractIntegrationTest.php
+++ b/Symfony/CS/Tests/AbstractIntegrationTest.php
@@ -42,7 +42,7 @@ use Symfony\CS\FixerInterface;
  * *** When the expected block is omitted the input is expected not to
  *     be changed by the fixers.
  *
- * @author SpacePossum
+ * @author SpacePossum <possumfromspace@gmail.com>
  *
  * @internal
  */
--- a/Symfony/CS/Tests/IntegrationTest.php
+++ b/Symfony/CS/Tests/IntegrationTest.php
@@ -14,7 +14,7 @@ namespace Symfony\CS\Tests;
 /**
  * Test that parses and runs the fixture '*.test' files found in '/Fixtures/Integration'.
  *
- * @author SpacePossum
+ * @author SpacePossum <possumfromspace@gmail.com>
  *
  * @internal
  */
--- a/Symfony/CS/Tests/IntegrationTest.php
+++ b/Symfony/CS/Tests/IntegrationTest.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests;
+
+/**
+ * Test that parses and runs the fixture '*.test' files found in '/Fixtures/Integration'.
+ *
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class IntegrationTest extends AbstractIntegrationTest
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected static function getFixturesDir()
+    {
+        return __DIR__.'/Fixtures/Integration';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected static function getTempFile()
+    {
+        return self::getFixturesDir().'/tmp/tmp.php';
+    }
+}
--- a/Symfony/CS/Fixer.php
+++ b/Symfony/CS/Fixer.php
@@ -90,6 +90,9 @@ class Fixer
         $this->fixers[] = $fixer;
     }
 
+    /**
+     * @return FixerInterface[]
+     */
     public function getFixers()
     {
         $this->sortFixers();
--- a/Symfony/CS/Tests/Fixtures/Integration/psr2.test
+++ b/Symfony/CS/Tests/Fixtures/Integration/psr2.test
@@ -1,0 +1,52 @@
+--TEST--
+PSR2 integration
+--CONFIG--
+level=psr2
+--INPUT--
+<?php
+namespace Vendor\Package;
+use FooInterface;
+use BarClass as Bar;
+use OtherVendor\OtherPackage\BazClass;
+class Foo extends Bar implements FooInterface{
+    public function sampleFunction($a, $b = null)
+    {
+        if ($a === $b) {
+            bar();
+        } elseif ($a > $b) {
+            $foo->bar($arg1);
+        } else {
+                BazClass::bar($arg2, $arg3);
+        }
+    }
+
+    static public  final function bar() {
+    // method body
+    }
+}
+--EXPECT--
+<?php
+namespace Vendor\Package;
+
+use FooInterface;
+use BarClass as Bar;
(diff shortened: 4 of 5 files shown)


Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+            throw new InvalidArgumentException(sprintf('Level not set in configuration "%s".', $fileName));
+        }
+
+        $fixer = new Fixer();
+        $fixer->registerBuiltInFixers();
+        $allFixers = $fixer->getFixers();
+
+        $fixers = array();
+        for ($i = count($allFixers) - 1; $i >= 0; --$i) {
+            /** @var FixerInterface $fixer */
```
GrahamCampbell: this shouldn't be needed

SpacePossum: it is needed.
Fixer's method `getFixers` has no PHPDoc, not does the returned property;
```php
    protected $fixers = array();
```

GrahamCampbell: Could you add phpdoc to that method please?

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+     */
+    public function testIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
+    {
+        $this->doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected);
+    }
+
+    public function getTests()
+    {
+        $fixturesDir = realpath($this->getFixturesDir());
+        $this->assertTrue(is_dir($fixturesDir), sprintf('Given fixture dir "%s" is not a directory.', $fixturesDir));
```
keradus: We should not test `getFixturesDir`, it is only a helper inside tests. Throw some kind of LogicException instead.

keradus:
It would be great to document `*.test` file structure

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+        $tests = array();
+
+        foreach (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($fixturesDir), \RecursiveIteratorIterator::LEAVES_ONLY) as $file) {
+            if (!preg_match('/\.test$/', $file)) {
+                continue;
+            }
+
+            $test = file_get_contents($file->getRealpath());
+            $fileName = $file->getFileName();
+            if (preg_match('/--TEST--[\r\n](.*?)\s--CONFIG--[\r\n](.*?)\s--INPUT--[\r\n](.*?[\r\n]*)(?:[\r\n]--EXPECT--\s(.*)|$)/s', $test, $match)) {
```
keradus:
```php
if (!preg_match...) {
    throw ...;
}

$testTitle = $match[1];
```

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+                $testTitle = $match[1];
+                $config = $this->getFixersFromConfig($fileName, $match[2]);
+                $input = $match[3];
+                $expected = isset($match[4]) ? $match[4] : null;
+                $tests[] = array($fileName, $testTitle, $config, $input, $expected);
```
keradus: I guess all that `testTitle, config, input, expected` variable aren't needed.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+            } else {
+                throw new InvalidArgumentException(sprintf('Test "%s" is not valid.', $fileName));
+            }
+        }
+
+        $this->assertNotEmpty($tests, sprintf('No tests found in fixtures dir "%s".', $fixturesDir));
```
keradus: Same here, it's not the error of code we try to test but the tests itself - that someone won't write tests well. I would throw an exception here.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+     * @param string           $testTitle
+     * @param FixerInterface[] $fixers
+     * @param string           $input
+     * @param string|null      $expected
+     */
+    protected function doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
+    {
+        $errorsManager = new ErrorsManager();
+        $fixer = new Fixer();
+        $fixer->setErrorsManager($errorsManager);
```
keradus: Why not `$errorsManager = $fixer->getErrorsManager();` ?

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+        $fs = new Filesystem();
+
+        $dir = $this->getTempDir();
+        $tmpDir = new \SplFileInfo($dir);
+        if (!$tmpDir->isDir()) {
+            $fs->mkdir($tmpDir);
+        }
+
+        $this->assertTrue($tmpDir->isDir(), sprintf('Given temp dir "%s" is not a directory.', $dir));
+        $this->assertTrue($tmpDir->isWritable(), sprintf('Given temp dir "%s" is not writable.', $dir));
```
keradus: It's not a real tests

Symfony/CS/Tests/AbstractIntegrationTest.php
GrahamCampbell: Yeh, exception.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+
+        $tmpFile = $tmpDir->getRealPath().'/tmp.php';
+        $this->assertNotFalse(@file_put_contents($tmpFile, $input), sprintf('Failed to write to tmp. file "%s".', $tmpFile));
+
+        $changed = $fixer->fixFile(new \SplFileInfo($tmpFile), $fixers, false, true, new FileCacheManager(false, null, $fixers));
+
+        if (null !== $expected) {
+            // read back the changed file content
+            $newContent = file_get_contents($tmpFile);
```
keradus: This shouldn't be needed. One IO operation more

Symfony/CS/Tests/IntegrationTest.php
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Symfony\CS\Tests;
+
+/**
+ * Test that parses and runs the fixture *.test files found in
+ * /Fixtures/Integration.
+ *
+ * @author SpacePossum
+ */
+class IntegrationTest extends AbstractIntegrationTest
```
keradus: What are the cases for extending `AbstractIntegrationTest` ? We have only one concrete class and it looks like it do all the job we may want - find all fixtures and test them.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+        $fs->remove($tmpFile);
```
keradus: There is no point in creating and removing file for each tests.
Too much IO operations.

Take a look at Linter - create one file at start, override a content when we want to test new content, and remove file at the end.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,260 @@
...
+                }
+            }
+        }
+
+        if (null === $config['level']) {
+            throw new InvalidArgumentException(sprintf('Level not set in configuration "%s".', $fileName));
+        }
+
+        $fixer = new Fixer();
+        $fixer->registerBuiltInFixers();
```
keradus: There should be one `$fixer` for all tests. Now we register built in fixers over and over again.

Symfony/CS/Tests/IntegrationTest.php
SpacePossum: This is an attempted to follow the suggestion of @stof here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1375#issuecomment-137393979
happy to revert

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: I did in the other PR, then it was requested to remove it again because it is not related to this test, ie. it needs to be done in another PR.

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: these are all used AFAIK

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: because that method does not exists in 1.x

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: check, done

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: if I do that it would be after `$fs->remove($tmpFile)`,
I guess this can be done after I refactored towards your RFC here  https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1384#discussion_r38820200

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: :+1:  done (I followed https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Test/IntegrationTestCase.php#L56 a bit to much..)

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: :+1:  done

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: test updated, it now re-write the tmp file every time

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: updated

Symfony/CS/Tests/AbstractIntegrationTest.php
keradus: :D 
my bad, sorry :D

Symfony/CS/Tests/IntegrationTest.php
keradus: Extended class that only override de facto 2 properties isn't much gain.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,226 @@
...
+            $test = file_get_contents($file->getRealpath());
+            $fileName = $file->getFileName();
+            if (!preg_match('/--TEST--[\r\n](.*?)\s--CONFIG--[\r\n](.*?)\s--INPUT--[\r\n](.*?[\r\n]*)(?:[\r\n]--EXPECT--\s(.*)|$)/s', $test, $match)) {
+                throw new InvalidArgumentException(sprintf('Test "%s" is not valid.', $fileName));
+            }
+
+            $tests[] = array($fileName, $match[1], $this->getFixersFromConfig($fileName, $match[2]), $match[3], isset($match[4]) ? $match[4] : null);
+        }
+
+        if (empty($tests)) {
```
keradus: if we return an empty array doesn't PHPUnit warn about that itself?

Symfony/CS/Tests/IntegrationTest.php
GrahamCampbell: Yeh, I agree. No point.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,226 @@
...
+            $labelValuePair = explode('=', $line);
+            if (2 !== count($labelValuePair)) {
+                throw new InvalidArgumentException(sprintf('Invalid configuration line "%s" in "%s".', $line, $fileName));
+            }
+
+            $label = strtolower(trim($labelValuePair[0]));
+            $value = trim($labelValuePair[1]);
+
+            switch ($label) {
+                case 'level' : {
```
keradus: I didn't saw that we use braces for cases.
It may even looking good, but we should be consistent.

.gitignore
```diff
@@ -1,3 +1,4 @@
 composer.lock
 vendor
 phpunit.xml
+Symfony/CS/Tests/Fixtures/Integration/tmp/*
```
keradus: It should not be needed. We should remove file at the end of tests.

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: no prloblemos :D 
let me know if I can make this easy as merge-able as possible ;)

keradus: I think we need to make this PR ready to merge first ;)
Then we will think about 2.x line.

keradus:
Travis failed.
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/78998867#L197

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: agreed ;)

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: yes, check removed

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: removed :+1:   (I still like those ;)  maybe a fixer for this would be nice... hmm.)

.gitignore
SpacePossum: revered

Symfony/CS/Tests/IntegrationTest.php
SpacePossum: >Extended class that only override de facto 2 properties isn't much gain.

I don't think that the amount of properties or methods you have to (or can) override says much about the gain of extending a class. 
In this case, just having to implement those two abstract methods should make it easy for others to setup a test to test their own set of test files in some other directory.

keradus: And how it would be easier than implementing non-abstract IntegrationTest ?

SpacePossum:
>There should be one $fixer for all tests. Now we register built in fixers over and over again.

This and remove the temp file is kind of a problem. PHPUnit has no hooks share objects between tests, nor does it have a hook to perform the deletion of the temp when the last test case has run.
The hook that is there `tearDownAfterClass` is static so it not useful since the temp file location is fetch dynamically.

Pointers about how to approach this are welcome :)

keradus:
Why name of file can't be static then? Same for fixers table.

Symfony/CS/Tests/IntegrationTest.php
SpacePossum: Not easier, the Integration test is not meant to be extended. That way it can still be changed within the PHP-CS-Fixer repo without being a BC-break, maybe I should make it final...

stof: @keradus AbstractIntegrationTest is meant to be part of the extension points for packages extending the tool with new fixers.
What we do in Symfony or Twig is that we put such classes in the ``Test`` namespace rather than ``Tests``: the class is part of the library API, not of the testsuite (and it is then used in the testsuite itself). This is even more important when the testsuite is in a separate folder which is not autoloadable at runtime (see Twig for instance)

keradus: Very well.
Then I suggest to merge it as-is, but having both class `@internal`, then, at 2.x the base class and fixture format will need to be refactred - and then we made base class `@api`.

After #1383 we should be very close to 2.x

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        if (!empty($config['fixers']) || !empty($config['--fixers'])) {
+            throw new \InvalidArgumentException(sprintf('Unknown fixers in configuration "%s".', implode(',', empty($config['fixers']) ? $config['--fixers'] : $config['fixers'])));
+        }
+
+        return $fixers;
+    }
+
+    public static function tearDownAfterClass()
```
keradus: from http://symfony.com/doc/current/contributing/code/standards.html

> Declare public methods first, then protected ones and finally private ones. The exceptions to this rule are the class constructor and the setUp and tearDown methods of PHPUnit tests, which should always be the first methods to increase readability;

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,233 @@
...
+                $fs = new Filesystem();
+                $fs->mkdir($dir, 0766);
+            }
+        }
+    }
+
+    /**
+     * @dataProvider getTests
+     */
+    public function testIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
```
keradus: params should be described, shortdoc would be cool too

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,233 @@
...
+    {
+        $this->doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected);
+    }
+
+    public function getTests()
```
keradus: output should be described, shortdoc would be cool too

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,233 @@
...
+
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\CS\ErrorsManager;
+use Symfony\CS\FileCacheManager;
+use Symfony\CS\Fixer;
+use Symfony\CS\FixerInterface;
+
+/**
+ * Integration test helper.
```
keradus: It is not a helper ;)
Also, long desc would be cool to descr what are integration tests.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1384#issuecomment-138062303

keradus:
It looks really great, I would be very happy to merge this PR after that few markers ;)

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,242 @@
...
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\CS\ErrorsManager;
+use Symfony\CS\FileCacheManager;
+use Symfony\CS\Fixer;
+use Symfony\CS\FixerInterface;
+
+/**
+ * Integration test helper. FIXME update this "It is not a helper ;)"
+ *
+ * FIXME add long description incl. fixtures file format
```
GrahamCampbell: todo

Symfony/CS/Tests/IntegrationTest.php
stof: @keradus why adding the feature in an internal way first to refactor it in 2.x then ? We should implement the feature in the expected way already (the internal implementation would of course change for 2.x, but not the user API)

Symfony/CS/Tests/AbstractIntegrationTest.php
keradus: changing this is pointless, it will be done before merge ;)

Symfony/CS/Tests/IntegrationTest.php
keradus: there will be major BC break on that feature after #1383

SpacePossum:
while the tests passes, it states:
```
[0Ktravis_time:start:06e52a26
[0K$ phpunit --verbose --coverage-clover build/logs/clover.xml
PHP Strict standards:  Static function Symfony\CS\Tests\AbstractIntegrationTest::getFixturesDir() should not be abstract in /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/Symfony/CS/Tests/AbstractIntegrationTest.php on line 116
```

is this something to worry about, I don't see why this is wrong, isn't this good use of late static binding?

keradus:
LSB is good, but in PHP one can not create method, which is both abstract and static.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,271 @@
...
+ * Integration test base class.
+ *
+ * This test searches for '.test' fixture files in the given directory.
+ * Each fixture file will be parsed and tested against the expected result.
+ *
+ * Fixture files have the following format:
+ *
+ * --TEST--
+ * Example test description
+ *  --CONFIG--
```
gharlan: one space before `--CONFIG--` should be removed

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,271 @@
...
+ * level=symfony|none|psr0|psr1|psr2|symfony
+ * fixers=fixer1,fixer2,...*
+ * --fixers=fixer3,fixer4,...**
```
gharlan: Why `--fixers` and not `-fixers`?
Or `fixers=fixer1,fixer2,-fixer3,-fixer4`?

keradus: I would not focus on that, we will change whole that --CONFIG-- section on 2.x line in favor of RulesSet

SpacePossum: >Why --fixers and not -fixers?

I took it from the command line options in the readme ;)

>fixers=fixer1,fixer2,-fixer3,-fixer4

I like that better, but if we are going to change it anyway, I leave it as is.

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: good catch, fixed

Symfony/CS/Tests/IntegrationTest.php
keradus: And clou is that #1383 will change user API we want to use here. No more levels etc

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,271 @@
...
+
+        return $tests;
+    }
+
+    /**
+     * Returns the full path to directory which contains the tests.
+     *
+     * @return string
+     */
+    abstract protected static function getFixturesDir();
```
keradus: Please, let me write a comment to remember that this method must not be both abstract and static.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,271 @@
...
+        $tests = array();
+
+        foreach (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($fixturesDir), \RecursiveIteratorIterator::LEAVES_ONLY) as $file) {
+            if (!preg_match('/\.test$/', $file)) {
+                continue;
+            }
+
+            $test = file_get_contents($file->getRealpath());
+            $fileName = $file->getFileName();
+            if (!preg_match('/--TEST--[\r\n](.*?)\s--CONFIG--[\r\n](.*?)\s--INPUT--[\r\n](.*?[\r\n]*)(?:[\r\n]--EXPECT--\s(.*)|$)/s', $test, $match)) {
```
keradus: `[\r\n]` ?
one char which may be `\n` (ok, we use that a lot!) or `\r` (we won't use single `\r` at all)

GrahamCampbell: yeh, we need to look for 0 or 1 \r and 1 \n

keradus: I would say it another way. Expect `\n` and nothing more. Follow PSR ;P

keradus: btw, I saw some sad systems in which you may use `\n\r` (not `\r\n`)

GrahamCampbell: hmmm, but then the tests will fail on a windows install out of the box won't they?

GrahamCampbell: isn't the default on windows to checkout using windows style and commit using unix style?

keradus: Don't know the default. And still don't know any dev that use `\r\n` in PHP code (even on Windows) or allow git to mess up with his files and change line ending. All IDEs on Win support `\n` as line end.

GrahamCampbell: > And still don't know any dev that use \r\n in PHP code (even on Windows) 

Yeh, I check out unix style on windows, but I think I had to change the default to do that.

SpacePossum: >And still don't know any dev that use \r\n in PHP code (even on Windows) or allow git to mess up with his files and change line ending.

I know of at least 5 developers/maintainers that do this (near me). The thing is IDEs like PHPStorm have a flag to not show white space changes. So as long as everyone on the team switches that one on (which is promoted) and doesn't look at the real changes (or github) and uses no PRs but just pushes to master; there is no problem at all ;)
(same for file permissions which Windows tends to change by default as well..)

back on topic: I'll remove the '\r' : )

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: What would you like me to add/change?
(It seems to work just fine)

keradus: It does not seems to work fine, you found it already:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1384#issuecomment-138591708

From #1382:
> I know, there is a problem to create abstract static method (only reasonable solution is to create it not abstract but throw sth like NotImplementedException.

(or make them non-static, but it is not a case here)

SpacePossum: Done, as this is indeed the only way out. Which is to bad because now we throw an exception during run time for something we could know is wrong on compile time.

I'm still puzzled about why this isn't allowed in PHP, since everything to support it seems to be there (after LSB was added).

keradus:
:+1:

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,277 @@
...
+
+                case 'fixers' :
+                case '--fixers' :
+                    $fixers = explode(',', $value);
+                    foreach ($fixers as $fixer) {
+                        $config[$label][] = strtolower(trim($fixer));
+                    }
+
+                    break;
+
```
GrahamCampbell: i thought our cs was to not have a blank line after break?

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,277 @@
...
+                        throw new \InvalidArgumentException(sprintf('Unknown level "%s" set in configuration in "%s", expected any of "%s".', $value, $fileName, implode(', ', array_keys($levelMap))));
+                    }
+
+                    if (null !== $config['level']) {
+                        throw new \InvalidArgumentException(sprintf('Cannot use multiple levels in configuration in "%s".', $fileName));
+                    }
+
+                    $config['level'] = $value;
+                    break;
+
```
GrahamCampbell: same

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: but have one above it? 
I always find these rules the most confusion ones, like the `return` white line rules of SF.

SpacePossum: In other words, if I puzzle a Fixer together that removes white lines below `break`, `return` and `continue` people would be OK with it to add to the repo?

keradus: +throws and configuration to override wich tokens should be and not fixed ;)

SpacePossum: very well, I write down the specs and come back later with a PR (might take some weeks ;) )

white lines are removed

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,274 @@
...
+                    if (null !== $config['level']) {
+                        throw new \InvalidArgumentException(sprintf('Cannot use multiple levels in configuration in "%s".', $fileName));
+                    }
+
+                    $config['level'] = $value;
+                    break;
+                case 'fixers' :
+                case '--fixers' :
+                    $fixers = explode(',', $value);
```
GrahamCampbell: uneeded asignment

SpacePossum: thanks, done

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -0,0 +1,273 @@
...
+ * Code to fix
+ * --EXPECT--
+ * Expected code after fixing***
+ *
+ *   * Additional fixers may be omitted.
+ *  ** Black listed filters may be omitted.
+ * *** When the expected block is omitted the input is expected not to
+ *     be changed by the fixers.
+ *
+ * @author SpacePossum
```
GrahamCampbell: missing email address

SpacePossum: I dont have one ^-^

gharlan: https://patch-diff.githubusercontent.com/raw/FriendsOfPHP/PHP-CS-Fixer/pull/1384.patch
line 2 ;)

SpacePossum: ha, nice find! 
oh damn, I was thinking I could PR anonymous ;)

keradus:
```
git show --format=%aE 9bd3da4
possumfromspace@gmail.com
```

works too ;P

keradus:
It is really a great job, thank you @SpacePossum !

keradus:
Ref #1384

SpacePossum:
Thank you all for reviewing and all the suggestions (and finding my email address back..;) )  :)
Nice merge, especially the one to master :+1: @keradus

keradus:
And now let me update my PR with rules/sets and FixerFactory :(

PR #1387, opened 2015-09-07
Made all test classes final

keradus:
Just a thought... shouldn't all tests be final?

PR #1388, opened 2015-09-08
Made all tests internal

keradus:
Continuation of #1387

### DIFF
--- a/Symfony/CS/Tests/AbstractFixerTest.php
+++ b/Symfony/CS/Tests/AbstractFixerTest.php
@@ -13,6 +13,8 @@ namespace Symfony\CS\Tests;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
  */
 final class AbstractFixerTest extends \PHPUnit_Framework_TestCase
 {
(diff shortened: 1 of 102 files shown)


keradus:
fabbot.io fails as always :(

stof:
@keradus what do you think about moving them to a separate ``tests`` folder at the root, which would not be autoloadable at runtime ?
This would cleanly separate the testsuite from the library, avoiding the need to mark anything as internal in the testsuite (as it would not be accessible at all anymore)

GrahamCampbell:
> @keradus what do you think about moving them to a separate tests folder at the root, which would not be autoloadable at runtime ?

I suggested this before, and @fabpot rejected the idea.

keradus:
> This would cleanly separate the testsuite from the library, avoiding the need to mark anything as internal in the testsuite (as it would not be accessible at all anymore)

keradus:
> This would cleanly separate the testsuite from the library, avoiding the need to mark anything as internal in the testsuite (as it would not be accessible at all anymore)

I can include classes from tests dir myself and use it, extend it ect. So marking them internal/final isn't pointless.

stof:
@keradus if you have to include them yourselves rather than relying on the autoloading, it is clear you are doing weird thing (and that you would ignore an ``@internal`` phpdoc anyway, which is harder to see than an autoloading failure)

keradus:
I' not saying here should we move tests or not, just saying that marking tests as final/internal is still gr8

keradus:
I agree that one should not include files from vendor himself, but if we mark tests as internal we won't lose anything

Issue #1386, opened 2015-09-07
ArrayElementSpaceFixer


gharlan:
An equivalent for [MethodArgumentSpaceFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/PSR2/MethodArgumentSpaceFixer.php) for array elements would be fine.

keradus:
:+1:

keradus:
Closing due to merge of #1484

PR #1389, opened 2015-09-08
[RFC] Added FinalInternalClassFixer

keradus:
Fixer that made all classes with `@internal` annotation a final.

Created during the work on #1386 and #1387

What do you think about it? Is it worth to continue work on that fixer?

### DIFF
--- a/Symfony/CS/Fixer/Contrib/FinalInternalClassFixer.php
+++ b/Symfony/CS/Fixer/Contrib/FinalInternalClassFixer.php
@@ -1,0 +1,83 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\DocBlock\DocBlock;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class FinalInternalClassFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_CLASS);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 1; 0 <= $index; --$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(T_CLASS)) {
+                continue;
+            }
+
+            $prevToken = $tokens[$tokens->getPrevMeaningfulToken($index)];
+
+            // ignore class if it is abstract or already final
+            if ($prevToken->isGivenKind(array(T_ABSTRACT, T_FINAL))) {
+                continue;
+            }
+
+            $docToken = $tokens[$tokens->getPrevNonWhitespace($index)];
+
+            // ignore class if it has no class-level docs
+            if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $doc = new DocBlock($docToken->getContent());
+
+            // ignore class if it is has no @internal annotation
+            if (empty($doc->getAnnotationsOfType('internal'))) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmmm. Can we also say that abstract internal classes shouldn't be made final?

GrahamCampbell:
Oh, I see you did. Just read the code rather than the description. ;)

keradus:
They simply cannot, abstract final is error ;)

GrahamCampbell:
Yeh, I know. I was just saying that your implementation shouldn't try to "make all internal classes final", but as I said, I then read the implementation, and you deal with abstract classes correctly.

keradus:
Yeah, just have answered and then saw your next comment :D

SpacePossum:
This would stop people from extending internal classes internally as well right? Isn't that a valid use case?

GrahamCampbell:
No, it's not. Internal classes should NEVER be used outside the library.

SpacePossum:
>No, it's not. Internal classes should NEVER be used outside the library.

I asked about: ' extending internal classes internally ', so I meant inside the library.

GrahamCampbell:
Ah, yeh, that's allowed if their abstract. :)

gharlan:
Hmm, and why not if they are not abstract?
It doesn't make sense to me. 

> Isn't that a valid use case?

We may do not have it in this repo, but for me it is a valid use case.

GrahamCampbell:
What's the status on this btw?

keradus:
Need a thought what to do when `@internal` class must not be marked as final.
Could we be aware of that situation inside of fixer's code?

keradus:
Or maybe, allowing the fixer to do his job blindly.
Mark the fixer as risky and don't run it by default.
Then indeed, could be some false positives.

If not, then I will simply throw this fixer away.

What do you think ?

gharlan:
> What do you think ?

Throw it away ;)

keradus:
Before:
> All internal classes should be final except abstract ones.
> Warning! This could change code behavior.

After:
> Non-abstract, internal classe that implements interface and is not Doctrine entity should be final.
> Warning! This could change code behavior.

Then, it won't do the work so much blindly and may be quite useful.

Slamdunk:
I would add

> Non-abstract, internal classes that implement interface __or extend a class__, and is not Doctrine entity should be final.

It is legit to use abstract classes (implementing interfaces) with the template method pattern (all methods signed as final, and one/some abstract protected method), so a simple class extending another class have to be considered as well to be final.

Definitively thumb up for this fixer.

keradus:
`all methods signed as final, and one/some abstract protected method`
but then, that class is abstract, right?
so it already matches `non-abstract` condition

Slamdunk:
I wasn't clear in my statement :grimacing: the subject of my intentions was the last part of the phrase.

I meant, not only classes that implement interfaces, but also classes that simply extends another class.
Why? Because they should extends only an abstract class, and not a real class.
So,
```php
class MyClass extends UnidentifiedClass
```
Should be marked as well as final no matter what `UnidentifiedClass` is, because the only legit ipothesis is that `UnidentifiedClass` is abstract.

SpacePossum:
:+1:  I like the current candidate selection, I would fix classes that marked as `@final` as well (even if these miss `@internal`)

SpacePossum:
Upping this to current state should not be that hard and adding the check for `@final`.
I would like this fixer, I know I didn't feel much for it at the start, but more and more I've valid use cases for this one. So, @keradus, could you update this one and wrap it up and ship it? :)

TomasVotruba:
Related fixer for inspiration: https://github.com/Slamdunk/php-cs-fixer-extensions/blob/master/lib/FinalInternalClassFixer.php

Symfony/CS/Fixer/Contrib/FinalInternalClassFixer.php
```diff
@@ -0,0 +1,83 @@
...
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(T_CLASS)) {
+                continue;
+            }
+
+            $prevToken = $tokens[$tokens->getPrevMeaningfulToken($index)];
+
+            // ignore class if it is abstract or already final
+            if ($prevToken->isGivenKind(array(T_ABSTRACT, T_FINAL))) {
```
keradus: `T_NEW` is missing

SpacePossum: you want me to pick this one up? 2 years in the making and I really like to get this on my (closed source) projects :)

keradus: that would be lovely :+1:

SpacePossum:
replaced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3183

PR #1390, opened 2015-09-08
Added ProjectCodeTest that tests if all classes inside tests are internal and final or abstract

keradus:
I have treated [that comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1387#issuecomment-138674838) as a request!

ping @GrahamCampbell

### DIFF
--- a/Symfony/CS/Tests/ProjectCodeTest.php
+++ b/Symfony/CS/Tests/ProjectCodeTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\CS\Tests;
 
 use Symfony\Component\Finder\Finder;
+use Symfony\CS\DocBlock\DocBlock;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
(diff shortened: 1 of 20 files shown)


GrahamCampbell:
Nice. :)

Symfony/CS/Tests/ProjectCodeTest.php
```diff
@@ -0,0 +1,83 @@
...
+            }
+        );
+    }
+
+    private function registerAllProjectClasses()
+    {
+        $finder = Finder::create()
+            ->files()
+            ->in(__DIR__.'/..')
+            ->exclude(array(
```
SpacePossum: I would add `->name('*.php')` as well, so `readme`, `.php_cs`, etc. can still be added later on into the directories (when needed).

SpacePossum:
Look good :)

Symfony/CS/Tests/ProjectCodeTest.php
```diff
@@ -0,0 +1,83 @@
...
+        if (null === $classes && null === $projectClasses) {
+            $this->registerAllProjectClasses();
+            $projectClasses = $this->getClasses('Symfony\\CS\\', get_declared_classes());
+        }
+
+        if (null === $classes) {
+            $classes = $projectClasses;
+        }
+
+        $ns = $prefix;
```
stof: ``$ns`` is unused

keradus:
Now test internal too ;)

PR #1391, opened 2015-09-09
Fixer::getLevelAsString is no longer static

keradus:
In non-tests code method is currently run in instance context and there is no way to mock static method.

Extracted from #1383, this little change looks quite ready and #1383 may have long review process.


SpacePossum:
:+1:

PR #1393, opened 2015-09-11
boost .travis.yml file

keradus:
Beceause nobody likes slow CI...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,10 +9,8 @@ matrix:
         - php: 5.4
         - php: 5.5
         - php: 5.6
-          env: 'COVERAGE=true'
         - php: 7.0
         - php: hhvm
-          env: 'COVERAGE=true'
 
 sudo: false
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmmm, though PHP 5.4 has different execution paths to 5.6 in some places, which is why coverage was running on every vesion, because we have version specific execution paths.

keradus:
We execute more path on newer and newer versions.
Do we indeed do sth only in old versions?

GrahamCampbell:
Well some paths are only executed on old versions?

keradus:
like?

GrahamCampbell:
```
        if (
            '' === $old
            || !$fileCacheManager->needFixing($this->getFileRelativePathname($file), $old)
            // PHP 5.3 has a broken implementation of token_get_all when the file uses __halt_compiler() starting in 5.3.6
            || (PHP_VERSION_ID >= 50306 && PHP_VERSION_ID < 50400 && false !== stripos($old, '__halt_compiler()'))
        ) {
            if ($this->eventDispatcher) {
                $this->eventDispatcher->dispatch(
                    FixerFileProcessedEvent::NAME,
                    FixerFileProcessedEvent::create()->setStatus(FixerFileProcessedEvent::STATUS_SKIPPED)
                );
            }

            return;
        }
```

keradus:
Then, 5.3, 5.6 and hhvm should be enought?

GrahamCampbell:
Yes, though do we actually have coverage for that anyway? if not, then we don't need 5.3.

keradus:
For now we have. Question is do we want to collect 5.3 (or 5.5) coverage on Travis.
For now we do not collect it for php7 (no support fod xdebug)

GrahamCampbell:
> For now we do not collect it for php7 (no support fod xdebug)

You don't need xdebug. On php 7, you can collect coverage using phpdbg and phpunit 5.

GrahamCampbell:
Travis only has an old vesion of phpunit though, and I don't think phpdbg is available - I've never tried.

keradus:
updated to run coverage on every available php versions, but still a bit faster ;)

GrahamCampbell:
:+1:

PR #1394, opened 2015-09-13
Stop mixing level from config file and fixers from CLI arg when one of fixers has dash

keradus:
**Stop mixing level from config file and fixers from CLI arg when one of fixers has dash**
Current behavior is very confusing and has so few user scenarios. The whole behavior changed when one add dash to fixers list.
So let us stop mixing the fixers configuration from CLI args and from config file.
One can still use `--level=psr2 --fixers=strict,-braces` to create target group of fixers from psr2 plus strict minus braces,
same with config file, but cannot mix fixers from CLI arg with level from some magic place.
Previous behavior description:
`--fixers=fixer1,fixer2,fixer3,fixer4,fixer5,fixer6,fixer7,fixer8`
which creates a group of 8 fixers (and ignore level set in config file)
and
`--fixers=fixer1,fixer2,fixer3,fixer4,-fixer5,fixer6,fixer7,fixer8`
which creates a group of fixers in the following steps:
  - get all fixers for level in config file (**unlike previously**!) or default one,
  - add 7 fixers from `--fixers` arg
  - remove fixer5 (from `--fixers` arg)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -153,13 +153,13 @@ using ``-name_of_fixer``:
 
 .. code-block:: bash
 
-    php php-cs-fixer.phar fix /path/to/dir --fixers=-short_tag,-indentation
+    php php-cs-fixer.phar fix /path/to/dir --level=psr2 --fixers=-short_tag,-indentation
 
 When using combinations of exact and blacklist fixers, applying exact fixers along with above blacklisted results:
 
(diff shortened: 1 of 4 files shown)


keradus:
extracted from #1383

GrahamCampbell:
:+1:

PR #1395, opened 2015-09-13
code grooming

keradus:


Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -213,14 +213,14 @@ protected function getFixersFromConfig($fileName, $config)
 
                     $config['level'] = $value;
                     break;
-                case 'fixers' :
```
GrahamCampbell: we need a fixer to remove that whitespace tbh

keradus: why not, go ahead!
my be tricky thanks to ternary operator

PR #1398, opened 2015-09-14
[2.0] Added more method separation tests

GrahamCampbell:


GrahamCampbell:
Good to have lots of test coverage. :)

Symfony/CS/Tests/Fixer/Symfony/MethodSeparationFixerTest.php
```diff
@@ -528,4 +528,14 @@ class ezcReflectionMethod extends ReflectionMethod {
 
         return $cases;
     }
+
+    public function testFixPlainFunction()
```
stof: don't create new tests. Add cases in the existing data providers instead.

keradus:
?

PR #1348, opened 2015-08-09
Linter - Prefer error output when throwing a linting exception

GrahamCampbell:
The error output provides information that's actually useful to debug the error.

PR #1400, opened 2015-09-14
appveyor.yml - add file

keradus:
Let's keep this PR minimal and then, if we want to, extend jobs run on AppVeyor in separate PR

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,30 @@
+build: off
+version: dev-{build}
+shallow_clone: false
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
I'm kind of :-1: here.

keradus:
Great argumentation ;)
We already had problems with Unix ws Windows - remember calling #1348 ? With AppVeyor we could cover such cases with utests.

sstok:
:+1:

PR #1366, opened 2015-08-29
Escaped PHP executable path in Process' command line

mdecaffmeyer:
In response to issue #1364

Issue #1364, opened 2015-08-27
Files that were not fixed due to errors reported during linting before fixing


mdecaffmeyer:
Whenever I try to format files with php-cs-fixer, I have the following message:

"Files that were not fixed due to errors reported during linting before fixing"

The files have correct syntax, it even used to work before on those same files.
When using --verbose option, I see that files are marked:
I-invalid file syntax, file ignored

The version is 2.0-DEV

keradus:
Please, try to fix a single file like `php php-cs-fixer fix my\file.php` until you find file that produces error, then please post the file. We will try to figure it out then

GrahamCampbell:
Or turn off linting, then look at the output. You might easily be able to see the file with the syntax error.

mdecaffmeyer:
```
<?php
class Test {
}
```
> php-cs-fixer fix --verbose Test.php
**I**
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes,
 F-fixed, E-error
Fixed all files in 0.040 seconds, 4.500 MB memory used
**Files that were not fixed due to errors reported during linting before fixing:**
   1) C:\data\websites\dokagenda-v3\Model\Test.php

How do I turn off linting?

GrahamCampbell:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Config/Config.php#L63

keradus:
Try setting php executable (available on 2.x line): https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Config/Config.php#L188

mdecaffmeyer:
1) With ```->setUsingLinter(false)``` it works, except if I specify a single file, then I have the same message as in intiial post.

```php-cs-fixer fix --verbose .```

Outputs me something like FFFFF etc.
When I executed it once more, it gave SSSSS etc.

```php-cs-fixer fix --verbose Test.php```

Gave me the intial error. Looks like ```->setUsingLinter(false)``` is only effective when fixing a directory of files, and thus doesn't work when trying to fix a single file.

But what exactly is linting and do I need it, enable it and stil try to find the error?

2) @keradus I don't know what i should put in this option, sorry.
[...]

PR #1403, opened 2015-09-17
Correctly escape the php binary when linting

GrahamCampbell:
Replaces #1366

### DIFF
--- a/Symfony/CS/Linter/Linter.php
+++ b/Symfony/CS/Linter/Linter.php
@@ -46,7 +46,7 @@ final class Linter implements LinterInterface
     {
         if (null === $executable) {
             $executableFinder = new PhpExecutableFinder();
-            $executable = $executableFinder->find();
+            $executable = $executableFinder->find(false);
 
             if (false === $executable) {
                 throw new UnavailableLinterException();
@@ -144,6 +144,14 @@ final class Linter implements LinterInterface
      */
     private function prepareCommand($path)
     {
-        return sprintf('%s -l %s', $this->executable, ProcessUtils::escapeArgument($path));
+        $executable = ProcessUtils::escapeArgument($this->executable);
+
+        if (defined('HHVM_VESION')) {
+            $executable .= ' --php';
+        }
+
+        $path = ProcessUtils::escapeArgument($path);
+
+        return sprintf('%s -l %s', $executable, $path);
     }
 }
--- a/Symfony/CS/Tests/Linter/LinterTest.php
+++ b/Symfony/CS/Tests/Linter/LinterTest.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\CS\Tests;
 
+use Symfony\Component\Process\ProcessUtils;
 use Symfony\CS\Linter\Linter;
 
 /**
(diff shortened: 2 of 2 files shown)


Symfony/CS/Linter/Linter.php
```diff
@@ -106,7 +106,8 @@ private function createProcessForFile($path)
             return $this->createProcessForSource(file_get_contents($path));
         }
 
-        $process = new Process(sprintf('%s -l %s', $this->executable, ProcessUtils::escapeArgument($path)));
+        $executable = ProcessUtils::escapeArgument($this->executable).(defined('HHVM_VESION') ? '--php' : '');
```
keradus: use `PhpExecutableFinder::findArguments`

keradus:
Could you add a tests for that fix please?

GrahamCampbell:
> Could you add a tests for that fix please?

Hmm, well, we already have integration tests don't we?

GrahamCampbell:
In that, we run php-cs-fixer on itself.

keradus:
> Hmm, well, we already have integration tests don't we?

It doesn't lint anything, so not a case here.


> In that, we run php-cs-fixer on itself.

And that run doesn't cover #1364 bug.

GrahamCampbell:
> It doesn't lint anything, so not a case here.

Yes it does. The other PR proved that it did because running php-cs-fixer was broken on hhvm.

Symfony/CS/Linter/Linter.php
GrahamCampbell: I don't really trust symfony's implementation here given that the find method doesn't work correctly if the path needs escaping and you're running anything that needs params.

keradus:
> Yes it does. The other PR proved that it did because running php-cs-fixer was broken on hhvm.

No it does not. HHVM was broken thanks to that param. HHVM didn't fail on tests but on runing tool in dry-run mode after tests were made.
Previous PR does not broke on keeping php in directory with space, as reported here: #1364

So it is great to have HHVM not failing, but still, tests with php-executable in directory with space are not covered by tests.

Symfony/CS/Linter/Linter.php
keradus: Executable as it is may not be escaped. When you find executable you may not want it escaped, it's up to your use case.

GrahamCampbell: > Executable as it is may not be escaped. When you find executable you may not want it escaped, it's up to your use case.

I don't really agree. If we're including the "arguments" appended to it, then it should definitely be quoted. I think that's a bug, but changing it would be a massive break. I'm not sure what symfony will actually do here.

GrahamCampbell: I doubt it would be broken on 2.x, but this stuff might change in 3.x, so that's why I don't want to depend on something I already feel is broken and I want changing.

keradus:
About #1364. Tests run well on repo, but running the code fails due to error in calling php from dir with space.

About this PR. It fix the bug it, but the facts that all tests passed won't prove it.
Why? Because of the same reason why tests won't fail on the bug you have just introduced - no tests for Linter class. The bug itself may even not be intuitive - HHVM on Travis doesn't use `--php`. And introduced bug? `HHVM_VESION` (not `HHVM_VERSION`) and missing space before `--php`. 
And both of that bugs may be ommited simply using `PhpExecutableFinder::findArguments`.
Proven here: https://travis-ci.org/keradus/PHP-CS-Fixer/jobs/81300374#L202

Ref: https://github.com/GrahamForks/PHP-CS-Fixer/pull/3

GrahamCampbell:
I'll look into this tomorrow.

keradus:
Please, could you rebase?

GrahamCampbell:
How does this look now?

keradus:
Better for sure.
But still it should be tested, so see https://github.com/GrahamForks/PHP-CS-Fixer/pull/3.
Eg here - after rebase you add missing space (one of bugs proven in my PR), but still not fix name of constant. Also, we still need a test to prove fix for #1364 itself.
All described issues are touched in my PR, and yes, it may not be perfect, but show the problems.

keradus:
could you fix the code to have tests passed on hhvm please ?

GrahamCampbell:
I'm massively busy this weekend and next week. I'll look into it after that.

keradus:
Very well: #1414 
it was just a one typo with const name in your code

keradus:
Closing due to merge of #1414

PR #1405, opened 2015-09-21
Linter - fix ignoring input parameter for constructor

keradus:
Linter - fix ignoring input parameter for constructor.
Also, add some tests.

### DIFF
--- a/Symfony/CS/Linter/Linter.php
+++ b/Symfony/CS/Linter/Linter.php
@@ -40,17 +40,17 @@ final class Linter implements LinterInterface
     private $executable;
 
     /**
-     * @param string|null $phpExecutable PHP executable, null for autodetection
+     * @param string|null $executable PHP executable, null for autodetection
      */
-    public function __construct($phpExecutable = null)
+    public function __construct($executable = null)
     {
-        if (null === $phpExecutable) {
+        if (null === $executable) {
             $executableFinder = new PhpExecutableFinder();
             $executable = $executableFinder->find();
-        }
 
-        if (empty($executable)) {
-            throw new UnavailableLinterException();
+            if (empty($executable)) {
+                throw new UnavailableLinterException();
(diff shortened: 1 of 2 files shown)


Symfony/CS/Tests/Linter/LinterTest.php
```diff
@@ -0,0 +1,42 @@
...
+    public function testLintSourceWithGoodCode()
+    {
+        $linter = new Linter();
+        $linter->lintSource('<?php echo 123;'); // no exception should be raised
+    }
+
+    /**
+     * @covers Symfony\CS\Linter\Linter::lintSource
+     *
+     * @expectedException Symfony\CS\Linter\LintingException
```
GrahamCampbell: We should definitely test the exception message too.

keradus: the message depends on PHP version, and here is a problem with Windows vs Unix

GrahamCampbell: Yes, indeed, but we can use phpunits regex message matching

keradus: done, let see how it will behave

Symfony/CS/Linter/Linter.php
```diff
@@ -40,17 +40,17 @@
...
     {
-        if (null === $phpExecutable) {
+        if (null === $executable) {
             $executableFinder = new PhpExecutableFinder();
             $executable = $executableFinder->find();
-        }
 
-        if (empty($executable)) {
-            throw new UnavailableLinterException();
+            if (empty($executable)) {
```
GrahamCampbell: why are we calling a function instead of doing a comparison?

keradus: dunno, the logic was already here

Issue #1407, opened 2015-09-22
PhpdocIndentFixer creates invalid comment syntax leading to comments being printed out


phofmann-trust:
PhpdocIndentFixer creates invalid comment syntax leading to comments being printed out.

The following code should not be modified:
```
<?php /** @var Foo $foo */ ?>
```
But it gets changed to:
```
<?php/** @var Foo $foo */ ?>
```

Because there is no white space after <?php the whole line will be printed out by PHP. So after running php-cs-fixer it is possible that there are PHP comments in between the HTML on a website / in the output of a program.

Test case for Symfony/CS/Tests/Fixer/Symfony/PhpdocIndentFixerTest.php PhpdocIndentFixerTest::provideDocblocks
```
        $cases[] = array('<?php /** @var Foo $foo */ ?>');
```

phofmann-trust:
This also seems to happen for the following code:
```
<?php /** foo */ ?>
```

keradus:
Hi!
Thank you for reporting!

Which version are you using?
(You may find it out by running `php-cs-fixer --version`)

phofmann-trust:
Oh, not the latest one - it's 1.9.3.

keradus:
Could you be so nice and test it out on newest 1.x@stable and 2.x@dev ?

GrahamCampbell:
I'd assume it's broken on both.

phofmann-trust:
I could quickly test 1.10. There it's broken too.

SpacePossum:
@phofmann-trust this should now be fixed on the 1.10 branch, could you give it another try?
(not the phar version yet ;) )

keradus:
Closing due to merge of #1407

phofmann-trust:
I tried out 1.10, it works now. Thanks! :+1:

PR #1410, opened 2015-09-24
PhpdocIndentFixer - Fix for open tag

SpacePossum:
This PR addresses this issue:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1407

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocIndentFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocIndentFixer.php
@@ -40,7 +40,8 @@ class PhpdocIndentFixer extends AbstractFixer
 
             // ignore inline docblocks
             if (
-                ($prevToken->isWhitespace(array('whitespaces' => " \t")) && !$tokens[$index - 2]->isGivenKind(T_OPEN_TAG))
+                $prevToken->isGivenKind(T_OPEN_TAG)
+                || ($prevToken->isWhitespace(array('whitespaces' => " \t")) && !$tokens[$index - 2]->isGivenKind(T_OPEN_TAG))
                 || $prevToken->equalsAny(array(';', '{'))
             ) {
                 continue;
--- a/Symfony/CS/Tests/Fixer/Symfony/PhpdocIndentFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/PhpdocIndentFixerTest.php
@@ -30,6 +30,8 @@ class PhpdocIndentFixerTest extends AbstractFixerTestBase
     {
         $cases = array();
 
+        $cases[] = array('<?php /** @var Foo $foo */ ?>');
+
         $cases[] = array('<?php /** foo */');
 
         $cases[] = array(


Symfony/CS/Fixer/Symfony/PhpdocIndentFixer.php
```diff
@@ -40,7 +40,8 @@ public function fix(\SplFileInfo $file, $content)
 
             // ignore inline docblocks
             if (
-                ($prevToken->isWhitespace(array('whitespaces' => " \t")) && !$tokens[$index - 2]->isGivenKind(T_OPEN_TAG))
+                $prevToken->isGivenKind(T_OPEN_TAG)
+                || ($prevToken->isWhitespace(array('whitespaces' => " \t")) && !$tokens[$index - 2]->isGivenKind(T_OPEN_TAG))
```
GrahamCampbell: surely we should allow other kinds of whitespace too?

GrahamCampbell: oh, wait, ignore me, lol

keradus:
looking nice ;)

GrahamCampbell:
This if statement is somewhat hard to read. Maybe it's worth moving it out to a protected method with multiple logical blocks in it rather than one massive ass statement.

SpacePossum:
@GrahamCampbell I don't see much value adding the overhead with a method. What do you think @keradus ?

keradus:
Each condition is written in separate line, if is clear for me.

But I'm also a perl developer, where you used to write code like
```perl
@dst = sort { $$a{nested_key} <=> $$b{nested_key} } map { $$_{flat_key} } @src if ($$one{need}[$the]{data});
````
;)

GrahamCampbell:
Well, I managed to missread it the first time, so I didn't find it very clear.

GrahamCampbell:
It made the diff hard to read in particular, more than the code before and after.

SpacePossum:
I see no problem with the Perl code ^-^

The diff. might be a bit hard, but as long as the end code is OK I think I rather keep this PR as is.

GrahamCampbell:
But, the point in a lot of symfony's code style, IS to make diffs easier to read, and I think that's very important.

SpacePossum:
I agree it is important, however I think most developers spend more time reading the code then the diffs of it.
It is not an excuse to mess up the history, but in this case I think the readability of the code outweighs that of the diff.

GrahamCampbell:
I think it would make BOTH more readable.

keradus:
I think that discussion has little point now...

stof:
> But, the point in a lot of symfony's code style, IS to make diffs easier to read, and I think that's very important.

Trailing commas in multiline arrays and not aligning assignment/array arrows are not about making the diff easier to read. They are about avoiding useless changes in diffs, which would also lead to more frequent conflicts (even more important in Symfony as we merge older branch into newer branches all the time, and we try to avoid conflicts between maintained branches whenever possible. this is also why we don't reorder existing use statements)

keradus:
Thank you @SpacePossum.

keradus:
Merged into 1.10, not 1.11 as requested.

SpacePossum:
thanks @keradus and the reviewers :)
sorry for targeting the wrong branch : |

PR #1414, opened 2015-09-25
Linter - fix escaping the php binary

keradus:
Fixes ##1364
Replaces #1366 and #1403


keradus:
Thank you @GrahamCampbell.

GrahamCampbell:
Thanks for finishing this off for me. ;)

PR #1416, opened 2015-09-26
Mark fixers as risky

### LINKED ISSUE #942: Proposal: mark fixer as "risky" (change meaning of code)
What do you think about marking fixers as risky?

Currently: psr0, strict, strict_param.
In future fixer that trim all trailing spaces (even inside string content), as PSR2 requires.

Then that fixer will run only if risky fixers are allowed by config/cli.

Then, in future, one may run all PSR2 fixers safely without changing meaning of code or run all PSR2 fixers with that risky flag and then apply all PSR2 rules, even if they change code meaning.

keradus:
Closes #942

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -160,6 +160,9 @@ When using combinations of exact and blacklist fixers, applying exact fixers alo
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
 
+The ``--allow-risky`` option allows you to set whether riskys fixer may run. Default value is taken from config file.
+Risky fixer is a fixer, which could change code behaviour. By default no risky fixers are run.
+
 The command can also read from standard input, in which case it won't
 automatically fix anything:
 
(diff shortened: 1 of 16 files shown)


stof:
Looks good to me

README.rst
```diff
@@ -160,6 +160,9 @@ When using combinations of exact and blacklist fixers, applying exact fixers alo
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
 
+The ``--allow-risky`` option allow you to set whether riskys fixer may run. Default value is taken from config file.
```
SpacePossum: `allows`

Symfony/CS/Config/Config.php
```diff
@@ -195,6 +196,24 @@ public function getPhpExecutable()
     /**
      * {@inheritdoc}
      */
+    public function getRiskyAllowed()
```
SpacePossum: maybe `isRiskyAllowed()`, following http://symfony.com/doc/current/contributing/code/conventions.html?

SpacePossum:
looking good :)

Symfony/CS/Config/Config.php
keradus: can't see `isFoo`, only `isEmptyFoo`

README.rst
keradus: yeah ;)

Symfony/CS/Config/Config.php
SpacePossum: fair enough :+1:  :)

Issue #1377, opened 2015-09-04
Double arrow align fail on sub array.


Soullivaneuh:
My original code:

```php
return new JsonResponse([
    'result' => true === $feedbackSubmitted ? 'OK' : 'KO',
    'html'   => $this->renderView('ticket/_feedback_modal.html.twig', [
        'ticket'            => $ticket,
        'feedbackForm'      => $feedbackForm->createView(),
        'feedbackSubmitted' => $feedbackSubmitted,
    ]),
]);
```

After CS:

```php
return new JsonResponse([
    'result'                => true === $feedbackSubmitted ? 'OK' : 'KO',
    'html'                  => $this->renderView('ticket/_feedback_modal.html.twig', [
        'ticket'            => $ticket,
        'feedbackForm'      => $feedbackForm->createView(),
        'feedbackSubmitted' => $feedbackSubmitted,
    ]),
]);
```

`result` and `html` should not be aligned with `renderView` method array parameter.

Using `v1.10`.

GrahamCampbell:
I've seen this issue too, and can confirm it.

SpacePossum:
Hi all,
I've tried to reproduce this on 1.10 and master using:
cd9fb23813a065dade8773d97df8766cd4b58ac9

but all looks OK. When running with all fixers on PSR2 level on 1.10 it works as well.
For finding/fixing this is it might help proving more info about the configuration used.

keradus:
Ref

keradus:
Closing thanks to #1422

GrahamCampbell:
:+1:

PR #1417, opened 2015-09-26
Fix for 1377

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1377

### DIFF
--- a/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
+++ b/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
@@ -12,6 +12,7 @@
 namespace Symfony\CS\Fixer\Contrib;
 
 use Symfony\CS\AbstractAlignFixer;
+use Symfony\CS\Tokenizer\Token;
 use Symfony\CS\Tokenizer\Tokens;
 
 /**
@@ -48,7 +49,7 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
         $this->deepestLevel = 0;
         $tokens = Tokens::fromCode($content);
 
-        $this->injectAlignmentPlaceholders($tokens);
+        $this->injectAlignmentPlaceholders($tokens, 0, count($tokens));
 
         return $this->replacePlaceholder($tokens, $this->deepestLevel);
     }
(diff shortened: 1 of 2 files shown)


Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
```diff
@@ -30,6 +30,25 @@ public function provideFixCases()
     {
         return array(
             array(
+                "<?php
+        return new JsonResponse([
```
keradus: it would be great to have third case with only long array syntax

Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
```diff
@@ -30,6 +30,25 @@ public function provideFixCases()
...
+            'result' => 'OK',
+            'html'   => renderView('views/my_view.html.twig', array(
+                'foo'    => 'bar',
+                'foofoo' => 42,
+            )),
+        ]);", ),
+            array(
+                "<?php
```
keradus:
```php
<?php
+    return new JsonResponse([
+        'result' => 'OK',
+        'html'   => renderView('views/my_view.html.twig', [
+            'foo'    => 'bar',
+            'foofoo' => 42,
+        ]),

+    ]);
```

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
```diff
@@ -72,15 +73,8 @@ public function getDescription()
      */
     private function injectAlignmentPlaceholders(Tokens $tokens, $startAt = null, $endAt = null)
     {
-        if (empty($startAt)) {
-            $startAt = 0;
-        }
-
-        if (empty($endAt)) {
```
keradus: why is that change?

SpacePossum: This way there is no need to test the input for every iteration, which makes the execution a bit faster.

Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
SpacePossum: good idea :)
added 3d3faebadfa5c29d9e51102033967899f88552b4

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
keradus: And make the interface uglier. But I see your point. And now... I can still use default null values :D

Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
SpacePossum: :+1: 
please see https://github.com/SpacePossum/PHP-CS-Fixer/commit/cd020438f83c5f60a7d6137a28aab5d31d060874

PR #1419, opened 2015-09-26
AlignDoubleArrowFixer, AlignEqualsFixer - fix priorities

keradus:
Fixes #1351

PR #1420, opened 2015-09-26
AlignDoubleArrowFixer, AlignEqualsFixer - add integration tests

keradus:
Add tests for #1419

### DIFF
--- a/Symfony/CS/Tests/AbstractIntegrationTest.php
+++ b/Symfony/CS/Tests/AbstractIntegrationTest.php
@@ -64,7 +64,7 @@ abstract class AbstractIntegrationTest extends \PHPUnit_Framework_TestCase
 
     public static function tearDownAfterClass()
     {
-        unlink(static::getTempFile());
+        @unlink(static::getTempFile());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -64,7 +64,7 @@ public static function setUpBeforeClass()
 
     public static function tearDownAfterClass()
     {
-        unlink(static::getTempFile());
+        @unlink(static::getTempFile());
```
keradus: If no tests were run then unlink will trigger non-existing file error.

cc @SpacePossum

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -236,7 +236,7 @@ protected function getFixersFromConfig($fileName, $config)
         }
 
         $fixers = array();
-        for ($i = count(self::$builtInFixers) - 1; $i >= 0; --$i) {
+        for ($i = 0, $limit = count(self::$builtInFixers); $i < $limit; ++$i) {
```
keradus: We must not change the order of fixers. `Fixer::getFixers()` will sort fixers, so we must use sorted version.

cc @SpacePossum

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: but how is that possible, if the .. oh I see, :+1: thanks for fixing :)

Symfony/CS/Tests/AbstractIntegrationTest.php
SpacePossum: good catch :+1:

SpacePossum:
:+1:

PR #1425, opened 2015-09-27
PhpdocInlineTagFixerTest - fix case when met inalid PHPDoc

### LINKED ISSUE #1347: Unbalanced parentheses in DocBlock cause error
I found a nice little edge case. If (for whatever reason) a DocBlock has unbalanced parentheses, PHP-CS-Fixer will error out and not be able to process that file. Example: 

[https://gist.github.com/samuelwilliams/ef214991b5d74a203d64](https://gist.github.com/samuelwilliams/ef214991b5d74a203d64)

```php
<?php
/**
 * {@link http://www.ietf.org/rfc/rfc1035.text)
 */
$someVar = 'hello';
```

`php-cs-fixer fix ./TestFile.php --dry-run --diff`

```txt
Fixed all files in 0.158 seconds, 6.500 MB memory used

Files that were not fixed due to internal error:
   1) ./TestFile.php
```

Can this behaviour be reasonably expected? There are legitimate reasons why one would want unbalanced parentheses in one's DocBlock, for instance: `The parameter should be on [0, 10)`.

keradus:
Fixes #1347

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocInlineTagFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocInlineTagFixer.php
@@ -38,7 +38,7 @@ final class PhpdocInlineTagFixer extends AbstractFixer
             // Make sure the tags are written in lower case, remove white space between end
             // of text and closing bracket and between the tag and inline comment.
             $content = preg_replace_callback(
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:+1:

SpacePossum:
:+1:
thanks :)

PR #1429, opened 2015-09-28
Dedicate public method for fixers by name

Soullivaneuh:
Permit to get fixers by name and do some checks manually.

Concrete case: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/pull/21

### DIFF
--- a/Symfony/CS/FixerFactory.php
+++ b/Symfony/CS/FixerFactory.php
@@ -76,6 +76,22 @@ final class FixerFactory
     }
 
     /**
+     * Get associative array of fixers with names as key.
+     *
+     * @return FixerInterface[]
+     */
+    public function getFixersByName()
+    {
+        $fixersByName = array();
+
+        foreach ($this->fixers as $fixer) {
+            $fixersByName[$fixer->getName()] = $fixer;
+        }
+
+        return $fixersByName;
+    }
+
+    /**
      * Register all built in fixers.
      *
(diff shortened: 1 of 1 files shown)


keradus:
Don't like the idea.
`getFixers` return fixers in good order - and getting fixers in other order create possibility to use them in wrong order, which will lead to errors.

I would suggest other method like `hasRule(string)`.

Also, when adding new public method remember to add tests for it.

Soullivaneuh:
I would suggest other method like `hasRule(string)`.

Why not. But in this case we should have `$fixersAsName` as a property and `getFixersAsName` private because this will be used twice. Do you agree?

keradus:
We could have `$fixersAsName` or not, it's not part of interface but implementation. If it would made implementation better why not.

Soullivaneuh:
Closed in favor of #1432.

Symfony/CS/FixerFactory.php
```diff
@@ -139,7 +155,7 @@ public function registerFixer(FixerInterface $fixer)
      */
     public function useRuleSet(RuleSetInterface $ruleSet)
     {
-        $fixersByName = array();
+        $fixersByName = $this->getFixersByName();
 
         foreach ($this->fixers as $fixer) {
```
GrahamCampbell: why is this code running twice?

PR #1430, opened 2015-09-28
Add rules

### LINKED ISSUE #1428: Config::addRules and Config::addRule
Could be useful to have fixers addition methods on Config class.

In some case, we could import config from a bridge like this: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge

And add some fixers like header_comment.

Actually, with 2.x, we have to do this:

```php
use SLLH\StyleCIBridge\ConfigBridge;

$header = 'test';

$config = ConfigBridge::create();

return $config
    ->setUsingCache(true)
    ->setRiskyAllowed(true)
    ->setRules(array_merge($config->getRules(), ['header_comment' => ['header' => $header]]))
;
```

Would be great if we could do this:

```php
return ConfigBridge::create()
    ->setUsingCache(true)
    ->setRiskyAllowed(true)
    ->addRules(['header_comment' => ['header' => $header]])
;
```

Or, even better, this:

```php
return ConfigBridge::create()
    ->setUsingCache(true)
    ->setRiskyAllowed(true)
    ->addRule('header_comment', ['header' => $header])
;
```

I can provide a PR if interested.

Soullivaneuh:
Closes #1428.


Soullivaneuh:
Sigh, wrong branch.

PR #1431, opened 2015-09-28
Implement Config::addRule and Config::addRules methods

### LINKED ISSUE #1428: Config::addRules and Config::addRule

Soullivaneuh:
Closes #1428.

### DIFF
--- a/Symfony/CS/Tests/Config/ConfigTest.php
+++ b/Symfony/CS/Tests/Config/ConfigTest.php
@@ -79,4 +79,42 @@ final class ConfigTest extends \PHPUnit_Framework_TestCase
 
         $this->assertSame($config, $config->setCacheFile('some-directory/some.file'));
     }
+
+    public function testAddRules()
+    {
+        $config = new Config();
+        $config->setRules(array(
+            'foo' => true,
+            'bar' => true,
+        ));
+
+        $config->addRules(array(
+            'foo' => false,
+            'something' => array('with' => 'array'),
+        ));
+
+        $this->assertSame(array(
+            'foo' => false,
+            'bar' => true,
+            'something' => array('with' => 'array'),
+        ), $config->getRules());
+    }
+
+    public function testAddRule()
+    {
+        $config = new Config();
+        $config->setRules(array(
+            'foo' => true,
+            'bar' => true,
+        ));
+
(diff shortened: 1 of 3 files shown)


Soullivaneuh:
AppVeyor needs to be re-launched.

Symfony/CS/Tests/Config/ConfigTest.php
```diff
@@ -79,4 +79,42 @@ public function testThatMutatorHasFluentInterface()
...
+
+    public function testAddRule()
+    {
+        $config = new Config();
+        $config->setRules(array(
+            'foo' => true,
+            'bar' => true,
+        ));
+
+        $config->addRule('foo', false);
```
keradus: And this test prove that this method is not to add rule but mutate it as well.
So this is de facto like `removeRule`.

keradus:
:-1:

Symfony/CS/Tests/Config/ConfigTest.php
Soullivaneuh: I can throw an exception instead if you want.

If not, I let your the liberty to close this PR and the related issue.

GrahamCampbell:
:-1: too, sorry.

keradus:
Closing due to discussion of #1428.

PR #1432, opened 2015-09-28
Fixer check on factory

Soullivaneuh:


Symfony/CS/FixerFactory.php
```diff
@@ -76,6 +83,20 @@ public function getFixers()
     }
 
     /**
+     * Check if fixer exists.
+     *
+     * @param string $name The rule name of the fixer
+     *
+     * @return bool
+     */
+    public function hasFixer($name)
```
keradus: `hasFixer` but input is name of... rule?

Soullivaneuh: This is quite confusing because it's still an array of Fixers.

What do you prefer? Description update? Rename the method to `hasRule`?

keradus: Look on interface - input is name of rule, output is bool.
What is done under the hood (like checking fixers) isn't crucial for interface

Symfony/CS/FixerFactory.php
```diff
@@ -139,7 +160,7 @@ public function registerFixer(FixerInterface $fixer)
      */
     public function useRuleSet(RuleSetInterface $ruleSet)
     {
-        $fixersByName = array();
+        $fixersByName = $this->getFixersByName();
 
         foreach ($this->fixers as $fixer) {
             $fixersByName[$fixer->getName()] = $fixer;
```
keradus: ?

Symfony/CS/FixerFactory.php
Soullivaneuh: Yes. So, `hasRule`?

Symfony/CS/FixerFactory.php
Soullivaneuh: Hmm... I think I forgot something. :-)

Symfony/CS/FixerFactory.php
keradus: why not

Soullivaneuh:
@keradus review fixed.

keradus:
Implementation is buggy.

```php
$factory->registerBuiltInFixers();
$factory->hasRule('return'); // true
$factory->useRuleSet(new RuleSet(['@PSR1' => true]));
$factory->hasRule('return'); // true, MUST be false
```

Soullivaneuh:
@keradus Fixed. Also added related test.

stof:
Tests are failing though, because of the usage of short array syntax (the tool supports PHP 5.3)

keradus:
yeah, my snippet was more a pseudocode than testcase ready to use ;)

Soullivaneuh:
@keradus @stof fixed.

Symfony/CS/FixerFactory.php
```diff
@@ -36,6 +36,13 @@
     private $fixers = array();
 
     /**
+     * Fixers by name.
+     *
+     * @var FixerInterface[]|null
```
keradus: what is the key is not documented

Symfony/CS/Tests/FixerFactoryTest.php
```diff
@@ -279,4 +279,24 @@ function ($name) {
 
         return $cases;
     }
+
+    public function testHasRule()
```
keradus: Not a must have, but it would be great to have it done on mocks instead of real fixers.
Now the fixer will fail if we rename the fixer name and forgot to change the test here. (second test too)
Any chances?

Symfony/CS/FixerFactory.php
Soullivaneuh: What do you mean?

Soullivaneuh: `FixerInterface[string]`? Is that working with IDEs?

Symfony/CS/Tests/FixerFactoryTest.php
Soullivaneuh: Will give a try.

Soullivaneuh: Don't understand, I did this based on other tests:

```php
public function testHasRule()
{
    $factory = new FixerFactory();
    $factory->registerBuiltInFixers();

    $f1 = $this->getMock('Symfony\CS\FixerInterface')
        ->expects($this->any())->method('getName')->willReturn('f1');
    $f2 = $this->getMock('Symfony\CS\FixerInterface')
        ->expects($this->any())->method('getName')->willReturn('f2');
    $f3 = $this->getMock('Symfony\CS\FixerInterface')
        ->expects($this->any())->method('getName')->willReturn('f3');
    $factory->registerFixer($f1);
    $factory->registerCustomFixers(array($f2, $f3));

    $this->assertTrue($factory->hasRule('f1'));
    $this->assertTrue($factory->hasRule('f2'));
    $this->assertTrue($factory->hasRule('f3'));
    $this->assertFalse($factory->hasRule('dummy'));
}
```

But I got:

```
1) Symfony\CS\Tests\FixerFactoryTest::testHasRule
Argument 1 passed to Symfony\CS\FixerFactory::registerFixer() must implement interface Symfony\CS\FixerInterface, instance of PHPUnit_Framework_MockObject_Builder_InvocationMocker given, called in /home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Tests/FixerFactoryTest.php on line 294 and defined

/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/FixerFactory.php:133
/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Tests/FixerFactoryTest.php:294
```

And it is a `FixerInterface` mock... any idea?

Symfony/CS/FixerFactory.php
keradus: TBH dunno how PSR handle it.
In the worst case - add to long descr.

Soullivaneuh: > In the worst case - add to long descr.

Done.

Symfony/CS/Tests/FixerFactoryTest.php
keradus: `willReturn` will return `PHPUnit_Framework_MockObject_Builder_InvocationMocker`, and this is what you got under `$f1`
What you  need undef `$f1` is result of `getMock`, which would be `PHPUnit_Framework_MockObject_MockObject`

Soullivaneuh: Ok Indeed.

That is finished, tests are now using mocks. :+1:

Symfony/CS/FixerFactory.php
```diff
@@ -175,4 +194,22 @@ private function sortFixers()
 
         return $this;
     }
+
+    /**
+     * Get associative array of fixers with names as key.
+     *
+     * @return FixerInterface[]
```
keradus: * @var FixerInterface[]|null Associative array of fixers with names as keys.

Symfony/CS/FixerFactory.php
```diff
@@ -36,6 +36,13 @@
     private $fixers = array();
 
     /**
+     * Fixers by name.
```
keradus: Would be great to say that it is lazy loaded when getter is used.
To prevent using the variable without getter directly in `FixerFactory`

Symfony/CS/Tests/FixerFactoryTest.php
```diff
@@ -279,4 +279,39 @@ function ($name) {
...
+        $factory = new FixerFactory();
+        $factory->registerBuiltInFixers();
+
+        $f1 = $this->getMock('Symfony\CS\FixerInterface');
+        $f1->expects($this->any())->method('getName')->willReturn('f1');
+        $factory->registerFixer($f1);
+
+        $this->assertTrue($factory->hasRule('f1'), 'Should have f1 fixer');
+
+        $factory->useRuleSet(new RuleSet(array('@PSR1' => true)));
```
keradus: If we remove `@PSR1` set then this will throw an exception.
Use mocks to create f1 and f2, then create ruleset without one of them.

Symfony/CS/FixerFactory.php
Soullivaneuh: Done.

Soullivaneuh:
Review corrected.

keradus:
Looks great @Soullivaneuh !
I hope I wasn't been picky

Soullivaneuh:
> I hope I wasn't been picky

I'm also quite picky on PR where I am repository contributor so I understand. ;-)

Soullivaneuh:
@keradus When this PR should be merged?

keradus:
Relax, I simply wasn't online ;)

keradus:
Thank you @Soullivaneuh.

Soullivaneuh:
> Relax, I simply wasn't online ;)

I was relax, simple question. ;-)

Thank you.

PR #1433, opened 2015-09-29
phpunit.xml.dist - disable gc

keradus:


SpacePossum:
out of pure curiosity, what is this benefit of stopping the default garbage collection?

keradus:
we don't need circular reference collector enabled on tests, therefore a bit speedup

stof:
you need to rebase your branch because of the branch protection you enabled (because of the status requirement actually)

keradus:
Sorry @stof, but I don't understand.
Branch protection required Travis and AppVeyor to pass, and they passed, so it doesn't block the merge.
Also, I just mark that 2 services as required, but that doesn't include administrators of repository.
Why the rebase is needed?

GrahamCampbell:
Because after rebasing, the check might fail.

stof:
@keradus the protection enforces the feature branch to be uptodate: 

> When using required status checks, your branch must be up-to-date to ensure your tests will pass after merging. The new “Update branch” button lets you merge the latest changes from the base branch of the Pull Request with one click.

I forgot that there is a way to disable the check for repo admins though.

sebastianbergmann:
Out of curiosity, do you have numbers that show how much the test suite is executed with the garbage collector disabled as compared to with the garbage collector enabled? Thanks.

PR #1434, opened 2015-09-29
Add Test\AccessibleObject class

keradus:
In some very edge cases we need to get non-public property/method.
So stop duplicate the code.

### DIFF
--- a/Symfony/CS/Test/AccessibleObject.php
+++ b/Symfony/CS/Test/AccessibleObject.php
@@ -24,6 +24,9 @@ final class AccessibleObject
         return new self($object);
     }
 
+    /**
+     * @param object $object
+     */
     public function __construct($object)
     {
         $this->object = $object;
--- a/Symfony/CS/Test/AccessibleObject.php
+++ b/Symfony/CS/Test/AccessibleObject.php
@@ -1,0 +1,77 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Test;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class AccessibleObject
+{
+    private $object;
+    private $reflection;
+
+    public static function create($object)
+    {
+        return new self($object);
+    }
+
+    public function __construct($object)
+    {
+        $this->object = $object;
+        $this->reflection = new \ReflectionClass($object);
+    }
+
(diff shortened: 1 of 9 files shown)


Symfony/CS/Test/AccessibleObject.php
```diff
@@ -0,0 +1,77 @@
...
+{
+    private $object;
+    private $reflection;
+
+    public static function create($object)
+    {
+        return new self($object);
+    }
+
+    public function __construct($object)
```
SpacePossum: Can't the constructor be `private`? I see no use of this constructor as is..dunno..

keradus: Sorry, don't understand. Could you elaborate please?

Symfony/CS/Test/AccessibleObject.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Test;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
```
SpacePossum: Don't you want this class to be marked as `internal` as well? I don't think you want to support uses of this class outside of the project?

Symfony/CS/Test/AccessibleObject.php
SpacePossum: I was to fast commenting on this, sorry, all good to me :+1: 
Is there a way to type hint `$object` in a PHPDoc as being something of an object (i.e. not scalar)?

Symfony/CS/Test/AccessibleObject.php
keradus: Personally I have no need for that, it's already final and functionality is so extra small that interface won't change.

SpacePossum: very well :)

SpacePossum:
:+1:

Symfony/CS/Test/AccessibleObject.php
keradus: done

Symfony/CS/Test/AccessibleObject.php
```diff
@@ -0,0 +1,80 @@
...
+        $method = $this->reflection->getMethod($name);
+        $method->setAccessible(true);
+
+        return $method->invokeArgs($this->object, $arguments);
+    }
+
+    public function __isset($name)
+    {
+        try {
+            return isset($this->$name);
```
stof: this looks broken to me. this will call ``__isset`` again

keradus: Great catch!
I added utests and fix that bug

PR #1440, opened 2015-10-10
Made AbstractFixerTestCase and AbstractIntegrationTestCase public

keradus:


keradus:
Fabbot failed as usual...

keradus:
any thoughts ?

keradus:
rebased

Soullivaneuh:
A note about BC break should be added on the upgrade gad IMO.

Althoug the class was marked as internal, it's still usable and probably use on 1.x.

keradus:
I don't get it. Yes, the classes were internal, now they are not - how could it be a BC break?
Changing internal interface isn't a BC break.

Also, we have stable 1.x, and this PR is for 2.x, so even if it was a BC breaker, we are already targetting branch with major version change.

Soullivaneuh:
> Yes, the classes were internal, now they are not - how could it be a BC break?

The class is renamed also. And even if it's marked as internal, it could be technically be used actually.

This should not, I agree, but it could AFAIK.

> Also, we have stable 1.x, and this PR is for 2.x, so even if it was a BC breaker, we are already targetting branch with major version change.

Yes, I was not talking about removing or handling BC break. Just put a note to `UPGRADE-2.0.md` file. :+1:

keradus:
> The class is renamed also. And even if it's marked as internal, it could be technically be used actually.
> This should not, I agree, but it could AFAIK.

If one decided to use internal code he should not be surprised when it changed. We do not promise to keep the compatibility of it, then it is not a BC break. No exceptions.


> Yes, I was not talking about removing or handling BC break. Just put a note to `UPGRADE-2.0.md` file. :+1:

Please, feel free to send a PR with that file, I will be more than happy to see it ;)

Symfony/CS/Test/AbstractIntegrationTestCase.php
```diff
@@ -137,7 +135,7 @@ protected static function getTempFile()
      * @param string           $input        Code to fix
      * @param string|null      $expected     Expected result or null if the input is expected not to change
      */
-    protected function doTestIntegration($testFileName, $testTitle, $fixers, $input, $expected = null)
+    protected function doTest($testFileName, $testTitle, $fixers, $input, $expected = null)
```
stof: does it need to be protected here ?

keradus:
Whole `Tests` dir could be moved, but it is a work for separate PR.

Symfony/CS/Test/AbstractIntegrationTestCase.php
keradus: maybe not needed, but one may like to override it, adding some extra checks

keradus:
Rebased, then I think it is ready to merge

Issue #1033, opened 2015-02-27
List of conflicting fixers


keradus:
See #910 for details.

~~fixer A turns off fixer B:~~
~~- BlanklineAfterOpenTagFixer -> NewlineAfterOpenTagFixer~~
~~- RemoveTrailingSpacesFixer -> TrailingSpacesFixer~~
~~- RemoveTrailingSpacesFixer -> WhitespacyLinesFixer~~

fixer A and B can't be run together
- PhpdocVarToType and PhpdocTypeToVar 
- LongArraySyntaxFixer and ShortArraySyntaxFixer
- ConcatWithSpacesFixer and ConcatWithoutSpacesFixer
- UnalignEqualsFixer and AlignEqualsFixer
- UnalignDoubleArrowFixer and AlignDoubleArrowFixer

GrahamCampbell:
You're missing a load of fixers from this list.

GrahamCampbell:
Both concatenation fixers conflict with each other, and the "type to var" and "var to type" fixers conflict too. Not sure what else conflicts.

keradus:
I didn't say it is a full list, just need some place to keep it.

gharlan:
* LongArraySyntaxFixer vs. ShortArraySyntaxFixer

gharlan:
* ConcatWithSpacesFixer vs. ConcatWithoutSpacesFixer

keradus:
thanks, updated

GrahamCampbell:
I did say this earlier. :)

> Both concatenation fixers conflict with each other

GrahamCampbell:
single_blank_line_before_namespace and no_blank_lines_before_namespace conflict too.

GrahamCampbell:
unary_operators_spaces and logical_not_operators_with_spaces

GrahamCampbell:
i think our unary_operators_spaces should be updated not to change `!`.

GrahamCampbell:
logical_not_operators_with_spaces also potentially conflicts with what ever fixer it is that ensures things are pushed up against the side of opening brackets in if stataments.

keradus:
> unary_operators_spaces and logical_not_operators_with_spaces

ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1306#issuecomment-122811837
(about similar `LogicalNotOperatorsWithSuccessorSpaceFixer`)

When `unary_operators_spaces` is run first there is no conflict.


[...]

PR #1443, opened 2015-10-11
combine PhpdocTypeToVarFixer and PhpdocVarToTypeFixer

keradus:
The easiest way to handle to conflicting fixers is not to have them ;)


keradus:
Could you take a look @GrahamCampbell ?

keradus:
fabbot is failing as always (cc @fabpot)

GrahamCampbell:
:-1:

Soullivaneuh:
Oh, so you merged it with an option for `var` or `type`?

@GrahamCampbell How this would be handled by StyleCI?

GrahamCampbell:
I'd rather we kept two fixers, BUT I think abstracting stuff to an abstract fixer class would be a good idea.

keradus:
Could you elaborate @GrahamCampbell ?
Both fixer touch same rule but in different manner, the only difference is configuration.
We have so many fixers that cannot be run together (not all listed, but see #1033), and creation abstraction for them is pointless - no gain to have 10 lines moved into other class. And merging fixers that do same job in different way protect us against conflicts.

keradus:
Any thoughts on that?
It could be a start on working on few more fixers that touch same rule but in different manner (eg how much whitespace should be before namespace)

GrahamCampbell:
:-1: for actually removing one of the fixers.

GrahamCampbell:
> Could you elaborate @GrahamCampbell ?

Like the other abstract fixers we have...

GrahamCampbell:
Like https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/AbstractPhpdocTypesFixer.php.

keradus:
Merging fixers together keep same code in one place and throw out the need of keeping list of conflicting fixers.
Creating abstraction for fixers but keeping two separated fixers doesn't give that.

GrahamCampbell:
I think it's worth it. Having fixers with different names is much simpler and clearer.

keradus:
Config vs inheritance - both are clear. If you only need an alias it may be done by `FixerFactory`.
But having multiple fixers that touch same rule isn't simpler if you think about handling conflicts. When you don't have multiple fixers for one rule there is no conflicts at all.

I hope your opinion isn't made due to that StyleCI doesn't support configuration yet.

Then we have a tie. Any other opinions?

GrahamCampbell:
> I hope your opinion isn't made due to that StyleCI doesn't support configuration yet.

No, I just think it's really inconsistent with the rest of php-cs-fixer and I really like the way php-cs-fixer already does things.

keradus:
It would be just a first step. All conflicting fixers would be merged. So no inconsistence.

And we already configure fixers, like telling which methods should be change. You would like a separate fixer for each method?

GrahamCampbell:
:-1: All conflicting fixers would be merged.

I prefer to have different fixers.

keradus:
Then, as I said...
Then we have a tie. Any other opinions?

sstok:
:+1: but the name should be changed as `phpdoc_type_to_var` is no longer correct then.

keradus:
maybe sth like `phpdoc_typehint_tag`, how do you think ?

GrahamCampbell:
If we change this, we have to change all the others.

keradus:
two fixers marged into new one as requested

GrahamCampbell:
is it not confusing that fixer tagged a symfony can fix to non symfony compliant standards?

keradus:
Fixer is not tagged as Symfony.
Rule (with it's fixer) is general, the group decides about used rules and how they are used.

`@Symfony` group contains `phpdoc_typehint_tag => [annotation=>type]`,
while some other group can contains `phpdoc_typehint_tag => [annotation=>var]`

GrahamCampbell:
The readme is missleading then.

keradus:
it shows that `phpdoc_typehint_tag` is used in `@Symfony` ruleset

keradus:
Closing as we decided to remove fixer that violates PSR5

PR #1448, opened 2015-10-12
Code grooming

keradus:


keradus:
fabbot fails as always :(
cc @fabpot

Symfony/CS/Fixer/Symfony/EmptyReturnFixer.php
```diff
@@ -26,7 +26,7 @@ public function fix(\SplFileInfo $file, $content)
     {
         $tokens = Tokens::fromCode($content);
 
-        foreach ($tokens->findGivenKind(T_RETURN) as $index => $token) {
+        foreach (array_keys($tokens->findGivenKind(T_RETURN)) as $index) {
```
GrahamCampbell: surely calling array_keys makes things slower because we're essentially iterating everything twice now

keradus: good point.

PR #1380, opened 2015-09-04
[2.x] PhpUnitConstructFixerTest

SpacePossum:
Fix the following (`Not`) cases as well.

```php
<?php $this->assertNotSame(false));
<?php $this->assertNotFalse(false);
```

```php
<?php $this->assertNotSame(true);
<?php $this->assertNotTrue(true);
```

```php
<?php $this->assertNotSame(false, $a);
<?php $this->assertNotFalse($a);
```

```php
<?php $this->assertNotSame(true, $a);
<?php $this->assertNotTrue($a);
```

PR #1449, opened 2015-10-13
PhpUnitConstructFixer - Fix more use cases

SpacePossum:
Fix the following (Not) cases as well.

```php
<?php $this->assertNotSame(false, $a);
<?php $this->assertNotFalse($a);
```

```php
<?php $this->assertNotSame(true, $a);
<?php $this->assertNotTrue($a);
```

```php
<?php $this->assertNotSame(false, $a);
<?php $this->assertNotFalse($a);
```

```php
<?php $this->assertNotSame(true, $a);
<?php $this->assertNotTrue($a);
```
Replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1380 (which was targeted at master and more of a mess).

### DIFF
--- a/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
@@ -62,36 +62,8 @@ class PhpUnitConstructFixerTest extends AbstractFixerTestBase
 
     public function provideTestFixCases()
     {
-        return array(
+        $cases = array(
             array('<?php $sth->assertSame(true, $foo);'),
-            array(
-                '<?php $this->assertTrue($a);',
-                '<?php $this->assertSame(true, $a);',
-            ),
-            array(
-                '<?php $this->assertTrue($a  , "true" . $bar);',
-                '<?php $this->assertSame(true  , $a  , "true" . $bar);',
-            ),
-            array(
-                '<?php $this->assertFalse(  $a, "false" . $bar);',
-                '<?php $this->assertSame(  false, $a, "false" . $bar);',
-            ),
-            array(
-                '<?php $this->assertNull(  $a  , "null" . $bar);',
-                '<?php $this->assertSame(  null, $a  , "null" . $bar);',
-            ),
-            array(
-                '<?php $this->assertNotNull(  $a  , "notNull" . $bar);',
-                '<?php $this->assertNotSame(  null, $a  , "notNull" . $bar);',
-            ),
-            array(
-                '<?php $this->assertFalse(  $a, "false" . $bar);',
-                '<?php $this->assertEquals(  false, $a, "false" . $bar);',
-            ),
-            array(
-                '<?php $this->assertNotNull(  $a  , "notNull" . $bar);',
-                '<?php $this->assertNotEquals(  null, $a  , "notNull" . $bar);',
-            ),
             array(
                 '<?php
     $this->assertTrue(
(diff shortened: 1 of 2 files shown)


SpacePossum:
fabbot tries to fix the test cases, this is not an issue IMO.

keradus:
Yes, fabbot is broken about that
cc @fabpot

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
```diff
@@ -23,76 +23,14 @@ class PhpUnitConstructFixerTest extends AbstractFixerTestBase
      */
     public function testFix($expected, $input = null)
     {
-        $fixer = $this->getFixer();
-
-        $fixer->configure(array(
```
keradus: why is this removed?

fabpot:
@keradus I would be more than happy to remove this check whenever php-cs-fixer does that for me :)

keradus:
@fabpot I think this fixer do what you want (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L87-L90)
(but only on 1.11@dev)

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
SpacePossum: merge gone wrong from changes of weeks ago... I'll revert sorry I missed this one before opening the PR

SpacePossum: restored and squashed

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
```diff
@@ -106,5 +78,42 @@ public function provideTestFixCases()
     );',
             ),
         );
+
+        $types = array('true', 'false', 'null');
+        // Equals' => true, 'NotEquals' => false); are transformed by PhpUnitStrictFixer to 'Same' and 'NotSame'
```
keradus: but this fixer can do it too, so why not test it?

SpacePossum: Good point, I cannot recall why I removed those from the initial PR.
Update the PR with `Equals` and I've simplified the test code a bit.

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
```diff
@@ -106,5 +78,27 @@ public function provideTestFixCases()
...
         );
+
+        return array_merge(
+            $cases,
+            $this->generateCases('<?php $this->assert%s(%s, $a, "%s", "%s");', '<?php $this->assert%s%s($a, "%s", "%s");'),
+            $this->generateCases('<?php $this->assert%s(%s, $a); //%s %s', '<?php $this->assert%s%s($a); //%s %s')
+        );
+    }
+
+    private function generateCases($fromTemplate, $toTemplate)
```
keradus: everywhere we use expected, input order.
it would be great to be consistent ;)

keradus:
:+1:

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
SpacePossum: fair enough, done and squashed :)

SpacePossum:
>@fabpot I think this fixer do what you want 

which is the same fixer being adjusted in this PR ^-^

Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php
keradus: That is why I don't like squashing when it is not needed.
sth goes wrong:
travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/85576570#L200

SpacePossum: I forgot to swap the inputs as well, things should be ok now :|

keradus:
Thank you @SpacePossum.

PR #1453, opened 2015-10-15
remove calling array_keys in foreach loops

keradus:


SpacePossum:
:+1: 

I would be nice if PHP 7 would support
```php
foreach($array as $key =>;) {
    //
}
```
but then again ;)

PR #1456, opened 2015-10-18
Added more phpdoc tests

GrahamCampbell:


keradus:
Adding more tests are great!
But why removing some of current tests?
eg
```php
/**
 * @var string Hello!
 */
```
is not the same as
```php
/** @var string Hello! */
```


Also, in mosts tests we strongly use `dataProvider`s, why you remove them here?

keradus:
ping @GrahamCampbell

GrahamCampbell:
Removing the esiting tests was unintentional. Adding separate methods was meant to improve readability.

GrahamCampbell:
I'll look at this at the weekend.

keradus:
Having providers is more common for this project and it's worth keep it for consistency.
It the test case is really tricky we can always comment what/why we test ;)

keradus:
ping @GrahamCampbell

Issue #1446, opened 2015-10-12
A class with a property/constant value which contains __DIR__ is not fixed


rybakit:
```php
// Foo.php

class Foo
{
    const BAZ = __DIR__.'/baz';

    function bar()
    {
        return "bar";
    }
}
```

Running

```sh
$ php-cs-fixer --level=symfony --dry-run --diff fix Foo.php
```

outputs

```sh
Fixed all files in 0.080 seconds, 5.750 MB memory used
```

After removing the `__DIR__` constant it starts outputting the expected result:

```
      ---------- begin diff ----------
      --- Original
      +++ New
      @@ @@
       <?php

       class Foo
       {
           const BAZ = '/baz';
      -
      -    function bar()
      +
      +    public function bar()           {
      -        return "bar";
      +        return 'bar';
           }
       }

      ---------- end diff ----------
```

SpacePossum:
Support for expressions in class consts was added in 5.6 (https://wiki.php.net/rfc/const_scalar_exprs)
Can you tell me what version of PHP you are running the fixer on and which version of the fixer itself are you using?
Could you run the command again with `-vvv` and check if it reports any relevant debug information?

rybakit:
Indeed, it's probably due to the PHP version mismatch. I tried to fix code written for php 5.6 under php 5.4. PHP CS Fixer version 1.10.1.

The output with `-vvv`:

```sh
I
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[0.075 s] /private/tmp/Foo.php

Fixed all files in 0.076 seconds, 5.750 MB memory used
```
It would be nice to output that file checking was skipped/aborted due to the invalid syntax, as the current message a bit misleading.

keradus:
The thing that you want is exactly what the `I` status is.

rybakit:
@keradus My concern is that the message is not accurate, it gives the false impression that everything is OK, when, in reality, all "fixed" files could have an invalid syntax.

keradus:
Invalid input is not fault of tool, if the file has syntax then it cannot be fixed, it is simply ommited.
And a message like `all files that are good and contain no syntax errors were fixed` overwhelmed.

rybakit:
> And a message like all files that are good and contain no syntax errors were fixed overwhelmed.

For example, in non-verbose mode PHPUnit outputs:
[...]

PR #1247, opened 2015-06-09
Refactor output

SpacePossum:
* Allow custom output through configuration
* more info on -v, -vv and -vvv for text
* check on is decorated everywhere

PR #1464, opened 2015-10-22
Add statistics about processed files to the output

rybakit:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1446


Symfony/CS/FixerFileProcessedStatsListener.php
```diff
@@ -0,0 +1,56 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS;
+
+/**
+ * @author Eugene Leonovich <gen.work@gmail.com>
```
keradus: Please add short descr.

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -422,7 +423,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
                 }
 
                 $fixEvent = $this->stopwatch->getEvent('fixFiles');
-                $output->writeln(sprintf('Fixed all files in %.3f seconds, %.3f MB memory used', $fixEvent->getDuration() / 1000, $fixEvent->getMemory() / 1024 / 1024));
+                $output->writeln(sprintf('Processed all files in %.3f seconds, %.3f MB memory used', $fixEvent->getDuration() / 1000, $fixEvent->getMemory() / 1024 / 1024));
+                if (!$showProgress) {
```
SpacePossum: Why not output these results when the process is shown as well? 
You can register 2 listeners, so your could always be added.

rybakit: My thoughts were that it will duplicate the status info. Actually, i did that in the very first implementation, but later realized that it's probably too much. But I don't mind to show the both results, if needed.

Symfony/CS/FixerFileProcessedEvent.php
```diff
@@ -39,13 +39,13 @@ class FixerFileProcessedEvent extends Event
...
-        self::STATUS_NO_CHANGES => array('symbol' => '.', 'description' => 'no changes'),
-        self::STATUS_FIXED => array('symbol' => 'F', 'description' => 'fixed'),
-        self::STATUS_EXCEPTION => array('symbol' => 'E', 'description' => 'error'),
-        self::STATUS_LINT => array('symbol' => 'E', 'description' => 'error'),
+        self::STATUS_UNKNOWN => array('symbol' => '?', 'name' => 'Unknown', 'description' => 'unknown'),
+        self::STATUS_INVALID => array('symbol' => 'I', 'name' => 'Invalid', 'description' => 'invalid file syntax, file ignored'),
+        self::STATUS_SKIPPED => array('symbol' => '', 'name' => 'Skipped',  'description' => ''),
+        self::STATUS_NO_CHANGES => array('symbol' => '.', 'name' => 'Unchanged', 'description' => 'no changes'),
+        self::STATUS_FIXED => array('symbol' => 'F', 'name' => 'Fixed', 'description' => 'fixed'),
+        self::STATUS_EXCEPTION => array('symbol' => 'E', 'name' => 'Error', 'description' => 'error'),
```
rybakit: I'm not sure about the 'Error' name. Maybe there are better alternatives?

keradus: `Exception` :D

Symfony/CS/Console/Command/FixCommand.php
keradus: IMO short stats should be show even when using progress indicator.
Even move that `Processed all files in %.3f seconds...` into `$eventListener` and we have whole a summary section

SpacePossum: sound like a plan :)
:+1:

rybakit: Then it will not be a stats listener anymore )

rybakit:
I added the `FixHelper` class and tried to keep changes in `FixCommand` to a minimum to ensure BC.

Symfony/CS/FixHelper.php
```diff
@@ -0,0 +1,278 @@
...
+
+namespace Symfony\CS;
+
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Stopwatch\Stopwatch;
+use Symfony\Component\Stopwatch\StopwatchEvent;
+
+/**
+ * Helps outputting status information when running the fix command.
```
keradus: Then it's not a helper, but class that do a specific task. No need to name it 'helper'

keradus:
I didn't suggesting that move you have done now because it is much bigger than you think and I wanted to keep this PR small to only change that summary section you want on beginning.

But OK, if you have time to work on it - it's great.
Take a look at #1247, take work from there and combine with work from here.
@SpacePossum, @rybakit please work on it together.

rybakit:
@keradus 
> I didn't suggesting that move you have done now

True, but I don't think that moving `Processed all files ..` part into the stats listener is a good idea, as it will make the listener aware of `StopwatchEvent` and force it to do something which is not directly belong to stats. By introducing `FixHelper` I tried to encapsulate the stopwatch (+ output) logic in one place instead of spreading it across different classes or/and polluting FixCommand even more.

> Take a look at #1247

My goal was to keep changes small as well as keep BC, but #1247 seems to break that. What is the current policy about BC and what parts of that PR you would like to see combined with this one?

Symfony/CS/FixerFileProcessedEvent.php
Soullivaneuh: > `Exception` :D

Oh yeah! \o/

keradus:
Moving whole output to separate class is great start. Then splitting formaters into separated classes for JSON, XML, TXT. Feel free to simply cherry pick changes from mentioned PR.

SpacePossum:
To be honest I don't feel much about picking up 1247 because no agreement was made about the architecture, i.e. outputting building using listeners vs. hard binding. I think that should be settled first before moving on with 1247.

keradus:
Exactly that was why I wasn't encourage you about that big changes at the beginning of this PR ;)

Symfony/CS/FixHelper.php
```diff
@@ -0,0 +1,278 @@
...
+     * @return mixed
+     */
+    public function fixFiles(\Closure $fixFiles, EventDispatcherInterface $eventDispatcher)
+    {
+        $listener = array($this, 'onFileProcessed');
+
+        $eventDispatcher->addListener(FixerFileProcessedEvent::NAME, $listener);
+        $this->stopwatch->start('fixFiles');
+
+        $result = $fixFiles();
```
keradus: Wow wow wow, you delegate fixing process this class? I thought it's purpose was to handle summary output (or now some more output), but not this class run a fix too? And by closure? IMO it's very bad and it reduces cleanliness of code and readability of flow. Sorry, but big :-1: for that solution.

I'm not saying that you should change the code ad hoc as soon you will read my comment, as you can see we already pretty sure that reformatting handling of output in `FixCommand` isn't easy.

GrahamCampbell: I agree with @keradus.

rybakit: Otherwise I have to write something like:

```php
$fixHelper->startFixing();
$this->fixer->fix(...);
$fixHelper->stopFixing();
```

which is not any better, or refactor `FixCommand` and break BC which I tried to avoid. But I'm open to any suggestions, if someone has a better one.

GrahamCampbell: Maybe just have a closure for the fix helper? Say `$fixHelper->fix(function () { $this->fixer->fix(...); });`.

rybakit: @GrahamCampbell that's what I actually did, see `FixHelper::fixFiles()`.

keradus:
Closing due to merge of #1824

PR #1465, opened 2015-10-27
Bump supported HHVM version, improve ShortEchoTagFixer on HHVM

keradus:


keradus:
yep, HHVM still failing

Symfony/CS/Fixer/PSR2/ElseifFixer.php
```diff
@@ -50,14 +50,6 @@ public function fix(\SplFileInfo $file, $content)
             $nextToken->clear();
         }
 
-        # handle `T_ELSE T_WHITESPACE T_IF` treated as single `T_ELSEIF` by HHVM
```
keradus: should start working from 3.7

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1329,39 +1329,9 @@ public function isMonolithicPhp()
...
 
-        /*
-         * Fix HHVM incompatibilities
-         */
-        $hhvmOpenTagsWithEcho = array();
-        $hhvmHashBangs = array();
-
-        if (defined('HHVM_VERSION')) {
-            /*
-             * HHVM parses '<?=' as T_ECHO instead of T_OPEN_TAG_WITH_ECHO
```
keradus: 3.8+

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1329,39 +1329,9 @@ public function isMonolithicPhp()
...
-                    $hhvmOpenTagsWithEcho[] = $token;
-                }
-            }
-
-            /*
-             * HHVM parses "#!/usr/bin/env php\n" as T_HASHBANG (not defined in
-             * PHP and T_HASHBANG. Moreover, HHVM does not define T_HASHBANG
-             * as a constant
-             *
-             * @see https://github.com/facebook/hhvm/issues/4810
```
keradus: 3.7+

GrahamCampbell:
The HHVM on travis is really old though. HHVM 3.9 is the current LTS, and 3.10 the latest.

Soullivaneuh:
Yeah, would be great to be able to set hhvm versions on Travis...

keradus:
If anyone know how to do that it would be great to share it with me, I can't find out how to do it :(

Soullivaneuh:
> If anyone know how to do that it would be great to share it with me, I can't find out how to do it :(

Yes, this what I was saying: It is not possible AFAIK. :-)

stof:
@Soullivaneuh the issue is that most builds on Travis are still running on Ubuntu Precise (12.04), which is not supported anymore by HHVM in recent versions. And their beta support of Ubuntu Trusty (14.04) does not come with PHP versions pre-installed yet, so I don't know which version of HHVM would get used

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1349,39 +1349,9 @@ public function isMonolithicPhp()
...
-            $tokens = self::fromCode("#!/usr/bin/env php\n");
-            if (!$tokens[0]->isGivenKind(T_INLINE_HTML)) {
-                $hashBangId = $tokens[0]->getId();
-                $hhvmHashBangs = $this->findGivenKind($hashBangId);
-            }
-        }
-
-        return 0 === count($kinds[T_INLINE_HTML]) + count($hhvmHashBangs) && 1 === count($kinds[T_OPEN_TAG]) + count($kinds[T_OPEN_TAG_WITH_ECHO]) + count($hhvmOpenTagsWithEcho);
+        return
+            0 === count($kinds[T_INLINE_HTML]) &&
```
GrahamCampbell: this would be fine on one line, infact prettier imo

.travis.yml
```diff
@@ -12,9 +12,12 @@ matrix:
         - php: 5.6
           env: SYMFONY_VERSION="~3.0@dev" COMPOSER_FLAGS="--prefer-stable"
         - php: 7.0
+        - php: hhvm-3.9
         - php: hhvm
 
-sudo: false
+sudo: required
```
Soullivaneuh: This can be specified on the concerned matrix target only: https://github.com/symfony/symfony/pull/18947/files

keradus: Great tip, thanks !

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1351,41 +1351,32 @@ public function isMonolithicPhp()
             return false;
         }
 
-        $kinds = $this->findGivenKind(array(T_OPEN_TAG, T_OPEN_TAG_WITH_ECHO, T_INLINE_HTML));
```
keradus: we are relying more and more on this function, let us not iterate through every single token if there is no need for that

GrahamCampbell:
LGTM, BUT, not for 1.11 IMO. This would be much better suited to 1.12 since it is not a bug fix as such. It is dropping support for ~15 versions of HHVM.

keradus:
doesn't matter
patch is for bugfixes, minor is for features.
refactoring and cleanups could go anywhere

GrahamCampbell:
This could be argued as a major breaking change in a patch release. Composer will install this for hhvm 3.6 users, and then it'll just straight up not work anymore.

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1351,41 +1351,32 @@ public function isMonolithicPhp()
...
+                )
+            ) {
+                if (++$openTagsCounter > 1) {
+                    return false;
+                }
             }
         }
 
-        return 0 === count($kinds[T_INLINE_HTML]) + count($hhvmHashBangs) && 1 === count($kinds[T_OPEN_TAG]) + count($kinds[T_OPEN_TAG_WITH_ECHO]) + count($hhvmOpenTagsWithEcho);
+        return true;
```
Soullivaneuh: This is indeed quite simpler. :trollface:

Symfony/CS/Tokenizer/Tokens.php
SpacePossum: Nice improvement!
Wouldn't be faster to iterate the tokens start at index 1 and if check for;
```php
if ($token->isGivenKind(array(T_INLINE_HTML, T_OPEN_TAG, T_OPEN_TAG_WITH_ECHO))) 
```

the reason is as the PHPDoc of the method states, the first token must be an open tag, so if you find an open tag after index 0 the file is not monolithic

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1351,41 +1351,32 @@ public function isMonolithicPhp()
...
-                }
+        $openTagsCounter = 0;
+
+        foreach ($this as $token) {
+            if ($token->isGivenKind(T_INLINE_HTML)) {
+                return false;
             }
 
-            /*
-             * HHVM parses "#!/usr/bin/env php\n" as T_HASHBANG (not defined in
```
SpacePossum: Nice to see this has been removed/dropped (https://3v4l.org/Quf63)
Maybe a good idea to add a test for this, something like:
```php
$this->assertFalse($tokens->isMonolithicPhp("#!/usr/bin/env php\n<?php // comment\n"));
```

Symfony/CS/Tokenizer/Tokens.php
keradus: code may be longer, but it's hell faster

keradus:
> This could be argued as a major breaking change in a patch release. Composer will install this for hhvm 3.6 users, and then it'll just straight up not work anymore.

Nope, major break is change of interface, here we don't change it. Bumping required PHP (or here - hhvm) version is not a BC breaker due to SEMVER specification.

But it's good point, I know how to specify requested PHP version in `composer.json`, but have no idea how to do this for hhvm

Symfony/CS/Tokenizer/Tokens.php
keradus: it's there: https://github.com/keradus/PHP-CS-Fixer/blob/8791b49cd6f714d0d0db19abf11485c862257f5e/Symfony/CS/Tests/Tokenizer/TokensTest.php#L870

Symfony/CS/Tokenizer/Tokens.php
keradus: i like that

GrahamCampbell:
To do it for HHVM, you have to actually require HHVM... so no PHP support...

keradus:
Then I don't care.

GrahamCampbell:
:trollface:

stof:
you can use a conflict rule to conflict with ``hhvm < 3.9``

keradus:
could you conflict with platform and not package ? :D

GrahamCampbell:
> you can use a conflict rule to conflict with hhvm < 3.9

That should work, yeh. :)

Symfony/CS/Tokenizer/Tokens.php
keradus: done

SpacePossum: :+1:  thanks, looks good :)

Soullivaneuh:
> you can use a conflict rule to conflict with hhvm < 3.9

More info: https://github.com/composer/composer/issues/5442#issuecomment-227652659

stof:
@keradus there is a ``hhvm`` package in composer, provided by the platform when running on HHVM (similar to the ``php`` package available all the time)

keradus:
done

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1351,41 +1351,24 @@ public function isMonolithicPhp()
...
+            if (
+                $this[$index]->isGivenKind(array(T_INLINE_HTML, T_OPEN_TAG, T_OPEN_TAG_WITH_ECHO))
+                || (
+                    /*
+                     * HHVM parses '<?=' as T_ECHO instead of T_OPEN_TAG_WITH_ECHO
+                     *
+                     * @see https://github.com/facebook/hhvm/issues/4809
+                     * @see https://github.com/facebook/hhvm/issues/7161
+                     */
+                    defined('HHVM_VERSION')
```
SpacePossum: moving the `defined` check out of the loop into a static var would be faster, although just a little bit I guess

SpacePossum:
looking good :)
:ship:

Symfony/CS/Tokenizer/Tokens.php
stof: this is a single lookup in the constant table, which cannot involve any autoloading code either (as PHP does not autoload constants, and HHVM will already have this constant defined so this will not trigger its constant autoloading logic either)

SpacePossum: You're right, thanks for the insights!
Never mind my comment, all good @keradus !

PR #1463, opened 2015-10-22
AlignDoubleArrowFixer - issue

Soullivaneuh:
@keradus @SpacePossum I'm sorry to say that #1450 is not completely resolved.

Here another proof. That code was not updated before 1.10.1/2.

PR #1466, opened 2015-10-28
Fix bad indent on align double arrow fixer

Soullivaneuh:
This is the same as #1463, I did something wrong with git.

### DIFF
--- a/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
+++ b/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
@@ -147,9 +147,7 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
                         $blockType = Tokens::detectBlockType($tokens[$arrayStartIndex]);
                         $arrayEndIndex = $tokens->findBlockEnd($blockType['type'], $arrayStartIndex);
 
-                        $arrayContent = $tokens->generatePartialCode($arrayStartIndex, $arrayEndIndex);
-
-                        if (false !== strpos($arrayContent, "\n")) {
+                        if ($tokens->isPartialCodeMultiline($arrayStartIndex, $arrayEndIndex)) {
                             break;
                         }
                     }
@@ -168,7 +166,7 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
     private function injectArrayAlignmentPlaceholders(Tokens $tokens, $from, $until)
     {
         // Only inject placeholders for multi-line arrays
-        if (false !== strpos($tokens->generatePartialCode($from, $until), "\n")) {
+        if ($tokens->isPartialCodeMultiline($from, $until)) {
             ++$this->deepestLevel;
             ++$this->currentLevel;
             $this->injectAlignmentPlaceholders($tokens, $from, $until);
--- a/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
+++ b/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
@@ -95,10 +95,7 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
                 $until = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $from);
                 $index = $until;
 
-                ++$this->deepestLevel;
-                ++$this->currentLevel;
-                $this->injectAlignmentPlaceholders($tokens, $from, $until);
-                --$this->currentLevel;
+                $this->injectArrayAlignmentPlaceholders($tokens, $from, $until);
                 continue;
             }
 
@@ -112,10 +109,7 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
                 $until = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_SQUARE_BRACE, $from);
                 $index = $until;
 
-                ++$this->deepestLevel;
-                ++$this->currentLevel;
-                $this->injectAlignmentPlaceholders($tokens, $from + 1, $until - 1);
-                --$this->currentLevel;
+                $this->injectArrayAlignmentPlaceholders($tokens, $from + 1, $until - 1);
                 continue;
             }
 
@@ -163,6 +157,22 @@ class AlignDoubleArrowFixer extends AbstractAlignFixer
                     ++$index;
                 }
             }
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $from
+     * @param int    $until
+     */
+    private function injectArrayAlignmentPlaceholders(Tokens $tokens, $from, $until)
+    {
+        // Only inject placeholders for multi-line arrays
+        if (false !== strpos($tokens->generatePartialCode($from, $until), "\n")) {
+            ++$this->deepestLevel;
+            ++$this->currentLevel;
+            $this->injectAlignmentPlaceholders($tokens, $from, $until);
+            --$this->currentLevel;
         }
     }
 }
--- a/Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/AlignDoubleArrowFixerTest.php
@@ -448,6 +448,118 @@ class AlignDoubleArrowFixerTest extends AbstractFixerTestBase
                 ;
                 ',
             ),
+            array(
+                '<?php
+    $dummy001 = $this->get("doctrine")->getRepository("AppBundle:Entity")->findBy(["server1" => $object], ["addedAt" => "DESC"], 5);
+    $foobar = $this->getDoctrine()->getRepository("AppBundle:Entity")->findBy(["server2" => $object], ["checkedAt" => "desc"], 50);
+    ',
+            ),
+            array(
+                '<?php
+    $dummy001 = $this->get("doctrine")->getRepository("AppBundle:Entity")->findBy(array("server1" => $object), array("addedAt" => "DESC"), 5);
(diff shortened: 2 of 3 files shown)


Soullivaneuh:
@keradus tests passed, ready to merge. :+1:

keradus:
I would like to get answer for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1463#discussion_r43188994 first
@cc GrahamCampbell

Soullivaneuh:
> I would like to get answer for #1463 (comment) first

I can understand, but this annoying bug is here from a long time and the fix is ready.

New tests passed and no regression with other tests, what is missing?

I would like to add tests later if @GrahamCampbell want it but I think this PR can be merged as is.

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
```diff
@@ -139,7 +139,7 @@ private function injectAlignmentPlaceholders(Tokens $tokens, $startAt, $endAt)
                 continue;
             }
 
-            if ($token->equals(',')) {
+            if ($token->equals(',') && !in_array($tokens[$index - 1]->getContent(), array(')', ']'), true)) {
```
keradus: use `equalsAny`

keradus:
> New tests passed and no regression with other tests, what is missing?

Votes with approval for this PR. Fix probably is good, but most of us just need more time than few hours to think about the bug, solution and discover edge cases

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
Soullivaneuh: Don't know about this method, will fix.

BTW, phpdoc should be improved to have better completion.

e.g.: `Tokens|Token[]`.

Soullivaneuh:
@keradus review corrected.

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
keradus: `Tokens` or `Token[]` or array of token prototype, where each prototype may be as well array like string.
eg:
```php
array(
    new Token(...),
    array(T_STRING, 'foo'),
    '('
)
```

Soullivaneuh: Well, `Tokens|Token[]|array` so. :-)

keradus: :+1:

Soullivaneuh:
@SpacePossum Could you please take a look of this one? Thanks!

GrahamCampbell:
What I was saying is you need to check these tokens actually belong to an array rather than:

`$foo[1],` where that would match a closing bracket and a comma.

GrahamCampbell:
Looks like we're missing a test case:

```php
<?php

function foo() {}

$bar = 42;

$foo = [
    'foo' => $bar[123],
    'a' => foo(),
];
```

GrahamCampbell:
If that gets fixed correctly after your change, than that's fine. I'd just like to be sure. :)

Soullivaneuh:
@GrahamCampbell I added your test, seems to be OK. :+1: 

cc @keradus

keradus:
looking on the condition, could you change the case into:

```diff
    $foo = [
        "test123" => "foo",
        "foo" => $bar[123],
        "a" => foo(),
+       "b" => 1,
    ];
```

?

GrahamCampbell:
:+1:

Soullivaneuh:
@keradus done.

SpacePossum:
Thanks for the PR. I've been testing a bit with it and I'm wondering what people think of the following two failing test cases.
Kindly note, these fail with and without your changes.

```php
            array(
                '<?php
    return array(
        self::STATUS_UNKNOWN    => array("symbol" => "?", "description" => "unknown"),
        self::STATUS_INVALID    => array("symbol123" => "III", "description" => "invalid file syntax, file ignored"),
    );',
                '<?php
    return array(
        self::STATUS_UNKNOWN => array("symbol" => "?", "description" => "unknown"),
        self::STATUS_INVALID    => array("symbol123" => "III", "description" => "invalid file syntax, file ignored"),
    );',
```

```php
            array(
                '<?php
    return array(
        self::STATUS_UNKNOWN    => array((1+11)=> "?", "description" => "unknown"),
        self::STATUS_INVALID    => array((2+3)=> "III", "description" => "invalid file syntax, file ignored"),
    );',
                '<?php
    return array(
        self::STATUS_UNKNOWN => array((1+11)=> "?", "description" => "unknown"),
        self::STATUS_INVALID    => array((2+3)=> "III", "description" => "invalid file syntax, file ignored"),
    );',
            ),
```

In other words, the inline array detection does not seem to work for all cases I think.

Soullivaneuh:
Thanks @SpacePossum, will take a look.

Soullivaneuh:
Sorry @SpacePossum, but what is the difference between your two pastes?

SpacePossum:
in case 1, hard index, space between '3"' and '=>' :
`array("symbol123" => "III",`

in case 2, 'dynamic' index, no space between ')' and '=>' :
`array((2+3)=> "III",`

Soullivaneuh:
Oh, didn't saw it. Need a :coffee:. :-)

Soullivaneuh:
I'm trying to check with line position:

```php
            if ($token->isGivenKind(T_ARRAY)) { // don't use "$tokens->isArray()" here, short arrays are handled in the next case
                $from = $tokens->getNextMeaningfulToken($index);
                $lineFrom = $tokens[$from]->getLine();
                $until = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $from);
                $lineUntil = $tokens[$until]->getLine();
                $index = $until;

                var_dump($lineFrom, $lineUntil);
                if ($lineUntil > $lineFrom) {
                    ++$this->deepestLevel;
                    ++$this->currentLevel;
                    $this->injectAlignmentPlaceholders($tokens, $from, $until);
                    --$this->currentLevel;
                }

                continue;
            }
```

But `getLine` always returns `NULL`.

@keradus do you know why?

keradus:
Yes. Never, never use `getLine`. As token prototype from `tokens_get_all` has it, `Token` class has it too. But when all our fixers do theirs stuff they change lines, adding new, removing old, and _do not_ update all lines of all tokens in collection. Do not rely on that method.
After said that, the method was removed on 2.x line.

Soullivaneuh:
Well, that's so bad. How can I deal with line? I don't know how to solve it with other way.

PHP code sniffer tokenizer have line management working well, maybe could we get inspiration of it?

keradus:
So far we need to handle line numbers in so few places, while keeping them up to date was costly, we decided to throw it away.
In that few places if we need to know if the tokens are in same line or not - we simply analyze content between, eg https://github.com/Soullivaneuh/PHP-CS-Fixer/blob/issue-1450-proof/Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php#L158

Soullivaneuh:
@SpacePossum @keradus fixed! :+1:

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
```diff
@@ -112,10 +116,14 @@ private function injectAlignmentPlaceholders(Tokens $tokens, $startAt, $endAt)
                 $until = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_SQUARE_BRACE, $from);
                 $index = $until;
 
-                ++$this->deepestLevel;
-                ++$this->currentLevel;
-                $this->injectAlignmentPlaceholders($tokens, $from + 1, $until - 1);
-                --$this->currentLevel;
+                // Inject placeholders only for multi lines array
+                if (false !== strpos($tokens->generatePartialCode($from, $until), "\n")) {
```
GrahamCampbell: maybe this needs abstracting to a method?

Soullivaneuh:
@GrahamCampbell suggestion added.

Symfony/CS/Fixer/Contrib/AlignDoubleArrowFixer.php
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1466#issuecomment-152152097 ;P

Soullivaneuh:
@keradus Are we now? :-)

keradus:
LGTM, just pleae get rid of that merge commit (eg rebase on 1.10)

Soullivaneuh:
Indeed, nicer history. :-)

PR ready to merge! :+1: 

I'm curious: Why a simple `rebase 1.10` can get rid of merge commits? And what would happen if I was already up to date from `1.10`?

keradus:
After rebasing on 1.10 the merge commit wouldn't bring any code changes (because you are working on top of 1.10), therefore it will be considered as empty commit, and while rebasing empty commits are ommited by default.
If your branch would be created on top of 1.10 - then you wouldn't have merge commit of 1.10 (that would mean the 1.10 has new commits and your branch is not created on top of 1.10).

keradus:
:+1:

SpacePossum:
looking good :+1:

Soullivaneuh:
Thanks. :-)

When this could be merged?

keradus:
Thank you @Soullivaneuh.

SpacePossum:
thanks @Soullivaneuh !

Soullivaneuh:
Thank you guys! :+1:

Soullivaneuh:
@keradus All PR seems to be [merged](https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestones/v1.10.3), do you have a date for v1.10.3 release?

keradus:
I was already trying to do it for few days... I have just change the country, please don't be impatient ;)

PR #1248, opened 2015-06-09
[WIP] Config resolver should rely on the ConfigInterface

SpacePossum:
In the docs it states people can provide their our instance of a ConfigInterface and the fixer will use it to run. However the ConfigResolver depends on the Config class. The diversion of the Config class with the ConfigInterface causes the resolver not being able to deal with an implementation of the ConfigInterface.

PR #1468, opened 2015-10-29
Add missing ConfigInterface::getHideProgress()

rybakit:
See: 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/ConfigurationResolver.php#L39
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/ConfigurationResolver.php#L87

### DIFF
--- a/Symfony/CS/ConfigurationResolver.php
+++ b/Symfony/CS/ConfigurationResolver.php
@@ -20,9 +20,24 @@ namespace Symfony\CS;
  */
 class ConfigurationResolver
 {
+    /**
+     * @var FixerInterface[]
+     */
     protected $allFixers;
+
+    /**
+     * @var ConfigInterface
+     */
     protected $config;
+
+    /**
+     * @var FixerInterface[]
+     */
     protected $fixers = array();
+
+    /**
+     * @var array
+     */
     protected $options = array(
         'fixers' => null,
         'level' => null,
--- a/Symfony/CS/ConfigurationResolver.php
+++ b/Symfony/CS/ConfigurationResolver.php
@@ -99,6 +99,8 @@ class ConfigurationResolver
 
     public function getProgress()
     {
+        // TODO: following condition should be removed on 2.0 line
+        // and method should be added to ConfigInterface
         if (!method_exists($this->config, 'getHideProgress')) {
             return $this->options['progress'];
         }
--- a/Symfony/CS/ConfigurationResolver.php
+++ b/Symfony/CS/ConfigurationResolver.php
@@ -99,6 +99,10 @@ class ConfigurationResolver
 
     public function getProgress()
     {
+        if (!method_exists($this->config, 'getHideProgress')) {
+            return $this->options['progress'];
+        }
+
         return $this->options['progress'] && !$this->config->getHideProgress();
     }
 
--- a/Symfony/CS/Tests/ConfigurationResolverTest.php
+++ b/Symfony/CS/Tests/ConfigurationResolverTest.php
@@ -68,7 +68,7 @@ class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
 
     public function testResolveFixersWithLevelConfig()
     {
-        $this->config->expects($this->once())->method('getLevel')
+        $this->config->expects($this->any())->method('getLevel')
             ->will($this->returnValue(FixerInterface::PSR1_LEVEL));
 
         $this->resolver->resolve();
@@ -81,10 +81,10 @@ class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
 
     public function testResolveFixersWithPositiveFixersConfig()
     {
-        $this->config->expects($this->once())->method('getLevel')
+        $this->config->expects($this->any())->method('getLevel')
             ->will($this->returnValue(FixerInterface::SYMFONY_LEVEL));
 
-        $this->config->expects($this->once())->method('getFixers')
+        $this->config->expects($this->any())->method('getFixers')
             ->will($this->returnValue(array('strict', 'strict_param')));
 
         $this->resolver->resolve();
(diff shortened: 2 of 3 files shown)


SpacePossum:
It is not, you can use the result of:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/ConfigurationResolver.php#L85

keradus:
@SpacePossum method you pointed need `ConfigInterface` to has `getHideProgress` method

rybakit:
@SpacePossum I didn't get how using the result of `getProgress()` can help here. Consider the following example:

```php
class FooConfig implements ConfigInterface
{
    public function getName() { return 'foo'; }
    public function getDescription() { return 'foobar'; }
    public function getFinder() { return []; }
    public function getLevel() { return FixerInterface::PSR2_LEVEL; }
    public function getFixers() { return []; }
    public function setDir($dir) {}
    public function getDir() {}
    public function addCustomFixer(FixerInterface $fixer) {}
    public function getCustomFixers() { return []; }
}

$resolver = new ConfigurationResolver();
$resolver->setConfig(new FooConfig());

$hideProgress = $resolver->getHideProgress(); // fatal error
```

SpacePossum:
I don't think the `show process` option should be part of the interface. Just like `is dry run` and the `verbosity level` . I'm not sure about `'use cache` and `generate diff`.
Anyway, there are more problems with the interface vs. the implementation and I would not add the `process` option to keep the interface as small as possible (it is already big and incomplete).

just for ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1248

rybakit:
@SpacePossum I agree that `ConfigInterface` needs refactoring, but that's out of the scope of this PR. This PR is about fixing broken `ConfigurationResolver` implementation, which relays on non-existing method. So either `getHideProgress()` should to be added to the interface or the method call should be removed from `ConfigurationResolver`.

keradus:
1.x: if method exists in config instance - call it, if not get default value.
2.x: ad method to interface

rybakit:
> 1.x: if method exists in config instance - call it, if not get default value.

What do you propose to do in such cases:

```php
class FooConfig implements ConfigInterface
{
    ...

    public function getHideProgress($requiredArg)
    {
    }
}
```
Is it assumed that `getHideProgress()` must not have any arguments and we can just do `method_exists()` check before calling it?

keradus:
Is sb make that method but with args then the method is not compatible with the call and it will fail.
But it is so edge case that I wouldn't care about it.

Symfony/CS/ConfigurationResolver.php
```diff
@@ -84,6 +99,10 @@ public function getFixers()
 
     public function getProgress()
     {
+        if (!method_exists($this->config, 'getHideProgress')) {
```
rybakit: An alternative approach is to check against the Config instance:

```php
if (!$this->config instanceof Config)) {
```

keradus: it will never happen
https://github.com/rybakit/PHP-CS-Fixer/blob/config_interface/Symfony/CS/ConfigurationResolver.php#L54

keradus: Oh, you mean `Config`, not `ConfigInterface`.
That could work too, but I think calling `method_exists` is more generic

rybakit: It's more generic indeed, but also more error prone (think of `getHideProgress($requiredArg)`). But yeah, it's not that important.

Symfony/CS/Tests/ConfigurationResolverTest.php
```diff
@@ -69,7 +68,8 @@ public function testResolveFixersReturnsEmptyArrayByDefault()
 
     public function testResolveFixersWithLevelConfig()
     {
-        $this->config->level(FixerInterface::PSR1_LEVEL);
+        $this->config->expects($this->once())->method('getLevel')
```
keradus: You don't know how many times the `getLevel` will be called, and the getter returns const value (the value shouldn't changed between calling).
I would suggest `->any()` here and in other places

Symfony/CS/Tests/ConfigurationResolverTest.php
```diff
@@ -257,9 +287,20 @@ public function testResolveProgressWithNegativeConfigAndPositiveOption()
 
     public function testResolveProgressWithNegativeConfigAndNegativeOption()
     {
-        $this->config->hideProgress(false);
+        $config = $this->getMock('Symfony\CS\Config\Config');
```
keradus: Mock is already created in `setUp` under `$this->config`.
No need to create new one and inject it to `$this->resolver`.

Symfony/CS/ConfigurationResolver.php
```diff
@@ -84,6 +99,10 @@ public function getFixers()
...
+            return $this->options['progress'];
```
keradus: Please, add a TODO like
`// TODO: following condition should be removed on 2.0 line and method should be added to ConfigInterface`

keradus:
Great work @rybakit !
Just raised some minors

Symfony/CS/Tests/ConfigurationResolverTest.php
rybakit: Agree. Fixed.

Symfony/CS/Tests/ConfigurationResolverTest.php
rybakit: It's needed because `ConfigInterface` doesn't have `getHideProgress()`.

keradus: Indeed.

keradus:
:+1:

keradus:
Thank you @rybakit.

keradus:
Merged into 1.10 line

PR #1470, opened 2015-10-29
Add blank line between methods

pyrech:
I'm not really satisfied with the fixer logic but it works. What do you think of it? Is it something that could be added to php-cs-fixer?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -537,6 +537,10 @@ Choose from the list of available fixers:
                         Add, replace or remove header
                         comment.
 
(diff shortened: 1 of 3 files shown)


keradus:
Please, take a look at [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Fixer/Symfony/MethodSeparationFixer.php) and tell the difference.

pyrech:
Huhuu, didn't find it -_-

Sorry for that

PR #1483, opened 2015-11-04
[1.10] Support "@property-read" and "@property-write"

GrahamCampbell:
Noticed this when I ran php-cs-fixer on Carbon. We now have this committed:

image

We need to get the other stuff fixed too basically. :)

### DIFF
--- a/Symfony/CS/DocBlock/Tag.php
+++ b/Symfony/CS/DocBlock/Tag.php
@@ -26,8 +26,9 @@ class Tag
     private static $tags = array(
         'api', 'author', 'category', 'copyright', 'deprecated', 'example',
         'global', 'internal', 'license', 'link', 'method', 'package', 'param',
-        'property', 'return', 'see', 'since', 'struct', 'subpackage', 'throws',
-        'todo', 'typedef', 'uses', 'var', 'version',
+        'property', 'property-read', 'property-write', 'return', 'see',
+        'since', 'struct', 'subpackage', 'throws', 'todo', 'typedef', 'uses',
+        'var', 'version',
     );
 
     /**
@@ -45,7 +46,7 @@ class Tag
     public function __construct($content)
     {
         $this->name = 'other';
-        preg_match_all('/@[a-zA-Z0-9_]+(?=\s|$)/', $content, $matches);
+        preg_match_all('/@[a-zA-Z0-9_-]+(?=\s|$)/', $content, $matches);
 
         if (isset($matches[0][0])) {
             $this->name = ltrim($matches[0][0], '@');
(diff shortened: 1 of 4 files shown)


joecohens:
:+1:

keradus:
:-1:

They were annotations from phpDocumentator.
We follow PSR-5 proposal, which doesn't allow such tags: https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#8-tags

GrahamCampbell:
Yeh, but in the real world, everyone uses it.

GrahamCampbell:
In the same way we're not meant to use `@var`, but we do...

SpacePossum:
:-1: for supporting the tags using the same fixers as for PSR5 by default. 
:+1: when using a separate fixer or a configuration option on the current fixer(s).

GrahamCampbell:
> when using a separate fixer or a configuration option on the current fixer(s).

We can't do that though, since stuff has changed that isn't even on the fixer level.

keradus:
any other thoughts? anyone?

GrahamCampbell:
Would need to go to 1.11, so see my other PR.

GrahamCampbell:
I've also got one more bug to fix in there too.

keradus:
even before deciding about version - there is no decision do we want to support those tags

GrahamCampbell:
Well, I think it's quite important we do. See the new test cases. The separation fixer is broken otherwise.

keradus:
that's why I want to have some more feedback of other users ;)

GrahamCampbell:
Sure. The other issue is still open, and people can still comment here. ;)

keradus:
OK, I'm :+1: for change.

If we don't add that we won't be able to add fixer that normalizes non-PSR5 `@property-*` tag into proper one ;P

GrahamCampbell:
Well, I think `@property` by itself isn't good enough. If someone only has `__set` implemented, but not `__get`, then saying a property that can be both read and written to is wrong, so we really do need these other two variants.

GrahamCampbell:
I don't think adding a fixer to convert these other variants to standard `@property` annotations as well as this improvement is such a bad idea though, for those than want total compliance with the current spec.

SpacePossum:
:+1:  for adding the changes to support more (non-PSR5) tags

GrahamCampbell:
Great. ;)

keradus:
It's always great to convert non-PSR5 tag to it's equivalent in PSR5 ;) but yeah, it's not that everyone need to use it!

GrahamCampbell:
Another reason to keep my abstract var -> type fixer thing. We can use it do this. :P

keradus:
And you still pointing out... my argument.
To have abstraction for similar changes, not for similar rules ;)

GrahamCampbell:
That's fine then. ;)

GrahamCampbell:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1487 is ready for review now I think. :)

PR #1487, opened 2015-11-07
[1.11] Added psr5 property fixer

GrahamCampbell:
If you want to consider #1483 as a new feature, then here's a PR for that case.

NB This PR is different because of fixer changes only present on 1.11.

### DIFF
--- a/Symfony/CS/AbstractPhpdocTagsFixer.php
+++ b/Symfony/CS/AbstractPhpdocTagsFixer.php
@@ -1,0 +1,49 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS;
+
+use Symfony\CS\DocBlock\DocBlock;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * This abstract fixer provides a base for fixers to rename tags.
+ *
+ * @author Graham Campbell <graham@mineuk.com>
+ */
+abstract class AbstractPhpdocTagsFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
+            $doc = new DocBlock($token->getContent());
+            $annotations = $doc->getAnnotationsOfType(static::$search);
+
+            if (empty($annotations)) {
+                continue;
+            }
+
+            foreach ($annotations as $annotation) {
+                $line = $doc->getLine($annotation->getStart());
+                $line->setContent(str_replace(static::$input, static::$output, $line->getContent()));
+            }
+
+            $token->setContent($doc->getContent());
+        }
+
+        return $tokens->generateCode();
+    }
+}
--- a/Symfony/CS/Fixer/Symfony/PhpdocTypeToVarFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocTypeToVarFixer.php
@@ -11,40 +11,33 @@
 
 namespace Symfony\CS\Fixer\Symfony;
 
-use Symfony\CS\AbstractFixer;
-use Symfony\CS\DocBlock\DocBlock;
-use Symfony\CS\Tokenizer\Tokens;
+use Symfony\CS\AbstractPhpdocTagsFixer;
 
 /**
  * @author Graham Campbell <graham@mineuk.com>
  */
-class PhpdocTypeToVarFixer extends AbstractFixer
+class PhpdocTypeToVarFixer extends AbstractPhpdocTagsFixer
 {
     /**
-     * {@inheritdoc}
+     * The tag to search for.
+     *
+     * @var string
      */
-    public function fix(\SplFileInfo $file, $content)
-    {
-        $tokens = Tokens::fromCode($content);
+    protected static $search = 'type';
 
-        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
-            $doc = new DocBlock($token->getContent());
-            $annotations = $doc->getAnnotationsOfType('type');
+    /**
+     * The input tag.
+     *
+     * @var string
+     */
+    protected static $input = '@type';
 
-            if (empty($annotations)) {
-                continue;
-            }
-
-            foreach ($annotations as $annotation) {
-                $line = $doc->getLine($annotation->getStart());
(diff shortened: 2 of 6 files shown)


GrahamCampbell:
Fixed the separation fixer now.

GrahamCampbell:
Also added a fixer to normalize property-read and property-write to property.

Symfony/CS/AbstractPhpdocTagsFixer.php
```diff
@@ -0,0 +1,49 @@
...
+
+use Symfony\CS\DocBlock\DocBlock;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * This abstract fixer provides a base for fixers to rename tags.
+ *
+ * @author Graham Campbell <graham@mineuk.com>
+ */
+abstract class AbstractPhpdocTagsFixer extends AbstractFixer
```
keradus: it should run before other phpdoc fixers, like aligning or sth. do we get that out of the box?

GrahamCampbell: The aligning fixers make sure the run after the others, yeh. ;)

keradus: gr8 ;)

Symfony/CS/AbstractPhpdocTagsFixer.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
```
keradus: I thought we decided to not go that way as `findGivenKind` isn't generator, or eevn iterator.

Symfony/CS/AbstractPhpdocTagsFixer.php
```diff
@@ -0,0 +1,49 @@
...
+            $doc = new DocBlock($token->getContent());
+            $annotations = $doc->getAnnotationsOfType(static::$search);
+
+            if (empty($annotations)) {
+                continue;
+            }
+
+            foreach ($annotations as $annotation) {
+                $line = $doc->getLine($annotation->getStart());
+                $line->setContent(str_replace(static::$input, static::$output, $line->getContent()));
```
keradus: abstraction level should know about that props

Symfony/CS/AbstractPhpdocTagsFixer.php
GrahamCampbell: Not sure what you mean? This code is identical to what was already there?

Symfony/CS/AbstractPhpdocTagsFixer.php
GrahamCampbell: Not sure what you mean here?

keradus: that there are no definitions of that properties (eg `$input`) in base class. then when wan't to override the class and fill the props he can't inherit the docs

Symfony/CS/Fixer/Symfony/PhpdocTypeToVarFixer.php
```diff
@@ -11,40 +11,33 @@
...
 
 /**
  * @author Graham Campbell <graham@mineuk.com>
  */
-class PhpdocTypeToVarFixer extends AbstractFixer
+class PhpdocTypeToVarFixer extends AbstractPhpdocTagsFixer
 {
     /**
-     * {@inheritdoc}
+     * The tag to search for.
```
keradus: like this, the docs should just be inherited

Symfony/CS/AbstractPhpdocTagsFixer.php
GrahamCampbell: How can I define them? We can't have abstract static properties?

keradus: I treat all static things as evil ones.
Do they need to be static?

If yes - instead of calling non-existing properties have some getters.

Symfony/CS/AbstractPhpdocTagsFixer.php
keradus: maybe we didn't change it everywhere then

Symfony/CS/AbstractPhpdocTagsFixer.php
GrahamCampbell: That has a performance impact. I wrote this in a way that has the best performance possible. That's why the input/search stuff is kind of duplicated too.

GrahamCampbell: Accessing a static property 1000 times is significanty faster than doing 1000 function calls.

keradus: you could cache the function result in static inside `fix` method.


one way or another if the base class need some structure the class itself must be aware of them.
then, again, why to keep it static?

GrahamCampbell: i thought about that, but actually, i can't, because the result for the extending classes is different, so the static cache inside the fix method would be wrong

keradus: 3rd time, why to keep it static? see no benefits

GrahamCampbell: Can we have non-static abstract properties?

GrahamCampbell: > Fatal error: Properties cannot be declared abstract

GrahamCampbell: > see no benefits

I see no difference to not having it static?

keradus: you could leave them null and describe them

GrahamCampbell: Even the, the phpdoc isn't actually the same for each property.

GrahamCampbell: Some of them are `string[]` in some classes, and `string` in others.

keradus: oh, that's sad

GrahamCampbell:
Rebased.

GrahamCampbell:
Working on a bigger PR that will address the inconsistant properties issue, although more of a side effect of the other refactoring. The motivation for this next incoming PR is to provide more abstraction at the docblock class level, to allow me to write more advanced fixers, like, for example, alphabetically sorting types that are chained with the `|`, or detecting redundant types like `false|bool` and fixing to `bool`. In the long term, I also plan to write a fixer that will analyze the default values in params and/or the actual type hint to check the phpdoc actually matches.

GrahamCampbell:
Sure. ;)

GrahamCampbell:
Will send in smaller chunks.

PR #1488, opened 2015-11-07
Added more inline phpdoc tests

GrahamCampbell:


keradus:
:+1:

Symfony/CS/Tests/Fixer/Symfony/PhpdocScalarFixerTest.php
```diff
@@ -272,4 +272,35 @@ public function testWrongCasedPhpdocTagIsNotAltered()
...
+    public function testInlineDoc()
+    {
+        $expected = <<<'EOF'
+<?php
+    /**
+     * Does stuffs with stuffs.
+     *
+     * @param array $stuffs {
+     *     @type bool $foo
+     *     @type int $bar
```
SpacePossum: will these be aligned by the `PhpdocAlignFixer` ?

GrahamCampbell: Yes, but that's unrelated to this PR?

keradus: I think that @SpacePossum asked for such a case, as you are extending them

GrahamCampbell: Right. Can we leave that to another PR, since I have nothing to do with the alignment fixer. TBH, I'd still like to see that rewritten to use the phpdoc classes we have for consistent behaviour.

keradus: Would adding that one tests, that you told would pass, be a problem?

> TBH, I'd still like to see that rewritten to use the phpdoc classes we have for consistent behaviour.

Go for it! :)

GrahamCampbell: > Go for it! :)

Ha. If I had the time, I'd love to. It's not a 5 min job I'm afraid. ;)

SpacePossum: it just got curious about it, that is all, now I know it will be done by the other fixer, so

>Can we leave that to another PR,

sure :) 
thanks for clearing it up :+1:

keradus:
Thank you @GrahamCampbell.

Issue #1462, opened 2015-10-22
Suggested new fixer: MethodArgumentDefaultValue


dereuromark:
Suggested fixer: MethodArgumentDefaultValue (besides existing MethodArgumentSpace)

Default arguments in front of non default ones should be fixed=removed, as they cannot be used.

```php
class TestClass
{
    public function x($foo = true, $bar)
    {
    }
}
```
Should be fixed to
```php
class TestClass
{
    public function x($foo, $bar)
    {
    }
}
```

keradus:
doesn't PHP 7 allow to use them like
```php
$a = new TestClass();
$a->x(default, 'foo')
```
?

Soullivaneuh:
> doesn't PHP 7 allow to use them like

PHP7 maybe, PHP5 not.

This is not the question so, this is like short and long array syntax. Depends the supported version too.

I think this fixer has his place on contrib section. :+1:

dereuromark:
Independent from the version this could also be considered a code smell in general, many things are allowed, but if that would be a good thing to do is usually a different topic :)

A rough but working version:

```php
/**
 * Fixer MethodArgumentDefaultValue
 *
 * @author Mark Scherer
 */
class MethodArgumentDefaultValueFixer extends AbstractFixer
{

    protected $defaultValueForbidden = false;

    /**
     * {@inheritdoc}
     */
    public function fix(\SplFileInfo $file, $content)
    {
        $tokens = Tokens::fromCode($content);

        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
            $token = $tokens[$index];

            // looking for start of brace and skip array
            if (!$token->equals('(') || $tokens[$index - 1]->isGivenKind(T_ARRAY)) {
                continue;
            }

            $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $index);

            // fix for method argument
            for ($i = $endIndex - 1; $i > $index; --$i) {
                if (!$tokens[$i]->equals(',')) {
                    continue;
                }

                $this->fixDefaultValues($tokens, $i, $endIndex - 1);
            }

            $this->fixDefaultValues($tokens, $i, $endIndex - 1);
        }

        return $tokens->generateCode();
    }

    /**
     * {@inheritdoc}
     */
    public function getDescription()
    {
        return 'In method arguments there must not be arguments with default values before non-default ones.';
    }

    /**
[...]

PR #1493, opened 2015-11-09
Added MethodArgumentDefaultValueFixer

lmanzke:
Hey guys, glad to be doing my first PR for you :).

My colleague @dereuromark opened an issue for a new fixer suggestion a while ago.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1462

Here is the PR implementing it. I took his solution and refactored it, please feel free to have a look.

I tried to make it as readable as possible. 

Maybe you have some suggestion on how finding the next occurrence of a Variable OR certain token might be done more conveniently. I thought I could use

```
$tokens->getNextTokenOfKind(T_VARIABLE) 
```

here, but it does not work that way ;).

Looking forward to your feedback!

Best regards,

Lucas

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -339,6 +339,12 @@ Choose from the list of available fixers:
                         Remove trailing commas in list
                         function calls.
 
+* **method_argument_default_value** [symfony]
+                        In method arguments there must
+                        not be arguments with default
+                        values before non-default
+                        ones.
+
 * **multiline_array_trailing_comma** [symfony]
                         PHP multi-line arrays should
                         have a trailing comma.
@@ -578,12 +584,6 @@ Choose from the list of available fixers:
 * **long_array_syntax** [contrib]
                         Arrays should use the long
                         syntax.
-
-* **method_argument_default_value** [contrib]
-                        In method arguments there must
-                        not be arguments with default
-                        values before non-default
-                        ones.
 
 * **multiline_spaces_before_semicolon** [contrib]
                         Multi-line whitespace before
--- a/README.rst
+++ b/README.rst
@@ -579,6 +579,12 @@ Choose from the list of available fixers:
                         Arrays should use the long
                         syntax.
 
+* **method_argument_default_value** [contrib]
+                        In method arguments there must
+                        not be arguments with default
+                        values before non-default
+                        ones.
+
 * **multiline_spaces_before_semicolon** [contrib]
                         Multi-line whitespace before
                         closing semicolon are
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -91,7 +91,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
             return false;
         }
 
-        while (!in_array($tokens[$nextRelevantTokenIndex]->getContent(), $this->functionDefinitionTerminatorTokens)) {
+        while (!in_array($tokens[$nextRelevantTokenIndex]->getContent(), $this->functionDefinitionTerminatorTokens, true)) {
             $nextRelevantTokenContent = $tokens[$tokens->getNextMeaningfulToken($nextRelevantTokenIndex)]->getContent();
 
             if (in_array($nextRelevantTokenContent, $this->argumentTerminatorTokens, true)) {
@@ -106,8 +106,9 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
 
     /**
      * @param Tokens $tokens
-     * @param int $index
-     * @param array $relevantTokens
+     * @param int    $index
+     * @param array  $relevantTokens
+     *
      * @return int|null
      */
     private function findNextVariableOrTokenOfKind(Tokens $tokens, $index, array $relevantTokens)
@@ -130,7 +131,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
 
     /**
      * @param array $values
-     * @param int $minimumValue
+     * @param int   $minimumValue
      *
      * @return int|null
      */
@@ -159,7 +160,8 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
 
     /**
      * @param Tokens $tokens
-     * @param int $index
+     * @param int    $index
+     *
      * @return bool
      */
     private function isDefaultArgumentAfterIndex(Tokens $tokens, $index)
@@ -172,7 +174,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
 
     /**
      * @param Tokens $tokens
-     * @param int $nextVariableIndex
+     * @param int    $nextVariableIndex
      */
     private function removeDefaultArgument(Tokens $tokens, $nextVariableIndex)
     {
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -186,14 +186,6 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
     /**
      * @return int
      */
-    public function getPriority()
-    {
-        return -49;
-    }
-
-    /**
-     * @return int
-     */
     public function getLevel()
     {
         return FixerInterface::CONTRIB_LEVEL;
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -218,11 +218,10 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
      */
     private function clearWhitespacesBeforeIndex(Tokens $tokens, $index)
     {
-        $currentIndex = $index - 1;
+        $token = $tokens[$index - 1];
 
-        while ($tokens[$currentIndex]->isGivenKind(T_WHITESPACE)) {
-            $tokens[$currentIndex]->clear();
-            --$currentIndex;
+        if ($token->isGivenKind(T_WHITESPACE)) {
+            $token->clear();
         }
     }
 
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -143,10 +143,10 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
         $currentIndex = $nextVariableIndex;
 
         $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($nextVariableIndex);
-        $nextMeaningfulTokenIndex = $tokens->getNextTokenOfKind($nextVariableIndex, array(array(310, 'null'), ',', ')'));
+        $nextMeaningfulTokenIndex = $tokens->getNextTokenOfKind($nextVariableIndex, array(array(T_STRING, 'null'), ',', ')'));
 
         if (
-            $tokens[$nextMeaningfulTokenIndex]->equals(array(310, 'null')) &&
+            $tokens[$nextMeaningfulTokenIndex]->equals(array(T_STRING, 'null')) &&
             !$tokens[$prevMeaningfulTokenIndex]->equalsAny($this->argumentBoundaryTokens)
         ) {
             return; //ignore typehinted null default values
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -22,7 +22,7 @@ use Symfony\CS\Tokenizer\Tokens;
 final class MethodArgumentDefaultValueFixer extends AbstractFixer
 {
     private $argumentBoundaryTokens = array('(', ',', ')', ';', '{', '}');
-    private $functionDefinitionTerminatorTokens = array(')', ';', '{', '}');
+    private $variableOrTerminatorTokens = array(array(T_VARIABLE), ')', ';', '{', '}');
     private $argumentTerminatorTokens = array(',', ')', ';', '{');
     private $defaultValueTokens = array('=', ')', ';', '{');
     private $immediateDefaultValueTokens = array('=', ')', ',', ';', '{');
@@ -63,7 +63,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
             $examinedIndex < $lastNonDefaultArgumentIndex &&
             $this->hasDefaultValueAfterIndex($tokens, $examinedIndex)
         ) {
-            $nextRelevantIndex = $this->findNextVariableOrTokenOfKind($tokens, $examinedIndex, $this->functionDefinitionTerminatorTokens);
+            $nextRelevantIndex = $tokens->getNextTokenOfKind($examinedIndex, $this->variableOrTerminatorTokens);
 
             if (!$tokens[$nextRelevantIndex]->isGivenKind(T_VARIABLE)) {
                 break;
@@ -87,7 +87,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
      */
     private function getLastNonDefaultArgumentIndex(Tokens $tokens, $index)
     {
-        $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $index, $this->functionDefinitionTerminatorTokens);
+        $nextRelevantTokenIndex = $tokens->getNextTokenOfKind($index, $this->variableOrTerminatorTokens);
 
         if (null === $nextRelevantTokenIndex) {
             return;
@@ -100,50 +100,10 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
                 $lastNonDefaultArgumentIndex = $nextRelevantTokenIndex;
             }
 
-            $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $nextRelevantTokenIndex, $this->functionDefinitionTerminatorTokens);
+            $nextRelevantTokenIndex = $tokens->getNextTokenOfKind($nextRelevantTokenIndex, $this->variableOrTerminatorTokens);
         }
 
         return $lastNonDefaultArgumentIndex;
-    }
-
-    /**
-     * @param Tokens $tokens
-     * @param int    $index
-     * @param array  $relevantTokens
-     *
-     * @return int|null
-     */
-    private function findNextVariableOrTokenOfKind(Tokens $tokens, $index, array $relevantTokens)
-    {
-        $variableIndices = array_keys($tokens->findGivenKind(T_VARIABLE));
-
-        $nextVariableIndex = $this->getFirstValueBiggerThan($variableIndices, $index);
-        $nextRelevantTokenIndex = $tokens->getNextTokenOfKind($index, $relevantTokens);
-
-        if (null === $nextVariableIndex) {
-            return $nextRelevantTokenIndex;
-        }
-
-        if (null === $nextRelevantTokenIndex) {
-            return $nextVariableIndex;
-        }
-
-        return ($nextVariableIndex < $nextRelevantTokenIndex) ? $nextVariableIndex : $nextRelevantTokenIndex;
-    }
-
-    /**
-     * @param array $values
-     * @param int   $minimumValue
-     *
-     * @return int|null
-     */
-    private function getFirstValueBiggerThan(array $values, $minimumValue)
-    {
-        foreach ($values as $value) {
-            if ($value > $minimumValue) {
-                return $value;
-            }
-        }
     }
 
     /**
@@ -188,7 +148,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
     }
 
     /**
-     * @return int
+     * {@inheritdoc}
      */
     public function getLevel()
     {
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -57,8 +57,10 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
     private function fixFunctionDefinition(Tokens $tokens, $index)
     {
         $examinedIndex = $tokens->getNextTokenOfKind($index, $this->argumentBoundaryTokens);
+        $lastNonDefaultArgumentIndex = $this->getLastNonDefaultArgumentIndex($tokens, $index);
+
         while (
-            $this->hasNonDefaultArgumentAfterIndex($tokens, $examinedIndex) &&
+            $examinedIndex < $lastNonDefaultArgumentIndex &&
             $this->hasDefaultValueAfterIndex($tokens, $examinedIndex)
         ) {
             $nextRelevantIndex = $this->findNextVariableOrTokenOfKind($tokens, $examinedIndex, $this->functionDefinitionTerminatorTokens);
@@ -69,7 +71,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
 
             if (
                 $this->isDefaultArgumentAfterIndex($tokens, $nextRelevantIndex - 1) &&
-                $this->hasNonDefaultArgumentAfterIndex($tokens, $nextRelevantIndex - 1)
+                $nextRelevantIndex - 1 < $lastNonDefaultArgumentIndex
             ) {
                 $this->removeDefaultArgument($tokens, $nextRelevantIndex);
             }
@@ -81,25 +83,27 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
      * @param Tokens $tokens
      * @param int    $index
      *
-     * @return bool
+     * @return int|null
      */
-    private function hasNonDefaultArgumentAfterIndex(Tokens $tokens, $index)
+    private function getLastNonDefaultArgumentIndex(Tokens $tokens, $index)
     {
         $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $index, $this->functionDefinitionTerminatorTokens);
 
         if (null === $nextRelevantTokenIndex) {
-            return false;
+            return;
         }
 
-        while (!$tokens[$nextRelevantTokenIndex]->equalsAny($this->functionDefinitionTerminatorTokens)) {
+        $lastNonDefaultArgumentIndex = null;
+
+        while ($tokens[$nextRelevantTokenIndex]->isGivenKind(T_VARIABLE)) {
             if ($tokens[$tokens->getNextMeaningfulToken($nextRelevantTokenIndex)]->equalsAny($this->argumentTerminatorTokens)) {
-                return true;
+                $lastNonDefaultArgumentIndex = $nextRelevantTokenIndex;
             }
 
             $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $nextRelevantTokenIndex, $this->functionDefinitionTerminatorTokens);
         }
 
-        return false;
+        return $lastNonDefaultArgumentIndex;
     }
 
     /**
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -117,7 +117,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
     private function isEllipsis(Tokens $tokens, $variableIndex)
     {
         if (PHP_VERSION_ID < 50600) {
-            return false;
+            return $tokens[$tokens->getPrevMeaningfulToken($variableIndex)]->equals('.');
         }
 
         return $tokens[$tokens->getPrevMeaningfulToken($variableIndex)]->isGivenKind(T_ELLIPSIS);
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -222,6 +222,7 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
 
         while ($tokens[$currentIndex]->isGivenKind(T_WHITESPACE)) {
             $tokens[$currentIndex]->clear();
+            --$currentIndex;
         }
     }
 
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -19,7 +19,7 @@ use Symfony\CS\Tokenizer\Tokens;
  * @author Mark Scherer
  * @author Lucas Manzke <lmanzke@outlook.com>
  */
-class MethodArgumentDefaultValueFixer extends AbstractFixer
+final class MethodArgumentDefaultValueFixer extends AbstractFixer
 {
     private $argumentBoundaryTokens = array('(', ',', ')', ';', '{', '}');
     private $functionDefinitionTerminatorTokens = array(')', ';', '{', '}');
@@ -38,13 +38,13 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
     /**
      * {@inheritdoc}
      */
-    public function fix(\SplFileInfo $file = null, $content)
+    public function fix(\SplFileInfo $file, $content)
     {
         $tokens = Tokens::fromCode($content);
-
-        $functionTokenIndices = array_keys($tokens->findGivenKind(T_FUNCTION));
-        foreach ($functionTokenIndices as $index) {
-            $this->fixFunctionDefinition($tokens, $index);
+        for ($i = 0, $l = $tokens->count(); $i < $l; ++$i) {
+            if ($tokens[$i]->isGivenKind(T_FUNCTION)) {
+                $this->fixFunctionDefinition($tokens, $i);
+            }
         }
 
         return $tokens->generateCode();
@@ -91,10 +91,8 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
             return false;
         }
 
-        while (!in_array($tokens[$nextRelevantTokenIndex]->getContent(), $this->functionDefinitionTerminatorTokens, true)) {
-            $nextRelevantTokenContent = $tokens[$tokens->getNextMeaningfulToken($nextRelevantTokenIndex)]->getContent();
-
-            if (in_array($nextRelevantTokenContent, $this->argumentTerminatorTokens, true)) {
+        while (!$tokens[$nextRelevantTokenIndex]->equalsAny($this->functionDefinitionTerminatorTokens)) {
+            if ($tokens[$tokens->getNextMeaningfulToken($nextRelevantTokenIndex)]->equalsAny($this->argumentTerminatorTokens)) {
                 return true;
             }
 
@@ -155,7 +153,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
         $nextTokenIndex = $tokens->getNextTokenOfKind($index, $this->defaultValueTokens);
         $nextToken = $tokens[$nextTokenIndex];
 
-        return $nextToken->getContent() === '=';
+        return $nextToken->equals('=');
     }
 
     /**
@@ -169,7 +167,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
         $nextTokenIndex = $tokens->getNextTokenOfKind($index, $this->immediateDefaultValueTokens);
         $nextToken = $tokens[$nextTokenIndex];
 
-        return $nextToken->getContent() === '=';
+        return $nextToken->equals('=');
     }
 
     /**
@@ -180,7 +178,7 @@ class MethodArgumentDefaultValueFixer extends AbstractFixer
     {
         $currentIndex = $nextVariableIndex;
 
-        while (!in_array($tokens[$currentIndex + 1]->getContent(), $this->argumentTerminatorTokens, true)) {
+        while (!$tokens[$currentIndex + 1]->equalsAny($this->argumentTerminatorTokens)) {
             $tokens[++$currentIndex]->clear();
         }
     }
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -1,0 +1,195 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\FixerInterface;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Mark Scherer
+ * @author Lucas Manzke <lmanzke@outlook.com>
+ */
+class MethodArgumentDefaultValueFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'In method arguments there must not be arguments with default values before non-default ones.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file = null, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $functionTokenIndices = array_keys($tokens->findGivenKind(T_FUNCTION));
+        foreach ($functionTokenIndices as $index) {
+            $this->fixFunctionDefinition($tokens, $index);
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     */
+    private function fixFunctionDefinition(Tokens $tokens, $index)
+    {
+        $examinedIndex = $tokens->getNextTokenOfKind($index, array('(', ',', ')'));
+        while (
+            $this->hasNonDefaultArgumentAfterIndex($tokens, $examinedIndex) &&
+            $this->hasDefaultValueAfterIndex($tokens, $examinedIndex)
+        ) {
+            $nextRelevantIndex = $this->findNextVariableOrTokenOfKind($tokens, $examinedIndex, array(')'));
+
+            if (')' === $tokens[$nextRelevantIndex]->getContent()) {
+                break;
+            }
+
+            if (
+                $this->isDefaultArgumentAfterIndex($tokens, $nextRelevantIndex - 1) &&
+                $this->hasNonDefaultArgumentAfterIndex($tokens, $nextRelevantIndex - 1)
+            ) {
+                $this->removeDefaultArgument($tokens, $nextRelevantIndex);
+            }
+            $examinedIndex = $nextRelevantIndex;
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    private function hasNonDefaultArgumentAfterIndex(Tokens $tokens, $index)
+    {
+        $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $index, array(')'));
+
+        if (null === $nextRelevantTokenIndex) {
+            return false;
+        }
+
+        while (')' !== $tokens[$nextRelevantTokenIndex]->getContent()) {
+            $nextRelevantTokenContent = $tokens[$tokens->getNextMeaningfulToken($nextRelevantTokenIndex)]->getContent();
+
+            if (in_array($nextRelevantTokenContent, array(',', ')'), true)) {
+                return true;
+            }
+
+            $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $nextRelevantTokenIndex, array(')'));
+        }
+
+        return false;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int $index
+     * @param array $relevantTokens
+     * @return int|null
+     */
+    private function findNextVariableOrTokenOfKind(Tokens $tokens, $index, array $relevantTokens)
+    {
+        $variableIndices = array_keys($tokens->findGivenKind(T_VARIABLE));
+
+        $nextVariableIndex = $this->getFirstValueBiggerThan($variableIndices, $index);
+        $nextRelevantTokenIndex = $tokens->getNextTokenOfKind($index, $relevantTokens);
+
+        if (null === $nextVariableIndex) {
+            return $nextRelevantTokenIndex;
+        }
+
+        if (null === $nextRelevantTokenIndex) {
+            return $nextVariableIndex;
+        }
+
+        return ($nextVariableIndex < $nextRelevantTokenIndex) ? $nextVariableIndex : $nextRelevantTokenIndex;
+    }
+
+    /**
+     * @param array $values
+     * @param int $minimumValue
+     *
+     * @return int|null
+     */
+    private function getFirstValueBiggerThan(array $values, $minimumValue)
+    {
+        foreach ($values as $value) {
+            if ($value > $minimumValue) {
+                return $value;
+            }
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    private function hasDefaultValueAfterIndex(Tokens $tokens, $index)
+    {
+        $nextTokenIndex = $tokens->getNextTokenOfKind($index, array('=', ')'));
+        $nextToken = $tokens[$nextTokenIndex];
+
+        return $nextToken->getContent() !== ')';
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int $index
+     * @return bool
+     */
+    private function isDefaultArgumentAfterIndex(Tokens $tokens, $index)
+    {
+        $nextTokenIndex = $tokens->getNextTokenOfKind($index, array('=', ')', ','));
+        $nextToken = $tokens[$nextTokenIndex];
+
+        return $nextToken->getContent() === '=';
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int $nextVariableIndex
+     */
+    private function removeDefaultArgument(Tokens $tokens, $nextVariableIndex)
+    {
+        $currentIndex = $nextVariableIndex;
+
+        while (!in_array($tokens[$currentIndex + 1]->getContent(), array(',', ')'), true)) {
+            $tokens[++$currentIndex]->clear();
+        }
+    }
+
+    /**
+     * @return int
+     */
+    public function getPriority()
+    {
+        return -49;
+    }
+
+    /**
+     * @return int
+     */
+    public function getLevel()
+    {
+        return FixerInterface::CONTRIB_LEVEL;
+    }
+}
--- a/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
@@ -142,6 +142,16 @@ final class MethodArgumentDefaultValueFixer extends AbstractFixer
     {
         $currentIndex = $nextVariableIndex;
 
+        $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($nextVariableIndex);
+        $nextMeaningfulTokenIndex = $tokens->getNextTokenOfKind($nextVariableIndex, array(array(310, 'null'), ',', ')'));
+
+        if (
+            $tokens[$nextMeaningfulTokenIndex]->equals(array(310, 'null')) &&
+            !$tokens[$prevMeaningfulTokenIndex]->equalsAny($this->argumentBoundaryTokens)
+        ) {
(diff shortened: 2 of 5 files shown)


Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+     */
+    public function getDescription()
+    {
+        return 'In method arguments there must not be arguments with default values before non-default ones.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file = null, $content)
```
dereuromark: `= null should be removed` - or you can run the fixer over itself. :) 
This should also be done for the other fixers in this repo if they contain this line.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * @dataProvider provideExamples
+     *
+     * @param string $expected
+     * @param string $input
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
```
dereuromark: newline missing above

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+class MethodArgumentDefaultValueFixerTest extends AbstractFixerTestBase
```
keradus: final

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+            $tokens[++$currentIndex]->clear();
+        }
+    }
+
+    /**
+     * @return int
+     */
+    public function getPriority()
+    {
+        return -49;
```
keradus: why is that?

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\FixerInterface;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Mark Scherer
+ * @author Lucas Manzke <lmanzke@outlook.com>
+ */
+class MethodArgumentDefaultValueFixer extends AbstractFixer
```
keradus: final

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file = null, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $functionTokenIndices = array_keys($tokens->findGivenKind(T_FUNCTION));
```
keradus: don't use `findGivenKind` for that purpose.
That way you are iterating here twice.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
dereuromark: why?

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+{
+    /**
+     * @dataProvider provideExamples
+     *
+     * @param string $expected
+     * @param string $input
```
keradus: or null

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+
+        $expected1 = <<<'EOT'
+                    <?php
+                        public function eFunction($foo, $bar, \SplFileInfo $baz, $x = 'default';
+
+                        public function fFunction($foo, $bar, \SplFileInfo $baz, $x = 'default');
+EOT;
+
+        return array(
+            array('<?php public function bFunction($foo, $bar)', '<?php public function bFunction($foo = null, $bar)'),
```
keradus: please, write the cases like
```php
array(
    '<?php expected...',
    '<?php input...',
),
```

then it's more readable to see the difference

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+     * @param int    $index
+     *
+     * @return bool
+     */
+    private function hasDefaultValueAfterIndex(Tokens $tokens, $index)
+    {
+        $nextTokenIndex = $tokens->getNextTokenOfKind($index, $this->defaultValueTokens);
+        $nextToken = $tokens[$nextTokenIndex];
+
+        return $nextToken->getContent() === '=';
```
keradus: 1. yoda with you should be
2. don't use getContent to check if the token is equals to sth. Use `equals` method

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+            array('<?php public function cFunction($foo, $bar, $baz)', '<?php public function cFunction($foo = false, $bar = \'bar\', $baz)'),
+            array('<?php public function dFunction($foo, $bar, $baz)', '<?php public function dFunction($foo = false, $bar, $baz)'),
+            array('<?php public function eFunction($foo, $bar, \SplFileInfo $baz, $x)', '<?php public function eFunction($foo = PHP_EOL, $bar, \SplFileInfo $baz = null, $x)'),
+            array('<?php public function eFunction($foo, $bar, \SplFileInfo $baz, $x = \'default\')', '<?php public function eFunction($foo, $bar = \'removedDefault\', \SplFileInfo $baz, $x = \'default\')'),
+            array($expected1, $input1),
```
keradus: just push them directly to array without that variables

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+EOT;
+
+        $expected1 = <<<'EOT'
+                    <?php
+                        public function eFunction($foo, $bar, \SplFileInfo $baz, $x = 'default';
+
+                        public function fFunction($foo, $bar, \SplFileInfo $baz, $x = 'default');
+EOT;
+
+        return array(
```
keradus: missing a case like
```php
array(
    '<?php function foo($bar, $c) {}',
    '<?php function foo($bar /* a */ = /* b */ 1, $c) {}',
),
```

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
dereuromark: You really should stop using Yoda, its a very bad practice - and instead use a sniffer/fixer to assert inline assignment is not possible (as we do).

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->makeTest($expected, $input);
+    }
+    /**
+     * @return array
+     */
+    public function provideExamples()
+    {
+        $input1 = <<<'EOT'
+                    <?php
+                        public function eFunction($foo, $bar, \SplFileInfo $baz, $x = 'default';
```
keradus: All of the cases are invalid PHP code!

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
dereuromark: Doesnt make it even harder to read?
I find your current practice in the project of doing this inline strings quite unreadable.
It would be better to more often use elaborate <<< examples.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: All fixers are final at 2.x.
We cannot make them final on 1.x to not introduce BC breaks, but all new fixers should be final.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
lmanzke: True ;)

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: The code should be consistent. If whole project use it - it should be used here too.

Also, I don't think we have assignment in such a places anywhere.

dereuromark: You can easily run a sniffer/fixer over it. I have written one. This way it is consistent then.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
keradus: When you have tons of different edge cases it's good to have many small cases instead few big ones.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
lmanzke: Ok, will add one, but probably tomorrow. Am curious to see if it would work on that, don't see a problem yet though.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
dereuromark: That makes it impossible for them to be extended... :/

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
keradus: using heredoc/nowdoc is OK, but here catch them into variables are pointless.
that is why the names of them aren't great

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: And it is consistent.
Don't want to use yoda? it's cool. want to use yoda? it's cool. but don't mix.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
keradus: I think it will pass, but it's great to have that test to protect the code to being corrupted during refactoring in the future

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
dereuromark: Yes, thats why I said you should stop doing it and fixing it (using an automated fixer/sniffer - at once).
Doesn't have to be today or tomorrow. But maybe in the near future.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $nextVariableIndex
+     */
+    private function removeDefaultArgument(Tokens $tokens, $nextVariableIndex)
+    {
+        $currentIndex = $nextVariableIndex;
+
+        while (!in_array($tokens[$currentIndex + 1]->getContent(), $this->argumentTerminatorTokens, true)) {
```
keradus: `equalsAny`

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,203 @@
...
+     */
+    private function hasNonDefaultArgumentAfterIndex(Tokens $tokens, $index)
+    {
+        $nextRelevantTokenIndex = $this->findNextVariableOrTokenOfKind($tokens, $index, $this->functionDefinitionTerminatorTokens);
+
+        if (null === $nextRelevantTokenIndex) {
+            return false;
+        }
+
+        while (!in_array($tokens[$nextRelevantTokenIndex]->getContent(), $this->functionDefinitionTerminatorTokens, true)) {
```
keradus: here and below - `equalsAny`

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: So far we do follow yoda condition and the PR should follow that too.
If you want to propose abandoning it - feel free to send separate PR for only that.

dereuromark: I first have to publish the fixer for it, though.
Actually both NoYoda and NoInlineAssignment including sniffs.
So far they are still in development phase and not quite public-ready :)

lmanzke: Well, it won't matter because I'll use equals anyway there. There won't be a comparison.

Is there a shortcut to check if the token is one of an array? Equals won't right cut it here.

lmanzke: Saw you already suggest it ;) equalsAny is what I was looking for

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
lmanzke: Did not know what number to put there. Any suggestions?

dereuromark:
Nice job!
I wonder if we should allso add an example with too less whitespace?

    public function hFunction($foo,$bar='removedValue',\SplFileInfo $baz,$x) {};

It should also remove the default value here properly without touching the rest of the whitespace.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * @return array
+     */
+    public function provideExamples()
+    {
+        return array(
+            array(
+                '<?php public function bFunction($foo, $bar) {}',
+                '<?php public function bFunction($foo = null, $bar) {}',
```
gharlan: please use valid php code

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
gharlan: If you do not know, try `0` (remove the method).

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     */
+    private function fixFunctionDefinition(Tokens $tokens, $index)
+    {
+        $examinedIndex = $tokens->getNextTokenOfKind($index, $this->argumentBoundaryTokens);
+        while (
+            $this->hasNonDefaultArgumentAfterIndex($tokens, $examinedIndex) &&
```
gharlan: Couldn't you determine the last non-default argument index once per function definition?

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: you should think if the fixer you are creating need to run after or before any other existing fixer ?

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,193 @@
...
+    {
+        $currentIndex = $nextVariableIndex;
+
+        while (!$tokens[$currentIndex + 1]->equalsAny($this->argumentTerminatorTokens)) {
+            $tokens[++$currentIndex]->clear();
+        }
+    }
+
+    /**
+     * @return int
```
keradus: inheritdoc

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
lmanzke: Yes, that would probably be an improvement. I'll change it a bit

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,80 @@
...
+                <<<'EOT'
+                    <?php
+                        function eFunction($foo, $bar, \SplFileInfo $baz, $x = 'default' {};
+
+                        function fFunction($foo, $bar, \SplFileInfo $baz, $x = 'default') {};
+EOT
+                ,
+                <<<'EOT'
+                    <?php
+                        function eFunction($foo, $bar, \SplFileInfo $baz, $x = 'default' {};
```
gharlan: missing `)`

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,193 @@
...
+     * @param int    $index
+     * @param array  $relevantTokens
+     *
+     * @return int|null
+     */
+    private function findNextVariableOrTokenOfKind(Tokens $tokens, $index, array $relevantTokens)
+    {
+        $variableIndices = array_keys($tokens->findGivenKind(T_VARIABLE));
+
+        $nextVariableIndex = $this->getFirstValueBiggerThan($variableIndices, $index);
```
keradus: `getNextTokenOfKind` should be faster than looking for all `T_VARIABLES` and getting first one.

lmanzke: getNextTokenOfKind does not work for T_VARIABLE

lmanzke: Its probably easier though just to iterate through the tokens until the first variable is encountered.

keradus: Try this out:
```php
$test = $tokens->getNextTokenOfKind($index, [[T_VARIABLE]]);
```

second argument is an array of tokens you are looking for.
you are looking for token like `[T_VARIABLE, $foo, 123]`, but don't care about name or line, so simplify it into `[T_VARIABLE]`. And then you create an array with your token: `[[T_VARIABLE]]`

lmanzke: Ok, didn't know this. I tried it with getNextTokenOfKind first and couldn't get it to work.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
lmanzke: This is actually intended, to check if in such an edge-case it does not do something strange.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
lmanzke: Ok, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/b0f8a6aaecdea9f98504eeb0ec4f60afe7bde0ad . Got rid of the function to find the next variable or specific token all along.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
gharlan: we do not fix invalid files, so we do not have to test them.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: Great!
TBH I just don't want to have such a nice `getFirstValueBiggerThan` method :D

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,157 @@
...
+            if ($tokens[$i]->isGivenKind(T_FUNCTION)) {
+                $this->fixFunctionDefinition($tokens, $i);
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * @param Tokens $tokens
```
dereuromark: If you do `@param Tokens|Token[] $tokens` you get typehinting on all `$tokens[$i]` etc.

lmanzke: I get those with the present version, too. Don't you?

dereuromark: Are you sure? Check the isGivenKind (is clickable?) in `$tokens[$i]->isGivenKind()` for example.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,157 @@
...
+        return 'In method arguments there must not be arguments with default values before non-default ones.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($i = 0, $l = $tokens->count(); $i < $l; ++$i) {
```
dereuromark: Would need a protected/private method wrapping to have typehinting on the $tokens[$i]->...
But no big deal I guess.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: I bet you are talking here about feature of PHPStorm, but looking on pure PHPDoc that notation would be wrong, because you cant pass `Token[]` to the method!

Typehinting of `$tokens[$nextRelevantIndex]` should be done thanks to https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tokenizer/Tokens.php#L23

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
lmanzke: If you don't have it this is most likely a bug on your end, because my PHPStorm understands it the way it is :smile: . Meaning isGivenKind() can be autocompleted.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
dereuromark: Didnt know that trick. Thx.

stof:
This fixer must not remove a ``null`` default value from a typehinted argument, even though it is followed by a required argument, as such default is about making the argument nullable, even though it cannot make it optional.
This case must be covered in the testsuite to ensure you don't break the code.

@keradus you should change the required status from appveyor as they splitted the status for pushes and PRs (as done by travis since a long time).

lmanzke:
@stof: I do not really see the difference between making an argument nullable and making it optional. Isn't a sign of poor design if nullable arguments have to be specified because they appear in the middle of the function definition?

I see your point, the fixer should not force you to refactor the code, but on the other side, this kind of refactoring should be done anyway.

lmanzke:
By the way, a general question on the tests: Why is it that expected and input must not be the same? Or is there a flag I have specify to point out that this behavior is intended?

I'd quite like to write some test-cases to test specifically that some constructions are **not** touched. However, that is disallowed, presumably as a precaution against tests that do not test anything by accident.

keradus:
```php
function foo(Bar $b = null, $c) {}

foo(new Bar(), 1); //OK
foo(null, 1); //OK too
```

if changed:
```php
function foo(Bar $b, $c) {}

foo(new Bar(), 1); //OK
foo(null, 1); //error
```

and that change the logic.

:+1: for @stof ;)

dereuromark:
@stof @keradus PHP makes all object attributes nullable either way. As such your examples are invalid as that is a default value, nothing to do with nullable.
To say that it can be nullable you should use the doc block `Object|null` instead.

keradus:
@lmanzke
take a look https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1495#issuecomment-155422233
also, going into `makeTest` method should also helps

dereuromark:
To clarify:

    function foo(Bar $b, $c) {}

Its perfectly valid to call `$this->foo(null, $c);` check it yourself

keradus:
@dereuromark 

```bash
ker@ad:/cli$ php --version
PHP 5.5.9-1ubuntu4.14 (cli) (built: Oct 28 2015 01:34:46) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
    with Xdebug v2.2.3, Copyright (c) 2002-2013, by Derick Rethans
ker@ad:/cli$ 
ker@ad:/cli$ cat test.php 
<?php

class Bar{}
function fnc(Bar $b, $c) {}

fnc(null, 1);
ker@ad:/cli$ 
ker@ad:/cli$ php -f test.php 
PHP Catchable fatal error:  Argument 1 passed to fnc() must be an instance of Bar, null given, called in /cli/test.php on line 6 and defined in /cli/test.php on line 4
PHP Stack trace:
PHP   1. {main}() cli/test.php:0
PHP   2. fnc() /cli/test.php:6
ker@ad:/cli$ 
```

lmanzke:
@keradus Thanks for clearing that up. However, "go into the method" is a somewhat invalid argument. That should clearly be something immediately obvious from the docblock, which is not even present on the AbstractFixerTestBase->makeTest(). Maybe adding that would avoid confusion in the feature. (Also, having inheritdoc and then there is no parent docblock seems strange)

keradus:
> Maybe adding that would avoid confusion in the feature. 

great catch, feel free to attach it!

dereuromark:
Wooh, ok. I could have sworn it was different in some cases.

lmanzke:
I added the makeTest docblock in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1496

keradus:
ping ;)

lmanzke:
I added the case described by @stof, but also encountered something odd:
If I use getNextTokenOfKind('null'), it does not find it, as I half expected. Null seems to have the token id 310, which is, according to tokenizer.php, T_VARIABLE.

At runtime, however, T_VARIABLE is 312. Did I look up the wrong definition?
If yes, where do I find the right one? Don't like using "310" as magic token id at all.

gharlan:
The IDs are different in different php versions.

https://3v4l.org/psqa6

`null` is a `T_STRING` token.

gharlan:
> All tokens listed below are also defined as PHP constants. Their value is automatically generated based on PHP's underlying parser infrastructure. This means that the concrete value of a token may change between two PHP versions.

http://php.net/manual/en/tokens.php

lmanzke:
Good to know ;). The function token_name is also a good catch ;). Will change that then.
Still interesting that if it's a simple string that $token->equals('null') will not work.

keradus:
Don't rely on int value itself.

also, `getNextTokenOfKind` get the argument as array of possible tokens, not single token

lmanzke:
@keradus: I know  that, just was a bit lazy. Of course what I meant was getNextTokenOfKind(array('null')) won't work, because the underlying usage of equals will not catch it.

keradus:
```php
<?php

$a = null;

```

```php
$firstNull = $tokens->getNextTokenOfKind(0, [[T_STRING, 'null']]);
var_dump($firstNull, $tokens[$firstNull]->toJson());
```

```bash
int(6)                 
string(100) "{         
    "id": 310,         
    "name": "T_STRING",
    "content": "null", 
    "line": 3,         
    "isArray": true    
}"                     
```

lmanzke:
@keradus I start getting the impression that you do not read my comments / commits. I already adapted this. 
Still, even if your code works, it shows that the tokenizer is bad and unreliable. The getNextTokenOfKind function works for 'function', why not for 'null'? Why is it necessary to use the prototype array construction there?

Anyway, are there any things I should still adapt?

keradus:
@lmanzke you said that this looking for null won't work, I show it works. Sorry, I don't saw your comment that you have already figured it out.
One way or another - great that you manage it.

> Still, even if your code works, it shows that the tokenizer is bad and unreliable. The getNextTokenOfKind function works for 'function', why not for 'null'? Why is it necessary to use the prototype array construction there?

Because there is `T_FUNCTION` and there is not `T_NULL`. Please, send thanks to PHP authors. 


> Anyway, are there any things I should still adapt?

Is the PR request ready from your side? If yes we will start a review ;)

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#issuecomment-155431991 is still to be done

dereuromark:
The test cases clearly show that this is not the case, that has been done already: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493/files#diff-5c2c894b7930f5a05b2dd0248e6c8d5bR82

keradus:
And still you have https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493/files#diff-5c2c894b7930f5a05b2dd0248e6c8d5bR37 .
You cannot assume that if a parameter is not typehinted as object it is not an object. It's safer to not delete `null` as default value.

If it will be removed for not-typehinted parameters then the fixer will be considered as risky (may change meaning of code!)

dereuromark:
> considered as risky

Maybe not a bad idea after all. In its original form it this fixer could be used to detect possible code smells early on.

But we could also fix it up to be non-risky and be done with it :)

keradus:
> In its original form it this fixer could be used to detect possible code smells early on.

Please remember that this is not mess detector but fixer.

lmanzke:
> If it will be removed for not-typehinted parameters then the fixer will be considered as risky (may 
> change meaning of code!)

I do not think that is true. That is the whole purpose if this fixer: If there is a `$a = null` somewhere in the function definition, the null does not make sense. It can be removed without changing the meaning of the code. The param is also nullable without the default value, but will always have to be passed if there are required arguments after the = null.

It is another case for typehinted arguments, though: Omitting the = null there would remove nullability and therefore change the behavior. So in this special case, if the argument is typehinted and default value is null, then it would change behavior, which is why it should leave it untouched.

However, it has to said that if something like this occurs, it is a code smell and should probably be refactored. But as this is not the purpose of the fixer, it leaves it as it is.

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,87 @@
...
+
+                        function fFunction($foo, $bar = 'removedValue', \SplFileInfo $baz, $x = 'default') {};
+EOT
+            ),
+            array(
+                '<?php function foo ($bar, $c) {}',
+                '<?php function foo ($bar /* a */ = /* b */ 1, $c) {}',
```
SpacePossum: Why are the comments removed?
What will this fixer do with something like:
```php
function foo(
  $a = 1, // test
  $b = 2, /* test*/
  $c = null, // abc
  $d 
) {}

**Comment by keradus** [comment]
@lmanzke , you are right, my bad.

File: Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php (hunk 22 above)
Comment (dereuromark) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: I agree, comments or non-related whitespace should not be removed (responsibility of other fixers if any).

Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: Some spaces should be removed, like:
`function foo($a =   1, $b) {}`
into:
`function foo($a, $b) {}`
is OK

`function foo($a    , $b) {}`
would be odd

Comment (dereuromark) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: This is not a whitespace or argument spacing fixer though.

Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: If it remove piece of could it should take care not only about one token, but about place itself.

Comment (dereuromark) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: When removing yes (every whitespace in between without comments). Otherwise it should skip.

Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: exactly, as in my example ;)

Comment (dereuromark) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: Indeed, but @SpacePossum s example comments would not be touched by this, as there the extra stuff is after the value (and not between variable and value).

Comment (lmanzke) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: I'll add test cases just to make sure.

Comment (SpacePossum) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: ah you are right :)  thank for making sure

**Comment by lmanzke** [comment]
Testcase added and rebased it on top of current 1.11, too. I can also squash it a bit, but would do that at the end.

**Comment by keradus** [comment]
(no need to squash yourself)

File: Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php (hunk 22 above)
Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: I still do not like that the comments after variable are removed.
IMO both `/* a */` and `/* b */` comments should be there after removing default value

Comment (gharlan) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: @keradus I agree

Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: in fact it looks like it is the last thing to do and good to merge then

Comment (gharlan) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: imo this should be fixed: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#discussion_r44389803

Comment (keradus) [inline Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php]: yeah, good point

**Comment by keradus** [comment]
@lmanzke it looks almost ready! Please, could you finish it off? I would be more than happy to merge it into 1.11

**Comment by lmanzke** [comment]
So the open things are the comments that should not be removed and the "wrong" test case?

**Comment by dereuromark** [comment]
@lmanzke Looks like it. Adding the closing braces sure makes sense as we only want to work with valid PHP anyway.

**Comment by keradus** [comment]
exactly ;)

File: Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php [hunk 23]
```diff
@@ -0,0 +1,167 @@
...
+    private function removeDefaultArgument(Tokens $tokens, $nextVariableIndex)
+    {
+        $currentIndex = $nextVariableIndex;
+
+        $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($nextVariableIndex);
+        $nextMeaningfulTokenIndex = $tokens->getNextTokenOfKind($nextVariableIndex, array(array(T_STRING, 'null'), ',', ')'));
+
+        if (
+            $tokens[$nextMeaningfulTokenIndex]->equals(array(T_STRING, 'null')) &&
+            !$tokens[$prevMeaningfulTokenIndex]->equalsAny($this->argumentBoundaryTokens)
```
gharlan: Please add a test case with a reference parameter:

```php
function foo(&$a = null, $b) {}
```

I'm not sure but I think it will fail because of the condition above.

gharlan: And please add a test case with uppercased `NULL`. I'm not sure if it will pass.

```php
function foo(\SplFileInfo $file = NULL, $foo) {}
```
```

File: Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php [hunk 24]
```diff
@@ -0,0 +1,167 @@
...
+        ) {
+            return; //ignore typehinted null default values
+        }
+
+        while (!$tokens[$currentIndex + 1]->equalsAny($this->argumentTerminatorTokens)) {
+            $tokens[++$currentIndex]->clear();
```
gharlan: Please add this test case:

```php
function foo($foo = array(array(1)), $bar) {}
```

Should fail because `$this->argumentTerminatorTokens` contains `)`.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
keradus: And `Null` ;D

lmanzke:
@gharlan Added your test case and have adapted it to make it pass.
Currently having another one that is causing trouble though:

`public function a($a = array('a' => 'b', 'c' => 'd'), $b) {}`

lmanzke:
Wow, strange. Can somebody reproduce that this test case actually **breaks** the tokenizer?
`public function a($a = array('a' => 'b', 'c' => 'd'), $b) {}`

Tokens::fromCode returns one token for that string which is quite odd

lmanzke:
Seems to be a problem in the token_get_all function used by Tokens::fromCode.
My PHP Version: 
`PHP 5.6.15 (cli) (built: Nov 13 2015 14:32:10)
Copyright (c) 1997-2015 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2015 Zend Technologies
    with Xdebug v2.3.3, Copyright (c) 2002-2015, by Derick Rethans`

lmanzke:
Ah see the error. Good old <?php is missing in front. Never mind ...

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
GrahamCampbell: or, in otherwords, case insensitive :P

lmanzke:
Ok, I now added all test cases mentioned here and refactored the fixer to make them work. If there are any other test cases anybody can think of :wink:

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,136 @@
...
+                        ) {}
+EOT
+            ),
+            array(
+                '<?php function foo($foo, $bar) {}',
+                '<?php function foo($foo = array(array(1)), $bar) {}',
+            ),
+            array(
+                '<?php public function a($a, $b) {}',
+                '<?php public function a($a = array(\'a\' => \'b\', \'c\' => \'d\'), $b) {}',
```
gharlan: invalid code (remove `public` or add a class around).

gharlan:
Missing this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#discussion_r45791923
(typehinted nullable argument with non-lowercase `NULL` (or `Null`).

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,136 @@
...
+                '<?php function foo($foo, $bar) {}',
+                '<?php function foo($foo = array(array(1)), $bar) {}',
+            ),
+            array(
+                '<?php function a($a, $b) {}',
+                '<?php function a($a = array(\'a\' => \'b\', \'c\' => \'d\'), $b) {}',
+            ),
+            array(
+                '<?php public function a($a, $b) {}',
+                '<?php public function a($a = [\'a\' => \'b\', \'c\' => \'d\'], $b) {}',
```
gharlan: invalid code

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,136 @@
...
+            ),
+            array(
+                '<?php public function a($a, $b) {}',
+                '<?php public function a($a = NULL, $b) {}',
```
gharlan: invalid code

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,136 @@
...
+            ),
+            array(
+                '<?php public function a($a, $b) {}',
+                '<?php public function a($a = Null, $b) {}',
```
gharlan: invalid code

Symfony/CS/Tests/Fixer/Contrib/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -0,0 +1,136 @@
...
+            ),
+            array(
+                '<?php public function a(&$a, $b) {}',
+                '<?php public function a(&$a = null, $b) {}',
```
gharlan: invalid code

dereuromark: I am pretty sure it was absolutely unnecessary to outline the same thing 5 times.

gharlan: My comment: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#discussion_r45812628

And the commit after that: https://github.com/lmanzke/PHP-CS-Fixer/commit/bcaee36cdd840ed16a89255bcbe389ac261cc05e

So I'm not sure. ;)

gharlan:
Oh, found another special case:

```php
function foo($a = 1, ...$b) {}
```

This is valid since PHP 5.6 and the default value must not be removed.
https://3v4l.org/EYN8Q

lmanzke:
Added a case for that and hope that all tests contain valid PHP code now.

lmanzke:
Ok, appveyor does not seem to know mb_convert_case (strange, exists since 4.3) and the T_ELLIPSIS constant is only available since 5.6 (makes sense). Will try to make sense of that tomorrow.

keradus:
As `mbstring` extension is not available by default in all envs and we * do not * rely on it.
Actually, we work to remove `ext-mbstring` from requirements in `compsoer.json` ;)

lmanzke:
Ok, tests pass again now as I put the ellipsis test cases into a separate method.

gharlan:
:+1: Thanks, very nice fixer!

lmanzke:
Seems I'm not quite there yet: 

`function a(\SplFileInfo &$a = null) {}` 

is valid php code (although senseless) and will probably fail. Verifying ... 

Would never have believed there are so many special cases! :wink:

lmanzke:
Actually, it passes. Added it to make sure though. 
Good to go.

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,236 @@
...
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     */
+    private function clearWhitespacesBeforeIndex(Tokens $tokens, $index)
+    {
+        $currentIndex = $index - 1;
+
+        while ($tokens[$currentIndex]->isGivenKind(T_WHITESPACE)) {
```
keradus: it is not standard to have few `T_WHITESPACE` tokens one after another. document why it is possible here

lmanzke: If that is the case, then the while can be removed, as it will only have one iteration ;)

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $variableIndex
+     *
+     * @return bool
+     */
+    private function isEllipsis(Tokens $tokens, $variableIndex)
+    {
+        if (PHP_VERSION_ID < 50600) {
+            return false;
```
keradus: not sure about it...
if one run the fixer under 5.5 it doesn't mean the code that is fixing isn't written in 5.6

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,236 @@
...
+     */
+    private function isTypehintedNullableVariable(Tokens $tokens, $variableIndex)
+    {
+        $prevMeaningfulTokenIndex = $tokens->getPrevTokenOfKind($variableIndex, array(array(T_STRING), ',', '('));
+        $nextMeaningfulTokenIndex = $tokens->getNextTokenOfKind($variableIndex, array(array(T_STRING), ',', ')'));
+
+        $lowerCasedNextContent = strtolower($tokens[$nextMeaningfulTokenIndex]->getContent());
+
+        return
+            $lowerCasedNextContent === 'null' &&
```
keradus: yoda with you should be

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,236 @@
...
+            $tokens[$prevMeaningfulTokenIndex]->isGivenKind(T_STRING);
```
keradus: revert the condition. `T_STRING` could be check cheaply, then check that `null`

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
gharlan: It is not possible to fix 5.6 codeunder 5.5.
Code is invalid, so the file would be skipped (after linting).

So Imo the solution above is ok.

keradus: I remember we already doing it in few places.
But in that case, we should at least have a test for 5.6 which be fixed (we have it!) and a test for <5.6 with `...` which would not be changed.

gharlan: > a test for <5.6 with ... which would not be changed.

It would be an invalid test (#1538). IMO we should avoid this.
The fixer would not fix the file under 5.5.

keradus: And we should check if he skip it indeed.
Not all use the linting option.

lmanzke: I thought there should be no invalid php code? For versions < 5.6, that would be the case if anything like ...$b occurs in a function definition. A test containing that would break that convention. The thing is it that the fixer would change PHP 5.5 code.

`<?php function a($a = 1, ...$b) {}`

would become

`<?php function a($a, ...b) {}` 

because the fixer would not 'understand' the ellipsis.

keradus: Exactly. When I try to run the fixer on input code with ... using 5.5 interpreter, fixer should not remove the default value. We need to make prevention for that to happen. Because not all users use linting

lmanzke: Ok that means there are invalid test cases, but at least it runs under 5.5. (and below)

README.rst
```diff
@@ -561,6 +561,12 @@ Choose from the list of available fixers:
                         Arrays should use the long
                         syntax.
 
+* **method_argument_default_value** [contrib]
```
keradus: Please move to Symfony

keradus:
:+1:

SpacePossum:
:+1:  thanks @lmanzke :)
before moving to Symfony level, I would like to know if @stof (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#issuecomment-155431991) is happy with the implementation

keradus:
The fixer already follows his suggestion

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,237 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $variableIndex
+     *
+     * @return bool
+     */
+    private function isEllipsis(Tokens $tokens, $variableIndex)
+    {
+        if (PHP_VERSION_ID < 50600) {
+            return $tokens[$tokens->getPrevMeaningfulToken($variableIndex)]->equals('.');
```
gharlan: @keradus do we really want to go this direction? Fixing weird token sequences (php 7 code tokenized with php 5 etc.)?

IMO if we do it here, we have to do it in many many other places.

Example: `$a <=> $b`
Here we have the spaceship operator (`T_SPACESHIP`), but php 5 tokenizer returns `<=` and `>` tokens, so we have to binary operators.
Result when fixing with php 5: `$a <= > $b`, which is invalid.

SpacePossum:
I don't think so, this test shows a default value `null` being removed because it is followed by a required argument https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493/files#diff-5c2c894b7930f5a05b2dd0248e6c8d5bR36

keradus:
check out https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1493#issuecomment-158511368 and following 4 comments

Symfony/CS/Fixer/Symfony/MethodArgumentDefaultValueFixer.php
```diff
@@ -0,0 +1,237 @@
...
+     * @param Tokens $tokens
+     * @param int    $variableIndex
+     *
+     * @return bool
+     */
+    private function isTypehintedNullableVariable(Tokens $tokens, $variableIndex)
+    {
+        $prevMeaningfulTokenIndex = $tokens->getPrevTokenOfKind($variableIndex, array(array(T_STRING), ',', '('));
+
+        if (!$tokens[$prevMeaningfulTokenIndex]->isGivenKind(T_STRING)) {
```
gharlan: missing array and callable typehints.

keradus: ping @lmanzke

Symfony/CS/Fixer/Contrib/MethodArgumentDefaultValueFixer.php
gharlan: ping @keradus

keradus: I'm ok if the PHP7 code won't be broken when fixing using PHP5 (no need to be fixed, just not broken)

gharlan: `$a <=> $b` is "fixed" to `$a <= > $b` (using php5), which **is** broken.
I'm sure I could find other examples where the code is broken after fixing with wrong php version.

keradus: One *bug* isn't a reason to introduce another one.

gharlan: My opinion: linting should not be optional.
We should not try to handle php code of a newer version than the current running version.

But: accepted, we do not agree here. ;)

keradus: I would be more than happy to have always linter mode on!
But - the performance loss is horrible :(

lmanzke:
Didn't realize that PHP 5.3 doesn't know callable

GrahamCampbell:
callable is new in php 5.4, yeh

lmanzke:
Okay - callable is now also fixed in 5.3 , but without the constant

keradus:
Thank you @lmanzke.

SpacePossum:
thanks @lmanzke :)  :+1:

lmanzke:
np :)

Issue #1494, opened 2015-11-10
[WIP][PSR-12] Declare strict_types is incorrectly reformatted


EspadaV8:
I realise this is a little early since PHP 7 isn't out yet and PSR-12 is still in progress, however, I thought it might be good to start some issue since I couldn't find any on here already.

With the following code

```php
<?php
declare(strict_types=1);

class Test
{
}
```

php-cs-fixer will currently insert a space between `declare` and the opening `(` brace. According to the current PSR-12 document (https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#example) there shouldn't be a space between.

Updating the fixer while it's still in draft would potentially be a pointless ping-pong back and forth but it would be nice if there were some work around that could be used in the meantime.

keradus:
yeah, if it added a space than it's a bug

keradus:
@EspadaV8 could you work on a fix?

EspadaV8:
I can give it a go, although I've not worked with tokenisers before. I tracked it down to the `fixSpaceAroundToken` method in the `BracesFixer` class. Would the best way to be to add a check just for the `strict_types` line? Other instances of using `declare` should probably follow the usual formatting rules (or should all `declare`s not have a space?

Here we don't want any spaces

```php
<?php
declare(strict_types=1);

class Test
{
}
```

But this should be formatted similar to `if`s, `while`s, etc.

```php
declare(ticks=1) {
    // entire script here
}
```

```php
declare (ticks=1) {
    // entire script here
}
```

nesl247:
They've updated the PSR-12 doc to include the formatting requirement: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#declare-statements-namespace-and-use-declarations

It is always the same, no spaces.

keradus:
Thanks!

ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1495#issuecomment-159739190

EspadaV8:
Closing because of e797b2e7

PR #1495, opened 2015-11-10
BracesFixer - added support for declare

EspadaV8:
This is a RP for #1494  Not sure this is the best/correct way to do something like this.

Basically, `declare(strict_types=1);` shouldn't have a space between the `declare` and opening `(` but other `declare`s should be formatted like `if`s, `where`s, etc..

The strict types declare should only be on the second line too, is this something that needs to be checked as well?

### DIFF
--- a/Symfony/CS/Fixer/PSR2/BracesFixer.php
+++ b/Symfony/CS/Fixer/PSR2/BracesFixer.php
@@ -22,6 +22,8 @@ use Symfony\CS\Tokenizer\Tokens;
  */
 class BracesFixer extends AbstractFixer
 {
+    private $singleLineWhitespaceOptions = array('whitespaces' => " \t");
+
     /**
      * {@inheritdoc}
      */
@@ -369,7 +371,7 @@ class BracesFixer extends AbstractFixer
 
             // Declare tokens don't follow the same rules are other control statements
             if ($token->isGivenKind(T_DECLARE)) {
-                $tokens->removeTrailingWhitespace($index);
+                $this->fixDeclareStatement($tokens, $index);
             } elseif ($token->isGivenKind($controlTokens) || $token->isGivenKind(T_USE)) {
                 $nextNonWhitespaceIndex = $tokens->getNextNonWhitespace($index);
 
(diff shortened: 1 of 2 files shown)


keradus:
> The strict types declare should only be on the second line too, is this something that needs to be checked as well?

even if, it should be separate fixer then

keradus:
when we have a case:
`array($expected, $input)`
then under the hood it tests both `array($expected, $input)` and `array($expected)`

So you don't need to add `array($expected)` on your own ;)

EspadaV8:
@keradus Removed the duplicated tests

Symfony/CS/Fixer/PSR2/BracesFixer.php
```diff
@@ -605,4 +606,27 @@ private function getFinalControlContinuationTokensForOpeningToken($openingTokenK
 
         return array();
     }
+
+    private function fixDeclareStrictTypes(Tokens $tokens)
+    {
+        $declares = $tokens->findGivenKind(T_DECLARE);
+
+        foreach ($declares as $index => $declare) {
+            if ($this->isStrictTypeDeclare($index, $tokens)) {
```
keradus: should the CS be different if sb declare sth else?

keradus: ping @EspadaV8

EspadaV8: Hi @keradus, sorry, on a short holiday atm. From what I remember, just the declare strict types block is different and should be formatted like `declare(strict_types=1);`. Other declares could be used like other block elements and should be rendered like this,

```php
declare (ticks = 200) {
    //... 
} 
```

Perhaps any `declare` on a single line should be formatted without spacing? e. g. `declare(ticks=200);` in which case the check should really be for single/block declares.

keradus: examples in PHP documentation doesn't have that space, but there is not such a case mentioned here: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md

EspadaV8: @keradus do you meant that there's no case for the block element formatting? In the PHP docs you can see that `ticks` can be used with curly bracers to apply to only a part of the file (which is invalid for `strict_types`) - http://php.net/manual/en/control-structures.declare.php#control-structures.declare

```php
<?php
// these are the same:

// you can use this:
declare(ticks=1) {
    // entire script here
}

// or you can use this:
declare(ticks=1);
// entire script here
?>
```

Since it's a control structure then I would assume it'd follow the same rules as the other control structure items in PSR 2 - https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#54-for - with the closest similar examle being the `for` section.

It could be the case that all single line `declare` statements are formatted the same (without spaces) but I couldn't find anything else mentioning this so I thought it best to just stick with what I could tell.

keradus: I would go to direction of PSR12 proposal.
https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#strict-types-namespace-and-use-declarations

So form without braces as you show.

The problem is that there is no standard for form with braces (or at least I can't find them), ref https://github.com/cs-extended/fig-standards/issues/24

EspadaV8: Strict types _only_ allows the single line version and it _has_ to be the first thing in a file.

Other `declare` statements _can_ be multiline statements though or they could be on a single line.

keradus: then my initial commit - fixer shouldn't care if it is a declaration for strict or sth else. should also work for other - as it is.
then don't understand the need for isStrictTypeDeclare check

EspadaV8: Ah okay.  I'll try to get some more tests written up just to clarify things in the morning and update the PR.

keradus:
ref https://github.com/cs-extended/fig-standards/commit/1ebe95b3aeea5554b2a0d860439e104bd8266fd9

EspadaV8:
I've cleaned up some of the code with that document in mind, however, I'm not sure how to go about making sure that there are no spaces within the brackets after the declare. I couldn't find another test that does something similar since the only thing I could think of would be formatting of e.g. `for`s, and there doesn't seem to be a fixer for that.

Any help in pointing out the fixers and tests for those would be great, thanks.

EspadaV8:
I found some code in the `FunctionDeclarationFixer` that seems to do something similar. I've reused some of that in the declare fixers so that it now removes any whitespace within the parenthesis.

keradus:
Please, rebase on 1.11 to get rid of that merge commit

EspadaV8:
That should fix the merge commit.

Symfony/CS/Fixer/PSR2/BracesFixer.php
```diff
@@ -367,7 +369,10 @@ private function fixSpaceAroundToken(Tokens $tokens)
         for ($index = $tokens->count() - 1; 0 <= $index; --$index) {
             $token = $tokens[$index];
 
-            if ($token->isGivenKind($controlTokens) || $token->isGivenKind(T_USE)) {
+            // Declare tokens don't follow the same rules are other control statements
```
keradus: as declare is not control statement. comment could be removed

EspadaV8: It's listed in the `getControlTokens` method - https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/PSR2/BracesFixer.php#L529 -  and it's under the control structures section of the PHP docs -  
http://php.net/manual/en/control-structures.declare.php

keradus: :(

OK, then nevermind...

keradus:
Thank you @EspadaV8.

PR #1496, opened 2015-11-10
Add docblock to AbstractFixerTestBase::makeTest

lmanzke:
Just a quick addition of a docblock for the makeTest function, explaining a bit more what to pass in.

Intended to avoid confusion in the future, feel free to make additions to it!

Best regards,

Lucas

### DIFF
--- a/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
+++ b/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
@@ -38,8 +38,9 @@ abstract class AbstractFixerTestBase extends \PHPUnit_Framework_TestCase
     }
 
     /**
-     * Tests if a fixer fixes a given input string to match the expected result. It is used both if you want to test
-     * if something is fixed or if it is not touched by the fixer.
+     * Tests if a fixer fixes a given input string to match the expected result.
+     *
+     * It is used both if you want to test if something is fixed or if it is not touched by the fixer.
      * It also makes sure that the expected output does not change when run through the fixer. That means that you do not
      * need two test cases like [$expected] and [$expected, $input] (where $expected is the same in both cases)
      * as the latter covers both of them.
--- a/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
+++ b/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
@@ -38,10 +38,18 @@ abstract class AbstractFixerTestBase extends \PHPUnit_Framework_TestCase
     }
 
     /**
+     * Tests if a fixer fixes a given input string to match the expected result. It is used both if you want to test
+     * if something is fixed or if it is not touched by the fixer.
+     * It also makes sure that the expected output does not change when run through the fixer. That means that you do not
+     * need two test cases like [$expected] and [$expected, $input] (where $expected is the same in both cases)
+     * as the latter covers both of them.
+     *
      * @param string              $expected The expected fixer output.
      * @param string|null         $input    The fixer input, or null if it should intentionally be equal to the output.
      * @param \SplFileInfo|null   $file     The file to fix, or null if unneeded.
      * @param FixerInterface|null $fixer    A fixer that should be used, or null if it should be inferred from the test name.
+     *
+     * @throws \InvalidArgumentException If $expected and $input are equal. Leave $input blank if that is intended.
      */
     protected function makeTest($expected, $input = null, \SplFileInfo $file = null, FixerInterface $fixer = null)
     {
--- a/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
+++ b/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
@@ -37,6 +37,12 @@ abstract class AbstractFixerTestBase extends \PHPUnit_Framework_TestCase
         return $files[$filename];
     }
 
+    /**
+     * @param string              $expected The expected fixer output.
+     * @param string|null         $input    The fixer input, or null if it should intentionally be equal to the output.
+     * @param \SplFileInfo|null   $file     The file to fix, or null if unneeded.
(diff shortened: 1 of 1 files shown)


keradus:
the docs doesn't say that method tests both `expoected === fix(input)` and `expected === fix(expected)`

lmanzke:
Where would you add this? I just added the note that input should be null if it is intentionally the same.
Would you just have some paragraph above explaining that?

keradus:
> Where would you add this? I just added the note that input should be null if it is intentionally the same.

But don't tell why. And that won't protect us for generating tests like `[$exp], [$exp, $inp]`.

Yeah, summary + description would make it!

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
```diff
@@ -37,6 +37,20 @@ protected function getTestFile($filename = __FILE__)
         return $files[$filename];
     }
 
+    /**
+     * Tests if a fixer fixes a given input string to match the expected result. It is used both if you want to test
```
GrahamCampbell: Please format this correctly.

lmanzke: Please be more specific. What is not to your liking?

SpacePossum: I think the first line should a short description which may not expend to the next line (I think it is a PSR 5 rule).

lmanzke: Thanks, that is much more helpful. Will look into it and update.

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
```diff
@@ -37,6 +37,21 @@ protected function getTestFile($filename = __FILE__)
         return $files[$filename];
     }
 
+    /**
+     * Tests if a fixer fixes a given input string to match the expected result.
+     *
+     * It is used both if you want to test if something is fixed or if it is not touched by the fixer.
```
GrahamCampbell: Please ensure lines don't exceed 80 characters in length, including the first 5 spaces, the star, and the space.

SpacePossum: why is that? 
PSR 2  `lines shoud be 80 characters or less.` ? (following `MUST NOT be a hard limit on line length`)
or did I miss something in PSR5?

lmanzke: I updated it to obey the restrictions, although 80 characters seems not much. Have a look and decide.

GrahamCampbell: No, it's Laravel's own code style that dictates this. We're very strict on comments and phpdoc actually. ;)

GrahamCampbell: Ignore my last comment. It's been a long day, LOL.

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
```diff
@@ -37,6 +37,30 @@ protected function getTestFile($filename = __FILE__)
...
+     * @param string|null         $input    The fixer input, or null
+     *                                      if it should intentionally
+     *                                      be equal to the output.
+     * @param \SplFileInfo|null   $file     The file to fix, or null
+     *                                      if unneeded.
+     * @param FixerInterface|null $fixer    The fixer to be used, or null
+     *                                      if it should be inferred
+     *                                      from the test name.
+     *
+     * @throws \InvalidArgumentException If $expected and $input are equal.
```
keradus: we do not use that annotation.
it would be usefull if it would be followed by all methods, but it does not

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
```diff
@@ -37,6 +37,30 @@ protected function getTestFile($filename = __FILE__)
...
     }
 
+    /**
+     * Tests if a fixer fixes a given string to match the expected result.
+     *
+     * It is used both if you want to test if something is fixed or if it
+     * is not touched by the fixer. It also makes sure that the expected
+     * output does not change when run through the fixer. That means that
+     * you do not need two test cases like
+     * [$expected] and
```
keradus: Lot of cool text, for me keeping them to 70 chars is pointless and make it less readable.
Especially that line.

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
SpacePossum: have one one me :beer:  ;)

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
lmanzke: So how to format it now? If there is no defined answer here, then it should be specified. The content is important, it should not be hindered by the format of the docblock.

keradus: I would just not keep to that ugly 70 limit.
Especially when the following code is wider.

GrahamCampbell: Yeh, makes sense to wrap most lines except where it looks stupid.

lmanzke: Then I'll just wrap it at the PHPStorm configured max line length, as that will probably be widely used.

SpacePossum: sounds like a plan :)  :+1:

keradus:
really nice :+1:

good job!

SpacePossum:
nice :+1:  thanks

keradus:
Thank you @lmanzke.

PR #1499, opened 2015-11-11
Adds common whitespace operator fixer

israelshirk:
Adds common whitespace fixer.  Probably duplicates a bunch of stuff.

One possible reformulation for resolving duplication would be to add spacing options on CLI or in config - I just needed a quick fix for this but would be glad to take a look into a prettier solution if that's useful to you guys.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,12 @@ Choose from the list of available fixers:
                         placed. Body of braces should
                         be properly indented.
 
+* **common_operator_whitespace**
+                        Common math, comparison, and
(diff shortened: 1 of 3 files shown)


gharlan:
What is the difference to the existing [`OperatorsSpacesFixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/Symfony/OperatorsSpacesFixer.php)?

israelshirk:
That'll do.  Thanks for the link.

PR #1501, opened 2015-11-13
Add class definition fixer.

SpacePossum:
Fix white spaces in classy definitions.

For example:
```php
<?php
interface     Test    extends TestInterface  ,  /* */   TestInterface2   {
}
```

to:

```php
<?php
interface Test extends TestInterface, /* */ TestInterface2 {
}
```

### DIFF
--- a/Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
@@ -42,7 +42,7 @@ final class ClassyDefinitionFixer extends AbstractFixer
                 if ($tokens[$j]->isWhitespace()) {
                     $content = $tokens[$j]->getContent();
                     if (false === $breakAt = strpos($content, "\n")) {
-                        if (',' === $tokens[$j + 1]->getContent()) {
+                        if ($tokens[$j + 1]->equals(',')) {
                             $tokens[$j]->clear();
                         } elseif (' ' !== $content) {
                             $tokens[$j]->setContent(' ');
--- a/Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class ClassyDefinitionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(Token::getClassyTokenKinds());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->getSize() - 1; $index > 0;--$index) {
+            if (!$tokens[$index]->isClassy()) {
+                continue;
+            }
+            $classyStart = $tokens->getNextTokenOfKind($index, array('{'));
+            for ($j = $index + 1; $j < $classyStart; ++$j) {
+                if ($tokens[$j]->isWhitespace()) {
+                    $content = $tokens[$j]->getContent();
+                    if (false === $breakAt = strpos($content, "\n")) {
+                        if (',' === $tokens[$j + 1]->getContent()) {
+                            $tokens[$j]->clear();
+                        } elseif (' ' !== $content) {
+                            $tokens[$j]->setContent(' ');
+                        }
+                    } else {
+                        $tokens[$j]->setContent(substr($content, $breakAt));
+                    }
+                }
+            }
+        }
+    }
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
I don't think we should allow comments in those places either tbh. I think it would be cool if the fixer remoed them.

keradus:
it's valid in PSR-2 matter:
```php
<?php
class ClassName extends ParentClass implements
    \ArrayAccess,  /* to have [] */
    \Countable,    // to count them as well ;)
    \Serializable  // as we want to serialize it!
{
```

Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
```diff
@@ -0,0 +1,65 @@
...
+        for ($index = $tokens->getSize() - 1; $index > 0;--$index) {
+            if (!$tokens[$index]->isClassy()) {
+                continue;
+            }
+            $classyStart = $tokens->getNextTokenOfKind($index, array('{'));
+            for ($j = $index + 1; $j < $classyStart; ++$j) {
+                if ($tokens[$j]->isWhitespace()) {
+                    $content = $tokens[$j]->getContent();
+                    if (false === $breakAt = strpos($content, "\n")) {
+                        if (',' === $tokens[$j + 1]->getContent()) {
```
keradus: ->equals

GrahamCampbell:
> it's valid in PSR-2 matter:

Yeh. Looks ugly though. Also, this fixer isn't in the PSR2 level. :P

Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
SpacePossum: Will change.
On a side note, I was wondering why tokens like these `,` and IIRC `:` after a `case` do not have an `id`/`kind`, just out of curiosity.

keradus:
it doesn't matter, it is a standard.

Soullivaneuh:
At least, this fixer should be named `class_definition_whitespaces` and another one can be propose to remove comments IMO.

I'm surprised to see this fixer on Contrib section. There is no valid PSR about class definition for that?

SpacePossum:
https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#41-extends-and-implements

There are few rules in PSR 2 about what should be on one line. There is nothing about the separation of the items on it, unless some other rules apply on these items?

keradus:
items?

SpacePossum:
I mean stuff like `implements` `class` `extends` ;)

Symfony/CS/Fixer/Contrib/ClassyDefinitionFixer.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        for ($index = $tokens->getSize() - 1; $index > 0;--$index) {
+            if (!$tokens[$index]->isClassy()) {
+                continue;
+            }
+            $classyStart = $tokens->getNextTokenOfKind($index, array('{'));
+            for ($j = $index + 1; $j < $classyStart; ++$j) {
+                if ($tokens[$j]->isWhitespace()) {
+                    $content = $tokens[$j]->getContent();
+                    if (false === $breakAt = strpos($content, "\n")) {
```
dereuromark: inline assignment and yoda together are bad, each for itself already is. It would be better to put the assignment on a line itself above.

SpacePossum: >inline assignment and yoda together are bad, each for itself already is. It would be better to put the assignment on a line itself above.

Can you elaborate on that?
I think yoda-style is common practice in some projects (including Symfony http://symfony.com/doc/current/contributing/code/standards.html), so why would it be `bad`?
How would putting the declaration above it improve something? Is it readability or is there some technical gain as well?

keradus: Internet is full of discussion is yoda bad or not, please do not discuss it here. We follow it, just like Symfony. One may not follow it, great. It's CS.

The real matter here is not to assign inside condition. Do it before, condition is only for checking conditions ;)

dereuromark: Yes, other places often by accident use single =.
By never using inline assignment you cannot make this mistake, making accidental inline assignment in conditions not possible by design and solving the problem at its root.

SpacePossum: alright I'll update it :)

SpacePossum:
PSR-2 states:
```
Control structure keywords MUST have one space after them; method and function calls MUST NOT.
```
So there are no rules for white space around between `class`, `interface`, `extends`, so this fixer should stay at contrib (unless Symfony agrees).

Renamed the class following the suggestion of @Soullivaneuh and did a fix following @dereuromark.

SpacePossum:
hmm forgot to re-gen the `readme`, will do later on.

keradus:
* choose right branch for PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/CONTRIBUTING.md#quick-guide
* tests failing
* Symfony level indeed

SpacePossum:
>choose right branch for PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/CONTRIBUTING.md#quick-guide

I know about that file. However I think I'm going to wait with this one until either 2.0 is released or a feature stop is in place for 1.x.

>tests failing

I noticed https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1501#issuecomment-157491192 :P

>Symfony level indeed

good to hear, I'll take that into account the next time.

keradus:
> I know about that file. However I think I'm going to wait with this one until either 2.0 is released or a feature stop is in place for 1.x.

why is that? what exactly do you need for this fixer from 2.x line that you don't have at 1.x line?
Fixer looks pretty ready, just send it to 1.x ;)

PR #1503, opened 2015-11-14
Psr0Fixer - fix case with comments lying around

GrahamCampbell:

PR #1504, opened 2015-11-14
[2.0] Revert broken change to the psr0 fixer

GrahamCampbell:
The tests added in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1503 fail without reverting this change. It seems to only be present on the master branch. Not sure why it was added? Ping @keradus.


GrahamCampbell:
Just for proof, I've sent my 1.10 change to 2.0 so we can see exactly what happens: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1505.

GrahamCampbell:
It doesn't seem to be that causing the other failures actually. Hmm.

PR #1281, opened 2015-07-02
[2.0] Allow installs along side symfony 3

GrahamCampbell:

PR #1510, opened 2015-11-17
Added Symfony 3.0 support

Ener-Getick:
This PR adds symfony 3.0 support. Several packages can't support symfony 3.0 because of this package ...

This was already discussed in #1281 but it was possible to have dev dependencies during the travis build.
This PR only allows ``symfony/symfony`` to be in a dev version and ``prefer-stable`` is used so when symfony 3.0 will be released, the travis build will use directly the stable version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,27 @@
 language: php
 
-php:
-    - 5.3
-    - 5.4
-    - 5.5
-    - 5.6
-    - 7.0
-    - hhvm
-
 matrix:
     fast_finish: true
     include:
         - php: 5.3
           env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
-    allow_failures:
+        - php: 5.4
+        - php: 5.5
+        - php: 5.6
+          env: SYMFONY_VERSION="~2.7"
+        - php: 5.6
+          env: SYMFONY_VERSION="~3.0@dev" COMPOSER_FLAGS="--prefer-stable"
         - php: 7.0
+        - php: hhvm
 
 sudo: false
 
 cache:
     directories:
         - $HOME/.composer/cache
+
+before_install:
+    - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
 
 install:
     - travis_retry composer update ${COMPOSER_FLAGS} --no-interaction
--- a/composer.json
+++ b/composer.json
@@ -16,12 +16,12 @@
     "require": {
         "php": ">=5.3.6",
         "ext-tokenizer": "*",
-        "symfony/console": "~2.3",
-        "symfony/event-dispatcher" : "~2.1",
-        "symfony/filesystem": "~2.1",
-        "symfony/finder": "~2.1",
-        "symfony/process": "~2.3",
-        "symfony/stopwatch": "~2.5",
+        "symfony/console": "~2.3|~3.0",
+        "symfony/event-dispatcher": "~2.1|~3.0",
+        "symfony/filesystem": "~2.1|~3.0",
+        "symfony/finder": "~2.1|~3.0",
+        "symfony/process": "~2.3|~3.0",
+        "symfony/stopwatch": "~2.5|~3.0",
         "sebastian/diff": "~1.1"
     },
     "require-dev": {


Soullivaneuh:
:+1:

GrahamCampbell:
:-1: for going to 1.10. Should go to 1.11 IMO.

GrahamCampbell:
:-1: for this anyway because the version selection doesn't actually work.

GrahamCampbell:
3.0 will end up getting installed by default as soon as it's stable so you're not even running the tests on 2.x on all php versions at that point.

Ener-Getick:
> 
:-1: for this anyway because the version selection doesn't actually work.

It allows to install symfony 3.0 so yes it works:
```json
{
    "require": {
        "fabpot/php-cs-fixer": "~1.0",
        "symfony/symfony": "3.0.x-dev"
    }
}
```
This actually doesn't work ...

>
3.0 will end up getting installed by default as soon as it's stable so you're not even running the tests on 2.x on all php versions at that point.

Symfony 3.0 only removes the deprecated interfaces. So something that works with symfony 3.0 works with symfony 2.8. Then there is a build that uses symfony 2.3, symfony respects bc so we are sure that the package works with symfony ``~2.3|~3.0``...
To be even more sure I added a build that uses symfony ``2.7.*``.

GrahamCampbell:
Better to have `~2.7` and `~3.0` IMO.

Ener-Getick:
For the branch, I targeted ``1.10`` for this to be released faster but I can create a new PR targeting ``1.11`` if you want.

GrahamCampbell:
> For the branch, I targeted 1.10 for this to be released faster but I can create a new PR targeting 1.11 if you want.

No need. It can be merged into either branch by @keradus. ;)

GrahamCampbell:
Just noticed you're requiring ALL of symfony there. Massive :-1: for that. Just require each component.

GrahamCampbell:
Once that's sorted, :+1: from me. ;)

Ener-Getick:
This is the easiest way and this is how all the bundles I know works (see several examples here https://docs.google.com/spreadsheets/d/1GOxJ2lNpLxN12YyZ87Bhsxb0iLAQ7e4UDZbJSly8Sy8/edit).

Another solution is to load the dev version of all the components but I'm not sure that this is better for this package (see https://github.com/dunglas/DunglasApiBundle/blob/1.0/.travis.yml#L26).

keradus:
> This is the easiest way and this is how all the bundles I know works

`throw new InvalidArgumentException('This is not the bundle.');`


When you download whole framework then you will be able to use classes in code that aren't even required in composer directly! Not to mention the performance impact to Travis...
Such solution won't be merged.

GrahamCampbell:
@keradus Exactly.

Soullivaneuh:
PHP-CS-Fixer requires many symfony dependencies. Requiring them one by one on travis would be painful.

I don't get the point of why required symfony full stack is bad? It is just for Travis, in order to test a specific version of all bundles.

As @Ener-Getick almost all library use this method without any problem.

:+1: for me.

Ener-Getick:
@keradus @GrahamCampbell Do you prefer an update of only the necessary dependencies ? (see 338ea52)

keradus:
> PHP-CS-Fixer requires many symfony dependencies.

nope, only 3 components.


> Do you prefer an update of only the necessary dependencies ? (see 338ea52)

better, what about modify composer.json file on the go to replace all packages `symfony/.*` using eg regex ? Then you won't need to manually list all of them in travis file

Ener-Getick:
Good idea ! I changed it.

mmoreram:
:+1: to merge this and to add a new tag :)

TomasVotruba:
[As for Symfony 3 bundles support](https://docs.google.com/spreadsheets/d/1GOxJ2lNpLxN12YyZ87Bhsxb0iLAQ7e4UDZbJSly8Sy8/edit#gid=0), every package using this package depends on this to be merged.

E.g. https://travis-ci.org/mmoreram/php-formatter/jobs/92008056

Ener-Getick:
Why @TomasVotruba? this change only removes symfony 2.2 support and symfony 3.0 is already supported apparently...

TomasVotruba:
@Ener-Getick Since which version?

stof:
> When you download whole framework then you will be able to use classes in code that aren't even required in composer directly

no you can't, because Travis requires the full framework only for a few jobs, not for all of them

keradus:
> no you can't, because Travis requires the full framework only for a few jobs, not for all of them

and in that jobs you would be able.

stof:
@keradus but the testsuite and the PHP-CS-Fixer codebase are the same in all jobs, so such missing dependency would still make the Travis build fail, even though one of the jobs is passing (and locally, it would also fail as there is no the extra symfony requirement)

keradus:
Yep, that's why it wasn't the only one argument.
Also, in such a case it would be confusing to new contributors that tests passed using Symfony 3.x and not using Symfony 2.x.

Ener-Getick:
Sorry @TomasVotruba I misunderstood.

@keradus @GrahamCampbell is there something else to do before this could be merged?

keradus:
Please rebase to get rid of that merge commit.

(From my side - I will need to check sth with this PR on myside)

Ener-Getick:
This is done @keradus

keradus:
sth is wrong now.

Take a look at number of jobs:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/92326598

When you added items to matrix: https://github.com/Ener-Getick/PHP-CS-Fixer/blob/SF3/.travis.yml#L11
you should remove them from php: https://github.com/Ener-Getick/PHP-CS-Fixer/blob/SF3/.travis.yml#L3
(also remember to move hhvm)

keradus:
:+1:

SpacePossum:
:+1:  lgtm :)

Ener-Getick:
ping

keradus:
Thank you @Ener-Getick.

TomasVotruba:
@keradus @Ener-Getick Thank you guys!

Soullivaneuh:
@keradus What is the planning date for 1.11 release?

If it's not for now, I think we should back-port this commit onto 1.10 branch and tag a last patch version.

Could be possible? Symfony 3.0 is released as stable now.

keradus:
PATCH semver is only for bugfix.

GrahamCampbell:
Yeh. This doesn't need backporting, and 1.11 isn't too far off release anyway.

keradus:
Waiting for last few PR to be finished

Soullivaneuh:
Well, make project compatible with higher stable vendor is a kind of bugfix IMHO. :-)

But if 1.11 will be out shortly, that's ok. :+1:

keradus:
nope, it is not the bug. you can still run the tool.

PR #1511, opened 2015-11-17
Clean up docs

SpacePossum:


keradus:
why not 1.11 ?

SpacePossum:
let me try again ;)

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1024,7 +1024,7 @@ public function isAnyTokenKindsFound(array $tokenKinds)
     /**
      * Check if token kind given as argument is found.
      *
-     * @param array
+     * @param int|string $tokenKind
```
SpacePossum: this fix must be done on master, the method it is not available on 1.x

keradus: what can be fix on 1.x line it should be fixed there. all the rest on master ;)

PR #1512, opened 2015-11-17
[1.11] Extended PHPDoc Work

GrahamCampbell:
Extends and replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1487.


Symfony/CS/DocBlock/Tag.php
```diff
@@ -26,30 +26,62 @@ class Tag
...
+    private $typesContent;
+
+    /**
      * Create a new tag instance.
      *
-     * @param string $content
+     * @param Line $line
      */
-    public function __construct($content)
+    public function __construct(Line $line)
```
keradus: couldn't tag may be multilined ?

GrahamCampbell: The "Annotation" can be, yeh, but the "Tag", no. ;)

GrahamCampbell: Ahh, wait, in reflection, I've put these methods on the wrong class. They should have gone on the annotation class, lol. I'll need to revert some of this. ;)

GrahamCampbell: I'll sort that tomorrow.

keradus: btw, isn't it a BC break in current form ?

GrahamCampbell: yeh, that too, lol

GrahamCampbell: Ok, I've made the changes now, but they still technically are breaking.

GrahamCampbell: We basically need to tag class to not be read-only anymore, so we need to pass through an actual line instance rather than just the content of a line.

GrahamCampbell:
How far off is the 2.0 release? Is it maybe worth me targeting this at 2.0 instead, and closing the other PR that's still open?

GrahamCampbell:
How many more 1.x minor releases do you have in mind?

keradus:
I do not plan that "there will be exactly 3 minor releases before 2.0", when there will be need of new minor then it will be released.
How far? You tell me. Plan what need to be done before 2.x was mentioned long ago, what left to do is public: https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestones/v2.0

GrahamCampbell:
Renaming the fixers should be trivial once we're all agreed on names. In terms of dealing with the conflicting stuff, I don't really have any time right now to look into that.

GrahamCampbell:
As long as people are happy with me sending more and more phpdoc fixers to only the master, that are dependent on this stuff, I'm happy to retarget this at the master. Otherwise, I'm not sure how we proceed, since I'd like to make the breaking changes to the constructor of the the tag class.

GrahamCampbell:
I guess that part could technically be moved to 2.x, and we only keep the tags stuff on here.

GrahamCampbell:
Will send in smaller chunks.

PR #1518, opened 2015-11-19
Added ClassDefinitionFixer

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1501

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,6 +197,12 @@ Choose from the list of available fixers:
                         placed. Body of braces should
                         be properly indented.
 
+* **class_definition** [PSR-2]
+                        Whitespace around the key
+                        words of a class, trait or
+                        interfaces definition should
+                        be one space.
+
 * **elseif** [PSR-2]
                         The keyword elseif should be
                         used instead of else if so
--- a/Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
+++ b/Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
@@ -1,0 +1,208 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\PSR2;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * Fixer for part of the rules defined in PSR2 ¶4.1 Extends and Implements.
+ *
+ * @author SpacePossum
+ */
+final class ClassDefinitionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = $tokens->getSize() - 1; $index > 0; --$index) {
+            if (!$tokens[$index]->isClassy()) {
+                continue;
+            }
+
+            $this->fixClassDefinition($tokens, $index, $tokens->getNextTokenOfKind($index, array('{')));
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Whitespace around the key words of a class, trait or interfaces definition should be one space.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before the TrailingSpacesFixer
+        return 21;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $start      Class definition token start index
+     * @param int    $classyOpen Class definition token end index
+     */
+    private function fixClassDefinition(Tokens $tokens, $start, $classyOpen)
+    {
+        // check if there is a `implements` part in the definition, since there are rules for it in PSR 2.
+        $implementsInfo = $this->getMultiLineInfo($tokens, $start, $classyOpen);
+
+        // 4.1 The extends and implements keywords MUST be declared on the same line as the class name.
+        if ($implementsInfo['numberOfInterfaces'] > 1 && $implementsInfo['multiLine']) {
+            $classyOpen += $this->ensureWhiteSpaceSeparation($tokens, $start, $implementsInfo['breakAt']);
+            $this->fixMultiLineImplements($tokens, $implementsInfo['breakAt'], $classyOpen);
+        } else {
+            $classyOpen -= $tokens[$classyOpen - 1]->isWhitespace() ? 2 : 1;
+            $this->ensureWhiteSpaceSeparation($tokens, $start, $classyOpen);
+        }
+    }
+
+    /**
+     * Returns an array with `implements` data.
+     *
+     * Returns array:
+     * * int  'breakAt'            index of the Token of type T_IMPLEMENTS for the definition, or 0
+     * * int  'numberOfInterfaces'
+     * * bool 'multiLine'
+     *
+     * @param Tokens $tokens
+     * @param int    $start
+     * @param int    $classyOpen
+     *
+     * @return array
+     */
+    private function getMultiLineInfo(Tokens $tokens, $start, $classyOpen)
+    {
+        $implementsInfo = array('breakAt' => 0, 'numberOfInterfaces' => 0, 'multiLine' => false);
+        $breakAtToken = $tokens->findGivenKind($tokens[$start]->isGivenKind(T_INTERFACE) ? T_EXTENDS : T_IMPLEMENTS, $start, $classyOpen);
+        if (count($breakAtToken) < 1) {
+            return $implementsInfo;
+        }
+
+        $implementsInfo['breakAt'] = key($breakAtToken);
+        $classyOpen = $tokens->getPrevNonWhitespace($classyOpen);
+        for ($j = $implementsInfo['breakAt'] + 1; $j < $classyOpen; ++$j) {
+            if ($tokens[$j]->isGivenKind(T_STRING)) {
+                ++$implementsInfo['numberOfInterfaces'];
+                continue;
+            }
+
+            if (!$implementsInfo['multiLine'] && ($tokens[$j]->isWhitespace() || $tokens[$j]->isComment()) && false !== strpos($tokens[$j]->getContent(), "\n")) {
+                $implementsInfo['multiLine'] = true;
+            }
+        }
+
+        return $implementsInfo;
+    }
+
+    /**
+     * Fix spacing between lines following `implements`.
+     *
+     * PSR2 4.1 Lists of implements MAY be split across multiple lines, where each subsequent line is indented once.
+     * When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.
+     *
+     * @param Tokens $tokens
+     * @param int    $breakAt
+     * @param int    $classyOpen
+     */
+    private function fixMultiLineImplements(Tokens $tokens, $breakAt, $classyOpen)
+    {
+        // implements should be followed by a line break, but we allow a comments before that,
+        // the lines after 'implements' are always build up as (comment|whitespace)*T_STRING{1}(comment|whitespace)*','
+        // after fixing it must be (whitespace indent)(comment)*T_STRING{1}(comment)*','
+        for ($index = $classyOpen - 1; $index > $breakAt - 1; --$index) {
+            if ($tokens[$index]->isWhitespace()) {
+                if ($tokens[$index + 1]->equals(',')) {
+                    $tokens[$index]->clear();
+                } elseif ($tokens[$index + 1]->isComment() && ' ' !== $tokens[$index]->getContent()) {
+                    $tokens[$index]->setContent(' ');
+                }
+            }
+
+            if ($tokens[$index]->isGivenKind(T_STRING)) {
+                $index = $this->ensureOnNewLine($tokens, $index);
+            }
+        }
+    }
+
+    /**
+     * Make sure the tokens are separated by a single space.
+     *
+     * @param Tokens $tokens
+     * @param int    $start
+     * @param int    $end
+     *
+     * @return int number tokens inserted by the method before the end token
+     */
+    private function ensureWhiteSpaceSeparation(Tokens $tokens, $start, $end)
+    {
+        $insertCount = 0;
+        for ($i = $end; $i > $start; --$i) {
+            if ($tokens[$i]->isWhitespace()) {
+                if (' ' !== $tokens[$i]->getContent()) {
+                    $tokens[$i]->setContent(' ');
+                }
+                continue;
+            }
+
+            if ($tokens[$i - 1]->isWhitespace() || "\n" === substr($tokens[$i - 1]->getContent(), -1)) {
+                continue;
+            }
+
+            if ($tokens[$i - 1]->isComment() || $tokens[$i]->isComment()) {
+                $tokens->insertAt($i, new Token(array(T_WHITESPACE, ' ')));
+                ++$insertCount;
+                continue;
+            }
+        }
+
+        return $insertCount;
+    }
+
+    private function ensureOnNewLine(Tokens $tokens, $index)
+    {
+        // while not whitespace and not comment go back
+        for (--$index; $index > 0; --$index) {
+            if (!$tokens[$index]->isGivenKind(array(T_NS_SEPARATOR, T_STRING))) {
+                break;
+            }
+        }
+
+        if ("\n" === substr($tokens[$index]->getContent(), -1)) {
+            return $index;
+        }
+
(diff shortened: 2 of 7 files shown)


keradus:
great to see this ;)

Symfony/CS/Tests/Fixer/Symfony/ClassDefinitionWhitespacesFixerTest.php
```diff
@@ -0,0 +1,82 @@
...
+class Abc {}
+',
+                 '<?php
+class    Abc    {}
+',
+             ),
+            array(
+                '<?php
+class Test extends \Exception
+             implements TestInterface
```
keradus: implements should go to the prev line

keradus:
Isn't this fixer fix PSR-2 ¶4.1. ?

SpacePossum:
> Isn't this fixer fix PSR-2 ¶4.1. ?

If I would make the fixer do; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1518#discussion_r45623820 it might get close ;)

I could not find a rule stating there must be one space between the class/interface/trait name and `implements` / `extends` etc., so this fixer does not format to a PSR rule I think.

keradus:
OK, so if we want to be so strict (it's OK with me) - please, could you add second fixer, that replace whitespaces around `extends` and `implements` to be not-multiline? like `__\n___` to `_____` ?
And then, run this fixer after it and it will convert it to `_` ;)

SpacePossum:
I like the idea, so I'm putting this one back to WIP status. I'll rework the fixer to apply PSR2 4.1 where possible. This will take a bit more time, so that expect a useful update soon ;)

keradus:
@SpacePossum could you please focus on this PR? I would like to merge it soon ! :D

SpacePossum:
@keradus I rewrote the whole thing, I'm going to puzzle on both performance and more test cases, 
but please early comments are welcome  :)
btw. this should fix to PSR-2 ¶4.1. now ^-^

README.rst
```diff
@@ -532,6 +538,11 @@ Choose from the list of available fixers:
                         Align equals symbols in
                         consecutive lines.
 
+* **case_parenthesis** [contrib]
```
SpacePossum: oh oops, ignore, there is still a fixer in my env. that should not be there ;)

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -397,9 +402,10 @@ public function findGivenKind($possibleKind)
...
         }
 
-        foreach ($this as $index => $token) {
+        for ($i = $start;  $i < $end; ++$i) {
+            $token = $this[$i];
             if ($token->isGivenKind($possibleKinds)) {
-                $elements[$token->getId()][$index] = $token;
+                $elements[$token->getId()][$i] = $token;
             }
         }
```
SpacePossum: these chagnes can be done in a diff. PR if wanted

keradus: no need for separate PR but need a tests in tokenizer component

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $start  Class definition token start index
+     * @param int    $end    Class definition token end index
+     */
+    private function fixClassDefinition(Tokens $tokens, $start, $end)
+    {
+        $implements = $tokens->findGivenKind(T_IMPLEMENTS, $start, $end);
+        if (count($implements) > 0) {
```
keradus: it is so big if...

Symfony/CS/Tokenizer/Tokens.php
SpacePossum: :+1: , please see 6137413d41212d521434a40583396a2ee4f391c8

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,185 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $start  Class definition token start index
+     * @param int    $end    Class definition token end index
+     */
+    private function fixClassDefinition(Tokens $tokens, $start, $end)
+    {
+        $implements = $tokens->findGivenKind(T_IMPLEMENTS, $start, $end);
+        // check if there is a `implements`  part in the definition, since there are rules for it in PSR 2.
+        if (count($implements) > 0) {
```
keradus: don;t know why git outdated my comment, but this if is so big.

SpacePossum: not sure how decomposition can be done in a meaningful way, there is not code repetition afaik.

keradus: even if there is no code duplication the flow is horrible to maintenance

SpacePossum:
fabbot is wrong IMO

GrahamCampbell:
> fabbot is wrong IMO

GrahamCampbell:
> fabbot is wrong IMO

Hmmm. That's not just fabbot. That's php-cs-fixer too.

GrahamCampbell:
Your phpdoc is also wrong though anyway.

GrahamCampbell:
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#7-describing-hashes

GrahamCampbell:
You must include a `$` in each variable name.

GrahamCampbell:
Also, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1557.

keradus:
If CS check won't pass after fixing your PHPDoc then please get rid of that PHPDoc hash and describe the content in long description.

keradus:
I would really love to have it in 1.11 and the PHPDoc won't stop us!

Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
```diff
@@ -0,0 +1,98 @@
...
+    TestInterface4,
+    TestInterface5, /**/
+    TestInterface6
+{
+}',
+                '<?php
+interface Test
+extends
+    TestInterface8
+  implements  /*a*/    /*b*/TestInterface1   ,  /* test */
```
gharlan: This is invalid. Interfaces can not implement other interfaces, they extend them.
So in interfaces you should handle `extends` like `implements` in classes.

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,218 @@
...
+    {
+        $added = 0;
+        // implements should be followed by a line break, but we allow a comments before that,
+        // the lines after 'implements' are always build up as (comment|whitespace)*T_STRING{1}(comment|whitespace)*','
+        // after fixing it must be (whitespace indent)(comment)*T_STRING{1}(comment)*','
+        for ($c = $end - 1; $c > $implementsAt - 1; --$c) {
+            if ($tokens[$c]->isWhitespace()) {
+                if ($tokens[$c + 1]->equals(',')) {
+                    $tokens[$c]->clear();
+                } elseif ($tokens[$c + 1]->isComment() && !$tokens[$c]->equals(' ')) {
```
gharlan: hmm, does this work?? `->equals(' ')`?
should be always `false`, because `' ' ` is not a valid token prototype.

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,218 @@
...
+ */
+final class ClassDefinitionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = $tokens->getSize() - 1; $index > 0;--$index) {
```
gharlan: Missing a space before `--`.

gharlan: We need a fixer for this ;)

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -383,12 +383,17 @@ public function findBlockEnd($type, $searchIndex, $findEnd = true)
      * Find tokens of given kind.
      *
      * @param int|array $possibleKind kind or array of kind
+     * @param int|null  $start        optional offset
```
gharlan: Shouldn't this be int only?

keradus: no

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: indeed, this is wrong, which also proofs the tests are not covering this cases... I'll rework this

Symfony/CS/Tokenizer/Tokens.php
GrahamCampbell: also, not optional then, so the description needs updating too

keradus: default value is `0`, so optional

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: feel free to raise an issue for this ;)

Symfony/CS/Tokenizer/Tokens.php
GrahamCampbell: ah, right - didn't read that far - the diff as viewed from the comment doesn't show it :(

GrahamCampbell: Also, I think we should ALWAYS use null as a default, like symfony and php do, so you can skip default values with a null as a convention.

keradus: tbh first time I red this i though it;s null too, just for the convention you mentioned

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: fixed, I'll file an issue if you don't ;)

Symfony/CS/Tokenizer/Tokens.php
SpacePossum: made it `int` only, default `0`, it is like `findSequence`, why would you default something to `null` if that always means you change it to `0`?

keradus: idea is to pass `($arg, null, 3)` - and then the dafault value for 2nd parameter is keeped.
sth like `default` keyword proposed for PHp7

Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
SpacePossum: :+1:  indeed
added (a failing) test as reminder to get this fixed
https://github.com/SpacePossum/PHP-CS-Fixer/commit/42c4c65e92e449e093b872b89f250d7ce937b15b

keradus:
@SpacePossum 
please, add a case:
```php
array(
    '<?php
    class Aaa implements Bbb, Ccc
    {
    }',
),
```

keradus:
https://github.com/SpacePossum/PHP-CS-Fixer/pull/5

:(

SpacePossum:
:(

@keradus your tests prove that setting the indenting before the class opening to `\n    ` might not be the right thing to do.
I suggest checking if there is one line break between the line with the class definition and the class opening '{', if there is leave as is, if there are more remove to one line break, if there is none add one break.
This will cause this fixer to move the class opening to the correct line, but it won't fix the indenting.
What you think?

keradus:
I can't understand why to move class opening at all?

SpacePossum:
```
PSR 2
4.1. Extends and Implements
The opening brace for the class MUST go on its own line; the closing brace for the class MUST go on the next line after the body.
```

you feel this is out of scope for this fixer?

keradus:
OK, if you want to fix it here - OK with me, but why move non-`{` tokens to new line ?

SpacePossum:
I removed the indenting part as well, so an `indenting fixer` should take care of that,
lets see how the tests do

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,199 @@
...
+     */
+    public function getDescription()
+    {
+        return 'White space around the key words of a class, trait or interfaces definition should be one space.';
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $start  Class definition token start index
+     * @param int    $end    Class definition token end index
```
keradus: what is it `class definition token end index` ?

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,199 @@
...
+        }
+    }
+
+    /**
+     * Returns an array with `implements` data.
+     *
+     * Returns array:
+     * int  'breakAt'            index of the Token of type T_IMPLEMENTS for the definition, or 0
+     * int  'numberOfInterfaces'
+     * bool 'multiLine'
```
keradus: add bullets like
- int
- int
- bool

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -0,0 +1,199 @@
...
+     * @param Tokens $tokens
+     * @param int    $start
+     * @param int    $end
+     *
+     * @return array
+     */
+    private function getMultiLineInfo(Tokens $tokens, $start, $end)
+    {
+        $implementsInfo = array('breakAt' => 0, 'numberOfInterfaces' => 0, 'multiLine' => false);
+        $implements = $tokens->findGivenKind($tokens[$start]->isGivenKind(T_INTERFACE) ? T_EXTENDS : T_IMPLEMENTS, $start, $end);
```
keradus: I can see extends here :)

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: it is the index of the last token that is part of the class definition

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: not sure what you want me to do here :)

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: From both phpdoc and your explanation i would say it is the index of last `}` in the following code:
```php
class Foo extends Bar implements Baz {
    public function xxx()
    {
    }
}
```

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: fix the var name, since it may not handle `T_IMPLEMENTS`

SpacePossum: :+1: 
renamed (`$end` as well)

keradus:
input file:
```php
<?php

class Aaa implements
    Symfony\CS\Tests\Fixer,\RFb,
    \Fcc1, \GFdd
{
}
```

fixing:

```bash
D:\GIT\fork\PHP-CS-Fixer ((detached from SpacePossum/1_11_feature_add_class_definition_fixer))
λ php php-cs-fixer fix issue -vv
F
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) Aaa.php (class_definition)
```

after first fix:
```php
<?php

class Aaa implements
    Symfony\CS\Tests\Fixer,
\RFb,
    \Fcc1, 
\GFdd
{
}

```
fixing:
```bash
D:\GIT\fork\PHP-CS-Fixer ((detached from SpacePossum/1_11_feature_add_class_definition_fixer))
λ php php-cs-fixer fix issue -vv
F
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) Aaa.php (trailing_spaces)
```

after second fix:
```php
<?php

class Aaa implements
    Symfony\CS\Tests\Fixer,
\RFb,
    \Fcc1,
\GFdd
{
}

```
fixing:
```bash
D:\GIT\fork\PHP-CS-Fixer ((detached from SpacePossum/1_11_feature_add_class_definition_fixer))
λ php php-cs-fixer fix issue -vv
.
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
```

------------------------------

priorities need to be set

SpacePossum:
after the second there are no changes which to be expected right?

keradus:
Thank you @SpacePossum.

SpacePossum:
\o/ 
thanks at all reviewers :)

PR #1519, opened 2015-11-19
[1.10] Fixed bug in tag regex

GrahamCampbell:


Symfony/CS/Tests/DocBlock/TagTest.php
```diff
@@ -37,6 +37,7 @@ public function provideNameCases()
             array('THROWSSS', "\t@THROWSSS\t  \t RUNTIMEEEEeXCEPTION\t\t\t\t\t\t\t\n\n\n"),
             array('other', ' *   @\Foo\Bar(baz = 123)'),
             array('expectedException', '     * @expectedException Exception'),
+            array('property-read', ' * @property-read integer $daysInMonth number of days in the given month'),
```
keradus: 1.10 doesn't support any tags with `-`, so this PR won't fix behaviour of any fixer

GrahamCampbell: yes and no

GrahamCampbell: we do "support" all tags, we just say they're not "valid"

GrahamCampbell: we should still be able to still return the correct tag with the getName method

although it would technically not change the operation of our fixers, it's still wrong, because `getName()` is returning `other` instead of `foo-bar` or whatever

GrahamCampbell:
Should I close this then?

keradus:
Merging other PR into 1.11 line has nothing to do with it.
In other PR there were more changes, and theoretically this one can still be merged into 1.10 line.

On way or another I want to release 1.11 at weekend and then life of 1.10 will ends.

GrahamCampbell:
Yes, but do you actually want it merged? I'm not bothered either way, and it sounded like you thought there was no point in merging, which is fair enough.

keradus:
fair enough

PR #1520, opened 2015-11-19
Code grooming

keradus:


Symfony/CS/Fixer/PSR2/VisibilityFixer.php
```diff
@@ -123,7 +123,7 @@ private function grabAttribsBeforeMethodToken(Tokens $tokens, $index)
      * @param Tokens $tokens Tokens collection
      * @param int    $index  token index
      *
-     * @return array array of grabbed attributes
+     * @return array<string, Token|null> array of grabbed attributes
```
GrahamCampbell: this syntax is not supported by our current fixers

GrahamCampbell: it's also not widely accepted as a standard

GrahamCampbell: the same way that PSR-5 says don't use `@var`, but we all do anyway

keradus: > this syntax is not supported by our current fixers

all is green


> it's also not widely accepted as a standard

accepted by PSR5 proposal


> the same way that PSR-5 says don't use `@var`, but we all do anyway

what?
https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#824-var

GrahamCampbell: It seems it really has been a while since I read the spec. Seems that change was reverted on 8th August 2014, lol. https://github.com/phpDocumentor/fig-standards/commit/61c695c43d1b2c713acbfb3302f284c518421bc5

GrahamCampbell: I think we should remove the var_to_type fixer in 2.x then since `@type` isn't a thing anymore.

keradus: feel free to send a PR

GrahamCampbell: i'll add an issue for it

GrahamCampbell: i'd like to do that after the other stuff has gone through and the branches have been merged

keradus: aren't they merged?

GrahamCampbell: no?

keradus: which one?

GrahamCampbell: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1487 and/or https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1512.

The 2nd one replaces the first one, but is breaking. I'm now suggesting take the non-breaking parts from the first, then apply them only to 1.11, then take the rest, and target the master.

keradus: oh, say you means PRs, not origin branches :D

GrahamCampbell: yeh :P

GrahamCampbell: well, both i guess :D

keradus: nope ;P

Symfony/CS/Tests/Fixer/Symfony/ArrayElementNoSpaceBeforeCommaFixerTest.php
```diff
@@ -78,26 +78,26 @@ public function testFixProvider()
             // multi line array
             array(
                 '<?php $x = [  "a" =>$a,
-                    "b"=>  
+                    "b"=>
```
SpacePossum: Are you sure having the trailing white space is not by design? (i.e. it tesst it is supported by the fixer being tested?)

keradus: I considered it. The fixer isn't touching it and doesn't care 'bout it.

SpacePossum: I see, but isn't the point of testing making sure that stays that way?

keradus: This cases tests completely different thing, I can't see any benefits of that spaces.

SpacePossum: :+1:  very well

Symfony/CS/Fixer/PSR2/VisibilityFixer.php
GrahamCampbell: still, i think that needs reverting - we shouldn't be using syntax our own code doesn't support

keradus: Then we should start supporting it

keradus: Extracted to separate PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1531

GrahamCampbell: > Then we should start supporting it

Easier said than done, but should be possible on the master branch when I'm finally done with the types stuff.

PR #1523, opened 2015-11-20
Added EchoToPrintFixer and PrintToEchoFixer

### LINKED ISSUE #1502: Feature request: print vs echo

Soullivaneuh:
Closes #1502.

> You know it so well... I think you want to make that fixer ! :D That's great ;P

image

If you have some concrete samples to take care of, please add it on comments. I'll add them on tests. :+1: 

Note: After that, those fixers could be merged into one with a direction option on 2.x branch.


Soullivaneuh:
On issue, I quoted from [w3school](http://www.w3schools.com/php/php_echo_print.asp):

> echo has no return value while print has a return value of 1 so it can be used in expressions

But the PHP [official documentation](http://php.net/manual/en/function.print.php#refsect1-function.print-returnvalues) says:

> Returns 1, always.

Since the return value is always 1 in any cases, I guess we could ignore it for conversion.

Soullivaneuh:
Well, forgot this.

```php
$ret = echo 'test';
```

Will throw syntax error. We have to take care of it.

README.rst
```diff
@@ -537,6 +537,10 @@ Choose from the list of available fixers:
                         with at least one whitespace
                         around.
 
+* **echo_to_print** [contrib]
+                        Converts echo functions to
```
sstok: `echo` is not a function its a language constructor, same goes for print (IIRC).

GrahamCampbell: :+1:

SpacePossum: `print is not actually a real function (it is a language construct) `
so yeah ;)

Soullivaneuh:
@keradus PR ready to review and merge. :+1:

Symfony/CS/Tests/Fixer/Contrib/PrintToEchoFixerTest.php
```diff
@@ -0,0 +1,78 @@
...
+                '<?php
+                echo("test");
+                ',
+                '<?php
+                print("test");
+                ',
+            ),
+        );
+    }
+
```
GrahamCampbell: cs

Symfony/CS/Tests/Fixer/Contrib/EchoToPrintFixerTest.php
```diff
@@ -0,0 +1,72 @@
...
+                '<?php
+                print("test");
+                ',
+                '<?php
+                echo("test");
+                ',
+            ),
+        );
+    }
+
```
GrahamCampbell: cs

Soullivaneuh:
CS corrected.

SpacePossum:
I would like to request a test cases for:
```php
function testFunction() {
  return print("test");
}

$a = testFunction();
$b += print($a);
$c='';
$c .= $b.print($a);
$d = print($c) > 0 ? 'a' : 'b';

```

keradus:
Wow, nice @Soullivaneuh !
I can see that PR is already in progress ;)

Soullivaneuh:
Ready for review.

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
```diff
@@ -0,0 +1,60 @@
...
+                    $canBeConverted = false;
+                    break;
+                }
+            }
+
+            if (false === $canBeConverted) {
+                continue;
+            }
+
+            $tokens[$echoIndex] = new Token(array(T_PRINT, 'print'));
```
gharlan: `overrideAt`

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
```diff
@@ -0,0 +1,60 @@
...
+        $tokens = Tokens::fromCode($content);
+
+        $echoTokens = $tokens->findGivenKind(T_ECHO);
+
+        foreach ($echoTokens as $echoIndex => $echoToken) {
+            $nextTokenIndex = $tokens->getNextMeaningfulToken($echoIndex);
+            $endTokenIndex = $tokens->getNextTokenOfKind($echoIndex, array(';'));
+            $canBeConverted = true;
+            for ($i = $nextTokenIndex; $i < $endTokenIndex; ++$i) {
+                if ($tokens[$i]->equalsAny(array(','))) {
```
gharlan: `equals`

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
```diff
@@ -0,0 +1,74 @@
...
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $printTokens = $tokens->findGivenKind(T_PRINT);
+
+        foreach ($printTokens as $printIndex => $printToken) {
+            $prevToken = $tokens[$tokens->getPrevMeaningfulToken($printIndex)];
+            if ($prevToken->equalsAny(array(
+                    [T_RETURN, 'return'],
```
gharlan: `array()`

Symfony/CS/Tests/Fixer/Contrib/PrintToEchoFixerTest.php
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+class PrintToEchoFixerTest extends AbstractFixerTestBase
```
keradus: all classes should be `final`, test classes should also be `@internal`

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
```diff
@@ -0,0 +1,74 @@
...
+                    [T_PLUS_EQUAL, '+='],
+                    [T_IS_EQUAL, '=='],
+                    [T_IS_IDENTICAL, '==='],
+                    '(',
+                    '=',
+                    ':',
+                    '.',
```
gharlan: The blacklist is not complete. Missing many operators etc.
I would use a whitelist instead: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/Symfony/PreIncrementFixer.php#L44

Symfony/CS/Tests/Fixer/Contrib/PrintToEchoFixerTest.php
Soullivaneuh: `final` key for both fixer and test classes?

Soullivaneuh: > test classes should also be `@internal`

I didn't find any example on tests classes.

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
Soullivaneuh: Thanks @gharlan, will take a look.

Soullivaneuh: @gharlan How would you manage the whitelist?

I mean, for most of case, the `print` function should be called with... nothing behind.

Maybe `Token` class has something like `isOperator`, will check that.

gharlan: I would use this:

```php
if (!$prevToken->equalsAny(array(';', '{', '}', array(T_OPEN_TAG)))) {
    continue;
}
```

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
```diff
@@ -0,0 +1,59 @@
...
+                    $canBeConverted = false;
+                    break;
+                }
+            }
+
+            if (false === $canBeConverted) {
+                continue;
+            }
+
+            $tokens[$echoIndex]->override(array(T_PRINT, 'print'));
```
gharlan: You should use `$tokens->overrideAt()` instead.

gharlan: (On 1.x there is no difference, but on 2.x there is: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tokenizer/Tokens.php#L852-L856)

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
Soullivaneuh: Found a simpler solution.

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
Soullivaneuh: Done.

Soullivaneuh:
This should be OK now.

Symfony/CS/Tests/Fixer/Contrib/PrintToEchoFixerTest.php
GrahamCampbell: yeh - it's only for new classes on 1.x. Everything is final and internal on 2.x.

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
gharlan: Hmm, why did you remove `{` and `}`?

Examples:

```php
if ($foo) {
    print 'foo';
}
print 'bar';
```

Prev tokens are `{` and `}`, and the print statements should be converted.

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
```diff
@@ -0,0 +1,59 @@
...
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $echoTokens = $tokens->findGivenKind(T_ECHO);
+
+        foreach ($echoTokens as $echoIndex => $echoToken) {
+            $nextTokenIndex = $tokens->getNextMeaningfulToken($echoIndex);
+            $endTokenIndex = $tokens->getNextTokenOfKind($echoIndex, array(';'));
```
gharlan: this is valid (without `;`):

```php
<div><?php echo 'foo' ?>
```

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
gharlan: Please add it as test case.

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
Soullivaneuh: This test pass without modification. I'll add it.

Symfony/CS/Fixer/Contrib/PrintToEchoFixer.php
Soullivaneuh: Thanks! Will correct it.

Soullivaneuh:
@gharlan review corrected.

gharlan:
Please add this test case:

```php
echo foo(1, 2);
```

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
gharlan: Yeah, but the for-loop should stop at `?>`.

Soullivaneuh:
Well, not working for `echo` to `print`.

Soullivaneuh:
And... fixed! :-)

gharlan:
Ok, next test case:

```php
echo ['foo', 'bar', 'baz'][$x];
```
;)

Soullivaneuh:
Ha ha ha ha... stop playing with me.

:heart:

gharlan:
I would fix these cases with `$tokens->findBlockEnd()`.

gharlan:
Another test case:

```php
<?php echo 'foo' ?>...<?php echo 'bar', 'baz' ?>
```

First `echo` should be converted.

I guess it will fail because of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1523#discussion_r45595696

Symfony/CS/Fixer/Contrib/EchoToPrintFixer.php
Soullivaneuh: Don't understand well. Please paste me a test that you think it will be buggy.

gharlan: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1523#issuecomment-158945664

```php
<?php echo 'foo' ?>...<?php echo 'bar', 'baz' ?>
```

keradus: don't read whole logic, but sth like:
`<div><?php echo XXX ?></div><div><?php echo YYY; ?></div>`
under `XXX` put echo statement that may be changed, and unde `YYY` put echo statement that can't be changed. Please notice that after `XXX` there are no `;`, and it is after `YYY`.
The current `getnextTokenOfKind` call will match the `;` after `YYY`, which will have an influence if the first `echo` could be converted or no

keradus: poor kerad, you should be silence next time ;)

Symfony/CS/Tests/Fixer/Contrib/EchoToPrintFixerTest.php
```diff
@@ -0,0 +1,124 @@
...
+                '<?php
+                echo foo(1, 2);
+                ',
+            ),
+            array(
+                '<?php
+                echo ["foo", "bar", "baz"][$x];
+                ',
+                '<?php
+                print ["foo", "bar", "baz"][$x];
```
gharlan: wrong order of expected and input

Symfony/CS/Tests/Fixer/Contrib/EchoToPrintFixerTest.php
```diff
@@ -0,0 +1,124 @@
...
+                ',
+            ),
+            array(
+                '<?php
+                <?php echo "foo" ?>...<?php echo "bar", "baz" ?>
+                ',
+                '<?php
+                <?php print "foo" ?>...<?php echo "bar", "baz" ?>
```
gharlan: wrong order of expected and input

keradus: and duplicated `T_OPEN_TAG`

gharlan:
The `EchoToPrintFixer` should run after `ShortEchoTagFixer`.

gharlan:
Another test case:

```php
print $foo ? 'foo' : 'bar';
```
(should be converted)

gharlan:
And this one (should pass without modification):

```php
@print foo();
```
Can not be converted, `@echo foo()` would result in a parse error.

Soullivaneuh:
> The `EchoToPrintFixer` should run after `ShortEchoTagFixer`.

Could you please elaborate why?

gharlan:
If someone uses both fixers the code should not change in a 2nd run. So the `ShortEchoTagFixer` (`<?=` to `echo`) has to run first

Soullivaneuh:
Oh ok I see. Any way to test it? ping @keradus

Soullivaneuh:
All tests should be OK now.

Please tell me if this is ready to merge, I'll squash and rebase in this case.

Soullivaneuh:
Tests passed, rebasing.

Soullivaneuh:
Done, RTM. :+1:

keradus:
> Oh ok I see. Any way to test it? ping @keradus

Yes.
Add test here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/FixerTest.php#L207
Also, you should add integration file here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/1.11/Symfony/CS/Tests/Fixtures/Integration

README.rst
```diff
@@ -611,6 +616,10 @@ Choose from the list of available fixers:
                         @var should always be written
                         as @type.
 
+* **print_to_echo** [contrib]
```
keradus: I bet it should be on Symfony level.

gharlan: :+1: 

[`print`](https://github.com/symfony/symfony/search?utf8=✓&q=print) vs. [`echo`](https://github.com/symfony/symfony/search?utf8=✓&q=echo&type=Code)

Soullivaneuh: Will do this.

Soullivaneuh:
@keradus print_to_echo fixer moved and priority tests added.

keradus:
Please, fix the Travis ;)

keradus:
hhvm could be failing due to ignoring short_open_tag

Soullivaneuh:
@keradus Well, I don't why the f*ck this is not working on 5.3... any idea?

keradus:
nope, but didn't investigate it, sorry, no time now :(

Soullivaneuh:
For HHVM, it seems to be failing because "PHP short open tags are not enabled.".

How to enable it properly on Travis?

keradus:
for hhvm see https://github.com/keradus/PHP-CS-Fixer/commit/766cb0ab07986c0196fd15409b727266827988e0
(from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1465)

keradus:
Also https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/Contrib/ShortEchoTagFixerTest.php#L31-L33

Soullivaneuh:
So why should I do? Waiting your fix or simply ignore test with HHVM?

keradus:
My PR is not a fix, it removes some workarounds of HHVM bugs. HHVM must fix that bug. They probably did it but on HHVM version not supported by Travis yet.
My PR is to keep bad hhvm-related code tracked and remove it when possible.

keradus:
yes, you should skip that one integration test on hhvm

Soullivaneuh:
Well, HHVM failed on **integration** test.

I don't know where to put the condition.

keradus:
Yes, on integration, on simple utests it is already skipped for short_echo fixer.

There is no such a place now, I think that you need, sadly, add the optional section in integration file and code to handle it :(

Soullivaneuh:
@keradus I tried a quick implementation to make a generic ignore of some integration files.

I don't think we need more, this is just a hack waiting HHVM fix.

Note: I tried with `HipHop VM 3.10.1 (rel)`, the bug with the short open tag is still here.

keradus:
Hacky... I don't like this... But I think we could accept it for now and later think for more general solution, if we have a need for it

keradus:
Then, the last thing is 5.3

Soullivaneuh:
> Then, the last thing is 5.3

For the moment, I clearly have no clue for this. :-(

Soullivaneuh:
If I comment the concerned assert, I get this:

```bash
$ php phpunit-old.phar Symfony/CS/Tests/IntegrationTest.php 
PHPUnit 4.8.18 by Sebastian Bergmann and contributors.

.F....

Time: 547 ms, Memory: 23.50Mb

There was 1 failure:

1) Symfony\CS\Tests\IntegrationTest::testIntegration with data set #1 ('echo_to_print,short_echo_tag.test', 'Integration of fixers: echo_t...o_tag.', array(Symfony\CS\Fixer\Contrib\ShortEchoTagFixer Object (), Symfony\CS\Fixer\Contrib\EchoToPrintFixer Object ()), '<div><?= "test" ?></div>\n', '<div><?php print "test" ?></div>\n')
Expected changes do not match result for "Integration of fixers: echo_to_print,short_echo_tag." in "echo_to_print,short_echo_tag.test".
--- Expected
+++ Actual
@@ @@
-<div><?php print "test" ?></div>
+<div><?= "test" ?></div>

/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Tests/AbstractIntegrationTest.php:165
/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Tests/IntegrationTest.php:48
/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Tests/AbstractIntegrationTest.php:77

FAILURES!
Tests: 6, Assertions: 20, Failures: 1.
```

Soullivaneuh:
@keradus If I remove the short open tag from the test file:

```
--INPUT--
<div><?php echo "test" ?></div>

--EXPECT--
<div><?php print "test" ?></div>
```

This work. Maybe related to `short_echo_tag` fixer?

Soullivaneuh:
I can confirmed it's related.

`short_echo_tag` fixer does not detect any `T_OPEN_TAG_WITH_ECHO` on my test content.

Soullivaneuh:
Content of $token in `short_echo_tag` while:

```php
<div><?= "test" ?></div>
```

Just this token, with hole code inside. WTH? :-)

Soullivaneuh:
@keradus I think we have a problem with `token_get_all` function on PHP 5.3. Please see my `test.php`:

```php
<?php

var_dump(token_get_all('<div><?= "test" ?></div>'));
```

And the result:

```bash
$ php test.php 
array(1) {
  [0]=>
  array(3) {
    [0]=>
    int(311)
    [1]=>
    string(24) "<div><?= "test" ?></div>"
    [2]=>
    int(1)
  }
}
```

What should I do? Ignoring this case for PHP 5.3? :trollface:

keradus:
check the php.ini: short_open_tag

keradus:
ShortOpenTagFixer works on 5.3 right? if so, the integration test should work too.

gharlan:
ShortOpenTagFixer yes, ShortEchoTagFixer no (if `short_open_tag` disabled on 5.3).

gharlan:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/Contrib/ShortEchoTagFixerTest.php#L23

Soullivaneuh:
Indeed, I enabled short_open_tag and this is working.

I think the best is to enable it on Travis. That will solve ShortEchoTagFixer and permit to remove `@requires` tag too.

Will investigate.

keradus:
or just made whole fixer to run only under 5.4+.
you can't know user's settings

Soullivaneuh:
Solution works on Travis.yml.

> or just made whole fixer to run only under 5.4+.

Yeah, but how? This is due to integration tests. It would make two hacking methods? That's quite ugly.

Soullivaneuh:
Will work to expends options on `--CONFIG--` section.

keradus:
> Will work to expends options on `--CONFIG--` section.

this is not a config. this would be a new section. If you wan't I can help with that on evening.

Soullivaneuh:
Well, if config is not config, what is the config? :-)

Anyway I almost finished my POC, I will let you review it.

keradus:
Config section is for things that will build your `Config` interface.
The things we need now is requirements that need to be fulfilled to run the integration test

Soullivaneuh:
Done @keradus, I let you appreciate it.

Symfony/CS/Tests/Fixtures/Integration/echo_to_print,short_echo_tag.test
```diff
@@ -0,0 +1,12 @@
+--TEST--
+Integration of fixers: echo_to_print,short_echo_tag.
+--CONFIG--
+level=none
+fixers=echo_to_print,short_echo_tag
+php=5.4
```
keradus: marker
(please ignore this comment)

keradus:
Could you make diff smaller? I can't see what changed in `getFixersFromConfig`.
Also, remember that this method is rewritten on 2.x line: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Test/AbstractIntegrationTestCase.php#L193

This (rewritting the method) is next reason why requirements should be moved to separated section (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1523/files#r45862969)

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -137,12 +146,21 @@ protected static function getTempFile()
      *
      * @param string           $testFileName Filename
      * @param string           $testTitle    Test title
+     * @param string[]         $config       Test file configuration
```
keradus: if I get it right, the config have only requirements ?


also, `$config['hhvm']` is compared to bool, so the typehinting here is wrong

Soullivaneuh: No, config has all keys -> values written on `--CONFIG--` section.

Soullivaneuh:
> Could you make diff smaller?

Not really, the diff is like that because I moved most of the logic from `getFixersFromConfig` to `getConfigFromSection`.

Soullivaneuh:
> This (rewritting the method) is next reason why requirements should be moved to separated section 

Oh I see. So what do you want? I propose a `--REQUIREMENTS--` or `--ENV--` section.

Soullivaneuh:
But if we add this section, we have to update all test files accordingly.

keradus:
Exactly, as I [said](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1523#issuecomment-159591243).

You can make this section optional, don't you ?

Soullivaneuh:
> You can make this section optional, don't you ?

Will try.

keradus:
the expected section is optional: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/AbstractIntegrationTest.php#L101

you can base on it

SpacePossum:
another pointer https://github.com/twigphp/Twig/blob/1.x/lib/Twig/Test/IntegrationTestCase.php#L90
(I based the first version of the integration tests here on the Twig one, the Twig one has an optional `condition`-block, don't know why I didn't add that one in the first place.. )

Soullivaneuh:
done @keradus.

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -270,4 +278,45 @@ protected function getFixersFromConfig($fileName, $config)
...
+     * Parses the '--REQUIREMENTS--' block of a '.test' file and determines requirements.
+     *
+     * @param string $fileName
+     * @param string $config
+     *
+     * @return array
+     */
+    protected function getRequirementsFromConfig($fileName, $config)
+    {
+        $requirements = array('hhvm' => true);
```
keradus: ?

keradus:
So much better now @Soullivaneuh
Thank you!

Symfony/CS/Tests/AbstractIntegrationTest.php
Soullivaneuh: I set default value for `hhvm` to avoid `isset` condition here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1523/files#diff-f387a0305e98fb7de24f5f8699ec1ec6R147

Soullivaneuh:
You're welcome! ;-)

Symfony/CS/Tests/AbstractIntegrationTest.php
keradus: And you didn't to if for PHP. it is not consistence

Soullivaneuh: Well, I can add it with `PHP_VERSION` constant.

keradus: or remove default value :)

Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -270,4 +278,45 @@ protected function getFixersFromConfig($fileName, $config)
...
+     * Parses the '--REQUIREMENTS--' block of a '.test' file and determines requirements.
+     *
+     * @param string $fileName
+     * @param string $config
+     *
+     * @return array
+     */
+    protected function getRequirementsFromConfig($fileName, $config)
+    {
+        $requirements = array('hhvm' => true, 'php' => PHP_VERSION);
```
keradus: php key requires to be at least with given version.
and... hhvm requires to run under hhvm? 

the documentation for possible requirements is needed

Soullivaneuh: Where should I put documentation? On docblock?

keradus: https://github.com/Soullivaneuh/PHP-CS-Fixer/blob/print-echo-fixer/Symfony/CS/Tests/AbstractIntegrationTest.php#L27-L43

Soullivaneuh: Done. :+1:

keradus: and still my question is without answer.

Soullivaneuh: > Tels which php minimum version is required and if hhvm is supported for the test.

What did you misunderstand?

keradus: reading only the docs I won't know what will happen when I set `hhvm=true` or `hhvm=false`

keradus: what if I don't set it at all? PHP requires PHP in at least given version, but hhvm is a setting not to require run under HHVM, but to allow run under it, then, the default value is not documented.

Soullivaneuh: I updated the doc. Sounds better?

Soullivaneuh:
Is this PR goot to merge?

keradus:
Thank you @Soullivaneuh.

Soullivaneuh:
You're welcome. :+1:

keradus:
btw, quite a nice work !

Soullivaneuh:
Thanks!

PR #1531, opened 2015-11-21
Update docs for few generic types

keradus:
We should start supporting generic types in docs.

connected: ref #1557

### DIFF
--- a/src/Fixer/ClassNotation/VisibilityRequiredFixer.php
+++ b/src/Fixer/ClassNotation/VisibilityRequiredFixer.php
@@ -185,7 +185,7 @@ class Sample
      * @param Tokens $tokens Tokens collection
      * @param int    $index  token index
      *
-     * @return array map of grabbed attributes, key is attribute name and value is array of index and clone of Token
+     * @return array<string, Token|null> map of grabbed attributes, key is attribute name and value is array of index and clone of Token
      */
     private function grabAttribsBeforeMethodToken(Tokens $tokens, $index)
     {
@@ -216,9 +216,9 @@ class Sample
      *
      * Token at given index is prepended by attributes.
      *
-     * @param Tokens $tokens      Tokens collection
-     * @param int    $memberIndex token index
-     * @param array  $attribs     map of grabbed attributes, key is attribute name and value is array of index and clone of Token
+     * @param Tokens                    $tokens      Tokens collection
+     * @param int                       $memberIndex token index
+     * @param array<string, Token|null> $attribs     map of grabbed attributes, key is attribute name and value is array of index and clone of Token
      */
     private function overrideAttribs(Tokens $tokens, $memberIndex, array $attribs)
     {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:-1: because we don't actually support them ourselves.

SpacePossum:
:+1: 
since PSR 5 does not provide an alternative for something like `array<int, string>`

GrahamCampbell:
> since PSR 5 does not provide an alternative

True. I'm :-1: only until I can actually get our fixers updated to support this stuff. Once the current phpdoc PRs have gone through, I can start working on it.

keradus:
@GrahamCampbell , that's why I extracted this into separate request and write:
> We should start supporting generic types in docs.

treat this as feature request.

GrahamCampbell:
k ;)

Symfony/CS/Fixer/PSR2/VisibilityFixer.php
```diff
@@ -61,9 +61,9 @@ public function getDescription()
      *
      * Token at given index is prepended by attributes.
      *
-     * @param Tokens $tokens  Tokens collection
-     * @param int    $index   token index
-     * @param array  $attribs array of token attributes
+     * @param Tokens                    $tokens  Tokens collection
+     * @param int                       $index   token index
+     * @param array<string, Token|null> $attribs array of token attributes
```
Soullivaneuh: Are you sure this is supported by all IDEs?

`string[]` and `Token[]` looks like to be more used and parsed.

keradus: `Token[]` is a list of Tokens. Flat one, without taking care of numbers. Here you don't have a list but a hash, with string as index. It is important.
`string[]` is bad as strings here are only the keys, `Tokens[]` is bad as (besides the keys matter) ignore the null possibility

Soullivaneuh: Oh yes, I didn't read the code correctly.

I found `Token[]` better but I agree, less complete. :-/

But this do not answer my question:

> Are you sure this is supported by all IDEs?


;-)

keradus: > > Are you sure this is supported by all IDEs?
>
> ;-)

Don't follow IDE notation, follow standardized notation!

[A short and incomplete list of IDEs](https://en.wikipedia.org/wiki/Comparison_of_integrated_development_environments#PHP) you may check. I could blindly bet - no, not supported by all IDEs, especially when not IDEs support PHPDoc itself.

SpacePossum:
please rebase, merge and :shipit:  :)

keradus:
it's not ready to be merged.
it's usage examples, but we do not really support them by our phpdoc fixers :(

SpacePossum:
>it's usage examples, but we do not really support them by our phpdoc fixers :(

do these break the tags? I would do this on 2.x and ship it ;)

keradus:
It may.
Before merging we should adjust docblock namespace to properly handle collection types (eg space inside type!) and then ensure all phpdoc fixers playing with type are still working properly

keradus:
thanks to #2858, rebased on 2.2

Issue #1480, opened 2015-11-02
Suggested new fixer: ShortCast


dereuromark:
A fixer we use as custom one, which is worth becoming a contrib one (as there is already a fixer for fixing `@return/param boolean/integer` to `bool/int`).
So this completes it for consistency:

```php
/**
 * Fixer ShortCast
 *
 * @author Mark Scherer
 */
class ShortCastFixer extends AbstractFixer
{

    /**
     * @var array
     */
    public static $matching = [
        '(boolean)' => '(bool)',
        '(integer)' => '(int)',
    ];

    /**
     * @param \SplFileInfo $file
     * @param string $content
     *
     * @return string
     */
    public function fix(\SplFileInfo $file, $content)
    {
        $tokens = Tokens::fromCode($content);
        foreach ($tokens as $index => $token) {
            // Don't use !!
            if ($token->getContent() === '!' && $tokens[$index - 1]->getContent() === '!') {
                $tokens[$index - 1]->setContent('');
                $tokens[$index]->setContent('(bool)');
                continue;
            }

            // Don't use long form
            if ($token->isCast()) {
                $content = $token->getContent();
                $key = strtolower($content);

                if (isset(self::$matching[$key])) {
                    $tokens[$index]->setContent(self::$matching[$key]);
                }
            }
        }

        return $tokens->generateCode();
    }

    /**
     * @return int
     */
    public function getPriority()
    {
        return -100;
    }

    /**
     * @return int
     */
    public function getLevel()
    {
        return FixerInterface::NONE_LEVEL;
    }

    /**
     * @return string
     */
    public function getDescription()
    {
        return 'Use short forms (bool) and (int) for casts and do not use !!.';
    }

}
```

GrahamCampbell:
Maybe we need the same deal with have with the phpdoc fixers, so "Integer" can be fixed to "int" too?

dereuromark:
And then extract that from the existing fixer? Because currently this is already fixed along with the other doc block changes. But I agree, it would be better to separate those two things.

keradus:
Please, could you send a PR?

GrahamCampbell:
Kind of hard to totally abstract, since the phpdoc fixer also includes fixes for sudo types, like "numeric", or "mixed".

dereuromark:
Wanted to first get a "OK, sounds useful" and maybe some preliminary feedback, like if the suggested name sounds OK. But then a PR should be possible very soon.

[...]

PR #1532, opened 2015-11-22
Added ShortBoolCastFixer

SpacePossum:
Early opening for `!!` to `(bool)` fixer,
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1480

Please let me know what (test) cases need more covering.

### DIFF
--- a/Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+class ShortBoolCastFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = count($tokens) - 1; $index > 1; --$index) {
+            if ($tokens[$index]->equals('!')) {
+                $index = $this->fixShortCast($tokens, $index);
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    private function fixShortCast(Tokens $tokens, $index)
+    {
+        for ($i = $index - 1; $i > 1; --$i) {
+            if ($tokens[$i]->equals('!')) {
+                $this->fixShortCastToBoolCast($tokens, $i, $index);
+                break;
+            }
+
+            if (!$tokens[$i]->isComment() && !$tokens[$i]->isWhitespace()) {
+                break;
+            }
+        }
+
+        return $i;
+    }
+
+    private function fixShortCastToBoolCast(Tokens $tokens, $start, $end)
+    {
+        for (; $start <= $end; ++$start) {
+            if (!$tokens[$start]->isComment()) {
+                $tokens[$start]->clear();
+            }
+        }
+
+        $tokens->insertAt($start, new Token(array(T_BOOL_CAST, '(bool)')));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Short cast bool using double exclamation mark should not be used.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before the SpacesCastFixer
+        return -9;
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
@@ -13,7 +13,7 @@ namespace Symfony\CS\Tests\Fixer\Contrib;
 
 use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
 
-class ShortBoolCastFixerTest  extends AbstractFixerTestBase
+class ShortBoolCastFixerTest extends AbstractFixerTestBase
 {
     /**
      * * @dataProvider provideCases
--- a/Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
@@ -1,0 +1,84 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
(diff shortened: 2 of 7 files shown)


GrahamCampbell:
What about the case `!(!$foo)`?

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
```diff
@@ -0,0 +1,45 @@
...
+class ShortBoolCastFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($i = count($tokens) - 1; $i > 1; --$i) {
+            if ($tokens[$i]->equals('!') && $tokens[$i - 1]->equals('!')) {
```
GrahamCampbell: what about comments in the way?

GrahamCampbell:
This fixer needs to be run before the fixer that adds a space after `(bool)`. That relation needs testing.

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
SpacePossum: good call,
```
<?php
$a = '0';
$b = !/*

    */!$a;

var_dump($b);
die;
```
is valid :/  who would write something like this ^-^

SpacePossum:
>What about the case !(!$foo)?

added to the test

>This fixer needs to be run before the fixer that adds a space after (bool). That relation needs testing.

priorities added

thanks for the feed back :+1:

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
GrahamCampbell: lol - there's always someone

SpacePossum: Not sure if should touch this code, i.e. remove the comment...
Maybe we should have a more generic fixer that removes comments from places where those make no sense, like in between operators and keyword.

gharlan: This should be touched by the fixer `! ! $a`

Symfony/CS/Tests/FixerTest.php
```diff
@@ -205,6 +205,7 @@ public function getFixersPriorityCases()
             array($fixers['php_unit_strict'], $fixers['php_unit_construct']),
             array($fixers['unary_operators_spaces'], $fixers['logical_not_operators_with_spaces']),
             array($fixers['unary_operators_spaces'], $fixers['logical_not_operators_with_successor_space']),
+            array($fixers['short_bool_cast'], $fixers['spaces_cast']),
```
keradus: it deserved integration test

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
SpacePossum: added, please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/599461876fa1be6a4de959d9ba6035c7ea455ea1#diff-4dae91f64e4ec75d2bf93f78a7b7deacR49

Symfony/CS/Tests/FixerTest.php
SpacePossum: if we can get this to Symfony level it will be more easy ;)

SpacePossum:
basic comment handling added, bucket list is an integration test I think?

Symfony/CS/Tests/FixerTest.php
keradus: Symfony integration test tests whole level.
The priority relation should be tested by integration test as well, it is not related with level

keradus:
please, rebase and add tests

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
```diff
@@ -0,0 +1,68 @@
...
+class ShortBoolCastFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($i = count($tokens) - 1; $i > 1; --$i) {
+            if ($tokens[$i]->equals('!')) {
```
dereuromark: it would be cleaner to return early (continue in this case) here instead of going to nest 5+ levels deep.

SpacePossum: thanks, reworked, rebased and squashed.

Symfony/CS/Fixer/Contrib/ShortBoolCastFixer.php
```diff
@@ -0,0 +1,76 @@
...
+        return $tokens->generateCode();
+    }
+
+    private function fixShortCast(Tokens $tokens, $index)
+    {
+        for ($i = $index - 1; $i > 1; --$i) {
+            if ($tokens[$i]->equals('!')) {
+                $this->fixShortCastToBoolCast($tokens, $i, $index);
+                break;
+            } elseif (!$tokens[$i]->isComment() && !$tokens[$i]->isWhitespace()) {
```
dereuromark: No elseif required after breaking, just a new line with if.

gharlan:
:+1:

dereuromark:
:+1:

gharlan:
IMO this can be in symfony level.

SpacePossum:
>IMO this can be in symfony level.

sounds good to me, what do you think @keradus (or should we ask someone over @ symfony?)

keradus:
Symfony indeed, sure of it

stof:
indeed, it can be at the Symfony level (even though I don't think we ever needed to apply such fix in Symfony itself)

keradus:
1 occurrence on 2.7 line, none on 2.8

Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+class ShortBoolCastFixerTest  extends AbstractFixerTestBase
```
stof: there is an extra space here

keradus: You see @SpacePossum , we really need the fixer of yours!

GrahamCampbell:
Just wondering... Would it be overkill to add De Morgan fixers? Does anyone have a preferred style for that sort of thing? `!$p && !$q` vs `!($q || $q)`?

keradus:
Please, don't go in that direction ;)

GrahamCampbell:
Haha. ;)

Symfony/CS/Tests/Fixer/Contrib/ShortBoolCastFixerTest.php
SpacePossum: yea I got so tired of doing wrong that I started on that fixer ^-^ 
fixed and move to Symfony level

keradus: Great!

// exactly, I often write 5-min "fixers" to do sth with whole codebase instead of sed'ing it

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks so much for merging @keradus : D  and thanks to all reviewers!

PR #1533, opened 2015-11-22
[master] Fix CS

SpacePossum:
After the last merges the CS of the Fixer itself is off.

### DIFF
--- a/Symfony/CS/Tests/Fixer/Symfony/MethodSeparationFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/MethodSeparationFixerTest.php
@@ -274,8 +274,7 @@ abstract class MethodTest3
     final private function method321a()
     {
     }
-}'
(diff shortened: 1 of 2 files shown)


Symfony/CS/Tests/Fixer/Symfony/MethodSeparationFixerTest.php
```diff
@@ -401,8 +400,7 @@ protected function getE(){echo 3;}
 private $a;
 
 function getF(){echo 4;}
-}'
-        , '<?php
+}', '<?php
```
keradus: I don't like that...

keradus:
Thank you for catching this. I will check all branches and fix what is needed.

keradus:
Thank you @SpacePossum

SpacePossum:
:+1: thank you @keradus :)

PR #1536, opened 2015-11-23
Typo fix

fabpot:


SpacePossum:
:+1:

keradus:
:+1:

Thanks @fabpot !

keradus:
Thank you @fabpot.

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -269,7 +269,7 @@ protected function configure()
 Create a build file to run ``php-cs-fixer`` on Travis. It's advisable to create a dedicated directory
 for PHP CS Fixer cache files and have Travis cache it between builds.
 
-    <?yml
+    <?yaml
```
stof: why is it ``<?`` ?

keradus: only internal marker for regexp

Issue #1522, opened 2015-11-20
Filename beginning with numbers produces internal error


tristanbettany:
This is an interesting one I came across today.

A filename beginning with numbers produces an internal error, remove the numbers and everything works fine. 

Maybe a more descriptive error in the internal error output could be handy? Took a while to find the cause of this.

Thanks :)

SpacePossum:
Hi,
Could you provide some more details. I tried to reproduce the problem, but failed to do so, see below.

```
possum@outerspace:~/work/PHP-CS-Fixer$ php php-cs-fixer fix --dry-run 123.php -vvv
Loaded config from "/home/possum/work/PHP-CS-Fixer/.php_cs"

Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[0.000 s] /home/possum/work/PHP-CS-Fixer/123.php

Fixed all files in 0.001 seconds, 5.500 MB memory used

possum@outerspace:~/work/PHP-CS-Fixer$ php php-cs-fixer fix --dry-run 123_abc.php -vvv
Loaded config from "/home/possum/work/PHP-CS-Fixer/.php_cs"
F
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) /home/possum/work/PHP-CS-Fixer/123_abc.php (eof_ending, header_comment)
Fixing time per file:
[0.055 s] /home/possum/work/PHP-CS-Fixer/123_abc.php

Fixed all files in 0.056 seconds, 6.250 MB memory used
````

can you tell which version of the fixer you use, which file name caused the problem (and maybe your PHP version as well?).

**Comment by tristanbettany** [comment]
~ $ php-cs-fixer -vvv fix ./015_test.php
E
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[0.480 s] /home/tristan.bettany/./015_test.php

Fixed all files in 0.483 seconds, 6.750 MB memory used

Files that were not fixed due to internal error:
   1) /home/tristan.bettany/./015_test.php

**Comment by tristanbettany** [comment]
Code Inside file:

```
<?php

namespace Fuel\Migrations;

class Test
{
  public function up()
  {
    try
    {
      echo 'Code';
    }
    catch (\Exception $e)
    {
      echo 'Error';
    }
  }
}
```

[...]

PR #1537, opened 2015-11-23
Psr0Fixer - ignore file without name or with name started by digit

keradus:
Fix for #1522

### DIFF
--- a/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
+++ b/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
@@ -152,7 +152,14 @@ class Psr0Fixer extends AbstractFixer implements ConfigAwareInterface
 
         $filenameParts = explode('.', $file->getBasename(), 2);
 
-        if (!isset($filenameParts[1]) || 'php' !== $filenameParts[1]) {
+        if (
+            // ignore file without name
+            '' === $filenameParts[0]
+            // ignore file with extension other than php
+            || (!isset($filenameParts[1]) || 'php' !== $filenameParts[1])
+            // ignore file started by digit (as class cannot be named that way)
+            || is_numeric($filenameParts[0][0])
+        ) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


Symfony/CS/Fixer/PSR0/Psr0Fixer.php
```diff
@@ -152,7 +152,14 @@ public function supports(\SplFileInfo $file)
...
         $filenameParts = explode('.', $file->getBasename(), 2);
 
-        if (!isset($filenameParts[1]) || 'php' !== $filenameParts[1]) {
+        if (
+            // ignore file without name
+            '' === $filenameParts[0]
+            // ignore file with extension other than php
+            || (!isset($filenameParts[1]) || 'php' !== $filenameParts[1])
+            // ignore file started by digit (as class cannot be named that way)
+            || is_numeric($filenameParts[0][0])
```
SpacePossum: Maybe `ctype_digit` would be better, since `is_numeric` will return true for hexadecimals for PHP 5 and not for PHP 7?

keradus: yeah, and we have already managed to remove ctype dep. ;)

SpacePossum:
LGTM :+1: 
of the record, how would this fixer handle other file names with characters not supported as a class name, like non-UTF-8 stuff?

keradus:
Please, give an example.

BTW
```bash
D:\GIT\fork
λ cat Żabka.php
<?php

class Żabka {}

D:\GIT\fork
λ php -l Żabka.php
No syntax errors detected in »abka.php
```

SpacePossum:
```php
class Psr0FixerTest extends AbstractFixerTestBase
{
    public function testIgnoreFileStartedByDigit()
    {
        $file = $this->getTestFile('$test.php');

        $expected = <<<'EOF'
<?php
namespace Aaa;
class Bar {}
EOF;

        $this->makeTest($expected, null, $file);
    }

```

I'm just trying this out here to be honest ;)

keradus:
@SpacePossum let us use same regex to validate classname as PHP does

SpacePossum:
sound like a solid idea @keradus :) 
A PSR4 fixer might benefit from this as well. Do you know where to find such regex, I don't but I'm happy to start looking.

GrahamCampbell:
I'll need to update my PSR4 PR with this too.

GrahamCampbell:
In, 2.x, maybe we need to ditch the PSR0 fixer, and just have a fixer to fix the namespace, and another to fix the filename.

keradus:
i like that thought

PR #1538, opened 2015-11-24
Added possibility to lint tests

gharlan:
Let us look how many test cases are invalid.

### DIFF
--- a/Symfony/CS/Tests/AbstractIntegrationTest.php
+++ b/Symfony/CS/Tests/AbstractIntegrationTest.php
@@ -17,6 +17,7 @@ use Symfony\CS\ErrorsManager;
 use Symfony\CS\FileCacheManager;
 use Symfony\CS\Fixer;
 use Symfony\CS\FixerInterface;
+use Symfony\CS\LintManager;
 
 /**
  * Integration test base class.
@@ -154,6 +155,12 @@ abstract class AbstractIntegrationTest extends \PHPUnit_Framework_TestCase
         }
         if (isset($requirements['php']) && version_compare(PHP_VERSION, $requirements['php']) < 0) {
             $this->markTestSkipped(sprintf('PHP %s (or later) is required.', $requirements['php']));
+        }
+
+        if (getenv('LINT_TEST_CASES')) {
+            $linter = new LintManager();
+            $lintProcess = $linter->createProcessForSource($input);
+            $this->assertTrue($lintProcess->isSuccessful(), $lintProcess->getOutput());
         }
 
         $errorsManager = new ErrorsManager();
--- a/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
+++ b/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
@@ -12,6 +12,7 @@
 namespace Symfony\CS\Tests\Fixer;
 
 use Symfony\CS\FixerInterface;
+use Symfony\CS\LintManager;
 use Symfony\CS\Tokenizer\Tokens;
 
 /**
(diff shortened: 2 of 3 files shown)


keradus:
I was thinking about that too ;)
Please, check the `$input` too.

gharlan:
PHP version | Invalid test cases
--- | ---
PHP 5.3 | 219
PHP 5.4 | 81
PHP 5.5 | 78
PHP 5.6 | 78
PHP 7.0 | 82

GrahamCampbell:
LOL

SpacePossum:
@gharlan could you add messages to the test so it is a bit more clear to the developers what failed?
If you agree with https://github.com/SpacePossum/PHP-CS-Fixer/commit/6c199933b39f87a74a97cab2ef5c0b3ef6cb0758 could you pick it on your branch? I'll drop the PR (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1545) and move on with the next bunch of tests that needs fixing :)
(en please rebase ;) )

gharlan:
@SpacePossum I will rebase when your PR is merged.

keradus:
good plan

SpacePossum:
:+1:  sounds good to me :)

gharlan:
Update:

PHP version | Invalid test cases
--- | ---
PHP 5.3 | 204
PHP 5.4 | 47
PHP 5.5 | 41
PHP 5.6 | 41
PHP 7.0 | 48

gharlan:
So, what is to do here?
Only linting on highest php version?

keradus:
OK, please lint only there, and when we handle it we will think about next step

SpacePossum:
shoud(n't) we lint the intergration test input as well?
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/AbstractIntegrationTest.php#L147)

GrahamCampbell:
> shoud(n't) we lint the intergration test input as well?

:+1:

gharlan:
> OK, please lint only there

How should I do this? hard coded check for php 7? environment variable? or..?

keradus:
What do you think about using an env for PHPUnit?

gharlan:
done. that way?

keradus:
Yeah, but please also add it to `phpunit.xml.dist` file with default value. Just to have listed env variables that are used inside phpunit tests.

gharlan:
> Yeah, but please also add it to phpunit.xml.dist file with default value.

done

> shoud(n't) we lint the intergration test input as well?

done

SpacePossum:
thanks! :+1:

gharlan:
I have a general question: When should I add my `@author` tag to a modified class and when not to?
Is there any rule?

keradus:
When you think that your contribution to the class is significant

gharlan:
At the moment we are linting: 

* input and expected of test cases
* input of integration tests

IMO we should either add the expected part of integration tests, or remove the expected part from the other test cases (when input is present).

By the way: I don't like the difference between integration tests and other test cases regarding input/expected. IMO we should change the integrations tests, so that the expected part is mandatory, and the input is optional. And we should swap input and expected (first expected, then input).

keradus:
> IMO we should either add the expected part of integration tests, or remove the expected part from the other test cases (when input is present).

We should check both.



> IMO we should change the integrations tests, so that the expected part is mandatory, and the input is optional. And we should swap input and expected (first expected, then input).

big :+1: for that, I actually trap into that few times!
Could you take care of it?

gharlan:
> Could you take care of it?

which branch?

keradus:
`AbstractIntegrationTest` is internal on 1.x so you could change it on 1.x

SpacePossum:
>And we should swap input and expected (first expected, then input).

does that mean swapping the parts in the fixtures file (i.e. `.test`) as well?

SpacePossum:
>input of integration tests

maybe lint the output as well?

gharlan:
yes

gharlan:
As #1594 is merged into 1.12, should I open a new pr for 1.12?

.travis.yml
```diff
@@ -12,6 +12,7 @@ matrix:
         - php: 5.6
           env: SYMFONY_VERSION="~3.0@dev" COMPOSER_FLAGS="--prefer-stable"
         - php: 7.0
+          env: LINT_TEST_CASES=1
```
keradus: I think that for now we could remove that line, merge the branch (so no changes for Travis) and then, when we get all test base valid, we will add it again. What do you think?

gharlan: good plan :+1: 
line removed.

keradus:
Thank you @gharlan.

gharlan:
@keradus there is a conflict with #1605.

keradus:
Could you ref it?

PR #1541, opened 2015-11-25
Psr0Fixer - Ignore filenames that are a reserved keyword or predefined constant

SpacePossum:
The PSR0 fixer should not try to fix files that are named a predefined constant or keyword,
like `interface.php`

### DIFF
--- a/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
+++ b/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
@@ -161,6 +161,11 @@ class Psr0Fixer extends AbstractFixer implements ConfigAwareInterface
             return false;
         }
 
+        $tokens = Tokens::fromCode(sprintf('<?php %s {}', $filenameParts[0]));
+        if ($tokens[1]->isKeyword() || $tokens[1]->isMagicConstant()) {
+            return false;
+        }
+
         // ignore stubs/fixtures, since they are typically containing invalid files for various reasons
         return !preg_match('{[/\\\\](stub|fixture)s?[/\\\\]}i', $file->getRealPath());
     }
--- a/Symfony/CS/Tests/Tokenizer/TokenTest.php
+++ b/Symfony/CS/Tests/Tokenizer/TokenTest.php
@@ -174,11 +174,45 @@ class TokenTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($this->getBraceToken()->isKeyword());
     }
 
-    public function testIsMagicConstant()
+    /**
+     * @param int    $tokenId
+     * @param string $content
+     * @param bool   $isConstant
+     *
+     * @dataProvider provideMagicConstantCases
+     */
+    public function testIsMagicConstant($tokenId, $content, $isConstant = true)
     {
-        $this->assertFalse($this->getBraceToken()->isMagicConstant());
-        $token = new Token(array(T_CLASS_C, '__CLASS__'));
-        $this->assertTrue($token->isMagicConstant());
+        $token = new Token(array($tokenId, $content));
+        $this->assertSame($isConstant, $token->isMagicConstant());
+    }
+
+    public function provideMagicConstantCases()
+    {
+        $cases = array(
+            array(T_CLASS_C, '__CLASS__'),
+            array(T_DIR, '__DIR__'),
+            array(T_FILE, '__FILE__'),
+            array(T_FUNC_C, '__FUNCTION__'),
+            array(T_LINE, '__LINE__'),
+            array(T_METHOD_C, '__METHOD__'),
+            array(T_NS_C, '__NAMESPACE__'),
+        );
+
+        if (defined('T_TRAIT_C')) {
+            $cases[] = array(T_TRAIT_C, '__TRAIT__');
+        }
+
+        foreach ($cases as $case) {
+            $cases[] = array($case[0], strtolower($case[1]));
+        }
+
+        foreach (array($this->getForeachToken(), $this->getBraceToken()) as $token) {
+            $cases[] = array($token->getId(), $token->getContent(), false);
+            $cases[] = array($token->getId(), strtolower($token->getContent()), false);
+        }
+
+        return $cases;
     }
 
     /**
--- a/Symfony/CS/Tests/Tokenizer/TokenTest.php
+++ b/Symfony/CS/Tests/Tokenizer/TokenTest.php
@@ -174,6 +174,13 @@ class TokenTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($this->getBraceToken()->isKeyword());
     }
 
+    public function testIsMagicConstant()
+    {
+        $this->assertFalse($this->getBraceToken()->isMagicConstant());
+        $token = new Token(array(T_CLASS_C, '__CLASS__'));
+        $this->assertTrue($token->isMagicConstant());
+    }
+
     /**
      * @dataProvider provideIsNativeConstantCases
      */
--- a/Symfony/CS/Tokenizer/Token.php
+++ b/Symfony/CS/Tokenizer/Token.php
@@ -234,17 +234,16 @@ class Token
     }
 
     /**
-     * Generate keywords array contains all keywords that exists in used PHP version.
+     * Generate array containing all keywords that exists in PHP version in use.
      *
-     * @return array
+     * @return array<int, int>
      */
     public static function getKeywords()
     {
         static $keywords = null;
 
         if (null === $keywords) {
-            $keywords = array();
-            $keywordsStrings = array('T_ABSTRACT', 'T_ARRAY', 'T_AS', 'T_BREAK', 'T_CALLABLE', 'T_CASE',
+            $keywords = self::getTokenKindsForNames(array('T_ABSTRACT', 'T_ARRAY', 'T_AS', 'T_BREAK', 'T_CALLABLE', 'T_CASE',
                 'T_CATCH', 'T_CLASS', 'T_CLONE', 'T_CONST', 'T_CONTINUE', 'T_DECLARE', 'T_DEFAULT', 'T_DO',
                 'T_ECHO', 'T_ELSE', 'T_ELSEIF', 'T_EMPTY', 'T_ENDDECLARE', 'T_ENDFOR', 'T_ENDFOREACH',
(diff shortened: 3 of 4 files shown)


Symfony/CS/Fixer/PSR0/Psr0Fixer.php
```diff
@@ -161,6 +160,12 @@ public function supports(\SplFileInfo $file)
             return false;
         }
 
+        $tokens = token_get_all(sprintf('<?php %s', $filenameParts[0]));
```
SpacePossum: if anyone knows a better way to check if a string is predefined constant or keyword in PHP, I'm happy to update :)

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tokenizer/Token.php#L241 may be helpful

SpacePossum: hmm afraid not so, it provides the `[int] type/kind - [int] value`, but I need to check against the string as used in the code itself.

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
```diff
@@ -161,6 +160,12 @@ public function supports(\SplFileInfo $file)
...
+        if ($tokens[1][0] !== T_STRING) {
```
keradus: btw, `isGivenKind`

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
```diff
@@ -161,6 +160,12 @@ public function supports(\SplFileInfo $file)
...
+            // do not allow PHP keywords as classy name
```
keradus: you check here not only keywords but also predefined constants

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
keradus: In tests you have list of 65 keywords. but there is 66 keywords. They are listed in method I referred to.
Create similar method to check if token is predefined constant.

Then, leave here that tokenizer call (just change it to `Tokens::fromCode` and in `if` below check `$tokens[1]->isKeyword()` and your new method `$tokens[1]->isPredefinedConstant()`

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
SpacePossum: yes,  otherwise the fixer will generate classes like this
```php
<?php 
// some name space

class __CLASS__ {

}
```
which is not valid php

SpacePossum: oh, the comment, check, I'll update

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
SpacePossum: :+1: 
should these be added to the list of predefined constants or do you want those on PHP7 only http://php.net/manual/en/reserved.other-reserved-words.php ?
btw. the same checks added to the class should be done for (parts of)  the namespace.

keradus: i think they should

SpacePossum:
lets see what Travis thinks, http://php.net/manual/en/reserved.other-reserved-words.php is still a TODO

Symfony/CS/Tokenizer/Transformer/ArrayTypehint.php
```diff
@@ -28,8 +28,11 @@ public function process(Tokens $tokens)
     {
         foreach ($tokens->findGivenKind(T_ARRAY) as $index => $token) {
             $nextIndex = $tokens->getNextMeaningfulToken($index);
-            $nextToken = $tokens[$nextIndex];
+            if (null === $nextIndex) {
+                return;
```
keradus: when does this happen? new test is needed.

SpacePossum: The PSR0 fixer generates `<?php class array` and ask the tokenizer for tokens of that code, then this will break down.

Can be tested by adding;
```php
            array(
                '<?php class array',
                array()
            ),
```
on line https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Tokenizer/Transformer/ArrayTypehintTest.php#L38

Symfony/CS/Fixer/PSR0/Psr0Fixer.php
```diff
@@ -161,6 +160,11 @@ public function supports(\SplFileInfo $file)
             return false;
         }
 
+        $tokens = Tokens::fromCode(sprintf('<?php %s', $filenameParts[0]));
```
keradus: `<?php %s {}`

Symfony/CS/Tokenizer/Transformer/ArrayTypehint.php
keradus: how it behaves after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1541/files#r45877095 ?

SpacePossum: kind of works, however `'CT_ARRAY_TYPEHINT'` is not in https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tokenizer/Token.php#L247 so the psr0 test fails. Should the custom token not be in that list?

keradus: should be

GrahamCampbell:
I'd prefer to ignore all these cases on all PHP versions since we should be able to fix code that will run ok on later PHP versions.

SpacePossum:
> I'd prefer to ignore all these cases on all PHP versions since we should be able to fix code that will run ok on later PHP versions.

I prefer that as well, however since we don't know have the tokens kind constants on newer versions we cannot filter on those.

keradus:
Please fix Travis

SpacePossum:
all green now :)

Symfony/CS/Tokenizer/Token.php
```diff
@@ -367,6 +395,13 @@ public function isNativeConstant()
         return $this->isArray && in_array(strtolower($this->content), $nativeConstantStrings, true);
     }
 
+    public function isMagicConstant()
```
keradus: need to be tested

keradus: and documented

SpacePossum: please see https://github.com/SpacePossum/PHP-CS-Fixer/commit/6812475fa12d114e997b4497b37b965d12dba40e

gharlan: Please add test for `__class__`. Should work, but I want to be sure. ;)

SpacePossum: sure thing :)
https://github.com/SpacePossum/PHP-CS-Fixer/commit/85732838eb4aa261387316931ea487345f7d0cb5
this way both lower and upper case for all magic constants are tested.. lets see what travis thinks..

Symfony/CS/Tokenizer/Token.php
```diff
@@ -268,6 +268,34 @@ public static function getKeywords()
...
+    {
+        static $magicConstants = null;
+
+        if (null === $magicConstants) {
+            $magicConstants = array();
+            $magicConstantsStrings = array(
+                'T_CLASS_C', 'T_DIR', 'T_FILE', 'T_FUNC_C', 'T_LINE', 'T_METHOD_C', 'T_NS_C', 'T_TRAIT_C',
+            );
+
+            foreach ($magicConstantsStrings as $keywordName) {
```
keradus: not to be picky, but DRY, move that common part to private helper

keradus:
:+1:

Symfony/CS/Tokenizer/Token.php
SpacePossum: fair point, working on it :)

SpacePossum: done, https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/62d0a345ba54c0b9767357083b047514d4734400
diff looks but odd...  should I move the private method to another place in the class?

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/CONTRIBUTING.md#projects-standards

keradus: http://symfony.com/doc/current/contributing/code/standards.html#structure

SpacePossum: check, I can move it after the weekend, keep you posted :)

Symfony/CS/Tests/Tokenizer/TokenTest.php
```diff
@@ -174,6 +174,13 @@ public function testIsKeywords()
         $this->assertFalse($this->getBraceToken()->isKeyword());
     }
 
+    public function testIsMagicConstant()
+    {
+        $this->assertFalse($this->getBraceToken()->isMagicConstant());
+        $token = new Token(array(T_CLASS_C, '__CLASS__'));
+        $this->assertTrue($token->isMagicConstant());
```
gharlan: I would like to have the lowercase variant here, too.

GrahamCampbell: Once that's done, :+1: from me.

Symfony/CS/Tokenizer/Token.php
SpacePossum: is this OK like this https://github.com/SpacePossum/PHP-CS-Fixer/commit/96a7ac99d74f8a9504c1648101088937ba19972c?

Symfony/CS/Tests/Tokenizer/TokenTest.php
SpacePossum: added in https://github.com/SpacePossum/PHP-CS-Fixer/commit/96a7ac99d74f8a9504c1648101088937ba19972c with some more test cases.

keradus:
Thank you @SpacePossum.

Issue #1539, opened 2015-11-24
Feature request: echo/print/case without parentheses


gharlan:
similar to `IncludeFixer`: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/Symfony/IncludeFixer.php

```php
echo('foo');
print('foo');

switch ($foo) {
    case('bar'): 
        // ..
        break;
}
```

to:

```php
echo 'foo';
print 'foo';

switch ($foo) {
    case 'bar': 
        // ..
        break;
}
```

SpacePossum:
This would be 3 fixers I guess @keradus ?

Little note for myself for test cases:
```php
$a = 5.1;
$b = 1.0;
switch($a) {
    case (int) $a < 1 : {
        echo 'leave alone';
        break;
    }
    case ($a < 2)/* test */: {
        echo 'fix 1';
        break;
    }
    case (3) : {
        echo 'fix 2';
        break;
    }
    case /**/(/**/ // test
        4
        /**/)//
        /**/: {
        echo 'fix 3';
        break;
    }
    case (((int)$b) + 4.1): {
        echo 'fix 4';
        break;
    }
    case ($b + 1) * 2: {
        echo 'leave alone';
        break;
    }
}
```

keradus:
if `echo` and `print` would be the same then it could be one fixer for printing and one for `case`, if not - 3 indeed.

GrahamCampbell:
Maybe we need a fixer to remove all comments :P
To be honest, why do we need them. If the code is any good, phpdoc will do the trick. :P

Soullivaneuh:
> if echo and print would be the same then it could be one fixer for printing and one for case

I'm :+1: for a `echo`/`print` parenthesis fixer and one another for `case`.

Soullivaneuh:
PRs for this issue:

* https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1544
* https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1548

gharlan:
@Soullivaneuh thanks!

gharlan:
new feature request: same fixer for `clone`/`return`/`yield` :)

keradus:
the logic between #1544 and #1548 is the same (I didn't expect they will be that much similar), things that are different:
- isGivenKind argument
- disallowed token after block end

after some unification:

```php
        $loops = [];
[...]

PR #1544, opened 2015-11-25
Print Echo parentheses fixer

Soullivaneuh:
#1539 for echo/print.

I don't know if this is related to any PSR. I put it on Contrib for the moment.

### DIFF
--- a/Symfony/CS/Fixer/Contrib/PrintEchoParenthesesFixer.php
+++ b/Symfony/CS/Fixer/Contrib/PrintEchoParenthesesFixer.php
@@ -8,7 +8,7 @@ use Symfony\CS\Tokenizer\Tokens;
 /**
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  */
-class PrintEchoParenthesesFixer extends AbstractFixer
+final class PrintEchoParenthesesFixer extends AbstractFixer
 {
     /**
      * {@inheritdoc}
--- a/Symfony/CS/Fixer/Contrib/PrintEchoParenthesesFixer.php
+++ b/Symfony/CS/Fixer/Contrib/PrintEchoParenthesesFixer.php
@@ -17,6 +17,27 @@ class PrintEchoParenthesesFixer extends AbstractFixer
     {
         $tokens = Tokens::fromCode($content);
 
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind([T_PRINT, T_ECHO])) {
+                continue;
+            }
+
+            $nextTokenIndex = $tokens->getNextMeaningfulToken($index);
+            if (null === $nextTokenIndex || !$tokens[$nextTokenIndex]->equals('(')) {
+                continue;
+            }
+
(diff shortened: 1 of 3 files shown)


gharlan:
test case: `echo (1 + 2) . $foo;`

Soullivaneuh:
@gharlan Will add it. For information:

```php
$foo = 'bar';

echo (1 + 2) . $foo."\n";
echo 1 + 2 . $foo."\n";
```

Output is the same, so we can remove parentheses. But is that really wanted?

I mean, parentheses of the complete argument should be removed. I don't think the others is the job of this fixer.

Furthermore, you can have some more tricky edge cases:

```php
echo (1 + 2) * 10, "\n"; // Output "30"
echo 1 + 2 * 10, "\n";   // Output "21"
```

gharlan:
@Soullivaneuh I agree, I even didn't know that the result would be the same (in my test case without parantheses) :D

So it was meant as an example which should not be changed by this fixer.

gharlan:
More test cases:

```php
echo ($a ? $b : $c); // should be changed
echo ($a ? $b : $c) ? $d : $e; // should not be changed
echo 10 * (2 + 3); // should not be changed
echo ('foo'), ('bar'); // i'm not sure
```

Soullivaneuh:
```
echo ('foo'), ('bar'); // i'm not sure
```

@gharlan For me, this should not be changed. The signature does not have parentheses, this is OK.

Soullivaneuh:
PR ready for review and merge.

keradus:
fix the project itself

Symfony/CS/Fixer/Contrib/PrintEchoParenthesesFixer.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+class PrintEchoParenthesesFixer extends AbstractFixer
```
keradus: final

Symfony/CS/Tests/Fixer/Contrib/PrintEchoParenthesesFixerTest.php
```diff
@@ -0,0 +1,107 @@
+<?php
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+class PrintEchoParenthesesFixerTest extends AbstractFixerTestBase
```
keradus: final, internal

Soullivaneuh:
> fix the project itself

Ok. Do you think this fixer should be on any PSR? Or Symfony rules?

Symfony/CS/Tests/Fixer/Contrib/PrintEchoParenthesesFixerTest.php
Soullivaneuh: One day, I will remember...

keradus:
Closing in favour of #1550

PR #1549, opened 2015-11-26
Return parentheses fixer

Soullivaneuh:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1539#issuecomment-159876370

PR #1548, opened 2015-11-26
[PSR-2] Switch case parentheses fixer

Soullivaneuh:
#1539 for switch cases.

Related PSR-2 rule: http://www.php-fig.org/psr/psr-2/#5-2-switch-case

PR #1550, opened 2015-11-26
Added UnneededControlParenthesesFixer

### LINKED ISSUE #1539: Feature request: echo/print/case without parentheses

Soullivaneuh:
Closes #1539 

Closes #1544 
Closes #1548 
Closes #1549
 
Abstract class to avoid duplicated logic.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -525,6 +525,10 @@ Choose from the list of available fixers:
                         placed adjacent to their
                         operands.
 
+* **unneeded_control_parentheses** [symfony]
+                        Removes unneeded parentheses
+                        around control statements.
+
 * **unused_use** [symfony]
                         Unused use statements must be
                         removed.
--- a/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
@@ -1,0 +1,122 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class UnneededControlParenthesesFixer extends AbstractFixer
+{
+    /**
+     * To be removed when PHP support will be 5.5+.
+     *
+     * @var string[] List of statements to fix.
+     */
+    private static $controlStatements = array(
+        'switch_case',
+        'echo_print',
+        'return',
+        'clone',
+        'yield',
+    );
+
+    private static $loops = array(
+        'switch_case' => array('lookupTokens' => T_CASE, 'neededSuccessors' => array(';', ':')),
+        'echo_print' => array('lookupTokens' => array(T_ECHO, T_PRINT), 'neededSuccessors' => array(';', array(T_CLOSE_TAG))),
+        'return' => array('lookupTokens' => T_RETURN, 'neededSuccessors' => array(';')),
+        'clone' => array('lookupTokens' => T_CLONE, 'neededSuccessors' => array(';', ':', ',', ')')),
+    );
+
+    /**
+     * Dynamic yield option set on constructor.
+     */
+    public function __construct()
+    {
+        // To be moved back on static when PHP support will be 5.5+
+        if (defined('T_YIELD')) {
+            self::$loops['yield'] = array('lookupTokens' => T_YIELD, 'neededSuccessors' => array(';', ')'));
+        }
+    }
+
+    /**
+     * @param array $controlStatements
+     */
+    public static function configure(array $controlStatements)
+    {
+        self::$controlStatements = $controlStatements;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        // Checks if specific statements are set and uses them in this case.
+        $loops = array_intersect_key(self::$loops, array_flip(self::$controlStatements));
+
+        foreach ($tokens as $index => $token) {
+            if (!$token->equals('(')) {
+                continue;
+            }
+
+            $blockStartIndex = $index;
+            $index = $tokens->getPrevMeaningfulToken($index);
+            $token = $tokens[$index];
+
+            foreach ($loops as $loop) {
+                if (!$token->isGivenKind($loop['lookupTokens'])) {
+                    continue;
+                }
+
+                $blockEndIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $blockStartIndex);
+                $blockEndNextIndex = $tokens->getNextMeaningfulToken($blockEndIndex);
+
+                if (!$tokens[$blockEndNextIndex]->equalsAny($loop['neededSuccessors'])) {
+                    continue;
+                }
+
+                if ($tokens[$blockStartIndex - 1]->isWhitespace() || $tokens[$blockStartIndex - 1]->isComment()) {
+                    $tokens[$blockStartIndex]->clear();
+                } else {
+                    // Adds a space to prevent broken code like `return2`.
+                    $tokens->overrideAt($blockStartIndex, array(T_WHITESPACE, ' '));
+                }
+                $tokens[$blockEndIndex]->clear();
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Removes unneeded parentheses around control statements.';
+    }
+
+    /**
+     * Should be run before trailing_spaces.
+     *
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return 30;
+    }
+}
--- a/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
@@ -16,6 +16,7 @@ use Symfony\CS\Tokenizer\Tokens;
 
 /**
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
 final class UnneededControlParenthesesFixer extends AbstractFixer
 {
@@ -90,12 +91,13 @@ final class UnneededControlParenthesesFixer extends AbstractFixer
                 }
 
                 if ($tokens[$blockStartIndex - 1]->isWhitespace() || $tokens[$blockStartIndex - 1]->isComment()) {
-                    $tokens[$blockStartIndex]->clear();
+                    $this->clearParenthesis($tokens, $blockStartIndex);
                 } else {
                     // Adds a space to prevent broken code like `return2`.
                     $tokens->overrideAt($blockStartIndex, array(T_WHITESPACE, ' '));
                 }
-                $tokens[$blockEndIndex]->clear();
+
+                $this->clearParenthesis($tokens, $blockEndIndex);
             }
         }
 
@@ -118,5 +120,24 @@ final class UnneededControlParenthesesFixer extends AbstractFixer
     public function getPriority()
     {
         return 30;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     */
+    private function clearParenthesis(Tokens $tokens, $index)
+    {
+        $tokens[$index]->clear();
+
+        if (
+            isset($tokens[$index - 1]) &&
+            isset($tokens[$index + 1]) &&
+            $tokens[$index - 1]->isWhitespace() &&
+            $tokens[$index + 1]->isWhitespace()
+        ) {
+            $tokens[$index - 1]->setContent($tokens[$index - 1]->getContent().$tokens[$index + 1]->getContent());
+            $tokens[$index + 1]->clear();
+        }
     }
 }
--- a/Symfony/CS/Tests/Fixer/Symfony/UnneededControlParenthesesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/UnneededControlParenthesesFixerTest.php
@@ -1,0 +1,391 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Symfony;
+
+use Symfony\CS\Fixer\Symfony\UnneededControlParenthesesFixer;
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ *
+ * @internal
+ */
+final class UnneededControlParenthesesFixerTest extends AbstractFixerTestBase
+{
+    private static $defaultStatements = null;
+
+    public static function setUpBeforeClass()
+    {
+        $controlStatementsProperty = new \ReflectionProperty('Symfony\CS\Fixer\Symfony\UnneededControlParenthesesFixer', 'controlStatements');
+        $controlStatementsProperty->setAccessible(true);
+        self::$defaultStatements = $controlStatementsProperty->getValue(null);
+    }
+
+    /**
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        // PHP <5.5 BC
+        if (version_compare(PHP_VERSION, '5.5', '<') && false !== strpos($input, 'yield')) {
+            $input = null;
+        }
+
+        // Default config. Fixes all statements.
+        UnneededControlParenthesesFixer::configure(self::$defaultStatements);
+        $this->makeTest($expected, $input);
+
+        // Empty array config. Should not fix anything.
+        UnneededControlParenthesesFixer::configure(array());
+        $this->makeTest($expected, null);
+
+        // Test with only one statement
+        foreach (self::$defaultStatements as $statement) {
+            UnneededControlParenthesesFixer::configure(array($statement));
+            $this->makeTest(
+                $expected,
+                $input && false !== strpos($input, $statement) ? $input : null
+            );
+        }
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                '<?php
+                yield "prod";
+                ',
+            ),
+            array(
+                '<?php
+                yield (1 + 2) * 10;
+                ',
+            ),
+            array(
+                '<?php
+                yield (1 + 2) * 10;
+                ',
+                '<?php
+                yield ((1 + 2) * 10);
+                ',
+            ),
+            array(
+                '<?php
+                yield "prod";
+                ',
+                '<?php
+                yield ("prod");
+                ',
+            ),
+            array(
+                '<?php
+                yield 2;
+                ',
+                '<?php
+                yield(2);
+                ',
+            ),
+            array(
+                '<?php
+                $a = (yield $x);
+                ',
+                '<?php
+                $a = (yield($x));
+                ',
+            ),
+            array(
+                '<?php
+                clone $object;
+                ',
+            ),
+            array(
+                '<?php
+                clone new Class();
+                ',
+            ),
+            array(
+                '<?php
+                clone $object;
+                ',
+                '<?php
+                clone ($object);
+                ',
+            ),
+            array(
+                '<?php
+                clone new Class();
+                ',
+                '<?php
+                clone (new Class());
+                ',
+            ),
+            array(
+                '<?php
+                foo(clone $a);
+                foo(clone $a, 1);
+                $a = $b ? clone $b : $c;
+                ',
+                '<?php
+                foo(clone($a));
+                foo(clone($a), 1);
+                $a = $b ? clone($b) : $c;
+                ',
+            ),
+            array(
+                '<?php
+                echo "foo";
+                print "foo";
+                ',
+            ),
+            array(
+                '<?php
+                echo (1 + 2) . $foo;
+                print (1 + 2) . $foo;
+                ',
+            ),
+            array(
+                '<?php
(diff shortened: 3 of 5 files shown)


Soullivaneuh:
I continue https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1539#issuecomment-159906171 here.

> Yes, the interface on 2.x line has changed, that's why it is MAJOR bump. There is really no need to create 5 fixers that do the same.

I was talking about 1.x with `HeaderComment`.

> Even if one keyword could be fixed on PSR and one not - then keep all PSR as default. Voila!

So all fixers on PSR2 folder?

keradus:
> I was talking about 1.x with HeaderComment.

Yes, and that method was removed on 2.x line.


> So all fixers on PSR2 folder?

if you didn't notice - I was still talking about _one_ fixer.

Soullivaneuh:
Ok so let's go for one fixer.

keradus:
Even now it's problematic - switch+case are together, print+echo too, but yield+return no, and clone is totally separated

Soullivaneuh:
@keradus done.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,71 @@
...
+final class UnneededControlParenthesesFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $loops = array(
```
keradus: Missing an option to configure which case user want to fix.

gharlan:
Missing test cases like:

```php
return($x);
case($x);
```

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
Soullivaneuh: Do you really want a option like `HeaderComment` does even if this system will be removed on master?

Personally, I would prefer adding options directly on master.

cc @GrahamCampbell for info.

Soullivaneuh:
Is `case($x);` valid? Oo

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
keradus: yes, I would like - to have logic handling the option in fixer already, and on master only change the setter method

gharlan:
https://3v4l.org/uDqGD

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
Soullivaneuh: Ok will work on it.

keradus: You can even mimic 2.x interface, just make configure method static

Soullivaneuh:
@keradus done.

Soullivaneuh:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/93369320#L424

Well, I don't know how to fix it.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1550/files#diff-88a34db2d4d0954ca3782d619b42b317R27
there is no such token on older PHP versoins

Soullivaneuh:
> there is no such token on older PHP versoins

Yes I know, this is why I set a `@requires` tag on fixer test.

This is not the problem. The problem is php-cs-fixer is ran at the end and fails on that.

How to fix/skip it?

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3d666f10aa4d1cc545877351dc6cbcf036b524b1/Symfony/CS/Fixer/PSR2/LowercaseConstantsFixer.php#L75-L77

Soullivaneuh:
Oh, indeed. :+1:

Soullivaneuh:
Well, my `T_YIELD` is on a static array. How should I properly set it with a condition?

Soullivaneuh:
Or a manual define outside the class.

Soullivaneuh:
Test passed.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,85 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\PSR2;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+// PHP <5.5 support
+if (!defined('T_YIELD')) {
```
keradus: Nope, do not do this. you will broke the behaviour of whole Tokenizer.

If you wan't to mock some constant you also need to mock its responsibility/functionality.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    private static $loops = array(
+        'switch_case' => array('lookupTokens' => T_CASE, 'neededSuccessors' => array(';', ':')),
+        'echo_print' => array('lookupTokens' => array(T_ECHO, T_PRINT), 'neededSuccessors' => array(';', array(T_CLOSE_TAG))),
+        'return' => array('lookupTokens' => T_RETURN, 'neededSuccessors' => array(';')),
+        'clone' => array('lookupTokens' => T_CLONE, 'neededSuccessors' => array(';')),
+        'yield' => array('lookupTokens' => T_YIELD, 'neededSuccessors' => array(';')),
+    );
+
+    public static function configure(array $controlStatements = null)
```
keradus: need to be covered by tests - what will be changed with given configuration. look at other configurable fixers.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
Soullivaneuh: Yes, but I can't dynamically define a static value.

What is your proposition? :-)

Soullivaneuh: I have an idea, but I'll have to store configuration too.

Soullivaneuh:
@keradus tests and code updated.

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,329 @@
...
+                '<?php
+                switch ($a) {
+                    case ("prod"):
+                        break;
+                }
+                ',
+            ),
+            array(
+                '<?php
+                case$x;
```
gharlan: IMO it should be `case $x` (same fore other keywords).
And please add test case for `case(2)`. `case2` would be invalid.

Soullivaneuh: > And please add test case for `case(2)`. `case2` would be invalid.

Indeed. In this case, I think I should add space if not present by default.

Soullivaneuh:
@gharlan Should be fixed now. :+1:

Soullivaneuh:
Tests passed, is this PR good to merge @keradus?

GrahamCampbell:
Might be an idea to rebase it to be 100% sure they do still pass.

keradus:
no need for review?

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,363 @@
...
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ *
+ * @internal
+ */
+final class UnneededControlParenthesesFixerTest extends AbstractFixerTestBase
+{
+    public function testConfigure()
+    {
+        $loopsProperty = new \ReflectionProperty('\Symfony\CS\Fixer\PSR2\UnneededControlParenthesesFixer', 'loops');
```
GrahamCampbell: No slash should be used on `Symfony\CS\Fixer\PSR2\UnneededControlParenthesesFixer`. PHP just trims it off under the hood.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        // To be moved back on static when PHP support will be 5.5+
+        if (defined('T_YIELD') && (empty(self::$controlStatements) || in_array('yield', self::$controlStatements, true))) {
```
keradus: calling this every time that `fix` method runs (so for everyfile) is bad

GrahamCampbell: Move it to the constuctor maybe?

Soullivaneuh:
> Might be an idea to rebase it to be 100% sure they do still pass.

Done.

> no need for review?

I think the fixer is OK but feel free! ;-)

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
Soullivaneuh: Who? loops are defined statically.

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,363 @@
...
+        $input = '<?php
+            echo ("foo");
+            case (true);
+            return (42);
+        ';
+
+        $this->makeTest($expected, $input);
+
+        UnneededControlParenthesesFixer::configure(array());
+        $loopsProperty->setValue($loops);
```
keradus: Changing the `loops` is bad.
It's internal for the fixer and you should not manipulate it.

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    /**
+     * @dataProvider provideFixCases
+     * @requires PHP 5.5
+     */
+    public function testFix($expected, $input = null)
```
keradus: This should be tested with configuration.
`testConfigure` didn't help.

see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L26

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
GrahamCampbell: This if statement is still evaluated though, each time,

GrahamCampbell: While it doesn't affect the asymptotic complexity, it does increase the runtime by a constant factor.

GrahamCampbell: I meant linearity of course there, constant factor is moving it to the constuctor. ;)

GrahamCampbell: So, it does affect the asymptotic complexity, lol...

I'll just shut up. :P

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
Soullivaneuh: This is for tests, if I don't do that the new configuration will be applied on classic tests.

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
Soullivaneuh: In your sample, you deactivate all options. How to test partially deactivation with that?

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
keradus: :+1: for constructor

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
keradus: Test sample where all options are active: github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L30-L35

Test sample where all options are inactive: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L38-L43

Test sample with only one option active (separately for each of them): https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L47-L53

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,363 @@
...
+
+        $this->makeTest($expected, $input);
+
+        UnneededControlParenthesesFixer::configure(array());
+        $loopsProperty->setValue($loops);
+    }
+
+    /**
+     * @dataProvider provideFixCases
+     * @requires PHP 5.5
```
gharlan: not all tests require php 5.5

gharlan:
More test cases:

```php
foo(clone($a));
$a = (yield($x));
```

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
keradus: then the classic tests should set configuration too, and all tests will use explicitly configuration

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
Soullivaneuh: Indeed, bad reading, thanks.

Soullivaneuh:
Thanks @gharlan for the relevant tests.

All reviews and tests are fixed. :+1:

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,104 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        // Checks if specific statements are set and uses them in this case.
+        $loops = empty(self::$controlStatements) ? self::$loops : array_intersect_key(self::$loops, array_flip(self::$controlStatements));
```
keradus: empty array shouldn't run all fixers

consider a case when your configure array is build dynamically and in some cases it computed as empty array to not fix anything

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,104 @@
...
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class UnneededControlParenthesesFixer extends AbstractFixer
+{
+    /**
+     * To be removed when PHP support will be 5.5+.
+     *
+     * @var array|null
+     */
+    private static $controlStatements = null;
```
keradus: what is the structure of it?
what are the default values (if I pass null to configure what will be fixed)?

Soullivaneuh:
@keradus Should be ok. `controlStatements` is always an array and an empty one will run nothing.

Symfony/CS/Tests/Fixer/PSR2/UnneededControlParenthesesFixerTest.php
```diff
@@ -0,0 +1,390 @@
...
+        // Default config. Fixes all statements.
+        UnneededControlParenthesesFixer::configure(self::$defaultStatements);
+        $this->makeTest($expected, $input);
+
+        // Empty array config. Should not fix anything.
+        UnneededControlParenthesesFixer::configure(array());
+        $this->makeTest($expected, null);
+
+        // Test with only some statements
+        $statements = array('switch_case', 'return');
```
keradus: Test sample with only one option active (separately for **each** of them): https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixer/Contrib/PhpUnitConstructFixerTest.php#L47-L53

Soullivaneuh: Done.

keradus:
Code looks nice for me ;)

Soullivaneuh:
> Code looks nice for me ;)

image

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
```diff
@@ -0,0 +1,110 @@
...
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        // Checks if specific statements are set and uses them in this case.
+        $loops = array_intersect_key(self::$loops, array_flip(self::$controlStatements));
+
+        foreach ($tokens as $index => $token) {
+            foreach ($loops as $loop) {
+                if (!$token->isGivenKind($loop['lookupTokens'])) {
```
gharlan: This could be optimized (avoiding to traverse `$loops` for every token).

Pseudo code:

```
foreach ($tokens) {
  if ($token != "(") {
    continue;
  }
  $prev = getPrevMeaningfulToken();
  foreach ($loops) {
    if (!$prev->isGivenKind($loop['lookupTokens'])) {
      continue;
    }
    // ..
  }
}
```

don't know if it's worth to change it.

gharlan:
Is it really PSR-2? I'm not sure.

Soullivaneuh:
> Is it really PSR-2? I'm not sure.

Is it surely for switch case. But since @keradus wants me for only one fixer, he said to put it on PSR.

In another side, it's consistent IMHO.

Symfony/CS/Fixer/PSR2/UnneededControlParenthesesFixer.php
Soullivaneuh: Done.

keradus:
I set to put it to PSR2 with proper default configuration

Soullivaneuh:
So do I have to change something on this PR or it's OK?

keradus:
Set [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1550/files) list to one matching PSR2

gharlan:
I would prefer to put this to symfony level with all types enabled.
In 2.x we can add some types to psr-2.

Soullivaneuh:
> I would prefer to put this to symfony level with all types enabled.
In 2.x we can add some types to psr-2.

I would :+1: for this solution.

keradus:
OK

keradus:
bug report, this should not happen:
```diff
-        return (
+        return 
             0 === strpos($method->name, 'get') &&
             3 < strlen($method->name) &&
             0 === $method->getNumberOfRequiredParameters()
-        );
+        ;
```

(the space after return should not be there)

GrahamCampbell:
Good catch @keradus.

keradus:
regress for 1.11 release ;P

Soullivaneuh:
@keradus Is that should be handled by this fixer? We have whitespace fixer for this AFAIK.

GrahamCampbell:
> Is that should be handled by this fixer? We have whitespace fixer for this AFAIK.

True. We just need to correct the fixer priorities IMO.

keradus:
yes and no.
The fixer should not fix issues that isn't it's responsibility.
But the fixer should not introduce new issues.
One may run only this one fixer - and if it fixes one issue and introduce other then it's not OK.

So if fixer need to remove the `(`, it should remove the space before brace if needed too.
But is should not care about spaces after return in all other places (places not touches by fixer)

GrahamCampbell:
> But the fixer should not introduce new issues.

Ah, yes, but who's to say that is an issue? That's only an issue by the definition that trailing whitespace is unwanted. Not everyone might have that definition, if you see what I mean? It's outside the scope of this fixer.

keradus:
Very well.
Then still need priority ensurance and integration test.

Soullivaneuh:
I added priority tests but I can't understand why it's failing.

Any idea?

gharlan:
Maybe a bug in `TrailingSpacesFixer`?

keradus:
nope

keradus:
Problem found by 2.x test (which, sadly, cannot be backported to 1.x easily).

Solved: https://github.com/Soullivaneuh/PHP-CS-Fixer/pull/2

Soullivaneuh:
@keradus Do we still need priority ensurance and tests with your fix?

Soullivaneuh:
> keradus modified the milestone: v1.12, v1.11 5 minutes ago

Too bad, we are closed to finish this PR. :-(

keradus:
Yes, my i fix bug in your fixer, but the need for tests the priority and integration are still in place.

Please, get rid of last commit (add author) (totally or squash it with first commit of yours).

I have started new release of 1.11. If Travis will soon be green for this PR I will incorporate it to 1.11.

Soullivaneuh:
> Please, get rid of last commit (add author) (totally or squash it with first commit of yours).

Like this?

keradus:
could be ;)

keradus:
Picky Travis is frozen ?

keradus:
please, rebase on 1.11, the branch is conflicted

Soullivaneuh:
@keradus done.

Seems Travis to be unfrozen! :smile:

Soullivaneuh:
Travis green, hurray ! :tada: 

And Windows still fetching repository... :trollface:

keradus:
Thank you @Soullivaneuh.

Soullivaneuh:
You are welcome. :+1:

PR #1245, opened 2015-06-05
Warn on xdebug, tweak "Fixed all files" message

SpacePossum:
Warn the user when xdebug is enabled and the verbosity level is `-vv` or up.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1239

Change "Fixed all files" message to;
"All filex fixed" or "All files checked" (when ran with --dry-run) 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/967

PR #1552, opened 2015-11-27
Warn when running with xdebug extension

SpacePossum:
Cant recall why I never finished this ^-^
Anyway following composer who warns it users about the impact of xdebug this will warn users in a similar way.
This time the warning is written to stderr (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1245#discussion_r31876821).


keradus:
Thank you @SpacePossum.

PR #1556, opened 2015-11-29
Enforce unix style line endings using .gitattributes

GrahamCampbell:
I also propose we export-ignore some stuff on 2.x, like, the massive tests folder.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf


GrahamCampbell:
NB, this is useful for windows users to ensure they don't accidentally use the wrong line endings.

keradus:
```
text=auto
Git will handle the files in whatever way it thinks is best. 
```

I don't like allowing GIT to guess the line ending and _change_ it.

GrahamCampbell:
We'll we'll notice if it's making bad changes, but so far, it's been a pain for me to develop on windows without it

keradus:
I develop on windows too and have just set `core.autocrlf=input` to _disallow_ GIT to manipulate line endings.
Also, all IDEs allows to set line ending type.

GrahamCampbell:
Hmmm. I guess we're not going to agree here then, lol.

keradus:
Why not? Please remember that I have already merged PRs that I wasn't happy about, but the community want it.

Also, what about `text eol=lf` ? Since we doesn't want git to guess (or at least I don't want), them we may told what the eol is. They are often used together, since first also detect which files are text (not binary), and second enforce correct eol

GrahamCampbell:
> text eol=lf

That's fine with me. ;)

GrahamCampbell:
How does that look now?

GrahamCampbell:
@keradus How does this look now?

AppVeyor seems stuck on https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-528 btw too.

keradus:
I'm neither :-1: nor :+1:
Let the others decide

gharlan:
±0 ;)

GrahamCampbell:
lol

SpacePossum:
seems like a tooling issue and not something caused/needs fixing by this project, but I would not mind having this in, so I'm +/-0 as well :)

keradus:
wel... anyone ?

GrahamCampbell:
Looks like we're going to have to assume everyone loves it! :trollface:

gharlan:
@GrahamCampbell +1
@keradus ±0
@gharlan ±0
@SpacePossum ±0

result: +1

keradus:
I wouldn't count author's vote, and if does then -1 from me.
We have a ±0 so far.

GrahamCampbell:
> I wouldn't count author's vote, and if would then -1 from me.

LOL. Hax.

sstok:
```
* text=auto

# Always use LF
core.autocrlf=lf
```

This should work perfect :+1: and configure the global config to use input.

keradus:
global config wouldn't be used in that case, why to change it?

What is the difference with the solution you provided and `* text eol=lf` ?

joecohens:
:+1:

keradus:
Hi @joecohens , nice to see you here! It's your first time, isn't it? I hope one day you will become a contributore of PHP CS Fixer as well as you are contributor of Style CI!

BTW, it would be great to see more vote from PHP CS Fixer contributor team ;) As it's will change (or not) the way you are working with repo.

joecohens:
Hi @keradus thanks for the welcome! I think it's my second time, for sure I'll see you more often :)

jbrooksuk:
:+1:

keradus:
ping @sstok ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1556#issuecomment-161060925

keradus:
ping @sstok ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1556#issuecomment-160975761

keradus:
hurrah, 3rd guy from StyleCI ;)
People, please, this is a settings for people who write the code in **this** repository, let **them** decide.

jbrooksuk:
> hurrah, 3rd guy from StyleCI ;)

We support one another, plus this makes sense for PHP-CS-Fixer anyway.

keradus:
Actually only folks from StyleCI want this, noone other see the benefits

GrahamCampbell:
> Actually only folks from StyleCI want this, noone other see the benefits

GrahamCampbell:
> Actually only folks from StyleCI want this, noone other see the benefits

Well, to be fair, it ultimately makes no difference for us. We can add it to our own version if we want.

I just think it makes it easier for people to contribute to PHP-CS-Fixer with this change.

keradus:
That's why I would like to have an opinion from people that actually contribute to PHP CS Fixer itself. And non of them actually follow that proposal. Maybe it is an issue of your env/global settings?
I don't say not, but it must serve the people who will use it.

@sstok proposal is really interesting too!

GrahamCampbell:
I'm fine using @sstok's idea instead. Both do the job.

sstok:
Setting the global configuration to "input" makes it work for all local repositories.
Setting the configuration locally ensures that someone with a "wrong" global configuration doesn't have to worry about wrong line-endings.

*All this hassle to support Windows, the only in the OS in the world that doesn't follow modern logic.*

keradus:
Thank you for the information @sstok .
I'm using the input global configuration for years and have no problem on Windows with line ending in git ;)

Changing the local is same as `text=auto eol=lf` in that case.

Symfony, Zend or PHPUnit do not use `.gitattributes` settings at all. Of course some repos use it, like Composer.
Take a look at [Composer .gitattributes](https://github.com/composer/composer/blob/master/.gitattributes). Maybe we should go that way?

Also, it may be tricky that `*.test` files will not follow all the rules :(

keradus:
Very well. Closing.

PR #1563, opened 2015-12-01
[WIP] Fix invalid PHP code samples in utests 

SpacePossum:
Early opening to keep an eye on the tests using Travis

### DIFF
--- a/Symfony/CS/Tests/Fixer/Symfony/MultilineArrayTrailingCommaFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/MultilineArrayTrailingCommaFixerTest.php
@@ -33,14 +33,34 @@ class MultilineArrayTrailingCommaFixerTest extends AbstractFixerTestBase
             array('<?php $x = array();'),
             array('<?php $x = array("foo");'),
             array('<?php $x = array("foo", );'),
-            array("<?php \$x = array(\n'foo',\n);", "<?php \$x = array(\n'foo'\n);"),
             array("<?php \$x = array('foo',\n);"),
-            array("<?php \$x = array('foo',\n);", "<?php \$x = array('foo'\n);"),
-            array("<?php \$x = array('foo', /* boo */\n);", "<?php \$x = array('foo' /* boo */\n);"),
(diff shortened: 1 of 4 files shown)


Symfony/CS/Tests/Fixer/Symfony/MultilineArrayTrailingCommaFixerTest.php
```diff
@@ -298,7 +260,78 @@ function foo(array $a)
...
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideExamples54()
+    {
+        return array(
+                // short syntax tests
+                array('<?php $x = array([]);'),
+                array('<?php $x = [[]];'),
```
keradus: funny thing that we actually _can_ handle short array even under PHP 5.4. And this is not the only fixer that do it perfectly.

SpacePossum: yea... while on that topic, are you sure this is the way to go? For example the short open tag fixers work just fine even when running without short tags being enabled..

keradus: The whole point of that fixer is to fix short open without need to have that flag enabled.

SpacePossum: in that case, we should not lint the input because it is should always fix... hmm catch 22...

keradus: or we should lint most input and only on high PHP version

SpacePossum: check, that would mean changing https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1538 I  guess? I proceed with this one when there is a conclusion about the test for tests :)

keradus: yep, it would be changes it.
most of changes, as in your prev PR, are good. but the short_open_tag fixer is great example to not change everything blindly.

SpacePossum:
when #1538 is updated, I'll start again.

PR #1567, opened 2015-12-02
PrintToEchoFixer - add to symfony rule set

gharlan:


keradus:
Good one.
Please, as you extend `@Symfony` set, could you extend `symfony` integration test as well?

gharlan:
Hmm, at the moment there is only a symfony integration test without any modification: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/Symfony/CS/Tests/Fixtures/Integration/symfony.test

I'm not sure how I should add test for PrintToEchoFixer

keradus:
add expect part ;)

gharlan:
This way?

keradus:
:+1:

SpacePossum:
:+1:

keradus:
Thank you @gharlan.

PR #1574, opened 2015-12-05
Add build and coverage status

dupuchba:
Add build and coverage status on the README.rst file to make the first page more "attractive".


keradus:
If we use all badges of tool we are using there should be about 3/4 more. But we do not use badges at all. Like all Symofny projects.

BTW, it is so great that you read the contributing file and know why is your PR failing on tests ;)

dupuchba:
Thx for your answer...

Issue #1564, opened 2015-12-01
Spacing of Tokens of a for-loop should be one space


SpacePossum:
```php
<?php
for($i=0;$i<4;--$i) {
    
}
```

to

```php
<?php
for($i = 0; $i < 4; --$i) {
    
}
```

http://www.php-fig.org/psr/psr-2/#5-4-for

for ref:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1518#discussion_r46253083

We need a fixer (fixing spaces around ";" in for-loop) for this ;)

gharlan:
I would also like if this is fixed:

```php
for ( $i = 0 ; $i < 4 ; --$i ) {}
```

SpacePossum:
agreed :)
working on a fixer at the moment

keradus:
Please, finish the opened one first ;)

SpacePossum:
>Please, finish the opened one first ;)

no worries, I'm on my max of the open PRs/issues I can handle right now ;)

SpacePossum:
When using `unalign_equals`, `spaces_before_semicolon` ,` parenthesis`
and updating the `OperatorsSpacesFixer` to fix to max one space fixing (with or without linebreak), we only need a fixer for `; --$i ` to `;--$i`.
So... does anyone know why the `OperatorsSpacesFixer` ensures one space but does not remove extra spacing?

What would be nice in the future is to move the logic of, for example `unalign_equals`, to a trait, that way we could create a new fixer, like a `loop` fixer, that is atomic (i.e. you don't need to run the `unalign_equals` on all the code), without code duplication. Just thinking out loud here ;)

keradus:
Loop fixer must not call `unalign_equals` on code itself. And remove few spaces only inside loop condition isn't hard.

> So... does anyone know why the OperatorsSpacesFixer ensures one space but does not remove extra spacing?

By design, since some don't want extra spaces to be removed.

SpacePossum:
>Loop fixer must not call `unalign_equals` on code itself. 

I figured ;)

>And remove few spaces only inside loop condition isn't hard.

Prop. not, but for adding spacing it is a bit more complex. My point is, the `unalign_equals` has the exact code I need, so I'll prop. copy paste it, which is not like DRY.. but since we cannot put that code in a trait the other option would be extending, which makes not much sense and would only work for the logic of one fixer.

>By design, since some don't want extra spaces to be removed.

[...]

PR #1575, opened 2015-12-05
[Feature, 1.11] Add 'for' semicolon fixer.

SpacePossum:
Add fixer for spacing around semicolons in a `for` loop

 @ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1564

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -207,6 +207,10 @@ Choose from the list of available fixers:
                         A file must always end with a
                         single empty line feed.
 
+* **for_loop_semicolon** [PSR-2]
+                        Spacing around semicolons in
+                        `for` control structures.
+
 * **function_call_space** [PSR-2]
                         When making a method or
                         function call, there MUST NOT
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
It'll end up in 1.12 btw. ;)

keradus:
as stated [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1564#issuecomment-161709858), there should be two fixers: one that remove spaces before `;`, one that add spaces after `;`, and both should not care is there is an `for` or not..

SpacePossum:
>one that add spaces after `;`, and `[...]` should not care is there is an for or not..

that one will create a lot of trailing whitespace.....

gharlan:
> one that remove spaces before `;`

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Fixer/Symfony/SpacesBeforeSemicolonFixer.php

> one that add spaces after `;`

that one is missing

> that one will create a lot of trailing whitespace.....

Add single space if there is no whitespace.

PR #1576, opened 2015-12-06
SpacesBeforeSemicolonFixer - do not remove space before semicolon if that space is after a semicolon

SpacePossum:
In order to create a fixer to add whitespace after a semicolon if needed, first this fixer should stop removing such space in order for the fixers not to become conflicting.

PR #1577, opened 2015-12-06
Added SpacesAfterSemicolonFixer

SpacePossum:
Add fixer for whitespace after semicolon fixing.

~~The changes in `Tokens` shows the problem if the fixer does not know if the semicolon is inside a `for` expression ;)~~

~~(ps. fixer is in conflict until https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1576 is resolved)~~

### DIFF
--- a/Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
+++ b/Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class SpacesAfterSemicolonFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = count($tokens) - 3; $index > 0; --$index) {
+            if (!$tokens[$index]->equals(';') || $tokens[$index + 2]->isComment()) {
+                continue;
+            }
+
+            if (!$tokens[$index + 1]->isWhitespace()) {
+                $tokens->insertAt($index + 1, new Token(array(T_WHITESPACE, ' ')));
+            } elseif (!$tokens[$index + 1]->equals(array(T_WHITESPACE, ' ')) && $tokens[$index + 1]->isWhitespace(array('whitespaces' => " \t"))) {
+                $tokens[$index + 1]->setContent(' ');
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Fix whitespace after a semicolon.';
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Symfony/SpacesAfterSemicolonFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/SpacesAfterSemicolonFixerTest.php
@@ -1,0 +1,229 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Symfony;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class SpacesAfterSemicolonFixerTest extends AbstractFixerTestBase
+{
+    /**
+     * @dataProvider provideCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideCases()
+    {
+        return array(
+            array(
+                '<?php
+                    test();
+                    $a; // test
+                ',
+            ),
+            array(
+                '<?php test();     ',
+            ),
+            array(
+                '<?php
+                    test();     // test
+                ',
+            ),
+            array(
+                '<?php test();       /* */ //',
+            ),
+            array(
+                '<?php
+                    test(); $a = 4;
+                ',
+                '<?php
+                    test();     $a = 4;
+                ',
+            ),
+            array(
+                '<?php
+                    test(); $b = 7;
+                ',
+                '<?php
+                    test();$b = 7;
+                ',
+            ),
+            array(
+                '<?php
+                    for (; ; ) {
+                    }
+                ',
+                '<?php
+                    for (;;) {
+                    }
+                ',
+            ),
+            array(
+                '<?php
+                    for (; ; ++$u1) {
+                    }
+                ',
+                '<?php
+                    for (;;++$u1) {
+                    }
+                ',
+            ),
+            array(
+                '<?php
+                    for (; $u2 < 0; ) {
+                    }
+                ',
+                '<?php
+                    for (;$u2 < 0;) {
+                    }
+                ',
+            ),
+            array(
+                '<?php
+                    for (; $u3 < 3; ++$u3) {
+                    }
+                ',
+                '<?php
+                    for (;$u3 < 3;++$u3) {
+                    }
+                ',
+            ),
+            array(
+                '<?php
+                    for ($u4 = 0; ; ) {
+                    }
+                ',
+                '<?php
+                    for ($u4 = 0;;) {
+                    }
+                ',
+            ),
+            array(
+                '<?php
+                    for ($u5 = 0; ; ++$u5) {
+                    }
+                ',
+                '<?php
+                    for ($u5 = 0;;++$u5) {
+                    }
+                ',
+            ),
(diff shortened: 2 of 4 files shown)


Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1196,14 +1196,14 @@ public function isBinaryOperator($index)
                 T_XOR_EQUAL => true,            // ^=
             );
             if (defined('T_POW')) {
-                $arrayOperators[T_POW] = true;          // **
-                $arrayOperators[T_POW_EQUAL] = true;    // **=
+                $arrayOperators[T_POW] = true; // **
+                $arrayOperators[T_POW_EQUAL] = true; // **=
```
gharlan: you should not remove spaces. only add one if it is missing.

SpacePossum: I'll update to check for a comment after the whitespace, otherwise this cannot be formatted to PSR2

```php
        for (;   $u3 < 3;   ++$u3) {
        }
```
to
```php
        for (; $u3 < 3; ++$u3) {
        }
```

gharlan: ok. didn't know that it is part of PSR-2.
But then the fixer should be in PSR-2 level, not symfony, right?

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,59 @@
...
+            if (!$tokens[$index]->equals(';') || $tokens[$index + 2]->isComment()) {
+                continue;
+            }
+
+            if (!$tokens[$index + 1]->isWhitespace()) {
+                $tokens->insertAt($index + 1, new Token(array(T_WHITESPACE, ' ')));
+                continue;
+            }
+
+            if ($tokens[$index + 1]->equals(' ')) {
```
gharlan: this is always false.

Symfony/CS/Tokenizer/Tokens.php
SpacePossum: Not sure, because the fixer works more generic now with the semicolons. For example it will do this which I don't think is in PSR2

```php
<?php
    test();     $a = 4;

```

```php
<?php
    test(); $a = 4;
                
```

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
SpacePossum: thank you update,
also added a test for the line break case

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,59 @@
...
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'There must be whitespace after a semicolon.';
```
keradus: it's tricky description
it won't put a space here:
```php
<?php
$f = 1;
```

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,59 @@
...
+            if (!$tokens[$index]->equals(';') || $tokens[$index + 2]->isComment()) {
+                continue;
+            }
+
+            if (!$tokens[$index + 1]->isWhitespace()) {
+                $tokens->insertAt($index + 1, new Token(array(T_WHITESPACE, ' ')));
+                continue;
+            }
+
+            if ($tokens[$index + 1]->getContent() === ' ') {
```
keradus: yoda

GrahamCampbell: image

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,59 @@
...
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+class SpacesAfterSemicolonFixer extends AbstractFixer
```
keradus: final

Symfony/CS/Tests/Fixer/Symfony/SpacesAfterSemicolonFixerTest.php
```diff
@@ -0,0 +1,169 @@
...
+ */
+
+namespace Symfony\CS\Tests\Fixer\Symfony;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author SpacePossum
+ */
+class SpacesAfterSemicolonFixerTest extends AbstractFixerTestBase
```
keradus: final, internal

Symfony/CS/Tests/Fixer/Symfony/SpacesAfterSemicolonFixerTest.php
```diff
@@ -0,0 +1,169 @@
...
+                    test(); $a = 4;
+                ',
+                '<?php
+                    test();$a = 4;
+                ',
+            ),
+            array(
+        '<?php
+        // 0 0 0
+        for (; ; ) {
```
keradus: TBH I get lost here. Maybe it would be more readable if we provide a separate test for each `for` notation?

keradus: And then no need for that `~binary` counting

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,59 @@
...
+            if (!$tokens[$index]->equals(';') || $tokens[$index + 2]->isComment()) {
+                continue;
+            }
+
+            if (!$tokens[$index + 1]->isWhitespace()) {
+                $tokens->insertAt($index + 1, new Token(array(T_WHITESPACE, ' ')));
+                continue;
+            }
+
+            if (' ' === $tokens[$index + 1]->getContent()) {
```
keradus: could be merged with first if.
Then, maybe the 2nd and 4th if could be merged and use `ensureWhitespaceAtIndex` method? (not sure about that multiline part!)

SpacePossum: Update the PR.
The `ensureWhitespaceAtIndex` is not going to work, since it does not check that when there is a whitespace the content is of the given one (one space in this case). So here it would not remove the whitespace that is to much :(

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,55 @@
...
+final class SpacesAfterSemicolonFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = count($tokens) - 3; $index > 0; --$index) {
```
keradus: `- 3` ?

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,55 @@
...
+            if (!$tokens[$index]->equals(';') || $tokens[$index + 2]->isComment() || ' ' === $tokens[$index + 1]->getContent()) {
```
keradus: do not use `getContent` for comparisons

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
```diff
@@ -0,0 +1,55 @@
...
+                continue;
+            }
+
+            if (!$tokens[$index + 1]->isWhitespace()) {
+                $tokens->insertAt($index + 1, new Token(array(T_WHITESPACE, ' ')));
+                continue;
+            }
+
+            if (false === strpos($tokens[$index + 1]->getContent(), "\n")) {
```
keradus: `->isWhitespace(array('whitespaces' => " \t"))`

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
SpacePossum: yes, 1 because it is the count, 1 because we look ahead one token in each iteration, and one because if the semicolon is the last of the tokens we don't want to add space after it anyway (`eof_ending` will take care of that)

Symfony/CS/Fixer/Symfony/SpacesAfterSemicolonFixer.php
SpacePossum: doesn't seem to work, in this case I'm looking for whitespaces that are:
- not a single space (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1577#discussion_r47026072)  
- do not contain a line break

I tried;

```php
if (!$tokens[$index + 1]->isWhitespace(array('whitespaces' => " \n"))) {
```

keradus: I gave you the solution to check is it multiline or not. check is it a single space must be done separately.

`" \n"` will check if the whitespace contains only `" "` and `"\n"`, eg `"       \n\n    \n    "` ;)

SpacePossum: : D
alrighty, let me know what you think of https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/48b98b6d65cde44b83c44cebaf257f3efece21cb

SpacePossum:
thanks for all pointer about the space check!

keradus:
Thank you @SpacePossum.

PR #1580, opened 2015-12-07
Added HeredocToNowdocFixer

gharlan:


Symfony/CS/Tests/Fixer/Contrib/HeredocToNowdocFixerTest.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideTestFixCases()
+    {
+        return array(
+            array(<<<'EOF'
+<?php $a = <<<'TEST'
+TEST;
```
stof: this test is useless. It is already covered by the later item (as ``makeTest`` always ensure that applying the fixer on the expected output triggers no additional change)

SpacePossum:
Test request

```php
            array(<<<'EOF'
<?php
$a = <<<TEST
'
TEST;

$a = <<<TEST
\$b = '<<<';
\$c = <=;
<<<'"$<<<'"
TEST;
EOF
            ),
```

```
1) Symfony\CS\Tests\Fixer\Contrib\HeredocToNowdocFixerTest::testFix with data set #0 ('<?php\n$a = <<<TEST\n'\nTEST;\n\n$...\nTEST;')
RuntimeException: Index invalid or out of range

/home/possum/work/PHP-CS-Fixer/Symfony/CS/Fixer/Contrib/HeredocToNowdocFixer.php:42
```

maybe I messed up the test?

gharlan:
> maybe I messed up the test?

yes. your code is invalid. it is missing a newline at end of code.
After end of heredoc/nowdoc there must be a newline (it is possible to put the newline after ";").

SpacePossum:
I see, re-tested, all :+1: thanks!

gharlan:
@stof could I move this fixer to symfony level?

gharlan:
ping, anyone else? :)

GrahamCampbell:
> should I move this fixer to symfony level?

Try running it on Symfony and see how many changes are made. If there are very few, or none, then yes. ;)

keradus:
If there are few also send a PR for lowest maintained branch

GrahamCampbell:
2.3, 2.7, 2.8, 3.0, master are the currently active branches on symfony atm btw. ;)

keradus:
yes, and CS fixes go from lowest.

GrahamCampbell:
Yeh, it does. I do sometimes wonder why they bother with CS on a bug-fix only branch, but meh, they do.

keradus:
To have diff between branches as little as possible.
Dont say it is good or not, just it is that way.

gharlan:
made a pr for symfony.

stof:
@keradus @GrahamCampbell the goal is to avoid having CS differences between branches, which would cause conflicts each time there is a change near such line which need to be merged into newer branches. We are merging maintained branches into each other on a weekly basis during the whole maintenance time (i.e. 3 years for LTS versions), so we don't want to have useless conflicts each time (which may lead to bad conflict resolution and introducing bugs or losing bug fixes in newer branches, as happened in the past already)

GrahamCampbell:
Sure. ;)

keradus:
Cool for me @stof, that is why I asked to send that PR to lowest branch ;)

Symfony/CS/Fixer/Contrib/HeredocToNowdocFixer.php
```diff
@@ -0,0 +1,74 @@
...
+
+            $content = $tokens[$index + 1]->getContent();
+            // regex: odd number of backslashes, not followed by dollar
+            if (preg_match('/(?<!\\\\)(?:\\\\{2})*\\\\(?![$\\\\])/', $content)) {
+                continue;
+            }
+
+            $this->convertToNowdoc($token);
+            $content = str_replace('\\\\', '\\', $content);
+            $content = str_replace('\\$', '$', $content);
```
keradus: one str_replace would be enough?

GrahamCampbell: indeed, you can pass arrays to param 1 and 2

Symfony/CS/Fixer/Contrib/HeredocToNowdocFixer.php
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            $content = $tokens[$index + 1]->getContent();
+            // regex: odd number of backslashes, not followed by dollar
+            if (preg_match('/(?<!\\\\)(?:\\\\{2})*\\\\(?![$\\\\])/', $content)) {
+                continue;
+            }
+
+            $this->convertToNowdoc($token);
+            $content = str_replace('\\\\', '\\', $content);
```
keradus: if this is needed to be done after `convertToNowdoc` call then convertion method result in invalid token

gharlan: No, there is no invalid token.

```php
<<<EOT
\\\\
EOT;

// after convertToNowdoc
// it is not invalid, but it is not the same content as before
<<<'EOT'
\\\\
EOT;

// after str_replace
// now it has the same meaning as before
<<<'EOT'
\\
EOT;
```

keradus: Then...
if this is needed to be done after convertToNowdoc call then convertion method result in  token with different meaning, therefore convertToNowdoc is corrupted (if you would have utests for that one method it will prove it)

keradus:
As one can see in referenced commits - Symfony level

GrahamCampbell:
Lol. Reference spam.

keradus:
It was for making sure we won't miss it ;P

Symfony/CS/Fixer/Contrib/HeredocToNowdocFixer.php
gharlan: The method just transforms a single `T_START_HEREDOC` token from heredoc to nowdoc notation. It is a low level method.

gharlan:
> As one can see in referenced commits - Symfony level

done

GrahamCampbell:
:+1:

keradus:
:+1:

keradus:
Thank you @gharlan.

SpacePossum:
nice!  :+1:  :)

PR #1583, opened 2015-12-07
Travis - update matrix

### LINKED ISSUE #1582: Travis: symfony versions
At the moment we are using this matrix:

PHP | Symfony
--- | ---
5.3 | 2.x (lowest)
5.4 | 2.8
5.5 | 3.0
5.6 | 2.8 and 3.0
7.0 | 3.0
hhvm | 3.0

Why are we running PHP 5.6 with Symfony 2.8 and 3.0, but PHP 5.5 and 7.0 only with Symfony 3.0?
It does not make sense to me.

keradus:
Closes #1582

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ matrix:
         - php: 5.4
         - php: 5.5
         - php: 5.6
-          env: SYMFONY_VERSION="~2.7"
-        - php: 5.6
-          env: SYMFONY_VERSION="~3.0@dev" COMPOSER_FLAGS="--prefer-stable"
         - php: 7.0
+          env: SYMFONY_VERSION="~2.8"
+        - php: 7.0
+          env: SYMFONY_VERSION="~3.0"
         - php: hhvm
 
 sudo: false


gharlan:
:+1:

GrahamCampbell:
What about testing php 5 with symfony 3?

gharlan:
> What about testing php 5 with symfony 3?

This is what we are doing.
Only 5.3 and 5.4 are running symfony 2 because symfony 3 requires php >= 5.5.

keradus:
PHP | Symfony
--- | ---
5.3 | 2.x (lowest)
5.4 | 2.8
5.5 | 3.0
5.6 | 3.0
7.0 | 2.8 and 3.0

GrahamCampbell:
Ahh, yes. What about hhvm and both symfony versions. :P

GrahamCampbell:
and, why does php 5.3 get the really old symfony version only?

keradus:
hhvm - I can't see any point of it. Even hhvm is extra old on Travis.
5.3 - look at composer flags for 5.3, it is on purpose to check if the tool will work under the min version of deps we required.

Soullivaneuh:
Maybe we can take profit to add SF 3.1 on allowed failures.

GrahamCampbell:
Bit overkill imo.

PR #1585, opened 2015-12-09
Code grooming - Improve utests code coverage

SpacePossum:
+ bit of SCA and typos fixes

### DIFF
--- a/Symfony/CS/Fixer/PSR2/LowercaseConstantsFixer.php
+++ b/Symfony/CS/Fixer/PSR2/LowercaseConstantsFixer.php
@@ -81,10 +81,6 @@ class LowercaseConstantsFixer extends AbstractFixer
             }
         }
 
-        if (null === $index) {
-            return true;
-        }
-
         $token = $tokens[$index];
 
         if ($token->equalsAny(array('{', '}'))) {
--- a/Symfony/CS/Fixer/Symfony/TernarySpacesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/TernarySpacesFixer.php
@@ -46,24 +46,24 @@ class TernarySpacesFixer extends AbstractFixer
                     }
                 } else {
                     // for `$a ? $b : $c` ensure space after `?`
-                    $this->ensureWhitespaceExistance($tokens, $index + 1, true);
+                    $this->ensureWhitespaceExistence($tokens, $index + 1, true);
                 }
 
                 // for `$a ? $b : $c` ensure space before `?`
-                $this->ensureWhitespaceExistance($tokens, $index - 1, false);
+                $this->ensureWhitespaceExistence($tokens, $index - 1, false);
 
                 continue;
             }
 
             if ($ternaryLevel && $token->equals(':')) {
                 // for `$a ? $b : $c` ensure space after `:`
-                $this->ensureWhitespaceExistance($tokens, $index + 1, true);
+                $this->ensureWhitespaceExistence($tokens, $index + 1, true);
 
                 $prevNonWhitespaceToken = $tokens[$tokens->getPrevNonWhitespace($index)];
 
                 if (!$prevNonWhitespaceToken->equals('?')) {
                     // for `$a ? $b : $c` ensure space before `:`
-                    $this->ensureWhitespaceExistance($tokens, $index - 1, false);
+                    $this->ensureWhitespaceExistence($tokens, $index - 1, false);
                 }
 
                 --$ternaryLevel;
@@ -81,7 +81,7 @@ class TernarySpacesFixer extends AbstractFixer
         return 'Standardize spaces around ternary operator.';
     }
 
-    private function ensureWhitespaceExistance(Tokens $tokens, $index, $after)
+    private function ensureWhitespaceExistence(Tokens $tokens, $index, $after)
     {
         $indexChange = $after ? 0 : 1;
         $token = $tokens[$index];
--- a/Symfony/CS/Tests/ApplicationTest.php
+++ b/Symfony/CS/Tests/ApplicationTest.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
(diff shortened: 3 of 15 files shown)


Symfony/CS/Fixer/PSR2/LowercaseConstantsFixer.php
```diff
@@ -81,10 +81,6 @@ private function isNeighbourAccepted(Tokens $tokens, $index)
             }
         }
 
-        if (null === $index) {
```
SpacePossum: is never true so can be removed

Symfony/CS/Fixer/Symfony/TernarySpacesFixer.php
```diff
@@ -81,7 +81,7 @@ public function getDescription()
         return 'Standardize spaces around ternary operator.';
     }
 
-    private function ensureWhitespaceExistance(Tokens $tokens, $index, $after)
+    private function ensureWhitespaceExistence(Tokens $tokens, $index, $after)
```
SpacePossum: just a little typo ;)

Symfony/CS/Tests/IntegrationTest.php
```diff
@@ -33,6 +33,6 @@ protected static function getFixturesDir()
      */
     protected static function getTempFile()
     {
-        return self::getFixturesDir().'/tmp/tmp.php';
+        return static::getFixturesDir().'/tmp/tmp.php';
```
SpacePossum: use late static binding so the method can be overridden

gharlan: You want to overwrite a final class? :)

SpacePossum: oh, ehmm no :)

I meant, it should call its own implementation of `getFixturesDir` and not the parent one (`AbstractIntegrationTestCase`).

SpacePossum:
fabbot.io fails because it tries to fix the tests cases by itself...

Symfony/CS/Tests/IntegrationTest.php
keradus: be we do not call it on parent-level, dont wee?

keradus:
@fabpot we have already released stable version of PHPUnit fixers.
Please, could you update fabbot.io to use them instead of regex ?

Symfony/CS/Tests/ApplicationTest.php
```diff
@@ -0,0 +1,26 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests;
+
+use Symfony\CS\Console\Application;
+
+/**
+ * @author SpacePossum
```
GrahamCampbell: invalid author tag

GrahamCampbell: Should be like this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/c784f68e77052939f594044f8d30e7ce1b564543/Symfony/CS/Tests/IntegrationTest.php#L17.

keradus: [it is valid](https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#examples-1)

GrahamCampbell: ok then, it's inconsistent with tags from this author already in the codebase

Symfony/CS/Tests/IntegrationTest.php
SpacePossum: indeed this was not needed, reverted (this one was needed https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1585/files#diff-f387a0305e98fb7de24f5f8699ec1ec6L92)

Symfony/CS/Tests/ApplicationTest.php
keradus: that's true ;)

Symfony/CS/Tests/ApplicationTest.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+class ApplicationTest extends \PHPUnit_Framework_TestCase
+{
+    public function testApplication()
+    {
+        $app = new Application();
+        $this->assertStringMatchesFormat('<info>PHP CS Fixer</info> version <comment>1.12-DEV</comment> by <comment>Fabien Potencier</comment>', $app->getLongVersion());
```
stof: this test is very fragile, as it must be updated for any release

Symfony/CS/Tests/DocBlock/AnnotationTest.php
```diff
@@ -238,4 +238,14 @@ public function testSetTypesOnBadTag()
 
         $tag->setTypes(array('string'));
     }
+
+    public function testGetTagsWithTypes()
+    {
+        $tags = Annotation::getTagsWithTypes();
+        $this->assertTrue(is_array($tags));
+        foreach ($tags as $tag) {
+            $this->assertTrue(is_string($tag));
```
stof: Use ``assertInternalType``

SpacePossum: thanks :)
updated

Symfony/CS/Tests/ToolInfoTest.php
```diff
@@ -0,0 +1,30 @@
...
+class ToolInfoTest extends \PHPUnit_Framework_TestCase
+{
+    public function testGetVersion()
+    {
+        $this->assertInternalType('string', ToolInfo::getVersion());
+    }
+
+    public function testIsInstallAsPhar()
+    {
+        $this->assertFalse(ToolInfo::isInstalledAsPhar());
```
keradus: does Composer could be checked ?

SpacePossum: not without faking all the logic that is the same as being tested, so it would not be a very useful test  :(

keradus: :(

keradus:
Thank you @SpacePossum.

PR #1591, opened 2015-12-12
UselessReturnFixer - Add

SpacePossum:
fabbot is wrong on this one, but a fix(er) for this is already PR'ed ;)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -702,6 +702,11 @@ Choose from the list of available fixers:
                         $strict param. Warning! This
                         could change code behavior.
 
+* **useless_return** [contrib]
+                        There should not be an empty
+                        return statement at the end of
+                        a function.
+
 
 The ``--config`` option customizes the files to analyse, based
 on some well-known directory structures:
--- a/Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
+++ b/Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
@@ -1,0 +1,108 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class UselessReturnFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $index = $tokens->getNextTokenOfKind($index, array(';', '{'));
+            if ($tokens[$index]->equals('{')) {
+                $this->fixFunction($tokens, $index, $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index));
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'There should not be an empty return statement at the end of a function.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before ReturnFixer, ExtraEmptyLinesFixer, WhitespacyLinesFixer and after EmptyReturnFixer and DuplicateSemicolonFixer.
+        return -18;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $start  Token index of the opening brace token of the function.
+     * @param int    $end    Token index of the closing brace token of the function.
+     */
+    private function fixFunction(Tokens $tokens, $start, $end)
+    {
+        for ($index = $end; $index > $start; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_RETURN)) {
+                continue;
+            }
+
+            $nextAt = $tokens->getNextMeaningfulToken($index);
+            if (!$tokens[$nextAt]->equals(';')) {
+                continue;
+            }
+
+            if ($tokens->getNextMeaningfulToken($nextAt) !== $end) {
+                continue;
+            }
+
+            $this->removeReturnStatement($tokens, $index, $nextAt);
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $start          Token index of the first token of the useless return.
+     * @param int    $semiColonIndex Token index of the semicolon token ending the useless return statement.
+     */
+    private function removeReturnStatement(Tokens $tokens, $start, $semiColonIndex)
+    {
+        $commentFound = false;
+        for ($index = $semiColonIndex; $index >= $start; --$index) {
+            if ($tokens[$index]->isComment()) {
+                $commentFound = true;
+                continue;
+            }
+
+            $tokens[$index]->clear();
+        }
+
+        // merge whitespace tokens if needed
+        if (!$commentFound && $tokens[$start - 1]->isWhitespace() && $tokens[$semiColonIndex + 1]->isWhitespace()) {
+            $tokens[$start - 1]->setContent($tokens[$start - 1]->getContent().$tokens[$semiColonIndex + 1]->getContent());
+            $tokens[$semiColonIndex + 1]->clear();
+        }
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Contrib/UselessReturnFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/UselessReturnFixerTest.php
@@ -1,0 +1,211 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class UselessReturnFixerTest extends AbstractFixerTestBase
+{
+    /**
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                '<?php
+    function b($b) {
+        if ($b) {
+            return;
+        }
+        /**/
+    }',
+                '<?php
+    function b($b) {
+        if ($b) {
+            return;
+        }
+        return /**/;
+    }',
+            ),
+            array(
+                '<?php
+    class Test2
+    {
+        private static function a($a)
+        {
+            if ($a) {
+                return;
+            }
+
+            $c1 = function() use ($a) {
+                if ($a)
+                    return;
+                if ($a > 1) return;
+                echo $a;
+                '.'
+            };
+            $c1();
+            '.'
+        }
+
+        private function test()
+        {
+            $d = function(){
+                echo 123;
+                '.'
+            };
+
+            $d();
+        }
+    }',
+                '<?php
+    class Test2
+    {
+        private static function a($a)
+        {
+            if ($a) {
+                return;
+            }
+
+            $c1 = function() use ($a) {
+                if ($a)
+                    return;
+                if ($a > 1) return;
+                echo $a;
+                return;
+            };
+            $c1();
+            return
+            ;
+        }
+
+        private function test()
+        {
+            $d = function(){
+                echo 123;
+                return;
+            };
+
+            $d();
+        }
+    }',
+            ),
+            array(
+                '<?php
+    function aT($a) {
+        if ($a) {
+            return;
+        }
+        '.'
+    }',
+                '<?php
+    function aT($a) {
+        if ($a) {
+            return;
+        }
+        return           ;
+    }',
+            ),
+            array(
+                '<?php return;',
+            ),
+            array(
+                '<?php
+    function c($c) {
+        if ($c) {
+            return;
+        }
+        //'.'
+    }',
+                '<?php
+    function c($c) {
+        if ($c) {
+            return;
+        }
+        return;//
+    }',
+            ),
+            array(
+                '<?php
+    class Test {
+
+        private static function d($d) {
+            if ($d) {
+                return;
+            }
+            }
+    }',
+                '<?php
+    class Test {
+
+        private static function d($d) {
+            if ($d) {
+                return;
+            }
+            return;}
+    }',
+            ),
+            array(
+                '<?php
+    interface FooInterface
+    {
+        public function fnc();
+    }',
+            ),
+            array(
+                '<?php
+    abstract class AbstractFoo
+    {
+        abstract public function fnc();
+        abstract public function fnc1();
+        static private function fn2(){}
+        public function fnc3() {
+            echo 1 . self::fn2();//{}
+        }
+    }',
+            ),
+            array(
+                '<?php
+    function foo () { }',
+            ),
+            array(
+                '<?php
+                function() {
+                     /**/
+                    '.'
+           /* a */   //
+                    '.'
+                }
+                ',
+                '<?php
+                function() {
+                    return  ; /**/
+                    return ;
+           /* a */  return; //
+                    return;
+                }
+                ',
+            ),
+        );
+    }
+}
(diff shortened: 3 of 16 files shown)


README.rst
```diff
@@ -659,6 +659,11 @@ Choose from the list of available fixers:
                         $strict param. Warning! This
                         could change code behavior.
 
+* **useless_return** [contrib]
```
GrahamCampbell: This should be symfony, right?

Symfony/CS/Tests/Fixtures/Integration/empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
```diff
@@ -0,0 +1,43 @@
...
+
+--EXPECT--
+<?php
+class Test {
+
+    private static function a($a) {
+        if ($a) {
+            return 1;
+        }
+
```
GrahamCampbell: this line should be removed, right?

GrahamCampbell: if the return is on it's own line, it would make sense to clear the whole line inc the newline char, right?

SpacePossum: The `extra_empty_lines` should pick this one up, we should add more config options to that one.
Which reminds me, this fixer should also run before the `return` fixer to prevent that one from adding an empty line.

GrahamCampbell: Yeh ;)

Symfony/CS/Tests/FixerTest.php
```diff
@@ -209,6 +209,9 @@ public function getFixersPriorityCases()
             array($fixers['short_bool_cast'], $fixers['spaces_cast']),
             array($fixers['unneeded_control_parentheses'], $fixers['trailing_spaces']), // tested also in: trailing_spaces,unneeded_control_parentheses.test
             array($fixers['class_definition'], $fixers['trailing_spaces']), // tested also in: class_definition,trailing_spaces.test
+            array($fixers['empty_return'], $fixers['useless_return']), // tested also in: empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
+            array($fixers['useless_return'], $fixers['whitespacy_lines']), // tested also in: empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
+            array($fixers['useless_return'], $fixers['extra_empty_lines']), // tested also in: empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
```
keradus: Please, also add integration test for only pair of fixers - one could run only 2 of them.

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,101 @@
...
+
+        // merge whitespace tokens if needed
+        if ($tokens[$index]->isWhitespace()) {
+            $content = $tokens[$index]->getContent();
+            for ($i = $index + 1; ; ++$i) {
+                if (!$tokens[$i]->isWhitespace()) {
+                    break;
+                }
+                $content .= $tokens[$i]->getContent();
+                $tokens[$i]->clear();
```
keradus: You need to merge 2 T_WHITESPACE being placed one after another. Where could be the situation with more than 2 ? I think this loop is not needed.

Symfony/CS/Tests/Fixer/Contrib/UselessReturnFixerTest.php
```diff
@@ -0,0 +1,107 @@
...
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+class UselessReturnFixerTest extends AbstractFixerTestBase
```
keradus: final

Symfony/CS/Tests/Fixer/Contrib/UselessReturnFixerTest.php
```diff
@@ -0,0 +1,107 @@
...
+{
+    /**
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideFixCases()
```
keradus: looking at the implementation it should work, but please add case with nested functions.

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $functionOpen = $tokens->getNextTokenOfKind($index, array('{'));
+            $this->fixFunction($tokens, $functionOpen, $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $functionOpen));
```
keradus: 1. iterate backward all tokens
2. for each T_FUNCTION iterate forward to find it's end
3. for each T_FUNCTION iterate backward from end to start

This means that for each function you are iterating through all its tokens... 3 times!
And or any 1st level nested function - 3x2, 2nd level nested function - 3x3 times and so on...

Here, you are not adding any tokens, only removing/merging. Then, you could Iterate forward on main loop and have no need to iterate few times over same tokens.

README.rst
```diff
@@ -659,6 +659,11 @@ Choose from the list of available fixers:
...
+                        These should not be an empty
```
keradus: `There should not (...)` ?

README.rst
SpacePossum: not sure, happy to move though :)

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,96 @@
...
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'There should not be an empty return statement at the end of a function.';
```
gharlan: why only at end of function, why not at end of file?

SpacePossum: I think that would make the fixer `risky`, for example:
```php
<?php
class Test {

    public function myTestFunction($a)
    {
        $b = $a;
        if ($a) {
            $b = include 'includeTest.php';
        }
        return $b;
    }
}

$test = new Test();
$b = $test->myTestFunction(true);
var_dump($b);
die;
```

`includeTest.php`

```php
<?php

if (time() < 0) {
    return 1;
}
return;
```

sure it isn't the most clean code IMHO, but it is valid and the fixer would break it.

gharlan: I thought that default return value of `include` is `null`. But it is wrong, default return value is `1`.

So I agree, would be risky and we should not change it.

keradus: Sadly, this is taken from perl. It is a confirmation that the module has been loaded :(

Please do not mess with file-level return statement in this one fixer. For sure it should not be handled in same fixer (non-risky + risky, no need to mark first part as risky as well).
Also, I don't think that it is needed at all. If one put empty return on non-function code then we could assume that he want to change that default `1` value ;)

gharlan: Yeah, we all agree here: nothing to change in this fixer and no need for another fixer.

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,96 @@
...
+        for ($index = $semiColonIndex; $index >= $start; --$index) {
+            if ($tokens[$index]->isComment()) {
+                continue;
+            }
+
+            $tokens[$index]->clear();
+        }
+
+        // merge whitespace tokens if needed
+        if ($tokens[$start - 1]->isWhitespace() && $tokens[$semiColonIndex + 1]->isWhitespace()) {
```
keradus: what if between `$start - 1` and `$semiColonIndex + 1` is that ommited comment ?

SpacePossum: good point, fixed :)

Symfony/CS/Tests/Fixer/Contrib/UselessReturnFixerTest.php
SpacePossum: thanks and done :)

README.rst
SpacePossum: ping @keradus, should I move it?

Symfony/CS/Tests/FixerTest.php
SpacePossum: not sure what such tests would add, the integration test already proves that changing any of the fixers priorities will cause troubles

keradus: Will be extra helpfull during any refactoring. It's always great to have more detailed tests as well, not only "big" one.

SpacePossum: check, added more tests :)

Symfony/CS/Tests/Fixtures/Integration/empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
```diff
@@ -0,0 +1,45 @@
...
+    private static function b($a) {
+        if ($a) {
+            return;
+        }
+
+        return null;
+    }
+}
+
+--EXPECT--
```
SpacePossum: @keradus I should swap the expected and input part right?

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
SpacePossum: please have another look, I took out some unneeded iterations

Symfony/CS/Tests/Fixtures/Integration/empty_return,useless_return,whitespacy_lines,extra_empty_lines.test
keradus: when #1605 will be merged

SpacePossum: :+1:  
which is RTM, so let me switch those test around :)
(expecting tests to fail)

keradus:
https://github.com/SpacePossum/PHP-CS-Fixer/pull/6

SpacePossum:
Thanks @keradus, picked in this branched, rebased and clean upped a bit. Lets see how the test do...
In the mean time could you think about https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1591#discussion_r47430303?

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,98 @@
...
+                continue;
+            }
+
+            if ($tokens->getNextMeaningfulToken($nextAt) !== $end) {
+                continue;
+            }
+
+            $this->removeReturnStatement($tokens, $index, $nextAt);
+
+            return;
```
GrahamCampbell: Isn't writing `break;` more conventional here. I think `return;` is a little confusing, especially if you missread the closing braces.

Symfony/CS/Tests/Fixer/Symfony/EmptyReturnFixerTest.php
```diff
@@ -30,7 +30,6 @@ public function provideExamples()
     {
         return array(
             // check correct statements aren't changed
-            array('<?php return;'),
```
GrahamCampbell: ?

keradus: The test is duplicated few lines later

keradus:
You squash it, could you point out the fix?

keradus:
Oh, there are no fix yet ;)

README.rst
keradus: @SpacePossum I wanted to think about it but the fixer doesn't work ;)

SpacePossum: there are about 6 cases on SF 2.3

   1) Bundle/WebProfilerBundle/Tests/Controller/ProfilerControllerTest.php (useless_return)
   2) Component/Intl/Data/Generator/LanguageDataGenerator.php (useless_return)
   3) Component/Intl/Data/Generator/CurrencyDataGenerator.php (useless_return)
   4) Component/Form/Tests/Extension/Validator/Type/BaseValidatorExtensionTest.php (useless_return)
   5) Component/Finder/Adapter/AbstractFindAdapter.php (useless_return)
   6) Component/Serializer/Tests/Normalizer/GetSetMethodNormalizerTest.php (useless_return)

SpacePossum:
>You squash it, could you point out the fix?

not squashed (yet ;) )
https://github.com/SpacePossum/PHP-CS-Fixer/commit/102eec2d7d06e72e9c4a1ef6469350d46390a1ba

README.rst
keradus: So standard procedure - send a PR to fix them.
Looking on the scale I think it will be merged, so Symfony level indeed.

keradus:
https://github.com/SpacePossum/PHP-CS-Fixer/pull/7 ?

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    private function removeReturnStatement(Tokens $tokens, $start, $semiColonIndex)
```
keradus: `$semicolonIndex`

phpdoc (at least typehinting)

keradus:
Please add:
```php
            array(
                '<?php
    function foo () { }',
            ),
            array(
                '<?php
    function foo () {
        '.'
    }',
                '<?php
    function foo () {
        return ;;
    }',
            ),
```

SpacePossum:
Thanks @keradus :)
You second test case is invalid, however it shows that the fixer should be ran after the `duplicate_semicolon`. 
I'll do some updating for this soon.

README.rst
SpacePossum: check, done (https://github.com/symfony/symfony/pull/17249)
lets see what comes back before I move the fixer.

keradus:
> You second test case is invalid

Why is that ?

SpacePossum:
yes, meant 3rd, I was counting from 0 I guess 0.o

README.rst
SpacePossum: I suggest leaving this Fixer at `contrib`, we can always move it to Symfony level later when they settled the return null vs. return void issue.

keradus: fine with me

keradus:
:D

keradus:
Please rebase and move integration tests into `priority` subdir

keradus:
Please @SpacePossum squash my commits into 1

SpacePossum:
I've squashed your commits.. but I'm sorry about the result...

gharlan:
from https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1672#issuecomment-171590838:

> the useless return fixer should be able to deal with multiple useless returns after each order (current missing feature)

this should be fixed

SpacePossum:
appveyor is down "Failed to provision build worker virtual machine in Google Compute Engine cloud. Try restarting build later." .. will try again later

I'll work on the multiples soon... would it stop merging though?

gharlan:
> I'll work on the multiples soon... would it stop merging though?

IMO yes. A fixer should not change anything in a second run.

SpacePossum:
>IMO yes. A fixer should not change anything in a second run.

didn't thought about that, agreed, I'll rework this :+1:

Symfony/CS/Fixer/Contrib/UselessReturnFixer.php
```diff
@@ -0,0 +1,108 @@
...
+        for ($index = $semiColonIndex; $index >= $start; --$index) {
+            if ($tokens[$index]->isComment()) {
+                $commentFound = true;
+                continue;
+            }
+
+            $tokens[$index]->clear();
+        }
+
+        // merge whitespace tokens if needed
```
keradus: it is the next place when I see this... please follow DRY

SpacePossum: This is done by a lot fixers, all in their own ways. I'm not sure where to refactor to, however I feel this fixer is a bit of a mess at the moment.

keradus: Feel free to clean it up ;)

SpacePossum: >I'm not sure where to refactor to

;)

keradus:
btw, please rebase ;)

SpacePossum:
>btw, please rebase ;)

:+1: done :)

keradus:
:+1:

before merge please squash commits (for some reason my one already lost almost all content?) - squash all into one (so even my one) ;)
then, please add fixer to `.php_cs` file

SpacePossum:
>for some reason my one already lost almost all content?

yea sorry about that, I've no idea how I kept your commit but lost a part of the content :(

I've added the fixer the `.php_cs`, squashed all commits into one.

keradus:
very well

PR #1594, opened 2015-12-14
Fix invalid PHP code samples in utests 

SpacePossum:
the saga continues;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1538

### DIFF
--- a/Symfony/CS/Tests/Fixer/Contrib/Php4ConstructorFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/Php4ConstructorFixerTest.php
@@ -965,23 +965,6 @@ EOF;
         $this->makeTest($expected);
     }
 
-    public function testStatic()
-    {
-        $expected = <<<'EOF'
-<?php
-
-static class Foo
-{
-    static function Foo()
-    {
-        // Do stuff
-    }
-}
-EOF;
-
-        $this->makeTest($expected);
-    }
-
     public function testParentMultipleClasses()
     {
         $expected = <<<'EOF'
--- a/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
@@ -82,7 +82,7 @@ class BracesFixerTest extends AbstractFixerTestBase
                 '<?php
     try {
         echo 1;
-    } catch (Exception $2) {
+    } catch (Exception $e) {
         echo 2;
     }',
                 '<?php
@@ -90,7 +90,7 @@ class BracesFixerTest extends AbstractFixerTestBase
     {
         echo 1;
     }
-    catch (Exception $2)
+    catch (Exception $e)
     {
         echo 2;
     }',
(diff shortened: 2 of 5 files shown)


GrahamCampbell:
:+1:

SpacePossum:
btw. is there a PHP version supporting static classes?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixer/Contrib/Php4ConstructorFixerTest.php#L973

and `try` without curly braces?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php#L711

stof:
@SpacePossum see the ``messages`` tab: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-738/messages (they could have opened it directly for better UX though)

stof:
And regarding static classes, this indeed does not make any sense. Regarding try without braces, I don't think so: https://3v4l.org/bZ10s

keradus:
Nope, no support for static-classes and try-without-{

SpacePossum:
thanks for the feedback @stof and @keradus (nice site to test code :) )
lets see AppVeyor gets a worker from google this time

Symfony/CS/Tests/Fixer/Symfony/MethodArgumentDefaultValueFixerTest.php
```diff
@@ -139,12 +139,17 @@ function foo(
...
+        if (defined('T_ELLIPSIS')) {
+            $cases[] = array(
+                '<?php function a($a = 1, \SplFileInfo ...$b) {}',
+            );
+            $cases[] = array(
+                '<?php function a($a = 1, ...$b) {}',
+            );
+        }
+
+        return $cases;
```
gharlan: should be reverted. The fixer has extra handling for php versions without `T_ELLIPSIS`.

SpacePossum: thank you,
reverted :)

Symfony/CS/Tests/Fixer/Contrib/Php4ConstructorFixerTest.php
```diff
@@ -965,23 +965,6 @@ public function __construct()
...
     }
 
-    public function testStatic()
-    {
-        $expected = <<<'EOF'
-<?php
-
-static class Foo
-{
-    static function Foo()
```
keradus: this case is important. Please recover it, remove `static` for class-level, but keep it for method-level

SpacePossum: Static constructors are not valid either (PHP 4-7) so the test will stay invalid.
I'll revert for now as to not stop this PR from going forward.

keradus: Yes, indeed, there are not static constructors. So this is only a static function. And the test is to prove that function called after class (so PHP4 constructor?) is not a constructor because it is static function, therefore is not changed into PHP5 constructor ;)

keradus:
:+1: for current progress !

Symfony/CS/Tests/Fixer/Contrib/Php4ConstructorFixerTest.php
gharlan: Imo the test case should be removed. it is not valid, result is a fatal error:

https://3v4l.org/CTJV0

keradus: OK, it's invalid on 4.x too, Thanks @gharlan!

@SpacePossum , please remove it one more time ;)

SpacePossum: re-reverted and squashed ^-^

keradus:
Thank you @SpacePossum.

PR #1597, opened 2015-12-15
MethodArgumentDefaultValueFixer - refactoring and fix closures with "use" clause

### LINKED ISSUE #1587: Broken method_argument_default_value fixer
Just tested this out on Laravel. The first change is wrong, but the second one is fine.

image

gharlan:
fixes #1587

### DIFF
--- a/Symfony/CS/Fixer/Symfony/MethodArgumentDefaultValueFixer.php
+++ b/Symfony/CS/Fixer/Symfony/MethodArgumentDefaultValueFixer.php
@@ -22,7 +22,7 @@ use Symfony\CS\Tokenizer\Tokens;
 final class MethodArgumentDefaultValueFixer extends AbstractFixer
 {
     private $argumentBoundaryTokens = array('(', ',', ';', '{', '}');
-    private $variableOrTerminatorTokens = array(array(T_VARIABLE), ';', '{', '}');
+    private $variableOrTerminatorTokens = array(array(T_VARIABLE), array(T_USE), ';', '{', '}');
     private $argumentTerminatorTokens = array(',', ')', ';', '{');
     private $defaultValueTokens = array('=', ';', '{');
     private $immediateDefaultValueTokens = array('=', ',', ';', '{');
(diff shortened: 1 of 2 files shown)


Symfony/CS/Fixer/Symfony/MethodArgumentDefaultValueFixer.php
```diff
@@ -22,7 +22,7 @@
 final class MethodArgumentDefaultValueFixer extends AbstractFixer
 {
     private $argumentBoundaryTokens = array('(', ',', ';', '{', '}');
-    private $variableOrTerminatorTokens = array(array(T_VARIABLE), ';', '{', '}');
+    private $variableOrTerminatorTokens = array(array(T_VARIABLE), array(T_USE), ';', '{', '}');
```
keradus: Sth is terrible wrong with this fixer and all that neighbours. 
`variableOrTerminatorTokens` - is `T_USE` a variable? No. Is it a terminator? No, as parameters list should already be terminated by `)`, but it isn't in the list even!

What is worst, the fixer works with following props:
```php
    private $argumentBoundaryTokens = array('(');
    private $variableOrTerminatorTokens = array(array(T_VARIABLE), array(T_USE), '{', );
    private $argumentTerminatorTokens = array(',');
    private $defaultValueTokens = array('=', ')');
    private $immediateDefaultValueTokens = array('=', '{');
```

I do not say that the props should be simplified into one I just provided.
Almost all usage of `getNextTokenOfKind` method in this fixer should be replaced by `getNextMeaningfulToken` and instead of searching for tokens of given kind - which may be way outside the parameters scope (like that `T_USE`) - decide what to do based on next token.

gharlan: Yeah, I agree. 
I just made a fast fix with current style. But I would prefer a refactoring, too.

gharlan: done

gharlan:
ping, anyone willing to do a review? :)

keradus:
I will need to take a deeper look into it as diff is crappy here

gharlan:
Yeah. Maybe it is easier to look first at my implementation without diff: https://github.com/gharlan/PHP-CS-Fixer/blob/madv-closures/Symfony/CS/Fixer/Symfony/MethodArgumentDefaultValueFixer.php

And then search the lines in diff where you want to comment something.

keradus:
That's the plan ;)

keradus:
It looks like all my concerns have been addressed.

:+1:

keradus:
Thank you @gharlan.

PR #1601, opened 2015-12-15
[WIP] UnneededControlParenthesesFixer - Inline case is not fixed

SpacePossum:
When trying to fix invalid test code I came across this test case for the `UnneededControlParenthesesFixer`;

```php
<?php
case ($x);
```

when 'fixed' to:
```php
<?php
switch($a){case ($x) : echo 1;}
```

it failed, expected:
```php
<?php
switch($a){case $x : echo 1;}
```

(no fix (yet), just reporting the test case)

### DIFF
--- a/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/UnneededControlParenthesesFixer.php
@@ -81,8 +81,7 @@ final class UnneededControlParenthesesFixer extends AbstractFixer
             }
 
             $blockStartIndex = $index;
-            $index = $tokens->getPrevMeaningfulToken($index);
-            $token = $tokens[$index];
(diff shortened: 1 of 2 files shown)


keradus:
Please, take a look at:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1734/files#diff-f01e54f3a6f8c27203f855c050c9eb86L321

SpacePossum:
thanks @keradus 
rebased and added the test, it fails locally. I guess it has to do with a one line `switch` statement
(rare case, but still ;) )

SpacePossum:
ping @Soullivaneuh @gharlan , please maybe one of you two could take a look? I'm not familiar with the fixer.

SpacePossum:
on checking again, there is no issue, I configured the fixer wrong.....

PR #1605, opened 2015-12-16
integration tests - swap EXPECT and INPUT (optional INPUT)

gharlan:


Symfony/CS/Tests/AbstractIntegrationTest.php
```diff
@@ -36,16 +36,16 @@
...
+ * Expected code after fixing
+ * --INPUT--
+ * Code to fix*****
  *
  *     * Additional fixers may be omitted.
  *    ** PHP minimum version. Default to current running php version (no effect).
  *   *** HHVM compliant flag. Default to true. Set to false to skip test under HHVM.
  *  **** Black listed filters may be omitted.
- * ***** When the expected block is omitted the input is expected not to
+ * ***** When the input block is omitted the expected part is expected not to
```
keradus: it is true even when input block is present

gharlan: good point.

"When the input block is omitted only the expected parts is checked to not be changed by the fixers."

Ok? Other proposal?

keradus: I would just say it is optional

keradus:
:+1:

GrahamCampbell:
:+1:

SpacePossum:
:+1:

keradus:
Thank you @gharlan.

Issue #1604, opened 2015-12-16
Performance Analysis


GrahamCampbell:
As PHP-CS-Fixer grows, it's definitely getting slower. I think we need to look at which fixers are the slowest. Maybe just run them one by one on symfony, then note the results. Basically, it would be good if we could mark certain fixers as "slow".

GrahamCampbell:
Also, one thing we could do is link the Docblock class to the actual token, then only re-parse it if it's content was changed, instead of parsing each phpdoc token multiple times during execution.

GrahamCampbell:
Or, even if we ran fixers together, it might cool to have one kind of way to collect performance details for each fixer.

GrahamCampbell:
If such a thing were to be done, I'd be happy to collect the metrics from StyleCI over a period of a week or something, and report back. That should be a fair bit of data.

stof:
It would be great to work towards releasing the v2, which is intended to be faster thanks to the candidate-based checking

> Also, one thing we could do is link the Docblock class to the actual token, then only re-parse it if it's content was changed, instead of parsing each phpdoc token multiple times during execution.

@GrahamCampbell or adding a cache for the docblock parsing (similar to cache for the code parsing). but anything related to caching docblock parsing would require ensuring that fixers avoid corrupting this cache, to avoid breaking other fixers.

SpacePossum:
>It would be great to work towards releasing the v2, which is intended to be faster thanks to the candidate-based checking

:+1:   
also for not re-tokenizing the code which is a great performance gain :)

keradus:
Actually, first we should measure the performance difference using _same_ fixers on 1.x and 2.x line.

GrahamCampbell:
> @GrahamCampbell or adding a cache for the docblock parsing (similar to cache for the code parsing). but anything related to caching docblock parsing would require ensuring that fixers avoid corrupting this cache, to avoid breaking other fixers.

Yeh, the idea is that any fixer that modified the contents of the doc token would flush the cache.

stof:
> also for not re-tokenizing the code which is a great performance gain

this is already the case in most cases in 1.x (thanks to the Tokens cache). What we actually avoid in 2.x is stringifying the tokens again and again

GrahamCampbell:
I'm not talking about tokenization, I'm talking about parsing in our "DocBlock" classes.

keradus:
Closing due to #1872

[...]

PR #1607, opened 2015-12-17
performance improvement

gharlan:
refs #1604

Here is the comparison for fixing this project itself (using given config, no cache, php 5.6):

https://blackfire.io/profiles/compare/e5a3a321-e328-426c-8a30-bde631dd70cf/graph

Before:

screenshot 2015-12-18 01 34 28

After:

screenshot 2015-12-18 01 33 11

### DIFF
--- a/Symfony/CS/Tests/Tokenizer/TokenTest.php
+++ b/Symfony/CS/Tests/Tokenizer/TokenTest.php
@@ -368,21 +368,6 @@ final class TokenTest extends \PHPUnit_Framework_TestCase
             array(true, array($brace, $foreach, array(T_FUNCTION, 'function'))),
             array(false, array($brace, $foreach, array(T_FUNCTION, 'Function'))),
             array(true, array($brace, $foreach, array(T_FUNCTION, 'Function')), false),
-            array(false, array($brace, $foreach, array(T_FUNCTION, 'Function')), array()),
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Pretty sexy/ :+1:

sstok:
:+1: :clap:

gharlan:
Made another improvement. What do you think? The feature was not used anywhere.

New comparison to current master:
https://blackfire.io/profiles/compare/b518187a-1e7a-4274-9c84-c644ab47f290/graph

GrahamCampbell:
Unfortunately, those stats are massively inflated on the reality though. I find that blackfire generally slows down all function calls by a fixed cost, so low cost functions have their call time inflated by blackfire. If you run this code without blackfire, you'll see the difference is much less, but still pretty decent. I found this out while trying to improve the performance of Laravel.

gharlan:
hmm, yes, you're right. Without blackfire the performance gain is much lower.

gharlan:
At the moment the performance gain on my machine is round about 12%.
(without blackfire, fixing project itself, php 5.6 and 7.0)

GrahamCampbell:
Yeh, still a very nice improvement though. :+1: :ship: :rocket:

SpacePossum:
:+1: :shipit: :)

keradus:
Thank you @gharlan.

PR #1608, opened 2015-12-17
Travis - change matrix order for faster results

gharlan:


gharlan:
if ordering by version is unimportant, we could use this matrix (fastest and most interesting builds at top):

```yml
matrix:
    fast_finish: true
    include:
        - php: hhvm
        - php: 7.0
        - php: 7.0
          env: SYMFONY_VERSION="~2.8"
        - php: 5.3
          env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
        - php: 5.4
        - php: 5.5
        - php: 5.6
```

keradus:
putting slowest jobs at the end is nightmare - it causes whole build to be slow.

gharlan:
My main target: get faster result when a build is failing, because then I have to change something in my pr.
And a second target: get a faster tendency. When hhvm, 7.0 and 5.3 are passing, there is a good chance that whole build will pass. So I prefer my second proposal.

gharlan:
Or this one:

```yml
matrix:
    fast_finish: true
    include:
        - php: 5.3
          env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
        - php: 7.0
        - php: hhvm
        - php: 5.4
        - php: 5.5
        - php: 5.6
        - php: 7.0
          env: SYMFONY_VERSION="~2.8"
```

GrahamCampbell:
I don't think what we have needs changing.

gharlan:
Ok. @keradus your comment was a :-1:, right?
Then we can close it ;)

keradus:
I'm :-1: for flip-flop the order as you done in commit.
But I'm **not** :-1: for changing the order, eg only move 7.0 on top (but I'm not :+1: for that neither, just :±0: ;) )

gharlan:
(updated to my current proposal..)

any other thoughts?

SpacePossum:
lgtm :+1:

.travis.yml
```diff
@@ -5,14 +5,13 @@ matrix:
     include:
         - php: 5.3
           env: 'COMPOSER_FLAGS="--prefer-stable --prefer-lowest"'
+        - php: 7.0
```
keradus: why removing the `SYMFONY_VERSION="~3.0"` ?

gharlan: imo it is useless

gharlan: later I would add `LINT_TEST_CASES=1` here.

keradus:
Thank you @gharlan.

PR #1610, opened 2015-12-18
HashToSlashCommentFixer - Add

SpacePossum:
Just a little fixer from my collection I think others might want to use once in a while (so on contrib level).
It changes single line comments starting with `#` to starting with `\\`.
PHP handles those the same so changing those should not be a problem.
For example

```php
<h1>This is an <?php #echo 123;?> example</h1>
```
to
```php
<h1>This is an <?php //echo 123;?> example</h1>
```

```php
<?php
    # test
```
to
```php
<?php
    // test
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -341,6 +341,11 @@ Choose from the list of available fixers:
                         function's argument and its
                         typehint.
 
+* **hash_to_slash_comment** [symfony]
+                        Single line comments should
+                        use double slashes (//) and
+                        not hash (#).
+
 * **include** [symfony]
                         Include/Require and file path
                         should be divided with a
--- a/Symfony/CS/Fixer/PSR2/ElseifFixer.php
+++ b/Symfony/CS/Fixer/PSR2/ElseifFixer.php
@@ -50,8 +50,8 @@ class ElseifFixer extends AbstractFixer
             $nextToken->clear();
         }
 
-        # handle `T_ELSE T_WHITESPACE T_IF` treated as single `T_ELSEIF` by HHVM
-        # see https://github.com/facebook/hhvm/issues/4796
+        // handle `T_ELSE T_WHITESPACE T_IF` treated as single `T_ELSEIF` by HHVM
+        // see https://github.com/facebook/hhvm/issues/4796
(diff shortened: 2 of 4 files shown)


gharlan:
> I think others might want to use

:+1: Thanks for this one.

README.rst
```diff
@@ -577,6 +577,11 @@ Choose from the list of available fixers:
                         with preg. Warning! This could
                         change code behavior.
 
+* **hash_to_back_slash_comment** [contrib]
+                        Single line comments should
+                        use double backslash (//) and
```
gharlan: `//` are not backslashes ;)
So the fixer name is wrong, too.

gharlan:
symfony level?

README.rst
SpacePossum: oh man... it is Friday right?

SpacePossum: maybe `hash_to_slashes_comment` ?

SpacePossum: renamed ;)

keradus:
multiline test should be added:
```php
            array(
                '<?php
                    // foo
                    // bar
                ',
                '<?php
                    # foo
                    # bar
                ',
            ),
```

keradus:
Also, I think that this should also be changed by fixer:
```php
            array(
                '<?php
                    // foo
                ',
                '<?php
                    ## foo
                ',
            ),
```

------------------

Perhaps even this:
```php
            array(
                '<?php
                    ////////// foo
                ',
                '<?php
                    ########## foo
                ',
            ),
```

keradus:
> symfony level?

yes

SpacePossum:
Test added, rebased, moved to Symfony, squashed

I disagree with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1610#issuecomment-167425472 so I won't change it, I thing we should not touch the content of comments (and even if this should be done, than it should not be done with this fixer)

keradus:
:+1:

keradus:
Thank you @SpacePossum.

SpacePossum:
thank you @keradus :+1:

Issue #889, opened 2014-12-17
Improving Fixer Names In 2.0


GrahamCampbell:
Here are my ideas. Please discuss! :)

| Current Name                       | Proposed Name                      |
|------------------------------------|------------------------------------|
| eof_ending                         | single_eol_at_eof                  |
| function_call_space                | no_spaces_after_method_name        |
| function_declaration               | closure_definition_spacing         |
| linefeed                           | unix_line_endings                  |
| multiple_use                       | single_statement_per_import        |
| parenthesis                        | no_spaces_inside_parenthesis       |
| trailing_spaces                    | no_trailing_whitespace             |
| visibility                         | visibility_required                |
| double_arrow_multiline_whitespaces | double_arrow_multiline_whitespace  |
| duplicate_semicolon                | no_duplicate_semicolons            |
| multiline_array_trailing_comma     | trailing_comma_multiline_array     |
| namespace_no_leading_whitespace    | no_leading_namespace_whitespace    |
| single_array_no_trailing_comma     | no_trailing_comma_single_array     |
| multiline_spaces_before_semicolon  | no_whitespace_before_semicolons    |

keradus:
This thread is very important.
We need to come back here after new pack of phpdoc fixers will be merged.

gharlan:
`UnaryOperatorsSpacesFixer` vs. `OperatorsSpacesFixer` vs. `TernarySpacesFixer`..

keradus:
@GrahamCampbell could you apply @stof's and @gharlan's propositions in first msg please?

keradus:
`visibility` vs `visibility_required`
this fixer does not only add missing visibility, but also handle sequence of visibility, abstract, static...

GrahamCampbell:
It also changes things like `var $foo;` to `public $foo;`.

GrahamCampbell:
> this fixer does not only add missing visibility, but also handle sequence of visibility, abstract, static...

Could we abstract the order changing stuff to another fixer?

keradus:
That would be great, but then:
- fixer that does ordering will need to be enhanced to support non-perfect syntax (eg without visibility) or force that extracted fixers to run
- sb will need to split the fixer

Both take time, and while idea is cool, the simply renaming is less time consuming.

SpacePossum:
I like the proposal from @GrahamCampbell for the new names :+1: , but  Im with stof and ceeram.

[...]

PR #1613, opened 2015-12-20
ScalarCastFixer - LowerCaseCastFixer - Add

SpacePossum:


Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
```diff
@@ -0,0 +1,64 @@
...
+                    break;
+                case 'integer' :
+                    $newContent = '(int)';
+                    break;
+                case 'double' :
+                case 'real' :
+                    $newContent = '(float)';
+                    break;
+                default :
+                    continue 2;
```
keradus: `hack x2`

GrahamCampbell: This also doesn't work correctly.

GrahamCampbell: `INT` will not get changed to `int`, yet `IntegER` will get changed to `int`.

gharlan: > INT will not get changed to int

look at `LowerCaseCastFixer`

GrahamCampbell: I see. Then the strtolower in this fixer is incorrect then.

GrahamCampbell: Just like in the phpdoc fixers, we should only modify it if the case is actually correct.

SpacePossum: >`hack x2`

?

About the casing; I like this better because the fixer does not rely on the `LowerCaseCastFixer`, making it useful without it.

keradus: continue in switch is ugly and hacky. `continue 2` is just a `goto` :/

keradus:
```php
$castFrom = strtolower(trim($castFrom));

$castMap = [
    'boolean' => 'bool,
    // ...
];

$castTo = $castMap[$castFrom];
```

GrahamCampbell: cast from should not be calling strtolower

GrahamCampbell: we shouldn't be matching incorrect case

GrahamCampbell: we should run after the fixer that normalizes the case

Symfony/CS/Tests/Fixer/Symfony/ScalarCastFixerTest.php
```diff
@@ -0,0 +1,80 @@
...
+                );
+            $cases[] =
+                array(
+                    sprintf('<?php $b= (%s)$d;', $to),
+                    sprintf('<?php $b= (%s)$d;', strtoupper($from)),
+                );
+            $cases[] =
+                array(
+                    sprintf('<?php $b=(%s) $d;', $to),
+                    sprintf('<?php $b=( %s) $d;', ucfirst($from)),
```
keradus: I do not like that space removal.

Fixer description:
> Cast should be written in
> lower case.

Nothing about removing spaces!

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
SpacePossum: If you make a fixer to fix the case the other way around, i.e. upper, than this fixer wouldn't work and you need to add a new one.
If you make one for upper first char. casing, this fixer wouldn't work, etc. 
It is a useless dependency adding nothing but complexity IMO, we don't this for fixers like AliasFunctionFixer as well (i.e. force a case fixer before it being useful).

Symfony/CS/Tests/Fixer/Symfony/ScalarCastFixerTest.php
SpacePossum: check, fixed :+1:

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
```diff
@@ -0,0 +1,60 @@
...
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = 0, $count = $tokens->count(); $index  < $count; ++$index) {
+            if (!$tokens[$index]->isCast()) {
+                continue;
+            }
+
+            static $castMap = array(
```
keradus: this could be taken outside the loop

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
```diff
@@ -0,0 +1,60 @@
...
+                'boolean' => 'bool',
+                'integer' => 'int',
+                'double' => 'float',
+                'real' => 'float',
+            );
+
+            $castFrom = substr($tokens[$index]->getContent(), 1, -1);
+            $castFrom = strtolower(trim($castFrom));
```
keradus: 41-42 could be combined

Symfony/CS/Fixer/Symfony/LowerCaseCastFixer.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $tokens = Tokens::fromCode($content);
+
+        for ($index = 0, $count = $tokens->count(); $index  < $count; ++$index) {
+            if (!$tokens[$index]->isCast()) {
+                continue;
+            }
+
+            $lowerContent = strtolower($tokens[$index]->getContent());
+            if ($lowerContent === $tokens[$index]->getContent()) {
```
keradus: `->setContent` will do it for you

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
SpacePossum: you still need it, because you need to deal with the white spaces (or not, but you want to :) )

Symfony/CS/Fixer/Symfony/LowerCaseCastFixer.php
SpacePossum:
```php
    /**
     * Set token's content.
     *
     * @param string $content
     */
    public function setContent($content)
    {
        // setting empty content is clearing the token
        if ('' === $content) {
            $this->clear();

            return;
        }

        $this->content = $content;
    }
```

it doesn't do any checks?

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
keradus: In matter of whitespaces the `str_replace` works the same as `str_ireplace`. The only difference between that 2 functions is case-sensitiveness

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
SpacePossum: done, thanks

Symfony/CS/Fixer/Symfony/LowerCaseCastFixer.php
keradus: 1.x line you don't need to check this and on 2.x line there is a check.

SpacePossum: I rather check it myself than relying on the implementation details of a method that promise nothin else than setting content.

keradus: then you will check it twice

Symfony/CS/Fixer/Symfony/ScalarCastFixer.php
GrahamCampbell: we need case sensitive replacing, replacing `(Boolean)` with `'bool'` is wrong, for the same reasons it was in the phpdoc fixers we have

keradus: Please do not hijack my comment ;P

In phpdoc the case may matter since one may have class `Boolean`. But casting is case-insensitive

GrahamCampbell: i don't see the need for two fixers here, because the 2nd fixer steps on the toes of the 1st

GrahamCampbell: just have a single fixer that normalizes casting, and we're done

Symfony/CS/Fixer/Symfony/LowerCaseCastFixer.php
keradus: btw, method promise that the content will  be the given arg, and indeed - it will. in both scenarios (same content/different content)

keradus:
:+1:

gharlan:
:+1:

README.rst
```diff
@@ -360,6 +360,10 @@ Choose from the list of available fixers:
                         Remove trailing commas in list
                         function calls.
 
+* **lower_case_cast** [symfony]
```
gharlan: maybe better `lowercase_casts` to match `lowercase_constants` and `lowercase_keywords`?

keradus: :+1:

SpacePossum: :+1:  done in https://github.com/SpacePossum/PHP-CS-Fixer/commit/de2705df1abfc2670967e002d79a9ce188e706c2#diff-88b99bb28683bd5b7e3a204826ead112L363

README.rst
```diff
@@ -495,6 +499,13 @@ Choose from the list of available fixers:
                         An empty line feed should
                         precede a return statement.
 
+* **scalar_cast** [symfony]
```
keradus: `schort_scalar_cast` maybe?

GrahamCampbell: I would say it's "correct" rather than "short", and the other types are aliases of the correct ones... but... php is so internally inconsistent, i have no idea what they actually intended to be the correct types

keradus: So whay you would say that `bool` is correct and `boolean` is... incorrect?
Both works, one may use only short form, one only long form.

GrahamCampbell: > So whay you would say that bool is correct and boolean is... incorrect?

I think that, but php itself gets it wrong all the time.

GrahamCampbell: Example: https://3v4l.org/QTLLB.

keradus: And we would change that `integer` into `int`.
So, as both form could be considered as correct, the choose of form is purely CS matter. Therefore, `short_scalar_cast` provides non-uncertain name

GrahamCampbell: > So, as both form could be considered as correct

That's what I have a problem with though, lol. IMO, there can only be one actual name.

I agree we should probably just choose `int` and `bool` since those seem to be conventionally used.

SpacePossum: cool, thanks for the thoughts both, renamed to `short_scalar_cast`

README.rst
gharlan: hmm, you have renamed it to `lowercase_cast` but my proposal was `lowercase_casts`..

SpacePossum: I see, if misread.. but than again, we have `spaces_cast` and `short_bool_cast`. 
Adding `lowercase_cast` and `short_scalar_cast` would be more inline of not using plural. I don't feel much for one or the other, but it would be nice to be consistent. It is also not upped in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/889 to rename the current cast fixer names. Let me know what you think (cc @keradus) and I rename accordingly.

gharlan: my proposal was plural because of `lowercase_constants` and `lowercase_keywords`. But I didn't look at other fixer names. 
So I'm not sure here, too ;)

keradus: Indeed, we should go in consistance way.
And then, I'm not sure which way is better

keradus:
Thank you @SpacePossum.

PR #1017, opened 2015-02-14
Add ignore of PHP-CS-Fixer cache files

SpacePossum:

PR #1614, opened 2015-12-20
.gitignore - add cache file

gharlan:


keradus:
When you run `php php-cs-fixer fix` no `.php_cs.cache` would be created!

GrahamCampbell:
It would on the master branch though. This is missing only on the master branch, right?

SpacePossum:
:+1: 

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1017 ;)

gharlan:
> When you run php php-cs-fixer fix no .php_cs.cache would be created!

Hmm, you're right, we are not using cache in this project.
The file was created by a misconfigured php-cs-fixer "tool" in my IDE.

PR #1618, opened 2015-12-20
Async linting

### LINKED ISSUE #1612: Async Linting
Just a thought, but could we save time after fixing each file by making it's linting asynchronous instead of waiting for it to complete before fixing the next file?

gharlan:
refs #1612

### DIFF
--- a/src/Fixer.php
+++ b/src/Fixer.php
@@ -17,6 +17,7 @@ use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Finder\SplFileInfo as FinderSplFileInfo;
+use Symfony\Component\Process\Process;
 use Symfony\Component\Stopwatch\Stopwatch;
 use Symfony\CS\Error\Error;
 use Symfony\CS\Error\ErrorsManager;
@@ -139,16 +140,24 @@ class Fixer
             $config->getRules()
         );
 
-        foreach ($config->getFinder() as $file) {
-            if ($file->isDir() || $file->isLink()) {
-                continue;
-            }
+        /** @var \Iterator|\IteratorAggregate $iterator */
+        $iterator = $config->getFinder();
+        if ($iterator instanceof \IteratorAggregate) {
+            $iterator = $iterator->getIterator();
+        }
+        $iterator->rewind();
 
-            $name = $this->getFileRelativePathname($file);
+        $nextFile = $this->getFileForFixing($iterator, $fileCacheManager);
+
+        while ($file = $nextFile) {
+            $iterator->next();
+            $nextFile = $this->getFileForFixing($iterator, $fileCacheManager);
+
+            $name = $this->getFileRelativePathname($file['file']);
 
             $this->stopwatch->start($name);
 
-            if ($fixInfo = $this->fixFile($file, $fixers, $dryRun, $diff, $fileCacheManager)) {
+            if ($fixInfo = $this->fixFile($file['file'], $file['content'], $fixers, $dryRun, $diff, $fileCacheManager, $file['lint_process'])) {
                 $changed[$name] = $fixInfo;
             }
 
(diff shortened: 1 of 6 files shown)


Symfony/CS/Fixer.php
```diff
@@ -139,18 +140,26 @@ public function fix(ConfigInterface $config, $dryRun = false, $diff = false)
...
 
-        foreach ($config->getFinder() as $file) {
-            if ($file->isDir() || $file->isLink()) {
-                continue;
-            }
+        $iterator = $config->getFinder();
+        if ($iterator instanceof \IteratorAggregate) {
+            $iterator = $iterator->getIterator();
+        }
+        /* @var \Iterator $iterator */
```
GrahamCampbell: you need `/**`

gharlan:
This is just a first try. So I think there will be some points that should be changed.
But I would be happy to get some comments.

GrahamCampbell:
This doesn't actually close `#1612`. It only partially implements it. The other idea in the PR, was to stop blocking after we've run the fixers, and just start fixing the next file, and let linting continue for the previous one in the background.

GrahamCampbell:
I appreciate that it might be a good idea to approach both ideas separately though.

Symfony/CS/Fixer.php
```diff
@@ -139,18 +140,26 @@ public function fix(ConfigInterface $config, $dryRun = false, $diff = false)
...
+        $iterator->rewind();
+
+        $nextFile = $this->getNextFile($iterator, $fileCacheManager);
```
GrahamCampbell: Any way this could be refactored to remove this duplicate line?

Symfony/CS/Linter/LinterInterface.php
```diff
@@ -23,7 +23,7 @@
      *
      * @param string $path
      */
-    public function lintFile($path);
+    public function startLintingFile($path);
```
GrahamCampbell: missing `@return` annotatio

GrahamCampbell:
Looking pretty nice so far though. ;)

gharlan:
> This doesn't actually close #1612.

ok, changed pr description (closes -> refs) ;)

GrahamCampbell:
> ok, changed pr description (closes -> refs) ;)

:+1:

Symfony/CS/Fixer.php
```diff
@@ -139,18 +140,26 @@ public function fix(ConfigInterface $config, $dryRun = false, $diff = false)
...
 
-            $this->stopwatch->start($this->getFileRelativePathname($file));
+        while ($file = $nextFile) {
+            $iterator->next();
+            $nextFile = $this->getNextFile($iterator, $fileCacheManager);
 
-            if ($fixInfo = $this->fixFile($file, $fixers, $dryRun, $diff, $fileCacheManager)) {
-                $changed[$this->getFileRelativePathname($file)] = $fixInfo;
+            $this->stopwatch->start($this->getFileRelativePathname($file['file']));
```
gharlan: I'm not sure where to put this line?
Before this pr the time started before checking cache file and linting.
Now it starts before waiting for linting process. So checking cache file and starting the linting process are not captured by the stopwatch. (Cached files are not captured at all.)

Symfony/CS/Fixer.php
gharlan: It was fixed by `PhpdocToCommentFixer`. ;)
I will move this line to variable declaration.

Soullivaneuh:
Nice start! :+1: 

Would not be great to add an option to choose the max number of running processes?

keradus:
This change doesn't work that way...

Symfony/CS/Fixer.php
```diff
@@ -158,9 +167,51 @@ public function fix(ConfigInterface $config, $dryRun = false, $diff = false)
         return $changed;
     }
 
-    public function fixFile(\SplFileInfo $file, array $fixers, $dryRun, $diff, FileCacheManager $fileCacheManager)
+    private function getNextFile(\Iterator $iterator, FileCacheManager $fileCacheManager)
```
stof: this is not really getting the next file. It is getting the current file of the iterator (your code gets the next file because it moves the iterator before calling it)

gharlan: yes and no.
It returns the next file of iterator (starting with current) which is not a link and which is not cached.

stof: well, my point is precisely related to ``starting with current``. the following code will give you the same file all the time:

```php
$file = $this->getNextFile($iterator, $fileCacheManager);
$file = $this->getNextFile($iterator, $fileCacheManager);
$file = $this->getNextFile($iterator, $fileCacheManager);
$file = $this->getNextFile($iterator, $fileCacheManager);
```

But the method name ``getNextFile`` tells me that I will get the next file each time (so 4 different files)

gharlan:
So, after the discussions in other issues/prs, do I understand it correctly, that I should pursue this pr with current concept?
And later we would think again about more advanced solutions.

keradus:
Linting one file ahead, which uses separated process (before and after this PR) is OK.

gharlan:
I would like to get a review of current status.
What is to do here?

keradus:
TBH it looks a lot overwhelmed for me. I would like to simplify it somehow

gharlan:
> TBH it looks a lot overwhelmed for me.

agree

> I would like to simplify it somehow

me too. But so far I had no idea how it could be done.

gharlan:
I tried to use `foreach` instead of iterator methods, but aborted it. I will try it again.

keradus:
What I was thinking about is sth like:
- have a class that will provide next and next file (Traversable interface at least)
- that class will encapsulate linting behaviour (let us call it `FileLintingIterator` for the discussion)
- FileLintingIterator will provide current file for iteration and start linting next one.

Then, in main `Fixer` class we will just use that new iterator instead of `Finder` iterator and almost whole `Fixer` logic will remain the same

keradus:
@gharlan @stof 
please, could you both take a look at #1666 ?

gharlan:
Concept looks much better, thanks!

PR #1623, opened 2015-12-20
AbstractIntegrationTest - lint both expected and input code

keradus:


keradus:
ping @gharlan 
Actually I was starting it too.

I think we should lint both input and output

gharlan:
> I think we should lint both input and output

actually they are both linted. When INPUT != EXPECT, EXPECT part is linted in this recursive call: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/AbstractIntegrationTest.php#L209

keradus:
Thank you

PR #1625, opened 2015-12-21
Travis - switch to trusty

keradus:
Before:
> Elapsed time 8 min 6 sec
> Total time for 28 min 17 sec

After:
> Elapsed time 6 min 43 sec
> Total time for 25 min

-------------------------------------------

Before:
> HipHop VM 3.5.0 (rel)

After:
> HipHop VM 3.6.6 (rel)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ matrix:
 
 sudo: false
 
+dist: trusty
+
 cache:
     directories:
         - $HOME/.composer/cache


stof:
?

keradus:
Please, just ignore the test PR ;)

keradus:
pnig @stof

keradus:
ref #1465

Soullivaneuh:
You learn me something, I didn't know we can choose the linux distribution.

Do you have related doc link?

GrahamCampbell:
I think using the newer linux distro, you don't get the container architecture though?

keradus:
Why you woudnt? If I get it right - you can use sudo, but you don't need to.

@Soullivaneuh actually you could just use "default" 12.x or upcoming 14.x, which after some time replace 12.

keradus:
Any thoughts?

PR #1626, opened 2015-12-21
[WIP] Async Fixing

GrahamCampbell:
I hope you can see where this is going. We can just have promises for the linting steps.


GrahamCampbell:
Let me know if you're interested in me continuing this.

GrahamCampbell:
NB The fulfilled promises in the code atm are just placeholders for when it will actually be made async.

keradus:
:+1: for using `proc_open` for linting
:-1: for using `proc_open` for fixing (we have heavy app start), other parallel mechanisms could be better here
:-1: for dropping 5.3.6 (I want the version 2 to support it)
:-1: for using not stable library

GrahamCampbell:
> :-1: for using proc_open for fixing (we have heavy app start), other parallel mechanisms could be better 

I totally agree there, and I never planned to do that.

GrahamCampbell:
> :-1: for using not stable library

I disagree that it's not stable.

GrahamCampbell:
> for dropping 5.3.6 (I want the version 2 to support it)

I think that's a small price to pay if we can significantly increase the performance.

GrahamCampbell:
This is a bit of a mess tbh, lol.

Anahkiasen:
> for dropping 5.3.6 (I want the version 2 to support it)

Wouldn't version 2 actually be the perfect moment to drop support for EOL versions of PHP?

keradus:
No. It is widely used (even in dev version).

GrahamCampbell:
Meh.

keradus:
It is pointless to say that 2.x is so awesome, works on 5.3.6 and then just drop it even if it could be used.
Changing interface to promises and async work is great, but it is a lot of work, more than it looks. Great thing for 3.x.

GrahamCampbell:
Yeh, I'm inclined to agree. I also ran into an issue with exceptions and errors being silenced.

GrahamCampbell:
It would be good to come back and review this situation again in a year or 2.

PR #1586, opened 2015-12-09
Add configuration exception classes and exit codes

SpacePossum:
replacement of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1500 on 1.12

PR #1627, opened 2015-12-21
FixCommand - fix output

keradus:
Bug introduced by #1586 (in 1.12-dev line!), which isn't released yet so I'm not treating it as a bug.

Before #1586 without `-v` there were list of fixed file, after there is _empty line per fixed file_.

```bash
λ php php-cs-fixer fix 1
   1) FixCommand.php
   2) ReadmeCommand.php
   3) SelfUpdateCommand.php
Fixed all files in 4.126 seconds, 6.250 MB memory used
```

```bash
λ php php-cs-fixer fix 1



Fixed all files in 4.097 seconds, 6.250 MB memory used
```

### DIFF
--- a/Symfony/CS/Console/Command/FixCommand.php
+++ b/Symfony/CS/Console/Command/FixCommand.php
@@ -422,8 +422,9 @@ EOF
                 }
 
                 foreach ($changed as $file => $fixResult) {
+                    $output->write(sprintf('%4d) %s', $i++, $file));
+
                     if ($fixerDetailLine) {
-                        $output->write(sprintf('%4d) %s', $i++, $file));
                         $output->write(sprintf($fixerDetailLine, implode(', ', $fixResult['appliedFixers'])));
                     }
 


SpacePossum:
:+1: 
so sorry about this one :(

keradus:
Merging things to master started tooking me like half an hour :(

keradus:
BTW, it would be so nice to test the console output :D

GrahamCampbell:
I think we need to hurry up and release 2.0.0 so you don't need to merge stuff all the time. Can we make 1.12 the last 1.x series?

SpacePossum:
oh no :( 
you want me to prep it? there is no rush getting to master..

SpacePossum:
>BTW, it would be so nice to test the console output :D

that should be do-able (and even more important when we split stdout to/and stderr), lets see. .. .. ;)

keradus:
@SpacePossum 
it's already there.
yep, Symfony has that kind of tests ;)

@GrahamCampbell 
This is the idea. This is why I am pushing some refactoring changes lately (eg pinging SpacePossum on things like that instead of new and new fixers).

keradus:
Or didn't want to start refactoring whole codebase to work async ;P

GrahamCampbell:
> Or didn't want to start refactoring whole codebase to work async ;P

Haha. That'll be a real mess if we actually ever do it.

keradus:
I have started rewriting it from scratch. The most changes will be done inside `Fixer` class (which manage fixing by fixers). And when not doing it like "what to change to made it async" but "lets write it from scratch" it will look better than now. A lot of work would be needed to write sth like pool manager that not using proc_open and had fallback to non-async solution.

keradus:
I have started rewriting it from scratch about year ago. It is more than possible.
The most changes will be done inside `Fixer` class (which manage fixing by fixers). And when not doing it like "what to change to made it async" but "lets write it from scratch" it will look better than now. A lot of work would be needed to write sth like pool manager that not using proc_open and had fallback to non-async solution.

PR #1628, opened 2015-12-22
Added OrderedClassElementsFixer

### LINKED ISSUE #832: Feature request: class' elements reordering
It would be great to have a fixer, that grab all class elements and put them in order: consts, properties, methods.

### LINKED ISSUE #1294: [RFC] Order methods in class by visibility
Would be nice to have a fixer to order methods in class by visibility: public, protected, private.

gharlan:
closes #1294 and closes #832.

### DIFF
--- a/src/Fixer/ClassNotation/OrderedClassElementsFixer.php
+++ b/src/Fixer/ClassNotation/OrderedClassElementsFixer.php
@@ -1,0 +1,391 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Gregor Harlan <gharlan@web.de>
+ */
+final class OrderedClassElementsFixer extends AbstractFixer
+{
+    /**
+     * @var array Array containing all class element base types (keys) and their parent types (values)
+     */
+    private static $typeHierarchy = array(
+        'use_trait' => null,
+        'public' => null,
+        'protected' => null,
+        'private' => null,
+        'constant' => null,
+        'constant_public' => array('constant', 'public'),
+        'constant_protected' => array('constant', 'protected'),
+        'constant_private' => array('constant', 'private'),
+        'property' => null,
+        'property_static' => array('property'),
+        'property_public' => array('property', 'public'),
+        'property_protected' => array('property', 'protected'),
+        'property_private' => array('property', 'private'),
+        'property_public_static' => array('property_static', 'property_public'),
+        'property_protected_static' => array('property_static', 'property_protected'),
+        'property_private_static' => array('property_static', 'property_private'),
+        'method' => null,
+        'method_static' => array('method'),
+        'method_public' => array('method', 'public'),
+        'method_protected' => array('method', 'protected'),
+        'method_private' => array('method', 'private'),
+        'method_public_static' => array('method_static', 'method_public'),
+        'method_protected_static' => array('method_static', 'method_protected'),
+        'method_private_static' => array('method_static', 'method_private'),
+    );
+
+    /**
+     * @var array Array containing special method types
+     */
+    private static $specialTypes = array(
+        'construct' => null,
+        'destruct' => null,
+        'magic' => null,
+        'phpunit' => null,
+    );
+
+    /**
+     * @var string[] Default order/configuration
+     */
+    private static $defaultOrder = array(
+        'use_trait',
+        'constant_public',
+        'constant_protected',
+        'constant_private',
+        'property_public',
+        'property_protected',
+        'property_private',
+        'construct',
+        'destruct',
+        'magic',
+        'phpunit',
+        'method_public',
+        'method_protected',
+        'method_private',
+    );
+
+    /**
+     * @var array Resolved configuration array (type => position)
+     */
+    private $typePosition;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $configuration = self::$defaultOrder;
+        }
+
+        $this->typePosition = array();
+        $pos = 0;
+        foreach ($configuration as $type) {
+            if (!array_key_exists($type, self::$typeHierarchy) && !array_key_exists($type, self::$specialTypes)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Unknown class element type "%s".', $type));
+            }
+
+            $this->typePosition[$type] = $pos++;
+        }
+
+        foreach (self::$typeHierarchy as $type => $parents) {
+            if (isset($this->typePosition[$type])) {
+                continue;
+            }
+
+            if (!$parents) {
+                $this->typePosition[$type] = null;
+                continue;
+            }
+
+            foreach ($parents as $parent) {
+                if (isset($this->typePosition[$parent])) {
+                    $this->typePosition[$type] = $this->typePosition[$parent];
+                    continue 2;
+                }
+            }
+
+            $this->typePosition[$type] = null;
+        }
+
+        $lastPosition = count($configuration);
+        foreach ($this->typePosition as &$pos) {
+            if (null === $pos) {
+                $pos = $lastPosition;
+            }
+            // last digit is used by phpunit method ordering
+            $pos *= 10;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(Token::getClassyTokenKinds());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($i = 1, $count = $tokens->count(); $i < $count; ++$i) {
+            if (!$tokens[$i]->isClassy()) {
+                continue;
+            }
+
+            $i = $tokens->getNextTokenOfKind($i, array('{'));
+            $elements = $this->getElements($tokens, $i);
+
+            if (!$elements) {
+                continue;
+            }
+
+            $sorted = $this->sortElements($elements);
+            $endIndex = $elements[count($elements) - 1]['end'];
+
+            if ($sorted !== $elements) {
+                $this->sortTokens($tokens, $i, $endIndex, $sorted);
+            }
+
+            $i = $endIndex;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Orders the elements of classes/interfaces/traits.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run before MethodSeparationFixer, NoBlankLinesAfterClassOpeningFixer and SpaceAfterSemicolonFixer
+        return 65;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     *
+     * @return array[]
+     */
+    private function getElements(Tokens $tokens, $startIndex)
+    {
+        static $elementTokenKinds = array(CT_USE_TRAIT, T_CONST, T_VARIABLE, T_FUNCTION);
+
+        ++$startIndex;
+        $elements = array();
+
+        while (true) {
+            $element = array(
+                'start' => $startIndex,
+                'visibility' => 'public',
+                'static' => false,
+            );
+
+            for ($i = $startIndex; ; ++$i) {
+                $token = $tokens[$i];
+
+                // class end
+                if ($token->equals('}')) {
+                    return $elements;
+                }
+
+                if ($token->isGivenKind(T_STATIC)) {
+                    $element['static'] = true;
+                    continue;
+                }
+
+                if ($token->isGivenKind(array(T_PROTECTED, T_PRIVATE))) {
+                    $element['visibility'] = strtolower($token->getContent());
+                    continue;
+                }
+
+                if (!$token->isGivenKind($elementTokenKinds)) {
+                    continue;
+                }
+
+                $type = $this->detectElementType($tokens, $i);
+                if (is_array($type)) {
+                    $element['type'] = $type[0];
+                    $element['name'] = $type[1];
+                } else {
+                    $element['type'] = $type;
+                }
+
+                $element['end'] = $this->findElementEnd($tokens, $i);
+                break;
+            }
+
+            $elements[] = $element;
+            $startIndex = $element['end'] + 1;
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return string|array type or array of type and name
+     */
+    private function detectElementType(Tokens $tokens, $index)
+    {
+        $token = $tokens[$index];
+
+        if ($token->isGivenKind(CT_USE_TRAIT)) {
+            return 'use_trait';
+        }
+
+        if ($token->isGivenKind(T_CONST)) {
+            return 'constant';
+        }
+
+        if ($token->isGivenKind(T_VARIABLE)) {
+            return 'property';
+        }
+
+        $nameToken = $tokens[$tokens->getNextMeaningfulToken($index)];
+
+        if ($nameToken->equals(array(T_STRING, '__construct'), false)) {
+            return 'construct';
+        }
+
+        if ($nameToken->equals(array(T_STRING, '__destruct'), false)) {
+            return 'destruct';
+        }
+
+        if (
+            $nameToken->equalsAny(array(
+                array(T_STRING, 'setUpBeforeClass'),
+                array(T_STRING, 'tearDownAfterClass'),
+                array(T_STRING, 'setUp'),
+                array(T_STRING, 'tearDown'),
+            ), false)
+        ) {
+            return array('phpunit', strtolower($nameToken->getContent()));
+        }
+
+        if ('__' === substr($nameToken->getContent(), 0, 2)) {
+            return 'magic';
+        }
+
+        return 'method';
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return int
+     */
+    private function findElementEnd(Tokens $tokens, $index)
+    {
+        $index = $tokens->getNextTokenOfKind($index, array('{', ';'));
+
+        if ($tokens[$index]->equals('{')) {
+            $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
+        }
+
+        for (++$index; $tokens[$index]->isWhitespace(" \t") || $tokens[$index]->isComment(); ++$index);
+
+        --$index;
+
+        return $tokens[$index]->isWhitespace() ? $index - 1 : $index;
+    }
+
+    /**
+     * @param array[] $elements
+     *
+     * @return array[]
+     */
+    private function sortElements(array $elements)
+    {
+        static $phpunitPositions = array(
+            'setupbeforeclass' => 1,
+            'teardownafterclass' => 2,
+            'setup' => 3,
+            'teardown' => 4,
+        );
+
+        foreach ($elements as &$element) {
+            $type = $element['type'];
+
+            if (array_key_exists($type, self::$specialTypes)) {
+                if (isset($this->typePosition[$type])) {
+                    $element['position'] = $this->typePosition[$type];
+                    if ('phpunit' === $type) {
+                        $element['position'] += $phpunitPositions[$element['name']];
+                    }
+                    continue;
+                }
+
+                $type = 'method';
+            }
+
+            if (in_array($type, array('constant', 'property', 'method'), true)) {
+                $type .= '_'.$element['visibility'];
+                if ($element['static']) {
+                    $type .= '_static';
+                }
+            }
+
+            $element['position'] = $this->typePosition[$type];
+        }
+
+        usort($elements, function (array $a, array $b) {
+            if ($a['position'] === $b['position']) {
+                // same group, preserve current order
+                return $a['start'] > $b['start'] ? 1 : -1;
+            }
+
+            return $a['position'] > $b['position'] ? 1 : -1;
+        });
+
+        return $elements;
+    }
+
+    /**
+     * @param Tokens  $tokens
+     * @param int     $startIndex
+     * @param int     $endIndex
+     * @param array[] $elements
+     */
+    private function sortTokens(Tokens $tokens, $startIndex, $endIndex, array $elements)
+    {
+        $replaceTokens = array();
+
+        foreach ($elements as $element) {
+            for ($i = $element['start']; $i <= $element['end']; ++$i) {
+                $replaceTokens[] = clone $tokens[$i];
+            }
+        }
+
+        $tokens->overrideRange($startIndex + 1, $endIndex, $replaceTokens);
+    }
+}
--- a/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
+++ b/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
@@ -1,0 +1,561 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\ClassNotation;
+
+use PhpCsFixer\Test\AbstractFixerTestCase;
+
+/**
+ * @author Gregor Harlan <gharlan@web.de>
+ *
+ * @internal
+ */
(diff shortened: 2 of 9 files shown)


gharlan:
WIP (missing many test cases etc.)

gharlan:
What do you think generally about the concept?

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+            $this->config[$type] = $pos;
+        }
+
+        $lastPosition = count($configuration);
+        foreach (self::$types as $type => $parents) {
+            if (isset($this->config[$type])) {
+                continue;
+            }
+
+            if ($parents && null !== $pos = $this->findConfigurationPosition($parents)) {
```
keradus: dont assign inside if statement...

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+        ++$startIndex;
+        $elements = array();
+
+        while (true) {
+            $element = array(
+                'start' => $startIndex,
+                'visibility' => 'public',
+                'static' => false,
+            );
+            for ($i = $startIndex; ; ++$i) {
```
keradus: logic is similar to https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tokenizer/Tokens.php#L452, maybe it could be combined?

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Gregor Harlan <gharlan@web.de>
+ */
+final class ClassElementsOrderFixer extends AbstractFixer
+{
+    private static $types = array(
```
keradus: this structure looks like a little nightmare, could you at least describe what is the value under key ?

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+        'method' => null,
+        'method_static' => array('method'),
+        'method_public' => array('method', 'public'),
+        'method_protected' => array('method', 'protected'),
+        'method_private' => array('method', 'private'),
+        'method_public_static' => array('method_static', 'method_public'),
+        'method_protected_static' => array('method_static', 'method_protected'),
+        'method_private_static' => array('method_static', 'method_private'),
+        'construct' => null,
+        'phpunit' => null,
```
keradus: ?

gharlan: These are the phpunit methods `setUp[BeforeClass]` and `tearDown[AfterClass]`.
Symfony standard requires that these methods are at top.

GrahamCampbell: What about the static setup methods?

keradus: Exactly @gharlan 
But which one would be first, second etc?

it is a big bucket...

gharlan: > But which one would be first, second etc?

order of these methods will not be changed. they are only moved to the desired place.

keradus: why not to order them ?

gharlan:
Some examples of possible configurations:

```php
array(
    'use_trait',
    'const',
    'public', // all public properties and methods, current order of public elements will not be changed
    'protected', // see public
    'private', // see private
)
```

```php
array(
    'use_trait',
    'const',
    'property', // all properties, current order will not be changed (regardless of visibility)
    'method', // see property
)
```

```php
array(
    'use_trait',
    'const',
    'property_public', // public properties
    'property_protected',
    'property_private',
    'construct', // if this is missing, constructor will be handled as normal method
    'method_public',
    'method_protected',
    'method_private',
)
```

```php
array(
    'use_trait',
    'const',
    'property_public_static', // if this is missing, public static properties will be handled like property_public
    'property_protected_static',
    'property_private_static',
    'property_public',
    'property_protected',
    'property_private',
    'construct',
    'method_public_static',
    'method_protected_static',
    'method_private_static',
    'method_public',
    'method_protected',
    'method_private',
)
```

gharlan:
If a main category is missing, the correspondent elements will be placed at bottom.

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
gharlan: Imo it is not necessary. But if you want it, I could change that.
Would you like to have a fixed order in `phpunit` group, or should it be configurable?

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
gharlan: I will add a docblock.

The keys are all possible configuration elements. Value is an array of parent keys. So if a key is missing in configuration, the elements will be handled like the parent ones.

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
keradus: `setUpBeforeClass, tearDownAfterClass, setUp, tearDown` ?

gharlan: I think, it makes sense ;)

keradus: Symfony one ;P

gharlan:
I want to use the symfony standard as default configuration.

> - Declare class properties before methods;
> - Declare public methods first, then protected ones and finally private ones. The exceptions to this rule are the class constructor and the setUp and tearDown methods of PHPUnit tests, which should always be the first methods to increase readability;

So, is this default config correct?

```php
private static $defaultOrder = array(
        'use_trait',
        'const',
        'property_public',
        'property_protected',
        'property_private',
        'phpunit',
        'construct',
        'method_public',
        'method_protected',
        'method_private',
    );
```

Or are there any (hidden) rules about static properties/methods?

sstok:
Wow amazing work so far :clap:

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+        'method_public',
+        'method_protected',
+        'method_private',
+    );
+
+    private $config;
+
+    public function __construct()
+    {
+        $this->configure(null);
```
SpacePossum: Are you sure you want to go through the default configuration init. on construction. Feels a bit unneeded if the Fixer is configured later on anyway by the FixerFactory IIRC?

great work so far :+1: :)

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+                    $element['type'] = 'const';
+                    $element['end'] = $this->findElementEnd($tokens, $i, array(';'));
+                    break;
+                }
+                if ($tokens[$i]->isGivenKind(T_VARIABLE)) {
+                    $element['type'] = 'property';
+                    $element['end'] = $this->findElementEnd($tokens, $i, array(';'));
+                    break;
+                }
+                if ($tokens[$i]->isGivenKind(T_FUNCTION)) {
```
SpacePossum: will this work with closures?

gharlan: It should work because closures are always within method bodies, and these are omitted by `findElementEnd`.
But we will see when I add necessary test cases. ;)

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
```diff
@@ -0,0 +1,305 @@
...
+            $elements = $this->getElements($tokens, $i);
+
+            if (!$elements) {
+                continue;
+            }
+
+            $sorted = $this->sortElements($elements);
+            $endIndex = $elements[count($elements) - 1]['end'];
+
+            if ($sorted !== $elements) {
```
gharlan: hmm, I need a non-strict comparison here, but `!=` was fixed to `!==`.

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
gharlan: Is the `configure` method always called?

gharlan: answer is: yes
fixed.

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
gharlan: added test case, it works.

gharlan:
some todos:

- tests, tests, tests
- priority (before `method_separation`), integration test
- order phpunit methods
- decide about default config
- maybe move to symfony

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
gharlan: Before I started to implement this part I looked at the method. IMO it would need more code to use that method than without using it, caused by special needs of this fixer. And the tokens would be traversed more often.

(at the moment this fixer traverse each token only once..)

Symfony/CS/Fixer/Contrib/ClassElementsOrderFixer.php
SpacePossum: thanks and thanks :)

gharlan:
should the fixer be renamed to `OrderedClassElementsFixer`?

keradus:
:+1: for rename

gharlan:
ready for review

SpacePossum:
losts of cool stuff :) 
I didn't see it in the tests, how does the fixer handle PHPDocs/comments of structures it is ordering?

gharlan:
@SpacePossum where is your comment? I answer anyway.. ;)

Comments are tested here: https://github.com/gharlan/PHP-CS-Fixer/blob/class-elements-order-fixer/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php#L146-L173
But I should add some more tests.

The cut is made before first line feed after `;` or `}`.
Some examples:

```
function a() {
} /* comment */ // comment<CUT>
// comment

function b() {
} /* multiline
comment */<CUT>

function c () {
}<CUT>
/* comment */

function d() {}<CUT> function e() {}
```

SpacePossum:
I removed the comment because I didn't look correctly, sorry for the noise :(
Thank you for explaining, I'm trying out the fixer atm, looking good :)

gharlan:
Additional note: the cut position is the reason why target branch is master. I need the transformer for `//` and `#` comments.

SpacePossum:
I did some testing and I'm wondering about this case.
Running on your branch with fixers `ordered_class_elements`, `no_blank_lines_after_phpdoc`.
(I was testing if the order would be important, turns out it isn't :+1: )

expected
```php
<?php

class Foo
{
    // test

    /**
     * public
     */
    public function bar() {}

    /**
     * protected
     */
    protected function baz() {}
}

```

input
```php
<?php

class Foo
{
    // test

    /**
     * protected
     */


    protected function baz() {}
    /**
     * public
     */


    public function bar() {}
}
```

diff;
```diff
@@ @@
 <?php
 
 class Foo
 {
-    // test
-
     /**
      * public
      */
     public function bar() {}
+    // test
 
     /**
      * protected
      */
     protected function baz() {}
 }

```


(btw. fabbot.io is wrong, Fixer has been fixed for this case a while back ;)  )

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
```diff
@@ -0,0 +1,398 @@
...
+                    continue;
+                }
+
+                $type = $this->detectElementType($tokens, $i);
+                if (is_array($type)) {
+                    $element['type'] = $type[0];
+                    $element['name'] = $type[1];
+                } else {
+                    $element['type'] = $type;
+                }
```
SpacePossum: missing blankline after `}` ;)

gharlan:
Question: What would you expect with this input:

```php
<?php

class Foo
{
    /**
     * protected
     */
    protected function baz() {}
   
    // test

    /**
     * public
     */
    public function bar() {}
}
```

SpacePossum:
I was thinking about it because of the method separation fixer findings on SF;
https://github.com/symfony/symfony/pull/17487/files#diff-a8f7745712c744a53020ed9c09bd68bcR979

These types of comments are tricky and in most of the cases people should be better of merging the comment in the PHPDoc. I'm +0 for leaving such comments at the current place or moving the comments along with the structure. I would like to ask for a utest for this case so the behavior of the fixer is clearly defined (and for all the usual reasons to have a test :) )

gharlan:
the fixer is very greedy. All comments above an element are treated as associated to the next class element.
So there are not any "neutral" parts between `{` and `}` of class body. Everything belongs to a class element.

We can not avoid some unwanted comment moves here. But in my opinion moving the comments along with the structure is the better choice in most cases.

gharlan:
added some more tests with comments

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
gharlan: done

SpacePossum:
Thanks for the update on the tests and the info! 

I've run the fixer on a bunch of projects, including some I know are pretty bad. All fixes were legit, I didn't spot missed cases, performance of the fixer is pretty good. One thing I didn't test thoroughly are all configuration options, but I thrust the utest on it.
Looking at the TODO list; 
I like the current default config; having it on SF level will be tricky since merging after the changes will be hard; the proof of the pudding is in the eating so maybe PR'ing the changes @ SF will give the conclusion.  
I'm :+1: for the current implementation; impressive fixer : D

gharlan:
thanks for testing! :+1: 

> having it on SF level will be tricky since merging after the changes will be hard; the proof of the pudding is in the eating so maybe PR'ing the changes @ SF will give the conclusion.

yeah, I guess the pr will not be merged, but I will open it anyway.

> I'm  for the current implementation; impressive fixer : D

thanks! ;)

tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
```diff
@@ -0,0 +1,534 @@
...
+            array(
+                <<<'EOT'
+<?php
+
+class Foo
+{
+    public const C1 = 1;
+    const C2 = 2;
+    public const C3 = 3;
+    protected const C4 = 4;
```
keradus: https://3v4l.org/38vqp

keradus: oh, 7.1 :D

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
```diff
@@ -0,0 +1,399 @@
...
+    private static $defaultOrder = array(
+        'use_trait',
+        'constant_public',
+        'constant_protected',
+        'constant_private',
+        'property_public',
+        'property_protected',
+        'property_private',
+        'phpunit',
+        'construct',
```
keradus: shouldn't this be `construct, phpunit` order ?

gharlan: hmm, don't know. My vote: +/-0

slavcodev:
Very useful fixer! Do you think it is worth also add the following rules:

+ All magic methods (__toString, __call, etc) should always be the first methods to increase readability, after the `__constructor`.
+ Ordering the overridden methods

tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
SpacePossum: just wait, we'll get `volatile` (and `virtual`) some time in PHP as well ;)

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
SpacePossum: +0 for swapping, maybe check with SF? (the [conversion](http://symfony.com/doc/current/contributing/code/standards.html#structure) does not give a conclusion)

SpacePossum:
>All magic methods 

+0

>Ordering the overridden methods

fixers have no knowledge about the class/trait hierarchy so this will not be possible

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
keradus: Actually I can't find any example in code.

> Declare public methods first, then protected ones and finally private ones. The exceptions to this rule are the class constructor and the setUp and tearDown methods of PHPUnit tests, which should always be the first methods to increase readability;

I just took order of elements in sentence. Nothing more strict.

keradus:
> Ordering the overridden methods

Even if we could I would be against it.

slavcodev:
> Even if we could I would be against it.

Can you comment more about your preference? This could be configurable.

keradus:
I don't like conflicting way of configure the fixer, and I have a strong feeling that it would be one of them.
But as I said - CS tool can't know about parent class (as it could even be in separated repository), so if one try to implement it there would be a lof of false negative fixes.

slavcodev:
Thanks, Got it.

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
gharlan: swapped phpunit and construct

keradus: thanks

keradus:
@gharlan what do you think about magic methods ?

gharlan:
would it be enough to have one special group "magic" which is used for all magic method except `__construct`?

keradus:
would `__construct` contains `__destruct` ?

gharlan:
don't know. 
at the moment: no.

keradus:
Symfony put desctuctor right after constructor

gharlan:
Or do we need something more flexible?

Example:

```
array(
        'use_trait',
        'constant_public',
        'constant_protected',
        'constant_private',
        'property_public',
        'property_protected',
        'property_private',
        ['method', '__construct'],
        ['method', '__destruct'],
        'method_public',
        'method_protected',
        'method_private',
        ['method', 'mySpecialMethodWhichShouldBeLastOne'],
)

**Comment by keradus** [comment]
Symfony uses: __construct, __destruct, __OTHER_MAGIC_METHODS.

I don't think that we need to handle user-input exception of order for methods (and not any element, which would be next exception).

**Comment by gharlan** [comment]
So should we include `__destruct` to "construct" group, and add a new group "magic" with all other magic methods?

**Comment by gharlan** [comment]
Or should we add "destruct" as own "group"?

**Comment by keradus** [comment]
Actually I'm fine for both.
I can see some cases when somebody could want to have it last, but if having it separated would need extra work don't focus on that

**Comment by SpacePossum** [comment]
:+1: for having the option to set `destructors` at the last position (i.e. own group). I don't expect it in this PR though, please just leave the option open so it can be added later (by whoever has the time) :)
Could we just not touch the `destructors` in the meantime? as grouping those with the `constructors` will make the fixer less flexible than it is atm.

**Comment by gharlan** [comment]
I will add "destruct" and "magic" group!

**Comment by gharlan** [comment]
fixer should stay in contrib level: https://github.com/symfony/symfony/pull/17980

do we want to use the fixer in this repo? Or is too complicated for mergings, too, @keradus?

**Comment by keradus** [comment]
> I will add "destruct" and "magic" group!

Thanks ;)


> do we want to use the fixer in this repo? Or is too complicated for mergings, too, @keradus?

Please do not activate the fixer for this repo by yourself. When it will be merged I will investigate the possibility to easily activate it.
For now merging a single fixer from 1.x line to 2.x line could take about 30 minutes...

File: src/Fixer/ClassNotation/OrderedClassElementsFixer.php [hunk 11]
```diff
@@ -0,0 +1,399 @@
...
+    {
+        if (null === $configuration) {
+            $configuration = self::$defaultOrder;
+        }
+
+        $this->typePosition = array();
+        $pos = 0;
+        foreach ($configuration as $type) {
+            if (!array_key_exists($type, self::$typeHierarchy)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Unknown class element type "%s".', $type));
```
keradus: Please add test for it

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
```diff
@@ -0,0 +1,399 @@
...
+            $parents = self::$typeHierarchy[$type];
+
+            if (!$parents) {
+                continue;
+            }
+
+            $pos = $this->findConfigurationPosition($parents);
+
+            if ($pos) {
+                return $pos;
```
keradus: it's never executed

gharlan:
> I will add "destruct" and "magic" group!

Should I add them to default order? `construct, destruct, magic, phpunit`?
Or maybe only add `destruct`?  `construct, destruct, phpunit`?

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
gharlan: good point! it is useless, will remove it.

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
gharlan: done

gharlan:
> I will add "destruct" and "magic" group!

done

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
```diff
@@ -0,0 +1,390 @@
...
+                array(T_STRING, 'tearDownAfterClass'),
+                array(T_STRING, 'setUp'),
+                array(T_STRING, 'tearDown'),
+            ), false)
+        ) {
+            return array('phpunit', strtolower($nameToken->getContent()));
+        }
+
+        if ('__' === substr($nameToken->getContent(), 0, 2)) {
+            return 'magic';
```
gharlan: Should I use a static array containing all magic method names, or just checking for names beginning with `__`?

http://php.net/manual/en/language.oop5.magic.php

> Caution
PHP reserves all function names starting with __ as magical. It is recommended that you do not use function names with __ in PHP unless you want some documented magic functionality.

keradus:
> Should I add them to default order? `construct, destruct, magic, phpunit`?
Or maybe only add destruct? `construct, destruct, phpunit`?

I would do it `construct, destruct, magic, phpunit`

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
keradus: checking `__` is enough

GrahamCampbell: > checking __ is enough

I don't agree. People could be using some crappy naming convention where _ means protected and __ means private.

keradus: Is there a way to get list of magic methods directly from PHP?
If not - I would treat all `__` prefixed methods as magic (so no need to code adjustment for newly introduced magic methods) and then, if one will follow that convention - he would simply not use that rule.

SpacePossum: >Is there a way to get list of magic methods directly from PHP?

AFAIK / could find, no. The methods-names are listed in the PHP src at a few [points](https://github.com/php/php-src/blob/master/Zend/zend_compile.h#L988), but I couldn't find an API function to retrieve those (would a easy pick PR @ PHP but wouldn't help you).

I'm :+1: for sorting all `__`-prefix methods as being magic, that is what PHP the docs states those should be anyway.

gharlan: my vote: :+1: for all `__` methods

gharlan:
> I would do it construct, destruct, magic, phpunit

I would like to use symfony standard as default even if the fixer is not added to symfony level.
Do you know if there is any cumulation of placement of magic methods in symfony code?

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1628#issuecomment-190912352

gharlan:
ah ok, sorry, forgot that comment ;)

then I will add "magic" to default config.

keradus:
:+1:

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
GrahamCampbell: > that is what PHP the docs states those should be anyway.

Surely we assume poor CS though?

keradus: for such case guy could turn off the rule

GrahamCampbell: lol

GrahamCampbell: meh, fine

keradus:
@gharlan please could you rebase? I would like to merge this.

SpacePossum:
:+1:  lets go :)

keradus:
@gharlan , it's outside this fixer to sort things also alphabetically in groups, eg:

from
```php
class Aaa
{
    public function fooC() {}
    public function fooB() {}
    private function fooA() {}
}
```

into
```php
class Aaa
{
    public function fooB() {}
    public function fooC() {}
    private function fooA() {}
}
```

gharlan:
At the moment it is not in scope of this fixer. Could be added, but in my opinion it should not be the default behavior.

keradus:
Thank you @gharlan.

SpacePossum:
:+1:  thanks @gharlan : D

PR #1630, opened 2015-12-22
Pass along the exception code.

SpacePossum:
Quick fix for (not yet released) configuration exception construction
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1586/files#r48208398

### DIFF
--- a/Symfony/CS/ConfigurationException/InvalidConfigurationException.php
+++ b/Symfony/CS/ConfigurationException/InvalidConfigurationException.php
@@ -23,10 +23,11 @@ use Symfony\CS\Console\Command\FixCommand;
 class InvalidConfigurationException extends \InvalidArgumentException
 {
     /**
-     * @param string $message
+     * @param string   $message
+     * @param int|null $code
      */
-    public function __construct($message)
+    public function __construct($message, $code = null)
     {
-        parent::__construct($message, FixCommand::EXIT_STATUS_FLAG_HAS_INVALID_CONFIG);
+        parent::__construct($message, null === $code ? FixCommand::EXIT_STATUS_FLAG_HAS_INVALID_CONFIG : $code);
     }
 }


Symfony/CS/ConfigurationException/InvalidConfigurationException.php
```diff
@@ -24,9 +24,10 @@ class InvalidConfigurationException extends \InvalidArgumentException
 {
     /**
      * @param string $message
+     * @param int    $code
      */
-    public function __construct($message)
+    public function __construct($message, $code = FixCommand::EXIT_STATUS_FLAG_HAS_INVALID_CONFIG)
```
GrahamCampbell: Maybe make the the default `null`, then do a ternary thing inside the code. That would be cleaner to me.

GrahamCampbell: Something like this:

```php
    /**
     * @param string   $message
     * @param int|null $code
     */
    public function __construct($message, $code = null)
    {
        parent::__construct($message, $code ?: FixCommand::EXIT_STATUS_FLAG_HAS_INVALID_CONFIG);
    }
```

keradus: :+1:

GrahamCampbell: Just what I'd do. D'no what anyone else thinks?

@keradus?

GrahamCampbell: Just missed your :+1: there. ;)

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1630#discussion_r48246011 ;P

keradus: yeah, 2 s difference :P

SpacePossum: that would not only check against `null` right? (like `0` would become the default as well?)

keradus: feel free to use strict comparision

SpacePossum: :+1: done

Symfony/CS/ConfigurationException/InvalidConfigurationException.php
```diff
@@ -24,9 +24,10 @@ class InvalidConfigurationException extends \InvalidArgumentException
 {
     /**
      * @param string $message
+     * @param int    $code
```
GrahamCampbell: int|null

SpacePossum: another down side of hiding the default form the signature

keradus: It is general style of default values - which in this one case indeed gives you nothing but consistency.

General issue:
`function foo($a, $b = 1, $c = 2) {}`
vs
`function foo($a, $b = null, $c = null) {}`

Try to pass $a and $c, but leave the $b as default in first example. You must know the default, and when it changes you need to change calls. In second example you just pass null. Everything would be cool if `default` keyword would be introduced, but it is not.
Also, that strategy of dealing is also used by native PHP functions.

GrahamCampbell: > another down side of hiding the default form the signature

I don't agree that it's a downside at all. Infact, I see it as better that the alternative.

GrahamCampbell: It allows you to skip defaults by passing null.

SpacePossum: agree to disagree, update :)

keradus:
:+1:

GrahamCampbell:
:+1:

gharlan:
:+1:

keradus:
Thank you @SpacePossum.

PR #1635, opened 2015-12-26
Removed symfony23 finder

GrahamCampbell:


sstok:
Explain?

GrahamCampbell:
Symfony use the default finder. Their config is committed in their code base...

GrahamCampbell:
https://github.com/symfony/symfony/blob/2.3/.php_cs

GrahamCampbell:
https://github.com/symfony/symfony/blob/2.7/.php_cs

sstok:
Ah, ok :)

:+1: then.

keradus:
What about Magento?
https://github.com/magento/magento2/blob/develop/.php_cs

GrahamCampbell:
I'd rather keep that to a different PR. Infact, I think we should remove that extra namespace we have too.

GrahamCampbell:
For config and finder.

keradus:
> I'd rather keep that to a different PR

why?
Clearing old configs/finders and keeping universal one seems to be quite atomic ;)

GrahamCampbell:
This probably needs a larger refactor. Someone else can take this if they want. If not, I'll get back to this when I have more time.

PR #1636, opened 2015-12-27
Better IDE support for custom tokens

gharlan:


Symfony/CS/Tokenizer/AbstractTransformer.php
```diff
@@ -55,8 +55,18 @@ public function registerCustomTokens()
...
     }
+
+    /**
+     * Next generated value for custom token.
+     *
+     * @return int
+     */
+    final protected static function getNextGeneratedCustomTokenValue()
+    {
+        return ++self::$lastGeneratedCustomTokenValue;
```
keradus: overcomplexity + performance loss?

Symfony/CS/Tokenizer/Transformer/ArrayTypehint.php
```diff
@@ -43,4 +43,12 @@ public function getCustomTokenNames()
     {
         return array('CT_ARRAY_TYPEHINT');
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function registerCustomTokens()
+    {
+        define('CT_ARRAY_TYPEHINT', self::getNextGeneratedCustomTokenValue());
```
keradus: in every Transformer you override method from `AbstractTransformer` to... do _the same_ thing.

keradus:
:-1:

Symfony/CS/Tokenizer/Transformer/ArrayTypehint.php
gharlan: In 2.x I would remove the method from `AbstractTransformer`.

keradus: keep sth simple and generic in one place or in 10... i prefer 1.

gharlan:
I'm not sure about this, too. Just an idea.

I don't like this:

bildschirmfoto 2015-12-27 um 12 00 57

I want to get auto completion etc.

Another idea: add a `Symfony/CS/Resources/custom_tokens` with content like this:

```php
define('CT_ARRAY_TYPEHINT', false);
define('CT_CLASS_CONSTANT', false);
define('CT_CURLY_CLOSE', false);
// ...
```

So the file would not be included, it would exist for ide support only.

Symfony/CS/Tokenizer/Transformer/ArrayTypehint.php
gharlan: I prefer a solution that is simple to use. And in my opinion the "undefined constants" are confusing.

gharlan: Of course, there should not be a performance loss, and I had to check this, if this pr had a chance to be merged.
But it seems that this is not the case.. ;)

keradus:
> And in my opinion the "undefined constants" are confusing.

this is problem o your IDE, not of PHP CS Fixer.

GrahamCampbell:
:-1:

GrahamCampbell:
Never optimize code for an IDE.

gharlan:
Ok, then closing ;)

PR #1640, opened 2015-12-31
IntegrationTest - fix directory separator

keradus:


Symfony/CS/Tests/IntegrationTest.php
```diff
@@ -25,14 +25,13 @@
      */
     protected static function getFixturesDir()
     {
-        return __DIR__.'/Fixtures/Integration';
+        return __DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'Integration';
     }
-
```
GrahamCampbell: this change looks wrong

keradus: yep

keradus: done, thx

PR #1641, opened 2015-12-31
AppVeyor/Travis - use GITHUB_OAUTH_TOKEN

keradus:


stof:
should it check whether the variable is defined ?

keradus:
I think it is not needed as it is defined directly in Travis configuration (not the file)

stof:
Well, if it is a secure env variable, it will not be available in PRs sent from a fork (to avoid allowing a PR to be sent which would dump the value of the env variable and so leak it)

keradus:
So... let us use the raw, readable format to allow all PRs use it ?

stof:
@keradus well, this can be discussed though. PRs will still benefit from the existing cache (as they reuse the cache of the target branch, which will have been populated by builds using the token), so they may not need the token.

keradus:
Indeed, in most cases PRs will use cache, but not always

keradus:
BTW, added token to AppVeyor as well

stof:
@keradus there is still a fallback to the slower clone then. the question is whether you want to have a github token available publicly (even though a token with no scope is not a big issue, it still allows people to make read requests on your behalf)

stof:
why have you cancelled the AppVeyor build ? you should let the build run as you changed the config file

keradus:
> there is still a fallback to the slower clone then.

So, when possible use cache, when not use dist with oauth (5000 limit), and clone is the last hope.
Isn't it nice?

> the question is whether you want to have a github token available publicly (even though a token with no scope is not a big issue, it still allows people to make read requests on your behalf)

I have no problem with that. That dummy token can be created for dummy account and always changed, right?

---------------

> why have you cancelled the AppVeyor build ?

I will restart it if we agreed to merge the PR.
For now I'm fighting with non-working AppVeyor on another PRs and AppVeyor is busy today (and I don't want to wait an hour to get my job done)

GrahamCampbell:
> For now I'm fighting with non-working AppVeyor on another PRs and AppVeyor is busy today (and I don't want to wait an hour to get my job done)

Is there not something else better out there? AppVeyor is always slow as hell.

keradus:
Feel free to find any ;)

GrahamCampbell:
> Feel free to find any ;)

Haha. Might be faster to just write one. :P

GrahamCampbell:
TBH, that's not too crazy a thought, lol.

keradus:
OK, then please focus on main topic of this PR ;)

keradus:
So...
:+1: ? ;)

stof:
:+1: from me

sstok:
> Is there not something else better out there? AppVeyor is always slow as hell.

Windows, with a 32bit php version (that needs to be reinstalled every time) what do you expect :)
At least it's free in price.

PR #1642, opened 2015-12-31
AppVeyor - install dev deps as well

keradus:


GrahamCampbell:
We need `--no-dev` still surely?

GrahamCampbell:
Just saw the description. Why would we want to do this?

keradus:
Backported from 2.x line. Without this tests won't run (we have fixtures for test as well, and without this PSR4 autoloader wont load things from tests dir)

keradus:
Also, we install devs on Travis ;)

GrahamCampbell:
> Also, we install devs on Travis ;)

That's because we actually use the dev-dependencies on travis. We do not use them on windows.

keradus:
`--no-dev` not only disable dependencies, but also autoloader.

GrahamCampbell:
Ah, of course. This PR is thus required for your PSR-4 changes to work.

PR #1643, opened 2015-12-31
AppVeyor - collect code coverage

keradus:


keradus:
refs:
- https://github.com/satooshi/php-coveralls/issues/171
- https://github.com/satooshi/php-coveralls/issues/172

appveyor.yml
```diff
@@ -12,18 +12,30 @@ cache:
...
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --prefer-source --ansi --no-interaction --no-dev
+    - php C:\tools\composer.phar update --no-interaction
 
 test_script:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\phpunit.phar --verbose
+    - php -c C:\tools\php\php-with_xdebug.ini C:\tools\phpunit.phar --verbose --coverage-clover build\logs\clover.xml
```
stof: do we really need to generate code coverage there too ?

GrahamCampbell: well, by that argument, we should only generate it on one php version on travis too

GrahamCampbell: the idea here is surely that execution paths on different php versions and OSs might not be the same

keradus: We already had some cases that works on unix but not on Windows, and as it great to fix them and test on Windows too, the fix may generate code that is reached only on Windows. So if possible - the code coverage should be collected as well.
For now PHP Coveralls won't work under Windows.

GrahamCampbell: > For now PHP Coveralls won't work under Windows.

lol, windows

keradus: No. The problem is PHP Coveralls.
It everything uses `/` in checks reading paths from disc, so all path-related checks fail

GrahamCampbell: The best sollution there is just call `realpath` rather than messing with the dir separator constant.

keradus: whatever solution - goto PHP Coveralls ;)

keradus:
Closing, as should be done with phpdbg instead of xdebug

PR #1646, opened 2016-01-01
Remove non-default Config/Finder classes

keradus:
Replaces #1635

PR #1647, opened 2016-01-01
Deprecate non-default Configs and Finders

keradus:
ref #1646

### DIFF
--- a/Symfony/CS/Config/MagentoConfig.php
+++ b/Symfony/CS/Config/MagentoConfig.php
@@ -20,6 +20,11 @@ class MagentoConfig extends Config
 {
     public function __construct()
     {
+        @trigger_error(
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
:+1:

SpacePossum:
:+1:

gharlan:
:+1:

GrahamCampbell:
Surely we need to deprecate the other 2 class too?

GrahamCampbell:
Since we're moving them?

GrahamCampbell:
We should move them by extension, then deprecate the originals shouldn't we?

keradus:
I was thinking about sth like that too, but I already have too many work in progress.
Could you send a PR ?

GrahamCampbell:
Sorry, have no time at all this week. ;(

PR #1649, opened 2016-01-01
Consider the php version when caching

### LINKED ISSUE #1619: Also included php version in cache
If someone updates from php 5.6 to 7.0, we probably want to consider the cache file stale.

GrahamCampbell:
Closes #1619.

### DIFF
--- a/src/FileCacheManager.php
+++ b/src/FileCacheManager.php
@@ -110,13 +110,13 @@ final class FileCacheManager
         return $result;
     }
 
-    private function isCacheStale($cacheVersion, $rules)
+    private function isCacheStale($php, $version, $rules)
     {
         if (!$this->isCacheAvailable()) {
             return true;
         }
 
-        return ToolInfo::getVersion() !== $cacheVersion || $this->rules !== $rules;
+        return PHP_VERSION !== $php || ToolInfo::getVersion() !== $version || $this->rules !== $rules;
     }
 
     private function readFromFile()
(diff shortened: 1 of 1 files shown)


keradus:
I don't like this.
I often store cache in CI cache and use it across all PHP versions.

GrahamCampbell:
Well, that's surely incorrect, because the behaviour of fixing a file on php 7 in php-cs-fixer is not the same as fixing on php 5.3.

keradus:
And 5.5.1 vs 5.5.2 is the same.
PHP7 is totally different case

GrahamCampbell:
Well, php 5.5.2 vs 5.5.1 might be different if some of them had a bug affecting it.

GrahamCampbell:
At the very least could we consider the major and minor versions, but not patch?

keradus:
I'm not against it, just say don't like it in current form ;)

Let the other say their opinion ;)

SpacePossum:
Like the idea :+1: 
Maybe use `PHP_VERSION_ID` over `PHP_VERSION`, it would take out out the extra bit (like RC etc.) I think?

GrahamCampbell:
> Maybe use PHP_VERSION_ID over PHP_VERSION, it would take out out the extra bit (like RC etc.) I think?

We want that extra part though, because RC1 !== RC2.

keradus:
BTW, having tests would be great ;)

keradus:
Any other thoughts?

keradus:
Any other thoughts?

keradus:
Anyone ?

GrahamCampbell:
Maybe a better way might be to select a cache file based on the php version?

keradus:
one can already choose non-default filename for cache file

GrahamCampbell:
I was meaning that php-cs-fixer could suffix that name with the php version.

GrahamCampbell:
To allow easy cashing across different php versions automatically.

keradus:
OK, sounds interesting

GrahamCampbell:
Just rebased this, but don't have time to do much else with it, so if you don't like it as is, feel free to just close this. I'm not massively fussed.

keradus:
Actually I was waiting for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1649#issuecomment-177268279

GrahamCampbell:
> Actually I was waiting for #1649 (comment)

GrahamCampbell:
> Actually I was waiting for #1649 (comment)

Yeh, well, not sure people would actually want that because it means they can't control the filename if we're arbitrarily appending stuff to it.

keradus:
Thank you @GrahamCampbell.

Issue #1579, opened 2015-12-07
using stdin and stdout


garak:
Is there any change to use the fixer with stdin/stdout?
The use case is: I don't want to pass a path and get the file(s) on that path to be fixed, I want to pass a content as stdin and get the fixed content as stdout

stof:
``php-cs-fixer fix -`` should do the work

GrahamCampbell:
Yeh, I'm pretty sure we got this fixed a while back?

garak:
Trying this:
``cat app/AppKernel.php > php-cs-fixer fix -``
and getting an error complaining that ``fix`` is not a file or directory

gharlan:
`|` instead of `>`

garak:
With pipe I'm getting this output:
``Loaded config from "/home/garak/foo/.php_cs"
   1) stdin.php
Checked all files in 0.152 seconds, 6.250 MB memory used``

And nothing else happens. The file is not fixed (but my purpose isn't that) and I don't get the fixed content as output

SpacePossum:
`Checked` mean `dry-run` is enabled, so no fixing is done

stof:
use the ``--diff`` option to get the diff in the output

stof:
hmm, except that all the progress info are also sent to stdout, rather than stderr, so there is no way to get the diff alone easily

sampablokuper:
Please can this issue be reopened? **It has not been fixed.**

@stof [wrote]:
> `cat file > php-cs-fixer fix -` should do the work

This, of course, is wrong, because `>` is an inappropriate shell redirection character to use here. Rather than piping the contents of `file` to the command `php-cs-fixer`, it would instead write the contents of `file` to a file called `php-cs-fixer`, creating the latter if necessary. But even we we use `|` instead of `>` per @gharlan 's [correction](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1579#issuecomment-162616905) of @stof's error, the functionality requested in @garak's [opening comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1579#issue-120824852) ("I don't want to pass a path and get the file(s) on that path to be fixed, I want to pass a content as stdin and get the fixed content as stdout") is still not fully present:

```
$ cat 1.php | php-cs-fixer fix --dry-run -
Loaded config default.
Using cache file ".php_cs.cache".
   1) php://stdin

[...]

PR #1654, opened 2016-01-05
Split output to stderr and stdout

SpacePossum:
For more correct testing I need https://github.com/symfony/symfony/pull/17255 to be resolved.

issue: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1579 (and others)

### DIFF
--- a/Symfony/CS/Console/Command/FixCommand.php
+++ b/Symfony/CS/Console/Command/FixCommand.php
@@ -287,8 +287,12 @@ EOF
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         // setup output
-        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : null;
-        if ($stdErr && extension_loaded('xdebug')) {
+        $stdErr = $output instanceof ConsoleOutputInterface
+            ? $output->getErrorOutput()
+            : ('txt' === $reporter->getFormat() ? $output : null)
+        ;
+
+        if (null !== $stdErr && extension_loaded('xdebug')) {
             $stdErr->writeln(sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.'));
         }
 
@@ -346,8 +350,8 @@ EOF
                 throw new InvalidConfigurationException(sprintf('The config file "%s" does not return a "Symfony\CS\ConfigInterface" instance. Got: "%s".', $configFile, is_object($config) ? get_class($config) : gettype($config)));
             }
 
-            if ('txt' === $input->getOption('format')) {
-                $output->writeln(sprintf('Loaded config from "%s"', $configFile));
+            if (null !== $stdErr && $configFile) {
+                $stdErr->writeln(sprintf('Loaded config from "%s".', $configFile));
             }
         } else {
             $config = $this->defaultConfig;
@@ -383,11 +387,11 @@ EOF
             ->resolve();
 
         $config->fixers($resolver->getFixers());
-        $showProgress = $resolver->getProgress();
+        $showProgress = null !== $stdErr && $resolver->getProgress();
 
         if ($showProgress) {
-            $fileProcessedEventListener = function (FixerFileProcessedEvent $event) use ($output) {
-                $output->write($event->getStatusAsString());
+            $fileProcessedEventListener = function (FixerFileProcessedEvent $event) use ($stdErr) {
+                $stdErr->write($event->getStatusAsString());
             };
 
             $this->fixer->setEventDispatcher($this->eventDispatcher);
@@ -401,7 +405,7 @@ EOF
         if ($showProgress) {
             $this->fixer->setEventDispatcher(null);
             $this->eventDispatcher->removeListener(FixerFileProcessedEvent::NAME, $fileProcessedEventListener);
-            $output->writeln('');
+            $stdErr->writeln('');
 
             $legend = array();
             foreach (FixerFileProcessedEvent::getStatusMap() as $status) {
@@ -410,7 +414,7 @@ EOF
                 }
             }
 
-            $output->writeln('Legend: '.implode(', ', array_unique($legend)));
+            $stdErr->writeln('Legend: '.implode(', ', array_unique($legend)));
         }
 
         $i = 1;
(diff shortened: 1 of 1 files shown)


keradus:
What about switching to [SymfonyStyle](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Style/SymfonyStyle.php) ?
`$io = new SymfonyStyle($input, $output);`

ref https://github.com/symfony/symfony/issues/15986

GrahamCampbell:
We'd have to raise our min console vesion.

keradus:
Raising dep version for just raising it is pointless. but rising it for functionality is OK ;)

keradus:
BTW, then you won't need to test if console printer works on our end, just rely on vendor code (and it's tests)

SpacePossum:
hmm even more work to be done in symfony/console... ;)

I don't like the current output construction, because if an output is given to a receiver/output writter it has to check if there is another output attached to it that is even more appropriate. This is causing code like this all over the place;
https://github.com/symfony/symfony/pull/15795/files#diff-425e56e3b866cb257f587805f7b2ee54R41
why isn't the error output passed to that method in the first place?
and why not set `hasErrOutput` on the `OutputInterface` that would at least remove the instance checks...?

anyway, `SymfonyStyle::error` still writes to stdout IIRC.. let me check the coming days :)

keradus:
> why isn't the error output passed to that method in the first place?

For ensuring that if it is possible the stdErr is used.
If it would be added to method's prototype then it would break BC, and if not (last & nullable param) then nobody would pass it and stdErr would not be used, and it is not the goal. Bit ugliness for nice output out of the box without changing user code.

keradus:
> anyway, `SymfonyStyle::error` still writes to stdout IIRC.. let me check the coming days :)

yes, and changing it will fix https://github.com/symfony/symfony/issues/15986 ;)

keradus:
> We'd have to raise our min console version.

To use SymfonyStyle we will need to bump to v2.7, which has raised PHP requirement:
```diff
-        "php": ">=5.3.3"
+        "php": ">=5.3.9"
```

I'm ok with that, since it bump only PATCH version

GrahamCampbell:
Maybe 2.x is a better place to make this change?

keradus:
Could you justify that?

SpacePossum:
>Could you justify that?

it would safe you the merge hell ;)

keradus:
There should not be a lot of differences between 1.x and 2.x since the output of both should be the same, but it is fine for me.

keradus:
What about creating our own Style on top of SymfonyStyle, so we could add more things there (like our progress) ?

stof:
SymfonyStyle is about formatting the output. Telling that some styles must go to stdout and some other styles go to stderr is wrong. This is misunderstanding the goal of stderr and stdout (naming them stderr and stdout does not help understand their goal either btw).
any progress reporting output should also go to stderr, not to stdout (it is not part of the output of the program, which would be piped to another program)

keradus:
Yep, we want to push the progress to stderr.
@stof, then what about https://github.com/symfony/symfony/issues/15986 ? Is it going to be the proposed way or the issue will be rejected ?

keradus:
@SpacePossum I think we won't hear anything about this from Symfony soon.
Are we able to handle it on our side at low cost?

SpacePossum:
>@SpacePossum I think we won't hear anything about this from Symfony soon.

Why do you think, all members are busy? I'll ping again, its been 10 days since my last proposal, hope I ping the right people ^-^
I could write a similar fix in the mean time for the fixer, but it would not be handy if my fix over at symfony gets turned down.

keradus:
@SpacePossum , https://github.com/symfony/symfony/pull/15794 has been merged. Would it help us?

SpacePossum:
@keradus that PR addresses some helpers we don't use ATM (and probably won't any time soon). 
I got an answer on 15986 from Fabien that contains enough pointer to finish it, I hope to find some time this weekend to do so.

SpacePossum:
My PR to SF to get testing options for both stdout and stderr has been merged, however it has been done on 3.x line only. I would like to suggest to move on without it by adding a bit of a polyfill until the fixer requirements reach the 3.x line of packages of SF.
I would really like to wrap this one up to close the issues related to outputting to the wrong stream(s) making it harder than needed to integrate the fixer with other applications. I would like to suggest to rebase and rework this after #1824 has been done as the PRs will have some conflicts among each other. Ping @keradus for thoughts on this :)

keradus:
fine for me

keradus:
ping @SpacePossum

SpacePossum:
I think a wait out #1824 since it will cover the output format testing. This means the tests here can be removed and this PR can focus on the used output streams only. I'll work on this soon to get the test logic in place we are going to need though :)

keradus:
#1824 is merged. please update

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -420,42 +420,42 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                        $stdErr->writeln(sprintf('[%.3f s] %s', $event->getDuration() / 1000, $file));
                     }
 
-                    $output->writeln('');
+                    $stdErr->writeln('');
                 }
 
                 $fixEvent = $this->stopwatch->getEvent('fixFiles');
-                $output->writeln(sprintf('%s all files in %.3f seconds, %.3f MB memory used', $input->getOption('dry-run') ? 'Checked' : 'Fixed', $fixEvent->getDuration() / 1000, $fixEvent->getMemory() / 1024 / 1024));
+                $stdErr->writeln(sprintf('%s all files in %.3f seconds, %.3f MB memory used', $input->getOption('dry-run') ? 'Checked' : 'Fixed', $fixEvent->getDuration() / 1000, $fixEvent->getMemory() / 1024 / 1024));
```
GrahamCampbell: seems wrong to print information to stderr when it's not an error

SpacePossum: It is not, the naming of the streams is confusing. It is better, IMO, to send all kind of information to stdErr and only parsable formatted info to stdOut

SpacePossum:
Reworked, please note this PR is now for ^1.12 only.

keradus:
you mean that non of those changes should be merged into 2.x, right?

SpacePossum:
>you mean that non of those changes should be merged into 2.x, right?

indeed, it is all good on 2.x :)

keradus:
Cool ;)

keradus:
Thank you @SpacePossum.

TomasVotruba:
I'm writing post about CLI output styles PHP packages use and I Googled this is the 1 place in PHP CS Fixer that mentions `SymfonyStyle`. 

It was declided due to min-requiremeent bump to Symfony 2.8.

Now the [min. version is Symfony 3.2](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1c10240da97479274fd40a136c3857ff94f7f93f/composer.json#L24), so it's ready :+1: 

Just FYI

Issue #1656, opened 2016-01-06
Broken switch case fixer


GrahamCampbell:
This should break the fixer:

```php
        switch ($json['type']) {
            case 'zikula-module':
            case 'zikula-theme':
            case 'zikula-plugin':
                break;
            default;

                return false;
```

SpacePossum:
maybe put this case in the test of the fixer and into a PR and we all see it break ;)

GrahamCampbell:
I didn't have much time so I just dumped it into an issue. Later on, I'll add a test case and look into why it's actually failing.

keradus:
To put some more valid code:
```php
switch($foo) {
    case '123';
    case '456';
        return 1;
    default;
        return 2;
}
```

the problem is that case may be followed by `;` as well, and we look for `:` only

SpacePossum:
can be closed :)

keradus:
thx

PR #1657, opened 2016-01-07
SwitchCaseSpaceFixer - Fix spacing between 'case' and semicolon

SpacePossum:
@see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1656

### DIFF
--- a/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
+++ b/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
@@ -38,7 +38,7 @@ final class SwitchCaseSpaceFixer extends AbstractFixer
                     ++$ternariesCount;
                 }
 
-                if ($tokens[$colonIndex]->equalsAny(array(':', ';'), false)) {
+                if ($tokens[$colonIndex]->equalsAny(array(':', ';'))) {
                     if (0 === $ternariesCount) {
                         break;
                     }
--- a/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
+++ b/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
@@ -38,7 +38,7 @@ final class SwitchCaseSpaceFixer extends AbstractFixer
                     ++$ternariesCount;
                 }
 
-                if ($tokens[$colonIndex]->equals(':') || $tokens[$colonIndex]->equals(';')) {
+                if ($tokens[$colonIndex]->equalsAny(array(':', ';'), false)) {
                     if (0 === $ternariesCount) {
                         break;
                     }
--- a/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
+++ b/Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
@@ -38,7 +38,7 @@ final class SwitchCaseSpaceFixer extends AbstractFixer
                     ++$ternariesCount;
                 }
 
-                if ($tokens[$colonIndex]->equals(':')) {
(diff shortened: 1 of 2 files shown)


Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
```diff
@@ -38,7 +38,7 @@ public function fix(\SplFileInfo $file, $content)
                     ++$ternariesCount;
                 }
 
-                if ($tokens[$colonIndex]->equals(':')) {
+                if ($tokens[$colonIndex]->equals(':') || $tokens[$colonIndex]->equals(';')) {
```
keradus: `equalsAny`

keradus:
btw, 1.11.x

keradus:
Please ignore fabbot.io this time

gharlan:
> btw, 1.11.x

The fixer does not exist in 1.11 ;)

keradus:
cool, thanks @gharlan !

Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
SpacePossum: check, done :)

Symfony/CS/Fixer/PSR2/SwitchCaseSpaceFixer.php
```diff
@@ -38,7 +38,7 @@ public function fix(\SplFileInfo $file, $content)
                     ++$ternariesCount;
                 }
 
-                if ($tokens[$colonIndex]->equals(':')) {
+                if ($tokens[$colonIndex]->equalsAny(array(':', ';'), false)) {
```
keradus: Plesae remove the last parameter, it wont be used in this case.

SpacePossum: the default is true, but it is not needed so false should be faster in execution doesn't it?

gharlan: No, a case-sensitive (`true`) check is faster than a case-insensitive (`false`).
But in this case we only have non-array tokens, so the parameter is not used:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tokenizer/Token.php#L117

SpacePossum: >Maybe you've misread the parameter as $caseInsensitive?

I did :|  thanks!

Symfony/CS/Tests/Fixer/PSR2/SwitchCaseSpaceFixerTest.php
```diff
@@ -272,6 +272,32 @@ public function provideFixCases()
...
+                '<?php
+                    switch($foo) {
+                        case 4  :  ; ;
+                        case 31 + test(";") ;  ; ; ;;
+                        case 1 + test(";") ; // ;
+                        case (1+2/*;*/) ;
+                        case 1  ;
+                        case 2 ;
+                            return 1;
+                        default;
```
keradus: would the space after `default` be removed?
(i bet yes, but nice to test;P)

SpacePossum: add https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/2586e04d2ebb217643c94a1ba60e9915bbe8c142 :)

keradus: thanks ;)

keradus:
Thank you @SpacePossum.

PR #1659, opened 2016-01-07
NativeFunctionCasingFixer - Add

SpacePossum:
This fixer ensures native functions (i.e. those defined by PHP itself) are called in the casing the function is declared.

FYI: There are zero hits on Symfony (3).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -615,6 +615,11 @@ Choose from the list of available fixers:
                         closing semicolon are
                         prohibited.
 
+* **native_function_casing** [contrib]
+                        Function defined by PHP should
+                        be called using the correct
+                        casing.
+
 * **newline_after_open_tag** [contrib]
                         Ensure there is no code on the
                         same line as the PHP open tag.
--- a/Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
+++ b/Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
@@ -26,6 +26,7 @@ final class NativeFunctionCasingFixer extends AbstractFixer
     {
         $tokens = Tokens::fromCode($content);
         for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            // test if we are at a function all
             if (!$tokens[$index]->isGivenKind(T_STRING)) {
                 continue;
             }
@@ -41,6 +42,12 @@ final class NativeFunctionCasingFixer extends AbstractFixer
                 continue;
             }
 
+            // do not though the function call if it is to a function in a namespace other than the default
+            if ($tokens[$functionNamePrefix]->isGivenKind(T_NS_SEPARATOR) && $tokens[$tokens->getPrevMeaningfulToken($functionNamePrefix)]->isGivenKind(T_STRING)) {
+                continue;
+            }
+
+            // test if the function call is to a native PHP function
             static $nativeFunctionNames = null;
             if (null === $nativeFunctionNames) {
                 $nativeFunctionNames = $this->getNativeFunctionNames();
--- a/Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
+++ b/Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class NativeFunctionCasingFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(T_STRING)) {
+                continue;
+            }
+
+            $next = $tokens->getNextMeaningfulToken($index);
+            if (!$tokens[$next]->equals('(')) {
+                $index = $next;
+                continue;
+            }
+
+            $functionNamePrefix = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$functionNamePrefix]->isGivenKind(array(T_DOUBLE_COLON, T_NEW, T_OBJECT_OPERATOR, T_FUNCTION))) {
+                continue;
+            }
+
+            static $nativeFunctionNames = null;
+            if (null === $nativeFunctionNames) {
+                $nativeFunctionNames = $this->getNativeFunctionNames();
+            }
+
+            $lower = strtolower($tokens[$index]->getContent());
+            if (!array_key_exists($lower, $nativeFunctionNames)) {
+                continue;
+            }
+
+            $tokens[$index]->setContent($nativeFunctionNames[$lower]);
+
+            $index = $next;
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Function defined by PHP should be called using the correct casing.';
+    }
+
+    private function getNativeFunctionNames()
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
This should be marked as risky.

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Function defined by PHP should be called using the correct casing.';
+    }
+
+    private function getNativeFunctionNames()
```
keradus: this is slow... 
no need to recalculate this for every single file

keradus:
Could this take adv. from AbstractFunctionReferenceFixer (#31617) ?

SpacePossum:
>This should be marked as risky.

Why? Function names (and calling those) is case insensitive

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
SpacePossum: is there an example how such a thing is done by another fixer?

GrahamCampbell:
> Why? Function names (and calling those) is case insensitive

Because you might be replacing other functions of the same name that are just namespaced.

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
gharlan: @keradus it is called only once: https://github.com/SpacePossum/PHP-CS-Fixer/blob/f2aff50364110df5be9bc3e267826726a6b62f67/Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php#L44-L47

But I would move the static variable inside this method.

SpacePossum:
>Could this take adv. from AbstractFunctionReferenceFixer (#31617) ?

Not sure, that code is useful for finding functions by the name of the function, here I do it the other way around; when a function call is found; check if it should be changed.

>Because you might be replacing other functions of the same name that are just namespaced.

that still wouldn't break the code?

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
GrahamCampbell: > it is called only once

It's called once... for each file.

GrahamCampbell: Might be an idea to just calculate this in the constructor. I dislike the static hacks.

keradus:
@SpacePossum check this out:
https://3v4l.org/ZJL1L

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
gharlan: hmm why?

gharlan:
I agree with @SpacePossum that it would not change code behavior.
But maybe there would be undesired changes anyway.

But is it "risky"?

stof:
> Because you might be replacing other functions of the same name that are just namespaced.

even these are case insensitive

README.rst
```diff
@@ -615,6 +615,11 @@ Choose from the list of available fixers:
                         closing semicolon are
                         prohibited.
 
+* **native_function_casing** [contrib]
```
stof: I would make it at the symfony level actually

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
keradus: Don't like the idea of putting this in constructor.
The `Fixer` (main one) create a lot of `*Fixer`s (like StrictFixer), then play with them, eg _remove_ the ones that have not been choosen to use.

Putting things into constructors will make them run every time, even if one not want to use that fixer.

README.rst
keradus: Risky fixer on Symfony level?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1659#issuecomment-169721563

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
GrahamCampbell: It would make sense to go into the constructor if we made it abstract.

README.rst
SpacePossum: it is not risky, you can change the casing in the example of both the call and the function name in any random way and it would not effect the code behavior

keradus: Please, add a test to PR:
```php
    array(
        '<?php
        namespace Bar {
            function strlen($str) {
                return "overriden" . \strlen($str);
            }
        
            echo strlen('foo');
        }',
    ),
```

gharlan: @keradus the tests should pass because `strlen` is the correct case ;)
maybe you want tests with overriden `StrLen` or something like this?

then it would not pass. but the change would not change code behavior
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1659#issuecomment-169722526)

GrahamCampbell: > then it would not pass. but the change would not change code behavior

True, but it hasn't made a correct change though. It's renamed a function that wasn't a native function.

SpacePossum: sure thing :)
https://github.com/SpacePossum/PHP-CS-Fixer/commit/4f58a28e8357bcb9c8e1b28b21a132b201b3808c

keradus: > maybe you want tests with overriden StrLen or something like this?

why not, please add them as well ;)

keradus: no, actually replace them :D

I updated :  https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1659/files#r49096853

sorry for confusion

gharlan: they will fail. same for this test case:

```php
    array(
        '<?php
        namespace Foo;

        echo strLen('foo'); // overriden in another file, so it should not be changed
    ),
```

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
SpacePossum: >It's called once... for each file.

This seems not to be the case, it is called only once (i.e. the fixer is re-used for each file).

README.rst
SpacePossum: Sorry all, I missed the comments after the initial request for the tests.
The last case from @gharlan (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1659#discussion_r49101273) will indeed be changed by the fixer. 
If it should be changed or not is up for discussion, however changing it will not break the code (i.e. change behavior). 
For example:

```php
<?php
namespace Bar {
    function strLen($str) {
        return 'overriden' . \strlen($str);
    }
}
```

Calling:

```php
namespace {
    echo \Bar\STRLEN('xxx');
}
```

```php
namespace {
    echo \Bar\strlen('xxx');
}
```

```php
namespace {
    echo \Bar\strLEN('xxx');
}
```

Or:

```php
namespace Bar{
    echo strlen('xxx');
}
```

```php
namespace Bar{
    echo STRLEN('xxx');
}
```

```php
namespace Bar{
    echo strLEn('xxx');
}
```

will all result in the same code, so not risky.

Let me check if this is touched by the fixer:
```php
namespace {
    echo \Bar\STRLEN('xxx');
}
```

the other cases like;

```php
namespace Bar{
    echo strLEn('xxx');
}
```

cannot be tested against by the fixer as it would require knowledge of the whole code base.

keradus: > cannot be tested against by the fixer as it would require knowledge of the whole code base

Isn't this one of original considering all function-replacing fixers as risky?
As we only know what happens in _single_ file?

gharlan: But in the other function-replacing fixers the change could change code behavior. Here not.
So strictly speaking it is not risky.

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
keradus: Not it is cached inside static variable inside `fix` method.
I would move that into this method - this will encapsulate result in cache in same place that it is computing it.

BTW, creating static var in `for` loop isn't prettiest solution ;)

keradus:
Thank you @SpacePossum.

Symfony/CS/Fixer/Contrib/NativeFunctionCasingFixer.php
SpacePossum: I see, indeed, thanks! and thanks for the clean up :)

PR #1663, opened 2016-01-10
DuplicateSemicolonFixer - Remove duplicate semicolons even if there are comments between those

SpacePossum:
With this PR duplicate semicolons will be removed even if there are comment between the semicolons. 

~~WIP because it seems there might be a priority issue with the multiple empty/blank and trailing whitespace on empty line fixers.~~ added

### DIFF
--- a/Symfony/CS/Fixer/Symfony/DuplicateSemicolonFixer.php
+++ b/Symfony/CS/Fixer/Symfony/DuplicateSemicolonFixer.php
@@ -41,7 +41,7 @@ class DuplicateSemicolonFixer extends AbstractFixer
                 continue;
             }
 
-            $prevIndex = $tokens->getPrevNonWhitespace($index);
+            $prevIndex = $tokens->getPrevMeaningfulToken($index);
 
             if (!$tokens[$prevIndex]->equals(';')) {
                 continue;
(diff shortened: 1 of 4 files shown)


SpacePossum:
I've a hard time trying to figure out why the coverage is reported as being down.

keradus:
> Coverage decreased (-0.01%) to 88.587%

This one? Ignore it, it's a measurement error.

SpacePossum:
I see, the test first failed but now its green :)
Btw. I'm planning to add more logic to this fixer on the master line to remove semicolons after close braces of control structures as well. I think it would be better to rename it to UseSemiColonFixer than, any thoughts?

keradus:
`;;` and `if (1) {};` are 2 different cases.
Add new fixer on 1.x line

gharlan:
IMO it should be one fixer, that removes empty statements (`/*nothing*/;`).
It does not matter what kind of statement is before the empty statement (ending with `;` or `}`).
So I would name it `NoEmptyStatementFixer`.

gharlan:
So I would also fix this case with the fixer:

```php
function foo() 
{
    ;
}

keradus:
:+1: for @gharlan 
Please, create complex fixer for 1.x line and send PR to remove trivial one for 2.x line.

keradus:
Thank you @SpacePossum.

PR #1666, opened 2016-01-11
new Runner, ProcessOutputInterface, DifferInterface and ResultInterface

keradus:
Introduce: new Runner, ProcessOutputInterface, DifferInterface and ResultInterface
Alternative for #1618 

```bash
D:\GIT\fork\PHP-CS-Fixer (master)
time php php-cs-fixer fix src
Loaded config from "D:\GIT\fork\PHP-CS-Fixer\.php_cs.dist"
Fixed all files in 42.924 seconds, 8.000 MB memory used
```

```bash
D:\GIT\fork\PHP-CS-Fixer (2.x-async_linting)
time php php-cs-fixer fix src
Loaded config from "D:\GIT\fork\PHP-CS-Fixer\.php_cs.dist"
Fixed all files in 27.270 seconds, 8.000 MB memory used
```


keradus:
This is just a proof of concept, if we agreed to go that way it will need more work like filling the docs and fix tests

gharlan:
At the moment the linting process is started for all files, regardless if they are cached or not..

keradus:
:(

GrahamCampbell:
> At the moment the linting process is started for all files, regardless if they are cached or not..

Yeh, there's no way this is allowed. ;(

src/Linter/ProcessResult.php
```diff
@@ -0,0 +1,47 @@
...
+namespace Symfony\CS\Linter;
+
+use Symfony\Component\Process\Process;
+
+/**
+ * TODO.
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class ProcessResult
```
stof: missing interface implementation

src/Linter/NullResult.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+namespace Symfony\CS\Linter;
+
+/**
+ * TODO.
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class NullResult
```
stof: missing interface implementation

stof: missing interface implementation

keradus:
Yeah, as I sad it is a POC, is it worth to go further ?

gharlan:
> is it worth to go further ?

:+1:

src/FileLintingIterator.php
```diff
@@ -0,0 +1,70 @@
...
+namespace Symfony\CS;
+
+use Symfony\CS\Linter\LinterInterface;
+
+/**
+ * TODO
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class FileLintingIterator extends \CachingIterator
```
gharlan: is it intended intended that there are two nested CashingIterators?

keradus: Nope, this one is enought. good catch

keradus:
> At the moment the linting process is started for all files, regardless if they are cached or not..

For the current master we have two *public* methods in main `Fixer` class:
- `fix` - for fixing everything
- `fixFile` for fixing single file
For now, the `fixFile` is taking care of cacheing and linting and `fix` simply iterate over all files and calls `fixFile` for each of them.
We want to lint one file ahead, so it should be done in `fix` method. But also we don't want to lint if cache is valid, so we need to move cache checking to `fix` as well.

For that I think we should encapsulate `fixFile` method and only expose interface to fix things in bulk process (one could always simply pass collection of files to be fixed with whole one element).
I have started to move logic from `fixFile` to `fix` and fixed the issue with cache.

It just show the direction, the refactoring of `fix` vs `fixFile` is not done yet.

What do you think ?

src/Fixer.php
```diff
@@ -166,22 +200,12 @@ public function fixFile(\SplFileInfo $file, array $fixers, $dryRun, $diff, FileC
...
-
-            return;
-        }
-
         try {
-            $this->linter->lintFile($file->getRealPath());
+            if (null !== $this->currentFileLintingResult) {
+                $this->currentFileLintingResult->check();
+            } else {
+                $this->linter->lintFile($file->getRealPath())->check();
```
keradus: this line could be removed when `fixFile` will become private

keradus:
Any thoughts?

gharlan:
I think, the direction is good! :+1:  ;)

keradus:
Then `Fixer` class needs a big rewrite...

keradus:
I would made it more state aware... passing everything to every method as parameter is pointless...

keradus:
Runner of fixers has been rewritten from scratch.
A bit more work needed for linting in async way

keradus:
(Please ignore 5.3 for now)

keradus:
Sorry for taking so long.
I have 5 branches with different approach for this problem. This one works the best.

I updated first post.

keradus:
@gharlan please, could you take a look?

src/Runner/Runner.php
```diff
@@ -10,100 +10,110 @@
...
      * @var ErrorsManager
      */
-    protected $errorsManager;
+    private $errorsManager;
 
     /**
-     * Linter instance.
-     *
-     * @var LinterInterface
+     * @var type FileCacheManager
```
gharlan: type?

src/Runner/FileLintingIterator.php
```diff
@@ -0,0 +1,87 @@
...
+            $this->nextResult = $this->handleItem($this->getInnerIterator()->current());
+        }
+    }
+
+    private function handleItem(\SplFileInfo $file)
+    {
+        return $this->linter->lintFile($file->getRealPath());
+    }
+
+    private function getFileRelativePathname(\SplFileInfo $file)
```
gharlan: method is not used

src/Runner/FileFilterIterator.php
```diff
@@ -0,0 +1,111 @@
...
+
+use PhpCsFixer\FileCacheManager;
+use PhpCsFixer\FixerFileProcessedEvent;
+use Symfony\Component\EventDispatcher\Event;
+use Symfony\Component\EventDispatcher\EventDispatcher;
+
+/**
+ * @internal
+ */
+final class FileFilterIterator extends \FilterIterator
```
gharlan: author?

gharlan:
looks very well, thanks for the work!

src/Runner/FileLintingIterator.php
keradus: I think it's residue from another approach I have tried, nice catch

keradus:
Thank you @gharlan , I applied your comments ;)

src/Differ/SebastianBergmannDiffer.php
```diff
@@ -0,0 +1,53 @@
...
+        $this->differ = new Differ();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function diff($old, $new)
+    {
+        return implode(
+            PHP_EOL,
```
GrahamCampbell: Might be an idea to have this configurable via an optional constructor param, so a windows system can output unix style line endings if desired.

keradus: This PR is already big, what you are referring to is not changed - it's also current, master implementation. It's outside the scope of this PR.
My preferred option would be to not have it at all ;)

PR #1670, opened 2016-01-12
Backport Finder and Config classes from 2.x line

keradus:
As on 2.x line we want to remove `Symfony\CS\Config\Config` and use only `Symfony\CS\Config` it would be great to allow to switch to it already on 1.x line, same for Finder.

// requested by @GrahamCampbell 

ref #1646

### DIFF
--- a/Symfony/CS/Config.php
+++ b/Symfony/CS/Config.php
@@ -131,6 +131,22 @@ class Config implements ConfigInterface
         return $this;
     }
 
+    public function addCustomFixers($fixers)
+    {
+        if (false === is_array($fixers) && false === $fixers instanceof \Traversable) {
+            throw new \InvalidArgumentException(sprintf(
+                'Argument must be an array or a Traversable, got "%s".',
+                is_object($fixers) ? get_class($fixers) : gettype($fixers)
+            ));
+        }
+
+        foreach ($fixers as $fixer) {
+            $this->addCustomFixer($fixer);
+        }
+
+        return $this;
+    }
+
     public function getCustomFixers()
     {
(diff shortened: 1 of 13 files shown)


Symfony/CS/Config.php
```diff
@@ -0,0 +1,155 @@
+<?php
```
keradus: This file was renamed from `Symfony/CS/Config/Config.php`, then:
- change namespace

Symfony/CS/Finder.php
```diff
@@ -0,0 +1,70 @@
+<?php
```
keradus: This file was renamed from `Symfony/CS/Finder/DefaultFinder.php`, then:
- change namespace
- stop searching for `*.xml` and `*.yml` as 2.x Finder does not do that, DefaultFinder will still search for them to not break BC

keradus:
ping @GrahamCampbell  ;)

keradus:
Any thoughts?

SpacePossum:
like the idea :)
however.. on 1.x this will break for others with configs like https://github.com/GeckoPackages/GeckoTwig/blob/master/.php_cs.dist#L14 ?

keradus:
Why is that?
The class is still there: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1670/files#diff-c3d879ff174f5aec5f64e9fdb6cfe4a9R22 , only deprecated

SpacePossum:
indeed, nevermind me :)
lgtm :+1:

SpacePossum:
Thanks :)

PR #1673, opened 2016-01-14
Add fixer suite support

PedroTroller:
Good morning.

So, here is the use case.
Imagine I have a HUGE set of custom fixers, and I wan't to provide this set as a plugin to PHP-CS-FIXER, then the customer have to add all my fixers one by one by calling the `addCustomFixer`.

Now, I can provider one (or more) fixers suite

ex:
```php
class MyCustomSuite implements FixerSuiteInterface
{
    public function getFixers()
    {
        return array(
            new MyCustumFixer1(),
            new MyCustumFixer2(),
            new MyCustumFixer3(),
            new MyCustumFixer4(),
            new MyCustumFixer5(),
            new MyCustumFixer6(),
        );
    }
}
```

```php
<?php

    ->in('src/')
;

    ->setUsingCache(true)
    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
    ->fixers([
        '..',
        'custum-1',
        'custum-2',
        'custum-3',
        'custum-4',
        'custum-5',
        'custum-6',
    ])
    ->addCustomFixer(new Project\CS\MyCustomSuite())
    ->finder($finder)
;
```

### DIFF
--- a/Symfony/CS/Config/Config.php
+++ b/Symfony/CS/Config/Config.php
@@ -15,6 +15,7 @@ use Symfony\CS\ConfigInterface;
 use Symfony\CS\Finder\DefaultFinder;
 use Symfony\CS\FinderInterface;
 use Symfony\CS\FixerInterface;
+use Symfony\CS\FixerSuiteInterface;
 
 /**
  * @author Fabien Potencier <fabien@symfony.com>
@@ -137,6 +138,15 @@ class Config implements ConfigInterface
     public function getCustomFixers()
     {
         return $this->customFixers;
+    }
+
+    public function addCustomFixerSuite(FixerSuiteInterface $suite)
+    {
+        foreach ($suite->getFixers() as $fixer) {
+            $this->addCustomFixer($fixer);
+        }
+
+        return $this;
     }
 
     public function hideProgress($hideProgress)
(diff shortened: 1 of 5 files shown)


Symfony/CS/ConfigInterface.php
```diff
@@ -82,4 +82,9 @@ public function addCustomFixer(FixerInterface $fixer);
      * @return FixerInterface[]
      */
     public function getCustomFixers();
+
+    /**
+     * Adds all fixers contained inside the given FixerSuiteInterface.
+     */
+    public function addCustomFixerSuite(FixerSuiteInterface $suite);
```
keradus: hurray, BC break!

Symfony/CS/Config/Config.php
```diff
@@ -139,6 +140,15 @@ public function getCustomFixers()
         return $this->customFixers;
     }
 
+    public function addCustomFixerSuite(FixerSuiteInterface $suite)
```
keradus: I would more likely go in direction to add sth like `addCustomFixers` (like we have `addCustomFixer` (without `s` on the end)) that will accept `array|Traversable`

SpacePossum:
Cool idea :+1: (not sure we can/should update the config interface on the 1.x line though)

Symfony/CS/Config/Config.php
SpacePossum: :+1:

PedroTroller:
I close the PR and I open a new one based on master branch : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1674

PR #1674, opened 2016-01-14
Config - add addCustomFixers method

PedroTroller:
Good morning.

So, here is the use case.
Imagine I have a HUGE set of custom fixers, and I wan't to provide this set as a plugin to PHP-CS-FIXER, then the customer have to add all my fixers one by one by calling the `addCustomFixer`.

Now, I can provider one (or more) fixers suite

ex:
```php
class MyCustomSuite extends AbstractFixerSuite
{
    public function getFixers()
    {
        return array(
            new MyCustumFixer1(),
            new MyCustumFixer2(),
            new MyCustumFixer3(),
            new MyCustumFixer4(),
            new MyCustumFixer5(),
            new MyCustumFixer6(),
        );
    }
}
```

```php
<?php

    ->in('src/')
;

    ->setUsingCache(true)
    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
    ->fixers([
        '..',
        'custum-1',
        'custum-2',
        'custum-3',
        'custum-4',
        'custum-5',
        'custum-6',
    ])
    ->addCustomFixers(new Project\CS\MyCustomSuite())
    ->finder($finder)
;
```

### DIFF
--- a/src/Config/Config.php
+++ b/src/Config/Config.php
@@ -131,6 +131,19 @@ class Config implements ConfigInterface
         return $this;
     }
 
+    public function addCustomFixers($fixers)
+    {
+        if (false === is_array($fixers) && false === $fixers instanceof \Traversable) {
+            throw new \InvalidArgumentException('Argument must be an array or a Traversable');
+        }
+
+        foreach ($fixers as $fixer) {
+            $this->addCustomFixer($fixer);
+        }
+
+        return $this;
+    }
+
     public function getCustomFixers()
     {
         return $this->customFixers;
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -79,6 +79,13 @@ interface ConfigInterface
     public function addCustomFixer(FixerInterface $fixer);
 
     /**
+     * Adds a suite of custom fixers.
+     *
+     * @param FixerInterface[]|\Traversable $fixers
+     */
+    public function addCustomFixers($fixers);
+
+    /**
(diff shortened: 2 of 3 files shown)


src/AbstractFixerSuite.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
keradus: I do not see any value of this class, I could use `Config::addCustomFixers` without it.

When I want to provide a nice class for my set of custom fixers I will still not need it. I will just create a class based on Traversable (eg IteratorAggregate). Also, if at all, then I would go in Interface direction. But then even less value of it.
Maybe the class-level docs would be a nice place to show value of this class.

PedroTroller: @keradus Class removed and test and documentation updated.

src/Config/Config.php
```diff
@@ -131,6 +131,15 @@ public function addCustomFixer(FixerInterface $fixer)
         return $this;
     }
 
+    public function addCustomFixers($fixers)
```
keradus: as it is not typehinted it would be great to validate it

PedroTroller: You mean as an array/traversable ?

keradus: yes

PedroTroller: Done

keradus:
Thank you @PedroTroller.

PR #1682, opened 2016-01-19
ElseifFixer - handle comments

SpacePossum:
Make `ElseifFixer` work even if there are comments used between `if` and `else`.

Some optimizing; 
* no need to check all tokens (the fixer looks for a token sequences of quite some tokens)
* don't iterate all tokens twice when on HHVM, but only re-modify tokens when needed (i.e. this fixer modified tokens).

### DIFF
--- a/Symfony/CS/Fixer/PSR2/ElseifFixer.php
+++ b/Symfony/CS/Fixer/PSR2/ElseifFixer.php
@@ -30,25 +30,31 @@ class ElseifFixer extends AbstractFixer
     public function fix(\SplFileInfo $file, $content)
     {
         $tokens = Tokens::fromCode($content);
+        foreach ($tokens as $index => $token) {
+            if (!$tokens[$index]->isGivenKind(T_ELSE)) {
+                continue;
+            }
 
-        foreach ($tokens->findGivenKind(T_ELSE) as $index => $token) {
-            $nextIndex = $tokens->getNextNonWhitespace($index);
-            $nextToken = $tokens[$nextIndex];
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
 
             // if next meaning token is not T_IF - continue searching, this is not the case for fixing
-            if (!$nextToken->isGivenKind(T_IF)) {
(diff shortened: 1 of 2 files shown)


SpacePossum:
TODO: fix HHVM

keradus:
@SpacePossum what is the state of this PR ?

SpacePossum:
waiting for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1705 and this one should be finished after that in no time :)

SpacePossum:
ping @keradus (and other reviewers), I think this one is ready

SpacePossum:
Appveyor is broken, its not the code

Symfony/CS/Fixer/PSR2/ElseifFixer.php
```diff
@@ -30,25 +30,31 @@ class ElseifFixer extends AbstractFixer
     public function fix(\SplFileInfo $file, $content)
     {
         $tokens = Tokens::fromCode($content);
+        for ($index = 7, $count = count($tokens); $index < $count - 1; ++$index) {
```
keradus: I really don't like that index juggling, I never know `why the hell it's 7 this time???`.
For real code-files it won't change the performance, but it's decreases readability comparing to simple foreach loop.

SpacePossum: It is always the same reason isn't it? Do not check tokens that are no candidate anyway.
I see no readability issues here at all.

Besides the code was a simple `foreach`:
```php
 foreach ($tokens->findGivenKind(T_ELSE) as $index => $token) {
```

however at least two main contributors told me in another PR to avoid `findGivenKind` where possible....

keradus: ` foreach ($tokens as $index => $token) {`
is more readable than magic `7`, at least for me...

SpacePossum: agree to disagree, updated, lets see how the tests do, I guess PHP 7 and/or HHVM will fail

keradus:
Thank you @SpacePossum

Issue #1672, opened 2016-01-13
unneeded "else" or "elseif" remover


GrahamCampbell:
### Fixer request.

Before:

```php
<?php

function stuffs($foo, $bar, $baz)
{
    if ($foo) {
        return $bar;
    } elseif ($baz) {
        return $foo;
    } else {
        return $baz;
    }
}
```

After:

```php
<?php

function stuffs($foo, $bar, $baz)
{
    if ($foo) {
        return $bar;
    }

    if ($baz) {
        return $foo;
    }

    return $baz;
}
```

SpacePossum:
I think I've a fixer for the useless `else` case in a heavily hacked PHP-CS-Fixer. There are some TODO's on that one... here are some thoughts ;)
- it must ran before the useless return fixer
- the useless return fixer should be able to deal with multiple useless returns after each order (current missing feature)
- it must ran before any kind of empty line / trailing whitespace fixer
- it must ran before any indenting fixer
- it must detect the useless else correctly for not being risky, which includes detecting the use of `GOTO` within the overall structure the `else` is defined in

SpacePossum:
another unneeded case
```php
if ($a) {
    throw new \Exception('');
} else {

}
```

gharlan:
same for `continue` and `break`

gharlan:
> it must detect the useless else correctly for not being risky, which includes detecting the use of GOTO within the overall structure the else is defined in

could you give an example with `goto` where code behavior would change?

SpacePossum:
I cannot, I'm not sure why I put that as a TODO in my code .. ... .

gharlan:
first after reading I thought "good point", but after more thought I could not find an example ;)

SpacePossum:
the "elseif" is still valid, although hard to make ;)

PR #1693, opened 2016-01-21
[WIP] Added NoUselessElseFixer

SpacePossum:
Add UselessElseFixer and `clear and merge whitespace tokens around cleared` methods. 

for ref; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1591#discussion_r50195317
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1693/files#diff-3dc73b7d0a0116aa5e987e3a2aad7a81R1414

part of https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1672

### DIFF
--- a/Symfony/CS/Fixer/Contrib/UselessElseFixer.php
+++ b/Symfony/CS/Fixer/Contrib/UselessElseFixer.php
@@ -1,0 +1,158 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractAlignFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+class UselessElseFixer extends AbstractAlignFixer
+{
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        // start from 4; minimum number of tokens that form a candidate for fixing; <?php if ( [] ) [5]
+        $elseTokens = $tokens->findGivenKind(T_ELSE, 4);
+        foreach ($elseTokens as $index => $elseToken) {
+            // `else if` vs. `else` check
+            if ($tokens[$tokens->getNextMeaningfulToken($index)]->isGivenKind(T_IF)) {
+                continue;
+            }
+
+            // clean up `else` if possible
+            $this->fixElse($tokens, $index);
+
+            // clean up `else` if it is an empty statement
+            if (!$tokens[$index]->isEmpty()) {
+                $this->fixEmptyElse($tokens, $index);
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'There should not be useless else cases.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before DuplicateSemicolonFixer, WhitespacyLinesFixer, ExtraEmptyLinesFixer and BracesFixer
+        return 15;
+    }
+
+    /**
+     * @param Tokens $tokens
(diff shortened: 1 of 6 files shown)


SpacePossum:
TODO: integration tests

current hits on Symfony; https://github.com/symfony/symfony/pull/17493

SpacePossum:
Only the integration tests and moving 2 tests to a required PHP < 5.4 are left todo.
Happy to take comments :)

Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -1403,4 +1403,49 @@ public function isMonolithicPhp()
...
     }
+
+    /**
+     * Clear tokens at given indexes.
+     *
+     * Merges whitespace tokens around the cleared tokens if needed
+     *
+     * @param int[] $indexes
+     */
+    public function clearTokens(array $indexes)
```
keradus: this vs `clearRange`

keradus: why the array? it's not clean to clear arbitrary indexes

SpacePossum: >clearRange
from - till

>clearTokens
individual indexes

or do you mean changing `clearRange` to merge possible whitespace tokens as well?

javiereguiluz:
This fixer looks interesting, but I think it needs to improve its output a bit to make it more useful.

If my original code is like this:

```php
<?php
    if ($a) {
        %s
    } else {
        echo 1;
    }
```

I expect the fixed code to be like this:

```php
<?php
    if ($a) {
        %s
    }

    echo 1;
```

But this fixer would output this:

```php
<?php
    if ($a) {
        %s
    }
        echo 1;
```

Could the fixer outdent the code that was originally inside the useless `else`? Thanks!

Symfony/CS/Tokenizer/Tokens.php
keradus: > > this vs `clearRange`
> 
> from - till

For end user it may be misleading which method to use.
Both are "clear some tokens", and then if you go inside you see that one will also merge some tokens, while the other not.

I would suggest to remove array (so pass _one_ index), and rename the method

keradus:
> // should be run before DuplicateSemicolonFixer, WhitespacyLinesFixer, ExtraEmptyLinesFixer

> Could the fixer outdent the code that was originally inside the useless else? Thanks!

So, the fixer should run before `BracesFixer` as well ;)

@javiereguiluz we have very single-responsibility fixers. First one will remove extra braces, second will reindent the code

javiereguiluz:
@keradus great! Thanks for the explanation.

Symfony/CS/Fixer/Contrib/UselessElseFixer.php
```diff
@@ -0,0 +1,174 @@
...
+                array(T_EXIT),
+                array(T_RETURN),
+                array(T_THROW),
+            )
+        );
+
+        if (null !== $candidateIndex && !$tokens[$candidateIndex]->equalsAny(array(';', '}'))) {
+            $this->clearElse($tokens, $index);
+
+            return;
```
SpacePossum: useless return

Symfony/CS/Fixer/Contrib/UselessElseFixer.php
```diff
@@ -0,0 +1,174 @@
...
+        return 'There should not be useless else cases.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // TODO
+        // should be run before DuplicateSemicolonFixer, WhitespacyLinesFixer, ExtraEmptyLinesFixer
```
SpacePossum: + BracesFixer

keradus:
@SpacePossum what is the state of this PR ?

SpacePossum:
waiting for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1705, after that I'll do some more testing :)

SpacePossum:
This needs integration tests.
I would like to ask all reviewers to have a look at the logic and current utest.

keradus:
@SpacePossum ?

SpacePossum:
Oh no I was cleaning up old branches and delete this one... I'll try to restore :|

SpacePossum:
ah man this is all wrong! I lost my fixes :(

PR #1696, opened 2016-01-22
[WIP]ThrowFixer - Add

lucasmichot:
Add a throw fixer.

---

Requires https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1766

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -712,6 +712,10 @@ Choose from the list of available fixers:
                         $strict param. Warning! This
                         could change code behavior.
 
+* **throw** [contrib]
+                        An empty line feed should
+                        precede a throw statement.
+
 
 The ``--config`` option customizes the files to analyse, based
 on some well-known directory structures:
--- a/Symfony/CS/Fixer/Contrib/ThrowFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ThrowFixer.php
@@ -1,0 +1,33 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractBlankLineAboveFixer;
+
+/**
+ * @author Lucas Michot <lucas@semalead.com>
+ */
+class ThrowFixer extends AbstractBlankLineAboveFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected static $token = T_THROW;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'An empty line feed should precede a throw statement.';
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Contrib/ThrowFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/ThrowFixerTest.php
@@ -1,0 +1,138 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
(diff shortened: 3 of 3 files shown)


GrahamCampbell:
Would be good to use our phpdoc classes.

GrahamCampbell:
Oh wait, misread what this was entirely, lol...

GrahamCampbell:
Is this code similar to the return fixer we have?

GrahamCampbell:
Might be an idea to include the authors in the class annotations, or better, make an abstract class?

lucasmichot:
@GrahamCampbell see https://github.com/lucasmichot/PHP-CS-Fixer/commit/ed503693aadab73ca9c74452fdccf92e350dd31c

SpacePossum:
good idea :+1: 
I would go for an abstract blankline above fixer which can be configured (on the master branch) like the reverse one `extra_empty_lines`. Changes are people want the blankline not just for `return` and `throw` but for all exit/return tokens; return, exit/die, throw, continue, break, goto.

lucasmichot:
Guys see https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/930b40ff0cfe3c8e54c55d8f231a84dcd1364972

@SpacePossum not completely sure what you mean.
Do you mean making one PR against master (for the abstract class only, then PR against 1.12 for making files using this abstract class ?

SpacePossum:
I would make one new fixer (so no abstract) on the 1.x branch that will a blankline above any of the tokens `return`, `exit/die`, `throw`, `continue`, `break`, `goto` following the some rules/conditions as the current `return`-fixer.
When that is done, we can remove the current `return`-fixer from master and add configuration options to your new fixer. People can than configure the new fixer to act the as the current `return`-fixer, or configure it to do even more.
I would advice that get more input and ideas on my idea from others (like ping @keradus) before putting work into this ;)

keradus:
:+1: for @SpacePossum proposition, I want to make the same ;)
Should go to 1.x line with possibility to configure as well (static method, like in `header` fixer).

keradus:
@lucasmichot you have created abstract fixer and then dummy `return` and `throw` fixers.
Please, instead of that follow @SpacePossum direction - leave the `return` fixer intact, create second fixer like `empty_line_above_ending_statement`, which can add that empty line before all of `return, exit/die, throw, continue, break, goto`, but may be configured (see PHPUnit fixers for example of how to do it).

(Then, on 2.x line, we will remove `return` fixer).

What do you think about it?

--------------------------------------------------------------------------

Also, be aware that when you just push the changes to PR nobody is notified about it. It's great to write a comment like "requested changes are done" or sth ;)

keradus:
@lucasmichot  what is the state of this PR ?

GrahamCampbell:
> Requires #1766

That requirement PR should probably be closed and moved into here.

lucasmichot:
OK @keradus, following @SpacePossum feedback.
Please check https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1766

keradus:
> OK @keradus, following @SpacePossum feedback : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1696#issuecomment-174189509.

How is that? Both @SpacePossum and me wrote to no **not** create abstract class. Then, again, both of us wrote to add new things on 1.12 line and remove on master branch, you did the other way around.

lucasmichot:
> good idea :+1: 
I would go for an abstract blankline above fixer which can be configured (on the master branch) like the reverse one extra_empty_lines.

keradus:
it may be worth to read newer comments as well... 
The very next comment of @SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1696#issuecomment-174189509
And then my comments.

Why you ignore them ?

lucasmichot:
> Why you ignore them ?

@keradus , I come back from vacations and pick the first comments in my mailbox from this PR. Sorry for the misunderstading

SpacePossum:
>I would go for an abstract blankline 

>I would make one new fixer (so no abstract)

I admit, this was me not being clear in using "abstract", and I'm sorry for that.
When talking 'abstract' as class property, I think it is not needed. I meant 'abstract' as one fixer with multiple configuration options (so not having dedicated fixers for each case).
Let me know if can express it better or anything.

lucasmichot:
NVM, my fault.
Sorry guys

PR #1700, opened 2016-01-24
Fixer - resolve import conflict

### LINKED ISSUE #1698: Fixer class broken
Line 18 `use Symfony\Component\Finder\Finder;` won't work because of a conflict. I get:

```
PHP Fatal error:  Cannot use Symfony\Component\Finder\Finder as Finder because the name is already in use in /home/ubuntu/PHP-CS-Fixer/src/Fixer.php on line 18
```

No idea why travis hasn't picked this up!

keradus:
Closes #1698

cc @GrahamCampbell

### DIFF
--- a/Symfony/CS/Fixer.php
+++ b/Symfony/CS/Fixer.php
@@ -13,7 +13,7 @@ namespace Symfony\CS;
 
 use SebastianBergmann\Diff\Differ;
 use Symfony\Component\EventDispatcher\EventDispatcher;
-use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\Finder as SymfonyFinder;
 use Symfony\Component\Finder\SplFileInfo as FinderSplFileInfo;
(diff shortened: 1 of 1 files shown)


Symfony/CS/Fixer.php
```diff
@@ -13,7 +13,7 @@
 
 use SebastianBergmann\Diff\Differ;
 use Symfony\Component\EventDispatcher\EventDispatcher;
-use Symfony\Component\Finder\Finder;
+use Symfony\Component\Finder\Finder as SymfonyFinder;
 use Symfony\Component\Finder\SplFileInfo as FinderSplFileInfo;
```
GrahamCampbell: Might make sense to alias this as SymfonySplFileInfo too @keradus?

keradus: both ways are good for me, if you want send a PR and I will merge it ;)

Issue #1521, opened 2015-11-19
Ditch `@type` Stuff


GrahamCampbell:
After the other phpdoc stuff I'm working on has gone through, I need to send a PR to the master branch to remove the `type` annotation from the list of valid types in the `Tag` class, and to remove the `var_to_type` contrib fixer.

keradus:
looks like already done.

GrahamCampbell:
Not done.

keradus:
No? I didn't found it on master branch.
Feel free to prepare the change and send a PR.

GrahamCampbell:
Both fixers remain:

1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Phpdoc/PhpdocTypeToVarFixer.php
2. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Phpdoc/PhpdocVarToTypeFixer.php

---

The proposal I had was to remove them both once we got to a point where nobody cared about type or had it used anywhere/

keradus:
The proposal you had was to remove one of them. If you want - feel free to send a PR.
But if your proposal is to wait for PSR-5 to be approved (like in 3 months? 6 months? year? never?), then it's pointless to keep the issue open. It could be recreated after year, if after that time we will be willing to finally make that change.

GrahamCampbell:
I don't think it'll ever go through TBH. :/

keradus:
so no reason to wait with this issue for PSR-5, right ?

PR #1703, opened 2016-01-24
PhpdocVarToTypeFixer - remove fixer

keradus:


GrahamCampbell:
I think it's probably best to leave this here until psr5 gets finalized.

keradus:
it won't be finalized for ages

gharlan:
:+1: for removing

GrahamCampbell:
What's the point in having either fixer if we don't have one for each direction?

keradus:
It's violates PSR-5 proposal. If PSR change the annotation (I really doubt it) then we could change the fixer and still keep one fixer that follows PSR.

GrahamCampbell:
> It's violates PSR-5 proposal.

So does the other fixer though. We say we only recognize valid PSR-5 annotations, so there is no such annotation type to even be fixed to var by that definition of strictly following PSR-5 only.

And, actually I disagree that we need to be so massively strict on being so connected to PSR-5 for the same reasons we have fixers that don't match up with symfony and/or PSR-2.

gharlan:
refs #1521

keradus:
good one @gharlan ;)

GrahamCampbell:
> good one @gharlan ;)

Yes, I do want it removing eventually, but I think 2.x is just too early. ;)

GrahamCampbell:
I think we should remove both fixers, and the definition of "type" from our phpdoc classes in 3.x, and keep them both in 2.x.

keradus:
3.x won't be soon, as well as official PSR5. No need to wait for both.

GrahamCampbell:
> 3.x won't be soon, as well as official PSR5. No need to wait for both.

I don't see the harm in leaving this fixer in the core for a year or two. It gives people the choice for what they want in their code base.

keradus:
replaced by #1779

PR #1708, opened 2016-01-26
Update composer.json requirements

keradus:


GrahamCampbell:
:-1: for doing this in a minor release.

GrahamCampbell:
Composer will just resolve php-cs-fixer 1.11 rather than 1.12 if someone on a later php version asked for 1.x now...

GrahamCampbell:
As far as I'm concerned, restricting version constraints requires a major version bump, but relaxing them does not.

keradus:
Where is the restrictions ?

GrahamCampbell:
> Where is the restrictions ?

GrahamCampbell:
> Where is the restrictions ?

You don't allow php 8+ but before you did.

GrahamCampbell:
Or, indeed php 6, lol.

keradus:
> You don't allow php 8+ but before you did.

Yeah, we don't support it :D 6.x as well.

GrahamCampbell:
Yes, but, 1.11 DOES support those versions, even if we didn't want it to, according to composer, if you see what I mean.

Forcing those versions off 1.12 results in undesired behaviour that 1.11.x is resolved instead.

keradus:
No, it does not. Yes, composer would install it. But we, as people behind the tool, does not promise it will work, as we indeed does not support it, not test it etc.

C'mon, if sb uses 6.x then yes, I want to disallow the tool to run under it.
And reasoning of yours just give my some thought that maybe even merge it into 1.11 line

GrahamCampbell:
> And reasoning of yours just give my some thought that maybe even merge it into 1.11 line

Then 1.11.0 still has the issue. :trollface:

keradus:
And then I think that only you will be affected by it.
Then c'mon, ask why composer install 1.11.0 on your PHP 6 and then some contributor (like... you? :D) will answer that we do not support PHP 6.

SpacePossum:
LGTM :+1: 
Ditching PHP 6 makes sense to me, even if this was incorrectly supported in the composer configuration before. This is not a BC since it was never supported, so it is a fix IMHO :)

PR #1710, opened 2016-01-27
LowercaseConstantFixer - Fix comment cases.

SpacePossum:
Little bug fix (so targetting 1.11).
The following cases are getting 'fixed' by the `LowercaseConstantFixer` while those should not be touched;
```php
<?php 
$x = new /**/False?>
<?php
Null/**/::test();
True//
  ::test();
```

The diff. looks more heavy because I've added a bunch of test cases using templating. 
(you gonna love PHP for allowing comments all over the place ;) )


keradus:
Thank you @SpacePossum.

Issue #906, opened 2014-12-19
Feature request: remove blank lines after control structure opening brace


keradus:
Feature request: remove blank lines after control structure opening brace.
level: contrib
Refs: #895

keradus:
Closing due to merge of #1712

PR #1712, opened 2016-01-28
NoExtraConsecutiveBlankLinesFixer - Remove blankline after curly brace open

SpacePossum:
Adds/updates:
- remove blankline after curly brace open if configured 
- remove blanklines between trait uses if configured
- updating note modified
- SF rule set updated
- fixer project configuration rules updated

Fixes
- remove current broken blankline between use imports fixing

Removes
- dedicated `remove_lines_between_uses` fixer (same can be done with this fixer using config)

Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/906

- [x] Do not fix when token is in 1 line lambda
- [ ] `Transformers::create` call is ugly (at that location)


tests/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -19,6 +19,64 @@
...
+                'use',
+                'curly_brace_open',
+            )
+        );
+
+        $this->doTest(
+            '<?php
+$a = function($e) use ($b) { while(3<1)break; while(3<1)continue; if (true) throw $e; return 1; };
+// test
+');
```
SpacePossum: @keradus what would be the rule here? It is pretty hard to skip closures blindly

keradus: maybe skip closure if it is one-liner ?

SpacePossum: hmm check..  just typing as a reminder to myself that case like this;

```php
function a() {
return function ($a) { if ($a) { return 1;}};

}
```

should be fixed based on the first `return` token and not the second

tests/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -19,6 +19,75 @@
...
+                'curly_brace_open',
+            )
+        );
+
+        $this->doTest(
+            '<?php
+$a = function($e) use ($b) { while(3<1)break; while(3<1)continue; if (true) throw $e; return 1; };
+
+
+// test
```
SpacePossum: updated, ping @keradus :)

tests/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -19,6 +19,75 @@
...
+
+                echo $a();
+                die;'
+        );
+    }
+
+    /**
+     * @dataProvider provideFix1Cases
+     */
+    public function testFix1($expected, $input = null)
```
keradus: 1 ?

UPGRADE.md
```diff
@@ -99,7 +99,7 @@ php_closing_tag                                | no_closing_tag
 phpdoc_params                                  | phpdoc_align
 phpdoc_short_description                       | phpdoc_summary
 remove_leading_slash_use                       | no_leading_import_slash
-remove_lines_between_uses                      | no_blank_lines_between_uses
+remove_lines_between_uses                      | no_extra_consecutive_blank_lines
```
keradus: it's not only renaming, please add 3rd column

src/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -52,26 +64,29 @@ public function configure(array $configuration = null)
             return;
         }
 
-        $this->tokenCallbackMap = array();
+        $this->tokenKindCallbackMap = array();
```
keradus: `tokenEquaqlsMap` should be cleared as well.

UPGRADE.md
keradus: so actually we combined 2 fixers into one, right ?

tests/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixerTest.php
keradus: cool !

src/Fixer/Symfony/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -43,6 +54,7 @@
      * - 'return' remove blank lines after a line with a 'return' statement
      * - 'throw' remove blank lines after a line with a 'throw' statement
      * - 'use' remove blank lines between 'use' import statements
+     * - 'curly_brace_open' remove blank lines after a curly opening brace ('{')
```
keradus: This option is so powerful !
Actually, you could now remove `NoBlankLinesAfterClassOpeningFixer` and move all it's tests here, tested with `curly_brace_open`. It works!

tests/FixerFactoryTest.php
```diff
@@ -216,13 +216,18 @@ public function getFixersPriorityCases()
             array($fixers['binary_operator_spaces'], $fixers['align_double_arrow']), // tested also in: align_double_arrow,binary_operator_spaces.test
             array($fixers['binary_operator_spaces'], $fixers['align_equals']), // tested also in: align_double_arrow,align_equals.test
             array($fixers['class_definition'], $fixers['no_trailing_whitespace']), // tested also in: class_definition,no_trailing_whitespace.test
+            array($fixers['no_unused_imports'], $fixers['no_extra_consecutive_blank_lines']), // tested also in: no_unused_imports,no_extra_consecutive_blank_lines.test
+            array($fixers['single_import_per_statement'], $fixers['no_unused_imports']), // tested also in: single_import_per_statement,no_unused_imports.test
+            array($fixers['no_leading_import_slash'], $fixers['ordered_imports']), // tested also in: no_leading_import_slash,ordered_imports.test
+            array($fixers['no_extra_consecutive_blank_lines'], $fixers['ordered_imports']),
+            array($fixers['no_unused_imports'], $fixers['no_leading_import_slash']),
+            array($fixers['single_import_per_statement'], $fixers['no_leading_import_slash']),
```
keradus: I don'y quite get that change. PR touched 2 fixers, but here are more fixers.
What am I missing?

keradus:
not sure is this change good:

```diff
         switch ($count = count($this->baseUrls)) {
             case 0:
                 return '';
-
             case 1:
                 return $this->baseUrls[0];
-
             default:
                 return $this->baseUrls[fmod(hexdec(substr(md5($path), 0, 10)), $count)];
         }
```

tests/FixerFactoryTest.php
SpacePossum: let me check, I started this branch before the big renaming of fixer so this could be a merge gone wrong.

GrahamCampbell:
> not sure is this change good:

I think that change is good.

SpacePossum:
>not sure is this change good:

I think it is ^-^
Anyway, it is not a result of the PR, this is current behavior even with the `return fixer` on 1.x I think.
So please feel free to report as an issue, but I would like to keep it out of this PR if that is ok.

keradus:
ping @SpacePossum

SpacePossum:
ping @keradus , ready for review :)
I think we should remove the `after break` configuration setting from SF (looking at the current SF codebase). Another thing to look at is the hack I've put in the configuration method of the fixer, I need the `CT_`'s being defined there; however currently this is not the case since the transformers are loaded later on in the flow.
Please re-read the description of the PR, I've updated it to reflect the current changes in the PR, thanks!

SpacePossum:
ping @keradus

keradus:
Anyone willing to make a review ?

keradus:
Thank you @SpacePossum.

PR #1732, opened 2016-01-31
move fixer classes from level subdirs to thematic subdirs

gharlan:
Do we want to remove the level subdirs (contrib/psr1/psr2/symfony)?


keradus:
:-1:
I don't want to have 100 fixers on one level.

Instead of it I would prefer to have dirs per type of changes, like dir for PhpDoc fixers, dir for phpunit fixers and so on.

gharlan:
> Instead of it I would prefer to have dirs per type of changes, like dir for PhpDoc fixers, dir for phpunit fixers and so on.

:+1:

GrahamCampbell:
:-1: for this as is

gharlan:
current state: https://github.com/gharlan/PHP-CS-Fixer/tree/remove-level-subdirs/src/Fixer
is this the right direction?

keradus:
I assume that `*_` are not-moved-yet fixers.
I like that direction.

keradus:
Also, at the end, I would move `FixerInterface` and `Abstract*Fixer` into `Fixer` dir as well.

gharlan:
> I assume that *_ are not-moved-yet fixers.

yes

gharlan:
all classes are moved now: https://github.com/gharlan/PHP-CS-Fixer/tree/remove-level-subdirs/src/Fixer
please discuss.. :)

GrahamCampbell:
Can we not get away with `Return` rather than `ReturnNotation` since php surely only reserves the class `Return`, not the namespace?

GrahamCampbell:
Same for the others, like `Array`.

GrahamCampbell:
Could we move the abstract phpdoc fixers into the phpdoc namespace?

GrahamCampbell:
I'm :+1: on the way this PR is moving btw.

gharlan:
> Can we not get away with Return rather than ReturnNotation since php surely only reserves the class Return, not the namespace?

https://3v4l.org/1RWsD

GrahamCampbell:
Sigh. I guess I didn't think about how tokenization would work.

keradus:
Maybe some new namespace like `BlankLine`, with:
- NoBlankLinesAfterPhpdocFixer (ReturnNotation)
- BlankLineBeforeReturnFixer.php (ReturnNotation)
?

Also, 3 of 4 Fixers from `Semicolon` are more about whitespaces, not semicolon.

keradus:
:+1:

keradus:
Thank you @gharlan.

SpacePossum:
:+1:  very nice :)

PR #1733, opened 2016-01-31
ConfigurationResolver - look for .php_cs file in cwd as well

keradus:


keradus:
Before:

```bash
vendor\bin\php-cs-fixer fix src/file.php
## no config file used, damn it!
```

```bash
vendor\bin\php php-cs-fixer fix src/file.php --config=.php_cs.dist
Loaded config from ".php_cs.dist" ## cool!
```

After:
```bash
vendor\bin\php php-cs-fixer fix src/file.php
Loaded config from ".php_cs.dist" ## cool!
```

SpacePossum:
nice :D  :+1:

Issue #1727, opened 2016-01-31
Question: phpdoc and trailing spaces


keradus:
Could trimming trailing spaces inside PHPDoc could change meaning of code?
In other words, if we had such fixer would it be risky ?

cc @GrahamCampbell

GrahamCampbell:
Trimming trailing "space" or "tab" characters from lines should not change the behaviour of phpdoc and should be safe.

keradus:
For now we have (following 1.x names):

> **trailing_spaces [PSR-2]**
> Remove trailing whitespace at the end of non-blank lines.

> **whitespacy_lines [symfony]**
> Remove trailing whitespace at the end of blank lines.

I would add:
`phpdoc_no_trailing_spaces` for PHPDocs purpose

and maybe, at some time:
`force_no_trailing_spaces` that will remove trailing whitespace from all places, but that may be risky.
```php
<?php

$content = 'foo    
bar';
$content === "foo    \nbar"; // true
```

GrahamCampbell:
> force_no_trailing_spaces that will remove trailing whitespace from all places, but that may be risky.

I like that fixer. Useful even if it's used as a guide to find those places in a code base where we can tweak it so as not to need the trailing whitespace if it were to change the behaviour.

gharlan:
> force_no_trailing_spaces that will remove trailing whitespace from all places, but that may be risky.

yes, would be risky. But I would like that fixer, too. :+1:

PR #1735, opened 2016-01-31
Added NoTrailingWhitespaceInCommentFixer

keradus:
ref #1727

### DIFF
--- a/Symfony/CS/Fixer/Symfony/NoTrailingWhitespaceInCommentFixer.php
+++ b/Symfony/CS/Fixer/Symfony/NoTrailingWhitespaceInCommentFixer.php
@@ -1,0 +1,47 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractAnnotationRemovalFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class NoTrailingWhitespaceInCommentFixer extends AbstractAnnotationRemovalFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        foreach ($tokens as $token) {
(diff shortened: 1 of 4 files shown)


Symfony/CS/Fixer/Symfony/NoTrailingWhitespaceInCommentFixer.php
```diff
@@ -0,0 +1,47 @@
...
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'There MUST be no trailing spaces inside comments.';
```
GrahamCampbell: Comments AND phpdoc.

keradus: Considering it strictly, phpdoc are special cases of comments.

GrahamCampbell: I don't consider that to be true actually. When someone says "comment", I don't include phpdoc in that definition in my mind.

keradus: `T_DOC_COMMENT`... a comment...

in the and I think it doesn't really matter...

I assume the rest of the PR is ok

GrahamCampbell: it's not a "comment", it's a "doc comment"

GrahamCampbell: Put it another way, the set of comments and doc comments are disjoint.

keradus: Not for me.
Like rectangle and squares.

But we don't need to agree about that I think,\

GrahamCampbell: A the set of doc comment can't possibly be a subset of the set of comments because surely that would mean that they'd both have to be tokenized as comments?

Anyway, I think we both understand each other's points, lol.

keradus: strings have few T_* as well, and we consider all of them as strings ;)

GrahamCampbell: frrr

GrahamCampbell: meh, consider it for the sake of clarity then

PR #1736, opened 2016-01-31
Integration tests - don't check priority for tests using short_tag fixer

### LINKED ISSUE #1615: Failing integration test when short_open_tag=on
When running `phpunit` locally, I get this failure:

```
1) Symfony\CS\Tests\IntegrationTest::testIntegration with data set #1 ('misc/php_closing_tag,short_tag.test', 'Integration of fixers: php_cl...t_tag.', array(Symfony\CS\Fixer\PSR1\ShortTagFixer Object (), Symfony\CS\Fixer\PSR2\PhpClosingTagFixer Object ()), array(false, '7.0.0'), '<?\n$a = 1;\n\n?>', '<?php\n$a = 1;')
Set priorities must be significant. If fixers used in reverse order return same output then the integration test is not sufficient or the priority relation between used fixers should not be set.
Failed asserting that two strings are not identical.
```

Reason: `short_open_tag` is enabled, so the priority of `ShortTagFixer` is irrelevant.

keradus:
Closes #1615

### DIFF
--- a/Symfony/CS/Tests/Fixtures/Integration/misc/php_closing_tag,short_tag.test
+++ b/Symfony/CS/Tests/Fixtures/Integration/misc/php_closing_tag,short_tag.test
@@ -3,6 +3,8 @@ Integration of fixers: php_closing_tag,short_tag.
 --CONFIG--
 level=none
 fixers=php_closing_tag,short_tag
+--SETTINGS--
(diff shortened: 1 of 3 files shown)


keradus:
ping @gharlan

gharlan:
Thanks!

PR #1737, opened 2016-01-31
RuleSet/FixerFactory - sort arrays content

keradus:


GrahamCampbell:
> sort arrays content

Sounds like an idea of a fixer. :trollface:

GrahamCampbell:
Would it be OCD to add a test to check the test arrays are sorted?

keradus:
yes

GrahamCampbell:
Well, if OCD needs must, it would be moved to a property. :D

PR #1741, opened 2016-02-02
FinerInterface - interface must extends Traversable

keradus:
Anything that extends `FinderInterface` cannot be used if it won't implement `Traversable` as well.


GrahamCampbell:
Seems legit. :+1:

GrahamCampbell:
Hmm, actually, this is a breaking change technically.

keradus:
No, a bugfix. It would be a BC breaker if one could use that without implementing `Traversable`

keradus:
Very well, closing in favour of  #1756

Issue #1719, opened 2016-01-29
Fixing a file excluded in the .php_cs config file should not occurs


tifabien:
I'm faced to a problem with [GrumPHP](https://github.com/phpro/grumphp) when I run php-cs-fixer task.

I created a .php_cs config file. In this config I excluded some autogenerated files. The problem is that when I run codestyle checks it fails on these autogenerated files. 
Behind the scene the command executed is the following : 
`vendor/binphp-cs-fixer --dry-run --config-file=.php_cs --verbose fix app/autoload.php` 

This `app/autoload.php` file is present in my config file
```<?php

return Symfony\CS\Config\Config::create()
    ->setUsingCache(true)
    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
    ->fixers(array('ordered_use'))
    ->finder(
        Symfony\CS\Finder\DefaultFinder::create()
            ->in(__DIR__)
            ->notPath('app/autoload.php')
    )
;
```
Is there a reason why .php_cs config file is not took into account when we invoke php-cs-fixer command with a specific file in param?

GrahamCampbell:
Not path cannot include a filename, it strictly means a path.

GrahamCampbell:
Maybe you want `notName('autoload.php');`?

tifabien:
Same result if I change notPath to notName.

It seems not path can accept filenames. it is the case in [Symfony/symfony](https://github.com/symfony/symfony/blob/master/.php_cs) repo

keradus:
This is other problem.

When you say that you want to fix `app/autoload.php` file by CLI argument, then yout finder from .php_cs file is overridden with path provided in console.

Do we want to somehow combine it? That could be a bit messy...

tifabien:
Indeed this could be a bit tricky to understand...
But this also is a bit heavy to check the whole project in the case of a pre-commit hook check.

How could we check files individually by applying .php_cs rules ?

keradus:
Yeah, I faced that too. I think that to achieve it we need to change the way of dealing with finder.

tifabien:
I agree :+1: 
Maybe we could introduce an new CLI option `--force` in order to overwrite finder if a file is provided.

keradus:
:-1: for 3 places to set same thing...
It would be just a hack...

Again, we already overriding the finder, that is the point...

tifabien:
Understood.
I'll try to find a hack for my use case !

[...]

PR #1632, opened 2015-12-24
Php Inspections (EA Extended): SCA for 1.12

kalessil:
Php Inspections (EA Extended): fixed following inspections complains

- methods names case mismatches
- strtr -> str_replace usage semantics
- un-needed parenthesises
- cascading str_replace calls
- not optimal if conditions
- simplifying ternary-expressions
- one-time use variables

PR #1742, opened 2016-02-02
path argument is used to create an intersection with existing finder

keradus:
ref #1719 

@tifabien , please, could you heavily play with this change?
(especially from CLI user side and your hooks)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
         - php: 5.5
         - php: 5.6
         - php: 7.0
-          env: SYMFONY_VERSION="~2.8"
+          env: SYMFONY_VERSION="^2.8"
 
 sudo: false
 
@@ -22,7 +22,7 @@ cache:
 
 before_install:
     - git config --global github.accesstoken 5e7538aa415005c606ea68de2bbbade0409b4b8c
-    - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
+    - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"^2[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
 
 install:
     - travis_retry composer update $COMPOSER_FLAGS --no-interaction
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -34,6 +34,12 @@ CLI options
               | --rules       | Rules to be used            | option was added
               | --using-cache | Does cache should be used   | option was added
 
+CLI argument
+------------
+
+On 1.x line `path` argument allows you to specify the dir/file that will be fixed.
+On 2.x line `path` argument is a bit different. It is a mask for finder you are using (defined in your configuration file or default one). Only files pointed by both finder and CLI `path` argument will be fixed.
+
 Exit codes
 ----------
 
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -406,32 +406,27 @@ final class ConfigurationResolver
         // - when we don't have valid finder - prepare new iterator
         $iterator = null;
         $currentFinder = $this->config->getFinder();
+        $path = realpath($this->path);
+        $isFile = is_file($path);
+
+        if ($isFile) {
+            $callback = function (\SplFileInfo $current) use ($path) {
+                return $current->getRealPath() === $path;
+            };
+        } else {
+            $path .= DIRECTORY_SEPARATOR;
+            $callback = function (\SplFileInfo $current) use ($path) {
+                return 0 === strpos($current->getRealPath(), $path);
+            };
+        }
 
         try {
-            $currentIteratorItems = array_map(
-                function (\SplFileInfo $current) {
-                    return $current->getRealPath();
-                },
-                iterator_to_array($currentFinder, false)
-            );
-
-            $nestedIterator = is_file($this->path)
-                ? new \ArrayIterator(array(new SymfonySplFileInfo(
-                    $this->path,
-                    substr(dirname($this->path), strlen($this->cwd) + 1),
-                    substr($this->path, strlen($this->cwd) + 1)
-                )))
-                : Finder::create()->in($this->path)->getIterator()
-            ;
-
             $iterator = new \CallbackFilterIterator(
-                $nestedIterator,
-                function (\SplFileInfo $current) use ($currentIteratorItems) {
-                    return in_array($current->getRealPath(), $currentIteratorItems, true);
-                }
+                $currentFinder instanceof \IteratorAggregate ? $currentFinder->getIterator() : $currentFinder,
+                $callback
             );
         } catch (\LogicException $e) {
-            if ($this->path) {
+            if ($isFile) {
                 $iterator = new \ArrayIterator(array(new SymfonySplFileInfo(
                     $this->path,
                     substr(dirname($this->path), strlen($this->cwd) + 1),
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -14,11 +14,14 @@ namespace PhpCsFixer\Console;
 
 use PhpCsFixer\ConfigInterface;
 use PhpCsFixer\ConfigurationException\InvalidConfigurationException;
+use PhpCsFixer\Finder;
 use PhpCsFixer\FixerFactory;
 use PhpCsFixer\FixerInterface;
 use PhpCsFixer\RuleSet;
 use PhpCsFixer\StdinFileInfo;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder as SymfonyFinder;
+use Symfony\Component\Finder\SplFileInfo as SymfonySplFileInfo;
 
 /**
  * The resolver that resolves configuration to use by command line options and config.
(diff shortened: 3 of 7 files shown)


tifabien:
I'll test it asap but at the first look it seems it will solve my problem :) 
Thanks a lot!

keradus:
General question. 
We could backport this change to 1.x as well (ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Console/Command/FixCommand.php#L356)
It may be considered as behaviour change, and for that BC break - because one asked for fixing excluded file and it won't be fixed.
Also, it may be considered as bug fix, so not BC break - because one asked to fix file, use configuration that exclude this file so finally it should not be changed.

Any thoughts about change itself and then target version ?
cc @stof @sstok @GrahamCampbell @SpacePossum @gharlan @ceeram and everybody  ;)

GrahamCampbell:
I think it's fine on 2.x only.

keradus:
It shouldn't. Please measure and show.
We will need to do it that way on 2.x when we will have cachingIterator to lint files before fixing them, and we need to ommit some files before linting

tifabien:
:+1: I tested my side and it works like a charm :) 

Small question, Do you know when you'll plan to release the 2.x?

Thanks a lot !

keradus:
Great to hear!
I'm still waiting for some more opinions on this PR.

About release - I won't give you exact date, but I think we are close:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestones/v2.0

src/FileCacheManager.php
```diff
@@ -49,6 +50,7 @@ public function __construct($isEnabled, $cacheFile, array $rules)
     {
         $this->isEnabled = $isEnabled;
         $this->cacheFile = $cacheFile;
+        $this->cacheFileRealDirName = dirname(realpath($cacheFile));
```
SpacePossum: does `realpath` work (correctly) in `phar`'s? (I recall there were some issues with it)

tests/Console/ConfigurationResolverTest.php
```diff
@@ -246,11 +246,53 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+
+        $this->assertCount(0, iterator_to_array($this->resolver->getConfig()->getFinder()));
+    }
+
+    public function testResolvePathWithFileThatIsExcludedByDir()
+    {
+        $dir = realpath(__DIR__.'/..');
+        $this->config->getFinder()
+            ->in($dir)
+            ->exclude(substr(realpath(__DIR__), strlen($dir) + 1));
```
gharlan: is realpath necessary here?

tests/Console/ConfigurationResolverTest.php
```diff
@@ -246,11 +246,53 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+        $dir = realpath(__DIR__);
+        $this->config->getFinder()
+            ->in($dir)
+            ->notPath(substr(__FILE__, strlen($dir) + 1));
+
+        $this->resolver
+            ->setOption('path', __FILE__)
+            ->resolve();
+
+        $this->assertCount(0, iterator_to_array($this->resolver->getConfig()->getFinder()));
```
gharlan: no need to call `iterator_to_array`, `assertCount` can handle traversables.

SpacePossum: but I doesn't handle iterators well on all versions (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1632#issuecomment-167264865), we could try this case out though :)

gharlan: interesting. didn't know that.

gharlan:
> Any thoughts about change itself and then target version ?

:+1: for change itself, but only 2.x

SpacePossum:
Besides https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1742#discussion_r52050731 I would like to have a little explaining about the new behavior if that is ok :)
(maybe we could at it to the docs somewhere as well?)

If I understand the issue correctly together with the PR; the fixer should only fix the given file / files in the directory passed on by command line argument, but still honor the finder exclude rules set in the config file (and fixers/level)?
 (and not apply possibly broader finder rules that would be out of the directory passed on by the command line?)

keradus:
@SpacePossum 
it would create an intersection 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1742/files#diff-62b374855f5961454b817b144c531e27R425

So, if you run the tool with both finder in configuration and path as CLI arguments only the files that would be found by both finders will be fixed.

SpacePossum:
Thanks for the explaining :)
So if one would black list directory `a/b/c` in its config and than run the fixer using `a/b/c` as input path by the command line, what would happen?

keradus:
Intersection would be empty.
That's the point of @tifabien #1719

If one configure his project to not touch sth, it should not be touched. If one decided to change that configuration - that's what config file is for.
Also remember that one can't exclude path by CLI.

tests/Console/ConfigurationResolverTest.php
keradus: it will normalize that `/..` part

SpacePossum:
I see, I would like the command line input to be leading in this case. I thought the issue reported was something like;
- config blacklists dir `a/b/c`
- run with command line `a/b`
- `a/b/c` is still being fixed

your PR will fix that, so :+1:  :)
However as a result of the PR, it will be impossible to fix files in `a/b/c` using the command line without changing the config file. If one would explicitly pass `a/b/c` as command line arg, I would expected the fixer to ran over dir. Its a tricky one, I think both solutions have + and -, so I'm still :+1: for the change as it fixes the issue for some people and +0 for the side effect :)

src/FileCacheManager.php
keradus: I never had any problems with it, could you elaborate?

SpacePossum: Cant recall any details, sorry man, ignore my comment

tests/Console/ConfigurationResolverTest.php
gharlan: I can not see that part here.. ;)

keradus: if `__DIR__would be in linked path then realpath will reveal real path that link point to.

If you have project forked in non-linked path then indeed, `$dir = realpath(__DIR__.'/..');` will only move one dir up and then second realpath is not needed. But if you would have in in linked path, then first realpath will reveal linked point, so the second one is needed.

If you have some nice way of normalizing `/foo/bar/.././baz` into `/foo/baz` I'm willing to listen, as it's the only reason I called realpath here.

src/FileCacheManager.php
keradus: OK, I think I may know what was in your mind.
calling realpath on phar:// stream won't provide useful data, but calling it (even inside phar file) on file that is outside phar (like here) is ok

keradus:
Done from my side.
It would be really nice to close it fast ;)

tests/Console/ConfigurationResolverTest.php
gharlan: hmm, symlinks should be resolved in `__DIR__`, so it should be always equivalent to `realpath(__DIR__)`.

http://php.net/manual/en/language.constants.predefined.php

`__FILE__`:
> The full path and filename of the file **with symlinks resolved**.

`__DIR__`:
> This is equivalent to `dirname(__FILE__)`.

.travis.yml
```diff
@@ -22,7 +22,8 @@ cache:
 
 before_install:
     - git config --global github.accesstoken 5e7538aa415005c606ea68de2bbbade0409b4b8c
-    - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
+    - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"^2[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
+    - cat composer.json
```
gharlan: is this change intended in this pr?

gharlan: ah, it is needed for the polyfill component, right?

keradus: exactly

tests/Console/ConfigurationResolverTest.php
keradus: thanks!

gharlan:
@keradus please take a look at https://github.com/keradus/PHP-CS-Fixer/pull/15

tests/Console/ConfigurationResolverTest.php
```diff
@@ -245,11 +244,53 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+
+        $this->assertSame(__DIR__, $this->resolver->getPath());
+    }
+
+    public function testResolvePathWithFileThatIsExcludedDirectly()
+    {
+        $dir = __DIR__;
+        $this->config->getFinder()
+            ->in($dir)
+            ->notPath(substr(__FILE__, strlen($dir) + 1));
```
gharlan: isn't it the same as `basename(__FILE__)`?

UPGRADE.md
```diff
@@ -34,6 +34,12 @@ CLI options
               | --rules       | Rules to be used            | option was added
               | --using-cache | Does cache should be used   | option was added
 
+CLI argument
+------------
+
+On 1.x line `path` argument allow you to specify the dir/file that will be fixed.
```
gharlan: "allows"

tests/Console/ConfigurationResolverTest.php
```diff
@@ -245,11 +244,52 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+
+        $this->assertCount(0, $this->resolver->getConfig()->getFinder());
+    }
+
+    public function testResolvePathWithFileThatIsExcludedByDir()
+    {
+        $dir = realpath(__DIR__.'/..');
+        $this->config->getFinder()
+            ->in($dir)
+            ->exclude(substr(__DIR__, strlen($dir) + 1));
```
gharlan: `basename(__DIR__)`

tests/Console/ConfigurationResolverTest.php
```diff
@@ -245,11 +244,52 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+
+        $this->assertCount(0, $this->resolver->getConfig()->getFinder());
+    }
+
+    public function testResolvePathWithFileThatIsNotExcluded()
+    {
+        $dir = __DIR__;
+        $this->config->getFinder()
+            ->in($dir)
+            ->notPath('foo-'.substr(__FILE__, strlen($dir) + 1));
```
gharlan: basename

tests/Console/ConfigurationResolverTest.php
```diff
@@ -245,11 +244,52 @@ public function testResolveConfigFileChooseFileWithInvalidFile()
...
+        $this->resolver
+            ->setOption('path', __FILE__)
+            ->resolve();
+
+        $this->assertCount(0, $this->resolver->getConfig()->getFinder());
+    }
+
+    public function testResolvePathWithFileThatIsExcludedByDir()
+    {
+        $dir = realpath(__DIR__.'/..');
```
gharlan: `dirname(__DIR__)`

keradus:
> @keradus please take a look at keradus#15

merged, thank you @gharlan !

SpacePossum:
looking good :)
:+1:

src/Console/ConfigurationResolver.php
```diff
@@ -390,13 +392,51 @@ private function resolveConfig()
...
+                $callback
+            );
+        } catch (\LogicException $e) {
+            $iterator = $isFile
+                ? new \ArrayIterator(array(new SymfonySplFileInfo(
+                    $this->path,
+                    substr(dirname($this->path), strlen($this->cwd) + 1),
+                    substr($this->path, strlen($this->cwd) + 1)
+                )))
+                : $currentFinder->in($this->path)
```
gharlan: Isn't it possible that `$currentFinder` is not a symfony finder instance, so the method `in` would not exist?

keradus:
looks ready

gharlan:
:+1:

keradus:
Thank you @gharlan.

gharlan:
thanks!

Issue #1706, opened 2016-01-25
Iterating file twice when provided path twice.


keradus:
When we set the path that should be fixed in both places (config file + [CLI](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Console/Command/FixCommand.php#L360)) tool will iterate through both paths separately, even if those are the same path.

Maybe not a bug, but it will slow down the tool if one use it that way (and I found that it is very common) - an no notice to user is show that he is doing sth wrong, and the fixing part tooks twice much as it should (since it is indeed fixing every single file twice).

I think it was originally reported by @GrahamCampbell as problem connected with symlinks, but we cannot reproduce it then.

GrahamCampbell:
I just use the following patch:

```
diff --git a/src/Fixer.php b/src/Fixer.php
index f020827..2e3acdb 100644
--- a/src/Fixer.php
+++ b/src/Fixer.php
@@ -137,12 +137,20 @@ public function fix(ConfigInterface $config, $dryRun = false, $diff = false)
             $config->getRules()
         );
 
+        $processed = array();
+
         foreach ($config->getFinder() as $file) {
-            if ($file->isDir() || $file->isLink()) {
+            $name = $this->getFileRelativePathname($file);
+
+            if (isset($protected[$name])) {
                 continue;
             }
 
-            $name = $this->getFileRelativePathname($file);
+            $processed[$name] = true;
+
+            if ($file->isDir() || $file->isLink()) {
+                continue;
+            }
 
             $this->stopwatch->start($name);
 
```

keradus:
Yes I saw that in StyleCI fork.

Even with that it's still iterating through every files twice.
The user call is bad, bad still we could do sth with it.

If we won't go with any other solution I would go that way indeed, but I have a strong feeling that it is fixing the symptom, not the reason :(

GrahamCampbell:
> but I have a strong feeling that it is fixing the symptom, not the reason :(

Yeh, it absolutely is. I just needed a fast hack to fix the worst effect of it.

keradus:
ref 1c9a7f4

PR #1743, opened 2016-02-03
Fixer - fix rare situation when file was visited twice

keradus:
Closing #1706
Developed during #1666

NOTE: during merging change target to 1.11


keradus:
cc @GrahamCampbell

PR #1744, opened 2016-02-03
Fixer - fix rare situation when file was visited twice

keradus:
Closing #1706
Developed during #1666


keradus:
cc @GrahamCampbell

GrahamCampbell:
Surely this is slower than the alternative I suggested?

keradus:
It shouldn't. Please measure and show.
We will need to do it that way on 2.x when we will have cachingIterator to lint files before fixing them, and we need to ommit some files before linting

keradus:
I tried to measure real impact this has on running the tool. The difference in measurement error rate

Issue #1746, opened 2016-02-03
Symfony Level + phpdoc_order leaves docblock not totally fixed.


cryode:
CS Fixer version: 1.11.1

Sample code:

```php
<?php

class Foo
{
    /**
     * @param Exception $exception
     * @param Closure $response
     * @param string $context
     * @return Foo
     * @throws \RuntimeException
     */
    public function handleError(\Exception $exception, \Closure $response, $context)
    {
        // implementation...
    }
}
```

Run only the required phpdoc fixers to achieve the results we want (aligned, ordered, separated, trimmed):

    php-cs-fixer fix example.php --fixers="phpdoc_params,phpdoc_order,phpdoc_separation,phpdoc_trim"

Achieve expected result:

```php
<?php

class Foo
{
    /**
     * @param Exception $exception
     * @param Closure   $response
     * @param string    $context
     *
     * @throws \RuntimeException
     *
     * @return Foo
     */
    public function handleError(\Exception $exception, \Closure $response, $context)
    {
        // implementation...
    }
}
```

However, running the Symfony level (which has all of those except order) plus `phpdoc_order` results in a mis-trim:

    php-cs-fixer fix example.php --level="symfony" --fixers="phpdoc_order"

```php
<?php

class Foo
{
    /**
     * @param Exception $exception
     * @param Closure   $response
     * @param string    $context
     *
     * @throws \RuntimeException
     * @return Foo
     *
     */
    public function handleError(\Exception $exception, \Closure $response, $context)
    {
        // implementation...
    }
}
```

It's correct except for that last line not being trimmed. Running the same command again will fix this as expected and you end up with the desired result.

I can only assume that this is caused either by an order of operations (trim running before something else and not being run again), or some other conflict with an existing Symfony level fixer. I would play with it more to try and figure it out for sure if it weren't the middle of a work day. :)

GrahamCampbell:
The order fixer doesn't run on symfony because symfony use a different order.

GrahamCampbell:
> However, running the Symfony level (which has all of those except order) plus phpdoc_order results in a mis-trim:

Does that line have a trailing space on it by any chance?

cryode:
> The order fixer doesn't run on symfony because symfony use a different order.

I understand why the order fixer isn't in the Symfony level. Just pointing out that all the fixers I'd like to run should be specified by combining the Symfony level with the `phpdoc_order` contrib fixer, yet the result is not the same.

[...]

PR #1747, opened 2016-02-04
Fixer - Fix ordering of fixer when both level and custom fixers are used

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1746

### DIFF
--- a/Symfony/CS/Fixer.php
+++ b/Symfony/CS/Fixer.php
@@ -78,7 +78,10 @@ class Fixer
         }
     }
 
-    public function registerCustomFixers($fixers)
+    /**
+     * @param FixerInterface[] $fixers
+     */
+    public function registerCustomFixers(array $fixers)
     {
         foreach ($fixers as $fixer) {
             $this->addFixer($fixer);
@@ -95,7 +98,7 @@ class Fixer
      */
     public function getFixers()
     {
-        $this->sortFixers();
+        $this->fixers = $this->sortFixers($this->fixers);
 
         return $this->fixers;
     }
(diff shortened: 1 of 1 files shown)


keradus:
what was the problem?

SpacePossum:
Inside the Fixer there is a lot of juggling with the fixers it should use, i.e. `$this->fixers` vs. `$fixers` from the configuration etc. I would suggest to make it more clean, i.e. use `$this->fixers` everywhere.
Because of this currently when you configure a level (on > master) and add some fixers, the fixers from the level are sorted but the complete set (level + added fixers) is not. This caused the fixers to be ran in the wrong order.

keradus:
Then please, could you make a utest to cover that?
I had strong feeling that it would need to be rewritten on 2.x line...

SpacePossum:
That test should cover the `FixCommand`, `Configuration(Resolver)` and `Fixer` because the provided fail-case did work as integration test, proving even more the complex interaction between the listed 3 parts.
I might find some time to prove the failure and fix by utest later this week or maybe next.

keradus:
I'm just afraid that if we merge it without utest and then merge 1.12 into 2.x line, which has completely different way preparing fixers (own Factory class), we could lost the fix.

keradus:
Tests would indeed be tricky and even not unit. Could you check if that error is already resolved on 2.x line ?

keradus:
Could you check if that error is already resolved on 2.x line ?

keradus:
ping @SpacePossum

SpacePossum:
Sorry about letting you wait.
Running the Fixer on master on the sample code provided with;
```
s1@possum:/home/possum/work/PHP-CS-Fixer# php php-cs-fixer fix test.php --rules=@Symfony,phpdoc_order
```

works as expected. So your factory class seems to work fine :)

keradus:
Cool. Knowing this is very important for merging 1.x->2.x

keradus:
Thank you @SpacePossum.

PR #1750, opened 2016-02-06
Add PhpdocSingleLineVarSpacingFixer

SpacePossum:
Fixes spacing around single line `@var` PHPDoc's.
For example:
Input
```php
                <?php
                    /**    @var   MyCass1 $test1      description   and more.*/
                    $test = 1;

                    /**    @var   MyCass2    description and    such. */
                    $test = 2;

                    /** @var   MyCass3    description.    */
                    $test = 3;
```
Output
```php
                <?php
                    /** @var MyCass1 $test1 description   and more. */
                    $test = 1;

                    /** @var MyCass2 description and    such. */
                    $test = 2;

                    /** @var MyCass3 description. */
                    $test = 3;
```

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocSingleLineVarSpacingFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocSingleLineVarSpacingFixer.php
@@ -1,0 +1,95 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * Fixer for part of rule defined in PSR5 ¶7.22.
+ *
+ * @author SpacePossum
+ */
+final class PhpdocSingleLineVarSpacingFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        /** @var Token $token */
+        foreach ($tokens as $index => $token) {
+            if ($token->isGivenKind(T_DOC_COMMENT)) {
+                $token->setContent($this->fixTokenContent($token->getContent()));
+                continue;
+            }
+
+            if (!$token->isGivenKind(T_COMMENT)) {
+                continue;
+            }
+
+            $content = $token->getContent();
+            $fixedContent = $this->fixTokenContent($content);
+            if ($content !== $fixedContent) {
+                $tokens->overrideAt($index, array(T_DOC_COMMENT, $fixedContent, $token->getLine()));
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * @param string $content
+     *
+     * @return string
+     */
+    private function fixTokenContent($content)
+    {
+        return preg_replace_callback(
+            '#^/\*\*[ \t]*@var[ \t]+(\S+)[ \t]*(\$\S+)?[ \t]*([^\n]*)\*/$#',
+            function (array $matches) {
+                $content = '/** @var';
+                for ($i = 1, $m = count($matches); $i < $m; ++$i) {
+                    if ('' !== $matches[$i]) {
+                        $content .= ' '.$matches[$i];
+                    }
+                }
+
+                $content = rtrim($content);
+
+                return $content.' */';
+            },
+            $content
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Single line @var PHPDoc should have proper spacing.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be ran after the PhpdocTypeToVarFixer.
+        return -10;
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Symfony/PhpdocSingleLineVarSpacingFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/PhpdocSingleLineVarSpacingFixerTest.php
@@ -64,16 +64,14 @@ final class PhpdocSingleLineVarSpacingFixerTest extends AbstractFixerTestBase
                     $test2 = 3;
 
                     class A {
+                        /** @var MyCass4 aa */
+                        public $test4 = 4;
 
-                    /** @var MyCass4 aa */
-                    public $test4 = 4;
+                        /** @var MyCass5 */
+                        public $test5 = 5;
 
-                    /** @var MyCass5 */
-                    public $test5 = 5;
-                    }
-
-                    /** @var MyCass6 */
-                    public $test6 = 6;
+                        /** @var MyCass6 */
+                        public $test6 = 6;
                     }
                 ',
                 '<?php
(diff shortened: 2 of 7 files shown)


GrahamCampbell:
It doesn't actually check that it IS single line?

Symfony/CS/Fixer/Contrib/PhpdocSingleLineVarSpacingFixer.php
```diff
@@ -0,0 +1,66 @@
...
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        /** @var Token $token */
+        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
+            $content = $token->getContent();
+            $content = preg_replace_callback(
+'               #/\*\*[ ]*@var[ ]+([\S]+)[ ]+(\$[\S]+)?[ ]+([^\n]*)\*/#',
```
gharlan: Unneccessary single char character classes..

SpacePossum: :+1: check, added tab as well

SpacePossum:
> It doesn't actually check that it IS single line?

I think it does, the line break in the regex prevents matching anything with a line break in it.

SpacePossum:
One file hit on SF, not worth PR'ing IMHO.

```diff
diff --git a/src/Symfony/Component/Process/ProcessPipes.php b/src/Symfony/Component/Process/ProcessPipes.php
index b2d09b8..33700cf 100644
--- a/src/Symfony/Component/Process/ProcessPipes.php
+++ b/src/Symfony/Component/Process/ProcessPipes.php
@@ -26,9 +26,9 @@ class ProcessPipes
     private $fileHandles = array();
     /** @var array */
     private $readBytes = array();
-    /** @var bool    */
+    /** @var bool */
     private $useFiles;
-    /** @var bool    */
+    /** @var bool */
     private $ttyMode;
 
     const CHUNK_SIZE = 16384;
```

Symfony/CS/Fixer/Contrib/PhpdocSingleLineVarSpacingFixer.php
```diff
@@ -0,0 +1,66 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        /** @var Token $token */
+        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
```
keradus: iterate through every tokens and check `isGivenKind`

Symfony/CS/Tests/Fixer/Contrib/PhpdocSingleLineVarSpacingFixerTest.php
```diff
@@ -0,0 +1,91 @@
...
+    public function test($required, $label)
+    {
+
+    }
+
+    /** @var   MyCass3
+    */
+    private $test0 = 3;
+
+    /** @var   MyCass3 aa       */
```
keradus: question: why we don't want this one to be fixed?

Symfony/CS/Fixer/Contrib/PhpdocSingleLineVarSpacingFixer.php
```diff
@@ -0,0 +1,66 @@
...
+            $content = $token->getContent();
+            $content = preg_replace_callback(
+'               #/\*\*[ \t]*@var[ \t]+([\S]+)[ \t]+(\$[\S]+)?[ \t]+([^\n]*)\*/#',
```
gharlan: `[\S]` -> `\S`

gharlan: and please add `^` and `$`.

SpacePossum: >[\S] -> \S

will do :)

>and please add ^ and $.

Not sure about those. We know the string will always start with '/**' and end with '*/' for being parsed as doc token. Both strings cannot occur more than once and the fixer doesn't care what is in front or beyond it. So what would be the benefit, would it process faster?

keradus: > Both strings cannot occur more than once 

https://3v4l.org/lassV

gharlan: > would it process faster?

I think so, because there are less possibilities that have to be checked by the regex engine.
It is faster for non-matching doc blocks, not for the matching ones.

> Both strings cannot occur more than once

The first (`/**`) can occur multiple times..

Symfony/CS/Tests/Fixer/Contrib/PhpdocSingleLineVarSpacingFixerTest.php
SpacePossum: we should ;)
updated!

Symfony/CS/Fixer/Contrib/PhpdocSingleLineVarSpacingFixer.php
SpacePossum: thanks for the pointers, great stuff to know :)

Symfony/CS/Fixer/Contrib/PhpdocSingleLineVarSpacingFixer.php
```diff
@@ -0,0 +1,70 @@
...
+
+        /** @var Token $token */
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $content = $token->getContent();
+            $content = preg_replace_callback(
+'               #^/\*\*[ \t]*@var[ \t]+(\S+)[ \t]+(\$*\S+)?[ \t]+([^\n]*)\*/$#',
```
keradus: - no need for spaces before `#`
- missing utests without space before `@` (`/**@var Foo*/`)
- I think we don't want to fix invalid docs with `$$$$$$$$var_name` in second capturing group

(btw, please move to Symfony level)

keradus:
Fixer should run after `phpdoc_type_to_var`

GrahamCampbell:
> Fixer should run after phpdoc_type_to_var

No, it should just recognize `@type` as well as `@var` like all the other fixers do?

keradus:
I don't have need to support non-standard annotation in every single fixer.
Actually, I just don't know what `@type` annotation is. This fixer follows PSR5 proposal and fixes one of its §, which says nothing about `@type`.

GrahamCampbell:
> This fixer follows PSR5 proposal and fixes one of its §, which says nothing about @type.

It did when I originally wrote the fixers though.

GrahamCampbell:
> And custom annotation could means anything, so it's better to not mess with it.

By that logic, it absolutely must be removed from our tag class then?

keradus:
No, as we support converting it to proper one if one follows old PSR-5 one day.

GrahamCampbell:
> No, as we support converting it to proper one if one follows old PSR-5 one day.

Well we can't have one rule for one, and one for another. We need to either change this fixer to deal with `@type`, or rip that stuff out the other fixers.

GrahamCampbell:
It sounds to me like you want to rip it out of the other fixers, but keep the converting fixer?

keradus:
I just don't like to propagate fixing that violates PSR-5 proposal

SpacePossum:
expecting test to fail because of add test without fix;

> I think we don't want to fix invalid docs with $$$$$$$$var_name in second capturing group

Symfony/CS/Tests/Fixer/Symfony/PhpdocSingleLineVarSpacingFixerTest.php
```diff
@@ -0,0 +1,137 @@
...
+    public function test($required, $label)
+    {
+
+    }
+
+    /** @var   MyCass3
+    */
+    private $test0 = 0;
+
+    /** @var      int           $$$$$$$custom */
```
keradus: OK, I didn't find a quick and nice way to deal with that one case. As it's extra rare we could give it a go. Please remove that part (`$test1`)

keradus:
One minor and good to go

Symfony/CS/Fixer/Symfony/PhpdocSingleLineVarSpacingFixer.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * Fixer for part of rule defined in PSR5 ¶7.22.
```
GrahamCampbell: This isn't quite true. PSR5 explicitly says that it doesn't define CS rules.

SpacePossum: that what the `part` hints at ;)

keradus:
Please @SpacePossum , could you handle non-working Travis ?

keradus:
Thank you @SpacePossum

PR #1617, opened 2015-12-20
DirConstantFixer, ModernizeTypesCastingFixer, RandomApiMigrationFixer: [WIP]

kalessil:
Relevant fixers status:

- IsNull: not included (still my FB only)
- AliasFunctionsFixer: needs to be re-created on top of AbstractFunctionReferenceFixer
- 1.12 base PR not possible (UTs crashed, interfaces changes?)

Changes:

- AbstractFunctionReferenceFixer interfaces were consolidated with existing
- Block constructs detection re-factored according hints
- Extended UTs cases

Still open:

- Risky/nor risky notes for all 3 fixers (@keradus : feedback needed)

PR #1751, opened 2016-02-06
FixerInterface::configure - method should always override configuration, not patch it

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1617#discussion_r49540026


keradus:
fabbot complains as always

SpacePossum:
:+1:  looking good :)

PR #1753, opened 2016-02-07
Performance improvement

GrahamCampbell:
We should mark the file as processed before we bother checking if it's valid so we don't double check files we no are invalid.

My example code did that already, but your iterator implementation currently does not.

### DIFF
--- a/src/UniqueFileIterator.php
+++ b/src/UniqueFileIterator.php
@@ -28,10 +28,6 @@ final class UniqueFileIterator extends \FilterIterator
         /** @var SplFileInfo $file */
         $file = $this->current();
 
-        if ($file->isDir() || $file->isLink()) {
-            return false;
-        }
-
         $path = $file->getRealPath();
 
         if (array_key_exists($path, $this->visitedElements)) {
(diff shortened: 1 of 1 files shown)


src/UniqueFileIterator.php
```diff
@@ -28,10 +28,6 @@ public function accept()
         /** @var SplFileInfo $file */
         $file = $this->current();
 
-        if ($file->isDir() || $file->isLink()) {
-            return false;
-        }
-
         $path = $file->getRealPath();
 
         if (array_key_exists($path, $this->visitedElements)) {
```
GrahamCampbell: isset is actually faster on php 5 than array_key_exists - do you want me to change this @keradus?

keradus:
it should go to 1.11, isn't it ?

src/UniqueFileIterator.php
keradus: why not. globally from at least 1.12 ;)

GrahamCampbell:
> it should go to 1.11, isn't it ?

Yeh.

GrahamCampbell:
See #1758.

PR #1757, opened 2016-02-07
Psr0Fixer - change way of configuring the fixer

keradus:
`ConfigInterface::getDir` is actually used only once. I want to get rid of it from `ConfigInterface`.
To do so let us replace the only usage with simply configuring the fixer.


SpacePossum:
nice, the interface is already so big I welcome any simplification :)

PR #1758, opened 2016-02-07
Performance improvement

GrahamCampbell:
We should mark the file as processed before we bother checking if it's valid so we don't double check files we no are invalid.

---

Replaces #1753.

PR #1759, opened 2016-02-07
UniqueFileIterator - performance improvement

GrahamCampbell:
We should mark the file as processed before we bother checking if it's valid so we don't double check files we no are invalid.

---

Replaces #1753 and #1758.


keradus:
Thank you @GrahamCampbell.

PR #1762, opened 2016-02-08
Remove ConfigInterface::getDir, ConfigInterface::setDir, Finder::setDir and whole FinderInterface

keradus:


keradus:
ref #1763

tests/Console/ConfigurationResolverTest.php
```diff
@@ -246,10 +246,10 @@ public function testResolvePathRelative()
     {
         $this->resolver
             ->setCwd(__DIR__)
-            ->setOption('path', 'Foo'.DIRECTORY_SEPARATOR.'Bar')
+            ->setOption('path', 'Command')
```
gharlan: why this change?

keradus: Because now it will call `finder->in(...)`, which would blow up with exception of non-existing `Foo\Bar` dir.

before it was only setting the property (which was later used in `finder->in(...)` call and then throw exception, but utest didn't went so far)

gharlan: ok

gharlan:
:+1:

Issue #1688, opened 2016-01-19
PhpDoc: no empty docblock


keradus:
Target: 1.x line
Motivation: after removing some tags (eg `@access`) the docblock could be empty.

Before:
```php
<?php

/**
 *
 *
 */
function baz()
{
}
```

After:
```php
<?php

function baz()
{
}
```

PR #1765, opened 2016-02-09
Added NoEmptyPhpdocFixer

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1688

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -411,6 +411,10 @@ Choose from the list of available fixers:
                         lines between docblock and the
                         documented element.
 
+* **no_empty_phpdoc** [symfony]
+                        There should not be empty
+                        PHPDoc blocks.
+
 * **object_operator** [symfony]
                         There should not be space
                         before or after object
--- a/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
+++ b/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
@@ -36,6 +36,9 @@ abstract class AbstractFixerTestBase extends \PHPUnit_Framework_TestCase
     protected function getFixer()
     {
         $name = 'Symfony\CS\Fixer'.substr(get_called_class(), strlen(__NAMESPACE__), -strlen('Test'));
+        if (!class_exists($name)) {
+            throw new \LogicException(sprintf('Fixer "%s" for test not found.', $name));
+        }
 
         return new $name();
     }
--- a/Symfony/CS/AbstractAnnotationRemovalFixer.php
+++ b/Symfony/CS/AbstractAnnotationRemovalFixer.php
@@ -25,8 +25,8 @@ abstract class AbstractAnnotationRemovalFixer extends AbstractFixer
      */
     public function getPriority()
     {
-        // must be run before the PhpdocSeparationFixer, PhpdocOrderFixer, and
-        // PhpdocTrimFixer
+        // must be run before the PhpdocSeparationFixer, PhpdocOrderFixer,
+        // PhpdocTrimFixer and PhpdocNoEmptyReturnFixer.
         return 10;
     }
 
--- a/Symfony/CS/Console/Command/SelfUpdateCommand.php
+++ b/Symfony/CS/Console/Command/SelfUpdateCommand.php
@@ -32,7 +32,7 @@ class SelfUpdateCommand extends Command
         $this
             ->setName('self-update')
             ->setAliases(array('selfupdate'))
-            ->setDescription('Update php-cs-fixer.phar to the latest version.')
+            ->setDescription('Update php-cs-fixer.phar to the latest stable version.')
             ->setHelp(<<<'EOT'
 The <info>%command.name%</info> command replace your php-cs-fixer.phar by the
 latest version from cs.sensiolabs.org.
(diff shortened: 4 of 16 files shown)


keradus:
`NoEmptyPhpdocFixer` ?

Symfony/CS/Tests/Fixer/Contrib/EmptyPhpdocFixerTest.php
```diff
@@ -0,0 +1,77 @@
...
+                    /**  *//** a *//**  */
+
+                    /**
+                    */
+
+                    /**
+                     *
+                     */
+
+                    /** ***
```
keradus: This is not empty docblock. it's odd, but not empty. The Summary is `***`.

README.rst
```diff
@@ -615,6 +615,10 @@ Choose from the list of available fixers:
                         construct to print if
                         possible.
 
+* **empty_phpdoc** [contrib]
```
keradus: Symfony?

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
```diff
@@ -36,6 +36,9 @@ public static function setUpBeforeClass()
     protected function getFixer()
     {
         $name = 'Symfony\CS\Fixer'.substr(get_called_class(), strlen(__NAMESPACE__), -strlen('Test'));
+        if (!class_exists($name)) {
+            $this->fail(sprintf('Fixer "%s" for test not found.', $name));
```
keradus: ???

Symfony/CS/Tests/Fixer/Contrib/EmptyPhpdocFixerTest.php
SpacePossum: I know, I don't like those in my projects.. do you want these?

Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php
SpacePossum: DX, readability of the fail is better than a fatal stack dump

keradus: fail mark the test as failed, and here we are not in test...

GrahamCampbell: Just throw an exaception?

Symfony/CS/Tests/Fixer/Contrib/EmptyPhpdocFixerTest.php
keradus: It's a user case thing. Taking things strictly - this is part of docblock, so it's not empty...
I don't like them too, but still...

GrahamCampbell: Maybe we should define empty more loosly has having nothing useful in them?

README.rst
SpacePossum: I'll ask them if they reply on my PR overthere, till than I'm not going to the trouble to check it all out. If they want it we can always move it to SF level.

Symfony/CS/Tests/Fixer/Contrib/EmptyPhpdocFixerTest.php
SpacePossum: very well, let me try out some more regex stuff.

SpacePossum:
closing till I figured out how to fix it (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1765#discussion_r52298636)

README.rst
keradus: No need to ask at first, I could send the PR (in the and you can treat PR as extra type of issue with code attached).
But seriously, I don't think there will be any violation for that rule.

The only case I met and why I want that fixer is our annotation cleaning fixers - eg when we remove `@access` annotation and there is nothing left in docblock

keradus:
@SpacePossum , please play with utests so the utests will define everything that should be fixed and how - when it will be done (or it is already?) I will help with regexp

SpacePossum:
Renamed, replaced the `fail` with `throw exception` (thanks @GrahamCampbell ) and added a bunch of integration tests (include case for removal of `@access` thanks @keradus).
Given that some of the tests are still running, I think the one thing (and most important thing) is to figure out the rule on which to remove a PHPDoc.
I like it as is, but I see the point @keradus is making.
It will be tricky to tweak though, so lets list some cases here.
I think we agree on these:
```php
/** */

/** 
*/

/** 
 *
 *
 */
```
However, the fixer also removes:
```php
/**
**/

/*******
 *
 *****/

/**
 * *
 */
```
I'm happy with both sets, so let me know were to aim for :)

keradus:
For 6 cases you provided I woudn't touch only the last one, in which `*` is summary part of PHPDoc. Odd, not nice, but still valid.

gharlan:
I like it as-is.

keradus:
@SpacePossum All cases you provided are also in utests, am I right ?

SpacePossum:
@keradus yes I think so, I checked and cannot think of anymore other cases, let me know I you have any in mind.

keradus:
Cool!
Please, move it to Symfony level

SpacePossum:
>Please, move it to Symfony level

:+1:  done :)

keradus:
Thank you @SpacePossum.

Issue #1768, opened 2016-02-10
php-cs-fixer internal error


glensc:
reproducer is here: https://gist.github.com/glensc/62561fe19c26ce2bac10

as mentioned here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1767#issuecomment-182452772

the -vvv does not give any useful information how to debug more or even know which fixer or which line or what internal error happened:

```
➔ php-cs-fixer.phar fix -vvv
Loaded config from "/home/glen/scm/php-cs-fixer/bug-1768/.php_cs"
You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.
E
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[2.007 s] color_smartytmpl.php

Fixed all files in 2.010 seconds, 6.500 MB memory used

Files that were not fixed due to internal error:
   1) color_smartytmpl.php
```

```
➔ php-cs-fixer.phar --version
PHP CS Fixer version 1.11.1 by Fabien Potencier (2c9f829)
```

keradus:
please provide version:
`php-cs-fixer --version`

glensc:
@keradus obviously i tested with latest code before even bothering reporting, and the full reproducer is cloneable, just run and see.

keradus:
and for me latest is 2.x@dev, for somebody 1.12, that why I'm asked.

Thank you for updating the report!

glensc:
the self-update does not upgrade further:

```
➔ php-cs-fixer.phar self-update
php-cs-fixer is already up to date.
```

glensc:
- HEAD aka v0.12 1a6c88f - broken
- master aka 2.0-DEV a674d59 - doesn't work as .php_cs references inexistent classes (`Class 'Symfony\CS\Finder\DefaultFinder' not found`)

keradus:
yep, it provide only stable releases
I will play with it and send PR with utest showing the error.

keradus:
2.x, as it is MAJOR change, has a lot of BC breakers ;)

keradus:
Handling of internal error protected you before creating invalid PHP file.

Then, what need to be fixed is #1769

keradus:
It **works** on 2.x line without any changes ! (thanks to our transformers)
It would work on 1.11 after 1769

PR #1769, opened 2016-02-10
FunctionDeclarationFixer - fix more cases

keradus:
ref #1768


glensc:
so, i removed those silly comments on lines 3&4 (of the gist on referenced issue), but it still gets internal error. what exactly is causing the error?

glensc:
nevemind. mislooked

keradus:
fixed

SpacePossum:
Looks like a solid fix :+1: 
I guess 1.x only? (i.e. the line break being dealt with by the transformers on 2.x?)
Do you know any other fixer might be effected as well?

keradus:
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1768#issuecomment-182594667

hard to guess

PR #1776, opened 2016-02-15
EofEndingFixer - new line on end line comment is allowed

Slamdunk:
The new line of a comment in the last line is indeed a new line and no other new lines should be attached.

As of yet, this code (I will expand line-feeds):
```php
<?php[\n]
return true;[\n]
// Comment[\n]
```
is fixed like
```php
<?php[\n]
return true;[\n]
// Comment[\n]
[\n]
```
Which breaks the statement of the fixer:
`A file must always end with a single empty line feed.`
and also the Unix file management.

### DIFF
--- a/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
+++ b/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
@@ -40,13 +40,25 @@ class EofEndingFixer extends AbstractFixer
             return $content;
         }
 
-        if ($token->isWhitespace()) {
-            $lineBreak = false === strrpos($token->getContent(), "\r") ? "\n" : "\r\n";
-            $token->setContent($lineBreak);
-        } elseif ($token->isComment() and '//' === substr($token->getContent(), 0, 2)) {
+        $isSingleLineComment = function (Token $token) {
+            return $token->isComment() and '//' === substr($token->getContent(), 0, 2);
+        };
+        $clearSingleLineComment = function (Token $token) {
             $content = $token->getContent();
             $content = rtrim($content, "\n")."\n";
             $token->setContent($content);
+        };
+
+        if ($token->isWhitespace()) {
+            if ($count > 1 and $isSingleLineComment($tokens[$count - 2])) {
+                $clearSingleLineComment($tokens[$count - 2]);
+                $token->clear();
+            } else {
+                $lineBreak = false === strrpos($token->getContent(), "\r") ? "\n" : "\r\n";
+                $token->setContent($lineBreak);
+            }
+        } elseif ($isSingleLineComment($token)) {
+            $clearSingleLineComment($token);
         } else {
             $tokens->insertAt($count, new Token(array(T_WHITESPACE, "\n")));
         }
--- a/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
+++ b/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
@@ -43,6 +43,10 @@ class EofEndingFixer extends AbstractFixer
         if ($token->isWhitespace()) {
             $lineBreak = false === strrpos($token->getContent(), "\r") ? "\n" : "\r\n";
             $token->setContent($lineBreak);
+        } elseif ($token->isComment() and '//' === substr($token->getContent(), 0, 2)) {
+            $content = $token->getContent();
+            $content = rtrim($content, "\n")."\n";
+            $token->setContent($content);
         } else {
             $tokens->insertAt($count, new Token(array(T_WHITESPACE, "\n")));
         }
--- a/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
+++ b/Symfony/CS/Fixer/PSR2/EofEndingFixer.php
@@ -41,7 +41,7 @@ class EofEndingFixer extends AbstractFixer
         }
 
         $isSingleLineComment = function (Token $token) {
-            return $token->isComment() and '/*' !== substr($token->getContent(), 0, 2);
+            return $token->isComment() && '/*' !== substr($token->getContent(), 0, 2);
         };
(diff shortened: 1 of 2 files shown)


Symfony/CS/Fixer/PSR2/EofEndingFixer.php
```diff
@@ -43,6 +43,10 @@ public function fix(\SplFileInfo $file, $content)
         if ($token->isWhitespace()) {
             $lineBreak = false === strrpos($token->getContent(), "\r") ? "\n" : "\r\n";
             $token->setContent($lineBreak);
+        } elseif ($token->isComment() and '//' === substr($token->getContent(), 0, 2)) {
```
SpacePossum: I think this part is not needed?

```php
 and '//' === substr($token->getContent(), 0, 2)
```

SpacePossum:
Thank you for the PR :)
It seems another fixer is effected with the good'ol linebreak-added-to-comment-token ^-^

keradus:
Yep, it should already works on 2.x line

Slamdunk:
@SpacePossum all the redundant code I added, like the one you pointed out, is due to the fact that `// A comment[\n]` is a single token, and not 2 separate tokens.

So, I found no other way to pass the 2 tests:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php#L87-L92
because of that way of PHP to tokenize the single line comment.

keradus:
Infos:
- that 2 tests are important for caching the tokenizer result, and in the end  - speed of whole tool.
- comment token with trailing new-lines is split into comment + whitespace tokens on 2.x line

Slamdunk:
@keradus Info nr 1: yup, I knew and understand, I was just point out that clearing a comment token and adding a new-line token it is not feasable because after cache clearing they will result in a single comment token with new-line at the end.

So the code can't be simple and clear because we need to fix tokens in the order and appearance they are before fixing.

Slamdunk:
To be more clear on the need of // check: https://3v4l.org/VlKOo

SpacePossum:
please a test for hash comments as well, I think these act the same as `//`
```php
<?php 
# comment
var_dump(token_get_all("<?php # comment\n"));
```

keradus:
`//` and `#` act the same way
`/*` is different, as it needs to be closed explicitly by `*/`

Slamdunk:
At the beginning I skipped the `#` comment because of the `HashToSlashCommentFixer` that I think everybody uses. Btw now is considered.

@SpacePossum there is no other way to check between `/*` and `//`.
Counting new-lines is ineffective: `// comment[\n]` ~ `/* comment[\n]*/`

keradus:
> At the beginning I skipped the # comment because of the HashToSlashCommentFixer that I think everybody uses. Btw now is considered.

Please do not assume that.

Symfony/CS/Tests/Fixer/PSR2/EofEndingFixerTest.php
```diff
@@ -102,6 +102,39 @@ public function provideCases()
 <?php
 
 ?>Inline2\r\n", ),
+            array(
+                "<?php return true;\n// A comment\n",
```
keradus: it's duplicated by next case, please remove

Symfony/CS/Tests/Fixer/PSR2/EofEndingFixerTest.php
```diff
@@ -102,6 +102,39 @@ public function provideCases()
...
+                "<?php return true;\n// A comment",
+            ),
+            array(
+                "<?php return true;\n// A comment\n",
+                "<?php return true;\n// A comment\n\n",
+            ),
+            array(
+                "<?php return true;\n# A comment\n",
```
keradus: same here

Symfony/CS/Tests/Fixer/PSR2/EofEndingFixerTest.php
```diff
@@ -102,6 +102,39 @@ public function provideCases()
...
+                "<?php return true;\n# A comment",
+            ),
+            array(
+                "<?php return true;\n# A comment\n",
+                "<?php return true;\n# A comment\n\n",
+            ),
+            array(
+                "<?php return true;\n/*\nA comment\n*/\n",
```
keradus: and here

Slamdunk:
It seems that Github inline comments don't like rebase: I was expecting the usual `[...] commented on an outdated diff [...]` but your comments are still active on different lines.

Btw as you can see, I committed the changes and rebased the branch.

keradus:
:+1:


Note to myself: extra cautious is needed when merging to 2.x line

Symfony/CS/Fixer/PSR2/EofEndingFixer.php
```diff
@@ -40,9 +40,25 @@ public function fix(\SplFileInfo $file, $content)
             return $content;
         }
 
+        $isSingleLineComment = function (Token $token) {
+            return $token->isComment() and '/*' !== substr($token->getContent(), 0, 2);
```
SpacePossum: please use `&&` over `and` (I think?)

keradus: yep

langj:
Is there somthing preventing this PR from being merged? I would like to get this fix.

SpacePossum:
I think it is OK, please note that people don't get notifications on commits alone. 
So @keradus might just have missed the fact that @Slamdunk added something to this PR.

keradus:
@SpacePossum that's right ;)

keradus:
Note to myself:
- merge to 1.11 instead of targeted 1.12
- simplify when merging to 2.0 (as it's not needed on 2.x line)

keradus:
Thank you @Slamdunk

PR #1781, opened 2016-02-16
Update README.rst

fitzel:
needs an Update, took me a while to figure out that the autoloader was not actually broken ;)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -751,12 +751,12 @@ to the default list of symfony-level fixers:
 
     <?php
 
-    $finder = Symfony\CS\Finder::create()
+    $finder = Symfony\CS\Finder\DefaultFinder::create()
         ->exclude('somedir')
         ->in(__DIR__)
     ;
 
-    return Symfony\CS\Config::create()
+    return Symfony\CS\Config\Config::create()
         ->fixers(array('strict_param', 'short_array_syntax'))
         ->finder($finder)
     ;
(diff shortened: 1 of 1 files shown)


SpacePossum:
Thanks for reporting. It seems the README is indeed not correct, not for 1.x and not for master as well.
Could you change the code that generates the README on 1.x?
(around https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Console/Command/FixCommand.php#L192) 

and than regenerate the README;
```
php php-cs-fixer readme > README.rs
```
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/CONTRIBUTING.md)

keradus:
Why it's not correct?

gharlan:
@fitzel I guess you are using 1.11.x, so you should look to that branch: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/README.rst

1.12 is not yet released, but there are new classes for config and finder, and the old ones are deprecated.
So the 1.12 readme is correct for 1.12.

GrahamCampbell:
Yeh, this PR isn't right.

keradus:
Thank you @fitzel, please follow @gharlan solution ;)

fitzel:
yep, thats my bad, sry guys! shouldve seen the version diff

PR #1783, opened 2016-02-17
Add NoWhileAsForLoopFixer.

SpacePossum:
Expected
```php
                <?php
                    while($a > 1) {
                        echo $a;
                    }
```

Input

```php
                <?php
                    for(;$a > 1;) {
                        echo $a;
                    }
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -662,6 +662,10 @@ Choose from the list of available fixers:
                         return statement at the end of
                         a function.
 
+* **no_while_as_for_loop** [contrib]
+                        Simple for-loops should be
+                        written as while-loops.
+
 * **ordered_use** [contrib]
                         Ordering use statements.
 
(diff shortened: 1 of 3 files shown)


gharlan:
`for (;;)` is valid, but `while ()` is not.

SpacePossum:
good catch :+1: 
updated in https://github.com/SpacePossum/PHP-CS-Fixer/commit/cf9ed85437eac7c4215419e7b0d20f336018ea59

Expected
```php
<?php
    while(true) {
        echo 1;
        break;
    }
```

Input
```php
<?php
    for(;;) {
        echo 1;
        break;
    }
```

SpacePossum:
maybe to rare of case to the code base?

gharlan:
> maybe to rare of case to the code base?

I wanted to add my ":+1:" just this moment. But maybe you're right.

gharlan:
And btw: thanks for all your work!!
But I think we should not add fixers with very rare use cases.

SpacePossum:
>And btw: thanks for all your work!!

Thanks! :)
I'll create a new public MIT repo with some rare fixers, just in case someone finds a need for those one day (other than me ^-^).

Thanks for reviewing!

PR #1784, opened 2016-02-17
Add NoUselessContinueCountFixer.

SpacePossum:
Expected
```php
    <?php
        while(true) {
            break  ;
        }
        while(true) {
            continue ;
        }
```

Input

```php
    <?php
        while(true) {
            break  1;
        }
        while(true) {
            continue 0;
        }
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -657,6 +657,10 @@ Choose from the list of available fixers:
                         before a namespace
                         declaration.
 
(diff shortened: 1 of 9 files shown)


gharlan:
Is this something you've seen in real code bases?

SpacePossum:
yes, however, may it to rare after all

PR #1787, opened 2016-02-18
[WIP] Add CombineConsecutiveImportsFixer.

SpacePossum:
Expected
```php
<?php
    global $a, $b;
```

Input
```php
<?php
    global $a;
   global $b;
```

TODO: priority checks and integration tests

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -609,6 +609,11 @@ Choose from the list of available fixers:
                         Align equals symbols in
                         consecutive lines.
 
+* **combine_consecutive_imports** [contrib]
+                        Calling import on multiple
+                        variables should be done in
+                        one call on one line.
+
 * **concat_with_spaces** [contrib]
                         Concatenation should be used
                         with at least one whitespace
--- a/Symfony/CS/Fixer/Contrib/CombineConsecutiveImportsFixer.php
+++ b/Symfony/CS/Fixer/Contrib/CombineConsecutiveImportsFixer.php
@@ -1,0 +1,127 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class CombineConsecutiveImportsFixer extends AbstractFixer
+{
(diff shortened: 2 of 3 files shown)


keradus:
:-1: for touching globals at all.

SpacePossum:
could you explain why?

keradus:
Globals should not be used at all. 
I don't like fixers that support bad practices.

GrahamCampbell:
I agree with @keradus.

SpacePossum:
What about short tags, we support those?

I would like it in, because I deal with a lot of legacy code which contains globals.
Anyway, the fixer is at contrib so I see no issue, however if no one else see a use in than lets drop it.

keradus:
short tags fixer is different. short tags fixer say "ok, so I will recognize your ugly short tag and fix it with full tag".

fixer in this PR is "ok, so I will recognize your ugly globals and keep them ugly"

SpacePossum:
>"ok, so I will recognize your ugly globals and keep them ugly"

"ok, so I will recognize your ugly globals and make them less ugly."

I'm not writting globals (here, and the fixer isn't neither), I just have to deal with those. The fixer is part of Friends Of PHP, globals are part of PHP. I really don't see why you are stonewalling this on contrib level,

keradus:
And then some less experienced developer, after having fun with globals and asked to stop doing it will say "but I saw that other use them too, there are even rules how you should write them!".
Really, I had situations like this few times.

You want to introduce coding style rule "combine your globals", while rule should be "don't use globals".

------------------------------------------------------------------------------------------------------------------------------------------

In the end remember that this tool is driven by community. There were few situations where I, personally, didn't like the PR but community does so we merge it. If the community will want it we will have it. For now there are just 3 opinions.

GrahamCampbell:
We should make some troll fixer for april 1st that just delete all global variables, and all variables that are snake_case. :P

SpacePossum:
Alright, I like to ask the community about 2 other rules as well before PR-ing it.

First;
Expected
```php
<?php
global $a, $b;
```
Input
```php
<?php
global $a,      $b   ;
```

Second;
Expected
```php
function test()
{
    global $t, $t1, $t2;

    return $t + $t1 + $t2;
}
```

Input:
```php
function test()
{
    global $t, $t1;
    return $t + $t1 + $t2;
}
```

keradus:
first one: 0
second one: :-1: (cmon, `$t2` is not global, don't create new one!)

SpacePossum:
update the case, I did not meant to create the new global; the fix is about having an empty line after the global at the start of a function

Slamdunk:
:-1:

gharlan:
:-1: for adding fixers related to `global`.

SpacePossum:
thanks all for your input :+1:

Slamdunk:
@SpacePossum to waste your hard work, you can reuse this code for [`static`](http://php.net/manual/it/language.variables.scope.php#language.variables.scope.static), which is totally acceptable :wink:

keradus:
:+1: for static

PR #1792, opened 2016-02-21
Add PhpUnitDedicateAssertFixer.

SpacePossum:
Expected
```
<?php
$this->assertEmpty($a);
$this->assertNotEmpty($a);
$this->assertFileExists($a);
$this->assertFileNotExists($a);
$this->assertNotNull($a);
$this->assertNull($a);
$this->assertNan($a);
$this->assertInternalType('float', $a);
```

Input
```php
<?php
$this->assertTrue(empty($a));
$this->ASSERTFALSE(empty($a));
$this->assertTrue(file_exists($a));
$this->assertFalse(file_exists($a));
$this->assertTrue(is_null($a));
$this->assertFalse(is_null($a));
$this->assertTrue(is_nan($a));
$this->assertTrue(is_real( $a));
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -689,6 +689,14 @@ Choose from the list of available fixers:
                         This could change code
                         behavior.
 
+* **php_unit_dedicate_assert** [contrib]
+                        PHPUnit assertions like
+                        "assertInternalType",
+                        "assertFileExists", should be
+                        used over "assertTrue".
+                        Warning! This could change
+                        code behavior.
+
 * **php_unit_strict** [contrib]
                         PHPUnit methods like
                         "assertSame" should be used
--- a/Symfony/CS/Fixer/Contrib/PhpUnitConstructFixer.php
+++ b/Symfony/CS/Fixer/Contrib/PhpUnitConstructFixer.php
@@ -35,6 +35,9 @@ final class PhpUnitConstructFixer extends AbstractFixer
         'assertNotSame' => 'fixAssertNegative',
     );
 
+    /**
+     * @param array<string, bool> $usingMethods
+     */
     public function configure(array $usingMethods)
     {
         foreach ($usingMethods as $method => $fix) {
@@ -90,10 +93,17 @@ final class PhpUnitConstructFixer extends AbstractFixer
      */
     public function getPriority()
     {
-        // should be run after the PhpUnitStrictFixer
+        // should be run after the PhpUnitStrictFixer and before PhpUnitDedicateAssertFixer.
         return -10;
     }
 
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     * @param string $method
+     *
+     * @return int|null
+     */
     private function fixAssertNegative(Tokens $tokens, $index, $method)
     {
         static $map = array(
@@ -105,6 +115,13 @@ final class PhpUnitConstructFixer extends AbstractFixer
         return $this->fixAssert($map, $tokens, $index, $method);
     }
 
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     * @param string $method
+     *
+     * @return int|null
+     */
     private function fixAssertPositive(Tokens $tokens, $index, $method)
     {
         static $map = array(
@@ -116,6 +133,14 @@ final class PhpUnitConstructFixer extends AbstractFixer
         return $this->fixAssert($map, $tokens, $index, $method);
     }
 
+    /**
+     * @param array<string, string> $map
+     * @param Tokens                $tokens
+     * @param int                   $index
+     * @param string                $method
+     *
+     * @return int|null
+     */
     private function fixAssert(array $map, Tokens $tokens, $index, $method)
     {
         $sequence = $tokens->findSequence(
--- a/Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
+++ b/Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
@@ -71,12 +71,16 @@ final class PhpUnitDedicateAssertFixer extends AbstractFixer
     );
 
     /**
-     * @param string[] $usingMethods
+     * @param array|null $configuration
      */
-    public function configure(array $usingMethods)
+    public function configure(array $configuration = null)
     {
+        if (null === $configuration) {
+            throw new InvalidFixerConfigurationException($this->getName(), 'Configuration is required.');
+        }
+
         $this->configuration = array();
-        foreach ($usingMethods as $method) {
+        foreach ($configuration as $method) {
             if (!array_key_exists($method, $this->fixMap)) {
                 throw new InvalidFixerConfigurationException($this->getName(), sprintf('Unknown configuration method "%s".', $method));
             }
--- a/Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
+++ b/Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
@@ -1,0 +1,260 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\ConfigurationException\InvalidFixerConfigurationException;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class PhpUnitDedicateAssertFixer extends AbstractFixer
+{
+    private $fixMap = array(
+        'array_key_exists' => array('assertArrayNotHasKey', 'assertArrayHasKey'),
+        'empty' => array('assertNotEmpty', 'assertEmpty'),
+        'file_exists' => array('assertFileNotExists', 'assertFileExists'),
+        'is_infinite' => array('assertFinite', 'assertInfinite'),
+        'is_nan' => array(false, 'assertNan'),
+        'is_null' => array('assertNotNull', 'assertNull'),
+        'is_array' => true,
+        'is_bool' => true,
+        'is_boolean' => true,
+        'is_callable' => true,
+        'is_double' => true,
+        'is_float' => true,
+        'is_int' => true,
+        'is_integer' => true,
+        'is_long' => true,
+        'is_​numeric' => true,
+        'is_object' => true,
+        'is_real' => true,
+        'is_​resource' => true,
+        'is_scalar' => true,
+        'is_string' => true,
+    );
+
+    private $configuration = array(
+        'array_key_exists',
+        'empty',
+        'file_exists',
+        'is_infinite',
+        'is_nan',
+        'is_null',
+        'is_array',
+        'is_bool',
+        'is_boolean',
+        'is_callable',
+        'is_double',
+        'is_float',
+        'is_int',
+        'is_integer',
+        'is_long',
+        'is_​numeric',
+        'is_object',
+        'is_real',
+        'is_​resource',
+        'is_scalar',
+        'is_string',
+    );
+
+    /**
+     * @param string[] $usingMethods
+     */
+    public function configure(array $usingMethods)
+    {
+        $this->configuration = array();
+        foreach ($usingMethods as $method) {
+            if (!array_key_exists($method, $this->fixMap)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Unknown configuration method "%s".', $method));
+            }
+
+            $this->configuration[] = $method;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        static $searchSequence = array(
+            array(T_VARIABLE, '$this'),
+            array(T_OBJECT_OPERATOR, '->'),
+            array(T_STRING),
+        );
+
+        $index = 1;
+        $candidate = $tokens->findSequence($searchSequence, $index);
+        while (null !== $candidate) {
+            end($candidate);
+            $index = $this->getAssertCandidate($tokens, key($candidate));
+            if (is_array($index)) {
+                $index = $this->fixAssert($tokens, $index);
+            }
+
+            ++$index;
+            $candidate = $tokens->findSequence($searchSequence, $index);
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'PHPUnit assertions like "assertInternalType", "assertFileExists", should be used over "assertTrue". Warning! This could change code behavior.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run after the PhpUnitConstructFixer.
+        return -15;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $assertCallIndex Token index of assert method call.
+     *
+     * @return int|int[] indexes of assert call, test call and positive flag, or last index checked.
+     */
+    private function getAssertCandidate(Tokens $tokens, $assertCallIndex)
+    {
+        $content = strtolower($tokens[$assertCallIndex]->getContent());
+        if ('asserttrue' === $content) {
+            $isPositive = 1;
+        } elseif ('assertfalse' === $content) {
+            $isPositive = 0;
+        } else {
+            return $assertCallIndex;
+        }
+
+        // test candidate for simple calls like: ([\]+'some fixable call'(...))
+        $assertCallOpenIndex = $tokens->getNextMeaningfulToken($assertCallIndex);
+        if (!$tokens[$assertCallOpenIndex]->equals('(')) {
+            return $assertCallIndex;
+        }
+
+        $testDefaultNamespaceTokenIndex = false;
+        $testIndex = $tokens->getNextMeaningfulToken($assertCallOpenIndex);
+
+        if (!$tokens[$testIndex]->isGivenKind(array(T_EMPTY, T_STRING))) {
+            if (!$tokens[$testIndex]->isGivenKind(T_NS_SEPARATOR)) {
+                return $testIndex;
+            }
+
+            $testDefaultNamespaceTokenIndex = $testIndex;
+            $testIndex = $tokens->getNextMeaningfulToken($testIndex);
+        }
+
+        $testOpenIndex = $tokens->getNextMeaningfulToken($testIndex);
+        if (!$tokens[$testOpenIndex]->equals('(')) {
+            return $testOpenIndex;
+        }
+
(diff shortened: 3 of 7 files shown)


SpacePossum:
SF acceptance PR made; https://github.com/symfony/symfony/pull/17872

Symfony/CS/Fixer/Contrib/PhpUnitConstructFixer.php
```diff
@@ -35,6 +35,9 @@
         'assertNotSame' => 'fixAssertNegative',
     );
 
+    /**
+     * @param array<string, bool> $usingMethods
```
GrahamCampbell: we can't use phpdoc of this format because it's not supported by php-cs-fixer, no is it widely accepted as valid syntax

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
```
GrahamCampbell: invalid format

Symfony/CS/Fixer/Contrib/PhpUnitConstructFixer.php
SpacePossum: with the exception of `Token.php` you mean?

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
SpacePossum: https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#72-author
`This tag MAY also contain an e-mail address.`

We had this discussion before, I don't see what is changed in the mean time that justifies having it again?

GrahamCampbell: The author is `SpacePossum <possumfromspace@gmail.com>`, as shown by git.

SpacePossum: if `git` is a source for meta data, than the tag is over complete and can be removed

GrahamCampbell: git doesn't handler file moves correctly though, so the author tag is still required

keradus: Guys, you had this discussion few times. If @SpacePossum don't want to expose his mail in codebase it's fine.

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+        $tokens->clearTokenAndMergeSurroundingWhitespace($openIndex);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($closeIndex);
+    }
+
+    /**
+     * @return array Token sequence
+     */
+    private function getSearchSequence()
+    {
+        static $sequence;
```
keradus: can't it's be declared `static $foo = [];` ?
I can't see the need for lazy initialization here

keradus: even better, no need for getter ;)

keradus:
What about assertInstanceOf?

SpacePossum:
>What about assertInstanceOf?

What about it?

README.rst
```diff
@@ -680,6 +680,14 @@ Choose from the list of available fixers:
                         This could change code
                         behavior.
 
+* **php_unit_no_dedicate_assert** [contrib]
```
keradus: I think `no` shouldn't be here

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        $searchSequence = $this->getSearchSequence();
+
+        $index = 1;
+        while (null !== $candidate = $tokens->findSequence($searchSequence, $index)) {
```
keradus: do not assign inside loop

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+        return -15;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $assertCallIndex Token index of assert method call.
+     *
+     * @return int|int[] indexes of assert call, test call and positive flag, or last index checked.
+     */
+    private function isAssertCandidate(Tokens $tokens, $assertCallIndex)
```
keradus: why `isFoo` returns non-bool value ?

SpacePossum:
TODO list:
`assertInfinite`
`assertInstanceOf`
`assertArrayHasKey`

there are more that could be done, but those would be very complex so I rather leave it with this set.

keradus:
Oh, now I see your point. I was thinking more about:
```diff
-$this->assertSame($a, get_class($b));
+$this->assertInstanceOf($a, $b);
```

but maybe it's out of scope, what do you think?

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        static $searchSequence;
+        if (null === $searchSequence) {
+            $searchSequence = array(
```
gharlan: could be simplified to `static $searchSequence = array(...`

keradus: actually even no great value of having it static

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
gharlan: :-1: In my opinion the assignment inside `while()` was more readable. 
Now we have the `findSequence` call twice.

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,216 @@
...
+        // test candidate for simple calls like: ([\]+'some fixable call'(...))
+        $assertCallOpen = $tokens->getNextMeaningfulToken($assertCallIndex);
+        if (!$tokens[$assertCallOpen]->equals('(')) {
+            return $assertCallIndex;
+        }
+
+        $testNsToken = false;
+        $test = $tokens->getNextMeaningfulToken($assertCallOpen);
+
+        if (!$tokens[$test]->isGivenKind(T_EMPTY) && !$tokens[$test]->isGivenKind(T_STRING)) {
```
gharlan: `!$tokens[$test]->isGivenKind(array(T_EMPTY, T_STRING))`

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,216 @@
...
+            if (!$tokens[$test]->isGivenKind(T_NS_SEPARATOR) || '\\' !== $tokens[$test]->getContent()) {
```
gharlan: `T_NS_SEPARATOR` is always `\`, isn't it?

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,216 @@
...
+
+        $testOpen = $tokens->getNextMeaningfulToken($test);
+        if (!$tokens[$testOpen]->equals('(')) {
+            return $testOpen;
+        }
+
+        $testClose = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $testOpen);
+
+        $assertCallClose = $tokens->getNextMeaningfulToken($testClose);
+        if (!$tokens[$assertCallClose]->equals(')') && !$tokens[$assertCallClose]->equals(',')) {
```
gharlan: `!equalsAny`

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,216 @@
...
+    {
+        list(
+            $isPositive,
+            $assertCallIndex,
+            $assertCallOpen,
+            $testNsToken,
+            $test,
+            $testOpen,
+            $testClose,
+            $assertCallClose) = $assertIndexes;
```
gharlan: IMO `)` should go to next line.

SpacePossum: thanks, also for all other great pointers! I've made changes following your suggestions :)

SpacePossum:
@keradus I'm going to try some stuff out, but to be honest I don't think I can work on the instance-cases. It feels a bit odd to draw a line somewhere, but it is about time mostly.
For example I would like to write a fixer to do;
expected
```php
$this->assertSame($a, is_array($b));
```

input
```php
$this->assertTrue($a === is_array($b));
```

etc. Its just its going to take more time than I've got.

keradus:
Yeah, for sure it's not worth to have the fixer over-complicated.
It was just "nice to have" ;)

keradus:
It's Symfony and not.
No risky fixer should be run by default.
Please exclude it from Symfony level to avoid that.

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,214 @@
...
+
+        switch ($content) {
+            case 'is_real':
+                $content = 'is_float';
+                // do not break here
+            case 'is_array':
+            case 'is_bool':
+            case 'is_float':
+            case 'is_int':
+            case 'is_string':
```
keradus: What about
```
is_​callable
is_​double
is_​integer
is_​long
is_​null
is_​numeric
is_​object
is_​resource
is_​scalar
```
?

Symfony/CS/Tests/Fixer/Symfony/PhpUnitDedicateAssertFixerTest.php
```diff
@@ -0,0 +1,180 @@
...
+            );
+
+            $cases[] = array(
+                sprintf('<?php $this->assertNotInternalType(\'%s\', $a);', $type),
+                sprintf('<?php $this->assertFalse(is_%s($a));', $type),
+            );
+        }
+
+        $cases[] = array(
+            '<?php $this->assertInternalType(\'float\', $a, "my message");',
```
keradus: Question: would internalType 'real' won't work?

SpacePossum: It would, I guess this is a left over from when it was a custom fixer.
This fixer should ran after `no_alias_functions` (on master only atm) to allow that fixer to rename `is_real` to `is_float`.

SpacePossum:
I might be good to have config options on this one, I've added some PHPUnit 5 only asserts.

Symfony/CS/Fixer/Symfony/PhpUnitDedicateAssertFixer.php
SpacePossum: nice list :)
I'll work on it (note to self: [PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/5.3/src/Framework/Constraint/IsType.php#L36]))

Symfony/CS/Tests/Fixer/Symfony/PhpUnitDedicateAssertFixerTest.php
keradus: I think it don't need to run after it, just handle the aliased function here.
We combine fixers in correct order to do sth that is not possible by one fixer itself. Here it's not that case.

SpacePossum: To be sure I understand correctly, the current mapping of the fixer for;
expected
```php
            <?php $this->assertInternalType('float', $a, "my message");'
```
input
```php
            <?php $this->assertTrue(is_real( $a), "my message");'
```

should be changed to expected:
```php
            <?php $this->assertInternalType('real', $a, "my message");'
```

keradus: yes.
priority still could be set to have it float, but it's cool to have the fixer working even for aliases.

Symfony/CS/Fixer/Contrib/PhpUnitNoDedicateAssertFixer.php
SpacePossum: @gharlan I agree. 
Assignments in `for`-loops and/or calling iterators to do internal (re-)assignments or calling other method/function who do something similar are all considered OK; however not inline. Anyway it is a conversion and I'm willing to follow it if that is what gets the fixer merged ;)

Symfony/CS/Tests/Fixer/Symfony/PhpUnitDedicateAssertFixerTest.php
SpacePossum: :+1:  thanks for clearing it up :)  I'll see what I can do to make the merge more easy on master (considering priorities regarding the `no_alias_functions` fixer).

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+        if (array_key_exists($content, $this->fixMap)) {
+            if (false !== $this->fixMap[$content][$isPositive]) {
+                $tokens[$assertCallIndex]->setContent($this->fixMap[$content][$isPositive]);
+                $this->removeFunctionCall($tokens, $testNsToken, $test, $testOpen, $testClose);
+            }
+
+            return $assertCallClose;
+        }
+
+        switch ($content) {
```
keradus: switch is a bit pointless now

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,221 @@
...
+            return $assertCallIndex;
+        }
+
+        // test candidate for simple calls like: ([\]+'some fixable call'(...))
+        $assertCallOpen = $tokens->getNextMeaningfulToken($assertCallIndex);
+        if (!$tokens[$assertCallOpen]->equals('(')) {
+            return $assertCallIndex;
+        }
+
+        $testNsToken = false;
```
keradus: ns?
is it a token or index ?

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
SpacePossum: Would you prefer a multi line if-statement or an `in_array` type check?

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
SpacePossum: NS = namespace, I'll refactor.

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
keradus: yes.
also, what if `content` doesn't match that cases?

SpacePossum: >yes

not sure I follow, which one would you prefer, `multi line if` or `in_array`? 

>also, what if content doesn't match that cases?

`content` is made lower case on line 160 so this should be no issue :)

keradus: `in_array`.

> > also, what if content doesn't match that cases?
> 
> content is made lower case on line 160 so this should be no issue :)

I didn't ask about casing but the cases you have listed in switch - as there is no default case.

SpacePossum: I see, all clear to me now. I'll update as suggested :)

SpacePossum:
Processed feedback and added configuration options. Happy to take more notes and ideas about the default config.

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
```diff
@@ -0,0 +1,260 @@
...
+        'is_real',
+        'is_​resource',
+        'is_scalar',
+        'is_string',
+    );
+
+    /**
+     * @param string[] $usingMethods
+     */
+    public function configure(array $usingMethods)
```
keradus: isn't the interface `array|null` ?

Symfony/CS/Tests/Fixer/Contrib/PhpUnitDedicateAssertFixerTest.php
```diff
@@ -0,0 +1,206 @@
...
+                    $this->assertTrue(is_infinite($a));
+            ',
+            null,
+            $fixer
+        );
+    }
+
+    /**
+     * @expectedException \Symfony\CS\ConfigurationException\InvalidFixerConfigurationException
+     * @expectedExceptionMessage [php_unit_dedicate_assert] Unknown configuration method "_unknown_".
```
keradus: Please always use `@expectedExceptionMessageRegExp`.
`@expectedExceptionMessage foo`
means
`@expectedExceptionMessageRegExp foo`
which is
`@expectedExceptionMessageRegExp ^.*foo.*$`
and not what we want, that is:
`@expectedExceptionMessageRegExp ^foo$`

keradus:
To be done on 2.x line (after this PR will be merged)
- Fixer must run after `no_alias_functions`

keradus:
Please fix 2 minors and :+1:

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
SpacePossum: What does `null` mean in this context, implicit set to default?

SpacePossum:
done, lets see how the tests do :)

Symfony/CS/Fixer/Contrib/PhpUnitDedicateAssertFixer.php
keradus: yes

keradus: maybe we could reconsider if the value should be nullable or not, but for now please be compatible with 2.x ;)

keradus:
Thank you @SpacePossum.

PR #1798, opened 2016-02-24
LintManager - meaningful tempnam

Slamdunk:
Killing the process manually leave the temporary file on the filesystem, as [`__destruct`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/LintManager.php#L31-L36) is not called.

In doubt, a meaningful name for the temporary file helps identifying them and manually getting rid of them if needed.

### DIFF
--- a/Symfony/CS/LintManager.php
+++ b/Symfony/CS/LintManager.php
@@ -66,7 +66,7 @@ class LintManager
     public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
-            $this->temporaryFile = tempnam('.', 'tmp');
+            $this->temporaryFile = tempnam('.', 'phpcsfixer_tmp_');
         }
 
         file_put_contents($this->temporaryFile, $source);
--- a/Symfony/CS/LintManager.php
+++ b/Symfony/CS/LintManager.php
@@ -66,7 +66,7 @@ class LintManager
     public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
-            $this->temporaryFile = tempnam('.', 'phpcsfixer_tmp_');
+            $this->temporaryFile = tempnam('.', 'cs_fix_tmp_');
         }
 
         file_put_contents($this->temporaryFile, $source);
--- a/Symfony/CS/LintManager.php
+++ b/Symfony/CS/LintManager.php
@@ -66,7 +66,7 @@ class LintManager
     public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
-            $this->temporaryFile = tempnam('.', 'cs_fix_tmp_');
+            $this->temporaryFile = tempnam('.', 'cs_fixer_tmp_');
         }
 
         file_put_contents($this->temporaryFile, $source);


Symfony/CS/LintManager.php
```diff
@@ -66,7 +66,7 @@ public function createProcessForFile($path)
     public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
-            $this->temporaryFile = tempnam('.', 'tmp');
+            $this->temporaryFile = tempnam('.', 'phpcsfixer_tmp_');
```
keradus: After this prefix on Windows would be `php`, which is even less meaning.

GrahamCampbell:
:-1:

Slamdunk:
Ok this is not a big deal, but let me show you what happens to me.
`tmp` is too often used everywhere for a temporary file, and in my business we ended up having tons of `tmp` files not cleaned after their use, and we often get stuck in the doubt if we can remove them or not. This happend also with this tool.

I have no preference on what the prefix should be, but we have a bit difficult times when we find untracked orphan `tmp*` files.

May we come to something slightly more verbose than just `tmp`?

keradus:
> having tons of tmp files not cleaned after their use, and we often get stuck in the doubt if we can remove them or not. 

aren't tmp files safe to remove, just from their nature?

Slamdunk:
@keradus this is true only if you consider all the PHP processes as terminated. In our business we use PHP even for long-running scripts.

keradus:
Then you have the files locked. But indeed, releasing could be messy if process was interrupted.

Slamdunk:
I'll let the community decide

SpacePossum:
+1, more clear and I see no drawback from the change

Symfony/CS/LintManager.php
```diff
@@ -66,7 +66,7 @@ public function createProcessForFile($path)
     public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
-            $this->temporaryFile = tempnam('.', 'tmp');
+            $this->temporaryFile = tempnam('.', 'cs_fix_tmp_');
```
keradus: if we want to have a name more verbose why not `cs_fixer_tmp_` ?

SpacePossum: :+1: even beter :)

SpacePossum:
@Slamdunk please squash your commits, I think it is RTM after that.

keradus:
:+1:

keradus:
@Slamdunk please don't squash your commits. It's not needed - if PR has only one autor then commits are automatically squashed during merging.
(Of course if you really want you can squash it on your side, but that's waste of time)

SpacePossum:
Sorry about that, force of habbit from work I guess.

:+1:

keradus:
Teach the people at work that they can squash commits during merging, it's not hard ;)

GrahamCampbell:
> Teach the people at work that they can squash commits during merging, it's not hard ;)

It would be nice if GitHub could actually support that during merging.

keradus:
Thank you @Slamdunk.

Issue #1801, opened 2016-02-25
Feature request to add a fixer for ordering use statements by length


kabirbaidhya:
I think it would be a nice to have a new fixer `ordered_use_by_length` in the package that organizes the `use` statements in the code by it's length to make the code look cleaner. 

Suppose we have this block of code: 

``` php
use Agent\Entity\Organization;
use Agent\Entity\Person;
use Auth\Entity\User;
use Crm\Creator\LeadCreator;
use Crm\Entity\CommercialRealEstateSpaceRequirement;
use Crm\Entity\Lead;
use Crm\Entity\LeadRole;
use Crm\Query\AvailableLeadStatusQuery;
use Crm\Query\CummingsUsersQuery;
use Crm\Query\Lead\ArchivedDetailsQuery;
```

And when applied it would change it to:

``` php
use Crm\Entity\Lead;
use Auth\Entity\User;
use Agent\Entity\Person;
use Crm\Entity\LeadRole;
use Crm\Creator\LeadCreator;
use Agent\Entity\Organization;
use Crm\Query\CummingsUsersQuery;
use Crm\Query\AvailableLeadStatusQuery;
use Crm\Query\Lead\ArchivedDetailsQuery;
use Crm\Entity\CommercialRealEstateSpaceRequirement;
```

I'm aware that there is already a fixer named `ordered_use` in `php-cs-fixer` but that orders the statements alphabetically. So, it's different as this is to **organize it by length**. 

I would suggest this new fixer to be named as `ordered_use_by_length` or `organize_use_by_length`. 

I know, not everybody would agree with this as many people rather stick with just ordering it alphabetically. But IMHO I think it would be nice to have this new feature as in big classes, it reduces code readability and cleanliness if we only organize it alphabetically. In our codebase we are doing it manually. 

Anyway, if you guys agree I can submit a  PR.

SpacePossum:
Hi,
I wouldn't use the ordering you propose, however there have been request to change the order-logic of this fixer before. So if this was to be done, maybe we could open up the fixer a bit. Allowing people to write there own sorter by allowing to extend the class or allow to configure a `sorter`-object (just typing some ideas)? I'm not in fav. of having a new fixer for this case.

keradus:
I disagree with reasoning. Ordering alphabetically increase readability as it groups by vendors, then packages, then modules and so on.

Question here: does anyone else need to sort `use` statements by length?

[...]

PR #1803, opened 2016-03-03
ordered_use_by_length fixer added

pbkyriak:


keradus:
Please see #1801

GrahamCampbell:
The following test case will fail:

```php
<?php

use Acme;
use Foo;
use Bar;
```

The output should be:

```php
<?php

use Bar;
use Foo;
use Acme;
```

GrahamCampbell:
I think. We should add it all the same. ;)

pbkyriak:
sorry missed that spot. can i try do the sorting option of the ordered_use fixer?

keradus:
Indeed, this should not be new fixer but option for existing one.

GrahamCampbell:
@keradus This should be closed then?

GrahamCampbell:
Any news here?

keradus:
Closing due to wrong approach and lack of authors interest

powelski:
@pbkyriak I'm cheering you up, this awesome feature is highly awaited :)

PR #1806, opened 2016-03-07
SelfAccessorFixer - skip anonymous classes

gharlan:


Symfony/CS/Tokenizer/Tokens.php
```diff
@@ -929,6 +929,28 @@ public function isArrayMultiLine($index)
...
+     * @param int $index
+     *
+     * @return bool
+     */
+    public function isAnonymousClass($index)
+    {
+        $token = $this[$index];
+
+        if (!$token->isClassy()) {
+            throw new \LogicException('No classy token at given index');
```
SpacePossum: why throw an exception for non-classy but return false for `trait` / `interface` ?

gharlan: > why throw an exception for non-classy

to be congruent to `isLambda` which throws an exception for token unequal to `T_FUNCTION`.
(I don't like the behavior. In my opinion it should return `false`.)

> but return false for trait / interface ?

to be pragmatic and allow code like this:

```php
if ($token->isClassy() && !$tokens->isAnonymousClass($index)) {
}
```
example: https://github.com/gharlan/PHP-CS-Fixer/blob/35e69bd27be0bc4c9c62b9c1efb6a663519a3c10/Symfony/CS/Fixer/Symfony/SelfAccessorFixer.php#L30

SpacePossum: I would like the `false` return better, but I follow your reasoning here :)

gharlan: I would vote :+1: for removing the exception in this pr (in `isAnonymousClass`), and removing it in `isLambda` in 2.x.

SpacePossum:
:+1: 

(having a test for something like `$a = new class(new class($d->a) implements B{}) extends C{};` would be a nice to have, but I'm sure it passes ;)  )

gharlan:
> `$a = new class(new class($d->a) implements B{}) extends C{};`

done. and added another test case.

Symfony/CS/Tokenizer/Tokens.php
keradus: I'm strongly against returning multiple types on output like `int|bool`

gharlan: hmm I think you've misread something. Who wants to return something like `int|bool`?

keradus: oh, bool or bool :D never mind ;)

gharlan: we want to return `false` instead of throwing an exception. So this should be possible without checking token kind before:

```php
if ($tokens->isAnonymousClass($anyTokenIndex) {}
```

gharlan: same fore `isLambda` (on 2.x for bc reasons)

Symfony/CS/Tests/Fixer/Symfony/SelfAccessorFixerTest.php
```diff
@@ -78,6 +78,9 @@ public function provideExamples()
                 // PHP < 5.4 compatibility: "self" is not available in closures
                 '<?php class Foo { function bar() { function ($a = Foo::BAZ) { new Foo(); }; } }',
             ),
+            array(
+                '<?php class Foo { function bar() { new class() { new Foo(); }; } }',
```
keradus: invalid code:
https://3v4l.org/Z3TRV

gharlan: good point. missing the linter (activated on 1.12) ;)

Symfony/CS/Tokenizer/Tokens.php
keradus: `isLambda` works like:
OK, I have a function and I want to check is it a lambda. So calling it on non-function token is providing an invalid argument, so an exception is rised.
So it's like `isThatFunctionALamba` but not `isThatTokenAFunctionThatIsLambdaFunctionAsWell`.

So here, to follow same principle, we should not allow to run method on interface / trait

gharlan: I don't see any benefit in it.
The name is `$tokens->isLambda($index)`. "Is under given index a lambda function?"

> So here, to follow same principle

I would change the principle.

But even when not changing the principle, I would be pragmatic regarding interface/trait. 
Then I would apply the principle this way:
"OK, I have a classy element and I want to check is it a anonymous class. So calling it on non-classy token is providing an invalid argument, so an exception is rised."

GrahamCampbell:
:+1:

GrahamCampbell:
Is this good to merge?

keradus:
Yes, @gharlan please, could you rebase?
I will try to merge few PRs this weekend

gharlan:
done

GrahamCampbell:
I'm inclined to say this is more of a feature and belongs in 1.12.

keradus:
Thank you @gharlan.

Issue #1808, opened 2016-03-07
Semicolon on same line fixer bug


GrahamCampbell:
Minimal test case:

INPUT:
```php
<?php

$foo->bar() // test
;
```

OUTPUT:

```php
<?php

$foo->bar() // test;
```

EXPECTED:

```php
<?php

$foo->bar(); // test
```

---

Issue present on latest master, and I assume on at least 1.12.x too.

GrahamCampbell:
:+1:

PR #1809, opened 2016-03-08
NoMultilineWhitespaceBeforeSemicolonsFixer - Semicolon should not be moved into comment

SpacePossum:
Bug on master only. 
The fixer is broken because of the transformer that splits up a comment with a trailing line break.

Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1808

### DIFF
--- a/src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php
+++ b/src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php
@@ -39,8 +39,13 @@ final class NoMultilineWhitespaceBeforeSemicolonsFixer extends AbstractFixer
             }
 
             $previous = $tokens[$index - 1];
+            if (!$previous->isWhitespace() || false === strpos($previous->getContent(), "\n")) {
+                continue;
+            }
 
-            if ($previous->isWhitespace() && false !== strpos($previous->getContent(), "\n")) {
+            if (("\n" === $previous->getContent()[0] || "\r" === $previous->getContent()[0]) && $tokens[$index - 2]->isComment()) {
+                $previous->setContent("\r" === $previous->getContent()[0] ? "\r\n" : "\n");
+            } else {
                 $previous->clear();
             }
         }
--- a/src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php
+++ b/src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php
@@ -43,8 +43,9 @@ final class NoMultilineWhitespaceBeforeSemicolonsFixer extends AbstractFixer
                 continue;
             }
 
-            if (("\n" === $previous->getContent()[0] || "\r" === $previous->getContent()[0]) && $tokens[$index - 2]->isComment()) {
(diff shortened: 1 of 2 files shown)


keradus:
Please fix CS

SpacePossum:
done :)

src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php
```diff
@@ -39,8 +39,13 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
             }
 
             $previous = $tokens[$index - 1];
+            if (!$previous->isWhitespace() || false === strpos($previous->getContent(), "\n")) {
+                continue;
+            }
 
-            if ($previous->isWhitespace() && false !== strpos($previous->getContent(), "\n")) {
+            if (("\n" === $previous->getContent()[0] || "\r" === $previous->getContent()[0]) && $tokens[$index - 2]->isComment()) {
+                $previous->setContent("\r" === $previous->getContent()[0] ? "\r\n" : "\n");
```
GrahamCampbell: php 5.3 doesn't support array access on the result of a function call

GrahamCampbell: you'll need to assign it to variable first

keradus: Or just call `substr`

GrahamCampbell: with long array syntax, and a true as the 3rd param :trollface:

keradus: I didn't want to write the code for him ;P

SpacePossum: lol, no :P

keradus:
@SpacePossum ?

SpacePossum:
IMHO this fixer shouldn't be on master in the first place. Using `spaces_before_semicolon` and `extra_empty_lines` will give better results. So I lost my interest beyond the point going through another round of reviews.

keradus:
Only concerns here were to not break fabpot / PHP 5.3. So nothing big.
So you are saying we don't need this fixer at all?

SpacePossum:
I think this fixer should be merged with `spaces_after_semicolon`, possible with some config options if one really cares, however I don't have the time.
Test are green now :)

PS. This stops the fixer from breaking stuff, however it would be nice to be tweaked at some point to do;
expected
```php
<?php
echo $a; // test
```
input
```php
<?php
echo $a // test

;
```

currently fixed to
```php
<?php
echo $a // test
;
```

keradus:
It's good enough then. But I don't think it will be done soon (combining with `spaces_after_semicolon`) as well.

I think it's ready to merge.

gharlan:
:+1:

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks all :)

PR #1810, opened 2016-03-10
Remove trailing whitespace from comments

simoheinonen:


SpacePossum:
Hi,
Are you sure the `no_trailing_whitespace_in_comment` isn't doing the same thing?

GrahamCampbell:
> Remove trailing whitespace from comments

This change doesn't work in multiline comments.

GrahamCampbell:
> Are you sure the no_trailing_whitespace_in_comment isn't doing the same thing?

Yes, this should be closed.

GrahamCampbell:
This change also cannot be allowed into this fixer anyway because it is not in PSR2, and this fixer is a strictly PSR-2 only fixer.

simoheinonen:
Hmm, yes that looks like what I'm after for. For some reason one comment wasn't fixed when I ran the fixer. I'll look more into it why that happened.

simoheinonen:
PSR2 2.3. Lines

> There MUST NOT be trailing whitespace at the end of non-blank lines.

GrahamCampbell:
PSR2 is referring to actual whitespace tokens. The whitespace at the end of comments is actually included as a comment, as it not treated as whitespace by PHP's parser.

SpacePossum:
thanks for reporting back @simoheinonen , good to hear you found the issue and a solution :)

PR #1811, opened 2016-03-10
Added NoSpacesInsideOfssetFixer

phansys:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -446,12 +446,11 @@ Choose from the list of available fixers:
                          or function name and the
                          opening parenthesis.
 
-* **no_spaces_inside_ofsset**
-                         There MUST NOT be a space
-                         after the opening offset
-                         square brace. There MUST NOT
-                         be a space before the closing
-                         offset square brace.
+* **no_spaces_inside_ofsset** [@Symfony]
+                         There MUST NOT be spaces
+                         between the offset square
+                         braces and its contained
+                         values.
 
 * **no_spaces_inside_parenthesis** [@PSR2, @Symfony]
                          There MUST NOT be a space
--- a/README.rst
+++ b/README.rst
@@ -446,6 +446,13 @@ Choose from the list of available fixers:
                          or function name and the
                          opening parenthesis.
 
+* **no_spaces_inside_ofsset**
+                         There MUST NOT be a space
+                         after the opening offset
+                         square brace. There MUST NOT
+                         be a space before the closing
+                         offset square brace.
+
 * **no_spaces_inside_parenthesis** [@PSR2, @Symfony]
                          There MUST NOT be a space
                          after the opening
--- a/src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
+++ b/src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
@@ -13,6 +13,7 @@
 namespace PhpCsFixer\Fixer\Whitespace;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
 /**
@@ -41,10 +42,10 @@ final class NoSpacesInsideOfssetFixer extends AbstractFixer
             $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE, $index);
 
             // remove space after opening `[`
-            $this->removeSpaceAroundToken($tokens, $index, 1);
+            $this->removeSpaceAroundToken($tokens[$index + 1]);
 
             // remove space before closing `]`
-            $this->removeSpaceAroundToken($tokens, $endIndex, -1);
+            $this->removeSpaceAroundToken($tokens[$endIndex - 1]);
         }
     }
 
@@ -57,20 +58,12 @@ final class NoSpacesInsideOfssetFixer extends AbstractFixer
     }
 
     /**
-     * Remove spaces on one side of the token at a given index.
+     * Remove spaces on one side of the token.
      *
-     * @param Tokens $tokens A collection of code tokens
-     * @param int    $index  The token index
-     * @param int    $offset The offset where to start looking for spaces
+     * @param Token $token
      */
-    private function removeSpaceAroundToken(Tokens $tokens, $index, $offset)
+    private function removeSpaceAroundToken(Token $token)
     {
-        if (!isset($tokens[$index + $offset])) {
-            return;
-        }
-
-        $token = $tokens[$index + $offset];
-
         if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
             $token->clear();
         }
--- a/src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
+++ b/src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
@@ -42,10 +42,10 @@ final class NoSpacesInsideOfssetFixer extends AbstractFixer
             $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE, $index);
 
             // remove space after opening `[`
-            $this->removeSpaceAroundToken($tokens[$index + 1]);
+            $this->removeWhitespaceToken($tokens[$index + 1]);
 
             // remove space before closing `]`
-            $this->removeSpaceAroundToken($tokens[$endIndex - 1]);
+            $this->removeWhitespaceToken($tokens[$endIndex - 1]);
         }
     }
 
@@ -58,13 +58,13 @@ final class NoSpacesInsideOfssetFixer extends AbstractFixer
     }
 
     /**
-     * Remove spaces on one side of the token.
+     * Removes the token if it is single line whitespace.
      *
      * @param Token $token
      */
-    private function removeSpaceAroundToken(Token $token)
+    private function removeWhitespaceToken(Token $token)
     {
-        if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
+        if ($token->isWhitespace(" \t") && false === strpos($token->getContent(), "\n")) {
             $token->clear();
         }
     }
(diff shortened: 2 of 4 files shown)


src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,94 @@
...
+
+namespace PhpCsFixer\Fixer\Whitespace;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Javier Spagnoletti <phansys@gmail.com>
+ *
+ * @internal
```
SpacePossum: no need to mark the fixer as internal, it is allowed to use the fixer elsewhere :)

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->equals('[')) {
+                continue;
+            }
+
+            $prevIndex = $tokens->getPrevMeaningfulToken($index);
+
+            // ignore parenthesis for T_ARRAY
+            if (null !== $prevIndex && $tokens[$prevIndex]->isGivenKind(T_ARRAY)) {
```
SpacePossum: I think it is impossible for the `$prevIndex` to be `null` here

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,94 @@
...
+        if (!isset($tokens[$index + $offset])) {
+            return;
+        }
+
+        $token = $tokens[$index + $offset];
+
+        if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
+            if (isset($tokens[$index + $offset - 1])) {
+                $prevToken = $tokens[$index + $offset - 1];
+                if ($prevToken->isComment() && false !== strpos($prevToken->getContent(), "\n")) {
```
SpacePossum: since you target `master` here, is this still needed with the comment transformer?

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
keradus: also, when we could have `[` after `array` ?

phansys:
Thanks for the feedback @SpacePossum. I've copied the main concept from `NoSpacesInsideParenthesisFixer`, and this is the reason for these undesired issues.
I'll update the PR with your comments ASAP.

phansys:
PR updated with proposed changes.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,78 @@
...
+     */
+    private function removeSpaceAroundToken(Tokens $tokens, $index, $offset)
+    {
+        if (!isset($tokens[$index + $offset])) {
+            return;
+        }
+
+        $token = $tokens[$index + $offset];
+
+        if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
```
keradus: When `strpos($token->getContent(), "\n")` would be true?
I can't see a test like that

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,78 @@
...
+            if (!$token->equals('[')) {
+                continue;
+            }
+
+            $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE, $index);
+
+            // remove space after opening `[`
+            $this->removeSpaceAroundToken($tokens, $index, 1);
+
+            // remove space after closing `]`
```
gharlan: after -> before

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Remove spaces on one side of the token at a given index.
+     *
+     * @param Tokens $tokens A collection of code tokens
+     * @param int    $index  The token index
+     * @param int    $offset The offset where to start looking for spaces
+     */
+    private function removeSpaceAroundToken(Tokens $tokens, $index, $offset)
```
keradus: just pass `Token $token` instead of that 3 params

phansys: @keradus, I don't know how to access `$tokens[$index + $offset]` from `Token`. I could add the check twice for every call at `fix()`, but doesn't seems a right approach.
Thank you in advance.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
phansys: Added test case.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
phansys: Thanks @gharlan, fixed.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
keradus:
```php
private function removeSpaceAroundToken(Token $token)
{
    if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
        $token->clear();
    }
}

Comment (keradus) [inline src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php]:
```php
$this->removeSpaceAroundToken($tokens[$index + 1]);
$this->removeSpaceAroundToken($tokens[$endIndex - 1]);
```

tests/Fixer/Whitespace/NoSpacesInsideOfssetFixerTest.php
```diff
@@ -0,0 +1,109 @@
...
+                '<?php
+$arr = [2,   2 , ];
+$var = $arr[0];',
+                '<?php
+$arr = [2,   2 , ];
+$var = $arr[ 0 ];',
+            ),
+            array(
+                '<?php
+$arr[2] = 3;',
```
keradus: Please add case like `$arr[    ] = 3;`

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
phansys: But what about the `isset()` call? Remember, I've copied the concept from [`NoSpacesInsideParenthesisFixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/5de5794c0409982c5ed36aaa1546d2fa6d5fdf24/src/Fixer/Whitespace/NoSpacesInsideParenthesisFixer.php#L80-L82).

keradus: Please provide me any case when there could be no token after opening `[` or before closing `]`.

phansys: Obviously, I think that isn't possible from an offset accessor, so that's the difference with the fixer from which I've copied this. Thanks.

tests/Fixer/Whitespace/NoSpacesInsideOfssetFixerTest.php
phansys: @keradus, done.

README.rst
```diff
@@ -429,6 +429,13 @@ Choose from the list of available fixers:
                          or function name and the
                          opening parenthesis.
 
+* **no_spaces_inside_ofsset**
+                         There MUST NOT be a space
```
keradus: could be simpliefied to one sentence

phansys: What about "There MUST NOT be spaces between offset square braces and its contained values."?

keradus:
Please add this rule to `@Symfony` ruleset

phansys:
@keradus, done.

phansys:
Given the last test I've added (string offset with spaces), I'm thinking if a better name could be "trim offsets" or something similar, because actually an offset can contain (internal) spaces.
What do you think?

phansys:
Rebased master. Ping @keradus.

SpacePossum:
:+1:

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        return 'There MUST NOT be spaces between the offset square braces and its contained values.';
+    }
+
+    /**
+     * Remove spaces on one side of the token.
+     *
+     * @param Token $token
+     */
+    private function removeSpaceAroundToken(Token $token)
```
gharlan: the method name is confusing. it does not remove space around the given token. it removes the token itself if it is a single line whitespace.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,71 @@
...
+    {
+        if ($token->isWhitespace() && false === strpos($token->getContent(), "\n")) {
```
gharlan: why not `$token->isWhitespace(" \t")`?

gharlan:
please add a test case with `$a[ $b[ 1 ] ]`

phansys:
Thank you @gharlan, PR updated.

src/Fixer/Whitespace/NoSpacesInsideOfssetFixer.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Removes the token if it is single line whitespace.
+     *
+     * @param Token $token
+     */
+    private function removeWhitespaceToken(Token $token)
+    {
+        if ($token->isWhitespace(" \t") && false === strpos($token->getContent(), "\n")) {
```
gharlan: the second par (after `&&`) is useless now

phansys: You're right @gharlan, I didn't noticed the default range for whitespaces at `isWhitespace()`. Updated.

keradus:
Thank you @phansys.

PR #1816, opened 2016-03-12
PhpUnitDedicateAssertFixer - configuration is not required anymore

keradus:
ref #1792
cc @SpacePossum


SpacePossum:
Just wondering if you want this as if the configuration is empty the fixer will go looking for cases to fix and when a candidate is found it figures nothing should be done.
Even if we want to allow empty configurations, maybe we should optimize the fixer(s) to not go looking for cases?

SpacePossum:
oh strike that, if the configuration is empty you use the default, cool :)   :+1:

PR #1818, opened 2016-03-13
Readme - travis: consider php version in cache filename

gharlan:
refs #1649


keradus:
:-1:
Travis cache is not shared between envs.

gharlan:
I thought it is shared between envs, but not between branches.
But it seems, you are right. One cache archive per branch/env.

Issue #1820, opened 2016-03-13
Remove NoEmptyCommentFixer::clearCommentToken


keradus:
ref #1786

keradus:
@SpacePossum please, could you take care of this ?

Issue #1817, opened 2016-03-12
Adjust PhpUnitDedicateAssertFixer priority on 2.x line


keradus:
Fixer must be run after `no_alias_functions`
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1792#issuecomment-194758297

keradus:
@SpacePossum please, could you take care of this ?

SpacePossum:
will do :)
(please tag the issue as 2.0 so it shows up for the 2.0 release)

keradus:
done

PR #1821, opened 2016-03-13
Ensure that PhpUnitDedicateAssertFixer runs after NoAliasFunctionsFixer, clean up NoEmptyCommentFixer

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1820
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1817

### DIFF
--- a/tests/Fixer/Comment/NoEmptyCommentFixerTest.php
+++ b/tests/Fixer/Comment/NoEmptyCommentFixerTest.php
@@ -66,10 +66,12 @@ final class NoEmptyCommentFixerTest extends AbstractFixerTestCase
             ),
             array(
                 '<?php
+                echo 2;
                     '.'
 echo 1;
                 ',
                 '<?php
+                echo 2;
                     //
 echo 1;
                 ',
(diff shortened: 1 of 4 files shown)


keradus:
:+1:

tests/Fixer/Comment/NoEmptyCommentFixerTest.php
```diff
@@ -66,10 +66,12 @@ public function provideFixCases()
             ),
             array(
                 '<?php
+                echo 2;
```
GrahamCampbell: how comes these fail on 1.x?

SpacePossum: These shouldn't, I've added the missing case just to make sure the fixer works on 2.x.

GrahamCampbell: Might make sense to omit them here, and PR this to 1.x. Also, it might be an idea to add a new test case rather than editing one so we have all cases properly covered, even if in the current implementation, it may be trivial to see that it works.

SpacePossum: in theory; sure :) 
however this changes does not remove a test case, it updates a duplicate to a better one.
I simply don't have the time to create another PR for a small case like this on 1.x, even more since I know it is not needed. I put it on my list of todo's, but I don't see this a stopper for this pr.

keradus:
Thank you @SpacePossum.

Issue #826, opened 2014-12-13
I get the message "Legend: ?-unknown, I-invalid file syntax" with the option --verbose


p4scu41:
I run:

```
php-cs-fixer.phar fix test.php --verbose
```

The file test.php content:

``` php
<?php
class Test
{
    public $param = null;

    public function test($param) {
        $this->param = $param;
    }
}
?>
```

After run the command line, I get the below message.

```
F
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
1) ~\test.php (php_closing_tag, braces, eof_ending)
Fixed all files in 0.339 seconds, 3.500 MB memory used
```

But, the file is fixed correctly.

The message displayed is correct?, because when I add the option  --format="xml", I get the same legend at the begining of the xml, and that cause a error at the moment to read by netbeans

PR #1824, opened 2016-03-14
Reporting extracted to separate classes

ekho:


keradus:
@SpacePossum please, could you take a diper look ? I remember once you did sth like this

SpacePossum:
cool idea :)
I'm happy to review this in depth. 
For one it is missing utest, which brings me to my own PR https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1654 which will have conflicts on this one. I'll try some stuff :)

ekho:
I'll add unit tests

src/Report/ReportInterface.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param Stopwatch $stopwatch
+     */
+    public function setStopwatch(Stopwatch $stopwatch);
+
+    /**
+     * Gets the application instance.
+     *
+     * @return Application An Application instance
+     */
+    public function getApplication();
```
keradus: why is it needed?

ekho: it used in JunitReport (see #1825 )

src/Report/JsonReport.php
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Report;
+
+use Symfony\Component\Console\Output\OutputInterface;
+
+class JsonReport extends AbstractReport
```
keradus: to all reports: final + internal

src/Console/Command/FixCommand.php
```diff
@@ -547,6 +400,27 @@ protected function execute(InputInterface $input, OutputInterface $output)
     }
 
     /**
+     * @param ConfigurationResolver $resolver
+     * @param InputInterface        $input
+     * @param OutputInterface       $output
+     *
+     * @return ReportInterface
+     */
+    protected function createReport(ConfigurationResolver $resolver, InputInterface $input, OutputInterface $output)
```
keradus: why do you need resolver here?

src/Report/XmlReport.php
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * @internal
+ */
+final class XmlReport extends AbstractReport
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function process(array $changed)
```
keradus: so, whole body is just one method...?

src/Console/Command/FixCommand.php
ekho: @keradus You mean that needs to pass $format instead of $resolver ?

src/Report/XmlReport.php
ekho: @keradus why not? It was just a simple refactoring. I extracted parts of `switch` into separate classes. At other time we can refactor this method. But I think it's too much for a single pr.

src/Console/Command/FixCommand.php
keradus: If you don't need anything else - yes. No need for that huge dependency. it would be easier for testing as well.

src/Report/XmlReport.php
keradus: still, one god method... :(

src/Console/Command/FixCommand.php
ekho: Ok, I'll fix this

src/Report/XmlReport.php
ekho: I still believe that too many changes for a single PR

keradus:
please don't mention PR number in commits

src/Report/XmlReport.php
```diff
@@ -0,0 +1,98 @@
...
+        $timeTotalXML = $dom->createElement('total');
+        $timeTotalXML->setAttribute('value', round($fixEvent->getDuration() / 1000, 3));
+        $timeXML->appendChild($timeTotalXML);
+
+        if (OutputInterface::VERBOSITY_DEBUG <= $verbosity) {
+            $timeFilesXML = $dom->createElement('files');
+            $timeXML->appendChild($timeFilesXML);
+            $eventCounter = 1;
+
+            foreach ($this->stopwatch->getSectionEvents('fixFile') as $file => $event) {
```
keradus: this loop won't do any iteration

ekho: I do not know. I copied it "as is" from FixCommand

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1826

ekho: thanks

src/Report/ReportInterface.php
```diff
@@ -0,0 +1,48 @@
...
+     * @param Stopwatch $stopwatch
+     */
+    public function setStopwatch(Stopwatch $stopwatch);
+
+    /**
+     * Process changed files array.
+     *
+     * @param array $changed
+     */
+    public function process(array $changed);
```
keradus: PHP CS Fixer may be use in different ways, not only as tool but also as library.
Then Report may be build in so many different way - for example without... console.
Report should be returned, not directly printed.
For same reason it should not know anything about input/output (since we may not have console at all)

ekho: It's a good idea, but for next refactoring

ekho: Here does not required console, but classes which implements Input**Interface** and Output**Interface** only

keradus: That's my point. Don't rely on it.

ekho:
I fixed all issues and added unit tests

tests/Report/JsonReportTest.php
```diff
@@ -0,0 +1,150 @@
...
+            ->method('getVerbosity')
+            ->willReturn(OutputInterface::VERBOSITY_DEBUG);
+
+        $actualJson = '';
+        $expectedJson = '{"files":[{"name":"someFile.php","appliedFixers":["some_fixer_name_here"],"diff":"this text is a diff ;)"}],"memory":2.5,"time":{"total":1.234}}';
+
+        $this->mockOutput
+            ->expects($this->once())
+            ->method('write')
+            ->with($this->callback(function ($json) use (&$actualJson) { return $actualJson = $json; }));
```
GrahamCampbell: cs

ekho: @GrahamCampbell Can you explain?
```bash
$ php php-cs-fixer fix -vvv tests/Report/JsonReportTest.php
Loaded config from "/home/ekho/dev/PHP-CS-Fixer/.php_cs.dist"
.
Fixed all files in 0.484 seconds, 6.000 MB memory used
```

keradus:
> I fixed all issues and added unit tests

not all: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1824/files#r56246828

ekho:
I think that we should not make such a big refactoring in single PR

keradus:
It's not big

src/Report/ReportInterface.php
ekho: If somebody wants to get report without prinitng he can write his own Output like this
```php
class NotPrintedOutput implements OutputInterface {
    private $buffer = '';
    public function write($line)
    {
        $this->buffer .= $line;
    }

    public function getBuffer()
    {
         return $this->buffer;
    }
}
```
and pass it to Report

keradus: Pointless.
Keep it simple, keep it stupid. Have one thing for one responsibility. One for generating report (like report class), one for printing it (eg in current FixCommand)

It would simplify a lot your implementation and utests.

tests/Report/JsonReportTest.php
GrahamCampbell: PHP-CS-Fixer doesn't have a fixer for this yet, but PSR-2 says that closures can't go on one line like that.

GrahamCampbell:
> I think that we should not make such a big refactoring in single PR

Yeh, this is pretty small tbh.

tests/Report/JsonReportTest.php
keradus: Not sure. Didn't PSR2 added an appendix for allowing this?

gharlan: No, One-liner are not allowed.
The appendix says that this is allowed:

```php
$app->get('/hello/{name}', function ($name) use ($app) { 
    return 'Hello '.$app->escape($name); 
});
```

https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide-meta.md#31---multi-line-arguments-09082013

keradus: Oh, right. Thanks !

keradus:
@ekho, a proposal.

I could help you a bit with requested changes.
But if you would like me to help then you will need to wait for #1666 is closed (I can't focus on both at once). So it would delay this PR.

What do you think?

keradus:
@ekho, a proposal.

I could help you a bit with requested changes.
But if you would like me to help then you will need to wait for #1666 is closed (I can't focus on both at once). So it would delay this PR.

What do you think?

ekho:
@keradus I made changes and squashed commits into single one. Review PR again pls

src/Console/Command/FixCommand.php
```diff
@@ -108,19 +115,43 @@ protected function configure()
...
+                        'using-cache',
+                        '',
+                        InputOption::VALUE_REQUIRED,
+                        'Does cache should be used (can be yes or no)',
+                        null
+                    ),
                     new InputOption('cache-file', '', InputOption::VALUE_REQUIRED, 'The path to the cache file'),
                     new InputOption('diff', '', InputOption::VALUE_NONE, 'Also produce diff for each file'),
-                    new InputOption('format', '', InputOption::VALUE_REQUIRED, 'To output results in other formats', 'txt'),
+                    new InputOption(
```
keradus: :-1: for inconsistency for having InputOption sometimes as one-liners and sometimes not.

src/Console/Command/FixCommand.php
```diff
@@ -312,19 +348,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-                'path' => $input->getArgument('path'),
-                'progress' => (OutputInterface::VERBOSITY_VERBOSE <= $verbosity) && 'txt' === $input->getOption('format'),
-                'using-cache' => $input->getOption('using-cache'),
-                'cache-file' => $input->getOption('cache-file'),
-                'format' => $input->getOption('format'),
-            ))
-            ->resolve()
-        ;
+            ->setOptions(
+                array(
```
keradus: why is that change for?

ekho: sorry, PHPStorm formated it wrong. Fixed

src/Console/Command/FixCommand.php
```diff
@@ -506,6 +434,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+     *
+     * @return ReportInterface
+     */
+    protected function createReport($format, InputInterface $input, OutputInterface $output)
+    {
+        $reportClass = self::$reports[$format];
+
+        /** @var ReportInterface $report */
+        $report = new $reportClass();
+        $report->setDecoratedOutput($output->isDecorated());
```
keradus: This is only for `text` report. it should not be part of ReportInterface.
Please remove it and simply split `text` report to `text` and `decorated-text` reporters.

src/Console/Command/FixCommand.php
```diff
@@ -506,6 +434,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $report->setShowAppliedFixers(OutputInterface::VERBOSITY_VERBOSE <= $output->getVerbosity());
+        $report->setShowDiff((bool) $input->getOption('diff'));
```
keradus: Please remove that and in places when you will decide to show diff or not use `if (!empty($fixResult['diff'])) {`

src/Report/ReportInterface.php
```diff
@@ -0,0 +1,54 @@
...
+     * @param bool $decoratedOutput
+     */
+    public function setDecoratedOutput($decoratedOutput);
+
+    /**
+     * Sets the Stopwatch.
+     *
+     * @param Stopwatch $stopwatch
+     */
+    public function setStopwatch(Stopwatch $stopwatch);
```
keradus: All the setters should not be part of interface.
For example: If one want to build hist own report that doesn't care about Stopwatch he should not be forced to create such method.

Take a look what is the reporter for - to create report. User of Reporter doesn't care how the report was created, if for example dry-run or stopwatch was used or not.

Please, remove the methods from interface. I would even go a bit farther and don't have them at abstract level as well - instead of that pass things that really matters for concrete reporters in constructor.

keradus:
> I made changes and squashed commits into single one

Please don't squash commits (it's easier to look at newest changes when you could see only at them, not at everything at once).
Also, please don't refer PR in commit log.

src/Console/Command/FixCommand.php
ekho: `DecoratedTextReport` and `TextReport` will different in string templates only.

May be I should make AbstractTextReport? Or should I remove `final` from `TextReport` and make `DecoratedTextReport` extends `TextReport` ?

ekho: @keradus

src/Report/ReportInterface.php
ekho: @keradus How can I pass dry-run or stopwatch to constructor if I know only about interface when creating report class.
```php
protected function createReport($format, InputInterface $input, OutputInterface $output)
    {
        $reportClass = self::$reports[$format];

        /** @var ReportInterface $report */
        $report = new $reportClass();
        $report->setShowAppliedFixers(OutputInterface::VERBOSITY_VERBOSE <= $output->getVerbosity());
        $report->setDryRun((bool) $input->getOption('dry-run'));
        $report->setStopwatch($this->stopwatch);

        return $report;
    }
```

For stopwatch I can create `StopwatchAwareInterface` and calls setter if report implements it. But what about dry-run or showAppliedFixers?

src/Console/Command/FixCommand.php
keradus: I'm fine for having one class with option, like mentioned here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1824#discussion_r56498208

src/Report/ReportInterface.php
keradus: You just won't have 
```
$reportClass = self::$reports[$format];
new $reportClass();
```

anymore, but full factory class that will handle creation of reports.

ekho:
@keradus is it that you want ? https://github.com/bankiru/PHP-CS-Fixer/commit/59e39eb05ef33e477891441a9fdb32456d480fae

ekho:
@keradus please review my changes

src/Report/JsonReport.php
```diff
@@ -0,0 +1,83 @@
...
+            $jFiles[] = $jfile;
+        }
+
+        $json = array(
+            'files' => $jFiles,
+        );
+
+        if ($this->stopwatch) {
+            $fixEvent = $this->stopwatch->getEvent('fixFiles');
+            $json['memory'] = round($fixEvent->getMemory() / 1024 / 1024, 3);
```
GrahamCampbell: would be cool to not be duplicating this rounding code everywhere

GrahamCampbell: maybe just have the event deal with that?

src/Report/JsonReport.php
```diff
@@ -0,0 +1,83 @@
...
+            $json['time'] = array(
+                'total' => round($fixEvent->getDuration() / 1000, 3),
```
GrahamCampbell: can we just get rid of the total index since that's all we're recording?

src/ReportFactory.php
```diff
@@ -0,0 +1,138 @@
...
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+use Symfony\Component\Finder\Finder as SymfonyFinder;
+use Symfony\Component\Finder\SplFileInfo;
+use Symfony\Component\Stopwatch\Stopwatch;
+
+/**
```
GrahamCampbell: missing author data

src/ReportInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+interface ReportInterface
```
GrahamCampbell: is this interface internal?

ekho: @GrahamCampbell no, it can be used outside to write custom reports

src/Report/JsonReport.php
ekho: Each report can have it own requirements about formatting

src/Report/JsonReport.php
ekho: May be. I kept structure as defined previously. @keradus ?

src/ReportFactory.php
ekho: thanks, i'll fix it

src/Console/Command/FixCommand.php
```diff
@@ -309,6 +315,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $verbosity = $output->getVerbosity();
         $resolver = new ConfigurationResolver();
+        $resolver->getReportFactory()
+            ->setAddAppliedFixers(OutputInterface::VERBOSITY_VERBOSE <= $output->getVerbosity())
+            ->setIsDecoratedOutput($output->isDecorated())
+            ->setStopwatch($this->stopwatch)
```
ekho: @keradus May be remove dependency of stopwatch and pass time and memory to method `process` ?

src/Report/JsonReport.php
```diff
@@ -0,0 +1,83 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Report;
+
+use PhpCsFixer\ReportInterface;
+use Symfony\Component\Stopwatch\Stopwatch;
+
+/**
```
GrahamCampbell: when i said missing author, i meant in all places, not just there

src/Report/JsonReport.php
keradus: I would keep it. @GrahamCampbell , in the future we could add more times, like measured time of resolving configuration or grouped time of fixing by fixer (and not file)

GrahamCampbell: k

src/Report/TextReport.php
```diff
@@ -0,0 +1,152 @@
...
+            trim($fixResult['diff'])
+        );
+    }
+
+    /**
+     * @return string
+     */
+    private function getFooter()
+    {
+        if (!$this->stopwatch) {
```
keradus: When could stopwatch be null? Don't we always have it in FixCommand?

ekho: When we using php-cs-fixer as library for example

src/ReportInterface.php
```diff
@@ -0,0 +1,35 @@
...
+    public function configure(array $options);
+
+    /**
+     * Process changed files array. Returns generated report.
+     *
+     * @param array $changed
+     *
+     * @return string
+     */
+    public function process(array $changed);
```
keradus: Report is not a service, it won't generate multiple reports.
Make the report state-aware and move things like `changed` into constructors.

src/Console/Command/FixCommand.php
keradus: :+1: for removing stopwatch dependency from Report.

But doing that please remember about:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1824/files#r56658635
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1824/files#r56659028

src/Console/ConfigurationResolver.php
```diff
@@ -161,6 +176,22 @@ public function getFormat()
     }
 
     /**
+     * @return ReportFactory
+     */
+    public function getReportFactory()
```
keradus: what?
why?
how does ReportFactory is connected to resolving configuration?
ReportFactory is not aware of configuration.

Instead of ReportFactory it should return ready to use Report generator

src/Report/TextReport.php
keradus: Report is used only in FixCommand, FixCommand always has it.

keradus: One way or another it would be simplified after removing stopwatch dependency ;)

src/Console/ConfigurationResolver.php
keradus: argh, you expose report as well.
You should not expose factory...

tests/ReportFactoryTest.php
```diff
@@ -0,0 +1,141 @@
...
+     * @covers PhpCsFixer\ReportFactory::setIsDecoratedOutput
+     * @covers PhpCsFixer\ReportFactory::setAddAppliedFixers
+     * @covers PhpCsFixer\ReportFactory::setStopwatch
+     */
+    public function testSetters()
+    {
+        $factory = new ReportFactory();
+        $factory->registerBuiltInReports();
+
+        $refObject = new \ReflectionObject($factory);
```
keradus: AccessibleObject

ekho: @keradus You mean https://github.com/etsy/phpunit-extensions/wiki/Helpers#accessibleobject ?

src/ReportInterface.php
GrahamCampbell: why are all implementations of it in here marked as internal then

src/Console/ConfigurationResolver.php
ekho: @keradus https://github.com/bankiru/PHP-CS-Fixer/commit/8c4cfdb9e5cc1d373647b168a2c61ec094293f8e  ???

ekho: @keradus up =)

GrahamCampbell:
Needs rebasing. ;)

src/Console/ConfigurationResolver.php
keradus: better ;)

ekho:
@keradus I pushed ReportBuilder commit here

src/ReportBuilder.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * @param string $format
+     *
+     * @return $this
+     */
+    public function setFormat($format)
+    {
+        $this->format = $format;
```
keradus: why some properties are named properties, but some are placed under `options` ?

keradus:
Please fix Travis

src/ReportBuilder.php
ekho: but `format` used for creating concrete report in builder, but `options` passes to this concrete report

src/Report/TextReport.php
```diff
@@ -0,0 +1,212 @@
...
+        $template = '';
+
+        if ($this->isDecoratedOutput) {
+            $template .= '<comment>      ---------- begin diff ----------</comment>';
+            $template .= PHP_EOL.'%s'.PHP_EOL;
+            $template .= '<comment>      ----------- end diff -----------</comment>';
+        } else {
+            $template .= '      ---------- begin diff ----------';
+            $template .= PHP_EOL.'%s'.PHP_EOL;
+            $template .= '      ----------- end diff -----------';
```
keradus: this report is formatted differently than original one... let me handle it...

ekho: There is very little difference :/

src/ReportBuilder.php
keradus: and what are we gaining for that BC break of output ?
ref https://github.com/bankiru/PHP-CS-Fixer/pull/1

src/Report/ReportFactory.php
```diff
@@ -0,0 +1,86 @@
...
+        /** @var string[] $builtInReports */
+        static $builtInReports;
+
+        if (null === $builtInReports) {
+            $builtInReports = array();
+
+            /** @var SplFileInfo $file */
+            foreach (SymfonyFinder::create()->files()->name('*Report.php')->in(__DIR__) as $file) {
+                $relativeNamespace = $file->getRelativePath();
+                $builtInReports[] = 'PhpCsFixer\\Report\\'.($relativeNamespace ? $relativeNamespace.'\\' : '').$file->getBasename('.php');
```
GrahamCampbell: Better to use `__namespace__` rather than hard coding it.

src/Report/JsonReport.php
```diff
@@ -0,0 +1,67 @@
...
+            $jFiles[] = $jfile;
+        }
+
+        $json = array(
+            'files' => $jFiles,
+        );
+
+        if (null !== $reportConfig->getTime()) {
+            $json['time'] = array(
+                'total' => round($reportConfig->getTime() / 1000, 3),
```
GrahamCampbell: Why can't we move rounding into the report config class?

GrahamCampbell: Also, I don't think "report config" is a good name for the class if it has information like this in it. That information litterally makes it a "report".

keradus:
I would be nice to have it finished

ekho:
I will try to find time for it on this week

src/Report/JsonReport.php
keradus: Please propose better naming so we could finally merge the PR

sstok: It's not about configuration it's a result, so what about `ReportSummary`?

And the `ReportInterface` indicates it holds information while it actually reports (like a reporter or ReportFormatter).

--
When class performs an action (rather then keeping data) it should end `-er` or `-or`,
Controller, Reporter, Configurator, Fixer. Factory being in exception or course ;)

Credit: https://youtu.com/--L0bS8n5ow?t=26m38s (highly recommended video)

keradus: btw, `youtu.com’s server DNS address could not be found.`

sstok: Hehe that's a bug in youtube's side then :)

keradus: Then, `ReportSummary` and `JsonReporter` ?

keradus:
@SpacePossum is there anything more that you will need from this PR ?

SpacePossum:
ping @ekho 
please consider https://github.com/bankiru/PHP-CS-Fixer/pull/2

SpacePossum:
Besides the little naming discussion I think https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1824/files#r57077995 is the last open thing?
This will close https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1654 as well and a bunch of others.
@ekho please rebase when you have the time :)

keradus:
The naming is last thing.

ekho:
@keradus @SpacePossum all done, plz review again

src/Report/JsonReporter.php
```diff
@@ -0,0 +1,67 @@
...
+            $jFiles[] = $jfile;
+        }
+
+        $json = array(
+            'files' => $jFiles,
+        );
+
+        if (null !== $reportSummary->getTime()) {
+            $json['time'] = array(
+                'total' => round($reportSummary->getTime() / 1000, 3),
```
GrahamCampbell: Can we move this rounding inside the summary class please? If it's missed out in some places, that's surely a bug?

GrahamCampbell: Oh, wait I see. Maybe a `getRoundedTime()` method would be cool then?

GrahamCampbell: or `getNormalizedTime()`, or something?

keradus:
@ekho, please, could you rebase on master (not merge) ? Then, squashing consecutive commits of same author would be great.

ekho:
@keradus, done

src/Console/Command/FixCommand.php
```diff
@@ -292,8 +302,8 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : null;
-        if ($stdErr && extension_loaded('xdebug')) {
+	$stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : $output;
+	if (extension_loaded('xdebug')) {
```
keradus: tabs?

ekho:
@keradus, done

ekho:
@keradus , I'll fix it today evening. And I found mistakes in rebasing

src/Console/Command/FixCommand.php
```diff
@@ -302,8 +302,8 @@ protected function configure()
...
             $stdErr->writeln(sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.'));
```
keradus: @SpacePossum , this must not be printed to stdout on not txt format, as it would break JSON/YML format.

Other changes in this commit are OK.

ekho:
@keradus @SpacePossum tabs fixed, mistakes of rebasing too.

src/Console/Command/FixCommand.php
```diff
@@ -292,8 +294,8 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : null;
-        if ($stdErr && extension_loaded('xdebug')) {
+        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : $output;
+        if (extension_loaded('xdebug')) {
```
keradus: @SpacePossum , this must not be printed to stdout on not txt format, as it would break JSON/YML format.

Other changes in this commit are OK.

SpacePossum: it is printen to `stdErr` (line 297), I'm not sure I follow. If there is no 2 stream support by the console, it will print to the same, the other option would be to not print the info at all... not sure what is better

keradus: scenario:
When `$output` is not `ConsoleOutputInterface`, then `$stdErr` will be `$output`.

When format is JSON (or YML), on `stdout` (so `$output`) we will print the JSON to the console, user may grab it and parse as JSON format.
When we add extra content to it, like warning about xdebug, the output will be impossible to parse as JSON. Then the whole output is corrupted and in fact useless.
Take a look at https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/826 and it's fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/837

SpacePossum: I understand the issue, so the RFC on this is when a console does not support multiple streams than only output the result of the reporter? Should be an easy fix

keradus: on txt format it could print more output (as we already do), like info about xdebug or loaded config file

SpacePossum: please see
https://github.com/bankiru/PHP-CS-Fixer/pull/3/files#r59013415

keradus:
Thank you @ekho ad @SpacePossum

PR #1826, opened 2016-03-14
Fixer - remove measuring fixing time per file

keradus:
Measuring whole fixing time is great.
Measuring fixing time per fixer would be nice, but we don't have it (at least yet).
But measuring fixing time per file is pointless.


GrahamCampbell:
:+1:

gharlan:
:+1:

Issue #1354, opened 2015-08-18
Extra spaces fixer


EvgenyOrekhov:
I'd love to have a fixer that removes all extra spaces except for indentation.
Before:
```php
function    sum($a,     $b)      {
    return       $a    +       $b;
}
```
After:
```php
function sum($a, $b) {
    return $a + $b;
}
```

keradus:
You have so many rules here...
For sure it should be done by many fixers - and we already do it (well, maybe not all spaces, but part of it for sure).
Then, we could just have a group with all of that fixers.

EvgenyOrekhov:
@keradus Maybe this fixer should have only two rules:
1. do not fix indentation
2. replace all adjoint multiple spaces with one
So it would work like:
Before:
```php
function    sum    (     $a   ,     $b       )       {
    return       $a    +       $b              ;
}
```
After:
```php
function sum ( $a , $b ) {
    return $a + $b ;
}
```

keradus:
Considering CS the `after` version is as bad as `before` one.

EvgenyOrekhov:
@keradus I don't think there are many codebases that have so many spaces all over the code, and basically I want a fixer that would fix `return $a  + $b;` to `return $a + $b;`.
I think this fixer with these two simple rules is easy to implement and it would play very nice when used in combination with other fixers.

keradus:
So what you want is a fixer similar to `operators_spaces` that trim extra spaces.
Great fixer indeed!

SpacePossum:
I think all the cases from the examples here are fixed now :)

Issue #1342, opened 2015-08-05
Ternary operator - extra spaces


ffsantos92:
Before:
 ```php
$result = ((isset($test)?true:false;
```
After:
```php
$result = ((isset($test) ? true : false;
```
=====================
Before:
```php
$result = ((isset($test)      ? true : false;
```
After:
```php
$result = ((isset($test)      ? true : false;
```
So, the problem is related with extra white spaces removal.

keradus:
Yes, it is that by design: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/Symfony/TernarySpacesFixer.php#L90

Question here (to all of us) - do we want to change multiple single-line whitespaces into single whitespace?
And leaving multi-line intact

gharlan:
> Question here (to all of us) - do we want to change multiple single-line whitespaces into single whitespace?
> And leaving multi-line intact

I would like it.

stof:
@keradus I think removing useless multi-spaces should be a dedicated fixer

keradus:
yeah @stof, indeed that would be the best solution

ffsantos92:
I agree with @stof.

GrahamCampbell:
:+1:

r3oath:
Has this fixer been implemented yet? I couldn't find it initially, but if it hasn't I may write it myself. Sick of jumping into projects where there's mental usage of unnecessary white space.

If I wrote it, it would naturally transform a line such as

`if  ( $something->isTrue()  && $something->isLegit( $var1,   'hello') )  {`

into 

`if ($something->isTrue() && $something->isLegit($var1, 'hello')) {`

SpacePossum:
I don't think this has been done and I think would be welcome :) 

For MD, try the block
```
```php
```

for example

```php
 if ( $something->isTrue() 

&&

 $something->isLegit( $var1, 'hello') ) {
```

SpacePossum:
ping @keradus this has been done, the TernarySpacesFixer will remove extra spaces after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2079

PR #1834, opened 2016-03-21
Remove unwanted whitespace fixer.

r3oath:
Removes any consecutive whitespace characters and replaces them with a single whitespace character instead. When used in conjunction with a fixer such as `no_spaces_inside_parenthesis` you can transform the code:

```php
if (   $var ==  9 && $foo !=   false ) {
    return;
}
``` 

into the much more satisfying to read and look at equivalent:

```php
if ($var == 9 && $foo != false) {
    return;
}
``` 

This also helps to resolve issues #1354 & #1342 - when used in conjunction with other existing whitespace fixers.


GrahamCampbell:
We need way more tests, including those with comments in them, because i think this will break stuff if we have a comment followed by a newline.

r3oath:
Sure, adding more test is easily done.

r3oath:
I've added numerous extra tests, including those with comments and comment-newline combo's. 

Out of curiosity, what did you think was going to break? This fixer only touches extra whitespace buried amongst functional code. I've run it against some of my own code in testing and haven't had any issues with it breaking anything.

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
```diff
@@ -0,0 +1,64 @@
...
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Remove extra whitespace buried inside of any SLOC.';
+    }
+
+    public function getPriority()
+    {
+        // Make sure this is generally run before any other whitespace fixer.
```
keradus: why is that?

r3oath: Helps normalize whitespace to a degree so that other whitespace fixers have a cleaner state to work with.

keradus: Does the other whitespace fixers wont work without it? Or they would work better with this one run first?
When you add non-standard priority you need to test it as well:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/FixerFactoryTest.php#L181

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
```diff
@@ -0,0 +1,64 @@
...
+ * @author Tristan Strathearn <r3oath@gmail.com>
+ */
+final class NoExtraWhitespaceFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
```
SpacePossum:
```php
return $tokens->isTokenKindFound(T_WHITESPACE);
```
?

tests/Fixer/Whitespace/NoExtraWhitespaceFixerTest.php
```diff
@@ -0,0 +1,119 @@
...
+                '<?php
+$a = 1; // Comment with    whitespace
+return;',
+                '<?php
+$a  =   1; // Comment with    whitespace
+return;',
+            ),
+            array(
+                '<?php
+$a = 1; // Comment with    whitespace\n\r',
```
SpacePossum: Wrong quotes for '\n\r', the string should have double quotes.
The characters need to be swapped as well, i.e. '\r\n'

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
GrahamCampbell: @SpacePossum that's maybe a waste of computation though - basically every file with have whitespace

SpacePossum: fair enough :)

tests/Fixer/Whitespace/NoExtraWhitespaceFixerTest.php
r3oath: tndcbv6

I'll fix that up. Seems like the small addition to the family keeping me up all night has made my brain crap itself.

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
r3oath: They'll work better. I haven't tested it against **every** whitespace fixer, so it may or may not be necessary to have such a hight priority. But having it run first shouldn't in any situation stop another whitespace fixer from working properly.

tests/Fixer/Whitespace/NoExtraWhitespaceFixerTest.php
SpacePossum: thanks for the update :+1: 
and good luck with catching up sleep ;)

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
r3oath: @keradus In terms of the priority tests you mentioned, what exactly am I testing for if this fixer is supposed to run before every other (whitespace) fixer, not just a specific subset of fixers? Seems a bit silly in this instance that we'd need to test this when a good set of unit tests on the priority system itself should ensure that it works.

r3oath:
@keradus @SpacePossum @GrahamCampbell - What's the status on this PR, is she good to go and i'll leave it in your good hands? Or are there any more modifications needed to be done by yours truly?

GrahamCampbell:
I'd be more comfortable with this if it didn't overlap at all with any existing fixers.

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
keradus: those tests would prove not that `priority system` works, but the priority you have set is correct.

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
```diff
@@ -0,0 +1,64 @@
...
+            }
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Remove extra whitespace buried inside of any SLOC.';
```
sstok: This can be more descriptive as not everyone knows what SLOC means, and there is no clear level what is considered "extra" whitespace. What if someone uses multi line if-statements?

https://github.com/rollerworks/search/blob/master/src/ConditionOptimizer/RangeOptimizer.php#L254-L258
https://github.com/rollerworks/search/blob/master/src/ConditionOptimizer/RangeOptimizer.php#L309-L310 (extreme example).

r3oath: Sure, the description can definitely be made less ambiguous. It only affects non-line breaking whitespace, so extra tabs, spaces or null characters buried inside of functional code. So in your examples, it wouldn't affect the multi line if's, it would only affect them if they were written like so:

```php
if ($range->isLowerInclusive()    !==   $value->isLowerInclusive() ||
    $range->isUpperInclusive() !==    $value->isUpperInclusive()
) {
    continue;
}
```

in which case it would reformat it like:

```php
if ($range->isLowerInclusive() !== $value->isLowerInclusive() ||
    $range->isUpperInclusive() !== $value->isUpperInclusive()
) {
    continue;
}
```

src/Fixer/Whitespace/NoExtraWhitespaceFixer.php
r3oath: As I eluded too in my previous comment, what would be considered a reliable test then? As it's currently set to come before any whitespace fixer... am I listing the order of this fixer then every other whitespace fixer currently available? or just choosing some random whitespace fixer I know the priority of and making sure it comes before that. Just seems this system doesn't make much sense unless you're writing something that very specifically had to be run before or after a limited set of fixers.

keradus: For now you are guessing the priority. Don't guess it. Grab the proper value and prove it's proper, like:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/FixerFactoryTest.php#L193
First, the assertion will check if the relation between to fixer priorities are the one you want, then, integration test will prove that the fixers must run in given order to make their work properly. If you won't find any case for that you don't need to set the priority.

keradus:
What is a state of this PR ?

SpacePossum:
Something like this is "fixed" as well I think
```php
<?php
$a = 1;      // I like these comments..
$abcdef = 3; // ... aligned
```

Pretty sure the fixer needs checking for  priority/conflicts for `align_double_arrow` and `align_equals`. It prop. should run after `no_useless_else` `no_useless_return` `no_empty_statement` and maybe others.

Beside those cases; what happens to code indented with spaces (i.e. not tabs)?

keradus:
General question there - do we want this fixer in the first place, or we call it to generic ?

SpacePossum:
IMHO this fixer is too generic, fixing all whitespace except some cases (which are missing ATM btw.) is hard to maintain (as new syntax features might be added by PHP and new code styles might be introduced by frameworks and such). In general the fixers we've got now dealing with whitespace work the other way around; find only specific cases to fix and fix when needed. This is also way more handy to configure for users (like keep indent spaces but remove double spaces between class definition elements for example).
So I'm :-1: on this for maintenance concerns and configuration complexity needed.

keradus:
Totally for @SpacePossum reasoning.
It will enfore to handle all whitespaces in same way, and we already have cases when we don't want to touch extra whitespaces. We need to handle whitespaces with gradation.

Thanks for your work and understanding @r3oath !

Issue #1830, opened 2016-03-19
broken return fixer removal


GrahamCampbell:
If there are no braces, then a mess can happen...

```php
<?php

if ($foo) 
    bar();
else
    return;

foo()
```

This valid code gets broken...

SpacePossum:
```php
<?php

if ($foo) 
    bar();
else
    return;

foo()
```

Which fixer is breaking this example? (NoUselessReturnFixer #1832 only checks inside methods/functions and only for the last tokens, so it shouldn't be that fixer...)

GrahamCampbell:
Ah, yeh. Take my example code, and put it inside a method. I meant to do that when i wrote the example.

GrahamCampbell:
New example:

```php
<?php

class Foo
{
    public function bar()
    {
        if ($baz)
            return $this->baz();
        else
            return null;
    }

    public function baz()
    {
        //
    }
}
```

GrahamCampbell:
the return null fixer fixes the return null to just return, and then the return fixer removes the return, thus causing a syntax error with the unexpected brace

SpacePossum:
I messed up the comment to auto close this issue I think (@ https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1836), anyway, @keradus this one can be closed now :)

keradus:
thanks for reminder and closing ;)

PR #1836, opened 2016-03-21
NoUselessReturnFixer - Do not remove return if last statement in short if statement

SpacePossum:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1830

bug @ 1.12

### DIFF
--- a/Symfony/CS/Fixer/Contrib/NoUselessReturnFixer.php
+++ b/Symfony/CS/Fixer/Contrib/NoUselessReturnFixer.php
@@ -79,6 +79,11 @@ final class NoUselessReturnFixer extends AbstractFixer
                 continue;
             }
 
+            $previous = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$previous]->equalsAny(array(array(T_ELSE), ')'))) {
+                continue;
+            }
+
             $tokens->clearTokenAndMergeSurroundingWhitespace($index);
             $tokens->clearTokenAndMergeSurroundingWhitespace($nextAt);
         }
--- a/Symfony/CS/Tests/Fixer/Contrib/NoUselessReturnFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/NoUselessReturnFixerTest.php
@@ -34,6 +34,44 @@ final class NoUselessReturnFixerTest extends AbstractFixerTestBase
         return array(
             array(
                 '<?php
+                    function bar($baz)
+                    {
+                        if ($baz)
+                            return $this->baz();
+                        else
+                            return;
+                    }',
+            ),
+            array(
+                '<?php
+                    function bar($baz)
+                    {
+                        if ($baz)
+                            return $this->baz();
+                        elseif($a)
+                            return;
+                    }',
+            ),
+            array(
+                '<?php
+                    function bar($baz)
+                    {
+                        if ($baz)
+                            return $this->baz();
+                        else if($a)
+                            return;
+                    }',
+            ),
+            array(
+                '<?php
+                    function bar($baz)
+                    {
+                        if ($baz)
+                            return;
+                    }',
+            ),
+            array(
+                '<?php
     function b($b) {
         if ($b) {
             return;


Symfony/CS/Fixer/Contrib/NoUselessReturnFixer.php
```diff
@@ -66,7 +66,10 @@ public function getPriority()
     private function fixFunction(Tokens $tokens, $start, $end)
     {
         for ($index = $end; $index > $start; --$index) {
-            if (!$tokens[$index]->isGivenKind(T_RETURN)) {
+            if (
+                !$tokens[$index]->isGivenKind(T_RETURN) ||
+                $tokens[$tokens->getPrevMeaningfulToken($index)]->isGivenKind(T_ELSE)
```
GrahamCampbell: what about elseifs

GrahamCampbell: are there any other statements that could be affected?

GrahamCampbell: what about just

```
if ($foo)
    return;
```

removing that will break stuff if it's at the end of a method

SpacePossum: :+1:  updated

GrahamCampbell:
:+1:

GrahamCampbell:
Is this good to merge?

SpacePossum:
(FYI I have/see no todo's on this one)

Symfony/CS/Fixer/Contrib/NoUselessReturnFixer.php
```diff
@@ -79,6 +79,11 @@ private function fixFunction(Tokens $tokens, $start, $end)
                 continue;
             }
 
+            $previous = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$previous]->equalsAny(array(array(T_IF), array(T_ELSE), ')'))) {
```
gharlan: `if return`??

SpacePossum: indeed useless, not sure what I was thinking here ^-^
removed

GrahamCampbell:
:+1:

SpacePossum:
AppVeyor is failing for no reason

keradus:
yep, good to merge

Symfony/CS/Tests/Fixer/Contrib/NoUselessReturnFixerTest.php
```diff
@@ -34,6 +34,44 @@ public function provideFixCases()
         return array(
             array(
                 '<?php
+                    function bar($baz)
+                    {
+                        if ($baz)
+                            return $this->baz();
+                        else
```
GrahamCampbell: in the case that we have nothing to execute here, then we could strip out the `else` too?

GrahamCampbell: `else` is a special case really, like, we can't do this for if, because the code there inside the check could do stuff

GrahamCampbell: that doesn't stop this PR being merged though - that's just something to consider for later

GrahamCampbell: i still think having access to the AST might help us make changes like that, in more general

GrahamCampbell: say, empty else blocks

SpacePossum: We could, but it is out of scope. Its a nice case between `useless else` and `useless return` ;)
Lets bug fix first and work/think about more fixable cases later

SpacePossum: >say, empty else blocks

the trick is of course the `else`-block not being empty, just useless since it is the last block and returns as PHP does on default (for now), 
having the AST would be nice (long wait though ;)  )

keradus: AST is tricky that it will provide logic meaning of code but not formatting

SpacePossum: That is true, but it might help to identify useless or not optimal code paths that can be fixed automatically.. but indeed we are at the border between CS and SCA fixing ;)

keradus:
Thank you @SpacePossum.

Soullivaneuh:
@keradus Any chance to have a new 2.0.0 alpha tag?

This fixer is very nice but can't be used without any risk with the actual alpha.

Thanks.

keradus:
this is 1.12, not 2.x

keradus:
marging 1.x to 2.x is not a backport.
this is 1.12@dev, please wait for it's release first ;)

Issue #1829, opened 2016-03-18
braces rule adds newline before ?>


tillkruss:
The `braces` rule turns the following code:

```
<?php echo $foobar; ?>
```
Into:
```
<?php echo $foobar;
?>
```

Is that intended?

I'm running version: `1.11.2`

tillkruss:
Upgrading from `1.10` to `1.11.2` fixed it.

GrahamCampbell:
Are you sure the braces fixer is doing that?

GrahamCampbell:
If it's not the braces fixer, please confirm that if you add multiple php open tags to the file, it doesn;t happen anymore.

GrahamCampbell:
Say:

```php
<?php echo $foobar; ?>
<?php echo $foobar; ?>
```

tillkruss:
Can you explain? I have loads of opening tags in the file. It's an old code base.

tillkruss:
When I uncomment the `braces` fixer it doesn't happen.

When I enable the `braces` fixer I see that the following fixers have been executed:

```
unary_operators_spaces, logical_not_operators_with_successor_space, braces
```

GrahamCampbell:
> Can you explain?

PHP fixer treats monolithic files differently. Feel free to flick through the code. It'll make more sense than me explaining it badly.

SpacePossum:
I tried some stuff, like
```php
<?php echo 1; ?>
<?php echo 2; ?>
<?php echo 3; ?>
<?php echo 4; ?>
```

however I cannot reproduce the issue you report.
Could you maybe provide the following to help us track down the issue:
- minimal code sample that goes wrong
- the configuration you use

tillkruss:
Sure, this is the smallest working example:

```
<?php

class Foo
{
    public function bar()
    {
        ?>

            <input type="checkbox" value="<?php echo $foobar; ?>"> Email

        <?php
    }
}

```

This is my configuration:

```
$fixers = [
    'blankline_after_open_tag',
    'braces',
    'concat_without_spaces',
    'double_arrow_multiline_whitespaces',
    'duplicate_semicolon',
    'elseif',
    'empty_return',
    'encoding',
    'eof_ending',
    'extra_empty_lines',
    'function_call_space',
    'function_declaration',
    'include',
    'indentation',
    'join_function',
    'line_after_namespace',
    'linefeed',
    'list_commas',
    'logical_not_operators_with_successor_space',
[...]

PR #1840, opened 2016-03-26
[Bug] BracesFixer - Do add a line before close tag

SpacePossum:
Test for https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1829

### DIFF
--- a/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
@@ -32,6 +32,20 @@ class BracesFixerTest extends AbstractFixerTestBase
         return array(
             array(
                 '<?php
+class Foo
+{
+    public function A()
+    {
+        ?>
+        Test<?php echo $foobar; ?>Test
+        <?php
+        $a = 1;
+    }
+}
+',
+            ),
+            array(
+                '<?php
     if (true) {
         $a = 1;
     } else {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
This will surely break most of our fixers?

SpacePossum:
I ran 1.12 with the default configuration of the this project on the sample code and only the `BracesFixer` touched. Maybe we should add this case to the SF2 integration test, not sure though..

GrahamCampbell:
When I said "this", I was more referring to that style of code, rather than that specific test case.

Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
```diff
@@ -32,6 +32,20 @@ public function provideFixControlContinuationBracesCases()
...
             array(
                 '<?php
+class Foo
+{
+    public function A()
+    {
+        ?>
+        Test<?php echo $foobar; ?>Test
+        <?php
+        $a = 1;
```
SpacePossum: When this line is removed, the fixer will add a white line.
The fix for this has a big impact on the performance of the fixer. If we've a transformer for the open tag this should not be an issue at all.

keradus: the problem is that adding new tokens to collection is slow, and doing it in transformer means doing it for all code.

SpacePossum: I understand, I propose to doc. this as a known edge case and not fix it, i.e. choose speed over fixing an edge case.

SpacePossum:
rebased, RTM if test are green

keradus:
btw, 1.12 not 1.11? why?

SpacePossum:
my mistake, I tried and the commit can be picked as is on 1.11 without any problems.

keradus:
OK, great !

keradus:
Thank you @SpacePossum.

PR #1843, opened 2016-03-26
FileFilterIterator - add missing import

### LINKED ISSUE #1842: Latest 2.x cached performance very poor
After the latest changes, the performance when all files are cached is much worse than before. Easily 20x slower than it was before on a code base ~400 files.

GrahamCampbell:


keradus:
Thank you @GrahamCampbell.

PR #1844, opened 2016-03-27
Fixed the runner iterators

GrahamCampbell:
1. The native `CachingIterator` and `FilterIterator` both require an `Iterator` as the first param, not just `Traversable`
2. We need to run the filter iterator first, not the linting one, otherwise we're linting files we know we can already skip. This leads to very poor performance.


GrahamCampbell:
Had this wrong in my head. They are of course already in the correct order.

PR #1845, opened 2016-03-27
FileCacheManager - Allow linting to determine the cache state too

### LINKED ISSUE #1831: linting issue
Fixing a file with linting turned off, then turning linting on, but not flushing the cache causes issues because php-cs-fixer is now regarding that file with invalid syntax as fine, because it skipped linting it, even though the previous run of php-cs-fixer broke the file.

GrahamCampbell:
Closes #1831.


keradus:
Thank you @GrahamCampbell.

PR #1846, opened 2016-03-27
FileFilterIterator - Corrected an iterator typehint

GrahamCampbell:
The native `CachingIterator` and `FilterIterator` both require an `Iterator`, not just `Traversable`.


keradus:
Thank you @GrahamCampbell.

PR #1847, opened 2016-03-28
Runner - always cache files, not only when something is changed

### LINKED ISSUE #1842: Latest 2.x cached performance very poor

gharlan:
I think, this should fix #1842.

/cc @GrahamCampbell

### DIFF
--- a/src/Runner/Runner.php
+++ b/src/Runner/Runner.php
@@ -210,14 +210,16 @@ final class Runner
                     }
                     throw new IOException(sprintf('Failed to write file "%s".', $file->getRealPath()), 0, null, $file->getRealPath());
                 }
-
-                $this->cacheManager->setFile($name, $new);
             }
 
             $fixInfo = array(
                 'appliedFixers' => $appliedFixers,
                 'diff' => $this->differ->diff($old, $new),
             );
+        }
+
+        if (!$this->isDryRun) {
+            $this->cacheManager->setFile($name, $new);
         }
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Yes, along with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1843, I think this should fix it.

GrahamCampbell:
I'll try it out and see.

GrahamCampbell:
Hmm, looking at your PR, it doesn't look right. dryRun doesn;t indicate if something was changed or not?

GrahamCampbell:
Oh, wait, sorry. I've missread the diff. It does look good!

GrahamCampbell:
Just tested this with my other imports patch, and they work.

:+1:

keradus:
Question: why we want to remember that file is fixed when in fact we didn't fix it?
With this PR:
- first run, with dry-run, it found that file should be fixed, but didn't fix it, then mark in cache
- second run, without dry-run, it skipped the file, even while file needs fixing

gharlan:
@keradus good point, updated the pr.

keradus:
great ;)
@GrahamCampbell is that solves performance issue you have reported?

GrahamCampbell:
With my import fix PR, yes it does.

keradus:
Thank you @gharlan.

PR #1848, opened 2016-03-29
DocBlock - Remove some old unused phpdoc tags

GrahamCampbell:
These aren't in the PSR Spec anymore (for a good while), and nobody uses them as far as I know.


keradus:
Thank you @GrahamCampbell.

GrahamCampbell:
:)

PR #1849, opened 2016-03-29
DocBlock - The category tag should be together with package

GrahamCampbell:


keradus:
Could you add some reference for that change?

GrahamCampbell:
Well, there's no reason for any of the grouping anyway. They're all totally arbitrary.

The category tag is directly related to the package tag though, so it would make sense to group it.

keradus:
Thank you @GrahamCampbell.

GrahamCampbell:
:)

PR #1850, opened 2016-03-29
Update README.rst

chrjean:
Fix namespace

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -772,12 +772,12 @@ to the default list of symfony-level fixers:
 
     <?php
 
-    $finder = Symfony\CS\Finder::create()
+    $finder = Symfony\CS\Finder\DefaultFinder::create()
         ->exclude('somedir')
         ->in(__DIR__)
     ;
 
-    return Symfony\CS\Config::create()
+    return Symfony\CS\Config\Config::create()
         ->fixers(array('strict_param', 'short_array_syntax'))
         ->finder($finder)
     ;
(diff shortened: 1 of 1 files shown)


keradus:
README on 1.12 branch (default one) is valid for v1.12.
If you are looking for readme for v1.11.x go for 1.11 branch

GrahamCampbell:
Why is it that people always install "dev-master"?

chrjean:
I do not install "dev-master" as you said.

I just run composer require fabpot/php-cs-fixer which install the 1.11 and I was expected to install the last one / the default one on github

GrahamCampbell:
> (BTW, I'll love the breaking change of the minor update on the next composer up)

We don't have any breaking changes in minor releases?

GrahamCampbell:
> I just run composer require fabpot/php-cs-fixer which install the 1.11 and I was expected to install the last one / the default one on github

It DOES do that?

keradus:
v1.11 is 1.11 branch, which is not default
v1.12@dev (no stable yet) is 1.12 branch, which is default one
dev-master is master branch and is not related to the discussion.

1.12 will introduce new classes, which you can't use as you use 1.11. Introducing new things for next MINOR release is not BC break.

chrjean:
I'll have to modify my .php_cs file because of the namespace change. If I don't make the change, I'll have a "Class Not Found" Exception.

I understand your "you just have to update the NS" but it's not expected on minor change

keradus:
No, you dont understand. On your 1.11 you should use `DefaultFinder`. `Finder`, on which you got class not found error, has been introduced on dev-1.12. But you are using 1.11, so you should stick with `DefaultFinder` (which is still available on 1.12)

chrjean:
Ok, I just saw that the 1.11 files stays here. My bad, sorry

PR #1851, opened 2016-03-30
Fix token equals for when both line numbers are <null>.

SpacePossum:
This PR fixes the `Token::equals` method wrongly returning `false` when calling it with two `Token`-objects which both have no value (i.e. `<null>` / `0` / `''`) for line number.
Secondly it adds an option to the `equals` method to ignore the `line number` completely as the value of it may be arbitrary in some cases.

### DIFF
--- a/Symfony/CS/Tests/Tokenizer/TokensTest.php
+++ b/Symfony/CS/Tests/Tokenizer/TokensTest.php
@@ -1232,6 +1232,48 @@ PHP;
     }
 
     /**
+     * @dataProvider provideTokenEquals
+     */
+    public function testTokenEquals(Token $token, $compare, $expected, $caseSensitive = true, $lineNumberSensitive = true)
+    {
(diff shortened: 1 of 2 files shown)


keradus:
I see no big value in that `$lineNumberSensitive`. Why do you need it? If you won't provide line number (so passing array, not object) it will be ignored.
Remember that on 2.x branch the line has been removed from `Token`.

SpacePossum:
>Remember that on 2.x branch the line has been removed from Token.

Good to know, I'll re-target my new fixer and than I have no need for the ignore flag.

I have a use case were I want to check if one `T_VARIABLE` token at line X is the same `T_VARIABLE` token on line Y. For example in this code I check the `$a` var tokens against the other:
```php
<?php
    function f1()
    {
        $a = 123;
        return $a;
    }
```
That is how I ran into this.

PR #1852, opened 2016-03-30
ConfigurationResolver - disallow empty rule name

keradus:
Closing #1713

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -350,8 +350,12 @@ final class ConfigurationResolver
         $rules = array();
 
         foreach (array_map('trim', explode(',', $this->options['rules'])) as $rule) {
+            if ('' === $rule) {
+                throw new InvalidConfigurationException('Empty rule name is not allowed.');
+            }
+
             if ('-' === $rule[0]) {
-                $rules[ltrim($rule, '-')] = false;
+                $rules[substr($rule, 1)] = false;
             } else {
                 $rules[$rule] = true;
             }


src/Console/ConfigurationResolver.php
```diff
@@ -350,6 +350,10 @@ private function parseRules()
         $rules = array();
 
         foreach (array_map('trim', explode(',', $this->options['rules'])) as $rule) {
+            if ('' === $rule) {
+                throw new InvalidConfigurationException('Empty rule name is not allowed.');
+            }
+
             if ('-' === $rule[0]) {
```
GrahamCampbell: We might want to be more careful here, since someone could have any number of `-` in front.

keradus: Then we are throwing exception for non-existing rule.

GrahamCampbell: It's still a bit confusing though. I think we should be throwing an invalid configuration exception if a dash occurs anywhere in the string apart from the first character.

keradus: then why `-` and not `^` or any other char?
In the end, "invalid char in rule name" is just a special case of "name of non-existing rule"

GrahamCampbell: meh, it's just that `---foo` is being treated as `-foo`, and that's not correct

keradus: `foo` is `{"foo": true}`
`-foo` is `{"foo": false}`
but
`---foo` is `{"--foo": false}`

GrahamCampbell: not by your code here - you just ltrim it?

GrahamCampbell:
```
---foo is {"--foo": false}
```

it should be that, but currently it's actually being treated as:

```
---foo is {"foo": false}
```

keradus: indeed, you are right.
but out of scope.
but changed ;)

PR #1853, opened 2016-03-31
[WIP] Add ReturnAssignmentFixer

SpacePossum:
Taking this fixer for a run by Travis.
It is still WIP, however please state if you do or do not want this fixer to be added to the fixer project (after it would be finished of course).
Target is master because I need the fixed `Token:equals`.
There are hits on SF in 16 files (2.3 current).

Fixes var assignments that are return directly, for example:

Expected:
```php
<?php
return 1;
```

Input:
```php
<?php
$a = 1;
return $a;
```

TODO;
- [ ] integration tests and priority fixing
- [ ] only fix within functions/methods and do not touch `static` or `global` variables

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -675,6 +675,11 @@ Choose from the list of available fixers:
                          with their mt_* analogs.
                          (Risky fixer!)
 
+* **return_assignment**
+                         Variables should not be
+                         assigned and directly
+                         returned.
+
 * **self_accessor** [@Symfony]
                          Inside a classy element
                          "self" should be preferred to
--- a/src/Fixer/ReturnNotation/ReturnAssignmentFixer.php
+++ b/src/Fixer/ReturnNotation/ReturnAssignmentFixer.php
@@ -34,13 +34,15 @@ final class ReturnAssignmentFixer extends AbstractFixer
      */
     public function fix(\SplFileInfo $file, Tokens $tokens)
     {
-        // stop at 4, smallest number of tokens that could be a candidate: "<?php $a=1;return $a;"
-        for ($index = count($tokens) - 1; null !== $index && $index > 4; --$index) {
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
Does this case fail?

```php
<?php

function foo ($bar)
{
    $a = 123;
    if ($bar)
        $a = 12345;
    return $a;
}
```

rob006:
You can't do that for any global var - `$a` can be used in other files and removing assignment can change behavior of code.

SpacePossum:
>Does this case fail?

nope, I've added your code to the utest https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1853/files#diff-92916bea59cd94be6dfa75d1d7d22ec9R133 thanks for pointing it out :)

>You can't do that for any global var - $a can be used in other files and removing assignment can change behavior of code.

Thanks for reporting, I going to look into it :+1:

GrahamCampbell:
I'd say global variables should NEVER be used, so this is fine. Maybe we could just mark the fixer as risky and say that crappy code may be modified incorrectly.

keradus:
We could easily detect if the variable is imported from globals

rob006:
> I'd say global variables should NEVER be used, so this is fine

You almost always have some global variables, because when you create file like
```php
<?php
$config = require __DIR__ . '/config.php';
App::run($config);
```
you have global var `$config`.

keradus:
I think we are already having some fixers that touch code only inside functions/methods

gharlan:
another "nice" test case (reference parameter):

```php
function foo(&$a)
{
    $a = 1;
    return $a;
}
```

SpacePossum:
thanks all for looking, I'll re-open after I've reworked this one (might take a while...)

PR #1855, opened 2016-03-31
FixCommand - fix passing NullLinter to Runner

keradus:
Fixes error when one does not use linter:

> PHP Fatal error:  Uncaught TypeError: Argument 5 passed to PhpCsFixer\Runner\Runner::__construct() must be an instance of PhpCsFixer\Linter\LinterInterface, null given, called in /xxx/vendor/fabpot/php-cs-fixer/src/Console/Command/FixCommand.php on line 346 and defined in /xxx/vendor/fabpot/php-cs-fixer/src/Runner/Runner.php:70


SpacePossum:
:+1:

PR #1858, opened 2016-04-01
Updated my email address

GrahamCampbell:
I requested this before, but now I don't even own the domain anymore, so this really does need changing.


keradus:
not a bugfix.
why won't you use email address that you won't lose?

GrahamCampbell:
I won't lose the alt-three one, and I don't release my personal gmail one to the internet like this.

OskarStark:
This should be merged IMO

GrahamCampbell:
Yeh, someone's parked the old domain now:

image

ChipWolf:
My fault allowing it to expire; assumed domain was deprecated. Originally was used for a startup we collaborated on, no longer in use as you can see.

GrahamCampbell:
@keradus Rebased, and ready for merge please. :)

PR #1862, opened 2016-04-07
FixerFactory - Do not allow conflicting fixers

### LINKED ISSUE #910: Add The Ability To Mark Fixers As Conflicting
We need a way to stop fixers such as the concatenation fixers as incompatible with each other, and prevent them from being run at the same time.

It would also be useful to allow fixers that depend on each other, and thus fixer Y can only be run if fixer X is enabled.

### LINKED ISSUE #1033: List of conflicting fixers

### LINKED ISSUE #1822: Enabling no_empty_comment fixer with single_blank_line_before_namespace adds empty line between header comment and namespace
When enabling the `no_empty_comment` fixer and when having enabled the `single_blank_line_before_namespace`, running `php-cs-fixer` adds an empty line  between header comment and namespace.

Generally, here's what I do, since I've been waiting for a release of `2.0.0` since #1076:

* regularly, I update `fabpot/php-cs-fixer`, pinning it to a commit in the `master` branch
* explicitly disable or enable fixers (if interesting)
* tag `refinery29/php-cs-fixer-config` so we can use a unified configuration across our projects

## Steps required to reproduce

* updated from https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/1121cf0...535e239 (see https://github.com/refinery29/php-cs-fixer-config/pull/70), which - without enabling any additional fixers - results in removing an empty line between header and namespace
* enabled `no_empty_comment` fixer (see https://github.com/refinery29/php-cs-fixer-config/pull/73), which results in adding an empty line between header and namespace again

:information_desk_person: Not sure if I'm doing it wrong, but I have experienced this a couple of times now when updating `fabpot/php-cs-fixer`, and to me it is unclear what the expected result actually should be with the set of fixers enabled as they are. We are using this configuration both in open and closed source projects, and I figured that with the combination of fixers

* with a header comment given and the `header_comment` fixer enabled, the `single_blank_line_before_namespace` should not kick in
* with no header comment given and the `header_comment` fixer enabled, the single_blank_line_before_namespace` should kick in

SpacePossum:
Closes #910
Closes #1033
Closes #1822

### DIFF
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -212,15 +212,15 @@ final class FixerFactory
     private function getFixersConflicts(FixerInterface $fixer)
     {
         static $conflictMap = array(
-                'short_array_syntax' => array('long_array_syntax'),
-                'align_double_arrow' => array('unalign_double_arrow'),
-                'align_equals' => array('unalign_equals'),
-                'concat_with_spaces' => array('concat_without_spaces'),
-                'echo_to_print' => array('print_to_echo'),
-                'long_array_syntax' => array('short_array_syntax'),
-                'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
-                'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
-                'psr0' => array('psr4'),
+            'short_array_syntax' => array('long_array_syntax'),
+            'align_double_arrow' => array('unalign_double_arrow'),
+            'align_equals' => array('unalign_equals'),
+            'concat_with_spaces' => array('concat_without_spaces'),
+            'echo_to_print' => array('print_to_echo'),
+            'long_array_syntax' => array('short_array_syntax'),
+            'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+            'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
+            'psr0' => array('psr4'),
         );
 
         $fixerName = $fixer->getName();
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -217,10 +217,8 @@ final class FixerFactory
             'align_equals' => array('unalign_equals'),
             'concat_with_spaces' => array('concat_without_spaces'),
             'echo_to_print' => array('print_to_echo'),
-            'long_array_syntax' => array('short_array_syntax'),
             'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
             'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
-            'psr0' => array('psr4'),
         );
 
         $fixerName = $fixer->getName();
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -138,8 +138,10 @@ final class FixerFactory
     {
         $fixers = array();
         $fixersByName = array();
+        $fixerConflicts = array();
 
-        foreach (array_keys($ruleSet->getRules()) as $name) {
+        $fixerNames = array_keys($ruleSet->getRules());
+        foreach ($fixerNames as $name) {
             if (!array_key_exists($name, $this->fixersByName)) {
                 throw new \UnexpectedValueException(sprintf('Rule "%s" does not exist.', $name));
             }
@@ -148,6 +150,20 @@ final class FixerFactory
             $fixer->configure($ruleSet->getRuleConfiguration($name));
             $fixers[] = $fixer;
             $fixersByName[$name] = $fixer;
+
+            $conflicts = array_intersect($this->getFixersConflicts($fixer), $fixerNames);
+            if (count($conflicts) > 0) {
+                $fixerConflicts[$name] = $conflicts;
+            }
+        }
+
+        if (count($fixerConflicts) > 0) {
+            $message = 'The rules contain conflicting fixers:';
+            foreach ($fixerConflicts as $fixer => $fixers) {
+                $message .= sprintf("\n- \"%s\" with \"%s\"", $fixer, implode('", "', $fixers));
+            }
+
+            throw new \UnexpectedValueException($message);
         }
 
         $this->fixers = $fixers;
@@ -191,5 +207,29 @@ final class FixerFactory
         }, $data);
 
         return $this;
+    }
+
+    /**
+     * @param FixerInterface $fixer
+     *
+     * @return string[]|null
+     */
+    private function getFixersConflicts(FixerInterface $fixer)
+    {
+        static $conflictMap = array(
+                'short_array_syntax' => array('long_array_syntax'),
+                'align_double_arrow' => array('unalign_double_arrow'),
+                'align_equals' => array('unalign_equals'),
+                'concat_with_spaces' => array('concat_without_spaces'),
+                'echo_to_print' => array('print_to_echo'),
+                'long_array_syntax' => array('short_array_syntax'),
+                'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+                'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
+                'psr0' => array('psr4'),
+        );
+
+        $fixerName = $fixer->getName();
+
+        return array_key_exists($fixerName, $conflictMap) ? $conflictMap[$fixerName] : array();
     }
 }
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -158,12 +158,7 @@ final class FixerFactory
         }
 
         if (count($fixerConflicts) > 0) {
-            $message = 'The rules contain conflicting fixers:';
-            foreach ($fixerConflicts as $fixer => $fixers) {
-                $message .= sprintf("\n- \"%s\" with \"%s\"", $fixer, implode('", "', $fixers));
-            }
-
-            throw new \UnexpectedValueException($message);
+            throw new \UnexpectedValueException($this->generateConflictMessage($fixerConflicts));
         }
 
         $this->fixers = $fixers;
@@ -231,5 +226,31 @@ final class FixerFactory
         $fixerName = $fixer->getName();
 
         return array_key_exists($fixerName, $conflictMap) ? $conflictMap[$fixerName] : array();
+    }
+
+    /**
+     * @param array<string, string[]> $fixerConflicts
+     *
+     * @return string
+     */
+    private function generateConflictMessage(array $fixerConflicts)
+    {
+        $message = 'Rule contains conflicting fixers:';
+        $report = array();
+        foreach ($fixerConflicts as $fixer => $fixers) {
+            // filter mutual conflicts
+            $report[$fixer] = array_filter(
+                $fixers,
+                function ($candidate) use ($report, $fixer) {
+                    return !array_key_exists($candidate, $report) || !in_array($fixer, $report[$candidate], true);
+                }
+            );
+
+            if (count($report[$fixer]) > 0) {
+                $message .= sprintf("\n- \"%s\" with \"%s\"", $fixer, implode('", "', $report[$fixer]));
+            }
+        }
+
+        return $message;
     }
 }
--- a/tests/FixerFactoryTest.php
+++ b/tests/FixerFactoryTest.php
@@ -391,7 +391,7 @@ final class FixerFactoryTest extends \PHPUnit_Framework_TestCase
     /**
      * @dataProvider provideConflictingFixersRules
      * @expectedException \UnexpectedValueException
-     * @expectedExceptionMessageRegExp #^The rules contain conflicting fixers:\n#
+     * @expectedExceptionMessageRegExp #^Rule contains conflicting fixers:\n#
      */
     public function testConflictingFixers(RuleSet $ruleSet)
     {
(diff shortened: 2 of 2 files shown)


keradus:
ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1033

src/Console/ConfigurationResolver.php
```diff
@@ -480,6 +519,15 @@ function (FixerInterface $fixer) {
         }
     }
 
+    private function getFixersConflictMap()
+    {
+        return array(
+            'concat_with_spaces' => array('concat_without_spaces'),
+            'phpdoc_var_to_type' => array('phpdoc_type_to_var'),
```
GrahamCampbell: could also confict with the general annotation fixer

src/Console/ConfigurationResolver.php
```diff
@@ -480,6 +519,15 @@ function (FixerInterface $fixer) {
         }
     }
 
+    private function getFixersConflictMap()
+    {
+        return array(
```
GrahamCampbell:
```
    const CONFLICTS = [
        'align_double_arrow'              => 'unalign_double_arrow',
        'align_equals'                    => 'unalign_equals',
        'concat_with_spaces'              => 'concat_without_spaces',
        'echo_to_print'                   => 'print_to_echo',
        'long_array_syntax'               => 'short_array_syntax',
        'no_blank_lines_before_namespace' => 'single_blank_line_before_namespace',
        'no_duplicate_semicolons'         => 'no_empty_statement',
        'phpdoc_type_to_var'              => 'phpdoc_var_to_type',
        'psr0'                            => 'psr4',
    ];
```

that's the full list of conflicts i have, excluding the two general fixers we have

SpacePossum: thank you, updated the PR :)

src/Console/ConfigurationResolver.php
```diff
@@ -480,6 +519,21 @@ function (FixerInterface $fixer) {
...
+            'short_array_syntax' => array('long_array_syntax'),
+            'align_double_arrow' => array('unalign_double_arrow'),
+            'align_equals' => array('unalign_equals'),
+            'concat_with_spaces' => array('concat_without_spaces'),
+            'echo_to_print' => array('print_to_echo'),
+            'long_array_syntax' => array('short_array_syntax'),
+            'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+            'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
+            'psr0' => array('psr4'),
```
GrahamCampbell: NB, this fixer does not exist yet

src/Console/ConfigurationResolver.php
```diff
@@ -459,6 +459,45 @@ private function resolveFixers()
     {
         $this->fixers = $this->fixerFactory->useRuleSet($this->ruleSet)->getFixers();
 
+        // test if there are conflicting fixers configured
+        $conflictMap = $this->getFixersConflictMap();
```
keradus: I would put it into fixer factory.
One could use fixer factory without configuration resolver, and with handling conflicts in resolver he will be able to create conflicting fixers array.

What do you think?

GrahamCampbell:
We should probably allow fixers themselves to know what they conflict with, otherwise we have incorrect measures in place for when people add custom fixers.

GrahamCampbell:
Also, we should mark the proxy fixers as conflicting with the fixer they proxy?

SpacePossum:
>We should probably allow fixers themselves to know what they conflict with, otherwise we have incorrect measures in place for when people add custom fixers.

Good point, I had discussed this with @keradus already and while it is an interesting solution it is also harder to make and maintain. If we would build that conflicting fixers would be aware of each other, both ways. This PR provides one solution, but indeed lacks the option to deal with custom fixers.

>Also, we should mark the proxy fixers as conflicting with the fixer they proxy?

I'm not sure what you mean, i.e. what issue is there?

src/Console/ConfigurationResolver.php
SpacePossum: do you only want the conflict map  (and fetch method) in the Factory or the logic to check for conflicts as well?

GrahamCampbell:
well the var to type type to var confict, but the general one could conflict too

src/Console/ConfigurationResolver.php
keradus: logic as well, to not allow to produce conflicting set by Factory itself.

keradus:
General fixers are very generic and could or could not conflict with bunch of fixers depending on configuration of that particular general fixer. I would not bother to mark them as conflicting - if one want to use general fixer he should know what is he doing.

I'm agree that one would be able to generate conflicting ruleset, but that's edge case that would made this PR bigger. I would keep it in few steps.
- introduce basic conflicts check (this PR)
- introduce possibility to mark fixers as conflicting in FixerInterface, thanks to which one could handle conflicts of custom fixers
- introduce heavy logic to check conflicts not by rule names, but also by rule configs

Last two should not be done in this PR, but in new one.
But only if one day we will decide to do it.
I have strong feeling that it would take a bit of time that could be used better, eg adding full support for PHP7.

src/Console/ConfigurationResolver.php
SpacePossum: I've moved the code around, let me know what you think.

src/FixerFactory.php
```diff
@@ -192,4 +208,55 @@ private function sortFixers()
...
+    /**
+     * @param FixerInterface $fixer
+     *
+     * @return string[]|null
+     */
+    private function getFixersConflicts(FixerInterface $fixer)
+    {
+        static $conflictMap = null;
+
+        if (null === $conflictMap) {
```
keradus: why not `static $conflictMap = array(...);` ?

SpacePossum: refactor left over ;)

tests/FixerFactoryTest.php
```diff
@@ -387,4 +387,16 @@ private function createFixerMock($name, $priority = 0)
...
+
+    /**
+     * @expectedException \UnexpectedValueException
+     * @expectedExceptionMessageRegExp #^The rules contain conflicting fixers:\n#
+     */
+    public function testConflictingFixers()
+    {
+        FixerFactory::create()
+            ->registerBuiltInFixers()
+            ->useRuleSet(new RuleSet(array('short_array_syntax' => true, 'long_array_syntax' => true)))
```
keradus: please add second test with reversed order

src/FixerFactory.php
```diff
@@ -192,4 +208,51 @@ private function sortFixers()
...
+     */
+    private function getConflictingFixers(FixerInterface $fixer, array $candidateFixerNames)
+    {
+        $conflictMap = $this->getFixersConflicts($fixer); // to be replaced in the future by calling the Fixer itself for its conflicts
+        if (null === $conflictMap) {
+            return;
+        }
+
+        $conflictingFixers = array();
+        foreach ($candidateFixerNames as $candidate) {
```
keradus: I would iterate over `conflcitMap` instead of `candidateFixerNames`, first one will almost always be smaller

keradus: actually...
```php
$conflictingFixers = array_intersect($conflictMap, $candidateFixerNames);
```

src/FixerFactory.php
```diff
@@ -192,4 +208,51 @@ private function sortFixers()
...
+                'echo_to_print' => array('print_to_echo'),
+                'long_array_syntax' => array('short_array_syntax'),
+                'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+                'phpdoc_type_to_var' => array('phpdoc_var_to_type'),
+                'psr0' => array('psr4'),
+        );
+
+        $fixerName = $fixer->getName();
+
+        return array_key_exists($fixerName, $conflictMap) ? $conflictMap[$fixerName] : null;
```
keradus: I would always return array for simpler implementation and strong typing.
After that `getConflictingFixers` will actually be a one-liner ;)

src/FixerFactory.php
SpacePossum: :)
done

src/FixerFactory.php
```diff
@@ -148,6 +150,20 @@ public function useRuleSet(RuleSetInterface $ruleSet)
...
+            $conflicts = array_intersect($this->getFixersConflicts($fixer), $fixerNames);
+            if (count($conflicts) > 0) {
+                $fixerConflicts[$name] = $conflicts;
+            }
+        }
+
+        if (count($fixerConflicts) > 0) {
+            $message = 'The rules contain conflicting fixers:';
+            foreach ($fixerConflicts as $fixer => $fixers) {
+                $message .= sprintf("\n\"%s\" with \"%s\"", $fixer, implode('", "', $fixers));
```
keradus: What about changing:
```
The rules contain conflicting fixers:
foo with bar, baz
aaa with bbb, ccc
```

into:
```
The rules contain conflicting fixers:
- foo with bar, baz
- aaa with bbb, ccc
```

?

tests/FixerFactoryTest.php
```diff
@@ -387,4 +387,22 @@ private function createFixerMock($name, $priority = 0)
...
+    public function testConflictingFixers(RuleSet $ruleSet)
+    {
+        FixerFactory::create()->registerBuiltInFixers()->useRuleSet($ruleSet);
+    }
+
+    public function provideConflictingFixersRules()
+    {
+        return array(
+            array(new RuleSet(array('short_array_syntax' => true, 'long_array_syntax' => true))),
+            array(new RuleSet(array('long_array_syntax' => true, 'short_array_syntax' => true))),
```
keradus: would also be great to check if we won't have sth like:
```
The rules contain conflicting fixers:
short_array_syntax with long_array_syntax
long_array_syntax with short_array_syntax
```

Sorry if you will find it over-specified, but I have strong feeling that during implementing logic on FixerInterface side we could be catch by that.

keradus:
2 minors + please rebase + :+1:

tests/FixerFactoryTest.php
SpacePossum: Good idea, updated, let me know :)
btw. I think changing the FixerInterface, AbstractFixer and the few conflicting fixers will be easy now (point 2 https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1862#issuecomment-209830341),  let me know if this PR is on track for that.

SpacePossum:
ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1822

keradus:
> ref: #1822

Like this PR is closing referred one?

src/FixerFactory.php
```diff
@@ -192,4 +203,54 @@ private function sortFixers()
...
+
+    /**
+     * @param FixerInterface $fixer
+     *
+     * @return string[]|null
+     */
+    private function getFixersConflicts(FixerInterface $fixer)
+    {
+        static $conflictMap = array(
+                'short_array_syntax' => array('long_array_syntax'),
```
keradus: too much spaces

tests/FixerFactoryTest.php
keradus: Actually, I would like to merge this PR soon, if you want to work on it it would be nice, but please do it in separate PR ;)

SpacePossum:
yes, the report was boiled down to https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1822#issuecomment-196365979 so this should fix it (IMO)

src/FixerFactory.php
SpacePossum: :+1:  done

keradus:
great!
from first thread:
> list is not complete

is it still true?

src/FixerFactory.php
```diff
@@ -192,4 +203,54 @@ private function sortFixers()
...
+     */
+    private function getFixersConflicts(FixerInterface $fixer)
+    {
+        static $conflictMap = array(
+            'short_array_syntax' => array('long_array_syntax'),
+            'align_double_arrow' => array('unalign_double_arrow'),
+            'align_equals' => array('unalign_equals'),
+            'concat_with_spaces' => array('concat_without_spaces'),
+            'echo_to_print' => array('print_to_echo'),
+            'long_array_syntax' => array('short_array_syntax'),
```
gharlan: it is a duplicate, isn't it?

GrahamCampbell: ye

SpacePossum: thanks, removed (together with non-existing psr4 fixer)

SpacePossum:
>> list is not complete

> is it still true?

I think it is complete, but happy to update (/accept PR's on the branch ;)  )

keradus:
Great, updated first thread ;)

keradus:
Thank you @SpacePossum.

PR #1864, opened 2016-04-09
Print non-report output to stdErr

keradus:
Appendix for #1824


keradus:
Tests are failing, replaced by #1867
cc @SpacePossum

PR #1866, opened 2016-04-09
Handle throwables too

GrahamCampbell:


src/Runner/Runner.php
```diff
@@ -168,14 +168,9 @@ private function fixFile(\SplFileInfo $file, LintingResultInterface $lintingResu
...
-            $this->dispatchEvent(
-                FixerFileProcessedEvent::NAME,
-                FixerFileProcessedEvent::create()->setStatus(FixerFileProcessedEvent::STATUS_EXCEPTION)
-            );
-
-            $this->errorsManager->report(new Error(Error::TYPE_EXCEPTION, $name));
-
-            return;
+            return $this->processException($name);
+        } catch (\Throwable $e) {
```
keradus: it's sad to use class that doesn't exists on some env like php5

GrahamCampbell: that's fine? php doesn't care if the class doesn't exist

GrahamCampbell: catching both exception and then throwable in this order is common place these days

GrahamCampbell: Example: https://github.com/laravel/framework/blob/v5.2.29/src/Illuminate/Queue/SyncQueue.php#L33-L45.

keradus: I know PHP won't crash on it, but still it's ugly. Want to play with PHP7 features? use polyfill

GrahamCampbell: there is no polyfill for this that's not "shit"

GrahamCampbell: this way here is by far the best way to handle this, and that's why everyone does it this way

keradus: https://github.com/symfony/polyfill/blob/master/src/Php70/Resources/stubs/Error.php

Trowable = Error + Exception

GrahamCampbell: > Trowable = Error + Exception

Not necessarily true. Something can implement Throwable without extending Error or Exception.

GrahamCampbell: This way here is future proof.

GrahamCampbell: Symfony itself doesn't even catch it's own polyfil errors, it catches throwable.

GrahamCampbell:
Is this good to merge @keradus?

keradus:
I don't like the solution but still see the value of it.

For that I'm waiting for any other contributors to give opinion ;)

src/Runner/Runner.php
```diff
@@ -232,6 +227,21 @@ private function fixFile(\SplFileInfo $file, LintingResultInterface $lintingResu
     }
 
     /**
+     * Process an exception that occurred.
+     *
+     * @param string $name
+     */
+    protected function processException($name)
```
gharlan: why not `private` like other methods in this class?

gharlan:
👍 (apart from my comment above)

src/Runner/Runner.php
GrahamCampbell: sorry, just a habit i have from laravel

GrahamCampbell: i'll get that changed

GrahamCampbell:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1882

PR #1867, opened 2016-04-09
Print non-report output to stdErr

keradus:
Appendix for #1824

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -320,7 +320,11 @@ EOF
             ->getReporter($resolver->getFormat())
         ;
 
-        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : 'txt' === $reporter->getFormat() ? $output : null;
+        $stdErr = $output instanceof ConsoleOutputInterface
+            ? $output->getErrorOutput()
+            : ('txt' === $reporter->getFormat() ? $output : null)
+        ;
+
         if (null !== $stdErr && extension_loaded('xdebug')) {
             $stdErr->writeln(sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.'));
         }
@@ -353,8 +357,8 @@ EOF
             $resolver->isDryRun()
         );
 
-        $progressOutput = $showProgress
-            ? new ProcessOutput($this->eventDispatcher)
+        $progressOutput = $showProgress && $stdErr
+            ? new ProcessOutput($stdErr, $this->eventDispatcher)
             : new NullOutput()
         ;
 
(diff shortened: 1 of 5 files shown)


src/Console/Command/FixCommand.php
```diff
@@ -318,10 +315,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         ;
 
+        $reporter = ReporterFactory::create()
+            ->registerBuiltInReporters()
+            ->getReporter($resolver->getFormat())
+        ;
+
+        $stdErr = $output instanceof ConsoleOutputInterface
+            ? $output->getErrorOutput()
+            : ('txt' === $reporter->getFormat() ? $output : null)
```
GrahamCampbell: I find nested ternary statements hard to read, and can be avoided.

GrahamCampbell: If it doesn't fit on one line, i'd always go for plain old if statements.

src/Console/Command/FixCommand.php
```diff
@@ -320,7 +320,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
             ->getReporter($resolver->getFormat())
         ;
 
-        $stdErr = ($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : 'txt' === $reporter->getFormat() ? $output : null;
+        $stdErr = $output instanceof ConsoleOutputInterface
+            ? $output->getErrorOutput()
+            : ('txt' === $reporter->getFormat() ? $output : null)
+        ;
```
keradus: @SpacePossum 

```php
($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : 'txt' === $reporter->getFormat() ? $output : null;
```
means
```php
(($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : 'txt' === $reporter->getFormat()) ? $output : null;
```
not
```php
($output instanceof ConsoleOutputInterface) ? $output->getErrorOutput() : ('txt' === $reporter->getFormat() ? $output : null);
```

-------------------

before my change this always returns `$output`

GrahamCampbell: good enough reason not to use nested ternary statements then :P

GrahamCampbell: people have already got confused

src/Console/Command/FixCommand.php
keradus: I find nested ternary statements nice to read as long as they are properly used.

src/Console/Command/FixCommand.php
keradus: @GrahamCampbell , this PR is only for allowing me merging #1824 without waiting another week. I will merge it as soon as CI will be green.

GrahamCampbell: well, it's replacing something broken with something also broken

keradus:
Thank you @SpacePossum.

Issue #1767, opened 2016-02-10
how to exclude a file


glensc:
README.md should include some example how to exclude file.

the examples show how to exclude directory, and that does not work for files, and it's not very obvious to know that 'exclude' means just directories.

glensc:
i.e i wish to exclude some file that prevents the process to run, and no cache file is written due that, causing subsequent runs to be slow

```
Files that were not fixed due to internal error:
   1) src/color_smartytmpl.php
```

GrahamCampbell:
See symfony's finder docs.

GrahamCampbell:
> Files that were not fixed due to internal error:

It would be useful for us to know why that was happening though so we can fix it.

glensc:
@GrahamCampbell the ticket is about asking to improve php-cs-fixer doc not about actual problem.
and even if your only response is "go see symfony docs" it should include direct link to the doc.

ps: the cache issue is solved, re-run with php-cs-fixer.phar 1.11.1 is okay

glensc:
regarding internal error, i ran with -vvv but that did not produce anything useful to understand what's wrong with the file, and looking at the file i do not see what could be wrong with it.

SpacePossum:
Unfortunately there is no way to get more info about the internal error at the moment, unless you work with the code of the fixer itself of cause.
Could you provide a minimal set of code on which the error occurs? It would help us debugging the issue. As @GrahamCampbell stated it would also help to know which fixers you are running.
About the blacklisting of files and adding more details to the doc; sounds like a good idea! Did you figure it out? Then maybe you could describe how to do it, maybe even in a PR for the docs? Otherwise I'm happy to pick up your notes and PR it :)

Torniojaws:
Yeah, it would be very handy to have a way (or document one, if it exists) to ignore certain .php files. Mostly when the contents are partials with only a few lines of inline PHP. 

You can easily end up with this:
`<li <?php if ($cur == 'news') { echo 'class="active"'; } ?>><a href="news">News</a></li>`

Becoming this:
```
                       <li <?php if ($cur == 'news') {
   echo 'class="active"';
} ?>><a href="news">News</a></li>
```

keradus:
You can configure your finder in your `.php_cs` file, eg:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/.php_cs#L25
https://github.com/symfony/symfony/blob/master/.php_cs#L19

[...]

PR #1870, opened 2016-04-11
Update README.rst

glensc:
add information about excluding files.

sample taken from
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1767#issuecomment-206833220

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -684,6 +684,7 @@ to the default list of symfony-level fixers:
 
     $finder = Symfony\CS\Finder\DefaultFinder::create()
         ->exclude('somedir')
+        ->notPath('src/Symfony/Component/Translation/Tests/fixtures/resources.php')
         ->in(__DIR__)
     ;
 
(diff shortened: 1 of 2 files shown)


keradus:
Please read two last points of https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/CONTRIBUTING.md#quick-guide

glensc:
> Regenerate readme: php php-cs-fixer readme > README.rst. Do not modify README.rst manually!

blah. annoying. but the changeset itself is okay?

keradus:
not annoying, as we are autogenerating a lot of file's content ;)

README.rst
```diff
@@ -692,6 +693,8 @@ to the default list of symfony-level fixers:
         ->finder($finder)
     ;
 
+NOTE: `exclude` will work only for directories, so if you need to exclude file, try `notPath`.
```
keradus: I would just link to symfony/finder docs.
For now it says about some functionallity, but not all. And recreating Symfony docs here is pointless.

glensc:
you omitted answer to question, do i assume `yes` then?

keradus:
My answer is around diff line it concerns

README.rst
glensc: symfony docs are hard to read, they are written probably in api point of view not actual real life examples, like here. and it also does not emphasis that `exclude` **does not work** for files.

glensc: maybe add `.gitignore` style matching? it is more stright forward to write and understand:

```
src/**.php
!src/Subdir
*.php
```

glensc:
updated commit.

README.rst
keradus: No. We do not have (and do not want to have) our own finder.
For us finder is anything that will implement `\Traversable`.

symfony/finder is only example

SpacePossum:
:+1:

keradus:
Thank you @glensc

PR #1873, opened 2016-04-13
Enhancement: Show path to cache file if it exists

localheinz:
This PR

* [x] shows the path to a cache file if it exists

💁 This seems useful to me to see if a cache file already exists, what do you think?

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -347,6 +347,13 @@ EOF
             }
         }
 
+        if (null !== $stdErr && $config->usingCache()) {
+            $cacheFile = $config->getCacheFile();
+            if (is_file($cacheFile)) {
+                $stdErr->writeln(sprintf('Using cache file "%s".', $cacheFile));
+            }
+        }
+
         $showProgress = $resolver->getProgress();
         $runner = new Runner(
             $config,


src/Console/Command/FixCommand.php
```diff
@@ -347,6 +347,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
+        if ($config->usingCache()) {
+            $cacheFile = $config->getCacheFile();
+            if (file_exists($cacheFile) && null !== $stdErr) {
```
keradus: `file_exists` will return true if directory with the same name exists. It could lead to problem since we are going to use it like a file.
Please use `is_file`.

src/Console/Command/FixCommand.php
```diff
@@ -347,6 +347,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                $stdErr->writeln(sprintf('Using existing cache file "%s".', $cacheFile));
```
keradus: how we could use non-existing file?

src/Console/Command/FixCommand.php
GrahamCampbell: php is pro at naming things

SpacePossum:
:+1: for the idea :)

src/Console/Command/FixCommand.php
keradus: actually dir is type of file ;)

GrahamCampbell: whoever designed filesystems is pro at naming things :P

SpacePossum: this even echo'ed into `assertFile` of PHPUnit, liked to have that one fail for directories even though those are files of some form  ;)

src/Console/Command/FixCommand.php
```diff
@@ -347,6 +347,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
             }
         }
 
+        if ($config->usingCache()) {
```
keradus: if we put `null !== $stdErr` on start we could save 2 methods calls and one IO call

localheinz:
@GrahamCampbell @keradus @SpacePossum 

Thank you for your feedback - incorporated the suggestions!

SpacePossum:
:+1:  thanks!

keradus:
Thank you @localheinz.

PR #1875, opened 2016-04-16
renamed Composer package

fabpot:
For 2.x, the package should be renamed to `friendsofphp/php-cs-fixer`. When this PR is merged, I will mark the current package as abandoned on Packagist and submit the new one.

For 1.x, everything will still work as before and anyone relying on the 1.x branch won't have to change anything.

For people relying on 2.x@dev, they will need to update their composer.json dependency. That seems fair as the code is in development.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -58,7 +58,7 @@ To install PHP-CS-Fixer, install Composer and issue the following command:
 
 .. code-block:: bash
 
-    $ ./composer.phar global require fabpot/php-cs-fixer
+    $ ./composer.phar global require friendsofphp/php-cs-fixer
 
 Then make sure you have ``~/.composer/vendor/bin`` in your ``PATH`` and
 you're good to go:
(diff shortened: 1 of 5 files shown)


hkdobrev:
Do you think you could tag alpha/beta/RC after this is merged?

keradus:
Thank you @fabpot.

SpacePossum:
@Seldaek out of curiosity, may I ask why you down voted the PR?

Seldaek:
Because IMO it's a waste of time for every user to make them switch, and it doesn't bring much added value really.

Seldaek:
Also what @fabpot said isn't entirely correct, if the old package is marked abandoned, then all 1.x users will start getting the abandoned notice as well. Of course nobody **has** to switch though.

SpacePossum:
Thank you for your explanation of your vote, I get it now :)

PR #1884, opened 2016-04-23
Add PhpdocRegularHeaderFixer to clean up phpdoc file headers

curry684:
We have a lot of legacy code with headers like this:
```php
<?php
/**
 * File generated by PhpStorm
 */
```
As the header is technically a docblock the HeaderCommentFixer will not touch it. This PR introduces an extra fixer which converts docblock headers to regular comment headers if they are the first non-whitespace in a monolithic file, and do not contain any actual annotations.


GrahamCampbell:
:-1:

GrahamCampbell:
our existing fixer just need modifying

GrahamCampbell:
this is totally broken with all our other fixers anyway...

GrahamCampbell:
OMG, SORRY! just read the code and understand what us does now :+1:

GrahamCampbell:
Please fix the fixer priority with the others fixers though. :)

curry684:
Oh I'll have some reading up to do on that then   :laughing:   I did consider adding this functionality to the header comment fixer, as I intend to use it together with that, but I figured it makes sense separately too and it would only complicate config options otherwise.

curry684:
Should be fixed now, also added a priority to the no-package header cleaner to ensure it always gets run before any of the others and they pick up a 'newly empty' docblock correctly. Added integration tests for both of them.

Symfony/CS/Fixer/Symfony/PhpdocNoPackageFixer.php
```diff
@@ -39,4 +39,15 @@ public function getDescription()
...
         return '@package and @subpackage annotations should be omitted from phpdocs.';
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        /*
+         * Needs higher priority than any of the fixers cleaning up empty docblocks or fixing headers.
```
GrahamCampbell: :-1:

GrahamCampbell: Just change the priority of the new fixer to run before ALL phpdoc fixers.

GrahamCampbell: Apart from the fixer that removes empty phpdoc.

Symfony/CS/Fixer/Contrib/PhpdocRegularHeaderFixer.php
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        /*
+         * Must be run before the HeaderCommentFixer if also present to ensure it picks up the header correctly.
+         */
+        return 5;
```
GrahamCampbell: Missing fixer priority tests.

Symfony/CS/Tests/Fixtures/Integration/priority/phpdoc_no_package,phpdoc_regular_header.test
```diff
@@ -0,0 +1,23 @@
+--TEST--
+Integration of fixers: phpdoc_regular_header,header_comment.
+--CONFIG--
+level=none
+fixers=phpdoc_no_package,phpdoc_regular_header
+--EXPECT--
+<?php
+/*
+ * Docblock without phpdoc
+ *
```
GrahamCampbell: this output is incorrect

curry684: Uhm you're right, will get onto it and squash commits. No more Sunday morning rush on this hehe.

curry684: No actually this is correct - first the `@package` annotation is stripped, then it no longer has annotations and is therefore converted to a regular comment.

GrahamCampbell: That is wrong IMO. It should be converted straight to a comment first, then not modified.

Symfony/CS/Fixer/Symfony/PhpdocNoPackageFixer.php
curry684: This is also not correct. The `@package` stripper specifically can leave an empty phpdoc behind, so the correct order is:
 - Strip `@package`
 - Convert to regular comment
 - Let the `header comment` fixer do its job (in this case removing it)

GrahamCampbell: why should we allow something which is not phpdoc to be process as if it is

Symfony/CS/Tests/Fixtures/Integration/priority/phpdoc_no_package,phpdoc_regular_header.test
curry684: But that completely defeats the purpose? Why would we leave docblock annotations in a non-docblock? In that order we would get from this case:
```php
<?php
/*
 * @package thisNeverGetsReadOrParsedByAnyonee
 */
```

Symfony/CS/Fixer/Symfony/PhpdocNoPackageFixer.php
GrahamCampbell: for example, why are you not allowing the whitespace phpdoc stripping fixer to run if you're allowing the package removal fixer to run

GrahamCampbell: also, in 2.x, we don't even have that fixer, as such

Symfony/CS/Tests/Fixtures/Integration/priority/phpdoc_no_package,phpdoc_regular_header.test
GrahamCampbell: i think that's fine

GrahamCampbell: we should really remove ALL annotations, not just `package`, since that annotation has nothing to do with file headers anymore than any others might do

curry684: I severely disagree with that a code fixer should generate mangled output like that. I'll close the PR then as that means we don't agree on the expected nor intended outcome of this fixer nor its interoperability with existing fixers.

keradus:
BTW, PSR Docblock proposal ask to document a file using docblock, not comment

curry684:
In that case the existing fixer is broken, as it happily inserts a fresh comment block before any existing docblock sections. It doesn't have all that much choice though as the following is completely valid:
```php
<?php
/**
 * Echo 'hello world'.
 */
function echoHelloWorld()
{
    echo 'Hello World';
}
```
In this case it cannot be distinguished reliably whether this is a file header for an undocumented function, or a function docblock in an undocumented file.

Quite an edge case though as usually it's going to be file header, then `namespace`, optionally `use`, then structural docblock.

keradus:
there is already an issue raised to allow header fixer use docblock instead of comment.
when fixer was made it wasn't standardized yet

PR #1886, opened 2016-04-24
PhpdocScalarFixer - Fix lowercase str to string too

GrahamCampbell:
Some languages, such as Python use `str` rather than `string`, and people that switch between a lot of languages can often lose track of details like this, so I think it would be a good idea for our fixers to help out for fixing `str` to `string`.

Sending to 2.0 since this is technically a breaking change.

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocScalarFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocScalarFixer.php
@@ -29,6 +29,7 @@ final class PhpdocScalarFixer extends AbstractPhpdocTypesFixer
         'double' => 'float',
         'integer' => 'int',
         'real' => 'float',
+        'str' => 'string',
     );
 
     /**
--- a/tests/Fixer/Phpdoc/PhpdocScalarFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocScalarFixerTest.php
@@ -108,6 +108,7 @@ EOF;
      *
      * @param int|float $integer
      * @param int|int[] $foo
+     * @param string|null $bar
      *
      * @return string|bool
      */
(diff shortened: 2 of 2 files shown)


jbrooksuk:
:+1:

keradus:
:+1:

Soullivaneuh:
> Sending to 2.0 since this is technically a breaking change.

Why this should be a breaking change?

GrahamCampbell:
> Why this should be a breaking change?

People might have a class named `str`, all lowercase, for some reason.

GrahamCampbell:
@keradus Is this good to merge?

Soullivaneuh:
> for some reason.

Is `@param str` a valid phpdoc type? If not, for me this is not a BC break but a fix that should be integrated to 1.x.

@DeeJaVu you voted :-1: on @GrahamCampbell comments. Could you please tell use your opinion?

keradus:
it is if it's your class name

Soullivaneuh:
> it is if it's your class name

Can you please elaborate? I don't understand. This fixer fix only phpdoc, no?

keradus:
> Is `@param str` a valid phpdoc type?

Yes, it's valid phpdoc type.

```php
class str {}

/**
 * @param str $s
 */
function (str $s) {}
```

Soullivaneuh:
Thanks @keradus, didn't think about this tricky point.

:+1: for in this case.

keradus:
Thank you @GrahamCampbell.

GrahamCampbell:
@Soullivaneuh, @DeeJaVu is downvoting everything I do on GitHub for no good reason. I'm going to report him to GitHub. This has been going on for weeks.

GrahamCampbell:
@keradus Thanks for merging. Have a good weekend. ;)

PR #1076, opened 2015-03-21
Enhancement: Allow to specify cache file

localheinz:
This PR

* [x] adds a mutator and accessor for the cache file to `Config`
* [x] requires said mutators and accessors to be implemented when implementing `ConfigInterface`
* [x] injects `cacheFile` into `FileCacheManager` and adjust it to work with it
* [x] adds a `cache-file` option to the `FixCommand` to allow to specify it from the console
* [x] resolve the value for `cacheFile`, whether specified by config or option

:bulb: This can be quite useful if you want to speed up the runs on Travis and cache `php_cs.cache` between runs.

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1074#issuecomment-84415033.

PR #1888, opened 2016-04-26
Cache refactoring, better cache handling in dry-run mode

localheinz:
This PR

* [x] extracts a bunch of objects related to caching
* [x] always sets the file hash, but clears it on dry-run when the file needs fixing

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1847.

💁 The problem is this:

* we perform a `--dry-run` on Travis to determine coding style issues
* if we don't store the file hashes during a `--dry-run`, we'll never have a cache containing useful data
* if we never have a cache containing useful data, then #1076 (which aimed to be able to cache the fixer cache between builds) is rendered useless


keradus:
The problem is different.
With dry-run cache should be read but not stored.
And that affects not only CI usages.

localheinz:
@keradus 

What do you suggest to do?

stof:
@localheinz the code updating the cache should be fixed to take the dry-run mode into account (not updating the cache in dry-run mode, but still reading it)

localheinz:
@keradus @stof 

Realizing that my change doesn't make sense since the exit code when not using `--dry-run` will always be `0`, so we definitely need something different.

The problem with only ever reading from the cache file, but never writing to it when using `--dry-run` in a CI environment the cache file will never contain any useful data, rendering it useless in that context.

I think we have the following options, then

* store hashes for both dry and actual run
* store the value of the previous run (dry run or actual run?)

Since the former probably doesn't make a lot of sense, how about the latter, than:

* if both the previous and the current run have the same value for `--dry-run`, read from and write to the cache
* if the previous and current run do not have the same value for `--dry-run`, invalidate the cache

What do you think?

keradus:
consecutive runs with dry-run option must always return same output

localheinz:
@keradus 

That renders the whole caching functionality useless, though, at least in a CI context.

localheinz:
@keradus 

I've amended the PR, what do you think?

GrahamCampbell:
:-1:

GrahamCampbell:
This overwrites caches for non-dry runs completely.

alucic:
@GrahamCampbell writing 2 comments every time keeps useless email count up, and it's pretty simple to write just one comment.

On the other side, both of your comments are useless, being solution oriented is valued more than spamming thumbs down

GrahamCampbell:
@alucic Thanks for your useless comment.

GrahamCampbell:
> being solution oriented is valued more than spamming thumbs down

No it's not. If I don't think something's suitable, I'll say it. I don't have to provide an alternative. I don't see anything coming from you?

localheinz:
The thing is, what are you going to do?

Whatever the context, you either run

```
vendor/bin/php-cs-fixer fix --dry-run
```

**or**

```
vendor/binb/php-cs-fixer fix
```

but never both, right?

For example:

1. Locally, we just run the former to fix coding style issues (no need to do a dry run first, then followed by an actual run).
2. In CI, we want to detect coding style issues, but not fix them.

In both cases, we want a quick run, so we really want to be able to use *a* cache without producing invalid results.

The only way in invalid result could be produced would be if we did a dry run followed by an actual run and skipped fixing a file which was detected as having coding style issues in the first run.

@keradus 

I agree that a dry run should always produce the same results insofar as that it should always complain about files which need fixing, otherwise, I don't care.

So how about this, then

* we store the value of the `--dry-run` option in the cache
* if both the previous and the current run have the same value for `--dry-run`, read from and write to the cache
* if the previous and current run do not have the same value for `--dry-run`, invalidate the cache
* if the current run is a `--dry-run` and a file needs fixing, set the hash to `-1` (or some other value that will never be returned by the hash function used to calculate the hash)

stof:
@localheinz I often run a dry-run locally before running a normal run a bit later.

localheinz:
@keradus @stof 

Ok, then how about c0597f9?

src/FileCacheManager.php
```diff
@@ -86,15 +86,22 @@ public function needFixing($file, $fileContent)
...
         }
 
         $file = $this->getRelativePathname($file);
 
-        $this->newHashes[$file] = $this->calcHash($fileContent);
+        $hash = $this->calcHash($fileContent);
+
+        // set hash to an impossible value in dry-run so that the file will definitely be inspected on the next run
+        if ($isDryRun && array_key_exists($file, $this->newHashes) && $this->newHashes[$file] !== $hash) {
+            $hash = -1;
```
localheinz: @keradus @stof 

Instead of setting to the hash to `-1`, we can of course set to something else, for example `null` or `false`.

What do you think?

localheinz:
@keradus @stof 

Care to take a look?

This fix literally shaves off hours of build time every day.

src/FileCacheManager.php
```diff
@@ -86,15 +86,24 @@ public function needFixing($file, $fileContent)
...
         }
 
         $file = $this->getRelativePathname($file);
 
-        $this->newHashes[$file] = $this->calcHash($fileContent);
+        $hash = $this->calcHash($fileContent);
+
+        // in dry run, clear hash for file that needs fixing so it will definitely be inspected on the next run
+        if ($isDryRun && isset($this->oldHashes[$file]) && $this->oldHashes[$file] !== $hash) {
+            unset($this->newHashes[$file]);
```
SpacePossum: Are we sure the offset `$this->newHashes[$file]` is always set here?
And maybe pass the `dry-run` flag on construction of the `FileCacheManager`?
Anyhow, good idea it will save me lots of processing time as well :+1:

stof: @SpacePossum not an issue. Unsetting undefined offsets is perfectly accepted by PHP (it will be a no-op)

localheinz:
@stof 

Amended!

SpacePossum:
:+1:

keradus:
@localheinz 
do I get it right what you want to achieve ?

In dry run mode:
File is cached but need a fixing - let us fix it on next non dry-run run.
File is cached and not need a fixing - do nothing.

Then, if you won't unset the key, but also won't override it, you will have the same effect.

For that, I would have 2 cache implementations - one read only, one read+write. first for dry-run, second for real run. And even 3rd for no-cache functionality available.

localheinz:
@keradus 

The aim of this PR is to save hours of running

```
$ vendor/bin/php-cs-fixer fix --dry-run
```

on Travis by actually caching the results between builds. Currently, if we run the above command with a cold cache on Travis, we never write to the cache, meaning that we're not actually caching at all.

Now, when we do a `--dry-run` and only cache the hashes of files which do not need fixing (that is, we unset the hash of files which *do* need fixing), we solve the caching problem and still produce the same results on consecutive runs.

Does that help?

localheinz:
@keradus 

Is this what you suggest?

```php
// only set the hash if this is not a dry run or the file does not need fixing
if (!$this->isDryRun || isset($this->oldHashes[$file]) && $this->oldHashes[$file] === $hash) {
    $this->newHashes[$file] = $hash;
}
```

keradus:
If I get it right, you rely on cache builded only on CI. Then, your cache file is always empty, as with dry-run tool never store the cache. For that you want to cache files that doesn't need fixing regardless dry-run mode.

You could also change your comment to `do not set the hash in dry-run when file needs fixing`.

Also, please remember to add tests for changed behaviour.

localheinz:
@keradus 

I've just run the tests with coverage, the only coverage for `FileCacheManager` is accidental - which kind of tests do you expect me to write?

keradus:
As we seen from discussion, it's not obvious when we should store information in cache file and when not.
Depending on isDryRun/isInOldCache/doesNeedAFixing we save cache information or not. I'm not asking to add utests for whole `CacheManager`, but at least for added part.

localheinz:
@keradus 

The problem with writing tests for this is that [`FileCacheManager::isCacheAvailable()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/FileCacheManager.php#L105-L114) evaluates to `false`  in a test context, regardless of whether caching is enabled or not. 

How about adding something like `ToolInfo::isTestContext()` to allow using the cache in a test context?

localheinz:
@keradus 

How about this:

* you provide tests for the current functionality in `master` and I follow up by rebasing this PR and adding tests for the changed functionality
* you just merge this, acknowledging that you made a mistake in #1847

keradus:
As the logic is not obvious and to be sure it works properly and wont be broken by further refactoring I would really want to have that tests.
After said that, for now I can't make tests for whole caching functionality soon enough, that is why I asked you to add logic only for changed things.

What would you say about decoupling things? Creating sth like
```php
interafce CacheDecider
{
    public function shouldRead();

    public function shouldWrite();
}
```

?
Then, whole your logic would be actually in very small and easy to test class.

localheinz:
@keradus 

I'm entirely sold on TDD, so 👍 - let me think about it!

keradus:
thank you @localheinz ! I'm really glad to hear that ;)

localheinz:
@keradus 

Here's a suggestion, what do you think about it?

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1890

keradus:
I'm not sure how you will achieve storing in cache even in dry-run mode the files that don't need to be fixed when you will be able to only choose NullCacheHandler or FileCacheHandler.
(I don't think it needed to have separate PR, as we don't see the solution and fix at once)

localheinz:
@keradus 

While I personally prefer small PRs, I don't mind pushing into this branch.

keradus:
usually yes, but having refactoring that will allow to fix a bug and then having separated PR with fix itself is tricky, when you can't see if refactoring indeed will allow for a fix ;) and since current fix is few liner, it won't be big PR when put together.

localheinz:
@keradus @stof

How about 0f69d2b as an approach for caching?

Extracted a bunch of objects with focussed responsibilities.

If you like this, happy to push the fix, too!

src/FileCacheManager.php
```diff
@@ -32,144 +35,76 @@
...
-                'linting' => $this->linting,
-                'rules' => $this->rules,
-                'hashes' => $this->newHashes,
-            )
-        );
-
-        if (false === @file_put_contents($this->cacheFile, $data, LOCK_EX)) {
-            throw new IOException(sprintf('Failed to write file "%s".', $this->cacheFile), 0, null, $this->cacheFile);
-        }
+        $this->cache->set($file, $this->calcHash($fileContent));
```
localheinz: @keradus @stof

The fix for the original issue would then probably be to

* inject `$isDryRun` into `FileCacheManager`
* adjust `setFile()`:

    ```php
    public function setFile($file, $fileContent)
    {
        $file = $this->getRelativePathname($file);

        if ($this->isDryRun && $this->cache->get($file) !== $this->calcHash($fileContent)) {
            $this->cache->clear($file);
            
            return;
        }

        $this->cache->set($file, $this->calcHash($fileContent));
    }
    ```

localheinz:
@keradus @stof 

Do you have time to take a look?

Build is failing on PHP5.3, maybe you have an idea what to do about it?

See https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/126861509#L226-L241.

src/FileCacheManager.php
```diff
@@ -32,144 +35,91 @@
  */
```
keradus: Description will need to be updated

src/Runner/Runner.php
```diff
@@ -83,8 +89,36 @@ public function __construct(
...
+            $isDryRun
+        );
+    }
+
+    /**
+     * @param ConfigInterface $config
+     *
+     * @return HandlerInterface
+     */
+    private function cacheHandler(ConfigInterface $config)
```
keradus: naming convension, it's not a getter to use shorthand method name, it creates a new instance. please rename the method, like `createFoo`

tests/FileCacheManagerTest.php
```diff
@@ -0,0 +1,714 @@
...
+
+        $manager = new FileCacheManager(
+            $handler,
+            $signature
+        );
+
+        unset($manager);
+    }
+
+    public function testNeedFixingReturnsTrueIfCacheHasNoNash()
```
keradus: `NoNash`

tests/FileCacheManagerTest.php
```diff
@@ -0,0 +1,714 @@
...
+            $isDryRun
+        );
+
+        $manager->setFile($file, $fileContent);
+    }
+
+    public function testSetFileUsesNormalizedFilePath()
+    {
+        $cacheFile = $this->file();
+        $file = __DIR__.DIRECTORY_SEPARATOR.'src'.DIRECTORY_SEPARATOR.'hello.php';
```
localheinz: @keradus 

How about extracting a collaborator here as well, for example `PathNormalizer`? Makes it easier to test, too.

keradus:
@localheinz in general, I do very like your approach and big thanks for you for the work you made so far!

Sadly, I have no option for now to check the 5.3 env, I will try to investigate it on week beginnig, as well as playing with your branch (I didn't find sth big that would need a change so far, but I want to check how the code will behave in few cases. Also, I'm not sure about some general method naming like `foo` vs `getFoo`, especially when having general `get` method)

src/Cache/CacheInterface.php
```diff
@@ -0,0 +1,53 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Cache;
+
+use Serializable;
```
gharlan: we do not import classes from root namespace, use `\Serializable` instead.

tests/FileCacheManagerTest.php
keradus: I was actually thinking about dropping usage of DIRECTORY_SEPARATOR and always use unix notation, and then just before asserting the value normalize the path. What do you think?

gharlan:
> Also, I'm not sure about some general method naming like foo vs getFoo, especially when having general get method)

http://symfony.com/doc/current/contributing/code/conventions.html#method-names

So it should be `getFoo`, and I think it is not problematic when also having general `get` method.
From link above:

> The usage of these methods are only allowed when it is clear that there is a main relation:
>
> * a CookieJar has many Cookie objects;
> * a Service Container has many services and many parameters (as services is the main relation, the naming convention is used for this relation);
> * a Console Input has many arguments and many options. There is no "main" relation, and so the naming convention does not apply.

We have second case. A class with main relation (`get`), and some other Relations (`getFoo`).

localheinz:
@gharlan @keradus @stof

Happy to apply more fixes if needed, please let me know what you think!

localheinz:
@gharlan @keradus @stof 

Any news on this one?

localheinz:
@gharlan @keradus @stof

Any chance of this getting merged soon?

gharlan:
I still think all getters should be renamed to `getX`.

src/Cache/Cache.php
```diff
@@ -0,0 +1,94 @@
...
+            return;
+        }
+
+        return $this->hashes[$file];
+    }
+
+    public function set($file, $hash)
+    {
+        if (!is_int($hash)) {
+            throw new \InvalidArgumentException('Value needs to be an integer');
```
SpacePossum: minor RFC

```php
throw new \InvalidArgumentException(sprintf('Value needs to be an integer, got "%s".', is_object($hash) ? get_class($hash) : gettype($hash)));
```

tests/Cache/NullHandlerTest.php
```diff
@@ -0,0 +1,67 @@
...
+            ->method($this->anything())
+        ;
+
+        $handler = new NullHandler();
+
+        $handler->write($cache);
+    }
+
+    /**
+     * @return \PHPUnit_Framework_MockObject_MockObject|CacheInterface
```
SpacePossum: this typehint looks odd to me

localheinz: @SpacePossum 

It's helpful insofar as that it helps with IDEs complaining about invalid types, `$handler->write($cache)` expects `$cache` to implement `CacheInterface`, if it's skipped from the `@return` docblock, the IDE complains.

And, after all, it's an implementation of `CacheInterface,

localheinz:
@SpacePossum 

Applied the suggestions, and increased the verbosity in `Signature` a bit, too.

tests/Cache/NullHandlerTest.php
SpacePossum: I see, thanks, nice trick! :+1:

src/Cache/FileHandler.php
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public function read()
+    {
+        if (!file_exists($this->file)) {
+            return;
+        }
+
+        $content = file_get_contents($this->file);
+        $cache = @unserialize($content);
```
keradus: Please, dont use unserialize on cache file - store file as JSON instead.
One thing is one could serialize any class he want to, as `allowed_classes` was added to PHP7 (And is not available for PHP5)
Second thing is that we sometimes read cache file by other script, and JSON format is universal.

src/FileCacheManager.php
```diff
@@ -32,159 +37,100 @@
...
+    /**
+     * @var DirectoryInterface
+     */
+    private $cacheDirectory;
+
+    /**
+     * @param HandlerInterface   $handler
+     * @param SignatureInterface $signature
+     * @param bool               $isDryRun
+     * @param DirectoryInterface $cacheDirectory
```
keradus: |null

src/Cache/NullHandler.php
```diff
@@ -0,0 +1,33 @@
...
+namespace PhpCsFixer\Cache;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+final class NullHandler implements HandlerInterface
+{
+    public function getFile()
```
keradus: violation of interface - this is suppose to return string.

src/Cache/NullHandler.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+    }
+
+    public function read()
```
keradus: same here but cache interface

src/Cache/HandlerInterface.php
```diff
@@ -0,0 +1,36 @@
...
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+interface HandlerInterface
+{
+    /**
+     * @return string
+     */
+    public function getFile();
```
keradus: why in interface? only FIleHandler needs it.
It's not needed for any other handler, eg if we add RedisHandler

src/Cache/NullHandler.php
localheinz: @keradus 

So what do you suggest to return here, an empty string?

src/Cache/NullHandler.php
localheinz: @keradus 

So what do you suggest to return here, a `NullCache` implementation of the `CacheInterface`?

src/FileCacheManager.php
localheinz: @keradus 

So what do you suggest to update the description to?

src/Cache/HandlerInterface.php
localheinz: @keradus 

It is implemented by the `NullHandler`.

src/FileCacheManager.php
```diff
@@ -32,159 +37,100 @@
...
-        // ignore corrupted serialized data
-        if (null === $data) {
-            return;
-        }
+        $file = $this->cacheDirectory->relativePathTo($file);
 
-        if (!isset($data['php'], $data['version'], $data['linting'], $data['rules'])) {
-            return;
-        }
+        $hash = $this->calcHash($fileContent);
```
keradus: this is slow operation.
please, perform it only when needed.

for example, we shouldn't need it when we are using `NullHandler`.
But so far, even with `NullHandler` we are using `Cache` and calculate hash here, in manager. Maybe moving it to `CacheInterface` would allow us to move that logic out of this class and introduce `NullCache` ?

src/Cache/FileHandler.php
SpacePossum: Is other applications reading the cache a real case? I mean than we need to start (considering) BC support on the cache format as well.
Using JSON will introduce overhead which can be avoided using serializing wouldn't it?

src/Cache/HandlerInterface.php
keradus: but `NullHandler` don't need it...
that's my point

keradus:
Please fix AppVeyor/Travis, it's all red.
Also, please rebase

src/Cache/FileHandler.php
localheinz: @keradus @SpacePossum 

The problem with using JSON here is that we'll have a whole new class of problems, like, validating the JSON format.

src/Cache/NullHandler.php
keradus: depending on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1888#discussion_r63658286
probably yes

for sure we need common interface

src/Cache/NullHandler.php
keradus: I suggest to not have it, let's discuss it here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1888#discussion_r63657564

src/Cache/FileHandler.php
keradus: That's why it's the second thing ;)
First one is more important.

what overhead ?
json_decode is often faster, and always more secure

src/Cache/NullHandler.php
localheinz: @keradus @SpacePossum 

How about making our lives a lot easier and introducing a `NullFileCacheHandler` instead?

src/Runner/Runner.php
```diff
@@ -82,15 +87,37 @@ public function __construct(
...
+                    PHP_VERSION,
+                    ToolInfo::getVersion(),
+                    $config->usingLinter(),
+                    $config->getRules()
+                ),
+                $isDryRun
+            );
+        }
+
+        return new NullFileCacheManager();
```
localheinz: @gharlan @keradus @SpacePossum 

How about this now?

src/FileCacheManager.php
localheinz: @keradus 

With the introduction of the `NullFileCacheManager`, this shouldn't be an issue anymore.

src/Cache/Cache.php
```diff
@@ -0,0 +1,115 @@
...
+            $data['version'],
+            $data['linting'],
+            $data['rules']
+        );
+
+        $cache = new self($signature);
+
+        $cache->hashes = $data['hashes'];
+
+        return $cache;
```
localheinz: @keradus 

How about this now?

src/Cache/FileHandler.php
keradus: @localheinz if JSON would be incorrect then assume cache file is not there.
Just like what you do now when you deserialize and get instance of other interface.
I would prefer to have invalid JSON than wrongly injected instance of not our class (as somebody could create his own class with his own code basing on our interface)

localheinz:
@keradus 

> Please fix AppVeyor/Travis, it's all red.

About the failing build, this is it, see https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/131065274#L226:

```
PHP Fatal error:  Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]]) in /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Cache/FileHandler.php on line 66
PHP Stack trace:
PHP   1. {main}() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:0
PHP   2. PHPUnit_TextUI_Command::main() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:722
PHP   3. PHPUnit_TextUI_Command->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:104
PHP   4. PHPUnit_TextUI_TestRunner->doRun() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:152
PHP   5. PHPUnit_Framework_TestSuite->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/TestRunner.php:398
PHP   6. PHPUnit_Framework_TestSuite->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestSuite.php:716
PHP   7. PHPUnit_Framework_TestCase->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestSuite.php:716
PHP   8. PHPUnit_Framework_TestResult->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:693
PHP   9. PHPUnit_Framework_TestCase->runBare() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestResult.php:609
PHP  10. PHPUnit_Framework_TestCase->runTest() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:737
PHP  11. ReflectionMethod->invokeArgs() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:860
PHP  12. PhpCsFixer\Tests\Cache\FileHandlerTest->testWriteThrowsIOExceptionIfFileCanNotBeWritten() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:860
PHP  13. PhpCsFixer\Cache\FileHandler->write() /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/tests/Cache/FileHandlerTest.php:120
PHP  14. Exception->__construct() /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Cache/FileHandler.php:66
Fatal error: Wrong parameters for Exception([string $exception [, long $code [, Exception $previous = NULL]]]) in /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Cache/FileHandler.php on line 66
Call Stack:
    0.0016     814792   1. {main}() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:0
    0.0190    1539696   2. PHPUnit_TextUI_Command::main() /home/travis/.phpenv/versions/5.3.29/bin/phpunit:722
    0.0190    1540448   3. PHPUnit_TextUI_Command->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:104
    1.1126   60884520   4. PHPUnit_TextUI_TestRunner->doRun() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/Command.php:152
    1.1415   61799808   5. PHPUnit_Framework_TestSuite->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/TextUI/TestRunner.php:398
    4.0079  161249784   6. PHPUnit_Framework_TestSuite->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestSuite.php:716
    4.0819  161660328   7. PHPUnit_Framework_TestCase->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestSuite.php:716
    4.0820  161660328   8. PHPUnit_Framework_TestResult->run() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:693
    4.0825  161661376   9. PHPUnit_Framework_TestCase->runBare() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestResult.php:609
    4.0833  161702576  10. PHPUnit_Framework_TestCase->runTest() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:737
    4.0834  161704064  11. ReflectionMethod->invokeArgs() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:860
    4.0834  161704096  12. PhpCsFixer\Tests\Cache\FileHandlerTest->testWriteThrowsIOExceptionIfFileCanNotBeWritten() phar:///home/travis/.phpenv/versions/5.3.29/bin/phpunit/phpunit/Framework/TestCase.php:860
    4.0837  161707184  13. PhpCsFixer\Cache\FileHandler->write() /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/tests/Cache/FileHandlerTest.php:120
    4.0853  161739728  14. Exception->__construct() /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Cache/FileHandler.php:66
```

You mentioned a while ago you would have a look into it - do you have a suggestion?

src/Cache/FileHandler.php
localheinz: @keradus 

Do you want to use https://github.com/justinrainbow/json-schema for validating the JSON?

keradus:
```json
"symfony/filesystem": "^2.4 || ^3.0",
```
will fix error detected by Travis.
4th argument (file) was added in 2.4, but so far we are supporting 2.1.
I'm fine with bumping it.

src/FileCacheManager.php
keradus: Sorry. What about describing caching logic in general docblock for FileCacheManager ?
As it's now a little more complicated

src/Cache/Cache.php
keradus: *this* like switching to JSON?
:+1:

localheinz: @keradus 

Exactly!

Was wondering whether you'd prefer to ensure that the required keys are available, too, for example, or the values hold data that makes sense.

tests/Cache/CacheTest.php
```diff
@@ -110,6 +110,45 @@ public function testFromJsonThrowsInvalidArgumentExceptionIfJsonIsInvalid()
...
+                'foo',
+                'bar',
+            ),
+            'hashes' => array(),
+        );
+
+        return array_map(function ($missingKey) use ($data) {
+            unset($data[$missingKey]);
+
+            return array(
```
keradus: :D

tests/Cache/CacheTest.php
```diff
@@ -110,6 +110,45 @@ public function testFromJsonThrowsInvalidArgumentExceptionIfJsonIsInvalid()
...
+     */
+    public function providerMissingData()
+    {
+        $data = array(
+            'php' => '5.5.5',
+            'version' => '2.0',
+            'linting' => false,
+            'rules' => array(
+                'foo',
+                'bar',
```
keradus: it shouldn't matter, but it's actually `'foo' => true` or `'bar' => [config]`

src/FileCacheManagerInterface.php
```diff
@@ -0,0 +1,46 @@
...
+ *  - fixer version changed,
+ *  - rules changed,
+ *  - file is new,
+ *  - file changed.
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+interface FileCacheManagerInterface
```
keradus: this interface is actually not needed to work on file. just our implementation do.
also, what about moving it (and implementations) to `Cache` subnamespace?

src/FileCacheManager.php
```diff
@@ -25,11 +25,11 @@
  * Cache is supported only for phar version and version installed via composer.
  *
  * File will be processed by PHP CS Fixer only if any of the following conditions is fulfilled:
- *  - cache is not available,
+ *  - cache is corrupt
  *  - fixer version changed,
  *  - rules changed,
- *  - file is new,
- *  - file changed.
```
keradus: so, some of the items are followed by `,` and some not ?

src/Cache/FileCacheManagerInterface.php
```diff
@@ -0,0 +1,46 @@
...
+ *  - fixer version changed,
+ *  - rules changed,
+ *  - file is new,
+ *  - file changed.
+ *
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+interface FileCacheManagerInterface
```
localheinz: @keradus 

The interface is implemented by `FileCacheManager` and `NullFileCacheManager`.

Maybe rename both to `CacheManagerInterface` and `NullCacheManager`?

src/FileCacheManager.php
localheinz: @keradus 

Dropped all of them now!

src/Cache/FileCacheManagerInterface.php
keradus: :+1:

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1888#discussion_r63702267

keradus:
:+1:

localheinz:
Thank you, everyone!

keradus:
after that great work of yours, could you double check if it solve the issue that raised this PR?

localheinz:
@keradus 

Hehe, good one, let me have a look again!

localheinz:
@keradus 

Checked this out right here

* https://github.com/refinery29/php-cs-fixer-config/pull/90

and it's using the cache just fine, see https://travis-ci.org/refinery29/php-cs-fixer-config/jobs/131158829#L311-L317:

```
$ if [[ "$WITH_CS" == "true" ]]; then vendor/bin/php-cs-fixer fix --config=.php_cs --verbose --diff --dry-run; fi
Loaded config from ".php_cs".
Using cache file "/home/travis/.php-cs-fixer/.php_cs.cache".
SS
Checked all files in 0.002 seconds, 6.500 MB memory used
```

👍

src/Cache/Directory.php
```diff
@@ -0,0 +1,50 @@
...
+
+    /**
+     * @param string $directoryName
+     */
+    public function __construct($directoryName)
+    {
+        $this->directoryName = $directoryName;
+    }
+
+    private function normalizePath($path)
```
gharlan: please use `public, protected, private` order for new classes.

src/Cache/Signature.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return $this->fixerVersion;
+    }
+
+    public function isLintingEnabled()
+    {
+        return $this->isLintingEnabled;
+    }
+
+    public function rules()
```
gharlan: why not `getRules`?

GrahamCampbell: i think to match up with other function names we have?

src/Cache/Signature.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        return $this->fixerVersion;
+    }
+
+    public function isLintingEnabled()
+    {
+        return $this->isLintingEnabled;
+    }
+
+    public function getRules()
```
localheinz: @GrahamCampbell 

Amended!

src/Cache/Directory.php
localheinz: @gharlan 

Sorted!

composer.json
```diff
@@ -18,7 +18,7 @@
         "ext-tokenizer": "*",
         "symfony/console": "^2.3 || ^3.0",
         "symfony/event-dispatcher": "^2.1 || ^3.0",
-        "symfony/filesystem": "^2.1 || ^3.0",
+        "symfony/filesystem": "^2.4 || ^3.0",
```
stof: if you drop Symfony 2.3, please bump to 2.7 directly. It does not make sense to use 2.4 here IMO, as it is unmaintained since years

keradus: On the other side, if we don't need 2.7 there is no need to enforce it, and ^2.4|^3 will work with 2.7 as well.

keradus: but no big deal here

keradus:
Thank you @localheinz.

SpacePossum:
cool! thanks so much @localheinz (and reviewers) :+1:  :)

localheinz:
Awesome, thanks everyone! :+1:

keradus:
thank you for your understanding for me being mean and demanding ;)

PR #1890, opened 2016-04-28
Enhancement: Implement and use cache handlers

localheinz:
This PR

* [x] extracts a `CacheHandler` interface, implemented by `FileCacheHandler` and `NullCacheHandler`, and injects either implementation into `FileCacheManager`, depending on whether caching is enabled and the environment allows it

Related to #1888.

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1888#issuecomment-215381346.

### DIFF
--- a/src/FileCacheHandler.php
+++ b/src/FileCacheHandler.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+use Symfony\Component\Filesystem\Exception\IOException;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+final class FileCacheHandler implements CacheHandler
+{
+    /**
+     * @var string
+     */
+    private $cacheFile;
+
+    public function __construct($cacheFile)
+    {
+        $this->cacheFile = $cacheFile;
+    }
+
+    public function willCache()
+    {
+        return true;
+    }
+
+    public function read()
+    {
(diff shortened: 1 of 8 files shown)


src/FileCacheHandler.php
```diff
@@ -0,0 +1,56 @@
...
+
+        return file_get_contents($this->cacheFile);
+    }
+
+    public function write($content)
+    {
+        $bytesWritten = @file_put_contents($this->cacheFile, $content, LOCK_EX);
+
+        if (false === $bytesWritten) {
+            throw new IOException(sprintf('Failed to write file "%s".', $this->cacheFile), 0, null, $this->cacheFile);
```
localheinz: @keradus 

Not sure how to test this, I tried to write tests yesterday using `mikey179/vfsstream`, but kind of failed using the `LOCK_EX` flag above.

src/FileCacheManager.php
```diff
@@ -43,15 +40,14 @@
     /**
      * Create instance.
      *
-     * @param bool   $isEnabled is cache enabled
-     * @param string $cacheFile cache file
-     * @param bool   $linting   is linting enabled
-     * @param array  $rules     array defining rules, format like one for ConfigInterface::setRules
+     * @param CacheHandler $cacheHandler handles read/write operations
+     * @param string       $cacheFile    cache file
```
localheinz: @keradus 

It would probably be cool to get rid of the `$cacheFile` parameter here, but it is used for determining the path to file names relative to it.

src/FileCacheHandler.php
```diff
@@ -0,0 +1,56 @@
...
+namespace PhpCsFixer;
+
+use Symfony\Component\Filesystem\Exception\IOException;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+final class FileCacheHandler implements CacheHandler
```
localheinz: @keradus 

Maybe worth moving those into their own namespace, for example `PhpCsFixer\FileCache`?

localheinz:
Closing in favour of picking these changes into #1888.

PR #1891, opened 2016-04-29
Feature namespace fixer

aschwin:
I had the need for a fixer that automatically would replace existing namespaces after moving around files.

This is the result I came up with. It works for me, and I would guess others would benefit from this as well.

Go ahead and make any changes and/or improvements on the code and structure.


GrahamCampbell:
:-1: since such coding standards are already deprecated.

GrahamCampbell:
Also, doesn't our PSR-0 fixer already sort this out?

aschwin:
The PSR-0 fixer checks if the namespace matches the path and suggests to move it in case it doesn't match. It does not fix the namespace according to the path.

But, can you tell me which coding standards should prevent one from keeping the namespace in sync with the path?

aschwin:
Ok, better yet. If the PSR-0 is supposed to fix this, why is it not working? I applied several namespaces to fix and the PSR-0 fixer did not fix it.

keradus:
It should be done by PSR0/PSR4 fixer. No need for new fixer. Closing

Issue #1892, opened 2016-05-03
Files that were not fixed due to internal error


drecken:
.php_cs and file (with php extension in project) attached

```
<?php

$finder = Symfony\CS\Finder\DefaultFinder::create()
    ->exclude([
        'storage',
    ])
    ->in(__DIR__ . '/app');

return Symfony\CS\Config\Config::create()
    ->setUsingCache(true)
    ->level(Symfony\CS\FixerInterface::PSR2_LEVEL)
    ->fixers([
        'return',
        'lowercase_constants',
        'multiline_array_trailing_comma',
        'short_array_syntax',
    ])
    ->finder($finder);
```

[Timezone.txt](https://github.com/FriendsOfPHP/PHP-CS-Fixer/files/247430/Timezone.txt)

keradus:
please check this out: #1893

PR #1893, opened 2016-05-03
BracesFixer - handle comments inside lambda function prototype

keradus:
ref #1892

### DIFF
--- a/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
@@ -18,45 +18,6 @@ use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
  */
 class BracesFixerTest extends AbstractFixerTestBase
 {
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Does it break with random phpdoc there too?

keradus:
random phpdoc is closed explicity by */, so no. also, its not T_COMMENT

Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
```diff
@@ -1142,6 +1142,39 @@ function &($a, $b) use ($selfName) {
...
+    };',
+                '<?php
+    $fnc = function ($a, $b) # sort offsets by -, 0, +
+    {
+        return 0;
+    };',
+            ),
+            array(
+                '<?php
+    $fnc = function ($a, $b) /* sort offsets by -, 0, + */ {
```
GrahamCampbell: Could you add a test with `/**` please?

Issue #1895, opened 2016-05-17
Console formatting injection with --ansi and --diff on code with "<error>....</error>" etc.


ktomk:
When I run php-cs-fixer over code that contains symfony console component typical formatting strings and I output with the --ansi optionand --diff (in my case to review with `less -R`) it looks like that the source-code displayed in the diff output is getting replaced with colors when it contains color formatting codes.

~~~
$ php-cs-fixer --version
PHP CS Fixer version 1.11.2 by Fabien Potencier (41f7015)
~~~

Example command-line:
~~~
$ php-cs-fixer fix --config-file build/.php_cs --dry-run --diff --ansi | less -R
~~~

Exemplary output:

![selection_013](https://cloud.githubusercontent.com/assets/352517/15330706/4b9701be-1c5e-11e6-8d05-cc19d4836876.png)

Strings appear as empty, but the concrete code is:

~~~
    /**
     * @param OutputInterface $output
     * @param string $name
     * @param string $value
     * @param Constraints\Collection $constraints The constraint(s) to validate against.
     *
     * @return string
     */
    private function validateArgument(OutputInterface $output, $name, $value, $constraints)
    {
        if (strlen($value)) {
            $errors = $this->validateValue($name, $value, $constraints);
            if ($errors->count() > 0) {
                $output->writeln('<error>' . $errors[0]->getMessage() . '</error>');
            } else {
                return $value;
            }
        }

        $question = '<question>' . ucfirst($name) . ': </question>';

         ...
~~~

SpacePossum:
The fixer doesn't pick up the `--ansi` flag itself I think, so if this is not a default flag on the SF console we might/could/should add it. 
Besides the flag we should always not replace the output with the colors on the `--diff` output itself.
Thanks for reporting I hope to look into this soon.

ktomk:
Well the diff output has some nice colourings that look legit (sections headers, pluses and minuses), it was just these places highlighted in the screen-shot I was wondering about the colours. It took me a moment to understand that they mark no changes but were caused by an injection. For me this looks like a data-encoding problem so I reported it.

keradus:
1. we dont have `--ansi` flag
2. isn't it because of that `less` call ?

SpacePossum:
The `--ansi` is supported by default through the console package of SF.
Please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1896

ktomk:
[...]

PR #1896, opened 2016-05-18
FixCommand - Fix escaping of diff output

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1895#issuecomment-220003344

after
after

before
before

### DIFF
--- a/Symfony/CS/Console/Command/FixCommand.php
+++ b/Symfony/CS/Console/Command/FixCommand.php
@@ -13,6 +13,7 @@
 namespace Symfony\CS\Console\Command;
 
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
@@ -404,8 +405,9 @@ EOF
             $this->eventDispatcher->addListener(FixerFileProcessedEvent::NAME, $fileProcessedEventListener);
         }
 
+        $isDiff = $input->getOption('diff');
         $this->stopwatch->start('fixFiles');
-        $changed = $this->fixer->fix($config, $input->getOption('dry-run'), $input->getOption('diff'));
+        $changed = $this->fixer->fix($config, $input->getOption('dry-run'), $isDiff);
         $this->stopwatch->stop('fixFiles');
 
         if ($showProgress) {
@@ -440,10 +442,32 @@ EOF
                         $output->write(sprintf($fixerDetailLine, implode(', ', $fixResult['appliedFixers'])));
                     }
 
-                    if ($input->getOption('diff')) {
+                    if ($isDiff) {
                         $output->writeln('');
                         $output->writeln('<comment>      ---------- begin diff ----------</comment>');
-                        $output->writeln($fixResult['diff']);
+
+                        if ($output->isDecorated()) {
+                            $diff = implode(
+                                PHP_EOL,
+                                array_map(
+                                    function ($string) {
+                                        $string = preg_replace('/^(\+){3}/', '<info>+++</info>', $string);
+                                        $string = preg_replace('/^(\+){1}/', '<info>+</info>', $string);
+                                        $string = preg_replace('/^(\-){3}/', '<error>---</error>', $string);
+                                        $string = preg_replace('/^(\-){1}/', '<error>-</error>', $string);
+                                        $string = str_repeat(' ', 6).$string;
+
+                                        return $string;
+                                    },
+                                    explode(PHP_EOL, OutputFormatter::escape($fixResult['diff']))
+                                )
+                            );
+
+                            $output->writeln($diff);
+                        } else {
+                            $output->writeln($fixResult['diff'], OutputInterface::OUTPUT_RAW);
+                        }
+
                         $output->writeln('<comment>      ---------- end diff ----------</comment>');
                     }
 
@@ -489,7 +513,7 @@ EOF
                         }
                     }
 
-                    if ($input->getOption('diff')) {
+                    if ($isDiff) {
                         $diffXML = $dom->createElement('diff');
                         $diffXML->appendChild($dom->createCDATASection($fixResult['diff']));
                         $fileXML->appendChild($diffXML);
@@ -497,7 +521,7 @@ EOF
                 }
 
                 $dom->formatOutput = true;
-                $output->write($dom->saveXML());
+                $output->write($dom->saveXML(), false, OutputInterface::OUTPUT_RAW);
                 break;
             case 'json':
                 $jFiles = array();
@@ -509,7 +533,7 @@ EOF
                         $jfile['appliedFixers'] = $fixResult['appliedFixers'];
                     }
 
-                    if ($input->getOption('diff')) {
+                    if ($isDiff) {
                         $jfile['diff'] = $fixResult['diff'];
                     }
 
@@ -540,7 +564,7 @@ EOF
                     $json['time']['files'] = $jFileTime;
                 }
 
-                $output->write(json_encode($json));
+                $output->write(json_encode($json), false, OutputInterface::OUTPUT_RAW);
                 break;
         }
 
--- a/Symfony/CS/Fixer.php
+++ b/Symfony/CS/Fixer.php
@@ -312,29 +312,17 @@ class Fixer
         return 'symfony';
     }
 
+    /**
+     * @deprecated Will be removed in the 2.0.
+     *
+     * @param string $old
+     * @param string $new
+     *
+     * @return string
+     */
     protected function stringDiff($old, $new)
     {
-        $diff = $this->diff->diff($old, $new);
-
-        $diff = implode(
-            PHP_EOL,
-            array_map(
-                function ($string) {
-                    $string = preg_replace('/^(\+){3}/', '<info>+++</info>', $string);
-                    $string = preg_replace('/^(\+){1}/', '<info>+</info>', $string);
-
-                    $string = preg_replace('/^(\-){3}/', '<error>---</error>', $string);
-                    $string = preg_replace('/^(\-){1}/', '<error>-</error>', $string);
-
-                    $string = str_repeat(' ', 6).$string;
-
-                    return $string;
-                },
-                explode(PHP_EOL, $diff)
-            )
-        );
-
-        return $diff;
+        return $this->diff->diff($old, $new);
     }
 
     /**
--- a/Symfony/CS/Tests/Console/TextDiffTest.php
+++ b/Symfony/CS/Tests/Console/TextDiffTest.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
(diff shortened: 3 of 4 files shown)


Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,48 @@
...
+            ),
+            array(
+                'decorated' => true,
+                'verbosity' => OutputInterface::VERBOSITY_DEBUG,
+            )
+        );
+
+        $output = $commandTester->getDisplay(true);
+
+        echo $output;
```
SpacePossum: oh maybe I should test the output here and not echo it... will fix later ^-^

ktomk:
Nice, thankfully **OutputFormatter** now supports escaping!

Symfony/CS/Fixer.php
```diff
@@ -312,31 +312,6 @@ public static function getLevelAsString(FixerInterface $fixer)
         return 'symfony';
     }
 
-    protected function stringDiff($old, $new)
```
SpacePossum: tech. speaking this is a BC break, however it is not a useful service point anyway and the functionality of the method itself has been changed as well. @keradus let me what you here.

SpacePossum:
Once we are happy here with this change I can prep. a version for master as well.

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -434,10 +437,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
                     }
 
-                    if ($input->getOption('diff')) {
+                    if ($isDiff) {
                         $output->writeln('');
                         $output->writeln('<comment>      ---------- begin diff ----------</comment>');
-                        $output->writeln($fixResult['diff']);
+
+                        $diff = $fixResult['diff'];
+                        if ($isDecorated) {
```
GrahamCampbell: the contents of this should probably be abstracted to their own method

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $command = new FixCommand();
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array(
+                'path' => __DIR__.'/TextDiffTestInput.test',
+                '--fixers' => 'concat_without_spaces',
+                '--dry-run' => true,
+                '--diff' => true,
+                //'--format' => 'json'
```
GrahamCampbell: ?

Symfony/CS/Console/Command/FixCommand.php
keradus: on 2.x it's onw class

Symfony/CS/Fixer.php
keradus: but BC is BC, it's not internal class.

ktomk:
@SpacePossum: Just ping me when you're confident, then I can give this a test-drive for direct feedback in my original scenario.

SpacePossum:
I propose to skip the test op Appveyor/Windows because all it adds is testing if the SF console component works correctly on it which is not what this test is for.
Trying to fix the test to pass on all WIndows versions used by Appveyor is a lot of work with no gain and would require a lot maintenance. 

@ktomk this should work, any reports/test results you can provide are welcome :+1:

keradus:
So, this won't work properly on Windows ?

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,76 @@
...
+        if ($decorated !== $commandTester->getOutput()->isDecorated()) {
+            $this->markTestSkipped(sprintf('Output should %sbe decorated.', $decorated ? '' : 'not'));
+        }
+
+        $this->assertStringMatchesFormat($format, $commandTester->getDisplay(true));
+    }
+
+    public function provideDiffReporting()
+    {
+        $r = array(
```
keradus: please, avoid one-char variables

SpacePossum:
>So, this won't work properly on Windows ?

It will; depending on the SF console component. The issue here is testing only. 

In the test I check what we output to the console. If the console supports colors we should escape the diff output itself to prevent the console component from replacing content in the diff. to color characters.
Which characters the console component uses to decorate the output is not of our concern, so testing one env. is good enough.

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -418,10 +420,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         }
 
         $i = 1;
-
```
GrahamCampbell: ?

SpacePossum: ? github differ slitch

SpacePossum: differ glitch

GrahamCampbell: no, that line has actually been deleted?

SpacePossum: oh, let me put it back, thanks

Symfony/CS/Console/Command/FixCommand.php
GrahamCampbell: :+1:

keradus:
This is why I don't like that exception.


It doesn't matter will I use standard System console like `cmd` / `Power Shell` or not, like `cmder`.

SpacePossum:
I start to suspect the decorating detecting of the output is not always done on CommandTests.
I'll see if can rewrite the expect part of the test, to bad as the fix seems to work.

keradus:
How could I check if fix indeed works on Windows?

ktomk:
As long as your Windows is compatible with the color codes, there should be no issue. The tests are most likely not well reflecting the changes, the coloring is done by the Symfony component which is well tested, point in case for this fix is to escape the code output properly so that it does not trigger any of the Symfony console color codes (that are acutally tags).

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,76 @@
...
+                false,
+                "%a-        \$output->writeln('' . 'a' . '');
++        \$output->writeln(''.'a'.'');%a",
+            ),
+        );
+
+        if (!getenv('APPVEYOR')) {
+            $cases[] = array(
+                true,
+                "%aTextDiffTestInput.test%a\033[%s-\033[%sm        \$output->writeln('<error>' . 'a' . '</error>');
```
keradus: maybe prepare this by OutputInterface instead of hardcoding the colors and it should be fine for both system?

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        $cases = array(
+            array(
+                false,
+                "%a-        \$output->writeln('' . 'a' . '');
++        \$output->writeln(''.'a'.'');%a",
+            ),
+        );
+
+        if (!getenv('APPVEYOR')) {
```
GrahamCampbell: Maybe it would be better to check the OS so we can run these tests on other systems that arn't CI systems?

keradus: or  make tests system independent

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -434,10 +436,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                                        $string = str_repeat(' ', 6).$string;
+
+                                        return $string;
+                                    },
+                                    explode(PHP_EOL, $diff)
+                                )
+                            );
+                        }
+
+                        $output->writeln($diff);
```
SpacePossum: still not sure how this could work for non-decorated consoles, the output does not remove escape characters (see line 440) when in non-decorated mode.

SpacePossum:
All cases covered, should ran in all consoles on all platforms.
Feedback welcome from people using a diff. console than Ubuntu default.

GrahamCampbell:
Much better. :+1:

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,80 @@
...
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class TextDiffTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @param string $expected
+     * @param string $format
+     * @param bool   $decorated
```
keradus: $isDecorated

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,80 @@
...
+                '--format' => $format,
+            ),
+            array(
+                'decorated' => $decorated,
+                'verbosity' => OutputInterface::VERBOSITY_NORMAL,
+            )
+        );
+
+        if ($decorated !== $commandTester->getOutput()->isDecorated()) {
+            $this->markTestSkipped(sprintf('Output should %sbe decorated.', $decorated ? '' : 'not'));
```
keradus: `%sbe` => `notbe`

Symfony/CS/Tests/Console/TextDiffTestInput.test
```diff
@@ -0,0 +1,26 @@
+<?php
```
keradus: Why not inside fixture dir ?

SpacePossum: I tried the fixture, but if you run this unit test on master from the root directory of the project the file gets excluded because of the intersection made based on the configuration file in the root directory (it was also excluded because it ended with `.test` I think).
I think that was why I moved it here, but indeed it is not needed. I'll refactor soon and process your other points as well. Thanks for reviewing :+1:

keradus:
so red

Symfony/CS/Tests/Console/TextDiffTest.php
```diff
@@ -0,0 +1,80 @@
...
+                'decorated' => $isDecorated,
+                'verbosity' => OutputInterface::VERBOSITY_NORMAL,
+            )
+        );
+
+        if ($decorated !== $commandTester->getOutput()->isDecorated()) {
+            $this->markTestSkipped(sprintf('Output should %sbe decorated.', $isDecorated ? '' : 'not '));
+        }
+
+        if ($decorated !== $commandTester->getOutput()->getFormatter()->isDecorated()) {
```
keradus: **is** is missing

SpacePossum:
..and green again, but yes I still have to move the file which is no fun trying to do using the github interface so I'll do it when Im back at some dev. setup ;)  (flagged WIP for the reason)

keradus:
All 6 new assertions were properly executed on my Win10 env with cmd, powershell and cmder consoles.
Don't focus on that file movement.
Please, prepare version for 2.x so I will use it during merging 1.12->master.
Then, I will merge this PR.

keradus:
Thank you @SpacePossum.

PR #1898, opened 2016-05-19
Use PHP_BINARY constant for syntax checking

Baldinof:
Currently the `LintManager` class check syntax by running `php -l path/`, it can cause some errors if the php binary has another name.


keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Linter/Linter.php

Baldinof:
@keradus sorry :smile:

PR #1231, opened 2015-05-28
Allow multiple files on fix command

Soullivaneuh:
PR related to #1230.

It's a proposition and I'm not sure of the good way to do this. Test may fail but it's not finished and I need some opinions.

Don't hesitate to give comment on my code for any mistake I could make! :+1: 

For the moment, php-cs-fixer fail with a regular file argument:

```bash
$ ./php-cs-fixer fix --level=symfony --verbose Symfony/CS/Console/ Symfony/CS/Finder/ Symfony/CS/Fixer/PSR0/Psr0Fixer.php 


                                                                                                                 
  [InvalidArgumentException]                                                                                     
  The "/home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Fixer/PSR0/Psr0Fixer.php" directory does not exist.  
                                                                                                                 


Exception trace:
 () at /home/sullivan/projects/fork/PHP-CS-Fixer/vendor/symfony/finder/Symfony/Component/Finder/Finder.php:677
 Symfony\Component\Finder\Finder->in() at /home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Finder/DefaultFinder.php:48
 Symfony\CS\Finder\DefaultFinder->setDir() at /home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Config/Config.php:89
 Symfony\CS\Config\Config->getFinder() at /home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Fixer.php:170
 Symfony\CS\Fixer->fix() at /home/sullivan/projects/fork/PHP-CS-Fixer/Symfony/CS/Console/Command/FixCommand.php:375
 Symfony\CS\Console\Command\FixCommand->execute() at /home/sullivan/projects/fork/PHP-CS-Fixer/vendor/symfony/console/Symfony/Component/Console/Command/Command.php:257
 Symfony\Component\Console\Command\Command->run() at /home/sullivan/projects/fork/PHP-CS-Fixer/vendor/symfony/console/Symfony/Component/Console/Application.php:874
 Symfony\Component\Console\Application->doRunCommand() at /home/sullivan/projects/fork/PHP-CS-Fixer/vendor/symfony/console/Symfony/Component/Console/Application.php:195
 Symfony\Component\Console\Application->doRun() at /home/sullivan/projects/fork/PHP-CS-Fixer/vendor/symfony/console/Symfony/Component/Console/Application.php:126
[...]

PR #1903, opened 2016-05-22
FixCommand - allow to pass multiple path argument

### LINKED ISSUE #1230: Proposal: Multiples files parameters
Could be awesome to run php-cs-fixer on some selected files like this:

```bash
php-cs-fixer fix src/path/to/file01.php src/path/to/file02.php src/path/to/file03.php 
```

Any chance to get this feature?

Thanks.

keradus:
Closes #1230
Replaces #1231

I still need to update readme file.

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -14,13 +14,13 @@ namespace PhpCsFixer\Console;
 
 use PhpCsFixer\ConfigInterface;
 use PhpCsFixer\ConfigurationException\InvalidConfigurationException;
+use PhpCsFixer\Finder;
 use PhpCsFixer\FixerFactory;
 use PhpCsFixer\FixerInterface;
 use PhpCsFixer\RuleSet;
 use PhpCsFixer\StdinFileInfo;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Component\Finder\Finder as SymfonyFinder;
-use Symfony\Component\Finder\SplFileInfo as SymfonySplFileInfo;
 
 /**
  * The resolver that resolves configuration to use by command line options and config.
@@ -91,7 +91,7 @@ final class ConfigurationResolver
         'config' => null,
         'dry-run' => null,
         'format' => 'txt',
-        'path' => null,
+        'path' => array(),
         'progress' => null,
         'using-cache' => null,
         'cache-file' => null,
@@ -162,7 +162,7 @@ final class ConfigurationResolver
     /**
      * Returns path.
      *
-     * @return string
+     * @return string[]
      */
     public function getPath()
     {
@@ -314,13 +314,15 @@ final class ConfigurationResolver
         }
 
         $path = $this->path;
-        if (is_file($path) && $dirName = pathinfo($path, PATHINFO_DIRNAME)) {
-            $configDir = $dirName;
-        } elseif ($this->isStdIn || null === $path) {
+
+        if ($this->isStdIn || 0 === count($path)) {
             $configDir = $this->cwd;
-            // path is directory
+        } elseif (1 < count($path)) {
+            throw new InvalidConfigurationException('For multiple paths config parameter is required.');
+        } elseif (is_file($path[0]) && $dirName = pathinfo($path[0], PATHINFO_DIRNAME)) {
+            $configDir = $dirName;
         } else {
-            $configDir = $path;
+            $configDir = $path[0];
         }
 
         $candidates = array(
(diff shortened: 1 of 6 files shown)


keradus:
Travis is dead ?

src/Console/ConfigurationResolver.php
```diff
@@ -401,52 +403,69 @@ private function resolveConfigPath()
...
-        $path = realpath($this->path);
+        $paths = array_filter(array_map(
+            function ($path) {
+                return realpath($path);
+            },
+            $this->path
+        ));
 
-        if (false === $path) {
+        if (empty($paths)) {
```
SpacePossum: This is no longer possible I think?
0 can only be with `stdIn` which is covered at line 400?

src/Console/ConfigurationResolver.php
```diff
@@ -401,52 +403,69 @@ private function resolveConfigPath()
             return;
         }
 
-        if (null === $this->path) {
+        if (empty($this->path)) {
```
SpacePossum: This is no longer possible I think?
0 length can only be with `stdIn` which is covered at line 400?

keradus: `$this->path` could be an empty array when no `path` argument was passed via CLI

src/Console/ConfigurationResolver.php
```diff
@@ -314,13 +314,15 @@ private function computeConfigFiles()
...
         $path = $this->path;
-        if (is_file($path) && $dirName = pathinfo($path, PATHINFO_DIRNAME)) {
-            $configDir = $dirName;
-        } elseif ($this->isStdIn || null === $path) {
+
+        if ($this->isStdIn || 0 === count($path)) {
             $configDir = $this->cwd;
-            // path is directory
+        } elseif (1 < count($path)) {
+            throw new InvalidConfigurationException('For multiple paths config parameter is required.');
```
mnapoli: Hi there, I cannot understand why this exception is thrown?

Basically when I try to run the fixer with multiple paths, it complains that I don't have a config. Why would I need a config in that case?

Thanks if you have any answer for that. I could work on a PR if that's a bug, but before I would like to understand.

keradus: It was a design decision.
Basically, if you provide a path but don't provide a config file, it's localization is automatically discovered. if you provide multiple path, autodiscovering is extra tricky and could lead to discover multiple config files, while tool can only use one config per run.

mnapoli: Thanks for the answer! That's not really in my interest ;) but at least I understand now.

To give some context I'm working on https://github.com/mnapoli/sniff What I want to do is basically the same thing as `.php_cs` except with a JSON file. My current approach is maybe not the best but at least I made it work, I run this:

```
php-cs-fixer path1 path2 … --config=null-config-file.php --rules=...
```

And here is the "null config file": https://github.com/mnapoli/sniff/blob/master/php-cs-null-config.php

keradus: yep, just specify the config. i would even rename the file into standard `.php_cs.dist`

mnapoli: Yes but projects using `sniff` should have a `.sniff.json` instead of `.php_cs`, which is why they don't have a config (but may have multiple paths).

That's why I wrote a "null config file" which returns an empty config: it's just so that I don't get the error *For multiple paths config parameter is required*.

Anyway I don't want to bore you with all that because that's my fault: I'm using php-cs-fixer in a weird way ;)

mfn: I'm coming from https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4024#issuecomment-750030447 and I didn't want to derail the other issue which is about something else.

> you do not need to change this line, just point which config file to use via cli parameter

Apologies @keradus , I couldn't follow this one.

> it's localization is automatically discovered

I also don't understand this part 😅 "localization" => text translation? 🤔 Nah, doesn't make sense.

---
So maybe this is about ambiguity of resolving possible multiple config files which isn't support, in case multiple files are provided to the fixer?

Because at least the `--help` output would signal to me multiple "paths" are supported:
```
Usage:
  fix [options] [--] [<path>...]
```

I see that there is a dedicated `--config` and usually I see a fallback to `.php_cs.dist`/`.php_cs`. Apologies, I don't quite understand. And the reason I'm asking, I think it would a useful to simple provide multiple files on the command line to fix.

Thank you for your patience with me :)

mfn: Was answered in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4024#issuecomment-750288219
> possible since creation of this github issue. just explicitly fingerpoint the config that should be used.

… and in fact like the message says, just use something like this; "ops":
`find … | xargs -n … -P … ./php-cs-fixer.phar --config=…`

Issue #1900, opened 2016-05-19
[master] Run on single file doesn't allow for all types and performance can be tweaked


SpacePossum:
For example:

```
/home/possum/work/PHP-CS-Fixer$ php php-cs-fixer fix tests/Fixtures/Integration/priority/no_useless_return,no_whitespace_in_blank_lines.test -vvv
```
Causes the iterator to check if the following paths are in fact `tests/Fixtures/Integration/priority/no_useless_return,no_whitespace_in_blank_lines.test`

```
[edit removed after comments]
```

which seems like an overkill since the one file to look for was given.

But more of an issue, if you tell the fixer explicitly to fix the `.test` file it won't do it because of the default Finder settings.

keradus:
1. as it's now building intersection, right?
2. to filter the extension by Finder, the Finder must be used.

SpacePossum:
1. so this is the results because there is a config file at the run location? ok, than it is a UX call, no issue than I guess?
2. can't we tweak the finder to always add the file given by the command to the result? I mean if the user provides it why not fix the file right?

keradus:
pseudocode:
foreach (modifiedFiles as file): php-cs-fixer fix file

so far, if the file is excluded by config then it's not fixed.
after your proposal it would, and that would actually revert whole intersection thing

SpacePossum:
>after your proposal it would, and that would actually revert whole intersection thing

That is not my intention, I understand now, I drop point 1 :)

But how about point 2, if the file is not dropped out by the intersection it can still be filtered out by extension. Would you allow the change that would be, in pseudo;
if the user passes a file by the commandline and it is not filtered by intersect, fix it no matter the extension

keradus:
actually finder filtering different things. Symfony one could filter by path, ext, content, size... 
I wouldn't like to introduce exceptions

SpacePossum:
I think the exceptions should be done on config level for projects like Symfony and not the other way around.
Now if one installs the fixer it is impossible to fix a file called `myFile.test` without changing the extension of the file or by writing a custom Finder in a configuration file.
I'm not asking to change the default behavior to start fixing `.test` files, just if some passes the file by the commandline to honor this and fix the file.

keradus:
Not sure is it that good we should go that way or not, what would you think:

[...]

PR #1910, opened 2016-05-25
[Master][Bug] Fix diff. output escaping

SpacePossum:
master version of;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1896

~~please note this is still **WIP**~~

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -106,6 +106,7 @@ final class ConfigurationResolver
     private $usingCache;
     private $cacheFile;
     private $ruleSet;
+    private $formats = array();
 
     public function __construct()
     {
@@ -431,8 +432,6 @@ final class ConfigurationResolver
         );
 
         foreach ($paths as $path) {
-            $isFile = is_file($path);
-
             if (is_file($path)) {
                 $pathsByType['file'][] = $path;
             } else {
@@ -514,8 +513,6 @@ final class ConfigurationResolver
 
     protected function resolveFormat()
     {
-        static $formats = array('txt', 'xml', 'json');
-
         if (array_key_exists('format', $this->options)) {
             $format = $this->options['format'];
         } elseif (method_exists($this->config, 'getFormat')) {
@@ -524,8 +521,8 @@ final class ConfigurationResolver
             $format = 'txt'; // default
         }
 
-        if (!in_array($format, $formats, true)) {
-            throw new InvalidConfigurationException(sprintf('The format "%s" is not defined, supported are %s.', $format, implode(', ', $formats)));
+        if (!in_array($format, $this->formats, true)) {
+            throw new InvalidConfigurationException(sprintf('The format "%s" is not defined, supported are %s.', $format, implode(', ', $this->formats)));
         }
 
         $this->format = $format;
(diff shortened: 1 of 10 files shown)


tests/TextDiffTest.php
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * @dataProvider provideDiffReporting
+     */
+    public function testDiffReportingDecorated($expected, $format, $isDecorated)
+    {
+        $command = new FixCommand();
+        $commandTester = new CommandTester($command);
+        $commandTester->execute(
+            array(
+                'path' => array(__DIR__.'/Fixtures/FixCommand/TextDiffTestInput.php'),
```
SpacePossum: the tests fail because;
1 - fix is not implemented
2 - file is not checked in the first place, @see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1900

src/Console/ConfigurationResolver.php
```diff
@@ -547,7 +546,6 @@ private function resolveIsStdIn()
     private function resolvePath()
     {
         $filesystem = new Filesystem();
-        $path = $this->options['path'];
```
keradus: nice catch!

SpacePossum:
**fabbot** is wrong on this one as it tries to fix a file in the `fixtures` directory.
If the other tests are green I think this is done

keradus:
Travis is red

SpacePossum:
>Travis is red

same issue as here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1919 (reported on gitter)
no idea what is wrong, but it seems related to the fixer caching files and hhvm,i.e. not related to the PRs

keradus:
It looks like we introduced invalid code snippet as test case.
Please, could you (or anyone willing) take care about this (in separated PR) ?

SpacePossum:
after your @keradus it is all green :) thanks! :+1: 
(not all, fabbot is still wrong but ey ;) )

src/Report/TextReporter.php
```diff
@@ -84,22 +82,28 @@ private function getDiff($isDecoratedOutput, array $fixResult)
...
         if ($isDecoratedOutput) {
-            $template .= '<comment>      ---------- begin diff ----------</comment>';
-            $template .= PHP_EOL.'%s'.PHP_EOL;
-            $template .= '<comment>      ----------- end diff -----------</comment>';
+            $template = '<comment>      ---------- begin diff ----------</comment>%s<comment>      ----------- end diff -----------</comment>';
+            $diff = implode(
+                PHP_EOL,
+                array_map(
+                    function ($string) {
+                        $string = preg_replace('/^(\+){3}/', '<info>+++</info>', $string);
```
keradus: Not sure about this.
This is SebastianBergmann diff library specific changes and shouldn't be applied for all (already existing or not yet) Differs.
What do you thinkg @SpacePossum ?

SpacePossum: Hmmm indeed, we should be OK for now but I see your point.
Are these string `diff` formats (like the one used on default by `git`) formalized in a way?

keradus: I don't think so, that why I moved the logic outside.
If some of them will share some logic we should deal with it on Differ level, not Reporter.

SpacePossum: But the reporters know about the output stream they are generation for, the differs don't.
I'll give it so more thought

keradus:
Thank you very much @SpacePossum 
branch used to merge #1896 to master

PR #1911, opened 2016-05-25
typo fix

OskarStark:


GrahamCampbell:
Please regen the readme too.

keradus:
BTW, is it a typo indeed?

> The meaning of “everything but” is quite literal; it means “everything except”, “everything with the exception of”. For example:
> 
> ```
> I eat everything but meat.
> She wants everything but happiness for other people.
> ```
> 
> which means
> 
> ```
> I eat everything except meat.
> She wants everything with the exception of happiness for other people.
> ```

GrahamCampbell:
Yeh, this actually isn't a typo. Both would be valid English though. Before the PR, the word `but` took the meaning `except`, but after, that was replaced with `but not`.

PR #1913, opened 2016-05-25
Introduce path-mode CLI option

### LINKED ISSUE #1900: [master] Run on single file doesn't allow for all types and performance can be tweaked

keradus:
Closes #1900

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -899,9 +899,9 @@ Then, add the following command to your CI:
 
 .. code-block:: bash
 
-    $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist `git diff --name-only $COMMIT_RANGE`
+    $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist --path-mode=intersection `git diff --name-only $COMMIT_RANGE`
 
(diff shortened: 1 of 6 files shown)


SpacePossum:
:+1:  nice work : D

keradus:
Thanks ;)
the only thing missing since night was default option value ;) (That's why CIs were red)

Have you only review it, or play with it as well? (just a question)

SpacePossum:
just a quick review, I'm rebasing my work on top of your branch and was checking of it was my test or your change : D

PR #1921, opened 2016-05-27
Strict types declaration fixer

Seldaek:
This adds a fixer that ensures all files start with

```
<?php declare(strict_types=1);

// ...
```

I also had to adjust the header_comment fixer because it shouldn't insert a header above the declare IMO.

The only problem I have is that the tests fail with:

```
4) PhpCsFixer\Tests\Fixer\Strict\StrictTypesFixerTest::testFixAddsToEmptyFile
The token at index 5 must be {
    "id": 317,
    "name": "T_LNUMBER",
    "content": 1,
    "isArray": true,
    "changed": false
}, got {
    "id": 317,
    "name": "T_LNUMBER",
    "content": 1,
    "isArray": true,
    "changed": true
}
```

I can't make sense of it though, as the code works and does the right thing but only this changed flag is true when it shouldn't or something. I don't know enough about internals to know how to fix this. Advice would be nice.

Another improvement I wanted to make is to allow a blacklist and/or whitelist to be configured on the fixer, like an array of files/regex to ignore or include, because this isn't strictly speaking coding style, it has runtime impact and you don't necessarily want to enforce this on ALL files. However isCandidate doesn't receive the file path, only the tokens, so it's hard to make a decision about the file at that point. I could do it in fix() I guess but wanted to ask first.

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -70,7 +70,7 @@ final class HeaderCommentFixer extends AbstractFixer
             return;
         }
 
-        $this->replaceHeaderComment($tokens, $oldHeaderIndex);
+        $this->replaceHeaderComment($tokens, $oldHeaderIndex, $newHeaderIndex);
     }
 
     /**
@@ -101,7 +101,7 @@ final class HeaderCommentFixer extends AbstractFixer
     }
 
     /**
-     * Find the header comment index.
+     * Find the closest whitespace block before the header comment index.
      *
      * @param Tokens $tokens
      *
@@ -110,14 +110,17 @@ final class HeaderCommentFixer extends AbstractFixer
     private function findHeaderCommentIndex(Tokens $tokens)
     {
         $index = $tokens->getNextNonWhitespace(0);
+        if (null !== $index && $tokens[$index]->isGivenKind(T_DECLARE)) {
+            $index = $this->skipDeclare($tokens, $index);
+        }
 
         if (null !== $index && $tokens[$index]->isGivenKind(T_COMMENT)) {
-            return $index;
+            return $index === 2 && $tokens[1]->isWhitespace() ? 1 : $index;
         }
     }
 
     /**
-     * Find the index where the header comment must be inserted.
+     * Find the closest whitespace block where the header comment must be inserted.
      *
      * @param Tokens $tokens
      *
@@ -126,13 +129,46 @@ final class HeaderCommentFixer extends AbstractFixer
     private function findHeaderCommentInsertionIndex(Tokens $tokens)
     {
         $index = $tokens->getNextNonWhitespace(0);
+        if (null !== $index && $tokens[$index]->isGivenKind(T_DECLARE)) {
+            $index = $this->skipDeclare($tokens, $index);
+        }
 
         if (null === $index) {
             // empty file, insert at the end
             $index = $tokens->getSize();
         }
 
-        return $index;
+        return $index === 2 && $tokens[1]->isWhitespace() ? 1 : $index;
+    }
+
+    /**
+     * Skips a declare(strict_type=1); statement as it is allowed to be before the header
+     *
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return int
+     */
+    private function skipDeclare(Tokens $tokens, $index)
+    {
+        $isOpen = false;
+        $isStrictTypes = false;
+        $originalIndex = $index;
+
+        while ($index = $tokens->getNextNonWhitespace($index)) {
+            if ($tokens[$index]->getContent() === '(') {
+                $isOpen = true;
+            }
+            if ($isOpen && $tokens[$index]->getContent() === 'strict_types') {
+                $isStrictTypes = true;
+            }
+            if ($tokens[$index]->getContent() === ';') {
+                $index++;
+                break;
+            }
+        }
+
+        return $isStrictTypes ? $index : $originalIndex;
     }
 
     /**
@@ -140,29 +176,34 @@ final class HeaderCommentFixer extends AbstractFixer
      *
      * @param Tokens $tokens
      * @param int    $oldHeaderIndex
+     * @param int    $headerInsertionIndex
      */
-    private function replaceHeaderComment(Tokens $tokens, $oldHeaderIndex)
+    private function replaceHeaderComment(Tokens $tokens, $oldHeaderIndex, $headerInsertionIndex)
     {
+        $headerEnd = null !== $oldHeaderIndex
+            ? $tokens->getNextNonWhitespace($oldHeaderIndex)
+            : $headerInsertionIndex - 1
+        ;
+
+        while (isset($tokens[$headerEnd + 1]) && ($tokens[$headerEnd + 1]->isWhitespace() || $tokens[$headerEnd + 1]->isGivenKind(T_COMMENT))) {
+            $headerEnd++;
+        }
+
         if ('' === $this->headerComment) {
             if ($oldHeaderIndex) {
-                $tokens->clearRange($oldHeaderIndex, $oldHeaderIndex + 1);
+                $tokens->clearRange($tokens->getNextNonWhitespace($oldHeaderIndex), $headerEnd);
             }
 
             return;
         }
 
         $headCommentTokens = array(
-            new Token(array(T_WHITESPACE, "\n")),
+            new Token(array(T_WHITESPACE, $headerInsertionIndex === 1 ? "\n" : "\n\n")),
             new Token(array(T_COMMENT, $this->headerComment)),
             new Token(array(T_WHITESPACE, "\n\n")),
         );
 
-        $newHeaderIndex = null !== $oldHeaderIndex
-            ? $oldHeaderIndex + 1
-            : $this->findHeaderCommentInsertionIndex($tokens) - 1
-        ;
-
-        $tokens->overrideRange(1, $newHeaderIndex, $headCommentTokens);
+        $tokens->overrideRange($headerInsertionIndex, $headerEnd, $headCommentTokens);
     }
 
     /**
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -142,7 +142,7 @@ final class HeaderCommentFixer extends AbstractFixer
     }
 
     /**
-     * Skips a declare(strict_type=1); statement as it is allowed to be before the header
+     * Skips a declare(strict_type=1); statement as it is allowed to be before the header.
      *
      * @param Tokens $tokens
      * @param int    $index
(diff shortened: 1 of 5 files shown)


src/Fixer/Strict/StrictTypesFixer.php
```diff
@@ -0,0 +1,106 @@
...
+            return;
+        }
+
+        $declareTokens = array(
+            new Token(array(T_OPEN_TAG, '<?php ')),
+            new Token(array(T_DECLARE, 'declare')),
+            new Token('('),
+            new Token(array(T_STRING, 'strict_types')),
+            new Token('='),
+            new Token(array(T_LNUMBER, 1)),
```
SpacePossum: The `1` is the problem of the failing test. The content should be a string, i.e. change it to `'1'` and you should be OK :)

I'll add a note to my TODO list to make the output of the failing test better (i.e. it would be nice if the output of the `toJson` method of `Token` would show `'1'` as content and not `1`)

Seldaek: OK thanks, fixed.

tests/Fixer/Strict/StrictTypesFixerTest.php
```diff
@@ -0,0 +1,105 @@
...
+    {
+        $input = "<?php\nphpinfo();\n?>\n<?";
+        $this->doTest($input);
+    }
+
+    public function testFixDoNotTouchFilesNotStartingWithOpenTag()
+    {
+        $input = " <?php\nphpinfo();\n";
+        $this->doTest($input);
+    }
```
SpacePossum: Would be nice to have a test with a file with header, both in PHPDoc and comment block and using both. Just to see if and where the `declare` gets inserted.
Have a test for short tag open would be nice (not that short tag is nice, but it is not removed from PHP 7).

Seldaek: Added a few tests to confirm all that.

GrahamCampbell:
Could we have an option to have it like this please:

image

GrahamCampbell:
More complete example of my prefered CS:

image

Seldaek:
I knew this'd come up, but I'd say please no. Believe me you'll get over it soon enough and this avoids shuffling every single file 2 lines below when opened. Besides if you use this to ensure the whole project is in strict mode then hiding it up there makes sense IMO, as you know it's there.

GrahamCampbell:
Could you add test cases showing this moves this up to the correct place if it already exists please?

GrahamCampbell:
I'd still rather we could allow both standards though.

Seldaek:
It's already there https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1921/files#diff-effdfb27184a14e565c6d45e17fea171R22

GrahamCampbell:
:)

SpacePossum:
the detection and moving is not always correct, for example.
Input:
```php
<?php

            /*test*/declare(strict_types=1)/**/;echo 123;
```
gets fixed to:
```php
<?php declare(strict_types=1);

/*test*/declare(strict_types=1)/**/;echo 123;
```

(with regards of the PHP horror coders ^-^)

Seldaek:
Ah indeed, it only detects it at the beginning, it doesn't scan the whole file for it. That'll make things a lot slower I guess, but I can do it if you think it's needed.

SpacePossum:
It is sad, however we can break valid code (no matter how bad written it is) unless we mark the fixer `risky`.

The scan only needs to start at the index of `$tokens->getNextMeaningfulToken(0)`.
Since the first token must be an open tag (`isMonolithicPhp` ensures that). The next meaningful must be the `declare`, or otherwise a `declare` in the file is invalid (and doesn't pass linting), so you can stop the scan there.
If the `null === nextMeaningFulIndex` than it is an empty file, so there is a little short cut for that edge case.

SpacePossum:
and please add a test for this case as well:
```php
<? //test
  declare      (
strict_types  =1
)




?>
test
```

Seldaek:
OK I added a fix for the one appearing after some comments, and added a test for that last one you mentioned but IMO those tests are kinda useless as it's all non-monolithic PHP so it doesn't even get parsed.

SpacePossum:
@Seldaek thanks for all the work!
Please consider https://github.com/Seldaek/PHP-CS-Fixer/pull/1/files for the StrrictTypeFixer.
I've modified the fixer so it can handle more cases.

src/Fixer/Strict/StrictTypesFixer.php
```diff
@@ -0,0 +1,170 @@
...
+            return;
+        }
+
+        // found, fix casing
+        $this->fixSequenceCasing($sequenceLocation);
+
+        end($sequenceLocation);
+        $sequenceEndIndex = $tokens->getNextMeaningfulToken(key($sequenceLocation));
+        if ($tokens[$sequenceEndIndex]->isGivenKind(T_CLOSE_TAG)) {
+            return; // don't move, comment placement might cause invalid code if moved
```
SpacePossum: Just realized this can be moved as long as the close tag is left in place and a semicolon is added to the moved `declare` statement. I might be able to PR this later on this week.

src/Fixer/Comment/HeaderCommentFixer.php
```diff
@@ -126,43 +129,81 @@ private function findHeaderCommentIndex(Tokens $tokens)
...
+            if ($isOpen && $tokens[$index]->getContent() === 'strict_types') {
+                $isStrictTypes = true;
+            }
+            if ($tokens[$index]->getContent() === ';') {
+                ++$index;
+                break;
+            }
+        }
+
+        return $isStrictTypes ? $index : $originalIndex;
```
gharlan: I think this will pass your check (`$isStrictTypes` will be `true`):

```php
declare(ticks=1) {
    const strict_types = 1;
}
```
;)

src/Fixer/Strict/StrictTypesFixer.php
```diff
@@ -0,0 +1,170 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // check for if declaration already in file
+        $sequence = $this->getDeclareStrictTypeSequence();
+        $searchIndex = $tokens->getNextMeaningfulToken(0);
+        if (null === $searchIndex) {
+            $this->insertNewSequence($tokens); // not found, insert declare statement
```
gharlan: `$searchIndex` is `null`, so there is no meaningful token in file (which should not be happen because of `isCandidate` check). But then you add the declare sequence (without php open tag).
I think something is wrong here..

src/Fixer/Comment/HeaderCommentFixer.php
SpacePossum: is this valid? and if it s, what does it mean? ;)

gharlan: hmm, the `const` is not valid there, but this is valid: https://3v4l.org/MZ2TH

keradus:
```php
declare(ticks=1);

const strict_types = 1;
```

src/Fixer/Strict/StrictTypesFixer.php
SpacePossum: indeed `$searchIndex` can never be `null` here, the whole if case can be removed, good catch :+1:

src/Fixer/Comment/HeaderCommentFixer.php
gharlan: @keradus but you example will not pass the check, because there is a `;` before `strict_types`.. ;)

Seldaek:
@SpacePossum if you want just open a new PR based off this one.. so you can freely tinker with it :) That'll be easier for you and me I guess, because I don't have a ton of time to spend on this. 

I just wanted to convert monolog to strict mode which I have done using my simple patch, now if you wanna spend hours figuring out all the insane syntax people could use to break the fixer I'm sure that's best for the project but I don't have time or energy for this :) It could also be merged as is because it works, and then improved later as/if needed.

src/Fixer/Strict/StrictTypesFixer.php
gharlan: hmm I was wrong!

```php
$searchIndex = $tokens->getNextMeaningfulToken(0);
```

I thought: if `$searchIndex` is `null`, there is no meaningful token in file.
But actually: there is no meaningful token after first token.

src/Fixer/Strict/StrictTypesFixer.php
```diff
@@ -0,0 +1,170 @@
...
+            }
+        } else {
+            $tokens->insertAt(1 + $endIndex, new Token(array(T_WHITESPACE, "\n")));
+        }
+
+        // transform "<?php\n" to "<?php " if needed
+        if (false !== strpos($tokens[0]->getContent(), "\n")) {
+            $tokens[0]->setContent(trim($tokens[0]->getContent()).' ');
+        } elseif ($tokens[1]->isWhitespace()) {
+            $tokens[1]->setContent(' ');
```
gharlan: I think we should remove the whitespace token to get `<?php declare..` (first space belongs to open tag token).

GrahamCampbell: Also, what happens if someone puts a different type of declare there?

src/Fixer/Comment/HeaderCommentFixer.php
keradus: huh https://3v4l.org/Op6dm ?

gharlan: I mean: `$isStrictTypes` will be `false` in your case.

keradus:
This fixer should be marked as risky

SpacePossum:
>This fixer should be marked as risky

Wouldn't it be possible to make it non-risky?

src/Fixer/Strict/StrictTypesFixer.php
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1921/files#diff-effdfb27184a14e565c6d45e17fea171R36

seems to work?

SpacePossum: >Also, what happens if someone puts a different type of declare there?

Nothing happens to that declare, the strict type declare is moved to be the first statement after the open tag.

keradus:
no, as `declare(strict_types=1)` changes how the code is executed

GrahamCampbell:
Yeh this is definitely risky. :)

SpacePossum:
I see, but that is the whole point :D I'm just not a big fan of warning people against themselves, but sure lets mark it risky :)

GrahamCampbell:
> I'm just not a big fan of warning people against themselves, but sure lets mark it risky :)

I think this is fine. People should know when a fixer is safe or not to avoid newcomers to php-cs-fixer from breaking their code by mistake.

keradus:
also, I pretty blindly run non-risky fixers against big codebase, and then activate risky fixers in smaller steps

Seldaek:
Closing in favor of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1949

Issue #1923, opened 2016-05-27
NoUselessElse fixer applies invalid fix


localheinz:
Noticed that applying the `no_useless_else` fixer on

```php
return array_map(function ($name, $options) use ($isMany) {
    if (array_key_exists('model', $options)) {
        $alias = $options['model'];
    } else {
        $alias = Inflector::singular($name);
    }

    return new Relation(
        $name,
        $alias,
        $isMany
    );
}, array_keys($value), array_values($value));
```

results in

```php
return array_map(function ($name, $options) use ($isMany) {
    if (array_key_exists('model', $options)) {
        $alias = $options['model'];
    }
    $alias = Inflector::singular($name);

    return new Relation(
        $name,
        $alias,
        $isMany
    );
}, array_keys($value), array_values($value));
```

Any idea? Using https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.0.0-alpha on

```
PHP 5.6.21 (cli) (built: Apr 29 2016 02:31:44)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
```

/cc @SpacePossum 

Related to #1773.

Soullivaneuh:
Also got an issue with `no_useless_else`

This:

```php
if (null !== $options['user']) {
    $qb = $er->getByOwnerQB($options['user']);
} else {
    $qb = $er->createQueryBuilder('s');
}
```

Is converted to this:

```php
if (null !== $options['user']) {
    $qb = $er->getByOwnerQB($options['user']);
}
$qb = $er->createQueryBuilder('s');
```

But in this case, `$qb` will be overridden even if `$options['user']` is not null.

localheinz:
@Soullivaneuh 

Looks like the exact same thing, right?! 😃

Soullivaneuh:
@localheinz Well, indeed. End of the day, you know... :-)

greg0ire:
"Me too" :trollface:

SpacePossum:
just a little longer, fix is on it way...

SpacePossum:
btw. @Soullivaneuh could you share a bit more of the code being changed incorrectly, it is prop. because of the tokens found just before the `if` (I hope, otherwise there is another issue..)

Soullivaneuh:
@SpacePossum Of course!

```php
public function configureOptions(OptionsResolver $resolver)
{
[...]

PR #1926, opened 2016-05-27
NoUselessElseFixer - fix wrong if handling

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1923

### DIFF
--- a/src/Fixer/ControlStructure/NoUselessElseFixer.php
+++ b/src/Fixer/ControlStructure/NoUselessElseFixer.php
@@ -96,12 +96,12 @@ final class NoUselessElseFixer extends AbstractFixer
                 return;
             }
 
-            $candidateIndex = $tokens->getTokenOfKindSibling(
+            $candidateIndex = $tokens->getPrevTokenOfKind(
                 $previous,
-                -1,
                 array(
                     ';',
                     array(T_CLOSE_TAG),
+                    array(T_IF),
                     array(T_BREAK),
                     array(T_CONTINUE),
                     array(T_EXIT),
@@ -113,8 +113,7 @@ final class NoUselessElseFixer extends AbstractFixer
 
             if (
                 null === $candidateIndex ||
-                $tokens[$candidateIndex]->equals(';') ||
-                $tokens[$candidateIndex]->isGivenKind(T_CLOSE_TAG) ||
+                $tokens[$candidateIndex]->equalsAny(array(';', array(T_CLOSE_TAG), array(T_IF))) ||
                 $this->isInConditional($tokens, $candidateIndex, $previousBlockStart)
             ) {
                 return;
@@ -211,7 +210,7 @@ final class NoUselessElseFixer extends AbstractFixer
      */
     private function isInConditional(Tokens $tokens, $index, $lowerLimitIndex)
     {
-        $candidateIndex = $tokens->getTokenOfKindSibling($index, -1, array(')', ';', ':'));
+        $candidateIndex = $tokens->getPrevTokenOfKind($index, array(')', ';', ':'));
         if ($tokens[$candidateIndex]->equals(':')) {
             return true;
         }
--- a/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
+++ b/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
@@ -705,6 +705,58 @@ else?><?php echo 5;',
                         echo 7;
                 ',
             ),
+            array(
+                '<?php
+                    $tmp = function($b){$b();};
+                    $a =1;
+                    return $tmp(function () use ($a) {
+                        if ($a) {
+                            $a++;
+                        } else {
+                            $a--;
+                        }
+                    });
+                ',
+            ),
+            array(
+                '<?php
+                    $tmp = function($b){$b();};
+                    $a =1;
+                    return $tmp(function () use ($a) {
+                        if ($a) {
+                            $a++;
+                        } elseif($a > 2) {
+                            return 1;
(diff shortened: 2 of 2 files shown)


SpacePossum:
@localheinz sorry for messing this up :(
This is an early opening PR since when it tried to find the problem based on your findings I ran into an `out range exception`, so not a false-positive fix. I've to dig a bit more to see if there is another cause.

If you have the time, could you retest the fixer with the changes in this PR on your code that was fixed incorrectly?

src/Fixer/ControlStructure/NoUselessElseFixer.php
```diff
@@ -100,21 +100,24 @@ private function fixElse(Tokens $tokens, $index)
...
+                    // positives
                     array(T_BREAK),
                     array(T_CONTINUE),
                     array(T_EXIT),
                     array(T_GOTO),
                     array(T_RETURN),
                     array(T_THROW),
+                    // negatives
+                    ';',
+                    array(T_IF),
```
SpacePossum: What would be even better/needed is to have a limit for the `getTokenOfKindSibling`method.

tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
```diff
@@ -22,6 +22,21 @@
  */
 final class NoUselessElseFixerTest extends AbstractFixerTestCase
 {
+    public function testLocalheinz()
+    {
+        $input =
+            '<?php
+                return $tmp(function () use $a {
```
localheinz: @SpacePossum 

Should be `use ($a)`!

😉

tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
```diff
@@ -22,6 +22,21 @@
  */
 final class NoUselessElseFixerTest extends AbstractFixerTestCase
 {
+    public function testLocalheinz()
```
keradus: ...

tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
```diff
@@ -22,6 +22,21 @@
...
+    {
+        $input =
+            '<?php
+                return $tmp(function () use $a {
+                    if ($a) {
+                        $a++;
+                    } else {
+                        $a--;
+                    }
+                };
```
localheinz: @SpacePossum 

You are also missing a closing `)` here!

SpacePossum: thanks! I'm re-doing this whole fix ATM, what I need is a limit on the `getTokenOfKindSibling` method. Hope to fix this evening!

localheinz: @SpacePossum 

Cool, let me know when I should take a look and test!

GrahamCampbell: :+1:

src/Tokenizer/Tokens.php
```diff
@@ -593,39 +594,50 @@ public function getPrevNonWhitespace($index, $whitespaces = null)
...
+     * @param int      $index         token index
+     * @param int      $direction     direction for looking, +1 or -1
+     * @param int|null $limit         null no limit
+     * @param array    $tokens        possible tokens
+     * @param bool     $caseSensitive perform a case sensitive comparison
      *
      * @return int|null
      */
-    public function getTokenOfKindSibling($index, $direction, array $tokens = array(), $caseSensitive = true)
+    public function getTokenOfKindSibling($index, $direction, $limit = null, array $tokens = array(), $caseSensitive = true)
```
SpacePossum: This is the only real point of interest in this PR.

keradus: why do we need it in the first place? what was wrong with checking it on the calling place?
it's not common in our code to check that limit as far as I can see

src/Tokenizer/Tokens.php
```diff
@@ -593,39 +594,50 @@ public function getPrevNonWhitespace($index, $whitespaces = null)
...
+     * @param bool     $caseSensitive perform a case sensitive comparison
+     * @param int|null $limit         false no limit
      *
      * @return int|null
      */
-    public function getPrevTokenOfKind($index, array $tokens = array(), $caseSensitive = true)
+    public function getPrevTokenOfKind($index, array $tokens = array(), $caseSensitive = true, $limit = null)
     {
-        return $this->getTokenOfKindSibling($index, -1, $tokens, $caseSensitive);
+        return $this->getTokenOfKindSibling($index, -1, $limit, $tokens, $caseSensitive);
```
keradus: also, having different order of params for general `getTokenOfKindSibling` method and it's aliases is not intuitive

keradus:
A lot of nice celanup, but thanks to them I lost what part is actually a bugfix and what only cleanup

Soullivaneuh:
@SpacePossum Could you add this to your tests? https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1923#issuecomment-222731341

Thanks.

SpacePossum:
sure thing @Soullivaneuh , sorry for the inconvenience (cc @greg0ire) :(
while we are talking, could take a look at https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1942 IIRC your are the author of the fixer?

SpacePossum:
ready for review

keradus:
:+1:

keradus:
Thank you @SpacePossum.

PR #1936, opened 2016-05-30
Update the package owner for composer, replacing fabpot by friendsofsymfony

Veve2:
Since fabpot/php-cs-fixer has been replaced by this one, the installation guide should use it.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ To install PHP-CS-Fixer, install Composer and issue the following command:
 
 .. code-block:: bash
 
-    $ ./composer.phar global require fabpot/php-cs-fixer
+    $ ./composer.phar global require friendsofphp/php-cs-fixer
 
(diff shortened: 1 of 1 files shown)


keradus:
This is not true for 1.x line

SpacePossum:
Hi,
Thanks for the PR, please note that the new version of PHP-CS-Fixer will be 2.0 released from master.
On master the README has been updated in a similar way as you propose.

GrahamCampbell:
@keradus How about we update the default branch to master here?

keradus:
:-1:

Again, we will be trapped by users who go for github starting page, read the readme, but then download stable version and "oh, it's not working!"

Veve2:
"This is not true for 1.x line"
How could I had write the patch to deal with this?
How can I test the patch for other branches?

GrahamCampbell:
@keradus The current package should not have been marked as abandoned on packagist yet.

keradus:
Patch is already done for branches that need it, closing.

PR #1939, opened 2016-05-30
BracesFixer - fix handling comment around control token

### LINKED ISSUE #1938: Test case that crashes php-cs-fixer
Latest 2.x. Enable the fixers: `braces` and `elseif` only.

**Input:**

```php
<?php

	if ($foo) {
		//
	}
	else
/*-- TROLOLOL --*/
	if ($bar) {}
```

keradus:
closes #1938

### DIFF
--- a/Symfony/CS/Fixer/PSR2/BracesFixer.php
+++ b/Symfony/CS/Fixer/PSR2/BracesFixer.php
@@ -450,7 +450,7 @@ class BracesFixer extends AbstractFixer
      */
     private function findParenthesisEnd(Tokens $tokens, $structureTokenIndex)
     {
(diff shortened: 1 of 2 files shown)


keradus:
cc @GrahamCampbell

GrahamCampbell:
:+1:

GrahamCampbell:
Could we add my other test case as an integration test?

keradus:
you know that it wasn't needed to use second fixer for this bug right ?

GrahamCampbell:
1.11 didn't crash for me on my test case with both fixers enabled.

keradus:
as it didn't crash for every possible code, i took the one that's causing the issues.

not sure if we want to go for every possible failing case and maintain it...

PR #1941, opened 2016-05-30
NoEmptyCommentFixer - fix middle line removal

keradus:
new case, not yet working


keradus:
@SpacePossum do we want to take care of it ?

SpacePossum:
:+1: 
We could, shouldn't be hard :)
Note to self to test the `#` version as well.

keradus:
@SpacePossum 
Please, could you handle it? Also, maybe you have some more tests on your mind basing on my example.

keradus:
Closing in favor of #1946

PR #1946, opened 2016-06-02
NoEmptyCommentFixer - Only remove complete empty comment blocks

SpacePossum:
master version of;
@see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1941

I've changed the spec a bit to only complete empty comment blocks.
Started on master because of the comment with line break transformers are helping a lot...

### DIFF
--- a/src/Fixer/Comment/NoEmptyCommentFixer.php
+++ b/src/Fixer/Comment/NoEmptyCommentFixer.php
@@ -20,22 +20,31 @@ use PhpCsFixer\Tokenizer\Tokens;
  */
 final class NoEmptyCommentFixer extends AbstractFixer
 {
-    /**
-     * {@inheritdoc}
-     */
-    public function isCandidate(Tokens $tokens)
-    {
-        return $tokens->isTokenKindFound(T_COMMENT);
-    }
+    const TYPE_HASH = 1;
+    const TYPE_DOUBLE_SLASH = 2;
+    const TYPE_SLASH_ASTERISK = 3;
 
     /**
      * {@inheritdoc}
      */
     public function fix(\SplFileInfo $file, Tokens $tokens)
     {
-        foreach ($tokens as $index => $token) {
-            if ($token->isGivenKind(T_COMMENT)) {
-                $this->fixComment($tokens, $index);
+        for ($index = 0, $count = count($tokens); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(T_COMMENT)) {
+                continue;
+            }
+
+            list($blockStart, $index, $isEmpty) = $this->getCommentBlock($tokens, $index);
+            if (false === $isEmpty) {
+                continue;
+            }
+
+            if ($tokens[$index]->isWhitespace() && false !== strpos($tokens[$index]->getContent(), "\n")) {
+                --$index;
+            }
+
+            for ($i = $blockStart; $i <= $index; ++$i) {
+                $tokens->clearTokenAndMergeSurroundingWhitespace($i);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/Fixer/Comment/NoEmptyCommentFixer.php
```diff
@@ -20,22 +20,31 @@
...
+                continue;
+            }
+
+            list($blockStart, $index, $isEmpty) = $this->getCommentBlock($tokens, $index);
+            if (false === $isEmpty) {
+                continue;
+            }
+
+            if ($tokens[$index]->isWhitespace() && false !== strpos($tokens[$index]->getContent(), "\n")) {
+                --$index;
```
keradus: why we are going backward once?

src/Fixer/Comment/NoEmptyCommentFixer.php
```diff
@@ -58,34 +67,105 @@ public function getPriority()
     }
 
     /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
```
keradus: should not be moved (as, alphabetic order inside methods with same visibility)

src/Fixer/Comment/NoEmptyCommentFixer.php
```diff
@@ -58,34 +67,105 @@ public function getPriority()
...
+                // single line comment starting with '#'
+                return 1 === preg_match('|^#\s*$|', $content);
+            case self::TYPE_SLASH_ASTERISK:
+                // comment starting with '/*' and ending with '*/' (but not a PHPDoc)
+                return 1 === preg_match('|^/\*\s*\*/$|', $content);
+            case self::TYPE_DOUBLE_SLASH:
+                // single line comment starting with '//'
+                return 1 === preg_match('|^//\s*$|', $content);
+            default:
+                throw new \UnexpectedValueException('Comment type detected failed.');
```
keradus: could never happen, as getCommentType return only 3 values (if it cannot detect it returns `//`)

keradus:
Awesome work ! Thank you @SpacePossum 
Please, could you rebase  and see my tiny comment ?

src/Fixer/Comment/NoEmptyCommentFixer.php
SpacePossum: but the methods same visibility are now in alphabetic order, you want `isCandidate` always above `fix` method ?

src/Fixer/Comment/NoEmptyCommentFixer.php
SpacePossum: not needed indeed, relic from 1.x branch, removed.

src/Fixer/Comment/NoEmptyCommentFixer.php
keradus: my bad, I ignored `is` prefix. please ignore me ;)

keradus:
Thank you @SpacePossum.

SpacePossum:
@keradus this fix should be backported to 1.x, however if I cannot do this in time for 1.12 release maybe we should remove the fixer for the 1.x line.

GrahamCampbell:
I don't think we really need to backport. Just document the fixer behaviour has changed in the upgrading guide.

keradus:
I don't see value for backporting this one as well, as it's sth in the middle of bugfix and specification update.
2.x in general change way of dealing with comments, so I don't see big need of describing this particular comment fixer change

SpacePossum:
very well, saves a lot of coding ^-^ thanks! :+1:

PR #1956, opened 2016-06-04
SelfUpdateCommand - don't update to non-stable version

### LINKED ISSUE #1925: SelfUpdateCommand updates to non-stable version
The currently hosted phar does not run.

jon@jon-bold-mac:~/repo$ php php-cs-fixer2 fix ./back-end/
PHP Fatal error:  Class 'Symfony\CS\Finder\DefaultFinder' not found in /home/jon/repo/back-end/.php_cs on line 3

Here's the contents of our .php_cs file.

<?php

	->exclude('config','storage','bootstrap')
	->in(__DIR__);

	->fixers(array('trailing_spaces','-psr0'))
	->finder($finder);

keradus:
Fixes #1925

This is not ultimate fix, as if one downloaded 2.x-alpha then he has it.
Also, general download URL is (still) serving 2.x-alpha.

Changing it is another issue.
In general we want to move phar files to github releases.

This PR fixes update command to not update to not-stable releases with current way of serving phar files.

### DIFF
--- a/Symfony/CS/Console/Command/SelfUpdateCommand.php
+++ b/Symfony/CS/Console/Command/SelfUpdateCommand.php
@@ -116,9 +116,13 @@ EOT
     {
         if ($this->checkIfVersionFileExists($major, $minor, $patch + 1)) {
             return $this->findBestVersion($major, $minor, $patch + 1);
-        } else if ($this->checkIfVersionFileExists($major, $minor + 1, 0)) {
+        }
+
+        if ($this->checkIfVersionFileExists($major, $minor + 1, 0)) {
             return $this->findBestVersion($major, $minor + 1, 0);
-        } else if ($this->checkIfVersionFileExists($major + 1, 0, 0)) {
+        }
+
+        if ($this->checkIfVersionFileExists($major + 1, 0, 0)) {
             return $this->findBestVersion($major + 1, 0, 0);
         }
 
--- a/Symfony/CS/Console/Command/SelfUpdateCommand.php
+++ b/Symfony/CS/Console/Command/SelfUpdateCommand.php
@@ -20,6 +20,7 @@ use Symfony\CS\ToolInfo;
  * @author Igor Wiedler <igor@wiedler.ch>
  * @author Stephane PY <py.stephane1@gmail.com>
  * @author Grégoire Pineau <lyrixx@lyrixx.info>
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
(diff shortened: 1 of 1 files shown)


Symfony/CS/Console/Command/SelfUpdateCommand.php
```diff
@@ -95,4 +103,43 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+    private function buildVersionFileUrl($major, $minor, $patch)
+    {
+        return sprintf('http://get.sensiolabs.org/php-cs-fixer-v%s.phar', $this->buildVersionString($major, $minor, $patch));
+    }
+
+    private function buildVersionString($major, $minor, $patch)
+    {
+        if (0 === $patch && $major < 2) {
+            return sprintf('%d.%d', $major, $minor);
```
keradus: actually, we could just relesae 1.12.0 instead of 1.12 and dont have that exception, as it won't affect versions 1.11(.0), 1.10(.0) or lower (as the patch will be release for 1.11.4+)

SpacePossum:
nice in between fix :)
question, wouldn't something like https://github.com/composer/semver help you here?

keradus:
I don't see how.
We don't need to compare version, `Semver` is not needed for us as well (things like `satisfiedBy`).
We only need to bump a version and see if it exists or not.

SpacePossum:
Very well :) 
I thought it might help as you don't have to parse the version string and up it yourself.

SpacePossum:
@keradus this fix doesn't seem to be in the 1.13 and master branch, not sure if this is done on purpose.

keradus:
Wow wow wow wow woow, really? :|:|:|

keradus:
the solution was modified on 1.11.5 to not use sensiolabs server as indicator of newest release, as it pointed to beta release.

```bash
keradus@keradus-Lap:~/Downloads$ php php-cs-fixer.phar --version
PHP CS Fixer version 1.11.5 by Fabien Potencier (d3d08b7)
keradus@keradus-Lap:~/Downloads$ php php-cs-fixer.phar self-update
php-cs-fixer updated.
keradus@keradus-Lap:~/Downloads$ php php-cs-fixer.phar --version
PHP CS Fixer version 1.12.3 by Fabien Potencier and Dariusz Ruminski (78a820c)
keradus@keradus-Lap:~/Downloads$ php php-cs-fixer.phar self-update
php-cs-fixer is already up to date.
```

SpacePossum:
ah so it was merged but than changed, clear, sorry for the noise!

Issue #1959, opened 2016-06-04
Handling semi-reserved words


keradus:
From PHP 7 some reserved words become semi-reserved, eg now we could make a constant named `public`, which for PHP 5 could be only visibility attribute.

This issue is to collect all talk in one place, so it replaces #1772, #1837, #1945 .

General thoughts collected from mentioned issues:
- code using on semi-reserved keywords is valid only for PHP7
- by default, PHP tokenizer get lost in given cases (like returning `T_PUBLIC`, which in general sucks!)
- we could pass `TOKEN_PARSE` option for `token_get_all` for PHP7 and don't for PHP5, then, all semi-reserved tokens will be treated as they should (eg as const's name, not visibility attribuet), and we won't need to handle all that kind exceptions by all fixers. win-win. Warning: we won't be able to properly handle PHP7 code when running the tool under PHP5, but we already advise to run tool with at least same version as the fixed code supports.

Also, as suggested, I can confirm (ref https://github.com/php/php-src/pull/1221) that tokenizer with passed `TOKEN_PARSE` option is in fact **linting** the code! So we wouldn't need call separated process to do so.
And then the best part. Currently, the linter is working in 2nd process, which in general boost tool a lot. After switching to linting using `TOKEN_PARSE`, we won't lint in parallel process, so it's sad... or not! As we will actually parse it once, our cache will already work for that and... the linting will be in fact no-cost at all! (Except that extending tokenizing time made by first paragraph, but we need it one way or another). Even better - parsing with `TOKEN_PARSE` doesn't took longer time or use more memory!

TODO:
- add `TOKEN_PARSE` for tokenizer when possible
- made new Linter and choose which one to use during runtime

(Disclaimer: working on solution)

----------

Marking as `bug`, as it will fix few reported bugs, and `enhancement`, as it will boos our linting

----------

Thank you @sebastianbergmann, @gharlan, @SpacePossum, @GrahamCampbell for discussion so far in related threads!

gharlan:
👍 Thanks for working on this!

keradus:
Bugpart fixed by #1970

keradus:
Enhancement part done by #1973

PR #1960, opened 2016-06-04
Fix invalid test cases

keradus:
Some kind of internal bug, end user is not affected by this.

It was discovered during #1959


SpacePossum:
sorry about that one, good catch! :+1:

keradus:
Updated tokenizer detected this :D

keradus:
few more cases

Issue #1957, opened 2016-06-04
SingleLineAfterImportsFixer broken


GrahamCampbell:
Consider the following input file:

```php
<?php namespace Foo;

use Bar\Baz;

/**
 * Foo.
 */
```

We'll end up with:

```
RuntimeException: Index invalid or out of range in /root/PHP-CS-Fixer/src/Fixer/Import/SingleLineAfterImportsFixer.php:69
Stack trace:
#0 /root/PHP-CS-Fixer/src/Runner/Runner.php(189): PhpCsFixer\Fixer\Import\SingleLineAfterImportsFixer->fix(Object(Symfony\Component\Finder\SplFileInfo), Object(PhpCsFixer\Tokenizer\Tokens))
#1 /root/PHP-CS-Fixer/src/Runner/Runner.php(141): PhpCsFixer\Runner\Runner->fixFile(Object(Symfony\Component\Finder\SplFileInfo), Object(PhpCsFixer\Linter\LintingResult))
#2 /root/PHP-CS-Fixer/src/Console/Command/FixCommand.php(364): PhpCsFixer\Runner\Runner->fix()
#3 /root/PHP-CS-Fixer/vendor/symfony/console/Command/Command.php(256): PhpCsFixer\Console\Command\FixCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#4 /root/PHP-CS-Fixer/vendor/symfony/console/Application.php(815): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#5 /root/PHP-CS-Fixer/vendor/symfony/console/Application.php(186): Symfony\Component\Console\Application->doRunCommand(Object(PhpCsFixer\Console\Command\FixCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#6 /root/PHP-CS-Fixer/vendor/symfony/console/Application.php(117): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#7 /root/PHP-CS-Fixer/php-cs-fixer(45): Symfony\Component\Console\Application->run()
#8 {main}
```

keradus:
[...]

PR #1961, opened 2016-06-05
[Bug] BraceFixer - fix no token before start.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1957#issuecomment-223771898

opening now to see Travis results

### DIFF
--- a/Symfony/CS/Fixer/Contrib/HeaderCommentFixer.php
+++ b/Symfony/CS/Fixer/Contrib/HeaderCommentFixer.php
@@ -63,6 +63,10 @@ class HeaderCommentFixer extends AbstractFixer
     {
         $tokens = Tokens::fromCode($content);
 
+        if (count($tokens) < 1 || !$tokens[0]->isGivenKind(T_OPEN_TAG)) {
+            return $content;
+        }
+
         if (!$tokens->isMonolithicPhp()) {
             return $content;
         }
--- a/Symfony/CS/Fixer/Contrib/NewlineAfterOpenTagFixer.php
+++ b/Symfony/CS/Fixer/Contrib/NewlineAfterOpenTagFixer.php
@@ -27,19 +27,19 @@ class NewlineAfterOpenTagFixer extends AbstractFixer
     {
         $tokens = Tokens::fromCode($content);
 
+        // ignore files with short open tag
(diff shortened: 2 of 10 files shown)


SpacePossum:
fabbot is wrong on this one.
Expecting HHVM to fail

SpacePossum:
HHVM failures become more and more interesting :D

keradus:
@SpacePossum 
are you sure that target should be 1.12 and not 1.11 ?

SpacePossum:
>are you sure that target should be 1.12 and not 1.11 ?

I started out on a bug on 1.12, but I think this might indeed needs a retarget to 1.11. I'll finish it first :)

Symfony/CS/Fixer/Contrib/HeaderCommentFixer.php
```diff
@@ -63,6 +63,10 @@ public function fix(\SplFileInfo $file, $content)
     {
         $tokens = Tokens::fromCode($content);
 
+        if (count($tokens) < 1 || !$tokens[0]->isGivenKind(T_OPEN_TAG)) {
```
keradus: Do we really call fixer with empty tokens collection?

keradus: if so, we should put that check to `AbstractFixer::isCandidate`

SpacePossum: >AbstractFixer::isCandidate

agreed, will do on 2.x :)

keradus: but first, do we?
> Do we really call fixer with empty tokens collection?

SpacePossum:
This PR fixes a bunch of issues including the one ref. in the description.
ping @GrahamCampbell as reporter of the stated issue.
All green now :)

keradus:
so, I should merge it to 1.11 ?

keradus:
nice fabbot failure btw...

SpacePossum:
Yeah fabbot is not as smart as it could be about spelling ;)
Anyway, here is a commit for 1.11 https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/533ac2065fd7f33fb82a6d097e8861d8fe1e4165

we can ditch this PR than as don't need the typo fixes in the comments of `NativeFunctionCasingFixer`

PR #1965, opened 2016-06-06
AbstractPsrAutoloadingFixer - fix edge case of halting compiler for PHP 5.3

keradus:


keradus:
cc @SpacePossum

SpacePossum:
nice :)
thanks!
are we in the clear with the other keywords and magic constants? I guess since the test are green :)

keradus:
yep, we have hack for halting compiler in other places too for 5.3

PR #1969, opened 2016-06-07
Force dependencies to work on PHP 5.3

GrahamCampbell:


keradus:
:-1:

this is not a good solution IMO, that way, regardless of env, you require the packages to be compatible with old PHP, even if we run composer under PHP7.
This flag may be used for build, but not for regular usage.

GrahamCampbell:
@keradus This flag is not inherited, so it doesn't affect anything when php-cs-fixer is used as a dependency. It only affects building it, or running it's own tests.

keradus:
exactly, it disallow us for testing newest packages that we claim to support, eg Symgony 3.x

GrahamCampbell:
> exactly, it disallow us for testing newest packages that we claim to support, eg Symgony 3.x

Yes it does. That is unfortunate. This PR is just a suggestion though. ;)

keradus:
It may be partial solution until we will attaching the phars for github releases.
In the meantime, could you remove that added part in Travis jobs ?

ajgarlag:
For me the best option is to require at least an LTS version of every symfony component, an then execute "composer update --prefer-lowest" before building the phar.

keradus:
which cannot be done untill we will modify way of building and storing phars, as I mentioned.

keradus:
@GrahamCampbell 

The current problem is that we will only test 5.3.6 compatible version of packages, eg never test Symfony 3.x components.

Please, for build matrix whenever we don't use `--prefer-lowest` add `--ignore-platform-reqs`.
That way, on 5.3 we will check lowest deps compatible with 5.3.6, and on all other we will test highest deps possible for given PHP version (ignoring 5.3.6 requirement, as it was tested in first job)

GrahamCampbell:
> The current problem is that we will only test 5.3.6 compatible version of packages, eg never test Symfony 3.x components.

Yeh. I know. D'no what to recommend here I'm afraid, other than having two composer.json files, which would really suck.

GrahamCampbell:
If it works, sure. I don't know how the phars are built.

keradus:
ping @GrahamCampbell

keradus:
?

keradus:
@GrahamCampbell  ?

GrahamCampbell:
My solution is crap. I only put it here to motivate something better. :P

keradus:
I thought we agreed on solution and we don't see anything better so far, so why closing ?

PR #1970, opened 2016-06-08
Tokens - handle semi-reserved PHP 7 keywords

keradus:
Fixing bug part of #1959 

Should be reviewed together with #1971 and #1972

### DIFF
--- a/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
+++ b/Symfony/CS/Fixer/PSR0/Psr0Fixer.php
@@ -157,12 +157,21 @@ class Psr0Fixer extends AbstractFixer implements ConfigAwareInterface
             (!isset($filenameParts[1]) || 'php' !== $filenameParts[1])
             // ignore file with name that cannot be a class name
             || 0 === preg_match('/^[a-zA-Z_\x7f-\xff][a-zA-Z0-9_\x7f-\xff]*$/', $filenameParts[0])
+            // ignore filename that will halt compiler (and cannot be properly tokenized under PHP 5.3)
+            || '__halt_compiler' === $filenameParts[0]
         ) {
             return false;
         }
 
-        $tokens = Tokens::fromCode(sprintf('<?php class %s {}', $filenameParts[0]));
-        if ($tokens[3]->isKeyword() || $tokens[3]->isMagicConstant()) {
+        try {
+            $tokens = Tokens::fromCode(sprintf('<?php class %s {}', $filenameParts[0]));
+
+            if ($tokens[3]->isKeyword() || $tokens[3]->isMagicConstant()) {
+                // name can not be a class name - detected by PHP 5.x
+                return false;
+            }
+        } catch (\ParseError $e) {
+            // name can not be a class name - detected by PHP 7.x
             return false;
         }
 
--- a/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
@@ -1301,6 +1301,31 @@ while (true) {
     }
 
     /**
+     * @dataProvider provide70Cases
+     * @requires PHP 7.0
+     */
+    public function test70($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provide70Cases()
+    {
+        return array(
+            array(
+                '<?php
+    class Foo
+    {
+        public function use()
+        {
+        }
+    }
+                ',
+            ),
+        );
+    }
+
+    /**
      * @dataProvider providePreserveLineAfterControlBrace
      */
     public function testPreserveLineAfterControlBrace($expected, $input = null)
--- a/Symfony/CS/Tests/Tokenizer/Transformer/ClassConstantTest.php
+++ b/Symfony/CS/Tests/Tokenizer/Transformer/ClassConstantTest.php
@@ -42,6 +42,10 @@ class ClassConstantTest extends AbstractTransformerTestBase
                 ),
             ),
             array(
+                '<?php echo X::bar;',
+                array(),
+            ),
+            array(
                 '<?php class X{}',
                 array(),
             ),
--- a/Symfony/CS/Tokenizer/Tokens.php
+++ b/Symfony/CS/Tokenizer/Tokens.php
@@ -137,18 +137,10 @@ class Tokens extends \SplFixedArray
             }
         }
 
-        $tokens = token_get_all($code);
+        $tokens = new self();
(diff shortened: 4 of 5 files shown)


keradus:
@GrahamCampbell you were one of those who reported the bug. Could you confirm the fix please?

GrahamCampbell:
@keradus https://github.com/padraic/mockery/blob/master/tests/Mockery/Fixtures/SemiReservedWordsAsMethods.php might be a good test case. :)

keradus:
It's min-case is aready there. No need to check every single keyword here,either we properly detecting semi-reserved keywords or not.

GrahamCampbell:
Seems to work:

```php
   1) SemiReservedWordsAsMethods.php (trailing_spaces, visibility, braces, eof_ending)
      --- Original
      +++ New
      @@ @@
       <?php

       namespace Mockery\Fixtures;

      -class SemiReservedWordsAsMethods
      +class SemiReservedWordsAsMethods
       {
      -    function callable() {}
      -    function class() {}
      -    function trait() {}
      -    function extends() {}
      -    function implements() {}
      -    function static() {}
      -    function abstract() {}
      -    function final() {}
      -    function public() {}
      -    function protected() {}
      -    function private() {}
      -    function const() {}
      -    function enddeclare() {}
      -    function endfor() {}
      -    function endforeach() {}
      -    function endif() {}
      -    function endwhile() {}
      -    function and() {}
      -    function global() {}
      -    function goto() {}
      -    function instanceof() {}
      -    function insteadof() {}
      -    function interface() {}
      -    function namespace() {}
      -    function new() {}
      -    function or() {}
      -    function xor() {}
      -    function try() {}
      -    function use() {}
      -    function var() {}
      -    function exit() {}
      -    function list() {}
      -    function clone() {}
      -    function include() {}
      -    function include_once() {}
      -    function throw() {}
      -    function array() {}
      -    function print() {}
      -    function echo() {}
      -    function require() {}
      -    function require_once() {}
      -    function return() {}
      -    function else() {}
      -    function elseif() {}
      -    function default() {}
      -    function break() {}
      -    function continue() {}
      -    function switch() {}
      -    function yield() {}
      -    function function() {}
      -    function if() {}
      -    function endswitch() {}
      -    function finally() {}
      -    function for() {}
      -    function foreach() {}
      -    function declare() {}
      -    function case() {}
      -    function do() {}
      -    function while() {}
      -    function as() {}
      -    function catch() {}
      -    function die() {}
      -    function self() {}
      -    function parent() {}
      +    public function callable()
      +    {
      +    }
      +    public function class()
      +    {
      +    }
      +    public function trait()
      +    {
      +    }
      +    public function extends()
      +    {
      +    }
      +    public function implements()
      +    {
      +    }
      +    public function static()
      +    {
      +    }
      +    public function abstract()
      +    {
      +    }
      +    public function final()
      +    {
      +    }
      +    public function public()
      +    {
      +    }
      +    public function protected()
      +    {
      +    }
      +    public function private()
      +    {
      +    }
      +    public function const()
      +    {
      +    }
      +    public function enddeclare()
      +    {
      +    }
      +    public function endfor()
      +    {
      +    }
      +    public function endforeach()
      +    {
      +    }
      +    public function endif()
      +    {
      +    }
      +    public function endwhile()
      +    {
      +    }
      +    public function and()
      +    {
      +    }
      +    public function global()
      +    {
      +    }
      +    public function goto()
      +    {
      +    }
      +    public function instanceof()
      +    {
      +    }
      +    public function insteadof()
      +    {
      +    }
      +    public function interface()
      +    {
      +    }
      +    public function namespace()
      +    {
      +    }
      +    public function new()
      +    {
      +    }
      +    public function or()
      +    {
      +    }
      +    public function xor()
      +    {
      +    }
      +    public function try()
      +    {
      +    }
      +    public function use()
      +    {
      +    }
      +    public function var()
      +    {
      +    }
      +    public function exit()
      +    {
      +    }
      +    public function list()
      +    {
      +    }
      +    public function clone()
      +    {
      +    }
      +    public function include()
      +    {
      +    }
      +    public function include_once()
      +    {
      +    }
      +    public function throw()
      +    {
      +    }
      +    public function array()
      +    {
      +    }
      +    public function print()
      +    {
      +    }
      +    public function echo()
      +    {
      +    }
      +    public function require()
      +    {
      +    }
      +    public function require_once()
      +    {
      +    }
      +    public function return()
      +    {
      +    }
      +    public function else()
      +    {
      +    }
      +    public function elseif()
      +    {
      +    }
      +    public function default()
      +    {
      +    }
      +    public function break()
      +    {
      +    }
      +    public function continue()
      +    {
      +    }
      +    public function switch()
      +    {
      +    }
      +    public function yield()
      +    {
      +    }
      +    public function function()
      +    {
      +    }
      +    public function if()
      +    {
      +    }
      +    public function endswitch()
      +    {
      +    }
      +    public function finally()
      +    {
      +    }
      +    public function for()
      +    {
      +    }
      +    public function foreach()
      +    {
      +    }
      +    public function declare()
      +    {
      +    }
      +    public function case()
      +    {
      +    }
      +    public function do()
      +    {
      +    }
      +    public function while()
      +    {
      +    }
      +    public function as()
      +    {
      +    }
      +    public function catch()
      +    {
      +    }
      +    public function die()
      +    {
      +    }
      +    public function self()
      +    {
      +    }
      +    public function parent()
      +    {
      +    }
       }
      -


      ---------- end diff ----------
```

It is odd that the line with the class definition is changing though, but doesn't appear changed?

GrahamCampbell:
Oh, wait, I see. It has a bit of trailing whitespace on the definition that the diff didn't show at all. Not even so i could highlight it with my mouse.

GrahamCampbell:
Tested the 2.x version and it works too. My tests were on php 7.0.7.

keradus:
isn't it general case of `braces` fixer? then, we have second fixer to remove extra trailing spaces.

GrahamCampbell:
> isn't it general case of braces fixer? then, we have second fixer to remove extra trailing spaces.

I just ran the symfony level. :P

keradus:
Then please try to reproduce it without this PR as well, as I don't see why this PR could introduce that behaviour

GrahamCampbell:
There's no bug, This works fine?

keradus:
didn't you said that you have got some odd whitespaces? or I understand it wrongly?
so, you confirm the fix ?

GrahamCampbell:
> didn't you said that you have got some odd whitespaces? or I understand it wrongly?

Yeh, then I added that was due to my input having whitespace I didn't know it had.

GrahamCampbell:
Everything appears to be working fine. Sorry for my ramblings. ;)

PR #1973, opened 2016-06-08
Introduce fast PHP7 based linter

keradus:
Closes enhancement part of #1959 .

Impact only on PHP 7

Command: `php php-cs-fixer fix -v`

**master**:
> Fixed all files in 19.708 seconds, 8.000 MB memory used
> Fixed all files in 19.647 seconds, 8.000 MB memory used
> Fixed all files in 19.824 seconds, 8.000 MB memory used

**this PR**:
> Fixed all files in 13.238 seconds, 8.000 MB memory used
> Fixed all files in 13.227 seconds, 8.000 MB memory used
> Fixed all files in 13.227 seconds, 8.000 MB memory used

**~33 %** faster

### DIFF
--- a/src/Runner/FileCachingLintingIterator.php
+++ b/src/Runner/FileCachingLintingIterator.php
@@ -20,7 +20,7 @@ use PhpCsFixer\Linter\LintingResultInterface;
  *
  * @internal
  */
-final class FileLintingIterator extends \IteratorIterator
+final class FileCachingLintingIterator extends \CachingIterator
 {
     /**
      * @var LintingResultInterface
@@ -31,6 +31,11 @@ final class FileLintingIterator extends \IteratorIterator
      * @var LinterInterface
      */
     private $linter;
+
+    /**
+     * @var LintingResultInterface
+     */
+    private $nextResult;
 
     public function __construct(\Iterator $iterator, LinterInterface $linter)
     {
@@ -48,8 +53,10 @@ final class FileLintingIterator extends \IteratorIterator
     {
         parent::next();
 
-        if ($this->valid()) {
-            $this->currentResult = $this->handleItem($this->current());
+        $this->currentResult = $this->nextResult;
+
+        if ($this->hasNext()) {
+            $this->nextResult = $this->handleItem($this->getInnerIterator()->current());
         }
     }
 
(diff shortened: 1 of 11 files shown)


gharlan:
What is happening when disabling linter on php 7?
If I see it correctly the code is linted anyway, because of `TOKEN_PARSE`. But the exception for invalid files is not catched, because `TokenizerLinter` is not used.

keradus:
Remember it's very early preview ;)
Indeed, on 7.x linter should be enforced.

gharlan:
I still think linting should be always enforeced and we should never fix invalid files (especially files from higher php versions).

keradus:
The problem is that for PHP5 you are not always able to enforce linting, as you need to find PHP executable path

GrahamCampbell:
I often want to turn off linting so I can debug bad fixers.

keradus:
one way or another for PHP7 linting would be always enabled thanks to Tokenizer change for semi-reserved keywords

GrahamCampbell:
What's the status here?

keradus:
I would guess it needs some dev time to be finished

src/Runner/Runner.php
```diff
@@ -128,7 +128,7 @@ public function fix()
         $finder = $config->getFinder();
         $finderIterator = $finder instanceof \IteratorAggregate ? $finder->getIterator() : $finder;
 
-        $collection = new FileLintingIterator(
+        $collection = new FileCachingLintingIterator(
```
GrahamCampbell: Guess this is yet to be implemented?

src/Runner/FileCachingLintingIterator.php
```diff
@@ -0,0 +1,80 @@
...
+
+use PhpCsFixer\Linter\LinterInterface;
+use PhpCsFixer\Linter\LintingResultInterface;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+final class FileLintingIterator extends \CachingIterator
```
GrahamCampbell: Typo in the filename here. :)

keradus:
cc @SpacePossum @GrahamCampbell

keradus:
Also, updated initial PR comment

GrahamCampbell:
Perhapse a linter factory would be better than having a linter wrapper class?

keradus:
It was my first implementation. Then I got sick of it to replace the usage everywhere so I decided to kept the original name as is and encapsulate decision which linter will be use for component, so one could treat Process/Tokenizer Linters only as private classes for Linter namespace (as if it would be possible in PHP, like in C#)

PR #1979, opened 2016-06-10
Added SingleLineClassDeclarationFixer

### LINKED ISSUE #1976: Single line class and method declaration fixers
Those fixers should be on the Symfony group according to the [coding standards](http://symfony.com/doc/current/contributing/code/standards.html#structure):

> * Declare the class inheritance and all the implemented interfaces on the same line as the class name;
> * Declare all the arguments on the same line as the method/function name, no matter how many arguments there are;

- [x] Class declaration: #1979 
- [x] Method declaration: #1980

Soullivaneuh:
Partially closes #1976

Question: Shall I add `trait` here too?

cc @javiereguiluz, you wrote the coding standard about that, what do you think?

I add some test cases. If you want to add more, please tell me! :+1:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -554,6 +554,12 @@ Choose from the list of available fixers:
                         blank line before a namespace
                         declaration.
 
+* **single_line_class_definition** [symfony]
+                        Declare the class inheritance
+                        and all the implemented
+                        interfaces on the same line as
+                        the class name.
+
 * **single_quote** [symfony]
                         Convert double quotes to
                         single quotes for simple
--- a/Symfony/CS/Fixer/Symfony/SingleLineClassDefinitionFixer.php
+++ b/Symfony/CS/Fixer/Symfony/SingleLineClassDefinitionFixer.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class SingleLineClassDefinitionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        $relevantTokenKinds = array(T_CLASS, T_INTERFACE);
+        if (version_compare(PHP_VERSION, '5.4', '>=')) {
+            $relevantTokenKinds[] = T_TRAIT;
+        }
+
+        foreach ($tokens->findGivenKind($relevantTokenKinds) as $givenTokens) {
+            foreach ($givenTokens as $index => $token) {
+                $beginIndex = $this->getBeginClassIndex($tokens, $index);
+                $braceIndex = $tokens->getNextTokenOfKind($beginIndex, array('{'));
+                $endIndex = $tokens->getPrevMeaningfulToken($braceIndex);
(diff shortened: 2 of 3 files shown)


SpacePossum:
>Question: Shall I add trait here too?

:+1:  and why not interfaces as well?

maybe as configuration option of the `class_definition`-fixer?

keradus:
and interfaces extending other interfaces?

javiereguiluz:
@Soullivaneuh it looks very nice 👍  Thanks for working on this.

Soullivaneuh:
> maybe as configuration option of the class_definition-fixer?

AFAIK, fixer options is only possible on 2.x.

Plus, I think it should not be on option.

It would be very weird to force inline declaration for classes, but not for traits, right?

@javiereguiluz Do we agree that this Symfony rules also applies for traits?

> and interfaces extending other interfaces?

Good catch! Will work on it. :+1:

javiereguiluz:
@Soullivaneuh I guess it applies to traits too to be consistent.

SpacePossum:
>AFAIK, fixer options is only possible on 2.x.

correct

>Plus, I think it should not be on option.

fair enough :)

>It would be very weird to force inline declaration for classes, but not for traits, right?

class_definition handles classes, interfaces and traits, would be great if your new would do that as well

Soullivaneuh:
> class_definition handles classes, interfaces and traits, would be great if your new would do that as well

Oh, so rename the fixer. Sounds good. :+1:

keradus:
> > AFAIK, fixer options is only possible on 2.x.
>
> correct

BTW, it was introduces to 2.x line as part of interface.
On 1.x line we already have configurations for few fixers, but not in part of interface and called statically, like header fixer

Soullivaneuh:
All fixed, please review! :+1:

Soullivaneuh:
@keradus Could you please explain this failing travis job? https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/136717844

SpacePossum:
that is because of typo in your test;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1979/files#diff-2cac64058aaace5a8ec0b1128005b97cR39
`implement` should be `implements`

On PHP7 (the version that is failing on Travis) the tests are linted. You can lint the test case on your machine as well if you want to prevent this from happening.

Soullivaneuh:
Thanks @SpacePossum, should be OK now.

SpacePossum:
I would still like to see the tests made in this PR moved to the `ClassDefinitionFixerTest`,
and based on config call a diff method here;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php#L143

The reason is basic DRY, the new fixer proposal has a lot in common with the current;
- finding candidates (should be the same)
- fixing white space before the classy token (should be the same)
- fixing white space after the classy token for `extends` (should be the same) and;
- fixing white space after the classy token for `implements` (should be the same)

So the only diff. is removing line breaks if there are any in a list of implementing interfaces.

keradus:
@SpacePossum you say that test classes are similar, are Fixer classes too ?

keradus:
Also, I still (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1979#issuecomment-225193893) think it should be one fixer...

Soullivaneuh:
@SpacePossum @keradus So if I understand well, the fixer should be merge onto `class_definition`

Is that right?

And what about #1980 in this case?

SpacePossum:
> So if I understand well, the fixer should be merge onto class_definition

I would prefer that yes :)
The link I provided shows where to hook:
```php
 $index = $this->ensureOnNewLine($tokens, $index);
```
->
```php
if (some config) {
     $index = $this->ensureOnSameine($tokens, $index); // new
} else {
     $index = $this->ensureOnNewLine($tokens, $index);
}

```

>And what about #1980 in this case?

lets discuss 1980 over there ;)

keradus:
ping @Soullivaneuh

Soullivaneuh:
Same opinion as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1980#issuecomment-230393449 ATM.

keradus:
any news here ?

keradus:
Closing due to merge of #2155

PR #1982, opened 2016-06-11
Added CombineConsecutiveIssetsFixer

### LINKED ISSUE #1977: Multiple argument isset refactor fixer
Isset cans take several arguments.

So a code line like this:

``` php
if (isset($criteria['date']) && isset($criteria['date']['query']) && isset($criteria['date']['params']))
```

Can be turned to this:

``` php
if (isset($criteria['date'], $criteria['date']['query'], $criteria['date']['params']))
```

I don't think the fixer will be risky, will be?

Soullivaneuh:
Closes #1977.

Feel free to add test cases on comments. :+1:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -616,6 +616,11 @@ Choose from the list of available fixers:
                         Align equals symbols in
                         consecutive lines.
 
+* **combine_consecutive_issets** [contrib]
+                        Calling isset on multiple
+                        items should be done in one
+                        call.
+
 * **combine_consecutive_unsets** [contrib]
                         Calling unset on multiple
                         items should be done in one
--- a/Symfony/CS/AbstractConsecutiveBuiltInFunctionsFixer.php
+++ b/Symfony/CS/AbstractConsecutiveBuiltInFunctionsFixer.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS;
+
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+abstract class AbstractConsecutiveBuiltInFunctionsFixer extends AbstractFixer
+{
+    /**
+     * @param Tokens $tokens
+     * @param int    $givenKind
+     */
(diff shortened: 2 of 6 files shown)


keradus:
Any chnaces to do some refactoring and unify logic shared for this and `combine_consecutive_unsets` fixers ? they ideas is the same, just different function to handle

Soullivaneuh:
@keradus What do you mean by "unify logic shared for this"?

The ideas are the same but the process is different:

* For unset, it merges multiple calls separated by `;`.
* For isset, it merges multiple calls separated by `&&`.

Soullivaneuh:
@keradus I did a quick refactor, but test are failing on unset.

Didn't have the time right now if this is justified or not, will do later.

If you have some clues to give me, I'm hearing! :wink:

SpacePossum:
failures seems legit, somewhere the comment-tokens are lost in the process of fixing

Soullivaneuh:
@SpacePossum I start reworking unset test on a separated commit.

Some comments are not placed by the same way, but I'm not sure this is really an issue.

For example, if we have this:

```php
<?php //1
    unset($foo/*;*/);
    unset(/*;*/$bar, $c ); //test
    unset($foobar  ); /* more comment test*/
    unset(  $foobar2);
```

And we use this fixer, I think that in any case the comment should be reworked manually. Because there is no "proper/recommended" way to place them after combine unset functions...

Am I wrong? Don't hesitate!

SpacePossum:
Disclaimer I didn't went through your changes and all of this is IMO ;)
IIRC I wrote the `unset` fixer doing something like;
- find `unset` call
- is there is a next one;
- move next one into previous one, leave comments where they are
- fix the white space inside the modified (first) `unset`
- repeat from 2

In respect to the comments I care only about;
- these shouldn't be inside the statement
- these shouldn't be removed

So I agree that it is OK to leave the comments to be reworked manually where needed.
I don't care much about what else is done to the comments and the white space around the comments, a best effort approach would do (i.e. the behavior should be the same each run and "sane").

If you rework the fixer it is OK to change some of it behavior related to the comments, so the tests might need to be reworked as well. I tend to add tests for all kind of case with white space and comments so it is clear what the fixer does with these and to test that the fixer will keep on working (i.e. doesn't throw exceptions etc.).

keradus:
any news here ?

keradus:
any news here ?

keradus:
@Soullivaneuh ?

julienfalque:
@Soullivaneuh This PR is getting old. I changed the target branch, any chance you can find some time to update it? :)

keradus:
ping @Soullivaneuh

keradus:
@Soullivaneuh , this PR got pretty outdated. While I do appreciate whole great work here, please let us know do you want to finish it and let us release it ;)

Soullivaneuh:
@keradus Sorry, time is precious and rare, as usual... I'll try to get a look as soon as possible.

keradus:
Closing due to merge of #3012

Issue #1592, opened 2015-12-12
Fixer Proposal


GrahamCampbell:
It would be good to have a fixer to change:

```php
<?php

function foo($baz)
{
    if ($baz) {
        var_dup('baz was truthy');
    }
    var_dup('hi there');
}

```

into:

```php
<?php

function foo($baz)
{
    if ($baz) {
        var_dup('baz was truthy');
    }

    var_dup('hi there');
}

```

gharlan:
-> #1122

keradus:
Closing as duplicate ;)

GrahamCampbell:
LOL. Duplicated myself.

PR #1985, opened 2016-06-11
Added ClassKeywordRemoveFixer

### LINKED ISSUE #1975: Class keyword fixers
Doc ref: http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class

My proposal is to create two fixers:
- [ ] `class_keyword`: Replace string like `'Foo\Bar'` by `Bar::class` and add the corresponding import.
- [x] `class_keyword_remove`: Replace all `Foo::class` by the string namespace, like `'Foo\Bar'`. #1985

The first one will be executed only if possible. Example with `'Foo\Bar'`, if a `Bar` class is already imported, this won't do nothing.

Those fixers have to be executed before `ordered_imports`.

Note: Be aware of classes with double dashes: `'Foo\\Bar\\Class'`

Edit: If you think you will need those fixers, vote :+1: on this message.

Soullivaneuh:
Partially closes #1975.

- [x] Make the fixer working
- [x] Integration test with `unused_use` fixer

Please don't hesitate for additional test cases!

### DIFF
--- a/Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
@@ -1,0 +1,202 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class ClassKeywordRemoveFixer extends AbstractFixer
+{
+    /**
+     * @var string[]
+     */
+    private $imports = array();
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        $this->replaceClassKeywords($tokens);
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * Replaces ::class keyword, namespace by namespace.
+     *
+     * It uses recursive method to get rid of token index changes.
+     *
+     * @param Tokens $tokens
+     * @param int    $namespaceNumber
+     */
+    private function replaceClassKeywords(Tokens $tokens, $namespaceNumber = -1)
+    {
+        $namespaceIndexes = array_keys($tokens->findGivenKind(T_NAMESPACE));
+
+        // Namespace blocks
+        if (!empty($namespaceIndexes) && isset($namespaceIndexes[$namespaceNumber])) {
+            $startIndex = $namespaceIndexes[$namespaceNumber];
+
+            $namespaceBlockStartIndex = $tokens->getNextTokenOfKind($startIndex, array(';', '{'));
+            $endIndex = $tokens[$namespaceBlockStartIndex]->equals('{')
+                ? $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $namespaceBlockStartIndex)
+                : $tokens->getNextTokenOfKind($namespaceBlockStartIndex, array(T_NAMESPACE));
+            $endIndex = $endIndex ?: $tokens->count() - 1;
+        } elseif (-1 === $namespaceNumber) { // Out of any namespace block
+            $startIndex = 0;
+            $endIndex = !empty($namespaceIndexes) ? $namespaceIndexes[0] : $tokens->count() - 1;
+        } else {
+            return;
+        }
+
+        $this->storeImports($tokens, $startIndex, $endIndex);
+        $tokens->rewind();
+        $this->replaceClassKeywordsSection($tokens, $startIndex, $endIndex);
+        $this->replaceClassKeywords($tokens, $namespaceNumber + 1);
+    }
+
+    /**
+     * @param Tokens $tokens
+     */
+    private function storeImports(Tokens $tokens, $startIndex, $endIndex)
+    {
+        $this->imports = array();
+
+        foreach ($tokens->getImportUseIndexes() as $index) {
+            if ($index < $startIndex || $index > $endIndex) {
+                continue;
+            }
+
+            $import = '';
+            while (($index = $tokens->getNextMeaningfulToken($index))) {
+                if ($tokens[$index]->equalsAny(array(';', '{')) || $tokens[$index]->isGivenKind(T_AS)) {
+                    break;
+                }
+
+                $import .= $tokens[$index]->getContent();
+            }
+
+            // Imports group (PHP 7 spec)
+            if ($tokens[$index]->equals('{')) {
+                $groupEndIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
+                $groupImports = array_map(
+                    'trim',
+                    explode(',', $tokens->generatePartialCode($index + 1, $groupEndIndex - 1))
+                );
+                foreach ($groupImports as $groupImport) {
+                    $groupImportParts = array_map('trim', explode(' as ', $groupImport));
+                    if (2 === count($groupImportParts)) {
+                        $this->imports[$groupImportParts[1]] = $import.$groupImportParts[0];
+                    } else {
+                        $this->imports[] = $import.$groupImport;
+                    }
+                }
+            } elseif ($tokens[$index]->isGivenKind(T_AS)) {
+                $aliasIndex = $tokens->getNextMeaningfulToken($index);
+                $alias = $tokens[$aliasIndex]->getContent();
+                $this->imports[$alias] = $import;
+            } else {
+                $this->imports[] = $import;
+            }
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     */
+    private function replaceClassKeywordsSection(Tokens $tokens, $startIndex, $endIndex)
+    {
+        $CTClassTokens = $tokens->findGivenKind(CT_CLASS_CONSTANT, $startIndex, $endIndex);
+        if (!empty($CTClassTokens)) {
+            $this->replaceClassKeyword($tokens, current(array_keys($CTClassTokens)));
+            $this->replaceClassKeywordsSection($tokens, $startIndex, $endIndex);
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $classIndex
+     */
+    private function replaceClassKeyword(Tokens $tokens, $classIndex)
+    {
+        $classEndIndex = $classIndex - 2;
+        $classBeginIndex = $classEndIndex;
+        while ($tokens[--$classBeginIndex]->isGivenKind(array(T_NS_SEPARATOR, T_STRING)));
+        ++$classBeginIndex;
+        $classString = $tokens->generatePartialCode($classBeginIndex, $classEndIndex);
+
+        $classImport = false;
+        foreach ($this->imports as $alias => $import) {
+            if ($classString === $alias) {
+                $classImport = $import;
+                break;
+            }
+
+            $classStringArray = explode('\\', $classString);
+            $namespaceToTest = $classStringArray[0];
+
+            if (0 === strcmp($namespaceToTest, substr($import, -strlen($namespaceToTest)))) {
+                $classImport = $import;
+                break;
+            }
+        }
+
+        $tokens->clearRange($classBeginIndex, $classIndex);
+        $tokens->insertAt($classBeginIndex, new Token(array(
+            T_CONSTANT_ENCAPSED_STRING,
+            "'".$this->makeClassFQN($classImport, $classString)."'",
+        )));
+    }
+
+    /**
+     * @param string|false $classImport
+     * @param string       $classString
+     *
+     * @return string
+     */
+    private function makeClassFQN($classImport, $classString)
+    {
+        if (false === $classImport) {
+            return $classString;
+        }
+
+        $classStringArray = explode('\\', $classString);
+        $classStringLength = count($classStringArray);
+        $classImportArray = explode('\\', $classImport);
+        $classImportLength = count($classImportArray);
+
+        if (1 === $classStringLength) {
+            return $classImport;
+        }
+
+        return implode('\\', array_merge(
+            array_slice($classImportArray, 0, $classImportLength - $classStringLength + 1),
+            $classStringArray
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Converts ::class keywords to FQCN strings.';
+    }
+}
--- a/README.rst
+++ b/README.rst
@@ -616,6 +616,10 @@ Choose from the list of available fixers:
                         Align equals symbols in
                         consecutive lines.
 
+* **class_keyword_remove** [contrib]
+                        Converts ::class keywords to
+                        FQCN strings.
+
 * **combine_consecutive_unsets** [contrib]
                         Calling unset on multiple
                         items should be done in one
--- a/Symfony/CS/Tests/Fixer/Contrib/ClassKeywordRemoveFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Contrib/ClassKeywordRemoveFixerTest.php
@@ -1,0 +1,205 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Sullivan Senechal <soullivaneuh@gmail.com>
+ */
+final class ClassKeywordRemoveFixerTest extends AbstractFixerTestBase
+{
+    /**
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                "<?php
+                use Foo\Bar\Thing;
+
+                echo 'Foo\Bar\Thing';
+                ",
+                "<?php
+                use Foo\Bar\Thing;
+
+                echo Thing::class;
+                ",
+            ),
+            array(
+                "<?php
+                use Foo\Bar;
+            
+                echo 'Foo\Bar\Thing';
+                ",
+                "<?php
+                use Foo\Bar;
+            
+                echo Bar\Thing::class;
+                ",
+            ),
+            array(
+                "<?php
+                use Foo\Bar\Thing as Alias;
+
+                echo 'Foo\Bar\Thing';
+                ",
+                "<?php
+                use Foo\Bar\Thing as Alias;
+
+                echo Alias::class;
+                ",
+            ),
+            array(
+                "<?php
+                use Foo\Bar\Dummy;
+                use Foo\Bar\Thing as Alias;
+
+                echo 'Foo\Bar\Dummy';
+                echo 'Foo\Bar\Thing';
+                ",
+                "<?php
+                use Foo\Bar\Dummy;
+                use Foo\Bar\Thing as Alias;
+
+                echo Dummy::class;
+                echo Alias::class;
+                ",
+            ),
+            array(
+                "<?php
+                echo '\DateTime';
+                ",
+                "<?php
+                echo \DateTime::class;
+                ",
+            ),
+            array(
+                "<?php
+                echo 'Thing';
+                ",
+                '<?php
+                echo Thing::class;
+                ',
+            ),
+            array(
+                "<?php
+                class Foo {
+                    public function amazingFunction() {
+                        echo 'Thing';
+                    }
+                }
+                ",
+                '<?php
+                class Foo {
+                    public function amazingFunction() {
+                        echo Thing::class;
+                    }
+                }
+                ',
+            ),
+            array(
+                "<?php
+                namespace A\B;
+
+                use Foo\Bar;
+
+                echo 'Foo\Bar';
+                ",
+                '<?php
+                namespace A\B;
+
+                use Foo\Bar;
+
+                echo Bar::class;
+                ',
+            ),
+            array(
+                "<?php
+
+                namespace A\B {
+
+                    class D {
+
+                    }
+                }
+
+                namespace B\B {
+                    class D {
+
+                    }
+                }
+
+                namespace C {
+                    use A\B\D;
+                    var_dump('A\B\D');
+                }
+
+                namespace C1 {
+                    use B\B\D;
+                    var_dump('B\B\D');
+                }
+                ",
+                '<?php
+
+                namespace A\B {
+
+                    class D {
+
+                    }
+                }
+
+                namespace B\B {
+                    class D {
+
+                    }
+                }
+
+                namespace C {
+                    use A\B\D;
+                    var_dump(D::class);
+                }
+
+                namespace C1 {
+                    use B\B\D;
+                    var_dump(D::class);
+                }
+                ',
+            ),
+            array(
+                "<?php
+                use Foo\\Bar\{ClassA, ClassB, ClassC as C};
+                use function Foo\\Bar\{fn_a, fn_b, fn_c};
+                use const Foo\\Bar\{ConstA, ConstB, ConstC};
+
+                echo 'Foo\\Bar\ClassB';
+                echo 'Foo\\Bar\ClassC';
+                ",
+                '<?php
+                use Foo\Bar\{ClassA, ClassB, ClassC as C};
+                use function Foo\Bar\{fn_a, fn_b, fn_c};
+                use const Foo\Bar\{ConstA, ConstB, ConstC};
+
+                echo ClassB::class;
+                echo C::class;
+                ',
+            ),
+        );
+    }
+}
--- a/Symfony/CS/Tests/Fixtures/Integration/priority/class_keyword_remove,unused_use.test
+++ b/Symfony/CS/Tests/Fixtures/Integration/priority/class_keyword_remove,unused_use.test
@@ -1,0 +1,15 @@
+--TEST--
+Integration of fixers: class_keyword_remove,unused_use.
+--CONFIG--
+level=none
+fixers=class_keyword_remove,unused_use
+--EXPECT--
+<?php
+
+echo 'Foo\Bar\Thing';
+
+--INPUT--
+<?php
+use Foo\Bar\Thing;
+
+echo Thing::class;


keradus:
I consider this fixer harmful and encouraging bad habits.
Changing `Foo\Bar` into `Bar::class` when possible (and that is tricky) would be awesome, but the other way around is not nice.

Soullivaneuh:
It's not about encouraging bad habits.

`::class` is not supported on PHP 5.3 and 5.4. We have projects with some maintained branches still supporting those versions. Keeping class string instead of `::class` will avoid use merge conflicts nightmare between branches.

BTW, the fixer is under the contrib section. So what is the problem? :wink:

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -0,0 +1,123 @@
...
+        $classImport = current(array_filter($this->imports, function ($import) use ($classString) {
+            $namespaceToTest = explode('\\', $classString)[0];
+
+            return 0 === strcmp($namespaceToTest, substr($import, - strlen($namespaceToTest)));
+        }));
+
+        $tokens->clearRange($classBeginIndex, $tokens->getNextMeaningfulToken($doubleColonIndex));
+        $tokens->insertAt($classBeginIndex, [
+            new Token("'"),
+            new Token($this->makeClassFQN($classImport, $classString)),
```
Soullivaneuh: The passed value is a string representing a class name:

* `Thing`
* `Foo\Bar\Thing`
* etc...

I don't understand the error: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/137230497#L283

Any idea? cc @keradus

SpacePossum: the sort array syntax is not going to work btw.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json#L17

SpacePossum: and the value should be one token with a string holding the class name, for example:
```
"'Foo\\Bar\\Thing'"
```
and not 3 tokens

Soullivaneuh: > the sort array syntax is not going to work btw.

Yes I know, this is a WIP, I will cleanup at the end.

And I don't this is not the problem, I targeted a job under PHP 5.5.

SpacePossum: and the `and not 3 tokens` ? didn't test but I think that is what Travis tries to tell you ;)

Soullivaneuh: With this:

```php
$tokens->insertAt($classBeginIndex, new Token("'".$this->makeClassFQN($classImport, $classString)."'"));
```

I got this:

```bash
1) Symfony\CS\Tests\Fixer\Contrib\ClassKeywordRemoveFixerTest::testFix with data set #0 ('<?php\n                use Foo...      ', '<?php\n                use Foo...      ')
The token at index 13 should be [318,"'Foo\\Bar\\Thing'"], got {
    "id": null,
    "name": null,
    "content": "'Foo\\Bar\\Thing'",
    "line": null,
    "isArray": false
}
```

And same for other test cases.

Soullivaneuh: Also with:

```php
$tokens->insertAt($classBeginIndex, new Token([T_STRING, "'".$this->makeClassFQN($classImport, $classString)."'"]));
```

Got:

```bash
1) Symfony\CS\Tests\Fixer\Contrib\ClassKeywordRemoveFixerTest::testFix with data set #0 ('<?php\n                use Foo...      ', '<?php\n                use Foo...      ')
The token at index 13 should be [318,"'Foo\\Bar\\Thing'"], got {
    "id": 310,
    "name": "T_STRING",
    "content": "'Foo\\Bar\\Thing'",
    "line": null,
    "isArray": true
}
```

The `318` id corresponds to a `T_DO` token, this make no sense...

Soullivaneuh: Of course, if I try:

```php
$tokens->insertAt($classBeginIndex, new Token([318, "'".$this->makeClassFQN($classImport, $classString)."'"]));
```

This works like a charm. But: :pig:

SpacePossum: try `T_CONSTANT_ENCAPSED_STRING`

Soullivaneuh:
I added an integration test case, the PR is ready for review. :+1:

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
keradus: for future: `token_name(318)`

keradus:
It's not - Travis is red

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
Soullivaneuh: @keradus Why I didn't think about that? --"

Soullivaneuh:
@keradus Should be fixed now.

keradus:
What will happen when I run the fixer on PHP 5.4 ?

Soullivaneuh:
> What will happen when I run the fixer on PHP 5.4 ?

https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/137301792

I think it works because I'm looking for `T_DOUBLE_COLON` which exists on 5.4 and then for `class` string.

GrahamCampbell:
This needs to be risky. I know of a PHP package that accepts partial class names as a paramater.

GrahamCampbell:
http://gentlero.bitbucket.org/bitbucket-api/

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -0,0 +1,123 @@
...
+                $import .= $tokens[$index]->getContent();
+            }
+            $this->imports[] = $import;
+        }
+    }
+
+    private function replaceClassKeywords(Tokens $tokens)
+    {
+        foreach ($tokens->findGivenKind(T_DOUBLE_COLON) as $index => $token) {
+            if ('class' !== $tokens[$tokens->getNextMeaningfulToken($index)]->getContent()) {
```
keradus: do not compare content. compare tokens. use `->equals`.

or...
```php
foreach ($tokens as $token) {
    if (!$token->isGivenKind(CT_CLASS_CONSTANT)) {
        continue;
    }

    // do your logic
}

**Comment by GrahamCampbell** [comment]
This is broken in the case that a class being imported conflicts with something else already in the namespace...

**Comment by GrahamCampbell** [comment]
> I consider this fixer harmful

Agreed. It fails on loads of cases and breaks the code too. (not the syntax, but will break when the autoloader starts requiring files)

File: Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php (hunk 2 above)
Comment (Soullivaneuh) [inline Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php]: Oh, this is maybe why `->equals` didn't work.

**Comment by Soullivaneuh** [comment]
> This needs to be risky. I know of a PHP package that accepts partial class names as a paramater.
> http://gentlero.bitbucket.org/bitbucket-api/

Could you please give me a sample?

> This is broken in the case that a class being imported conflicts with something else already in the namespace...

Could you elaborate? Or even better, give me a sample to add on tests?

>> I consider this fixer harmful

> Agreed.

Then don't use it. This is why it's on the contrib section. :wink: 

> It fails on loads of cases

Which one? I will happy to update tests.

> and breaks the code too. (not the syntax, but will break when the autoloader starts requiring files)

Not sure to understand, sorry. :-/

**Comment by keradus** [comment]
autoloader... nice catch!

**Comment by Soullivaneuh** [comment]
> autoloader... nice catch!

Still didn't understand the problem with the autoloader... ^^

**Comment by Soullivaneuh** [comment]
> > What will happen when I run the fixer on PHP 5.4 ?

> https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/137301792

Sorry for that, test are skipped. But they works, I removed the requirement.

**Comment by GrahamCampbell** [comment]
@Soullivaneuh Consider this case:

FILE: src/Foo/Bar.php
```php
<?php

namespace Baz/Foo;

class Bar
{
    // such class, much amaze
}
```

FILE: src/Foo/Oops.php
```php
<?php

namespace Baz/Foo;

class Oops
{
    public function lol()
    {
        return 'Baz\Bar';
    }
}
```

See how importing `Baz\Bar` will break everything.

Soullivaneuh:
> See how importing `Baz\Bar` will break everything.

Except... I don't import anything here?

Soullivaneuh:
This PR is for `class_keyword_remove` fixer, not `class_keyword`. #1975

Soullivaneuh:
Test passed: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/137307080

I squashed my commits.

Soullivaneuh:
I added some phpdoc comments.

For my side, the Pull Request is ready.

If you have some tests to give me, tell me! :+1:

keradus:
@GrahamCampbell 
Actually just calling `::class` won't call autoloader.
What's more, you could do `NonExistingClass::class` and get name of non existing class...

```bash
keradus@keradus-workLap:~/tmp$ cat A.php 
<?php

class A
{
}

echo __FILE__ . "\n";

keradus@keradus-workLap:~/tmp$ 
keradus@keradus-workLap:~/tmp$ cat index.php 
<?php

spl_autoload_register(function ($class) { require $class . '.php'; });

new A();

echo __FILE__ . "\n";

keradus@keradus-workLap:~/tmp$ 
keradus@keradus-workLap:~/tmp$ php -f index.php 
/home/keradus/tmp/A.php
/home/keradus/tmp/index.php
```

```bash
keradus@keradus-workLap:~/tmp$ cat A.php 
<?php

class A
{
}

echo __FILE__ . "\n";

keradus@keradus-workLap:~/tmp$ 
keradus@keradus-workLap:~/tmp$ cat index.php 
<?php

spl_autoload_register(function ($class) { require $class . '.php'; });

A::class;

echo __FILE__ . "\n";

keradus@keradus-workLap:~/tmp$ 
keradus@keradus-workLap:~/tmp$ php -f index.php 
/home/keradus/tmp/index.php
```

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     */
+    private function replaceClassKeywords(Tokens $tokens)
+    {
+        foreach ($tokens->findGivenKind(T_DOUBLE_COLON) as $index => $token) {
+            if (!$tokens[$tokens->getNextMeaningfulToken($index)]->isGivenKind(CT_CLASS_CONSTANT)) {
```
keradus: no need for checking `T_DOUBLE_COLON`.

Also, as in my prev provided example, dont use `findGivenKind` to iterate over whole collection, it will decrease the speed.

Soullivaneuh:
> What's more, you could do `NonExistingClass::class` and get name of non existing class...

Well, it not out fault if someone is using a non existing class, is it?

BTW, the fixer won't break anything:

```php
<?php

var_dump(NonExistingClass::class);
var_dump('NonExistingClass');
var_dump(NonExistingClass::class === 'NonExistingClass');
```

```bash
sullivan@sweethome:~/projects/fork/FriendsOfPHP/PHP-CS-Fixer(class-keyword-remove)$ php test.php 
string(16) "NonExistingClass"
string(16) "NonExistingClass"
bool(true)
```

This corresponds to this test case:

```php
array(
    "<?php
    echo 'Thing';
    ",
    '<?php
    echo Thing::class;
    ',
),
```

For the autoload case, AFAIK, `new A();` instantiates a new object, `A::class;` does... nothing special?

So I don't see the risk here, or I don't get the point. ;-)

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
Soullivaneuh: Indeed. The CT_CLASS_CONSTANT is not detected by my IDE, this is weird and why I didn't find it earlier.

Soullivaneuh:
@keradus updated. Is this what you need?

GrahamCampbell:
> Actually just calling ::class won't call autoloader.

Yeh, I know it won't. I didn't say it would? I'm saying that it could conflcit with existing autoloaded definitions.

GrahamCampbell:
What if we have `$foo->('Bar\Baz')`, but that is not a class name.

GrahamCampbell:
To be completely honest, I think this fixer is a horrible idea, sorry.

SpacePossum:
>What if we have $foo->('Bar\Baz'), but that is not a class name.

The fixer is the other way I think, i.e. `$foo->('Bar::class')`?

SpacePossum:
To be honest I wouldn't be using this fixer as it would "help" back porting PHP code, which I hope never to do.

Soullivaneuh:
> What if we have `$foo->('Bar\Baz')`, but that is not a class name.

This fixer would do **nothing** on this piece of code.

If you want to, I can add it on the test cases, but I think this will be pointless.

> The fixer is the other way I think, i.e. `$foo->('Bar::class')`?

Nearly true. Considering the `Bar\Baz`, it converts `$foo->(Baz::class)` onto `$foo->(`Bar\Baz`)`.

> To be completely honest, I think this fixer is a horrible idea, sorry.
> To be honest I wouldn't be using this fixer as it would "help" back porting PHP code, which I hope never to do.

Well OK. I love short array syntax and I love `::class` notation, really. This is **not** why I would like to add this fixer.

The only reason I would like to use it is to reduce merge conflict. Imagine two branches of a project: `branch-old` and `branch-new`, both still maintained.

`branch-old` support PHP 5.3+, `branch-new` PHP 5.5+.

If we convert this piece of code present on `branch-old`:

```php
$result = class_exist('Foo\Bar')
    ? 'thing'
    : 'another';
```

To this on `branch-new`:

```php
$result = class_exist(Bar::class)
    ? 'thing'
    : 'another';
```

And, for a bugfix or whatever, we have to change the class name on both `branch-old` and `branch-new`, we would start to do this on `branch-old`:

```php
$result = class_exist('Foo\AnotherBar')
    ? 'thing'
    : 'another';
```

What will happen when I'll merge `branch-old` to `branch-new`? A merge conflict.

And this is just a sample with a 3 lines piece of code. I let you imagine a thousand lines project with all class name replaced by the new `::class` notation on `branch-new`...

That's all. This is the only reason.

And I will stop arguing now. I really don't see why this should be discussed.

The fixer is **technically** acceptable and will not break any code.

You don't like it? Just don't use it, this is why I put it on the contrib section.

This is the case for lot of fixer. For example, some want aligned arrow, some other don't want to. This why we have both on this project and it works very well.

BTW, I would like to work on the reverse of this fixer (re-read #1975, I have two tasks). Because yes, when we will not have any maintained branch supporting PHP <5.5, I would love to get rid of this old and ugly syntax to go on `::class`

I really hope you will understand.

Now please let's concentrate on if the fixer is technically reliable. Another test cases are still welcomed.

keradus:
> The fixer is technically acceptable and will not break any code.
> You don't like it? Just don't use it

There is other thing. Maintenance. Even if you made hard work to create a fixer, then sb will need to maintain it. That is why we put the fixers that more than one people will use. Because later we need to maintain it. So if we need to make our life harder and maintain more code, we want that code to be useful for more people. That's why we usually discuss/vote about fixers requests/ideas before implementing them.
I created a lot of my own fixers to do one-time job or have them only in one project, but I know that there is no reason to keep them here.

I would strongly recommend to freeze the work on that fixer and come back to discussing feature request in 1975 itself.

Soullivaneuh:
> There is other thing. Maintenance. Even if you made hard work to create a fixer, then sb will need to maintain it. That is why we put the fixers that more than one people will use. Because later we need to maintain it.

This fixer will be dispatched on all Sonata projects (https://github.com/sonata-project/dev-kit/issues/166) while PHP 5.3 will be supported.

And it will still be supported for a while (https://github.com/sonata-project/dev-kit/issues/153) so trust me, it will be maintained, at least by me, for a moment.

Plus, this fixer is quite small, no big things to maintain here.

To finish, this fixer can be simply remove once you will drop support for PHP <5.5 on this project, because keeping this one would indedd not making any sense.

I understand you want to merge fixer that will be used. This is the case of this one, for at least 30 projects.

> I created a lot of my own fixers to do one-time job or have them only in one project, but I know that there is no reason to keep them here.

Yeah but in this case there is a public need and it would be great to have this one to use it with additional tools, like StyleCI.

On the other way, is that easy to create/**use** per-project custom fixer? Any documentation for that?

Thanks.

Soullivaneuh:
I would add that my merge conflict argument for this fixer is receivable. Maybe not liked by everyone but **technically** receivable and can make sens. Is that not sufficient to accept it?

keradus:
> This fixer will be dispatched on all Sonata projects (sonata-project/dev-kit#166) while PHP 5.3 will be supported.

without changes made by fixer your jobs based on 5.3 will break

-----------------------------------------

Again, let us focus on feature request issue to decide should we or should we not have this fixer.
Disclaimer: I won't vote, I leave it to community.

GrahamCampbell:
@keradus My vote is :-1: for allow ANY implementation in. I don't like the idea. Sorry @Soullivaneuh.

Soullivaneuh:
> without changes made by fixer your jobs based on 5.3 will break

What do you mean? The fixer work under PHP 5.3: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/137342439

And if the PHP syntax is invalid, the fixer will just flag it as invalid.

The goal is not to avoid `::class` on PHP 5.3 branches but on all branches for the reason I said previously: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-225724643

> @keradus My vote is :-1: for allow ANY implementation in. I don't like the idea. Sorry @Soullivaneuh.

As you said, you don't like. But I don't think this is a matter of taste...

@keradus I hope the decision would be based on +1 vote uniquely. Because if 700 persons does not like this feature but 300 peoples want this for example, it's steel 300 persons wanted this feature.

This is also why this kind of fixer is under contrib and, AFAIK, PHP-CS-Fixer is a tool to apply fixer, not to define which CS style we should accept or not, like PSR or Symfony coding standard.

GrahamCampbell:
> But I don't think this is a matter of taste...

I'm not saying I don't use `::class` in my code, I'm saying it makes no sense to have a fixer to change every occurrence, some of the changes being wrong anyway.

keradus:
> > without changes made by fixer your jobs based on 5.3 will break

> What do you mean? 

You say you need a fixer for lower PHP version. If you won't have a fixer your 5.5 code will simply fail on 5.3. It's not about coding standards but language syntax.

keradus:
> This is also why this kind of fixer is under contrib and, AFAIK, PHP-CS-Fixer is a tool to apply fixer, not to define which CS style we should accept or not, like PSR or Symfony coding standard.

And especially for that one could have his own fixers.
If the rule would be used by many - let them vote and have the rule, awesome. If the vote will be used by few - let them have it locally.

Soullivaneuh:
> You say you need a fixer for lower PHP version. If you won't have a fixer your 5.5 code will simply fail on 5.3. It's not about coding standards but language syntax.

Please re-read https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-225724643.

This is not about the branch with php 5.3+ support, but the new one with php 5.5+ support. This is to avoid some PR with `::class` fixer rename that will produce **merge conflicts**.

> I'm not saying I don't use ::class in my code,

Are you **really** sure you understood the goal of this PR? This one does not add `::class` but **removes** them.

In the case I explained in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-225724643, this fixer will be here for **preventive** reason. Because of course, there is actually no `::class` token anywhere on Sonata because we still support PHP 5.3 on **all** branches. But this is gonna be changed.

> I'm saying it makes no sense to have a fixer to change every occurrence, some of the changes being wrong anyway.

Well, change the `long_array_syntax` to `short_array_syntax` on your config, it will do the same thing: Change every occurrence.

Soullivaneuh:
> If the vote will be used by few - let them have it locally.

* Could you please define "few"? Because I don't think every user are looking for all issues of this project...
* From https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-225840869: On the other way, is that easy to create/use per-project custom fixer? Any documentation for that?

Thanks.

greg0ire:
> You say you need a fixer for lower PHP version. If you won't have a fixer your 5.5 code will simply fail on 5.3. It's not about coding standards but language syntax.

Hi! Sonata member here… if I can help shedding a light on this situation : we want to avoid people using `::class` on the unstable branch even if the unstable branch does not support 5.3 anymore. The goal is avoiding unnecessary differences between branches, because these will make changes harder to merge from legacy stable, to stable, end then to unstable.

If someone were to use `::class` in the unstable branch, the build would no fail on 5.3 because unstable has no job for 5.3 at all, so we cannot expect the unit tests to make us notice that. This is why this is needed.

GrahamCampbell:
Hmmm. I guess I'm actually neutral on having a fixer to remove `::class`, but very, very, very negative on having a fixer to go the other way.

SpacePossum:
I'm still +0, I simply have no need for it. I see there are others that have, I wish them good luck with it. The reason I do not to up vote it is that I've some concerns about the life time of the fixer and the amount of support on maintenance can be expected.

On the topic of `string` to `::class`, I've big reservation about that one. It is simply impossible to do this reliable based on the tokens of a single file. Even a best effort approach will result in a lot of false positives, which in time results in a lot of issues reported. Placing the fixer in contrib and marking it risky will not stop this from happen. I think it would be a great fixer that have, but unfortunately impossible to make.

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @param Tokens $tokens
+     */
+    private function storeImports(Tokens $tokens)
+    {
+        foreach ($tokens->getImportUseIndexes() as $index) {
+            $import = '';
+            while (($index = $tokens->getNextMeaningfulToken($index))) {
+                if ($tokens[$index]->equals(';')) {
```
SpacePossum: imports can end by a close tag (`T_CLOSE_TAG`) as well, for example (and please at to your PHP 7 tests);
```php
<?php use const some\test\{ConstA, ConstB, ConstC};
use function some\test\{fn_a, fn_b, fn_c} ?>
```

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -0,0 +1,129 @@
...
+                    break;
+                }
+                $import .= $tokens[$index]->getContent();
```
SpacePossum: missing empty line after `}`

SpacePossum:
Test request:
```php
<?php

namespace A\B {

    class D {

    }
}

namespace B\B {
    class D {

    }
}

namespace C {
    use A\B\D;
    var_dump(D::class);
}

namespace C1 {
    use B\B\D;
    var_dump(D::class);
}
```

Soullivaneuh:
> I've some concerns about the life time of the fixer

Simple: This fixer can be removed when PHP-CS-Fixer will drop support for PHP 5.3 and 5.4.

> and the amount of support on maintenance can be expected.

As I said, we have 30 Sonata projects. All will have legacy and maintained branch supporting PHP 5.3. So the fixer will be maintains at least by us for a while.

And, this is not a complicated fixer, not hard to maintain IMHO.

Thanks for the diff comments and the test request, I'll work on it. :+1:

SpacePossum:
> This fixer can be removed when PHP-CS-Fixer will drop support for PHP 5.3 and 5.4.

Which is BC break as other might start depend on it, so we need to take into consideration before adding.

Soullivaneuh:
> Which is BC break as other might start depend on it, so we need to take into consideration before adding.

Dropping PHP version support is BC break according to your rules AFAIK.

So dropping PHP version will be done on next major. The fixer removal can be done at the same time.

After that, I said it can be removed because it will make no really sense. But this one works great on ANY  PHP version.

SpacePossum:
>Dropping PHP version support is BC break according to your rules AFAIK.

Indeed, but introducing one BC break doesn't automatically justify another. Each BC break should be considered on itself.

>After that, I said it can be removed because it will make no really sense. But this one works great on ANY PHP version.

didn't question that :)

Soullivaneuh:
> Indeed, but introducing one BC break doesn't automatically justify another. Each BC break should be considered on itself.

Not sure to understand. How this cannot be justify? Example for a new major reasons:

* Drop PHP <5.5 support
* Drop `class_keyword_remove` as this is now irrelevant

And if you think PHP <5.5 support dropping is not a justification to remove this fixer, well just not remove it. :-)

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
Soullivaneuh: Great! And I suppose `use const ` and `use function ` should not be considered as it's not classes?

Am I right?

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
Soullivaneuh: You mean this?

```php
if ($tokens[$index]->equals(';')) {
    break;
}

$import .= $tokens[$index]->getContent();
```

Is that a CS requirement? This was not done by php-cs-fixer.

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
SpacePossum: Indeed :)
But please test something like as well, these are class uses
```php
use some\test\{ClassA, ClassB, ClassC as C};
use some\test\{ClassD, ClassE, ClassF}?>
```

Thanks for adding the test, however you need a little PHP7 req. wrapper for when the test are run on < PHP7

Example;
```php
/**
     * @dataProvider provide70Cases
     * @requires PHP 7.0
     */
    public function test70($expected, $input = null)
    {
         $this->makeTest($expected, $input);
    }
 
    public function provide70Cases()
    {
        return array(
            array()
        );
    }
```

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
SpacePossum: Fixer is not written, for ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1592

GrahamCampbell: yes, it is a CS requirement

Soullivaneuh: Right, good to know. :+1:

Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php
```diff
@@ -78,12 +85,17 @@ private function replaceClassKeyword(Tokens $tokens, $classIndex)
...
+            if ($classString === $alias) {
+                return $import;
+            }
+
             $classStringArray = explode('\\', $classString);
             $namespaceToTest = $classStringArray[0];
 
             return 0 === strcmp($namespaceToTest, substr($import, -strlen($namespaceToTest)));
-        }));
+        }, ARRAY_FILTER_USE_BOTH));
```
SpacePossum: `ARRAY_FILTER_USE_BOTH`
was added in PHP 5.6, so you cannot use it on the min. supported version of PHP 5.3, sorry

Soullivaneuh: Argh, so bad... :-/

Soullivaneuh:
I don't know how to resolve this fail: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/137739385#L303

If I try this:

```php
private function replaceClassKeywords(Tokens $tokens)
{
    foreach ($tokens->findGivenKind(CT_CLASS_CONSTANT) as $index => $token) {
        var_dump($tokens[$index]);
        $this->replaceClassKeyword($tokens, $index);
    }
}
```

I got this:

```
object(Symfony\CS\Tokenizer\Token)#50 (4) {
  ["content":"Symfony\CS\Tokenizer\Token":private]=>
  string(5) "class"
  ["id":"Symfony\CS\Tokenizer\Token":private]=>
  int(10004)
  ["isArray":"Symfony\CS\Tokenizer\Token":private]=>
  bool(true)
  ["line":"Symfony\CS\Tokenizer\Token":private]=>
  int(5)
}
object(Symfony\CS\Tokenizer\Token)#56 (4) {
  ["content":"Symfony\CS\Tokenizer\Token":private]=>
  string(2) "::"
  ["id":"Symfony\CS\Tokenizer\Token":private]=>
  int(384)
  ["isArray":"Symfony\CS\Tokenizer\Token":private]=>
  bool(true)
  ["line":"Symfony\CS\Tokenizer\Token":private]=>
  int(6)
}
```

The second should be `class`, not `::`, should not be?

Any idea of why?

Soullivaneuh:
Ok apparently caused by tokens modification. Fixed that with a recursive way.

keradus:
About PHP version.
If one have a package version 1.1 which reqs PHP:^5.3 he could have package version 1.2 which reqs PHP:^5.4 or even PHP:^7. Bumping requirements is not BC breaker due to official SEMVER.
I totally could understand the opposite thinking and if it would be up to me I would do it that way, but it's not.

Soullivaneuh:
@keradus This is why I said "according to **your** rules", but apparently it's not.

In this case, the fixer could stay here after version bumping because it will still working, and remove it on a next major.

greg0ire:
> That **would** be considered compatible since it does not affect the public API. Software that explicitly depends on the same dependencies as your package **should** have their own dependency specifications and the author will notice any conflicts. Determining whether the change is a patch level or minor level modification depends on whether you updated your dependencies in order to fix a bug or introduce new functionality. I **would usually** expect additional code for the latter instance, in which case it’s obviously a minor level increment.

So this looks a bit up-to-you in fact, doesn't it ?

keradus:
@Soullivaneuh I was referring only to bumping php version, not to remove the fixer itself

Soullivaneuh:
> @Soullivaneuh I was referring only to bumping php version, not to remove the fixer itself

Yes ofc! This was just a related addition. :wink:

keradus:
@greg0ire so for that, bumping deps is not BC breaker ;) and introducing BC break version to only bump deps is pointless.

greg0ire:
> bumping deps is not BC breaker

bumping deps **should** not be a BC breaker. It's still up-to you IMO

Soullivaneuh:
> @greg0ire so for that, bumping deps is not BC breaker ;) and introducing BC break version to only bump deps is pointless.

This depends of your branch management IMHO. For your case, it's OK because you have one branch per minor.

In Sonata, we dropped this when and we only maintain major branches. So we decide to bump big dependencies like SF or PHP only on new majors.

But you are right, according to semver, we are not forced to do that. :wink:

GrahamCampbell:
> But you are right, according to semver

How is this related to php-cs-fixer?

GrahamCampbell:
People can do what they want with BC in their own projects. That surely has no bearing on what fixers we implement?

Soullivaneuh:
> How is this related to php-cs-fixer?

PHP-CS-Fixer seems to follows semver rules.

> People can do what they want with BC in their own projects. That surely has no bearing on what fixers we implement?

Not sure to understand. We are talking about PHP-CS-Fixer, not own projects.

This is about https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-226112782.

BTW, @keradus answer me about that, we should not diverge the debate on this PR. :+1:

Soullivaneuh:
I fixed last tests and rebased the PR.

Nothing else to do IMHO.

As you can see with the upvotes on #1975 and this PR, this fixer give some interests.

This contrib fixer can add a real benefit for big project maintainers who care about avoiding merge conflicts and time as possible. This is the case for Sonata projects.

Would be really great to have this merged after review. :+1: 

Thanks.

Soullivaneuh:
AppVeyor fail not related to my PR.

keradus:
@GrahamCampbell what would be your vote after time that passed?

GrahamCampbell:
I'd give it a "meh".

Soullivaneuh:
So we have a "meh" and some :+1:.

@keradus What is your final decision? If you don't want to merge it, could you please at least give me a hint of how to properly integrate custom fixers globally as you said in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1985#issuecomment-225909188?

Thanks.

keradus:
Thank you @Soullivaneuh.

Soullivaneuh:
Thank you too @keradus!

PR #1992, opened 2016-06-15
Update README to instruct use of new repository 'friendsofphp'

vukanac:
Update `README` to instruct use of new repository `friendsofphp` instead of `fbpot`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ To install PHP-CS-Fixer, install Composer and issue the following command:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


keradus:
Hello @vukanac !
First, thank you for your willingness and contribution !

Your PR checks are red - it's because you didn't follow contribution guide that say to not modify readme file as is, as it's autogenerated via `readme` command.

Then, this PR was send for 1.x line, which is still supposed to work under `fabpot` vendor. Change of vendor was done for 2.x line and it's already updated in the readme.

For the following reasons I close this PR.
Thank you.

Issue #1119, opened 2015-03-31
[Meta] PHP 7 features


gharlan:
We should add test cases to relevant fixers for those new features:
- [x] new operators ([coalesce](https://wiki.php.net/rfc/isset_ternary), [spaceship](https://wiki.php.net/rfc/combined-comparison-operator)) #1064 
- [x] [Return Type Declarations](https://wiki.php.net/rfc/return_types)
- [x] [Anonymous Class Support](https://wiki.php.net/rfc/anonymous_classes): #1806, #1881, #1894, #1989, ...
- [x] [Group Use Declarations](https://wiki.php.net/rfc/group_use_declarations): ordered_use, unused_use etc.., and maybe we should add a new transformer for the braces (I'm not sure); #1989, #2235
- [ ] [Uniform Variable Syntax](https://wiki.php.net/rfc/uniform_variable_syntax): there are some new token sequences which may break some fixers
- [x] [Scalar Type Hints](https://wiki.php.net/rfc/scalar_type_hints_v5): I guess they are unproblematic because they are not so different to existing type hints
- [x] Semi-reserved keywords #1970

7.1 extracted: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2090

keradus:
Nice list !

SpacePossum:
hope I'm not de-railing this, but I have some questions about PHP7 support.

```php
use some\a\{ClassA, ClassB, ClassC as C};
use function some\a\{fn_a, fn_b, fn_c};
use const some\a\{ConstA, ConstB, ConstC};
```

when modifying related fixers to support this I wandered; do we want some sorting order on these?
Like `classes`, `functions` `const`

Should the groups be broken up into single `use` statements?
If not, should we sort the parts inside the braces alphabetically? 

Should the visibility fixer support the new `const` visibility introduced in PHP 7.1?  
(for ref.: https://wiki.php.net/rfc/class_const_visibility)

GrahamCampbell:
The `new_with_braces` fixer is also broken with anonymous classes.

keradus:
could you send a fix for it @GrahamCampbell ?

SpacePossum:
>The new_with_braces fixer is also broken with anonymous classes.

addressed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1989

SpacePossum:
```php
<?php
function arraysSum(array ...$arrays): INT
{
[...]

PR #1995, opened 2016-06-15
Rely on own phpunit, not one from CI service

keradus:
cc @SpacePossum

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
 
 script:
     - cp $HOME/xdebug.ini $HOME/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini || return 0
-    - phpunit --verbose --coverage-clover build/logs/clover.xml
+    - vendor/bin/phpunit --verbose --coverage-clover build/logs/clover.xml
     - phpenv config-rm xdebug.ini || return 0
     - php php-cs-fixer --diff --dry-run -v fix
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,12 +21,15 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
+    - mv $HOME/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini $HOME/xdebug.ini || return 0
     - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
 
 install:
+    - travis_retry composer global require hirak/prestissimo
     - travis_retry composer update ${COMPOSER_FLAGS} --no-interaction
 
 script:
+    - cp $HOME/xdebug.ini $HOME/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini || return 0
     - phpunit --verbose --coverage-clover build/logs/clover.xml
     - phpenv config-rm xdebug.ini || return 0
     - php php-cs-fixer --diff --dry-run -v fix
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "sebastian/diff": "~1.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "@stable",
+        "phpunit/phpunit": "^4.5|^5",
         "satooshi/php-coveralls": "^0.7.1"
     },
     "autoload": {
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "sebastian/diff": "~1.1"
     },
     "require-dev": {
-        "satooshi/php-coveralls": "0.7.*@dev"
+        "phpunit/phpunit": "@stable",
+        "satooshi/php-coveralls": "^0.7.1"
     },
     "autoload": {
         "psr-4": { "Symfony\\CS\\": "Symfony/CS/" }


SpacePossum:
Thanks so much for picking this up!
I don't know much about the changes you made, I simply don't have the experience :)
I'll try to learn a bit about it, but looks good!

composer.json
```diff
@@ -25,7 +25,8 @@
         "sebastian/diff": "~1.1"
     },
     "require-dev": {
-        "satooshi/php-coveralls": "0.7.*@dev"
+        "phpunit/phpunit": "@stable",
```
GrahamCampbell: Wildcards are a bad idea.

GrahamCampbell: `@stable` is exactly the same as `*`

GrahamCampbell: This of course breaks prefer lowest too.

GrahamCampbell: Perhaps `~4.0|~5.0` will work fine here?

keradus: :+1: for `^4|^5` !

keradus:
@SpacePossum please temporary merge this PR into your once that has problems with AppVeyor and let see if it will help

.travis.yml
```diff
@@ -21,13 +21,16 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
+    - mv $HOME/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini $HOME/xdebug.ini || return 0
     - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'
 
 install:
+    - travis_retry composer global require hirak/prestissimo
```
GrahamCampbell: This actually ends up slower surely? It takes just as long to install this globally as it does to install the rest.

SpacePossum:
thanks, done over @ https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1949 (please note I made changes because that PR targets master, but as a POC it should be useful for now)

composer.json
```diff
@@ -25,7 +25,8 @@
         "sebastian/diff": "~1.1"
     },
     "require-dev": {
-        "satooshi/php-coveralls": "0.7.*@dev"
+        "phpunit/phpunit": "^4.5|^5",
```
Soullivaneuh: :-1: If we are able to not make it dependents, don't do it.

This will install a **lot** of dependencies, making composer longer to resolve.

Plus, all developer does not work with vendor/bin/phpunit.

What about directly download the `.phar` on travis?

Example: https://github.com/sonata-project/dev-kit/blob/a4574f4ff3428bdd89ecf1c2f4f0190227abb250/project/.travis/install_test.sh#L4-L13

composer.json
```diff
@@ -25,7 +25,8 @@
...
+        "satooshi/php-coveralls": "^0.7.1"
```
Soullivaneuh: Same here, you can get the `.phar`: https://github.com/sonata-project/dev-kit/blob/a4574f4ff3428bdd89ecf1c2f4f0190227abb250/project/.travis/install_test.sh#L15-L17

.travis.yml
Soullivaneuh: You mean about the time to insstall this library?

Dependencies for this one is very light, it should not be a problem.

@keradus I would ever say that the `travis_retry` should not be here.

composer.json
keradus: Disclaimer: I'm one of those who release php-coveralls

keradus:
So, this fixes AppVeyor bug of #1949 @SpacePossum 
(related Travis fail is due to quick cherry-pick without solving conflicts on .travis.yml)

.travis.yml
keradus: I see no reason to drop it.

composer.json
Soullivaneuh: Oh OK. :+1: 

And then? Don't you recommend to use `.phar`?

I'm curious to know why. :wink:

composer.json
keradus: I don't like fetching some deps by Composer and some not. You don't like it? Cool. You are not enforced to download dev deps.
Composer is nicer way to describe required version. like not supporting 4.1

composer.json
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1995#discussion_r67325506

use whatever you prefer ;)

SpacePossum:
Indeed, AppVeyor is happy, so :ship: this :)
(Maybe in some other life I try to figure out want went wrong with the previous setup)

composer.json
SpacePossum: because the dependencies (and conflicts) are not managed using `phar`'s as these are using composer (for ref.: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-2529)

keradus:
AppVeyor break it's PHPUnit, I remember that it happens once with Travis as well.
So, mergy ?

composer.json
Soullivaneuh: > I don't like fetching some deps by Composer and some not.

As you want but IMHO, it's more a tool than a "dependency".

> You don't like it? Cool. You are not enforced to download dev deps.

Not really true for all projects. Some can require dev packages other than phpunit and we can select which one we want to install.

By installing with composer, you get dependencies **described** by phpunit, not locked by `composer.lock` like for `.phar`. This mean you increase the issue risk to get PHPUnit not working because of a sub-dependency update.

So this is not only a matter of taste for me. :wink: 

> Composer is nicer way to describe required version. like not supporting 4.1

Well, if you not support it, tests will fail. :-)

composer.json
Soullivaneuh: @SpacePossum This could be indeed an issue. But could be.

This is why I said on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1995#discussion_r67324037:

> If **we are able to** not make it dependents, don't do it.

BTW, a WIP is here concerning this issue: https://github.com/sebastianbergmann/phpunit/issues/2015

Soullivaneuh: @SpacePossum BTW, concerning your ref, this is unrelated.

Related issue: https://github.com/sebastianbergmann/phpunit/issues/2219

.travis.yml
Soullivaneuh: `travis_retry` is good for long with no output jobs.

This command would not take more than 10 minutes to execute.

Plus, `travis_retry` drop the original output of the command, harder to read.

This is as you want, but I suggest you to remove this. :+1:

Soullivaneuh:
So no concerns about my last comment? https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1995#discussion_r67326444 :-/

FYI, if you made this PR to fix this error: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-2529

Please not this is not related at all. This was an issue from PHPUnit: https://github.com/sebastianbergmann/phpunit/issues/2219

composer.json
keradus: 1. tool could be a dependency
2. to develop a project you should have a phpunit. you may have global one - cool, but we don't promise that our project is compatible with it.
3. theoretically, but such an edge case...
4. it;s better to say "this one will work, you may use it" than "some phpunit could work, we dont know, try, and if it fails - well... we dont know which version works"

keradus:
@Soullivaneuh answered.
To be honest I see no value in continuing this topic, it won't bring any new value. I like to be consistent and have one way of dealing with deps, you like to be minimalistic. I see good and bad points of both approaches.

If you have some power and willing please focus on something important, like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1119

composer.json
Soullivaneuh: > it;s better to say "this one will work, you may use it" than "some phpunit could work, we dont know, try, and if it fails - well... we dont know which version works"

Or simply rely on the last stable version of PHPUnit. :wink: 

> theoretically, but such an edge case...

Are you talking about the risk to get PHPUnit not working because of a sub-dependency update?

Not such an edge case as you think. :wink: 

Anyway, at least I have an answer. :-)

PR #1998, opened 2016-06-17
Cache should use absolute pathname

Slamdunk:
When using a finder like this in the `.php_cs` file:

``` php
    ->in(__DIR__ . '/path_one')
    ->in(__DIR__ . '/path_two')
;
```

And in both path you have the same file name (e.g. `bootstrap.php`) the `php-cs-fix` keeps trying to fix the second one (and subsequents) because the cache is saved using the relative pathname instead of the absolute, and so this https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v1.11.4/Symfony/CS/FileCacheManager.php#L65-L67 keeps beeing triggered if the files are different.

### DIFF
--- a/Symfony/CS/FileCacheManager.php
+++ b/Symfony/CS/FileCacheManager.php
@@ -34,6 +34,7 @@ class FileCacheManager
 
     private $dir;
     private $isEnabled;
+    private $isCacheAvailable;
     private $fixers;
     private $newHashes = array();
     private $oldHashes = array();
@@ -61,6 +62,8 @@ class FileCacheManager
             return true;
         }
 
+        $file = $this->getRelativeFilename($file);
+
         if (!isset($this->oldHashes[$file])) {
             return true;
         }
@@ -81,6 +84,8 @@ class FileCacheManager
             return;
         }
 
+        $file = $this->getRelativeFilename($file);
+
         $this->newHashes[$file] = $this->calcHash($fileContent);
     }
 
@@ -91,13 +96,11 @@ class FileCacheManager
 
     private function isCacheAvailable()
     {
-        static $result;
-
-        if (null === $result) {
-            $result = $this->isEnabled && (ToolInfo::isInstalledAsPhar() || ToolInfo::isInstalledByComposer());
+        if (null === $this->isCacheAvailable) {
+            $this->isCacheAvailable = $this->isEnabled && (ToolInfo::isInstalledAsPhar() || ToolInfo::isInstalledByComposer());
         }
 
-        return $result;
+        return $this->isCacheAvailable;
     }
 
     private function isCacheStale($cacheVersion, $fixers)
@@ -155,5 +158,15 @@ class FileCacheManager
         if (false === @file_put_contents($this->dir.self::CACHE_FILE, $data, LOCK_EX)) {
             throw new IOException(sprintf('Failed to write file "%s".', self::CACHE_FILE), 0, null, $this->dir.self::CACHE_FILE);
         }
+    }
+
+    private function getRelativeFilename($file)
+    {
+        $dir = realpath($this->dir);
+        if (0 === strpos($file, $dir)) {
+            return '.'.substr($file, strlen($dir));
+        }
+
+        return $file;
     }
 }
--- a/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
+++ b/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
@@ -1,0 +1,124 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests;
+
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\SplFileInfo as FinderSplFileInfo;
+use Symfony\CS\FileCacheManager;
+use Symfony\CS\Fixer;
+
+class FileCacheManagerIntegrationTest extends \PHPUnit_Framework_TestCase
+{
+    protected function setUp()
+    {
+        $this->fs = new Filesystem();
+        $this->rootDir = __DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'FileCacheManagerIntegrationTest';
+
+        $this->fs->mkdir($this->rootDir);
+        $this->isCacheAvailable = $this->isCacheAvailable();
+
+        // Never ever rely on last test results
+        $this->clearFileCacheManagerIntegrationTestFiles();
+
+        $this->customConfigDir = $this->rootDir.DIRECTORY_SEPARATOR.'custom_config_dir';
+
+        $this->files = array(
+            'one' => $this->rootDir.DIRECTORY_SEPARATOR.'path_one'.DIRECTORY_SEPARATOR.'CommonFilename.php',
+            'two' => $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php',
+        );
+
+        $this->fs->mkdir($this->rootDir);
+        $this->fs->mkdir($this->customConfigDir);
+
+        $fs = $this->fs;
+        array_walk($this->files, function ($filename) use ($fs) {
+            $fs->mkdir(dirname($filename));
+            file_put_contents($filename, "<?\n// {$filename}\n");
+        });
+    }
+
+    protected function tearDown()
+    {
+        $this->clearFileCacheManagerIntegrationTestFiles();
+    }
+
+    private function clearFileCacheManagerIntegrationTestFiles()
+    {
+        $this->fs->remove($this->rootDir);
+    }
+
+    private function isCacheAvailable()
+    {
+        $fileCacheManager = new FileCacheManager(true, $this->rootDir, array());
+        $reflectionIsCacheAvailable = new \ReflectionMethod($fileCacheManager, 'isCacheAvailable');
+        $reflectionIsCacheAvailable->setAccessible(true);
+        $isCacheAvailable = $reflectionIsCacheAvailable->invoke($fileCacheManager);
+        $reflectionIsCacheAvailable->setAccessible(false);
+
+        // We don't want FileCacheManager::__destruct() interferences
+        unset($reflectionIsCacheAvailable, $fileCacheManager);
+
+        return $isCacheAvailable;
+    }
+
+    /**
+     * @dataProvider provideScenarios
+     */
+    public function testCacheScenarios($insideCacheFileDir)
+    {
+        if (!$this->isCacheAvailable) {
+            $this->markTestSkipped('Cache functionality not supported');
+        }
+
+        $configDir = $this->rootDir;
+        if (!$insideCacheFileDir) {
+            $configDir = $this->customConfigDir;
+        }
+
+        $fixer = new Fixer();
+        $fixer->addFixer(new Fixer\PSR1\ShortTagFixer());
+        $fixers = $fixer->getFixers();
+
+        $files = array_map(function ($filename) {
+            return new FinderSplFileInfo(
+                $filename,
+                dirname($filename),
+                basename($filename)
+            );
+        }, $this->files);
+
+        $fileCacheManager = new FileCacheManager(true, $configDir, $fixers);
+        $this->assertNotNull($fixer->fixFile($files['one'], $fixers, false, false, $fileCacheManager));
+        $this->assertNotNull($fixer->fixFile($files['two'], $fixers, false, false, $fileCacheManager));
+
+        // This unset() calls FileCacheManager::saveToFile() and is needed for the test
+        unset($fileCacheManager);
+
+        $fileCacheManager = new FileCacheManager(true, $configDir, $fixers);
+        $this->assertFalse($fileCacheManager->needFixing($files['one']->getRealpath(), file_get_contents($files['one']->getRealpath())));
+        $this->assertFalse($fileCacheManager->needFixing($files['two']->getRealpath(), file_get_contents($files['two']->getRealpath())));
+
+        // This unset() calls FileCacheManager::saveToFile() and let the tearDown clean
+        // everything after the test has ended; otherwise FileCacheManager::__destruct()
+        // would leave some files after the run
+        unset($fileCacheManager);
+    }
+
+    public function provideScenarios()
+    {
+        return array(
+            'cache-stored-with-pathname-relative-to-cachefile-if-in-subfolder' => array(true),
+            'cache-stored-with-common-pathname-relative-to-cachefile-if-outside-cachefile-folder' => array(false),
+        );
+    }
+}
--- a/Symfony/CS/Fixer.php
+++ b/Symfony/CS/Fixer.php
@@ -174,7 +174,7 @@ class Fixer
 
         if (
             '' === $old
-            || !$fileCacheManager->needFixing($this->getFileRelativePathname($file), $old)
+            || !$fileCacheManager->needFixing($file->getRealpath(), $old)
             // PHP 5.3 has a broken implementation of token_get_all when the file uses __halt_compiler() starting in 5.3.6
             || (PHP_VERSION_ID >= 50306 && PHP_VERSION_ID < 50400 && false !== stripos($old, '__halt_compiler()'))
         ) {
@@ -290,7 +290,7 @@ class Fixer
             }
         }
 
-        $fileCacheManager->setFile($this->getFileRelativePathname($file), $new);
+        $fileCacheManager->setFile($file->getRealpath(), $new);
 
         if ($this->eventDispatcher) {
             $this->eventDispatcher->dispatch(


keradus:
@localheinz , you have just refactored whole cache
Please, could you take a look ?

localheinz:
@keradus 

Will do!

keradus:
thanks ! I do really appreciate that !

localheinz:
@keradus 

Questions I have

* Will storing the absolute path be a problem (for example, for people who check in the cache file `.php_cs.cache`)?
* Would it make sense to store a path relative to the configuration file `.php_cs.`?

localheinz:
@Slamdunk 

Problem reproduced, looking into it!

Slamdunk:
> Would it make sense to store a path relative to the configuration file .php_cs.?

This bug apart [and this project apart], cache is a dangerous idea (even though essential), so to work without unexpected behaviours it always has to be tightened to the most specificity as possible. I would completely remove anything "relative" from it.

keradus:
@localheinz 
I would really like to have paths relative. When I move whole project from one place to another I don't want to have a need to recreate the cache from scratch :(

@Slamdunk 
what is the dangerous part?

GrahamCampbell:
This could mess up if people run php-cs-fixer passing in a path to a specific part of a project, right?

localheinz:
@GrahamCampbell @keradus @Slamdunk 

In both, the cause of the problem is in [`Symfony\CS\Fixer::getFileRelativePathname()](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer.php#L278-L285)`:

```php
private function getFileRelativePathname(\SplFileInfo $file)
{
    if ($file instanceof FinderSplFileInfo) {
        return $file->getRelativePathname();
    }

    return $file->getPathname();
}
```

The problem could be fixed by returning the path relative to the configuration file here.

Slamdunk:
@keradus false positive cache hit

Here there is no danger, but in the cache theory, a false positive cache hit may be very dangerous depending on the context.

localheinz:
@GrahamCampbell @keradus @Slamdunk 

Noticed another thing (unrelated to this issue, though), hashes for files which previously existed, but don't exist anymore, are kept in the cache forever, shall we fix this?

keradus:
> The problem could be fixed by returning the path relative to the configuration file here.

And for now - we took the relative path to what location? (see 1.12 and 2.x please)


> Noticed another thing (unrelated to this issue, though), hashes for files which previously existed, but don't exist anymore, are kept in the cache forever, shall we fix this?

I'm aware of it. For me it's not big thing, just inconvenience. I agree it would be nice to remove them from cache file, but only with having fast and small memory consumption, while keeping cache shared between consecutive fix run with Intersection finder.

keradus:
I still don't like passing real path instead of relative one. It's way less flexible

Slamdunk:
I'll try to use .php_cs file path (on next days as I'm out today). Will it be acceptable?

GrahamCampbell:
I prefer the real one. :P

keradus:
I encountered cases when cache file is shared between different env. Having path absolute would practically disallow it while bringing no value instead.

Slamdunk:
@keradus may you explain the details of that cases?
Written like that, I see two flaws: actual cache not taking account of environment variables, and developers that share the cache.

keradus:
> actual cache not taking account of environment variables

why cache should depend on them? PHP CS Fixer doesn't use env vars


> and developers that share the cache.

useful especially on beginning of working with big monolitic project

Slamdunk:
@keradus I stripped out the project dir from the path in the cache manager to make it relative and fix the issue for now. Some notes:

1. The stripping is done in the `FileCacheManager` and not in the `Fixer` because I think it's a due of the Cache
1. I would prefer much more in the master branch to typehint `\SplFileInfo` in the `FileCacheManager` methods so that becomes its whole responsability to take care of it, and `Fixer` could become agnostic

Symfony/CS/FileCacheManager.php
```diff
@@ -86,6 +90,11 @@ private function calcHash($content)
         return crc32($content);
     }
 
+    private function getRelativePath($file)
+    {
+        return str_replace($this->dir, './', $file);
```
localheinz: @Slamdunk 

This is already taken care of, see `Directory::getRelativePathTo()`. 

As it stands without the most recent commit, even when passing in the absolute path to the `FileCacheManager`, paths relative to the cache file are stored within the cache file.

Slamdunk:
@localheinz I did'nt get what you say, but I just discovered `$dir` in FileCacheManager may be null in a lot of contexts, so I just removed that commit.

Slamdunk:
@keradus @localheinz I changed the behaviour and now:

1. If the file under fixing is inside the configured path, use the pathname relative to the configured path
1. Otherwise use the full pathname

Example of the keys used by FileCacheManager:
```
/root/project/subfolder/.php_cs
/root/project/subfolder/classes/MyClass.php     => ./classes/MyClass.php
/root/project/MyOtherClass.php                  => /root/project/MyOtherClass.php
```

The two may also occur in conjunction: the relative path is preceeded by a dot, so no false positive can occur if the relative path of a file is the same of the full path of another file.
The bug is fixed, and you can move around your project without losing cache.

Some notes about the tests:

1. The test covers the bug: the two different behaviours explained above cannot be tested as the signature of the FileCacheManager doesn't allow to do much without changing it
1. A full folder and files tree is created and removed at every run because the files need to change to write a meaningful test. I tried to rewrite it with some virtual filesystem like [vfs](http://vfs.bovigo.org/) but got the file_get_contents related error
1. The static behaviour of FileCacheManager::isCacheAvailable makes the class untestable if someone tries to test it first with $isEnables = false; I had to change that: I see no drawback since I still cache the value, and in a run there is only one instance of FileCacheManager
1. To guess if the cache is available without a public method and without copy-paste, I had to access it via Reflection

Slamdunk:
May I ask a feedback on my last comment?

SpacePossum:
@localheinz could you take a look?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+{
+    protected function setUp()
+    {
+        $this->fs = new Filesystem();
+        $this->rootDir = __DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'FileCacheManagerIntegrationTest';
+
+        $this->fs->mkdir($this->rootDir);
+        $this->isCacheAvailable = $this->isCacheAvailable();
+
+        $this->clearFileCacheManagerIntegrationTestFiles();
```
localheinz: Do we really need to clear the files here, or would it be sufficient to clear the files in `tearDown()`?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+     * @dataProvider provideScenarios
+     */
+    public function testCacheScenarios($insideCacheFileDir)
+    {
+        if (!$this->isCacheAvailable) {
+            $this->markTestSkipped('Cache functionality not supported');
+        }
+
+        $configDir = $this->rootDir;
+        if (!$insideCacheFileDir) {
```
localheinz: Instead of using this condition here, can we adjust the data provider instead to return the corresponding directory?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->rootDir = __DIR__.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'FileCacheManagerIntegrationTest';
+
+        $this->fs->mkdir($this->rootDir);
+        $this->isCacheAvailable = $this->isCacheAvailable();
+
+        $this->clearFileCacheManagerIntegrationTestFiles();
+
+        $this->customConfigDir = $this->rootDir.DIRECTORY_SEPARATOR.'custom_config_dir';
+        $this->fileOne = $this->rootDir.DIRECTORY_SEPARATOR.'path_one'.DIRECTORY_SEPARATOR.'CommonFilename.php';
+        $this->fileTwo = $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php';
```
localheinz: Instead of storing the file names in two fields, would it make sense to use one field instead, for example

```php
$this->files = array(
    $this->rootDir.DIRECTORY_SEPARATOR.'path_one'.DIRECTORY_SEPARATOR.'CommonFilename.php',
    $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php',
);
```

This way, we could simplify both test setup as well as the test itself.

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+        if (!$insideCacheFileDir) {
+            $configDir = $this->customConfigDir;
+        }
+
+        $fixer = new Fixer();
+        $fixer->addFixer(new \Symfony\CS\Fixer\PSR1\ShortTagFixer());
+        $fixers = $fixer->getFixers();
+
+        $splFileOne = new FinderSplFileInfo($this->fileOne, dirname($this->fileOne), basename($this->fileOne));
+        $splFileTwo = new FinderSplFileInfo($this->fileTwo, dirname($this->fileTwo), basename($this->fileTwo));
```
localheinz: Following the comment above, how about just mapping the field `$fieldNames` like this:

```php
/* @var FinderSplFileInfo[] $files */
$files = array_map(function ($fileName) {
    return new FinderSplFileInfo(
        $fileName,
        dirname($fileName),
        basename($fileName)
    );
}, $this->fileNames);
```

Then we could just iterate over the `$files` array below, both for warming up the cache, and asserting that the files do not need fixing. What do you think?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->clearFileCacheManagerIntegrationTestFiles();
+
+        $this->customConfigDir = $this->rootDir.DIRECTORY_SEPARATOR.'custom_config_dir';
+        $this->fileOne = $this->rootDir.DIRECTORY_SEPARATOR.'path_one'.DIRECTORY_SEPARATOR.'CommonFilename.php';
+        $this->fileTwo = $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php';
+
+        $this->fs->mkdir($this->rootDir);
+        $this->fs->mkdir($this->customConfigDir);
+        $this->fs->mkdir(dirname($this->fileOne));
+        $this->fs->mkdir(dirname($this->fileTwo));
```
Slamdunk: I don't understand what you mean: can you explain it to me better please?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
Slamdunk: I thought alot about it when I wrote the test, and chose this form because in the data provider I'm declaring what and why I am testing, and in the test I adjast the variables in one direction or another.
It's not strictly a data provider, but a way to create two tests without duplicating the code, and I didn't want the data provide know the internal parameters of the test.

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+            $this->markTestSkipped('Cache functionality not supported');
+        }
+
+        $configDir = $this->rootDir;
+        if (!$insideCacheFileDir) {
+            $configDir = $this->customConfigDir;
+        }
+
+        $fixer = new Fixer();
+        $fixer->addFixer(new \Symfony\CS\Fixer\PSR1\ShortTagFixer());
```
Slamdunk: Done

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->assertNotNull($fixer->fixFile($splFileOne, $fixers, false, false, $fileCacheManager));
+        $this->assertNotNull($fixer->fixFile($splFileTwo, $fixers, false, false, $fileCacheManager));
+
+        unset($fileCacheManager);
+
+        $fileCacheManager = new FileCacheManager(true, $configDir, $fixers);
+        $this->assertFalse($fileCacheManager->needFixing($splFileOne->getRealpath(), file_get_contents($splFileOne->getRealpath())));
+        $this->assertFalse($fileCacheManager->needFixing($splFileTwo->getRealpath(), file_get_contents($splFileTwo->getRealpath())));
+
+        unset($fileCacheManager);
```
Slamdunk: Yes, I wrote a comment about it

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
localheinz: Ha, of course!

We dynamically declare a few fields here, but we never declare them. Personally, I would prefer to declare these fields:

```diff
diff --git a/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php b/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
index 1040ca4..99aa69a 100644
--- a/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
+++ b/Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
@@ -19,6 +19,31 @@

 class FileCacheManagerIntegrationTest extends \PHPUnit_Framework_TestCase
 {
+    /**
+     * @var Filesystem
+     */
+    private $fs;
+
+    /**
+     * @var string
+     */
+    private $rootDir;
+
+    /**
+     * @var bool
+     */
+    private $isCacheAvailable;
+
+    /**
+     * @var string
+     */
+    private $customConfigDir;
+
+    /**
+     * @var array
+     */
+    private $files;
+
     protected function setUp()
     {
         $this->fs = new Filesystem();
```

This has the advantage that you can

* search for usages of these fields (using PHPStorm)
* rename these fields (using PHPStorm refactor)
* avoid false negatives (using PHPStorm inspections)

Does that make sense?

As I said, not sure what the rules are in this repository. I have been working in other repositories where the maintainers had strong opinions on not declaring fields in tests.

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,129 @@
...
+            'two' => array(
+                'filename' => $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php',
+                'content' => "<?\n// two\n",
+            ),
+        );
+
+        $this->fs->mkdir($this->rootDir);
+        $this->fs->mkdir($this->customConfigDir);
+
+        array_map(function (array $file) {
```
localheinz: I believe it's a matter of taste, but personally, I prefer to use `array_map()` only when I am interested in the return value of it, otherwise I would use `array_walk()`, for example.

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+        $this->fs->mkdir($this->rootDir);
+        $this->isCacheAvailable = $this->isCacheAvailable();
+
+        // Never ever rely on last test results
+        $this->clearFileCacheManagerIntegrationTestFiles();
+
+        $this->customConfigDir = $this->rootDir.DIRECTORY_SEPARATOR.'custom_config_dir';
+
+        $this->files = array(
```
localheinz: Was thinking, could we simplify this to

```php
$this->files = array(
    $this->rootDir.DIRECTORY_SEPARATOR.'path_one'.DIRECTORY_SEPARATOR.'CommonFilename.php',
    $this->rootDir.DIRECTORY_SEPARATOR.'path_two'.DIRECTORY_SEPARATOR.'CommonFilename.php',
);
```

and then later

```php
array_walk($this->files, function ($file) {
    $this->fs->mkdir(dirname($file));
    file_put_contents($file, "<?\n// $file\n",);
});
```

?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,129 @@
...
+            return new FinderSplFileInfo(
+                $file['filename'],
+                dirname($file['filename']),
+                basename($file['filename'])
+            );
+        }, $this->files);
+
+        $fileCacheManager = new FileCacheManager(true, $configDir, $fixers);
+        $this->assertNotNull($fixer->fixFile($files['one'], $fixers, false, false, $fileCacheManager));
+        $this->assertNotNull($fixer->fixFile($files['two'], $fixers, false, false, $fileCacheManager));
```
localheinz: How about changing this to

```php
foreach ($files as $file) {
    $this->assertNotNull($fixer->fixFile($file, $fixers, false, false, $fileCacheManager));
}
```

?

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
```diff
@@ -0,0 +1,129 @@
...
+
+        // This unset() calls FileCacheManager::saveToFile() and is needed for the test
+        unset($fileCacheManager);
+
+        $fileCacheManager = new FileCacheManager(true, $configDir, $fixers);
+        $this->assertFalse($fileCacheManager->needFixing($files['one']->getRealpath(), file_get_contents($files['one']->getRealpath())));
+        $this->assertFalse($fileCacheManager->needFixing($files['two']->getRealpath(), file_get_contents($files['two']->getRealpath())));
```
localheinz: How about changing this to

```php
foreach ($files as $file) {
    $this->assertFalse($fileCacheManager->needFixing($file->getRealpath(), file_get_contents($file->getRealpath())));
}
```

?

The advantage of this and above is that if you feel like doing this with any other number than two files, you only need to modify the array of `$this->files` in a single place.

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
Slamdunk: Done

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
Slamdunk: Yup, better

Symfony/CS/Tests/FileCacheManagerIntegrationTest.php
Slamdunk: I consider this very dangerous: if for any reason in the future the `$files` variable will result empty, the test will pass silently because no assertion will be done.

It happened some times in our business and in a bunch of OSS projects: I strongly discourage dynamic assertions.

SpacePossum: > if for any reason in the future the $files variable will result empty, the test will pass silently because no assertion will be made.

legit concern, how about adding `$this->assertCount(2, $files)` before the loop?

keradus:
Again, the paths should be relative to config file. Making paths absolute is making cache too much restricted.
On the same side, I'm fine with having the solution only for 2.x line, where issue happen as well, but which has a lot of caching code rewritten.

keradus:
Closed due to merge of #2357

PR #1999, opened 2016-06-17
Runner - No need to determine relative file name twice

localheinz:
This PR

* [x] determines the relative pathname for a file once only

Somewhat related to #1998.


keradus:
It could be from some 1.12 merge conflict.
Or it appears duplicated on 1.12 as well ?

localheinz:
@keradus 

Nope, only in `master`.

keradus:
thanks for checking

keradus:
Thank you @localheinz.

PR #2004, opened 2016-06-17
Fix: Pass in real path to FileCacheManager

localheinz:
This PR
- [x] passes in the real path of a file to the `FileCacheManager`

Follows #1998, but fixes it for `master`.


Slamdunk:
Add tests like #1998 please, otherwise in a couple of month someone will change it without getting noticed about this bug.

Btw, if #1998 is going to be merged, I would have PRed to master too

SpacePossum:
@localheinz why close these cache related PR's? These looked useful

keradus:
Closed due to merge of #2357

PR #2006, opened 2016-06-19
Update composer install in readme

sitawit:
When install with fabpot/php-cs-fixer, the composer suggests to use friendsofphp/php-cs-fixer instead.


GrahamCampbell:
Hi, thanks, but the name here is actually correct. This tool is getting renamed from 2.x only, so the name as been updated on the master branch only so far. :)

PR #1955, opened 2016-06-04
Introduce CT_ARRAY_INDEX_CURLY_BRACE_OPEN and CT_ARRAY_INDEX_CURLY_BRACE_CLOSE

keradus:
Also, fixes #1909

PR #2019, opened 2016-06-28
Fixing newlines added after curly brace string index access

### LINKED ISSUE #2018: php-cs-fixer adding a newline after brace of string character access
Given this PHP code:

```php
<?php

if (true) {
    $property = substr($name, 3);
    $property{0} = strtolower($property{0});
    $property = (string) $property;
}
```

And run with this config:

```php
    ->fixers([
        '-concat_without_spaces',
        '-single_quote',
        '-unalign_double_arrow',
        '-unalign_equals',
        'short_array_syntax',
        'align_equals',
        'align_double_arrow'
    ]);
```

Using php-cs-fixer 1.11.3 results in this output:

```php
<?php

if (true) {
    $property    = substr($name, 3);
    $property{0}
    = strtolower($property{0});
    $property    = (string) $property;
}
```

Here's the output with --verbose flag showing which fixers were applied.

> 1) test.php (align_equals, braces)

I believe the output should be this:

```php
<?php

if (true) {
    $property    = substr($name, 3);
    $property{0} = strtolower($property{0});
    $property    = (string) $property;
}
```

jaydiablo:
Fixes #2018

This fixes my most basic use case where I ran into an issue of extra
new lines, so it only deals with assignment of a string character to
another value (see the test for an example of what I mean).  There may
be other cases where newlines are added for this string index access
syntax, though there are language limitations as to what you can do to
a string offset.

### DIFF
--- a/Symfony/CS/Fixer/PSR2/BracesFixer.php
+++ b/Symfony/CS/Fixer/PSR2/BracesFixer.php
@@ -215,7 +215,8 @@ class BracesFixer extends AbstractFixer
                         // and it is not a `${"a"}->...` and `${"b{$foo}"}->...` situation
                         !($nestToken->equals('}') && $tokens[$nestIndex - 1]->equalsAny(array('"', "'", array(T_CONSTANT_ENCAPSED_STRING)))) &&
                         // and it is not a `$var{0} = ` situation (character access on string)
-                        !($nestToken->equals('}') && $nextNonWhitespaceNestToken->equals('='))
+                        // TODO: remove on 2.x line
+                        !($nestToken->equals('}') && $nextNonWhitespaceNestToken->equalsAny(array('=', array(T_OBJECT_OPERATOR))))
                     ) {
                         if ($nextNonWhitespaceNestToken->isGivenKind($this->getControlContinuationTokens()) || $nextNonWhitespaceNestToken->isGivenKind(T_CLOSE_TAG)) {
                             $whitespace = ' ';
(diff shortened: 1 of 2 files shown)


keradus:
Already fixed on 2.x line with #1955 
Please, add the following utest:
```
           array(
                '<?php
                    if (1) {
                        echo $items{0}->foo;
                        echo $collection->items{1}->property;
                    }
                ',
            ),
```

Symfony/CS/Tests/Fixer/PSR2/BracesFixerTest.php
```diff
@@ -1367,4 +1367,24 @@ public function providePreserveLineAfterControlBrace()
             ),
         );
     }
+
+    /**
+     * @dataProvider provideDontAddNewLineAfterCurlyBraceOfStringCharacterAccess
+     */
+    public function testDontAddNewLineAfterCurlyBraceOfStringCharacterAccess($expected, $input = null)
```
keradus: TODO: remove on 2.x line

Symfony/CS/Fixer/PSR2/BracesFixer.php
```diff
@@ -213,7 +213,9 @@ function ($item) {
                         // and it is not a `Foo::{bar}()` situation
                         !($nestToken->equals('}') && $nextNonWhitespaceNestToken->equals('(')) &&
                         // and it is not a `${"a"}->...` and `${"b{$foo}"}->...` situation
-                        !($nestToken->equals('}') && $tokens[$nestIndex - 1]->equalsAny(array('"', "'", array(T_CONSTANT_ENCAPSED_STRING))))
+                        !($nestToken->equals('}') && $tokens[$nestIndex - 1]->equalsAny(array('"', "'", array(T_CONSTANT_ENCAPSED_STRING)))) &&
+                        // and it is not a `$var{0} = ` situation (character access on string)
```
keradus: TODO: remove on 2.x line

keradus:
Thank you @jaydiablo.

PR #2020, opened 2016-06-28
Added PhpdocAnnotationWithoutDotFixer

keradus:
ref https://github.com/symfony/symfony/pull/19198

### DIFF
--- a/Symfony/CS/Fixer/Contrib/Php4ConstructorFixer.php
+++ b/Symfony/CS/Fixer/Contrib/Php4ConstructorFixer.php
@@ -315,7 +315,7 @@ class Php4ConstructorFixer extends AbstractFixer
      *     - endIndex (int): The index of the function/method end.
      *     - bodyIndex (int): The index of the function/method body.
      *     - modifiers (array): The modifiers as array keys and their index as
-     *       the values, e.g. array(T_PUBLIC => 10).
+     *       the values, e.g. array(T_PUBLIC => 10)
      */
     private function findFunction(Tokens $tokens, $name, $startIndex, $endIndex)
     {
--- a/README.rst
+++ b/README.rst
@@ -428,6 +428,11 @@ Choose from the list of available fixers:
                         surrounded by at least one
                         space.
 
+* **phpdoc_annotation_without_dot** [symfony]
+                        Phpdocs annotation
+                        descriptions should not end
+                        with a full stop.
+
 * **phpdoc_indent** [symfony]
                         Docblocks should have the same
                         indentation as the documented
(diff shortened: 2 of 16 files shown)


fabpot:
LGTM

SpacePossum:
can we ditch other stop characters as well?  (`'.。!?¡¿！？'`)

keradus:
https://github.com/symfony/symfony/pull/19198#issuecomment-228971663

GrahamCampbell:
You should also lower case the first letter too then, because it's not a sentence.

SpacePossum:
>You should also lower case the first letter too then, because it's not a sentence.

not all people write English only, so the rule doesn't always apply.

GrahamCampbell:
What about those people who, like symfony?

SpacePossum:
>What about those people who, like symfony?

the fixer might be used by other people than symfony people

rob006:
First word might be the name of class, constant, acronym, etc., you can not safely change case of the first letter.

GrahamCampbell:
Don't make the first word that then. I never do that because of that that exact issue. Allow me to have correctly formatted sentences.

SpacePossum:
you can make an issue out of every new fixer and/or configuration option you don't want, but since no one forces you to use it, what is the point?

keradus:
@GrahamCampbell if you want to put sentence to param description it is great ! Don't use the fixer that force you to not have it ;)

Symfony/CS/Fixer/Contrib/Php4ConstructorFixer.php
```diff
@@ -315,7 +315,7 @@ private function getWrapperMethodSequence(Tokens $tokens, $method, $startIndex,
      *     - endIndex (int): The index of the function/method end.
      *     - bodyIndex (int): The index of the function/method body.
      *     - modifiers (array): The modifiers as array keys and their index as
-     *       the values, e.g. array(T_PUBLIC => 10).
+     *       the values, e.g. array(T_PUBLIC => 10)
```
GrahamCampbell: This removal is invalid

GrahamCampbell: we should only remove on singleline descriptions, since this one has multiple sentences

keradus: the thing is that due to Symfony standards you shouldn't have sentence here in the first places.
(btw, single line could have multiple sentences as well, not connected)

GrahamCampbell: Yeh. I would suggest removing the last line only if there was not a dot earlier on, but that wouldn't be reliable.

PR #2025, opened 2016-06-29
NoShortEchoTagFixer - adjust isCandidate check for hhvm

keradus:
hhvm on 2.x line is broken :( let us fix it !

### DIFF
--- a/src/Fixer/PhpTag/NoShortEchoTagFixer.php
+++ b/src/Fixer/PhpTag/NoShortEchoTagFixer.php
@@ -26,7 +26,13 @@ final class NoShortEchoTagFixer extends AbstractFixer
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_OPEN_TAG_WITH_ECHO);
+        return (
+            $tokens->isTokenKindFound(T_OPEN_TAG_WITH_ECHO)
+            || (
+                defined('HHVM_VERSION')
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
:+1: :ship:

src/Fixer/PhpTag/NoShortEchoTagFixer.php
```diff
@@ -26,7 +26,17 @@
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_OPEN_TAG_WITH_ECHO);
+        return $tokens->isTokenKindFound(T_OPEN_TAG_WITH_ECHO)
+            /*
+             * HHVM parses '<?=' as T_ECHO instead of T_OPEN_TAG_WITH_ECHO
+             *
+             * @see https://github.com/facebook/hhvm/issues/4809
```
GrahamCampbell: You can't use phpdoc annotations when this isn't `T_DOC_COMMENT`.

SpacePossum: >You can't use phpdoc annotations when this isn't T_DOC_COMMENT.

there are no rules for comment content we follow, there is no need also, this PR proves that

keradus: I could use anything inside regular comment...

PR #2026, opened 2016-06-29
Changing composer install location

### LINKED ISSUE #2014: package name outdated in composer global install instructions?
I installed via composer on windows with the following command:

composer global require fabpot/php-cs-fixer

going off of the install instructions in the readme, but during the install I was presented with this message:

"Package fabpot/php-cs-fixer is abandoned, you should avoid using it. Use friendsofphp/php-cs-fixer instead." 

Should the package name be updated in the readme? I can submit a pull request to update the readme if so.

jaydiablo:
Fixes #2014

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ To install PHP-CS-Fixer, install Composer and issue the following command:
 
 .. code-block:: bash
 
-    $ ./composer.phar global require fabpot/php-cs-fixer
+    $ ./composer.phar global require friendsofphp/php-cs-fixer
 
 Then, make sure you have ``~/.composer/vendor/bin`` in your ``PATH``, and
 you're good to go:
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks, but this doesn't need changing on 1.x.

jaydiablo:
Why not?  Composer does complain as the referenced ticket mentions, and 1.11 shows as the default branch when you hit https://github.com/FriendsOfPHP/PHP-CS-Fixer, so people will see the fabpot package location rather than friendsofphp.

GrahamCampbell:
The old package is not deprecated yet. The packagist warning is too early.

jaydiablo:
I see, the plan is to deprecate when 2.0 is released?

Should the referenced ticket be closed then with wontfix?

GrahamCampbell:
There are over 10 tickets now.

PR #2039, opened 2016-07-04
NoExtraConsecutiveBlankLinesFixer - Fix curly brace open false positive

SpacePossum:
This fixes the wrong fixing of cases like:
```php
<?php

function test(){}
function test1(){}
echo $a;
```

It add the options:
- `square_brace_open`
- `parenthesis_brace_open`

All other changes are minor, such as adding missing docs and utest, optimize some stuff and up the DX by adding better exception messages.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -17,6 +17,7 @@ return PhpCsFixer\Config::create()
         'combine_consecutive_unsets' => true,
         'header_comment' => array('header' => $header),
         'long_array_syntax' => true,
+        'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block'),
         'no_useless_else' => true,
         'no_useless_return' => true,
         'ordered_class_elements' => true,
@@ -27,7 +28,6 @@ return PhpCsFixer\Config::create()
         'psr4' => true,
         'strict_comparison' => true,
         'strict_param' => true,
-        'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'curly_brace_open'),
     ))
     ->finder(
         PhpCsFixer\Finder::create()
--- a/src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
+++ b/src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
@@ -55,7 +55,9 @@ final class NoExtraConsecutiveBlankLinesFixer extends AbstractFixer
      * - 'throw' remove blank lines after a line with a 'throw' statement
      * - 'use' remove blank lines between 'use' import statements
      * - 'useTrait' remove blank lines between 'use' trait statements
-     * - 'curly_brace_open' remove blank lines after a curly opening brace ('{')
+     * - 'curly_brace_block' remove blank lines after a curly opening block brace ('{') and/or end block brace ('}')
+     * - 'parenthesis_brace_block' remove blank lines after a parenthesis opening block brace ('(') and/or end block brace (')')
+     * - 'square_brace_block' remove blank lines after a square opening block brace ('[') and/or end block brace (']')
      *
      * @param string[]|null $configuration
      */
@@ -90,8 +92,14 @@ final class NoExtraConsecutiveBlankLinesFixer extends AbstractFixer
                 case 'useTrait':
                     $this->tokenKindCallbackMap[CT_USE_TRAIT] = 'removeBetweenUse';
                     break;
-                case 'curly_brace_open':
-                    $this->tokenEqualsMap['{'] = 'fixAfterToken';
+                case 'curly_brace_block':
+                    $this->tokenEqualsMap['{'] = 'fixStructureOpenCloseIfMultiLine'; // i.e. not: CT_ARRAY_INDEX_CURLY_BRACE_OPEN
+                    break;
+                case 'parenthesis_brace_block':
+                    $this->tokenEqualsMap['('] = 'fixStructureOpenCloseIfMultiLine'; // i.e. not: CT_BRACE_CLASS_INSTANTIATION_OPEN
+                    break;
+                case 'square_brace_block':
+                    $this->tokenKindCallbackMap[CT_ARRAY_SQUARE_BRACE_OPEN] = 'fixStructureOpenCloseIfMultiLine'; // typeless '[' tokens should not be fixed (too rare)
                     break;
                 default:
                     throw new InvalidFixerConfigurationException($this->getName(), sprintf('Unknown configuration item "%s" passed.', $item));
@@ -104,7 +112,7 @@ final class NoExtraConsecutiveBlankLinesFixer extends AbstractFixer
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_WHITESPACE);
+        return true;
     }
 
     /**
@@ -154,6 +162,8 @@ final class NoExtraConsecutiveBlankLinesFixer extends AbstractFixer
             }
 
             $this->$callback($index);
+
+            return;
         }
     }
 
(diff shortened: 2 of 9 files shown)


keradus:
@SpacePossum , is it still WIP ?

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -91,8 +91,15 @@ public function configure(array $configuration = null)
                     $this->tokenKindCallbackMap[CT_USE_TRAIT] = 'removeBetweenUse';
                     break;
                 case 'curly_brace_open':
-                    $this->tokenEqualsMap['{'] = 'fixAfterToken';
+                    $this->tokenEqualsMap['{'] = 'fixAfterTokenIfMultiLineBody'; //'fixAfterToken'; // i.e. not: CT_ARRAY_INDEX_CURLY_BRACE_OPEN
```
keradus: comment could be cleaned up

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -91,8 +91,15 @@ public function configure(array $configuration = null)
...
                     break;
+                case 'square_brace_open':
+                    $this->tokenKindCallbackMap[CT_ARRAY_SQUARE_BRACE_OPEN] = 'fixAfterTokenIfMultiLineBody'; // typeless '[' tokens should not be fixed (to rare)
+                    break;
+                case 'parenthesis_brace_open':
+                    $this->tokenEqualsMap['('] = 'fixAfterTokenIfMultiLineBody'; // i.e. not: CT_BRACE_CLASS_INSTANTIATION_OPEN
+                    break;
+                // case 'after_coma', 'before_comma'
```
keradus: ?

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
SpacePossum: will do :+1:

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
SpacePossum: left over from `local hacked`  fixer version, will delete :)

SpacePossum:
update the comments and PR description

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -91,7 +91,13 @@ public function configure(array $configuration = null)
                     $this->tokenKindCallbackMap[CT_USE_TRAIT] = 'removeBetweenUse';
                     break;
                 case 'curly_brace_open':
-                    $this->tokenEqualsMap['{'] = 'fixAfterToken';
+                    $this->tokenEqualsMap['{'] = 'fixAfterTokenIfMultiLineBody'; // i.e. not: CT_ARRAY_INDEX_CURLY_BRACE_OPEN
+                    break;
+                case 'square_brace_open':
+                    $this->tokenKindCallbackMap[CT_ARRAY_SQUARE_BRACE_OPEN] = 'fixAfterTokenIfMultiLineBody'; // typeless '[' tokens should not be fixed (to rare)
```
keradus:
```diff
-to rare
+too rare
```

keradus: I don't get the comment.
`[` is different than `CT_ARRAY_SQUARE_BRACE_OPEN`. one is for opening array, second for opening index.

keradus:
:+1: for code but comments

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
SpacePossum: I think something like:
```php
$a[

1,
2,
3,
];
```

happens (I'm looking @ Laravel configuration stuff).
But something like;
```php
$a = $b[

1];
```

is rare and would get fixed to;

```php
$a = $b[
1];
```

which is not much better ;)

keradus: in first one you meant array, right? (`$a = [ ...`)

I agree to fix only first one, for second one we should have separate rule.

------------

so, we will fix
```php
$a = [

1,
];
```

into

```php
$a = [
1,
];
```

---------------

what about 

```php
$a = [
1,

];
```
and
```php
$a = array(

1,
);
```

?

SpacePossum: Indeed typo in first case, good catch :+1: 
Your cases are correct related to the current state of the PR.
The last to sample could be fixed, but will take more time to develop and amend to this PR.
Let me think about those two to see how hard it would be (i.e. at config rule `after comma`, `before ] if within block`).

keradus: for me it's a must have if we want to fix blank line after opening array by `[` we need to fix it after opening array be `array(`.
before closing would be nice as well, and actually could be single option together with after opening.

SpacePossum: after open `array(` will be fixed using the new option `parenthesis_brace_open` I think, but I'll add a test for to be sure

SpacePossum: confirmed, 
```php
<?php
array(

1
);`
```

```php
<?php
array(
1
);`
```

with this fixer and `parenthesis_brace_open`
I'll commit a test for this soon

SpacePossum:
I've changed the PR, no it will remove extra blank lines after an opening token and before the related closing token. Checked on SF as well and sure if this is what they want, maybe I should make the before and after token both optional.

.php_cs.dist
```diff
@@ -13,24 +13,11 @@ EOF;
...
-        'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'curly_brace_open'),
+        'no_extra_consecutive_blank_lines' => array( 'parenthesis_brace_block', 'square_brace_block',  'curly_brace_block'),
     ))
     ->finder(
         PhpCsFixer\Finder::create()
             ->exclude('tests/Fixtures')
-            ->in(__DIR__)
+            ->in(__DIR__.'/./../symfony/src')
     )
 ;
```
SpacePossum: all to be reverted ofcourse

SpacePossum:
ready for reviewing

SpacePossum:
fabbot is wrong
for the failing tests see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2163

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -55,7 +55,10 @@
      * - 'throw' remove blank lines after a line with a 'throw' statement
      * - 'use' remove blank lines between 'use' import statements
      * - 'useTrait' remove blank lines between 'use' trait statements
-     * - 'curly_brace_open' remove blank lines after a curly opening brace ('{')
+     * Blocks:
```
keradus: please remove that line, it looks like we have one key like `options` and second like `blocks`, while we have only one options ;)

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -154,6 +163,8 @@ private function fixByToken(Token $token, $index)
             }
 
             $this->$callback($index);
+
+            return;
```
keradus: good catch !

tests/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -632,6 +666,79 @@ public function provideOneAndInLineCases()
...
+            ),
+            array(
+                'curly_brace_block',
+                "<?php switch(\$a){\ncase 1:echo 789;}",
+                "<?php switch(\$a){\n   \ncase 1:echo 789;}",
+            ),
+            array(
+                'parenthesis_brace_block',
+                "<?php is_int(\n1);\nfunction test(\n\$a,\n\$b,\n\$c\n)\n{\n\n\n}",
+                "<?php is_int(\n\n1);\nfunction test(\n\n\$a,\n\$b,\n\$c\n\n\n)\n{\n\n\n}",
```
keradus: could we unroll `\n`, so it would be easy to see what's changed ?

keradus:
great work !
some minors and :+1:

please rebase

SpacePossum:
I don't understand why github is warning for the conflicts, the commit applies to current master branch without any issue.

keradus:
maybe your master is outdated ?

keradus:
Actually it's not only github.
both CI are claiming they cannot apply PR patch to master to perform the builds

keradus:
rebased.

SpacePossum:
awesome, thanks for the correct rebasing!

keradus:
Thank you @SpacePossum.

PR #2038, opened 2016-07-03
[WIP] [bugfix] Check for file encoding before try to apply fixers

phansys:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

I think this breaks BC since it requires `ext-mbstring` in order to work. Also, I don't know how is this tool behaving under all cases where file encoding can be detected.
I've found this issue using `dev-master` version, but I think it applies here at `1.11` too.

```
PHP Fatal error:  Method PhpCsFixer\DocBlock\Line::__toString() must not throw an exception, caught ErrorException: Method PhpCsFixer\DocBlock\Line::__toString() must return a string value in vendor/friendsofphp/php-cs-fixer/src/DocBlock/DocBlock.php on line 177
PHP Stack trace:
PHP   1. {main}() vendor/friendsofphp/php-cs-fixer/php-cs-fixer:0
PHP   2. Symfony\Component\Console\Application->run() vendor/friendsofphp/php-cs-fixer/php-cs-fixer:45
PHP   3. Symfony\Component\Console\Application->doRun() vendor/symfony/console/Application.php:123
PHP   4. Symfony\Component\Console\Application->doRunCommand() vendor/symfony/console/Application.php:192
PHP   5. Symfony\Component\Console\Command\Command->run() vendor/symfony/console/Application.php:844
PHP   6. PhpCsFixer\Console\Command\FixCommand->execute() vendor/symfony/console/Command/Command.php:259
PHP   7. PhpCsFixer\Runner\Runner->fix() vendor/friendsofphp/php-cs-fixer/src/Console/Command/FixCommand.php:370
PHP   8. PhpCsFixer\Runner\Runner->fixFile() vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php:141
PHP   9. PhpCsFixer\Fixer\Phpdoc\PhpdocAnnotationWithoutDotFixer->fix() vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php:198
PHP  10. PhpCsFixer\DocBlock\DocBlock->getContent() vendor/friendsofphp/php-cs-fixer/src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php:55
PHP  11. implode() vendor/friendsofphp/php-cs-fixer/src/DocBlock/DocBlock.php:177
```
[...]

PR #2040, opened 2016-07-04
PhpdocAnnotationWithoutDotFixer - fix crash on odd character

keradus:
Replaces and fixes #2038

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
@@ -43,7 +43,12 @@ class PhpdocAnnotationWithoutDotFixer extends AbstractFixer
             foreach ($annotations as $annotation) {
                 if ($annotation->getTag()->valid()) {
                     $line = $doc->getLine($annotation->getEnd());
-                    $line->setContent(preg_replace('/[.。](\s+)$/u', '\1', $line->getContent()));
+
+                    $content = preg_replace('/[.。](\s+)$/u', '\1', $line->getContent());
+
+                    if (null !== $content) {
+                        $line->setContent($content);
+                    }
                 }
             }
             $token->setContent($doc->getContent());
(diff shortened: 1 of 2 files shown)


Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
```diff
@@ -43,7 +43,12 @@ public function fix(\SplFileInfo $file, $content)
             foreach ($annotations as $annotation) {
                 if ($annotation->getTag()->valid()) {
                     $line = $doc->getLine($annotation->getEnd());
-                    $line->setContent(preg_replace('/[.。](\s+)$/u', '\1', $line->getContent()));
+
+                    preg_match('/^(.*[.。])(\s*)$/', $line->getContent(), $matches);
```
SpacePossum: you don't need the `/u` flag anymore? (out of interest)

Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
```diff
@@ -43,7 +43,12 @@ public function fix(\SplFileInfo $file, $content)
...
+
+                    if (0 !== count($matches)) {
+                        $line->setContent(mb_substr($matches[1], 0, -1).$matches[2]);
```
SpacePossum: the `mb` package is not installed on a default PHP install, so we should at it to `composer.json` as req.?

keradus: check the newest commit ;P

Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
keradus: new commit ;)

Symfony/CS/Fixer/Symfony/PhpdocAnnotationWithoutDotFixer.php
keradus: yeah, it's not working as well ;/

keradus:
```php
PHP Notice:  iconv_strlen(): Detected an illegal character in input string in /home/keradus/github/PHP-CS-Fixer/vendor/symfony/polyfill-mbstring/Mbstring.php on line 338
```

keradus:
I think that is the best we can do. If `preg_replace` will crash there is not much we can do.

Yes, we could add some fancy logic and do everything on very low level, but it's pointless

SpacePossum:
thanks! :+1:

PR #2047, opened 2016-07-05
Fix: NativeFunctionCasingFixer should leave class name alone

localheinz:
This PR

* [x] adds a failing test for the `NativeFunctionCasingFixer`

🙎 Not sure, maybe it helps?


SpacePossum:
it is because:
```php
            array(
                '<?php
                    new \STRTOLOWER();
                ',
            ),
```

is a constructor call to the class `STRTOLOWER` which is not a native function :)

localheinz:
@SpacePossum 

Noticed this in a legacy project using Kohana, and there, they have a `Date` class.

Here

```php
new \Date()
```

is fixed to 

```php
new \date();
```

localheinz:
Closing in favor of #2048.

PR #2049, opened 2016-07-05
Use official package name

mnabialek:
Use official package name instead of legacy one

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -54,7 +54,7 @@ To install PHP-CS-Fixer, install Composer and issue the following command:
 
 .. code-block:: bash
 
-    $ ./composer.phar global require fabpot/php-cs-fixer
+    $ ./composer.phar global require friendsofphp/php-cs-fixer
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Thanks, but this isn't actually correct. This name is only for 2.x.

mnabialek:
@GrahamCampbell  So I don't know why but when I've upgraded to 1.11.4 I'm getting:

> Package fabpot/php-cs-fixer is abandoned, you should avoid using it. Use friendsofphp/php-cs-fixer instead.

GrahamCampbell:
Yeh, that message is correct. It's just saying that the name is changing when 2.x comes out, so be prepared.

GrahamCampbell:
It should not have been marked as abandoned until 2.x is out IMO.

Soullivaneuh:
Or maybe add a hint on the documentation just below the require example?

BTW, `composer require friendsofphp/php-cs-fixer:1.0.*` command actually works without any problem.

So maybe we can consider changing it?

Issue #2053, opened 2016-07-07
ClassDefinitionFixer incorrectly fixes valid multi-line "implements" sections with only one interface


ezzatron:
In [PSR-2 section 4.1. Extends and Implements](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#41-extends-and-implements):

> Lists of implements MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one interface per line.

But `ClassDefinitionFixer` currently only allows multi-line `implements` definitions when there is more than 1 interface name. This additional restriction is **not** part of the PSR-2 spec.

The relevant code can be found on [this line](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/5de5794c0409982c5ed36aaa1546d2fa6d5fdf24/src/Fixer/ClassNotation/ClassDefinitionFixer.php#L76):

```php
if ($implementsInfo['numberOfInterfaces'] > 1 && $implementsInfo['multiLine']) {
```

This bug is currently causing issues with the use of multi-line `implements` definitions that are long because of the length of class names, and not the number of implemented interfaces:

```php
class VeryLongClassNameWithLotsOfLetters extends AnotherVeryLongClassName implements
    VeryLongInterfaceNameThatIDontWantOnTheSameLine
{
}
```

The above is erroneously "fixed" to become:

```php
class VeryLongClassNameWithLotsOfLetters extends AnotherVeryLongClassName implements VeryLongInterfaceNameThatIDontWantOnTheSameLine
{
}
```

I tried to create a PR, but ran into some issues with an existing test case that broke. I was not sure of the intent of the test case, so I thought it better just to report the bug instead.

SpacePossum:
Thanks for reporting!
I cannot recall I took this case into consideration when writing the fixer, so yes this could be a valid bug report. Feel free to open your PR including the failing test case, please add [WIP] in the title and tag me, happy to help you out :)

ezzatron:
Okay, no worries.

sharifzadesina:
This is same with more than one interface.

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2155

keradus:
Closing due to merge of #2155

PR #2054, opened 2016-07-07
[WIP] Fixed ClassDefinitionFixer behavior for single interfaces on multiple lines.

ezzatron:
@SpacePossum Here's my attempt at a fix for #2053.

The breaking test is this example:

```php
<?php
class/**/Test123
extends  /**/        \RuntimeException    implements

TestZ{
}
```

I'm actually not sure whether this should be treated as multi-line, or single-line. It could probably go either way I guess.


keradus:
What about 1.12 line? If this bug affects also it your PR should be targeted for 1.12 branch instead of master (2.x line)

ezzatron:
I didn't even know that branch existed, since the default branch is `1.11`, which doesn't have `ClassDefinitionFixer`. I'll have to re-submit the PR when I have more time.

keradus:
ping @ezzatron

ezzatron:
@keradus see #2101

keradus:
closing in favor of #2101

PR #2055, opened 2016-07-07
composer require friendsofphp/php-cs-fixer

gonzakpo:
composer require friendsofphp/php-cs-fixer


gharlan:
number 10 :)
-> #2049

PR #2056, opened 2016-07-07
Travis - run nightly PHP

keradus:
Let us run 7.1@dev on Travis, as we already have few cases for it.
master branch check: https://travis-ci.org/keradus/PHP-CS-Fixer/builds/143149413

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,15 @@ matrix:
     include:
         - php: 5.3
           env: DEPLOY=yes COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
-        - php: 7.0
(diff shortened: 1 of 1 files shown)


SpacePossum:
>Does travis not allow us to ask for 7.1 rather than nightly?

I don't think there is an alias setup'ed by Travis of 7.1 to the current PHP build (as was done for 7.0 until 7.0 was released).

Nice idea btw., we could leave the nightly in as is even when 7.1 stable is out right?

keradus:
after 7.1 will be officially added as 7.1 and not nightly then let see what will be the nightly. It 7.2 then great to have, if only just 7.1.X then doesn't matter.

PR #2058, opened 2016-07-08
DeclareStrictTypesFixer - prove bug

keradus:
I see no reason why we remove that empty line...

### DIFF
--- a/tests/Fixer/Strict/DeclareStrictTypesFixerTest.php
+++ b/tests/Fixer/Strict/DeclareStrictTypesFixerTest.php
@@ -85,6 +85,7 @@ class A {
             ),
             array(
                 '<?php declare(strict_types=1);
+
                 phpinfo();',
                 '<?php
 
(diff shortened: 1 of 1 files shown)


keradus:
any thoughts @Seldaek @SpacePossum ?

Seldaek:
Nope I don't see why it would need to touch that space.

keradus:
awesome ! (especially as it conflicting with other rules).

could you prepare a fix for fixer using my utests please?

Seldaek:
Eh I really can't at the moment I only have my phone next few days. If
someone else can great :)

keradus:
I see.
@SpacePossum , would you be able to play with it please?

SpacePossum:
I played around with a bit, but is harder than expected because of the line break attached to the open tag... I would say lets park this one until we figured out; https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2062
(to prevent wasting dev. time on doing the same thing twice)

keradus:
replaced by #3597

PR #2069, opened 2016-07-11
Add semicolon after instruction fixer

SpacePossum:
Fixer that makes sure every instruction is terminated with a semicolon.

(On master only because I need CurlyBraceTransformer.)

### DIFF
--- a/tests/Fixer/Semicolon/SemicolonAfterInstructionFixerTest.php
+++ b/tests/Fixer/Semicolon/SemicolonAfterInstructionFixerTest.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\Semicolon;
+
+use PhpCsFixer\Test\AbstractFixerTestCase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class SemicolonAfterInstructionFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideCases
+     * @requires PHP 5.4
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideCases()
+    {
+        return array(
+            array(
+                '<?php $a = [1,2,3]; echo $a{1}; ?>',
+                '<?php $a = [1,2,3]; echo $a{1} ?>',
+            ),
+            array(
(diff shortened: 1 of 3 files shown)


SpacePossum:
missing case:
```php
<?= 1 ?>
```

tests/Fixer/Semicolon/SemicolonAfterInstructionFixerTest.php
```diff
@@ -0,0 +1,68 @@
...
+                "<?php echo 123;\n\t?>",
+                "<?php echo 123\n\t?>",
+            ),
+            array(
+                '<?= 1; ?>',
+                '<?= 1 ?>',
+            ),
+            array(
+                "<?= '1_'; /**/    ?>",
+                "<?= '1_' /**/    ?>",
```
keradus: T_INLINE_HTML on our PHP 5.3 matrix

tests/Fixer/Semicolon/SemicolonAfterInstructionFixerTest.php
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class SemicolonAfterInstructionFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideCases
+     * @requires PHP 5.4
```
keradus: - no need for require 5.4 when only one case is failing on 5.3 and all the rest are working
- so, we test whole fixer only on 5.4+, but run it on 5.3+. then, on 5.3 we have no idea how the fixer will behave (as it still runs, but we dont check how)
- isn't whole thing more about `short_open_tag` instead of PHP version ?

SpacePossum: >isn't whole thing more about short_open_tag instead of PHP version ?

it is, I look at;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/Contrib/ShortEchoTagFixerTest.php#L23

for a quick fix, but will refactor!

keradus: Short echo tag fixer is different.
Whole fixer is suppose to work even with `short_open_tag` disabled (here, there is no need for it)

keradus: nope, I mistook it with short_tag

keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2071

keradus:
Thank you @SpacePossum.

SpacePossum:
:+1:  
btw. it awesome that you delete the branch as well, keeping everything clean and the development cycle fast : D

keradus:
oh, sorry, it wasn't intended :( just a habit...
I will try to not do that in future, in any case you could restore the branch (eg from github interface)

SpacePossum:
its all good man :)

Issue #2070, opened 2016-07-11
Bug in lowercase_keywords with PHP7


tgabi333:
It seems that PHP7 does not fix `Foo:CLASS` to `Foo:class`

See: https://github.com/tgabi333/PHP-CS-Fixer/commit/b797c5469a305f32ec998a4355038e507a7fcef6 and https://travis-ci.org/tgabi333/PHP-CS-Fixer/builds/143955227

GrahamCampbell:
That's not a lowercase keyword in the same sense that `true` and `false` is?

SpacePossum:
It is a keyword, but used for something else than `true` and `false`,
for ref.: http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class

Since the introduction all methods of casing of `::class` are allowed (https://3v4l.org/c2ogC), so this can be fixed by a fixer.

GrahamCampbell:
Ok, right. :+1: for adding this to the existing fixer then.

tgabi333:
It is currenlty works for PHP5 but not for PHP7

tgabi333:
Here is the problem: https://3v4l.org/YYNPS

Came with this commit: https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/0668080d9dbfa31b0b61badf94a7c4b0bc5e5004

SpacePossum:
so the token for `class` changed from T_CLASS to T_STRING on PHP7... hmm :)
thanks for the info!
btw. do you plan for writing a PR to fix this on the PHP-CS-Fixer?

tgabi333:
I would love to help you. I'm still investigating the root problem. Found this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/search?utf8=%E2%9C%93&q=CT_CLASS_CONSTANT

SpacePossum:
I think the root might be that the custom CT_CLASS_CONSTANT token type is not listed as keyword here;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tokenizer/Token.php#L264
(but I didn't look into it in depth to see what might happen if it was added)

tgabi333:
I found too, thank you @SpacePossum , tomorrow i will create a proper PR for this. 

https://travis-ci.org/tgabi333/PHP-CS-Fixer/builds/143990898

keradus:
remember, 1.11 for bugfix.

PR #2072, opened 2016-07-12
LowercaseKeywordsFixer - handle CT_CLASS_CONSTANT

tgabi333:
Two changes plus tests:
`CT_CLASS_CONSTANT ` is recognized case insensitive mode
`CT_CLASS_CONSTANT ` was added to list of known keywords

Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2070

### DIFF
--- a/Symfony/CS/Tokenizer/Token.php
+++ b/Symfony/CS/Tokenizer/Token.php
@@ -270,7 +270,7 @@ class Token
                 'T_INTERFACE', 'T_ISSET', 'T_LIST', 'T_LOGICAL_AND', 'T_LOGICAL_OR', 'T_LOGICAL_XOR',
                 'T_NAMESPACE', 'T_NEW', 'T_PRINT', 'T_PRIVATE', 'T_PROTECTED', 'T_PUBLIC', 'T_REQUIRE',
                 'T_REQUIRE_ONCE', 'T_RETURN', 'T_STATIC', 'T_SWITCH', 'T_THROW', 'T_TRAIT', 'T_TRY',
-                'T_UNSET', 'T_USE', 'T_VAR', 'T_WHILE', 'T_YIELD', 'CT_ARRAY_TYPEHINT',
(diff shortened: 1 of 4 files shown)


SpacePossum:
:+1:

GrahamCampbell:
Maybe add one more tokenizer unit test too?

keradus:
like ?

Soullivaneuh:
> like ?

Maybe `::CLASS`? But I'm not sure about the benefit...

tgabi333:
@Soullivaneuh There is two test already, one is `::class` and the other is `::cLaSS` in `ClassConstantTest`

As i see another testcase would only test the case sensitve flag of `Token::equalsAny`

keradus:
you didn't change `equalsAny` so no need to change the tests for it

GrahamCampbell:
Looks good. 👍

Symfony/CS/Tokenizer/Token.php
```diff
@@ -270,7 +270,7 @@ public static function getKeywords()
                 'T_INTERFACE', 'T_ISSET', 'T_LIST', 'T_LOGICAL_AND', 'T_LOGICAL_OR', 'T_LOGICAL_XOR',
                 'T_NAMESPACE', 'T_NEW', 'T_PRINT', 'T_PRIVATE', 'T_PROTECTED', 'T_PUBLIC', 'T_REQUIRE',
                 'T_REQUIRE_ONCE', 'T_RETURN', 'T_STATIC', 'T_SWITCH', 'T_THROW', 'T_TRAIT', 'T_TRY',
-                'T_UNSET', 'T_USE', 'T_VAR', 'T_WHILE', 'T_YIELD', 'CT_ARRAY_TYPEHINT',
+                'T_UNSET', 'T_USE', 'T_VAR', 'T_WHILE', 'T_YIELD', 'CT_ARRAY_TYPEHINT', 'CT_CLASS_CONSTANT',
```
GrahamCampbell: Note that this is already present on 2.x.

keradus:
Thank you @tgabi333.

SpacePossum:
thanks @tgabi333 !

PR #2067, opened 2016-07-09
Added DeclareEqualNormalizeFixer

keradus:
Closes #2064

PR #2073, opened 2016-07-12
DeclareStrictTypesFixer - Remove fix CS fix logic from fixer.

SpacePossum:
With the `DeclareEqualNormalizeFixer` (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2067 :+1: ) in place the CS fixing part of the declare strict fixer can (and should) be removed to make it more atomic.


keradus:
Thank you @SpacePossum.

Issue #2075, opened 2016-07-14
New linter doesn't show the line number


GrahamCampbell:
The php 7 tokenizer linter doesn't show the line number it occurred on.

GrahamCampbell:
Old linter: `Syntax error, unexpected ')' in lol.php on line 5.`.

New linter: `Syntax error, unexpected ')'.`.

GrahamCampbell:
Doesn't help that PHP has a bug where if you get the parse error as a string, it doesn't insert the line number correctly. Compare these two outputs:

1. https://3v4l.org/gMGNN
2. https://3v4l.org/t6BWp

GrahamCampbell:
Note that the current code in php-cs-fixer calls the `getMessage` method, which is not the same as the to string implementation.

GrahamCampbell:
Oh, wait, I think this is a bug in 3v4l.org trying to string out sensitive content about itself.

GrahamCampbell:
I've sent a PR to address this.

GrahamCampbell:
Hmm, my PR is incorrect since it clearly just gets the current file from the call stack rather than the one we're checking. Since we're only passing PHP the actual file content, and not the file, it doesn't look like we can really fix this in this way. The PHP bug with the bad replacement still exists (but isn't our problem).

---

I'd recommend we try to use the old tokeniser to generate the error after we use the new one to check each file originally.

keradus:
See no value in it. From tool point of view the file is valid or not. we don't care why. we don't care on which line.
We could grab some info why, but don't need to waste time on it.
It's user responsibility to have valid files, and if we tell him that file is not valid he could lint it himself.

GrahamCampbell:
It's useful if we break their files and the post-fixing linter finds an error.

keradus:
If we break it we don't save it.
And for debugging it's way to few information to have only output from linter one way or another.

hackel:
Interestingly, I came here to report that it didn't indicate the filename that caused it to fail.  I *do* get a line number, though:
`PHP Parse error:  syntax error, unexpected '$this' (T_VARIABLE), expecting ';' or '{' in  on line 32`

keradus:
Unrelated, as new linter is available only on v2.
Please create separated issue providing full command you have run (with `-vvv`) and it's output .

keradus:
@GrahamCampbell ref #2088

PR #2076, opened 2016-07-14
Mimic the output of the original linter

GrahamCampbell:
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2075 for full details. Also, note, as described in this issue that PHP has a bug that it's to string implementation has a different behaviour when called by the shutdown handler than in normal execution, hence the manual construction of the error text.

### DIFF
--- a/src/Linter/TokenizerLintingResult.php
+++ b/src/Linter/TokenizerLintingResult.php
@@ -37,8 +37,17 @@ final class TokenizerLintingResult implements LintingResultInterface
      */
     public function check()
     {
-        if (null !== $this->error) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Hmm, my PR is incorrect since it clearly just gets the current file from the call stack rather than the one we're checking. Since we're only passing PHP the actual file content, and not the file, it doesn't look like we can really fix this in this way. The PHP bug with the bad replacement still exists (but isn't our problem).

---

I'd recommend we try to use the old tokeniser to generate the error after we use the new one to check each file originally.

PR #2078, opened 2016-07-14
Added SilencedDeprecationErrorFixer

HeahDude:
The `SilencedDeprecationErrorFixer` prepends the silent operator `@` to
`trigger_error` calls only using `E_USER_DEPRECATED`.

### DIFF
--- a/Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
+++ b/Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
@@ -1,0 +1,63 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Jules Pietri <jules@heahprod.com>
+ */
+final class SilencedDeprecationErrorFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+            if (!$token->equals(array(T_STRING, 'trigger_error'), false)) {
+                continue;
+            }
+
+            $start = $index;
+            $prev = $tokens->getPrevMeaningfulToken($start);
+            if ($tokens[$prev]->isGivenKind(T_NS_SEPARATOR)) {
+                $start = $prev;
+                $prev = $tokens->getPrevMeaningfulToken($start);
+            }
+
+            if ($tokens[$prev]->isGivenKind(T_STRING) || $tokens[$prev]->equals('@')) {
+                continue;
+            }
+
+            $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $tokens->getNextTokenOfKind($index, array(T_STRING, '(')));
+            if ($tokens[$tokens->getPrevMeaningfulToken($end)]->equals(array(T_STRING, 'E_USER_DEPRECATED'))) {
+                $tokens->insertAt($start, new Token('@'));
+            }
+        }
+
+        return $tokens->generateCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'Ensures deprecation notices are silenced. Warning! This could change code behavior.';
+    }
+}
--- a/Symfony/CS/Tests/Fixer/Symfony/SilencedDeprecationErrorFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/SilencedDeprecationErrorFixerTest.php
@@ -1,0 +1,67 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Symfony;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Jules Pietri <jules@heahprod.com>
+ *
+ * @internal
+ */
+final class SilencedDeprecationErrorFixerTest extends AbstractFixerTestBase
+{
+    /**
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->makeTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                '<?php trigger_error("This is a deprecation warning."); ?>',
+            ),
+            array(
+                '<?php trigger_error("This is a deprecation warning.", E_USER_WARNING); ?>',
+            ),
+            array(
+                '<?php A\B\trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
+            array(
+                '<?php \A\B/* */\trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
+            array(
+                '<?php @trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+                '<?php trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
+            array(
+                '<?php @\trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+                '<?php \trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
+            array(
+                '<?php echo "test";@trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+                '<?php echo "test";trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
+            array(
+                '<?php //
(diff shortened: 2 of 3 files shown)


SpacePossum:
I would like a fixer to remove all silences (highly risky), but the other way around? not so sure...

Anyway looking at the implementation I would like to give some pointers.
Finding the cases is not very efficiently done, something like this is faster;
```php
for ($index = $tokens->count() - 1; $index >= 0; --$index) {
    $token = $tokens[$index];
    if (!$token->isGivenKind(T_STRING) || 'trigger_error' !== $token->getContent()) {
        continue;
    }

    // fixing logic
}
``` 

Looking directly around the token candidate using `+1` and `-1` offset causes missed case like;
```php
<?php /**/trigger_error/**/(/**/'',/**/    E_USER_DEPRECATED  /**/ ); ?>
```

using something like `$tokens->getNextMeaningfulToken` and `$tokens->getPrevMeaningfulToken` will help you.
...and there is no need to use a description as key for the return test cases ;)

HeahDude:
Ok @SpacePossum, thanks for the review, I work on it :)

HeahDude:
@SpacePossum I've addressed your comments.

keradus:
don't call `getContent` to compare token's content

HeahDude:
@keradus I don't get what you mean.

SpacePossum:
>don't call getContent to compare token's content

Means that is not the best way to check if a token is what you want. Test a token based on its type (if possible, most of them can) `Token::getType` if you want to know if the content matches _as well_ use `token:equals`. Check out the PHPDoc of last method, it will help you :)
(the most powerful and useful for you case it that the `equals` method also has a `$caseSensitive` flag, so you catch `triGgEr_erRor` and such).
Please add cases with white space, casing and comment variations and you see for yourself :)

HeahDude:
@SpacePossum ok thanks!

SpacePossum:
Thanks for all the work and time you spend on this fixer.
Looking at the fixer it will silence errors (which is risky as it will change the behavior of the code) for calls to `trigger_error` while that functions' only responsibility is to trigger an error. You might as well remove the whole `trigger_error` statement. Silencing errors can be done using other PHP functionality, which imo would be better than using a CS fixer to modify the code.
I see no use case for this fixer and I doubt the Symfony community does, therefor I'm -1 on adding it to the code base.

keradus:
Symfony **does** silent `E_USER_DEPRECATED` error (and not others, as stated in this PR).
As for regular usage of application you don't want to be noticed by that.
Then, in PHPUnit Symfony play with this error: https://github.com/symfony/phpunit-bridge/blob/master/DeprecationErrorHandler.php

keradus:
any news here ?

Symfony/CS/Tests/Fixer/Symfony/SilencedDeprecationErrorFixerTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        return array(
+            array(
+                '<?php trigger_error("This is a deprecation warning."); ?>',
+            ),
+            array(
+                '<?php trigger_error("This is a deprecation warning.", E_USER_WARNING); ?>',
+            ),
+            array(
+                '<?php @trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
```
keradus: please remove, it duplicates last case

HeahDude:
@keradus I've missed the time, I'll finish it this week-end. Thanks!

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
```diff
@@ -0,0 +1,52 @@
...
+        $tokens = Tokens::fromCode($content);
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+            if (!$token->isGivenKind(T_STRING) || 'trigger_error' !== $token->getContent()) {
+                continue;
+            }
+
+            $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $tokens->getNextTokenOfKind($index, array(T_STRING, '(')));
+            if (
+                'E_USER_DEPRECATED' === $tokens[$tokens->getPrevMeaningfulToken($end)]->getContent()
```
phansys: Just for any weird case, we could also check against the constant value (`16384`).

keradus: ?

phansys: ```in_array($tokens[$tokens->getPrevMeaningfulToken($end)]->getContent(), array('E_USER_DEPRECATED', '16384'), true)```

keradus: :-1:

Symfony/CS/Tests/Fixer/Symfony/SilencedDeprecationErrorFixerTest.php
HeahDude: I thought we need to test one to fix and one to let as is, don't we?

SpacePossum: >and one to let as is

you don't need that one :)
let say you have the following test input for `makeTest`; array('A', 'B').
Then the test will first check 'B' is changed into 'A' by the fixer and than test if 'A' is not changed by fixer.
(for ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixer/AbstractFixerTestBase.php#L97)

HeahDude:
Ok I've updated using `equals` and removed the duplicated test. Thanks for your reviews!

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+            if (!$token->isGivenKind(T_STRING) || !$token->equals(array(T_STRING, 'trigger_error'), false)) {
```
SpacePossum: I think you no longer need the givenKind check anymore

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
```diff
@@ -0,0 +1,52 @@
...
+                continue;
+            }
+
+            $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $tokens->getNextTokenOfKind($index, array(T_STRING, '(')));
+            if (
+                $tokens[$tokens->getPrevMeaningfulToken($end)]->equals(array(T_STRING, 'E_USER_DEPRECATED'))
+                && !$tokens[$tokens->getPrevMeaningfulToken($index)]->equals('@')
```
SpacePossum: doing this check first be starting to look for the start and end of the body will speed up the fixer

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
HeahDude: Is `equals` optimized enough?

Symfony/CS/Tests/Fixer/Symfony/SilencedDeprecationErrorFixerTest.php
```diff
@@ -0,0 +1,47 @@
...
+                '<?php trigger_error("This is a deprecation warning.", E_USER_DEPRECATED); ?>',
+            ),
```
SpacePossum: please add a test with diff. casing and comments. Something like;
```php
'<?php //
Trigger_Error/**/("This is a deprecation warning.", E_USER_DEPRECATED/***/); ?>';
```

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
SpacePossum: good call! lets wait for @keradus pov :)

SpacePossum:
since `trigger_error` is a function, do we want to detect namespaced versions as well?
i.e.:
```php
<?php
\trigger_error("This is a deprecation warning.", E_USER_DEPRECATED);

\A\B\trigger_error("This is a deprecation warning.", E_USER_DEPRECATED);  // false positive ATM
?>
```

HeahDude:
>since trigger_error is a function, do we want to detect namespaced versions as well?

I think we should remain in the symfony scope.

Symfony/CS/Fixer/Symfony/SilencedDeprecationErrorFixer.php
keradus: yes

HeahDude:
Comments addressed :)

GrahamCampbell:
This fixer must be marked as risky surely. Also, trigger_error could be locally defined too?

SpacePossum:
>This fixer must be marked as risky surely.

:+1: 

>could be locally defined too?

it can, so checking and not fixing something like `\A\B\trigger_error(...)` would be nice (i.e. less false positives and less risky)

HeahDude:
I've updated to handle namespaced calls.

SpacePossum:
Looking good, time to wrap it up :)

Please sort the public methods alphabetical and as last add an empty line between the `if` statements (around line 47).

:+1: 

Note @ keradus, on the master line this fixer must be marked as risky and therefor needs a re-gen of the readme ;)

HeahDude:
@SpacePossum Done, thanks for your reviews :)

SpacePossum:
LGTM :)
:+1:

keradus:
> Note @ keradus, on the master line this fixer must be marked as risky and therefor needs a re-gen of the readme ;)

Then `Warning! This could change code behavior.` is missing from fixer descr

SpacePossum:
Oh, I thought it was added during the readme generation when a fixer is risky.. silly me ;)

HeahDude:
@keradus fixed.

keradus:
@SpacePossum , on 2.x it is. but if we say it's a risky fixer on 1.x we need to handle it manually.

keradus:
:+1:

SpacePossum:
:+1:

keradus:
Thank you @HeahDude.

HeahDude:
Thank you for merging, and thank you all for your guidance and help :)

SpacePossum:
Thanks at @HeahDude  sorry for -1 vote first, nice work man :D

PR #2079, opened 2016-07-15
TernarySpacesFixer - Remove multiple spaces

SpacePossum:
I'm not sure if this is a bug report or not. I would like cases like the following to be fixed by the fixer as well:
```php
$a = $b    ?   1   :   2;
```

Currently the fixer won't touch these cases and a [test](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Tests/Fixer/Symfony/TernarySpacesFixerTest.php#L42
) exists to make sure of that. 
This PR change the fixer so it will fix the example to:
```php
$a = $b ? 1 : 2;
```


keradus:
If I remember correctly existing whitespaces was ignored to not break multiline ternaries.
With your update they are not made single line ;)

:+1:

keradus:
Thank you @SpacePossum.

PR #974, opened 2015-01-17
Add EregToPregFixer

mbeccati:
ext/ereg has been deprecated since a long time and will be unbundled in PHP7:

https://wiki.php.net/rfc/remove_deprecated_functionality_in_php7#extereg

The fixer helps with the migration by substituting eregs with pregs.

This PR is based on #973 and #972

PR #2082, opened 2016-07-18
Added MbStrFunctionsFixer

Slamdunk:
This fixer acts like `mbstring.func_overload = 2`

Reference page: http://php.net/manual/en/mbstring.overload.php

Some notes:

1. `mbstring.func_overload = 4` related functions are left untouched since they are deprecated and removed from PHP 7
1. `mbstring.func_overload = 1` mail() to mb_send_mail() is left untouched since involves a more complex refactoring in the user application
1. Like [EregToPregFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/974) this fixer does **not** handle namespaced overrided functions like: `namespace MyNamespace { function strlen() { return 1; } }` since it is impossible to distinguish, in another source file, if the function called refers to custom or native one.
1. The Fixer is marked Risky

### DIFF
--- a/Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
@@ -17,7 +17,7 @@ use Symfony\CS\Tokenizer\Tokens;
 /**
  * @author Filippo Tessarotto <zoeslam@gmail.com>
  */
-class MbStrFunctionsFixer extends AbstractFixer
+final class MbStrFunctionsFixer extends AbstractFixer
 {
     /**
      * @var array the list of the string-related function names and their mb_ equivalent
--- a/Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
+++ b/Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class MbStrFunctionsFixer extends AbstractFixer
+{
+    /**
+     * @var array the list of the string-related function names and their mb_ equivalent
+     */
+    private static $functions = array(
+        'strlen' => 'mb_strlen',
+        'strpos' => 'mb_strpos',
+        'strrpos' => 'mb_strrpos',
+        'substr' => 'mb_substr',
+        'strtolower' => 'mb_strtolower',
+        'strtoupper' => 'mb_strtoupper',
+        'stripos' => 'mb_stripos',
+        'strripos' => 'mb_strripos',
+        'strstr' => 'mb_strstr',
+        'stristr' => 'mb_stristr',
+        'strrchr' => 'mb_strrchr',
+        'substr_count' => 'mb_substr_count',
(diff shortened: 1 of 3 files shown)


Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class MbStrFunctionsFixer extends AbstractFixer
+{
+    /**
+     * @var array the list of the string-related function names and their mb_ equivalent
+     */
+    private static $functions = array(
```
GrahamCampbell: Maybe make this configurable?

GrahamCampbell: Actually, don't we already have an abstraction on renaming functions? Just use that?

Slamdunk: The only configuration I consider usefull is to distinguish mbstring.func_overload 1, 2 and 4, like I described in the PR.
To me it does not make any sense to convert i.e. `strlen` but not `strpos`

Slamdunk: About second node: missed that, can you link it to me please?

keradus: maybe https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/AbstractFunctionReferenceFixer.php ?

Slamdunk: My PR is into `1.11`: is it ok for you if I leave this PR here to be merged on `1.11` and open another one for `master` using the linked abstract class?

Symfony/CS/Tests/Fixer/Contrib/MbStrFunctionsFixerTest.php
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public function provideCases()
+    {
+        return array(
+            array('<?php $x = "strlen";'),
+            array('<?php $x = Foo::strlen("bar");'),
+            array('<?php $x = $foo->strlen("bar");'),
+
+            array('<?php $x = mb_strlen("bar");', '<?php $x = strlen("bar");'),
```
keradus: * please add a case with `\strlen` usage
* please add a cases with other changed methods (so tests won't be green if one remove half of `$functions`)

Symfony/CS/Fixer/Contrib/MbStrFunctionsFixer.php
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class MbStrFunctionsFixer extends AbstractFixer
```
keradus: please mark it final

Symfony/CS/Tests/Fixer/Contrib/MbStrFunctionsFixerTest.php
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class MbStrFunctionsFixerTest extends AbstractFixerTestBase
```
keradus: please mark it final and internal

Slamdunk:
- Added more test cases, both positive and negative
- Marked all classes as final and the test as internal
- Opened https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2122 for `master` branch

keradus:
Thank you @Slamdunk.

Issue #1899, opened 2016-05-19
[FR] Remove useless keywords on `final` classes


SpacePossum:
I would like to write and add a fixer that removes the `final` keyword from methods of a `final` declared class. ~~Secondly on a `final` class `protected` methods can safely be made `private` I think.~~

gharlan:
> Secondly on a final class protected methods can safely be made private I think.

No: https://3v4l.org/GNsrP

SpacePossum:
check, thanks :+1:

Soullivaneuh:
Looks good. We can also remove `final` keyword from `private` method.

A discussion about this was made in Sonata group: https://github.com/sonata-project/dev-kit/issues/185

We find that a `private` method can be "overridden". I said "overridden" because it's not really an override, just the possibility to have the same name. But it was broken if method signatures differ between the child and the parent.

At the end, we supposed it's a PHP bug, and this one was [confirmed](https://bugs.php.net/bug.php?id=72496) and [patched](https://github.com/php/php-src/pull/1980).

So we can say adding the `final` keyword to a private method is completely useless and should be removed.

I think this have a place to this fixer, what do you think?

SpacePossum:
```php
<?php
class Foo
{
    final private function test()
    {
        
    }
}

final class Bar extends Foo
{
    final private function test()
    {
        
    }
}
```

https://3v4l.org/6vgbs

If I get it right, it should be allowed to "override" a final private method after the PR you pointed out is merged and new PHP versions released?
If so I'm happy to see the `final` from private methods removed by a fixer, but not until the new PR versions are out so we can test against this (since before that the code is invalid).

Soullivaneuh:
> If I get it right, it should be allowed to "override" a final private method after the PR you pointed out is merged and new PHP versions released?

Should be. They don't tested with final but about method signature change (should not be an impact).

But yes it should be allowed.

> If so I'm happy to see the `final` from private methods removed by a fixer, but not until the new PR versions are out so we can test against this (since before that the code is invalid).

You're right. It should be out on PHP 5.6.24 (RC at the moment): https://github.com/php/php-src/commit/08777e9615868cf2acf4dc2a86a7ef21106beddd

SpacePossum:
I'm closing this as I use a diff. system for keeping tracking on stuff I want to do one day when I've time. Please don't reopen, I like to have a small list of issues on GH to manage. Thanks!

PR #2084, opened 2016-07-19
Added ProtectedToPrivateFixer

Slamdunk:
This fixer converts all attributes and method from protected to private when possible, so if and only if the class:
1. Is final
2. Is not abstract
3. Does not extend any other classes
4. Is not a trait
5. Does not use traits

So only where inheritance is absent

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -560,9 +560,7 @@ Choose from the list of available fixers:
    @var should always be written as @type.
 
 * **protected_to_private** [contrib]
-                        Converts protected variables
-                        and methods to private where
-                        possible.
+   Converts protected variables and methods to private where possible.
 
 * **short_array_syntax** [contrib]
    PHP arrays should use the PHP 5.4 short-syntax.
--- a/README.rst
+++ b/README.rst
@@ -559,6 +559,11 @@ Choose from the list of available fixers:
 * **phpdoc_var_to_type** [contrib]
    @var should always be written as @type.
 
+* **protected_to_private** [contrib]
+                        Converts all protected
+                        variables and methods to
+                        private if needed.
+
 * **short_array_syntax** [contrib]
    PHP arrays should use the PHP 5.4 short-syntax.
 
--- a/README.rst
+++ b/README.rst
@@ -560,9 +560,9 @@ Choose from the list of available fixers:
    @var should always be written as @type.
 
 * **protected_to_private** [contrib]
-                        Converts all protected
-                        variables and methods to
-                        private if needed.
+                        Converts protected variables
+                        and methods to private where
+                        possible.
 
 * **short_array_syntax** [contrib]
    PHP arrays should use the PHP 5.4 short-syntax.
--- a/Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
@@ -17,6 +17,7 @@ use Symfony\CS\Tokenizer\Tokens;
 
 /**
  * @author Filippo Tessarotto <zoeslam@gmail.com>
+ * @author SpacePossum
  */
 final class ProtectedToPrivateFixer extends AbstractFixer
 {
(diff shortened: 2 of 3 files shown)


SpacePossum:
Thanks for working on this!

for ref.: closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1899

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,114 @@
...
+namespace Symfony\CS\Fixer\Contrib;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Token;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class ProtectedToPrivateFixer extends AbstractFixer
```
SpacePossum: please make the class `final`, we do this on the 2.x/master line for all fixers and also for new fixers on 1.x

Symfony/CS/Tests/Fixer/Contrib/ProtectedToPrivateFixerTest.php
```diff
@@ -0,0 +1,132 @@
...
+ */
+
+namespace Symfony\CS\Tests\Fixer\Contrib;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+class ProtectedToPrivateFixerTest extends AbstractFixerTestBase
```
SpacePossum: please make this test `final` as well and add an `@internal` tag to the class description doc.

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,114 @@
...
+                $prevTokenIndex = $tokens->getPrevMeaningfulToken($prevTokenIndex);
+                $prevToken = $tokens[$prevTokenIndex];
+            }
+
+            if (!$prevToken->isGivenKind(T_PROTECTED)) {
+                continue;
+            }
+
+            $prevToken->clear();
+            $tokens->insertAt(
```
SpacePossum: maybe you can use `Tokens::overrideAt`?

SpacePossum:
The `use` detection is off I think, lets say we have something like this:
```php
<?php

final class A {
    protected function A() {

    }
}

use \DateTime as Temp;
$d = new Temp();
var_dump($d);
```

keradus:
@SpacePossum 
how?
> I would like to write and add a fixer that removes the final keyword from methods of a final declared class. 

this fixer does not play with `final` keyword removal

SpacePossum:
I think the `protected` for the method can be made `private` in my sample, 
but it won't be fixed because the fixer doesn't check if the `use` is within the `class` body.

Slamdunk:
@SpacePossum your example is covered by the last test case: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2084/files#diff-745abf3775862e89aad00d0c09ce4d33R128

In the meantime I've pushed the changed requested.

SpacePossum:
Thanks for the updates :)
About the sample; I don't think it is. Please try it the other way around, add a test with a class definition first (which should be fixed) and _below_ it an `use` import.

Slamdunk:
You are right: the fixer does not cover any case that has something after the first class.
I'm trying to fix it.

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,114 @@
...
+
+            $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
+            $prevToken = $tokens[$prevTokenIndex];
+
+            if ($prevToken->isGivenKind(T_ABSTRACT)) {
+                return true;
+            }
+
+            $useIndex = $tokens->getNextTokenOfKind($index, array(array(T_USE)));
+            if ($useIndex > 0) {
```
SpacePossum: @Slamdunk move this check and change it to
```php
if ($useIndex < classOpeningIndex) {
    return false;
}
```

that should do it :)

Slamdunk: Yes, but it's not enough, this fixer should also work for:
```php
class MyClass
{
    protected $var;
}
final class MyClass
{
    protected $var;
}
```
Fixing the latter, but this involves a lot of refactoring of my original code.
I'm working on it, using somthing like [Php4ConstructorFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Fixer/Contrib/Php4ConstructorFixer.php#L30-L34) approach

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,111 @@
...
+     * @param Tokens $tokens the Tokens instance
+     *
+     * @return bool
+     */
+    private function skipClass(Tokens $tokens, $classIndex)
+    {
+        $prevTokenIndex = $tokens->getPrevMeaningfulToken($classIndex);
+        $prevToken = $tokens[$prevTokenIndex];
+
+        if ($prevToken->isGivenKind(T_ABSTRACT) or !$prevToken->isGivenKind(T_FINAL)) {
```
SpacePossum: please use `||` over `or` in this case

Slamdunk:
Changes done

README.rst
```diff
@@ -629,6 +629,11 @@ Choose from the list of available fixers:
                         @var should always be written
                         as @type.
 
+* **protected_to_private** [contrib]
+                        Converts all protected
+                        variables and methods to
+                        private if needed.
```
keradus: it's not needed, it's only possible

keradus:
Could anyone take a look at the fixer, see it's value and make a review ?

Slamdunk:
After rebasing the builds fail because of a [header_comment fix need](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/154087852#L573) but not only my files fail, also `Symfony/CS/ShutdownFileRemoval.php` is involved and is not part of this PR.

Even if the change is trivial, I am not confortable in touching unrelated files.
Let me know what you want me to do about this.

keradus:
@Slamdunk great catch, I must mess up the merge 1.11->1.12
it's not big change, please apply in this pr, thanks !

Slamdunk:
Done

SpacePossum:
@Slamdunk please consider https://github.com/Slamdunk/PHP-CS-Fixer/pull/1

SpacePossum:
:+1: 
(but lets wait for the test result, btw. please rebase)

Slamdunk:
Everything done

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,109 @@
...
+     * @param int    $classOpenIndex
+     * @param int    $classCloseIndex
+     */
+    private function fixClass(Tokens $tokens, $classIndex, $classOpenIndex, $classCloseIndex)
+    {
+        for ($index = $classOpenIndex + 1; $index < $classCloseIndex; ++$index) {
+            $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
+            $prevToken = $tokens[$prevTokenIndex];
+
+            if ($prevToken->isGivenKind(T_STATIC)) {
```
gharlan: why is this extra part needed for `T_STATIC`?
Just skip it like all other tokens apart from `T_PROTECTED`.

Slamdunk:
> Why? Could you give an example for this case, where it would break the behavior? 

https://3v4l.org/6BdYa
In PHP 7 `$v1` raises no error, but `$v2` does

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
Slamdunk: Yup, legacy copy+paste

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
```diff
@@ -0,0 +1,109 @@
...
+     * @param int    $classIndex
+     * @param int    $classOpenIndex
+     * @param int    $classCloseIndex
+     *
+     * @return bool
+     */
+    private function skipClass(Tokens $tokens, $classIndex, $classOpenIndex, $classCloseIndex)
+    {
+        $prevToken = $tokens[$tokens->getPrevMeaningfulToken($classIndex)];
+        if ($prevToken->isGivenKind(T_ABSTRACT) || !$prevToken->isGivenKind(T_FINAL)) {
```
Slamdunk: Yup, legacy copy+paste

gharlan:
> https://3v4l.org/6BdYa
> In PHP 7 $v1 raises no error, but $v2 does

Thanks! Didn't know that. 
Theoretically you could change protected methods to private ones anyway in such a case.

Slamdunk:
You are right, but honestly I'd prefer to leave all trait stuff untouched.
Inheritance, both horizontally and vertically, is just a source of mess, even more when changing PHP version.

Symfony/CS/Fixer/Contrib/ProtectedToPrivateFixer.php
Slamdunk: No more "outdated diff bla bla", github?

keradus:
Replaced by #2249

PR #2088, opened 2016-07-20
TokenizerLintingResult - expose line number of parsing error

### LINKED ISSUE #2075: New linter doesn't show the line number

keradus:
closes #2075

### DIFF
--- a/src/Linter/TokenizerLintingResult.php
+++ b/src/Linter/TokenizerLintingResult.php
@@ -38,7 +38,11 @@ final class TokenizerLintingResult implements LintingResultInterface
     public function check()
     {
         if (null !== $this->error) {
-            throw new LintingException($this->error->getMessage(), $this->error->getCode(), $this->error);
+            throw new LintingException(
+                sprintf('PHP Parse error: %s on line %d.', $this->error->getMessage(), $this->error->getLine()),
+                $this->error->getCode(),
+                $this->error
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
That doesn't work?

GrahamCampbell:
That gives you the line in our code that threw the error.

keradus:
I don't think so. Code proves the other way

GrahamCampbell:
Oh, cool. I guess I was a bit drunk when I tested this then. That was the reason I closed my last PR.

GrahamCampbell:
👍 then to the idea, but I'd prefer it if we slightly matched PHP's format more closely - doesn't have to be exactly the same tho. :P

GrahamCampbell:
How about this?

```php
throw new LintingException(
    sprintf('%s on line %d.', $this->error->getMessage(), $this->error->getLine()),
    $this->error->getCode(),
    $this->error
);
```

keradus:
doesn't matter...

SpacePossum:
:+1:  looking good :)

GrahamCampbell:
> doesn't matter...

Definitely does. The message you have there is horrible IMO.

keradus:
it's copy-pasted from `php -l file` on php7

Issue #2035, opened 2016-07-03
"no whitespaces between offset accessors"


phansys:
Input:
```php
<?php
$foo [0] [1]   [2] = 3;
$bar = $foo [0] [1]   [2];
$baz [0]
     [1]
     [2] = 3;
$foobar = $foo {0} [1]   {2};
$withComments [0] // here is a comment
     [1] // and here is another
     [2] = 3;
```
Expected:
```php
<?php
$foo[0][1][2] = 3;
$bar = $foo[0][1][2];
$baz[0][1][2] = 3;
$foobar = $foo{0}[1]{2};
$withComments [0] // here is a comment
     [1] // and here is another
     [2] = 3;
```

keradus:
multiline should be addressed as well
`{` for indexes too

phansys:
@keradus, I've added more use cases, but really I don't know what to do in some cases, by instance, if there are comments between these accessors.

keradus:
For sure not remove them ;) I will ignore whitespaces between `]` and `[` if there is comment between.

SpacePossum:
@keradus can be closed since https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2089 has been merged :)

keradus:
thanks

PR #2089, opened 2016-07-22
Add `no_spaces_around_offset` fixer

phansys:
| Q             | A
| ------------- | ---
| Branch        | master
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #2035
| License       | MIT
| Doc PR        |

**NOTE TO MERGER**: This PR depends on #2093 and must be merged after it.


tests/Fixer/Whitespace/NoSpacesBetweenOffsetFixerTest.php
```diff
@@ -0,0 +1,123 @@
...
+{
+    /**
+     * @dataProvider provideCases
+     */
+    public function testFixSpaceInsideOffset($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function testLeaveNewLinesAlone()
```
keradus: I'm not sure that this is about new lines, it's more about comments?
as here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2089/files#diff-fd7fb655f90d8659624713f1c4edcec1R101  - new lines are removed

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return $tokens->isAnyTokenKindsFound(array('[', CT_ARRAY_INDEX_CURLY_BRACE_OPEN));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        /* @var $token \PhpCsFixer\Tokenizer\Token */
```
keradus: 1. no need for FQCN, it's imported
2. no need for typehinting, it's typehinted here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Tokenizer/Tokens.php#L26

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
```diff
@@ -0,0 +1,60 @@
...
+        foreach ($tokens as $index => $token) {
+            if (!$token->equalsAny(array(new Token('['), new Token(array(CT_ARRAY_INDEX_CURLY_BRACE_OPEN, '{'))))) {
```
keradus: 1. no need to create 2 new objects for every single loop iteration
2. no need to create 2 new tokens at all

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
```diff
@@ -0,0 +1,60 @@
...
+                continue;
+            }
+
+            $prevMeaningfulIndex = $tokens->getPrevNonWhitespace($index);
+            $prevMeaningfulToken = $tokens[$prevMeaningfulIndex];
+            if ($prevMeaningfulToken->equals('=') || $prevMeaningfulToken->isComment()) {
```
keradus: `equals('=')` is not needed

SpacePossum:
nice idea for a fixer :)

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
phansys: The inline docblock was removed.

tests/Fixer/Whitespace/NoSpacesBetweenOffsetFixerTest.php
phansys: I've updated this method name. I forgot to rename it after I copied the example from another fixer.

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
phansys: Removed.

src/Fixer/Whitespace/NoSpacesBetweenOffsetFixer.php
phansys: If I use `{` instead of `new Token(array(CT_ARRAY_INDEX_CURLY_BRACE_OPEN, '{'))`, the fixer fails:
```
--- Expected
+++ Actual
@@ @@
 <?php
-$foo{0}{1}{2} = 3;
+$foo {0} {1}   {2} = 3;
```
Maybe you have a better approach for this situation, really I don't know how this happens.

phansys: Ping @keradus.

keradus: @phansys 
`if (!$token->equalsAny(array('[', array(CT_ARRAY_INDEX_CURLY_BRACE_OPEN, '{')))) {`

phansys: Thank you @keradus, done.

keradus:
:+1: but please move to `symfony` ruleset

phansys:
Moved.

keradus:
You don't need to squash the commits. I do it during merging.

Please fix CI.

phansys:
I think we should use another name instead "no_spaces_between_offset", since this fixer don't removes only spaces between offsets, also the space between an offset and another valid expression. What about "no_offset_surrounding_spaces" or something similar?

keradus:
what about `no_spaces_outside_offset` ?
or even combine it with `no_spaces_inside_offset` and create options which spaces should be removed, all by default.

phansys:
Combining both fixers seems to be the better approach, but how can name the resultant fixer? "no_offset_spaces"?

keradus:
or `no_spaces_around_offset`
I'm fine for both

phansys:
If I allow the curly braces syntax for the "inside" fixer, it throws `InvalidArgumentException: Invalid param $startIndex - not a proper block start`. How can I avoid it?

SpacePossum:
>If I allow the curly braces syntax for the "inside" fixer, it throws InvalidArgumentException: Invalid param $startIndex - not a proper block start. How can I avoid it?

Its because the block type is not added yet to the types for which one can search the end and start. I've added there in this PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2093

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
```diff
@@ -0,0 +1,122 @@
...
+        foreach ($tokens as $index => $token) {
+            if (!$token->equalsAny(array('[', array(CT_ARRAY_INDEX_CURLY_BRACE_OPEN)))) {
+                continue;
+            }
+
+            if (in_array('inside', $this->configuration, true)) {
+                if ($token->equals('[')) {
+                    $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE, $index);
+                } else {
+                    $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index, true);
```
keradus: why once we use 3rd param and once not ?

phansys: It was just a try to make it work @keradus, I'll update these calls after #2093 is merged.

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
```diff
@@ -0,0 +1,122 @@
...
+    {
+        return $tokens->isAnyTokenKindsFound(array('[', CT_ARRAY_INDEX_CURLY_BRACE_OPEN));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
```
keradus: no need to iterate over whole collection if there is no elements in config

```php
if (empty($this->configuration)) {
    return;
}
```

tests/Fixer/Whitespace/NoSpacesAroundOffsetFixerTest.php
```diff
@@ -0,0 +1,304 @@
...
+            ),
+        );
+    }
+
+    /**
+     * @dataProvider provideConfigurationCases
+     */
+    public function testFixWithConfiguration(array $configuration, $expected)
+    {
+        static $input = <<<'EOT'
```
keradus: move input to provider as well.

keradus:
:+1: except minor comment and if green after dep merge

SpacePossum:
please update the https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md as well since the fixer is renamed/replaced.

keradus:
that was my first thought, but the replaced fixer was introduced on 2.x (even not 1.12), which is not yet released ;)

SpacePossum:
ha, very well! sorry people, Monday morning Im gonna get some coffee and stop spamming wrong review suggestions ;)

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
```diff
@@ -0,0 +1,122 @@
...
+
+        $this->configuration = $configuration;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(array('[', CT_ARRAY_INDEX_CURLY_BRACE_OPEN));
```
stof: this should also contain a check for the empty configuration (return ``false`` as it is not a candidate when nothing is configured)

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
stof: is the check still needed here if we handle it in ``isCandidate`` ?

keradus: one place is enough for sure. why you propose to use `isCandidate` instead ?

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
SpacePossum: shouldn't the fixer throw a config exception if the config is empty? It not something that can change during runtime anyway.

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
stof: @keradus isn't ``isCandidate`` meant to exclude filters early when we know for sure that they don't need to run ? The empty configuration is such a case

keradus: The initial meaning was 'is given Tokens collection a candidate for a fixer'.
I don't insist any option here, just wanna make it common

stof: Well, if this fixer is configured to do nothing, no collection is a candidate for being fixed...

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
keradus: does any of existing fixers throw it ?

keradus:
please rebase

phansys:
Rebased.

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Comment/HeaderCommentFixer.php#L271

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php#L37

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRenameFixer.php#L37

but other don't, so we can go anyway I guess :)

phansys: You mean to throw an exception instead of use the default configuration if the user doesn't provide any fixer config?

SpacePossum: If the user doesn't provide configuration I would expect the default to be used, but if the user provides an empty (or invalid) configuration I would expect a configuration exception to be thrown.
But we can go both ways (since it is an edge case anyway) :)

phansys:
CI failures are not related.

src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php
phansys: I've added an exception for empty configuration.

keradus:
restarted

phansys:
Is there something preventing to merge this PR?

tests/Fixer/Whitespace/NoSpacesAroundOffsetFixerTest.php
```diff
@@ -0,0 +1,325 @@
...
+
+function someFunc()    {   $someVar = [];   }
+EOF;
+        $this->doTest($expected);
+    }
+
+    public function provideOutsideCases()
+    {
+        return array(
+            array(
```
SpacePossum: test request for the basic form:
```php
$a = $b   [0]    ;
```

phansys: Addressed.

SpacePossum: thanks :)
I was playing around with your fixer (will PR soon, minor details) and was wondering,
wouldn't you expect the fixer with `outside`-config to fix to:
```php
$a = $b[0];
```

and not:
```php
$a = $b[0]    ;
```
?

phansys: Thanks @SpacePossum.
Regarding your question, I'm trying to not touch here any other token that isn't part of the offset get/set, so I added [another simple test](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2089/commits/3640265778c079a1a92c06256452a688228b73c9) in order to expose that fact.

SpacePossum: I check, so it is more like `inside` and `before` (rather than `outside`), all good :+1:

SpacePossum:
tested on the following code bases;
- Symfony master, one valid fix
- composer 1.12, zero hits
- Twig 1.x, zero hits
- Silex master, zero hits

I'm :+1: for the current version of this PR, nice fixer @phansys  :)

(On a side note, maybe we should open up a meta ticket to normalize the way we do fixer configuration on 2.x before that one gets released)

keradus:
Thank you @phansys.

PR #2093, opened 2016-07-24
Tokens - add block type BLOCK_TYPE_ARRAY_INDEX_CURLY_BRACE

SpacePossum:
So finding the start and end for the custom token types can be done.


src/Tokenizer/Tokens.php
```diff
@@ -33,6 +33,7 @@ class Tokens extends \SplFixedArray
     const BLOCK_TYPE_ARRAY_SQUARE_BRACE = 4;
     const BLOCK_TYPE_DYNAMIC_PROP_BRACE = 5;
     const BLOCK_TYPE_DYNAMIC_VAR_BRACE = 6;
+    const BLOCK_TYPE_ARRAY_CURLY_BRACE = 7;
```
keradus: `ARRAY_CURLY_BRACE` vs `ARRAY_INDEX_CURLY_BRACE`

keradus:
Thank you @SpacePossum.

PR #2095, opened 2016-07-24
Transformers - add required PHP version

keradus:


src/Tokenizer/Transformer/ImportTransformer.php
```diff
@@ -40,6 +40,14 @@ public function getCustomTokenNames()
     /**
      * {@inheritdoc}
      */
+    public function getRequiredPhpVersionId()
+    {
+        return 70000;
```
gharlan: 50600

src/Tokenizer/Transformer/ArraySquareBraceTransformer.php
```diff
@@ -40,6 +40,16 @@ public function getCustomTokenNames()
     /**
      * {@inheritdoc}
      */
+    public function getRequiredPhpVersionId()
+    {
+        // short array syntax was introduced in PHP 5.4, but the fixer is smart
+        // enough to handel it even before 5.4
+        return 50000;
```
SpacePossum: maybe we could define the min. PHP version of the Fixer (5.3.6) on the Application class and return that as default on an AbstractTransformer class?

SpacePossum:
:+1:

tests/Tokenizer/Transformer/ImportTransformerTest.php
```diff
@@ -23,6 +23,7 @@
 {
     /**
      * @dataProvider provideProcessCases
+     * @requires PHP 7.0
```
gharlan: 5.6 ;)

src/Tokenizer/Transformer/ArraySquareBraceTransformer.php
keradus: that was my first implementation. then i throw it away.
why? because then it would be so easy to ever overwrite it and always apply all transformers, even when not needed. and waste time...

SpacePossum: makes sense, cool :)  ignore my comment

PR #2100, opened 2016-07-27
Update Readme to use friendsofphp package not fabpot

AdenFraser:


SpacePossum:
thanks, but please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2055 and others :)

PR #2104, opened 2016-08-01
Explicitly set finder when using directory in args

MattKetmo:
When you use a `.php_cs` config file which set a finder plus you pass a path as argument of the fix command, the behavior differ if that path is a file or a directory.

For instance using the PHP-CS-Fixer project:

```bash
# This will apply the rules from the config file to a **single** PHP file.
./php-cs-fixer fix --config-file .php_cs Symfony/CS/Console/Command/FixCommand.php
```

```bash
# When using a directory the rules are applied on the **whole project**
# (and not just the path given in args).
./php-cs-fixer fix --config-file .php_cs Symfony/CS/Console/Command/
```

This patch force the finder to be overridden when specifying a directory in args.

### DIFF
--- a/Symfony/CS/Console/Command/FixCommand.php
+++ b/Symfony/CS/Console/Command/FixCommand.php
@@ -337,6 +337,7 @@ EOF
         } elseif ($stdin) {
             $config->finder(new \ArrayIterator(array(new StdinFileInfo())));
         } elseif (null !== $path) {
+            $config->finder(new \DirectoryIterator($path));
             $config->setDir($path);
         }
 


keradus:
1. no utests to prove the bug
2. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.11/Symfony/CS/Config/Config.php#L85

keradus:
btw, works on 2.x

Symfony/CS/Console/Command/FixCommand.php
```diff
@@ -337,6 +337,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         } elseif ($stdin) {
             $config->finder(new \ArrayIterator(array(new StdinFileInfo())));
         } elseif (null !== $path) {
+            $config->finder(new \DirectoryIterator($path));
```
keradus: with this change it won't use DefaultFinder

MattKetmo:
Indeed the issue seems fixed in 2.x (actual master branch).

```
# This checks only the given directory
./php-cs-fixer fix --config .php_cs.dist src/Console/Command/
```

I'm fine with it (I don't have much time to investigate more for 1.x and create a proper test).

Thanks.

Issue #2106, opened 2016-08-03
UnalignEqualsFixer - ShortArraySyntaxFixer - priority bug


SpacePossum:
Input

```php
<?php
$a =  array  ();
```

after run 1;
```php
<?php
$a =   [];
// unalign_equals, short_array_syntax
```

after run 2;
```php
<?php
$a = [];
// unalign_equals
```

```bash
PHP CS Fixer version 1.11.5 by Fabien Potencier (d3d08b7)
PHP 7.0.9-1+deb.sury.org~xenial+1 (cli) ( NTS )
```

keradus:
Great catch !
It looks easy.
Please, could you send a fix @SpacePossum ?

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2108

PR #2108, opened 2016-08-03
ShortArraySyntaxFixer, TernarySpacesFixer, UnalignEqualsFixer - fix priority bug

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2106#issuecomment-237187939

turns out the token clear done by the `short_array_syntax` fixer causes problems because the `unalign_equals` expects one white space token (and not multiple in a row or with cleared tokens in between)

### DIFF
--- a/Symfony/CS/Fixer/Contrib/ShortArraySyntaxFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ShortArraySyntaxFixer.php
@@ -53,7 +53,7 @@ class ShortArraySyntaxFixer extends AbstractFixer
      */
     public function getPriority()
     {
-        // should be run before the UnalignEqualsFixer.
+        // should be run before the UnalignEqualsFixer and TernarySpacesFixer.
         return 1;
     }
 }
--- a/Symfony/CS/Fixer/Contrib/ShortArraySyntaxFixer.php
+++ b/Symfony/CS/Fixer/Contrib/ShortArraySyntaxFixer.php
@@ -31,10 +31,10 @@ class ShortArraySyntaxFixer extends AbstractFixer
             $openIndex = $tokens->getNextTokenOfKind($index, array('('));
             $closeIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $openIndex);
 
-            $token->clear();
-
             $tokens->overrideAt($openIndex, '[');
             $tokens->overrideAt($closeIndex, ']');
+
+            $tokens->clearTokenAndMergeSurroundingWhitespace($index);
         }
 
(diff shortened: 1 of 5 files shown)


Symfony/CS/Tests/Fixtures/Integration/priority/short_array_syntax,unalign_equals.test
```diff
@@ -0,0 +1,12 @@
+--TEST--
+Integration of fixers: short_array_syntax,unalign_equals.test.
+--CONFIG--
+level=none
+fixers=short_array_syntax,unalign_equals
+--INPUT--
```
SpacePossum: needs an requires annotation for PHP 5.4.0

SpacePossum:
Maybe I can prove the need for the merge white space tokens in the test of the fixer itself as well (the integration test proves it). Note to myself to check other fixers that call `clear` as well.

Symfony/CS/Fixer/Contrib/ShortArraySyntaxFixer.php
```diff
@@ -47,4 +47,12 @@ public function getDescription()
     {
         return 'PHP arrays should use the PHP 5.4 short-syntax.';
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return 1;
```
keradus: please put a comment describing why we need non-default priority

keradus:
It would be great ;)

SpacePossum:
test added :)

I'll check the other fixers later if I can find some time, which won't be soon...  so don't wait for it ;)

GrahamCampbell:
👍

SpacePossum:
note to self to update the intergration test with:
```php
<?php
$b = [];
$a = $b + array(  );
$a = $b ?   array ( ) :   array (  ); 
$a = $b ==   array (  ) ? 1 : 2;
$a = $b ===   array (  ) ? 1 : 2;
$a = $b !=   array ( )  ? 1 : 2;
$a = $b !==   array ( )  ? 1 : 2;
$a = $b <>  array (  )   ? 1 : 2;
```

I'll be a away / busy for the upcoming weeks so proceed with this PR however you people want, I think it is RTM but you could wait/provide a PR for the other cases

SpacePossum:
this should be it I think :)

GrahamCampbell:
Nice!

keradus:
Great work !

keradus:
Thank you @SpacePossum.

GrahamCampbell:
# ALERT!

This breaks everything if you have neither align fixer turned on. It unaligns everything.

PR #2111, opened 2016-08-04
.travis.yml - rename PHP 7.1 env

keradus:


GrahamCampbell:
👍

SpacePossum:
:+1: 

(but why not leave nightly in with allow fail, then when 7.1 stable is released we will be testing 7.2 beta's and RC's out of the box...)

GrahamCampbell:
Little point. I'd recommend just adding in 7.2 in later on when we actually start to care.

keradus:
exactly
I add it as nightly instead of 7.1 as it was the only option to use 7.1 on travis back there

PR #2113, opened 2016-08-05
[POC/WIP] Add FunctionTypehintDefaultFixer

SpacePossum:
Hi all,

Just an early opening to see how the test do, but more important to see what the community thinks :)
Currently the fixer does something like:
```php
<?php 
//expected
function a( array $a1 = array()){}
```

```php
<?php 
// input
function a( $a1 = array()){}
```

Next up would be for PHP7
```php
// expected
function a(int $a = 1) {
    return $a;
}
```

```php
// input
function a($a = 1) {
    return $a;
}
```

This is of cause all very risky.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -648,6 +648,11 @@ Choose from the list of available fixers:
                         with preg. Warning! This could
                         change code behavior.
 
+* **function_typehint_default** [contrib]
+                        Function arguments with
+                        defaults should be typehinted.
+                        (Risky fixer!)
+
 * **header_comment** [contrib]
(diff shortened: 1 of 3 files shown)


keradus:
```php
function a($a = 1) {
    return $a;
}
```

you don't know should it be int or not. int may be one of valid values, but may not be the only one.


What do you think about doing it on top of other indicator - not default value, but phpdoc typehinting ?

SpacePossum:
will work on this some more if I find the time, moving to backlog for now, thanks all for looking!

PR #2118, opened 2016-08-08
SilencedDeprecationErrorFixer - adjust level

keradus:
ref #2078
no risky fixers should be enabled by default

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -551,11 +551,6 @@ Choose from the list of available fixers:
                         "(double)" and "(real)" as
                         "(float)".
 
-* **silenced_deprecation_error** [symfony]
-                        Ensures deprecation notices
-                        are silenced. Warning! This
-                        could change code behavior.
-
 * **single_array_no_trailing_comma** [symfony]
                         PHP single-line arrays should
                         not have trailing comma.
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
👍

HeahDude:
Hey, we discussed this with Fabien, and from his pov, not changing the code in that specific case is a bug that should be fixed by this fixer, so it should not be risky at all.

GrahamCampbell:
This is definitely risky. The fact that it fixes a "bug" by definition means it's changed the behaviour.

HeahDude:
But changing the behavior is not risky when you expect to fix a bug, right?

GrahamCampbell:
Yes it is. Anything other than a cs fix is risky. A bug fix is risky.

HeahDude:
I understand your rules, but life, sometimes, is not binary. Concretely what is risky here?

GrahamCampbell:
That is the definition of risky here. This is a code style fixing tool, not a static analyzer.

keradus:
For Symfony not-silenced deprecation error is a bug, fixer do the job and Symfony is happy. Awesome.
We do the same - for use comparing with `==` is a bug, `===` should be used. We have a fixer that does that. Awesome.

But in both cases - sb else may not consider this as bugs, and since this is changing the behaviour we need to warn user about that.

HeahDude:
@keradus so the new ruleset you are adding (#2119) is the only way to get it running on the symfony repo easily?

keradus:
no, one could also use CLI `rules` arg or define ruleset in config file ;)
But I thought it would be easier if that ready to use ruleset would exists.

HeahDude:
Thanks ;)

PR #2119, opened 2016-08-08
Create @Symfony:risky ruleset

keradus:
ref #2118

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -545,7 +545,7 @@ Choose from the list of available fixers:
 * **ordered_imports**
                          Ordering use statements.
 
-* **php_unit_construct**
+* **php_unit_construct** [@Symfony:risky]
                          PHPUnit assertion method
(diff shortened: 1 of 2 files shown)


SpacePossum:
:+1:

SpacePossum:
Do we need an overall integration test for this ruleset?

keradus:
Symfony integration tests was created on 1.x line with the following reasoning:
- oh boy, we have so many fixers, some of them may conflicts one another ! or have priority issues ! we don't know
- we have priority tests, but only for those we know that should be in given order
- let us create a big integration tests with a lot of fixers
- oh, most of the fixers are inside Symfony ruleset, let us use it!

So the reasoning was to test most fixers, not to test Symfony ruleset.

Here, the case is different - only 3 rules

SpacePossum:
Thanks! :)

PR #2124, opened 2016-08-11
Add TernaryToNullCoalescingFixer

Slamdunk:
Change wherever possible [ternary operator to null coalescing operator](http://php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op):

``` php
// Before
$x = isset($var) ? $var : 'ops';
// After
$x = $var ?? 'ops';
```

It reduces code duplication, so it should always be favoured.

The fixer is marked as **risky** because of this example (https://3v4l.org/Fjddg):

``` php
// Before
$a = 1;
var_dump(isset($a) ? $a : isset($b) ? $b : false);
// Outputs:
// Notice: Undefined variable: b
// NULL

// After
$a = 1;
var_dump($a ?? $b ?? false);
// Outputs:
// int(1)
```

This only occurs if you concatenate ternary operator without parenthesis, that is an insane usage of ternary operator: the fixer actually fix the code to the right intent, but has to be marked as risky.

### DIFF
--- a/tests/Fixer/Operator/TernaryToNullCoalescingFixerTest.php
+++ b/tests/Fixer/Operator/TernaryToNullCoalescingFixerTest.php
@@ -1,0 +1,126 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\Operator;
+
+use PhpCsFixer\Test\AbstractFixerTestCase;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ *
+ * @internal
+ */
+final class TernaryToNullCoalescingFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideCases
+     * @requires PHP 7.0
+     *
+     * @param mixed      $expected
+     * @param null|mixed $input
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideCases()
+    {
+        return array(
+            array('<?php $x = isset($a) ? $a[1] : null;'),
+            array('<?php $x = isset($a) and $a ? $a : "";'),
+            array('<?php $x = "isset($a) ? $a : null";'),
+            array('<?php $x = isset($a) ? $$a : null;'),
+            array('<?php $x = isset($a) ? "$a" : null;'),
+            array('<?php $x = isset($a) ?: false;'),
+            array('<?php $x = $a ? $a : isset($b) ? $b : isset($c) ? $c : "";'),
+            array('<?php $x = $y ?? isset($a) ? $a : null;'),
+            array('<?php $x = isset($a) ?: $b;'),
+            array('<?php $x = isset($a, $b) ? $a : null;'),
+            array('<?php $x = $a && isset($b) ? $b : null;'),
+            array('<?php $x = $a & isset($b) ? $b : null;'),
+            array('<?php $x = ! isset($a) ? $a : null;'),
+            array('<?php $x = false === isset($a) ? $a : 2;'),
+            array('<?php $x = 4 * isset($a) ? $a : 2;'),
+            array('<?php $x = 3 ** isset($a) ? $a : 2;'),
+            array('<?php $x = 1 | isset($a) ? $a : 2;'),
+            array('<?php $x = (array) isset($a) ? $a : 2;'),
+            array('<?php $x = isset($a[++$i]) ? $a[++$i] : null;'),
+            array('<?php $x = function(){isset($a[yield]) ? $a[yield] : null;};'),
+            array('<?php $x = isset($a[foo()]) ? $a[foo()] : null;'),
+            array('<?php $x = isset($a[$callback()]) ? $a[$callback()] : null;'),
+
+            array(
+                '<?php $x = $a ?? null;',
+                '<?php $x = isset($a) ? $a : null;',
+            ),
+            array(
+                '<?php $x = $a ?? 1; $y = isset($b) ? "b" : 2; $x = $c ?? 3;',
+                '<?php $x = isset($a) ? $a : 1; $y = isset($b) ? "b" : 2; $x = isset($c) ? $c : 3;',
+            ),
+            array(
+                '<?php $x = $a[ $b[ "c"  ]]   ?? null;',
+                '<?php $x = isset   (  $a[$b["c"]]) ?$a[ $b[ "c"  ]]   : null;',
+            ),
+            array(
+                '<?php $x = $a ?? $b[func(1, true)];',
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
> The fixer is marked as risky because of this example (https://3v4l.org/Fjddg):

It shouldn't be risky because it's broken. That's not what risky means.

This fixer should correct any precedence issues it causes with brackets.

Slamdunk:
> This fixer should correct any precedence issues it causes with brackets.

I'm sorry I don't understand, how
```php
isset($a) ? $a : isset($b) ? $b : false
```
should be fixed?

GrahamCampbell:
Yeh, sure it should?

GrahamCampbell:
`isset($a) ? $a : isset($b) ? $b : false` needs fixing to `$a ?? isset($b) ? $b : false`, but no further, since this is evaluated as:

`($a ?? isset($b)) ? $b : false`

since the original was evaluated as:

`(isset($a) ? $a : isset($b)) ? $b : false`.

---


Fixing to `$a ?? $b ?? false` would be wrong since that's not the same at all.

Slamdunk:
Ok, I excluded any subsequent ternary operator that isn't rounded by parenthesis, so the risky flag can be removed.

Thanks for the hint

tests/Fixer/Operator/TernaryToNullCoalescingFixerTest.php
```diff
@@ -0,0 +1,75 @@
...
+                '<?php $x = $a ?? ($b ?? "");',
+                '<?php $x = isset($a) ? $a : (isset($b) ? $b : "");',
+            ),
+            array(
+                '<?php $x = ($a ?? isset($b)) ? $b : "";',
+                '<?php $x = (isset($a) ? $a : isset($b)) ? $b : "";',
+            ),
+            array(
+                '<?php $x = $a ?? isset($b) ? $b : isset($c) ? $c : "";',
+                '<?php $x = isset($a) ? $a : isset($b) ? $b : isset($c) ? $c : "";',
```
keradus: add case like 
```php
<?php $x = isset($a) /*a*/ ? /*b*/ $a /*c*/ : /*d*/ isset($b) /*e*/ ? /*f*/ $b /*g*/ : /*h*/ isset($c) /*i*/ ? /*j*/ $c : "";
```

Slamdunk: As of yet this case would be fixed like:
```php
<?php $x = $a /*c*/ ?? /*d*/ isset($b) /*e*/ ? /*f*/ $b /*g*/ : /*h*/ isset($c) /*i*/ ? /*j*/ $c /*k*/ : /*l*/ "";
```
Syntactically ok, but losing `/*a*/` and `/*b*/`: where should they go?
Something like this would be ok?

```php
<?php $x = /*a*/ /*b*/ $a /*c*/ ?? /*d*/ isset($b) /*e*/ ? /*f*/ $b /*g*/ : /*h*/ isset($c) /*i*/ ? /*j*/ $c /*k*/ : /*l*/ "";
```

keradus: TBH I don't care where the comments would go, but they must not be removed

Slamdunk:
Comments preserved, logics simplified.

keradus:
@GrahamCampbell , could you please play with a fixer and make a full review ?

GrahamCampbell:
Will do at the weekend. I'll try my best to break it. :trollface:

SpacePossum:
ping @GrahamCampbell did you find time to test?

SpacePossum:
>$x = isset($a[yield]) ? $a[yield] : null;

is not valid PHP ;) please use https://3v4l.org/TDKnd (note that HHVM is acting diff. :/ )

gharlan:
complete example: https://3v4l.org/KsmSR

Slamdunk:
All higher precedence operators are now excluded because the behaviour actually changes:
```php
$x = $a && isset($b) ? $b : null;
$x = $a & isset($b) ? $b : null;
$x = ! isset($a) ? $a : null;
$x = false === isset($a) ? $a : 2;
$x = 4 * isset($a) ? $a : 2;
$x = 3 ** isset($a) ? $a : 2;
$x = 1 | isset($a) ? $a : 2;
$x = (array) isset($a) ? $a : 2; // That's OMG, but cannot be touched
```
Also every changing code and function calls are excluded:
```php
$x = isset($a[++$i]) ? $a[++$i] : null;
$x = function(){isset($a[yield]) ? $a[yield] : null;};
$x = isset($a[foo()]) ? $a[foo()] : null;
$x = isset($a[$callback()]) ? $a[$callback()] : null;
```
**NOTE**: what I don't like in my implementations for this cases is that I had to check everything "manually", i.e. explicitly writing `T_ARRAY_CAST`, `T_BOOLEAN_AND` etc.
I tried to use `TokensAnalyzer` without success.

If you have better ideas and/or cases ping me.

Slamdunk:
@keradus squashed and rebased. Is there a chance of seeing this on [v2.0.0 milestone](https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestone/4)?

keradus:
1. is it still risky ? or riskness has been fixed ?
2. we need anyone first to review it (cc @GrahamCampbell @gharlan , maybe you could ?)

Slamdunk:
Riskness has been fixed

keradus:
Great to hear ;)

Slamdunk:
@SpacePossum I'm periodically rebasing this PR, and now I'm getting this error in tests:
```
1) PhpCsFixer\Tests\FixerFactoryTest::testShortFixerDefinition
Not valid amount of fixers using ShortFixerDefinition.
If this test is failing it means one of those scenario occurred:
- you introduced new Fixer using `ShortFixerDefinition`, you should use `FixerDefinition` instead,
- you update the Fixer to stop using `ShortFixerDefinition`, you should decrease the guard value.
```
Can you help me to fix it? I can't figure out how 😕

SpacePossum:
Hi,

For new fixers we added test to make sure these come with a description.
You might want to add some like this to your fixer:

```php
    /**
     * {@inheritdoc}
     */
    {
        return new FixerDefinition(
            'Use null coalescing operator (??) wherever possible.',
            array(
                new VersionSpecificCodeSample(
                    "<?php\n\$sample = isset(\$a) ? \$a : \$b;",
                    new VersionSpecification(70000)
                ),
            )
        );
    }
```

includes are like:
```php
use PhpCsFixer\FixerDefinition\VersionSpecification;
use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
```

Slamdunk:
Thank you, now everything is ok.

Note: there is no evidence on https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.0.0/doc/COOKBOOK-FIXERS.md about the new `getDefinition()` method and the `FixerDefinition` class, it would be very usefull for newcomers.

SpacePossum:
Ah yes, the always the docs ^_^
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2404

Slamdunk:
Backtick operator adopted in the description.

SpacePossum:
@Slamdunk please see https://github.com/Slamdunk/PHP-CS-Fixer/pull/2

keradus:
Thank you @Slamdunk.

keradus:
Thank you @SpacePossum

Issue #2098, opened 2016-07-25
Mirroring issues


JBleijenberg:
Whenever i try to create a mirror of this package (the original package is required by Magento2) with Satis, i get the following error

`  [PharException]
  tar-based phar "/private/var/folders/8c/rtrl_2sj3ldggkpnrcpjr3c40000gn/T/composer_archive5795f901a61ca.tar" cannot be created, filename "tests/Fixtures/Integration/misc/simplified_null_return,no_useless_return,no_whitespace_in_blank_line,blank_line_before_return,no_e
  xtra_consecutive_blank_lines.test" is too long for tar file format`

What is going wrong here?
(I know what's going wrong here, the filename is too long. But please fix it)

SpacePossum:
which app/spec. states the name is too long?

keradus:
http://php.net/manual/en/phar.fileformat.tar.php

we are building phars without any problems.

also, why there is tests folder in your phar ?

JBleijenberg:
@keradus because it's in the package on packagist. And satis repackages eveything when it creates a mirror.

JBleijenberg:
Btw. I can still use 1.11, after that i get this error

JBleijenberg:
@SpacePossum Satis uses Composer to recreate the package. That's where the error pops up

SpacePossum:
I see, would a `.gitattributes` help?

keradus:
also, do you know how phar file is created? as I assume it don't use our config for box.

Slamdunk:
`.gitattributes` is coupled with git archive, so if the phar is compiled from source retrieved in any method except the zip, the error remains.

keradus:
the error is very tricky to reproduce. we also build phar file and we have no issues with it

keradus:
@JBleijenberg could you check if this is still an issue on 1.12 ?

JBleijenberg:
@keradus it works in 1.12
But just a question out of curiosity, why is your master branch completely different from v1.12?
When i require >= 1.11, composer will satis will break when it tries to repackage dev-master.

keradus:
Great to here it's working! In that case I'm closing this issue, big thanks for raising it and checking it!

The `master` branch is dev line of 2.x version.
I have no idea why satis will fallbacking 1.11 to dev-master, for me it's wrong behaviour

eyalzek:
[...]

PR #2128, opened 2016-08-12
[git] Add .gitattributes to remove unneeded files

Slamdunk:


keradus:
this should be sent to 1.x line

Slamdunk:
I sent it to master on purpose because master has a clear `/tests` folder, on `1.x` instead I thought it would be strange to have a `./Symfony/CS/Tests export-ignore`.

keradus:
Why strange ?
And it should solve a problem with too long paths in phar file right ? and that problem ocucrs on 1.x as wel..

Slamdunk:
I'm not aware of the phar problem you are talking about, btw https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2128

keradus:
ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2098
is it related?

Slamdunk:
No: that bug may or may not emerge depending on how the source is retrieved.

By the way, as stated in #2098 `.gitattributes` only affects zip archive creation: no drawbacks as long as all the production code remains available within the zip

PR #2131, opened 2016-08-15
Use friendsofphp as the vendor in composer

minuwan:
Use "friendsofphp" as the vendor instead of "fabpot" in composer installs.


keradus:
Travis is failing

PR #2138, opened 2016-08-16
composer.json - disallow to run on PHP 7.2+

keradus:
Requesting for comments.

As all new 7.X versions could brings new tokens (or reuse existing token in different place) it's impossible to say that we will properly handle those new tokens with existing code.
That could lead to wrong changes in file, ignoring file or crashing the fixer.
Instead, let not guess that probably we will works on 7.2+ code, but when we will be sure we do just update supported PHP versions.

Nice side effect is that we would be able to introduce new transformers for new tokens without breaking BC and bumping MAJOR version.

### DIFF
--- a/Symfony/CS/Resources/phar-stub.php
+++ b/Symfony/CS/Resources/phar-stub.php
@@ -15,8 +15,8 @@ if (defined('HHVM_VERSION_ID')) {
         fwrite(STDERR, "HHVM needs to be a minimum version of HHVM 3.9.0\n");
         exit(1);
     }
-} elseif (!defined('PHP_VERSION_ID') || PHP_VERSION_ID < 50306) {
-    fwrite(STDERR, "PHP needs to be a minimum version of PHP 5.3.6\n");
+} elseif (!defined('PHP_VERSION_ID') || PHP_VERSION_ID < 50306 || PHP_VERSION_ID >= 70200) {
+    fwrite(STDERR, "PHP needs to be a minimum version of PHP 5.3.6 and maximum version of PHP 7.1.*\n");
     exit(1);
 }
 
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.3.6 || ^7.0",
+        "php": "^5.3.6 || >=7.0 <7.2",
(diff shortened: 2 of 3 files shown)


GrahamCampbell:
Maybe go with `^5.3.6 || ^7.0 <7.2` or `>=5.3.6 <7.2`?

keradus:
First, I wanna have `^5.3.6 || ^7.0 <7.2` but it's not working (https://semver.mwl.be/#?package=symfony%2Fsymfony&version=%5E2%20%7C%7C%20%5E3.0%20%3C3.1&minimum-stability=stable)
The later one looks ugly for me, like we would support crappy v6 :/

And what do you think about idea itself ?

SpacePossum:
:+1: for the idea :) 
it would require at least one new release for each new PHP version, even when no changes are needed in the app. Will you also add runtime checks for the PHP version for the phar releases?

keradus:
done

SpacePossum:
Lgtm :+1:

keradus:
thanks @SpacePossum (btw, please check gitter private msg)

Symfony/CS/Resources/phar-stub.php
```diff
@@ -15,8 +15,8 @@
         fwrite(STDERR, "HHVM needs to be a minimum version of HHVM 3.9.0\n");
         exit(1);
     }
-} elseif (!defined('PHP_VERSION_ID') || PHP_VERSION_ID < 50306) {
-    fwrite(STDERR, "PHP needs to be a minimum version of PHP 5.3.6\n");
+} elseif (!defined('PHP_VERSION_ID') || PHP_VERSION_ID < 50306 || PHP_VERSION_ID >= 70200) {
```
GrahamCampbell: These checks are not the same as the composer checks though?

keradus: yes, with this check one could run on php6. i don't care.

GrahamCampbell: these checks are exactly what i suggested for the composer file since they were more concise, but you didn't want them there. that's really why i'm asking

keradus: preferred installed way is composer.
in composer.json we also specify supported version

PR #2142, opened 2016-08-16
Fixed deployments

GrahamCampbell:
Surely this'll fix it?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,9 @@ matrix:
     fast_finish: true
     include:
         - php: 5.3
-          env: DEPLOY=yes COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+        - php: 5.3
+          env: DEPLOY=yes
         - php: 5.4
         - php: 5.5
         - php: 5.6


GrahamCampbell:
Seems box broke this in their 2.7.3 update when they dropped support for php 5.3 and 5.4.

keradus:
it's not fixing anything... before deploy we install deps once again without prefer lowest but with platform.

keradus:
again, they didn't.
https://github.com/box-project/box2/issues/151
they just wrongly build phar file, installation via composer on 5.3 works fine

GrahamCampbell:
Yeh. I didn't read the first error properly. I understand now what's going on. Even thgen, I'd still recommend this PR even though it doesn't fix this per say.

GrahamCampbell:
> they just wrongly build phar file, installation via composer on 5.3 works fine

They intentionally built it like that according to their release notes.

keradus:
Then it would only add new job to build matrix, pointless

GrahamCampbell:
No, it means our phar doesn't use old dependencies?

keradus:
our phar use as new deps as possible that will still works on 5.3.6

GrahamCampbell:
No it doesn't. It uses prefer lowest, use uses symfony 2.3.0? Am i going crazy?

keradus:
Yes it does. It uses 2.6.x

keradus:
if you don't believe just unpack .phar file and see the files.

GrahamCampbell:
2.6.0?

GrahamCampbell:
I'll take a closer look later. For now, I'll trust that you're right. I have no reason to assume otherwise really. :)

keradus:
depends what is the newest 2.6 during phar creation.

PR #2144, opened 2016-08-16
Remove temporary files not deleted by destructor on failure

### LINKED ISSUE #2022: PHP-CS-Fixer leaves a temporary file at the root of the project
When using the 1.11.4 version, a ``cs_fixer_tmp_...`` file stays at the root of the project at the end of the run

adawolfa:
This PR should fix #2022

LintManager creates temporary files and deletes them in destructor. In certain circumstances (i. e. fatal error), the destructor not get called. I have solved this by registering shutdown function, which unlinks files left in index.

I'm aware this is not _clean_ solution and atm I have no idea how to write utest for this behavior. Any suggestions are welcome.

### DIFF
--- a/Symfony/CS/LintManager.php
+++ b/Symfony/CS/LintManager.php
@@ -28,10 +28,23 @@ class LintManager
      */
     private $temporaryFile;
 
+    /**
+     * Shutdown files removal handler.
+     *
+     * @var ShutdownFileRemoval
+     */
+    private $shutdownFileRemoval;
+
+    public function __construct()
+    {
+        $this->shutdownFileRemoval = new ShutdownFileRemoval();
+    }
+
     public function __destruct()
     {
         if (null !== $this->temporaryFile) {
             unlink($this->temporaryFile);
+            $this->shutdownFileRemoval->detach($this->temporaryFile);
         }
     }
 
@@ -67,6 +80,7 @@ class LintManager
     {
         if (null === $this->temporaryFile) {
             $this->temporaryFile = tempnam('.', 'cs_fixer_tmp_');
+            $this->shutdownFileRemoval->attach($this->temporaryFile);
         }
 
         file_put_contents($this->temporaryFile, $source);
--- a/Symfony/CS/ShutdownFileRemoval.php
+++ b/Symfony/CS/ShutdownFileRemoval.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of the PHP CS utility.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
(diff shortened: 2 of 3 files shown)


Symfony/CS/LintManager.php
```diff
@@ -67,11 +83,22 @@ public function createProcessForSource($source)
     {
         if (null === $this->temporaryFile) {
             $this->temporaryFile = tempnam('.', 'cs_fixer_tmp_');
+            self::$temporaryFiles[$this->temporaryFile] = $this->temporaryFile;
```
keradus: `[foo => foo, bar => bar]` :(

adawolfa: Should I replace value by a TRUE, or should I use in_array?

keradus:
Nice work so far!

I would move that logic to separated class.
sth like
```php
interface ShutdownFileRemoval
{
    public function add(string $file);
    public function remove(string $file);
    public function remove();
}
```

Each instance of `ShutdownFileRemoval` would register it's own shutdown function.

Then, you will have no static methods. A lot easier to test.

Also, thanks to this, you will be apply to apply same logic in `AbstractIntegrationTest` about `$tmpFile`

Symfony/CS/LintManager.php
keradus: check out my last proposal

adawolfa:
OK, stay tuned.

Symfony/CS/ShutdownFileRemoval.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * Removes a file from shutdown removal.
+     *
+     * @param string $path
+     */
+    public function detach($path)
+    {
+        if ($key = array_search($path, $this->files, true)) {
```
keradus: do not assign inside condition

Symfony/CS/ShutdownFileRemoval.php
```diff
@@ -0,0 +1,66 @@
...
+        }
+    }
+
+    /**
+     * Removes attached files.
+     */
+    public function clean()
+    {
+        foreach ($this->files as $file) {
+            @unlink($file); // @ - intentionally
```
keradus: then describe why it's intentionally

Symfony/CS/ShutdownFileRemoval.php
```diff
@@ -0,0 +1,66 @@
...
+ */
+
+namespace Symfony\CS;
+
+/**
+ * Handles left over temporary files removal.
+ *
+ * @author Adam Klvač <adam@klva.cz>
+ */
+class ShutdownFileRemoval
```
keradus: final + `@internal`

keradus:
Please update integration test as well ;)

keradus:
Thank you @adawolfa.

PR #2145, opened 2016-08-17
Specified line endings in .gitattributes.

adawolfa:
I suggest specifing line endings into `.gitattributes` file, otherwise hundreds of tests don't pass.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,5 @@
 .travis.yml export-ignore
 appveyor.yml export-ignore
 phpunit.xml.dist export-ignore
+*.php text eol=lf
+*.test text eol=lf


keradus:
https://github.com/composer/composer/blob/master/.gitattributes

keradus:
?

adawolfa:
I have overlooked the link target, I'm sorry! Aren't those lines sufficient?

keradus:
Currently the control of white characters are checked by fixer itself.
If we want to go with git as well - why not use more advanced checks?

adawolfa:
Well, I think those additional options don't make any difference when checking out files. Line ending option does. When I switched EOL to LF via git config, I had to re-checkout all files. Then all tests passed.

keradus:
it disallows you to add the file with tabs, for example

SpacePossum:
if we start with `.gitattributes` I'm :+1: for making it as complete as possible, so adding the other lines as well

keradus:
@adawolfa
could you update ?

SpacePossum:
btw. if take
```
*.yml   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=2
```

we should update;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/.travis.yml
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/appveyor.yml

as these are indented with 4 and not 2 spaces

keradus:
```diff
-*.yml   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=2
+*.yml   text eol=lf whitespace=blank-at-eol,blank-at-eof,space-before-tab,tab-in-indent,tabwidth=4
```

magic ;)

keradus:
Replaced by #2188

PR #2149, opened 2016-08-18
Stop using the deprecated DefaultFinder on Config

Soullivaneuh:
You deprecated this class but you are still using it.

This produce not wanted and not solvable deprecation errors like this: https://travis-ci.org/Soullivaneuh/php-cs-fixer-styleci-bridge/jobs/153204311#L392

Related failing PR from my bridge: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/pull/74

### DIFF
--- a/Symfony/CS/Config.php
+++ b/Symfony/CS/Config.php
@@ -12,8 +12,6 @@
 
 namespace Symfony\CS;
 
-use Symfony\CS\Finder\DefaultFinder;
-
 /**
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Katsuhiro Ogawa <ko.fivestar@gmail.com>
@@ -37,7 +35,7 @@ class Config implements ConfigInterface
         $this->description = $description;
         $this->level = FixerInterface::SYMFONY_LEVEL;
         $this->fixers = array();
-        $this->finder = new DefaultFinder();
+        $this->finder = new Finder();
         $this->customFixers = array();
     }
 


Symfony/CS/Config.php
```diff
@@ -12,7 +12,6 @@
 
 namespace Symfony\CS;
 
-use Symfony\CS\Finder\DefaultFinder;
```
GrahamCampbell: Too many blank lines here now. :P

Soullivaneuh:
Fixed CS.

keradus:
The reason why we have introduced new class instead of modifying exisitng one is that the changes we have made are BC breakers.
For compatibility reasons we still use original finder, but we also deprecated it and encourage to use new one. If you are ready to use new one, that will works in different way - specify it in you config file.

keradus:
Closing as stated.

Soullivaneuh:
@keradus I can understand, but how shall I fix the failing test so? https://travis-ci.org/Soullivaneuh/php-cs-fixer-styleci-bridge/builds/153204307

Thanks.

keradus:
Configure Finder in `.php_cs` file.
You could use our new one (Finder instead of DefaultFinder) or one of your own.
Be aware that they will behave differently (that's why we didn't switch to not break BC)

keradus:
Also, you could not treat deprecation notices as errors

Soullivaneuh:
I'm not talking about `.php_cs` but the bridge which create a config.

For your 1.12 version, I use `\Symfony\CS\Config::create();` as recommended. But this method use your deprecated `DefaultFinder` which trigger a deprecation error.

This is why I provided this PR. The error is triggered even if we make sure to use the right class.

Concerned line: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/pull/74/files#diff-b77b80825f9a37eb99f8064048f80d9fR113

> Also, you could not treat deprecation notices as errors

This is not an acceptable solution because I won't be warned of deprecation at all, even for my usage.

Soullivaneuh:
See also the related test: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/blob/2406ac6ab7a243e1b4d4f5f73c12e5905c629877/tests/ConfigBridgeTest.php#L14

I can't mark it as legacy because it is not, and I can't fix the deprecation because it's triggered from your code. :-)

Soullivaneuh:
I also tried that:

```php
$config = \Symfony\CS\Config::create()->finder(new \Symfony\CS\Finder());
```

This change nothing, still have the error.

And this is logic, because the `trigger_error` is placed on the constructor, which is called on the config constructor.

SpacePossum:
Indeed, always creating a default `finder` in the configuration class is not needed (as it might be set to something else using the `setter` before using the `finder`). 
Only creating the finder (of type `DefaultFinder` for BC reasons) in the `getFinder`-method if the `finder` is `null` is not possible, since it would be a BC break (as the `finder` is a `protected` attr. of the class and it would change that is always set as it is now).
Changing the `finder` to the new `finder` class would be a BC break as well (as prop. in this PR).
So for now there is not much we can do, I would ignore the warning. On the master branch I hope we ditch the protected fields and only create the `finder` when needed (and not always as it is now)

keradus:
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2244/commits/3f67e32fdcdf4d2c2f4bfeef438253a5fda9df15

Soullivaneuh:
@keradus Good thing, but only for 2.0.

I'll mark tests as legacy waiting dropping 1.x support form my library.

PR #2101, opened 2016-07-28
[WIP] Fixed ClassDefinitionFixer behavior for single interfaces on multiple lines.

ezzatron:
@keradus This is a re-submit of #2054 targeted against the 1.12 branch. Original PR text:

@SpacePossum Here's my attempt at a fix for #2053.

The breaking test is this example:

```php
<?php
class/**/Test123
extends  /**/        \RuntimeException    implements

TestZ{
}
```

I'm actually not sure whether this should be treated as multi-line, or single-line. It could probably go either way I guess.

PR #2155, opened 2016-08-26
ClassDefinitionFixer - overhaul

SpacePossum:
Replaces: 
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2101
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1979

Closes:
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2053
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1976 1 of 2

Addition fixes:
- (proper) support for anonymous classes
- (proper) support for more namespace madness

TODO
- [ ] master version
- [x] configuration options
- [x] check/test priority
- [x] check/test against other code bases (master SF and Twig tested, no hits)


GrahamCampbell:
Why delete that integration test?

SpacePossum:
The new fixer doesn't have the priority issue the current one has, so the test has become false.

GrahamCampbell:
Would be good to keep the test though to prove it still works.

SpacePossum:
we don't have a test to prove there is _not_ a priority issue, as the possible combinations with all the fixers against each would be extreme

keradus:
Actually, we have:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/1.12/Symfony/CS/Tests/Fixtures/Integration/misc

For sure it's not needed to have all combinations, but in rare cases when we want to check if bunch of fixers are working properly together, even without priority issue, we still could do that.

SpacePossum:
For now I get;
```
1) Symfony\CS\Tests\IntegrationTest::testIntegration with data set #43 (Symfony\CS\Test\IntegrationCase Object (...))
Set priorities must be significant. If fixers used in reverse order return same output then the integration test is not sufficient or the priority relation between used fixers should not be set.
Failed asserting that two strings are not identical.
```

so that makes it legit to remove the priority test.
But sure I can make something else if it is needed.

keradus:
I didn't go that deep, I am not saying is it needed in this case or not.
But it's possible, eg: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixtures/Integration/misc/php_closing_tag%2Cshort_tag.test#L7

SpacePossum:
ready for review

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -24,17 +25,61 @@
...
+     * @var array<string, bool>
+     */
+    private static $defaultConfig = array(
+        // put class declaration on one line
+        'singleLine' => false,
+        // if a classy extends or implements only one element than put it on the same line
+        'singleItemSingleLine' => false,
+        // if an interface extends multiple interfaces declared over multiple lines put each interface on its own line
+        'multiLineExtendsEachSingleLine' => false,
+    );
```
keradus: so, could I pass multiple true ?

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
```diff
@@ -24,17 +25,61 @@
...
+
+    /**
+     * @var array
+     */
+    private static $config;
```
keradus: what is default configuration?
what option is PSR-2 compatible ?

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: depends on what you want, I struggled with the naming of the options so I'm open for other names :)

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: current default is part of PSR 2 (spacing), but not all of it (i.e. make multi line etc.)

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: So I could use as config this array?
```php
 array(
        // put class declaration on one line
        'singleLine' => true,
        // if a classy extends or implements only one element than put it on the same line
        'singleItemSingleLine' => true,
        // if an interface extends multiple interfaces declared over multiple lines put each interface on its own line
        'multiLineExtendsEachSingleLine' => true,
    )
```

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: hard to see it, as default config is false everywhere... i got lost

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: yes, that would work fine :)

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
SpacePossum: SF2:
```php
private static $defaultConfig = array(
    'singleLine' => true,
    'singleItemSingleLine' => false, // doesn't matter
    'multiLineExtendsEachSingleLine' => false, // doesn't matter
);
```

PSR2:
```php
private static $defaultConfig = array(
    'singleLine' => false,
    'singleItemSingleLine' => false,
    'multiLineExtendsEachSingleLine' => true,
);
```

Current default of the fixer:
```php
private static $defaultConfig = array(
    'singleLine' => false,
    'singleItemSingleLine' => true,
    'multiLineExtendsEachSingleLine' => true,
);
```

Symfony/CS/Fixer/PSR2/ClassDefinitionFixer.php
keradus: how could I "put all on one line" and "put interfaces in their own lines" at the same time ?

SpacePossum: that config wouldn't make sense...
I can make some changes, not sure where to go though

Symfony/CS/Tests/Fixtures/Integration/priority/class_definition,trailing_spaces.test
```diff
@@ -1,22 +0,0 @@
...
-{
-}
-
---INPUT--
-<?php
-class Aaa implements
-    Symfony\CS\Tests\Fixer,\RFb,
-    \Fcc1, \GFdd
-{
-}
```
keradus: could we recover that test back and put it in `misc` and test it without priority check ?

keradus:
Please validate the input to configuration.
Having all false is pointless, having conflicting settings is bad as well.

BTW, the RuleSet will need to be updated on 2.x line. Also, default setting on 2.x should be pure PSR2 as usuall.

SpacePossum:
>Having all false is pointless

It is not, the fixer will still fix the spacing between keywords following PSR2. It will however not fix multiline/single line, it will not touch that part of classy definitions.

> having conflicting settings is bad [sic].

Good point, I'll update the configuration check.

>BTW, the RuleSet will need to be updated on 2.x line. Also, default setting on 2.x should be pure PSR2 as usuall.

Yes, making a master version of this has been on the TODO list of the PR since it was opened, I'll follow up on that as soon as 1.x has been done :)

keradus:
:+1:

keradus:
@SpacePossum is something still needed here? Is this PR a WIP or not ?

SpacePossum:
IMHO this is RTM :)

GrahamCampbell:
I thought PSR-2 was a strict subset of Symfony, so how come symfony have not followed psr-2 in that one case we have here?

keradus:
PSR2: **if** you would write it in many lines, then ...
Symfony: don't write it in many lines

SpacePossum:
PSR 2 
>Lists of implements MAY be split across multiple line

SF doesn't want that, so they do follow PSR2, but you can configure the fixer to follow PSR2 in another way

GrahamCampbell:
Ah, I see. :+1:

keradus:
Thank you @SpacePossum.

keradus:
Awesome work @SpacePossum !
just got merged into 2.x line, please prepare PR for rulesets ;)

Issue #2156, opened 2016-08-26
Delete tmp file, handle exception


SpacePossum:
```
[10:42 ]-[possum@cosmos1]-[/home/possum/work/PHP-CS-Fixer]-[git 1_12_ClassDef] 
$ php php-cs-fixer fix -vvv test.php 
You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.
Loaded config from "/home/possum/work/PHP-CS-Fixer/.php_cs".
F
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) /home/possum/work/PHP-CS-Fixer/test.php (header_comment, class_definition, braces, eof_ending)
Fixing time per file:
[0.181 s] /home/possum/work/PHP-CS-Fixer/test.php

Fixed all files in 0.185 seconds, 6.000 MB memory used
PHP Fatal error:  Uncaught ErrorException: unlink(/home/possum/work/PHP-CS-Fixer/cs_fixer_tmp_OtfzdY): No such file or directory in /home/possum/work/PHP-CS-Fixer/Symfony/CS/LintManager.php:47
Stack trace:
#0 [internal function]: {closure}(2, 'unlink(/home/po...', '/home/possum/wo...', 47, Array)
#1 /home/possum/work/PHP-CS-Fixer/Symfony/CS/LintManager.php(47): unlink('/home/possum/wo...')
#2 [internal function]: Symfony\CS\LintManager->__destruct()
#3 {main}
  thrown in /home/possum/work/PHP-CS-Fixer/Symfony/CS/LintManager.php on line 47
PHP Stack trace:
PHP   1. {main}() /home/possum/work/PHP-CS-Fixer/php-cs-fixer:0
PHP   2. Symfony\Component\Console\Application->run() /home/possum/work/PHP-CS-Fixer/php-cs-fixer:44
```

```
$ php-cs-fixer -V
PHP CS Fixer version 1.12.0 by Fabien Potencier and Dariusz Ruminski (ddac737)
```

```
$ php -v
PHP 7.0.9-1+deb.sury.org~xenial+1 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.9-1+deb.sury.org~xenial+1, Copyright (c) 1999-2016, by Zend Technologies
[...]

PR #2158, opened 2016-08-26
LintManager - use best effort approach when cleaning up

SpacePossum:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2156

If the file cannot be unlinked PHP will raise a warning which becomes an exception.

### DIFF
--- a/Symfony/CS/LintManager.php
+++ b/Symfony/CS/LintManager.php
@@ -31,7 +31,7 @@ class LintManager
     public function __destruct()
     {
         if (null !== $this->temporaryFile) {
-            unlink($this->temporaryFile);
+            @unlink($this->temporaryFile);
         }
     }
 


keradus:
I think it would hide the error, not fix the reason.

keradus:
I would go with this, as ultimately we cannot be sure file would be properly removed, but maybe we may sat 777 or lock the name between different processes ?

GrahamCampbell:
We need to register a shutdown function since if there's a fatal error, descrutors aren't called.

SpacePossum:
>I think it would hide the error, not fix the reason.

Indeed, but there is nothing else that can be tried or done.

>We need to register a shutdown function since if there's a fatal error, descrutors aren't called.
:+1:  there is one which uses the same best effort approach;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2144/files#diff-939f23610a6ec73c9a9b1d14c8bbf509R65

(btw. @keradus this is missing on master?)

keradus:
what missing, `ShutdownFileRemoval` usage in general or... ?

SpacePossum:
I grepped on `register_shutdown_function` on master but it didn't return anything, is this https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2144 merged to master?

keradus:
as told in private, merged 1.x->2.x few days ago.

SpacePossum:
>merged 1.x->2.x few days ago.

awesome, thanks so much! :)

PR #2162, opened 2016-08-29
Fix issue where an exception is thrown if the cache file exists but is empty.

ikari7789:
If you attempt to use PHP CS Fixer and have it set to use a cache, and the cache file exists, but is empty, it will throw an exception about a missing array key since ```unserialize()``` does not fail for an empty string.

### DIFF
--- a/Symfony/CS/FileCacheManager.php
+++ b/Symfony/CS/FileCacheManager.php
@@ -123,7 +123,7 @@ class FileCacheManager
         $data = @unserialize($content);
 
         // ignore corrupted serialized data
-        if (null === $data) {
+        if (!is_array($data)) {
             return;
         }
 


SpacePossum:
Isn't the problem here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2162/files#diff-12b3be4cb17c0a53ac8a34fe6e2d8a2cR131

```php
        // ignore corrupted serialized data
        if (null === $data) {
            return;
        }
```

as
```php
$a = @unserialize(null);
var_dump($a);

$a = @unserialize('');
var_dump($a);
```

all give `false` and not `null`

GrahamCampbell:
We should just change the check to `is_array`

ikari7789:
That seems logical. Don't know how I even missed that. Had the doc open for unserialize and everything.

SpacePossum:
:ship: :+1:

Symfony/CS/FileCacheManager.php
```diff
@@ -123,7 +123,7 @@ private function readFromFile()
         $data = @unserialize($content);
 
         // ignore corrupted serialized data
-        if (null === $data) {
+        if (false === is_array($data)) {
```
GrahamCampbell: !is_array is more conventional

ikari7789: I was only trying to keep in line with the rest of the code in the file.

keradus:
Thank you @ikari7789.

PR #2164, opened 2016-08-30
OperatorsSpacesFixer - Do not unalign double arrow and equals operators

SpacePossum:
fixes the last N reported issues

### DIFF
--- a/Symfony/CS/Fixer/Symfony/OperatorsSpacesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/OperatorsSpacesFixer.php
@@ -18,6 +18,7 @@ use Symfony\CS\Tokenizer\Tokens;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author SpacePossum
  */
 class OperatorsSpacesFixer extends AbstractFixer
 {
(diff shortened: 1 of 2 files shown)


keradus:
I would put normalizing spaces around `=` and `=>`as option instead of removing it, false by default.
Then, on master true by default, and remove that `unalign_=/=>` fixers.

keradus:
> fixes the last N reported issues


which one? so after this we will know what to close ;)

Symfony/CS/Fixer/Symfony/OperatorsSpacesFixer.php
```diff
@@ -77,4 +55,52 @@ public function getDescription()
     {
         return 'Binary operators should be surrounded by at least one space.';
     }
+
+    private function fixWhiteSpaceAroundOperator(Tokens $tokens, $index)
+    {
+        // do not change the alignment of `=>` or `=`, see `(un)align_double_arrow`, `(un)align_equals`
```
keradus: I would put normalizing spaces around `=` and `=>`as option instead of removing it, false by default.
Then, on master true by default, and remove that `unalign_=/=>` fixers.

keradus:
Info:
I have changed the target branch, 1.11 is not supported anymore.

Symfony/CS/Fixer/Symfony/OperatorsSpacesFixer.php
SpacePossum: I pref. this way because it would restores behavior to the previous way the fixer worked. 
Adding configuration would make the fixer more complex without much benefits. This is because on 1.x you can create the same result using the alignment fixers; the configuration will not be portable 1-0-1 for 2.x line so people will need to update their configuration anyway.
My suggestion is to merge this PR to 1.x and than I make a master version which will use configuration options.

keradus: very well

keradus:
:+1:

keradus:
Thank you @SpacePossum.

keradus:
merged into 2.x, @SpacePossum please do the needful

SpacePossum:
nice, thanks! I'll start working on a master version:)

PR #2169, opened 2016-09-07
Prefix all bash commands with dollar sign for consistency

bcremer:
Some of the command examples used the dollar sign in front of the command, some others did not. To be consistent with the symfony docs  all commands were prefixed with the dollar sign.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,13 +30,13 @@ your system:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


keradus:
why not.

read the contributing file !

bcremer:
Replaced be https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2170.

PR #2179, opened 2016-09-13
BinaryOperatorSpacesFixer - add (un)align configuration options

SpacePossum:


src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -54,6 +95,21 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
+            $i = 0;
+
+            /** @var AbstractAlignFixerHelper $helper */
+            foreach ($this->alignFixerHelpers as $helper) {
+                if ($tokens->isChanged()) {
+                    $tokens->clearEmptyTokens();
+                }
+
+                $helper->fix($file, $tokens);
+                ++$i;
```
SpacePossum: todo - delete useless $i

keradus:
Not sure if introducing possibility to align operators inside `BinaryOperatorSpaces` is good idea.
It makes `BinaryOperatorSpaces` more complex. We could unalign all operators in same way, but we cannot do it easily for aligning them.

Even with single fixer one could don't want to unalign `=`, but also don't want to align `=`, as he is using different logic. For now he dont have option to leave `=` as is and don't touch it.

SpacePossum:
So I misread your suggestion to add configuration, no problem.
If I understand correctly I would suggest to apply https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2164 to master after all and than merge the unalign and the align counter part fixers and add configuration to those?

keradus:
> So I misread your suggestion to add configuration, no problem.

no no, the configuration should be there

> If I understand correctly I would suggest to apply #2164 to master after all and than merge the unalign and the align counter part fixers and add configuration to those?

for sure unaligning fixers should be combined and then configurable.
not sure what to do with alignment ones, as both of them have very specific logic.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -24,6 +26,45 @@
...
+
+    /**
+     * @var AbstractAlignFixerHelper[]
+     */
+    private $alignFixerHelpers = array();
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
```
keradus: please document the input

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -24,6 +26,45 @@
 final class BinaryOperatorSpacesFixer extends AbstractFixer
 {
     /**
+     * @var array<string, bool>
+     */
+    private $configuration;
+
+    private static $defaultConfiguration = array(
+        'unalign_equals' => true,
+        'unalign_double_arrow' => true,
```
keradus:
```
'align_double_arrow' => false,
'align_equals' => false,
```

SpacePossum:
>Even with single fixer one could don't want to unalign =, but also don't want to align =, as he is using different logic. For now he dont have option to leave = as is and don't touch it.

fixed, by allowing `null` as configuration value of an item (`align_double_arrow` and/or `align_equals`)

keradus:
aaah, that's the thing.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -111,4 +180,21 @@ private function isDeclareStatement(Tokens $tokens, $index)
 
         return false;
     }
+
+    private function runHelperFixers(\SplFileInfo $file, Tokens $tokens)
+    {
+        $helpersToRunCount = count($this->alignFixerHelpers);
+        if (0 === $helpersToRunCount) {
+            return;
+        }
```
keradus: all 4 lines could be removed, that early return doesn't provide any good here, the same will happen if array is empty

keradus:
Upgrade guide needs to be updated

SpacePossum:
Done...
damn this PR got big, lucky this one is the biggest of merging conflicting Fixers ;)

keradus:
Thank you @SpacePossum.

PR #2181, opened 2016-09-13
Disallow to disable linter

keradus:
Disallow to disable linter, as it's required by 2.x tokenizer and transformer


GrahamCampbell:
It would be cool to have a way to disable the fast linting though. :)

GrahamCampbell:
I mean, it would make no difference for me, just thinking about other people using the tool. :)

keradus:
it's outside of scope of this PR

SpacePossum:
please fix the project itself :)

:+1: for the idea

PR #2186, opened 2016-09-15
[POC/Test] HHVM test

SpacePossum:


SpacePossum:
note: 1.11.0 test is a false positive

SpacePossum:
I think we have enough to move forward with the issue, i.e. both 1.11.x and 1.12.x won't work on HHVM.
I don't use HHVM in production and given the lack of others raising issues I would say this is the same for most of the user base of the fixer. Secondly I don't like HHVM breaking the same thing over and over again.
So I close this; if someone else wants to work on this I'm happy to assist up to some level.
Thanks all!

Issue #2171, opened 2016-09-07
Internal error with comment between use statements


bcremer:
When a `use` statement is disconnected with a comment that ends with an comma, an Internal error occures.

Please see the following testscript to reproduce:

```bash
$ cat > TestFile.php <<EOL
<?php
use Some\ClassName,
// some comment that ends with comma,
SomeOther\FanceClassName;

class Foo
{

}
EOL

php -l TestFile.php
php-cs-fixer fix -vvv TestFile.php
No syntax errors detected in TestFile.php
E
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[0.031 s] /tmp/TestFile.php

Fixed all files in 0.032 seconds, 4.000 MB memory used

Files that were not fixed due to internal error:
   1) /tmp/TestFile.php
```

I stumbled upon this error while fixing the following file of a real live project: https://github.com/shopware/shopware/blob/v5.2.6/engine/Shopware/Components/Model/EventSubscriber.php#L27

keradus:
```
php --version
php-cs-fixer --version
cat .php_cs
cat .php_cs.dist
```

?

bcremer:
Sure

```
php --version
php-cs-fixer --version
PHP 7.0.10 (cli) (built: Aug 21 2016 07:35:18) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.10, Copyright (c) 1999-2016, by Zend Technologies
PHP CS Fixer version 1.12.0 by Fabien Potencier and Dariusz Ruminski (ddac737)
```

The error happens also on with the latest `master` but with a slightly different message:
"`Files that were not fixed due to errors reported during linting after fixing"

keradus:
Great, do you want to propose a bugfix for it ? :)

bcremer:
Sorry, I have limited time and the error scope is rather limited.
I am sure however, that the error is caused by this line: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.0.0-alpha/src/Fixer/Import/SingleImportPerStatementFixer.php#L46

Hope this helps.

SpacePossum:
[...]

PR #2187, opened 2016-09-16
MultipleUseFixer - Fix handling comments

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2171


SpacePossum:
For master some of the fixers which have a priority related to the MultipleUserFixer have been renamed. The comment @ `getPriority` needs to be updated, integration test files needs to be updated as well.
Fixers add/removed on master need to be checked against the new implementation of the MultipleUserFixer.

keradus:
Thank you @SpacePossum.

PR #2191, opened 2016-09-20
PhpdocToCommentFixer - fix false positive for docblock of variable

### LINKED ISSUE #1715: phpdoc_to_comment bug?
using 2.x

```php
/**
 * @var Foo
 */
$foo = createFoo();
```

should this really be treated as not phpdocblock ?

keradus:
Fixes #1715

https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#description-21

> The `@var` tag MUST contain the name of the element it documents. An exception to this is when property declarations only refer to a single property. In this case the name of the property MAY be omitted.

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocToCommentFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocToCommentFixer.php
@@ -167,11 +167,7 @@ class PhpdocToCommentFixer extends AbstractFixer
     {
         $nextIndex = $tokens->getNextMeaningfulToken($variableIndex);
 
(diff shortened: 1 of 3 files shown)


SpacePossum:
will this resolve the priority issue (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1715#issuecomment-248524932) as well or is that out of scope for this PR?

keradus:
it was raised after my PR was made ;)
it may not even be an issue - i guess we have proper priority for those 2 fixers, but yeah, integration test would be nice

keradus:
integrity check added. no priority issue

PR #2193, opened 2016-09-21
UnneededControlParenthesesFixer - Fix more return cases.

SpacePossum:
Cases like these can be fixed as well:

```php
<?php
//
return(2)?>
```

This solves a not tested priority issue on master with the `semicolon_after_instruction`


keradus:
priority issue ? smells like integration tests

SpacePossum:
>priority issue ? smells like integration tests

could do, but if this PR is merged there is no longer a need ;)

keradus:
integration tests are to protect us from priority issues we faced once.

SpacePossum:
I cannot create a priority for `semicolon_after_instruction` and `UnneededControlParenthesesFixer` on 1.12 because `semicolon_after_instruction` is introduced on master.
The fix however can be done on 1.12.
How do you want me to proceed?

keradus:
prepare integration test for master only ;)

SpacePossum:
here you go ;)

keradus:
Thanks @SpacePossum

Issue #2090, opened 2016-07-22
[Meta] PHP 7.1 features


SpacePossum:
PHP 7.1.0 Beta 1 is [released](http://php.net/archive/2016.php#id2016-07-21-1) with a [feature stop](https://www.reddit.com/r/PHP/comments/4tw8m0/php_710_beta_1_released/d5ks9e6) so we can list the things we need to take care of or could support.
- [ ] `void` return type [RFC](https://wiki.php.net/rfc/void_return_type)
- [ ] `Iterable` pseudo-type accepting any array or object implementing `Traversable` [RFC](https://wiki.php.net/rfc/iterable)
- [x] Support Class Constant Visibility [RFC](https://wiki.php.net/rfc/class_const_visibility)
- [x] Added a form of the `list()` construct where keys can be specified [RFC](https://wiki.php.net/rfc/list_keys)
- [x] Added `[] =` as alternative construct to `list() =` [RFC](https://wiki.php.net/rfc/short_list_syntax)
- [x] Added `nullable` types. [RFC](https://wiki.php.net/rfc/nullable_types)
- [x] String offset access now supports negative references, which will be counted from the end of the string [RFC](https://wiki.php.net/rfc/negative-string-offsets)
- [x] ASCII 0x7F delete control character is no longer permitted in unquoted identifiers in source code
- [x] multiple catch [RFC](https://wiki.php.net/rfc/multiple-catch)
- [x] rand() aliased to mt_rand() and srand() aliased to mt_srand()

Sources:
https://github.com/php/php-src/blob/php-7.1.0beta1/NEWS
https://github.com/php/php-src/blob/php-7.1.0beta1/UPGRADING

Thanks to all who worked on PHP7.1!

SpacePossum:
ping @dshafik it seems the support for class constant visibility is missing from;
https://github.com/php/php-src/blob/php-7.1.0beta1/UPGRADING
it is listed in https://github.com/php/php-src/blob/php-7.1.0beta1/NEWS#L373 (PHP 7.1.0alpha1).
Am I correct to assume it will still be part of 7.1?

keradus:
I Added multiple-catch  to the list

SpacePossum:
perfect, filling in the blanks :)

HeahDude:
Also, it seems that the fixer for spaces in ternary condition also adds a space in nullable type hints :(

keradus:
>  Added nullable types. RFC

[...]

PR #2196, opened 2016-09-23
PhpdocTypesFixer - support iterable type

GrahamCampbell:
PHP 7.1 will have the `iterable` type. We should treat it like the other types we normalize.

Refs:
* https://wiki.php.net/rfc/iterable
* https://github.com/php/php-src/pull/1941
* https://github.com/php/php-src/blob/php-7.1.0RC2/UPGRADING#L25-L27

### DIFF
--- a/Symfony/CS/Fixer/Symfony/PhpdocTypesFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpdocTypesFixer.php
@@ -34,6 +34,7 @@ final class PhpdocTypesFixer extends AbstractPhpdocTypesFixer
         'float',
         'int',
         'integer',
+        'iterable',
         'mixed',
         'null',
         'object',
--- a/Symfony/CS/Tests/Fixer/Symfony/PhpdocTypesFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/Symfony/PhpdocTypesFixerTest.php
@@ -91,6 +91,30 @@ EOF;
         $this->makeTest($expected, $input);
     }
 
+    public function testIterableFix()
+    {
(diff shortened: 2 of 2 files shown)


SpacePossum:
please add missing details here https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2090

keradus:
isn't it's there? 2nd position of initial comment

keradus:
Nice !

Could you please add small integration test for 7.1 (like this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/Symfony/CS/Tests/Fixtures/Integration/misc/PHP7.test) ?

I'm not saying to add all 7.1 (which of course would be nice), but at least `iterable`

GrahamCampbell:
> Could you please add small integration test for 7.1 

Why PHP 7.1. This phpdoc type can be used on any version, the same way `string` is used on PHP 5 even though it can't be type hinted until 7.0, and `mixed` isn't even a real type on any version.

keradus:
iterable is not only for typedoc.
add `function foo(iterable $bar) {}` in integration test like we did for 7.0 syntax changes.

GrahamCampbell:
This fixer change has nothing to do with actual typehints though, so in the interest of keeping this atomic, I shouldn't make an unrelated change.

keradus:
then this PR would make false sense of feeling we support the iterable, even if we would support it only in doctype. big -1 for that

keradus:
Thank you @GrahamCampbell.

SpacePossum:
awesome, thanks @GrahamCampbell !

PR #2198, opened 2016-09-24
FileCacheManager - fix exception message and undefined property

j0k3r:
It occurs when the cache can't be written.


keradus:
Great catch !
I think it's from some not well done 2.x backporting

keradus:
Thank you @j0k3r.

PR #2199, opened 2016-09-24
make fixers less risky

SpacePossum:
Input:

```php
<?php

intval();
die;
```

Output:

```php
<?php

(int) ;
die;
```

This PR output:

```php
<?php

intval();
die;
```

(i.e. no change)

This PR makes some of the function replacement fixer less risky as these will now check if the number of arguments are as expected before changing the code.

### DIFF
--- a/src/Fixer/Alias/RandomApiMigrationFixer.php
+++ b/src/Fixer/Alias/RandomApiMigrationFixer.php
@@ -22,12 +22,14 @@ use PhpCsFixer\Tokenizer\Tokens;
 final class RandomApiMigrationFixer extends AbstractFunctionReferenceFixer
 {
     /**
-     * @var string[]
+     * @var array
      */
-    private $configuration = array(
-        'rand' => 'mt_rand',
-        'srand' => 'mt_srand',
-        'getrandmax' => 'mt_getrandmax',
+    private $configuration;
+
+    private static $defaultConfiguration = array(
+        'rand' => array('alternativeName' => 'mt_rand', 'argumentCount' => array(0, 2)),
+        'srand' => array('alternativeName' => 'mt_srand', 'argumentCount' => array(0, 1)),
+        'getrandmax' => array('alternativeName' => 'mt_getrandmax', 'argumentCount' => array(0)),
     );
 
     /**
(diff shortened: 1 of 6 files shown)


src/Fixer/Alias/RandomApiMigrationFixer.php
```diff
@@ -21,14 +21,16 @@
  */
 final class RandomApiMigrationFixer extends AbstractFunctionReferenceFixer
 {
+    public static $defaultConfig = array(
```
keradus: why public?
why static?

src/Fixer/Alias/RandomApiMigrationFixer.php
```diff
@@ -38,16 +40,24 @@ public function configure(array $configuration = null)
         static $validFunctions = array('rand', 'srand', 'getrandmax');
```
keradus: why end user is supposed to provide number of arguments ?
we allows to replace only those 3 functions, and they have always same number of arguments. asking user to fill them is pointless. in same time, asking user for more info changing the configure array structure, so BC problem...

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -58,14 +58,20 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
                 // analysing cursor shift
                 $currIndex = $openParenthesis;
 
+                $parametersCount = $this->countArguments($tokens, $openParenthesis, $closeParenthesis);
+
                 // special case: intval with 2 parameters shall not be processed (base conversion)
-                if ('intval' === $functionIdentity) {
-                    $parametersCount = $this->countArguments($tokens, $openParenthesis, $closeParenthesis);
+                if ('intval' === $functionIdentity && 2 === $parametersCount) {
                     if ($parametersCount > 1) {
```
keradus: `$parametersCount === 2 && $parametersCount > 1`
... ? :|

src/Fixer/Alias/RandomApiMigrationFixer.php
SpacePossum: there is BC problem on master right?

keradus: again:

> why end user is supposed to provide number of arguments ?

----

> there is BC problem on master right?

OK, no issue as it was not on 1.x, nothing to describe in update guide

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
SpacePossum: fixed

src/Fixer/Alias/RandomApiMigrationFixer.php
SpacePossum: let me think about for a while

keradus: don't change the input structure and keep number of indexes in separated structure

keradus:
Thank you @SpacePossum.

Issue #2028, opened 2016-06-30
Fixer for missing root namespaces in @expectedException


ro0NL:
From https://github.com/symfony/symfony/issues/19237

> I noticed this at the workflow component and decided to do a quick scan
> 
> Im talking about `@expectedException \Exception` vs. `@expectedException Exception`. Without `use` statements the latter works for phpunit but breaks editor click-throughs, at least phpstorm.
> 
> Searched with `@expectedException\s+[^\\\s]` with 34 results on master (not checked for `use` statements). Perhaps something to automate?
> 
> Not sure about [this](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Tests/Controller/ControllerNameParserTest.php#L111) one :-)

I would be nice if we can force consistent behaviour, i.e. convert everything to use statements, _or_ everything to FQCN's without use statements.

SpacePossum:
I would like this a lot :)
This fixer should be marked as risky I think, for example:
```php
    /**
     * @expectedException UnexpectedValueException
     */
    public function testRegisterFixerWithOccupiedName()
    {
        a();
    }
```

when fixed to:

```php
    /**
     * @expectedException \UnexpectedValueException
     */
    public function testRegisterFixerWithOccupiedName()
    {
        a();
    }
```

if `a()` would throw;
```php
namespace My\Poor\Named\Exception;

class UnexpectedValueException extends \Exception
{
}
```

then the behavior is changed.

ro0NL:
Well... if `a()` throws `\My\Poor\Named\Exception\UnexpectedValueException` it should be 

`@expectedException \My\Poor\Named\Exception\UnexpectedValueException`

**or**

```
use My\Poor\Named\Exception\UnexpectedValueException
@expectedException UnexpectedValueException
```

GrahamCampbell:
PHPUnit does not read impors.

```
use My\Poor\Named\Exception\UnexpectedValueException
@expectedException UnexpectedValueException
```

That will expect `UnexpectedValueException` not `My\Poor\Named\Exception\UnexpectedValueException`.

SpacePossum:
indeed
```php
<?php
namespace My\Poor\Named\Exception;

class InvalidArgumentException extends \Exception
{

}

class MyTest extends \PHPUnit_Framework_TestCase
{
    /**
     * @expectedException InvalidArgumentException
     */
    public function testException()
    {
        a();
    }
}

[...]

PR #2200, opened 2016-09-24
Added fixer for PHPUnit's @expectedException annotation

ro0NL:
See https://github.com/symfony/symfony/pull/20043 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2028

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -433,6 +433,10 @@ Choose from the list of available fixers:
    PHPUnit assertions like "assertInternalType", "assertFileExists", should
    be used over "assertTrue". (Risky fixer!)
 
+* **php_unit_expected_exception_annotation_with_root_namespace** [@Symfony]
+   PHPUnit @expectedException annotation should be a FQCN including a root
+   namespace.
+
 * **php_unit_strict**
    PHPUnit methods like "assertSame" should be used instead of
    "assertEquals". (Risky fixer!)
--- a/src/Fixer/PhpUnit/PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer.php
@@ -1,0 +1,61 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Roland Franssen <franssen.roland@gmail.com>
+ */
+final class PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return 'PHPUnit @expectedException annotation should be a FQCN including a root namespace.';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before NoUnusedImportsFixer
+        return -9;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $token->setContent(preg_replace('~^(\s*\*\s*@expectedException\h+)(\w.*)$~m', '$1\\\\$2', $token->getContent()));
+        }
+    }
+}
--- a/tests/Fixtures/Integration/priority/php_unit_expected_exception_annotation_with_root_namespace,no_unused_imports.test
+++ b/tests/Fixtures/Integration/priority/php_unit_expected_exception_annotation_with_root_namespace,no_unused_imports.test
@@ -1,0 +1,28 @@
+--TEST--
+Integration of fixers: php_unit_expected_exception_annotation_with_root_namespace,no_unused_imports.
+--CONFIG--
+{"php_unit_expected_exception_annotation_with_root_namespace": true, "no_unused_imports": true}
+--EXPECT--
+<?php
+
+class Foo extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @expectedException \Exception
+     */
+    public function testBar()
+    { }
+}
+
+--INPUT--
+<?php
+use Some\Exception;
+
+class Foo extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @expectedException Exception
+     */
+    public function testBar()
+    { }
+}
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -121,6 +121,7 @@ final class RuleSet implements RuleSetInterface
             'unalign_equals' => true,
             'unary_operator_spaces' => true,
             'whitespace_after_comma_in_array' => true,
+            'php_unit_expected_exception_annotation_with_root_namespace' => true,
         ),
         '@Symfony:risky' => array(
             'php_unit_construct' => true,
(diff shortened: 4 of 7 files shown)


SpacePossum:
LGTM :)

Not sure about the name as it is kind of vague IMHO (i.e. does this fixer remove multiple `@` like `@@Expec..`, does it fix the casing of the tag like `@ExpectedException`? I know it doesn't and shouldn't ;)) and all the classnames of the PHPDoc fixers are prefixed with `Phpdoc` ATM.

ro0NL:
What about `PhpUnit\PhpdocExpectedExceptionToFqcnFixer`

SpacePossum:
nice one :)
I expect the tests to fail because you need to re-gen the readme though ;)

phansys:
This should run before `no_unused_imports`, WDYT?

ro0NL:
About the readme.. im not sure but the diff is pretty huge on my machine when running
`php-cs-fixer readme > README.rst`

```diff
 * **align_double_arrow**
-                         Align double arrow symbols in
-                         consecutive lines.
+                           Align double arrow symbols
+                           in consecutive lines.

etc.
```

@phansys yes..

SpacePossum:
the big diff is because the name you picked for the fixer is longer than the current longest so the indent changes :)
no worries, re-gen and commit ^-^

phansys:
The diff is fine for sure @ro0NL, since the command re generates the whole file (re aligning all fixer's descriptions by instance) according the changes you made. You can test it ensuring it's fine in your local environment by running `phpunit`.

ro0NL:
Should we call this `PhpUnitPhpdocExpectedExceptionWithRootNamespaceFixer`? :sweat_smile:

src/Fixer/PhpUnit/PhpdocExpectedExceptionWithRootNamespaceFixer.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Roland Franssen <franssen.roland@gmail.com>
+ */
+class PhpdocExpectedExceptionWithRootNamespaceFixer extends AbstractFixer
```
keradus: final

src/Fixer/PhpUnit/PhpdocExpectedExceptionWithRootNamespaceFixer.php
```diff
@@ -0,0 +1,54 @@
...
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $token->setContent(preg_replace_callback('~^(\s*\*\s*@expectedException\s+)([^\s\\\\].*)$~m', function (array $matches) {
```
keradus: why callback version ?
why set content when regexp won't change anything ?

keradus:
> This should run before `no_unused_imports`, WDYT?

then the integration test is missing

src/Fixer/PhpUnit/PhpdocExpectedExceptionWithRootNamespaceFixer.php
ro0NL: This is simple string normalization.. what do you propose?

ro0NL: Im not fully aware of inner workings, but i noticed calling `setContent` is marking tokens as changed. so yes, lets avoid that when content is not changed :+1:

ro0NL:
The readme breaks at https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2200/files#diff-88b99bb28683bd5b7e3a204826ead112R197 

see https://github.com/ro0NL/PHP-CS-Fixer/blob/4cfee8b13364bc130ef80c32bd508b79641b37a1/README.rst

src/Fixer/PhpUnit/PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        foreach ($tokens as $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $content = $token->getContent();
+            $newContent = preg_replace('~^(\s*\*\s*@expectedException\s+)([^\s\\\\].*)$~m', '$1\\\\$2', $content);
+
+            if ($newContent !== $content) {
```
SpacePossum: check is not needed,
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Tokenizer/Token.php#L479

tests/Fixtures/Integration/priority/php_unit_expected_exception_annotation_with_root_namespace,no_unused_imports.test
```diff
@@ -0,0 +1,28 @@
...
+use Some\Exception;
+
+class Foo extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @expectedException Exception
+     */
+    public function testBar()
+    { }
+}
```
SpacePossum: please add the prio test to this list/test as well;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/FixerFactoryTest.php#L193

src/Fixer/PhpUnit/PhpdocExpectedExceptionWithRootNamespaceFixer.php
```diff
@@ -31,6 +31,15 @@ public function getDescription()
     /**
      * {@inheritdoc}
      */
+    public function getPriority()
+    {
+        // should be run before NoUnusedImportsFixer
+        return -9;
```
keradus: add test here as well: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/FixerFactoryTest.php#L278

ro0NL:
Ready? :) Do we need a 1.12 backport for symfony?

README.rst
```diff
@@ -175,657 +175,1079 @@ automatically fix anything:
...
+                                        MUST be
+                                        enclosed by
+                                        braces. Braces
+                                        should be
+                                        properly
+                                        placed. Body
+                                        of braces
+                                        should be
+                                        properly
+                                        indented.
```
keradus: i dont like putting one word per statement...
sth to be resolved in new PR.

nothing to worry about for you @ro0NL ;)

keradus:
add to Symfony ruleset (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/RuleSet.php#L54), also update Symfony integration test to cover this new fixer as well

keradus:
https://github.com/ro0NL/PHP-CS-Fixer/pull/1

phansys:
Why not to use the naming defined at the [PHP's standard resolution rules](http://php.net/manual/en/language.namespaces.rules.php) instead?

ro0NL:
@phansys because it doesnt apply to `@expectedException`, it's intended to be a FQCN. However i agree this is often unclear and causes false positives like https://github.com/symfony/symfony/pull/20028/files#diff-63ff7c7532d7434f326bdd6dfd265f35

The missing root namespace is an editor problem (PHPStorm doesnt recognize it as a known class), the assertion itself works as phpunit doesnt care (it assumes FGCN anyway).

Normalizing to FQCN with root namespace shows the wrong use statements (due `no_unused_imports`) which should make things clear enough..

keradus:
Please rebase.

Do you want to prepare version for 1.x line ?

keradus:
yes, ubot runs 1.x

> I initially targeted master because of "new feature"

Why you didn't read this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.12/CONTRIBUTING.md#quick-guide ?

If you care about having the fixer released soon you should target 1.x, if not then 2.x is OKish

ro0NL:
Aargh, my bad :) ill try to rebase 1.12 later.. until then 2.x is ok.

phansys:
@ro0NL, sorry, maybe I've not explained my point in a clear way regarding the name. What about to call it `PhpUnit\PhpdocExpectedExceptionFqcnFixer`? I'm against the concept of "root" here, since it isn't used at as part of the namespace name definitions.

ro0NL:
@phansys agree. Should we still include "annotation" in the name.. so it's clear this is not about `setExpectedException()`.

On 2nd thought, _should_ this be about `setExpectedException()`? We could also fix/normalize calls like `setExpectedException('Some\Exception');` to `'\Some\Exception'` for the sake of consistency.

keradus:
@ro0NL Please put whole work in one line (1.x ?) and freeze the second one (2.x ?)

ro0NL:
Closing in favor of #2208.

PR #2208, opened 2016-09-25
Added fixer for PHPUnit's @expectedException annotation

ro0NL:
In favor of #2200
- [x] Rename to `PhpUnitFqcnAnnotationFixer`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -349,6 +349,10 @@ Choose from the list of available fixers:
 * **operators_spaces** [symfony]
    Binary operators should be surrounded by at least one space.
 
+* **php_unit_fqcn_annotation** [symfony]
+   PHPUnit @expectedException annotation should be a FQCN including a root
+   namespace.
+
 * **phpdoc_annotation_without_dot** [symfony]
    Phpdocs annotation descriptions should not end with a full stop.
 
--- a/Symfony/CS/Fixer/Symfony/PhpUnitFqcnAnnotationFixer.php
+++ b/Symfony/CS/Fixer/Symfony/PhpUnitFqcnAnnotationFixer.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Fixer\Symfony;
+
+use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
+
+/**
+ * @author Roland Franssen <franssen.roland@gmail.com>
+ */
+final class PhpUnitFqcnAnnotationFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        foreach ($tokens as $token) {
+            if ($token->isGivenKind(T_DOC_COMMENT)) {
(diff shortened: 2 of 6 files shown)


Symfony/CS/Fixer/Symfony/PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer.php
```diff
@@ -0,0 +1,53 @@
...
+final class PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, $content)
+    {
+        $tokens = Tokens::fromCode($content);
+
+        foreach ($tokens->findGivenKind(T_DOC_COMMENT) as $token) {
```
keradus: don't use find for iterating. it's slower than regular loop.

ro0NL: Revised.

keradus:
>  Only fix classes extending \PHPUnit_Framework_TestCase

Nope, we don't do that assumption in any of our PHPUnit fixer.
`PHPUnit_Framework_TestCase` could not be used directly but custom child of it is used.
Eg we are using `PhpCsFixer\Test\AbstractFixerTestCase` in most of our tests files.

ro0NL:
`is_subclass_of`? I mean.. we shouldnt just update every call to `setExpectedException`... right?

keradus:
When you fix a code you have only file-level context.

If given class is defined outside of file, you have no idea what it is. It may be even defined in different repo

ro0NL:
Yeah.. but we should verify the class/file we're updating is actually a test class/file or am i missing something..

SpacePossum:
>Normalize setExpectedException('Some\Exception'); to '\Some\Exception'

I would do that in another fixer, it use will be limitted because of: https://github.com/sebastianbergmann/phpunit/issues/2074#issuecomment-183472385 (i.e. it will be removed from PHPUnit itself)

> Only fix sub classes from \PHPUnit_Framework_TestCase

I agree with @keradus here, there is no way solid figuring out if a class (or some include / trait) is part of a `PHPUnit_Framework_TestCase`-class inherence by looking at the tokens of that single file.

> Rename to PhpUnitExpectedExceptionFixer

only if your fixer also fixes the casing ;) (I would like that feature btw., but still think it is out of scope)

keradus:
> Rename to `PhpUnitExpectedExceptionFixer`

as we don't modifying `->expectException` I would name it `PhpUnitFqcnAnnotationFixer`.
changing string passed to method is different thing and it should be separated fixer/PR.

ro0NL:
:+1: lets keep things simple.

Symfony/CS/Tests/Fixer/Symfony/PhpUnitExpectedExceptionAnnotationWithRootNamespaceFixerTest.php
```diff
@@ -0,0 +1,51 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace Symfony\CS\Tests\Fixer\Symfony;
+
+use Symfony\CS\Tests\Fixer\AbstractFixerTestBase;
+
+/**
+ * @author Roland Franssen <franssen.roland@gmail.com>
```
keradus: add @internal annotation

keradus:
raname and ready to merge ?

ro0NL:
Worst idea ever.. but i think it's ok :)

keradus:
> Worst idea ever..

the name? it was just a proposal to have it short and not connected to function call with same name that annotation.
propose better one ;)

ro0NL:
I meant renaming via github :smile: 

This is ready :tada:

keradus:
:+1:

SpacePossum:
:+1: :ship: thanks @ro0NL

keradus:
Thank you @ro0NL.

PR #2209, opened 2016-09-27
LinefeedFixer - Fix in a safe way

SpacePossum:
The fixer should not touch all line breaks.

### DIFF
--- a/Symfony/CS/Fixer/PSR2/LinefeedFixer.php
+++ b/Symfony/CS/Fixer/PSR2/LinefeedFixer.php
@@ -13,11 +13,13 @@
 namespace Symfony\CS\Fixer\PSR2;
 
 use Symfony\CS\AbstractFixer;
+use Symfony\CS\Tokenizer\Tokens;
 
 /**
  * Fixer for rules defined in PSR2 ¶2.2.
  *
  * @author Fabien Potencier <fabien@symfony.com>
+ * @author SpacePossum
  */
 class LinefeedFixer extends AbstractFixer
 {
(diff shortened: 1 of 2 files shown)


keradus:
then we need a second fixer (risky one) that will fix all the rest...

SpacePossum:
>then we need a second fixer (risky one) that will fix all the rest...

I think this is a bug fix, so do not need such a fixer for BC reasons. Do you think someone really wants it?

gharlan:
> Do you think someone really wants it?

I know at least one: me!

keradus:
me too .actually most of community. the only thing is that it's risky.

we have exactly the same problem with trailing spaces. PSR requires to not have them in non-empty lines - even if it's part of string var. but then it modifies a content, so it's risky. but still, most of us wanna follow that.

keradus:
It's bugfix but also shrinking functionality. Now one cannot fix linefeed inside var content. could you prepare the new fixer as well please? it's really important for us ! :)

SpacePossum:
alrighty, would a config flag work as well or you really want a diff. fixer?

keradus:
I was thinking about the config as well, but then you need to dynamically say if the fixer is risky or not depends on config.

SpacePossum:
I'm kind of happy with this PR as is, but if we need another fixer please provide names for both fixers and I'll update this PR :)

keradus:
as discussed it could be an option. the only thing to keep in mind is risky vs not risky... depends on used options

keradus:
OK, I treat it as bugfix after thinking about it. Missing functionality is raised here: #2224

keradus:
:+1:

keradus:
Thank you @SpacePossum.

PR #2217, opened 2016-10-10
SelfUpdateCommand - Print version of update fixer

SpacePossum:
If the fixer is updated by the command print out the version it was updated to.


keradus:
Thank you @SpacePossum.

Issue #2195, opened 2016-09-22
[RFC] var_dump/dump fixer


akovalyov:
This RFC is for var_dump/dump removal fixer. While a lot of devs debug via var_dump(or via  Symfony/VarDumper dump), it is dangerous to have it in code, especially if it appears in production . So, why not to introduce these functions removal? What do you think?

SpacePossum:
Hi and thanks for your suggestion :)

How about `var_export` or the `CURLOPT_VERBOSE` flag?
The thing is that we would not be fixing code style, but modifying the way some code works. This is not always a bad thing IMO. However in this case we cannot determine if the `var_dump` is indeed a forgotten debug statement (and the function is not overloaded) or is part of the logic that is essential to the application. 
Therefor I would not add it to the core of the application because it is out of scope and risky. However such a fixer is relatively easy to write and can exists as a custom fixer for your needs.
But lets see what the community thinks :)

akovalyov:
@SpacePossum thank you for your feedback.

 I agree with you that such a functionality might be out of scope of the project. 

Right now I can't judge what is more risky for developer - keeping statements in the code, which are commonly used for debugging, or use the fixer as such. 

However, I think that contrib fixers are usually enabled explicitly by developer, which means that one understands all the consequences of it. 

Anyway, the reason why I've created the issue is that I've already done this fixer for private purposes. If community wants to have that, I will be more than happy to contribute to an awesome tool. If there is no need/it is our of scope of the project, I will just use my fork.

buffcode:
Also consider that those debug statements, especially `var_export($var, true)` could be used to save debug data / traces in production whenever an unexpected error occurs.

akovalyov:
Short summary from PR discussion - too risky to have it in core. For those, who want to have that - https://github.com/akovalyov/DebugStatementsFixers.

PR #2218, opened 2016-10-12
Dump fixer

akovalyov:
Resolution for #2195

### DIFF
--- a/Symfony/CS/Fixer/Contrib/RemoveVarDumpFixer.php
+++ b/Symfony/CS/Fixer/Contrib/RemoveVarDumpFixer.php
@@ -32,12 +32,9 @@ final class RemoveVarDumpFixer extends AbstractFixer
         $end = $tokens->count() - 1;
 
         foreach (['dump', 'var_dump'] as $function) {
-            // the sequence is the function name, followed by "(" and a quoted string
-            $seq = array(array(T_STRING, $function), '(');
-
             $currIndex = 0;
             while (null !== $currIndex) {
-                $match = $tokens->findSequence($seq, $currIndex, $end, false);
+                $match = $tokens->findSequence(array(array(T_STRING, $function), '('), $currIndex, $end, false);
 
                 // did we find a match?
                 if (null === $match) {
@@ -45,10 +42,9 @@ final class RemoveVarDumpFixer extends AbstractFixer
                 }
 
                 $match = array_keys($match);
-                $currIndex = $match[1];
 
-                $funcStart = $tokens->getPrevTokenOfKind($currIndex, array(';'));
-                $funcEnd = $tokens->getNextTokenOfKind($currIndex, array(';'));
+                $funcStart = $tokens->getPrevTokenOfKind($match[1], array(';'));
+                $funcEnd = $tokens->getNextTokenOfKind($match[1], array(';'));
                 for ($i = $funcStart + 1; $i <= $funcEnd; ++$i) {
                     $tokens[$i]->clear();
                 }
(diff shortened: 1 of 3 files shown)


keradus:
1. risky
2. `var_dump` with 2nd parameter should not be removed

SpacePossum:
```
\a\b\c\dump($a);
```

should not be removed either. 

There is some nice stuff in the `AbstractFunctionReferenceFixer` you could you use and make it more easy to do what you want :)

GrahamCampbell:
> Dump fixer

Best name ever. 😆

Symfony/CS/Fixer/Contrib/RemoveVarDumpFixer.php
```diff
@@ -0,0 +1,81 @@
...
+
+                if (null === $matches) {
+                    break;
+                }
+                $match = array_keys($matches);
+
+                $funcStart = $tokens->getPrevNonWhitespace($match[0]);
+
+                $funcEnd = $tokens->getNextTokenOfKind($match[1], array(';'));
+                for ($i = $funcStart + 1; $i <= $funcEnd; ++$i) {
```
SpacePossum: There is a shorthand for this;

```php
$tokens->clearRange()
```

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Tokenizer/Tokens.php#L1041

Symfony/CS/Fixer/Contrib/RemoveVarDumpFixer.php
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Quick search for dump/var_dump existence.
+     *
+     * @param string $content
+     *
+     * @return bool
+     */
+    private function hasDump($content)
+    {
+        return false !== stripos($content, 'dump');
```
SpacePossum: this will make the fixing case sensitive (i.e. not remove `Var_dump()`),
removing this and using `AbstractFunctionReferenceFixer` will prevent this

akovalyov:
@keradus @SpacePossum sorry for being out of this PR for some time, but now I have a little bit of it to finalize it. But first I would like to hear your opinion - do you think it is needed in core and is there a chance for it to be merged? If yes - then I will proceed with comments resolution and any suggestions/comments which may raise, otherwise let's just close it. Looking forward to hear your thoughts on it!

SpacePossum:
IMO it is a very risky fixer going beyond the scope of the tool, i.e. fix code style and optimize code where possible. However the acceptable 'risky level' is of course opinionated.
But lets here from the community!

GrahamCampbell:
Yeh, this is very risky, because function calls affect the program state in PHP. Removing them could cause serious problems.

GrahamCampbell:
Not to mention, that statements can be used as well as just expressions, so even if php was pure, there could still be assignments within the arguments, affecting the state.

GrahamCampbell:
I'm 👎 on having this accepted as a cs fixer in here.

akovalyov:
I agree with you, that this too risky and  is most likely out of scope of this awesome tool. 
I've extracted it in small lib - https://github.com/akovalyov/DebugStatementsFixers. Those, who are interested, can always use that. 
Thanks for your feedback and time!

curry684:
As this issue is still high on the list when Googling for 'how to prohibit accidentally committing dump statements` - there is an awesome PHPStan plugin that allows you to define forbidden functions and constructs, and is out of the box default configured to do exactly what we want here: https://github.com/ekino/phpstan-banned-code
```bash
composer require --dev ekino/phpstan-banned-code
```

kubawerlos:
@curry684 you can even automate removing it with [CommentedOutFunctionFixer](https://github.com/kubawerlos/php-cs-fixer-custom-fixers#CommentedOutFunctionFixer) and [NoCommentedOutCodeFixer](https://github.com/kubawerlos/php-cs-fixer-custom-fixers#NoCommentedOutCodeFixer).

PR #2220, opened 2016-10-12
Doctrine annotation fixers

julienfalque:
Various Doctrine annotations related fixers. Any thoughts?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -302,6 +302,97 @@ Choose from the list of available rules:
 
   *Risky rule: risky when the function ``dirname()`` is overridden.*
 
+* **doctrine_annotation_braces**
+
+  Doctrine annotations without arguments must use the configured syntax.
+
+  Configuration options:
+
+  - ``ignored_tags`` (``array``): list of tags that must not be treated as Doctrine
+    Annotations; defaults to ``['abstract', 'access', 'code', 'deprec',
+    'encode', 'exception', 'final', 'ingroup', 'inheritdoc', 'inheritDoc',
+    'magic', 'name', 'toc', 'tutorial', 'private', 'static', 'staticvar',
+    'staticVar', 'throw', 'api', 'author', 'category', 'copyright',
+    'deprecated', 'example', 'filesource', 'global', 'ignore', 'internal',
+    'license', 'link', 'method', 'package', 'param', 'property',
+    'property-read', 'property-write', 'return', 'see', 'since', 'source',
+    'subpackage', 'throws', 'todo', 'TODO', 'usedBy', 'uses', 'var',
+    'version', 'after', 'afterClass', 'backupGlobals',
+    'backupStaticAttributes', 'before', 'beforeClass',
+    'codeCoverageIgnore', 'codeCoverageIgnoreStart',
+    'codeCoverageIgnoreEnd', 'covers', 'coversDefaultClass',
+    'coversNothing', 'dataProvider', 'depends', 'expectedException',
+    'expectedExceptionCode', 'expectedExceptionMessage',
+    'expectedExceptionMessageRegExp', 'group', 'large', 'medium',
+    'preserveGlobalState', 'requires', 'runTestsInSeparateProcesses',
+    'runInSeparateProcess', 'small', 'test', 'testdox', 'ticket', 'uses',
+    'SuppressWarnings', 'noinspection', 'package_version', 'enduml',
+    'startuml', 'fix', 'FIXME', 'fixme', 'override']``
+  - ``syntax`` (``'with_braces'``, ``'without_braces'``): whether to add or remove
+    braces; defaults to ``'without_braces'``
+
+* **doctrine_annotation_indentation**
+
+  Doctrine annotations must be indented with four spaces.
+
+  Configuration options:
+
+  - ``ignored_tags`` (``array``): list of tags that must not be treated as Doctrine
+    Annotations; defaults to ``['abstract', 'access', 'code', 'deprec',
+    'encode', 'exception', 'final', 'ingroup', 'inheritdoc', 'inheritDoc',
+    'magic', 'name', 'toc', 'tutorial', 'private', 'static', 'staticvar',
+    'staticVar', 'throw', 'api', 'author', 'category', 'copyright',
+    'deprecated', 'example', 'filesource', 'global', 'ignore', 'internal',
+    'license', 'link', 'method', 'package', 'param', 'property',
+    'property-read', 'property-write', 'return', 'see', 'since', 'source',
+    'subpackage', 'throws', 'todo', 'TODO', 'usedBy', 'uses', 'var',
+    'version', 'after', 'afterClass', 'backupGlobals',
+    'backupStaticAttributes', 'before', 'beforeClass',
+    'codeCoverageIgnore', 'codeCoverageIgnoreStart',
+    'codeCoverageIgnoreEnd', 'covers', 'coversDefaultClass',
+    'coversNothing', 'dataProvider', 'depends', 'expectedException',
+    'expectedExceptionCode', 'expectedExceptionMessage',
+    'expectedExceptionMessageRegExp', 'group', 'large', 'medium',
+    'preserveGlobalState', 'requires', 'runTestsInSeparateProcesses',
+    'runInSeparateProcess', 'small', 'test', 'testdox', 'ticket', 'uses',
+    'SuppressWarnings', 'noinspection', 'package_version', 'enduml',
+    'startuml', 'fix', 'FIXME', 'fixme', 'override']``
+
+* **doctrine_annotation_spaces**
+
+  Fixes spaces in Doctrine annotations.
+
+  Configuration options:
+
+  - ``around_argument_assignments`` (``bool``): whether to fix spaces around
+    argument assignment operator; defaults to ``true``
+  - ``around_array_assignments`` (``bool``): whether to fix spaces around array
+    assignment operators; defaults to ``true``
+  - ``around_commas`` (``bool``): whether to fix spaces around commas; defaults to
+    ``true``
+  - ``around_parentheses`` (``bool``): whether to fix spaces around parentheses;
+    defaults to ``true``
+  - ``ignored_tags`` (``array``): list of tags that must not be treated as Doctrine
+    Annotations; defaults to ``['abstract', 'access', 'code', 'deprec',
+    'encode', 'exception', 'final', 'ingroup', 'inheritdoc', 'inheritDoc',
+    'magic', 'name', 'toc', 'tutorial', 'private', 'static', 'staticvar',
+    'staticVar', 'throw', 'api', 'author', 'category', 'copyright',
+    'deprecated', 'example', 'filesource', 'global', 'ignore', 'internal',
+    'license', 'link', 'method', 'package', 'param', 'property',
+    'property-read', 'property-write', 'return', 'see', 'since', 'source',
+    'subpackage', 'throws', 'todo', 'TODO', 'usedBy', 'uses', 'var',
+    'version', 'after', 'afterClass', 'backupGlobals',
+    'backupStaticAttributes', 'before', 'beforeClass',
+    'codeCoverageIgnore', 'codeCoverageIgnoreStart',
+    'codeCoverageIgnoreEnd', 'covers', 'coversDefaultClass',
+    'coversNothing', 'dataProvider', 'depends', 'expectedException',
+    'expectedExceptionCode', 'expectedExceptionMessage',
+    'expectedExceptionMessageRegExp', 'group', 'large', 'medium',
+    'preserveGlobalState', 'requires', 'runTestsInSeparateProcesses',
+    'runInSeparateProcess', 'small', 'test', 'testdox', 'ticket', 'uses',
+    'SuppressWarnings', 'noinspection', 'package_version', 'enduml',
+    'startuml', 'fix', 'FIXME', 'fixme', 'override']``
+
 * **elseif** [@PSR2, @Symfony]
 
   The keyword ``elseif`` should be used instead of ``else if`` so that all
--- a/composer.json
+++ b/composer.json
@@ -15,6 +15,7 @@
     ],
     "require": {
         "php": "^5.3.6 || >=7.0 <7.2",
+        "doctrine/annotations": "^1.2",
         "ext-tokenizer": "*",
         "sebastian/diff": "^1.1",
         "symfony/console": "^2.3 || ^3.0",
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -1,0 +1,217 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+use PhpCsFixer\Doctrine\Annotation\Tokens as DoctrineAnnotationTokens;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\Tokenizer\Token as PhpToken;
+use PhpCsFixer\Tokenizer\Tokens as PhpTokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * Base class for Doctrine annotation fixers.
+ *
+ * @internal
+ */
+abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, PhpTokens $phpTokens)
+    {
+        /** @var PhpToken $docCommentToken */
+        foreach ($phpTokens->findGivenKind(T_DOC_COMMENT) as $index => $docCommentToken) {
+            if (!$this->nextElementAcceptsDoctrineAnnotations($phpTokens, $index)) {
+                continue;
+            }
+
+            $tokens = DoctrineAnnotationTokens::createFromDocComment(
+                $docCommentToken,
+                $this->configuration['ignored_tags']
+            );
+            $this->fixAnnotations($tokens);
+            $docCommentToken->setContent($tokens->getCode());
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(PhpTokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * Fixes Doctrine annotations from the given PHPDoc style comment.
+     *
+     * @param DoctrineAnnotationTokens $doctrineAnnotationTokens
+     */
+    abstract protected function fixAnnotations(DoctrineAnnotationTokens $doctrineAnnotationTokens);
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        $ignoredTags = new FixerOptionBuilder('ignored_tags', 'List of tags that must not be treated as Doctrine Annotations.');
+        $ignoredTags = $ignoredTags
+            ->setAllowedTypes(array('array'))
+            ->setAllowedValues(array(function ($values) {
+                foreach ($values as $value) {
+                    if (!is_string($value)) {
+                        return false;
+                    }
+                }
+
+                return true;
+            }))
+            ->setDefault(array(
+                 // PHPDocumentor 1
+                 'abstract',
+                 'access',
+                 'code',
+                 'deprec',
+                 'encode',
+                 'exception',
+                 'final',
+                 'ingroup',
+                 'inheritdoc',
+                 'inheritDoc',
+                 'magic',
+                 'name',
+                 'toc',
+                 'tutorial',
+                 'private',
+                 'static',
+                 'staticvar',
+                 'staticVar',
+                 'throw',
+
+                 // PHPDocumentor 2
+                 'api',
+                 'author',
+                 'category',
+                 'copyright',
+                 'deprecated',
+                 'example',
+                 'filesource',
+                 'global',
+                 'ignore',
+                 'internal',
+                 'license',
+                 'link',
+                 'method',
+                 'package',
+                 'param',
+                 'property',
+                 'property-read',
+                 'property-write',
+                 'return',
+                 'see',
+                 'since',
+                 'source',
+                 'subpackage',
+                 'throws',
+                 'todo',
+                 'TODO',
+                 'usedBy',
+                 'uses',
+                 'var',
+                 'version',
+
+                 // PHPUnit
+                 'after',
+                 'afterClass',
+                 'backupGlobals',
+                 'backupStaticAttributes',
+                 'before',
+                 'beforeClass',
+                 'codeCoverageIgnore',
+                 'codeCoverageIgnoreStart',
+                 'codeCoverageIgnoreEnd',
+                 'covers',
+                 'coversDefaultClass',
+                 'coversNothing',
+                 'dataProvider',
+                 'depends',
+                 'expectedException',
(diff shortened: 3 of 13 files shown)


localheinz:
@julienfalque 

If this works, that would kick butt! 👍

Symfony/CS/Doctrine/Annotation/Tokens.php
```diff
@@ -0,0 +1,342 @@
...
+            return false;
+        }
+
+        $identifier = $this[$index + 1]->getContent();
+        $tag = new Tag('@'.$identifier);
+        if ($tag->valid()) {
+            return false;
+        }
+
+        $phpUnitAnnotations = array(
```
keradus: I would rather list all valid doctrine's annotations instead of phpunit ones...

julienfalque: Unfortunately there is no exhaustive list of valid Doctrine annotations because one can create custom annotations, see [Doctrine Annotation documentation](http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html#annotation-classes).

Maybe this should be an option with the current list as a sensible default value?

keradus:
anyone ?

GrahamCampbell:
Probably too may fixers. Would be better to make use of config.

julienfalque:
@GrahamCampbell what do you mean by _use of config_?

GrahamCampbell:
> what do you mean by use of config?

Fixers can be "configurable". See the array align fixers on 2.x line for example.

SpacePossum:
A lot of awesome work, thanks @julienfalque
Having some many things in one PR makes it hard to review. I'm _not_ stating to split this up, but it a lot to process. You might want to follow up on @GrahamCampbell suggestion, checking the master line which  has (way better) support for configuration of fixers. One candidate would be to combine `doctrine_annotation_with_braces` and `doctrine_annotation_without_braces`. The reason would be that one would always choose one of those, but never both (i.e. those are conflicting fixers). It would be better to have fixer that you can configure, like: `doctrine_annotation_braces` fixer with config for either `with` or `without` braces.

julienfalque:
The patch was originally submitted for 1.12, I didn't know that fixers were way more configurable on master, this is pretty cool :)

I was wondering if it would be better to have a single `DoctrineAnnotationFixer` with several options like `fix_arguments_spaces`, `fix_assignment_spaces` and so on, instead of these separate fixers. WDYT?

SpacePossum:
I can see why you picked the 1.x line, that's how we described the contribution flow so :+1: 

Looking at the current state of the project I _think_ you PR will be processed faster on the master line, because the focus seems:
- answer question from the community/end-users
- bug fixes on 1.x line (especially if those bugs are also on the 2.x line)
- wrap up 2.x release
- look at new features in PR's

This is by no way official, but to me it looks like the current state. This is still a lot of work and not that many people working on it so please be patient with us :)

I would unofficially advice you to create a branch from master and add the fixers on there. After that merge the `doctrine_annotation_with_braces` and `doctrine_annotation_without_braces`. I'm not sure about merging the other atm, but happy to look at those later. If you don't want to take your changes with doing that work I can totally see that.

keradus:
> I would unofficially advice you to create a branch from master and add the fixers on there. 

This PR is already targeting `master` branch, I think it was even rebased ;)

SpacePossum:
>This PR is already targeting master branch

I need coffee and stand corrected ;)

julienfalque:
I merged `doctrine_annotation_with_braces` and `doctrine_annotation_without_braces` fixers into a single, configurable `doctrine_annotation_braces` fixer and ignored tags are now configurable via the fixers instead of a static method. I also rebased the branch onto `master` to include all the (impressive) work done there :)

Should I merge more fixers?

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,276 @@
...
+use PhpCsFixer\Tokenizer\Token as PhpToken;
+use PhpCsFixer\Tokenizer\Tokens as PhpTokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * Base class for Doctrine annotation fixers.
+ */
+abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements ConfigurableFixerInterface
+{
+    private $ignoredTags;
```
keradus: and the type is... ?

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,276 @@
...
+use PhpCsFixer\Doctrine\Annotation\Tokens as DoctrineAnnotationTokens;
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
+use PhpCsFixer\Tokenizer\Token as PhpToken;
+use PhpCsFixer\Tokenizer\Tokens as PhpTokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * Base class for Doctrine annotation fixers.
+ */
+abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements ConfigurableFixerInterface
```
keradus: please mark it with internal annotation

src/Doctrine/Annotation/Token.php
```diff
@@ -0,0 +1,159 @@
...
+    const T_STRING = 6;
+    const T_FLOAT = 7;
+    const T_IDENTIFIER = 8;
+    const T_AT = 9;
+    const T_CLOSE_CURLY_BRACES = 10;
+    const T_CLOSE_PARENTHESIS = 11;
+    const T_COMMA = 12;
+    const T_EQUALS = 13;
+    const T_FALSE = 14;
+    const T_NAMESPACE_SEPARATOR = 15;
```
keradus: question:
are there no ready to use doctrine-annotation-parser ?
eg one used by doctrine itself ?

What I am thinking, you make a lot of work to make it working, but i'm afraid about maintaining it.
As you see, there are not much doctrine-users here that could at least review your fixers (not from code point of view, but from annotation syntax point of view)

src/Doctrine/Annotation/Token.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+
+namespace PhpCsFixer\Doctrine\Annotation;
+
+/**
+ * A Doctrine annotation token.
+ */
+class Token
+{
+    const T_NONE = 1;
```
keradus: what's the none ?
only whitespaces ?

src/Doctrine/Annotation/Tokens.php
```diff
@@ -0,0 +1,308 @@
...
+     * {@inheritdoc}
+     */
+    public function offsetSet($index, $token)
+    {
+        if (null === $index) {
+            $index = count($this);
+            $this->setSize($this->getSize() + 1);
+        }
+
+        if (!$token instanceof Token) {
```
keradus: validation should be done before changing the size

README.rst
```diff
@@ -226,6 +226,29 @@ Choose from the list of available rules:
...
+
+* **doctrine_annotation_braces**
+   | Doctrine annotations without arguments must use the configured syntax.
+   | *Rule is: configurable.*
+
+* **doctrine_annotation_declaration**
+   | There must not be any space around Doctrine annotation braces.
+   | *Rule is: configurable.*
+
+* **doctrine_annotation_indentation**
```
keradus: question: would any of those fixers conflicting ?

README.rst
```diff
@@ -226,6 +226,29 @@ Choose from the list of available rules:
    | Replaces dirname(__FILE__) expression with equivalent __DIR__ constant.
    | *Rule is: risky.*
 
+* **doctrine_annotation_argument_space**
```
keradus: question: would any of this fixers run before/after another ?
hint: make an integration test using all doctrine_* fixers to see they work as expected when one use all of them

src/Doctrine/Annotation/Token.php
sstok: https://github.com/doctrine/annotations you should be able to at least use there Lexer.

julienfalque: I agree, I did try to use https://github.com/doctrine/annotations in the first place, without success. I'll try harder.

src/AbstractDoctrineAnnotationFixer.php
julienfalque: PHPDoc added.

src/AbstractDoctrineAnnotationFixer.php
julienfalque: Done.

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,276 @@
...
+     * @return DoctrineAnnotationTokens
+     */
+    private function getDoctrineAnnotationTokens(PhpToken $input)
+    {
+        static $regex;
+
+        if (null === $regex) {
+            $regex = sprintf(
+                '/(%s)/i',
+                implode(')|(', array(
```
julienfalque: > no need for static, php automatically cache regexes

I couldn't find documentation about this, but I made the changes anyway.

src/Doctrine/Annotation/Token.php
julienfalque: `T_NONE` is for whitespaces and meaningless words in the PHP comment, e.g. in
```php
/**
 * Some words
 */
```
`T_NONE` would apply to _Some_ and _words_

src/Doctrine/Annotation/Tokens.php
julienfalque: Done.

README.rst
julienfalque: Integration test added.

src/Doctrine/Annotation/Token.php
julienfalque: I remember why I couldn't use `doctrine/annotations`: the `DocLexer` ignores any token that is not relevant for annotation parsing while we need all the tokens here. The problem is that the class is `final` so we can't extend it to tweak it for our needs. I'm not sure what to do here.

src/Doctrine/Annotation/Token.php
keradus: so, `T_WHITESPACE` and `T_STRING` ?

julienfalque: `T_STRING` is for tokens that are actually parsed as strings by Doctrine Annotations. In this exemple, the words would be ignored and treated just like whitespaces, but they are not whitespaces, so I think `T_WHITESPACE` is not suitable either.

Maybe I should just add a comment next to the constant to make things clearer?

keradus: documentation is always welcome ;)

julienfalque:
I'm stuck with `doctrine/annotations` integration: `DocLexer` ignores meaningless tokens so we can't use it as is. But the class is final, we can't extend it and tweak it. Any suggestion?

julienfalque:
I managed to integrate `doctrine/annotations`' lexer, which allowed me to slightly simplify the code.
I also merged all spacing-related fixers into a single one and improved descriptions.

julienfalque:
I managed to integrate `doctrine/annotations`' lexer, which allowed me to slightly simplify the code.
I also merged all spacing-related fixers into a single one and improved descriptions.

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,234 @@
...
+     *
+     * @return DoctrineAnnotationTokens
+     */
+    private function getDoctrineAnnotationTokens(PhpToken $input)
+    {
+        $tokens = new DoctrineAnnotationTokens();
+        $tokens->setIgnoredTags($this->ignoredTags);
+
+        $lexer = new DocLexer();
+        $lexer->setInput($content = $input->getContent());
```
keradus: please initialize `$content` outside of method parameters

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,234 @@
...
+        while (null !== $token = $lexer->peek()) {
+            if (DocLexer::T_NONE === $token['type']) {
+                continue;
+            }
+
+            if (DocLexer::T_STRING === $token['type']) {
+                $token['value'] = '"'.str_replace('"', '""', $token['value']).'"';
+            }
+
+            if ($lastPosition !== $newPosition = $token['position']) {
```
keradus: do not assign value inside condition

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -0,0 +1,234 @@
...
+
+        return array_key_exists($index, $analyzer->getClassyElements());
+    }
+
+    /**
+     * @param PhpToken $input
+     *
+     * @return DoctrineAnnotationTokens
+     */
+    private function getDoctrineAnnotationTokens(PhpToken $input)
```
keradus: I would move it outside of this class (eg into `Annotation\Tokens::createFrom...`)

src/AbstractDoctrineAnnotationFixer.php
julienfalque: I made a bunch of these, do you want me to fix them all or just this one?

keradus: yes, all of them.
I just found it pointless to complain about every single them individually, making a lot of spam comments ;)

julienfalque: Just wanted to be sure :)
I fixed all.

src/AbstractDoctrineAnnotationFixer.php
julienfalque: Done.

src/AbstractDoctrineAnnotationFixer.php
julienfalque: Moved to `Tokens::createFromDocComment()`.

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
```diff
@@ -0,0 +1,247 @@
...
+            return;
+        }
+
+        $configuration = array_merge(array(
+            'around_parentheses' => true,
+            'around_commas' => true,
+            'around_assignments' => true,
+        ), $configuration);
+
+        $this->aroundParentheses = (bool) $configuration['around_parentheses'];
```
keradus: please take a look at other configurable fixers.
if setting is not a bool, raise an exception. if key is not there, use default value.

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
```diff
@@ -0,0 +1,247 @@
...
+            $this->aroundCommas = true;
+            $this->aroundAssignments = true;
+
+            parent::configure(null);
+
+            return;
+        }
+
+        if (!count($configuration)) {
+            parent::configure($configuration);
```
keradus: this method is calling parent in 3 different places
let us call it once on top of the method.

then, it will not confusing if `ignored_tags` parameter obligatory when `$configuration` is not an array or not, because here, in this method, you convert `$configuration` to `null` if there is empty array, which is not an obvious flow

composer.json
```diff
@@ -15,6 +15,7 @@
     ],
     "require": {
         "php": "^5.3.6 || >=7.0 <7.2",
+        "doctrine/annotations": "^1.2",
```
Slamdunk: Can we put this as a [sugget](https://getcomposer.org/doc/04-schema.md#suggest)?

Those who will use these fixers probably already have it, those who don't will not install another scheme-tree invain

julienfalque: I agree but shouln't we have it as `required` to build the Phar file?

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
julienfalque: I refactored configuration handling: now everything is done in `AbstractDoctrineAnnotationFixer` using protected methods `getDefaultConfiguration()` and `validateConfiguration()` that are overridden is child classes.

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
julienfalque: See my comment in previous review.

composer.json
keradus:
```
keradus@keradus:~/github/PHP-CS-Fixer$ du -sh vendor/
19M	vendor/
keradus@keradus:~/github/PHP-CS-Fixer$ composer require doctrine/annotations:^1.2 -q
keradus@keradus:~/github/PHP-CS-Fixer$ du -sh vendor/
19M	vendor/
```

GrahamCampbell: > I agree but shouln't we have it as required to build the Phar file?

Should go in require dev then?

GrahamCampbell: Probably a bad idea to add this dependency since they've already dropped php 5.3->5.5 support, so I can only see this causing future dependency conflicts.

julienfalque: > I can only see this causing future dependency conflicts.

 I think it's fine in the case PHP-CS-Fixer is used as a Composer dependency: PHP <5.6 users will have `doctrine/annotations` 1.2 installed while PHP 5.6+ users will have version 1.3+ which is supposed to be backward compatible. I don't know if this is an issue regarding the Phar version of PHP-CS-Fixer though.

> Should go in require dev then?

It won't work in the case PHP-CS-Fixer is used as a Composer dependency in a project that does not require `doctrine/annotations` itself, so I would either have it in `require`, or move it to `suggest` as @Slamdunk proposed, making it explicit that this must be manually installed if wanted.

keradus: no, version you requested is 1.2, which will work on all versions that we do support.
while building phar, we set up platform config for composer, so it will install newest version that will still work on lowest supported platform.

keradus: `annotations` + required by it `lexer` packages are both very light. let us keep it in `requirements`.

GrahamCampbell: My point was not that we couldn't install it, it was that newer versions will not support old php, and then we'll have no choice but to upgrade when ever there are a mountain of conflicts with other people wanting a 2.x version of this package, but we need 1.x.

GrahamCampbell: Rather, I'm referring to the futureproofing of this decision.

keradus: this is pitfall of composer. While experiencing conflicts of cli apps I suggest to use phar version.

Slamdunk: @keradus My point wasn't about weigth of packages. The point is that I've seen too many times a project that offers 100 functionalities, you want to use 1 of them, and you end up with an entire operative system to use just a hundredth of it.

For example, Zend has already taken this approach: [zend-view](https://github.com/zendframework/zend-view/blob/master/composer.json) **can** operate dozens of plugin, and they are all in `require-dev`, but you are free to choose which one to install.
The [zend-cache](https://github.com/zendframework/zend-cache/blob/master/composer.json) does this with extensions too.

TomasVotruba:
It seems stuck. What needs to be done here?

julienfalque:
I think this requires more tests and reviews. Also I'd like #2500 to be merged first to leverage it here.

keradus:
@julienfalque, do you have some doctrine-familiar friends (eg from projects you work on) that could at least doublecheck the principles of rules you have implemented? like is the code after fixes is nice here ?

julienfalque:
I don't think my current coworkers are familiar with Doctrine Annotations, but it's a quite widespread lib so there are probably people here that use it and could help.

keradus:
As you see, it's hard to get their feedback here

julienfalque:
I'm thinking about splitting this PR into smaller parts, would it help?

hkdobrev:
I'll make sure to test it out this weekend on our project which we've recently updated with even more Doctrine annotations including from Doctrine extensions. I even created a reminder to test this PR out on Saturday :)

keradus:
Thank you @hkdobrev , it would be really appreciated !

@juliendufresne , I'm not sure it would be helpful here. We first need to confirm rules, splitting PR won't help here. Then we need to review the code, but if splitted (like 3x rule + annotation classes) then you don't see the whole context here

hkdobrev:
Here are a few things I've noticed when running the fixers on our codebase:

1. The fixers are changing some whitespace in legacy PHPDoc annotations which are not desirable.

```php
/**
 * @copyright  (c) 2011-2017 Acme Inc.
 */
```
is changed to:
```php
/**
 * @copyright(c) 2011-2017 Acme Inc.
 */
```
Before the annotation was only `@copyright` and the content was aligned to the above annotation content. It seems the fixer is recognising the annotation as a Doctrine annotation with arguments incorrectly.

I can see `@copyright` is already in the list of PHPDoc annotations ignored by default. Do you think you could look into what's causing this?

2. Options in Doctrine annotations have some unwanted whitespace changed:

Before:
```php
/**
 * @ORM\Column(type="integer", nullable=true, options={"unsigned":true})
 */
```
After:
```php
/**
 * @ORM\Column(type="integer", nullable=true, options={"unsigned" : true})
 */
```

Notice there is a space added before the colon character in the options.

---

Other than these, it works as expected, fixes what it needs to fix and the configuration works as expected 🎉

julienfalque:
@hkdobrev Thank you for testing this!

> The fixers are changing some whitespace in legacy PHPDoc annotations

Yes, ignored tags were not correctly handled in some parts of the fixers. I fixed this, added tests and updated the default list of ignored tags.

> Options in Doctrine annotations have some unwanted whitespace changed

This is expected. For now, the fixer removes spaces around argument assignment operator and forces one space around array assignment operator. I made this choice based on the syntax used in Symfony documentation, see [@Route and @Method](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/routing.html) for example.

I think adding options to change this behavior makes sense, but I'd rather do it in a separate PR later.

* * *

I tried running the fixers on PHP-CS-Fixer code itself. No change should have occured (as we don't use Doctrine Annotation internally) but I actually got one change. I think there is a bug in Doctrine Annotation's lexer, I'll investigate this. For now, I just added the corresponding failing test cases.

hkdobrev:
Just wanted to mention I've tried it again after the latest updates from @julienfalque and everything works fine :)

julienfalque:
> I think adding options to change this behavior makes sense, but I'd rather do it in a separate PR later.

Actually I think I'm going to implement the options in this PR, as I will split the `around_assignments` option into `around_argument_assignments` and `around_array_assignments` options, which would require to deprecate the current one.

julienfalque:
@hkdobrev I splitted option `around_assignments` into `around_argument_assignments` and `around_array_assignments`. Now you should be able to run the fixer without having unexpected whitespace added around colons.

Note that this will not remove spaces around colons when already there. Thinking about it, this would require to change all options to support setting whether you (don't) want a space before and/or after a `(`, before and/or after a `)`, and so on. It should not be very complicated to implement, but I'm not sure it's worth it.

keradus:
ing @julienfalque

julienfalque:
I'd like to be sure tests cover as much cases as possible. If anybody can test this on other projects and report issues, that would be very helpful :)

keradus:
We tried it over last 6 months (!!!) and only @hkdobrev did (good of himself!).
If we will wait for next user, it will take next 6 months.
Let you assure that rules are as good as possible, then we will release and collect bug reports, if any.

hkdobrev:
Yeah, release early and often :) People would start gradually using those and the early adopters would notice the changes and report back if there are any problems.

julienfalque:
Then I think it's ok to merge it right now :)

SpacePossum:
than please rebase :)

keradus:
Thank you @julienfalque.

teohhanhui:
I think the "ignored tags" solution is really clumsy and hackish. Couldn't we check for `@Annotation` on the annotation class itself?

julienfalque:
Sadly we can't currently because the tool has no way to load the annotation class and thus cannot check whether it's an actual `@Annotation`.

TomasVotruba:
I made one more rule for nested annotations:

```diff
 /**
- * @ORM\Table(name="table_name", indexes={@ORM\Index(name="...", columns={"..."}), @ORM\Index(name="...", columns={"..."})})
+ * @ORM\Table(name="table_name", indexes={
+ * @ORM\Index(name="...", columns={"..."}), 
+ * @ORM\Index(name="...", columns={"..."})
+ * })
  */
final class SomeEntity
{
}
```


https://github.com/symplify/symplify/pull/2078

TomasVotruba:
> I think the "ignored tags" solution is really clumsy and hackish. Couldn't we check for `@Annotation` on the annotation class itself?

Agreed the solution is broken. It breaks many non-annotations like `@type` etc. Better opt-in for `Doctrine\ORM\Annotation*` - 1 line, rather than guess all the words in the worlds people can use for annotations.

PR #2227, opened 2016-10-19
Stop polluting global namespace with CT

keradus:
Practically done.
2 non-functional changes still to be done (marked as TODO in code)


src/Console/Command/FixCommand.php
```diff
@@ -287,8 +286,6 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        Transformers::create(); // make sure all transformers have defined custom tokens since configuration might depend on it
```
keradus: that was the couse of whole refactoring

SpacePossum:
:+1: awesome :)

src/Tokenizer/CT.php
```diff
@@ -0,0 +1,53 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
```
keradus: actually not, it may be used in 3rd party fixers

GrahamCampbell:
I'm actually not so sure this is needed? Why does this matter if this is just a tool?

GrahamCampbell:
Surely it's just making branch merges really hard now?

keradus:
To not pollute global namespace with some constants at one, quite random runtime execution point.
So basically fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2227#discussion_r84184603

PR #2232, opened 2016-10-24
Add a test for style of varaible decalration : var

daiglej:
Bug fix: 
Symfony\CS\Fixer\Symfony\PhpdocToCommentFixer 

Was converting doc-block from old style "var" property declaration


SpacePossum:
:+1:

keradus:
Thank you @daiglej.

PR #2234, opened 2016-10-25
Added some failing PHP7 test cases for ordered_use

pascal-hofmann:
This PR adds two test cases for ordered_use fixer and group use declarations (PHP7).

Test case 1: Imports inside group use declarations should be ordered too.
Test case 2: The fixer generates invalid PHP.

### DIFF
--- a/tests/Fixer/Import/OrderedImportsFixerTest.php
+++ b/tests/Fixer/Import/OrderedImportsFixerTest.php
@@ -639,6 +639,30 @@ use A\B;
 use const some\a\{ConstA, ConstB, ConstC};
 ',
             ),
+            array(
+                '<?php
+use A\B;
+use some\a\{  ClassA as A,ClassB, ClassC};
+use const some\a\{ConstA,ConstC,ConstB};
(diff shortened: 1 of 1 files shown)


keradus:
Great catch @pascal-hofmann !
are you able to provide fix as well? it would be great !

pascal-hofmann:
I tried fixing the fixer, but failed. I think especially the second test case I added won't be easy to fix. Unfortunately I can't afford to put more time into this. :-(

keradus:
@SpacePossum , maybe you are able to help here ?

SpacePossum:
Bug fix should go to 1.12 I think?
I played with the fixer for an hour or so, but it is not the easiest one to fix, will look more into this soon I hope. Thanks for reporting @pascal-hofmann

SpacePossum:
@keradus with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2235 merged this one can be closed I think, unless @pascal-hofmann wants to rebase and add more stuff?

pascal-hofmann:
Nope. Closed! :-)

PR #2244, opened 2016-10-28
Config cleanup

keradus:
Goal is to clean up the Config and it's usage

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -12,6 +12,7 @@
 
 namespace PhpCsFixer\Console;
 
+use PhpCsFixer\Cache\CacheManagerInterface;
 use PhpCsFixer\Cache\FileCacheManager;
 use PhpCsFixer\Cache\FileHandler;
 use PhpCsFixer\Cache\NullCacheManager;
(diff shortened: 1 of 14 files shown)


SpacePossum:
thanks for picking this up, will be a great improvement : )

SpacePossum:
Out of curiosity, do the attr. of the `Config` class need to be/stay `protected`? It feels like hard endpoints to maintain (BC) without a lot of gain.

keradus:
Valid question.
I want to make a Config a DTO object without many deps, clean it up from unneeded props and then let us see.

keradus:
https://github.com/keradus/PHP-CS-Fixer/pull/16 is merged into this PR

src/Console/ConfigurationResolver.php
```diff
@@ -37,17 +45,17 @@
     const PATH_MODE_INTERSECTION = 'intersection';
 
     /**
-     * @var bool
+     * @var null|bool
```
SpacePossum: can this really be `null` and what would it mean?

keradus: Property is lazy loaded and should be accessed only via getter, which output is not nullable.

keradus:
Thank you @SpacePossum.

src/Console/ConfigurationResolver.php
SpacePossum: :+1: I see now

PR #2248, opened 2016-10-30
Added a fixer to add a New Line before phpdoc (black list). 

daiglej:
Added a fixer to add a New Line before phpdoc. 

Will not add a new line before:
- File level doc-block.
- First doc-block inside curly brackets.
- Consecutive Inline @var

Example of consecutive inline @var :

```
<?php

/** @var SomeClass $foo */
/** @var ComeOtherClass $bar */
list($foo, $bar) = SomeWeird::factory($input);
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -576,6 +576,10 @@ Choose from the list of available rules:
 * **single_blank_line_before_namespace** [@Symfony]
    | There should be exactly one blank line before a namespace declaration.
 
+* **single_blank_line_before_phpdoc**
+   | Adds a blank line before phpdoc, excepts for: first element in curly
+   | brackets "{ }", file level phpdoc and consecutive inline @var.
+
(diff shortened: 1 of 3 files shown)


keradus:
For first to minutes it looks overcomplicated.
Why not add missing new line if the character before docblock is `;` ?

daiglej:
Sorry for the long answer, it's been a very busy week.
Yes I think that should do it, I will post an update soon.

daiglej:
So I created an alternate version here : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2263

In ended up being not that much simpler, since I soon realized I wanted a blank line between the file doc and the class doc, but not between 2 consecutive inline @var

keradus:
Closing in fav. of #2263

Issue #2250, opened 2016-11-01
priority issue: protected_to_private vs ordered_class_elements


keradus:
`protected_to_private` should run before `ordered_class_elements`

PR #2251, opened 2016-11-01
ProtectedToPrivateFixer - fix priority, fix comments with new fixer names

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2250


SpacePossum:
Why don't we disable fabbot.io for the project?
We do already ready run the fixer as part of the required tests to pass so there is no real value and we would lose the false negatives on PRs passing.

keradus:
it's funny way for me to guess which version of CS Fixer it is using ;)

keradus:
Thank you @SpacePossum.

PR #2252, opened 2016-11-01
ClassDefinitionFixer - Set configuration of the fixer in the RuleSet of SF.

SpacePossum:
These are the last things for the master branch after #2155 was merged to 1.12
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2155#issuecomment-257109271

### DIFF
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -57,6 +57,7 @@ final class RuleSet implements RuleSetInterface
             'blank_line_after_opening_tag' => true,
             'blank_line_before_return' => true,
             'cast_spaces' => true,
+            'class_definition' => array('singleLine' => true),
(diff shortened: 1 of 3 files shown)


keradus:
does default configuration of fixer match PSR-2 ?

SpacePossum:
>does default configuration of fixer match PSR-2 ?

yes, on a side note all other options of the fixer are PSR2 compliant. However the configuration options deals with parts of PSR-2 that state _may_ and not _must_ (which is what the Fixer does on default) :)

keradus:
sure, but the default option must not  change PSR-2 valid into another PSR-2 valid syntax.
Great !

keradus:
Thank you @SpacePossum.

PR #1919, opened 2016-05-26
EmptyReturnFixer - move fixer outside of Symfony level

keradus:
ref #1638

cc @stof

PR #2255, opened 2016-11-02
Use empty_return

julienfalque:
Using a `return null` statement at the end of a function will make fabbot.io complain and suggest a patch. Using `empty_return` fixer will prevent this.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -16,6 +16,7 @@ return Symfony\CS\Config::create()
     // use default SYMFONY_LEVEL and extra fixers:
     ->fixers(array(
         'combine_consecutive_unsets',
+        'empty_return',
         'header_comment',
         'long_array_syntax',
         'no_useless_else',


keradus:
Hi !

I'm not sure about your intentions here.
Do you want **this** project to follow `empty_return` rule on **this** project codebase, or you want to extend `SYMFONY_LEVEL` to use that rule (so fabbot will follow it on all repos) ?

In any cases - after PHP7 this rule is problematic and was removed from both - Symfony ruleset and this project settings: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1919

julienfalque:
I don't like the `empty_return` rule but it looks like _fabbot.io_ actually uses it. On this [report](http://fabbot.io/report/FriendsOfPHP/PHP-CS-Fixer/2220/eb5373e43fa442f1a5a5eee3aa22720343b6bdff), all `return null` statements are replaced with `return`. Is this expected?

If yes, then this project should follow the same rules to avoid such report. Otherwise, can we remove `empty_return` from _fabbot.io_'s rules?

keradus:
> fabbot.io actually uses it

fabbot uses outdated version of CS Fixer.

> this project should follow the same rules to avoid such report

I disagree. fabbot.io hook is not required by this project, but we still want to run it.

> can we remove empty_return from fabbot.io's rules?

PHP CS Fixer is a tool used by many CI services. One of them is fabbot. It's up to service's maintainers to update the tool and I believe they will update it shortly.

keradus:
Closing due to discussion

Issue #1638, opened 2015-12-29
Return fixer should be risky


GrahamCampbell:
The `return null;` to `return;` fixer we have needs to be marked as risky as it could break code on php 7.1.

See https://wiki.php.net/rfc/void_return_type.

Also note that the rfc actually changes the definition of `void` from what we thought it was before. It now strictly means nothing rather than just garbage.

keradus:
Could you provide example of risky change as well?

GrahamCampbell:
Just have a fuction marked as return void, and put `return null;` in it. It's beahviour changes before and after the fixer is run.

GrahamCampbell:
One could argue the code that said `return null;` was broken already, but who's to say the intention of the original code wasn't to raise a type error.

GrahamCampbell:
We've made a change to the code behaviour, if you see what I mean.

keradus:
https://3v4l.org/h0tn2

need to be documented well why it is risky ;)

GrahamCampbell:
> 3v4l.org

That website doesn't contain the void return type code that's new in php 7.1.

GrahamCampbell:
Only HHVM 3.11 actually gave an output of anything close to what php 7.1.0 will give when it's released. The old version of php 7.1 still thought void was the name of an object.

keradus:
I'm ignoring hhvm...

Behavior on php7.1 will be same way, just a bit different Error will be thrown, isn't it?

GrahamCampbell:
Yeh, but in terms of recognising void as a type rather than a class name, it was closer.

GrahamCampbell:
We'll need to wait for the december version of php 7.1 to make it to 3v4l.org

keradus:
The problem exists even without change to PHP, as showed on 3v4l.

GrahamCampbell:
https://3v4l.org/1h9P9

GrahamCampbell:
not really - there's still a type error before and after on 7.0

keradus:
So let us see what error will be thrown, maybe it will be in same inheritance line.

stof:
Well, actually, this fixer should be removed from the Symfony level if this RFC is accepted in PHP 7, because the rule defined in the Symfony coding standards will become invalid (I already don't like this rule)

stof:
OK, actually, the RFC is already accepted. So we should start a discussion in Symfony instead.

GrahamCampbell:
[...]

PR #2256, opened 2016-11-03
EmptyReturnFixer - it's now risky fixer due to null vs void

keradus:
ref #1638 
https://3v4l.org/7HscE


SpacePossum:
:+1:  :ship:

PR #2262, opened 2016-11-05
Fixer for new line before doc block version 2

daiglej:


keradus:
wow , sth went terribly wrong here

keradus:
replaced by #2263

Issue #2260, opened 2016-11-04
SelfUpdateCommand - Do not auto update to next major


SpacePossum:
RFC

I think it would be nice to not auto update 1.x to 2.x (when 2.x is released) when running the `SelfUpdateCommand`.
Showing a message that a new major is released with a link to the `UPGRADE.md` would be nice. Additional we could add a `-f` - flag to allow forcing to updating to next major release.

The reason why I think it would be good is that when people have a (custom) config file it needs updating or the Fixer won't run. I would like to prevent getting a lot of people reporting:

```
$ php php-cs-fixer fix -vvv
PHP Fatal error:  Uncaught Error: Class 'Symfony\CS\Fixer\Contrib\HeaderCommentFixer' not found in /home/possum/work/PHP-CS-Fixer/.php_cs:13
```

keradus:
:+1:

GrahamCampbell:
Maybe we should still show a notice if a new stable version is available, but the major version of the new version is greater?

SpacePossum:
Sure, check it out https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2264/files#diff-b48dda0c2ec7e050caee3b14d174b17dR85 :)

GrahamCampbell:
👍

glensc:
This issue comes up when searching `Symfony\CS\Fixer\Contrib\HeaderCommentFixer`. would be nice to add here link for `> 1` syntax for same code.

SpacePossum:
Little general help for all who come here because of the error message:

For upgrading from 1.x to 2.x please checkout the upgrade guide;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/UPGRADE.md

Feel free to ask more on this topic or ask for help at the community;
https://gitter.im/FriendsOfPHP/PHP-CS-Fixer

PR #2264, opened 2016-11-05
SelfUpdateCommand - Do not update to next major version by default

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2260

### DIFF
--- a/Symfony/CS/Console/Command/SelfUpdateCommand.php
+++ b/Symfony/CS/Console/Command/SelfUpdateCommand.php
@@ -14,6 +14,7 @@ namespace Symfony\CS\Console\Command;
 
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\CS\ToolInfo;
 
@@ -33,10 +34,16 @@ class SelfUpdateCommand extends Command
         $this
             ->setName('self-update')
             ->setAliases(array('selfupdate'))
+            ->setDefinition(
+                array(
+                    new InputOption('--force', '-f', InputOption::VALUE_NONE, 'Force update to next major version if available.'),
+                )
+            )
             ->setDescription('Update php-cs-fixer.phar to the latest stable version.')
             ->setHelp(<<<'EOT'
 The <info>%command.name%</info> command replace your php-cs-fixer.phar by the
-latest version from cs.sensiolabs.org.
+latest version released on:
+<comment>https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases</comment>
 
 <info>$ php php-cs-fixer.phar %command.name%</info>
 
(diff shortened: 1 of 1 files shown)


Symfony/CS/Console/Command/SelfUpdateCommand.php
```diff
@@ -56,28 +64,46 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            return 0;
         }
 
-        $remoteFilename = $this->buildVersionFileUrl($remoteTag);
+        $remoteVersionParsed = $this->parseVersion($remoteTag);
+        $currentVersionParsed = $this->parseVersion($currentVersion);
+
+        if ($remoteVersionParsed[0] > $currentVersionParsed[0] && true !== $input->getOption('force')) {
+            $output->writeln(sprintf('<info>A new major version of php-cs-fixer is available</info> (<comment>%s</comment>)', $remoteTag));
+            $output->writeln(sprintf('<info>Before upgrading please read</info> https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/%d.%d/UPGRADE.md', $remoteTag[0], $remoteTag[1]));
```
keradus: `$remoteTag` is a string, eg `v1.12.3`. accessing it's idxs is not good idea.
Didn't you want to use `$remoteVersionParsed` here instead ?

SpacePossum: indeed, good catch! I did some renaming and this one went wrong indeed

Symfony/CS/Console/Command/SelfUpdateCommand.php
```diff
@@ -83,7 +83,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         if ($remoteVersionParsed[0] > $currentVersionParsed[0] && true !== $input->getOption('force')) {
             $output->writeln(sprintf('<info>A new major version of php-cs-fixer is available</info> (<comment>%s</comment>)', $remoteTag));
-            $output->writeln(sprintf('<info>Before upgrading please read</info> https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/%d.%d/UPGRADE.md', $remoteTag[0], $remoteTag[1]));
+            $output->writeln(sprintf('<info>Before upgrading please read</info> https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/%d.%d/UPGRADE.md', $remoteVersionParsed[0], $remoteVersionParsed[1]));
```
keradus: Actually, please don't assume branch name. Just use tag.
it would look like `https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.0.0-alpha/UPGRADE.md`

SpacePossum: :+1: updated, didn't know github supported that :)

SpacePossum:
fabbot.io is wrong, this PR is ready

keradus:
Thank you @SpacePossum.

PR #2265, opened 2016-11-05
Add custom dumper for Token.

SpacePossum:
dump

left is new, right is default of SF dumper when using `dump($token)` in a unit test


keradus:
I never needed to do this.
Instead, I would rather implement `Token::__debugInfo()`

SpacePossum:
will try diff. approach (or abandon the idea all together)

PR #2266, opened 2016-11-05
config negative

SpacePossum:
not sure if we want this, otherwise lets fix the exception message on config something like
```php
<?php

$header = <<<'EOF'
This file is part of PHP CS Fixer.

(c) Fabien Potencier <fabien@symfony.com>
    Dariusz Rumiński <dariusz.ruminski@gmail.com>

This source file is subject to the MIT license that is bundled
with this source code in the file LICENSE.
EOF;

    ->setRiskyAllowed(true)
    ->setRules(array(
        '@Symfony' => true,
        '-braces',
    ))
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->exclude('tests/Fixtures')
            ->in(__DIR__)
    )
;
```

current exception message is 
```
  [UnexpectedValueException]  
  Rule "0" does not exist.    
```

### DIFF
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -269,7 +269,13 @@ final class RuleSet implements RuleSetInterface
         } while ($hasSet);
 
         // filter out all rules that are off
-        $rules = array_filter($rules);
+        $rules = array_filter(
+            $rules,
+            function($name) use ($rules) {
+                return '-' !== $name[0] && $rules[$name] && !array_key_exists('-'.$name, $rules);
+            },
+            ARRAY_FILTER_USE_KEY
+        );
 
         $this->rules = $rules;
 
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -48,7 +48,7 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
      * @expectedException              \PhpCsFixer\ConfigurationException\InvalidConfigurationException
      * @expectedExceptionMessageRegExp /^Unknown option name: "foo"\.$/
      */
-    public function atestSetOptionWithUndefinedOption()
+    public function testSetOptionWithUndefinedOption()
     {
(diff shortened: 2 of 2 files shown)


src/RuleSet.php
```diff
@@ -269,7 +269,13 @@ private function resolveSet()
         } while ($hasSet);
 
         // filter out all rules that are off
-        $rules = array_filter($rules);
+        $rules = array_filter(
+            $rules,
+            function($name) use ($rules) {
+                return '-' !== $name[0] && $rules[$name] && !array_key_exists('-'.$name, $rules);
+            },
+            ARRAY_FILTER_USE_KEY
```
keradus: nope, `ARRAY_FILTER_USE_KEY` was introduced in 5.6

keradus:
we do not allow `'-braces',` format in config file. please do not make exception.
key must be the rule name, value the rule enable/disable flag/config.
because if we would allow `'-braces'`, (which is value, not key!), then we would need to also allow `'braces'` (again value, not a key)..
I agree this is a bug, but we should make better exception that try to figure it out what user wanted to achieve and try to fix it for him

src/RuleSet.php
SpacePossum: :(

keradus: will send a proposal in 2 minutes

SpacePossum:
Makes sense, I would say it is not even a bug than.
What we could improve is letting the Fixer check on reading the configuration file if all items in the `setRules(array` are key-value pairs.
If not check if the key starts with `-` and create a nice exception message, otherwise create some other message. Secondly we could than throw a `ConfigurationException`.
What you think?

keradus:
I'm not against the message if one use `-`, but I don't need it as well.

Meanwhile - let me validate `RuleSet` : #2267

SpacePossum:
closing in fav. of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2267

PR #2267, opened 2016-11-05
RuleSet - validate set

keradus:
```
D:\GIT\fork\PHP-CS-Fixer (2_ruleset)
λ php php-cs-fixer fix -v
Loaded config default from "D:\GIT\fork\PHP-CS-Fixer\.php_cs.dist".


  [InvalidArgumentException]
  Missing value for "-braces" rule/set.
```

### DIFF
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -169,6 +169,12 @@ final class RuleSet implements RuleSetInterface
 
     public function __construct(array $set = array())
     {
+        foreach ($set as $key => $value) {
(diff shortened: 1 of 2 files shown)


SpacePossum:
could those exceptions be `InvalidConfigurationException`? If not I happy with the current PR as well :)
:+1:

keradus:
I say it would RuleSet should not be aware about configuration.
It may be catch during creation via resolver and then rethrow as configuration error

PR #2274, opened 2016-11-09
ShortArraySyntaxFixer/LongArraySyntaxFixer - Merge conflicting fixers

SpacePossum:


keradus:
nice so far ;)
(btw, please check gitter)

SpacePossum:
problem here is that the integration test will resolve the fixers _and_ configure those before checking the requirements set in the test.

keradus:
Why it is a problem ?

SpacePossum:
I check if the configuration of the fixer makes sense based on the env., so this will fail:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2274/files#diff-00016b9c4b10354d36ed225366173552R58

even if I set the requirements of the test on 5.4

keradus:
then why not simply check requirements before creating fixers? (what actually makes perfect sense)

SpacePossum:
>then why not simply check requirements before creating fixers? (what actually makes perfect sense)

thats the plan :)   :+1:

SpacePossum:
This PR also fixes using `setFormat()` in a configuration file as it was not used by the Fixer (i.e. it would always run with the value given on commandline or fall back on 'txt').

keradus:
could we separated that fix? PR getting bigger and bigger already, and that bugfix is not directly related

SpacePossum:
Expect Travis PHP5.3 to fail because of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2285

keradus:
rebased

keradus:
Thank you @SpacePossum.

PR #2277, opened 2016-11-09
Remove trailing spaces

keradus:


SpacePossum:
:+1: I blame PHPStorm for no longer removing all trailing space as it used to ;)

PR #2278, opened 2016-11-09
Concat*Fixer - unify concat fixers

SpacePossum:


SpacePossum:
fabbot.io is wrong, PR is ready

src/Fixer/Operator/ConcatSpaceFixer.php
```diff
@@ -0,0 +1,138 @@
...
+            throw new InvalidFixerConfigurationException($this->getName(), 'Missing "spacing" configuration.');
+        }
+
+        switch ($configuration['spacing']) {
+            case 'one':
+                $this->fixCallback = 'fixConcatenationToSingleSpace';
+
+                break;
+            case 'none':
+                $this->fixCallback = 'fixConcatenationToNoSpace';
```
keradus: would be great to not call methods by string. really hard to perform static analysis then

tests/Fixer/Operator/ConcatSpaceFixerTest.php
```diff
@@ -16,20 +16,40 @@
...
+        $this->getFixer()->configure(array('a' => 1));
+    }
+
+    /**
+     * @expectedException \PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException
+     * @expectedExceptionMessageRegExp #^\[concat_space\] "spacing" configuration must be "one" or "none".$#
+     */
+    public function testInvalidConfigValue()
+    {
+        $this->getFixer()->configure(array('spacing' => 'tabs'));
```
keradus: this will modify the instance of Fixer between tests. make a clone of it before modifying

src/Fixer/Operator/ConcatSpaceFixer.php
SpacePossum: I noticed, as some SCA will give the `unused private method`-warning. 
While I don't care that much about it I was looking into the PHP internals what the cost is for a dynamic name method call vs. a defined one, but I haven't concluded that little trip down the internals road ;)
I did a few fixers this way and we could stop doing that and update those already accepted at some point.
There are than 2 other ways to do this:
1 - check the configuration each time a token of interest is found to see what method should be called to fix it, with the cost that the answer to that will always be same during execution
2 - within the configuration method create an private interfaced object what will be called during fixing that does the fixing. This will be at the cost of creating more classes/code which means more to maintain.

I've no strong feeling about any of the 3 solutions, but the solution `1` feels like the least optimal.

tests/Fixer/Operator/ConcatSpaceFixerTest.php
SpacePossum: This will be a repeated pattern in the test of fixers that are configurable. Maybe we can add a protected method to the base test class to get a fresh clone of the fixer being tested?

keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2288

src/Fixer/Operator/ConcatSpaceFixer.php
keradus:
```php
$callbackMap = [
    'none' => function () { $this->fixConcatenationToNoSpace(); }
];
```
lambda proxy method to not hide function usages.

keradus:
Sadly 5.3 is failing, same reason for which i didn't merge phpmd branch :/
please fix

keradus:
Sorry, i don't like it. just use the ugly previous form of `$this->$callback`, and then we will clean this up properly without that delegator while dropping 5.3

SpacePossum:
(as told, I didn't like it either)

keradus:
rebased + added upgrade guide

keradus:
Thank you @SpacePossum.

PR #2148, opened 2016-08-17
Make Order use fixer configurable and added sort by length option

DarkaOnLine:

PR #2280, opened 2016-11-10
Configurable OrderedImportsFixer

DarkaOnLine:


GrahamCampbell:
Lots of failing tests now.

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -782,15 +782,15 @@ public function testInvalidConfigWithMultipleSortAlgorithm()
...
+        $this->getFixer()->configure(array(
             'typesOrder' => null,
             'sortAlgorithm' => new \stdClass(),
         ));
     }
 
     public function testFixByLength()
     {
-        OrderedImportsFixer::configure(array(
+        $this->getFixer()->configure(array(
```
keradus: that way you modify the fixer instance between test methods. dont

keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2288

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -21,13 +22,104 @@
...
+     */
+    private static $supportedSortAlgorithms = array(self::SORT_ALPHA, self::SORT_LENGTH);
+
+    /**
+     * Configuration settings for sort algorithm and sor types.
+     *
+     * array['sortAlgorithm']   string defines sort algorithm
+     * array['typesOrder']      int[]|null defines import types order or null to sort only by algorithm
+     *
+     * @param array $configuration
```
keradus: not really

DarkaOnLine: array|null ?

DarkaOnLine: string[]|null ?

keradus: yep. check the docs of parent class/interface.
null is meaningful here. it is supposed to reset fixer to use default settings.

keradus: sorry, not sure what do you mean ?

DarkaOnLine: Ups. Sorry looked into the wrong class .

DarkaOnLine: Changed. Alos added `Use 'null' for sorting alphabetically by default.` hope it's ok ?

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -68,10 +68,12 @@
     /**
      * Configuration settings for sort algorithm and sor types.
      *
+     * Use 'null' for sorting alphabetically by default.
```
keradus: implementation of the method is not following that.

DarkaOnLine: But I am setting default here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2280/files#diff-80d6131ebed1ab07edef75d2d7ad6c82R59 

Or you think it's better to do like this:

```php
    self::$sortAlgorithm = self::SORT_ALPHA;
    return;
}
```

Or remove the comment ?

keradus: regardless there will be that comment or not - this behaviour is described in interface, so all classes that implements it must follow that.

currently: 

```php
$f = new OrderedImportsFixer(); // configuration is default
$f->configure(null); // configuration is still default
$f->configure($customArrayConfig); // configuration is custom
$f->configure(null); // configuration is still custom, while null should reset it do default one
```

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -21,13 +22,106 @@
...
+     * @var string[]
+     */
+    private static $supportedSortTypes = array(self::IMPORT_TYPE_CLASS, self::IMPORT_TYPE_CONST, self::IMPORT_TYPE_FUNCTION);
+
+    /**
+     * Default sort algorithm.
+     *
+     * @var string
+     */
+    private static $sortAlgorithm = self::SORT_ALPHA;
```
keradus: this is configuration for a instance, should not be shared between all instances. no need for static here.

DarkaOnLine:
@keradus is there anything else I can do ?

SpacePossum:
someone please kick the Travis tests again, it failed because it got stalled

keradus:
@DarkaOnLine https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2280#discussion_r87712447 is still not resolved

SpacePossum:
@GrahamCampbell can you take a look (i think this a laravel standard)?

DarkaOnLine:
@SpacePossum I think he already reviewed it here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2148#issuecomment-258633327

keradus:
And then have been modified and adjusted for master branch, and then have few things raised to be updated (already done), so would be great to have final review, even if it would be quick one ;)

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -251,4 +408,23 @@ private function getNewOrder(array $uses, Tokens $tokens)
...
+     *
+     * @return array
+     */
+    private function sortByAlgorithm($indexes)
+    {
+        switch ($this->sortAlgorithm) {
+            case self::SORT_LENGTH:
+                uasort($indexes, array($this, 'sortByLength'));
+                break;
+            default:
```
keradus:
```
case self::SORT_ALPHA:
    uasort(....);
    break;
default:
    throw new Invalid...
```

DarkaOnLine: But we already checking if correct sort type is passed in configuration: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2280/files/da3106b7fbb3249ad15e8698905805acf933efd7#diff-80d6131ebed1ab07edef75d2d7ad6c82R120

and by default, we are sorting alphabetically

keradus: awesome, so, the config value is validated and is always great.
and then, one will be playing around and may, even by accident, just do: `$this->sortAlgorithm = 'foobarbaz'` ;)

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -417,11 +417,14 @@ private function getNewOrder(array $uses, Tokens $tokens)
...
         switch ($this->sortAlgorithm) {
+            case self::SORT_ALPHA:
+                uasort($indexes, array($this, 'sortAlphabetically'));
+                break;
             case self::SORT_LENGTH:
                 uasort($indexes, array($this, 'sortByLength'));
                 break;
             default:
-                uasort($indexes, array($this, 'sortAlphabetically'));
+                throw new InvalidFixerConfigurationException('ordered_use', sprintf('Sort algorithm is invalid. Should be one of the: "%s"', implode('", "', $this->supportedSortAlgorithms)));
```
keradus: here, it's not setting up the onfiguration so this exception is overused.
just rise logic exception

DarkaOnLine: Changed: e30fdb1

DarkaOnLine:
@GrahamCampbell,  please review this :pray:

GrahamCampbell:
This looks reasonable in principle. I'll need a bit of time to test this in more detail.

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -42,34 +42,32 @@
...
      */
-    private static $sortTypesOrder = null;
+    private $sortTypesOrder = null;
 
     /**
      * Array of supported sort types in configuration.
      *
      * @var string[]
      */
-    private static $supportedSortTypes = array(self::IMPORT_TYPE_CLASS, self::IMPORT_TYPE_CONST, self::IMPORT_TYPE_FUNCTION);
```
DarkaOnLine: But @keradus suggested to remove statics : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2280#pullrequestreview-8316205

So what should I do? :)

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -1497,6 +1499,8 @@ public function provide70CasesByLength()
     /**
      * @dataProvider provide70TypesOrderAndLength
      * @requires PHP 7.0
+     * @param mixed      $expected
+     * @param null|mixed $input
```
DarkaOnLine: fixed

src/Fixer/Import/OrderedImportsFixer.php
GrahamCampbell: I don't think he meant this one though. Ping @keradus.

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -692,4 +693,935 @@ public function provide70Cases()
...
     }
+
+    /*
+    |--------------------------------------------------------------------------
+    | Test sorting by length
+    |--------------------------------------------------------------------------
+    */
+
+    /**
+     * @expectedException \PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException
```
GrahamCampbell: Please remove all these annotations, and put use the newer syntax please, as updated in the rest of the codebase. See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2381.

DarkaOnLine:
@keradus can we move forward ?

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -692,4 +693,940 @@ public function provide70Cases()
...
+        $this->setExpectedException(
+            'PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException',
+            '[ordered_imports] Sort algorithm is invalid. Should be one of the: "alpha", "length"'
+        );
+
+        $this->fixer->configure(array(
+            'typesOrder' => null,
+            'sortAlgorithm' => array(OrderedImportsFixer::SORT_ALPHA, OrderedImportsFixer::SORT_LENGTH),
+        ));
+    }
```
keradus: should be one test method with data provider.

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -22,12 +24,106 @@
...
+
+        $sortAlgorithm = $configuration['sortAlgorithm'];
+        // If no configuration was passed, stick to default.
+        if (null === $sortAlgorithm) {
+            return;
+        }
+
+        // Check if passed sort type is supported.
+        if (!is_string($sortAlgorithm) || !in_array($sortAlgorithm, $this->supportedSortAlgorithms, true)) {
+            throw new InvalidFixerConfigurationException('ordered_imports', sprintf('Sort algorithm is invalid. Should be one of the: "%s"', implode('", "', $this->supportedSortAlgorithms)));
```
keradus: final `.` is missing

DarkaOnLine: Fixed d99cb3d

GrahamCampbell:
Needs rebasing again. Might be an idea to actually "rebase", rather than having merge commits. Will make it much easier going forward.

GrahamCampbell:
image

keradus:
Thank you @DarkaOnLine.

PR #2282, opened 2016-11-10
IntegrationCase is now aware about RuleSet but not Fixers

keradus:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2274#issuecomment-259640406

PR #2285, opened 2016-11-11
IntegrationCase is now aware about RuleSet but not Fixers

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2282
thanks @keradus !


keradus:
Thank you @SpacePossum.

Issue #1476, opened 2015-11-01
[FR] @link to @see fixer


SpacePossum:
PSR 5 `@link` is deprecated in fav. of `@see`.
(https://github.com/phpDocumentor/fig-standards/blob/master/proposed/phpdoc.md#810-link-deprecated)

Would a fixer for this be welcome here?

keradus:
:+1:

GrahamCampbell:
@keradus Yet another reason why we should have the var to type fixer abstract. :trollface:

GrahamCampbell:
We can just use that fixer to fix this case too.

keradus:
No, it is not a reason to create abstraction for `var-type`. An abstraction could be for common functionality, not for different rules. Yep, you could extract common code that replaces tags.

PR #2286, opened 2016-11-11
Phpdoc*Fixer - unify rename fixers

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1476


src/Fixer/Phpdoc/PhpdocNoAliasTagFixer.php
```diff
@@ -0,0 +1,120 @@
...
+        foreach ($configuration as $from => $to) {
+            if (!is_string($from)) {
+                throw new InvalidFixerConfigurationException($this->getName(), 'Tag to replace must be a string.');
+            }
+
+            if (!is_string($to)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Tag to replace to from "%s" must be a string.', $from));
+            }
+
+            if (1 !== preg_match('#^[^\s]+$#', $to) || false !== strpos($to, '*/')) {
```
keradus: sorry, i don't get this line
especially part with phpdoc ending.

why not check is it matchin `[a-zA-Z0-9]` ? spaces are not allowed, but `foo*bar` is not allowed as well.

tests/Fixer/Phpdoc/PhpdocNoAliasTagFixerTest.php
```diff
@@ -0,0 +1,243 @@
...
+            ),
+        );
+    }
+
+    /**
+     * @dataProvider provideVarToTypeCases
+     */
+    public function testVarToTypeFix($expected, $input = null)
+    {
+        $this->getFixer()->configure(array(
```
keradus: same issue here... or maybe we could solve it another way than cloning ?

SpacePossum: fixed by rebasing on top of you fix

keradus:
UPGRADE file needs to be updated

README.rst
```diff
@@ -469,6 +465,10 @@ Choose from the list of available rules:
 * **phpdoc_no_access** [@Symfony]
    | @access annotations should be omitted from phpdocs.
 
+* **phpdoc_no_alias_tag** [@Symfony]
```
GrahamCampbell: not really a good name tbh, given you rename things to their aliases if you want

src/Fixer/Phpdoc/PhpdocNoAliasTagFixer.php
```diff
@@ -0,0 +1,120 @@
...
+ */
+final class PhpdocNoAliasTagFixer extends AbstractFixer
+{
+    /**
+     * @var array<string, string>
+     */
+    private $configuration;
+
+    private static $defaultConfiguration = array(
+        'property-read' => 'property',
```
GrahamCampbell: these are not alias renames for example

SpacePossum: so what should the name be in your opinion?

keradus:
Thank you @SpacePossum.

PR #2287, opened 2016-11-11
NoExtraConsecutiveBlankLinesFixer - fix bug that removes empty line already after scope end

keradus:
Proving the bug from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2282#discussion_r87481482

@GrahamCampbell could you take a look to fix it ?


keradus:
fixed

SpacePossum:
nice find and fix :) thanks @keradus !

PR #2288, opened 2016-11-11
FixerInterface::configure(null) reset fixer to use default configuration

keradus:


keradus:
no need to cloning after this 
cc @SpacePossum @DarkaOnLine

SpacePossum:
awesome, thanks!

PR #2289, opened 2016-11-11
Add PhpdocAddMissingParamAnnotationFixer

keradus:
Very early preview.

Goal: fixer will add `mixed` phpdoc `@param` annotation for parameter that is not yet documented.
`mixed` is not meaningful indeed, but it will point to the user where he should adjust the documentation and change `mixed` to something

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
@@ -1,0 +1,239 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Phpdoc;
+
+use PhpCsFixer\AbstractFunctionReferenceFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\WhitespacesFixerConfigAwareInterface;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class PhpdocAddMissingParamAnnotationFixer extends AbstractFunctionReferenceFixer implements WhitespacesFixerConfigAwareInterface
+{
+    /**
+     * @var array
+     */
+    private $configuration;
+
+    private static $defaultConfiguration = array(
+        'only_untyped' => true,
+    );
+
+    /**
+     * @param null|array<string, bool> $configuration
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $this->configuration = self::$defaultConfiguration;
+
+            return;
+        }
+
+        foreach ($configuration as $key => $value) {
+            if (!array_key_exists($key, self::$defaultConfiguration)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('"%s" is not handled by the fixer.', $key));
+            }
+
+            if (!is_bool($value)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Expected boolean got "%s".', is_object($value) ? get_class($value) : gettype($value)));
+            }
+
+            $configuration[$key] = $value;
+        }
+
+        $this->configuration = $configuration;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return false;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = 0, $limit = $tokens->count(); $index < $limit; ++$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            if (1 === preg_match('/inheritdoc/i', $token->getContent())) {
+                continue;
+            }
+
+            $index = $tokens->getNextMeaningfulToken($index);
+
+            if (null === $index) {
+                return;
+            }
+
+            while ($tokens[$index]->isGivenKind(array(
+                T_ABSTRACT,
+                T_FINAL,
+                T_PRIVATE,
+                T_PROTECTED,
+                T_PUBLIC,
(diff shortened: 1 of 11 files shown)


src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,200 @@
...
+            }
+
+            $lines = $doc->getLines();
+            $lastLine = array_pop($lines);
+
+            preg_match('/^(\s*).*$/', $lastLine->getContent(), $matches);
+            $indent = $matches[1];
+
+            foreach ($arguments as $argument) {
+                if ($this->config['only_untyped'] && '' !== $argument['type']) {
```
keradus: move to separated loop, then we could have early `continue` and no need for splitting/joining lines ;)

keradus: done

keradus:
The main part is there.

What's missing is integration tests (for priority fixes) and apply the tool for codebase (actually, tests folder only)

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,227 @@
...
+            if (null === $index) {
+                return;
+            }
+
+            while ($tokens[$index]->isGivenKind(array(
+                T_PRIVATE,
+                T_PROTECTED,
+                T_PUBLIC,
+                T_STATIC,
+                T_VAR,
```
SpacePossum: `final` and `abstract` are missing?

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,227 @@
...
+
+            if ($token->isGivenKind(T_VARIABLE)) {
+                $sawName = true;
+                $info['name'] = $token->getContent();
+                continue;
+            }
+
+            if ($token->equals('=')) {
+                $sawEq = true;
+                continue;
```
SpacePossum: missing white line above `continue` (same as @ 210)

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,227 @@
...
+            }
+
+            if (!$sawName) {
```
SpacePossum: negative check can be removed if paths swapped

SpacePossum:
nice idea :)

do you think of having the configuration option `non-type-only => true` would be of value, it seems the SF community tends to move that way

keradus:
what do you mean by `non-type-only` ?
`'only_untyped' => true` will make fixer to add annotation for parameters that is not typed in prototype

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,227 @@
...
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = 0, $limit = $tokens->count(); $index < $limit; ++$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            if (1 === preg_match('/inheritdoc/i', $token->getContent())) {
```
keradus: one may not want to run `phpdoc_inline_tag`.

SpacePossum:
I meant the exact thing you described, thanks! : )

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,227 @@
...
+            if (!count($arguments)) {
+                continue;
+            }
+
+            $doc = new DocBlock($token->getContent());
+
+            foreach ($doc->getAnnotationsOfType('param') as $annotation) {
+                $pregMatched = preg_match('/^[^$]+(\$\w+).*$/s', $annotation->getContent(), $matches);
+
+                if (1 !== $pregMatched) {
```
keradus: yes and no. `preg_match` is manipulating input parameter `$matches`, and since I don't want to ever modify any value inside a condition, I put it outside.

So, I treat it like it would be this:
```php
$matches = proper_preg_match(...);
if (error_get_last()) {
    continue;
}
```

keradus:
Thanks for review @SpacePossum , I think it's ready now. Please, could you take a look again ?

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
```diff
@@ -0,0 +1,239 @@
...
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\WhitespacesFixerConfigAwareInterface;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ */
+final class PhpdocAddMissingParamAnnotationFixer extends AbstractFunctionReferenceFixer implements WhitespacesFixerConfigAwareInterface
+{
+    /**
+     * @var array
```
SpacePossum:
```
array<string, bool>
```

src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php
SpacePossum: granted, but one might use `inheritdoc` and not `@inheritdoc`, in that case it should not pass this check right?

tests/Fixtures/Integration/misc/phpdocs.test-out.php
```diff
@@ -0,0 +1,25 @@
...
+     * @var array
+     */
+    public $baz;
+
+    /**
+     * Foo.
+     *
+     * @param int   $fo  this is int
+     * @param float $bar this is float
+     * @param mixed $qux
```
SpacePossum: why is `$baz` not added?

tests/FixerFactoryTest.php
```diff
@@ -289,6 +289,7 @@ public function getFixersPriorityCases()
             array($fixers['no_blank_lines_after_phpdoc'], $fixers['single_blank_line_before_namespace']), // tested also in: no_blank_lines_after_phpdoc,single_blank_line_before_namespace.test
             array($fixers['php_unit_fqcn_annotation'], $fixers['no_unused_imports']), // tested also in: php_unit_fqcn_annotation,unused_use.test
             array($fixers['protected_to_private'], $fixers['ordered_class_elements']), // tested also in: protected_to_private,ordered_class_elements.test
+            array($fixers['phpdoc_add_missing_param_annotation'], $fixers['phpdoc_align']), // tested also in: phpdoc_add_missing_param_annotation,phpdoc_align.test
```
SpacePossum: \+ after `phpdoc_no_alias_tag`, when configured to change some tag to `@param`

SpacePossum:
btw. tests are failing.
Don't we have a fixer to sort `@params` to match the order of params as defined but the function/method signature?

keradus:
Not yet, but this is natural direction after this one is released ;)

PR #2290, opened 2016-11-12
FixCommand - fix progress

keradus:


keradus:
cc @SpacePossum

SpacePossum:
:+1: nice fix!

In another PR I'll prep. some unit tests to test all CLI input with and without a config file input to check the resolver

PR #2291, opened 2016-11-12
Make fixers ready to use directly after creation

keradus:
Currently, if fixer is configurable then it's config claims to be an array, but actually it's a null until someone will call `configure`, so calling `->fix` will crash the execution.. This is bad and after mine change a default configuration is in use directly after instance creation.

Exception is header comment fixer, as the only possible default configuration is to drop the header, which is very dangerous.

### DIFF
--- a/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
+++ b/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
@@ -13,7 +13,6 @@
 namespace PhpCsFixer\Fixer\Phpdoc;
 
 use PhpCsFixer\AbstractFixer;
-use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
 use PhpCsFixer\DocBlock\DocBlock;
 use PhpCsFixer\Tokenizer\Tokens;
 
@@ -34,7 +33,7 @@ final class GeneralPhpdocAnnotationRemoveFixer extends AbstractFixer
     public function configure(array $configuration = null)
     {
         if (null === $configuration) {
-            throw new InvalidFixerConfigurationException($this->getName(), sprintf('Configuration is required.'));
+            $this->configuration = array();
         }
 
         $this->configuration = $configuration;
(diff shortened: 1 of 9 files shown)


src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
```diff
@@ -34,7 +33,7 @@
     public function configure(array $configuration = null)
     {
         if (null === $configuration) {
-            throw new InvalidFixerConfigurationException($this->getName(), sprintf('Configuration is required.'));
+            $this->configuration = array();
```
SpacePossum: this is overriden on the next statement,
did you mean
```
$configuration = array();
```
?

src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
```diff
@@ -63,6 +62,10 @@ public function getPriority()
      */
     public function fix(\SplFileInfo $file, Tokens $tokens)
     {
+        if (!count($this->configuration)) {
```
SpacePossum: I don't like the option to use a fixer without it doing anything because the config makes no sense, it will list in the results that the fixer didn't change anything and not as a `non-useful-configured-fixer`.

src/Fixer/Phpdoc/GeneralPhpdocAnnotationRenameFixer.php
```diff
@@ -34,7 +33,7 @@
     public function configure(array $configuration = null)
     {
         if (null === $configuration) {
-            throw new InvalidFixerConfigurationException($this->getName(), sprintf('Configuration is required.'));
+            $this->configuration = array();
```
SpacePossum: same, `$this->configuration` will still end up being `null` ;)

src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
keradus: missing return statement, thanks

PR #2292, opened 2016-11-12
GeneralPhpdocAnnotation*Fixer::configure - add missing return

keradus:
fix #2291
thanks @SpacePossum


SpacePossum:
:+1: that was fast : D

PR #2296, opened 2016-11-13
ProjectCodeTest - make test part of regular testsuite, not standalone one

keradus:


tests/ProjectCodeTest.php
```diff
@@ -25,73 +26,70 @@
...
 
-        foreach ($finder as $file) {
-            include_once $file;
-        }
+        $files = array_map(
+            function (SplFileInfo $file) {
+                return sprintf(
+                    '%s\\%s%s%s',
+                    __NAMESPACE__,
+                    strtr($file->getRelativePath(), DIRECTORY_SEPARATOR, "\\"),
```
GrahamCampbell: incorrect quotes used

keradus: no need for manually reporting things that are reported by Travis ;)

PR #2297, opened 2016-11-13
Use phpdbg instead of xdebug

keradus:
Before: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/175546739
After: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/175568326

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,14 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
-    # XDebug preparation:
-    # - for tag building for release we don't need to collect code coverage, let us turn off xdebug completely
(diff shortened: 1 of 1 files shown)


SpacePossum:
nice idea :)
btw. why `TASK_TESTS` this is always set and on `TASK_SCA` do we detect `null`-able vs. `void` as well?

keradus:
> TASK_TESTS

yes, it's always set, but as one of our task type, it could be disable on any env in the future


> TASK_SCA

no, currently it is just checking run with declare...strict... (that's why we need utests when we do SCA),
not easy to check null vs void when there is no return type in prototype at source.

PR #2298, opened 2016-11-14
ConfigurationResolver - grooming

SpacePossum:
- resolve differ in configuration resolver, not fin fix command
- configuration resolver return `null` for cache file if caching is disabled
- add test to test command line args. override default config
- add test to test config file overrides default config
- typo in readme
- configuration resolver; sort method names


src/Console/Command/FixCommand.php
```diff
@@ -87,7 +87,7 @@ public function __construct()
     }
 
     /**
-     * {@inheritdoc}
+     * @see Command
```
keradus: please revert

keradus:
:+1: if green

SpacePossum:
Travis timeout on 7, other versions all green (including 7.1)

keradus:
Thank you @SpacePossum.

PR #2300, opened 2016-11-14
Simplify rule set

SpacePossum:


.php_cs.dist
```diff
@@ -13,19 +13,17 @@ EOF;
...
         'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block'),
         'no_useless_else' => true,
         'no_useless_return' => true,
         'ordered_class_elements' => true,
         'ordered_imports' => true,
-        'php_unit_construct' => true,
-        'php_unit_dedicate_assert' => true,
         'php_unit_strict' => true,
         'psr4' => true,
+        'silenced_deprecation_error' => false,
```
keradus: actually we silence them as well, no ?

SpacePossum: indeed we do, I searched in master.... ; )

keradus:
So, now we are getting random hhvm failures ? new hhvm update on Travis env or what ?

SpacePossum:
->
https://github.com/sebastianbergmann/object-enumerator/blob/1.0.0/src/Enumerator.php#L47
->
https://github.com/sebastianbergmann/recursion-context/blob/master/src/Context.php#L141

maybe that last ` $end[0]` access is a problem as it is not checked?

keradus:
:+1:

src/RuleSet.php
```diff
@@ -53,6 +53,7 @@
         ),
         '@Symfony' => array(
             '@PSR2' => true,
+            'array_syntax' => array('syntax' => 'long'),
```
keradus: I was thinking about it. Symfony keeps that rule only for framework project to not have merging hell with old branches that supported 5.3.3.

keradus:
Thank you @SpacePossum.

GrahamCampbell:
The symfony ruleset is still wrong.  Not all symfony projects do this. Only selected ones like the main framework.

SpacePossum:
>The symfony ruleset is still wrong

wrong how? It was not changed in this PR

GrahamCampbell:
Sorry, wrong PR. :P

Issue #2302, opened 2016-11-15
[2.x] ClassDefinitionFixer - issues with anonymous classes


gharlan:
Input:
```php
new class($this->prop) {
};
```

Expected: no change

Result:
```php
new class($this -> prop)
{
};
```

* spaces around `->` are wrong
* opening brace is wrong -> this is fixed by `BracesFixer`, but IMO the `ClassDefinitionFixer` should no put the opening brace to a newline for anonymous classes

keradus:
`php-cs-fixer --version` ?

gharlan:
`PHP CS Fixer version 2.0.0-DEV by Fabien Potencier and Dariusz Ruminski`

current master

keradus:
thanks !
could you work on a fix please ? :)

SpacePossum:
is this also wrong on 1.x line?

SpacePossum:
arrow fixing is wrong on 1.13 as well, braces seem ok

keradus:
#2303 has been merged.
is there sth left to be done here ?

gharlan:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2303/files#r88790767

PR #2303, opened 2016-11-15
ClassDefinitionFixer - Anonymous classes fixing

SpacePossum:
ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2302

### DIFF
--- a/Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
+++ b/Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
@@ -154,6 +154,14 @@ final class ClassDefinitionFixerTest extends AbstractFixerTestBase
                 "<?php \$a = new class(10) extends SomeClass implements SomeInterface, D\n{};",
                 "<?php \$a = new    class(10)     extends\nSomeClass\timplements    SomeInterface, D {};",
             ),
+            array(
+                "<?php \$a = new class(\$this->prop)\n{};",
+                '<?php $a = new class(   $this->prop   ){};',
+            ),
+            array(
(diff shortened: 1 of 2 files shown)


keradus:
Thank you @SpacePossum.

Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
```diff
@@ -154,6 +154,14 @@ public function provideAnonymousClassesCases()
                 "<?php \$a = new class(10) extends SomeClass implements SomeInterface, D\n{};",
                 "<?php \$a = new    class(10)     extends\nSomeClass\timplements    SomeInterface, D {};",
             ),
+            array(
+                "<?php \$a = new class(\$this->prop)\n{};",
+                '<?php $a = new class(   $this->prop   ){};',
```
gharlan: IMO the fixer should not add the `\n` before `{` for anonymous classes.

gharlan:
thank you @SpacePossum!

Symfony/CS/Tests/Fixer/PSR2/ClassDefinitionFixerTest.php
keradus: I agree. It would follow PSR-2 anonymous functions standards + PSR-12 anonymous classes standards

SpacePossum: Stopping the fixer to do that should be relative issue, 
changing it to format to PSR 12 (including multiple line support with `use` case) will take some work.
Hope to look into this soon.

PR #2304, opened 2016-11-15
DX: use PHPMD

keradus:


src/StdinFileInfo.php
```diff
@@ -46,6 +46,8 @@ public function getATime()
 
     public function getBasename($suffix = null)
     {
+        unset($suffix);
```
SpacePossum: what is the added value in unsetting these params?

keradus: PHPMD complaining about unused parameter without that. And actually thanks to that rule we removed few parameters that are not in use indeed. Trade off.

keradus:
OK, whatever I will do it will crash on 5.3. I guess we need to wait with that for dropping of 5.3

src/StdinFileInfo.php
keradus: you could put annotation for a code to ignore some rule in some line, but it's putting a tool-related comments into code, which is also not the best ;)

SpacePossum: Trying out the whitelist might work, I did setup it up for a big project. Its like with code coverage, you want to know if things will get worse when a PR is merged, not if the project is in perfect spot.
The comment thing might not be the best either, but it would choose it above no-ops ;)

SpacePossum:
You dislike `switch` that much : D
Anyway, all good. You sure you don't want to add more rules like `GotoStatement` and such?

.travis.yml
```diff
@@ -68,9 +68,14 @@ install:
     - travis_retry composer update $DEFAULT_COMPOSER_FLAGS $COMPOSER_FLAGS
     - composer info -D | sort
 
+before_script:
+    - if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then COMMIT_RANGE=$TRAVIS_COMMIT_RANGE; else COMMIT_RANGE="HEAD~..HEAD"; fi;
+    - if [ $TASK_SCA == 1 ]; then export COMMIT_SCA_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`; fi
+
 script:
+    - if [ $TASK_SCA == 1 ] && [ -n "$COMMIT_SCA_FILES" ]; then vendor/bin/phpmd `echo $COMMIT_SCA_FILES | sed 's/ /,/g'` text phpmd.xml --exclude src/Resources,tests/Fixtures; fi;
```
julienfalque: You should mention this in the CONTRIBUTING.md file.

keradus:
> You dislike switch that much : D

> I'm not a big fan of replacing switches with array maps but this looks like a nice addition :)

for me it's combination of `goto` with `==` (when you cannot enforce using `===`). see the issues we had recently with `case false` and `null` value

.travis.yml
keradus: I am thinking about starting having some proper task runner...

keradus:
> Anyway, all good. You sure you don't want to add more rules like `GotoStatement` and such?

feel free to propose (eg just a list) more of them.

I remember that I have locally few more rules that could be applied on 2.3 branch (due to dropped old PHP versions)

.travis.yml
julienfalque: Something like `make` would be great indeed :)

keradus: I always try to go with non-unix only solution ;)
anyway, definietly not in this PR.

SpacePossum:
all good, however the tests fail because of missing env. var to ignore the PHP constraint.

src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -94,38 +94,6 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     }
 
     /**
-     * @param Tokens $tokens
-     * @param int    $endIndex
-     */
-    private function fixWhiteSpaceAroundSequence(Tokens $tokens, $endIndex)
```
keradus: unused method :D

src/Console/ConfigurationResolver.php
```diff
@@ -253,7 +249,21 @@ public function getConfigFile()
     public function getDiffer()
     {
         if (null === $this->differ) {
-            $this->differ = false === $this->options['diff'] ? new NullDiffer() : new SebastianBergmannDiffer();
+            $mapper = array(
+                'null' => function () { return new NullDiffer(); },
+                'sbd' => function () { return new SebastianBergmannDiffer(); },
+            );
```
keradus: short one is missing here

keradus: actually no, it was reverted on 2.2 line

PR #2306, opened 2016-11-17
DeclareStrictTypesFixer - do not move tokens

SpacePossum:
This is step one to resolve the "file opening fixing" conflict we have with the current fixers.
With this PR the DeclareStrictTypesFixer will:
- insert the sequence if not found, and fix the white space around the new sequence
- fix the casing of 'strict_types=1' (as we have no other fixer for it, nor would it be useful ATM to split this functionality IMO)

It will no longer:
- move an existing sequence
- fix white around an existing sequence or within such sequence
- fix the code style of an existing sequence (with a small exception, see above)

So with a bit of pain in my heart.. here we go  ;)


keradus:
> move an existing sequence

agree here

> fix white around an existing sequence or within such sequence
> fix the code style of an existing sequence (with a small exception, see above)

but why drop that ?

keradus:
Thank you @SpacePossum.

PR #2307, opened 2016-11-17
Update .editorconfig

SpacePossum:


keradus:
:+1:

.editorconfig
```diff
@@ -5,3 +5,9 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[/Symfony/CS/Tests/Fixtures/**/*]
+end_of_line = ignore
+insert_final_newline = ignore
+indent_style = ignore
+trim_trailing_whitespace = ignore
```
keradus: does `ignore` config value a valid one?
I cannot find it https://github.com/editorconfig/editorconfig/search?utf8=✓&q=ignore

localheinz: Maybe this helps: http://stackoverflow.com/a/30539967/1172545?

localheinz: What I meant is to create an `.editorconfig` with the following content

```yml
root = true
```

in [`/Symfony/CS/Tests/Fixtures`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/1.13/Symfony/CS/Tests/Fixtures) - maybe that helps?

SpacePossum: >does ignore config value a valid one?

keradus: so, would "official way" works for us ?

keradus:
Thank you @SpacePossum.

PR #2312, opened 2016-11-18
RuleSet - sort rules

localheinz:
This PR

* [x] asserts rules are sorted properly in set definitions
* [x] keeps rules sorted in `RuleSet` and wraps their configuration (if available), as well as keeping that sorted, too

💁‍♂️ There's a small PhpStorm plugin (see https://plugins.jetbrains.com/plugin/5919) that helps doing that, and keeping an array sorted makes it easier to find items - what do you think?


SpacePossum:
LGTM :)
With 2.x not being released yet this shouldn't cause merging issue I think, so why not.
Should we add a test to make sure the sets will stay sorted or what it be a bit of an overkill?

keradus:
actually, the current, master form is kinda sorted if you would follow original names, like the content was not reordered when one changed the name of fixer to not have issues with merging 1.x->2.x

keradus:
Thank you @localheinz.

PR #2314, opened 2016-11-19
no_empty_comment: preserve paragraphs

pschultz:
Make NoEmptyCommentFixer preserve paragraphs in a sequence of single
line comments. E.g. do not change comments like the following

    // Introductory statement
    //
    // Lengthy description goes here.
    //
    // Descriptions may also be formatted as multiple paragraphs.

For all intents and purposes, this is a single comment that is clearly
not empty. It should not be changed into multiple comments that are
separated by blank lines.

### DIFF
--- a/Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
+++ b/Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
@@ -134,12 +134,33 @@ final class NoEmptyCommentFixer extends AbstractFixer
         $tokens->overrideAt($index, array(T_WHITESPACE, "\n", $tokens[$index]->getLine()));
     }
 
+    /**
+     * isSurroundedBySingleLineComments checks whether the preceding and
+     * succeeding lines of the token at $index contain a single-line comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
     private function isSurroundedBySingleLineComments(Tokens $tokens, $index)
     {
         return $this->hasSingleLineCommentSibling($tokens, $index, -1) &&
                $this->hasSingleLineCommentSibling($tokens, $index,  1);
     }
 
+    /**
+     * hasSingleLineCommentSibling checks whether the preceding or succeeding
+     * line of the token at $index contains a single-line comment. Which line
+     * is checked depends on $direction.
+     *
+     * @param Tokens $tokens
+     * @param int    $index
+     * @param int    $direction either -1 to check the preceding line, or +1 to
+     *                          check the succeeding line
+     *
+     * @return bool
+     */
     private function hasSingleLineCommentSibling(Tokens $tokens, $index, $direction)
     {
         do {
--- a/Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
+++ b/Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
@@ -41,7 +41,7 @@ final class NoEmptyCommentFixer extends AbstractFixer
      */
     public function getDescription()
     {
-        return 'There should not be an empty comments.';
+        return 'There should not be an empty comment.';
     }
 
     /**
@@ -64,6 +64,10 @@ final class NoEmptyCommentFixer extends AbstractFixer
         // single line comment starting with '#'
         if ('#' === $content[0]) {
             if (preg_match('|^#\s*$|', $content)) {
+                if ($this->isSurroundedBySingleLineComments($tokens, $index)) {
+                    return;
+                }
+
                 $this->clearCommentToken($tokens, $index);
             }
 
@@ -73,6 +77,10 @@ final class NoEmptyCommentFixer extends AbstractFixer
         // single line comment starting with '//'
         if ('/' === $content[1]) {
             if (preg_match('|^//\s*$|', $content)) {
+                if ($this->isSurroundedBySingleLineComments($tokens, $index)) {
+                    return;
+                }
+
                 $this->clearCommentToken($tokens, $index);
             }
 
@@ -124,5 +132,42 @@ final class NoEmptyCommentFixer extends AbstractFixer
         // else
         // override with whitespace token linebreak
         $tokens->overrideAt($index, array(T_WHITESPACE, "\n", $tokens[$index]->getLine()));
+    }
+
+    private function isSurroundedBySingleLineComments(Tokens $tokens, $index)
+    {
+        $line = $tokens[$index]->getLine();
+        $prev = $tokens->getPrevNonWhitespace($index);
+        $next = $tokens->getNextNonWhitespace($index);
+
+        return $this->isSingleLineCommentOnLine($tokens, $prev, $line - 1) &&
+               $this->isSingleLineCommentOnLine($tokens, $next, $line + 1);
+    }
+
+    private function isSingleLineCommentOnLine(Tokens $tokens, $index, $line)
+    {
+        if ($index === null) {
+            return false;
+        }
+
+        $token = $tokens[$index];
+
+        if (false === $token->isComment()) {
+            return false;
+        }
+
+        if ($line !== $token->getLine()) {
+            return false;
+        }
+
+        if ($token->getContent()[0] === '#') {
+            return true;
+        }
+
+        if ($token->getContent()[1] === '/') {
+            return true;
+        }
+
+        return false;
     }
 }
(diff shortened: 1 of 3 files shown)


Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -125,4 +133,41 @@ private function clearCommentToken(Tokens $tokens, $index)
...
+            return false;
+        }
+
+        $token = $tokens[$index];
+
+        if (false === $token->isComment()) {
+            return false;
+        }
+
+        if ($line !== $token->getLine()) {
```
SpacePossum: please don't rely on this method, is removed on master since it does not work as expected

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -125,4 +133,41 @@ private function clearCommentToken(Tokens $tokens, $index)
...
+            return false;
+        }
+
+        if ($token->getContent()[0] === '#') {
+            return true;
+        }
+
+        if ($token->getContent()[1] === '/') {
```
SpacePossum: this cannot be done on PHP 5.3 and please use Yoda style

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -125,4 +133,41 @@ private function clearCommentToken(Tokens $tokens, $index)
         // override with whitespace token linebreak
         $tokens->overrideAt($index, array(T_WHITESPACE, "\n", $tokens[$index]->getLine()));
     }
+
+    private function isSurroundedBySingleLineComments(Tokens $tokens, $index)
+    {
+        $line = $tokens[$index]->getLine();
+        $prev = $tokens->getPrevNonWhitespace($index);
```
SpacePossum: please test previous (or next) line first and early return if possible, than check the other line if still needed

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -41,7 +41,7 @@ public function fix(\SplFileInfo $file, $content)
      */
     public function getDescription()
     {
-        return 'There should not be an empty comments.';
+        return 'There should not be an empty comment.';
```
SpacePossum: please revert (together with the README), the fixer will still fix all comments that are applicable

SpacePossum:
Hi!
I like the idea, however I have some questions about how you build it :)

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
pschultz: I don't understand. This is just a grammar fix. Do you prefer the plural?

> There should not be any empty comments.

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
pschultz: I don't know how else to do this. It is important that comments be on the immediate next or previous line. Do you have a specific solution in mind that doesn't require `getLine()`?

By the way, `NoEmptyCommentFixer` has been rewritten on master, so this patch won't apply anyway.

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
pschultz: Fixed in 32a3bd4

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
SpacePossum: I prefer the plural :)

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
pschultz: Turns out that the line number checks are not necessary at all. I've added tests to prove that in 1291ec9.

pschultz: I guess I spoke too soon. Looking at the test failure now.

pschultz: Fixed in 8fbd89a

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -125,4 +133,29 @@ private function clearCommentToken(Tokens $tokens, $index)
         // override with whitespace token linebreak
         $tokens->overrideAt($index, array(T_WHITESPACE, "\n", $tokens[$index]->getLine()));
     }
+
+    private function isSurroundedBySingleLineComments(Tokens $tokens, $index)
```
pschultz: You meant the doc comment, I asssume. Added in 75cddc3.

Symfony/CS/Fixer/Symfony/NoEmptyCommentFixer.php
```diff
@@ -125,4 +133,50 @@ private function clearCommentToken(Tokens $tokens, $index)
...
+     */
+    private function hasSingleLineCommentSibling(Tokens $tokens, $index, $direction)
+    {
+        do {
+            $index += $direction;
+            if (false === isset($tokens[$index])) {
+                return false;
+            }
+
+            $token = $tokens[$index];
```
SpacePossum:
```php
if (!$token->isWhitespace() && !$token->isComment()) {
    return false;
}
```

keradus:
any news here ?

pschultz:
I'm really busy at work. I hope to get to this later today.

GrahamCampbell:
Maybe this is worth targeting at master now, because it's kinda a feature, right?

GrahamCampbell:
Ping @pschultz. :)

SpacePossum:
@pschultz Hi I'm closing this as your initial report seems to be resolved in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1946

Your PR also is complementary as it would fix:
```php
/**
 *
 * a
 *
 * b
 * 
 */
```

to

```php
/**
 * a
 *
 * b 
 */
```


which is currently not done.
If you would like the first and last "empty" of a comment to be removed as well, please let me know than we can work on that.
Thanks for reporting and even more for putting time and effort in this PR! : )

SpacePossum:
Hi @pschultz 
Your PR targets the 1.x line which we no longer maintain. I looked at your changes and compared those against the current working of the fixer. The paragraphs you like to preserve as your sample in the initial description of your PR are now already preserved, this is because of this [PR](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1946) which has been merged on the 2.x line.
The difference in functionality between your proposal and the current state is the following;
input
```php
# 
# a
#
# b
#

#
```
your way of fixing fixes to:
```php
# a
#
# b
```
current way of fixing
```php
#
# a
#
# b
#
```

I'm closing this now because of the wrong target branch and duplicate logic in your PR compared to the current state. 
Please let me know if you plan on a PR for the 2.x line to make it possible to remove the first and last empty line of a comment block. Happy to help with that!
Thanks for your PR, idea and putting time and thoughts into it, much appreciated!

pschultz:
We don't plan on updating to 2.x in the foreseeable future, so I won't be working on that.

SpacePossum:
Thanks for letting me know, I'll backlog the enhancement :)
 If at some point you want help with upgrading let me know, happy to help out

keradus:
btw @pschultz , why not to upgrade? lack of time or some other issue ?

pschultz:
1.x and 2.x are incompatible in many ways (CLI, config file) and we don't have the capacity do find a good way to upgrade.

keradus:
did you try out upgrade guide ?
if so and still have problems, show your config file and CLI command you are currently using ;)

pschultz:
The point is that there's one php-cs-fixer "service" running, fixing many independent PHP repositories, each with their own config file. We'll have to build something that supports both versions at the same time, since it's not feasible to update all PHP repos at once.

keradus:
had that case one. on your CI service you could have different workers - like `PHP CS Fixer v1` and `PHP CS Fixer v2`. but then you need to adjust the connection to specify which repo <=> which version of tool.
If you want, you could do auto-discovery of that via config file - `.php_cs` for v1, `.php_cs.dist` for v2.

Then, you update repos one by one - each time adjusting the config file name.

Hope it will help ;)

PR #2315, opened 2016-11-19
Token - Deprecate getLine()

SpacePossum:


Symfony/CS/Tokenizer/Token.php
```diff
@@ -205,6 +205,8 @@ public function getId()
     }
 
     /**
+     * @deprecated Will be removed in the 2.0
```
keradus: First line should be summary, not annotation, that's why Travis is failing

SpacePossum: what if your summary is an annotation? ;)

keradus:
Thank you @SpacePossum.

PR #2318, opened 2016-11-20
*TestCase - Reduce visibility of setUp()

localheinz:
This PR

* [x] reduces the visibility of `setUp()` from `public` to `protected`

💁‍♂️ This is the visibility as declared on the parent - unless I am missing something, there's probably no need to increase it.

### DIFF
--- a/src/Test/AbstractFixerTestCase.php
+++ b/src/Test/AbstractFixerTestCase.php
@@ -42,7 +42,7 @@ abstract class AbstractFixerTestCase extends \PHPUnit_Framework_TestCase
      */
     private $fixerClassName;
 
-    public function setUp()
+    protected function setUp()
     {
         $this->linter = $this->getLinter();
         $this->fixer = $this->createFixer();
--- a/src/Test/AbstractIntegrationTestCase.php
+++ b/src/Test/AbstractIntegrationTestCase.php
@@ -90,7 +90,7 @@ abstract class AbstractIntegrationTestCase extends \PHPUnit_Framework_TestCase
         self::$fileRemoval->delete($tmpFile);
     }
 
-    public function setUp()
+    protected function setUp()
     {
         $this->linter = $this->getLinter();
     }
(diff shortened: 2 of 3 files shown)


keradus:
good idea. please do it globally:

```bash
D:\GIT\fork\PHP-CS-Fixer (master)
λ grep -r "function setUp" src tests | grep "public"
src/Test/AbstractFixerTestCase.php:    public function setUp()
src/Test/AbstractIntegrationTestCase.php:    public static function setUpBeforeClass()
src/Test/AbstractIntegrationTestCase.php:    public function setUp()
tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php:    public static function setUpBeforeClass() {}
tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php:    public function setUp() {}
tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php:    public static function setUpBeforeClass() {}
tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php:    public function setUp() {}
tests/Fixer/ControlStructure/NoUnneededControlParenthesesFixerTest.php:    public static function setUpBeforeClass()

D:\GIT\fork\PHP-CS-Fixer (master)
λ grep -r "function tearDown" src tests | grep "public"
src/Test/AbstractIntegrationTestCase.php:    public static function tearDownAfterClass()
```

localheinz:
@keradus 

The methods `setupBeforeClass()` and `tearDownAfterClass()` also have `public` visibility on the parent class, see 

* [`PHPUnit_Framework_TestCase::setupBeforeClass()`](https://github.com/sebastianbergmann/phpunit/blob/5.6/src/Framework/TestCase.php#L2249-L2256)
* [`PHPUnit_Framework_TestCase::tearDownAfterClass()`](https://github.com/sebastianbergmann/phpunit/blob/5.6/src/Framework/TestCase.php#L2298-L2305)

The other examples are from test case code examples, and there's no clear relation to the code examples extending `PHPUnit_Framework_TestCase` - shall I change them anyway?

keradus:
oh damn
sure, in that case please change it in fixture as well, so I won't get that false positive next time ;)

SpacePossum:
:+1:

keradus:
Thank you @localheinz.

PR #2320, opened 2016-11-21
Clear up status code on 1.x

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2231

PR #2321, opened 2016-11-21
Add exit status 1 to doc.

SpacePossum:


keradus:
:+1:

README.rst
```diff
@@ -762,12 +762,13 @@ Exit codes
...
 Exit code is build using following bit flags:
 
-*  0 OK
-*  4 Some files have invalid syntax (only in dry-run mode)
-*  8 Some files need fixing (only in dry-run mode)
-* 16 Configuration error of the application
-* 32 Configuration error of a Fixer
-* 64 Exception raised within the application
+*  0 OK.
+*  1 PHP/HHVM minimal requirement not matched.
```
SpacePossum: We've to check the error/exception handling in the Application class. IIRC Symfony will use the exception code when existing an uncaught exception. Did you spot it?

keradus: exit status `1` (and 2, and 128+) has special meaning, that's why we don't use it inside application since v2.
`Catchall for general errors`.
if error occurs in our app - we provide proper error.
if sth goes really wrong - we use `1`.

Maybe sth like `general error occur, eg pHP/HHVM ... not matched` ?

SpacePossum: >if sth goes really wrong - we use 1.

maybe leave that as is, and use a diff. code for environment requirement not match, i.e. change this https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/php-cs-fixer#L22 ?

keradus:
Thank you @SpacePossum , used during merging #2320 into master !

PR #2327, opened 2016-11-25
HeaderCommentFixer - Do not allow invalid header to be configured.

SpacePossum:
This should not be used as header input:
```
'header' => '*/ $a; /**',
```

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -39,8 +39,8 @@ final class HeaderCommentFixer extends AbstractFixer implements WhitespacesFixer
     /** @var string */
     private $headerComment;
 
-    /** @var int */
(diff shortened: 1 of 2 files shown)


SpacePossum:
anyone got time for a review?

keradus:
general question:
we support
```
/**
 * Foo.
 */
```

and

```
/*
 * Foo.
 */
```.

Is
```
// Foo.
```
a proper header (taking PSR-5 into consideration) ?
if so, then when we will want to introduce possibility to use `//` type then your solution will be problematic

SpacePossum:
Fair enough, I could update to test that the configured header exists out of comments and white space only, where the header must start and end with a comment. This will still leave it open for possible conflicts (multiple blank lines, trailing white space etc.), but that would be user/config error. WDYT?

SpacePossum:
I'll try something else.

PR #2332, opened 2016-11-28
New colours of diff on console

keradus:
Before:
before

----------------------------------------

After:
after


SpacePossum:
:+1:

PR #2335, opened 2016-11-29
don't register self-update command when running non-phar distribution

remicollet:
Just because this may looks strange:

```
$ php-cs-fixer list
  ...
  self-update  Update php-cs-fixer.phar to the latest stable version.
  selfupdate   Update php-cs-fixer.phar to the latest stable version.

$ php-cs-fixer self-update
Self-update is available only for PHAR version.
```

### DIFF
--- a/Symfony/CS/Console/Application.php
+++ b/Symfony/CS/Console/Application.php
@@ -17,6 +17,7 @@ use Symfony\CS\Console\Command\FixCommand;
 use Symfony\CS\Console\Command\ReadmeCommand;
 use Symfony\CS\Console\Command\SelfUpdateCommand;
 use Symfony\CS\Fixer;
+use Symfony\CS\ToolInfo;
 
 /**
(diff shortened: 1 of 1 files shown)


keradus:
From https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/710#issuecomment-62793466 :
> If one run self-update command on non-phar version he should get info that this action is allowed only with phar version.
Otherwise one who read README file and want to simply run self-update as described will get no command found.

Big :-1: for having different commands depending on installation type

remicollet:
BTW, composer or PHPUnit already do this way.

SpacePossum:
>Self-update is available only for PHAR version.

I like this message, it gives some info about what or why.

keradus:
Closing due to discussion and lack of further communication

GrahamCampbell:
I agree that conditional command registration seems like a hack.

PR #2337, opened 2016-11-29
Normalize rule naming

keradus:


src/ConfigInterface.php
```diff
@@ -105,6 +105,8 @@ public function getUsingCache();
     /**
      * Adds a suite of custom fixers.
      *
+     * Name of custom fixer should follow `VendorName/rule_name` convention.
```
keradus: This is the expected pattern

src/FixerNameValidator.php
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @return bool
+     */
+    public function isValid($name, $isCustom)
+    {
+        if (!$isCustom) {
+            return 1 === preg_match('/^[a-z][a-z0-9_]*$/', $name);
+        }
+
+        return 1 === preg_match('/^[A-Z][a-zA-Z0-9]*\/[a-z][a-z0-9_]*$/', $name);
```
SpacePossum: why not allow vendor names to start with lower case char.?

keradus: 1. there must must be unique separator for vendor and rule. in your example, vendor was `gecko-packages/cs` or `gecko-packages`? nobody know
2. I want to avoid mixed experience, when sometimes rule is following that conversion, and sometimes another. Reusing patterns like `PascalCase/snake_case` looks good to me,

keradus: I would really want to not give multiple options, for not providing different experience for users.
So either `PascalCase` or `snake_case` for a vendor. Not both of them at same time.

Also, our current set naming is PascalCase (with optional `:risky` label).
Here, for vendor, I choose PascalCase to have this in future: `CompanyName/@CorpoRules`

SpacePossum: alright, let see for a bit what the community thinks :)

PR #2338, opened 2016-11-29
Drop hacks for unsupported HHVM

keradus:


SpacePossum:
:+1:  
on a side note, with non of the release phars working on HHVM do we want to keep supporting it all?

keradus:
For now I would keep it with minimal support.
If it's installable via composer but not phar - this is problem of hhvm and we won't fix it.
If feature is working everywhere  but not on hhvm due to hhvm bugs - let us try to disable feature for hhvm only.
Also, I'm intrested in keeping that minimal support only for newest hhvm we could check at CI.

PR #2339, opened 2016-11-29
Add some Fixer descriptions

SpacePossum:
replaces: https://github.com/keradus/PHP-CS-Fixer/pull/22

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -587,7 +587,7 @@ Choose from the list of available rules:
    | Replace all <> with !=.
 
 * **strict_comparison**
-   | Comparison should be strict.
+   | Comparisons should be strict.
    | *Rule is: risky.*
 
 * **strict_param**
--- a/README.rst
+++ b/README.rst
@@ -556,7 +556,7 @@ Choose from the list of available rules:
    | *Rule is: risky.*
 
 * **simplified_null_return**
-   | A return statement wishing to return nothing should be simply "return".
+   | A return statement wishing to return "void" should not return "null".
    | *Rule is: risky.*
 
 * **single_blank_line_at_eof** [@PSR2, @Symfony]
--- a/README.rst
+++ b/README.rst
@@ -175,7 +175,8 @@ automatically fix anything:
 Choose from the list of available rules:
 
 * **array_syntax**
-   | PHP arrays should be declared using the configured syntax.
+   | PHP arrays should be declared using the configured syntax (requires PHP
+   | >= 5.4 for short syntax).
    | *Rule is: configurable.*
 
 * **binary_operator_spaces** [@Symfony]
@@ -218,7 +219,7 @@ Choose from the list of available rules:
    | Equal sign in declare statement should not be surrounded by spaces.
 
 * **declare_strict_types**
-   | Force strict types declaration in all files.
+   | Force strict types declaration in all files. Requires PHP >= 7.0.
    | *Rule is: risky.*
 
 * **dir_constant**
@@ -512,7 +513,7 @@ Choose from the list of available rules:
    | @var and @type annotations should not contain the variable name.
 
 * **pow_to_exponentiation** [@PHP56Migration, @PHP70Migration, @PHP71Migration]
-   | Converts 'pow()' to '**'.
+   | Converts 'pow()' to '**' operator. Requires PHP > 5.6.
    | *Rule is: risky.*
 
 * **pre_increment** [@Symfony]
--- a/README.rst
+++ b/README.rst
@@ -513,7 +513,7 @@ Choose from the list of available rules:
    | @var and @type annotations should not contain the variable name.
 
 * **pow_to_exponentiation** [@PHP56Migration, @PHP70Migration, @PHP71Migration]
-   | Converts 'pow()' to '**' operator. Requires PHP > 5.6.
+   | Converts 'pow()' to '**' operator. Requires PHP >= 5.6.
    | *Rule is: risky.*
 
 * **pre_increment** [@Symfony]
--- a/src/Fixer/Alias/PowToExponentiationFixer.php
+++ b/src/Fixer/Alias/PowToExponentiationFixer.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer\Fixer\Alias;
 
 use PhpCsFixer\AbstractFunctionReferenceFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
@@ -70,6 +72,21 @@ final class PowToExponentiationFixer extends AbstractFunctionReferenceFixer
     /**
      * {@inheritdoc}
      */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Converts \'pow()\' to \'**\' operator. Requires PHP > 5.6.',
+            null,
+            array(new CodeSample("<?php\n pow(\$a, 1);")),
+            null,
+            null,
+            'Risky when the function \'pow()\' function is overridden.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getPriority()
     {
         // must be run before BinaryOperatorSpacesFixer
@@ -79,9 +96,9 @@ final class PowToExponentiationFixer extends AbstractFunctionReferenceFixer
     /**
      * {@inheritdoc}
      */
-    protected function getDescription()
+    public function isRisky()
     {
-        return 'Converts \'pow()\' to \'**\'.';
+        return true;
     }
 
     /**
--- a/src/Fixer/Alias/PowToExponentiationFixer.php
+++ b/src/Fixer/Alias/PowToExponentiationFixer.php
@@ -14,7 +14,7 @@ namespace PhpCsFixer\Fixer\Alias;
 
 use PhpCsFixer\AbstractFunctionReferenceFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
-use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\ShortFixerDefinition;
 use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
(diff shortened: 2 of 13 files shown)


src/Fixer/Alias/PowToExponentiationFixer.php
```diff
@@ -70,6 +72,21 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
      */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Converts \'pow()\' to \'**\'.',
+            null,
+            array(new CodeSample("<?php\n pow(\$a, 1);")),
+            null,
+            null,
+            'Risky when the function \'pow()\' function is overloaded.'
```
keradus: overloaded function is when a function has multiple possible interfaces.
like `pow(int):int` and `pow(float):float`
it's not supported by PHP language syntax.

here, you meant that function is overridden, so one has replaced default implementation of it.

src/Fixer/ArrayNotation/ArraySyntaxFixer.php
```diff
@@ -80,26 +82,41 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
      * {@inheritdoc}
      */
-    public function getPriority()
+    public function getDefinition()
     {
-        // should be run before the BinaryOperatorSpacesFixer and TernaryOperatorSpacesFixer.
-        return 1;
+        return new FixerDefinition(
+            'PHP arrays should be declared using the configured syntax.',
+            null,
```
keradus: maybe put here that Fixer requires 5.4 to works ?

src/Fixer/Alias/PowToExponentiationFixer.php
```diff
@@ -70,6 +72,21 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
     /**
      * {@inheritdoc}
      */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Converts \'pow()\' to \'**\'.',
```
keradus: to ** operator

src/Fixer/CastNotation/ShortScalarCastFixer.php
```diff
@@ -60,8 +54,20 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
      * {@inheritdoc}
      */
-    protected function getDescription()
+    public function getDefinition()
     {
-        return 'Cast "(boolean)" and "(integer)" should be written as "(bool)" and "(int)". "(double)" and "(real)" as "(float)".';
+        return new FixerDefinition(
+            'Cast "(boolean)" and "(integer)" should be written as "(bool)" and "(int)". "(double)" and "(real)" as "(float)".',
+            null,
+            array(new CodeSample("<?php\n\$a = (boolean) \$b; \$a = (integer) \$b; \$a = (boolean) \$b; \$a = (double) \$b;"))
```
keradus: duplicated `boolean`, missing `float`.
i would split them to be followed line by line

src/Fixer/ControlStructure/NoUselessElseFixer.php
```diff
@@ -57,18 +59,24 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
-    public function getPriority()
+    public function getDefinition()
     {
-        // should be run before NoWhitespaceInBlankLineFixer, NoExtraConsecutiveBlankLinesFixer, BracesFixer and after NoEmptyStatementFixer.
-        return 25;
+        return new FixerDefinition(
+            'There should not be useless else cases.',
+            null,
+            array(
+                new CodeSample("<?php\nif (\$a) {\n    return 1;\n} else {\n    return 2;\n}"),
```
keradus: don't hesitate to use `'` instead of `"` and real line endings instead of `\n` when it would be more readable ;)

src/FixerDefinition/FixerDefinition.php
```diff
@@ -36,9 +36,9 @@ public function __construct(
         $summary,
         $description,
         array $codeSamples,
-        $configurationDescription,
+        $configurationDescription = null,
         array $defaultConfiguration = null,
-        $riskyDescription
+        $riskyDescription = null
```
keradus: I would really want to avoid optional parameters.
I used `array $defaultConfiguration = null` just because I cannot use `?array $defaultConfiguration`

tests/FixerFactoryTest.php
```diff
@@ -380,15 +379,38 @@ public function testFixerDefinitions(FixerInterface $fixer)
             $this->markTestIncomplete('ShortFixerDefinition does not contains all needed information.');
         }
 
-        $this->assertNotEmpty($definition->getCodeSamples(), 'Code samples are required.');
+        $samples = $definition->getCodeSamples();
+        $this->assertInternalType('array', $samples, 'Code samples must be an array.');
```
keradus: return type of `getCodeSamples` is `array`, always.
not sure that we need to check if one is violating the interface... we don't do it anywhere else

tests/FixerFactoryTest.php
```diff
@@ -380,15 +379,38 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+        $this->assertNotEmpty($samples, 'Code samples are required.');
```
keradus: maybe move that check do Definition constructor ?

tests/FixerFactoryTest.php
keradus: same about types checking later in code

tests/FixerFactoryTest.php
```diff
@@ -380,15 +379,38 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+            $code = $sample->getCode();
+            $this->assertStringIsNotEmpty($code);
+            $config = $sample->getConfiguration();
+            if (null !== $config) {
+                $this->assertInternalType('array', $config, 'Sample config must be an array or null.');
+            }
+
+            // TODO
+            // test code is valid PHP
+            // test code is candidate for fixer
```
keradus: execute the `$fixer->fix` and see if input is different than output.
in the meantime, Tokenizer will raise an exception for invalid PHP on PHP7

src/FixerDefinition/FixerDefinition.php
SpacePossum: it gets kind of bloated to add some dummy data for each fixer that is not risky.

keradus: I'm not saying that $riskyDescription is not nullable.
Let think in PHP7.1 world.
`$riskyDescription` may be `string` or `null`.
So the type in prototype is `?string $riskyDescription`.
And that typehinting is already in method phpdoc.

What I'm saying that I don't like to have default value here.
`?string $riskyDescription` is good for me
`?string $riskyDescription = null` is not

SpacePossum: >Let think in PHP7.1 world.

Kind of hard here ;)  anyway this keeps the SCA happy for now

tests/FixerFactoryTest.php
SpacePossum: for now this is a nice DX when adding the descriptions, I see no harm in keeping this.

SpacePossum:
IMHO this PR is ready

SpacePossum:
don't know what is wrong with appveyor

README.rst
```diff
@@ -555,7 +556,7 @@ Choose from the list of available rules:
    | *Rule is: risky.*
 
 * **simplified_null_return**
-   | A return statement wishing to return nothing should be simply "return".
+   | A return statement wishing to return "void" should not return "null".
```
SpacePossum: the current description makes no sense, because:
1 - you cannot return `nothing` (unless you have class named like that), so a statement cannot wish to return it
2 - we do not check the willingness/wish of the statement to return 'nothing'

src/FixerDefinition/FixerDefinition.php
SpacePossum: here you go https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/180062009#L467

README.rst
keradus: and the new one:
`A return statement wishing to return "void" should not return "null".`

is suggesting that we do check the willingness

SpacePossum: >and the new one: [...] is suggesting that we do check the willingness

yes, like the old one, so 

>maybe just don't modify the description

doesn't solve it.

Going back and forth over copy makes no sense to me, just tell me what to put in.
Same with:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2339#discussion_r90223642
You don't want the defaults set to `null`, I don't want to pass `null` everywhere, the result is the tests breaking. So what to do with this PR, its getting time consuming without much process.

src/FixerDefinition/FixerDefinition.php
keradus: this is not static code analysis (like phpmd) problem. it's missing argument in your code.

README.rst
keradus: description:
indeed, none of them check the willingness, but only new description says about `statement wishing to`..., which is not correct.
original descr is not good, but IMO new one is more misleading.
I would not change the description in this PR as it's problematic and then...
maybe we could remove it, as it way problematic after introducing `void`...

---------

passing `null` vs default `null`.
As I already written before, especially for definition of the fixer, I would demand a developer to be aware that there is a parameter and he should think about value for it, instead of just "oh, it's optional parameter, i don't need to pass it even if it would make sense".
it's not going back and forward. i'm pretty consistent about my expectations.

SpacePossum: Disagree, as why as developer shoud I care about a param that does not apply to my case? Here why pass risky description if the fixer is not risky?

SpacePossum: Btw. We have test to force risky description if the fixer is risky

keradus: to prove that it does not apply willingly
utests in this repo wont test custom fixers which have a definition.


I just don't want to have default parameters at DTO objects...

keradus: very well and very well.
also, i moved required parameters on very beginning

src/Fixer/ArrayNotation/ArraySyntaxFixer.php
```diff
@@ -91,10 +91,12 @@ public function getDefinition()
                     "<?php\n[1,2];",
                     array('syntax' => 'long')
                 ),
+                /* @TODO That code should be in use, but for now it will fail on lower PHP version...
```
keradus: FYI @SpacePossum

keradus:
Thank you @SpacePossum.

PR #2347, opened 2016-12-01
UPGRADE.md - Fix multi-row description

drAlberT:
Avoid description to be divided in to more then one row

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -31,7 +31,7 @@ CLI options
 --fixers        |                 | Coding standard fixers                                                    | option was removed, see --rules
 --format        | --format        | Choose format                                                             |
 --level         |                 | Coding standard level                                                     | option was removed, see --rules
-                | --path-mode     | Should the finder from config be overriden or intersected with `path` arg | option was added
+                | --path-mode     | Should the finder from config be <br/>overriden or intersected with `path` arg | option was added
                 | --rules         | Rules to be used                                                          | option was added
                 | --using-cache   | Does cache should be used                                                 | option was added
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -20,21 +20,20 @@ Finally, the caching mechanism is enabled by default.
 CLI options
 -----------
 
-1.x           | 2.0           | Description                                     | Note
-------------- | ------------- | ----------------------------------------------- | ----
-              | --allow-risky | Are risky fixers allowed                        |
-              | --cache-file  | The path to the cache file                      | option was added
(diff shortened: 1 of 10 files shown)


UPGRADE.md
```diff
@@ -20,21 +20,20 @@ Finally, the caching mechanism is enabled by default.
...
+              | --allow-risky | Are risky fixers allowed                               |
+              | --cache-file  | The path to the cache file                             | option was added
+--config      |               | Config class codename                                  | option was removed
+--config-file | --config      | The path to a .php_cs file                             | option was renamed
+--diff        | --diff        | Show diff                                              |
+--dry-run     | --dry-run     | Run in dry-run mode                                    |
+--fixers      |               | Coding standard fixers                                 | option was removed, see --rules
+--format      | --format      | Choose format                                          |
+--level       |               | Coding standard level                                  | option was removed, see --rules
+              | --path-mode   | Should `path` argument override the finder from config | option was added
```
GrahamCampbell: Might be more readable as "Should the `path` arg override the finder from config"

keradus:
> Avoid description to be divided in to more then one row

what for? this is perfectly valid MD syntax

drAlberT:
@keradus it is valid, of course .. but the resulting output is splitted on two *different* rows, having different colors ... 

Before PR:
screen shot 2016-12-06 at 09 32 40

After PR:
screen shot 2016-12-06 at 09 33 19

UPGRADE.md
SpacePossum: Like the idea of not taking two rows for one line of text (two lines would be ok).
The intersect part got lost (and is an important change) in the new text, I think this should be added again.

drAlberT:
WIP .. wait guys .. going to find a better solution .. sorry for noise .. not having a shell at the moment, but using GH editor is going to create a lot of confusion on this PR ..

Will stash ASAP

SpacePossum:
:+1:

keradus:
Table doesn't look nice now:
https://github.com/drAlberT/PHP-CS-Fixer/blob/28df22ba370db84f5de9a4ebf6a77bc5f9fadfe4/UPGRADE.md#cli-options

compare to
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md#cli-options

the description column is way to wide.
line breaker for long description is needed

SpacePossum:
@drAlberT please throw in a `<br/`> and check how it looks ;)

keradus:
Thank you @drAlberT.

PR #2350, opened 2016-12-02
Update .editorconfig

vinkla:
Added better support for `*.md` and `*.yml` files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,7 +10,3 @@ trim_trailing_whitespace = true
 
 [*.md]
 trim_trailing_whitespace = false
-
-[*.yml]
-indent_style = space
-indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,16 @@
 root = true
+
 [*]
+charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_style = space
+indent_size = 2


.editorconfig
```diff
@@ -1,7 +1,16 @@
...
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.yml]
+indent_style = space
+indent_size = 2
```
keradus: Please use 4 instead

vinkla:
4 spaces in Yaml? Isn't 2 required?

SpacePossum:
>4 spaces in Yaml? Isn't 2 required?

it is not required, please see http://www.yaml.org/spec/current.html#id2519916

vinkla:
Done.

GrahamCampbell:
:+1:

.editorconfig
```diff
@@ -1,7 +1,12 @@
...
 [*]
+charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
vinkla: `CHANGELOG.md`, `CONTRIBUTING.md` and `UPGRADE.md`.

keradus: why do we want to have trailing whitespaces at those files ?

vinkla: In Markdown, two spaces in the end of a line means line-break.

https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#lines

keradus: we don't rely on that feature. i would not allow for trailing spaces anywhere except fixture files

vinkla: Which means you don't want this?

keradus: indeed, i would drop special rule for .md

vinkla:
I'll just go ahead on close this all together.

PR #2351, opened 2016-12-02
Enhancement: Allow to configure return_type_declaration rule

localheinz:
This PR

* [x] allows to configure the `return_type_declaration` rule

💁‍♂️ Some people prefer to have a space before `:`, hehe. For reference, also see https://github.com/php-fig/fig-standards/pull/733.

### DIFF
--- a/src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
+++ b/src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer\Fixer\FunctionNotation;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\FixerDefinition\VersionSpecification;
 use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
@@ -22,8 +24,44 @@ use PhpCsFixer\Tokenizer\Tokens;
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
-final class ReturnTypeDeclarationFixer extends AbstractFixer
+final class ReturnTypeDeclarationFixer extends AbstractFixer implements ConfigurableFixerInterface
 {
+    /**
+     * @var string
+     */
+    private $config;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $this->config = 'none';
+
+            return;
+        }
+
+        $key = 'space_before';
+        $values = array(
+            'one',
+            'none',
+        );
+
+        if (!array_key_exists($key, $configuration) || !in_array($configuration[$key], $values, true)) {
+            throw new InvalidFixerConfigurationException(
+                $this->getName(),
+                sprintf(
+                    'Configuration must define "%s" being "%s".',
+                    $key,
+                    implode('" or "', $values)
+                )
+            );
+        }
+
+        $this->config = $configuration[$key];
+    }
+
     /**
      * {@inheritdoc}
      */
@@ -36,8 +74,17 @@ final class ReturnTypeDeclarationFixer extends AbstractFixer
                 continue;
             }
 
-            if ($tokens[$index - 1]->isWhitespace()) {
-                $tokens[$index - 1]->clear();
+            $previousToken = $tokens[$index - 1];
+
+            if ($previousToken->isWhitespace()) {
+                if ('none' === $this->config) {
+                    $previousToken->clear();
+                } else {
+                    $previousToken->setContent(' ');
+                }
+            } elseif ('one' === $this->config) {
+                $tokens->ensureWhitespaceAtIndex($index, 0, ' ');
+                ++$index;
             }
 
             ++$index;
@@ -50,14 +97,29 @@ final class ReturnTypeDeclarationFixer extends AbstractFixer
      */
     public function getDefinition()
     {
+        $versionSpecification = new VersionSpecification(70000);
+
         return new FixerDefinition(
-            'There should be no space before colon and one space after it in return type declaration. Requires PHP >= 7.',
+            'There should be one or no space before colon, and one space after it in return type declarations, according to configuration.',
             array(
                 new VersionSpecificCodeSample(
-                    '<?php function foo(int $a):string {}',
-                    new VersionSpecification(70000)
+                    "<?php\nfunction foo(int \$a):string {};",
+                    $versionSpecification
                 ),
-            )
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(int \$a): string {};",
+                    $versionSpecification,
+                    array('space_before' => 'none')
+                ),
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(int \$a): string {};",
+                    $versionSpecification,
+                    array('space_before' => 'one')
+                ),
+            ),
+            null,
+            "Configuration must have one element 'space_before' with value 'none' (default) or 'one'.",
+            array('space_before' => 'none')
         );
     }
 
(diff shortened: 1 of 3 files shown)


SpacePossum:
I like the idea :)

About failing tests, please decrease the guard, it would be a valid change as you up'ped the number of descriptions :+1: 

The other one is related to a limitation we have on the testing process itself. Currently when testing if a code sample is valid it is not possible to define the PHP requirement(s) of the sample. I would like suggest to run these test on PHP 7 only for now, as it would suit more (/most) cases, but that needs some discussion first.

localheinz:
Thank you, @SpacePossum!

I like it a lot, and I am very happy it's finally been released. Great work! 👍

localheinz:
@keradus @SpacePossum 

I have rebased this branch now following #2354!

SpacePossum:
thanks, looking good :)

tests/Fixer/FunctionNotation/ReturnTypeDeclarationFixerTest.php
```diff
@@ -22,18 +22,47 @@
 final class ReturnTypeDeclarationFixerTest extends AbstractFixerTestCase
 {
     /**
+     * @expectedException \PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException
+     * @expectedExceptionMessageRegExp #^\[return_type_declaration\] Configuration must define "space_before" being "one" or "none".$#
```
keradus: please set expected via method instead annotation.
there are plans to drop it on phpunit 6

tests/Fixer/FunctionNotation/ReturnTypeDeclarationFixerTest.php
```diff
@@ -22,18 +22,47 @@
...
+        $this->fixer->configure(array('s' => 9000));
+    }
+
+    /**
+     * @dataProvider testFixProviderWithSpaceBeforeNone
+     *
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @requires PHP 7.0
```
keradus: could be moved to class-level annotation

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
```diff
@@ -53,8 +103,31 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
+                    $versionSpecification,
+                    array('space_before' => 'none')
+                ),
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(): string {};",
+                    $versionSpecification,
+                    array('space_before' => 'one')
+                ),
+            ),
+            null,
```
localheinz: Ha, not sure what to put here, can you help?

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
```diff
@@ -53,8 +103,31 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
     {
-        return 'There should be no space before colon and one space after it in return type declaration.';
+        $versionSpecification = new VersionSpecification(70000);
+
+        return new FixerDefinition(
+            'There should be one or no space before colon, and one space after it in return type declarations, according to configuration.',
+            array(
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo() : string {};",
+                    $versionSpecification
```
localheinz: Hmm.

Just ran

```
$ ./php-cs-fixer describe return_type_declaration

Description of return_type_declaration rule.
There should be one or no space before colon, and one space after it in return type declarations, according to configuration.

Fixer is configurable.
Configuration must have one element 'space_before' with value 'none' (default) or 'one'.
Default configuration: ['space_before' => 'none'].

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ @@
    <?php
   -function foo() : string {};
   +function foo(): string {};

 * Example #2. Fixing with configuration: ['space_before' => 'none'].
   --- Original
   +++ New
   @@ @@
    <?php
   -function foo() : string {};
   +function foo(): string {};

 * Example #3. Fixing with configuration: ['space_before' => 'one'].
   --- Original
   +++ New
   @@ @@
    <?php
   -function foo(): string {};
   +function foo() : string {};
```

Is there something I have missed here?

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
keradus: I would put sth like :  `'Rule is applied only on PHP 7+ environment.'` 
Could be anything, not insisting on exact content, just want to have required php printed ;)

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
SpacePossum: if you run it on PHP < 7 it is empty ;)

localheinz: Ha, good point!

localheinz: That should probably be a fix against `master`, or do you have another idea? 

That is, following #2354, if we end up with no code examples satisfied by the current version, we display a note about that?

keradus: sure, note is ok.
if you want to do it in separated pr im fine with that. just be aware that i will wait with merging this pr until new one will be merged to not provide php5 users bad experience

keradus: actually yeah, fix will go for 2.0.1, and rule enhancement for 2.1.0, so definitely 2nd pr is needed ;)

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
```diff
@@ -53,8 +103,31 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
+                ),
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(): string {};",
+                    $versionSpecification,
+                    array('space_before' => 'one')
+                ),
+            ),
+            'Rule is applied only in a PHP 7+ environment.',
+            "Configuration must have one element 'space_before' with value 'none' (default) or 'one'.",
+            array('space_before' => 'none')
```
localheinz: I was thinking, this kind of smells, what do you think about

* adding a field to configurable fixers with the default configuration, then using that here
* streamlining the way configuration is resolved in `configure()`

What do you think?

localheinz: Probably should be done after #2366, when we know what the default configurations are, hehe.

keradus: in most cases the field is already there and is used during resolving.

but i see your point that its duplicated

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
```diff
@@ -13,15 +13,56 @@
...
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $this->config = 'none';
+
+            return;
+        }
+
+        $key = 'space_before';
+        $values = array(
```
SpacePossum: can be one line (but no strong opinion)

tests/Fixer/FunctionNotation/ReturnTypeDeclarationFixerTest.php
```diff
@@ -18,22 +18,50 @@
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  *
  * @internal
+ *
+ * @requires PHP 7.0
```
SpacePossum: :+1:  nice :)

SpacePossum:
@localheinz please rebase and resolve the conflicts :)

localheinz:
@keradus @SpacePossum 

Rebased, and incorporated suggestions, as well as a default configuration as suggested in #2382.

SpacePossum:
awesome, shippit :)

keradus:
Thank you @localheinz.

localheinz:

SpacePossum:
thank you @localheinz, nice! :)

PR #2352, opened 2016-12-02
Corrected method visibility

GrahamCampbell:
I'm not sure if this is what you had in mind?


SpacePossum:
sure, shipit :)
the goal is to have all those method removed anyway ;)

GrahamCampbell:
Maybe we should also add an internal test to check the visibility of this method?

keradus:
no need for doing that:
- they are outside the interface, if one uses them it is his risk
- with time, we will remove all of them

GrahamCampbell:
Makes sense. :)

keradus:
Thank you @GrahamCampbell.

PR #2354, opened 2016-12-03
Enhancement: Allow to specify minimum and maximum PHP versions for code samples

localheinz:
This PR

* [x] allows to compose a `CodeSample` and a `VersionSpecification` into a `VersionSpecificCodeSample`, which will then be ignored in the `DescribeCommand` when the `VersionSpecification` isn't satisfied by the current PHP version

Related to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2351#issuecomment-264530922.

### DIFF
--- a/src/Fixer/ArrayNotation/ArraySyntaxFixer.php
+++ b/src/Fixer/ArrayNotation/ArraySyntaxFixer.php
@@ -17,6 +17,8 @@ use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
 use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
@@ -91,12 +93,11 @@ final class ArraySyntaxFixer extends AbstractFixer implements ConfigurableFixerI
                     "<?php\n[1,2];",
                     array('syntax' => 'long')
                 ),
-                /* @TODO That code should be in use, but for now it will fail on lower PHP version...
-                new CodeSample(
+                new VersionSpecificCodeSample(
                     "<?php\narray(1,2);",
+                    new VersionSpecification(50400),
                     array('syntax' => 'short')
                 ),
-                */
             ),
             null,
             'Configure to use "long" or "short" array declaration syntax.',
--- a/src/FixerDefinition/VersionSpecificCodeSample.php
+++ b/src/FixerDefinition/VersionSpecificCodeSample.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class VersionSpecificCodeSample implements VersionSpecificCodeSampleInterface
+{
+    /**
+     * @var CodeSampleInterface
+     */
+    private $codeSample;
+
+    /**
+     * @var null|VersionSpecificationInterface
+     */
+    private $versionSpecification;
+
+    /**
+     * @param string                        $code
+     * @param VersionSpecificationInterface $versionSpecification
+     * @param null|array                    $configuration
+     */
+    public function __construct(
+        $code,
+        VersionSpecificationInterface $versionSpecification,
+        array $configuration = null
+    ) {
+        $this->codeSample = new CodeSample($code, $configuration);
+        $this->versionSpecification = $versionSpecification;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCode()
+    {
+        return $this->codeSample->getCode();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfiguration()
+    {
+        return $this->codeSample->getConfiguration();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isSuitableFor($version)
+    {
+        if (null === $this->versionSpecification) {
+            return true;
+        }
+
+        return $this->versionSpecification->isSatisfiedBy($version);
+    }
+}
--- a/src/FixerDefinition/VersionSpecificCodeSample.php
+++ b/src/FixerDefinition/VersionSpecificCodeSample.php
@@ -23,7 +23,7 @@ final class VersionSpecificCodeSample implements VersionSpecificCodeSampleInterf
     private $codeSample;
 
     /**
-     * @var null|VersionSpecificationInterface
+     * @var VersionSpecificationInterface
      */
     private $versionSpecification;
 
--- a/src/FixerDefinition/VersionSpecificCodeSample.php
+++ b/src/FixerDefinition/VersionSpecificCodeSample.php
@@ -62,10 +62,6 @@ final class VersionSpecificCodeSample implements VersionSpecificCodeSampleInterf
      */
     public function isSuitableFor($version)
     {
-        if (null === $this->versionSpecification) {
-            return true;
-        }
-
         return $this->versionSpecification->isSatisfiedBy($version);
     }
 }
--- a/src/FixerDefinition/VersionSpecificCodeSampleInterface.php
+++ b/src/FixerDefinition/VersionSpecificCodeSampleInterface.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Moeller <am@localheinz.com>
+ */
+interface VersionSpecificCodeSampleInterface extends CodeSampleInterface
+{
+    /**
+     * @param int $version
+     *
+     * @return bool
+     */
+    public function isSuitableFor($version);
+}
--- a/src/FixerDefinition/VersionSpecification.php
+++ b/src/FixerDefinition/VersionSpecification.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+final class VersionSpecification implements VersionSpecificationInterface
+{
+    /**
+     * @var int|null
+     */
+    private $minimum;
+
+    /**
+     * @var int|null
+     */
+    private $maximum;
+
+    /**
+     * @param int|null $minimum
+     * @param int|null $maximum
+     *
+     * @throws \InvalidArgumentException
+     */
+    public function __construct($minimum = null, $maximum = null)
+    {
+        if (null === $minimum && null === $maximum) {
+            throw new \InvalidArgumentException('Either minimum or maximum need to be specified');
+        }
+
+        if (null !== $minimum && (!is_int($minimum) || 1 > $minimum)) {
+            throw new \InvalidArgumentException('Minimum needs to be either null or an integer greater than 0');
+        }
+
+        if (null !== $maximum && (!is_int($maximum) || 1 > $maximum)) {
+            throw new \InvalidArgumentException('Minimum needs to be either null or an integer greater than 0');
+        }
+
+        if (null !== $maximum && null !== $minimum && $maximum < $minimum) {
+            throw new \InvalidArgumentException('Maximum should not be less than the minimum');
+        }
+
+        $this->minimum = $minimum;
+        $this->maximum = $maximum;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isSatisfiedBy($version)
+    {
+        if (null !== $this->minimum && $version < $this->minimum) {
+            return false;
+        }
+        if (null !== $this->maximum && $version > $this->maximum) {
+            return false;
+        }
+
+        return true;
+    }
+}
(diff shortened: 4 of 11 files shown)


src/FixerDefinition/CodeSampleInterface.php
```diff
@@ -26,4 +26,9 @@ public function getCode();
      * @return null|array
      */
     public function getConfiguration();
+
+    /**
+     * @return null|int
+     */
+    public function getRequiredVersion();
```
GrahamCampbell: Adding new method to interfaces is breaking. 2.x is stable now.

GrahamCampbell: This interface was not marked as internal.

localheinz: Hmm, do you have an idea how to solve the problem?

We could extend the interface, and have a concretion that implements the extended interface, what do you think?

GrahamCampbell: Hmm, depends how much we want to respect the BC promise. Maybe it's worth making a 3.x branch for this? I don't think we should just add more interfaces, though, that's maybe not such a bad idea - would definitely be a solution.

Really something for @keradus for decide. Each solution has pros and cons, and the release timeline for 3.x affects things.

keradus: breaking bc is not an option. extended interface is needed

SpacePossum: Maybe we should add an upper and lower bound, as some fixers fix code not valid on PHP7 for example.

tests/FixerFactoryTest.php
```diff
@@ -385,6 +385,10 @@ public function testFixerDefinitions(FixerInterface $fixer)
         $sampleCounter = 0;
         foreach ($samples as $sample) {
             ++$sampleCounter;
+
+            if (!$sample->isSuitableForVersion(PHP_VERSION_ID)) {
+                continue;
+            }
             $this->assertInstanceOf('PhpCsFixer\FixerDefinition\CodeSampleInterface', $sample, sprintf('[%s] Sample #%d', $fixer->getName(), $sampleCounter));
```
SpacePossum: Should be placed a bit lower, after the instance and type checks

tests/FixerDefinition/CodeSampleTest.php
```diff
@@ -0,0 +1,77 @@
...
+namespace PhpCsFixer\Tests\FixerDefinition;
+
+use PhpCsFixer\FixerDefinition\CodeSample;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ */
+final class CodeSampleTest extends \PHPUnit_Framework_TestCase
```
SpacePossum: Nice!

src/FixerDefinition/CodeSample.php
```diff
@@ -28,13 +28,20 @@
...
      * @param string     $code
      * @param null|array $configuration
+     * @param null|int   $requiredVersion
      */
-    public function __construct($code, array $configuration = null)
+    public function __construct($code, array $configuration = null, $requiredVersion = null)
     {
         $this->code = $code;
         $this->configuration = $configuration;
+        $this->requiredVersion = $requiredVersion;
```
SpacePossum: Lower bound default can be set to the PHP min. version required by the fixer tool.

keradus: there is no php ver required by a fixer

keradus: oh, tool, not rule ?
what for ?

tests/FixerFactoryTest.php
```diff
@@ -398,6 +400,10 @@ public function testFixerDefinitions(FixerInterface $fixer)
                 }
             }
 
+            if ($sample instanceof VersionedCodeSample && !$sample->isSuitableForVersion(PHP_VERSION_ID)) {
```
keradus: Do it before calling `$fixer->confgure(...)` (so just after `assertStringIsNotEmpty`), then tests will pass.

tests/FixerDefinition/VersionedCodeSampleTest.php
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @internal
+ */
+final class VersionedCodeSampleTest extends \PHPUnit_Framework_TestCase
+{
+    public function testImplementsVersionedCodeSampleInterface()
+    {
+        $reflection = new \ReflectionClass('PhpCsFixer\FixerDefinition\VersionedCodeSample');
+
+        $this->assertTrue($reflection->implementsInterface('PhpCsFixer\FixerDefinition\VersionedCodeSampleInterface'));
```
keradus: question: what is a benefit of this test ?
if concrete class would not follow interface, the other tests would fail

tests/FixerDefinition/VersionedCodeSampleTest.php
```diff
@@ -0,0 +1,84 @@
...
+    public function testConstructorSetsValues()
+    {
+        $code = '<php echo $foo;';
+        $configuration = array(
+            'foo' => 'bar',
+        );
+
+        $codeSample = new VersionedCodeSample(
+            $code,
+            $configuration
```
keradus: no test with 3rd parameter ?

src/FixerDefinition/CodeSampleInterface.php
```diff
@@ -26,4 +26,11 @@ public function getCode();
      * @return null|array
      */
     public function getConfiguration();
+
+    /**
+     * @param int $version
+     *
+     * @return bool
+     */
+    public function isSuitableForVersion($version);
```
keradus: this was moved to `Versioned...`, should be removed from here

SpacePossum:
:+1: looking good :)

src/Fixer/ArrayNotation/ArraySyntaxFixer.php
```diff
@@ -91,12 +93,11 @@ public function getDefinition()
                     "<?php\n[1,2];",
                     array('syntax' => 'long')
                 ),
-                /* @TODO That code should be in use, but for now it will fail on lower PHP version...
-                new CodeSample(
```
keradus: please do the same for `PowToExponentiationFixer`.

src/FixerDefinition/VersionSpecificCodeSample.php
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * @param string                             $code
+     * @param null|array                         $configuration
+     * @param VersionSpecificationInterface|null $versionSpecification
+     */
+    public function __construct(
+        $code,
+        array $configuration = null,
+        VersionSpecificationInterface $versionSpecification = null
```
keradus: Wait... so here, we are inside `VersionSpecificCodeSampleInterface` implementation.
Why `VersionSpecific...` could be set up to be specific-less ?
then, one should use pure `CodeSampleInterface`

src/FixerDefinition/VersionSpecification.php
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * @var int|null
+     */
+    private $maximum;
+
+    public function __construct($minimum = null, $maximum = null)
+    {
+        $this->minimum = $minimum;
+        $this->maximum = $maximum;
```
keradus: I would say that having 2 nulls is an invalid argument exception

src/FixerDefinition/VersionSpecification.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * @var int|null
+     */
+    private $minimum;
+
+    /**
+     * @var int|null
+     */
+    private $maximum;
```
keradus: @SpacePossum you have asked for it, @localheinz you have accept that request.
but my question was not been aswered

Why we need to have maximum PHP version ?
We do not have any fixer that would specify that

tests/FixerDefinition/CodeSampleTest.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    public function testImplementsCodeSampleInterface()
+    {
+        $reflection = new \ReflectionClass('PhpCsFixer\FixerDefinition\CodeSample');
```
keradus: let me repeat the question
what is a benefit for checking that class implements an interface?
if that would be violated, the compiler-time parameter typechecking will raise an error and test would not pass as well.

tests/FixerDefinition/VersionSpecificCodeSampleTest.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    /**
+     * @dataProvider providerIsSuitableForVersionUsesVersionSpecification
+     *
+     * @param bool $isSatisfied
+     */
+    public function testIsSuitableForUsesVersionSpecification($isSatisfied)
+    {
+        $version = 9000;
```
keradus: should be part of data provider.
data provider that provide only expected value, but no input values, looks weird for me

src/FixerDefinition/VersionSpecification.php
SpacePossum: forward support, depending on how this will turns out:
https://wiki.php.net/rfc/deprecations_php_7_2
for example. If in that RFC (or any other) features are removed from PHP which are valid now, we cannot set a max. fix level. So if `(binary)$a;` is no longer valid syntax on 7.2 or 8.0, we can longer removed that cast when one is running the tool on 7.2/8.0, but we could still do it on >7.1

tests/FixerDefinition/CodeSampleTest.php
SpacePossum: yes, this might be overcomplete.

src/FixerDefinition/VersionSpecification.php
localheinz: Why, maybe we just don't care about the version?

src/FixerDefinition/VersionSpecificCodeSampleInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Moeller <am@localheinz.com>
+ *
+ * @internal
```
keradus: `CodeSampleInterface` is not internal, as it's supposed to be used by users writing custom fixers.
not sure why this, extended version, should be marked as internal ...

src/FixerDefinition/VersionSpecification.php
SpacePossum: maybe, but than why use this :) 
one could than use `CodeSample` and not `VersionSpecificCodeSample`

src/FixerDefinition/VersionSpecificCodeSample.php
localheinz: Well, that would just mean that we need to make `$configuration` a required value, right? Or should I modify the order of arguments, and make `$versionSpecification` the second argument and required?

localheinz:
@GrahamCampbell @keradus @SpacePossum 

I think I implemented all of your suggestions, let me know what you think!

src/FixerDefinition/VersionSpecificCodeSample.php
SpacePossum: swapping would make sense to me :)  not sure what Keradus thinks though

src/FixerDefinition/VersionSpecificationInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
```
SpacePossum: can be removed :)

src/Fixer/ArrayNotation/ArraySyntaxFixer.php
SpacePossum: done by @localheinz , thanks :)

src/FixerDefinition/VersionSpecification.php
```diff
@@ -0,0 +1,74 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerDefinition;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
```
keradus: most DTO here are to be used for users that write custom fixers, so yes.

keradus:
@localheinz please don't squash the commits.
now instead of taking a look at changes only we need to read whole PR again

src/FixerDefinition/VersionSpecification.php
keradus: the problem here would be if we would convert sth to binary on our fixer.
we don't do that.

localheinz: Would you prefer me to remove `$maximum` for now?

src/FixerDefinition/VersionSpecificCodeSample.php
keradus: TBH I treat `$configuration` as `?array $configuration` without default value. just php5 don't accept that ;)

cc @localheinz @SpacePossum

src/FixerDefinition/VersionSpecification.php
keradus: i don't want to blindly go in one direction, whichever it would be.
that's why i'm looking for reasoning ;)

SpacePossum: >the problem here would be if we would convert sth to binary on our fixer.

Not only in that case.
If we have a fixer that would deal with `(binary)` we want to add an example here, but if `(binary)` is no longer valid on a high PHP version the example won't lint and so fails.
Same for the PHP4 constructors, if those would be removed (i.e. no longer lint) on PHP8, the fixer can only give valid samples for versions below that.

keradus: @localheinz 
that's exactly why I asked for discussion about it instead of requiring the change you have now made.
now I see and understand it as a valid concern and `max` should be there, please revert last commit

src/FixerDefinition/VersionSpecificCodeSample.php
```diff
@@ -0,0 +1,67 @@
...
+ */
+final class VersionSpecificCodeSample implements VersionSpecificCodeSampleInterface
+{
+    /**
+     * @var CodeSampleInterface
+     */
+    private $codeSample;
+
+    /**
+     * @var null|VersionSpecificationInterface
```
keradus: not nullable anymore

src/FixerDefinition/VersionSpecification.php
localheinz: Done!

keradus: Thank you !

tests/FixerDefinition/VersionSpecificationTest.php
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * @expectedException \InvalidArgumentException
+     */
+    public function testConstructorRequiresEitherMinimumOrMaximum()
+    {
+        new VersionSpecification();
+    }
+
+    /**
+     * @expectedException \InvalidArgumentException
```
keradus: instead of annotation use `setExpectedException` method please

keradus:
:+1: for PR , I like your codestyle @localheinz 
I treat this as documentation-related change, therefore I gonna merge this into 2.0.x line. Anything against ?

SpacePossum:
looks good to me! Love to see this merged as I'm having more and more cases where I need this : D

localheinz:
Thank you, @GrahamCampbell, @keradus and @SpacePossum!

> 👍 for PR , I like your codestyle @localheinz

Thank you, I appreciate it, @keradus!

keradus:
Thank you @localheinz.

SpacePossum:
\o/  thanks @localheinz :)

PR #2359, opened 2016-12-05
Add PhpdocNoUselessInheritdocFixer

SpacePossum:
Early opening, need to fix the priority tests.
For ref.: https://github.com/twigphp/Twig/pull/2254/files

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocNoUselessInheritdocFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocNoUselessInheritdocFixer.php
@@ -1,0 +1,192 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Phpdoc;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Remove inheritdoc tags from classy that does not inherit.
+ *
+ * @author SpacePossum
+ */
+final class PhpdocNoUselessInheritdocFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // min. offset 4 as minimal candidate is @: <?php\n/** @inheritdoc */class min{}
+        for ($index = 1, $count = count($tokens) - 4; $index < $count; ++$index) {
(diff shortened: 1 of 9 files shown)


keradus:
what if documented `inheritdoc` method comes from trait, not parent class ? could it be a case ?

SpacePossum:
>what if documented inheritdoc method comes from trait, not parent class ? could it be a case ?

good point, I'll add a test for candidate if a class uses a trait

localheinz:
Nice!

SpacePossum:
all points addressed

src/Fixer/Phpdoc/PhpdocNoUselessInheritdocFixer.php
```diff
@@ -0,0 +1,193 @@
...
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // min. offset 4 as minimal candidate is @: <?php\n/** @inheritdoc */class min{}
+        for ($index = 1, $count = count($tokens) - 4; $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(array(T_CLASS, T_INTERFACE))) {
+                continue;
+            }
+
+            $index = $this->fixClassy($tokens, $index);
```
keradus: remove negation from condition, move this inside the condition, 2 lines less and no negation ;)

src/Fixer/Phpdoc/PhpdocNoUselessInheritdocFixer.php
```diff
@@ -0,0 +1,193 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Classy that does not inherit must not have inheritdoc tags.',
+            array(
+                new CodeSample("<?php\n/** {@inheritdoc} */\nclass MyClass\n{\n}"),
+            )
```
keradus: please add a case with a method to show that it also works for them

tests/FixerFactoryTest.php
```diff
@@ -300,6 +300,11 @@ public function getFixersPriorityCases()
             array($fixers['protected_to_private'], $fixers['ordered_class_elements']), // tested also in: protected_to_private,ordered_class_elements.test
             array($fixers['phpdoc_add_missing_param_annotation'], $fixers['phpdoc_align']), // tested also in: phpdoc_add_missing_param_annotation,phpdoc_align.test
             array($fixers['phpdoc_no_alias_tag'], $fixers['phpdoc_add_missing_param_annotation']), // tested also in: phpdoc_no_alias_tag,phpdoc_add_missing_param_annotation.test
+            array($fixers['phpdoc_no_useless_inheritdoc'], $fixers['no_empty_phpdoc']), // tested also in: phpdoc_no_useless_inheritdoc,no_empty_phpdoc.test
+            array($fixers['phpdoc_no_useless_inheritdoc'], $fixers['phpdoc_inline_tag']), // tested also in: phpdoc_no_useless_inheritdoc,phpdoc_inline_tag.test
+            array($fixers['phpdoc_to_comment'], $fixers['phpdoc_no_useless_inheritdoc']), // tested also in: phpdoc_to_comment,phpdoc_no_useless_inheritdoc.test
+
+                        // FIXME if skip priority, do it anyway to see if the claim correct
```
keradus: ?

keradus:
Thank you @SpacePossum.

Issue #2341, opened 2016-11-30
ClassKeywordRemoveFixer incorrectly leaves leading Backslash


rvock:
The ClassKeywordRemoveFixer leaves the backslash prefix infront of classes, if it was in the source:
[ClassKeywordRemoveFixerTest.php#L85](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ac04a510bed5407e91664f8a37b9d58072d96768/Symfony/CS/Tests/Fixer/Contrib/ClassKeywordRemoveFixerTest.php#L85)

This is incorrect. `::class` always returns the fully qualified name without a leading backslash:
```bash
$ php -r 'echo \DateTime::class;'
DateTime
```

This means the testcase is incorrect and the generated strings will be incorrect.

GrahamCampbell:
> This is incorrect. ::class always returns the fully qualified name without a leading backslash:

No it's not. The fixer is correct.

GrahamCampbell:
Compare `php -r 'namespace Foo; echo DateTime::class;` with `php -r 'namespace Foo; echo \DateTime::class;`

SpacePossum:
>Compare php -r 'namespace Foo; echo DateTime::class; with php -r 'namespace Foo; echo \DateTime::class;

done https://3v4l.org/gXD63

i.e. `echo \DateTime::class;` is not `echo '\DateTime';`

GrahamCampbell:
Oh, I see what you mean now. Sorry! I was thinking the fixer was going the other direction, so we'd have to have the slash for safety.

keradus:
Thanks for a report @rvock !
Do you want to work on a fix for that ?

rvock:
I think, it's enough to change one line:
[ClassKeywordRemoveFixer.php#L151](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/4cdef406cccdc0e3db79932a883f1c0531610861/Symfony/CS/Fixer/Contrib/ClassKeywordRemoveFixer.php#L151), but I am not sure, if this works with other imports:
```php
$classString = ltrim($tokens->generatePartialCode($classBeginIndex, $classEndIndex), '\\');
```

I might find some time during the weekend to test.

PR #2360, opened 2016-12-05
bug #2341 ClassKeywordRemoveFixer incorrectly leaves leading Backslash

rvock:
I adjusted the test-case and removed the leading backslashes from the classname.

### DIFF
--- a/src/Fixer/LanguageConstruct/ClassKeywordRemoveFixer.php
+++ b/src/Fixer/LanguageConstruct/ClassKeywordRemoveFixer.php
@@ -159,7 +159,7 @@ final class ClassKeywordRemoveFixer extends AbstractFixer
         $classBeginIndex = $classEndIndex;
         while ($tokens[--$classBeginIndex]->isGivenKind(array(T_NS_SEPARATOR, T_STRING)));
         ++$classBeginIndex;
-        $classString = $tokens->generatePartialCode($classBeginIndex, $classEndIndex);
+        $classString = ltrim($tokens->generatePartialCode($classBeginIndex, $classEndIndex), '\\');
 
         $classImport = false;
         foreach ($this->imports as $alias => $import) {
--- a/tests/Fixer/LanguageConstruct/ClassKeywordRemoveFixerTest.php
+++ b/tests/Fixer/LanguageConstruct/ClassKeywordRemoveFixerTest.php
@@ -90,7 +90,7 @@ final class ClassKeywordRemoveFixerTest extends AbstractFixerTestCase
             ),
             array(
                 "<?php
-                echo '\DateTime';
+                echo 'DateTime';
                 ",
                 "<?php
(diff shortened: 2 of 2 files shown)


SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2341

(for ref. and autoclosing on merge)

src/Fixer/LanguageConstruct/ClassKeywordRemoveFixer.php
```diff
@@ -159,7 +159,7 @@ private function replaceClassKeyword(Tokens $tokens, $classIndex)
         $classBeginIndex = $classEndIndex;
         while ($tokens[--$classBeginIndex]->isGivenKind(array(T_NS_SEPARATOR, T_STRING)));
         ++$classBeginIndex;
-        $classString = $tokens->generatePartialCode($classBeginIndex, $classEndIndex);
+        $classString = ltrim($tokens->generatePartialCode($classBeginIndex, $classEndIndex), '\\');
```
keradus: I would rather move the `$classBeginIndex` if its `T_NS_SEPARATOR` than creating bigger partial code and then shrink it back.

rvock: Do you mean like this?
```php
$classStringBeginIndex = $tokens[$classBeginIndex]->isGivenKind([T_NS_SEPARATOR]) ? $classBeginIndex + 1 : $classBeginIndex;
$classString = $tokens->generatePartialCode($classStringBeginIndex, $classEndIndex);
```

This would require a new index-variable, because the `$classBeginIndex` is still required to call clearRange and insertAt later in the method.

rvock:
@SpacePossum
Should I change my commit to include the "closes"-keyword? I thought maybe bug # is enough in the commit message, because there are other commits, that are missing "closes" in their commit-message...

SpacePossum:
@rvock no need to change the message.
(commit messages typically don't contain closes refs. in this project anyway)
All good thanks :+1:  lets focus on code changes :)

src/Fixer/LanguageConstruct/ClassKeywordRemoveFixer.php
SpacePossum: Looks like a good idea, I wouldn't worry about the extra variable.
I'm not sure about the fixed +1 offset, please add this case to the tests:

```php
<?php

class A
{
}

echo \/**/A::class;
echo \/**/A::
//
class;

echo \/**/A::#123
//
class;
```

https://3v4l.org/be3T8

keradus: from what i shortly checked recently we don't handle comments in `::class` construction anyway in this fixer
different bug, still nice to fix

PR #2361, opened 2016-12-06
ConfigurationResolver - Reject unknown rules

localheinz:
This PR

* [x] asserts that unknown rules are rejected by the `ConfigurationResolver`
* [x] rejects unknown rules 
* [x] adjusts an unrelated test that references an unknown (possibly renamed) rule

💁‍♂️ Not sure if this is the nicest way, but at least it is a way without breaking BC!

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -128,6 +128,11 @@ final class ConfigurationResolver
     private $usingCache;
 
     /**
+     * @var FixerFactory
+     */
+    private $fixerFactory;
+
+    /**
      * ConfigurationResolver constructor.
      *
      * @param ConfigInterface $config
@@ -513,15 +518,15 @@ final class ConfigurationResolver
      */
     private function createFixerFactory()
     {
-        static $fixerFactory = null;
-
-        if (null === $fixerFactory) {
+        if (null === $this->fixerFactory) {
             $fixerFactory = new FixerFactory();
             $fixerFactory->registerBuiltInFixers();
             $fixerFactory->registerCustomFixers($this->getConfig()->getCustomFixers());
+
+            $this->fixerFactory = $fixerFactory;
         }
 
-        return $fixerFactory;
+        return $this->fixerFactory;
     }
 
     /**
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -529,7 +529,10 @@ final class ConfigurationResolver
     private function getRuleSet()
     {
         if (null === $this->ruleSet) {
-            $this->ruleSet = new RuleSet($this->parseRules());
+            $rules = $this->parseRules();
+            $this->validateRules($rules);
+
+            $this->ruleSet = new RuleSet($rules);
         }
 
         return $this->ruleSet;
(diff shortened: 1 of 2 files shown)


src/RuleSetInterface.php
```diff
@@ -30,7 +30,7 @@ public static function create(array $set = array());
      *
      * @param string $rule name of rule
      *
-     * @return array|null
+     * @return array|bool
```
keradus: no. bc breaker

localheinz: Do you have a suggestion how to consistently reject undefined rules even when they are disabled without a BC break?

keradus: `RuleSet` is to resolve rule. one part of resolving is removing rules from sets that were disabled after all.
I see no value in returning disabled rules by RuleSet.

If wanted to validate, it should be validated somewhere else.

localheinz:
@keradus 

Closing, I'll solve this elsewhere for now!

src/RuleSetInterface.php
keradus: @localheinz 
Don't get me wrong. I see your point about DX and I agree with it.
I just don't find changing `RuleSet` a valid solution.

localheinz: @keradus 

Not a problem, no worries! Already implemented a solution. 👍

localheinz:
I have reopened this PR following @SpacePossum's suggestion in https://github.com/refinery29/php-cs-fixer-config/pull/127#issuecomment-265205756.

😆 Let me know what you think!

SpacePossum:
LGTM, thanks @localheinz : D

tests/Console/ConfigurationResolverTest.php
```diff
@@ -868,6 +868,22 @@ public function testResolveRulesWithOption()
...
+    public function testResolveRulesWithUnknownRules()
+    {
+        $this->setExpectedException(
+            'PhpCsFixer\ConfigurationException\InvalidConfigurationException',
+            'The rules contain unknown fixers (foo, bar).'
+        );
+
+        $resolver = new ConfigurationResolver(
+            $this->config,
+            array('rules' => 'foo,-bar'),
```
keradus: please add existing rule here as well, to be sure tat ex msg is not containing everything

src/Console/ConfigurationResolver.php
```diff
@@ -563,6 +566,48 @@ private function parseRules()
...
+        $ruleSet = RuleSet::create(array_map(function () {
+            return true;
+        }, $rules));
+
+        /* @var string[] $availableFixers */
+        $configuredFixers = array_keys($ruleSet->getRules());
+
+        $fixerFactory = new FixerFactory();
+        $fixerFactory->registerBuiltInFixers();
+        $fixerFactory->registerCustomFixers($this->getConfig()->getCustomFixers());
```
keradus: `$fixerFactory` creation is duplicated from `getFixers` method

src/Console/ConfigurationResolver.php
```diff
@@ -563,6 +566,48 @@ private function parseRules()
...
+
+        $reflection = new \ReflectionProperty('PhpCsFixer\FixerFactory', 'fixersByName');
+        $reflection->setAccessible(true);
```
keradus: eee, no. don't play with privates that you do not own

keradus:
:+1: for idea, some more work is needed

src/Console/ConfigurationResolver.php
localheinz: Hmm. The problem is, how do I find out which fixers have been registered by name? Should I add an accessor to `FixerFactory`, then?

src/Console/ConfigurationResolver.php
```diff
@@ -513,6 +509,18 @@ private function computeConfigFiles()
     }
 
     /**
+     * @return FixerFactory
+     */
+    private function createFixerFactory()
+    {
+        $fixerFactory = new FixerFactory();
```
keradus: shall be cached, no need for 2 instances while resolving the configuration

src/Console/ConfigurationResolver.php
```diff
@@ -513,9 +513,13 @@ private function computeConfigFiles()
      */
     private function createFixerFactory()
     {
-        $fixerFactory = new FixerFactory();
-        $fixerFactory->registerBuiltInFixers();
-        $fixerFactory->registerCustomFixers($this->getConfig()->getCustomFixers());
+        static $fixerFactory = null;
```
keradus: now it will be cached across different instances of `ConfigurationResolver`. very problematic for tests and using `PHP-CS-Fixer` as library. static is global, and then you base it on config, which may be different for different `ConfigurationResolver` instances

src/Console/ConfigurationResolver.php
```diff
@@ -513,6 +509,22 @@ private function computeConfigFiles()
     }
 
     /**
+     * @return FixerFactory
+     */
+    private function createFixerFactory()
+    {
+        static $fixerFactory = null;
```
keradus: now it will be cached across different instances of `ConfigurationResolver`. very problematic for tests and using `PHP-CS-Fixer` as library. static is global, and then you base it on config, which may be different for different `ConfigurationResolver` instances

keradus:
Thank you @localheinz.

localheinz:

SpacePossum:
thanks @localheinz :)

Issue #2365, opened 2016-12-07
Broken "Add support for user-defined whitespaces"


GrahamCampbell:
Causes `ErrorException·Undefined offset: 1` on some inputs. I can't confirm exactly how to reproduce this just yet, but I will do soon.

The issue is happening in `PhpCsFixer\DocBlock\Line::addBlank`:

![image](https://cloud.githubusercontent.com/assets/2829600/20967504/b84c3784-bc77-11e6-8a22-80670d6baa96.png)

Called inside `PhpCsFixer\Fixer\Phpdoc\PhpdocSeparationFixer::ensureAreSeparate`:

![image](https://cloud.githubusercontent.com/assets/2829600/20967516/cdd7abf6-bc77-11e6-8a65-66961be8da11.png)

ktomk:
@GrahamCampbell: Sounds like there was no regex match as that is unchecked.

    $matches += [1 => '', 2 => ''];

to provide defaults. If it's not expected to not match, perhaps turning this into a more saying exception would help next time to easier reproduce it.

GrahamCampbell:
But, this code didn't crash before the change, and now it does.

PR #2377, opened 2016-12-09
PhpdocSeparationFixer - Ignore incorrect PHPDoc

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2365

@GrahamCampbell happy debugging!

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocSeparationFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocSeparationFixerTest.php
@@ -24,6 +24,11 @@ final class PhpdocSeparationFixerTest extends AbstractFixerTestCase
 {
     public function testFix()
     {
+        $this->doTest('<?php
+/** @param EngineInterface $templating 
+*@return void
+*/');
+
         $expected = <<<'EOF'
 <?php
     /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Ah, thank you. :)

GrahamCampbell:
Well, I mean, naively just reverting to the original algorithm if the new one doesn't match the correct number of items works:

image

image

SpacePossum:
:+1:  however it would be nice if we could avoid the second `preg_match*` though

tests/Fixer/Phpdoc/PhpdocSeparationFixerTest.php
```diff
@@ -24,6 +24,11 @@
 {
     public function testFix()
     {
+        $this->doTest('<?php
+/** @param EngineInterface $templating
```
keradus: i would not treat it as proper docblock

still, tool must not crash on it, but i woulnd't say it's crucial to "fix" it

keradus:
Thank you @SpacePossum.

GrahamCampbell:
I don't like the way this fix is implemented. You made the line class silently fail. It should be the responsibility of the fixer to decide what to do imo.

keradus:
we quickly fix the tool to not crash, which was the main concern.
if you want to provide a fix to change phpdoc in that scenario - go ahead ;)

PR #2381, opened 2016-12-10
Tests - Do not use annotations for asserting exceptions

localheinz:
This PR

* [x] uses `setExpectedException()` and siblings instead of annotations for asserting exceptions to be thrown

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2351#pullrequestreview-12317080.

### DIFF
--- a/tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
+++ b/tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
@@ -429,33 +429,37 @@ EOF;
         $this->doTest($expected, $input);
     }
 
-    /**
-     * @expectedException \PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException
-     * @expectedExceptionMessageRegExp /^\[visibility_required\] Expected string got "NULL".$/
-     */
     public function testInvalidConfigurationType()
     {
+        $this->setExpectedExceptionRegExp(
+            'PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException',
+            '/^\[visibility_required\] Expected string got "NULL".$/'
+        );
+
         $this->fixer->configure(array(null));
     }
 
-    /**
-     * @expectedException \PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException
-     * @expectedExceptionMessageRegExp /^\[visibility_required\] Unknown configuration item "_unknown_", expected any of "property", "method", "const".$/
-     */
     public function testInvalidConfigurationValue()
     {
+        $this->setExpectedExceptionRegExp(
+            'PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException',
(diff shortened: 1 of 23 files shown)


tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
```diff
@@ -460,7 +460,6 @@ public function testInvalidConfigurationValueForPHPVersion()
             '/^\[visibility_required\] Invalid configuration item "const" for PHP ".+".$/'
         );
 
-
```
localheinz: 👍

keradus:
Thank you @localheinz.

garak:
This is a regression, since `setExpectedException` and similar were deprecated in phpunit sometimes ago, and there were removed in phpunit 6.

julienfalque:
This only concerns internal tests, which use PHPUnit 4 and 5 (composer.json has `"phpunit/phpunit": "^4.5|^5"`).

garak:
It's a regression anyway, because sooner or later we'll have to switch to a more recent (and still supported) version of phpunit

keradus:
moving from annotation to code is a good move. when move was done, the new method was not there yet.
we cannot move to it easily without bumping phpunit version, which we cannot done yet due to php reqs

garak:
Current php requirements are indeed compatible with phpunit 6.
I'm not against moving from annotation to code, but it should have been done without relying on deprecated methods.

SpacePossum:
Would've been nice, but not possible, don't get me wrong I would like to see the new method used as well.

What would be great IMO is when PHPUnit would release new minor updates of the older versions with the new method and deprecating the old one. That way we could move forward without this problem.
But than again this wasn't done so we are stuck for now.

keradus:
> Current php requirements are indeed compatible with phpunit 6.

no, PHPUnit requires PHP 7+, this project works on 5+ as well

garak:
@keradus that's what I said: being php 7 > 5, current requirements allow phpunit 6.
Anyway, I see the need for keeping the older (and unsupported) releases of php, but we're discussing about **dev** requirements, right? We can be more loose on that side.
Anyway, even if we don't upgrade to phpunit 6 *now*, it would be nice being able to do it at some point in the future, so getting deprecations is bad

localheinz:
🤦‍♂️

SpacePossum:
What I'm saying is that if new minor versions of older PHPUnit releases are released which contain the new method and deprecate the old one (which would be a good thing, not bad), then everyone could update their code using the new methods (and classes of that matter).
The problem is not in this repo, it is PHPUnit not providing an update path for using both 4,5 and 6.

garak:
Indeed, the old code (with annotations) could run in 4,5, and 6.
There are other things that are not working in 6, (like old names without namespace), but this is another question.

keradus:
@garak we came so far with dropping deprecations as it was possible keeping phpunit running on all our build matrix. it one will just update phpunit to version 6, we won't be able to use it.

Issue #1122, opened 2015-04-01
Empty line after control structures fixer


GrahamCampbell:
We need this fixer. :)

Soullivaneuh:
In which preset this one should be added? Symfony? cc @javiereguiluz

keradus:
Symfony is not fully follow that rule.

like
```php
$foo = ...;
if ($foo) {
    ...
}
```

Soullivaneuh:
@keradus If I well understand, with this code:

```php
$foo = ...;
if ($foo) {
    ...
}
$bar = $foo;
```

With Symfony CS, it should be:

```php
$foo = ...;
if ($foo) {
    ...
}

$bar = $foo;
```

With the fixer asked by @GrahamCampbell, is should be:

```php
$foo = ...;

if ($foo) {
    ...
}

$bar = $foo;
```

Right?

Well, in this case, why not two fixers?

* `empty_line_after_control_structure` -> On Symfony
* `empty_line_before_control_structure` -> On Contrib

What do you think?

SpacePossum:
one with config, but only after the current issue/PR count is down ^_^

SpacePossum:
Having two fixers on 1.x and one on master will be a merge hell.

Having one fixer is more logical to me, it prevents code duplication, having less code to write, debug, maintain and document, 
The diff with those three fixers is those have diff. logic to find and test candidate tokens to fix;
namespace, open tag and return.
The new fixers would not, both fixers would have the same logic to find candidates; control structure.
Btw. refactor of `blank_line_before_return` is already proposed to supported more cases using configuration on master.

Soullivaneuh:
> Having two fixers on 1.x and one on master will be a merge hell.

Agree. :+1: 

Keeping two fixers on master could be great too IMHO.

> Having one fixer is more logical to me, it prevents code duplication, having less code to write, debug, maintain and document, 
> The diff with those three fixers is those have diff. logic to find and test candidate tokens to fix;
namespace, open tag and return.

Why not an abstract fixer class for common code? This is already done.

SpacePossum:
>Why not an abstract fixer class for common code? This is already done.

you would still analyze to find the same candidates twice for each file, slowing down the app

Soullivaneuh:
> you would still analyze to find the same candidates twice for each file, slowing down the app

Oh, I see your concern.

keradus:
[...]

PR #2384, opened 2016-12-11
Add BlankLineBeforeStatementFixer

localheinz:
This PR

* [x] allows to configure the `ReturnNotation\BlankLineBeforeReturnFixer`
* [x] renames the `ReturnNotation\BlankLineBeforeReturnFixer` to  `Whitespace\BlankLineBeforeControlStatementFixer`
* [x] provides a deprecated `ReturnNotation\BlankLineBeforeReturnFixer`, proxying to the `Whitespace\BlankLineBeforeControlStatementFixer`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -236,19 +236,20 @@ Choose from the list of available rules:
   Ensure there is no code on the same line as the PHP open tag and it is
   followed by a blank line.
 
-* **blank_line_before_control_statement** [@Symfony]
-
-  An empty line feed should precede a control statement.
-
-  Configuration options:
-
-  - ``statements`` (``array``): list of control statements to fix; defaults to
-    ``['return']``
-
 * **blank_line_before_return**
 
   An empty line feed should precede a return statement (deprecated, use
-  ``blank_line_before_control_statement`` instead).
+  ``blank_line_before_statement`` instead).
+
+* **blank_line_before_statement** [@Symfony]
+
+  An empty line feed must precede any configured statement.
+
+  Configuration options:
+
+  - ``statements`` (``array``): list of statements which must be must be preceded by
+    an empty line; defaults to ``['break', 'continue', 'declare', 'return',
+    'throw', 'try']``
 
 * **braces** [@PSR2, @Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -236,9 +236,19 @@ Choose from the list of available rules:
   Ensure there is no code on the same line as the PHP open tag and it is
   followed by a blank line.
 
-* **blank_line_before_return** [@Symfony]
+* **blank_line_before_control_statement** [@Symfony]
 
-  An empty line feed should precede a return statement.
+  An empty line feed should precede a control statement.
+
+  Configuration options:
+
+  - ``statements`` (``array``): list of control statements to fix; defaults to
+    ``['return']``
+
+* **blank_line_before_return**
+
+  An empty line feed should precede a return statement (deprecated, use
+  ``blank_line_before_control_statement`` instead).
 
 * **braces** [@PSR2, @Symfony]
 
--- a/src/Fixer/ClassNotation/OrderedClassElementsFixer.php
+++ b/src/Fixer/ClassNotation/OrderedClassElementsFixer.php
@@ -93,12 +93,14 @@ final class OrderedClassElementsFixer extends AbstractFixer implements Configura
 
             if (!$parents) {
                 $this->typePosition[$type] = null;
+
                 continue;
             }
 
             foreach ($parents as $parent) {
                 if (isset($this->typePosition[$parent])) {
                     $this->typePosition[$type] = $this->typePosition[$parent];
+
                     continue 2;
                 }
             }
@@ -268,11 +270,13 @@ final class Example
 
                 if ($token->isGivenKind(T_STATIC)) {
                     $element['static'] = true;
+
                     continue;
                 }
 
                 if ($token->isGivenKind([T_PROTECTED, T_PRIVATE])) {
                     $element['visibility'] = strtolower($token->getContent());
+
                     continue;
                 }
 
@@ -289,6 +293,7 @@ final class Example
                 }
 
                 $element['end'] = $this->findElementEnd($tokens, $i);
+
                 break;
             }
 
@@ -391,6 +396,7 @@ final class Example
                     if ('phpunit' === $type) {
                         $element['position'] += $phpunitPositions[$element['name']];
                     }
+
                     continue;
                 }
 
--- a/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
+++ b/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
@@ -12,17 +12,20 @@
 
 namespace PhpCsFixer\Fixer\ReturnNotation;
 
-use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\AbstractProxyFixer;
+use PhpCsFixer\Fixer\Whitespace\BlankLineBeforeControlStatementFixer;
 use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Tokenizer\Token;
-use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\WhitespacesFixerConfig;
 
 /**
+ * @deprecated
+ *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
  */
-final class BlankLineBeforeReturnFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+final class BlankLineBeforeReturnFixer extends AbstractProxyFixer implements WhitespacesAwareFixerInterface
 {
     /**
      * {@inheritdoc}
@@ -30,7 +33,7 @@ final class BlankLineBeforeReturnFixer extends AbstractFixer implements Whitespa
     public function getDefinition()
     {
         return new FixerDefinition(
-            'An empty line feed should precede a return statement.',
+            'An empty line feed should precede a return statement (deprecated, use `blank_line_before_control_statement` instead).',
             [new CodeSample("<?php\nfunction A()\n{\n    echo 1;\n    return 1;\n}")]
         );
     }
@@ -38,58 +41,16 @@ final class BlankLineBeforeReturnFixer extends AbstractFixer implements Whitespa
     /**
      * {@inheritdoc}
      */
-    public function getPriority()
+    public function setWhitespacesConfig(WhitespacesFixerConfig $config)
     {
-        // should be run after NoUselessReturnFixer
-        return -19;
+        $this->proxyFixer->setWhitespacesConfig($config);
     }
 
     /**
      * {@inheritdoc}
      */
-    public function isCandidate(Tokens $tokens)
+    protected function createProxyFixer()
     {
-        return $tokens->isTokenKindFound(T_RETURN);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
-    {
-        $lineEnding = $this->whitespacesConfig->getLineEnding();
-
-        for ($index = 0, $limit = $tokens->count(); $index < $limit; ++$index) {
-            $token = $tokens[$index];
-
-            if (!$token->isGivenKind(T_RETURN)) {
-                continue;
-            }
-
-            $prevNonWhitespaceToken = $tokens[$tokens->getPrevNonWhitespace($index)];
-
-            if (!$prevNonWhitespaceToken->equalsAny([';', '}'])) {
-                continue;
-            }
-
-            $prevIndex = $index - 1;
-            $prevToken = $tokens[$prevIndex];
-
-            if ($prevToken->isWhitespace()) {
-                $parts = explode("\n", $prevToken->getContent());
-                $countParts = count($parts);
-
-                if (1 === $countParts) {
-                    $tokens[$prevIndex] = new Token([T_WHITESPACE, rtrim($prevToken->getContent(), " \t").$lineEnding.$lineEnding]);
-                } elseif (count($parts) <= 2) {
-                    $tokens[$prevIndex] = new Token([T_WHITESPACE, $lineEnding.$prevToken->getContent()]);
-                }
-            } else {
-                $tokens->insertAt($index, new Token([T_WHITESPACE, $lineEnding.$lineEnding]));
-
-                ++$index;
-                ++$limit;
-            }
-        }
+        return new BlankLineBeforeControlStatementFixer();
     }
 }
--- a/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
+++ b/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
@@ -13,14 +13,14 @@
 namespace PhpCsFixer\Fixer\ReturnNotation;
 
 use PhpCsFixer\AbstractProxyFixer;
-use PhpCsFixer\Fixer\Whitespace\BlankLineBeforeControlStatementFixer;
+use PhpCsFixer\Fixer\Whitespace\BlankLineBeforeStatementFixer;
 use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\WhitespacesFixerConfig;
 
 /**
- * @deprecated since 2.4, replaced by BlankLineBeforeControlStatementFixer
+ * @deprecated since 2.4, replaced by BlankLineBeforeStatementFixerTest
  *
  * @todo To be removed at 3.0
  *
@@ -35,7 +35,7 @@ final class BlankLineBeforeReturnFixer extends AbstractProxyFixer implements Whi
     public function getDefinition()
     {
         return new FixerDefinition(
-            'An empty line feed should precede a return statement (deprecated, use `blank_line_before_control_statement` instead).',
+            'An empty line feed should precede a return statement (deprecated, use `blank_line_before_statement` instead).',
             [new CodeSample("<?php\nfunction A()\n{\n    echo 1;\n    return 1;\n}")]
         );
     }
@@ -53,7 +53,7 @@ final class BlankLineBeforeReturnFixer extends AbstractProxyFixer implements Whi
      */
     protected function createProxyFixer()
     {
-        $fixer = new BlankLineBeforeControlStatementFixer();
+        $fixer = new BlankLineBeforeStatementFixer();
         $fixer->configure(['statements' => ['return']]);
 
         return $fixer;
--- a/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
+++ b/src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
@@ -20,7 +20,9 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\WhitespacesFixerConfig;
 
 /**
- * @deprecated
+ * @deprecated since 2.4, replaced by BlankLineBeforeControlStatementFixer
+ *
+ * @todo To be removed at 3.0
  *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  * @author Andreas Möller <am@localheinz.com>
@@ -51,6 +53,9 @@ final class BlankLineBeforeReturnFixer extends AbstractProxyFixer implements Whi
      */
     protected function createProxyFixer()
     {
-        return new BlankLineBeforeControlStatementFixer();
+        $fixer = new BlankLineBeforeControlStatementFixer();
+        $fixer->configure(['statements' => ['return']]);
+
+        return $fixer;
     }
 }
--- a/src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
+++ b/src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
@@ -1,0 +1,259 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Whitespace;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class BlankLineBeforeControlStatementFixer extends AbstractFixer implements ConfigurableFixerInterface, WhitespacesAwareFixerInterface
+{
+    /**
+     * @var array
+     */
+    private static $defaultConfiguration = [
+        'return',
+    ];
+
+    /**
+     * @var array
+     */
+    private static $tokenMap = [
+        'break' => T_BREAK,
+        'continue' => T_CONTINUE,
+        'declare' => T_DECLARE,
+        'do' => T_DO,
+        'else' => T_ELSE,
+        'elseif' => T_ELSEIF,
+        'for' => T_FOR,
+        'foreach' => T_FOREACH,
+        'if' => T_IF,
+        'include' => T_INCLUDE,
+        'include_once' => T_INCLUDE_ONCE,
+        'require' => T_REQUIRE,
+        'require_once' => T_REQUIRE_ONCE,
+        'return' => T_RETURN,
+        'switch' => T_SWITCH,
+        'throw' => T_THROW,
+        'try' => T_TRY,
+        'while' => T_WHILE,
+    ];
+
+    /**
+     * @var array
+     */
+    private $fixTokenMap = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $configuration = self::$defaultConfiguration;
+        }
+
+        foreach ($configuration as $controlStatement) {
+            if (!is_string($controlStatement) || !array_key_exists($controlStatement, self::$tokenMap)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf(
+                    'Expected one of "%s", got "%s".',
+                    implode('", "', array_keys(self::$tokenMap)),
+                    is_object($controlStatement) ? get_class($controlStatement) : gettype($controlStatement)
+                ));
+            }
+
+            $this->fixTokenMap[$controlStatement] = self::$tokenMap[$controlStatement];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+
+        for ($index = 0, $limit = $tokens->count(); $index < $limit; ++$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(array_values($this->fixTokenMap))) {
+                continue;
+            }
+
+            $prevNonWhitespaceToken = $tokens[$tokens->getPrevNonWhitespace($index)];
+
+            if (!$prevNonWhitespaceToken->equalsAny([';', '}'])) {
+                continue;
+            }
+
+            $prevToken = $tokens[$index - 1];
+
+            if ($prevToken->isWhitespace()) {
+                $parts = explode("\n", $prevToken->getContent());
+                $countParts = count($parts);
+
+                if (1 === $countParts) {
+                    $prevToken->setContent(rtrim($prevToken->getContent(), " \t").$lineEnding.$lineEnding);
+                } elseif (count($parts) <= 2) {
+                    $prevToken->setContent($lineEnding.$prevToken->getContent());
+                }
+            } else {
+                $tokens->insertAt($index, new Token([T_WHITESPACE, $lineEnding.$lineEnding]));
+
+                ++$index;
+                ++$limit;
+            }
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'An empty line feed should precede a control statement.',
+            [
+                new CodeSample(
+                    '<?php
+function A() {
+    echo 1;
+    return 1;
+}'
+                ),
+                new CodeSample(
+                    '<?php
+switch ($foo) {
+    case 42:
+        $bar->process();
+        break;
+    case 44:
+        break;
+}',
+                    [
+                        'break',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+foreach ($foo as $bar) {
+    if ($bar->isTired()) {
+        $bar->sleep();
+        continue;
+    }
+}',
+                    [
+                        'continue',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+$i = 0;
+do {
+    echo $i;
+} while ($i > 0);
+',
+                    [
+                        'do',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+$a = 9000;
+if (true) {
+    $foo = $bar;
+}',
+                    [
+                        'if',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+
+if (true) {
+    $foo = $bar;
+    return;
+}',
+                    [
+                        'return',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+$a = 9000;
+switch ($a) {
+    case 42:
+        break;
+}',
+                    [
+                        'switch',
+                    ]
+                ),
+                new CodeSample(
+'<?php
+if (null === $a) {
+    $foo->bar();
+    throw new \UnexpectedValueException("A cannot be null");
+}',
+                    [
+                        'throw',
+                    ]
+                ),
+                new CodeSample(
+'<?php
+$a = 9000;
+try {
+    $foo->bar();
+} catch (\Exception $exception) {
+    $a = -1;
+}',
+                    [
+                        'try',
+                    ]
+                ),
+            ],
+            null,
+            sprintf(
+                'Provide an array with any of "%s".',
+                implode('", "', array_keys(self::$tokenMap))
+            ),
+            self::$defaultConfiguration
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run after NoUselessReturnFixer
+        return -19;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(array_values($this->fixTokenMap));
+    }
+}
--- a/src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
+++ b/src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
@@ -13,9 +13,11 @@
 namespace PhpCsFixer\Fixer\Whitespace;
 
 use PhpCsFixer\AbstractFixer;
-use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
-use PhpCsFixer\Fixer\ConfigurableFixerInterface;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerConfiguration\FixerOptionValidatorGenerator;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Token;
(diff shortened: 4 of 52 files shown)


src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,228 @@
...
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class BlankLineBeforeControlStatementFixer extends AbstractFixer implements ConfigurableFixerInterface, WhitespacesAwareFixerInterface
```
localheinz: It's entirely copied from the `BlankLineBeforeReturnStatementFixer`, only added the token map, configuration, and description.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * @var array
+     */
+    private static $tokenMap = array(
+        'break' => T_BREAK,
+        'continue' => T_CONTINUE,
+        'if' => T_IF,
+        'return' => T_RETURN,
+        'switch' => T_SWITCH,
```
SpacePossum: maye `throws` as well?

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,228 @@
...
+            }
+
+            $this->fixTokenMap[$controlStatement] = self::$tokenMap[$controlStatement];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
```
SpacePossum: please move after the getters (i.e. sort public the methods A-Z, normally I wouldn't care, but I'm re-ordering those everywhere atm)

SpacePossum:
Like the idea :+1: 

Maybe its time we can start deprecating fixers. This would be mean removing them from the documentation but adding those to on upgrading file. Triggering deprecation warning when deprecated fixer is configured. Wonder what people think about it.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
localheinz: Also added `try`!

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,255 @@
...
+
+                ++$index;
+                ++$limit;
+            }
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
```
localheinz: Added a docblock here, however, in projects I work on I usually omit a docblock of this kind if

* it is overriding a method in a parent class (that has a docblock)
* implements an interface (that has a docblock)

as I believe they do not really add any value - aren't they automatically inherited?

Not sure, wasn't there a PR for adding a fixer that removes 

```php
/**
 * {@inheritdoc}
*/
```

as well?

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
keradus: very big :-1: on that.

Please do the following commits (so it will be easier to review by not having all added, all removed, but sth modified):

1. change the behaviour of existing fixer to have that configuration (ignore the old name)
2. rename the fixer
3. create fixer with old name that will be a proxyfixer using new fixer with given configuration, fixer shall be makred as deprecated

localheinz: Done!

README.rst
```diff
@@ -190,6 +190,10 @@ Choose from the list of available rules:
    | Ensure there is no code on the same line as the PHP open tag and it is
    | followed by a blank line.
 
+* **blank_line_before_control_statement**
+   | An empty line feed should precede a control statement.
+   | *Rule is: configurable.*
```
keradus: please remove deprecated fixer from `@Symfony` ruleset and include this one instead

src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
```diff
@@ -13,55 +13,37 @@
...
 use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\WhitespacesFixerConfig;
 
 /**
+ * @deprecated
+ *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
  */
 final class BlankLineBeforeReturnFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
```
keradus: what about extending `AbstractProxyFixer` ?

keradus:
Please, at least in `::getDefinition` method of deprecated fixer add `summary` to mention that fixer has been deprecated.

localheinz:
@keradus 

Incorporated all of the suggested changes, thanks for reviewing!

src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
```diff
@@ -26,25 +26,12 @@
...
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function fix(\SplFileInfo $file, Tokens $tokens)
-    {
-        $this->fixer->fix($file, $tokens);
-    }
+    protected $proxyFixer;
```
keradus: already in base class, no need to redeclare

src/Fixer/ReturnNotation/BlankLineBeforeReturnFixer.php
```diff
@@ -12,83 +12,51 @@
...
-                }
-            } else {
-                $tokens->insertAt($index, new Token(array(T_WHITESPACE, $lineEnding.$lineEnding)));
-
-                ++$index;
-                ++$limit;
-            }
-        }
-    }
+    protected $proxyFixer;
```
keradus: already in base class, no need to redeclare

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
keradus: why... make a fixer that would reorder them instead ;)
i hope that is what you meant :D

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,240 @@
...
+    $a = -1;
+}',
+                    array(
+                        'try',
+                    )
+                ),
+            ),
+            null,
+            sprintf(
+                'Provide an array with any of "%s".',
```
keradus: I would rather make it an object instead of flat array.
like `['statements' => [try, continue]]` instead of just `[try, continue]`, so we will be able to add different options in the future (like eg "how many lines")

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,240 @@
...
+     * @var array
+     */
+    private static $defaultConfiguration = array(
+        'return',
+    );
+
+    /**
+     * @var array
+     */
+    private static $tokenMap = array(
```
keradus: looking at the fixer name, all control statements should be here:
http://php.net/manual/en/language.control-structures.php

SpacePossum: :+1:, but - `else`, - `elseif`, - `alternative syntax`,- `goto`, - `declare` (prevent conflict)

localheinz: Will update later!

localheinz: This needs more tests, and some of the cases are a bit ugly, I believe:

* `else` (with braces)
* `elseif` (with braces)
* `while` (with braces)

Maybe more.

localheinz: Urgh, alternative syntax!

SpacePossum: don't bother alternative syntax, try not to break it and thats it ;)

keradus:
@localheinz 
what is the state here? how could we continue here?

localheinz:
@keradus 

Ha, right, need to take a closer look at the extra control statements!

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1046 @@
...
+    public function providerFixWithElse()
+    {
+        return array(
+            array(
+'<?php
+if (true) {
+    echo $b;
+} 
+
+else {
```
localheinz: So this is what passes the tests, but personally, I would want it to fix this to

```php
<?php
if (true) {
    echo $b;

} else {
    echo $c;
}
```

Well, actually, I wouldn't want it to be fixed to that, but I _think_ that maybe that is what other would want this to be fixed to, what do you think?

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1046 @@
...
+    public function providerFixWithElseIf()
+    {
+        return array(
+            array(
+'<?php
+if ($c) {
+    echo $b;
+} 
+
+elseif ($b) {
```
localheinz: So this is what passes the tests, but personally, I would want it to fix this to

```php
<?php
if ($c) {
    echo $b;

} elseif ($b) {
    echo $c;
}
```

(See above)

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1046 @@
...
+include_once "foo.php";',
+            ),
+            array(
+'<?php
+$foo = $bar;
+
+include_once "foo.php";',
+'<?php
+$foo = $bar;
+include_once "foo.php";',
```
localheinz: What I think is missing here is cases where we have multiple `include_once` followed by each other, _personally_ I would only want to have a blank line before the first in a block of `include_once` statements.

What do you think?

(Same applies to `include`, 'require`, `require_once`)

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,263 @@
...
+        'foreach' => T_FOREACH,
+        'if' => T_IF,
+        'include' => T_INCLUDE,
+        'include_once' => T_INCLUDE_ONCE,
+        'require' => T_REQUIRE,
+        'require_once' => T_REQUIRE_ONCE,
+        'return' => T_RETURN,
+        'switch' => T_SWITCH,
+        'throw' => T_THROW,
+        'try' => T_TRY,
```
localheinz: Should we also add `catch` and `finally`?

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: if you mean fixing by removing the empty line than I would suggest don't bother, we've a fixer for that ;)

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
SpacePossum: I'm -1 on those, however since these can be configured I'm happy with it any way :)

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: hmm.. I don't think people have a lot of these types of blocks, therefor I wonder if it worth the additional dev. time and processing time to detect such cases.
If you want to go for it, please do so, but for me it would not be a show stopper for this PR

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1046 @@
...
+        break;
+}',
+            ),
+            array(
+'<?php
+switch ($a) {
+    case 42:
+        $foo = $bar;
+
+        break;
```
julienfalque: When inside a switch, I also remove any blank lines between `break` and next `case`/`default`, should this fixer do that too?

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
julienfalque: If #2500 is merged, this will be the case anyway, doing it here will avoid a deprecation in the future if this is merged first.

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: I follow the same style, so I get the suggestion :)
However I think this fixer should be limited to adding lines and not also remove lines.
You're suggestion would be a great addition to `no_extra_consecutive_blank_lines`, which already has a lot of configuration options so this one should be able to fit it.

SpacePossum:
this fixer is possibly in conflict with the `no_extra_consecutive_blank_lines`

Example input
```php
<?php
if ($c) {
    echo $b;
} elseif ($b) {
    echo $c;
}
```

by configuration this fixer fixes to
```php
<?php
if ($c) {
    echo $b;

} elseif ($b) {
    echo $c;
}
```

by configuration `no_extra_consecutive_blank_lines` (`curly_brace_block`) fixes the fixed output back to the input.
Not sure how to deal with this, we could do more advanced configuration checking for conflicts.

Maybe for now lets more without worrying :)

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
julienfalque: I just looked at `no_extra_consecutive_blank_lines` and it already accepts a `break` option that seems to do exactly that, so I think we're good :)

julienfalque:
I think it doesn't make sense to configure both at the same time anyway.

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: funny that I start forgetting what I made ;)

SpacePossum:
@localheinz could you remove the trailing white space in BlankLineBeforeControlStatementFixer?
After that and a rebase, I think only https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2384#discussion_r100136355 is left? TBH I think we can do without that case. 
I would really like to do a final review soon and get this in :)

julienfalque:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2384#discussion_r96342585 should be addressed too :)

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
keradus: yep, please do the needful

localheinz:
Sorry for the delay, picking this up again!

keradus:
Please rebase

SpacePossum:
Hi @localheinz how are you?
Your branch is bitten by some refactors I'm afraid, but a rebase shouldn't be very hard.
Let me know if you going to pick this one up soon, otherwise I'm happy to continue on your great work :)

localheinz:
@SpacePossum 

Ha, if you would like to take over, that would be great!

keradus:
@localheinz , I did squash all your commits into one and rebase it on newest master (during that I had to remove trailing whitespaces as side effect, as I didn't wanted to change my git config to allow them)
Then, I make some adjustments to make it works with currect codebase.

@SpacePossum @julienfalque , anything critical here? don't be picky, I bet we won't introduce new tests and so on here. I just wanna help finishing and merging this PR.
Only think I'm currently thinking about is deciding about default configuration and configuration for us in `.php_cs.dist` file.

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1020 @@
...
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Test\AbstractFixerTestCase;
+use PhpCsFixer\WhitespacesFixerConfig;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ * @coversNothing
```
julienfalque: Not correct.

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -107,6 +107,7 @@ public function getFixersPriorityCases()
             [$fixers['no_useless_return'], $fixers['no_whitespace_in_blank_line']], // tested also in: no_useless_return,no_whitespace_in_blank_line.test
             [$fixers['no_useless_return'], $fixers['no_extra_consecutive_blank_lines']], // tested also in: no_useless_return,no_extra_consecutive_blank_lines.test
             [$fixers['no_useless_return'], $fixers['blank_line_before_return']], // tested also in: no_useless_return,blank_line_before_return.test
+            [$fixers['no_useless_return'], $fixers['blank_line_before_control_statement']],
```
SpacePossum: note @ self; todo rename no_empty_phpdoc,no_extra_consecutive_blank_lines.test

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,265 @@
...
+            $prevNonWhitespaceToken = $tokens[$tokens->getPrevNonWhitespace($index)];
+
+            if (!$prevNonWhitespaceToken->equalsAny([';', '}'])) {
+                continue;
+            }
+
+            $prevToken = $tokens[$index - 1];
+
+            if ($prevToken->isWhitespace()) {
+                $parts = explode("\n", $prevToken->getContent());
```
SpacePossum: note @ self; todo `$parts` is only used for counting, so a line break (substring) count should be enough

SpacePossum:
looking good,
I would leave the default to `return` for now, we can always apply more to own config later after discussion and PR that separately.

tests/AutoReview/FixerFactoryTest.php
keradus: yeah, originally there was a comment with file, but actually no file was commited :(

keradus:
I rather think about defualt config (not only for our project), to not change it in future that soon (as it would be BC breaker)

localheinz:
I still don't have tests for all of control statements, urgh.

localheinz:
@keradus 

Running the tests fails with

```
PhpCsFixer\Tokenizer\Token::clearChanged is deprecated and will be removed in 3.0: 15033x
    6147x in IntegrationTest::testIntegration from PhpCsFixer\Tests
    5260x in FixerTest::testFixerDefinitions from PhpCsFixer\Tests\AutoReview
    1119x in DescribeCommandTest::testDescribeCommand from PhpCsFixer\Tests\AutoReview
    500x in ConfigTest::testCustomConfig from PhpCsFixer\Tests
    273x in BlankLineBeforeControlStatementFixerTest::testFixWithBreak from PhpCsFixer\Tests\Fixer\Whitespace
    222x in BlankLineBeforeControlStatementFixerTest::testFixWithContinue from PhpCsFixer\Tests\Fixer\Whitespace
    156x in BlankLineBeforeControlStatementFixerTest::testFixWithElseIf from PhpCsFixer\Tests\Fixer\Whitespace
    153x in BlankLineBeforeControlStatementFixerTest::testFixWithWhile from PhpCsFixer\Tests\Fixer\Whitespace
    132x in BlankLineBeforeControlStatementFixerTest::testFixWithElse from PhpCsFixer\Tests\Fixer\Whitespace
    120x in BlankLineBeforeControlStatementFixerTest::testFixWithTry from PhpCsFixer\Tests\Fixer\Whitespace
    111x in BlankLineBeforeReturnFixerTest::testFix from PhpCsFixer\Tests\Fixer\ReturnNotation
    108x in BlankLineBeforeControlStatementFixerTest::testFixWithDefaultConfiguration from PhpCsFixer\Tests\Fixer\Whitespace
    108x in BlankLineBeforeControlStatementFixerTest::testFixWithReturn from PhpCsFixer\Tests\Fixer\Whitespace
    87x in BlankLineBeforeControlStatementFixerTest::testFixWithThrow from PhpCsFixer\Tests\Fixer\Whitespace
    75x in BlankLineBeforeControlStatementFixerTest::testFixWithSwitch from PhpCsFixer\Tests\Fixer\Whitespace
    72x in BlankLineBeforeReturnFixerTest::testMessyWhitespaces from PhpCsFixer\Tests\Fixer\ReturnNotation
    72x in BlankLineBeforeControlStatementFixerTest::testFixWithReturnAndMessyWhitespaces from PhpCsFixer\Tests\Fixer\Whitespace
    66x in BlankLineBeforeControlStatementFixerTest::testFixWithIf from PhpCsFixer\Tests\Fixer\Whitespace
    60x in BlankLineBeforeControlStatementFixerTest::testFixWithDo from PhpCsFixer\Tests\Fixer\Whitespace
    48x in BlankLineBeforeControlStatementFixerTest::testFixWithDeclare from PhpCsFixer\Tests\Fixer\Whitespace
    36x in BlankLineBeforeControlStatementFixerTest::testFixWithRequireOnce from PhpCsFixer\Tests\Fixer\Whitespace
    36x in BlankLineBeforeControlStatementFixerTest::testFixWithIncludeOnce from PhpCsFixer\Tests\Fixer\Whitespace
    36x in BlankLineBeforeControlStatementFixerTest::testFixWithInclude from PhpCsFixer\Tests\Fixer\Whitespace
    36x in BlankLineBeforeControlStatementFixerTest::testFixWithRequire from PhpCsFixer\Tests\Fixer\Whitespace

PhpCsFixer\Tokenizer\Token::setContent is deprecated and will be removed in 3.0: 59x
    10x in DescribeCommandTest::testDescribeCommand from PhpCsFixer\Tests\AutoReview
    10x in FixerTest::testFixerDefinitions from PhpCsFixer\Tests\AutoReview
    4x in IntegrationTest::testIntegration from PhpCsFixer\Tests
    3x in BlankLineBeforeControlStatementFixerTest::testFixWithReturn from PhpCsFixer\Tests\Fixer\Whitespace
    3x in BlankLineBeforeControlStatementFixerTest::testFixWithBreak from PhpCsFixer\Tests\Fixer\Whitespace
    3x in BlankLineBeforeControlStatementFixerTest::testFixWithContinue from PhpCsFixer\Tests\Fixer\Whitespace
    3x in BlankLineBeforeReturnFixerTest::testFix from PhpCsFixer\Tests\Fixer\ReturnNotation
    3x in BlankLineBeforeControlStatementFixerTest::testFixWithDefaultConfiguration from PhpCsFixer\Tests\Fixer\Whitespace
    2x in BlankLineBeforeReturnFixerTest::testMessyWhitespaces from PhpCsFixer\Tests\Fixer\ReturnNotation
    2x in BlankLineBeforeControlStatementFixerTest::testFixWithWhile from PhpCsFixer\Tests\Fixer\Whitespace
    2x in BlankLineBeforeControlStatementFixerTest::testFixWithElse from PhpCsFixer\Tests\Fixer\Whitespace
    2x in BlankLineBeforeControlStatementFixerTest::testFixWithElseIf from PhpCsFixer\Tests\Fixer\Whitespace
    2x in BlankLineBeforeControlStatementFixerTest::testFixWithReturnAndMessyWhitespaces from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithTry from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithThrow from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithSwitch from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithIncludeOnce from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithRequireOnce from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithRequire from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithInclude from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithDo from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithDeclare from PhpCsFixer\Tests\Fixer\Whitespace
    1x in BlankLineBeforeControlStatementFixerTest::testFixWithIf from PhpCsFixer\Tests\Fixer\Whitespace
```

Not sure, what do I need to adjust here?

SpacePossum:
>PhpCsFixer\Tokenizer\Token::setContent

this one should be replaced with something like;
`$tokens[$index] = new Token(...);`

>PhpCsFixer\Tokenizer\Token::clearChanged 

not sure what changes in this PR triggers that one, I'll have a look

keradus:
done

SpacePossum:
there is one option for `while`, however I think some people might like to differentiate;
```php
<?php
do { 
    echo 1;  
} while($a); // no empty line above
```

and
```php
<?php

echo 1;

while($a) { // line above is wanted
}
```

keradus:
> > PhpCsFixer\Tokenizer\Token::clearChanged
>
> not sure what changes in this PR triggers that one, I'll have a look

using `->setContent` causes to use `clearChanged`, as Tokens collection starts to work in legacy mode

SpacePossum:
check, thanks, that explains the high number of warnings :)

SpacePossum:
all green now

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1085 @@
...
+        ];
+    }
+
+    /**
+     * @dataProvider providerFixWithReturnAndMessyWhitespaces
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFixWithReturnAndMessyWhitespaces($expected, $input = null)
```
julienfalque: Instead of using specific test cases for different indent and line ending, I would reuse `providerFixWithReturn()` data provider and `str_replace("\n", "\r\n", ...)` on input and expected codes before calling `$this->doTest(...)`. I would also do that for all other keywords.

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
julienfalque: I would say that multiple include/require still happen quite a lot, but mostly on legacy codebases, probably not using Composer. So I guess it's fine if we don't support this scenario.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
julienfalque: 👍 for `catch` and `finally`, they are quite similar to `else` or `elseif` so why not support them as well?

SpacePossum: I'm -1 on `else` and `elseif` TBH, you know of any (major) code base / community formatting to?

```php
if ($a) {
    echo 1;

} else {
    echo 2;
}
```

```php
try {
    $a();
    $b();

} catch(\Exception $e) {

}
```

adding options and tests for such options and documentation all adds up in effort and time, so if the only reason is `similarity` but no real-live sample than I'm -1

julienfalque: Yes, the only reason is that `else` and `elseif` are already there. But I agree, it's better to just drop them all instead.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,264 @@
...
+        'return',
+    ];
+
+    /**
+     * @var array
+     */
+    private static $tokenMap = [
+        'break' => T_BREAK,
+        'continue' => T_CONTINUE,
+        'declare' => T_DECLARE,
```
SpacePossum: I don't think there is such a conflict as `DeclareStrictTypesFixer` inserts the `declare` as first statement (as required by PHP). Since this fixer only insert a line if there is either a `;` or `{` before the statement it cannot be in conflict.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,264 @@
...
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+
+        for ($index = 0, $limit = $tokens->count(); $index < $limit; ++$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(array_values($this->fixTokenMap))) {
```
SpacePossum: done

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
SpacePossum: done

SpacePossum:
TODO
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2384#issuecomment-307339864

and 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2384#issuecomment-306478618
(would do that as last thing as it produces a lot of changes making reviewing less fun)

julienfalque:
> there is one option for while, however I think some people might like to differentiate;
> 
> ```php
> <?php
> do { 
>     echo 1;  
> } while($a); // no empty line above
> ```
>
> (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2384/files#diff-c53d4c3333692f270d9f520f2d4a8c3bR1009 this line whould not be added, btw. should it be added between `while` and `}` and not above `}` ?)

I think it's the same as `else`, `elseif`, `catch` and `finally` so like you said, the fixer should not add a line here.

SpacePossum:
updated, fixer no longer adds a line above a T_WHILE if part of do-while

keradus:
Travis is sad

SpacePossum:
>Travis is sad

As discussed, PHPMD reports error on code not touched, stopping all other tests from running.
Lets make a call on default config first, than I deal with the SCA fallout

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: the current way is inline with the other tests in the repo, I see no reason to divert from it now

SpacePossum:
rebased, squashed, fixed unrelated PHPMD failure, added more tests.

@julienfalque please review again and update your previous preview accordingly. If you want more test case please provide those explicit cause I'm a bit confused about some suggestions you made in your comments.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,269 @@
...
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('statements', 'List of control statements to fix.'))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([
+                    (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf(array_keys(self::$tokenMap)),
+                ])
+                ->setDefault([
+                    'break',
+                    'continue',
+                    'do',
```
julienfalque: I still think `do`, `for`, `foreach`, `if`, `switch` and `while` should not be part of the default value. They are often preceded by variable initialization or something similar and adding a newline does not _always_ improve readability in such case IMO.

This statement could apply for other keywords, but less likely so I would keep them.

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
```diff
@@ -0,0 +1,1085 @@
...
+$foo = "bar";
+
+do {
+} while (true);',
+'<?php
+$foo = "bar";
+do {
+} while (true);',
+            ],
+        ];
```
julienfalque: I think this one still needs to be addressed. Some tests don't cover cases like:
```php
if ($foo) {
    for ($i = 0; $i < 9000; ++$i) {
        foo();
    }
}
```
which should not be changed, i.e. assert that the fixer does not add a newline before the `for` loop because it is at the beginning of a control structure.

julienfalque:
> If you want more test cases please provide those explicit cause I'm a bit confused about some suggestions you made in your comments.

Some tests don't cover cases like:
```php
if ($foo) {
    for ($i = 0; $i < 9000; ++$i) {
        foo();
    }
}
```
which should not be changed, i.e. assert that the fixer does not add a newline before the `for` loop because it is at the beginning of a control structure.

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -0,0 +1,269 @@
...
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class BlankLineBeforeControlStatementFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
```
keradus: control statement? `include_once` is not

julienfalque:
The fixer should not add newline after a comment, e.g.:
```php
// foo
if ($foo) { }
```
should not be changed.

Reading the code, it already handles such case but it looks like test cases are missing, except for `return`. Could you please add cases for other keywords?

tests/Fixer/Whitespace/BlankLineBeforeControlStatementFixerTest.php
SpacePossum: done

SpacePossum:
added the test, side note I don't see much value in testing all possible code constructs;
like

```php
if ($foo) {
    for ($i = 0; $i < 9000; ++$i) {
        foo();
    }
}
```

```php
while ($foo) {
    for ($i = 0; $i < 9000; ++$i) {
        foo();
    }
}
```

```php
do {
    for ($i = 0; $i < 9000; ++$i) {
        foo();
    }
}while ($foo);
```

there are just more of the same and the number of combinations is insane

SpacePossum:
test will fail because I didn't run the fixer on the repo yet, lets finish the discussion about the default configuration and the `number of tests`+1 cycle first

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
```diff
@@ -22,30 +22,23 @@
...
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  * @author Andreas Möller <am@localheinz.com>
+ * @author SpacePossum
  */
 final class BlankLineBeforeControlStatementFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
```
keradus: control statement? `include_once` is not

keradus:
:+1: for default config

keradus:
> these are just more of the same and the number of combinations is insane

not claiming to do it here, but we already generate test cases from templates providing different keywords ;)

src/Fixer/Whitespace/BlankLineBeforeControlStatementFixer.php
SpacePossum: fair point, shall we update the fixer to match the name (remove `include*` etc. as config option) or rename to something else?

keradus: `include` etc were added for some reason. do we find this reason valid? if so, then I would rename the fixer

julienfalque: IMO supporting statements like `include` or `require` is nice and maybe we could add more in the future. I would rename the fixer too.

SpacePossum: so `BlankLineBeforeControlStatementFixer.php` => `BlankLineBeforeStatementFixer.php` ?
like to hear other/better suggestions

localheinz: Sounds good to me!

src/Fixer/ClassNotation/OrderedClassElementsFixer.php
```diff
@@ -93,12 +93,14 @@ public function configure(array $configuration = null)
...
                 $this->typePosition[$type] = null;
+
                 continue;
             }
 
             foreach ($parents as $parent) {
                 if (isset($this->typePosition[$parent])) {
                     $this->typePosition[$type] = $this->typePosition[$parent];
+
                     continue 2;
```
SpacePossum: ^_^

SpacePossum:
all done, lets :ship: ?

src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'An empty line feed should precede a configured statement.',
```
julienfalque: I would update the description: _An empty line feed should precede **configured statements**._

src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('statements', 'List of control statements to fix.'))
```
julienfalque: Please remove _control_ here as well.

src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
SpacePossum: well thats an English language catch I think; as here there is a one-on-one relation which will be applied multiple times. I.e. 
>An empty line feed should precede configured statements.

to me suggests that the line would _only_ be added if there is more than one statement, which is not true.
So;

>An empty line feed should precede any configured statement.

?

SpacePossum:
TODO: regen README after copy discussion is done

src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
julienfalque: Much better 👍

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1122
(granted, only the part above a control statement, not after)

keradus:
Thank you @localheinz.

localheinz:

SpacePossum:
thanks @localheinz :D

keradus:
meh, it tooks only half year ;D

SpacePossum:
haha, oh well better to get it right the first time than rushing it :P

localheinz:
🤓

PR #2385, opened 2016-12-11
Order PHPDoc Params

localheinz:
This PR

* [x] adds a failing test case for the `PhpdocAddMissingParamAnnotationFixer`

💁‍♂️ Came across this when I tried to find out what is going to happen when applying this fixer on a legacy code base

### 1st Run

```php
$config = [
    'only_untyped' => true,
];
```

### 2nd Run

```php
$config = [
    'only_untyped' => false,
];
```

In the example the first parameter of a method was typed, the following weren't, so in the first run all parameters were added to the docblock (except the first, typed one), and in the 2nd run, the first, typed parameter was added to the docblock, but as last parameter.


keradus:
I do not consider this as a bug.
responsibility of a rule is to add missing annotation, it is always added at last place.
There were already requested to have a rule for sorting the params by order in method prototype, but this would be the other fixer (independent of adding missing params).

If you want to continue with this PR please make a new rule.

SpacePossum:
like to see that new fixer :)  i.e. order `@param`'s in the other of the function/method declaration

localheinz:
Hmm, the question is, what should a `PhpdocSortParamAnnotationFixer` do when

* there are extra parameters in the docblock (move past existing ones?)
* there are parameters missing (ignore those?)

Which makes me think that we should probably also have a `PhpdocRemoveExtraParamAnnotationFixer`, shouldn't we?

localheinz:
Maybe it would make more sense to address all of these issues with a single, configurable fixer?

```php
$config = [
    'add-missing',
    'remove-extra',
    'sort',
];
```

GrahamCampbell:
> Maybe it would make more sense to address all of these issues with a single, configurable fixer?

:-1: I think it's nice to separate the concerns, as it were. :P

localheinz:
@GrahamCampbell 

Definitely easier to test, hehe!

SpacePossum:
:+1: for config, otherwise we have 3+ fixers looking for the same candidates slowing down the tool.

localheinz:
Looks like adding an annotation for a missing parameter is the easy part, removing and sorting requires more guessing about where, for example, a comment regarding a parameter ends.

keradus:
what is the state here @localheinz ?

localheinz:
@keradus 

I think I will need some help with removing and sorting `@param` annotations, can you help me with that?

keradus:
sadly, not in January :(

localheinz:
@keradus 

No worries, I will try to give this a shot myself (as time permits), and then we can revisit!

keradus:
ping @localheinz 
any chances here ?

SpacePossum:
ping @localheinz please let us know if you've some time to work on this, no pressure :)

localheinz:
@SpacePossum 

Apologies, let me take a look again!

SpacePossum:
I'm closing this one in favor of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3909
but please all, hop over to that one and help out with comments providing insights, reviews, whatever :)
thanks for all the input here so far!

PR #2386, opened 2016-12-12
PHP7.1 Integration test - Add features added in PHP7.1.

SpacePossum:


GrahamCampbell:
What happens if we add a multicatch?

SpacePossum:
>What happens if we add a multicatch?

https://github.com/SpacePossum/PHP-CS-Fixer/blob/640e37c5938f8e94a238a8baddb5144f6803155a/tests/Fixtures/Integration/misc/PHP7_1.test#L21

it is already part of the test (note that currently the main function of the test is to ensure the fixer doesn't break down or breaks code, not to prove it fixes all 7.1 features)

GrahamCampbell:
great //0

GrahamCampbell:
great :)

keradus:
Thank you @SpacePossum.

PR #2394, opened 2016-12-13
Update .gitattributes

SpacePossum:
remove `benchmark.sh` and `box.json` from the export


keradus:
Thank you @SpacePossum.

PR #2395, opened 2016-12-14
NoEmptyCommentFixer - Fix typo

fritz-c:


SpacePossum:
Thanks for PR'ing this typo!
The README.rst is generated so the fix should be done within the fixer itself as well.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Comment/NoEmptyCommentFixer.php#L70
Let me know if you've the time to do this, otherwise I can pick it up later.

fritz-c:
Yeah, I can get that spot tomorrow (Japan time).

fritz-c:
Updated.

SpacePossum:
:+1: thanks :)

keradus:
Thank you @fritz-c.

PR #2398, opened 2016-12-14
Fix handling trailing linebreak added to close tag.

SpacePossum:
I would really like to have a transformer for both the open and close tag Tokens.
It would make the fixers way easier, it would also make the tokenizer more reusable outside the tool.
The performance hit is not that big btw. and we could optimize using the candidate appr. for the transformers as well I think.


SpacePossum:
invalid

keradus:
?

SpacePossum:
I thought
```
<?php
//
?>

```

would result in a linebreak echo'd, but it isn't, PHP is strange some times ;)

PR #2402, opened 2016-12-15
Token - Add __debugInfo()

SpacePossum:


src/Tokenizer/Token.php
```diff
@@ -67,6 +67,17 @@ public function __construct($token)
         }
     }
 
+    public function __debugInfo()
+    {
+        return array(
+            "\0+\0content" => $this->content,
```
keradus: why `"+"` ?

SpacePossum: this is the prefix PHP uses itself (for private attr.)

keradus: i played with it, i was able to put anything here - not only "+", but literally, any content.
why "+" ? is it standardized for privates (not for public btw?) ? any ref would be nice !

SpacePossum:
```php
<?php
class A
{
    private $content;
}

class B
{
    private $content;
    public function __debugInfo()
    {
        return array(
            "\0+\0content" => $this->content,
            "-content" => $this->content,
            "+content" => $this->content,
        );
    }
}

var_dump(new A(), new B());
die;
```

```bash
$ php /home/possum/work/symfony/test.php

/home/possum/work/symfony/test.php:20:
class A#1 (1) {
  private $content =>
  NULL
}
/home/possum/work/symfony/test.php:20:
class B#2 (3) {
  private $content =>
  NULL
  public $-content =>
  NULL
  public $+content =>
  NULL
}


```

(note the first `private` in the vardump of second object).

Something similar is done in the [caster](https://github.com/symfony/var-dumper/blob/master/Caster/Caster.php#L35) of SF VarDumper. I can't find any PHP docs as ref. as I think the output format is not formalized.

keradus: what php are you using ?
https://3v4l.org/nvjvI

here, the output is totally different, and between `\0` is the name of the class which declared a property

SpacePossum: hmm odd, did you try it in a console, it looks good in Ubuntu default console

keradus:
```
keradus@keradus:/tmp$ uname -a
Linux keradus 4.9.0-040900-generic #201612111631 SMP Sun Dec 11 21:33:00 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
keradus@keradus:/tmp$ php -v
PHP 7.1.0RC3 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.1.0-dev, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.1.0RC3, Copyright (c) 1999-2016, by Zend Technologies
keradus@keradus:/tmp$ php -f php.php 
object(A)#1 (1) {
  ["content":"A":private]=>
  NULL
}
object(B)#2 (3) {
  ["content":"+":private]=>
  NULL
  ["-content"]=>
  NULL
  ["+content"]=>
  NULL
}
```

SpacePossum: hmm weird, lets drop this PR for now, I feel like its not worth it to get my head around ;)

Issue #2336, opened 2016-11-29
ClassDefinitionFixer - Anonymous classes format by PSR12


keradus:
Sadly, #2328 needs to be reapplied on master branch.
I was unable to merge it in normal workflow from 1.13 to master.

@SpacePossum , please, could you create a PR against master one more time ?

PR #2328, opened 2016-11-25
ClassDefinitionFixer - Anonymous classes format by PSR12

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2302

PR #2403, opened 2016-12-15
ClassDefinitionFixer - Anonymous classes format by PSR12

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2336
master version of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2328

Fixes
- anonymous classes definitions
- edge case with `#` comment within the definition

Please review @gharlan (let me know if cases are missing)

### DIFF
--- a/tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
+++ b/tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
@@ -18,6 +18,8 @@ use PhpCsFixer\Tokenizer\Tokens;
 use PhpCsFixer\WhitespacesFixerConfig;
 
 /**
+ * @author SpacePossum
+ *
  * @internal
  */
 final class ClassDefinitionFixerTest extends AbstractFixerTestCase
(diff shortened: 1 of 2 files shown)


tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
```diff
@@ -41,16 +43,17 @@ public function testConfigureDefaultToNull()
     }
 
     /**
-     * @param string $expected PHP source code
-     * @param string $input    PHP source code
+     * @param string     $expected PHP source code
+     * @param string     $input    PHP source code
+     * @param null|mixed $config
```
keradus: `null|array` ?

SpacePossum: ah yes, PHPStorm guessed wrong ;)

GrahamCampbell:
Anonymous classes need to be in the configuration, otherwise, we'll have to remove this from the psr ruleset.

keradus:
they are in the spirit of psr-2 anonymous methods.
the thing is that when document was created, anonymous classes was not possible to made

GrahamCampbell:
Right, but the psr-2 spec officially doesn't enforce anything, only psr-12, right?

SpacePossum:
>Anonymous classes need to be in the configuration, otherwise, we'll have to remove this from the psr2 ruleset.

I vote to remove it from the PSR2 set than as imo there is no point in splitting up the fixer.

keradus:
almost nobody uses psr-12

SpacePossum:
I don't get it, the was done on 1.x and merge, than requested to do on master and now it is a problem?

keradus:
i don't have problem with it, I am waiting for @gharlan check as you asked him ;)

keradus:
Thank you @SpacePossum.

PR #2405, opened 2016-12-15
Transformers - Must be final

SpacePossum:


src/Tokenizer/Transformer/TypeColonTransformer.php
```diff
@@ -23,6 +23,8 @@
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  *
  * @internal
+ *
+ * @final
```
keradus: feel free to mark it as final on class-syntax, it's already internal

tests/Tokenizer/Transformer/TransformersTest.php
```diff
@@ -0,0 +1,72 @@
...
+ * @author SpacePossum
+ */
+final class TransformersTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @param \ReflectionClass $transformerRef
+     *
+     * @dataProvider provideReflectedTransformers
+     */
+    public function testTransformersAreFinal(\ReflectionClass $transformerRef)
```
keradus: please provide class name instead of `ReflectionClass`

example from other test what would be displayed with RC:

```
245) PhpCsFixer\Tests\ProjectCodeTest::testThatSrcClassesNotExposeProperties with data set #243 (ReflectionClass Object (...))
```

so, it's impossible to see which class looking at provided case

tests/Tokenizer/Transformer/TransformersTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+        return $transformersArray;
+    }
+
+    /**
+     * @param \ReflectionClass $item
+     *
+     * @return string[]
+     */
+    private function getPublicMethodNames(\ReflectionClass $item)
```
keradus: not needed anymore

tests/Tokenizer/Transformer/TransformersTest.php
SpacePossum: that is not true, there is one insert that can fail and displays the class name that is reflected, so all info is in the message

keradus: but it's inside msg of assertion, which may be there or not for new assertions in given test.
at the same time, test indicator should be meaningful, currently all of them are just showing `ReflectionClass`

tests/Tokenizer/Transformer/TransformersTest.php
```diff
@@ -0,0 +1,60 @@
...
+ * @author SpacePossum
+ */
+final class TransformersTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @param TransformerInterface $transformer
+     *
+     * @dataProvider provideTransformers
+     */
+    public function testTransformersAreFinal(TransformerInterface $transformer)
```
keradus: maybe sth like this:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/ProjectCodeTest.php#L179
but for src ?

then, i guess we will need few exceptions that will be marked as todo for 3.x line

SpacePossum: might work, I'll just wait until you're finished with that route to prevent needing to revert more out of this PR

SpacePossum: the route of adding something like this to the `ProjectCodeTest`

keradus:
@SpacePossum 
closing intentionally ?

SpacePossum:
>even if big ProjectCodeTest check the methods, it doesnt check the final

ah I missed out, I thought it would be one big awesome test :)

keradus:
That test does not exist yet, and actually a lot of classes are not final and making them final is BC breaker

keradus:
Thank you @SpacePossum.

PR #2413, opened 2016-12-19
Update Symfony ruleset

fabpot:
We've just upgraded PHP-CS-Fixer to 2.0 and we now have a lot of recommendations that we don't want to get on Symfony. The `heredoc` one being one of them.

Another rule we don't want, but I didn't find where this is enabled, is the conversion of one-line anonymous functions to multiline ones. Anyone knowns which rule we should disable to avoid that (http://fabbot.io/report/symfony/symfony/20907/a2c7e0634dc7052562b0d42bff5b141b638b7d18 has an example)? Thanks.


SpacePossum:
>[...] we don't want to get on Symfony. The heredoc one being one of them.

The was proposed to the SF community and accepted here:
https://github.com/symfony/symfony/pull/17086

so I'm not sure what we could've done more.

> is the conversion of one-line anonymous functions to multiline ones.

This is done by the `braces` fixer, which cannot be configured ATM.
The fixer is following [PSR-2](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#6-closures) so that is why it ended up in the SF set as default.

keradus:
@fabpot , your further input about `heredoc` is needed to proceed.
Rule was approved and apply on framework repo, so at one point it was accepted. Are you sure to remove it?

lambdas - #2421

fabpot:
I've just changed the configuration of Symfony to exclude it (symfony/symfony@d559e26).

I still think it was a mistake to accept such a change in Symfony.

keradus:
Then please apply both changes here and regenerate readme, as @SpacePossum asked.

fabpot:
Done

SpacePossum:
:+1:

keradus:
Thank you @fabpot.

PR #2414, opened 2016-12-19
Add PhpdocReturnSelfReferenceFixer

SpacePossum:


keradus:
but... those are different
`this` - return exactly the same instance
`self` - return new instance of exactly same type
`static` - return new instance of same type or any subtype

SpacePossum:
Well, if all your fellow dev's know it and did it right you might not need the fixer to replace that ;)
Anyway `$this` != `this`,  `$self` != `self`, so there all kind of cases you could fix

keradus:
for handling `$` and `@` i'm :+1:
but for changing self - static, I'm :-1:

keradus:
Thank you @SpacePossum.

src/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixer.php
```diff
@@ -0,0 +1,196 @@
...
+        return new FixerDefinition(
+            'The type of `@return` annotations of methods returning a reference to itself must the configured one.',
+            array(new CodeSample('
+<?php
+class Sample
+{
+    /**
+     * @return this
+     */
+    public function test()
```
staabm: shouldn't this fixer also turn

```
class Sample
{
    /**
     * @return Sample
     */
    public function test()
    {
        return $this;
    }
}'
```

into

```
class Sample
{
    /**
     * @return this
     */
    public function test()
    {
        return $this;
    }
}'
```

??

PR #2416, opened 2016-12-20
Allow config name on static construct.

SpacePossum:


GrahamCampbell:
Could be argued to be a breaking change, since it's breaking extension of the `create` method, since the `Config` class is neither internal, nor final, and that method is not private.

keradus:
Indeed, if one extends `Config` this will break his code.
:-1:

also, one could already done that by `(new Config('name'))->set...`, just with newer php version

SpacePossum:
breaking how?

GrahamCampbell:
> breaking how?

Current code: https://3v4l.org/GMXeP. New code: https://3v4l.org/gjuRR.

PHP 7 error:

```
Warning: Declaration of Bar::baz() should be compatible with Foo::baz($name = 'default') in /in/gjuRR on line 17
```

PHP 5 error:

```
Strict Standards: Declaration of Bar::baz() should be compatible with Foo::baz($name = 'default') in /in/gjuRR on line 17
```

HHVM error:

```
Uncaught exception: std::bad_alloc
```

SpacePossum:
yes, my point is I target master here, which is for BC breaks (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/CONTRIBUTING.md 'Master if you make changes to the code that are not backward compatible.'). So when this is merged it will only effect people who use master which is not a concern IMO.

GrahamCampbell:
That guide is out of date, because master is 2.0.x right now, now 3.0.x.

SpacePossum:
>That guide is out of date, because master is 2.0.x right now, now 3.0.x.

so we can not do any BC breaks till when? this is stalling process for no reason

GrahamCampbell:
Well, I'm not saying that, I'm just saying this can't be merged as is, if master is 2.0.x, and I imagine 3.x is ages away.

SpacePossum:
If a new 2.x branch is made from master this can be merged with no side effects.
SO why would it this or 3.x be ages away? We have no LTS stuff going on.

GrahamCampbell:
> So why would this or 3.x be ages away? We have no LTS stuff going on.

Just a feeling I have. :trollface: 

> If a new 2.x branch is made from master this can be merged with no side effects.

True, true. @keradus

keradus:
having 3.x just for this is not a good reason,
especially when same thing could be achieved by using newer php syntax...

PR #2428, opened 2016-12-27
SCA with Php Inspections (EA Extended)

kalessil:


src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
```diff
@@ -70,15 +70,6 @@ public function getDefinition()
     /**
      * {@inheritdoc}
      */
-    public function getPriority()
```
SpacePossum: note sure about this one, the comment is nice

src/Fixer/Alias/PowToExponentiationFixer.php
```diff
@@ -100,14 +100,6 @@ public function getPriority()
     }
 
     /**
-     * {@inheritdoc}
-     */
-    public function isRisky()
```
SpacePossum: ncie catch (together with the one in `MbStrFunctionsFixer`)

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
kalessil: Nice but seems not relevant anymore (method same as parents').

src/Fixer/Alias/PowToExponentiationFixer.php
kalessil: My plugin forced me to do this %)

src/Cache/Cache.php
```diff
@@ -98,7 +98,7 @@ public static function fromJson($json)
             ));
         }
 
-        $requiredKeys = array(
+        static $requiredKeys = array(
```
keradus: please provide reasoning.
it has value for older php, but currently if an array is not modified, it's declared only once in memory

src/DocBlock/Annotation.php
```diff
@@ -157,7 +157,7 @@ public function getTypes()
      */
     public function setTypes(array $types)
     {
-        $pattern = '/'.preg_quote($this->getTypesContent()).'/';
+        $pattern = '/'.preg_quote($this->getTypesContent(), '/').'/';
```
keradus: reasoning, like new unit test, is missing.
please provide any case when value we quoting here actually contains `/`.

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
keradus: yep, it was added because of comment.
if removed, the comment needs to be added in every mentioned fixer.

src/RuleSet.php
```diff
@@ -21,7 +21,7 @@
  */
 final class RuleSet implements RuleSetInterface
 {
-    private $setDefinitions = array(
+    private static $setDefinitions = array(
```
keradus: :-1: for staticness.
in general, i hate static things.

i agree that sometimes it's nice. but in all "nice" cases - we will basically use `const` instead of `static` with never PHP ;)

here - please revert. why? `setDefinitions` is not supposed to be constant, it is supposed to be extended by users in future (same like we register custom fixers, we want to register custom rules)

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
kalessil: Then probably relocating the coment into \PhpCsFixer\AbstractFixer::getPriority will work (knowledge is preserved, no duplicated code)?

SpacePossum: the comment is specific for the fixer extending the abstract, so that would not be possible (or the same)

src/DocBlock/Annotation.php
kalessil: The reasoning is the fact that preg_quote is not escaping '/' by default and here used exactly this separator. I didn't check deeper than at this level.

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
kalessil: Hm, abstract fixers' method returns the same priority (0). So by default fixers should behave as described in the comment. Or?

src/Cache/Cache.php
kalessil: For my understanding: are you referencing to packed array optimizations in PHP7+?

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
SpacePossum: Yes, they should behave as the PHPDoc describe :)
The comment I ref. to is the one inside the method (line 75)
```
 // should be run after NoEmptyPhpdocFixer, [...] and NoUselessElseFixer.
```

that info is only relevant to this fixer

kalessil: I see, then let's revert this.

src/Cache/Cache.php
keradus: no, non-editable arrays in 5.6, maybe 5.5 as well.
request for reasoning stays

src/DocBlock/Annotation.php
keradus: you need to escape that `/` only when `getTypeContent` will return sth that may contain `/`. it it does not, not need for time wasting looking for `/` to escape it.
is it a case here? if so, provide utest for that

kalessil: I checked the logic behind, you are right here - under the hood regex filters alfa-numeric.

For the sake of clear code, I'd suggest keeping this. Please confirm if you'd prefer reverting preg_quote changes.

kalessil:
@keradus : static local variables and newly added preg_quote 2nd parameters are reverted now.

Would you please check if the rest part of the PR looks good?

src/DocBlock/Annotation.php
keradus: I'm slightly -0 here, no big deal.
just wanted to have solid reasoning behind that change (eg bugfix of some edge case we were not aware here, but as you checked - it's not a case here, thanks)

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -47,8 +47,9 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
 
         $usesOrder = array();
         foreach ($namespacesImports as $uses) {
-            $usesOrder = array_replace($usesOrder, $this->getNewOrder(array_reverse($uses), $tokens));
+            $usesOrder[] = $this->getNewOrder(array_reverse($uses), $tokens);
         }
+        $usesOrder = call_user_func_array('array_replace', $usesOrder);
```
keradus: yeah, it's faster indeed, i just hate calling function by strings... harder to refactor and analyse the code... (like is the function called anywhere)

keradus:
Thank you @kalessil.

src/Fixer/Import/OrderedImportsFixer.php
kalessil: Hm, I need to check if PSs' find usages will work with array_replace::class (if this valid at all), but your point is very good one here - let see if there is a better alternative

PR #2429, opened 2016-12-27
FixCommand: allow ConfigInterface injection

Slamdunk:
Hi, version 1 allowed [custom building](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v1.13.1/Symfony/CS/Console/Command/FixCommand.php#L85) of FixCommand, functionality lost in version 2.

In our company we have dozens of application using this (awesome) project, and we make available a set of custom fixers for all this project through a custom `php-cs-fixer` bin trivially customized.

If the command cannot be touched, all the customization have to be done in the `.php_cs` file, a not maintainable way for dozens of apps.

This PR retain current behaviour, allowing for custom config injection during instantiation, and allowing to serve help also for custom fixers.

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -73,14 +73,14 @@ final class FixCommand extends Command
      */
     private $defaultConfig;
 
-    public function __construct()
+    public function __construct(ConfigInterface $config = null)
     {
-        parent::__construct();
-
-        $this->defaultConfig = new Config();
+        $this->defaultConfig = $config ?: new Config();
         $this->errorsManager = new ErrorsManager();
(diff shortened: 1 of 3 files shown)


keradus:
why you company needs to inject it, instead of creating `MyCompanyConfig` that will be returned by `.php_cs` file ?

localheinz:
@slamdunk

As @keradus suggested, this is a great and easy way, and this is how we have been doing it for a while - for reference, see https://github.com/refinery29/php-cs-fixer-config.

Slamdunk:
Ok for the Config.

For the help message, it's a bit counter-intuitive to use fixers that are not listed.
I hope `.php_cs` Config could be use to generate help message too.

Issue #2254, opened 2016-11-02
alignment of dollar-sign with variadic parameters


havvg:
Hi there, is there a way to have the dollar-sign be aligned with variadic parameters?

Currently, the fixer yields this patch:

```diff
@@ -89,7 +89,7 @@ final class PortalVariantVoter implements VoterInterface
     /**
      * Disables variants on a specific portal.
      *
-     * @param string    $portal
+     * @param string $portal
      * @param string[] ...$articleNumber
      *
      * @return PortalVariantVoter
Stage this hunk [y,n,q,a,d,/,K,g,e,?]?
```

However I would like to have this result:

```diff
@@ -89,7 +89,7 @@ final class PortalVariantVoter implements VoterInterface
     /**
      * Disables variants on a specific portal.
      *
-     * @param string    $portal
+     * @param string      $portal
      * @param string[] ...$articleNumber
      *
      * @return PortalVariantVoter
Stage this hunk [y,n,q,a,d,/,K,g,e,?]?
```

keradus:
Nobody brings that for our PHPDoc fixer yet. I like the idea. Could you send a PR proposal ?

PR #2430, opened 2016-12-28
PhpdocAlignFixer - Fix alignment of variadic params.

SpacePossum:
Contrary to https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2254 I think the `...` are part of the variable, like `&` is, and therefor the alignment should not be done on the `$` but on `...`.
Making the fixer work case insensitive makes it atomic (i.e. no longer depended on `phpdoc_types`).
On a side note [PSR5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#713-param) is not conclusive on this topic.

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
@@ -512,4 +512,78 @@ EOF;
 
         $this->doTest($expected, $input);
     }
+
+    /**
+     * @param string $expected
+     * @param string $input
+     *
+     * @requires PHP 5.6
+     * @dataProvider provideVariadicCases
+     */
+    public function testVariadicParams($expected, $input)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideVariadicCases()
+    {
+        return array(
+            array(
+                '<?php
+final class Sample
+{
+    /**
+     * @param int[] $a    A
+     * @param int   &$b   B
+     * @param array ...$c C
+     */
+    public function sample2($a, &$b, ...$c)
+    {
+    }
+}
+',
+            '<?php
+final class Sample
+{
+    /**
+     * @param int[]       $a  A
+     * @param int          &$b B
+     * @param array ...$c    C
+     */
+    public function sample2($a, &$b, ...$c)
+    {
+    }
+}
+',
+            ),
+                        array(
+                '<?php
+final class Sample
+{
+    /**
+     * @param int     $a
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Annotations are case sensitive. You should first run the other fixer I wrote which normalizes annotation names.

SpacePossum:
>Annotations are case sensitive. You should first run the other fixer I wrote which normalizes annotation names.

You mean [phpdoc_types](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2430#issue-197805359)? Well I didn't and the fixer still works so all good :)

GrahamCampbell:
> Well I didn't and the fixer still works so all good :)

What I'm saying is that it's incorrect to treat `@return` the same as `@Return`, because they are not the same.

SpacePossum:
>What I'm saying is that it's incorrect to treat `@return` the same as `@Return`, because they are not the same.

So the `phpdoc_types` is either wrong or risky?
Or maybe it doesn't matter at all and this fixer won't break anything just like `phpdoc_types`. 
If you want me to change anything here please provide an (common/widely) accepted standard showing this is wrong or a test proving an issue.
These discussions are typically leading nowhere and take a lot of time that can used better.

GrahamCampbell:
> So the phpdoc_types is either wrong or risky?

Yeh, probably should be marked as risky.

tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
```diff
@@ -512,4 +512,78 @@ public function testFixUnicode()
...
+    {
+        return array(
+            array(
+                '<?php
+final class Sample
+{
+    /**
+     * @param int[] $a    A
+     * @param int   &$b   B
+     * @param array ...$c C
```
keradus: this is not valid. name of argument is `$c`, not `...$c`

keradus:
please revert annotation casing

tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
SpacePossum: >btw, this is not valid. name of argument is $c, not ...$c

where do the `&` and `...` go if not before the name?

keradus: maybe it should not be there at all, but it's not cs thing

keradus:
Thank you @SpacePossum.

PR #1187, opened 2015-05-11
[WIP] Feature new fixer oneline method argument

rokerkony:
Closes #1184

PR #2435, opened 2016-12-30
Added one line method argument fixer

greg0ire:
This is a carry of #1187 

# TODO

- [ ] improve description ?
- [ ] fix new test cases

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -424,6 +424,10 @@ Choose from the list of available rules:
 * **object_operator_without_whitespace** [@Symfony]
    | There should not be space before or after object T_OBJECT_OPERATOR.
 
+* **one_line_method_argument**
+   | Make sure method/function arguments are on a single line, or on each
+   | their own.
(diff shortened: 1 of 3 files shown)


SpacePossum:
Hi again :)
My suggestion for using `method_argument_space` with configuration is because that fixer already deals with spacing within method calls. I see you points [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1187#issuecomment-269794435), but I'm affraid have a new fixer might be conflicting with the current. Maybe worth it to check out `method_argument_space` to see if there is overlap in functionality and move from there (i.e. maybe remove some from the current and add a new fixer, or add more to the current fixer by configuration option).

greg0ire:
Might be best indeed. I'm going to stop there today, but I'll have a look at `method_argument_space` later.

greg0ire:
See #2440

PR #2438, opened 2016-12-31
Mysql to Mysqli fixer

theUniC:
Hi,

I've created a fixer to translate ```mysql_*``` style functions to ```mysqli_*``` style functions. Since not all mysql functions have a direct mysqli translation there're a few unsupported functions that don't get touched:

* `mysql_create_db`
* `mysql_db_name`
* `mysql_db_query`
* `mysql_drop_db`
* `mysql_escape_string`
* `mysql_field_flags`
* `mysql_field_len`
* `mysql_field_name`
* `mysql_field_table`
* `mysql_field_type`
* `mysql_list_dbs`
* `mysql_list_fields`
* `mysql_list_processes`
* `mysql_list_tables`
* `mysql_pconnect`
* `mysql_query`
* `mysql_real_escape_string`
* `mysql_result`
* `mysql_select_db`
* `mysql_tablename`
* `mysql_unbuffered_query`

Hope you find it useful!

Best,
Christian.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -525,8 +525,17 @@ Choose from the list of available fixers:
    Multi-line whitespace before closing semicolon are prohibited.
 
 * **mysql_to_mysqli** [contrib]
-   Replace deprecated `mysql_*` functions with mysqli_* equivalents.
-   Warning: This could change code behaviour.
+   Replace deprecated `mysql_*` functions with mysqli_* equivalents. Not
+   all functions are supported since there are a few functions that don't
+   have a direct mysqli translation. The unsupported ones are:
+   `mysql_create_db`, `mysql_db_name`, `mysql_db_query`, `mysql_drop_db`,
+   `mysql_escape_string`, `mysql_field_flags`, `mysql_field_len`,
+   `mysql_field_name`, `mysql_field_table`, `mysql_field_type`,
+   `mysql_list_dbs`, `mysql_list_fields`, `mysql_list_processes`,
+   `mysql_list_tables`, `mysql_pconnect`, `mysql_query`,
+   `mysql_real_escape_string`, `mysql_result`, `mysql_select_db`,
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
This fixer already exists in the form of the alias function fixer.

GrahamCampbell:
On 2.x.

theUniC:
Oh! Didn't know it! Also could not find it on the master branch. Can you point me there?

GrahamCampbell:
It's enabled by you passing in configuration to the fixer.

GrahamCampbell:
The fixer is called `no_alias_functions`.

theUniC:
I already noticed the presence of this fixer but I'm afraid it's not the same. ```mysql_*``` functions are not aliases of ```mysqli_*``` functions. mysql functions can only be replaced by mysqli functions when number of arguments match.

GrahamCampbell:
Good point. I'm not sure if 1.x is still accepting new fixers though. Ping @keradus

keradus:
no, 1.x is features frozen

theUniC:
OK! So would be better to PR to master branch?

SpacePossum:
2.x/master would be better :)

>I already noticed the presence of this fixer but I'm afraid it's not the same. mysql_* functions are not aliases of mysqli_* functions. mysql functions can only be replaced by mysqli functions when number of arguments match.

Actually this is also true for the alias functions, i.e. if the number of arguments do not match the expected it is not an alias and should not be fixed, but this is another issue ;)

You can take a look at the [MbStrFunctionsFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.0.0/src/Fixer/Alias/MbStrFunctionsFixer.php#L270 as it also checks the number of arguments before replacing one method call with another.

keradus:
Closing as this PR is staled.
Feel free to open new one for 2.x

PR #2431, opened 2016-12-29
MethodArgumentSpaceFixer: allow to retain multiple spaces after comma

Slamdunk:
PR https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/998 changed `MethodArgumentSpaceFixer` with a BC break. We have tons of code like this one:
```php
// Symfony command specifications
$this
    ->setName('company:command')
    ->setDescription('Our command')
    ->addOption('xxxxxyyy',             null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxxyyyyyy',          null, InputOption::VALUE_REQUIRED,  'desc [...]')
    ->addOption('xxxxxyyyy',            null, InputOption::VALUE_REQUIRED,  'desc [...]')
    ->addOption('xxxxxauto',            null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxx',                null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxxi',               null, InputOption::VALUE_REQUIRED,  'desc [...]')
    ->addOption('xxxxxyyyyyyyyyyy',     null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxxyyyyyyyyyy',      null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxxyyyyyyyyyy',      null, InputOption::VALUE_NONE,      'desc [...]')
    ->addOption('xxxxxyyyyyyyyyyyyy',   null, InputOption::VALUE_NONE,      'desc [...]')
;

// Values objects
$columns = new Excel\Collection(array(
    new Excel\Column('xxx',                        'DESC 1111',             5, new Excel\Style\Text()),
    new Excel\Column('xxxx',                       'DESC',                 15, new Excel\Style\Text()),
    new Excel\Column('xxx',                        'DES',                  15, new Excel\Style\Text()),
    new Excel\Column('xxxxxxxxxxxxxxxxxxxxxx',     'DESC 2222222',         20, new Excel\Style\CC()),
    new Excel\Column('xxxxxxxxxxxxxxxxxxxxxxx',    'DESC 333333333333',    40, new Excel\Style\Text()),
    new Excel\Column('xxxxxxxxxxx',                'DESC 444444',          15, new Excel\Style\Text()),
    new Excel\Column('xxxxxxxxxxxxxxxxxxxx',       'DESC 5555555555555',   10, new Excel\Style\Text()),
[...]

Issue #770, opened 2014-11-25
Wrapped array elements indentation not fixed.


jian-wu:
Hi, 

I came across an issue where elements in array that are wrapped with double indentation, which isn't fixed by the fixer.

```
<?php
// expected
$var = array(
    'a' => 1,
    'b' => 2,
);
// actual
$var = array(
        'a' => 1,
        'b' => 2,
);
```

I've tried adding this case to the indentation fixer but it does not fail on unit test.

Can someone please verify or let me know it's not an issue?

Thanks.

harikt:
I have something similar.

The code was having tabs and when I ran to convert to psr-2 it changed to spaces . But there is 8 spaces. See

```php
return $this->render(
            'AcmeShopBundle:Random:index.html.twig',
array(
'number' => $number
                )
        );
```

the above code is very ugly and it is unable to make it look nice :( .

keradus:
Closing due to amazing work of #3135 by @julienfalque

PR #2440, opened 2016-12-31
MethodArgumentSpaceFixer - add ensure_fully_multiline option

greg0ire:
Todo : 

- [x] fix broken tests
- [x] add more test cases
    - [x] for indented methods (currently indention is hardcoded and not preserved)
    - [x] for multiline arguments
        - [x] for multiline array arguments
        - [x] for multiline function arguments
    - [x] for fix needed near opening parenthesis
    - [x] for fix needed near closing parenthesis
    - [x] for different OS line endings
    - [x] for different indention chars
    - [x] for different indention levels
    - [x] for nested arrays
- [x] make the fixer optional

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -541,10 +541,15 @@ Choose from the list of available rules:
 * **method_argument_space** [@PSR2, @Symfony]
 
   In method arguments and method call, there MUST NOT be a space before
-  each comma and there MUST be one space after each comma.
+  each comma and there MUST be one space after each comma. Argument lists
+  MAY be split across multiple lines, where each subsequent line is
+  indented once. When doing so, the first item in the list MUST be on the
+  next line, and there MUST be only one argument per line.
 
   Configuration options:
 
+  - ``ensure_fully_multiline`` (``bool``): ensure every argument of a multiline
+    argument list is on its own line; defaults to ``false``
   - ``keep_multiple_spaces_after_comma`` (``bool``): whether keep multiple spaces
     after comma; defaults to ``false``
 
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -56,12 +56,10 @@ final class Cache implements CacheInterface
     public function set($file, $hash)
     {
         if (!is_int($hash)) {
-            throw new \InvalidArgumentException(
-                sprintf(
+            throw new \InvalidArgumentException(sprintf(
                 'Value needs to be an integer, got "%s".',
                 is_object($hash) ? get_class($hash) : gettype($hash)
-            )
-            );
+            ));
         }
 
         $this->hashes[$file] = $hash;
--- a/src/Cache/Cache.php
+++ b/src/Cache/Cache.php
@@ -56,9 +56,11 @@ final class Cache implements CacheInterface
     public function set($file, $hash)
     {
         if (!is_int($hash)) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
                 'Value needs to be an integer, got "%s".',
-                is_object($hash) ? get_class($hash) : gettype($hash))
+                is_object($hash) ? get_class($hash) : gettype($hash)
+            )
             );
         }
 
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -106,7 +106,9 @@ final class DescribeCommand extends Command
 
             throw new \InvalidArgumentException(sprintf(
                 '%s %s not found.%s',
-                ucfirst($e->getType()), $name, null === $alternative ? '' : ' Did you mean "'.$alternative.'"?'
+                ucfirst($e->getType()),
+                $name,
+                null === $alternative ? '' : ' Did you mean "'.$alternative.'"?'
             ));
         }
     }
--- a/src/Console/Command/SelfUpdateCommand.php
+++ b/src/Console/Command/SelfUpdateCommand.php
@@ -43,7 +43,8 @@ final class SelfUpdateCommand extends Command
                 ]
             )
             ->setDescription('Update php-cs-fixer.phar to the latest stable version.')
-            ->setHelp(<<<'EOT'
+            ->setHelp(
+                <<<'EOT'
 The <info>%command.name%</info> command replace your php-cs-fixer.phar by the
 latest version released on:
 <comment>https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases</comment>
--- a/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
+++ b/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
@@ -37,7 +37,8 @@ final class ModernizeTypesCastingFixer extends AbstractFunctionReferenceFixer
     $a = doubleval($b);
     $a = strval ($b);
     $a = boolval($b);
-'),
+'
+            ),
             ],
             null,
             'Risky if any of the functions `intval`, `floatval`, `doubleval`, `strval` or `boolval` are overridden.'
--- a/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
+++ b/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
@@ -38,8 +38,7 @@ final class ModernizeTypesCastingFixer extends AbstractFunctionReferenceFixer
     $a = strval ($b);
     $a = boolval($b);
 '
-            ),
-            ],
+            )],
             null,
             'Risky if any of the functions `intval`, `floatval`, `doubleval`, `strval` or `boolval` are overridden.'
         );
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -323,6 +323,6 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
      */
     private function isNewLine(Token $token)
     {
-        return $token->isWhitespace() && (false !== strpos($token->getContent(), "\n"));
+        return $token->isWhitespace() && false !== strpos($token->getContent(), "\n");
     }
 }
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -248,7 +248,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
     }
 
     /**
-     * Makes sure there is a whitespace at the given location
+     * Makes sure there is a whitespace at the given location.
      *
      * @param Tokens $tokens      The token stream to modify
      * @param int    $index       where to insert the whitespace
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -103,7 +103,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
             if ($token->equals('(')) {
                 $meaningfulTokenBeforeParenthesis = $tokens[$tokens->getPrevMeaningfulToken($index)];
                 if (!$meaningfulTokenBeforeParenthesis->isKeyword()
-                    || $meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
+                    || $meaningfulTokenBeforeParenthesis->isGivenKind([T_LIST, T_FUNCTION])) {
                     if ($this->fixFunction($tokens, $index) && $this->configuration['ensure_fully_multiline']) {
                         if (!$meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
                             $this->ensureFunctionFullyMultiline($tokens, $index);
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -113,14 +113,14 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('keep_multiple_spaces_after_comma', 'Whether keep multiple spaces after comma.'))
                 ->setAllowedTypes(['bool'])
-                ->setDefault(false) // @TODO should be true at 3.0
+                ->setDefault(false)
                 ->getOption(),
             (new FixerOptionBuilder(
                 'ensure_fully_multiline',
                 'Ensure every argument of a multiline argument list is on its own line'
             ))
                 ->setAllowedTypes(['bool'])
-                ->setDefault(false)
+                ->setDefault(false) // @TODO should be true at 3.0
                 ->getOption(),
         ]);
     }
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -67,14 +67,14 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
                     ['ensure_fully_multiline' => true]
                 ),
                 new CodeSample(
-                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample(1,  2);",
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample('foo',    'foobarbaz', 'baz');\nsample('foobar', 'bar',       'baz');",
                     [
                         'ensure_fully_multiline' => true,
                         'keep_multiple_spaces_after_comma' => true,
                     ]
                 ),
                 new CodeSample(
-                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample(1,  2);",
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample('foo',    'foobarbaz', 'baz');\nsample('foobar', 'bar',       'baz');",
                     [
                         'ensure_fully_multiline' => true,
                         'keep_multiple_spaces_after_comma' => false,
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -186,7 +186,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         $indentation = $existingIndentation.$this->whitespacesConfig->getIndent();
         $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
         if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
-            $tokens->addNewlineAndIndent(
+            $this->addNewlineAndIndent(
                 $tokens,
                 $endFunctionIndex,
                 $existingIndentation,
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -14,6 +14,7 @@ namespace PhpCsFixer\Fixer\FunctionNotation;
 
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
@@ -27,7 +28,7 @@ use PhpCsFixer\Tokenizer\Tokens;
  *
  * @author Kuanhung Chen <ericj.tw@gmail.com>
  */
-final class MethodArgumentSpaceFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+final class MethodArgumentSpaceFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
     /**
      * Method to insert space after comma and remove space before comma.
@@ -47,7 +48,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
     public function getDefinition()
     {
         return new FixerDefinition(
-            'In method arguments and method call, there MUST NOT be a space before each comma and there MUST be one space after each comma.',
+            'In method arguments and method call, there MUST NOT be a space before each comma and there MUST be one space after each comma. Argument lists MAY be split across multiple lines, where each subsequent line is indented once. When doing so, the first item in the list MUST be on the next line, and there MUST be only one argument per line.',
             [
                 new CodeSample(
                     "<?php\nfunction sample(\$a=10,\$b=20,\$c=30) {}\nsample(1,  2);",
@@ -60,6 +61,17 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
                 new CodeSample(
                     "<?php\nfunction sample(\$a=10,\$b=20,\$c=30) {}\nsample(1,  2);",
                     ['keep_multiple_spaces_after_comma' => true]
+                ),
+                new CodeSample(
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,\n    2);",
+                    ['ensure_fully_multiline' => true]
+                ),
+                new CodeSample(
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);",
+                    [
+                        'ensure_fully_multiline' => true,
+                        'keep_multiple_spaces_after_comma' => true,
+                    ]
                 ),
             ]
         );
@@ -81,8 +93,14 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         for ($index = $tokens->count() - 1; $index > 0; --$index) {
             $token = $tokens[$index];
 
-            if ($token->equals('(') && !$tokens[$index - 1]->isGivenKind(T_ARRAY)) {
-                $this->fixFunction($tokens, $index);
+            if ($token->equals('(')) {
+                $meaningfulTokenBeforeParenthesis = $tokens[$tokens->getPrevMeaningfulToken($index)];
+                if (!$meaningfulTokenBeforeParenthesis->isKeyword()
+                    || $meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
+                    if ($this->fixFunction($tokens, $index) && $this->configuration['ensure_fully_multiline']) {
+                        $this->ensureFunctionFullyMultiline($tokens, $index);
+                    }
+                }
             }
         }
     }
@@ -95,6 +113,13 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('keep_multiple_spaces_after_comma', 'Whether keep multiple spaces after comma.'))
                 ->setAllowedTypes(['bool'])
+                ->setDefault(false) // @TODO should be true at 3.0
+                ->getOption(),
+            (new FixerOptionBuilder(
+                'ensure_fully_multiline',
+                'Ensure every argument of a multiline argument list is on its own line'
+            ))
+                ->setAllowedTypes(['bool'])
                 ->setDefault(false)
                 ->getOption(),
         ]);
@@ -105,10 +130,14 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
      *
      * @param Tokens $tokens             Tokens to handle
      * @param int    $startFunctionIndex Start parenthesis position
+     *
+     * @return bool whether the function is multiline
      */
     private function fixFunction(Tokens $tokens, $startFunctionIndex)
     {
         $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
+        $isMultiline = $this->isNewline($tokens[$startFunctionIndex + 1])
+            || $this->isNewline($tokens[$endFunctionIndex - 1]);
 
         for ($index = $endFunctionIndex - 1; $index > $startFunctionIndex; --$index) {
             $token = $tokens[$index];
@@ -127,8 +156,101 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
 
             if ($token->equals(',')) {
                 $this->fixSpace2($tokens, $index);
+                if (!$isMultiline && $this->isNewline($tokens[$index + 1])) {
+                    $isMultiline = true;
+                    break;
+                }
             }
         }
+
+        return $isMultiline;
+    }
+
+    private function ensureFunctionFullyMultiline(Tokens $tokens, $startFunctionIndex)
+    {
+        // find out what the indentation is
+        $searchIndex = $startFunctionIndex;
+        do {
+            $prevWhitespaceTokenIndex = $tokens->getPrevTokenOfKind(
+                $searchIndex,
+                [[T_WHITESPACE]]
+            );
+            $searchIndex = $prevWhitespaceTokenIndex;
+        } while ($prevWhitespaceTokenIndex
+            && false === strpos($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n")
+        );
+        $existingIndentation = $prevWhitespaceTokenIndex
+            ? ltrim($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n\r")
+            : '';
+
+        $indentation = $existingIndentation.$this->whitespacesConfig->getIndent();
+        $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
+        if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
+            $tokens->insertAt(
+                $endFunctionIndex,
+                new Token([
+                    T_WHITESPACE,
+                    $this->whitespacesConfig->getLineEnding().$existingIndentation,
+                ])
+            );
+            ++$endFunctionIndex;
+        }
+
+        for ($index = $endFunctionIndex - 1; $index > $startFunctionIndex; --$index) {
+            $token = $tokens[$index];
+
+            // skip nested method calls and arrays
+            if ($token->equals(')')) {
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $index, false);
+                continue;
+            }
+
+            // skip nested arrays
+            if ($token->isGivenKind(CT::T_ARRAY_SQUARE_BRACE_CLOSE)) {
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_ARRAY_SQUARE_BRACE, $index, false);
+                continue;
+            }
+
+            if ($token->equals(',')) {
+                $this->fixNewline($tokens, $index, $indentation);
+            }
+        }
+        $this->fixNewLine($tokens, $startFunctionIndex, $indentation, false);
+    }
+
+    /**
+     * Method to insert newline after comma or opening parenthesis.
+     *
+     * @param Tokens $tokens
+     * @param int    $index       index of a comma
+     * @param string $indentation the indentation that should be used
+     * @param bool   $override    whether to override the existing character or not
+     */
+    private function fixNewline(Tokens $tokens, $index, $indentation, $override = true)
+    {
+        if ($this->isNewline($tokens[$index + 1]) || $tokens[$index + 1]->isComment()) {
+            return;
+        }
+        if ($tokens[$index + 2]->isComment()) {
+            $nextMeaningfulTokenIndex = $tokens->getNextMeaningfulToken($index + 2);
+            if (!$this->isNewLine($tokens[$nextMeaningfulTokenIndex - 1])) {
+                $tokens->insertAt(
+                    $nextMeaningfulTokenIndex,
+                    new Token([
+                        T_WHITESPACE,
+                        $this->whitespacesConfig->getLineEnding().$indentation,
+                    ])
+                );
+            }
+
+            return;
+        }
+        $method = $override ? 'overrideAt' : 'insertAt';
+
+        $tokens->$method($index + 1, new Token([
+            T_WHITESPACE,
+            $this->whitespacesConfig->getLineEnding().$indentation,
+        ]));
     }
 
     /**
@@ -190,5 +312,17 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         $content = $tokens[$index + 1]->getContent();
 
         return $content !== ltrim($content, "\r\n");
+    }
+
+    /**
+     * Checks if token is new line.
+     *
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNewLine(Token $token)
+    {
+        return $token->isWhitespace() && (false !== strpos($token->getContent(), "\n"));
     }
 }
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -186,12 +186,11 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         $indentation = $existingIndentation.$this->whitespacesConfig->getIndent();
         $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
         if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
-            $tokens->insertAt(
+            $this->ensureWhitespaceAtIndex(
+                $tokens,
                 $endFunctionIndex,
-                new Token([
-                    T_WHITESPACE,
-                    $this->whitespacesConfig->getLineEnding().$existingIndentation,
-                ])
+                $existingIndentation,
+                false
             );
             ++$endFunctionIndex;
         }
@@ -234,26 +233,42 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         if ($tokens[$index + 2]->isComment()) {
             $nextMeaningfulTokenIndex = $tokens->getNextMeaningfulToken($index + 2);
             if (!$this->isNewLine($tokens[$nextMeaningfulTokenIndex - 1])) {
-                $tokens->insertAt(
+                $this->ensureWhitespaceAtIndex(
+                    $tokens,
                     $nextMeaningfulTokenIndex,
-                    new Token([
-                        T_WHITESPACE,
-                        $this->whitespacesConfig->getLineEnding().$indentation,
-                    ])
+                    $indentation,
+                    false
                 );
             }
 
             return;
         }
 
-        call_user_func(
-            [$tokens, $override ? 'overrideAt' : 'insertAt'],
-            $index + 1,
-            new Token([
-                T_WHITESPACE,
-                $this->whitespacesConfig->getLineEnding().$indentation,
-            ])
-        );
+        $this->ensureWhitespaceAtIndex($tokens, $index + 1, $indentation, $override);
+    }
+
+    /**
+     * Makes sure there is a whitespace at the given location
+     *
+     * @param Tokens $tokens      The token stream to modify
+     * @param int    $index       where to insert the whitespace
+     * @param string $indentation the indentation that should be used
+     * @param bool   $override    whether to override the existing character or not
+     */
+    private function ensureWhitespaceAtIndex(Tokens $tokens, $index, $indentation, $override)
+    {
+        $whitespaceToken = new Token([
+            T_WHITESPACE,
+            $this->whitespacesConfig->getLineEnding().$indentation,
+        ]);
+
+        if ($override) {
+            $tokens->overrideAt($index, $whitespaceToken);
+
+            return;
+        }
+
+        $tokens->insertAt($index, $whitespaceToken);
     }
 
     /**
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -270,7 +270,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         ]);
 
         if ($override) {
-            $tokens->overrideAt($index, $whitespaceToken);
+            $tokens[$index] = $whitespaceToken;
 
             return;
         }
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -186,7 +186,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         $indentation = $existingIndentation.$this->whitespacesConfig->getIndent();
         $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
         if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
-            $this->ensureWhitespaceAtIndex(
+            $tokens->addNewlineAndIndent(
                 $tokens,
                 $endFunctionIndex,
                 $existingIndentation,
@@ -233,7 +233,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         if ($tokens[$index + 2]->isComment()) {
             $nextMeaningfulTokenIndex = $tokens->getNextMeaningfulToken($index + 2);
             if (!$this->isNewLine($tokens[$nextMeaningfulTokenIndex - 1])) {
-                $this->ensureWhitespaceAtIndex(
+                $this->addNewlineAndIndent(
                     $tokens,
                     $nextMeaningfulTokenIndex,
                     $indentation,
@@ -244,7 +244,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
             return;
         }
 
-        $this->ensureWhitespaceAtIndex($tokens, $index + 1, $indentation, $override);
+        $this->addNewlineAndIndent($tokens, $index + 1, $indentation, $override);
     }
 
     /**
@@ -255,7 +255,7 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
      * @param string $indentation the indentation that should be used
      * @param bool   $override    whether to override the existing character or not
      */
-    private function ensureWhitespaceAtIndex(Tokens $tokens, $index, $indentation, $override)
+    private function addNewlineAndIndent(Tokens $tokens, $index, $indentation, $override)
     {
         $whitespaceToken = new Token([
             T_WHITESPACE,
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -105,7 +105,9 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
                 if (!$meaningfulTokenBeforeParenthesis->isKeyword()
                     || $meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
                     if ($this->fixFunction($tokens, $index) && $this->configuration['ensure_fully_multiline']) {
-                        $this->ensureFunctionFullyMultiline($tokens, $index);
+                        if (!$meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
+                            $this->ensureFunctionFullyMultiline($tokens, $index);
+                        }
                     }
                 }
             }
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -245,12 +245,15 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
 
             return;
         }
-        $method = $override ? 'overrideAt' : 'insertAt';
 
-        $tokens->$method($index + 1, new Token([
-            T_WHITESPACE,
-            $this->whitespacesConfig->getLineEnding().$indentation,
-        ]));
+        call_user_func(
+            [$tokens, $override ? 'overrideAt' : 'insertAt'],
+            $index + 1,
+            new Token([
+                T_WHITESPACE,
+                $this->whitespacesConfig->getLineEnding().$indentation,
+            ])
+        );
     }
 
     /**
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -67,10 +67,17 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
                     ['ensure_fully_multiline' => true]
                 ),
                 new CodeSample(
-                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);",
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample(1,  2);",
                     [
                         'ensure_fully_multiline' => true,
                         'keep_multiple_spaces_after_comma' => true,
+                    ]
+                ),
+                new CodeSample(
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,  \n    2);\nsample(1,  2);",
+                    [
+                        'ensure_fully_multiline' => true,
+                        'keep_multiple_spaces_after_comma' => false,
                     ]
                 ),
             ]
--- a/src/Fixer/PhpUnit/PhpUnitFqcnAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitFqcnAnnotationFixer.php
@@ -74,7 +74,8 @@ final class MyTest extends \PHPUnit_Framework_TestCase
         foreach ($tokens as $index => $token) {
             if ($token->isGivenKind(T_DOC_COMMENT)) {
                 $tokens[$index] = new Token([T_DOC_COMMENT, preg_replace(
-                    '~^(\s*\*\s*@(?:expectedException|covers|coversDefaultClass|uses)\h+)(\w.*)$~m', '$1\\\\$2',
+                    '~^(\s*\*\s*@(?:expectedException|covers|coversDefaultClass|uses)\h+)(\w.*)$~m',
+                    '$1\\\\$2',
                     $token->getContent()
                 )]);
             }
--- a/src/Fixer/Whitespace/IndentationTypeFixer.php
+++ b/src/Fixer/Whitespace/IndentationTypeFixer.php
@@ -33,8 +33,7 @@ final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwa
     {
         return new FixerDefinition(
             'Code MUST use configured indentation type.',
-            [new CodeSample("<?php\n\nif (true) {\n\techo 'Hello!';\n}"),
-        ]
+            [new CodeSample("<?php\n\nif (true) {\n\techo 'Hello!';\n}")]
         );
     }
 
--- a/src/Fixer/Whitespace/IndentationTypeFixer.php
+++ b/src/Fixer/Whitespace/IndentationTypeFixer.php
@@ -34,7 +34,8 @@ final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwa
         return new FixerDefinition(
             'Code MUST use configured indentation type.',
             [new CodeSample("<?php\n\nif (true) {\n\techo 'Hello!';\n}"),
-        ]);
+        ]
+        );
     }
 
     /**
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -38,7 +38,7 @@ final class RuleSet implements RuleSetInterface
             'line_ending' => true,
             'lowercase_constants' => true,
             'lowercase_keywords' => true,
-            'method_argument_space' => true,
+            'method_argument_space' => ['ensure_fully_multiline' => true],
             'no_closing_tag' => true,
             'no_spaces_after_function_name' => true,
             'no_spaces_inside_parenthesis' => true,
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -72,6 +72,7 @@ final class RuleSet implements RuleSetInterface
             'include' => true,
             'lowercase_cast' => true,
             'magic_constant_casing' => true,
+            'method_argument_space' => true,
             'method_separation' => true,
             'native_function_casing' => true,
             'new_with_braces' => true,
--- a/src/Test/IntegrationCase.php
+++ b/src/Test/IntegrationCase.php
@@ -128,13 +128,17 @@ final class IntegrationCase
         if (!is_string($name)) {
             throw new \InvalidArgumentException(sprintf(
                 'Requirement key must be a string, got "%s".',
-                is_object($name) ? get_class($name) : gettype($name).'#'.$name));
+                is_object($name) ? get_class($name) : gettype($name).'#'.$name
+            ));
         }
 
         if (!array_key_exists($name, $this->requirements)) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
                 'Unknown requirement key "%s", expected any of "%s".',
-                $name, implode('","', array_keys($this->requirements)))
+                $name,
+                implode('","', array_keys($this->requirements))
+            )
             );
         }
 
--- a/src/Test/IntegrationCase.php
+++ b/src/Test/IntegrationCase.php
@@ -133,13 +133,11 @@ final class IntegrationCase
         }
 
         if (!array_key_exists($name, $this->requirements)) {
-            throw new \InvalidArgumentException(
-                sprintf(
+            throw new \InvalidArgumentException(sprintf(
                 'Unknown requirement key "%s", expected any of "%s".',
                 $name,
                 implode('","', array_keys($this->requirements))
-            )
-            );
+            ));
         }
 
         return $this->requirements[$name];
--- a/src/Test/IntegrationCaseFactory.php
+++ b/src/Test/IntegrationCaseFactory.php
@@ -91,16 +91,20 @@ final class IntegrationCaseFactory
         ]);
 
         if (!is_string($parsed['indent'])) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
                 'Expected string value for "indent", got "%s".',
-                is_object($parsed['indent']) ? get_class($parsed['indent']) : gettype($parsed['indent']).'#'.$parsed['indent'])
+                is_object($parsed['indent']) ? get_class($parsed['indent']) : gettype($parsed['indent']).'#'.$parsed['indent']
+            )
             );
         }
 
         if (!is_string($parsed['lineEnding'])) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
                 'Expected string value for "lineEnding", got "%s".',
-                is_object($parsed['lineEnding']) ? get_class($parsed['lineEnding']) : gettype($parsed['lineEnding']).'#'.$parsed['lineEnding'])
+                is_object($parsed['lineEnding']) ? get_class($parsed['lineEnding']) : gettype($parsed['lineEnding']).'#'.$parsed['lineEnding']
+            )
             );
         }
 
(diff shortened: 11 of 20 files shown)


tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -223,6 +223,23 @@ function xyz(
     );
 ",
             ),
+            'test half-multiline function gets full-multiline' => array(
```
greg0ire: Other test cases use comments for this, but I think this will yield better error messages.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -223,6 +223,23 @@ function xyz(
...
+                <<<'EXPECTED'
```
greg0ire: Are you ok with HEREDOC here? I find it is nicer to look at because the indention is no longer broken.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -54,24 +54,14 @@ public function testFixProvider()
...
                 "<?php xyz(\$a=10 , \$b=20 ,\t \$c=30);",
             ),
-            // test method call with \n not affected
-            array(
-                "<?php xyz(\$a=10, \$b=20,\n                    \$c=30);",
-            ),
-            // test method call with \r\n not affected
-            array(
-                "<?php xyz(\$a=10, \$b=20,\r\n                    \$c=30);",
-            ),
```
greg0ire: These method calls should definitely be affected, but by the newline fixer.

greg0ire:
I'm kind of stuck with this strange message : 

```
1) PhpCsFixer\Tests\Fixer\FunctionNotation\MethodArgumentSpaceFixerTest::testFix with data set "argument after multiline comment" ('<?php\nfunction xyz(\n    $a=10...0) {\n}', '<?php\nfunction xyz(\n    $a=10..
Code build on input code must match expected code.
--- Expected
+++ Actual
@@ @@
 <?php
 function xyz(
     $a=10, /* multiline comment
              not at the end of line
-            */
+            */
     $b=2000,
     $a2=10 /* multiline comment
              not at the end of line
             */,
     $b2=2000,
     $c=30) {
 }

/home/greg/dev/PHP-CS-Fixer/src/Test/AbstractFixerTestCase.php:138
/home/greg/dev/PHP-CS-Fixer/tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php:32
```

I can't spot what the problem is.

GrahamCampbell:
> I can't spot what the problem is.

The 2nd line has an extra space on it in that diff. After the `/`, though not in your copy/paste of it there. I guess GitHub trimmed it.

greg0ire:
Thanks! It seems to be a bug with neovim : when I run `! phpunit %` from it I don't see the space.

src/Console/Command/DescribeCommand.php
```diff
@@ -84,7 +84,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
             throw new \InvalidArgumentException(sprintf(
                 '%s %s not found.%s',
-                ucfirst($e->getType()), $name, null === $alternative ? '' : ' Did you mean "'.$alternative.'"?'
+                ucfirst($e->getType()),
+                $name,
+                null === $alternative ? '' : ' Did you mean "'.$alternative.'"?'
```
greg0ire: Looks like PHP-CS-Fixer has some PSR2 violations 😅

src/Fixer/PhpUnit/PhpUnitFqcnAnnotationFixer.php
```diff
@@ -30,7 +30,8 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
         foreach ($tokens as $token) {
             if ($token->isGivenKind(T_DOC_COMMENT)) {
                 $token->setContent(preg_replace(
-                    '~^(\s*\*\s*@(?:expectedException|covers|coversDefaultClass|uses)\h+)(\w.*)$~m', '$1\\\\$2',
+                    '~^(\s*\*\s*@(?:expectedException|covers|coversDefaultClass|uses)\h+)(\w.*)$~m',
+                    '$1\\\\$2',
```
greg0ire: Existing violation

src/Test/IntegrationCase.php
```diff
@@ -139,7 +139,8 @@ public function getRequirement($name)
         if (!array_key_exists($name, $this->requirements)) {
             throw new \InvalidArgumentException(sprintf(
                 'Unknown requirement key "%s", expected any of "%s".',
-                $name, implode('","', array_keys($this->requirements)))
+                $name,
+                implode('","', array_keys($this->requirements)))
```
greg0ire: Existing violation

greg0ire:
Question : should it be the responsability of this fixer to reindent things properly?

Consider this code snippet : 

```php
<?php
functionCall(array(
    'array1' => 'array2'
)
);
```

Currently, the fixer produces this : 

```php
<?php
functionCall(
    array(
    'array1' => 'array2'
    )
);
```

It it its responsability to properly reindent this? Or is there some other fixer that should take care of this?

greg0ire:
> Is it its responsability to properly reindent this? Or is there some other fixer that should take care of this?

@SpacePossum @GrahamCampbell can you please advise on this?

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -97,6 +102,48 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+
+            if ($this->isNewLine($token)) {
+                $isMultiline = true;
+            }
+        }
+
+        return $isMultiline;
+    }
+
+    private function ensureFunctionFullyMultiline(Tokens $tokens, $startFunctionIndex)
```
SpacePossum: yes, no worries, follow the interfaces and name things a bit logical and all good :)

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
SpacePossum: not sure I understand

src/Console/Command/DescribeCommand.php
SpacePossum: I'm surprised there aren't more of there TBH. I apply this exception in my projects as well,
for example
```php
<?php
$a = sprintf(
    '%d %d %d %d %d %d %d %d', 
    $a, $a, $a, $a, $a, $a, $a, $a
);
```

vs.

```php
<?php
$a = sprintf(
    '%d %d %d %d %d %d %d %d', 
    $a,
    $a, 
    $a, 
    $a,
    $a, 
    $a, 
    $a, 
    $a
);
```

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: This was testing that this piece of code is not affected. Now that the fixer also fixes newlines, the piece of code *should* be affected, so I removed the test.

SpacePossum:
>Is it its responsability to properly reindent this? Or is there some other fixer that should take care of this?

I would say no, other fixers should take care of that.
This _does_ mean you need to get the priority right for the fixer (which means tests (FixerFactoryTest and integration/fixture tests))

>make the fixer optional? @SpacePossum you talked about this, what did you mean exactly? How can a fixer be optional?

IIRC I was talking about not creating a new fixer, but add the functionality of the fixer to an existing fixer.
I think there a clear advantages for doing that, but than again not my call ;)

greg0ire:
> I would say no, other fixers should take care of that.
This does mean you need to get the priority right for the fixer (which means tests (FixerFactoryTest and integration/fixture tests))

Thanks for your insight, I'll try to have a look at that.

greg0ire:
@SpacePossum when I create a file with 

```php
<?php

functionCall(
    array(
    'array1' => 'array2',
    )
);
```

and run `php-cs-fixer fix --level=psr2 myfile.php`, nothing gets fixed. Is indenting outside php-cs-fixer's scope ?

SpacePossum:
>Is indenting outside php-cs-fixer's scope ?

not out of scope, but this case is prop. like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/770 (so a feature is missing)

greg0ire:
@SpacePossum : ok so let's keep things atomic (if that's what is meant in the cookbook), and let the indention be handled by some other fixer someday. I'm done, this is ready for review.

greg0ire:
> ERROR: Empty checksums are no longer allowed by default for non-secure sources. Please ask the maintainer to add checksums to this package. In the meantime if you need this package to work correctly, please enable the feature allowEmptyChecksums, provide the runtime switch '--allow-empty-checksums', or pass in checksums at runtime (recommended - see 'choco install -?' / 'choco upgrade -?' for details). It is strongly advised against allowing empty checksums for non-internal HTTP/FTP sources.

Looks like you have to add checksums for `http://windows.php.net/downloads/releases/archives/php-7.0.9-nts-Win32-VC14-x64.zip`

greg0ire:
Or ignore them like in #2443

src/Cache/Cache.php
```diff
@@ -58,8 +58,8 @@ public function set($file, $hash)
         if (!is_int($hash)) {
             throw new \InvalidArgumentException(sprintf(
                 'Value needs to be an integer, got "%s".',
-                is_object($hash) ? get_class($hash) : gettype($hash))
-            );
+                is_object($hash) ? get_class($hash) : gettype($hash)
+            ));
```
Soullivaneuh: Should not this be on a separated commit?

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -43,7 +43,8 @@ protected function configure()
                 )
             )
             ->setDescription('Update php-cs-fixer.phar to the latest stable version.')
-            ->setHelp(<<<'EOT'
+            ->setHelp(
+                <<<'EOT'
```
Soullivaneuh: I'm not sure this is correct. It's not really a multi-line arguments IMHO, we only have one.

src/Cache/Cache.php
greg0ire: I don't think so, this is the very problem this fixer is fixing. Not fixing it in that commit would make the commit invalid (the build for this commit wouldn't be green).

Soullivaneuh:
Looks good, thanks for this PR! :+1: 

I would separate the fixer update and the CS fix of the project onto two separated commit, but the decision goes to the maintainer.

src/Console/Command/SelfUpdateCommand.php
greg0ire: Heredoc arguments can't be anything else than multi-line arguments, because as you can't write this: https://3v4l.org/RUlJB

greg0ire:
@keradus please review this

keradus:
sadly, not a chance soon.
anyone else is willing ?

greg0ire:
>  anyone else is willing ?

*crickets*

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -96,7 +103,100 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+            }
+
+            // skip nested arrays
+            if ($token->isGivenKind(CT::T_ARRAY_SQUARE_BRACE_CLOSE)) {
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_ARRAY_SQUARE_BRACE, $index, false);
+                continue;
             }
+
+            if ($token->equals(',')) {
+                $this->fixNewline(
```
julienfalque: I think newlines are not needed here.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -96,7 +103,100 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+            $searchIndex = $prevWhitespaceTokenIndex;
+        } while (
+            $prevWhitespaceTokenIndex
+            && strpos($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n") === false
+        );
+        $existingIndentation = $prevWhitespaceTokenIndex ?
+            ltrim($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n") :
+            '';
+
+        $indentation = $existingIndentation.'    ';
```
julienfalque: This is not valid in case the file uses different indentation style (e.g. tabs or two spaces). Please add test cases that cover that.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -96,7 +103,100 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+        $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
+        if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
+            $tokens->insertAt(
+                $endFunctionIndex,
+                new Token(array(T_WHITESPACE, "\n$existingIndentation"))
```
julienfalque: This is not valid in case the file uses line endings different than `\n` (e.g. `\r\n`). Please add test cases that cover that.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -223,6 +220,109 @@ function xyz(
...
+        'test code that should not be affected (because not a function nor a method)' => array(
+            <<<'UNAFFECTED'
+<?php
+if (true &&
+    true
+    ) {
+    // do whatever
+}
+UNAFFECTED
+            ),
```
julienfalque: Please add test cases with nested function calls/arrays.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: I'm ok with it, but who are you asking?

julienfalque: Anybody having an opinion about this :)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -156,4 +256,16 @@ private function isCommentLastLineToken(Tokens $tokens, $index)
...
+    /**
+     * Checks if token is new line.
+     *
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNewLine(Token $token)
+    {
+        return trim($token->getContent(), " \r\t\0\x0B") === "\n";
```
greg0ire: I think it would do something different... this checks if a given token contains a newline, not if it only contains that, right?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.php_cs.dist#L24

greg0ire: @keradus I did use NOWDOC

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: You're right, but you confused me for a moment ;)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
julienfalque: That's what `\R` matches: line endings (`\n`, `\r` and `\r\n`).

greg0ire:
5 months later, with changes all over the codebase, this was of course very painful to rebase. I made a separate, ultra mergeable PR so that this does not happen again : #2779 
I don't know if I will ever finish this one, but I sure won't touch it until #2779 is merged.

julienfalque:
I'm very sorry that the rebase was complicated.

Unfortunately we try to avoid fixing code style violations without implementing a fixer so we are not going to merge #2779 for now. As it does not include the fixer, it would not prevent new violations to be introduced until this PR is merged as well anyway.

Also we don't want to use another code style tool as it might have unwanted rules, and most wanted rules are probably covered by PHP CS Fixer itself.

So please, let's focus on this PR for now, we'll try to review it more often so you don't need to handle a new epic rebase :)

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -134,59 +134,79 @@ public function testFixProvider()
...
-                                */ $b=2000,
-                        $a2=10 /* multiline comment
-                                 not at the end of line
-                                */ , $b2=2000,
-                        $c=30) {
-                    }',
+            ),
+            'multiline comments also must be ignored' => array(
+                <<<'EXPECTED'
+<?php function xyz(
```
keradus: :-1: for all that changes.
having code indented actually have a reason - we were trapped few times that fixer works when code was not extra intended, but crashed when it was.

like this works:
```php
if (1) {
    a();
}
```

but this not:
```php
    if (1) {
        a();
    }
````

you could achieve proper argument separation without changing indentation

**Comment by keradus** [comment]
PR title:
> Add newline argument fixer

Where is that new fixer?

Instead, one of existing fixers were modified. It now does more that it claims to do in definition.
It was supposed to only remove spaces before comma/add single space after it. Now, it adds new lines. It's nice to add missing new lines in general, but it's not the purpose of that fixer. And then it started to touch arrays as well. That violates fixer defined behaviour. It's a bc breaker.

**Comment by greg0ire** [comment]
@keradus here is the rationale behind not creating a new fixer : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2435#issuecomment-269795217

**Comment by greg0ire** [comment]
> Also we don't want to use another code style tool as it might have unwanted rules

I think using phpcs with the PSR2 standard would by definition not have any unwanted rules, right?

> So please, let's focus on this PR for now, we'll try to review it more often so you don't need to handle a new epic rebase :)

Ok, first things first : should I use a new fixer or modify an existing one? @keradus @SpacePossum , you seem to have conflicting opinions on the matter, which should it be?

**Comment by keradus** [comment]
from that comment:

> remove some from the current and add a new fixer

So, make new fixer that will have nothing in common with existing one, so there would not be 2 fixers with partially same functionality. Existing one is removing the space before `,`, add one after `,` when needed, New one is adding `\n` when needed, but not removing spaces before `,`.
Or...

> or add more to the current fixer by configuration option

have a config option, so by default behaviour of existing fixer will be the same, and then it could be altered by providing different configuration.

-------------------------

2 ways to go was proposed, none was choosen.

**Comment by keradus** [comment]
> @keradus @SpacePossum , you seem to have conflicting opinions on the matter, which should it be?

no, that is exactly the same opinion we presented. you just didn't pick it.

**Comment by greg0ire** [comment]
Ok, I don't think I have it in me to achieve this properly, I think someone else will do a better job, sorry.

**Comment by keradus** [comment]
You just need to run `ensureFunctionFullyMultiline` conditionally if `$this->configuration['multiline']` (or sth like this), second config for handling arrays together with methods.
Take a look on others configurable fixers, it's easy to apply that solution

**Comment by greg0ire** [comment]
Ok I thought for a moment I had to rewrite it all. Maybe I will take a look at that and finish it.

**Comment by greg0ire** [comment]
I updated the TODO list

File: src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php (hunk 14 above)
Comment (greg0ire) [inline src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php]: That's right. And that's not what `isNewLine` does, right? So... this is resolved?

File: tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php (hunk 15 above)
Comment (greg0ire) [inline tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php]: Ok, I will restore the indention, but keep the nowdoc

File: tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php [hunk 16]
```diff
@@ -134,59 +134,79 @@ public function testFixProvider()
...
                     $b=2000,//comment2
-                    $c=30) {
+                    $c=30
+                ) {
                 }',
             ],
             'multiline comments also must be ignored (I)' => [
-                '<?php function xyz(
+                <<<'EXPECTED'
+<?php function xyz(
```
greg0ire: @keradus am I allowed to reindent this? it looks weird.

SpacePossum: Changing the indent of a nowdoc changes the value of the content. In this case adding white space before the open tag effectivly changing the test case. I'm not sure why this string needs changing to nowdoc here?

greg0ire: I know it does, but I don't think that's really what's under test here. 

> I'm not sure why this string needs changing to nowdoc here?

It doesn't, it's just that I think nowdoc is better in this case, for various reasons, the first of which is : you are more sure of what you are testing. I bet the author of this meant to test

```
    function ()
    {
    }
```

and not

```
function ()
    {
    }
```

keradus: as I stated here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440#discussion_r116956645 , different indent is intended

Sure, it's not needed for every single test case, but for some of them it was important, even crucial, as it was testing reported bugs.
Let us not investigating "was it important in given case" for every test case you reindent. 
That's why golden rule with utests stated - add new tests, but don't modify existing ones.

greg0ire: > different indent is intended

You stated that extra indent was intended, not mixed indent.

> That's why golden rule with utests states - add new tests, but don't modify existing ones.

I guess since my feature is optional now, I'm no longer forced to change these tests, so I'll revert them back to how they were written

README.rst
```diff
@@ -545,6 +545,8 @@ Choose from the list of available rules:
 
   Configuration options:
 
+  - ``ensure_fully_multiline`` (``bool``): ensure every argument of a multiline
```
keradus: please extend definition by adding example with that option

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
julienfalque: This is used to detect newlines around arguments, right? What if the token contains several line endings (e.g. `\n\n`)?

greg0ire: I guess it no longer works. I'll add a test case.

greg0ire: Done and fixed with `strpos`

README.rst
greg0ire: Done. I also appended the rule as described in the PSR-2 guide in the description

keradus:
please avoid squashing commits. now, it's very hard to figure out what is the lates change and to see latest changes one need to read whole diff again...

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -61,6 +62,10 @@ public function getDefinition()
                     "<?php\nfunction sample(\$a=10,\$b=20,\$c=30) {}\nsample(1,  2);",
                     ['keep_multiple_spaces_after_comma' => true]
                 ),
+                new CodeSample(
+                    "<?php\nfunction sample(\$a=10,\n    \$b=20,\$c=30) {}\nsample(1,\n    2);",
+                    ['ensure_fully_multiline' => true]
```
keradus: what if one would set `['ensure_fully_multiline' => true, 'keep_multiple_spaces_after_comma => true]` ?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -250,6 +263,212 @@ function xyz(
...
+                ['ensure_fully_multiline' => true],
+            ],
+            'test multiline array arguments do not trigger multiline' => [
+                <<<'EXPECTED'
+<?php
+defraculate(1, array(
+    'a',
+    'b',
+    'c',
+), 42);
```
greg0ire: @julienfalque you asked met to add tests with nested arrays, what do you mean exactly? Can you give an example?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,6 +37,18 @@ public function testFix($expected, $input = null, array $configuration = null)
...
             $this->fixer->configure($configuration);
         }
+        $indent = '    ';
+        if (strpos((string) $input, "\t") !== false) {
+            $indent = "\t";
+        }
+        if (preg_match('/\n  \S/', (string) $input)) {
+            $indent = '  ';
+        }
+        $lineEnding = strpos((string) $input, "\r") === false ? "\n" : "\r\n";
```
julienfalque: I retrieved your branch and rewrote the tests:
* removed the `\r` from the `test violation after opening parenthesis, with \r\n line endings` case
* added a new test that uses the same data provider and replaces all `\n` with `\r\n`

```php
/**
 * @dataProvider testFixProvider
 */
public function testFix(/* ... */)
{
    // current test without the \r check and always using \n as line ending
}

/**
 * @dataProvider testFixProvider
 */
public function testFixWithDifferentLineEndings(/* ... */)
{
    // ...
    $expected = str_replace("\n", "\r\n", $expected);
    if (null !== $input) {
        $input = str_replace("\n", "\r\n", $input);
    }

    $this->fixer->setWhitespacesConfig(new WhitespacesFixerConfig(
        $indent,
        "\r\n"
    ));

}
```

This way, all cases are tested with `\n` **and** `\r\n`. I then got one failure, didn't investigate it for now, could you please have a look?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
julienfalque: I meant something like:
```php
foo('a',
    'b',
    [
        'c',
        'd', bar('e', 'f'),
        baz('g',
            ['h',
                'i',
            ]),
    ]);
```
which may be fixed like this:
```php
foo(
    'a',
    'b',
    [
        'c',
        'd', bar('e', 'f'),
        baz(
            'g',
            ['h',
                'i',
            ]
        ),
    ]
);
```

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -97,6 +104,13 @@ protected function createConfigurationDefinition()
                 ->setAllowedTypes(['bool'])
                 ->setDefault(false)
                 ->getOption(),
+            (new FixerOptionBuilder(
+                'ensure_fully_multiline',
+                'Ensure every argument of a multiline argument lists is on its own line'
```
greg0ire: You could also write "each and every", so I guess this means both are probably equally valid

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -187,4 +309,16 @@ private function isCommentLastLineToken(Tokens $tokens, $index)
...
+    /**
+     * Checks if token is new line.
+     *
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNewLine(Token $token)
+    {
+        return strpos(trim($token->getContent(), " \r\t\0\x0B"), "\n") !== false;
```
greg0ire: > Is it possible that someone still uses \r as line endings?

Isn't that what MacOS users use?

SpacePossum: I theory one could use `\r` only, but this very uncommon (https://en.wikipedia.org/wiki/Newline), typically `\n` for Unix and MacOS, `\r\n` for Windows. We don't support `\r` only, so don't worry about it.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: Could you please make a PR on my repo?

greg0ire: (or push directly on this branch, I think you have the necessary permissions)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
keradus: PHP CS Fixer actually support only `\n` and `\r\n` - we don't check other new-line combinations in rest of the tool.

`\r` was used by classic MacOS (before 2001 release of MacOs X)

greg0ire: TIL. Thanks!

greg0ire:
> please avoid squashing commits. now, it's very hard to figure out what is the lates change and to see latest changes one need to read whole diff again...

@keradus GH has this feature that lets you view changes since your last page view, but it can only be used once. I will refrain from squashing commits from no on.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: I added an example for the former, but the latter already documented, right? I mean since `false` is the default value for `keep_multiple_spaces_after_comma`... or do I need to explicitly add the option line 67?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: I copy/pasted your input, it works exactly as specified :)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: done

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -97,6 +104,13 @@ protected function createConfigurationDefinition()
                 ->setAllowedTypes(['bool'])
                 ->setDefault(false)
                 ->getOption(),
+            (new FixerOptionBuilder(
+                'ensure_fully_multiline',
+                'Ensure every argument of a multiline argument lists is on its own line'
+            ))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
```
greg0ire: done

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -109,7 +109,7 @@ protected function createConfigurationDefinition()
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('keep_multiple_spaces_after_comma', 'Whether keep multiple spaces after comma.'))
                 ->setAllowedTypes(['bool'])
-                ->setDefault(false)
+                ->setDefault(false) // TODO default to true
```
keradus:
```diff
-// TODO default to true
+// @TODO should be true at 3.0
```

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +146,111 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+            if ($token->isGivenKind(CT::T_ARRAY_SQUARE_BRACE_CLOSE)) {
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_ARRAY_SQUARE_BRACE, $index, false);
+                continue;
+            }
+
+            if ($token->equals(',')) {
+                $this->fixNewline($tokens, $index, $indentation);
+            }
+        }
+        if (!$this->isNewline($tokens[$startFunctionIndex + 1])) {
```
greg0ire: Yes, I did it by adding a new parameter. Is that what you were thinking of? Or did you want something that would override the existing token only if it is a whitespace?

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,6 +37,18 @@ public function testFix($expected, $input = null, array $configuration = null)
         if (null !== $configuration) {
             $this->fixer->configure($configuration);
         }
+        $indent = '    ';
+        if (strpos((string) $input, "\t") !== false) {
```
greg0ire: done

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: done

greg0ire:
@julienfalque I will probably have some time to spend on this tonight, if you can create the PR to my repo before, that would be great, and of course, not a big deal if you can't :)

greg0ire:
Nevermind @julienfalque I did something similar (remove the test data instead of altering it, called the first method with different arguments), and if fails indeed.

greg0ire:
@julienfalque fixed
@keradus @julienfalque please review and tell me to squash if necessary.

greg0ire:
Let's trigger a new build : https://twitter.com/packagist/status/867084819276148736

greg0ire:
All green :)

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,10 +37,50 @@ public function testFix($expected, $input = null, array $configuration = null)
...
+        $indent = '    ';
+        $lineEnding = "\n";
+        if (null !== $input) {
+            if (strpos((string) $input, "\t") !== false) {
+                $indent = "\t";
+            }
+            if (preg_match('/\n  \S/', (string) $input)) {
+                $indent = '  ';
+            }
+            if (strpos($input, "\r") !== false) {
```
julienfalque: This `if` can be dropped now.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,10 +37,50 @@ public function testFix($expected, $input = null, array $configuration = null)
         if (null !== $configuration) {
             $this->fixer->configure($configuration);
         }
+        $indent = '    ';
+        $lineEnding = "\n";
```
julienfalque: This value can now be passed directly where used.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,10 +37,50 @@ public function testFix($expected, $input = null, array $configuration = null)
...
+    public function testFixWithDifferentLineEndings(
+        $expected,
+        $input = null,
+        array $configuration = null
+    ) {
+        if (null !== $input) {
+            $input = str_replace("\n", "\r\n", $input);
+        }
+
+        return $this->testFix(
```
julienfalque: Missing `$this->fixer->setWhitespacesConfig()` call here.

greg0ire: It doesn't miss : it's called inside testFix, with the correct parameters.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: No it can't see reason below

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: Same as below

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
julienfalque: My bad, didn't notice the new test was calling the other one.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -81,8 +93,10 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
         for ($index = $tokens->count() - 1; $index > 0; --$index) {
             $token = $tokens[$index];
 
-            if ($token->equals('(') && !$tokens[$index - 1]->isGivenKind(T_ARRAY)) {
-                $this->fixFunction($tokens, $index);
+            if ($token->equals('(') && !$tokens[$index - 1]->isGivenKind([T_ARRAY, T_WHITESPACE])) {
```
SpacePossum: Not sure what is special/diff between array with and without white space in front of it.
Didn't test, but I think there is an existing bug in the code which is dealt with here in way that does not cover all cases.
 maybe this should be something like;
 ```php
&& !$tokens->getPrevMeaningfulToken($index)->isGivenKind(T_ARRAY)
```

please add a test like
```php
<?php
$a = array/**/(  1);
$a = array/**/( 12,    7);
$a = array/***/(123,  7);
$a = array (        1,
2);
```

this case should _not_ be touched/fixed

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +153,103 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
 
             if ($token->equals(',')) {
                 $this->fixSpace2($tokens, $index);
+                if (!$isMultiline && $this->isNewline($tokens[$index + 1])) {
+                    $isMultiline = true;
```
SpacePossum: can this be an early return? there is no need to keep iterating I think

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +153,103 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+        $searchIndex = $startFunctionIndex;
+        do {
+            $prevWhitespaceTokenIndex = $tokens->getPrevTokenOfKind(
+                $searchIndex,
+                [[T_WHITESPACE]]
+            );
+            $searchIndex = $prevWhitespaceTokenIndex;
+        } while (
+            $prevWhitespaceTokenIndex
+            && strpos($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n") === false
```
SpacePossum: please make this Yoda style

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +153,103 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+        );
+        $existingIndentation = $prevWhitespaceTokenIndex ?
```
SpacePossum: we use this style in the repo
```php
    $existingIndentation = $prevWhitespaceTokenIndex 
        ? ltrim($tokens[$prevWhitespaceTokenIndex]->getContent(), "\n\r") 
        : ''
    ;
```

please update :)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -187,4 +308,16 @@ private function isCommentLastLineToken(Tokens $tokens, $index)
...
+    /**
+     * Checks if token is new line.
+     *
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNewLine(Token $token)
+    {
+        return strpos($token->getContent(), "\n") !== false;
```
SpacePossum: please yoda this :)

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,10 +37,50 @@ public function testFix($expected, $input = null, array $configuration = null)
         if (null !== $configuration) {
             $this->fixer->configure($configuration);
         }
+        $indent = '    ';
+        $lineEnding = "\n";
+        if (null !== $input) {
+            if (strpos((string) $input, "\t") !== false) {
```
SpacePossum: >@param null|string $input

so either the cast is overcomplete (`null !== $input` so must be s string), or the PHPDoc is of

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: There is no need for Yoda here, `strpos(...) = false` will just crash. Kill the Jedi.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: Will do.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: Makes no sense, see above.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
greg0ire: I forgot to remove the cast

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
SpacePossum: there is no need, like for most CS rules ;)  however there is a standard in the repo

greg0ire: Sad to hear it was adopted here too, and also in the instances where there is no risk for assignment. IMO this is outdated, horrible style that makes no sense when you have tests and code review but we already discussed this so I guess I'll just indulge to the crime against humanity this time.

SpacePossum: thanks for understanding and updating :)  :+1:

greg0ire: I will :)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: done

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: Added a break

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: This seems to have an impact on `if` / `while` statements that are multiline

greg0ire: Is there a better way to detect a function declaration / function call? If I end up listing all token kinds here this will get hairy.

greg0ire: `isKeyword`, maybe

greg0ire: `list` is a keyword it seems...

greg0ire: I settled for not keywords except list

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: done

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: Done
yikes

greg0ire:
I rebased yesterday evening. Can you please review again and test @julienfalque ?

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -105,10 +130,15 @@ protected function createConfigurationDefinition()
      *
      * @param Tokens $tokens             Tokens to handle
      * @param int    $startFunctionIndex Start parenthesis position
+     *
+     * @return bool whether the function is multiline
      */
     private function fixFunction(Tokens $tokens, $startFunctionIndex)
     {
+        $isMultiline = false;
```
julienfalque: This line is not needed as the variable is overridden below.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +157,103 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+        $this->fixNewLine($tokens, $startFunctionIndex, $indentation, false);
+    }
+
+    /**
+     * Method to insert newline after comma or opening parenthesis.
+     *
+     * @param Tokens $tokens
+     * @param int    $index       index of a comma
+     * @param string $indention   the indention that should be used
+     * @param mixed  $indentation
```
julienfalque: duplicated `@param` (because the previous line has a typo).

greg0ire:
The build seems to fail because of this deprecation:

```
The Symfony\Component\Process\ProcessUtils::escapeArgument() method is deprecated since version 3.3 and will be removed in 4.0. Use a command line array or give env vars to the Process::start/run() method instead: 13x

    5x in ProcessLinterTest::testPrepareCommand from PhpCsFixer\Tests\Linter

    4x in ProcessLinterTest::testLintFile from PhpCsFixer\Tests\Linter

    4x in ProcessLinterTest::testLintSource from PhpCsFixer\Tests\Linter
```

greg0ire:
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2812

tests/AbstractDoctrineAnnotationFixerTestCase.php
```diff
@@ -86,7 +86,10 @@ private function withClassDocBlock($comment)
 %s
 class FooClass
 {
-}', $comment, false);
```
julienfalque: I'm not sure those cases should be fixed: newlines are within strings only, not around arguments. Also, a newline is now required after opening braces.

If those changes were made by the new behavior introduced in this PR, then you should add test cases to cover this.

tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
```diff
@@ -722,22 +724,28 @@ private function provideClassyCases($classy)
...
 %s/**/B //
-/**/ {}', $classy, $classy
+/**/ {}',
+                    $classy,
+                    $classy
                 ),
             ],
             [
-                sprintf('<?php
+              sprintf(
```
julienfalque: Indentation is wrong.

tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
```diff
@@ -762,13 +770,17 @@ private function provideClassyExtendingCases($classy)
                 sprintf('<?php
 %s Test124 extends
 \Exception
-{}', $classy),
```
julienfalque: This `sprintf` call and the one below were not multiline and did not need to be fixed. You should either revert these changes or add the missing newline after the opening brace.

tests/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixerTest.php
```diff
@@ -135,7 +135,12 @@ public function AB($self)
         return $this; // %s
      }
 }
-', $input, $input, $expected, $input);
+',
```
julienfalque: This `sprintf` call and the one below were not multiline and did not need to be fixed. You should either revert these changes or add the missing newline after the opening brace.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -36,10 +37,50 @@ public function testFix($expected, $input = null, array $configuration = null)
         if (null !== $configuration) {
             $this->fixer->configure($configuration);
         }
+        $indent = '    ';
+        $lineEnding = "\n";
+        if (null !== $input) {
+            if (false !== strpos($input, "\t")) {
+                $indent = "\t";
+            }
+            if (preg_match('/\n  \S/', $input)) {
```
julienfalque: `elseif`?

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -95,6 +113,13 @@ protected function createConfigurationDefinition()
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('keep_multiple_spaces_after_comma', 'Whether keep multiple spaces after comma.'))
                 ->setAllowedTypes(['bool'])
+                ->setDefault(false) // @TODO should be true at 3.0
```
greg0ire: wow good catch

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -81,8 +93,14 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
         for ($index = $tokens->count() - 1; $index > 0; --$index) {
             $token = $tokens[$index];
 
-            if ($token->equals('(') && !$tokens[$index - 1]->isGivenKind(T_ARRAY)) {
-                $this->fixFunction($tokens, $index);
+            if ($token->equals('(')) {
+                $meaningfulTokenBeforeParenthesis = $tokens[$tokens->getPrevMeaningfulToken($index)];
+                if (!$meaningfulTokenBeforeParenthesis->isKeyword()
+                    || $meaningfulTokenBeforeParenthesis->isGivenKind(T_LIST)) {
+                    if ($this->fixFunction($tokens, $index) && $this->configuration['ensure_fully_multiline']) {
```
greg0ire: I think the OTV is ok in this case

julienfalque: What's _OTV_?

greg0ire: One-time-variable (https://a-nickels-worth.blogspot.fr/2016/04/a-guide-to-naming-variables.html)

greg0ire:
I think the actual expected fix is

```
<?php
functionCall(
    'a', /* comment */
    'b',
    'c'
);
```
The space between 'a', /* comment */ is added by `fixSpace2`, that I did not write.

tests/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixerTest.php
greg0ire: I think I will revert all the changes and try the modified fixer on it.

tests/Fixer/ClassNotation/ClassDefinitionFixerTest.php
greg0ire: This one still needs to be fixed.

tests/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixerTest.php
greg0ire: This is the same issue I think, I will add tests for multiline strings

tests/AbstractDoctrineAnnotationFixerTestCase.php
greg0ire: Just added this code snippet as a failing test.

src/Fixer/Whitespace/IndentationTypeFixer.php
```diff
@@ -33,7 +33,8 @@ public function getDefinition()
         return new FixerDefinition(
             'Code MUST use configured indentation type.',
             [new CodeSample("<?php\n\nif (true) {\n\techo 'Hello!';\n}"),
-        ]);
+        ]
+        );
```
greg0ire: This is not the best fix, but the original code is very wrong in the first place...

src/Test/IntegrationCase.php
```diff
@@ -128,13 +128,17 @@ public function getRequirement($name)
...
 
         if (!array_key_exists($name, $this->requirements)) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
                 'Unknown requirement key "%s", expected any of "%s".',
-                $name, implode('","', array_keys($this->requirements)))
+                $name,
+                implode('","', array_keys($this->requirements))
+            )
```
greg0ire: The indention is wrong, but I think we established it wasn't the responsability of this fixer. And it looks like there is no fixer that can fix it afterwards.

tests/Report/TextReporterTest.php
```diff
@@ -63,7 +63,10 @@ public function testGenerateNoErrors()
 
     public function testGenerateSimple()
     {
-        $expectedReport = str_replace("\n", PHP_EOL, <<<'TEXT'
+        $expectedReport = str_replace(
+            "\n",
+            PHP_EOL,
+            <<<'TEXT'
```
greg0ire: A heredoc / nowdoc cannot end with something else than a newline. This means that any function call containing one should be multiline. This looks right to me.

greg0ire:
@julienfalque please review again :)

src/Test/IntegrationCase.php
julienfalque: The `sprintf(` part should remain on the same line as before so that indentation stays valid. Then both closing parentheses would be on the same line too. Same for similar changes in other files.

tests/Report/TextReporterTest.php
julienfalque: Please add a test case that covers this.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -187,4 +309,16 @@ private function isCommentLastLineToken(Tokens $tokens, $index)
...
+    /**
+     * Checks if token is new line.
+     *
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNewLine(Token $token)
+    {
+        return $token->isWhitespace() && (false !== strpos($token->getContent(), "\n"));
```
greg0ire: I know but I think it makes things a bit more clear. Should I remove them anyway?

src/Test/IntegrationCase.php
greg0ire: I know, and agree, but do you want me to fix this manually, or to change the fixer?
Originally, the code was like this:
```
if (!array_key_exists($name, $this->requirements)) {
            throw new \InvalidArgumentException(sprintf(
                'Unknown requirement key "%s", expected any of "%s".',
                $name,
                implode('","', array_keys($this->requirements)))
             );
         }
```

tests/Report/TextReporterTest.php
greg0ire: Done

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -37,7 +37,8 @@ public function getDefinition()
     $a = doubleval($b);
     $a = strval ($b);
     $a = boolval($b);
-'),
+'
+            ),
```
greg0ire: I think so too, but should I do this manually or with the fixer. I kept it like that because I don't want to hide the fixer's shortcomings, and I want to know if this is supposed to be its responsability or not. If it is, I don't think it's going to be a really simple thing.

julienfalque: The issue here is that both the array and the arguments list must be fixed. But indeed, the fixer you are working on should only care about arguments list, so please fix this manually instead.

Implementing a fixer for the array would be a nice-to-have though.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
julienfalque: IMO they don't, please remove them.

src/Test/IntegrationCase.php
greg0ire: I fixed it manually.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +156,101 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+                        $this->whitespacesConfig->getLineEnding().$indentation,
+                    ])
+                );
+            }
+
+            return;
+        }
+        $method = $override ? 'overrideAt' : 'insertAt';
+
+        $tokens->$method($index + 1, new Token([
```
keradus: `$foo->$bar()` syntax is not advised

greg0ire: I replaced that with use `call_user_func`

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -127,8 +156,104 @@ private function fixFunction(Tokens $tokens, $startFunctionIndex)
...
+                        $this->whitespacesConfig->getLineEnding().$indentation,
+                    ])
+                );
+            }
+
+            return;
+        }
+
+        call_user_func(
+            [$tokens, $override ? 'overrideAt' : 'insertAt'],
```
keradus: don't call method by string. it's impossible to detect method usage here.

keradus: `ensureWhitespaceAtIndex` perhaps ?

greg0ire: > ensureWhitespaceAtIndex perhaps ?

??
Do you want me to rename `fixNewline` like this?

greg0ire: Oh I think you want me to move just the `call_user_func` part to a private method. Will do.

greg0ire: Done.

keradus: `ensureWhitespaceAtIndex` is existing method in `Tokens` collection.
Yes, I wanted you to not call method via string like you did.

Please, either use `Tokens::ensureWhitespaceAtIndex` if it fits your case or rename your method to shrink down confusion

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
julienfalque: Please don't forget this one ;)

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -234,26 +233,42 @@ private function fixNewline(Tokens $tokens, $index, $indentation, $override = tr
...
+
+    /**
+     * Makes sure there is a whitespace at the given location
+     *
+     * @param Tokens $tokens      The token stream to modify
+     * @param int    $index       where to insert the whitespace
+     * @param string $indentation the indentation that should be used
+     * @param bool   $override    whether to override the existing character or not
+     */
+    private function ensureWhitespaceAtIndex(Tokens $tokens, $index, $indentation, $override)
```
keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440#discussion_r120087946

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: I renamed it because `ensureWhitespaceAtIndex` does many things that I don't do there.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -186,7 +186,7 @@ private function ensureFunctionFullyMultiline(Tokens $tokens, $startFunctionInde
         $indentation = $existingIndentation.$this->whitespacesConfig->getIndent();
         $endFunctionIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $startFunctionIndex);
         if (!$this->isNewline($tokens[$endFunctionIndex - 1])) {
-            $this->ensureWhitespaceAtIndex(
+            $tokens->addNewlineAndIndent(
```
keradus: that was exactly what I was talking about :D

greg0ire:
Rebased and green. Please review again.

src/Test/IntegrationCaseFactory.php
```diff
@@ -91,16 +91,20 @@ private function determineConfig($config)
         ]);
 
         if (!is_string($parsed['indent'])) {
-            throw new \InvalidArgumentException(sprintf(
+            throw new \InvalidArgumentException(
+                sprintf(
```
julienfalque: These changes (here and above in this file) are invalid: `sprintf(` part should remain on the previous line. You may group the closing parentheses instead.

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: Too many comments, it got drowned :P

greg0ire:
@julienfalque sorry, should be ok now.

greg0ire:
>  julienfalque approved these changes a day ago 

This made me so happy! @keradus , hopefully final review and merge?

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
keradus: what is a value of setting up `keep_multiple_spaces_after_comma` (to any `true` or `false`) when `ensure_fully_multiline=true` is set ?

what is a difference between `ensure_fully_multiline=true keep_multiple_spaces_after_comma=true` and `ensure_fully_multiline=true keep_multiple_spaces_after_comma=false`
?

keradus:
old, not resolved:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440/files#r117374636

src/RuleSet.php
```diff
@@ -38,7 +38,7 @@
             'line_ending' => true,
             'lowercase_constants' => true,
             'lowercase_keywords' => true,
-            'method_argument_space' => true,
+            'method_argument_space' => ['ensure_fully_multiline' => true],
```
keradus: put `'method_argument_space' => true,` to `@Symfony` and `'method_argument_space' => ['ensure_fully_multiline' => true],` to `.php_cs.dist`

keradus:
please rebase

src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
greg0ire: I think I understand what your concern is now, I will push something about that soon.

julienfalque:
I think the new code samples are wrong regarding the `keep_multiple_spaces_after_comma` option. See #2431 for details about it.

greg0ire:
@keradus pushed new examples and fix Travis
@julienfalque I don't understand what you mean.

Here is the output for the samples I added if that helps:

 * Example #5. Fixing with configuration: ['ensure_fully_multiline' => true, 'keep_multiple_spaces_after_comma' => true].

```diff
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ @@
    <?php
   -function sample($a=10,
   -    $b=20,$c=30) {}
   -sample(1,  
   -    2);
   +function sample(
   +    $a=10,
   +    $b=20,
   +    $c=30
   +) {}
   +sample(
   +    1,  
   +    2
   +);
    sample(1,  2);
   ----------- end diff -----------
```

 * Example #6. Fixing with configuration: ['ensure_fully_multiline' => true, 'keep_multiple_spaces_after_comma' => false].
```diff
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ @@
    <?php
   -function sample($a=10,
   -    $b=20,$c=30) {}
   -sample(1,  
   -    2);
   -sample(1,  2);
   +function sample(
   +    $a=10,
   +    $b=20,
   +    $c=30
   +) {}
   +sample(
   +    1,
   +    2
   +);
   +sample(1, 2);
   ----------- end diff -----------
```
The last line is the most important here IMO

julienfalque:
Oh right, with the last line they are valid, sorry. Is it possible to update them with something more visible? I bet I'm not the last one to miss it 😋

greg0ire:
Are you happier with this @julienfalque ?

julienfalque:
Better, thanks :)
I'm not sure about leaving trailing spaces when there is a newline though. Shouldn't `keep_multiple_spaces_after_comma` apply only for single-line arguments lists?

greg0ire:
> I'm not sure about leaving trailing spaces when there is a newline though.

It doesn't. What makes you think trailing spaces are left?

julienfalque:
I ran the following command on your branch:
```
$ ./php-cs-fixer describe method_argument_space
```
Here is the output I got for the code sample 5:
```
 * Example #5. Fixing with configuration: ['ensure_fully_multiline' => true, 'keep_multiple_spaces_after_comma' => true].
   --- Original
   +++ New
   @@ @@
    <?php
   -function sample($a=10,
   -    $b=20,$c=30) {}
   -sample(1,  
   -    2);
   +function sample(
   +    $a=10,
   +    $b=20,
   +    $c=30
   +) {}
   +sample(
   +    1,  
   +    2
   +);
    sample('foo',    'foobarbaz', 'baz');
    sample('foobar', 'bar',       'baz');
```

If you select the multiline `sample` function, you'll see that it still has two trailing spaces after the first argument. Anyway I just tried adding a similar test case on master branch and it failed as well. So I think the bug lies in `keep_multiple_spaces_after_comma` rather than in the changes introduced here.

julienfalque:
See #2827. I tried applying the patch to your branch and it fixed the issue as well. So the bug is not introduced by this PR and you don't need to worry about it :)

greg0ire:
a) good catch! And b) glad to read that!

greg0ire:
Rebased on your PR. Please review again <sup>(and <sup>merge <sup>?</sup></sup>)</sup>

SpacePossum:
Looking better and better, almost there!

For test coverage completeness, please add the following 2 cases, I would like to see what the fixer does :)

```php
<?php
$a = function#
(#
#
$a#
#
,#
#
$b,$c#
#
)#
use ($b,
$c,$d) {
};
```

```php
<?php
// no fix
list($a,
    $b, $c) = $a;
isset($a,
$b, $c);
unset($a,
$b, $c);
array(1,
    2,3
);
```

greg0ire:
@SpacePossum done. The first piece of code is unaffected, and I slightly changed the code so that the second isn't affected too, because of your `no fix` comment.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -588,6 +588,48 @@ class FooClass
     // do whatever
 }
 UNAFFECTED
+            ,
+                null,
+                ['ensure_fully_multiline' => true],
+            ],
+            'test ungodly code' => [
```
SpacePossum: love the label :)

greg0ire:
@SpacePossum should the first test be affected and if yes, what should be the output?

SpacePossum:
I theory;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440/files/7be5f5234f1b132dff8866c2db435378bea84c9c..582a8f0ad1ef6c0c5c37364951bbda2238105fb8#diff-c2d4cc2096dd614d141880c5036ee9b0R605

this one should've been made multi line I think?

(I don't care about multi line-ing the `use ()` part, its not part of PSR either and can be done later in other PR if someone really wants it (i.e. not out of `completeness` reasons ;) ))

greg0ire:
> this one should've been made multi line I think?

I think so too, not sure why it wasn't fixed...

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -588,6 +588,48 @@ class FooClass
...
+            <<<'UNAFFECTED'
+<?php
+$a = function#
+(#
+#
+$a#
+#
+,#
+#
+$b,$c#
```
keradus: it shall be split indeed

greg0ire: This is a closure @julienfalque , I didn't notice that.

greg0ire: Fixed and rebase on the latest master

greg0ire:
Please review again :pray:

keradus:
@greg0ire please guide me with answer so I wouldn't need to dig inside diff ;)

>  for multiline array arguments

Does fixer in this PR touch whitespaces in arrays or not?
If yes, does fixer touch them on master ?

greg0ire:
@keradus it doesn't, and proves it:

```
            'arrays with whitespace inside' => [
                <<<'UNAFFECTED'
<?php
$a = array/**/(  1);
$a = array/**/( 12,
7);
$a = array/***/(123,  7);
$a = array (        1,
2);
UNAFFECTED
            ,
                null,
```

>  If yes, does fixer touch them on master ?

N/A

keradus:
cool, please add it to tests ;) so it won't be accidentally changed in future

keradus:
Thank you @greg0ire.

julienfalque:
🤘

Thanks @greg0ire for this great feature!

SpacePossum:
thanks @greg0ire , lots of awesome work : D   :+1:

greg0ire:
Thank you so much for the numerous reviews and patient hints!

eclipxe13:
Congrats! Please, take a look at #2958

sustmi:
@greg0ire: Correct me if I am wrong, but I believe that by declaring the `ensure_fully_multiline` option as `true` by default you introduced a backward incompatible change and it should not have been released in v2.x according to Semantic Versioning.

We experienced that our code that was OK with friendsofphp/php-cs-fixer-2.3.2 and now it is not passing style-checks with v2.5.0. In my opinion such situations should not happen.

The right approach would be to add the new BC-breaking functionality but leave it disabled by default in v2.x and enable it by default in v3.x.

Anyway, thanks for the work! 👍 
(Please, do not take me wrong. I just wanted to let you know for future improvements.)

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440/files#diff-4cfdb64e15658046dbe0cda315ed9a99R132

default value for new option is **false** especially to keep BC.
there's a TODO to change default option to **true** on **3.0**, so when we could break BC.

sustmi:
It is `true` by default when you use `@PSR2` ruleset: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.5.0/src/RuleSet.php#L41

keradus:
`@PSR2` ruleset is reflecting PSR-2 document.
Whenever we do enhance our support for PSR-2 document, ruleset is updated.
In your scenario, you were:
- saying you follow PSR2
- used PSR2 ruleset
after update, it appear that you were not fully following PSR2 document

SpacePossum:
If you configured you're project to be `@PSR2` compliant we try to fix the project as good as we can to that standard. When we get better at it we release a new version and your project gets fixed better.
This BC compliant, we fixed the code to `PSR2` in the previous version and now we do a better job at it.

If you want to prevent this from happening you should configure the fixer with a custom configuration stating which fixers and their what configuration you want.

sustmi:
I guess that you are right. If the behaviour before this merge was not obeying PSR-2 then this is actually not a feature but a bug-fix. So it is probably OK that it was included in non-major release.

Thank you for the explanation.

keradus:
1. each bugfix is bcbreaker in the end
2. not everything is under SemVer. like rule desc is not keeped under bc promise, same with rulesets. single rules are

Issue #2442, opened 2017-01-01
no_alias_functions renamed user defined join() method to implode()


ilijastuden:
When `no_alias_functions` is included in the list of rules, it will rename user defined `join()` method. For example, there's intereface:

```php
<?php

interface JoinInterface
{
    public function &join();
}
```

and class that implements that interface.

PHP CS Fixer will rename both methods (signature from interface, and implementation from the concrete class) to `implode()`.

PHP 7.1, Code Style Fixer 2.0.

keradus:
@ilijastuden could you send a PR with testcases proving bug and, if possible, a fix for it ?

PR #2444, opened 2017-01-02
MbStrFunctionsFixer - handle return reference in method declaration

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2442


src/Fixer/Alias/NoAliasFunctionsFixer.php
```diff
@@ -52,6 +52,12 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
     {
         /** @var $token \PhpCsFixer\Tokenizer\Token */
         foreach ($tokens->findGivenKind(T_STRING) as $index => $token) {
+            // check mapping hit
+            $tokenContent = strtolower($token->getContent());
```
SpacePossum: this move is for code execution speed only

keradus:
Thank you @SpacePossum.

PR #2212, opened 2016-10-05
add no_yoda_contition to contrib fixers

niklongstone:
No Yoda condition for ==, !=, ===, and !==

example:
Given: `1 === $mayTheForceBeWithYou`
The result should be: `$mayTheForceBeWithYou === 1`

PR #701, opened 2014-11-07
Add simplified YodaConditionsFixer

keradus:
See:
- #700 for temporary remove YodaConditionsFixer
- #693 for discuss what need to be changed in fixer and discussion
- #698 for some little fixes

PR #2446, opened 2017-01-03
Add YodaStyleFixer

SpacePossum:
Be formatted in yoda style (or not), code should.

This fixer can be configured to:
- make code Yoda style, or:
- make code no longer Yoda style, or:
- don't run the fixer if you don't care

Fixer can be configured to change compare statements:
- `equal`: `==` and `!==` (`<>`)
- `identical`: `===` and `!==`
- `less_and_greater`: `<=`, `>=`, `<` and `>`

You can combine any of the operator groups, if you don't like any of the groups don't run the fixer.

Default:
Fix to Yoda style for operators `equal` and `identical`.

Replaces;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2212
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/701


tests/Fixer/ControlStructure/YodaStyleFixerTest.php
```diff
@@ -0,0 +1,266 @@
...
+                '<?php return "" === $this->myObject->{$index};',
+                '<?php return $this->myObject->{$index} === "";',
+            ),
+            array(
+                '<?php return "" === $this->myObject->{$index}->a;',
+                '<?php return $this->myObject->{$index}->a === "";',
+            ),
+            array(
+                '<?php return self::MY_CONST === self::$myVariable;',
+                '<?php return self::$myVariable === self::MY_CONST;',
```
greg0ire: A constant is… constant. You can't assign it by mistake, because [you simply can't assign it](https://3v4l.org/GSj1T), so Yoda is useless here. Shouldn't be fixed IMO.

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
```diff
@@ -0,0 +1,266 @@
...
+        } else {
+            $this->doTest($input, $expected);
+        }
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            // Don't fix cases.
+            array('<?php $a = 1 === 1;'),
```
greg0ire: Is the `$a =` part necessary?

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
```diff
@@ -0,0 +1,266 @@
...
+        $config['yoda'] = false;
+        $this->fixer->configure($config);
+        $this->doTest($input, $expected);
+    }
+
+    public function provideLessGreaterCases()
+    {
+        return array(
+            array(
+                '<?php $a = 3 <= $b;',
```
greg0ire: Same here, see https://github.com/symfony/symfony-docs/pull/5381 (and disclaimer, I might be a Sith lord)

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
SpacePossum: stricktly no, but as a test it has value IMO

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
SpacePossum: How is this the same as what? i.e. assignment of `$a` is based on var. `$b`

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
SpacePossum: Not sure I see the issue, you don't have run the fixer if you don't want it to be changed to yoda,
or you can run it with the configuration option `'yoda' => false` to make it non-yoda.

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
greg0ire: Same as [the second comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2446#discussion_r94458043) : Yoda shields you from nothing, does it?

SpacePossum: thats not the point, you get the option to make it yoda or not-yoda (or not run the fixer).

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
greg0ire: The issue is that Yoda is supposed to shield you from doing an assignment when you don't need to. But here, since you will get an error message as show in my previous comment, it is not worth it, which means it might be better to use "natural" (at least for english people) order. We're not from Dagobah after all. So to sum up, Yoda should only apply IMO to expressions like `something == 42` (or even `$x === 42`, why not), if and only if `something = 42` does not throw an error message.

SpacePossum: all rules you describe here can be configured on this fixer so it fixes the code as you want, so I see no issue

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
greg0ire: > you can even config the fixer to change only some statements so exclude <, >, <=, => if you want

That's a great feature, I missed that. I personally think that Yoda is a bad practice, but if I were forced to use it in a project where I contribute (like the sonata libs), I would try to have it configured with `==` only (would be great to have this as the default). So nevermind this comment.

> CS is not always/only about logic, sometimes it really is just style

Yeah but in this particular case, no one uses that because it looks good IMO, it's just to spare accidental mistakes (so benefits the writer, but not the reader), and is pretty useless if you have unit tests.

greg0ire: > would be great to have this as the default

If I read the code correctly it seems to be :)

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
greg0ire: If by that you mean that there are some people who are going to want Yoda even when it makes no sense (when none of the operands are assignable, in this particular case), but just because they prefer to have Yoda everywhere because they like it like that then indeed, there is no issue.

greg0ire:
Nice branch name ^^

Slamdunk:
👍 for the fixer, and 👍 for leaving `less_and_greater` out of the default config

SpacePossum:
What is left here is to decide to add more configuration options for setting which kind of comparisons should be fixed. Based on previous discussions it seems some people would like to target no-vars like `(array(2))` and `self::MY_CONST` while other don't. Like to get some feedback so this fixer can finished :)

orlangur:
@SpacePossum thanks for pointing to this PR, I tried to search for "yoda" and "cond" with no luck and thought `YodaConditionFixer` was renamed to something less predictable.

Why Yoda-style is default one in AutomateEverything era? Most of experienced programmers have got some professional deformation and read `null === $something` better than `$something === null`?

SpacePossum:
The rationale behind the defaults in this project (and this PR) are based on:
- current CS in the project
- expected most used values

I added the options to prevent discussion about what is best and why. With the options everyone can use the fixer as wanted. So lets see what options are missing, what bugs are left in this PR or where other improvements can be made here.

keradus:
Please rebase

hkdobrev:
Any updates on this? Should just conflicts and build be fixed or is there more to this? Thank you!

SpacePossum:
rebased
@hkdobrev there are a few cases still failing (related to array destruction) and some tests are missing for new PHP syntax features (fixer was written before PHP 7 launched).
Please let me know if you can help out finishing this :)  (or anybody else, all help is welcome here)

keradus:
Sadly, tests are failing

SpacePossum:
all green... cause... the test that was failing was because the `expected` and `output` were swapped...
: D

Anyway, what is left is test cases for all the new syntax stuff PHP7 and up provided.
Therefor I would ask all people here to come up with nice test cases :)

SpacePossum:
TODO: test on mulitple repo's to see if all good

keradus:
@SpacePossum `Symfony` follows yoda, could you please test it on that repo?

SpacePossum:
https://github.com/SpacePossum/symfony/pull/1
:tada:

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -0,0 +1,609 @@
...
+            }
+
+            if ($current->isGivenKind([T_LIST, CT::T_DESTRUCTURING_SQUARE_BRACE_OPEN, CT::T_DESTRUCTURING_SQUARE_BRACE_CLOSE])) {
+                return false;
+            }
+
+            $next = $tokens[$tokens->getNextMeaningfulToken($index)];
+
+            // self:: or ClassName::
+            if ($current->isGivenKind(T_STRING) && $next->isGivenKind(T_DOUBLE_COLON)) {
```
julienfalque: In these `if` statements you check some expected token kinds and increment `$index` with hardcoded values, assuming the next token is at this position, which is not always true (you provided cases in tests). Wouldn't it be better to store the actual position of the next meaningful token and reuse it to increment `$index`?

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -0,0 +1,609 @@
...
+                ++$index;
+                $expectString = true;
+
+                continue;
+            }
+
+            return true;
+        }
+
+        return true; // unreachable statement ATM, kept for SCA and support for PHP changes in the future.
```
julienfalque: Why not replacing the other `return true;` statement above with `break;`?

src/Fixer/ControlStructure/YodaStyleFixer.php
SpacePossum: this is a direct copy of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/701/files#diff-fe2f2514a225293636fa7d1795d7a67dR153 (that is why there are some many authors in the fixer listed),
I always read it as, we found start of a statement that we know is at least X tokens, so lets skip the known amount, from there we check again in the loop. In the end the result should always be the same.

SpacePossum: I think I've updated following all your other suggestions.

tests/Fixer/ControlStructure/YodaStyleFixerTest.php
```diff
@@ -0,0 +1,493 @@
...
+            [
+                '<?php return "" === $this->myArray[$index]->a;',
+                '<?php return $this->myArray[$index]->a === "";',
+            ],
+            [
+                '<?php return "" === $this->myObject->  {$index};',
+                '<?php return $this->myObject->  {$index} === "";',
+            ],
+            [
+                '<?php return "" === $this->myObject->{$index}->a;',
```
julienfalque: Please add a similar test case without the braces around the dynamic property, e.g.:
```php
<?php return $this->myObject->$index->a === "";
```

SpacePossum:
I found a bug :(
related to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2446#discussion_r136873586
working on it

SpacePossum:
all done.. I hope

SpacePossum:
both fabbot and Travis failures are false-negatives, sadly

keradus:
wow, now we get broken pipe on 5.6? wtf?

keradus:
Thank you @SpacePossum.

localheinz:
Nice work!

kubawerlos:
@keradus shouldn't this fixer be part of Symfony rule set?

keradus:
@kubawerlos , yoda style is part of Symfony styleguide indeed. Yet, `yoda` could be understand very differently, that is why we have configuration options here (yet not all different understanding of `yoda` is covered by options, there are many edge understanding of this rule).

Shall this fixer (configured somehow) be part of Symfony ruleset? Yes, as soon as Symfony as a group would be willing to have it - either directly (by sending a PR that would change a ruleset) or indirectly (by accepting https://github.com/symfony/symfony/pull/24123 )

SteveJobzniak:
I don't understand the configuration options... 

```
yoda_style [@Symfony]

Write conditions in Yoda style or not based on configuration.

Configuration options:

equal (bool, null): change equal (==, !=) statements; defaults to true
identical (bool, null): change identical (===, !==) statements; defaults to true
less_and_greater (bool, null): change less and greater than (<, <=, >, >=) statements; defaults to null
```

They all say: "Change X (if true) or not (if false)". Apparently that's not right. According to the 1st post in this ticket, it does something like:

true = Force to yoda style.
false = Force to non-yoda style.
null = Do nothing.

The docs need a little rewrite, perhaps?

keradus:
Hi.
Feel free to raise PR with description improvement :+1:

SteveJobzniak:
@keradus Hi. Alright! https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3070 :-)

devdrops:
Hi!

If this PR is closed, why it's mentioned [as part of the latest release](https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.6.0) instead of #3070?

I mean, I got confused reading the release's description :dizzy_face:

hkdobrev:
@devdrops it was merged manually via https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/73fddc63275df7298d409a2e957c8fd688af1227.

PR #2447, opened 2017-01-04
.php_cs.dist - update local CS config

keradus:


.php_cs.dist
```diff
@@ -27,6 +27,7 @@ return PhpCsFixer\Config::create()
         'no_useless_return' => true,
         'ordered_class_elements' => true,
         'ordered_imports' => true,
+        'phpdoc_order' => true,
```
localheinz: Do you want to keep rules sorted here?

.php_cs.dist
```diff
@@ -29,6 +29,7 @@ return PhpCsFixer\Config::create()
         'ordered_imports' => true,
         'php_unit_strict' => true,
         'phpdoc_add_missing_param_annotation' => true,
+        'phpdoc_order' => true,
```
localheinz: :+1:

PR #2448, opened 2017-01-05
[WIP][PHP 7.1] Add ListSyntaxFixer

SpacePossum:
question, should we add a new custom token for `[` (and its counter part `]`) of the [new short syntax](https://wiki.php.net/rfc/short_list_syntax) for array destructing assignment?
(I think so btw. ;) )


keradus:
like https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Tokenizer/Transformer/SquareBraceTransformer.php#L25 ?

SpacePossum:
oh yeah like that, haha oh man can someone bring me coffee :)  thanks!

PR #2452, opened 2017-01-06
Provide rules definitions

keradus:


SpacePossum:
:+1: looking good!

src/Fixer/ArrayNotation/ArraySyntaxFixer.php
```diff
@@ -107,7 +107,7 @@ public function getDefinition()
                 ),
             ),
             null,
-            'Configure to use "long" or "short" array declaration syntax.',
+            'The following can be configured: `only_untyped => "long"|"short"`',
```
SpacePossum: this is not correct I think
(ping @keradus )

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2466

PR #2456, opened 2017-01-10
Introduce @PhpCsFixer ruleset

keradus:


keradus:
does it make sense ?

GrahamCampbell:
Err, no?

GrahamCampbell:
The ruleset of this tool is not one other people are going to want to use I imagine?

SpacePossum:
looking at my own projects (I use with just a few minor changes) I'm +1,
but I'm not sure about how much others use it.

keradus:
I also use it in every single project.
Our ruleset is basically: Symfony hard rules (so `@Symfony`), Symfony soft rules (applied only in new codebase, not in whole repo, due to merging issues) and strict fixers

SpacePossum:
please rebase and :ship: :)

teohhanhui:
This will only create more confusion for the users. :smile:

julienfalque:
I'm not sure I want this feature.

The name suggests that the ruleset is internal, or at least it does not really give a clue about the rules one can expect from it.

It enables the `array_syntax` rule with `long` syntax while most recent projects will use the `short` syntax so they will still need to override this rule.

Creating this ruleset will require to maintain BC on it afterwards. This mean we won't be able to change anything in it except adding new rules. So if we want to change the way we fix the project itself, we will have to override some rules in our config file instead of tweaking the ruleset, defeating its purpose.

SpacePossum:
@keradus sorry man, gonna close this one after discussion. Lets see if we can get up with another way to make it easy to share this ruleset as I do like/use it : )

PR #2459, opened 2017-01-12
BracesFixer – Add option for keeping opening brackets on the same line

jtojnar:
Initially, I intended to add separate options for classy constructs and functions but since most people probably want the same setting for both and having two options would mean need for even more tests, I have decided to stay with single config value.

### DIFF
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -434,7 +434,7 @@ class Foo
             }
 
             if ($token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)) {
-                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
+                if (self::LINE_SAME === $this->configuration['position_after_functions_and_oop_constructs'] && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
                     $ensuredWhitespace = ' ';
                 } else {
                     $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
@@ -453,7 +453,7 @@ class Foo
                         $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
                     }
                 } else {
-                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
+                    if (self::LINE_SAME === $this->configuration['position_after_functions_and_oop_constructs'] && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
                         $ensuredWhitespace = ' ';
                     } else {
                         $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -31,8 +31,14 @@ use PhpCsFixer\Tokenizer\TokensAnalyzer;
  */
 final class BracesFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
+    /**
+     * @internal
+     */
     const NEXT_LINE = 'next';
 
+    /**
+     * @internal
+     */
     const SAME_LINE = 'same';
 
     /**
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -434,10 +434,10 @@ class Foo
             }
 
             if ($token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)) {
-                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
-                    $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
-                } else {
+                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
                     $ensuredWhitespace = ' ';
+                } else {
+                    $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
                 }
 
                 $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, $ensuredWhitespace);
@@ -453,10 +453,10 @@ class Foo
                         $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
                     }
                 } else {
-                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
-                        $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
-                    } else {
+                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
                         $ensuredWhitespace = ' ';
+                    } else {
+                        $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
                     }
 
                     $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, $ensuredWhitespace);
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -34,12 +34,12 @@ final class BracesFixer extends AbstractFixer implements ConfigurationDefinition
     /**
      * @internal
      */
-    const NEXT_LINE = 'next';
+    const LINE_NEXT = 'next';
 
     /**
      * @internal
      */
-    const SAME_LINE = 'same';
+    const LINE_SAME = 'same';
 
     /**
      * {@inheritdoc}
@@ -127,7 +127,7 @@ class Foo
     }
 }
 ',
-                    array('position_after_functions_and_oop_constructs' => self::SAME_LINE)
+                    array('position_after_functions_and_oop_constructs' => self::LINE_SAME)
                 ),
             )
         );
@@ -164,8 +164,8 @@ class Foo
 
         $positionAfterFunctionsAndOopConstructs = new FixerOptionBuilder('position_after_functions_and_oop_constructs', 'whether the opening brace should be placed on "next" or "same" line after classy constructs (non-anonymous classes, interfaces, traits, methods and non-lambda functions).');
         $positionAfterFunctionsAndOopConstructs = $positionAfterFunctionsAndOopConstructs
-            ->setAllowedValues(array(self::NEXT_LINE, self::SAME_LINE))
-            ->setDefault(self::NEXT_LINE)
+            ->setAllowedValues(array(self::LINE_NEXT, self::LINE_SAME))
+            ->setDefault(self::LINE_NEXT)
             ->getOption()
         ;
 
@@ -434,9 +434,9 @@ class Foo
             }
 
             if ($token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)) {
-                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::NEXT_LINE) {
+                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
                     $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
-                } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::SAME_LINE) {
+                } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME) {
                     $ensuredWhitespace = ' ';
                 }
 
@@ -453,9 +453,9 @@ class Foo
                         $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
                     }
                 } else {
-                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::NEXT_LINE) {
+                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
                         $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
-                    } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::SAME_LINE) {
+                    } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME) {
                         $ensuredWhitespace = ' ';
                     }
 
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -31,6 +31,10 @@ use PhpCsFixer\Tokenizer\TokensAnalyzer;
  */
 final class BracesFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
+    const NEXT_LINE = 'next';
+
+    const SAME_LINE = 'same';
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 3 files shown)


jtojnar:
AppVeyor seems to be broken.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -26,9 +28,45 @@
...
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            $this->config = self::$defaultConfiguration;
+
+            return;
+        }
+
+        if (array_key_exists('classy_constructs', $configuration)) {
+            if (!in_array($configuration['classy_constructs'], array('same', 'next'), true)) {
```
GrahamCampbell: probably best to define constants for next and same, rather than using strings

jtojnar: It would be better but none of the other fixers use constants.

GrahamCampbell: > It would be better but none of the other fixers use constants.

Yes they do?

GrahamCampbell: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Import/OrderedImportsFixer.php#L35-L43

jtojnar: At least [ConcatSpace](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Operator/ConcatSpaceFixer.php#L33) and [ArraySyntax](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/ArrayNotation/ArraySyntaxFixer.php#L38) do not. But I agree it is better and will add it.

jtojnar:
Constants added and rebased.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -30,23 +30,32 @@
...
-     * @param array<string, bool>|null $configuration
+     * @param array|null $configuration
+     *
+     * @throws InvalidFixerConfigurationException
      */
     public function configure(array $configuration = null)
     {
-        if (null === $configuration) {
-            $this->configuration = self::$defaultConfiguration;
+        $this->configuration = self::$defaultConfiguration;
```
keradus: no. as docs stated, null will reset the configuration, array will override it, not patch it on top of default values.

jtojnar: [Oh](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f3baf72eb2f58bf275b372540f5b47d25aed910f/src/Fixer/ConfigurableFixerInterface.php), I did not see that. I will change it, though I find this behaviour much more convenient than having to specify all or nothing. Maybe in the future it could be changed.

keradus: ok. better.
you could patch default config (we usually do it like 
`$configuration = array_merge(self::$defaultConfiguration, $configuration);`
), but not current one.

so, if default is `[foo => 1, bar => 2]`. then
```
$f->configure([foo => 3]); // configured as [foo => 3, bar => 2]
$f->configure([bar => 4]); // configured as [foo => 1, bar => 4], not as [foo => 3, bar => 4]

Comment (jtojnar) [inline src/Fixer/Basic/BracesFixer.php]: That makes sense, updated.

File: src/Fixer/Basic/BracesFixer.php [hunk 3]
```diff
@@ -55,12 +64,17 @@ public function configure(array $configuration = null)
                 throw new InvalidFixerConfigurationException($this->getName(), sprintf('"%s" is not handled by the fixer.', $functionName));
             }
 
-            if (!is_bool($replacement)) {
+            if ('allow_single_line_closure' === $functionName && !is_bool($replacement)) {
                 throw new InvalidFixerConfigurationException($this->getName(), sprintf('Expected bool got "%s".', is_object($replacement) ? get_class($replacement) : gettype($replacement)));
+            } elseif ('classy_constructs' === $functionName && !in_array($replacement, $this->supportedBracePositions, true)) {
```
keradus: no need for `else` after `throw`

jtojnar: Fixed

src/Fixer/Basic/BracesFixer.php
```diff
@@ -66,7 +66,9 @@ public function configure(array $configuration = null)
 
             if ('allow_single_line_closure' === $functionName && !is_bool($replacement)) {
                 throw new InvalidFixerConfigurationException($this->getName(), sprintf('Expected bool got "%s".', is_object($replacement) ? get_class($replacement) : gettype($replacement)));
-            } elseif ('classy_constructs' === $functionName && !in_array($replacement, $this->supportedBracePositions, true)) {
+            }
+
+            if ('classy_constructs' === $functionName && !in_array($replacement, $this->supportedBracePositions, true)) {
```
keradus: For some weird reason keys was named wrongly.
As you are touching the could, could you rename `functionName` and `replacement` vars please ?

jtojnar: Done.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -30,17 +30,26 @@
...
+    private $supportedBracePositions = array(self::NEXT_LINE, self::SAME_LINE);
+
     /**
      * @var array
      */
     private $configuration;
 
     private static $defaultConfiguration = array(
         'allow_single_line_closure' => false,
+        'classy_constructs' => self::NEXT_LINE,
```
keradus: what about methods ?
what about control statements like `if`, `do` etc ?

jtojnar: Methods are affected, I could not come up with a better name, hopefully the documentation will clear it up. Control statements are not affected. According to [PSR-2 survery](http://www.php-fig.org/psr/psr-2/#a3-survey-results), majority of use cases will be covered by this configuration option anyway.

jtojnar: Would `brace_after_oop_constructs` be a better name?

keradus: new braces for `if` statement is not `oop constructor`

jtojnar: As I wrote before, this does not cover `if` statements. See the documentation: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2459/files#diff-0b7fa2dc2594eae7a18836044e3287d8R172

Though it does affect non-anonymous functions since they are handled by the same code as methods. I cannot come up with a better hypernym, though.

jtojnar: Does `position_after_functions_and_oop_constructs` sound better?

keradus:
1 month old PR. I would need some comments here from community ! ;)

eduardodgarciac:
Hi, when do you will add this feature?

keradus:
the voice of community is missing here :(

src/Fixer/Basic/BracesFixer.php
```diff
@@ -30,17 +30,26 @@
...
+    private $supportedBracePositions = array(self::NEXT_LINE, self::SAME_LINE);
+
     /**
      * @var array
      */
     private $configuration;
 
     private static $defaultConfiguration = array(
         'allow_single_line_closure' => false,
+        'position_after_functions_and_oop_constructs' => self::NEXT_LINE,
```
julienfalque: I find the option name very unclear. And as it only accepts two values, I would rather make it a boolean option.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -30,17 +30,26 @@
...
     );
 
     /**
-     * @param array<string, bool>|null $configuration
+     * @param array|null $configuration
```
julienfalque: This PHPDoc doesn't need to be changed in this PR IMO.

src/Fixer/Basic/BracesFixer.php
jtojnar: Well, there are three options:
* Disabled
* Enforce same line
* Enforce next line

src/Fixer/Basic/BracesFixer.php
SpacePossum: because of the the new config option it changed to
```php
/** @param array<string, string|bool> $configuration */
```

so it needs to be updated to something ;)

julienfalque: Ah you're right, sorry.

src/Fixer/Basic/BracesFixer.php
SpacePossum: how to use `disabled`? the key in the default configuration which is merged with given configuration, for given configuration only `self::NEXT_LINE` and `self::SAME_LINE` are allowed?

julienfalque: How do you pass the "disabled" value to the option? As far as I understand, the option value must be either `next` or `same`, and defaults to `next` when not passed.

tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -22,14 +22,31 @@
  */
 final class BracesFixerTest extends AbstractFixerTestCase
 {
+    private static $configurationOopPositionSameLine = array('position_after_functions_and_oop_constructs' => 'same');
+
+    public function testInvalidConfigurationClassyConstructs()
+    {
+        $this->setExpectedExceptionRegExp(
+            'PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException',
+            '#^\[braces\] Position of the opening brace is invalid. Should be one of: "next", "same".$#'
```
SpacePossum: I think the dots need escaping here?

src/Fixer/Basic/BracesFixer.php
jtojnar: Sorry, I was thinking about the fixer having `string $configuration` but since there is more options now, it is not actually possible anymore.

tests/Fixer/Basic/BracesFixerTest.php
jtojnar: Fixed.

src/Fixer/Basic/BracesFixer.php
jtojnar: Regarding the boolean option, [other](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1fe4e61f89f48313ae229761c0da915acb2bea78/src/Fixer/Whitespace/NoSpacesAroundOffsetFixer.php#L39) [fixers](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/bf0f42981f8a89708ed9e16c33cb3a876d146f3d/src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php#L53) [seem](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/7651b56fab166dd9adf468886f07fdae97384ec7/src/Fixer/LanguageConstruct/DeclareEqualNormalizeFixer.php#L49) [to](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f6101b0b733772fdf283c573591433a6828b3fc0/src/Fixer/Alias/NoMixedEchoPrintFixer.php#L52) [use](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ec4b981a72075f10d2665c0ede7666f81df2b443/src/Fixer/Import/OrderedImportsFixer.php#L72) strings for binary choices as well. The booleans might just be confusing – the opposite of enforcing a brace placement is not-enforcing it, not enforcing other placement.

I address splitting the option in the OP, almost nobody wants to control the position of the class braces individually (according the [PSR-2 survey](www.php-fig.org/psr/psr-2/#a3-survey-results)) and it would only complicate the whole thing further.

julienfalque: I suggested to use a boolean option because your current implementation seems to allow two forced positions for the braces but not to leave them untouched. But if you actually want to support disabling this as well, then a string option makes sense, altought I think another fixer has a similar option and uses `true|false|null`.

jtojnar: Yes, but for me boolean implies on – off behaviour. This is more of a A – B situation.

mstanglx:
This is definitely an option our team is looking forward to having in this tool! Can't wait.

allejo:
definitely something I'd use 👍

keradus:
Anyone willing to make a review ?

src/Fixer/Basic/BracesFixer.php
```diff
@@ -403,7 +456,13 @@ function ($item) {
             }
 
             if ($token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)) {
-                $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, $this->whitespacesConfig->getLineEnding().$indent);
+                if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
+                    $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
+                } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME) {
```
julienfalque: The option accepts only two values, so I would use a simple `else` instead of `elseif` here.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -416,7 +475,13 @@ function ($item) {
                         $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
                     }
                 } else {
-                    $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, $this->whitespacesConfig->getLineEnding().$indent);
+                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_NEXT) {
+                        $ensuredWhitespace = $this->whitespacesConfig->getLineEnding().$indent;
+                    } elseif ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME) {
```
julienfalque: Same here.

SpacePossum:
can you add this case to tests so I it is clear how comments are handled?

```php
<?php
class Something # a
{
    public function sth(): string //        
    {
        return function (int $foo) use ($bar): string { return $bar; };
    }
}

function C /**/ //    # /**/
{
}

function C /** 
*
*/
{
}

```

jtojnar:
Fixed.

jtojnar:
@SpacePossum Your test case is not valid PHP code.

SpacePossum:
updated here https://3v4l.org/uCQQI  , thanks!

jtojnar:
@SpacePossum Are the trailing spaces intentional?

SpacePossum:
yes, they are :)

src/Fixer/Basic/BracesFixer.php
```diff
@@ -416,7 +475,13 @@ function ($item) {
                         $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
                     }
                 } else {
-                    $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, $this->whitespacesConfig->getLineEnding().$indent);
+                    if ($this->configuration['position_after_functions_and_oop_constructs'] === self::LINE_SAME && !$tokens[$startBraceIndex - 2]->isGivenKind(T_COMMENT)) {
```
keradus: With you Yoda should be.

jtojnar: Fixed.

mstanglx:
Hey all, good job on this. Any ides when it will get merged to the main branch?
Thanks.

SpacePossum:
@mstanglx There is one known issue with this PR, after that I think we are close to merging. Stay put :)

jtojnar:
@SpacePossum I already did the changes suggested in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2459#issuecomment-285203911 in c16c01d. Also, the trailing whitespace is not important, since it is a [part of the comment](https://3v4l.org/dJuIL).

allejo:
Any updates?

SpacePossum:
@jtojnar please consider https://github.com/jtojnar/PHP-CS-Fixer/pull/1
and please rebase :)

jtojnar:
Rebased, though CI is frozen and I cannot test locally.

jtojnar:
Rebased.

SpacePossum:
one last detail:
```
if (isset($configuration)) {
```
to
```
if (null !== $configuration) {
```

in the unit tests.
After that this is ready to merge IMO :)

keradus:
Thank you @jtojnar.

SpacePossum:
thanks @jtojnar , nice new option!

mstanglx:
Thank you @jtojnar and @SpacePossum !!!

Issue #2420, opened 2016-12-22
Consider fixer to qualify global functions


bendavies:
See http://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/ for background and benefit.

Probably quite difficult considering a user could have overwritten global functions in the current namespace.

SpacePossum:
Finding candidates can be done using something like the [NativeFunctionCasingFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Casing/NativeFunctionCasingFixer.php) does.
But yes, it will be hard (if not impossible) to see if a candidate is a call to function with the same name in a different namespace by looking at the tokens alone.

keradus:
I wonder what would be better - add `\` or `import` statement.

Still, if one will override the function in same namespace but different file, we would break the code

localheinz:
@bendavies 

Take a look at https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2462!

@keradus 

>I wonder what would be better - add `\` or import statement.

Probably depends on whether people like a ton of imports!

> Still, if one will override the function in same namespace but different file, we would break the code

Declared the fixer in the PR as risky, hehe!

bendavies:
Implemented!

PR #2462, opened 2017-01-13
Add NativeFunctionInvocationFixer

localheinz:
This PR

* [x] implements a configurable`NativeFunctionInvocationFixer`

Related to #2420.

💁‍♂️ For reference, see 

* https://twitter.com/Ocramius/status/811504929357660160
* http://php.net/manual/en/language.namespaces.faq.php#language.namespaces.faq.shortname2
* http://veewee.github.io/blog/optimizing-php-performance-by-fq-function-calls/

>Function or constant names that do not contain a backslash like name can be resolved in 2 different ways.
>
>First, the current namespace name is prepended to _name_.
>
>Finally, if the constant or function _name_ does not exist in the current namespace, a global constant or function _name_ is used if it exists.

### DIFF
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -141,7 +141,7 @@ final class NativeFunctionInvocationFixer extends AbstractFixer implements Confi
      */
     public function getDefinition()
     {
-        $riskyDescription = <<<TXT
+        $riskyDescription = <<<'TXT'
 Rule is risky when a function with the same name as a native function exists in the current namespace.
 +Two major situations when it could happen are:
 +* polyfill is used to provide function from higher PHP version, if PHP CS Fixer would been run also on that higher PHP version, but software is also desired to work on lower PHP version that naively doesn't have that polyfilled function
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -185,7 +185,7 @@ function baz($options)
                 ),
             ),
             null,
-            'Configure names of functions to exclude, for example, when using a polyfill.',
+            'Configure names of functions to exclude, for example, when mocking.',
             self::$defaultConfiguration,
             $riskyDescription
         );
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -88,50 +88,10 @@ final class NativeFunctionInvocationFixer extends AbstractFixer implements Confi
 
         $indexes = array();
 
-        $openBracesCount = 0;
-        $isNamespaceWithBraces = false;
-        $isWithinNamespace = false;
-
         for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
             $token = $tokens[$index];
 
             $tokenContent = $token->getContent();
-
-            if ('{' === $tokenContent) {
-                ++$openBracesCount;
-
-                continue;
-            }
-
-            if ('}' === $tokenContent) {
-                --$openBracesCount;
-
-                if (true === $isWithinNamespace
-                    && true === $isNamespaceWithBraces
-                    && 0 === $openBracesCount
-                ) {
-                    $isWithinNamespace = false;
-                }
-
-                continue;
-            }
-
-            if ($token->isGivenKind(T_NAMESPACE)) {
-                $namespaceDeclarationEndIndex = $tokens->getNextTokenOfKind($index, array(
-                    ';',
-                    '{',
-                ));
-
-                if ($tokens[$namespaceDeclarationEndIndex]->getContent() === '{') {
-                    $isNamespaceWithBraces = true;
-                }
-
-                $isWithinNamespace = true;
-            }
-
-            if (!$isWithinNamespace) {
-                continue;
-            }
 
             // test if we are at a function call
             if (!$token->isGivenKind(T_STRING)) {
@@ -182,42 +142,30 @@ final class NativeFunctionInvocationFixer extends AbstractFixer implements Confi
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Add leading `\` before function invocation of internal function within namespaces to speed up resolving.',
+            'Add leading `\` before function invocation of internal function to speed up resolving.',
             array(
                 new CodeSample(
 '<?php
 
-namespace Foo;
-
-class Bar
+function baz($options) 
 {
-    public function baz($options) 
-    {
-        if (!array_key_exists("foo", $options)) {
-            throw new \InvalidArgumentException();
-        }
-        
-        return json_encode($options);
+    if (!array_key_exists("foo", $options)) {
+        throw new \InvalidArgumentException();
     }
     
+    return json_encode($options);
 }'
                 ),
                 new CodeSample(
 '<?php
 
-namespace Foo;
-
-class Bar
+function baz($options) 
 {
-    public function baz($options) 
-    {
-        if (!array_key_exists("foo", $options)) {
-            throw new \InvalidArgumentException();
-        }
-        
-        return json_encode($options);
+    if (!array_key_exists("foo", $options)) {
+        throw new \InvalidArgumentException();
     }
     
+    return json_encode($options);
 }',
                     array(
                         'exclude' => array(
@@ -238,10 +186,7 @@ class Bar
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isAllTokenKindsFound(array(
-            T_NAMESPACE,
-            T_STRING,
-        ));
+        return $tokens->isTokenKindFound(T_STRING);
     }
 
     /**
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -141,6 +141,14 @@ final class NativeFunctionInvocationFixer extends AbstractFixer implements Confi
      */
     public function getDefinition()
     {
+        $riskyDescription = <<<TXT
+Rule is risky when a function with the same name as a native function exists in the current namespace.
++Two major situations when it could happen are:
++* polyfill is used to provide function from higher PHP version, if PHP CS Fixer would been run also on that higher PHP version, but software is also desired to work on lower PHP version that naively doesn't have that polyfilled function
++* function is mocked during tests execution, eg mocking `time` function - in that case after applying the rule src code will always use original, unmocked function
++To deal with described situation provide a configuration with function names you want to preserve unchanged.'
+TXT;
+
         return new FixerDefinition(
             'Add leading `\` before function invocation of internal function to speed up resolving.',
             array(
@@ -177,7 +185,7 @@ function baz($options)
             null,
             'Configure names of functions to exclude, for example, when using a polyfill.',
             self::$defaultConfiguration,
-            'Risky if a function with the same name as a native function exists in the current namespace, or a polyfill is used.'
+            $riskyDescription
         );
     }
 
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -233,7 +233,7 @@ class Bar
                 ),
             ),
             null,
-            'Configure names of functions to exclude',
+            'Configure names of functions to exclude, for example, when using a polyfill.',
             self::$defaultConfiguration,
             'Risky if a function with the same name as a native function exists in the current namespace, or a polyfill is used.'
         );
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -154,7 +154,7 @@ class Bar
             null,
             null,
             null,
-            'Risky if a function with the same name as a native function exists in the current namespace'
+            'Risky if a function with the same name as a native function exists in the current namespace, or a polyfill is used.'
         );
     }
 
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -28,7 +28,7 @@ final class NativeFunctionInvocationFixer extends AbstractFixer implements Confi
     /**
      * @var string[]
      */
-    private $exclude = array();
+    private $configuration = array();
 
     /**
      * @var array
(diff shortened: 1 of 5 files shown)


localheinz:
Because I'm lazy, I tried replacing

As a heads-up, with some experimenting, I used PhpStorm regular expressions to find and replace, replacing

```
(^\s+|return\s|,\s|=\s|=>\s|\?\s|:\s|!|\()(?!(array|empty|exit|do|for|foreach|if|while|isset|list|unset|__))(\w+)\(
```

with

```
$1\\$3\(
```

keradus:
In general, I am for the change, but not in this form. It's hard to maintain and keep it consistent.

Instead, I propose:
1. PR with a rule to add that `\` without dummy regexes that will fail on edge cases
2. then apply it on repo
3. and keep rule running under CI

localheinz:
@keradus 

Sounds good, let me take a look at it!

localheinz:
Actually missing quite a bunch of functions.

SpacePossum:
did someone compare if with the import of these function (`use function \array_keys;` etc.) also gives the same benefits?

localheinz:
@keradus @SpacePossum 

How about this one?

keradus:
@SpacePossum it shall, as engine will, as well as with `echo \implode($x)`, do function lookup, but it will take it from import

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,191 @@
...
+            throw new \InvalidArgumentException();
+        }
+    }
+    
+}'
+            )),
+            null,
+            null,
+            null,
+            'Risky if a function with the same name as a native function exists in the current namespace'
```
keradus: While I find this fixer really useful, it's not only case when it't risky. And I would say that the 2nd case is more tricky.

Let me use `is_iterable` function:
```php
var_dump(is_iterable($foo));
```

Then run the fixer on 7.1 env:
```php
var_dump(\is_iterable($foo));
```

Then run the code on 5.6 env - it will fail, because of missing function.
And if I would run original code - it will succeed, as missing function is polyfilled via `symfony/polyfill`.

Please describe a risk about defined function in given env and polyfilled functions.
Also, please add a configuration that I do not want to be prefixed with `\`, so I could set `is_iterable` as exception and still use the rule

localheinz:
@keradus 

Updated! 🤓

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,270 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class NativeFunctionInvocationFixer extends AbstractFixer implements ConfigurableFixerInterface
+{
+    /**
+     * @var array
```
keradus: array of ?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,270 @@
...
+        'exclude' => array(),
+    );
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            return;
```
keradus: this should reset `$exclude` to original, empty state

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,270 @@
...
+}',
+                    array(
+                        'exclude' => array(
+                            'json_encode',
+                        ),
+                    )
+                ),
+            ),
+            null,
+            'Configure names of functions to exclude',
```
keradus: provide reasoning why one would do that (polyfills)

tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
```diff
@@ -69,6 +69,53 @@ public function providerInvalidConfigurationElement()
         );
     }
 
+    public function testConfigureResetsExclude()
```
keradus: what about making it global for all fixers somehow ?

localheinz: Not sure I understand, can you explain?

keradus: this test method is checking that `null` is causing a reset of configuration for this one Fixer class.
what about making it general, to run on all Fixers ?

localheinz: Sounds like a good idea, but how can we do that?

Should we have an abstract test case for configurable fixers?

keradus: We put that kind of tests here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/FixerFactoryTest.php#L380  (not the best place for sure)
More tricky question is how to generalize the test for every fixer without a need to actually provide example configs for every single of them.

localheinz: Hehe, yes - let me take a look at this as soon as I have time.

What do you think about this PR so far, though, is there anything else you think should be done?

keradus: let me review it once more

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,272 @@
...
+
+        $indexes = \array_reverse($indexes);
+
+        $namespaceSeparator = new Token(array(
+            T_NS_SEPARATOR,
+            '\\',
+        ));
+
+        foreach ($indexes as $index) {
+            $tokens->insertAt($index, $namespaceSeparator);
```
keradus: Nice approach to not iterate backwards.
I would put `array_reverse` directly before `foreach` look to emphasise that we do insert in reverse order.
 
But, you must not insert same element at different indexes. This is token object. If one will decide in another fixer to change one of them, he will change all of them instead.

src/Test/AbstractFixerTestCase.php
```diff
@@ -171,8 +171,10 @@ protected function doTest($expected, $input = null, \SplFileInfo $file = null)
             $this->assertNull($fixResult, '->fix method must return null.');
         }
 
+        $actual = $tokens->generateCode();
+
         $this->assertThat(
-            $tokens->generateCode(),
+            $actual,
```
keradus: btw, huh ?

keradus:
I let myself fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2462/files#r96804912

tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
```diff
@@ -0,0 +1,409 @@
...
+',
+            ),
+            array(
+'<?php
+
+class WithoutNamespace
+{
+    public function bar($foo)
+    {
+        return json_encode($foo);
```
keradus: why not to put `\` even if code is not in namespace ?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,266 @@
...
+                        'exclude' => array(
+                            'json_encode',
+                        ),
+                    )
+                ),
+            ),
+            null,
+            'Configure names of functions to exclude, for example, when using a polyfill.',
+            self::$defaultConfiguration,
+            'Risky if a function with the same name as a native function exists in the current namespace, or a polyfill is used.'
```
keradus: and mocks, like `ClockMock`

tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
localheinz: I'm fine removing the namespace checking, shall I?

localheinz: I figured that if the code isn't in a namespace, then it's of no use adding the `\`, as we are already in the root namespace. That is, an attempt to resolve a function name in the same namespace will already yield the actual function, then, as it is the root namespace.

What do you think?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
localheinz: Need to update?

tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
keradus: it will not change the performance indeed, but IMO it's better to provide same style and DX, so less-experienced users will not get confused.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
keradus: yep, I would like to ask for an update

localheinz: Before I update, can you just explain real quick what you mean, though, please? I don't get it, hehe!

keradus: maybe sth like this ?

```diff
-'Risky if a function with the same name as a native function exists in the current namespace, or a polyfill is used.'
+'Rule is risky when a function with the same name as a native function exists in the current namespace.
+Two major situations when it could happen are:
+* polyfill is used to provide function from higher PHP version, if PHP CS Fixer would been run also on that higher PHP version, but software is also desired to work on lower PHP version that naively doesn't have that polyfilled function
+* function is mocked during tests execution, eg mocking `time` function - in that case after applying the rule src code will always use original, unmocked function
+To deal with described situation provide a configuration with function names you want to preserve unchanged.'
```

keradus: Hmmm... actually, a polyfilling may not be an issue here:
https://3v4l.org/fkiWD

Could you verify ?

SpacePossum:
@localheinz can you fix the 5.6 issue, it is a test thing only I thing :)
and please rebase as well.

Love to see this in!
(as I would like to build upon it to add config option to import rather than to add '\' for all native calls)

keradus:
> as I would like to build upon it to add config option to import rather than to add '' for all native calls

let us make it first as adding `\` (as it's simpler and easier), then with time we could add option to decide to import them instead of prefixing.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
keradus: Could one verify my thoughts ?
cc @localheinz , @SpacePossum

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,211 @@
...
+    /**
+     * @return string[]
+     */
+    private function getInternalFunctionNames()
+    {
+        $definedFunctions = \get_defined_functions();
+
+        return \array_map(function ($name) {
+            return \strtolower($name);
+        }, $definedFunctions['internal']);
```
Slamdunk: Minor improvement: https://github.com/localheinz/PHP-CS-Fixer/pull/4

localheinz:
Ha, sorry for the delay, will take a look tomorrow!

localheinz:
Tests should pass, take a look!

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,217 @@
...
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        $riskyDescription = <<<'TXT'
+Rule is risky when a function with the same name as a native function exists in the current namespace.
++Two major situations when it could happen are:
++* polyfill is used to provide function from higher PHP version, if PHP CS Fixer would been run also on that higher PHP version, but software is also desired to work on lower PHP version that naively doesn't have that polyfilled function
++* function is mocked during tests execution, eg mocking `time` function - in that case after applying the rule src code will always use original, unmocked function
++To deal with described situation provide a configuration with function names you want to preserve unchanged.'
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2462#discussion_r98139966

keradus: If my commit will work on 5.3 at Travis it would mean that the `polyfill` part is not valid, proving me being wrong and nobody check it himself :d
Still, `mocking` part is valid one way or another.

keradus: Travis is happy. @localheinz , could you adjust PR please ?
(also, please keep my .travis.yml change)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,217 @@
...
+    {
+        if (null === $configuration) {
+            $this->exclude = array();
+
+            return;
+        }
+
+        $key = 'exclude';
+
+        if (!\array_key_exists($key, $configuration) || !\is_array($configuration[$key])) {
```
SpacePossum: what if I don't want to exclude anything?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        static $internalFunctionNames = null;
+
+        if (null === $internalFunctionNames) {
+            $internalFunctionNames = $this->getInternalFunctionNames();
```
SpacePossum: what if I want to add a `\` to a function I declared in the global namespace but didn't load with the Fixer project? (i.e. non-native function but still declared in global namespace?)

keradus: cannot auto-find them, right? maybe direct config for it

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
localheinz: I believe it's not an issue!

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
localheinz: Give me a moment!

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,285 @@
...
+
+        if (empty($configuration)) {
+            throw new InvalidFixerConfigurationException(
+                $this->getName(),
+                'Configuration must define "exclude" and/or "include" as an array.'
+            );
+        }
+
+        if (\array_key_exists('include', $configuration) && \array_key_exists('exclude', $configuration)) {
+            $contradictingFunctions = \array_intersect(
```
localheinz: Not sure about this one, because actually, to determine contradictions, the function names should be normalized. Should we even care about contradicting configurations?

keradus: if one would set same function name under both keys then yes, I think we should raise an configuration error

localheinz: Should we worry about different casing?

For example:

```php
$configuration = [
    'exclude' => [
        'foo_bar',
    ],
    'include' => [
        'Foo_Bar',
    ],
];
```

keradus: do we care about casing when we decide to add `\` ?
here, should be the same behaviour

keradus:
@SpacePossum , I am not sure about that `include`... it's not good practice to create global functions... I'm not for encouraging people to use them by adding setting just for them

SpacePossum:
@keradus fair enough :) @localheinz please ignore my comment :+1:

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -0,0 +1,232 @@
...
+}',
+                    array(
+                        'exclude' => array(
+                            'json_encode',
+                        ),
+                    )
+                ),
+            ),
+            null,
+            'Configure names of functions to exclude, for example, when using a polyfill.',
```
keradus: mocking, not polyfilling.
how to configure them, at which key?

keradus:
@localheinz, is there anything left in this PR, or ready for merge ?

localheinz:
@keradus 

I can't think of anything, if you are all happy with it, let's 🚀 it!

keradus:
Thank you @localheinz.

localheinz:
Thank you, @keradus, @Slamdunk, and @SpacePossum!

SpacePossum:
thanks @localheinz , awesome new fixer!

PR #2463, opened 2017-01-13
Deal with cache signature rules encoding

GrahamCampbell:


GrahamCampbell:
D'no what we wanna do about the conditional function existence checks there?

keradus:
remove them and require `symfony/polyfill-mbstring` and `symfony/polyfill-xml`

src/Cache/Signature.php
```diff
@@ -73,4 +73,19 @@ public function equals(SignatureInterface $signature)
...
     }
+
+    private static function utf8Encode(array $data)
+    {
+        if (!function_exists('utf8_encode') || !function_exists('mb_detect_encoding')) {
+            return $data;
+        }
+
+        array_walk_recursive($data, function (&$item) {
+            if (is_string($item) && !mb_detect_encoding($item, 'utf-8', true)) {
```
keradus: do we need detection ? why?

GrahamCampbell:
> symfony/polyfill-xml

On a side note, that function was actually moved out of ext-xml in either php 7.1, or 7.2.

keradus:
7.2 is not there, 7.1 is just one version.
sure, so, they are in code in 7.1 ?

keradus:
ping @GrahamCampbell

keradus:
ping @GrahamCampbell

keradus:
ping @GrahamCampbell
really ?

src/Cache/Signature.php
SpacePossum: The problem with the `utf8_encode` function is that it might produce garbled output when the input is already UTF8 encoded. However it should still be an UTF8 encoded string so could be used as "signature part" to check against.

Looking at this PR, and I try not to talk anyone down here, I wonder, for the cache we need a signature so we can test if it is still valid. Can we do this without introducing the new dependencies? For example, if we serialize the configuration and MD5 (or other hashing, security is not an issue here) that, can we use that as way of comparing if the cache is invalid? (I know it is slower than json encoding, but it would be without new dependencies and we "only" need to do it one time for each run).

Besides that, I think if this PR solves an issue the current way, than we should add mb_string and xml as dependencies, otherwise the issue still remains for users without it. Next the the polyfil for 5.5 can go IMHO and replaced by detecting if the function is available. The details it provides are nice and OK, but not worth the additional depend. as users prop. won't use the additional information to fix the cache file.

tests/Cache/CacheTest.php
```diff
@@ -148,32 +150,48 @@ public function providerMissingData()
...
-
         $cache = new Cache($signature);
 
         $file = 'test.php';
         $hash = crc32('hello');
 
         $cache->set($file, $hash);
 
-        /* @var Cache $cached */
+        /** @var Cache $cached */
```
SpacePossum: maybe this type hint can be removed as a whole as the method `Cache::fromJson` already type hints the return value correctly?

src/Cache/Signature.php
SpacePossum: I just saw polyfill for 55 is already in composer.json, so please ignore that part of my comment.

GrahamCampbell:
> sure, so, they are in core in 7.1 ?

D'no. Don't really care. Doesn't matter.

GrahamCampbell:
We have the polyfil anyways. What do I actually need to do here?

keradus:
@SpacePossum , any concerns here ?

src/Cache/Signature.php
```diff
@@ -73,4 +73,19 @@ public function equals(SignatureInterface $signature)
 
         return true;
     }
+
+    private static function utf8Encode(array $data)
+    {
+        if (!function_exists('utf8_encode') || !function_exists('mb_detect_encoding')) {
```
SpacePossum: the function `utf8_encode` will always exist in this repo, as we polyfill it:
https://github.com/symfony/polyfill-xml/blob/master/bootstrap.php#L14

SpacePossum:
My concerns are about edge cases we won't likely to run into any time soon, nor can I back those concerns up with any real input or test. So for that those should be dismissed.

The changes here look good to me, these do improve what we have now :)
Thanks @GrahamCampbell 

I suggest merging this as is (although resolving https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2463/files#r109024293 would be nice).

src/Cache/Signature.php
keradus: what about requiring `symfony/polyfill-mbstring` and remove both checks, having cache works all the time...

GrahamCampbell: My personal feeling is I'd rather not add extra dependencies that we don't really need. Maybe add the extension in the "suggest" block?

keradus: and then...
> if you don't install suggested dependency the code is buggy and you cannot rely on caching feature...

GrahamCampbell: True, but basically nobody runs into this particular issue, but then we make everyone load a dependency they don't need either because this isn't an issue for them, or because they don't need the polyfill.

keradus: or because they were not aware that the caching is not working properly

keradus:
Thank you @GrahamCampbell for taking care of this.
As here are 2 commits of yours, I squashed them and picked into original PR.

PR #2481, opened 2017-01-27
Fix priorities between declare_strict_types and blank_line_after_opening_tag

### LINKED ISSUE #2492: Adding script_types right after <?php
If I run the fixer on files without `declare(strict_types=1);`, it will add it, but right after the opening `<?php` tag which itself thinks is wrong, if I re-run it, it will fix it again to be as expected. This means all fixes must be run twice.

**The PHP version you are using:**
PHP 7.1.1

**PHP CS Fixer version you are using:**
PHP CS Fixer 2.0.0 by Fabien Potencier and Dariusz Ruminski

**The command you use to run the fixer**
`vendor/bin/php-cs-fixer fix --diff -v --using-cache false`

and the configuration file you are using if any

    <?php
    declare(strict_types = 1);
        ->exclude('vendor')
        ->exclude('var')
        ->notPath('bin/symfony_requirement')
        ->in(__DIR__);
    
        ->setRiskyAllowed(true)
        ->setRules(
            [
                '@Symfony' => true,
                'declare_strict_types' => true,
                'php_unit_strict' => true,
                'phpdoc_order' => true,
                'psr4' => true,
                'strict_comparison' => true,
                'strict_param' => true,
            ]
        )
        ->setFinder($finder);

**The fixed version of that code after you run the fixer**

    <?php declare(strict_types=1);
    class Foo
    {
    }

**and the version you expected.**

    <?php
    
    declare(strict_types=1);
    class Foo
    {
    }

juliendufresne:
declare_strict_types must be run before blank_line_after_opening_tag is run.
Otherwise it will add a new line after the php tag and then add declare(strict_type=1); right after the php tag

PS: I didn't run php-cs-fixer fix because lots of tests files would be modified.

fixes #2492

### DIFF
--- a/src/Fixer/Strict/DeclareStrictTypesFixer.php
+++ b/src/Fixer/Strict/DeclareStrictTypesFixer.php
@@ -75,8 +75,8 @@ final class DeclareStrictTypesFixer extends AbstractFixer implements Whitespaces
      */
     public function getPriority()
     {
-        // must ran before SingleBlankLineBeforeNamespaceFixer.
-        return 1;
+        // must ran before SingleBlankLineBeforeNamespaceFixer and BlankLineAfterOpeningTagFixer.
+        return 2;
     }
 
     /**
--- a/tests/FixerFactoryTest.php
+++ b/tests/FixerFactoryTest.php
@@ -292,6 +292,7 @@ final class FixerFactoryTest extends \PHPUnit_Framework_TestCase
             array($fixers['no_useless_else'], $fixers['no_trailing_whitespace']), // tested also in: no_useless_else,no_trailing_whitespace.test
             array($fixers['no_useless_else'], $fixers['no_whitespace_in_blank_line']), // tested also in: no_useless_else,no_whitespace_in_blank_line.test
             array($fixers['declare_strict_types'], $fixers['single_blank_line_before_namespace']), // tested also in: declare_strict_types,single_blank_line_before_namespace.test
+            array($fixers['declare_strict_types'], $fixers['blank_line_after_opening_tag']), // tested also in: declare_strict_types,blank_line_after_opening_tag.test
             array($fixers['array_syntax'], $fixers['binary_operator_spaces']), // tested also in: array_syntax,binary_operator_spaces.test
             array($fixers['array_syntax'], $fixers['ternary_operator_spaces']), // tested also in: array_syntax,ternary_operator_spaces.test
             array($fixers['class_keyword_remove'], $fixers['no_unused_imports']), // tested also in: class_keyword_remove,no_unused_imports.test
--- a/tests/Fixtures/Integration/priority/declare_strict_types,blank_line_after_opening_tag.test
+++ b/tests/Fixtures/Integration/priority/declare_strict_types,blank_line_after_opening_tag.test
@@ -1,7 +1,7 @@
 --TEST--
 Integration of fixers: declare_strict_types,blank_line_after_opening_tag.
 --RULESET--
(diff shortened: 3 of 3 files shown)


GrahamCampbell:
> PS: I didn't run php-cs-fixer fix because lots of tests files would be modified.

Huh, what?

GrahamCampbell:
If this is a bug fix, it should be targeted at 2.0.

SpacePossum:
LGTM :)
:+1:

GrahamCampbell:
Other than the target branch, 👍 from me too.

GrahamCampbell:
The target can normally be resolved on merge tho.

juliendufresne:
> Huh, what? That's not right. Travis runs php-cs-fixer on itself successfully.
@GrahamCampbell My bad on that. In the [CONTRIBUTING.md](/FriendsOfPHP/PHP-CS-Fixer/CONTRIBUTING.md) file I read "`php php-cs-fixer fix .`" instead of "`php php-cs-fixer fix`."
Using the dot overide the Finder specified in the .php_cs.dist

juliendufresne:
> If this is a bug fix, it should be targeted at 2.0.

I rebased my work against the 2.0 branch.  
Sorry for the mistake.

keradus:
>declare_strict_types must be run before blank_line_after_opening_tag is run.
> Otherwise it will add a new line after the php tag and then add declare(strict_type=1); right after the php tag


Your integration test is not confirming that statement.
Adding only itests, without changing the priority:
https://github.com/keradus/PHP-CS-Fixer/pull/23
https://travis-ci.org/keradus/PHP-CS-Fixer/builds/196465517

juliendufresne:
Nice catch!  
It was a matter of ordering between fixer. Since I defined the declare_strict_types before the blank_line_after_opening_tag, it runs it before.

Now this should work as expected.

Do you want me to change the name of the file?
From  
`declare_strict_types,blank_line_after_opening_tag.test`  
to
`blank_line_after_opening_tag,declare_strict_types.test`

juliendufresne:
Nice catch! I fixed it. Thank you.

Indeed, in my integration test I declared the `declare_strict_types` rule before the `blank_line_after_opening_tag` so php-cs-fixer runs the first one before the later and tests passed without requiring to change the order.

When I switch the two fixers and keep the same priority, the test fails so the integration test is good now.

Do you want me to change the name of the file?
From  
`declare_strict_types,blank_line_after_opening_tag.test`  
to
`blank_line_after_opening_tag,declare_strict_types.test`

keradus:
Please, could you remove a priority change for a fixer, push it, allow travis to crash, then fix it and push fix again as separated commit ?

juliendufresne:
But when I do this, I have to comment a lines checking that priorities are not the same.

keradus:
feel free to temporary do it to prove that test will fail on fixing part.

juliendufresne:
I rebased my work:
* the first commit should fail in travis (proof the bug exists)
* the second one should fix the issue

Hope this will be accepted as I am facing lots of issue with this priority bug (and it requires me to run php-cs-fixer twice in order to fix it)

Note: I didn't have to remove the assertion I talked about.

juliendufresne:
Do you know why the appveyor check is not running for lots and lots PR?

keradus:
Thanks @juliendufresne  !
Please, take a look at #2483 before we will merge your PR.
Ah, ignore AppVeyor queue busy all the time.

keradus:
:+1: if Travis will be green

keradus:
Thank you @juliendufresne.

SpacePossum:
Thanks @juliendufresne nice fix :)

PR #2483, opened 2017-01-30
DX: AbstractIntegrationTestCase does not use IntegrationCase::shouldCheckPriority, logic is now automated and method is now deprecated

keradus:
inspired by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2481#issuecomment-275987557

2nd commit is there to show how it works and will be reverted when PR will be reviewed.


keradus:
cc @SpacePossum @juliendufresne

SpacePossum:
Like the idea :+1: 

I think this should only be done for tests with more than one fixer (there may not be such case in this project, but one may have custom ones).

Surprised not more tests are failing, IIRC I tried to run the test suit with `checkPriority = true` on all fixtures and some failed.

src/Test/AbstractIntegrationTestCase.php
```diff
@@ -243,39 +245,41 @@ protected function doTest(IntegrationCase $case)
...
+                count(array_unique(array_map(
+                    function (FixerInterface $fixer) {
+                        return $fixer->getPriority();
+                    },
+                    $fixers
+                ))),
+                sprintf(
+                    'Rules priorities are not differential enough. If rules would be used in reverse order then final output would be different than the expected one. For that, different priorities must be set up for used rules to ensure stable order of them. In "%s".',
+                    $case->getFileName()
+                )
```
juliendufresne: **I might be totally wrong**
I'm guessing this check will not work when you have more than 2 fixers with different priority but 2 of them with the same priority.

The real issue here is that we do an array_reverse but we don't sort by priority (something we do in the real run).  
We should only reverse fixers of the same priority and assert that the output remains the same.

- check that there are some fixers with the same priority. If not, skip the test
- sort fixers by priority, preserving the same order
- reverse fixers orders priority by priority
- assert output is the same as the original one.

SpacePossum: When fixers have the same priority but the output diff.'s. when running with the fixers in other order than there is a problem, i.e. the fixers should've diff. priority so the test helps finding those cases.

juliendufresne: **I might be totally wrong**
I'm guessing this check will not work when you have more than 2 fixers with different priority but 2 of them with the same priority.

The real issue here is that we do an array_reverse but we don't sort by priority (something we do in the real run).  
We should only reverse fixers of the same priority and assert that the output remains the same.

- check that there are some fixers with the same priority. If not, skip the test
- sort fixers by priority, preserving the same order
- reverse fixers orders priority by priority
- assert output is the same as the original one.

juliendufresne: Original comment (sorry for messing with this)
> **I might be totally wrong**
I'm guessing this check will not work when you have more than 2 fixers with different priority but 2 of them with the same priority.

The real issue here is that we do an array_reverse but we don't sort by priority (something we do in the real run).  
We should only reverse fixers of the same priority and assert that the output remains the same.

- check that there are some fixers with the same priority. If not, skip the test
- sort fixers by priority, preserving the same order
- reverse fixers orders priority by priority
- assert output is the same as the original one.

SpacePossum: Ah I get, yes :)
so this should only be done if the test contains a pair (or more)  fixers with same prioriy
and if so, only change the priority of those fixers

juliendufresne: i.e, only when:

```
$numberOfFixers = count($fixers);
$numberOfPriorities = count(array_unique(array_map(
  function (FixerInterface $fixer) {
    return $fixer->getPriority();
  },
  $fixers
)));

if ($numberOfFixers > 1 && $numberOfPriorities !== $numberOfFixers) {
  // run the tests here
  // only shuffle fixers of the same priorities
  // output should not diverge from original fixer
}
```

keradus:
> I think this should only be done for tests with more than one fixer (there may not be such case in this project, but one may have custom ones).

good point. on the same side, maybe just throw exception on integration test where there is only one rule, as this won't be an integration test at all... but yeah, it may be considered as a bc breaker.

> Surprised not more tests are failing, IIRC I tried to run the test suit with checkPriority = true on all fixtures and some failed.

didn't notice that. especially because of it I send that new integration test that would fail in described case to prove that mechanism is working.

src/Test/AbstractIntegrationTestCase.php
keradus: don't see big value of it to be honest.
integration test usually contains only 2 fixers, reversing their order is enough.

Anything else than ABC->CBA, eg ABC->BAC is different situation. we could add it, but CBA would most probably cover that scenario.

keradus: > so this should only be done if the test contains a pair (or more) fixers with the same priority

now: it the output is different for reverse order, check that priorities are different
your approach: if the priorities are the same, check would the output will be the same with reverse order

I think both ways would achieve what I want to have. Why later is better ?

SpacePossum: >integration test usually contains only 2 fixers

all `set` tests contain multiple fixers ([example](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.0/tests/Fixtures/Integration/set/@Symfony_whitespaces.test))

For the order, let say I got these Fixers [name:prio]
A:1
B:0
C:0
D:-1

when revered ordered run DCBA changes _are_ expected, when run as ACBD changes are _not_ expected.

SpacePossum:
>maybe just throw exception on integration test where there is only one rule

please don't, I like the integration test for one fixer as I can put a lot of complex cases in there which I don't want to type out in utest ;)

src/Test/AbstractIntegrationTestCase.php
keradus: purpose of `set` is not to check priorities, but general overcome
as I said, we could have it, but I would rather split it to separated PR (feel free to contribute anyone!), as this PR has to fix another concern raised by ref. PR.

SpacePossum: Maybe I miss read the PR than, as now the `AbstractIntegrationTestCase` will always test in reverse order now so the `set` tests are effected as well aren't they?

Secondly since there is no assert for when the content is the same in reverse order we loose the detection of invalid order dependency claims (i.e. `Set priorities must be significant`)?

keradus: > Maybe I miss read the PR than, as now the AbstractIntegrationTestCase will always test in reverse order now so the set tests are effected as well aren't they?

master version:
- test reverse order only when `checkPriority` is true (in my PR - always)
- `set` may or may not be tested in revers order, i don't know, i don't care - they are outside of `priority` subdir

> Secondly since there is no assert for when the content is the same in reverse order we loose the detection of invalid order dependency claims 

I add it at first, and then realize there are some problems where A and B fixers could run in any order, but have different priorities because of relations A<->C and B<->D

SpacePossum:
if we skip the one-fixer cases than I'm :+1: (after tests are fixed of course)

src/Test/IntegrationCase.php
```diff
@@ -168,9 +168,19 @@ public function getTitle()
...
      * @return bool
+     *
+     * @deprecated since v2.1, on ~2.1 line IntegrationTest check whether different priorities are required is done automatically, this method will be removed on v3.x
      */
     public function shouldCheckPriority()
     {
-        return $this->settings['checkPriority'];
+        @trigger_error(
+            sprintf(
+                'The "%s" class is deprecated. You should stop using it, as it will be removed in 3.0 version.',
```
keradus: done ;)

keradus:
one-fixer itest checked

src/Test/AbstractIntegrationTestCase.php
```diff
@@ -269,14 +262,27 @@ function (FixerInterface $fixer) {
...
+                            return $fixer->getPriority();
+                        },
+                        $fixers
+                    ))),
+                    sprintf(
+                        'Rules priorities are not differential enough. If rules would be used in reverse order then final output would be different than the expected one. For that, different priorities must be set up for used rules to ensure stable order of them. In "%s".',
+                        $case->getFileName()
+                    )
+                );
+            }
```
juliendufresne: Ok so this will only check the case where 2 (or more) fixers have the same priority and the fixer output are different.

When does that happens? I thought we were doing those kind of tests when priorities are different (or maybe I was in a specific use case).

What I said earlier was: since we are running the fixer twice (in the right and reversed order) and we know that fixers without the same priority will generate diff, why don't we use this second run to check that two fixers with the same priority generate the same fix?

I've also noticed (thanks to your PR) that if fixers are specified in an order different from their priorities, the IntegrationTest will produce wrong fix.

keradus: > I thought we were doing those kind of tests when priorities are different (or maybe I was in a specific use case).

We do unit tests for every rule separately.
We do integration tests for multiple rules combined if we think (or if it was proven) that they are conflicting one another. Sometimes they do, sometimes they dont. Then, this extra check will check that if they are conflicting indeed, the execution order is stable (so priorities are different).

-----------------------

> What I said earlier was: since we are running the fixer twice (in the right and reversed order) and we know that fixers without the same priority will generate diff, why don't we use this second run to check that two fixers with the same priority generate the same fix?

The purpose of the check is not "what should be the priority" but "what should be the output".
For that, our main concern is to get proper output - setting priorities are not the goal, but a way to achieve it.
If the outcome is the same regardless of order - it's good and we don't need to do anything else.
If the outcome is different for different orders - we need to check that priorities are set so we will always use same order.

-----------------------

> I've also noticed (thanks to your PR) that if fixers are specified in an order different from their priorities, the IntegrationTest will produce wrong fix.

There are two phases of itest:
- test in main order
- test in reverse order

main order is an order in which the Fixer tool itself will execute the rules - and they are sorted by https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Test/AbstractIntegrationTestCase.php#L312 , not but order of json config in itest file.

keradus:
good to merge then ?

SpacePossum:
Looks good to me :)

btw. why doesn't this trigger the deprecation warning in the utests here?

keradus:
not sure what do you refer to.
where would you trigger it?

SpacePossum:
oh yeah nevermind, :+1: :ship:

keradus:
no no, seriously, what was it ?

SpacePossum:
Just me not thinking correct.
My initial though was that there should be deprecation warnings triggered as long as we've integration tests with something like this in it: 
```
--SETTINGS--
{"checkPriority": false}
```

but that is not true, only calling AbstractIntegrationTestCase::shouldCheckPriority will trigger the warning. So users will not be notified the setting is no longer used by the tool, which is OK as users might still want to use the settings within there own test setup.

keradus:
ok. btw, for that, i drop them from our ifiles

src/Test/AbstractIntegrationTestCase.php
juliendufresne: Ok. I got lost on the code and didn't see where it sorts the fixer during the integration test although I looked. Now I see where fixers are sorted.

So if we suppose we have two fixers with different priorities, main order or reverse order will actually do the same fixes in the same order. All my discussion was because I didn't see where we sort the fixers in the integration test.

src/Test/AbstractIntegrationTestCase.php
```diff
@@ -269,14 +262,27 @@ function (FixerInterface $fixer) {
...
 
-            $this->assertNotSame(
-                $fixedInputCode,
-                $fixedInputCodeWithReversedFixers,
-                sprintf('Set priorities must be significant. If fixers used in reverse order return same output then the integration test is not sufficient or the priority relation between used fixers should not be set. In "%s".', $case->getFileName())
-            );
+            // If output is different depends on rules order - we need to verify that the rules are ordered by priority.
+            // If not, any order is valid.
+            if ($fixedInputCode !== $fixedInputCodeWithReversedFixers) {
+                $this->assertGreaterThan(
```
juliendufresne: I might be wrong again but IMHO, if the output is different in reversed order, we always have an issue that should be fixed using priorities.

Since no matter their orders are, fixers are first sorted by priority, the only reason why output is different in reversed order is because 2 fixers shares the same priority.

Shouldn't we just test that no matter the order of fixers, we always have the same output?
I don't want to bother you with me not understanding the code in details but since you asked me to have a look at this PR, I want it to be clear in my mind :)

src/Test/AbstractIntegrationTestCase.php
keradus: no. if we have 2 fixers with different priorities, itest will perform fixing process using different order

src/Test/AbstractIntegrationTestCase.php
keradus: wrong basing you made your conclusion on.

main `Runner` class perform fixing using rules
`FixerFactory` convert json describing rules one want to use for `FixerInterface[]` instances and **sort** them

IntegrationTest will take `fixers` from `FixerFactory` - that is already sorted.
Then, we run `Runner` with `fixers` in proper order to generate `fixedInputCode`.
Then, we run `Runner` with `fixers` in reverse order to generate `fixedInputCodeWithReversedFixers`

juliendufresne: Ok then. My bad again :)

SpacePossum:
:+1: thanks!

PR #2486, opened 2017-01-31
Add FunctionToConstantFixer

SpacePossum:
```diff
--- Original
+++ New
@@ @@
<?php
-echo phpversion();
-echo pi();
-echo php_sapi_name();
+echo PHP_VERSION;
+echo M_PI;
+echo PHP_SAPI;
```

### DIFF
--- a/src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
+++ b/src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
@@ -1,0 +1,202 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\LanguageConstruct;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolverRootless;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerConfiguration\FixerOptionValidatorGenerator;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class FunctionToConstantFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @var string[]
+     */
+    private static $availableFunctions = array(
+        'phpversion' => 'PHP_VERSION',
+        'php_sapi_name' => 'PHP_SAPI',
+        'pi' => 'M_PI',
+    );
+
+    /**
+     * @var array<string, string>
+     */
+    private $functionsFixMap;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->functionsFixMap = array();
+        foreach ($this->configuration['functions'] as $key) {
+            $this->functionsFixMap[$key] = self::$availableFunctions[$key];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 4; $index > 0; --$index) {
+            $candidate = $this->getReplaceCandidate($tokens, $index);
+            if (null === $candidate) {
+                continue;
+            }
+
+            $this->fixFunctionCallToConstant(
+                $tokens,
+                $index,
+                $candidate[0], // brace open
+                $candidate[1], // brace close
+                $candidate[2]  // replacement
+            );
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Replace core functions calls returning constants with the constants.',
+            array(new CodeSample("<?php\necho phpversion();\necho pi();\necho php_sapi_name();")),
+            null,
+            'Risky when any of the configured functions to replace are overridden.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should run before NativeFunctionCasingFixer
+        // must run before NoExtraConsecutiveBlankLinesFixer, NoSinglelineWhitespaceBeforeSemicolonsFixer, NoTrailingWhitespaceFixer and NoWhitespaceInBlankLineFixer
+        // must run after NoSpacesAfterFunctionNameFixer and NoSpacesInsideParenthesisFixer
+
+        return 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_STRING);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        $generator = new FixerOptionValidatorGenerator();
+        $functionNames = array_keys(self::$availableFunctions);
+        $functions = new FixerOptionBuilder('functions', 'List of function names to fix.');
+        $functions = $functions
+            ->setAllowedTypes(array('array'))
+            ->setAllowedValues(array(
+                $generator->allowedValueIsSubsetOf($functionNames),
+            ))
+            ->setDefault($functionNames)
+            ->getOption()
+        ;
+
+        return new FixerConfigurationResolverRootless('functions', array($functions));
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     * @param int    $braceOpenIndex
+     * @param int    $braceCloseIndex
+     * @param string $replacementConst
+     */
+    private function fixFunctionCallToConstant(Tokens $tokens, $index, $braceOpenIndex, $braceCloseIndex, $replacementConst)
+    {
+        $tokens->clearTokenAndMergeSurroundingWhitespace($braceCloseIndex);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($braceOpenIndex);
+        $tokens[$index]->clear();
+        $tokens->insertAt($index, new Token(array(T_STRING, $replacementConst)));
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return null|array
+     */
+    private function getReplaceCandidate(Tokens $tokens, $index)
+    {
+        // test if we are at a function all
+        if (!$tokens[$index]->isGivenKind(T_STRING)) {
+            return null;
+        }
+
+        $braceOpenIndex = $tokens->getNextMeaningfulToken($index);
+        if (!$tokens[$braceOpenIndex]->equals('(')) {
+            return null;
+        }
+
+        $braceCloseIndex = $tokens->getNextMeaningfulToken($braceOpenIndex);
+        if (!$tokens[$braceCloseIndex]->equals(')')) {
+            return null;
+        }
+
+        $functionNamePrefix = $tokens->getPrevMeaningfulToken($index);
+        if ($tokens[$functionNamePrefix]->isGivenKind(array(T_DOUBLE_COLON, T_NEW, T_OBJECT_OPERATOR, T_FUNCTION))) {
+            return null;
+        }
+
+        if ($tokens[$functionNamePrefix]->isGivenKind(T_NS_SEPARATOR)) {
+            // skip if the call is to a constructor or to a function in a namespace other than the default
+            $prevIndex = $tokens->getPrevMeaningfulToken($functionNamePrefix);
+            if ($tokens[$prevIndex]->isGivenKind(array(T_STRING, T_NEW))) {
+                return null;
+            }
+        }
+
+        // test if the function call is to a native PHP function
+        $lowerContent = strtolower($tokens[$index]->getContent());
+        if (!array_key_exists($lowerContent, $this->functionsFixMap)) {
+            return null;
+        }
+
+        return array(
+            $braceOpenIndex,
+            $braceCloseIndex,
+            $this->functionsFixMap[$lowerContent],
+        );
+    }
+}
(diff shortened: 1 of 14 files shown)


tests/Fixer/LanguageConstruct/FunctionToConstantFixerTest.php
```diff
@@ -0,0 +1,166 @@
...
+                '.'
+                ;',
+                '<?php echo
+                phpversion
+                (
+                )
+                ;',
+            ),
+            'Global namespaced.' => array(
+                '<?php echo \PHP_VERSION;',
```
GrahamCampbell: hmmm, kinda odd i think

SpacePossum: please elaborate, anything you want to see changed?

keradus: @GrahamCampbell ?

GrahamCampbell: No leading slash.

GrahamCampbell: I so no reason why we should want to actually keep it, since it's context is different to that of the replacement.

keradus: same as `phpversion()` was translated into `PHP_VERSION`, the `\phpversion()` was updated keeping `\` intact.

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should run before NativeFunctionCasingFixer
+        // must run before NoExtraConsecutiveBlankLinesFixer, NoSinglelineWhitespaceBeforeSemicolonsFixer, NoTrailingWhitespaceFixer and NoWhitespaceInBlankLineFixer
+        // must run after NoSpacesAfterFunctionNameFixer and NoSpacesInsideParenthesisFixer
```
keradus: btw, why is that?
if there would be some spaces, cannot we ignore them in this fixer and don't have that priority conflict ?

SpacePossum: we can ignore the white space, but that way we would have diff. end result based on which fixers are run first

SpacePossum:
Aside from squashing my last commit into the first I don't thing there is more to do here.
Please let me know if otherwise (@ reviewers).

keradus:
I would like to play with a fixer on my env first. I hope to do it tmr.

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,189 @@
...
+            return;
+        }
+
+        $this->configuration = array();
+        foreach ($configuration as $key) {
+            if (!is_string($key) || !array_key_exists($key, self::$defaultConfiguration)) {
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Configuration key must any of "%s".', implode('", "', self::$defaultConfiguration)));
+            }
+
+            $this->configuration[strtolower($key)] = self::$defaultConfiguration[$key];
```
keradus: if key is not lowered, then it will not be a key in `$defaultConfiguration`, so it would trigger exception
:+1: for triggering exception (config should be valid in the first place), but that means no need for calling `strtolower`

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,189 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Replace core functions calls returning constants with the constants.',
+            array(new CodeSample("<?php\necho phpversion();\necho pi();\necho php_sapi_name();")),
+            '',
+            'Configure any of `phpversion`, `php_sapi_name` and `pi`.',
```
keradus: ref #2500 
missing ~root key in configuration

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,189 @@
...
+        $tokens->insertAt($index, new Token(array(T_STRING, $replacementConst)));
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return array|false
+     */
+    private function isReplaceCandidate(Tokens $tokens, $index)
```
keradus: `isFoo` should return boolean value, not data.
please convert it to `getReplaceCandidate : ?array`

keradus:
@SpacePossum @julienfalque this one is conflicting with #2500 
which one we want to merge first ?

julienfalque:
The configuration handling will have to be updated but the options defined here are already compatible with #2500 as they use an array with string keys. It would really be a problem if the configuration was an array of tokens with numeric keys, e.g. `['foo', 'bar']` but it's not the case here.

You can merge this one first, I'll then rebase my PR and update the code.

keradus:
OK, then this one will go first.

(but still, ~root config key is missing, eg for current configuration: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2486/files#diff-8734f36b18f3357b8502815bc6d869f1R102 )

SpacePossum:
Maybe better to do 2500 first, after that I can work on the configuration of this fixer as I know what the new configuration structure will be. No rush on any of these two PR's : )

keradus:
for me, this PR is almost RTM (3 comments only). 2500 is way further

SpacePossum:
@julienfalque please let me know if the new config structure makes sense :)

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -144,7 +157,7 @@ private function fixFunctionCallToConstant(Tokens $tokens, $index, $braceOpenInd
      *
      * @return array|false
```
keradus: `null|array`

julienfalque:
Yep, it's now compatible, thx :)

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+
+/**
+ * @author SpacePossum
+ */
+final class FunctionToConstantFixer extends AbstractFixer implements ConfigurableFixerInterface
+{
+    /**
+     * @var array<string, string>
+     */
+    private static $defaultConfiguration = array(
```
julienfalque: Should be updated to use booleans.

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+        'phpversion' => 'PHP_VERSION',
+        'php_sapi_name' => 'PHP_SAPI',
+        'pi' => 'M_PI',
+    );
+
+    /**
+     * @var array<string, string>
```
julienfalque: `array<string, bool>`

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+    {
+        if (null === $configuration) {
+            $this->configuration = self::$defaultConfiguration;
+
+            return;
+        }
+
+        $this->configuration = array();
+        foreach ($configuration as $key => $value) {
+            if (!is_string($key) || !array_key_exists($key, self::$defaultConfiguration)) {
```
julienfalque: Isn't the `array_key_exists` check sufficient?

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+                throw new InvalidFixerConfigurationException($this->getName(), sprintf('Configuration key must be any of "%s".', implode('", "', array_keys(self::$defaultConfiguration))));
+            }
+
+            if (!is_bool($value)) {
+                throw new InvalidFixerConfigurationException(
+                    $this->getName(),
+                    sprintf(
+                        'Configuration value must be an bool, got "%s" for "%s".',
```
julienfalque:
```diff
-an bool
+a bool
```

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+                        is_object($value) ? get_class($value) : (gettype($value).'#'.$value),
+                        $key
+                    )
+                );
+            }
+
+            if ($value) {
```
julienfalque: This whole `if` looks odd to me, I would rather do the following after the `foreach`:

```php
$this->configuration = array_merge(self::$defaultConfiguration, $configuration);
```

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Replace core functions calls returning constants with the constants.',
+            array(new CodeSample("<?php\necho phpversion();\necho pi();\necho php_sapi_name();")),
+            '',
+            'Configure any of `phpversion`, `php_sapi_name` and `pi`.',
+            array_keys(self::$defaultConfiguration),
+            'Risky when the any of the configured functions to replace are overridden.'
```
julienfalque:
```diff
-Risky when the any of the configured functions to replace are overridden.
+Risky when any of the configured function to replace is overridden.
```

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
SpacePossum: why?

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
SpacePossum: it is always `<string, string>` ATM

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
julienfalque: Sorry, I missed the fact that the public configuration uses booleans but internally you still use the mapping. Some of my comments assumed you used booleans everywhere, you can ignore them :)

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
SpacePossum: and then filter out the `false` values again and for the left `true` values resolve the replacement value? 
The reason why this became odd is because the configuration requires some value. Maybe that is what I don't understand, why can't we just pass list of `strings` any more?

julienfalque: What we can't do anymore with #2500 is passing an array with keys that don't match option names (numeric keys or undefined options).

Here we had this: `['foo' => 'FOO', 'bar' => 'BAR']`, which can be seen as _"the configuration is a mapping of functions with constants"_. Strictly speaking, it _is_ compatible with #2500, assuming the option names are `foo` and `bar`.

But if you want to support replacing more functions, you have to create a new option for each. And if you want to add an option to tweak the behavior instead of extending supported functions, you end up mixing "mapping" options with "behavior" options. This is why I suggested passing the mapping in a dedicated `replacements` option (or any better name).

Thinking again, I think boolean options here are not the best solution because you may still mix "mapping" and "behavior" options, and you lose the ability to choose the constants.

SpacePossum: >Here we had this: ['foo' => 'FOO', 'bar' => 'BAR']

This was never used as the configuration format for this fixer.

The configuration here was;
`['funcA', 'funcB', ...  ]`
In the configure method this was mapped to
`['funcA' => 'funcA replacement', 'funcB'  => 'funcA replacement', ...  ]`

Now this is replaced with, configuration:
`['funcA' => bool, 'funcB' => bool, ...  ]`
Mapping; to the same format

The proposed `array_merge` will not result in the same mapping.

>Thinking again, I think boolean options here are not the best solution because you may still mix "mapping" and "behavior" options

I see no point in the booleans either, passing the function names did the trick just as well, however it was requested to change it; "Or store the map internally and use boolean options:" (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2486#issuecomment-286410636). 
Restructuring the config format for new options (so not new replacements) can be done when needed.

>and you lose the ability to choose the constants.

Nothing is lost, as it was never added. It is not a generic find-replace-function-with-constant-fixer by design.

tests/Fixer/LanguageConstruct/FunctionToConstantFixerTest.php
```diff
@@ -0,0 +1,166 @@
...
+            ),
+            'Overridden function.' => array(
+                '<?php if (!function_exists("phpversion")){function phpversion(){}}?>',
+            ),
+            'phpversion only' => array(
+                '<?php echo PHP_VERSION; echo php_sapi_name(); echo pi();',
+                '<?php echo phpversion(); echo php_sapi_name(); echo pi();',
+                array(
+                    'phpversion' => true,
+                    'php_sapi_name' => false,
```
keradus: what is the value of passing a key with `false` ?

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
julienfalque: > The configuration here was;
> ['funcA', 'funcB', ... ]

Ok so I definitely misread your original code, I'm deeply sorry for that.

For such listing, the easiest replacement to be compatible with #2500 would be to wrap the list in a dedicated option, e.g.:

```php
['functions' => ['funcA', 'funcB', ...]]
```

Same passed values, but nested in an array with a string key instead of numeric keys.

SpacePossum: no worries, thanks for the hints, will update later :+1:

tests/Fixer/LanguageConstruct/FunctionToConstantFixerTest.php
SpacePossum: there isn't, there was some confusion about the new config structure, will be solved in another way

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
SpacePossum: https://3v4l.org/W2dod

julienfalque: A bit extreme but ok :D

SpacePossum:
I re-did the configuration options, @julienfalque can you take a look again?

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -0,0 +1,202 @@
...
+        $functions = $functions
+            ->setAllowedTypes(array('array'))
+            ->setAllowedValues(array(
+                $generator->allowedValueIsSubsetOf($functionNames),
+            ))
+            ->setDefault($functionNames)
+            ->getOption()
+        ;
+
+        return new FixerConfigurationResolverRootless('functions', array($functions));
```
julienfalque: The `FixerConfigurationResolverRootless` is used to deprecate options that are not compatible with the new configuration resolver. As the option is new here, you should use `FixerConfigurationResolver` instead.

keradus: good catch, done

SpacePossum: thank you both, I've update the test as well

SpacePossum:
all green :)
thanks @keradus , interesting commit history for this PR :)

keradus:
Thank you @SpacePossum.

PR #2493, opened 2017-02-03
FixCommand - Output warning message when both config and rules options are passed

SpacePossum:


GrahamCampbell:
Probably best to target this at 2.1 since it's a feature?

keradus:
please focus on code rather the target ;) (for most cases) I could easily change the target during merging.

src/Console/Command/FixCommand.php
```diff
@@ -141,20 +144,26 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        }
+        if (null !== $stdErr) {
+            if (extension_loaded('xdebug')) {
+                $stdErr->writeln(sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.'));
+            }
 
-        $configFile = $resolver->getConfigFile();
+            if (null !== $passedConfig && null !== $passedRules) {
+                $stdErr->writeln(array(
+                    sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'When passing both "--config" and "--rules" the configuration file not to be used.'),
```
keradus: config file is used, but not ruleset defined in it. (eg, finder will be used)

src/Console/Command/FixCommand.php
```diff
@@ -141,20 +144,26 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                    sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'Passing both options is deprecated; version v3.0 PHP-CS-Fixer will exit with an configuration exit code.'),
```
keradus:
```diff
-configuration exit code
+configuration error exit code
```
?

keradus:
:+1: but minors

keradus:
I see no feature here, just better DX with more verbose CLI self-documentation comments

keradus:
@SpacePossum , would you mind adding the same warning for path when not in intersect mode ?

keradus:
Thank you @SpacePossum.

PR #2494, opened 2017-02-03
FixCommand - Support rules with params

### LINKED ISSUE #2423: [Question] How to execute rules with params via command line
I am trying to apply only a rule, without create a .php_cs config file, for example:

    $ php vendor/friendsofphp/php-cs-fixer/php-cs-fixer --rules=array_syntax mysrc

But that will take the default value for array_syntax value (which is syntax => long).

I would like use params via command line, something like:

    $ php vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --rules=array_syntax[syntax=>short] mysrc

That should convert all the arrays with syntax short (that is array() to [] ) which is more syntax sugar and clean for improve readability. But this obviously doesn't work and it doesn't take the "syntax" param as supported.

Alternative ways (using quoted parameters) could be this one:

 $ php vendor/friendsofphp/php-cs-fixer/php-cs-fixer fix --rules=array_syntax['syntax' => 'short'] mysrc

In fact this produces a weird exception (probably a bug) assuming that is a path:

    [PhpCsFixer\ConfigurationException\InvalidConfigurationException]  
  The path "=" is not readable.

But this is mostly a regresion, because in previous versions the documentation says that you can use short_array_syntax that was "deprecated" as you can read in [UPGRADE](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/UPGRADE.md) file, but it is only possible to use the values via config file and don't via parameters in command line?

ptcong:
According to issue #2423, i made this PR.
```
--rules='{"@PSR2": true, "class_keyword_remove": true, "concat_space": {"spacing": "none"}, "dir_constant": true}'
```

Closes #2423, #2502, #2510


SpacePossum:
Hi and thanks for the PR!
Please provide a utest for some cases. For example I think if someone passes a string that is not valid json this will result in an exception/error being raised.

ptcong:
When json_decode a string, it will returns null if failed, doesn't throw anything. So if user passes a string, it will run the code block under.

keradus:
Hi. Still, utest is required

ptcong:
I've just added test for this case.

SpacePossum:
(Please fix the tests by fixing the code style of the project itself.)

I think it would be nice improvement to allow JSON (that decodes to `array`), or rules comma separated list of fixer names.
This PR aims for that, however it would be nice to check if either one of two formats is used.
Now if you pass invalid JSON (or valid but resolves to non-array) the user does not get an error message that makes sense, because the tool will try to explode the invalid JSON input. WDYT?

src/Console/ConfigurationResolver.php
```diff
@@ -588,6 +588,12 @@ private function parseRules()
             return $this->getConfig()->getRules();
         }
 
+        $jsonRules = json_decode($this->options['rules'], true);
+
+        if (is_array($jsonRules)) {
```
keradus: exit on invalid user input instead of silently going forward

ptcong: Ok.

src/Console/ConfigurationResolver.php
```diff
@@ -588,10 +588,15 @@ private function parseRules()
             return $this->getConfig()->getRules();
         }
 
-        $jsonRules = json_decode($this->options['rules'], true);
+        $rules = trim($this->options['rules']);
 
-        if (is_array($jsonRules)) {
-            return $jsonRules;
+        if (in_array($rules[0], array('[', '{', '"', "'"), true)) {
```
keradus: valid json object must starts from `{`

src/Console/ConfigurationResolver.php
```diff
@@ -588,10 +588,15 @@ private function parseRules()
...
+            $rules = json_decode($rules, true);
+            if (!is_array($rules)) {
```
keradus: if it was decoded we already know it's an assoc object as it has to starts with `{` (due to prev comment).
then, consume `json_last_error` instead

src/Console/ConfigurationResolver.php
```diff
@@ -588,10 +588,15 @@ private function parseRules()
...
+                throw new InvalidConfigurationException('Invalid rules input.');
```
keradus: details to the user are needed

src/Console/ConfigurationResolver.php
```diff
@@ -590,10 +590,11 @@ private function parseRules()
 
         $rules = trim($this->options['rules']);
 
-        if (in_array($rules[0], array('[', '{', '"', "'"), true)) {
+        if ($rules[0] === '{') {
             $rules = json_decode($rules, true);
-            if (!is_array($rules)) {
-                throw new InvalidConfigurationException('Invalid rules input.');
+            if ($error = json_last_error()) {
+                $error = function_exists('json_last_error_msg') ? json_last_error_msg() : "Error $error";
```
keradus: drop checking method existence please

src/Console/ConfigurationResolver.php
```diff
@@ -588,6 +588,17 @@ private function parseRules()
             return $this->getConfig()->getRules();
         }
 
+        $rules = trim($this->options['rules']);
+
+        if ($rules[0] === '{') {
+            $rules = json_decode($rules, true);
+            if ($error = json_last_error()) {
```
SpacePossum: Although I like it, please don't assign inside the condition.
And please use `sprintf` to construct the exception message.

tests/ConfigTest.php
```diff
@@ -25,6 +26,41 @@
  */
 final class ConfigTest extends \PHPUnit_Framework_TestCase
 {
+    public function testConfigRulesUsingJson()
```
SpacePossum: nice cases :)
when you are done with the creating all tests please split up the cases in separate methods, pref. using dataproviders where handy.

src/Console/ConfigurationResolver.php
```diff
@@ -588,6 +588,17 @@ private function parseRules()
...
+                throw new InvalidConfigurationException("Invalid JSON rules input: Error $error");
```
keradus: drop checking method existence please
add `symfony/polyfill-php55` to requirements

src/Console/ConfigurationResolver.php
keradus: `if (null !== json_last_error())`

src/Console/ConfigurationResolver.php
ptcong: But the library still supports PHP 5.3 while `json_last_error_msg` required 5.5

> PHP needs to be a minimum version of PHP 5.3.6.

keradus: so, for the 3rd time,

> add `symfony/polyfill-php55` to requirements

after this, `json_last_error_msg` will work on 5.3

SpacePossum: >But the library still supports PHP 5.3 while json_last_error_msg required 5.5

yes, thats why `add symfony/polyfill-php55 to requirements`  ;)

ptcong: Thanks

composer.json
```diff
@@ -23,7 +23,8 @@
         "symfony/finder": "^2.2 || ^3.0",
         "symfony/polyfill-php54": "^1.0",
         "symfony/process": "^2.3 || ^3.0",
-        "symfony/stopwatch": "^2.5 || ^3.0"
+        "symfony/stopwatch": "^2.5 || ^3.0",
+        "symfony/polyfill-php55": "^1.3"
```
keradus: how did you add it?
why https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json#L37 didn' t work ?

ptcong: Sorry, i just added it manually :D

tests/ConfigTest.php
```diff
@@ -25,6 +26,47 @@
...
+        $config = new Config();
+        $configResolver = new ConfigurationResolver(
+            $config, array(
+                'rules' => '{"array_syntax": {"syntax": "short"}, "cast_spaces": true}',
+            ),
+            getcwd()
+        );
+        $fixers = $configResolver->getFixers();
+        $this->assertInstanceOf(\PhpCsFixer\Fixer\ArrayNotation\ArraySyntaxFixer::class, $fixers[0]);
+        $this->assertInstanceOf(\PhpCsFixer\Fixer\CastNotation\CastSpacesFixer::class, $fixers[1]);
```
keradus: * please do not rely on the order here, `getFixers` may return fixers in different order depends on priority of the fixers
* please check if the config was applied

tests/ConfigTest.php
```diff
@@ -25,6 +26,47 @@
...
+    public function testConfigRulesUsingInvalidJson()
+    {
+        $config = new Config();
+        $configResolver = new ConfigurationResolver(
+            $config, array(
+                'rules' => '{blah',
+            ),
+            getcwd()
+        );
+        $this->expectException(\PhpCsFixer\ConfigurationException\InvalidConfigurationException::class);
```
keradus: please set up the expectation on top of test method

keradus:
:+1; after travis will be fixed (one cannot use `::class` on lower PHP versions)

SpacePossum:
:+1:
(after test is green)

keradus:
Thank you @ptcong.

PR #2495, opened 2017-02-03
DX: Travis - check for trailing spaces

keradus:


keradus:
so far this is a check only, it will fail for now

PR #2496, opened 2017-02-03
DX: Travis - check for trailing spaces

keradus:


keradus:
so far this is a check only, it will fail for now

keradus:
fixed

SpacePossum:
:+1: 

(sad PHPStorm can no longer be configured to trim lines in string literals)

PR #2498, opened 2017-02-03
Fixes #2497 no_closing_tag: strip closing tag with whitespace to EOF

### LINKED ISSUE #2497: Allow no_closing_tag to strip closing tag with whitespace between it and EOF
Currently `no_closing_tag` works only for monolithic PHP files (without `T_INLINE_HTML`).

I have some PHP files that are detected to be not monolitic, because of whitespace after closing php tag `?>`, for example:

```
<?php

echo "...";

?>
\n
\n
\n
```

I would like this closing tag to also be removed, so after running `php-cs-fixer` it should be:

```
<?php

echo "...";
```

cronfy:
This PR is made after #2497 . It updates no_closing_tag fixer, allowing whitespace between closing tag and EOF to be stripped.

### DIFF
--- a/src/Fixer/PhpTag/NoClosingTagFixer.php
+++ b/src/Fixer/PhpTag/NoClosingTagFixer.php
@@ -40,7 +40,7 @@ final class NoClosingTagFixer extends AbstractFixer
         list($index, $token) = each($closeTags);
 
         switch (true) {
-            case $index == count($tokens) - 2:
+            case $index === count($tokens) - 2:
                 $nextToken = $tokens[$tokens->getNextNonWhitespace($index)];
                 if (!$nextToken->isGivenKind(T_INLINE_HTML)) {
                     return;
@@ -50,7 +50,7 @@ final class NoClosingTagFixer extends AbstractFixer
                 }
                 $nextToken->clear();
                 break;
-            case $index == count($tokens) - 1:
+            case $index === count($tokens) - 1:
                 break;
             default:
                 return;


keradus:
not sure did I get your intentions.

> allowing whitespace between closing tag and EOF to be stripped.

you did not provide test case with whitespace between `?>` and EOF.

the case you have provided is already there in non-monolitic version:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/Fixer/PhpTag/NoClosingTagFixerTest.php#L57-L65

the change in code you have introduced is to start modifying non-monolitic files, sth different that what you described in PR title/description.
what is your intention?

if intentions are to hande non-monolitic files, then it's BC breaking change 
and apart of that, non-monolitic files could be:
- php-based templates
- weird php files without clean purpose
for first case, in templates, each opening tag should have closing tag
for second case, we don't want to touch them

cronfy:
We discussed it at #2497 . After some time I can review my intentions, so basically I want to strip closing php tag in files like this:

```
<?php

// php code

?>
\n
\n
\n
```

But PHP-CS-Fixer considered it to be non-monolithic and did not fix it.

keradus:
then it's totally different what you have implemented here.
please make new, risky fixer that will:
- check if there is **one** inline html, that is at the end of file and contains only whitespaces, if so - remove it
- make integration test + priority utest that will make sure that new fixer works nicely with existing fixer you have just modified

keradus:
ping @cronfy
are you willing to finish this PR ?

cronfy:
@keradus no, sorry, I am unable to do it in a near time.

SpacePossum:
ping @cronfy any plans on finishing this, no rush or pressure, just trying to get a feeling on where some PRs stand ATM :)

cronfy:
@SpacePossum sorry, no: I am not PHP-CS-Fixer user anymore, so I afraid I can't do this job good.

SpacePossum:
No worries @cronfy , thanks for letting us know, I'm sure someone will pick this one up and finish it.
Sad to hear you're no longer a user of the fixer though!

SpacePossum:
Closing as this PR got out dated with the project and the author stated no longer using the project.
If anyone wants to picks this up, please lets us know :)

Issue #2482, opened 2017-01-27
Configuration resolver


julienfalque:
In configurable fixers, resolving the configuration and handling default values is always done by hand (see #2382 or #2220). Do you think it would make sense to use a dedicated class that would hold the default options, validate the values etc.? It could be an internal class (`ConfigurationResolver`, `ConfigurationBag`?). Or we could use an existing lib like https://github.com/symfony/options-resolver.

What do you think? If agreed, I'm willing to work on it.

localheinz:
I had the exact same thoughts! :+1:

SpacePossum:
sounds like a good idea, maybe a small POC would help to see what everyone thinks?

julienfalque:
Sure, working on it :)

localheinz:
It would be very nice, of course, if both default configuration and possible configuration options (along with explanations, if need be) could be extracted, so that they can be automatically compiled into `README.md` or at least when running the `describe` command. 

What do you think?

julienfalque:
Yes I plan to do that, I think it's possible with `symfony/options-resolver`.

keradus:
:+1: for idea. we already discussed it while working on 1.x line, but at that time we had only few configurable fixers so we decided to postpone that work and bring 2.0 sooner.
I will be more than happy to see this.

PR #2500, opened 2017-02-04
Configuration resolver

julienfalque:
Implements #2482 using `symfony/options-resolver` component.

Configurable fixers implement a new `ConfigurationDefinitionFixerInterface` that extends `ConfigurableFixerInterface` to add a `defineConfiguration()` method. This method must return an `OptionsResolver` instance that defines the configuration options. The interface is temporary (thus deprecated) and to be merged into `ConfigurableFixerInterface` in 3.0.

A few deprecations are introduced: some fixers can be configured by passing a list of strings, e.g. `['foo', 'bar']` but the `symfony/options-resolver` component is not designed for this. In such case, passing the list as the root configuration array is deprecated and a new dedicated option has been created to pass the list, e.g. `['option' => ['foo', 'bar']]`. This also allows to add new options to these fixers in the future. Fixers that handle such configuration will automatically convert the passed array and trigger a deprecation notice. Also, as `symfony/options-resolver` only deals with arrays, passing `NULL` to set the default configuration is deprecated, an empty array must be passed instead.

**Fix command help, before:**

fix-help-old

**Fix command help, after:**

fix-help-new

**Describe command, before:**

describe-old

**Describe command, after:**

describe-new

See also the [new README file](https://github.com/julienfalque/PHP-CS-Fixer/blob/configuration-resolver/README.rst).


GrahamCampbell:
Should probably be targeted at master, not 2.0.

GrahamCampbell:
> With this PR, FixerInterface has 4 new methods

Adding methods to the interface is a major breaking change is it not?

julienfalque:
Woops, of course it is 😅

I reverted `ConfigurableFixerInterface` removal and added a new `ConfigurationDefinitionFixerInterface` with the new methods instead. Both interfaces are deprecated and to be removed and merged into `FixerInterface` in 3.0.

julienfalque:
By the way, does anyone know why AppVeyor queues all PHP-CS-Fixer's builds and never runs them?

julienfalque:
Rebased onto master as introducing deprecations requires a minor version incrementation.

README.rst
```diff
@@ -182,480 +182,594 @@ automatically fix anything:
 
 Choose from the list of available rules:
 
-* **array_syntax**
-   | PHP arrays should be declared using the configured syntax (requires PHP
-   | >= 5.4 for short syntax).
-   | *Rule is: configurable.*
+**array_syntax**
```
keradus: why dropping the bullet ?

julienfalque: Because I was not sure that having the configuration options as a sublist would be nice, but I checked and it's ok, so I restored the bullets.

julienfalque:
I added configuration details in the `describe` command output. By the way, do you think I should add same colors for configuration details in the `fix` command help?

julienfalque:
I replaced `WhitespacesFixerConfig` by configuration options `indent` and `line_ending`. The `WhitespacesFixerConfig` class and all methods using it are now deprecated.

`FixerFactory::setWhitespacesConfig()` has no replacement yet. I'm thinking about something like a `setDefaultOptions()` method.

keradus:
> > I replaced WhitespacesFixerConfig by configuration options indent and line_ending. The WhitespacesFixerConfig class and all methods using it are now deprecated.

Please don't make such changes before alignment on it.
Big :-1: for that. One thing is that it is pointless to have this config specified for each fixer separately (repeating it for half of the fixers in docs is pointless), but bigger thing is or now one could configure different whitespaces for different fixers.
Please revert that change.

julienfalque:
> One thing is that it is pointless to have this config specified for each fixer separately (repeating it for half of the fixers in docs is pointless), but bigger thing is now one could configure different whitespaces for different fixers.

I didn't though about this and I completely agree. I removed the related commit.

The point was to have a single configuration entry point and I still think it would be great, but I'll try to figure out a better solution later, maybe in a separate PR.

keradus:
Thank you
Totally :+1: for not doing everything in one PR ;)

julienfalque:
Added description to all configuration options and tweaked the README.

Last thing I'd like to do is to deprecate `getConfigurationDescription()` and `getDefaultConfiguration()` methods from `FixerDefinitionInterface` since the details are also available through the new configuration definition. Do you agree?

src/AbstractFixer.php
```diff
@@ -80,6 +86,63 @@ public function supports(\SplFileInfo $file)
...
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        if (null === $configuration) {
+            @trigger_error(
+                'Passing NULL to set default configuration is deprecated and will not be supported in 3.0, use an empty array instead.',
+                E_USER_DEPRECATED
+            );
```
keradus: :-1: for this change. If we want to deprecate sth, I would rather split this method into 2 - one to set config, one to reset it

src/Fixer/Alias/NoMixedEchoPrintFixer.php
```diff
@@ -24,14 +22,9 @@
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  * @author SpacePossum
  */
-final class NoMixedEchoPrintFixer extends AbstractFixer implements ConfigurableFixerInterface
+final class NoMixedEchoPrintFixer extends AbstractFixer
```
keradus: dropping interface is a BC breaker

src/Fixer/ConfigurationDefinitionFixerInterface.php
```diff
@@ -0,0 +1,42 @@
...
+     * @return OptionsResolver $configurationDefinition
+     */
+    public function getConfigurationDefinition();
+
+    /**
+     * Returns whether the fixer has configuration options.
+     *
+     * @return bool
+     */
+    public function isConfigurable();
```
keradus: what for? if the fixer is not configurable, it must not follow this interface

src/AbstractFixer.php
julienfalque: Passing an empty array would still reset to default configuration anyway, and this is what this new method would do. I'm not sure adding it is relevant.

src/Fixer/Alias/NoMixedEchoPrintFixer.php
julienfalque: This is because the interface is implemented by `AbstractFixer` now, but I'm going to revert that change.

src/Fixer/ConfigurationDefinitionFixerInterface.php
julienfalque: Because I wanted this interface to be merged in `FixerInterface` in 3.0 so that fixers use this method instead of an interface to tell whether they are configurable.

src/AbstractFixer.php
```diff
@@ -12,32 +12,38 @@
...
+use Symfony\Component\OptionsResolver\Exception\ExceptionInterface;
+use Symfony\Component\OptionsResolver\Exception\MissingOptionsException;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  *
  * @internal
  */
-abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
+abstract class AbstractFixer implements FixerInterface, ConfigurationDefinitionFixerInterface, DefinedFixerInterface
```
julienfalque: Reverting.

src/AbstractFixer.php
keradus: If we want to have a deprecation let us make it the way it will follow separation of concerns. If not, then no need for introducing deprecation.

src/Fixer/ConfigurationDefinitionFixerInterface.php
keradus: Then the getter (`getConfigurationDefinition` or sth) could throw Exception (as it will do it even if `isConfigurable` will return `false`, as both methods are public). Is it not ?

julienfalque: Yep, this method was useful only because the interface was implemented by all fixers via `AbstractFixer`. As I am reverting this, the method becomes useless and I'm going to remove it anyway.

julienfalque:
I reverted the implementation of `ConfigurationDefinitionFixer` in `AbstractFixer`. Now, only fixers that are actually configurable implement it so I removed the `isConfigurable()` method from it.

I also added tests on the internal `OptionsResolver` that extends Symfony's one.

GrahamCampbell:
Not sure how this simplifies things. There's twice as much code now...

GrahamCampbell:
That said, the new help output is very nice. :P

julienfalque:
@GrahamCampbell You're right, maybe some parts could be simplified (thinking about `OptionsResolver::setNormalizer()` usages for example). I'll have a look at it.

keradus:
side request: @juliendufresne , if possible, please avoid squashing.
it's hard to see the changes you made across the time if the commits get squashed and one need to read whole diff again.
thanks !

julienfalque:
@keradus I use `git commit --fixup` a lot, is it ok if I still push fixup commits so we can squash them before merging?

keradus:
sure, don't bother about squashing on your side. we do it automatically during merging anyway.
(the only exception is when there are many authors to not remove authorship)

julienfalque:
@keradus When I edit the original post of the PR, fabbot.io checks are triggered while the code didn't change, is it expected?

SpacePossum:
fabbot.io works in mysterious ways, it's not something to worry about ;)

keradus:
new github hook msg is send, so new processing is executed

keradus:
@SpacePossum , we have just release `v2@stable`, I don't want to have this a BC breaker for 3.0, it will make maintainable 2.x<=>3.x harder. I don't wanna have BC breakers for at least next few months and I believe this PR could be achieved without breaking BC.

src/FixerDefinition/FixerDefinition.php
```diff
@@ -40,6 +40,15 @@ public function __construct(
         array $defaultConfiguration = null,
         $riskyDescription = null
     ) {
+        if (6 === func_num_args()) {
+            @trigger_error('Arguments #5 and #6 of FixerDefinition::__construct() is deprecated and will be dropped in 3.0, use argument #4 instead.', E_USER_DEPRECATED);
+        } elseif (5 === func_num_args()) {
+            @trigger_error('Argument #5 of FixerDefinition::__construct() is deprecated and will be dropped in 3.0.', E_USER_DEPRECATED);
+        } else {
+            $riskyDescription = $configurationDescription;
```
keradus: Why `$riskyDescription = $configurationDescription` ?

keradus: ah, as you try to remove not 5&6 arg, but 4&5

julienfalque:
> I believe this PR could be achieved without breaking BC

I can remove some of the deprecations introduced in this PR for the moment but I think they make sense. However, there is one that cannot be avoided because of `symfony/options-resolver` not handling some configuration cases as described in the OP.

I do understand the maintenance burden this PR may create so maybe it's wiser to wait before merging. But on the other hand, having full details of configuration options right in the README file seems valuable to me.

keradus:
@julienfalque, deprecations without breaking a BC are OK. Like triggering warnings or introducing new interfaces.
In my prev msg I refer to fact I remove the milestone, not to your code ;)

keradus:
@julienfalque, sorry for very small output from my side.

I think I do want to have here for sure is no BC breakers (deprecations and new interfaces are OK, as stated). For sure I would like to have more time and review whole PR once in the future, but for now it would be great if anyone from community would help reviewing this every-day bigger PR.

julienfalque:
@keradus Sure, no problem. I think I'm done here and I won't push more changes, unless I'm requested to. Anyway, thanks for the feedback you already gave :)

SpacePossum:
great @julienfalque :)
can you do the reviewers one more favor and rebase one more time, I hope to pick this one up shortly.

julienfalque:
Done.

src/AbstractFixer.php
SpacePossum: for completeness, as discussed @ https://github.com/julienfalque/PHP-CS-Fixer/pull/1#discussion_r102440937 this deprecation is indeed valid and should stay in this PR :+1:

src/OptionsResolver.php
```diff
@@ -0,0 +1,401 @@
...
+
+use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
+use Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver as BaseOptionsResolver;
+
+/**
+ * @internal
+ */
+class OptionsResolver extends BaseOptionsResolver
```
SpacePossum: please make the class `final`

julienfalque:
Adressed most of the review comments. Note that with `symfony/phpunit-bridge` being used for tests, the tested code will **not** be allowed to throw deprecation notices that are not explicitly expected in the related tests anymore.

Last point: as @SpacePossum noticed, there is no upgrade path to avoid using code that is deprecated in this PR. While I agree it would be better to have one, I wonder if it's really worth it: the only deprecated code that we can't avoid using is `ConfigurationDefinitionFixerInterface`, which is an interface so using it does not actually trigger any deprecation notices. @SpacePossum suggested to remove the `@deprecated` tag for now and add it again just before releasing `3.0`, which we may forget to do. So, is it worth it?

src/Fixer/Alias/NoMixedEchoPrintFixer.php
```diff
@@ -13,25 +13,20 @@
...
  * @author Sullivan Senechal <soullivaneuh@gmail.com>
  * @author SpacePossum
  */
-final class NoMixedEchoPrintFixer extends AbstractFixer implements ConfigurableFixerInterface
+final class NoMixedEchoPrintFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
-     * @deprecated will not be publicly available on 3.0
-     */
-    public static $defaultConfig = array('use' => 'echo');
```
keradus: why this was removed ?

src/Fixer/ConfigurationDefinitionFixerInterface.php
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+namespace PhpCsFixer\Fixer;
+
+use PhpCsFixer\OptionsResolver;
+
+/**
+ * @deprecated Will be merged into ConfigurableFixerInterface in 3.0
+ */
+interface ConfigurationDefinitionFixerInterface extends ConfigurableFixerInterface
```
keradus: I'm not sure is it a good way of doing things...

We have:
- `ConfigurableFixerInterface`
- `ConfigurationDefinitionFixerInterface extends ConfigurableFixerInterface`

what we want is: later is newest form how we want to specif configuration, all should stop using original one
what we does: newly added interface is deprecated, so one should not use it. as one won't use it, he will still use original interface


We should not add interface, that is deprecated from the very beginning of its existence.
We could add new interface and deprecate existing one.

src/Fixer/Alias/NoMixedEchoPrintFixer.php
julienfalque: I didn't see it was public, I'm gonna update the deprecation message instead.

src/Fixer/ConfigurationDefinitionFixerInterface.php
julienfalque: I know, the point was to add a new method to `ConfigurableFixerInterface` but this is a BC break, hence the new interface deprecated because it will be merged into the current one in 3.0. I removed the `@deprecated` tag.

Current state is:
* `ConfigurableFixerInterface` not deprecated;
* `ConfigurationDefinitionFixerInterface` extends `ConfigurableFixerInterface` with a new method, not deprecated.

So now I can deprecate `ConfigurableFixerInterface`, making `ConfigurationDefinitionFixerInterface` the interface to use in 3.0. But I think `ConfigurableFixerInterface` is a better name, I would like to keep it in 3.0 and remove `ConfigurationDefinitionFixerInterface` instead. So what about having:

* `ConfigurableFixerInterface` (not ?) deprecated, will have a new method in 3.0;
* `ConfigurationDefinitionFixerInterface` extends `ConfigurableFixerInterface`, not deprecated, will be deprecated in 3.0 and removed in 4.0.

Is such process desirable? worth it? too complicated? Does it make sense to deprecate an interface while it is not to be removed later?

keradus: @juliendufresne 

Please, don't go that far to think about v4, it won't be there for months, if not years.

current state:
We have 2 interfaces for same thing, user will have no idea which one he should use.
We should deprecate the one that we don't want to be used, here, it would be the original one.

Please:
- deprecate `ConfigurableFixerInterface`, as we don't want user to use it

What we will do in v4:
- drop deprecations, so we will remove `ConfigurableFixerInterface`
- we could make BC breakers, like renaming `ConfigurationDefinitionFixerInterface` (maybe into `ConfigurableFixerInterface`)

julienfalque: Fair enough, I added the deprecation.

julienfalque:
I think I adressed all last reviews comment.

src/Fixer/Comment/HeaderCommentFixer.php
```diff
@@ -13,63 +13,70 @@
...
+use Symfony\Component\OptionsResolver\Options;
 
 /**
  * @author Antonio J. García Lagar <aj@garcialagar.es>
  * @author SpacePossum
  */
-final class HeaderCommentFixer extends AbstractFixer implements ConfigurableFixerInterface, WhitespacesAwareFixerInterface
+final class HeaderCommentFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
-    const HEADER_PHPDOC = 'PHPDoc';
```
julienfalque: I will restore these constants. Removing them is not needed here and is a BC break anyway.

hkdobrev:
I'm not into the codebase of the PHP CS Fixer enough to help here, but the changes from this PR seems really good, so is there anything else needed here? I see AppVeyor is failing but seems like a temporary networking problem.

@julienfalque Thank you for all the work!

julienfalque:
Sadly AppVeyor isn't as reliable as Travis and is very slow. Wish we could run Travis on Windows and get rid of AppVeyor.

src/ConfigurationException/InvalidConfigurationException.php
```diff
@@ -24,11 +24,16 @@
...
 {
     /**
-     * @param string   $message
-     * @param int|null $code
+     * @param string          $message
+     * @param int|null        $code
+     * @param \Exception|null $previous
      */
-    public function __construct($message, $code = null)
+    public function __construct($message, $code = null, \Exception $previous = null)
```
keradus: in `\Exception` definition, it is `\Throwable`, not `\Exception`

src/Console/Command/ReadmeCommand.php
```diff
@@ -216,9 +216,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $help = str_replace('%command.name%', $command->getName(), $help);
         $help = preg_replace('#</?(comment|info)>#', '``', $help);
         $help = preg_replace('#^(\s+)``(.+)``$#m', '$1$2', $help);
-        $help = preg_replace('#^ \* ``(.+)``#m', '* **$1**', $help);
-        $help = preg_replace("#^\n( +)#m", "\n.. code-block:: bash\n\n$1", $help);
-        $help = preg_replace("#^\.\. code-block:: bash\n\n( +<\?(\w+))#m", ".. code-block:: $2\n\n$1", $help);
+        $help = preg_replace('#^ \* ``(.+)``(.*?\n)#m', "* **$1**$2\n", $help);
+        $help = preg_replace('#^   \\| #m', '  ', $help);
+        $help = preg_replace('#^   \\|#m', '', $help);
+        $help = preg_replace('#^(?=  \\*Risky rule.\\*$)#m', "\n", $help);
```
keradus: why such a change here and not in `getHelp` method ?

src/ConfigurationException/InvalidConfigurationException.php
julienfalque: Good catch, but won't it break on PHP 5.x? If so, this exception class is only supposed to wrap `OptionsResolver`'s exceptions, so it should not be a big issue.

src/Console/Command/ReadmeCommand.php
julienfalque: Because I wanted to keep both `fix --help` output and README file nice. Which change would like to move exactly? Everything?

src/OptionsResolver.php
```diff
@@ -0,0 +1,401 @@
...
+
+use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
+use Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException;
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver as BaseOptionsResolver;
+
+/**
+ * @internal
+ */
+final class OptionsResolver extends BaseOptionsResolver
```
keradus: This class adds support for `rootConfigurationOption`.
Is it the only reasoning for this class ?

src/OptionsResolver.php
```diff
@@ -0,0 +1,401 @@
...
+
+    /**
+     * Adds a normalizer for the given option.
+     *
+     * @param string   $option
+     * @param \Closure $normalizer
+     *
+     * @return $this
+     */
+    public function addNormalizer($option, \Closure $normalizer)
```
keradus: this is used only once.
why can't we rely only on setNormalizer ?

src/Console/Command/ReadmeCommand.php
keradus: (please freeze any update of readme file until further notice)

src/OptionsResolver.php
julienfalque: No: the base `OptionsResolver` class does not provide all the getters I added. It also provides descriptions support.

src/OptionsResolver.php
julienfalque: Don't remember. I'll have a look, maybe I can remove it.

keradus: If the only purpose of this method is to use `unbind` helper then drop both of them.
Nice workaround, but we gonna drop 5.3 anyway

src/AbstractFixer.php
```diff
@@ -26,6 +30,11 @@
 abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
 {
     /**
+     * @var array
```
keradus: actually, it's `Options`

keradus: actually, it may be null

src/OptionsResolver.php
keradus: Sadly, I do not like the approach here. Don't get me wrong, I don't have better solution ( at least yet ), but copy-paste all private properties and duplicate the code to handle them is a pretty big boilerplate. Imagine that sb would fix a bug in original implementation - we would not be aware of this in our duplicated code.

`OptionsResolver` class is not final, but yet it uses private properties.
Could you propose to make the properties protected instead ?

src/OptionsResolver.php
julienfalque: I needed it to implement `setAllowedValueIsSubsetOf()`. I'll check if I can implement it differently.

src/AbstractFixer.php
julienfalque: It can be `null` but not `Options`, the result of `OptionsResolver::resolve()` is an array which can contain any kind of value. I can use `array<string, mixed>|null` here if you want.

src/OptionsResolver.php
keradus: Could you elaborate? I don't see connection between those 2 functions:

```
λ grep -r addNormalizer src\Fixer  | wc -l
      1
λ grep -r setAllowedValueIsSubsetOf src\Fixer  | wc -l
      8
```

src/OptionsResolver.php
keradus: Another option would be not to inherit, but to compose.

Quick draft:

```php
final class FixerConfigurationResolver
{
    private $options; // FixerOptions[]
    private $optionsResolver; // Symfony class instance we use in following methods

    public function setRoot(string $name) { ... } // to be removed on 3.x line
    public function setOption(FixerOption $option) { ... }
    public function getOptions() :FixerOption[] { ... }
    public function resolve(array $opts) :array {...}
}

final class FixerOption // just a DTO
{
// required
    private $name;
    private $description;

// optional
    private $default;
    private $normalizer;
    private $validator;

    // getters, setters
}
```

that way we do not modify or rely on internal Symfony code.

src/OptionsResolver.php
julienfalque: `setAllowedValueIsSubsetOf()` defines a normalizer, so a subsequent call to `setNormalizer()` would override it, hence the `addNormalizer()` method. But I'll find a better way.

src/OptionsResolver.php
julienfalque: Yes, I really don't like this approach neither. Your idea seems way better, will do that.

src/OptionsResolver.php
keradus: wow, don't mix what is allowed with normalizer. it's different purpose.

replace `setAllowedValueIsSubsetOf` should be relaced with:
```php
->setAllowedTypes('opt', 'array')
->setAllowedValues('opt', function (...) {...})
```

src/OptionsResolver.php
keradus: be aware that `FixerConfigurationResolver` will be part of public interface of this package, same for `FixerOption`.
Make them final with interfaces.

julienfalque: Do we really need interfaces for that?

keradus: If we have a interface for a fixer itself, then an interface for configuring it, and then we will have a non-interfaced final class it will break open principle here. Remember that everywhere here end-user may want to create his own Fixer (rule class). Then, he will use our DTO. Great, but he cannot modify the behaviour, if for some reason he would want to. Then, with interface, he could provide his own logic.

keradus: This (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500#discussion_r104912822) proposal would make a lot of code change in this PR (update of all configs definitions). Before doing that it would be great if one could double check the idea.

cc @SpacePossum , willing to take a look ?

src/AbstractFixer.php
```diff
@@ -34,7 +43,7 @@ public function __construct()
     {
         if ($this instanceof ConfigurableFixerInterface) {
             try {
-                $this->configure(null);
+                $this->configure(array());
```
keradus: I pointed it here to show it.
`getConfigurationDefinition` must not create new instance all the time it is called. (Especially with new approach of creating bunch of DTO objects).

Currently, we generate new instances for one fixer when:
1. class is created, default configuration is applied
2. class is configured with user-provided config

No need in real tool run for doubling the object instantation.
Then, we also test it:
3. With every tests execution in `setUp`
4. With some tests execution when we provide custom config.

Also, applying the default config should be ultra fast. 

---------------

Proposal: lazy loading.
in whole src/fixers/*/ rename `getConfigurationDefinition` to `createConfigurationDefinition` and add new `getConfigurationDefinition` to this abstract class that will use `createConfigurationDefinition` if the instance was not yet created.

src/OptionsResolver.php
keradus: cc @SpacePossum , willing to take a look at my proposal (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500#discussion_r104912822) ? (not necessarily whole PR)

SpacePossum: sure thing, didn't find the time yet

julienfalque: Implemented.

src/OptionsResolver.php
julienfalque: Fixed.

src/AbstractFixer.php
julienfalque: Done.

src/AbstractFixer.php
julienfalque: I agree and tried to do that but I faced two issues.

First, it has unexpected results with configuration definitions that depend on some context of the fixer. It made `header_comment` configuration throw errors because it depends on the `WhitespacesFixerConfig` instance of the fixer, which can change between `configure()` calls. It can be fixed by forcing this fixer to always create a new configuration definition instance instead of reusing the first one.

But the second issue is an increasing of memory usage: I wasn't able to run the test suite without exceeding the memory limit on my setup. The limit is the default 128M so it's not high but the test suite usually requires ~40M. I'll have to investigate more to determine if this memory issue is a direct consequence of this change or something else that requires some optimization first. It looks like this only concerns the test suite though, as running the tool on the project didn't show a noticeable change in the consumed memory.

Anyway, applying this change didn't seem to have an impact on performance either, so I'm not sure it's worth it.

src/Console/Command/CommandHelp.php
```diff
@@ -269,6 +270,21 @@ public static function toString($value)
         return var_export($value, true);
     }
 
+    public static function getDisplayableAllowedValues(FixerOption $option)
+    {
+        if (null !== $allowed = $option->getAllowedValues()) {
```
keradus: please avoid side effects.
condition should not assign value

src/Console/Command/CommandHelp.php
```diff
@@ -269,6 +270,21 @@ public static function toString($value)
         return var_export($value, true);
     }
 
+    public static function getDisplayableAllowedValues(FixerOption $option)
```
keradus: what is the output ?

src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @param string $optionName
+     *
+     * @throws \LogicException when the option is already defined
+     *
+     * @return $this
+     *
+     * @deprecated will be removed in 3.0
+     */
+    public function mapRootConfigurationTo($optionName)
```
keradus: this method shoud be extracted to separated class that will inherit from this class. that class should be not only deprecated but also internal. that way we wont produce new deprecated code already by dev users

src/Console/Command/CommandHelp.php
julienfalque: Ah right, I forgot this one. I will assign the variable before the `if`. Just out of curiosity, why do you want to avoid such syntax? IMO it's more convenient and readable.

src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -0,0 +1,116 @@
...
+
+use Symfony\Component\OptionsResolver\Options;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+final class FixerConfigurationResolver implements FixerConfigurationResolverInterface
+{
+    /**
+     * @var array<string, FixerOptionInterface>
+     */
+    private $options = array();
```
keradus: we have an assoc array here. we don't need the keys that much. what about replacing it with `SplObjectStorage` (used as Set, not as Map) ?

julienfalque: The keys are just used internally to easily know whether an option is defined using `array_key_exists()`. Not sure how using `SplObjectStorage` would be better, but why not?

src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -0,0 +1,116 @@
...
+    private $root;
+
+    /**
+     * @param FixerOptionInterface $option
+     *
+     * @throws \LogicException when the option is already defined
+     *
+     * @return $this
+     */
+    public function addOption(FixerOptionInterface $option)
```
keradus: maybe make class immutable and pass options to constructor only ? wdyt?

src/FixerConfiguration/FixerConfigurationResolver.php
keradus: or, see next comment

src/FixerConfiguration/FixerConfigurationResolver.php
julienfalque: Having this method can be useful when a set of fixers share common options but some of them have specific ones too: the common options can be defined in an abstract parent class and the child can still add new options. I'd like to use that in #2220 :)

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,225 @@
...
+
+    /**
+     * Unbinds the given closure to avoid memory leaks. See {@see https://bugs.php.net/bug.php?id=69639 Bug #69639} for
+     * details.
+     *
+     * @param \Closure $closure
+     *
+     * @return \Closure
+     */
+    private function unbind(\Closure $closure)
```
keradus: let us drop this method. we dont care about 5.3

julienfalque: Well, this method is actually useful on all PHP versions supported here... except PHP 5.3 which does not bind `$this` to `Closure`s by default 😁.

Maybe I can find another workaround though, or just get rid of the `setAllowedValueIsSubsetOf()` method but it's kind of convenient.

src/Console/Command/CommandHelp.php
keradus: good practice. too easy to write
`if `($foo = bar()) {}`
then you don't know did one wanted to just assign and check is it truly or compare function result to variable

src/FixerConfiguration/FixerConfigurationResolver.php
keradus: btw: input could not be an array but any `iterable`

src/Console/Command/CommandHelp.php
julienfalque: Yep, I only do this in case I have both assignment **and** comparison. In `if ('foo' === $bar = baz()) { }` there is no ambiguity.

mlocati: I find far more readable the following code:
```php
$bar = baz();
if ('foo' === $bar) {
```

julienfalque: Maybe we could write a fixer to force one way or the other 😋

tests/ProjectCodeTest.php
```diff
@@ -81,6 +83,15 @@ function (\ReflectionClass $interface) {
             'PhpCsFixer\Config' => array('create'),
             'PhpCsFixer\Fixer\FunctionNotation\MethodArgumentSpaceFixer' => array('fixSpace'),
             'PhpCsFixer\Fixer\Import\OrderedImportsFixer' => array('sortingCallBack'),
+            'PhpCsFixer\FixerConfiguration\FixerConfigurationResolver' => array('addOption', 'mapRootConfigurationTo'),
+            'PhpCsFixer\FixerConfiguration\FixerOption' => array(
+                'removeDefault',
```
keradus: this one does not exist

tests/FixerConfiguration/FixerOptionTest.php
```diff
@@ -0,0 +1,125 @@
...
+
+        $function = function () {};
+        $option->setAllowedValues($function);
+        $this->assertSame(array($function), $option->getAllowedValues());
+    }
+
+    public function testSetAllowedValueIsSubsetOf()
+    {
+        $option = new FixerOption('foo', 'Bar.');
+        $this->assertSame($option, $option->setAllowedValueIsSubsetOf(array('baz', 'qux')));
```
keradus: You check only fact that interface is fluent.
Not the logic of given method

src/FixerConfiguration/FixerOption.php
keradus: ah, I misunderstand the code.
:+1: for unbinding

src/FixerConfiguration/FixerConfigurationResolver.php
julienfalque: Actually this would require `FixerConfigurationResolver` to not be `final`, is that ok?

keradus: no, let s think about it further

src/AbstractFixer.php
keradus: memory issue could be figured out. let us focus on first one first.

> It made header_comment configuration throw errors because it depends on the WhitespacesFixerConfig instance of the fixer, which can change between configure() calls.

It should not be done that way.
It's possible to run both methods - `configure` and `setWhitespacesConfig`.
If one would call: `setWhitespacesConfig`, `configure` - it works.
If one would call: `setWhitespacesConfig`, `configure`, `setWhitespacesConfig`, `configure` - it works.
If one would call `configure`, `setWhitespacesConfig` - it doesn't work, because configuration is set to outdated, previous whitespaces - one need to call `configure` all after all times he would call `setWhitespacesConfig` for `header_comment` Fixer.
I hope I show you the shortcoming of current implementation.

Both methods must be independent. I think you must remove that normalizer and call it not during applying configuration , but during using the value from config

src/Console/Command/CommandHelp.php
julienfalque: Assignment extracted.

src/Console/Command/CommandHelp.php
julienfalque: PHPDoc added.

src/Console/Command/CommandHelp.php
```diff
@@ -270,9 +270,17 @@ public static function toString($value)
...
+     *
+     * @param FixerOption $option
+     *
+     * @return array|null
+     */
     public static function getDisplayableAllowedValues(FixerOption $option)
     {
-        if (null !== $allowed = $option->getAllowedValues()) {
+        $allowed = $option->getAllowedValues();
+        if (null !== $allowed) {
```
keradus: not only here, in other places as well

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,225 @@
...
+    {
+        if (!$this->hasDefault()) {
+            throw new \LogicException('No default value defined.');
+        }
+
+        return $this->default;
+    }
+
+    /**
+     * @param string[]|string|null $allowedTypes
```
keradus: let us be strict here. consider this as PHP 7.1 code.
what type could you typehint in method prototype?
`?string $allowed` is cool, `?array $allowed` as well. but you cannot typehint `array|string|null $allowed`

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function getAllowedTypes()
+    {
+        return $this->allowedTypes;
+    }
+
+    /**
+     * @param string[]|string|null $allowedValues
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500/files#r105841300

tests/Fixer/LanguageConstruct/DeclareEqualNormalizeFixerTest.php
```diff
@@ -72,13 +85,16 @@ public function provideCases()
     }
 
     /**
+     * @param array $config
+     * @param mixed $expectedMessage
```
keradus: why mixed ?

src/FixerConfiguration/FixerOption.php
keradus: and then, you break this interface by passing function ;)

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,225 @@
...
+    public function setAllowedValueIsSubsetOf(array $allowedArrayValues)
+    {
+        $option = $this->name;
+
+        return $this
+            ->setAllowedTypes('array')
+            ->setAllowedValues($this->unbind(function ($values) use ($option, $allowedArrayValues) {
+                foreach ($values as $value) {
+                    if (!in_array($value, $allowedArrayValues, true)) {
+                        throw new InvalidOptionsException(sprintf(
```
keradus: why not `return false` ?

julienfalque: Would work too, but the error message would be something like _Option "foo" with value "array" is not valid_, which is less helpful that the error message here.

src/FixerConfiguration/FixerOption.php
julienfalque: I copy-pasted this from the previous one and forgot to update the PHPDoc. It should be `mixed` here.

keradus: :-1: for mixed

src/FixerConfiguration/FixerOption.php
julienfalque: This is what Symfony OptionsResolver supports: `string|string[]`, thought we should just support the same. I added `null` as well but I'm not sure it's useful actually.

src/FixerConfiguration/FixerOption.php
julienfalque: Why?

tests/ProjectCodeTest.php
julienfalque: This was a method I wrote and then removed, but I forgot to remove it from here too.

src/FixerConfiguration/FixerOption.php
keradus: no need for mimics symfony here. i prefer to be strict, then Zend releases PHP 7 which allows for that as well ;)
symfony code was created ages ago

src/FixerConfiguration/FixerOption.php
keradus: `mixed` could be **anything**. eg `resource` (like img handler). Do you handle it here?

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    /**
+     * Sets the given option to only accept an array with a subset of the given values.
+     *
+     * @param array $allowedArrayValues
+     *
+     * @return $this
+     */
+    public function setAllowedValueIsSubsetOf(array $allowedArrayValues)
```
keradus: Sorry, I would totally drop it.
This method breaks so many things. It's a convenient helper, but it's nothing more. It doesn't provide any functionallity.
And it provides issues.

There are 2 concepts in `OptionsResolver` component:
- normalizers - which **modify** value (eg ensuring there is no trailing spaces in value)
- validators (`setAllowed...`) - which **check** value (eg if they are in set or passing function)

If I would like to set up few validators:
```php
$foo->setAllowedValues([
    [$array, $of, $allowed, $values],
    function ($value) { /* validates value */ },
]);
$foo->setAllowedValueIsSubsetOf([...]);
```

what is the effect? first two validators were lost with 3rd one.

One could try to collect all validators, but then those methods wouldn't be setters, they would be adders. I don't like that.
IMO, `setAllowedValueIsSubsetOf` should be removed.

src/FixerConfiguration/FixerOption.php
julienfalque: Ok, so just `string[]` (typehint `array`)?

src/FixerConfiguration/FixerOption.php
julienfalque: Passing a `resource` does not make sense but it won't break validation, I'm fine with that. Do you want to be more strict anyway?

keradus: strict is good
loosly is bad
preferably do what could be achieved in prototype in 7.1

src/FixerConfiguration/FixerOption.php
keradus: :+1:

src/FixerConfiguration/FixerOption.php
keradus: Both msg provide same value

src/FixerConfiguration/FixerOption.php
keradus: ref https://github.com/julienfalque/PHP-CS-Fixer/pull/2

src/AbstractFixer.php
keradus: issue with `header_comment` fixed.
could you investigate memory issue when remembering the definition ?

src/Console/Command/CommandHelp.php
```diff
@@ -279,16 +337,6 @@ function (FixerInterface $a, FixerInterface $b) {
...
-
-            if ($fixer->isRisky()) {
-                $attributes[] = 'risky';
-            }
-
-            if ($fixer instanceof ConfigurableFixerInterface) {
-                $attributes[] = 'configurable';
-            }
-
             $description = wordwrap($description, 72, "\n   | ");
```
keradus: what is the difference between this function and newly introduced `wordwrap` method ?

julienfalque: The added method does not include style tags when couting lines length.

src/Fixer/ControlStructure/NoUnneededControlParenthesesFixer.php
```diff
@@ -71,17 +55,27 @@ public function __construct()
...
+                'clone',
+                'continue',
+                'echo_print',
+                'return',
+                'switch_case',
+                'yield',
+            ))
+        ;
+
+        return new FixerConfigurationResolverRootless('control_statements', array(
```
keradus: could be just `statements`

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -29,28 +31,27 @@
...
      */
-    private static $defaultConfiguration = array(
+    private $availableTokens = array(
+        'break',
+        'continue',
         'extra',
+        'return',
+        'throw',
+        'use',
+        'useTrait',
```
keradus: is it possible to somehow make this deprecated in favor of `use_trait` ?

tests/FixerConfiguration/FixerConfigurationResolverTest.php
```diff
@@ -0,0 +1,166 @@
...
+            throw $exception;
+        });
+
+        $catched = null;
+        try {
+            $configuration->resolve(array('foo' => '1'));
+        } catch (InvalidOptionsException $catched) {
+        }
+
+        $this->assertSame($exception, $catched);
```
keradus: why not `expectException` instead of `$catched , try-catch and assertion` ?

README.rst
```diff
@@ -197,482 +197,773 @@ automatically fix anything:
...
-   | Binary operators should be surrounded by at least one space.
-   | *Rule is: configurable.*
+
+  Binary operators should be surrounded by at least one space.
+
+  Configuration options:
+
+  - ``align_double_arrow`` (``NULL``, ``false``, ``true``): whether to apply, remove or
+    ignore double arrows alignment; defaults to ``false``
+  - ``align_equals`` (``NULL``, ``false``, ``true``): whether to apply, remove or ignore
```
keradus: It should be `null, false, true`, not `NULL, false, true`.

PS yes, `var_export` is stupid

tests/FixerConfiguration/FixerConfigurationResolverTest.php
julienfalque: To check that the catched exception is exactly the thrown one, but I'm fine with `expectException()` if you want.

keradus: Then I think the test is over engineered.  Aren't you testing OptionResolver of Symfony here ?

src/Console/Command/CommandHelp.php
```diff
@@ -288,7 +288,15 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
             });
 
-            sort($allowed);
+            $toString = function ($value) {
+                return call_user_func('PhpCsFixer\Console\Command\CommandHelp::toString', $value);
+            };
+            usort(
+                $allowed,
+                function ($valueA, $valueB) use (&$toString) {
+                    return strcasecmp($toString($valueA), $toString($valueB));
```
keradus: why not just
```php
return `strcasecmp(CommandHelp::toString($valueA), CommandHelp::toString($valueB));`
```
?

src/AbstractFixer.php
```diff
@@ -104,8 +110,12 @@ public function configure(array $configuration = null)
             $configuration = array();
         }
 
+        if (null === $this->configurationDefinition) {
+            $this->configurationDefinition = $this->getConfigurationDefinition();
```
keradus: with this approach one will still call `getConfigurationDefitinion()` and create new object all the time.
it is getter that should provide cached object.

tests/FixerConfiguration/FixerOptionTest.php
```diff
@@ -112,6 +116,10 @@ public function testGetNormalizers()
         $option = new FixerOption('foo', 'Bar.');
         $this->assertNull($option->getNormalizer());
 
+        if (PHP_VERSION_ID >= 50400) {
+            $this->markTestIncomplete('Assertion below fails due to closure unbinding');
+        }
```
keradus: then, let us drop those assertions

tests/FixerConfiguration/FixerConfigurationResolverTest.php
julienfalque: No, this makes sure that `FixerConfigurationResolver::resolve()` does not replace/wrap the thrown exception. But yes, maybe not very important to test that, I'll use `setExpectedException()` instead.

src/FixerConfiguration/FixerOption.php
```diff
@@ -50,13 +50,40 @@
...
+                if ($allowedValue instanceof \Closure) {
+                    $allowedValue = $this->unbind($allowedValue);
+                }
+            }
+        }
+
         $this->name = $name;
         $this->description = $description;
+        $this->isRequired = $isRequired;
+        $this->default = $default;
```
keradus:
```php
if ($isRequired && $default) {
    throw new...
}
```
?

src/FixerConfiguration/FixerOptionBuilder.php
```diff
@@ -0,0 +1,126 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerConfiguration;
+
+final class FixerOptionBuilder
```
keradus: I'm wondering... maybe internal ?

src/Console/Command/CommandHelp.php
julienfalque: I guess I didn't drink enough coffee when writing this :p

keradus: i just hate calling methods by string :D

tests/FixerConfiguration/FixerOptionTest.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->assertNull($option->getAllowedValues());
+
+        $option = new FixerOption('foo', 'Bar.', true, null, null, array('baz'));
+        $this->assertSame(array('baz'), $option->getAllowedValues());
+
+        $option = new FixerOption('foo', 'Bar.', true, null, null, array('baz', 'qux'));
+        $this->assertSame(array('baz', 'qux'), $option->getAllowedValues());
+
+        if (PHP_VERSION_ID >= 50400) {
+            $this->markTestIncomplete('Assertion below fails due to closure unbinding');
```
keradus: then, let us drop those assertions

src/AbstractFixer.php
keradus: 2nd issue solved

src/FixerConfiguration/FixerOptionBuilder.php
julienfalque: Why? It could be useful to devs that write custom fixers.

tests/FixerConfiguration/FixerOptionTest.php
julienfalque: Actually I rewrote similar assertions in another test to make them compatible, forgot to rewrite this one.

src/Console/Command/CommandHelp.php
julienfalque: Agreed, this is a hacky workaround for PHP 5.3 that does not allow calling `self::foo()` in closures.

julienfalque: Just didn't remember I could just replace `self` with the actual class name 😁.

src/FixerConfiguration/FixerOptionBuilder.php
keradus: why they can't use `FixerOption` ?

julienfalque: They can, just as we can too. But the builder is more convenient to avoid passing all the arguments to `FixerOption` constructor.

src/AbstractFixer.php
```diff
@@ -110,7 +110,7 @@ public function configure(array $configuration = null)
             $configuration = array();
         }
 
-        if (null === $this->configurationDefinition) {
+        if (null === $this->configurationDefinition || $this instanceof WhitespacesAwareFixerInterface) {
```
keradus: https://github.com/julienfalque/PHP-CS-Fixer/pull/7

src/AbstractFixer.php
```diff
@@ -80,6 +95,44 @@ public function supports(\SplFileInfo $file)
...
+        if (null === $configuration) {
+            @trigger_error(
+                'Passing NULL to set default configuration is deprecated and will not be supported in 3.0, use an empty array instead.',
+                E_USER_DEPRECATED
+            );
+
+            $configuration = array();
+        }
+
+        if (null === $this->configurationDefinition || $this instanceof WhitespacesAwareFixerInterface) {
```
keradus: https://github.com/julienfalque/PHP-CS-Fixer/pull/7

src/AbstractFixer.php
```diff
@@ -80,6 +95,44 @@ public function supports(\SplFileInfo $file)
...
+            $this->configurationDefinition = $this->getConfigurationDefinition();
```
keradus: with this approach one will still call `getConfigurationDefitinion()` and create new object all the time.
it is getter that should provide cached object.

src/AbstractFixer.php
```diff
@@ -80,6 +95,44 @@ public function supports(\SplFileInfo $file)
...
+                'Passing NULL to set default configuration is deprecated and will not be supported in 3.0, use an empty array instead.',
+                E_USER_DEPRECATED
+            );
+
+            $configuration = array();
+        }
+
+        if (null === $this->configurationDefinition) {
+            $this->configurationDefinition = $this->getConfigurationDefinition();
+        }
```
keradus: with this approach one will still call `getConfigurationDefitinion()` and create new object all the time.
it is getter that should provide cached object.

src/FixerConfiguration/FixerOption.php
```diff
@@ -0,0 +1,176 @@
...
+                if ($allowedValue instanceof \Closure) {
+                    $allowedValue = $this->unbind($allowedValue);
+                }
+            }
+        }
+
+        $this->name = $name;
+        $this->description = $description;
+        $this->isRequired = $isRequired;
+        $this->default = $default;
```
keradus: for some reason GitHub shows it outdated, but it's not:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500#discussion_r106368247

src/AbstractFixer.php
```diff
@@ -41,9 +41,9 @@
     protected $whitespacesConfig;
 
     /**
-     * @var FixerConfigurationResolver|null
+     * @var FixerConfigurationResolverInterface|null
      */
-    private $configurationDefinition;
+    protected $configurationDefinition;
```
keradus: Not sure if we want to split `AbstractFixer`. If extracting configurable part, one could also want to extract whitespaceConfigurable part. but then we would end up in 2 new abstract classes, and then `header_comment` want to inherit both of them...
Please rethink that, if we really want to move that part away, we should do it fully, like moving `configure` into `AbstractConfigurableFixer`, this property and handling in constructor

src/AbstractFixer.php
```diff
@@ -26,15 +31,25 @@
...
+
+    /**
      * @var WhitespacesFixerConfig
      */
     protected $whitespacesConfig;
 
+    /**
+     * @var FixerConfigurationResolverInterface|null
+     */
+    protected $configurationDefinition;
```
keradus: Not sure if we want to split `AbstractFixer`. If extracting configurable part, one could also want to extract whitespaceConfigurable part. but then we would end up in 2 new abstract classes, and then `header_comment` want to inherit both of them...
Please rethink that, if we really want to move that part away, we should do it fully, like moving `configure` into `AbstractConfigurableFixer`, this property and handling in constructor

src/Console/Command/CommandHelp.php
```diff
@@ -298,9 +353,53 @@ function (FixerInterface $a, FixerInterface $b) {
                 $help .= sprintf(" * <comment>%s</comment>\n   | %s\n", $fixer->getName(), $description);
             }
 
-            if (count($attributes)) {
-                sort($attributes);
-                $help .= sprintf("   | *Rule is: %s.*\n", implode(', ', $attributes));
+            if ($fixer->isRisky()) {
+                $help .= "   | *Risky rule.*\n";
```
keradus: maybe replace it with sth like `Rule may be risky: $descriptionFromDefinition` ?

src/Console/Command/ReadmeCommand.php
keradus: readme is fine for me :+1:

src/FixerConfiguration/FixerOption.php
keradus: resolved.

keradus:
Thank you @julienfalque.

julienfalque:
Thanks everyone for helping me on this!

SpacePossum:
@julienfalque awesome work, great addition to the tool! :D  thanks so much!

Issue #2504, opened 2017-02-05
FunctionDeclarationFixer – Make space after (anonymous) function configurable


jtojnar:
[PSR-2](http://www.php-fig.org/psr/psr-2/#closures) speaks quite explicitly about spaces with regards to lambdas:

>Closures MUST be declared with a space after the `function` keyword

It would be nice, if this were configurable similarly to how `concat_space` is. That is choosing `closure_function_spacing` from `none` and `one`.

PR #2505, opened 2017-02-06
FunctionDeclarationFixer – Make space after anonymous function configurable

jtojnar:
Closes: #2504

### DIFF
--- a/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
+++ b/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
@@ -28,8 +28,14 @@ use PhpCsFixer\Tokenizer\TokensAnalyzer;
  */
 final class FunctionDeclarationFixer extends AbstractFixer implements ConfigurableFixerInterface
 {
+    /**
+     * @internal
+     */
     const SPACING_NONE = 'none';
 
+    /**
+     * @internal
+     */
     const SPACING_ONE = 'one';
 
     private $supportedSpacings = array(self::SPACING_NONE, self::SPACING_ONE);
--- a/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
+++ b/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
@@ -149,7 +149,7 @@ final class FunctionDeclarationFixer extends AbstractFixer implements Configurab
                 $tokens[$startParenthesisIndex - 1]->clear();
             }
 
-            if ($this->configuration['closure_function_spacing'] === self::SPACING_NONE && $isLambda) {
+            if ($isLambda && self::SPACING_NONE === $this->configuration['closure_function_spacing']) {
                 // optionally remove whitespace after T_FUNCTION of a closure
                 // eg: `function () {}` => `function() {}`
                 if ($tokens[$index + 1]->isWhitespace()) {
--- a/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
+++ b/src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer\Fixer\FunctionNotation;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\CT;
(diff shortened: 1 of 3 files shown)


SpacePossum:
Thanks for picking this up.
As this is a PSR2 rule I wonder if we need config at all and not just always remove the space, but ey thats just my vote ;)

jtojnar:
Well, PSR states that there **should be** a space. This option allows the opposite.

tests/Fixer/FunctionNotation/FunctionDeclarationFixerTest.php
```diff
@@ -161,6 +171,141 @@ function foo() /* bar */
      * @param string      $expected
      * @param null|string $input
      *
+     * @dataProvider provideClosureFunctionSpacingNoneCases
+     */
+    public function testFixClosureFunctionSpacingNone($expected, $input = null)
```
keradus: should not exist.
modify `testFix` to allow optional `array $configuration = null` and pass it in data provider

src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
```diff
@@ -24,11 +26,53 @@
...
+
+    private $supportedSpacings = array(self::SPACING_NONE, self::SPACING_ONE);
+
     private $singleLineWhitespaceOptions = " \t";
 
     /**
+     * @var array
+     */
+    private static $defaultConfiguration = array(
+        'closure_function_spacing' => self::SPACING_ONE,
```
keradus: `one` or `none` is preferred by PSR ?

jtojnar: `one` is preferred.

tests/Fixer/FunctionNotation/FunctionDeclarationFixerTest.php
jtojnar: Do you mean something like this?

```php
<?php
public function provideXCases()
{
	$spacingNone = array('closure_function_spacing' => 'none'),

	$shared = array_map_cross(function ($configuration, $case) {
		return array($configuration) + $case;
	}, array(null, $spacingNone), array(
		array('c'),
		array('expected d', 'd'),
	));

	return $shared + array(
		array(null, 'a'),
		array(null, 'expected b', 'b'),
		array($spacingNone, 'expected a', 'a'),
		array($spacingNone, 'b'),
	);
}

public function testX($configuration, $expected, $input = null)
{
	if (isset($configuration)) {
		$this->fixer->configure($configuration);
	}

}
```

jtojnar: Do you mean something like this?
```php
<?php
public function provideXCases()
{
	$spacingNone = array('closure_function_spacing' => 'none'),

	$defaultConfigCases = self::casesForConfigurations(array(null), array(
		array('a'),
		array('expected b', 'b'),
	));

	$spacingNoneCases = self::casesForConfigurations(array($spacingNone), array(
		array('expected a', 'a'),
		array('b'),
	));

	$sharedCases = self::casesForConfigurations(array(null, $spacingNone), array(
		array('c'),
		array('expected d', 'd'),
	));

	return $defaultConfigCases + $spacingNoneCases + $sharedCases;
}

public function testX($configuration, $expected, $input = null)
{
	if (isset($configuration)) {
		$this->fixer->configure($configuration);
	}

}

/**
 * @param array $configurations
 * @param array $cases
 *
 * @return array
 */
public static function casesForConfigurations($configurations, $cases)
{
	$result = array();

	foreach ($configurations as $configuration) {
		foreach ($cases as $case) {
			$result[] = array($configuration) + $case;
		}
	}

	return $result;
}
```

keradus: we use config as 3rd param ;)
I don't see a need for `casesForConfigurations` method or merging arrays via `+`.

just

```php
function provide...() {
    return [
        [ $expected, $input, ],
        [ $expected, $input, $config, ],
    ];
}
```

jtojnar: How do you handle when `$expected === $input`? Can you link me to some test file that uses this?

As for the helper method,

```php
$defaultConfigCases = self::casesForConfigurations(array(null), array(
	array('a'),
	array('expected b', 'b'),
));

$spacingNoneCases = self::casesForConfigurations(array($spacingNone), array(
	array('expected a', 'a'),
	array('b'),
));

$sharedCases = self::casesForConfigurations(array(null, $spacingNone), array(
	array('c'),
	array('expected d', 'd'),
));

return $defaultConfigCases + $spacingNoneCases + $sharedCases;
```

looks much more cleaner than

```php
return array(
	array('a'),
	array('expected b', 'b'),
	array('expected a', 'a'),
	array('b'),
	array('c'),
	array('expected d', 'd'),
	array('c', 'c', $spacingNone),
	array('expected d', 'd', $spacingNone),
));
```

especially when there are more than two shared cases.

keradus: > `$expected === $input`

input is nullable, so 
`[ $expected, null, null ]` (which is the same as ` [ $expected ]`
or
`[ $expected, null, $config ]`

-----------------------

strongly disagree, with later it way easier to add more cases (without need to change return statement evaluation) an we could use custom labels, like
`return [
    'no changes' => [$expected],
    'changes' => [$expected, $input],
    'changes with one space' => [$expected, $input, $config],
    'changes without space' => [$expected, $input, $config],
]'`

jtojnar: Changed as per your suggestion.

src/Fixer/FunctionNotation/FunctionDeclarationFixer.php
```diff
@@ -24,11 +26,53 @@
  *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
-final class FunctionDeclarationFixer extends AbstractFixer
+final class FunctionDeclarationFixer extends AbstractFixer implements ConfigurableFixerInterface
 {
+    const SPACING_NONE = 'none';
```
SpacePossum: Not sure about adding these (public) `const`'s here, for the other fixers we didn't do that either.
Maybe we could add those `WhitespacesFixerConfig` and re-use over other fixers? (don't do that in this PR though for now, lets discuss first ;) ), otherwise I would opt for removing and just using the hardcoded string.

jtojnar: Rationale for using the constants is here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2459#discussion_r95901213

Originally, I wanted to use private `const`s but apparently they are only supported by PHP 7.1: https://3v4l.org/ZKbBJ

keradus: :+1: for constant, mark it as with `@internal` annotation please.

jtojnar: Fixed.

SpacePossum: thanks so much, this PR is getting closer to getting merged :)
I would like to do some testing, could you please rebase on top of the master branch?

jtojnar: Rebased.

SpacePossum:
@keradus after rebase RTM, however should the configuration options here also be changed to match https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500?

julienfalque:
They already are compatible with #2500.

SpacePossum:
alright than, @jtojnar please rebase :)

jtojnar:
@SpacePossum rebased.

tests/Fixer/FunctionNotation/FunctionDeclarationFixerTest.php
```diff
@@ -22,14 +22,31 @@
...
+
     /**
      * @param string      $expected
      * @param null|string $input
+     * @param null|array  $configuration
      *
      * @dataProvider provideCases
      */
-    public function testFix($expected, $input = null)
+    public function testFix($expected, $input = null, $configuration = null)
```
keradus: everywhere - please add typehinting for new parameter, it's array, so `array $configuration = null`.

jtojnar: Fixed

keradus:
Thank you @jtojnar.

PR #2506, opened 2017-02-06
SCA

SpacePossum:


src/Fixer/Import/NoLeadingImportSlashFixer.php
```diff
@@ -30,20 +30,20 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
     {
         $foundNamespace = $tokens->findGivenKind(T_NAMESPACE);
         if (empty($foundNamespace)) {
-            return;
+            return; // wrong see test4.php and test5.php
```
SpacePossum: note to self, remove comment and create new PR with fix

src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
```diff
@@ -86,6 +81,7 @@ public function getPriority()
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_OPEN_TAG);
+        // ignore files with short open tag and ignore non-monolithic files
+        return $tokens[0]->isGivenKind(T_OPEN_TAG) && $tokens->isMonolithicPhp();
```
keradus: this makes `isCandidate` slow instead of extra fast

keradus: 2nd thing, `->fix` method must still works properly if one forgot to call `->isCandidate`

src/Fixer/PhpTag/LinebreakAfterOpeningTagFixer.php
```diff
@@ -66,6 +61,7 @@ public function getDefinition()
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_OPEN_TAG);
+        // ignore files with short open tag and ignore non-monolithic files
+        return $tokens[0]->isGivenKind(T_OPEN_TAG) && $tokens->isMonolithicPhp();
```
keradus: same here

src/Fixer/PhpTag/NoClosingTagFixer.php
```diff
@@ -69,6 +60,6 @@ public function getDefinition()
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_CLOSE_TAG);
+        return $tokens->isTokenKindFound(T_CLOSE_TAG) && $tokens->isMonolithicPhp();
```
keradus: and here

src/Fixer/ClassNotation/ProtectedToPrivateFixer.php
```diff
@@ -82,7 +81,7 @@ public function getPriority()
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isAnyTokenKindsFound(Token::getClassyTokenKinds());
+        return $tokens->isTokenKindFound(T_CLASS);
```
SpacePossum: not handled by the fixer; https://github.com/SpacePossum/PHP-CS-Fixer/blob/ffc59269d3750ffd71db6fbe2d5dad2f87a2516b/src/Fixer/ClassNotation/ProtectedToPrivateFixer.php#L34
Besides I think you can safely change a `protected` method to `private` in a Trait.

src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
SpacePossum: would be nice to cache the `isMonolithicPhp` result within `Tokens` for performance, I'll have a look

SpacePossum: >2nd thing, ->fix method must still works properly if one forgot to call ->isCandidate

this is not the case for a bunch of fixers, but I see your point

src/Fixer/ClassNotation/ProtectedToPrivateFixer.php
keradus: aaah, right ;)

src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
keradus: Please revert this change for now, and caching could be postponed to separated PR

keradus: from public interface, one could still run ->fix without ->isCandidate, right? later one is just a hint "it may be worth to run it" / " no need for run it for sure"

SpacePossum: yes, however I noticed a bunch of fixers rely on the candidate check, same for example that a lot of fixers won't run correctly on an empty token collection while this is valid (i.e. empty file)

keradus:
Thank you @SpacePossum.

PR #2508, opened 2017-02-06
Add JSON output to describe command

mlocati:
[This is more a feature request than a real pull request]

What about adding an option to the describe command to make it output details in JSON format?
This could be very helpful in some automated environment.

I also added a new `configurationSchema` property to the `FixerDefinitionInterface`: it can be used by automated tools to know the format of the fixers configuration (using the [JSON Schema notation](http://json-schema.org/)).

Here's a sample output of `php-cs-fixer describe --json @Symfony`:

```json
{
    "ruleset": "@Symfony",
    "rules": {
        "binary_operator_spaces": {
            "align_double_arrow": false,
            "align_equals": false
        },
        "blank_line_after_namespace": true,
        "blank_line_after_opening_tag": true,
        "blank_line_before_return": true,
        "braces": {
            "allow_single_line_closure": true
        },
        [...omissis...]
    }
}
```

And a sample output of `php-cs-fixer describe --json braces`:

```json
{
    "name": "braces",
    "summary": "The body of each structure MUST be enclosed by braces. Braces should be properly placed. Body of braces should be properly indented.",
    "description": "",
    "priority": -25,
    "risky": false,
    "riskyDescription": "",
    "configurationDescription": "The `allow_single_line_closure` key could be set to `true` to allow for single line lambda notation.",
    "defaultConfiguration": {
        "allow_single_line_closure": false,
        "codeSamples": {
            "code": "<?php\n$positive = function ($item) { return $item >= 0; };\n$negative = function ($item) {\n                return $item < 0; };\n",
            "configuration": {
                "allow_single_line_closure": true
            }
        }
    },
    "configurationSchema": {
        "properties": {
            "allow_single_line_closure": {
                "type": "boolean"
            }
        },
        "additionalProperties": false
    },
    "codeSamples": []
}
```

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -29,6 +29,7 @@ use PhpCsFixer\Tokenizer\Tokens;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -62,6 +63,7 @@ final class DescribeCommand extends Command
                 )
             )
             ->setDescription('Describe rule / ruleset.')
+            ->addOption('json', 'j', InputOption::VALUE_NONE, 'Format output in JSON format')
         ;
     }
 
@@ -71,14 +73,18 @@ final class DescribeCommand extends Command
     protected function execute(InputInterface $input, OutputInterface $output)
     {
         $name = $input->getArgument('name');
+        $jsonOutput = $input->getOption('json');
         try {
             if ('@' === $name[0]) {
-                $this->describeSet($output, $name);
+                $this->describeSet($output, $name, $jsonOutput);
 
                 return;
             }
-
-            $this->describeRule($output, $name);
+            if ($jsonOutput) {
+                $this->describeRuleJSON($output, $name);
+            } else {
+                $this->describeRule($output, $name);
+            }
         } catch (DescribeNameNotFoundException $e) {
             $alternative = $this->getAlternative($e->getType(), $name);
             $this->describeList($output, $e->getType());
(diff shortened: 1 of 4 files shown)


mlocati:
PS: the only breaking change is [this one](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2508/commits/75fd6a61202dc9f826f41da862cb519fcdb271af#diff-9dc398d8e540fd164493c346866acdd6R57) (adding a new `getConfigurationSchema` method to `FixerDefinitionInterface`).
In order to keep BC, we could remove that method from `FixerDefinitionInterface` and replace [this line](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2508/commits/75fd6a61202dc9f826f41da862cb519fcdb271af#diff-88cad9ab17d8e3b415e648e3fd31f3d2R238) with this one:

```php
if ($fixer instanceof ConfigurableFixerInterface && method_exists($fixer, 'getConfigurationSchema')) {
```

(BTW, does anybody implements a `FixerDefinitionInterface` as an alternative to the current `FixerDefinition`?)

SpacePossum:
There is work going on here to formalize the configuration of the fixers here  https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2500. I know your PR aim to output the configuration in JSON (together with other details that don't require BC breaking changes), maybe these PR's could be complimentary to each other?

mlocati:
@SpacePossum Having a structured way to define fixers arguments would simplify everything a lot, since we don't have to describe/check the arguments twice (in the `configure` and in the `getDefinition` method): an `OptionsResolver` (defined just once) could be used for both purposes.

I'll leave this PR open as a reminder, either for myself and for others, so that we can talk about JSON output once we're done with #2500

src/Console/Command/DescribeCommand.php
```diff
@@ -199,7 +205,55 @@ private function describeRule(OutputInterface $output, $name)
...
+        /** @var FixerInterface $fixer */
+        $help = array(
+            'name' => $fixer->getName(),
+            'summary' => '',
+            'description' => '',
+            'priority' => $fixer->getPriority(),
+            'risky' => $fixer->isRisky(),
+            'riskyDescription' => '',
+            'configurationDescription' => '',
+            'defaultConfiguration' => ($fixer instanceof ConfigurableFixerInterface) ? false : null,
```
GrahamCampbell: brackets redundant here actually

src/FixerDefinition/FixerDefinitionInterface.php
```diff
@@ -48,4 +48,11 @@ public function getRiskyDescription();
      * @return CodeSampleInterface[]
      */
     public function getCodeSamples();
+
+    /**
+     * Configuration Schema (null for non-configurable fixer, false if not implemented)
+     *
+     *  @return array|null|false
+     */
+    public function getConfigurationSchema();
```
GrahamCampbell: This is a major breaking change to a public facing interface.

mlocati:
@GrahamCampbell I don't think this PR needs fixes, IMHO it will be discarded when #2500 will be merged
(we'll need a complete rewrite of this code)...
BTW, about the breaking change: see [my comment above](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2508#issuecomment-277632104).

SpacePossum:
ping @mlocati 2500 has been merged :)
@keradus what do you think about the idea behind this PR?

keradus:
Regardless implementation - question is do we want to have json formatter for `describe` command. I don't see much value of it yet.

mlocati:
> ping @mlocati 2500 has been merged :)

Finally! My mailbox has been flooded by tons of notifications about #2500 :wink:

mlocati:
> do we want to have json formatter for describe command. I don't see much value of it yet.

Do you mean that the subject of this PR should be moved to a new command, or that this whole argument is useless for you?

In the fist case, I've not any opinion.
In the second case: think for instance at 3rd party tools like style.ci, having an automated way to list the rulesets/rules and knowing their parameters is a must...

keradus:
any cases for real users ?

GrahamCampbell:
👎 Don't really see the point in this.

keradus:
@mlocati , please show codebase of real case integration with PHP CS Fixer that would benefit from this command, in other case I'm big :-1: on this as it's not needed for regular user at all, but yet would need to be maintained.

mlocati:
there's no use case I can see for end users

keradus:
what about first part?
> codebase of real case integration with PHP CS Fixer that would benefit from this command

mlocati:
I don't understand: would you like to see some code about how this integration could be done?

Meroje:
[Pronto](https://github.com/mmozuras/pronto) would benefit from an integration with php-cs-fixer if it was possible to grab the information this PR seeks to make available

keradus:
codebase case that could benefit does makes sense.
In Proto I see how it may display violation info, and we do provide fixing report in json (or anything)

GrahamCampbell:
> In Pronto I see how it may display violation info, and we do provide fixing report in json (or anything)

I really hate tools that post crap about code style and coverage as comments on GitHub. Totally backwards in my opinion.

Meroje:
> and we do provide fixing report in json (or anything)
Only fixing report is provided, there's no description of the violations

GrahamCampbell:
PHP-CS-Fixer is great, because it just fixes the issues, no feedback spam.

Meroje:
> as comments on GitHub

that's optional, what's really great with these tools is the aggregation of all the reports, from css to php and event speed regressions

mlocati:
Thanks to #2500, we can extract the details of every fixer.
So, the features of this PR are not required (if users that want these details can work with PHP sources and have time to study the php-cs-fixer code).

I just started a [project](https://mlocati.github.io/php-cs-fixer-configurator/) to help people configure php-cs-fixer ([here's the source code](https://github.com/mlocati/php-cs-fixer-configurator)).
At the moment, it does nothing except printing out the list of fixers and giving details about them. BTW I'll improve it, so that it will be able to generate configuration files (eg `.php_cs` / `.php_cs.dist` files, options for style.ci, ...).

So, I think this PR can be closed.

SpacePossum:
Thanks @mlocati for getting back on this. Best of luck with website, looking good already! :)

PR #2516, opened 2017-02-07
IsNullFixerTest - set exception expectation on top

keradus:


julienfalque:
I'm not in favor of standardizing such changes. Setting the exception expectation at the very beginning of the test means that anything in the test is allowed to throw said exception, including the fixer instanciation. Here it is quite unlikely to happen, but still.

keradus:
on the same side, we test one method in one test method.
For that, it is not allowed in this project to set up exception expectation and call multiple methods ;)

SpacePossum:
>For that, it is not allowed in this project to set up exception expectation and call multiple methods ;)

Agreed, however it cannot be helped in all cases. For example one might need to call a constructor to create an object where you want to call that one method on, the constructor might throw the exception as well while it shouldn't. 
I prefer setting the expected exception just before calling what is supposed to throw the exception.

localheinz:
For reference, also see https://thephp.cc/news/2016/02/questioning-phpunit-best-practices.

keradus:
Closing **without** merging due to discussion.
Thanks guys !

SpacePossum:
:+1: please know it was not personal :)  thanks for PR'ing!

keradus:
of course, reasoning was nice ;)

PR #2520, opened 2017-02-08
Add OpeningTagPlusEchoToShortEchoTagFixer

jtojnar:
This commit adds OpeningTagPlusEchoToShortEchoTagFixer which nicely complements the NoShortEchoTagFixer. This is useful for PHP templates.

### DIFF
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -259,6 +259,7 @@ final class FixerFactory
     {
         static $conflictMap = array(
             'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+            'no_short_echo_tag' => array('opening_tag_plus_echo_to_short_echo_tag'),
         );
 
         $fixerName = $fixer->getName();
--- a/README.rst
+++ b/README.rst
@@ -462,6 +462,9 @@ Choose from the list of available rules:
    | There should not be space before or after object ``T_OBJECT_OPERATOR``
    | ``->``.
 
+* **opening_tag_plus_echo_to_short_echo_tag**
+   | Replace ``<?php echo`` with short-echo ``<?=`` syntax.
+
 * **ordered_class_elements**
    | Orders the elements of classes/interfaces/traits.
    | *Rule is: configurable.*
--- a/src/Fixer/PhpTag/OpeningTagPlusEchoToShortEchoTagFixer.php
+++ b/src/Fixer/PhpTag/OpeningTagPlusEchoToShortEchoTagFixer.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpTag;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Vincent Klaiber <hello@vinkla.com>
+ */
+final class OpeningTagPlusEchoToShortEchoTagFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $i = count($tokens);
+        $lastEcho = null;
+
+        while ($i--) {
+            $token = $tokens[$i];
+
+            if ($token->isWhitespace()) {
+                continue;
+            } elseif ($token->isGivenKind(T_ECHO)) {
+                $lastEcho = $i;
+                continue;
+            } elseif (!$token->isGivenKind(T_OPEN_TAG)) {
+                $lastEcho = null;
+                continue;
+            }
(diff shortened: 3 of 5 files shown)


jtojnar:
The tests fail on HHVM due to https://github.com/facebook/hhvm/issues/7161

https://3v4l.org/a8YLJ

SpacePossum:
Not sure about how popular this fixer will become.
If this is going to be accepted as part of the fixers set of this repo, at least the tests must be fixed and the fixer must be marked as conflicting with `NoShortEchoTagFixer`.

src/FixerFactory.php
```diff
@@ -259,6 +259,7 @@ private function getFixersConflicts(FixerInterface $fixer)
     {
         static $conflictMap = array(
             'no_blank_lines_before_namespace' => array('single_blank_line_before_namespace'),
+            'no_short_echo_tag' => array('opening_tag_plus_echo_to_short_echo_tag'),
```
keradus: I would rather make one configurable fixer.
Like `short_echo_tag` with configuration which version to use.

keradus:
Valid question, anyone would like to have that fixer in the first place?

jtojnar:
We use PHP template syntax in [selfoss](https://github.com/SSilence/selfoss/pull/866). And apparently we aren’t the [only one](http://platesphp.com/).

jtojnar:
As for the tests failing on HHVM, that is problem of the test runner itself and should probably be fixed in a separate pull request.

keradus:
For that I would say one may want to have full open tag + echo in regular
php code and short echo in templates. Then, main challenge here is to
distinguish the files
-- 

Dariusz Rumiński

jtojnar:
Hopefully, this won’t be very frequent. [PSR-1](http://www.php-fig.org/psr/psr-1/#overview) even forbids defining symbols and having side effects in the same file. Additionally, the templates can use different configuration file.

src/FixerFactory.php
jtojnar: I pondered this for quite some time and I am still not this should be merged into `short_echo_tag` fixer. I understand the fixer as something that changes one aspect of code to one of specific styles in a certain context. Each of these fixers, however, is applied in different context and has a purpose of its own:

`no_echo_tag` fixer is applicable almost everywhere and aims to prevent (most often accidental) use of `<?=`. On the other hand `opening_tag_plus_echo_to_short_echo_tag` will only be useful for templates and is actually meant for harmonization.

I understand, though, that this is somewhat subjective distinction.

orlangur:
@keradus,
> Valid question, anyone would like to have that fixer in the first place?

I have no idea who is not using short echo tag [since PHP 5.4](http://softwareengineering.stackexchange.com/a/151694) and what was the story behind `NoShortEchoTagFixer` introduction. Just like there is no reason to use long array syntax unless you need to support PHP < 5.4 (when even PHP 5.6 is [not supported](http://php.net/supported-versions.php) by itself already).

> one may want to have full open tag + echo in regular php code and short echo in templates

What is "template"? How can `<?php echo` appear in regular php code not being template?

keradus:
good point, indeed, i didn't saw `<?php echo` outside of template ages ago.
template in very common understanding, like php file which only purpose is to provide html (or other) content

SpacePossum:
@keradus if this fixer would be added, how do you want the current conflict with `no_short_echo_tag` to be handled?

keradus:
@SpacePossum , do we want to accept such rule in first place ?

Answering question, what about one fixer, that will:
- for monolitic files - always use `<?php`
- for non monolitic files, depends on config - do nothing, use `<?php` or `<?=`

orlangur:
@keradus will such fixer also absorb what currently `NoShortEchoTagFixer` does?

keradus:
It would be extension of `NoShortEchoTagFixer` indeed.
But before modification, let us align direction.

SpacePossum:
_if_ going forward than indeed one fixer with config options seems the best way. Are there any examples of frameworks/communities that use the short hand notation?

orlangur:
Surely, just want to fully understand direction you propose.

I was always wondering why `NoShortEchoTagFixer` even exist and who preferred `<?php echo` over `<?=` and why :) (same for YodaConditionFixer while AntiYoda seems much more human-readable to me, but that's another story).

Looking at what fixers touching `echo` we have currently I propose to absorb proposed in this PR logic and `no_short_echo_tag`, `no_mixed_echo_print` into one configurable fixer `normalize_echo`.

SpacePossum:
(YodaConditionFixer is not in the repo, one proposed allows for anti-Yoda-fixing-by-config so please upvote that one ;) )

I would not move `no_mixed_echo_print` into fixers about open-tags, these are used in diff. context.

orlangur:
@SpacePossum what would be the downside of combining? (like, for which use case provided by single fixer behavior would be undesired)

With configuration like
```
'echo-style' => 'print',
'echo-tag-style' => 'short-echo-tag',
```

should do the trick for all currently supported cases and decrease collisions between different fixers touching same things as to me.

jtojnar:
> Are there any examples of frameworks/communities that use the short hand notation?

I know about us (selfoss) and anyone using [Plates](http://platesphp.com/).

jtojnar:
I do not really care about this fixer much, since templates can be only fixed in a limited fashion anyway. If someone wants to finish this, feel free.

keradus:
Thank you for your input here @jtojnar !

I am closing this PR as it was not finished and there were not big community willingness to have this rule.
I'm not saying this rule would not be accepted at all, but it's not ready in current state. If anyone is willing to have this rule feel free to read the discussion and finish PR.

mlocati:
See also #2990

mlocati:
See also #4493

PR #2522, opened 2017-02-08
Fix docs and small code issues

keradus:


src/AbstractFunctionReferenceFixer.php
```diff
@@ -67,7 +67,7 @@ protected function find($functionNameToSearch, Tokens $tokens, $start = 0, $end
         $matches = $tokens->findSequence($candidateSequence, $start, $end, false);
         if (null === $matches) {
             // not found, simply return without further attempts
-            return;
+            return null;
```
keradus: doc is claiming we will not return void ;)

tests/FixerFactoryTest.php
```diff
@@ -415,7 +415,10 @@ public function testFixerDefinitions(FixerInterface $fixer)
             $this->assertStringIsNotEmpty($definition->getConfigurationDescription(), sprintf('[%s] Configuration description is required.', $fixer->getName()));
             $default = $definition->getDefaultConfiguration();
             $this->assertInternalType('array', $default, sprintf('[%s] Default configuration must be an array.', $fixer->getName()));
-            $this->assertNotEmpty('array', $default, sprintf('[%s] Default configuration is required.', $fixer->getName()));
+
+            if (!in_array($fixer->getName(), array('general_phpdoc_annotation_remove', 'psr0'), true)) {
+                $this->assertNotEmpty($default, sprintf('[%s] Default configuration is required.', $fixer->getName()));
```
keradus: i don't wanna focus now on how to handle those exceptions better , but we were calling assertion wrongly

keradus:
@fabpot, fabbot is still wrong about phpunit assertion :(
previously repored here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2466#issuecomment-277249672

fabpot:
@keradus fixed, for real this time :) fabbot is green!

src/RuleSet.php
```diff
@@ -210,7 +210,7 @@ public function getRuleConfiguration($rule)
         }
 
         if ($this->rules[$rule] === true) {
```
SpacePossum: while you're here; yoda ;)

src/Test/IntegrationCase.php
```diff
@@ -93,11 +93,6 @@ public function __construct(
         $this->inputCode = $inputCode;
     }
 
-    public static function create()
```
SpacePossum: BC break?

tests/FixerFactoryTest.php
SpacePossum: nice catch :_  :+1:

SpacePossum:
thanks @fabpot :)

src/Test/IntegrationCase.php
keradus: i was thinking about it. yes and not. you cannot use this method. anyway. it will crash after calling it

keradus:
@fabpot thank you, hopefully ;)

src/Test/IntegrationCase.php
SpacePossum: hmm indeed :)
:+1: for removing

keradus: it is not an interface (we don't rely on the interface :D)
it is final (nobody could extend it)

the only way that removing of this method would break sth is when one actually run this method, but for that he already has a PHP crash. safe to remove

src/AbstractFunctionReferenceFixer.php
GrahamCampbell: It's the same thing coming out of the function in practice. the void type doesn't actually exist, it's just used by the type checker in php 7.1 to assert all return statements have no statement following them.

GrahamCampbell: A function typed to return void actually returns `null`.

keradus: If I would compare it to perl - PHP7 void make a function work in empty context, so no return data is passed - null or not. you cannot grab the outcome of that function.
I agree that from functional point of view there is no difference, still, I wanted to solve issue raised by SCA.

GrahamCampbell: > still, I wanted to solve issue raised by SCA.

The static analyzer is wrong IMO, since we both agree the semantics of what the function returns is the same, so by definition, it was already correct.

keradus: i still see semantic difference with `@return null` and `return null;` vs `return;`

GrahamCampbell: There is no semantic difference, as defined by PHP's semantics. Both functions have return value `NULL`.

keradus: > Fatal error: A void function must not return a value

anyway, great to see you being active. would be great on focus on sth meaningful instead, eg https://github.com/FriendsOfPHP/PHP-CS-Fixer/pulls/GrahamCampbell

GrahamCampbell: > Fatal error: A void function must not return a value

That's not the semantics of the function, that's just the syntax restrictions.

GrahamCampbell: What I mean is `f() = NULL` even if `f () -> void`.

Issue #2511, opened 2017-02-06
Remove useless method comment


OskarStark:
It would be cool to have fixer like this:

## Before
```php
    /**
     * Constructor.
     *
     * @param CurrencyDataTransformer $currencyTransformer
     */
    public function __construct(CurrencyDataTransformer $currencyTransformer)
    {
        $this->currencyTransformer = $currencyTransformer;
    }
```

## After
```php
    /**
     * @param CurrencyDataTransformer $currencyTransformer
     */
    public function __construct(CurrencyDataTransformer $currencyTransformer)
    {
        $this->currencyTransformer = $currencyTransformer;
    }
```

because `Constructor.` is useless imo, or is this still possible?

Best regards and thank you,
Oskar

SpacePossum:
I would say your sample PHPDoc is overcomplete as a whole ;) 

Anyway removing 'Constructor' (with or without trailing dot) as PHPDoc summary from constructors is fine by me :)
Removing `@return` annotations from constructors as well would be nice, but might be out-of-scope of the PR, although having a `NoUselessStuffInAPHPDocOfConstructorsFixer` could contain both ;)

keradus:
not only constructor is a case.
```
/**
 * Class Foo.
 */
final class Foo
{
    /**
     * Constructor
     */
    public function __construct() {}

    /**
     * Destructor
     */
    public function __destruct() {}
}
```

keradus:
@OskarStark , could you send a PR with proposal ?

OskarStark:
Is there a fixer which could be used as an example to start working on a PR?

SpacePossum:
I would go for;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.0/src/Fixer/Phpdoc/PhpdocAddMissingParamAnnotationFixer.php

that fixer
- searches for PHPDocs
- check if the doc belong to a function (new one should be methods only)
- check some things about the function (new one needs to check the name of the method only (case insensitive))
- it checks and parses the PHPDoc (new needs to do that as well)
- fixes the PHPDoc if needed

so I think most parts that you need are in there, however there is also a lot of stuff in there you can ignore as it is unrelated to the what the new fixer needs.

OskarStark:
> @OskarStark , could you send a PR with proposal ?

I checked the example, but to be honest, I currently don't have enough time to dive into it

OskarStark:
Could be helpful ;-)

[...]

Issue #1859, opened 2016-04-04
[FR] Remove @return from constructor PHPDocs


SpacePossum:
Any thoughts about of this fixer/change is welcome?

GrahamCampbell:
We already have a fixer to remove return vid annotations.

SpacePossum:
Do those fix cases like there?

```php
class A
{
    /**
     * @return A
     */
    public function __construct()
    {
    }
}
```

GrahamCampbell:
Well, that phpdoc is incorrect, because constructors return whatever you return from them, and that one there is `void`. I know there was some controversy over this for PHP 7.1, and I really hope their shitty idea of forcefully changing what those magic methods return doesn't actually make it throgh.

SpacePossum:
>Well, that phpdoc is incorrect,

indeed, that is why I would like to have fixer that removes the `@return` from constructors, these have no value IMHO

SvenRtbg:
@GrahamCampbell Constructors don't return anything - that's what makes them special in PHP. The instance of the class is already created, the constructor merely has the chance to initialize some fields or trigger some action, but you'll never see this code assinging `null` to $expectedNull:

    class ThatClass {
    public function __construct() {
        return null;
    }
    // ...
    $expectedNull = new ThatClass();

Some people add an `@return` annotation stating the return value would be "ThatClass", which is also wrong, but does point less into the wrong direction. However, IDEs evaluating annotations might get convinces that a constructor with a `@return` annotation but without an actual `return` statement might contain bugs.

I'd support the code fixer to offer the removal of `@return` annotations from constructors.

SpacePossum:
I'm closing this as I use a diff. system for keeping tracking on stuff I want to do one day when I've time. Please don't reopen, I like to have a small list of issues on GH to manage. Thanks!

PR #2526, opened 2017-02-09
[WIP] Add PhpdocMagicMethodReturnAnnotationFixer

SpacePossum:
[WIP] so please don't review as of yet, thanks!

ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1859

@OskarStark as follow up on https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2511#issuecomment-277920446 I opened the PR early (it is more or less a raw copy paste from a custom fixer of mine). The fixer here can be used as example as it finds constructors with PHPDocs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -502,6 +502,11 @@ Choose from the list of available rules:
 * **phpdoc_inline_tag** [@Symfony]
    | Fix phpdoc inline tags, make inheritdoc always inline.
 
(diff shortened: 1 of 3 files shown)


SpacePossum:
I'll follow up on this one when I've the time.

Issue #2514, opened 2017-02-07
RFC stop on first style violation in dry-run mode


keradus:
CI case
When we have a project that uses PHP CS Fixer on CI, it will fail on style violation (so, `--dry-run` mode).
For that, most probably we could stop on first violation instead of checking whole project - for sure as a separated CLI argument switch.

What do you think ?

SpacePossum:
like PHPUnit's `--stop-on-error` CLI option ? :+1:

keradus:
yeah, but rather `--stop-on-violation` ;)

GrahamCampbell:
> yeah, but rather --stop-on-violation ;)

How would that work? Stop once a file fails, or stop once a fixer wants to change something in a file.

keradus:
stop on detection of cs violation

GrahamCampbell:
Your implementation stops once all the fixers have completed running on a file, rather than as soon as one fixer has found something wrong.

keradus:
fine for me. then it won't be like guy stucks 10 times fixing same files, but different rules.

SpacePossum:
would be really be interested in a single rule violation in file over all violations in one file?

keradus:
cannot get it. too many `be` ;)

SpacePossum:
I'll rephrase: I think a user wants the tool to stop if a violation is found. If this happens the user wants the tool to lists _all_ violations in that file and not just of one rule.

hope its better ;)

keradus:
so, same like me

SpacePossum:
like you proposed, so not like your https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2514#issuecomment-279404413  ^_^

keradus:
Ah, I meant my original proposal is fine for me, as Graham's one would couse 10 times fixing for same file

keradus:
Then, please review #2532

SpacePossum:
closing as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2532 has been merged,
thanks for the work @keradus !

PR #2532, opened 2017-02-11
FixCommand - add stop-on-violation option

keradus:
Closing #2514


SpacePossum:
I think a line of text added to the README @ sub chapter `Usage` would be nice.
But with or without it; I'm :+1: nice idea, nice implementation

keradus:
Do you want to propose one ?

SpacePossum:
At sub chapter [`Usage`](https://github.com/FriendsOfPHP/PHP-CS-Fixer#usage) we provide information on the CLI options/arguments.
So I suggest to add at the end of this sub chapter;
`--stop-on-violation` stop PHP-CS-Fixer on the first detection of a file that needs to be fixed.

PR #2415, opened 2016-12-19
Add IsNullFixer

kalessil:
- [x] Re-create the fixer
- [x] Fix bugs discovered in #1304 (assignments, ternaries)
- [x] Add option for yoda/no-yoda fixing style

PR #2534, opened 2017-02-11
Enhancement: Provide edge case for is_null fixer

localheinz:
This PR

* [x] provides an edge case for the `is_null` fixer

Follows #2415.

💁‍♂️ Came across this when enabling it for a legacy code base where example code was

```php
if (is_null($news_keywords_meta) == true || empty($news_keywords_meta) == true) {
    return false;
}
```

Code was then fixed to 

```php
if (null === $news_keywords_meta == true || empty($news_keywords_meta) == true) {
    return false;
}
```

but change was not applied because it resulted in a linting error.

@kalessil Do you have an idea?

### DIFF
--- a/src/Fixer/LanguageConstruct/IsNullFixer.php
+++ b/src/Fixer/LanguageConstruct/IsNullFixer.php
@@ -96,10 +96,18 @@ final class IsNullFixer extends AbstractFixer implements ConfigurationDefinition
                 }
             }
 
+            /* edge cases: is_null() followed/preceded by ==, ===, !=, !==, <> */
+            $parentLeftToken = $tokens[$tokens->getPrevMeaningfulToken($isNullIndex)];
+            $parentRightToken = $tokens[$tokens->getNextMeaningfulToken($referenceEnd)];
+            $parentOperations = array(T_IS_EQUAL, T_IS_NOT_EQUAL, T_IS_IDENTICAL, T_IS_NOT_IDENTICAL);
+            $wrapIntoParentheses = $parentLeftToken->isGivenKind($parentOperations) || $parentRightToken->isGivenKind($parentOperations);
+
             if (!$isContainingDangerousConstructs) {
-                // closing parenthesis removed with leading spaces
-                $tokens->removeLeadingWhitespace($referenceEnd);
-                $tokens[$referenceEnd]->clear();
+                if (!$wrapIntoParentheses) {
+                    // closing parenthesis removed with leading spaces
+                    $tokens->removeLeadingWhitespace($referenceEnd);
+                    $tokens[$referenceEnd]->clear();
+                }
 
                 // opening parenthesis removed with trailing spaces
                 $tokens->removeLeadingWhitespace($matches[1]);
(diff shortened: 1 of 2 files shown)


kalessil:
@localheinz : not yet, will take a look at the case today

localheinz:
Thank you, @kalessil!

keradus:
nice case.
This is exactly why we have
> change was not applied because it resulted in a linting error.

;)

kalessil:
Okay, should be pretty easy to fix I'll make a PR to @localheinz feature branch when ready (somewhere at this week).

Note for myself: preserve `()` if `is_null` prepended/followed by `==, ===, !=, !==`

orlangur:
@kalessil, I was surprised seeing `IsNullFixer` as a risky one and now I guess here is the reason :)

Will it become safe after your change or there is some additional reasoning to keep it risky?

kalessil:
@orlangur : heh, not really because of this.

It's possible to override is_null (e.g. via` use function origin as alias;`) the fixer (and similar) is considered risky.

orlangur:
@kalessil, thanks for pointing out. Will stick to sniff forbidding `is_null` then rather then `php-cs-fixer --dry-run`.

SpacePossum:
>Note for myself: preserve () if is_null prepended/followed by ==, ===, !=, !==

for completeness, also check `<>`, not sure if more operators not to be checked.

So bunch of more tests:
```php
    array('<?php $x = is_null($x) <> true;'),
    array('<?php $x = is_null($x) === ($a);'),
    array(
        '<?php $x = $a/**/===#
    is_null($x);'
    ),
    array('<?php $a = is_null($a,$b,$c);'), // diff problem, param count not checked, lower prio but nice if resolved
```

good luck!

kalessil:
OMG people are still aware of `<>` ... Thank you @SpacePossum

SpacePossum:
ping @kalessil , no worries if you've no time :)

kalessil:
pong @SpacePossum. I'm burned out, need 1-2 weeks to recover and get back to this...

localheinz:
Take your time, @kalessil!

SpacePossum:
don't think about this one, take your time man and get well!

kalessil:
Here we go with the fix: https://github.com/localheinz/PHP-CS-Fixer/pull/5

keradus:
I saw 10 commits from @kalessil , now they are gone. what's going on ?

localheinz:
@keradus 

Hang on a second!

keradus:
was I too fast ? :P

localheinz:
Messed something up here, should be all good now!

keradus:
failing, failing everywhere

localheinz:
@keradus 

Not sure if I messed something up here, how about we scrap the entire thing and then let @kalessil take over with a new PR? That would give @kalessil the freedom to work on this without the need to open a PR against my branch. The initial commit with the failing test case can easily be cherry-picked.

What do you think, @kalessil?

keradus:
Fine for me,
Whatever works for both of you ;)

kalessil:
Oki-doki: #2627

PR #2535, opened 2017-02-12
Make .gitignore entries more specific

julienfalque:
- `vendor` is supposed to be a directory;
-  ignored targets are at the project's root only, also prevents some `.php_cs` files in the tests fixtures to be covered by the `.gitignore` while still being tracked.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-.php_cs
-composer.lock
-phpunit.xml
-vendor
+/.php_cs
+/composer.lock
+/phpunit.xml
+/vendor


keradus:
btw, dir is a file with extra attribute on FS level, so it doesn't matter would there be trailing `/` or not

julienfalque:
I know what directories are ;)

Git has a special handling for ignore entries that end with a `/` as these will only match directories instead of any file. So this entry can be read like _"Ignore the `vendor` directory"_ instead of _"Ignore `vendor`"_ and it looks better to me. But as the result here will be the same, I can remove it if you prefer.

SpacePossum:
out of curiosity, would git ignore a file called `vendor` now?

julienfalque:
If we keep the ending `/`: no. But is it to ever happen?

SpacePossum:
its nice if the file 'vendor' is excluded, as it should not exist so :+1:

julienfalque:
Rebased onto `2.1` and removed ending `/` to vendor entry.

keradus:
> its nice if the file 'vendor' is excluded, as it should not exist so 

but `vendor` dir is existing dir for a reason - composer will create it. that is why we exclude it.
`vendor` file is not there, as well as `foobarbaz` file. no need for special handling of it

keradus:
Thank you @julienfalque.

PR #2536, opened 2017-02-12
DX suggest update to v2 on composer level

keradus:


SpacePossum:
:+1:  ship :)

keradus:
nope, sadly, suggesting installing same package is not working

PR #2541, opened 2017-02-14
README.rst - provide download link for latest version

keradus:


SpacePossum:
I prefer. `latest` over `newest`, but thats just me. So I'm :+1: any way :)

Issue #2543, opened 2017-02-15
Rule set subtraction doesn't work


Bilge:
PHP 5.6.18
PHP CS Fixer 2.1.0 by Fabien Potencier and Dariusz Ruminski

The documentation seems to suggest it is possible to write `--rules=@Symfony,-@PSR1`. Although it is not clearly specified what this means, I expect it would use the Symfony rule set, less the PSR-1 set. However, this configuration does not work: overlapping PSR-1 rules remain enabled.

keradus:
Could you please send a PR with test proving the bug and hopefully a bugfix ?

Bilge:
I don't know how to do that.

keradus:
Closing due to merge of #2545

PR #2545, opened 2017-02-15
RuleSet - change resolvement

SpacePossum:
for ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2543

### DIFF
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -16,20 +16,15 @@ use PhpCsFixer\Config;
 use PhpCsFixer\Console\Command\FixCommand;
 use PhpCsFixer\Console\ConfigurationResolver;
 use PhpCsFixer\Finder;
-use PhpCsFixer\RuleSet;
-use PhpCsFixer\Test\AccessibleObject;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * @author Katsuhiro Ogawa <ko.fivestar@gmail.com>
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
- * @author SpacePossum
  *
  * @internal
  *
  * @covers \PhpCsFixer\Console\ConfigurationResolver
- * @covers \PhpCsFixer\RuleSet::resolveSet
- * @covers \PhpCsFixer\RuleSet::resolveSubset
  */
 final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
 {
@@ -948,148 +943,6 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
         );
     }
 
-    /**
-     * @param array $expected
-     * @param array $rules
-     *
-     * @dataProvider provideResolveRulesCases
-     */
-    public function testResolveRules(array $expected, array $rules)
-    {
-        $ruleSet = $this->createRuleSetToTestWith($rules);
-
-        $this->assertSameRules($expected, $ruleSet->getRules());
-    }
-
-    public function provideResolveRulesCases()
-    {
-        return array(
-            '@Foo + C\' -D' => array(
-                array('A' => true, 'B' => true, 'C' => 56),
-                array('@Foo' => true, 'C' => 56, 'D' => false),
-            ),
-            '@Foo + @Bar' => array(
-                array('A' => true, 'B' => true, 'D' => 34, 'E' => true),
-                array('@Foo' => true, '@Bar' => true),
-            ),
-            '@Foo - @Bar' => array(
-                array('B' => true),
-                array('@Foo' => true, '@Bar' => false),
-            ),
-            '@A - @E (set in set)' => array(
-                array('AA' => true), // 'AB' => false, 'AC' => false
-                array('@A' => true, '@E' => false),
-            ),
-            '@A + @E (set in set)' => array(
-                array('AA' => true, 'AB' => '_AB', 'AC' => 'b', 'Z' => true),
-                array('@A' => true, '@E' => true),
-            ),
-            '@E + @A (set in set) + rule override' => array(
-                array('AC' => 'd', 'AB' => true, 'Z' => true, 'AA' => true),
-                array('@E' => true, '@A' => true, 'AC' => 'd'),
-            ),
-            'nest single set' => array(
-                array('AC' => 'b', 'AB' => '_AB', 'Z' => 'E'),
-                array('@F' => true),
-            ),
-            'Set reconfigure rule in other set, reconfigure rule.' => array(
-                array(
-                    'AA' => true,
-                    'AB' => true,
-                    'AC' => 'abc',
-                ),
-                array(
-                    '@A' => true,
-                    '@D' => true,
-                    'AC' => 'abc',
-                ),
-            ),
-            'Set reconfigure rule in other set.' => array(
-                array(
-                    'AA' => true,
-                    'AB' => true,
-                    'AC' => 'b',
-                ),
-                array(
-                    '@A' => true,
-                    '@D' => true,
-                ),
-            ),
-            'Set minus two sets minus rule' => array(
-                array(
-                    'AB' => true,
-                ),
-                array(
-                    '@A' => true,
-                    '@B' => false,
-                    '@C' => false,
-                    'AC' => false,
-                ),
-            ),
-            'Set minus two sets' => array(
-                array(
-                    'AB' => true,
-                    'AC' => 'a',
-                ),
-                array(
-                    '@A' => true,
-                    '@B' => false,
-                    '@C' => false,
-                ),
-            ),
-            'Set minus rule test.' => array(
-                array(
-                    'AA' => true,
-                    'AC' => 'a',
-                ),
-                array(
-                    '@A' => true,
-                    'AB' => false,
-                ),
-            ),
-            'Set minus set test.' => array(
-                array(
-                    'AB' => true,
-                    'AC' => 'a',
-                ),
-                array(
-                    '@A' => true,
-                    '@B' => false,
-                ),
-            ),
-            'Set to rules test.' => array(
-                array(
-                    'AA' => true,
-                    'AB' => true,
-                    'AC' => 'a',
-                ),
-                array(
-                    '@A' => true,
-                ),
-            ),
-            '@A - @C' => array(
-                array(
-                    'AB' => true,
-                    'AC' => 'a',
-                ),
-                array(
-                    '@A' => true,
-                    '@C' => false,
-                ),
-            ),
-            '@A - @D' => array(
-                array(
-                    'AA' => true,
-                    'AB' => true,
-                ),
-                array(
-                    '@A' => true,
-                    '@D' => false,
-                ),
-            ),
-        );
-    }
-
     public function testResolveRulesWithUnknownRules()
     {
         $this->setExpectedException(
@@ -1183,45 +1036,6 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($resolver->getUsingCache());
         $this->assertNull($resolver->getCacheFile());
         $this->assertSame('xml', $resolver->getReporter()->getFormat());
-    }
-
-    private function createRuleSetToTestWith(array $rules)
-    {
-        static $testSet = array(
-            '@A' => array(
-                'AA' => true,
-                'AB' => true,
-                'AC' => 'a',
-            ),
-            '@B' => array(
-                'AA' => true,
-            ),
-            '@C' => array(
-                'AA' => false,
-            ),
-            '@D' => array(
-                'AC' => 'b',
-            ),
-            '@E' => array(
-                '@D' => true,
-                'AB' => '_AB',
-                'Z' => true,
-            ),
-            '@F' => array(
-                '@E' => true,
-                'Z' => 'E',
-            ),
-            '@Foo' => array('A' => true, 'B' => true, 'C' => true, 'D' => 12),
-            '@Bar' => array('A' => true, 'C' => false, 'D' => 34, 'E' => true, 'F' => false),
-        );
-
-        $ruleSet = new RuleSet();
-        $reflection = new AccessibleObject($ruleSet);
-        $reflection->setDefinitions = $testSet;
-        $reflection->set = $rules;
-        $reflection->resolveSet();
-
-        return $ruleSet;
     }
 
     private function assertSameRules(array $expected, array $actual, $message = '')
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -984,6 +984,14 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
                 array('AA' => true, 'AB' => '_AB', 'AC' => 'b', 'Z' => true),
                 array('@A' => true, '@E' => true),
             ),
+            '@E + @A (set in set) + rule override' => array(
+                array('AC' => 'd', 'AB' => true, 'Z' => true, 'AA' => true),
+                array('@E' => true, '@A' => true, 'AC' => 'd'),
+            ),
+            'nest single set' => array(
+                array('AC' => 'b', 'AB' => '_AB', 'Z' => 'E'),
+                array('@F' => true),
+            ),
             'Set reconfigure rule in other set, reconfigure rule.' => array(
                 array(
                     'AA' => true,
@@ -1198,6 +1206,10 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
                 '@D' => true,
                 'AB' => '_AB',
                 'Z' => true,
+            ),
+            '@F' => array(
+                '@E' => true,
+                'Z' => 'E',
             ),
             '@Foo' => array('A' => true, 'B' => true, 'C' => true, 'D' => 12),
             '@Bar' => array('A' => true, 'C' => false, 'D' => 34, 'E' => true, 'F' => false),
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -977,11 +977,11 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
                 array('@Foo' => true, '@Bar' => false),
             ),
             '@A - @E (set in set)' => array(
-                array('AA' => true, 'AB' => true), // , 'AC' => false
+                array('AA' => true), // 'AB' => false, 'AC' => false
                 array('@A' => true, '@E' => false),
             ),
             '@A + @E (set in set)' => array(
-                array('AA' => true, 'AB' => true, 'AC' => 'b', 'Z' => true),
+                array('AA' => true, 'AB' => '_AB', 'AC' => 'b', 'Z' => true),
                 array('@A' => true, '@E' => true),
             ),
             'Set reconfigure rule in other set, reconfigure rule.' => array(
@@ -1196,6 +1196,7 @@ final class ConfigurationResolverTest extends \PHPUnit_Framework_TestCase
             ),
             '@E' => array(
                 '@D' => true,
+                'AB' => '_AB',
                 'Z' => true,
             ),
             '@Foo' => array('A' => true, 'B' => true, 'C' => true, 'D' => 12),
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -16,6 +16,8 @@ use PhpCsFixer\Config;
 use PhpCsFixer\Console\Command\FixCommand;
 use PhpCsFixer\Console\ConfigurationResolver;
 use PhpCsFixer\Finder;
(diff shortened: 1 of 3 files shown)


Bilge:
This test seems to pass, so does that prove my report is incorrect?

keradus:
I would say so. If not, please extend your report. Maybe sth is missing

Bilge:
In my case I was subtracting `@PSR2`, not `@PSR1`. I don't know if that should make any difference.

keradus:
nope, mechanism is the same

SpacePossum:
Subtracting set from set with same rule but with diff. config for that rule does not work as expected.
Removing rule from set does work.
Please review!

tests/Console/ConfigurationResolverTest.php
```diff
@@ -868,6 +869,123 @@ public function testResolveRulesWithOption()
         );
     }
 
+    /**
+     * @param string[] $rules
+     * @param string[] $negativeRules
+     *
+     * @dataProvider provideSubtractionCases
+     */
+    public function testResolveRulesWithNegativeSet(array $rules, array $negativeRules)
```
keradus: Please redesign the test, now it's messy.
What is the expected value here? nobody knows.

Input should be `string $rules, array $expectedRules`

keradus:
Sorry, but without https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2545/commits/24981c9b23792403fea2299cda683c1d17f58c85#r105905980 I cannot evaluate is the generated ruleset is properly resolved

tests/Console/ConfigurationResolverTest.php
SpacePossum: I'll have a look later, IIRC I made the test like this to prevent a lot of repeating code and being depended too much on what rules are in the current sets, as it would make the tests fragile.

keradus: 1. maybe possible to not rely on real seats, but register mocked ones (with only few rules) ?
2. even moving "semi-resolving" from test method to data provider would make test method getting raw `input, expected` params, test method should avoid having code with logic

SpacePossum: Looking at the code; to do it the nice way we should move the rule set definitions out of the `RuleSet` class and introduce something like `RuleSetDefinitions` or `RuleSetRepository` and inject that into `RuleSet`. Maybe even better to introduce a `RuleSetResolver` and inject the `RuleSet` and `RuleSetDefiniton` and let the resolver figure out what the set of rules should be.
Currently I don't have the time to make a POC on this though, so maybe park this PR for a bit.

keradus: I would not introduce such a big change into this PR.
Still, I do believe that it's possible to move logic outside of `test*` method

SpacePossum: Re-wrote, but the number of reflection needed indicate something might be off.
Let me know if you've another strat. in mind for the tests.

tests/Console/ConfigurationResolverTest.php
```diff
@@ -961,12 +1070,104 @@ public function testResolveConfigFileOverridesDefault()
...
+        $setDefinitionsReflectionProperty->setAccessible(true);
+        $setDefinitionsReflectionProperty->setValue($ruleSet, $testSet);
+
+        $setReflectionProperty = $reflection->getProperty('set');
+        $setReflectionProperty->setAccessible(true);
+        $setReflectionProperty->setValue($ruleSet, $rules);
+
+        $resolveSetReflectionMethod = $reflection->getMethod('resolveSet');
+        $resolveSetReflectionMethod->setAccessible(true);
+        $resolveSetReflectionMethod->invoke($ruleSet);
```
keradus: what about using `AccessibleObject` ?

tests/Console/ConfigurationResolverTest.php
```diff
@@ -961,12 +1070,104 @@ public function testResolveConfigFileOverridesDefault()
...
+
+        return $ruleSet;
+    }
+
     private function assertSameRules(array $expected, array $actual, $message = '')
```
keradus: why can't we just compare 2 arrays ? I don't think we really need that verbose message

SpacePossum:
all done

tests/Console/ConfigurationResolverTest.php
```diff
@@ -868,6 +870,100 @@ public function testResolveRulesWithOption()
...
+            'Set reconfigure rule in other set.' => array(
+                array(
+                    'AA' => true,
+                    'AB' => true,
+                    'AC' => 'b',
+                ),
+                array(
+                    '@A' => true,
+                    '@D' => true,
+                ),
```
keradus: please add input ruleset:
```
                array(
                    '@A' => true,
                    '@D' => false,
                ),
```

SpacePossum: :+1: done

tests/Console/ConfigurationResolverTest.php
```diff
@@ -961,6 +961,26 @@ public function provideResolveRulesCases()
                     '@A' => true,
                 ),
             ),
+            '@A - @C' => array(
```
keradus: in `@A` rule `AA` is on
in `@C` rule `AA` is off

`@A - @C` should still has `AA` on, as we didn't extract any set that contains this rule

tests/Console/ConfigurationResolverTest.php
```diff
@@ -961,6 +961,26 @@ public function provideResolveRulesCases()
...
+                array(
+                    'AB' => true,
+                    'AC' => 'a',
+                ),
+                array(
+                    '@A' => true,
+                    '@C' => false,
+                ),
+            ),
+            '@A - @D' => array(
```
keradus: in `@A` rule `AC=a`
in `@D` rule `AC=b`

`@A - @D` should still has `AC=a`, as nobody turn off this configuration and we allow one set to overwrite configuration of rule from different set

tests/Console/ConfigurationResolverTest.php
SpacePossum:
```
            '@A' => array(
                'AA' => true,
                'AB' => true,
                'AC' => 'a',
            ),
```
minus
```
            '@C' => array(
                'AA' => false,
            ),
```

AA, AB, AC - AA = AB, AC

tests/Console/ConfigurationResolverTest.php
SpacePossum: ? how could you turn of rule AC than?

tests/Console/ConfigurationResolverTest.php
keradus: no, as
`(new Ruleset([ 'AA' => false ]))->getRules() === (new Ruleset([]))->getRules()`

tests/Console/ConfigurationResolverTest.php
keradus: `[ AC => false ]`

tests/Console/ConfigurationResolverTest.php
SpacePossum:
```
'AA' => true,
'AB' => true,
'AC' => 'a',
```
minus
```
'AA' => false,
```
so 'AA' should be out as 'AA => false' is the way to turn it off (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2545#discussion_r107535379)?

keradus: same for both.

`@X = [AA => true]`, `@Y = [AA => false]`

`@X,@Y => [AA => true] + [AA => false]` => `[AA => false]`
but
`@X,-@Y => [AA => true] - [AA => false]` => `[AA => true]`

SpacePossum: >@X,-@Y => [AA => true] - [AA => false] => [AA => true]

keradus: OK, then I think we need more opinion on this case:

Should `[FOO => true, BAR =>false]` means the same as simply `[FOO => true]` or not, when it is part of ruleset definition like PSR or PHPMigration ?

julienfalque: IMO, if `@A - @B` means "rules from A that are not in B", then having an explicitly disabled rule in B is like not having it at all. I would expect `['foo' => true] - ['foo' => false]` to result in `['foo' => true]`.

SpacePossum: > explicitly disabled rule in B is like not having it at all.

and than you have it in the result anyway?

julienfalque: Yep.

SpacePossum: Alright, I've reverted all changes to the resolve logic, leaving the test.
Oddly it only shows the original reported issue, however I'm confused if this is an issue or not.
I put this PR to a POC/question and leave it to others to resolve because after re-doing this PR N-times I'm blind for what the result should be.

keradus: I don't like squashing, now we have only one commit and don't see the history, it's harder to see it and play with it.

so, guys, let say I want to make `@Foo` ruleset, and then `@Bar` ruleset, which is the same as `@Foo` but without one rule (and having some extra rule). How could I define `@Bar` set?
```
'@Bar' => {
    '@Foo` => true,
    'add_this_rule` => true,
    'remove_this_rule` => ???
},
```
?

julienfalque: > How could I define `@Bar` set?

Should be:
```
'@Bar' => {
    '@Foo` => true,
    'add_this_rule` => true,
    'remove_this_rule` => false,
},
```
right? Isn't this the current way to do it?

> then, what would be the difference between `@X,@Y` and `@X,-@Y` ?

That's why I think explicitly disabled rules should be ignored in such case.

keradus: ah, right, this is current way, issue is when one subtract not a single rule, but a set (that disable rule as well)
```
{ '@Foo': true, } : @Foo' => { 'AA' => true, 'NOOP' => true, }
{ '@Bar': true, } : @Bar => { 'AA' => false, 'NOOP' => true, }
{ '@Bar': false, } : -@Bar => { 'AA' => true, NOOP' => false, } // as direct revert
```

Remember that `,` means `take later`, so:
- `{ 'X' => true, 'X' => true } = { 'X' => true }`
- `{ 'Y' => true, 'Y' => false } = { 'Y' => false }`
- `{ 'Z' => false, 'Z' => true } = { 'Z' => true }`

IMO:
- `{ '@Foo': true, '@Bar': true } => { 'NOOP' => true, }`, as:
`{ '@Foo': true, '@Bar': true } = { 'AA' => true, 'NOOP' => true, 'AA' => false, 'NOOP' => true, } => {'NOOP' => true, }`
- `{ '@Foo': true, '@Bar': false } => { 'AA' => true, }`, as:
`{ '@Foo': true, '@Bar': false } => { '@Foo': true, '-@Bar': true, } => { 'AA' => true, 'NOOP' => true, 'AA' => true, 'NOOP' => false, } => { 'NOOP' => true, }`

keradus: ping @SpacePossum @julienfalque

SpacePossum: I'm still breaking my head around this, hope to write something tomorrow

julienfalque: @keradus I think I agree with your last comment except for this point:

```
{ '@Bar': false, } : -@Bar => { 'AA' => true, NOOP' => false, } // as direct revert
```

IMO, if your configuration only includes a disabled ruleset with no further rules, the resolved result should be an empty list. In your exemple, this enable `AA` rule.

keradus: let that something be discussion, not the code. better to discusses in advance

keradus: @julienfalque , but then it would be like:
`{ '@Foo': true, '@Bar': false } => { '@Foo': true, }` :| weird...

julienfalque: Not to me, I read it like this:
* `{ '@Foo': true, '@Bar': false }` => rules from `@Foo` that are not in `@Bar` => some rules from `@Foo`;
* `{ '@Bar': false }` => rules from nothing that are not in `@Bar` => nothing

keradus:

tests/Console/ConfigurationResolverTest.php
pilgerone: I agree with @keradus on `[ AC => false ]` being the most consistent way to turn AC off right now. That is considering we don't want to change the way rules that contain arrays are set.

But what if we set rules that contain arrays the same way we set regular rules? For example:
in `@A` rule `AC=['a' => true]`
in `@D` rule `AC=['b' => true]`

It would be more consistent and would allow to fine tune which rules are turned on and off. For example:
in `@A` rule `AC=['a' => true]`
in `@D` rule `AC=['b' => true, 'a' => false]`
This would work as @SpacePossum proposed and be fairly explicit.

keradus: @pilgerone , go for main discussion:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2545#discussion_r107527563

it has more complicated cases.

tests/Console/ConfigurationResolverTest.php
keradus:
```
{ '@Bar': false, } : -@Bar => { 'AA' => true, NOOP' => false, } // as direct revert
```

this is revertion by simple rule
> if your configuration only includes a disabled ruleset with no further rules, the resolved result should be an empty list. In your exemple, this enables AA rule.

it would make exception of how to understand sets and resolve them.
if set contains only rule disabling and you don't want to disable or enable them - don't use the set

pilgerone: **It is** more complicated. :no_mouth: So I'm dropping in kinda awkwardly in hope to help however I can.

Okay, so I partially agree with @julienfalque on `@Foo - @Bar` should read as `remove from @Foo all rules that are present in @Bar`. E.g.: considering:

`@Foo => { 'A' => true, 'B' => true, 'C' => true}`
`@Bar => { 'A' => true, 'C' => false}`
then `{@Foo - @Bar} => {'B' => true, 'C' => true}`

In cases of subtraction, the most coherent, imho, would be to ignore rules set to false within the subrahend. Therefore the `'C' => false` being ignored. I imagine someone doing something like `@Symfony - @PSR2` and not expecting to have some rule turned off in `@PSR2` being automatically turned on.

So it would be like mathematical union, intersection and subtraction between groups: 
shaded-part-represents-b-a

I believe the effect of setting a group of rules as `false` should be `disable all rules within the group`. So, in cases of sequence definition, I'd go with:

`@Foo => { 'A' => true, 'B' => true, 'C' => true}`
`@Bar => { 'A' => true, 'C' => false}`
then `{@Foo = true,  @Bar = false} => {'B' => true}`

That being said, isn't the subtraction between groups of rules a bit of an edge case? It could be accomplished setting the rules independently. Seems like a bit of over engineering. Or maybe someone could provide some use cases?

What do you guys think?

SpacePossum: My proposal was and is your later example;
```
@Foo => { 'A' => true, 'B' => true, 'C' => true}
@Bar => { 'A' => true, 'C' => false}
then {@Foo = true, @Bar = false} => {'B' => true}
```

So; `@A - @B` being; remove from `@A` all rules in `@B`, no matter how the rule is configured in `@B`.
This also solves this;
```
@Foo => { 'A' => true, 'B' => true, 'C' => true}
@Bar => { 'A' => true, 'D' => false}
```
then `@Foo - @Bar = {'B' => true, 'C' => true}` , i.e. `D` is _not_ enabled as result of subtracting a negative rule (i.e. `-- !== +`)
This would be consistent to;
```
@Foo => { 'A' => 'foo'}
@Bar => { 'A' => 'bar'}
```
then `@Foo - @Bar = {}` , which currently resolves to `{'A' => 'foo'}`

The last one is important, because if you want to disable rule `A` when using `@Foo` than currently it is not possible using _subtraction_ from another _set_. 
(There is other way; subtract the rule only)
So should one than _add_ a set where the rule is _negative_ (i.e. `false`; `+- = -`)?

pilgerone: >```
> @Foo => { 'A' => 'foo'}
> @Bar => { 'A' => 'bar'}
> ```
> then `@Foo - @Bar = {}` , which currently resolves to `{'A' => 'foo'}`

This feels really weird to me. It's like subtracting an apple from an orange. I'd go with  the current resolution: `A` would only be removed if values match. For example:
```
@Foo => { 'A' => 'foobar'}
@Bar => { 'A' => 'foobar'}
@Foo - @Bar => {}
```
But
```
@Foo => { 'A' => 'foobar'}
@Bar => { 'A' => 'bar'}
@Foo - @Bar => { 'A' => 'foobar'}
```
Since it wouldn't be able to find a `'A' => 'bar'` to remove.
```
@Foo => { 'A' => ['foo', 'bar']}
@Bar => { 'A' => ['bar']}
@Foo - @Bar => { 'A' => ['foo']}
```
Would make perfect sense, though.

And for me there is a difference between `@Foo - @Bar` and `@Foo = true, @Bar = false`. The first removes from `@Foo` all rules set to true in `@Bar` (consider as if a rule set to false in `@Bar` doesn't exist). The later applies all rules defined in `@Foo` **and then** sets all rules present in `@Bar` as false. The first is an operation while the second is an assignment.

To minimize confusion, I believe it would be better to have only assignments and no subtractions. Unless there're some use cases in which ruleSet subtraction would make so much sense it could justify the whole mess. :man_shrugging:

SpacePossum:
```
@Foo => { 'A' => ['foo', 'bar']}
@Bar => { 'A' => ['bar']}
@Foo - @Bar => { 'A' => ['foo']}
```
the issue with it is, that for `A` both `[foo, bar]` and `[bar]` might be valid config, however this doesn't guarantee `[foo]` is a valid option. This adds another layer of complexity of which I wonder is worth to resolve.

If we would only remove rules if the config match exactly, than we are stuck with some odd cases.
Like;
```
@SF{'header' => 'some SF copyright statment'}
-
@CommentFixers{'header' => false}
```

would still result in adding the header, which I think is odd.

pilgerone: Analyzing the issue more calmly today I think there was a bit of a misunderstanding.

We don't have a `@Foo -@Bar` anywhere. What we have is a `@Foo, -@Bar` (notice the comma). Which is the command line way of saying `Use @Foo but disable @Bar`. Which is equivalent to `{'@Foo': true, '@Bar': false}` on a json or a 
```
    ->setRules(array(
        '@Foo' => true,
        '@Bar' => false
    )
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->exclude('tests/Fixtures')
            ->in(__DIR__)
    );
```

So there is no operation between sets, just plain old **enable this**, **disable that**. I couldn't find any Rule Set on **RuleSet.php** with a `'rule' => false` either. Allowing a Rule Set to forcibly disable a specific rule seems a bit unnecessary and can lead to this sort of confusion we're having, imho.

@SpacePossum, then a `--rules=@SF, @CommentFixers` would  behave as you intend and there would be no such thing as a `--rules=@SF minus @CommentFixers`. That is considering an hypothetical Rule Set that has a `'header' => false` arbitrarily.

What do you guys think?

SpacePossum:
After considering all input for reviewers and other members I made a first version of the new code for the set resolving. I know this might not be what everyone wants or expect, however I feel this is the most consistent way of resolving. 
This leave the following for this PR
- reviews (please do not request logic changes, that discussion has been done)
- see why some integration tests now fail
- make a call on the release version, is this a bug fix or feature, I can see both sides but lean to bug fix after all

tests/Console/ConfigurationResolverTest.php
```diff
@@ -943,6 +948,140 @@ public function testResolveRulesWithOption()
...
+    {
+        $ruleSet = $this->createRuleSetToTestWith($rules);
+
+        $this->assertSameRules($expected, $ruleSet->getRules());
+    }
+
+    public function provideResolveRulesCases()
+    {
+        return array(
+            '@Foo + C\' -D' => array(
```
keradus: why extra `'` in the middle ?

tests/Console/ConfigurationResolverTest.php
```diff
@@ -943,6 +948,140 @@ public function testResolveRulesWithOption()
...
+     * @param array $expected
+     * @param array $rules
+     *
+     * @dataProvider provideResolveRulesCases
+     */
+    public function testResolveRules(array $expected, array $rules)
+    {
+        $ruleSet = $this->createRuleSetToTestWith($rules);
+
+        $this->assertSameRules($expected, $ruleSet->getRules());
```
keradus: shouldn't this test be in `RuleSetTest` ?

SpacePossum: moved :)
as a bonus, with the new added logic for testing the resolving of sets, more logic of the RuleSet class can be tested (so no logic changes, just more test coverage)

keradus:
Thank you @SpacePossum.

PR #2548, opened 2017-02-18
Improved specificity of gitignore rules

Bilge:
These rules all specify files in the root directory so they should be preceded by a slash (`/`) so they do not match in any other directory. `vendor` is a directory specifier so it should be suffixed by a slash (`/`) so it does not match files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
-.php_cs
-composer.lock
-phpunit.xml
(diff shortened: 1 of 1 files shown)


keradus:
Please rebase on top of current master.

Bilge:
Someone already did this...? OK, then.

PR #2549, opened 2017-02-18
Travis maintainability improvements

Bilge:
This PR keeps code lines under 120 chars and simplifies some syntax to improve maintainability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,10 +46,13 @@ cache:
         - $HOME/.composer/cache
 
 before_install:
+    # Stop on first error.
+    - set -e
+
     # check phpdbg
     - phpdbg --version 2> /dev/null || { echo 'No phpdbg'; export TASK_TESTS_COVERAGE=0; }
 
-    # turn off XDebug
+    # Turn off Xdebug. Ignores failure on HHVM builds due to missing xdebug.ini.
     - phpenv config-rm xdebug.ini || true
 
     # validate tasks configuration
@@ -104,7 +107,7 @@ script:
 
     - if ((TASK_SCA)); then git checkout . -q; fi
 
-    - if (($TASK_CS)); then php php-cs-fixer --diff --dry-run -v fix; fi
+    - if ((TASK_CS)); then php php-cs-fixer --diff --dry-run -v fix; fi
 
 after_success:
     - if ((TASK_TESTS_COVERAGE)); then vendor/bin/coveralls -v; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ before_install:
     - if ((!TASK_TESTS & TASK_TESTS_COVERAGE)); then travis_terminate 1; fi
 
     # for building a tag release we don't need to run SCA tools, collect code coverage or self-fix CS
-    - $TRAVIS_TAG && export TASK_SCA=0 TASK_TESTS_COVERAGE=0 TASK_CS=0
+    - if [[ -n $TRAVIS_TAG ]]; then export TASK_SCA=0 TASK_TESTS_COVERAGE=0 TASK_CS=0; fi
 
     # Composer: boost installation
     - travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,7 +56,7 @@ before_install:
     - phpenv config-rm xdebug.ini || true
 
     # validate tasks configuration
-    - if ((!TASK_TESTS & TASK_TESTS_COVERAGE)); then travis_terminate 1; fi
+    - if ((!TASK_TESTS && TASK_TESTS_COVERAGE)); then travis_terminate 1; fi
 
     # for building a tag release we don't need to run SCA tools, collect code coverage or self-fix CS
     - if [[ -n $TRAVIS_TAG ]]; then export TASK_SCA=0 TASK_TESTS_COVERAGE=0 TASK_CS=0; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -50,21 +50,22 @@ before_install:
     - phpdbg --version 2> /dev/null || { echo 'No phpdbg'; export TASK_TESTS_COVERAGE=0; }
 
     # turn off XDebug
-    - phpenv config-rm xdebug.ini || return 0
+    - phpenv config-rm xdebug.ini || true
 
     # validate tasks configuration
-    - if [ $TASK_TESTS == 0 ] && [ $TASK_TESTS_COVERAGE == 1 ]; then travis_terminate 1; fi
+    - if ((!TASK_TESTS & TASK_TESTS_COVERAGE)); then travis_terminate 1; fi
```
keradus: `((` is for arithmetic operations, this is not that case
`[` is standard POSIX syntax for tests, please keep it.

.travis.yml
```diff
@@ -50,21 +50,22 @@ before_install:
...
 
     # for building a tag release we don't need to run SCA tools, collect code coverage or self-fix CS
-    - if [ $TRAVIS_TAG ]; then export TASK_SCA=0; fi
-    - if [ $TRAVIS_TAG ]; then export TASK_TESTS_COVERAGE=0; fi
-    - if [ $TRAVIS_TAG ]; then export TASK_CS=0; fi
+    - $TRAVIS_TAG && export TASK_SCA=0 TASK_TESTS_COVERAGE=0 TASK_CS=0
```
keradus: :-1: for using AND operator to conditionally execute statements

.travis.yml
Bilge: http://stackoverflow.com/a/26920580/1041515

keradus:
:-1:
having multiple commands in conditional bash script written that way will not fail the build if one of the command fails

.travis.yml
keradus: I'm not saying it won't run, I'm saying that `[` is supposed to be use for logic checks.

Bilge:
@keradus To which block of code are you referring when you state command failure will not fail the build?

keradus:
the ones with multiple commands, as said.

http://steven.casagrande.io/articles/travis-ci-and-if-statements/

Bilge:
But there are none with multiple commands, except when which calls `travis_terminate`, in which the exit code of that branch doesn't matter.

.travis.yml
Bilge: If that were the case, "arithmetic" expressions would not include logic operators. Logic is not arithmetic.

.travis.yml
```diff
@@ -74,18 +78,39 @@ install:
...
+            -not -path './.git/*' \
+            -not -path './vendor/*' \
+            -not -path './tests/Fixtures/*' \
+            -exec egrep -nH ' $' {} \;
+          )"
+        if [[ -n $files_with_trailing_spaces ]]; then
+          echo "$files_with_trailing_spaces"
+          travis_terminate 1
+        else echo No trailing spaces detected.
+        fi
```
keradus: everywhere here, there are multiple commands

Bilge: I have added `set -e` to the top of the script.

keradus: you did added multiple commands, eg in `TASK_SCA` task

Bilge:
It seems the only code you like is your own.

keradus:
It's good I omitted any reasoning and arguments ;)

Bilge:
You *have* omitted any **objective** reasoning. The only *reasoning* supplied is purely subjective. 

>👎 for executing multiple commands in condition

Non-issue when `set -e` is in effect.

>👎 for using set -e in global Travis scope

That's correct; you haven't explained why this is a problem. If [it's good enough for Microsoft](https://github.com/Microsoft/tolerant-php-parser/blob/e8d47239fc20ae19a50713cb79a5fd37760d4b3b/.travis.yml#L24), it's good enough for me.

👎 for dropping standard POSIX conditions to use sometimes arithmetic calculation syntax ((()and sometimes bash syntax ([[)

It is [recommended to use Bash conditions in Bash](http://mywiki.wooledge.org/BashFAQ/031). Unless you are targeting `sh` there is no reason to be artificially constrained to traditional test syntax. There is no need for this script to be portable on older systems because it is specifically designed to target Travis only, which is a consistent environment that we know supports Bash syntax. Ergo your desire to use "POSIX conditionals" is purely a subjective and personal one, which happen to be more error prone (e.g. word splitting) and less concise than the alternatives.

👎 for adding 25 lines

Lines are not a resource that needs to be conserved. As stated in the opening summary, the objective of this PR is to make the Travis file more maintainable by breaking long lines down in many, shorter lines. If you think it has gone too far in the other direction, please identify which ones should be condensed.

SpacePossum:
>This PR keeps code lines under 120 chars and simplifies some syntax to improve maintainability.

We've no rule about 120 characters for a line, not as CS nor is this required by the environment we run the tests on. So what is objective reasoning behind this?

> simplifies some syntax to improve maintainability.

I see no simplification, but that is just me and how I read the file. Code simplicity is not objective.
IMHO for maintainability it is better to have small diffs. in the history.

So please let me know if this PR improves anything else, does it solve a problem? Does it add new features? Is this better for the future? Maybe I just oversee it, I'm not that familiar with travis you see.

Bilge:
If the current code is perfect it is impossible for me, or anyone else, to improve upon it.

keradus:
> > does it solve a problem? Does it add new features? Is this better for the future? Maybe I just oversee it, I'm not that familiar with travis you see.
>
> If the current code is perfect it is impossible for me, or anyone else, to improve upon it.

beautiful

Issue #2529, opened 2017-02-10
Line and column where error occurred should be shown in output


Bilge:
* PHP 5.6.18
* PHP CS Fixer 2.0.1 by Fabien Potencier and Dariusz Ruminski

`vendor/bin/php-cs-fixer --allow-risky=no --rules=array_syntax --using-cache=yes -vvv --dry-run fix foo.php`

---

As a minimum, this tool should provide information about the line, and preferably the column, where an error occurred. Output like this is not useful for those wishing to manually fix errors:

>   1) foo.php (array_syntax)

I have no idea where to start looking in the file to fix the issue.

SpacePossum:
`--diff` will show/list all changes that would be made if you would drop the `--dry-run` flag.

For adding alternative or additional output by the tool please see also the other RFC's like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299 for example.

SpacePossum:
closed by the merge of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3162

Issue #299, opened 2014-02-07
Add an output format that prints the line numbers


sk-:
Currently there are only two reporting formats, just the filename and the errors (with verbose mode) and diff, which shows which lines need to be modified.

However, if one wants to modify the file manually it is really time consuming as there are no indications on where on the files these errors are.

Additionally, this feature would be really useful for the integration of this amazing tool and git-lint (https://github.com/sk-/git-lint/). Git-lint is a lint aggregator that shows only errors on modified lines.
Disclaimer: I'm the author of git-lint

lukasoppermann:
@keradus Has this been added? I can't really see lines yet, and this sometimes is a big problem, when in large files you get errors like this. 

```
      --- Original
      +++ New
      @@ @@
           }
      -
       }
```

keradus:
This feature request has never been :+1:'ed and no one has been working on it.
I'm not saying that we won't merge it if it will be made, but there were no interest on it yet except @sk- .

If you want, please start working on that feature and for sure we will help you with your PR ;)

lukasoppermann:
Ahh, I see, well, sorry, I have no idea how to do it. :(

sstok:
#1370

Bilge:
The issue seems to be that fixers themselves do not report the location at which they attempt to modify a file, because they can't, because `FixerInterface` specifies that fixers only receive `Tokens` that do not contain information about lines in a file. Therefore, the only way to know where errors were found is to let the fixer make its changes and run a diff, making a note of the lines in the source file that were changed.

This library currently has diff functionality (but is only invoked when the `--diff` option is specified). However, it uses [sebastianbergmann/diff](https://github.com/sebastianbergmann/diff) which is a crude unified diff implementation; so crude it [does not include any information about the range of lines affected](https://github.com/sebastianbergmann/diff/blob/7a31372a53d0c52bb2d4e95cf600338735fb4a36/src/Differ.php#L101) as in a [normal unified diff](https://en.wikipedia.org/wiki/Diff_utility#Unified_format).

[...]

Issue #1409, opened 2015-09-24
Tokenizer as a standalone component


soullivaneuh:
The Tokenizer of this project is good and could be useful elsewhere IMO.

Would be nice to separate it on a standalone component.

What do you think?

keradus:
Out of curiosity... what is your scenario to use Tokenizer without PHP CS Fixer?

ref #565

Soullivaneuh:
Well, that's a little and crappy use case for me: I have a dependency with another library just for configuration set on constants but I can't add it because of higher PHP version constraint.

So I have to make a command that download the file, parse it with Tokenizer and recreate it with the needed format.

keradus:
If I get you right - downgrade PHP source code ? :D

Soullivaneuh:
I would prefer to say : Keep support for devs using old and crappy PHP versions. ;-)

boekkooi:
Another option would be to allow a project like [PHP CodeSniffer](https://github.com/squizlabs/PHP_CodeSniffer) to use the same Tokenizer or in my case I could maybe use it for something like [PHP-CS-Checker](https://github.com/boekkooi/PHP-CS-Checker)

julienfalque:
I think it's not worth it: the internal tokenizer is tightly linked to how PHP CS Fixer works, making it a separate package would mostly increase maintainance costs for no real gain. One can use [nikic/php-parser](https://packagist.org/packages/nikic/php-parser) or simply have another `token_get_all()` wrapper.

PR #2550, opened 2017-02-19
Added changed line numbers to Tokens and Text/JSON reporters

Bilge:
Proof of concept for adding line numbers to output. Currently only text and JSON output shows the line numbers (XML and Junit are left to the interested programmer). Addresses #2529, #299.

### Sample text output

>   1) src\Console\Command\FixCommand.php (array_syntax lines 83, 280; class_definition line 39)

### Sample JSON output

```js
{
  "files": [
    {
      "name": "src\\Console\\Command\\FixCommand.php",
      "appliedFixers": {
        "array_syntax": [
          83,
          280
        ],
        "class_definition": [
          39
        ]
      }
    }
  ],
  "time": {
    "total": 4.032
  },
  "memory": 8
}
```

### BC breaks

⚠️ Breaks BC for modified formatters because `appliedFixers` structure was changed from a list of strings to a map of integers (line numbers) with string keys (fixer names). I think this is a better structure but if a BC break is unacceptable, or to back-port this feature to 2.x, we can copy this data to a different name (instead of `appliedFixers`). I still think modifying the structure of `appliedFixers` should be the preferred format for the next major version since it avoids duplication of fixer names.

### TODO
* Fix tests.

### Future enhancements

* Update the text formatter to list runs of consecutive lines as groups, e.g. instead of "lines 1, 2, 3, 4, 5", report: "lines 1-5".

### DIFF
--- a/src/Runner/Runner.php
+++ b/src/Runner/Runner.php
@@ -169,9 +169,9 @@ final class Runner
                 $fixer->fix($file, $tokens);
 
                 if ($tokens->isChanged()) {
+                    $appliedFixers[$fixer->getName()] = $tokens->findChangedLines();
                     $tokens->clearEmptyTokens();
                     $tokens->clearChanged();
-                    $appliedFixers[] = $fixer->getName();
                 }
             }
         } catch (\Exception $e) {
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -834,6 +834,9 @@ class Tokens extends \SplFixedArray
                 throw new \InvalidArgumentException('Must not add empty token to collection.');
             }
 
+            // Mark inserted token as changed.
+            $items[$i]->setChanged(true);
+
             $this[$i + $index] = $items[$i];
         }
     }
@@ -968,7 +971,7 @@ class Tokens extends \SplFixedArray
         $this->setSize(count($tokens));
 
         foreach ($tokens as $index => $token) {
-            $this[$index] = new Token($token);
+            $this[$index] = $previousToken = new Token($token, isset($previousToken) ? $previousToken : null);
         }
 
         $transformers = Transformers::create();
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -413,7 +413,7 @@ class Tokens extends \SplFixedArray
     }
 
     /**
-     * Finds all unique changed lines numbers.
+     * Finds all changed lines numbers.
      *
      * @return int[] Line numbers
      */
@@ -423,10 +423,13 @@ class Tokens extends \SplFixedArray
 
         foreach ($this as $index => $token) {
             if ($token->isChanged()) {
-                $lines[] = $this->findNearestPrevTokenLineNumber($index);
+                $lines[] = $this->backtrackToNearestLineNumber($index);
             }
         }
 
+        sort($lines);
+
+        // Call to array_values ensures keys are contiguous, especially for JSON output.
         return array_values(array_unique($lines));
     }
 
@@ -1160,23 +1163,27 @@ class Tokens extends \SplFixedArray
     }
 
     /**
-     * Finds the line number of the first token previous to the one at the specified index, including itself.
+     * Backtracks through the token list to find the line number of the nearest token, including itself, starting at
+     * the specified token index.
      *
      * @param int $index Token index
      *
      * @return int Line number
      */
-    private function findNearestPrevTokenLineNumber($index)
+    private function backtrackToNearestLineNumber($index)
     {
-        if ($index <= 0) {
-            return 1;
-        }
+        $newlines = 0;
 
-        if (null !== $line = $this[$index]->getLineNumber()) {
-            return $line;
-        }
+        do {
+            $token = $this[$index];
+            $newlines += $token->countNewlines();
 
-        return $this->findNearestPrevTokenLineNumber(--$index);
+            if (null !== $originalLineNumber = $token->getOriginalLineNumber()) {
+                return $originalLineNumber + $newlines;
+            }
+        } while (--$index > 0);
+
+        return 1;
     }
 
     /**
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -413,6 +413,24 @@ class Tokens extends \SplFixedArray
     }
 
     /**
+     * Finds all unique changed lines numbers.
+     *
+     * @return int[] Line numbers
+     */
+    public function findChangedLines()
+    {
+        $lines = array();
+
+        foreach ($this as $index => $token) {
+            if ($token->isChanged()) {
+                $lines[] = $this->findNearestPrevTokenLineNumber($index);
+            }
+        }
+
+        return array_values(array_unique($lines));
+    }
+
+    /**
      * Find tokens of given kind.
      *
      * @param int|array $possibleKind kind or array of kind
@@ -1136,6 +1154,26 @@ class Tokens extends \SplFixedArray
         }
 
         $this[$nextIndex]->clear();
+    }
+
+    /**
+     * Finds the line number of the first token previous to the one at the specified index, including itself.
+     *
+     * @param int $index Token index
+     *
+     * @return int Line number
+     */
+    private function findNearestPrevTokenLineNumber($index)
+    {
+        if ($index <= 0) {
+            return 1;
+        }
+
+        if (null !== $line = $this[$index]->getLineNumber()) {
+            return $line;
+        }
+
+        return $this->findNearestPrevTokenLineNumber(--$index);
     }
 
     /**
--- a/src/Report/JunitReporter.php
+++ b/src/Report/JunitReporter.php
@@ -125,7 +125,7 @@ final class JunitReporter implements ReporterInterface
         if ($shouldAddAppliedFixers) {
             $failureContent = "applied fixers:\n---------------\n";
 
-            foreach ($fixResult['appliedFixers'] as $appliedFixer) {
+            foreach (array_keys($fixResult['appliedFixers']) as $appliedFixer) {
                 $failureContent .= "* {$appliedFixer}\n";
             }
         } else {
--- a/src/Report/TextReporter.php
+++ b/src/Report/TextReporter.php
@@ -62,7 +62,16 @@ final class TextReporter implements ReporterInterface
     {
         return sprintf(
             $isDecoratedOutput ? ' (<comment>%s</comment>)' : ' (%s)',
-            implode(', ', $fixResult['appliedFixers'])
+            implode(
+                '; ',
+                array_map(
+                    function ($fixer, $lines) {
+                        return sprintf('%s line%s %s', $fixer, count($lines) !== 1 ? 's' : '', implode(', ', $lines));
+                    },
+                    array_keys($fixResult['appliedFixers']),
+                    $fixResult['appliedFixers']
+                )
+            )
         );
     }
 
--- a/src/Report/XmlReporter.php
+++ b/src/Report/XmlReporter.php
@@ -81,7 +81,7 @@ final class XmlReporter implements ReporterInterface
     {
         $appliedFixersXML = $dom->createElement('applied_fixers');
 
-        foreach ($fixResult['appliedFixers'] as $appliedFixer) {
+        foreach (array_keys($fixResult['appliedFixers']) as $appliedFixer) {
             $appliedFixerXML = $dom->createElement('applied_fixer');
             $appliedFixerXML->setAttribute('name', $appliedFixer);
             $appliedFixersXML->appendChild($appliedFixerXML);
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -37,6 +37,13 @@ class Token
     private $id;
 
     /**
+     * Line number from which this token was parsed.
+     *
+     * @var int|null
+     */
+    private $line;
+
+    /**
      * If token prototype is an array.
      *
      * @var bool
@@ -61,6 +68,11 @@ class Token
             $this->isArray = true;
             $this->id = $token[0];
             $this->content = $token[1];
+
+            // Line number is unavailable for some tokens.
+            if (isset($token[2])) {
+                $this->line = $token[2];
+            }
         } else {
             $this->isArray = false;
             $this->content = $token;
@@ -232,6 +244,16 @@ class Token
     public function getId()
     {
         return $this->id;
+    }
+
+    /**
+     * Gets the line number from which this token was parsed.
+     *
+     * @return int|null Line number
+     */
+    public function getLineNumber()
+    {
+        return $this->line;
     }
 
     /**
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -37,11 +37,11 @@ class Token
     private $id;
 
     /**
-     * Line number from which this token was parsed.
+     * Line number from which this token was originally parsed.
      *
      * @var int|null
      */
-    private $line;
+    private $originalLineNumber;
 
     /**
      * If token prototype is an array.
@@ -61,8 +61,8 @@ class Token
      * Initializes this instance with the specified token prototype, and optionally, the previous token. When the
      * previous token is specified, the line number for single-character tokens is derived from the previous token.
      *
-     * @param string|array $token token prototype
-     * @param Token|null $previousToken Optional. Previous token.
+     * @param string|array $token         token prototype
+     * @param Token|null   $previousToken Optional. Previous token.
      */
     public function __construct($token, $previousToken = null)
     {
@@ -73,18 +73,18 @@ class Token
 
             // Line number may be unspecified for fabricated tokens.
             if (isset($token[2])) {
-                $this->line = $token[2];
+                $this->originalLineNumber = $token[2];
             }
         } else {
             $this->isArray = false;
             $this->content = $token;
 
             if ($previousToken) {
-                $this->line = $previousToken->getLineNumber();
+                $this->originalLineNumber = $previousToken->getOriginalLineNumber();
 
                 // When previous line contained one or more newlines, this token must appear that many lines ahead.
-                if ($previousToken->isWhitespace() && $newlineCount = substr_count($previousToken->content, "\n")) {
-                    $this->line += $newlineCount;
+                if ($newlineCount = $previousToken->countNewlines()) {
+                    $this->originalLineNumber += $newlineCount;
                 }
             }
         }
@@ -138,6 +138,16 @@ class Token
     public function clearChanged()
     {
         $this->changed = false;
+    }
+
+    /**
+     * Counts the number of newlines (\n) in whitespace tokens.
+     *
+     * @return int Number of newlines
+     */
+    public function countNewlines()
+    {
+        return $this->isWhitespace() ? substr_count($this->content, "\n") : 0;
     }
 
     /**
@@ -258,13 +268,13 @@ class Token
     }
 
     /**
-     * Gets the line number from which this token was parsed.
+     * Gets the line number from which this token was originally parsed.
      *
      * @return int|null Line number
      */
-    public function getLineNumber()
+    public function getOriginalLineNumber()
     {
-        return $this->line;
+        return $this->originalLineNumber;
     }
 
     /**
@@ -502,11 +512,11 @@ class Token
     /**
      * Sets the changed flag to the specified value.
      *
-     * @param bool $changed True if this token has changed, otherwise false.
+     * @param bool $changed true if this token has changed, otherwise false
      */
     public function setChanged($changed)
     {
-        $this->changed = (bool)$changed;
+        $this->changed = (bool) $changed;
     }
 
     /**
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -58,24 +58,35 @@ class Token
     private $changed = false;
 
     /**
-     * Constructor.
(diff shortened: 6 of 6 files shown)


Bilge:
> You set it up during creation of Token, but you do not maintain it during fixing process.

Why should we? We do not want to know what the line number of the modified file is; we want to know the original line number of the source file. Ergo, additions are irrelevant.

keradus:
No, that means the data attached to token is not a token line, but some ugly original token line and not real line in tokens collection. That show bad way of achieving the goal, as attached data is not valid during whole lifecycle of instance.

src/Tokenizer/Tokens.php
```diff
@@ -1139,6 +1157,26 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
     }
 
     /**
+     * Finds the line number of the first token previous to the one at the specified index, including itself.
+     *
+     * @param int $index Token index
+     *
+     * @return int Line number
+     */
+    private function findNearestPrevTokenLineNumber($index)
```
keradus: method should not exist, as with your approach Token instance should have the line property always valid. here, you hack the code because you don't have it always provided. bad approach

Bilge:
You're certainly a very vocal critic but you haven't proposed any alternative solutions. What do you suggest instead?

keradus:
It's not just a criticism, I point you concrete and real gaps and issues.
Valid solution would be to have nice differ that would provide lines. Fact of not having differ that would match our needs is not an excuse for deficient solution

src/Tokenizer/Tokens.php
Bilge: Do you not understand or do you choose not to understand? As I [tried to explain](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2550#issuecomment-280913912), there are **two versions** of a file: the **original version** and the **fixed version**. We *do not* want to report the lines of the fixed version because they are ephemeral and meaningless to a user wishing to manually fix the original source file; they are only acquainted with the **original file**. Ergo it is not only a huge waste of time, effort and processing power to track the lines of the modified tokens but it is also **strictly incorrect** to do so.

keradus: So, again, the outcome you wanna have and place you put the solution is 2 completely different things. Token is independent of runner. Keep things independent - if you want to have some data in some class - the data must be valid all the time. If it cannot be obtained, it proves you put it in wrong place.

Bilge:
[I have explained why diffing is not a valid solution](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299#issuecomment-280888207) in the related ticket. In summary, there is no good diff tool for PHP, which you already seem to be acutely aware of, so I'm not sure why you're suggesting that as a better implementation.

I do not believe re-parsing a string diff would be a better solution even if there was a good diff tool available since doing so is less efficient than using the token collection and the line numbers already provided by the token parser. Converting the tokens to a diff and then parsing it again must be worse performance because it requires two additional transformations.

I have provided a working proof of concept for the token parser solution but you only seem prepared to criticize without providing a valid alternative. You are not doing a good code review if you only identify problems but never propose solutions.

If the only solution available is a bad one it's better than none at all. For what it's worth, I don't think it's so bad. I think this PR needs some improvements but I'm not inclined to make them until we're on the same page. Let's discuss feasible improvements that build on this foundation instead of throwing it all away.

cc @fabpot

src/Tokenizer/Tokens.php
Bilge: I have not mixed any concerns here, the code is separated just as it was before. You are correct that there are some limitations of the line tracking code but those limitations suit the current use case, which is to track the lines of the source file.

If there becomes a use case to track the lines of the modified file then this code could be extended to provide that improvement. For now it is not necessary, and I suspect that requirement will never exist, so it is certainly not worth the extra work at this stage.

keradus:
> I have provided a working proof of concept for the token parser solution but you only seem prepared to criticize without providing a valid alternative.

that's the point. it's not working, and I have provided you cases for it.

Bilge:
What steps are needed to make this PR passable? So far your only suggestion has been to trash it and write a differ instead. If you're prepared to merge this PR, what changes need to be made?

SpacePossum:
>If the only solution available is a bad one it's better than none at all.

I disagree, having features we know do not work correctly should not be added.

> I think this PR needs some improvements but I'm not inclined to make them until we're on the same page. Let's discuss feasible improvements that build on this foundation instead of throwing it all away with no valid alternative on the table.

Sounds good to me, what do you propose to deal with tokens that are;
- moved to another line in a file; and/or
- removed or replaced with other tokens; and/or
- created with multiple lines of content which has changed partly; 

As example for the last point:
```diff
--- Original
+++ New
@@ @@
 <?php
 class D extends G
 {
     /**
-     * 1
+     * 1.
      *
      * @return
-     *
-     *
      */
     public function __construct(\SplFileInfo $a)
     {
     }
 }
```

You are critical to the other solution, so what do you propose for these cases?

src/Runner/Runner.php
```diff
@@ -169,9 +169,9 @@ private function fixFile(\SplFileInfo $file, LintingResultInterface $lintingResu
                 $fixer->fix($file, $tokens);
 
                 if ($tokens->isChanged()) {
+                    $appliedFixers[$fixer->getName()] = $tokens->findChangedLines();
```
keradus: btw, this does change the output structure of `Runner::fix`, which is a BC breaker

Bilge: Isn't that exactly what I wrote in the issue summary?

>:warning: Breaks BC for modified formatters because appliedFixers structure was changed from a list of strings to a map of integers (line numbers) with string keys (fixer names).

Bilge:
@SpacePossum Have you tested your example? What is the result? I cannot test it at the moment, but I expect it would produce the expected result: line 5 was changed. This is because, unlike your description, there are not multiple lines of content but in fact just one line of content which was overridden. In the override case the line number is preserved.

The difficult case is the removal case because, in the absence of the token, we have no token to track the change! However, we could track removals as a separate list and merge them together when the list of changed lines are requested.

keradus:
there need to be one mechanism for tracking change of token, removal of token and modification of it

Bilge:
From my brief analysis, it seems you never remove tokens from the collection at all; their content is simply blanked. If that is the case then removals should already be reported correctly without any modification.

keradus:
eg `else_if` fixer removes space between `else` and `if`. token is removed.

SpacePossum:
> This is because, unlike your description, there are not multiple lines of content but in fact just one line of content which was overridden.

My sample clearly shows 4 lines changed (3 moved, 1 modified), all in the same token. So I think my description is correct.
As much as I like to see this feature added I'm not sure this is the best way to do it, that is all.

Bilge:
> I'm not sure this is the best way to do it, that is all.

* Have you tested your case to prove my code fails? If so, what is the result? If not, why are you claiming my code is bad without even trying it?
* If my code is bad, propose an alternative. What is the best way?

keradus:
@Bilge , you need to change approach, as one of our user already stated in different issue.

Bilge:
@keradus 

* Why do I need to change my approach? 
* What do I need to change my approach to?

Bilge:
As expected, my code already works for the removal case.

@keradus Re: `ElseifFixer`:

```php
<?php

class Foo
{
    public function test()
    {
        if ($x = 0) {
            $x = 1;
        } else if ($y = 1) {
            $y = 0;
        }
    }
}
```

>   1) test.php (header_comment line 2; elseif line 9)

Bilge:
@SpacePossum 

> $cat test.php

```php
<?php

class D extends G
{
    /**
     * 1
     *
     * @return
     *
     *
     */
    public function __construct(\SplFileInfo $a)
    {
    }
}
```

>$ ./php-cs-fixer fix --dry-run -v test.php
>   1) test.php (header_comment line 2; phpdoc_summary line 5; phpdoc_trim line 5)

SpacePossum:
```php
<?php
class A{
    private function test(
        $a,
    $b
    ){}public function testa(){
        echo 123;
    }
}
```

```
F
   1) test123.php (ordered_class_elements lines 2, 3, 4, 5, 6, 7; method_separation line 5; blank_line_after_opening_tag lines ; header_comment line 1; function_declaration lines ; class_definition lines ; braces lines 2, 3)

```

So here is an example. I tested it as you can see, you can recreate the test as all is provided.
Lets take a look out the line numbers:
>ordered_class_elements lines 2, 3, 4, 5, 6, 7;

3 and 6 or OK, 2, 4, 5 and 7 are wrong

>method_separation line 5;

5 is wrong

>blank_line_after_opening_tag

no lines

>header_comment

is OK

>function_declaration

no lines

>class_definition lines

no lines

>class_definition lines ; braces lines 2, 3

3 is wrong

total: OK 3, wrong 6, 3 missing.

This is only on simple same with a few fixers. All tests I did I got the wrong numbers.

Now you can conclude I'm being negative, only like my own code, not providing objective reasoning and in general I should provide you all you need to get this working as I review your PR and provide test cases.
However from the start of this PR the issues with this approach have been pointed out and an alternative was provided. The alternative being developing another diff. report builder.  No one asks you to write that report builder. No one states it is the _only_ solution.
However that doesn't mean this PR is acceptable either. This PR is not diff. from any other. The community reviews and if issues are spotted these have to be addressed before it can be merged.

Bilge:
@SpacePossum I never claimed the implementation is perfect. On the contrary, I know it isn't. I still think it's salvageable and a little more work is needed to fix the outstanding issues. You seem to think it's non-salvageable and we should just throw it all away. Why is that?

src/Tokenizer/Tokens.php
```diff
@@ -1139,6 +1157,26 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
...
+    {
+        if ($index <= 0) {
+            return 1;
```
Bilge: @SpacePossum It seems this code is not working correctly. This is supposed to ensure that if we hit the start of the file the line number must be line 1, but sometimes we receive a blank value, so something is not quite right here.

Bilge:
@SpacePossum Re:
>ordered_class_elements lines 2, 3, 4, 5, 6, 7;
3 and 6 or OK, 2, 4, 5 and 7 are wrong

I'm not sure how you can ever assert only 3 and 6 are correct. Since the methods are transposed the entire body of the class is effectively rewritten in swapping the method order. Therefore lines 3-8 are modified. Both my implementation and your proposed diff algorithm would view this change the same way; there is no way to reduce it down to just the method declaration line.

keradus:
btw, not sure if there is any actual value of having modifed lines per rule instead of just modified lines per whole fixing process

Bilge:
@keradus There should be absolutely no doubt that lines per fixer comprise essential information. The use-case is a user whom wants to fix files manually. If they have a mismatched list of fixers and modified lines they have no idea which fixer was applied to each line so they are no more enabled than they are currently in resolving each failed check.

keradus:
main purpose of this tool is to automatically fix violations.
going in direction to ask user to fix it manually is kinda not purpose of this tool...

Bilge:
@keradus [That's absolutely not what you said to the guy who first reported this issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299).

@sk- says:

>if one wants to **modify the file manually** it is really time consuming as there are no indications on where on the files these errors are.

You say:

>If you want, **please start working on that feature** and for sure **we will help you with your PR ;)**

In reality all you do is take a massive shit on anyone who isn't already an established contributor. I'm trying my best implement this feature and, counter to your claim, you have done absolutely **nothing** to help and on top of all that you're now pretending the entire feature is out of scope. Why are you being like this?

keradus:
You need to be ready to encounter criticism. But it was not empty one - I show you cases you did not cover and why. For some reason, it's hard for me (and what I saw in other issues not only me) to came to nice conclusion with you. Maybe I just require solution that would be too good than needed. Maybe not.

But yeah, not all PRs requires my accompaniment, let me do a step backward and let this PR be handled by community.

SpacePossum:
I do know how it feels to up a lot of effort into something you really care about just to see it being criticized. My PR's didn't always make it, others took forever to complete. People had trouble with the amount of PR's I opened and problems with all of my PR's for various reasons.

I also know how it feels to spend a lot of time reviewing and testing PR's, helping on PR's, answering questions, fixing and maintaining code (of which some I didn't write nor care about). Not all of that was always greeted with thanks from the community, let alone on first sight.

For this PR I tried from the beginning to express my concerns about the approach and through the process try to show by examples that it might not be test way in my opinion.
Holding on and trying to get your idea across to me is something I appreciate, though the style sometimes makes me feel like I want give up.

So my thoughts on the changes in this PR (so not on you, nor the time you put into it):
- I think the FR can be done using a different differ. All mechanics for this are on place. Yes it will a lot of work and I'm not asking anyone to do it.
- I would like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1409 to happen at some point. Therefor I would like to see a clear separation of concerns. In my view a token does not need to know where it is, not in the collection, not in a file, nor where it came from. Not all tokens come from a file for example.
- Trying to determine where a token originated from is pretty hard. For example the transformers may already moved, replaced or replaced tokens. Fixers may do something similar.

You may value all those points differently and you are free and welcome to state so. Just as I hope you allow me to state my thoughts. Lets hope we get more input from the community.

Bilge:
@keradus @SpacePossum Let me start by saying: I appreciate your input, from both of you. Second, I do not enjoy fighting with each other; I want to discuss this PR instead. That said, it's clear we still have a few issues to discuss before we can get to that.

### On criticism

You both believe I cannot accept criticism. Ironically, I seldom hear the criticism that I cannot accept criticism. It may be the case that I could do better here, but allow me to share my point of view. I believe the bulk of criticism I have received hitherto has been non-constructive. [As I alluded to earlier](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2550#issuecomment-280957297), a good code review should do two things:

1. Identify a problem.
2. Propose a solution.

Most of the criticism I have received only does one of these things. If you identify a problem but don't propose any solution you're relying on the PR author to find a solution, but there may not be one, in which case the value in identifying the problem is diminished.

The bigger issue occurs when solutions are proposed without identifying a problem. If you offer alternative solutions but don't explain why the current solution is problematic, you're just offering an alternative with no compelling reason for switching to it. You might think the reasons are so *obvious* you don't need to explain them, but you do: what is obvious to you is not necessarily obvious to everyone else, and suffice it to say, some of your assumptions may be faulty but we cannot begin to discuss them if they are not written down.

In both cases the key thing that's missing is objective reasoning underpinning your critique. Why should I, or anyone else, accept criticism that lacks justification? If we cannot begin an objective discussion regarding the pros and cons of each viable approach we cannot begin to move forward.

With that said, let's move on to discuss the valid criticisms of this PR.

### On diffing

I am consistently being told that my solution should be thrown out for a diff analyser. Before I began working on this PR I [explained why this approach should be better](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299#issuecomment-280888207). Basically, it was because there isn't a good diff solution available right now. However, at that time I did not realise how much more elegant token tracking would be, and I have since come to realise that it is by far the better approach despite the fact you are both telling me otherwise. Allow me to explain.

When analysing the [method moving scenario](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2550#issuecomment-281072054) it occurred to me that a diff is a terrible solution to this problem because the diff will report different affected lines depending on the content and structure of the methods. The token analyser, in contrast, should consistently report the correct range of lines regardless of how the methods are composed. The reason for this is a diff has no idea how the fixed file arrived at its final state, whereas the token tracker does know, because it's keeping track of the tokens individually throughout their transformations.

In the specific example of the `ordered_class_elements` fixer, reporting lines `3, 4, 5, 6, 7, 8` as modified would be correct (it is currently off by one due to a bug which can easily be fixed). Moreover, we could analyse sequences of consecutive lines to condense them down to runs expressed as `3-8` for example, to make it more readable. This is not possible using a diff solution.

### On separation of concerns

I do not think this PR precludes creating a standalone tokenizer. It is important to realize the tokenizer is really just a wrapper around [token_get_all](http://php.net/manual/en/function.token-get-all.php). Note that this tokenizer provides line numbers, and indeed, it is these very line numbers I am using in this PR. It seems intuitive to me to provide the complete set of information returned by the API instead of arbitrarily discarding some of it, particularly when it's very useful data such as line numbers.

I think the confusion has occurred because I called the field *line* and the accessor, *getLineNumber*, which may create the expectation of some developers that this method should always provide the current and correct line number for that token, which it does not. I believe this is a mistake on my part but is purely a naming problem rather than an implementation problem: the field should be renamed to *originalLineNumber* and the the accessor, *getOriginalLineNumber*. In this way there is no doubt that this line number is from the source file only and is immutable: it will not change.

If there is a need to track new line numbers for changed tokens that feature can be added at a later date, if it is considered appropriate at that time for whatever use case it is intended to support. It is certainly not required for the use case this PR intends to support so I will not be adding it.

### On token tracking

To address @SpacePossum's final point:

>Trying to determine where a token originated from is pretty hard. For example the transformers may already moved, replaced or replaced tokens. Fixers may do something similar.

I have to strongly disagree. Tracking where a token originated from is very easy because that's exactly what `token_get_all` gives us. Furthermore, the implementation of the `Tokens` collection is really nice because we do not have to worry about removed or replaced tokens: in both cases the `Tokens::override` method is called which marks the token as changed and preserves the line number *even in the removal case*. That's right, removed tokens are just overridden with blank content so we do not lose the ability to track them!

---

For these reasons I still think token tracking is a far stronger solution than diff analysing. It's more accurate, faster and a proof of concept is available now where one does not exist for the alternative. It's not perfect, but I'm confident we can fix it without too much more effort.

All I'm asking for is your support. By support I am not asking for you to code anything for me, though you can if you wish; I'm just asking for you to be working with me instead of against me. I cannot be motivated to fix these issues until I get some confirmation that doing so will result in a merge of the final, working version of the code. I just want to be confident I am not wasting my time by working further on this. Please continue to provide constructive criticism and together I believe we can solve this.

Bilge:
@SpacePossum 
>    blank_line_after_opening_tag
> no lines

The reason this happens is because `BlankLineAfterOpeningTagFixer` only inserts a token. Inserted tokens have their `$changed` flag set to `false`. Since this token as marked as unchanged it is not picked up by `Tokens::findChangedLines`. I believe this is incorrect: inserted tokens *should* be marked as changed, however it is not currently possible to do so because the setter doesn't exist. I think we need to add a `setChanged` method to `Token` for this case.

Bilge:
* I have added line numbers to single-character tokens to improve line tracking accuracy. The line number of single-character tokens is derived by analysing the previous token.

* I added change tracking for inserted token by setting their changed flag to `true`.

A little more work needs to be done to improve line number accuracy reported by the backtracer.

Bilge:
I have fixed the backtracker so it counts any newlines it passes over. With this change, this is no longer a proof of concept&mdash;I am happy to submit this as the finished piece of work. Of course, tests still need to be fixed and probably some new ones written. Please let me know if you find any issues.

Bilge:
@keradus Please re-review this when you have the time. Regarding your original points:

>non-array tokens like ; don't have line property, only array tokens like T_IF have

This is no longer the case. I have added line number tracking for single-character tokens to improve accuracy of the changed lines reporter.

>when adding new token we don't set line property

This is still true because we do not need this feature to report on changed lines. This can be added later if needed for some other use case.

>when changing new lines we don't update line of following tokens

This PR tracks the original line numbers only. We do not need to track line numbers of new tokens.

src/Runner/Runner.php
Bilge: One possible optimization here: we could use `$tokens->findChangedLines()` in lieu of the guarding `$tokens->isChanged()` check, since there will always be changed lines if the tokens collection has changed. Essentially, the `isChanged` guard is obsolete and also inefficient because both methods might loop over the entire token collection.

However, if we want to have an option to disable line numbers the guard would still be needed. Therefore this optimization is a question of whether we should always produce line numbers or not.

keradus:
Is there is any other voice from community about proposed solution for tracking changed lines except mine and SpacePossum one?

SpacePossum:
I'm closing this PR after a lengthy discussion so I keep my reasoning for closing it a bit short as to not repeat myself to much ;)

This PR introduces changes that, after some tweaking, could be used to add line numbers to the diff-output. While that is a feature we would like to have, the cost of this approach is to high.
The roadmap for the fixer is to split out the tokenizer (and the Token class and transformers with it), such that other implementations can be used (and the tokenizer can be reused for other tooling). 
As part of that I don't want to require (or even recommend) that the line numbers of tokens are tracked. 
This makes the implementation of a tokenizer easier and less dependent on PHP's native `token_get_all`, it also keeps it more easy to create token transformers.
So, I'm not judging the quality of the code or even the concept, however it is not a feature that fits in the future plan of the fixer. After almost 3 months no people have responded in favor of this approach and therefor I'm not convinced. I'm still open to hear from others of course.

This PR was not always about the changes itself, but also about how maintainers/members responded on it. For this I would like to thank you in part, because it got me thinking about how I react to PR's, issues, communicate future plans etc., how I could expressed myself better and how I can help prevent contributors spending their valuable time on features that are not likely going to make it.
As such I hope I do a bit now and plans to update the `contribute` doc's are in the making as well. It does not mean I agree with all that is sad, but leaving the wrong expression is still something that should and needs to be addressed.

Keeping this PR open would suggest this PR is still be considered to be merged, while atm it is not.
However closing a PR does not mean it must be the end of it, so anyone is still free to comment on this when something is left out. Also for fellow members and maintainers as this comment and PT-closing is done IMHO.

PR #2552, opened 2017-02-21
Initial compatibility with PHP 7.2-DEV

keradus:


julienfalque:
👍 but shouldn't we allow it to fail?

keradus:
I would rather want to have nightly fixed and then keep it green all the time.
After that, I don't want to allow users to break it with new PRs.

julienfalque:
Nightlies are very unstable. Once you have fixed all errors and the code runs fine on nightly, you can't be sure that it won't break again because of some bug or unexpected change introduced later in nightly. You may end up having failing builds on PRs that don't even change the code.

I think getting feedback on whether the code runs on nightlies is great, as it will allow to anticipate on future PHP releases. But IMO it should not be a mandatory requirement of the build to succeed. At least for now. Maybe later, when we will be closer to a new PHP release, would be more appropriate.

keradus:
> You may end up having failing builds on PRs that don't even change the code.

I agree it may happen, and it would be a prio then to fix it for project maintainers.

Let us first focus on having that nightly build running green

julienfalque:
> it would be a prio then to fix it for project maintainers.

Alright then 👍

SpacePossum:
how about adding;
```
 - 7.0snapshot
 - 7.1snapshot
```
(for ref.: https://github.com/travis-ci/travis-ci/issues/7268#issuecomment-277669166)

keradus:
I see no value in adding `7.1snapshot` while we have added latest stable of `7.1` already.
`7.1@dev` is not that much different for us than `7.1@stable` in context of code syntax.
What we care about is `7.2` vs `7.1`.

SpacePossum:
Sad HHVM never fix the iterator issue, IMHO this is the best away around.
Thanks @keradus :+1:

.travis.yml
```diff
@@ -20,6 +20,8 @@ matrix:
           env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: 7.1
           env: TASK_SCA=1
+        - php: nightly
```
GrahamCampbell: better to say 7.2, surely?

keradus: is it already there? when started work on this PR, it was not

SpacePossum: [nope](https://secure.php.net/supported-versions.php), so lets wait for official release to set the release version, PHP might just jump another major to 9 given its history ;)

Issue #1990, opened 2016-06-13
Short diff


Phyks:
Hi,

I am using PHP-CS with a combination of `--dry-run` and `--diff` for automated tests. However, diff is really verbose and outputs the diff of the whole file (starting from the error). Would it be possible to have only a short diff, suppressing common lines, as well?

For instance, for a file having an encoding issue (typically UTF-8 with BOM), it will display the +/- on the first line but will also display the common part, that this the whole rest of the file, which is not very readable.

Thanks

keradus:
1. you don't need to show diff to have check failed.
2. but you probably want to have it to see why it fails, cool ;) for 1.x there is no plan to allow using different diff's (a lot of work needed), but there is a plan for 2.x line for allowing it (needed refactoring is already done and merged).

In the end, I think what you want is #299 .
Also, with proper implementation of #299 you will be able to choose between PHPUnit diff, Phpspec diff or your own, custom one.

Phyks:
Indeed, #299 seems to match perfectly with what I need in the end. I focused on having a short diff, then did not find it when I searched for issues.

Closing as a duplicate, will subscribe to and follow #299.

Phyks:
Another interesting thing with short diffs I have just came across is when you have many files (say 200 for example) that you want to modify, but want to check the diff. You only want the diff to show the actual modifications, as displaying 200 files in your terminal is not really useful.

My current way of handling it is to do a fix run and then run git diff, so post-check rather than pre-check.

keradus:
I agree that this feature would be nice ;) Just need sb who will implement this. Maybe you want to help ?

Phyks:
I guess we should adapt https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/da56b8c83664dfa7f805a9dab2d29b44f4acf869/Symfony/CS/Fixer.php#L316. A proof of concept is available [here](https://gist.github.com/Phyks/407ca2054a3f0452119c8a0ddd0645c6#file-usage-example).

Let me know what do you think about it. I do not know the codebase of PHP-CS-Fixer at all, I am not sure how to integrate it and / or add an option for short diffs.

keradus:
1. Create new differ class here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master/src/Differ
that will call phpspec differ (new dependency)
[...]

PR #2554, opened 2017-02-23
Add short diff. output format

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1990
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2007

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -180,6 +180,7 @@ Complete configuration for rules can be supplied using a ``json`` formatted stri
 
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
+Optionally the ``--diff`` can be used to specify the diff. output format; ``--diff=sbd`` (default) or ``--diff=sbd-short``.
 
 The ``--allow-risky`` option allows you to set whether risky rules may run. Default value is taken from config file.
 Risky rule is a rule, which could change code behaviour. By default no risky rules are run.
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         "php": "^5.3.6 || >=7.0 <7.2",
         "doctrine/annotations": "^1.2",
         "ext-tokenizer": "*",
-        "sebastian/diff": "^1.1",
+        "sebastian/diff": "^1.4",
         "symfony/console": "^2.4 || ^3.0",
         "symfony/event-dispatcher": "^2.1 || ^3.0",
         "symfony/filesystem": "^2.4 || ^3.0",
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -107,7 +107,6 @@ final class ConfigurationResolver
         'cache-file' => null,
         'config' => null,
         'diff' => null,
-        'differ' => null,
         'dry-run' => null,
         'format' => null,
         'path' => array(),
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -103,18 +103,19 @@ final class ConfigurationResolver
      */
     private $options = array(
         'allow-risky' => null,
+        'cache-file' => null,
         'config' => null,
+        'diff' => null,
+        'differ' => null,
         'dry-run' => null,
         'format' => null,
         'path' => array(),
         'path-mode' => self::PATH_MODE_OVERRIDE,
-        'using-cache' => null,
-        'cache-file' => null,
         'rules' => null,
-        'diff' => null,
-        'verbosity' => null,
-        'stop-on-violation' => null,
         'show-progress' => null,
+        'stop-on-violation' => null,
+        'using-cache' => null,
+        'verbosity' => null,
     );
 
     private $cacheFile;
@@ -253,7 +254,25 @@ final class ConfigurationResolver
     public function getDiffer()
     {
         if (null === $this->differ) {
-            $this->differ = false === $this->options['diff'] ? new NullDiffer() : new SebastianBergmannDiffer();
+            switch ($this->options['diff']) {
+                case false:
+                    $this->differ = new NullDiffer();
+
+                    break;
+                case 'sbd':
+                    $this->differ = new SebastianBergmannDiffer(true);
+
+                    break;
+                case 'sbd-short':
+                    $this->differ = new SebastianBergmannDiffer(false);
+
+                    break;
+                default:
+                    throw new InvalidConfigurationException(sprintf(
+                        'Differ must be "sbd" or "sbd-short", got "%s".',
+                        $this->options['diff']
+                    ));
+            }
         }
 
         return $this->differ;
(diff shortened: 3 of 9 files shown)


composer.json
```diff
@@ -16,7 +16,7 @@
     "require": {
         "php": "^5.3.6 || >=7.0 <7.2",
         "ext-tokenizer": "*",
-        "sebastian/diff": "^1.1",
+        "sebastian/diff": "^1.4",
```
SpacePossum: for ref.: https://github.com/sebastianbergmann/diff/pull/29

SpacePossum:
With appveyor up again this should be all green.
This is RTM IMHO

src/Console/Command/FixCommandHelp.php
```diff
@@ -79,6 +79,7 @@ public static function getHelpCopy()
 
 A combination of <comment>--dry-run</comment> and <comment>--diff</comment> will
 display a summary of proposed fixes, leaving your files unchanged.
+Optionally the <comment>--diff</comment> can be used to specify the diff. output format; ``--diff diff`` (default) or ``--diff diff-short``.
```
keradus: `--diff=diff` is werid

keradus:
could you show example before and after please ?

src/Console/Command/FixCommandHelp.php
SpacePossum: I'm out of ideas TBH

julienfalque: What about `normal` vs. `short` instead of `diff` vs. `short-diff`?

SpacePossum: not sure, as what is `normal` is in eye of the beholder ;)

keradus: `sbd` (so `SebastianBergmannDiffer) and `short-sbd` ?

keradus:

please rebase

src/Console/Command/FixCommandHelp.php
julienfalque: `SebastianBergmannDiff` is an internal detail of how diff is implemented, this should not be exposed to end user IMO. Plus, neither `SebastianBergmannDiff` nor `sbd` tells me what the actual result will look like.

tests/Console/ConfigurationResolverTest.php
```diff
@@ -944,6 +944,61 @@ public function testResolveCommandLineInputOverridesDefault()
...
+     */
+    public function testResolveDiffer($expected, $differConfig)
+    {
+        $resolver = new ConfigurationResolver(
+            $this->config,
+            array('diff' => $differConfig),
+            ''
+        );
+
+        $this->assertInstanceOf($expected, $resolver->getDiffer());
```
julienfalque: This test does not guarantee that the differ instance uses the correct mode. Not sure how we could do that though.

tests/Console/ConfigurationResolverTest.php
```diff
@@ -944,6 +944,61 @@ public function testResolveCommandLineInputOverridesDefault()
...
+                'diff-short',
+            ),
+        );
+    }
+
+    public function testUnknownDiffConfiguration()
+    {
+        $this->setExpectedExceptionRegExp(
+            '\PhpCsFixer\ConfigurationException\InvalidConfigurationException',
+            '#^Differ must be \"diff\" \(default\) or \"diff-short\", got \"_unknown_\"\.$#'
```
julienfalque: `"` character does not need to be escaped.

tests/Console/ConfigurationResolverTest.php
```diff
@@ -944,6 +944,61 @@ public function testResolveCommandLineInputOverridesDefault()
...
+            array(
+                '\PhpCsFixer\Differ\SebastianBergmannDiffer',
+                'diff-short',
+            ),
+        );
+    }
+
+    public function testUnknownDiffConfiguration()
+    {
+        $this->setExpectedExceptionRegExp(
```
julienfalque: I would define the expected exception *after* the resolver instantiation.

src/Console/ConfigurationResolver.php
```diff
@@ -251,7 +251,25 @@ public function getConfigFile()
...
+                    $this->differ = new SebastianBergmannDiffer(true);
+
+                    break;
+                case 'diff-short':
+                    $this->differ = new SebastianBergmannDiffer(false);
+
+                    break;
+                default:
+                    throw new InvalidConfigurationException(sprintf(
+                        'Differ must be "diff" (default) or "diff-short", got "%s".',
```
julienfalque: Not sure if saying that `diff` is the default is relevant here.

README.rst
```diff
@@ -181,6 +181,7 @@ Complete configuration for rules can be supplied using a ``json`` formatted stri
 
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
+Optionally the ``--diff`` can be used to specify the diff. output format; ``--diff diff`` (default) or ``--diff diff-short``.
```
julienfalque: The syntax for other options in the document is `--name=value`, could you please use it here as well for consistency?

tests/Console/ConfigurationResolverTest.php
SpacePossum: we can't, that is why I left it out

julienfalque: Maybe we can test the result of `$resolver->getDiffer()->diff($old, $new)`?

keradus: @SpacePossum , we can by introduce new class `ShortSebastianBergmannDiffer`

@julienfalque , this is a task for `*DifferTest` classes

SpacePossum: fixed

src/Console/ConfigurationResolver.php
```diff
@@ -103,18 +104,19 @@
      */
     private $options = array(
         'allow-risky' => null,
+        'cache-file' => null,
         'config' => null,
+        'diff' => null,
+        'differ' => null,
```
keradus: this key is not used

src/Console/ConfigurationResolver.php
```diff
@@ -253,7 +255,25 @@ public function getConfigFile()
     public function getDiffer()
     {
         if (null === $this->differ) {
-            $this->differ = false === $this->options['diff'] ? new NullDiffer() : new SebastianBergmannDiffer();
+            switch ($this->options['diff']) {
+                case false:
```
keradus: could this option be also null ?

keradus:
Travis is sad

src/Console/ConfigurationResolver.php
SpacePossum: Yes, since we allow passing an `array` it could be even stranger like Resources.
 However, since we pass directly values as parsed by InputOption it should not be possible to be `null`. Updating to support sounds fair to me though, I'll make the change.

src/Console/ConfigurationResolver.php
SpacePossum: good catch, left over from previous build up

src/Console/ConfigurationResolver.php
keradus: if it's not possible to get it from `InputOption` then current way is fine for me. we have default case with exception

keradus:
Thank you @SpacePossum.

PR #2555, opened 2017-02-23
Separate align binary operators and space binary operators

crhallberg:
Hello! Congratulations on the 2.x release!

We use PHP-CS-Fixer for the [VuFind](https://github.com/vufind-org/vufind) project and we saw that the new update completely changed one of the fixers we use, the binary space fixer. We talked to the community and they would prefer the original functionality, with the additional option to align.

I've done my best attempt to separate the binary aligning fixer into the align-unalign fixer new to 2.0 and a fixer that ensures that there is at least one space on either side of the binary operators.


crhallberg:
I am having a hard time telling if the errors that travis and appveyor are the result of this PR. Also, how did you get the Update Branch functionality here on GitHub? That would be handy on our project.

julienfalque:
As far as I understand, your PR basically renames `binary_operator_spaces` to `align_operator_spaces` (which is a BC break) and then recreates the same `binary_operator_spaces` but without alignment feature. I'm not sure I understand what's the point of this, I think you just need to update your configuration instead.

SpacePossum:
Hi @crhallberg ,

Thanks for you proposal by PR!

I looked at the project and the configuration you've there now for the 1.x line.
As you saw `operators_spaces` was rename to `binary_operator_spaces` (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.0.0/UPGRADE.md).

In that process we also merged two of the `alignment` fixers into `binary_operator_spaces`. As a result of that you can no longer use the `binary_operator_spaces` without stating what you want related to the alignment.
Configuring both alignment options (`align_equals`, `align_double_arrow`) to `false` will give you the functionality as you have on the 1.x with one exception;
for example
```php
$a = 1    + 2;
```
was not fixed on 1.x
On 2.x it becomes
```php
$a = 1 + 2;
```

Your PR suggest to go back to the old situation. This would be a BC break but would also reintroduce the problem that lead to the merging of the fixers of the first place.
We have more RFC's for the `binary_operator_spaces` so its great when this is worked on. Could you maybe elaborate on why you want to split up the fixers?

On a side note to answer your questions
>Also, how did you get the Update Branch functionality here on GitHub?

I don't think we did anything special for this, AFAIK this is rolled out for all GH repo's, however it might not always show (for example when merging would lead to conflicts)

>I am having a hard time telling if the errors that travis and appveyor are the result of this PR

They are, these are related to the new fixer. Not to worry though, when we know what will be test way to move forward I'm happy to help out with those.

crhallberg:
@julienfalque @SpacePossum Thanks for clarifying a few things. It's true, this PR was an attempt to bring back the old functionality. My apologies for the BC break, I just didn't want to make an empty request.

I'm not sure what the previous problems were that led to a merging of this functionality but the option to check spacing without alignment is something our dev team will definitely miss. No hard feelings if this isn't the best approach to restore this functionality.

I'll be here when you've decided how to move forward. :+1:

demiankatz:
@SpacePossum, just to elaborate on the reasoning behind this pull request: in our project, we like to make alignment of `=` and `=>` optional. There are some cases where aligning improves readability, and others where it causes problems. We would like to ensure that operators have at least one space on each side, but we don't want to force people to ALWAYS or NEVER align; with our codebase, that decision requires some human discretion. The old 1.x behavior was helpful for this; the 2.x is a bit more strict than we would like.

Obviously we have the option of simply turning off `binary_operator_spaces`, but it would be nice to be able to keep the benefits that work for us without being forced to enforce alignment standards one way or the other.

In any case, thanks for taking a look at this, and for working on what we have found to be a very helpful tool!

SpacePossum:
@crhallberg  @demiankatz thanks for the details about the reason behind your PR.
Currently alignment is optional, however this does indeed come with non-optional un-alignment of other binary operators (so, indeed `always` or `never`).
Knowing what people are missing is of great value to know where further development effort is appreciated, so thanks for taking the time to talk here with the community, especially by putting time and effort into a PR!

I also took a shot in the form of a PR to fix some issues with this fixer and to bring more options for it (back) into the tool, https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573  Please don't read this as I'm telling how things should be done, I'm just exploring what routes are feasible and would match peoples request/requirements.

demiankatz:
@SpacePossum, no worries about "telling how things should be done." We opened this to start a conversation and appreciate someone more familiar with the project proposing an alternative solution!

A quick glance at #2573 suggests to me that it should meet our needs, if I'm reading correctly that `align` means "always", `align_single_space` means "never" and `align_single_space_minimal` refers to the legacy behavior. My only suggestion is that perhaps calling this the "alignment strategy" isn't the clearest possible terminology, since we're only actually aligning things in one case. Perhaps "whitespace strategy" would be a broader, more descriptive term, and then you could use values like `align_operators`, `single_space` and `single_space_minimal`. Anyway, just a thought!

In any case, thanks again for taking a time to consider our use case. I'll wait and see if @crhallberg has any further thoughts, but if he does not object, I'd be fine with closing this and focusing on finishing up #2573. If there's anything else we can do to help, please let us know.

crhallberg:
I agree that the new PR seems like a better way forward. I left a comment over there to help clarify what I think we're expecting. Thank you!

demiankatz:
Status update: we are currently waiting for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573 to be completed and merged so that we can move forward with this. It should give us the flexible alignment functionality that we desire.

keradus:
Closes due to merge of #2573

Issue #2457, opened 2017-01-11
code moved into comment when bracket is after comment


thedarkwinter:
The PHP version you are using:
```
PHP 7.0.8-0ubuntu0.16.04.3 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.8-0ubuntu0.16.04.3, Copyright (c) 1999-2016, by Zend Technologies
```

PHP CS Fixer version you are using:
```
PHP CS Fixer version 2.0.0 by Fabien Potencier and Dariusz Ruminski (f3baf72)
```

`php-cs-fixer fix reproduce.php`

* Code
```
<?php
if (1==1) // this is terrible code :P
{ $a = 'hello'; }
echo $a;
```

* Becomes >>

```
<?php
if (1==1) { // this is terrible code :P $a = 'hello';
}
echo $a;
```

Its code breaking... hopefully this helps you dig it out!

Thanks,
Michael

SpacePossum:
Thanks for the report!
Confirmed on master, `braces` fixer does this.

SpacePossum:
(ping @keradus please tag as bug as I does brake code on fixing :|  )

keradus:
labeled

PR #2557, opened 2017-02-24
BracesFixer - Better comments handling

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2457

### DIFF
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -155,8 +155,8 @@ $negative = function ($item) {
             if ($token->isGivenKind($controlTokens)) {
                 $prevIndex = $this->findParenthesisEnd($tokens, $index);
             } elseif (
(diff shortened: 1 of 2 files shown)


src/Fixer/Basic/BracesFixer.php
```diff
@@ -179,17 +179,22 @@ private function fixCommentBeforeBrace(Tokens $tokens)
             }
 
             $tokenTmp = $tokens[$braceIndex];
-            $trimIndex = $tokens->getPrevNonWhitespace($braceIndex);
-            $tokens[$trimIndex]->setContent(rtrim($tokens[$trimIndex]->getContent()));
```
SpacePossum: my guess is this is legacy code from 1.x branch when comments still had a trailing line break, as opposed to on 2.x where these are now two tokens

keradus:
Thank you @SpacePossum.

Issue #2472, opened 2017-01-22
Error while using binary_operator_spaces 


AntoineLemaire:
Each time I want set `align_double_arrow` or `align_equals` to true in my `.php_cs`, I got an error:

```
Loaded config default from "/var/www/project/.php_cs".
E
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error

Fixed all files in 0.013 seconds, 6.000 MB memory used

Files that were not fixed due to errors reported during fixing:
   1) /var/www/project/CSTestController.php

```

With this file https://gist.github.com/AntoineLemaire/d8153d17fca3b1e67ef2dc98810180b3

And this `.php_cs`

```
<?php

return PhpCsFixer\Config::create()
    ->setUsingCache(false)
    ->setRules(array(
        '@Symfony' => true,
        'array_syntax' => array('syntax' => 'short'),
        'unalign_equals' => false,
        'phpdoc_order' => true,
        'phpdoc_to_comment' => false,
        'binary_operator_spaces' => array(
            'align_double_arrow' => true,
            'align_equals' => true,
        ),
    ))
;
```
I'm using php 7.0.13 
```
PHP 7.0.13-0ubuntu0.16.04.1 (cli) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.13-0ubuntu0.16.04.1, Copyright (c) 1999-2016, by Zend Technologies
```
and php-cs-fixer 2.0.0

```
php-cs-fixer -vvv fix /var/www/project/CSTestController.php
```

SpacePossum:
I failed to reproduce the error with a checkout version @ v2.0.0 using PHP 7.1.0-5+deb.sury.org~xenial+1 (cli) ( NTS ) on Ubuntu 16.04.1 LTS. Anyone else got time to help out?

keradus:
@AntoineLemaire , would it be fine to you to reveal the `/var/www/project/CSTestController.php` file content ?

SpacePossum:
He did, see the link to the gist ;)

keradus:
ah, sorry

AntoineLemaire:
It's work with 7.0.15-1
```
[...]

PR #2558, opened 2017-02-24
require symfony/polyfill-xml

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2472


keradus:
why not suggest `ext-xml` for performance ?

keradus:
Thank you @SpacePossum.

PR #2560, opened 2017-02-25
PhpdocNoAliasTagFixer - Fix circular replacements detection

julienfalque:
Using 
```php
['a' => 'b', 'c' => 'a']
```
triggers an error but using 
```php
['a' => 'b', 'b' => 'c']
```
does not.


SpacePossum:
Looking good :) please escape the dot of exception message in the regex, other than that this looks good to me, thanks

keradus:
Thank you @julienfalque.

PR #2562, opened 2017-02-26
Add schema.json

keradus:
we already have schema for junit, let us add one for json

### DIFF
--- a/doc/schema.json
+++ b/doc/schema.json
@@ -4,7 +4,7 @@
     "properties": {
         "files": {
             "type": "array",
-            "uniqueItems": false,
+            "uniqueItems": true,
             "items": {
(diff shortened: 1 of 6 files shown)


doc/schema.json
```diff
@@ -0,0 +1,47 @@
...
+                    "appliedFixers": {
+                        "type": "array",
+                        "uniqueItems": true,
+                        "items": {
+                            "type": "string"
+                        }
+                    }
+                },
+                "required": [
+                    "name"
```
SpacePossum: `appliedFixers` is required as well I think?

keradus: it's there only when one run command with verbose option, if not, then the key is not present

PR #2563, opened 2017-02-26
Add deprecation notices tests

julienfalque:
~Requires #2500.~


keradus:
I think you could easily extract bridge and new tests to be not dependent on referred PR.
If not (or it would require more work), let us freeze this PR until it's parent is merged, at least in terms of reviewing it.

julienfalque:
Yeah, I've been a bit lazy here ;)
Rebasing.

SpacePossum:
:+1:

keradus:
Thank you @julienfalque.

Issue #2484, opened 2017-01-30
Improve linter failure reporting


curry684:
As mentioned in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2090#issuecomment-275627493 - when linting fails before running a fix operation the output only states the file that causes an error:
```
Files that were not fixed due to errors reported during linting before fixing:
   1) MyBundle/Entity/SomeEntity.php
```
The `LintingException` caught [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3a394e5b8dd8c11065d4c04223714a443beb4a55/src/Runner/Runner.php#L143) does however include the full details of the problem, specifically line number and exact error. It should be logged and shown in the error report.

SpacePossum:
for ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1176

Issue #1270, opened 2015-06-18
Suggestion of new feature: Show name and path of invalid files


jatubio:
Hi,

When i get one message like this:

```
D:\Proyectos\Repositorios_Git\Laravel\Salud Digital\sdf>pflaravel -vvv
Loaded config from ".php_cs"
FII
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) tests\functional\boletin\BoletinControllerStoreTest.php (method_argument_space)
Fixing time per file:
[0.002 s] app\Console\Commands\Inspire.php
[0.001 s] app\Console\Kernel.php
[0.001 s] app\Events\Event.php
(...)
```

I can't get the names of 'I' (invalid) files. I tried with `-v`, `-vv` and `-vvv` without results.

Show invalid files would be great.

funivan:
We can see files with errors at the bottom. My command: `./console.php cs:fix --dry-run --diff -vvv`
```
.EI..
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
Fixing time per file:
[0.021 s] Controller/Domains.php
[0.030 s] Controller/SpamRegs.php
[0.018 s] Controller/Banners.php
[0.023 s] Controller/SpamIps.php

Fixed all files in 0.095 seconds, 8.250 MB memory used

Files that were not fixed due to internal error:
   1) Controller/SpamRegs.php
```

Possible it will be better to see all files list with time and statuses. For example
```
[0.021 s] .  Controller/Domains.php
[0.030 s] E  Controller/SpamRegs.php
[0.018 s] E  Controller/Banners.php
[0.023 s] I  Controller/SpamIps.php

```

keradus:
Showing files that didn't changed is pointless to me, why you want to show them ?

jatubio:
Yes, @funivan, i like something like this :)

@keradus, i you think that there is no need to show files that didn't changed, why show the `I` about them? I think it's better have one option to see what files are ignored or don't nothing about them (not even `I`). Now, for example, i have some projects with a lot of files. And, when I run PHP-CS-Fixer on them, I get a `I` in then last file and I'm very curious to know what they are. Since projects seems to run perfectly (Although having one file with 'invalid file sintax').

funivan:
Other way is eventDispathcer. 
[...]

Issue #1176, opened 2015-04-27
Output exceptions when verbosity level is increased


evert:
I'm getting exceptions on several files (internal error), but I don't see a way to further debug those errors.

I assume that there's an exception being caught, it would be great if I could see it.

keradus:
Yes, they are hidden for end-user.
Please, take a look into [ErrorsManager](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1.8/Symfony/CS/ErrorsManager.php)

evert:
Do you have any additional documentation about using that?

keradus:
No (just one that's inside file itself), this is class just to store that info you saw on console.
When I want to debug I just add some dumping in `report` method.

BTW, in 2.x version I would like to have some autogenerated bug reports with all info that is not needed for end-user, but essential to fix it.

PR #2566, opened 2017-02-27
Show more details on errors and exceptions.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1270
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1176
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2484

No BC breaks as we have no promise on stdErr ouput

### DIFF
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -56,19 +56,34 @@ final class ErrorOutput
         foreach ($errors as $i => $error) {
             $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
             if ($showDetails && null !== $e = $error->getSource()) {
-                $this->output->writeln(array(
-                    '      <bg=yellow;fg=black;>Details</>',
-                    sprintf('      <comment>Class</comment>    %s', $this->prepareOutput(get_class($e))),
-                    sprintf('      <comment>Message</comment>  %s', $this->prepareOutput($e->getMessage())),
-                    sprintf('      <comment>Code</comment>     %d', $e->getCode()),
-                    sprintf('      <comment>File</comment>     %s:%d', $this->prepareOutput($e->getFile()), $e->getLine()),
-                ));
+                $class = sprintf('[%s]', get_class($e));
+                $message = $e->getMessage();
+                if (0 !== $code = $e->getCode()) {
+                    $message .= " ($code)";
+                }
+
+                $length = max(strlen($class), strlen($message));
+                $lines = array(
+                    '',
+                    $class,
+                    $message,
+                    '',
+                );
+
+                $this->output->writeln('');
+                foreach ($lines as $line) {
+                    if (strlen($line) < $length) {
+                        $line .= str_repeat(' ', $length - strlen($line));
+                    }
+
+                    $this->output->writeln(sprintf('      <error>  %s  </error>', $this->prepareOutput($line)));
+                }
 
                 if ($showTrace && !$e instanceof LintingException) { // stack trace of lint exception is of no interest
-                    $this->output->writeln('      <bg=yellow;fg=black;>Stack trace</>');
+                    $this->output->writeln('');
                     $stackTrace = $e->getTrace();
                     foreach ($stackTrace as $trace) {
-                        $this->outputTrace($trace, '      ');
+                        $this->outputTrace($trace);
                     }
                 }
             }
@@ -76,25 +91,23 @@ final class ErrorOutput
     }
 
     /**
-     * @param array  $trace
-     * @param string $indent
+     * @param array $trace
      */
-    private function outputTrace(array $trace, $indent)
+    private function outputTrace(array $trace)
     {
-        if (isset($trace['file'])) {
-            $this->output->writeln(sprintf('%s<comment>File</comment>      %s:%d', $indent, $this->prepareOutput($trace['file']), $trace['line']));
-        }
-
         if (isset($trace['class'], $trace['type'], $trace['function'])) {
             $this->output->writeln(sprintf(
-                '%s<comment> Method</comment>   %s%s%s',
-                $indent,
+                '      <comment>%s</comment>%s<comment>%s()</comment>',
                 $this->prepareOutput($trace['class']),
                 $this->prepareOutput($trace['type']),
                 $this->prepareOutput($trace['function'])
             ));
         } elseif (isset($trace['function'])) {
-            $this->output->writeln(sprintf('%s<comment> Function</comment> %s', $indent, $this->prepareOutput($trace['function'])));
+            $this->output->writeln(sprintf('      <comment>%s()</comment>', $this->prepareOutput($trace['function'])));
+        }
+
+        if (isset($trace['file'])) {
+            $this->output->writeln(sprintf('        in <info>%s</info> at line <info>%d</info>', $this->prepareOutput($trace['file']), $trace['line']));
         }
     }
 
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -51,39 +51,50 @@ final class ErrorOutput
             $process
         )));
 
-        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
-        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
+        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
+        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_DEBUG;
         foreach ($errors as $i => $error) {
             $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
-            if ($showDetails && null !== $e = $error->getSource()) {
-                $class = sprintf('[%s]', get_class($e));
-                $message = $e->getMessage();
-                if (0 !== $code = $e->getCode()) {
-                    $message .= " ($code)";
-                }
-
-                $length = max(strlen($class), strlen($message));
-                $lines = array(
-                    '',
-                    $class,
-                    $message,
-                    '',
-                );
-
-                $this->output->writeln('');
-                foreach ($lines as $line) {
-                    if (strlen($line) < $length) {
-                        $line .= str_repeat(' ', $length - strlen($line));
+            if ($showDetails) {
+                $e = $error->getSource();
+                if (null !== $e) {
+                    $class = sprintf('[%s]', get_class($e));
+                    $message = $e->getMessage();
+                    $code = $e->getCode();
+                    if (0 !== $code) {
+                        $message .= " ($code)";
                     }
 
-                    $this->output->writeln(sprintf('      <error>  %s  </error>', $this->prepareOutput($line)));
-                }
+                    $length = max(strlen($class), strlen($message));
+                    $lines = array(
+                        '',
+                        $class,
+                        $message,
+                        '',
+                    );
 
-                if ($showTrace && !$e instanceof LintingException) { // stack trace of lint exception is of no interest
                     $this->output->writeln('');
-                    $stackTrace = $e->getTrace();
-                    foreach ($stackTrace as $trace) {
-                        $this->outputTrace($trace);
+
+                    foreach ($lines as $line) {
+                        if (strlen($line) < $length) {
+                            $line .= str_repeat(' ', $length - strlen($line));
+                        }
+
+                        $this->output->writeln(sprintf('      <error>  %s  </error>', $this->prepareOutput($line)));
+                    }
+
+                    if ($showTrace && !$e instanceof LintingException) { // stack trace of lint exception is of no interest
+                        $this->output->writeln('');
+                        $stackTrace = $e->getTrace();
+                        foreach ($stackTrace as $trace) {
+                            if (isset($trace['class'], $trace['function']) && 'Symfony\Component\Console\Command\Command' === $trace['class'] && 'run' === $trace['function']) {
+                                $this->output->writeln('      [ ... ]');
+
+                                break;
+                            }
+
+                            $this->outputTrace($trace);
+                        }
                     }
                 }
             }
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -1,0 +1,113 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Console\Output;
+
+use PhpCsFixer\Error\Error;
+use PhpCsFixer\Linter\LintingException;
+use Symfony\Component\Console\Formatter\OutputFormatter;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class ErrorOutput
+{
+    /**
+     * @var OutputInterface
+     */
+    private $output;
+
+    /**
+     * @var bool
+     */
+    private $isDecorated;
+
+    public function __construct(OutputInterface $output)
+    {
+        $this->output = $output;
+        $this->isDecorated = $output->isDecorated();
+    }
+
+    /**
+     * @param string  $process
+     * @param Error[] $errors
+     */
+    public function listErrors($process, array $errors)
+    {
+        $this->output->writeln(array('', sprintf(
+            'Files that were not fixed due to errors reported during %s:',
+            $process
+        )));
+
+        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
+        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
+        foreach ($errors as $i => $error) {
+            $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
+            if ($showDetails && null !== $e = $error->getSource()) {
+                $this->output->writeln(array(
+                    '      <bg=yellow;fg=black;>Details</>',
+                    sprintf('      <comment>Class</comment>    %s', $this->prepareOutput(get_class($e))),
+                    sprintf('      <comment>Message</comment>  %s', $this->prepareOutput($e->getMessage())),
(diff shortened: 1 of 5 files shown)


src/Error/Error.php
```diff
@@ -47,28 +47,43 @@
...
-     * @param string $filePath
+     * @var \Exception|\Error|\Throwable|null
      */
-    public function __construct($type, $filePath)
+    private $source;
+
+    /**
+     * @param int                               $type
+     * @param string                            $filePath
+     * @param \Exception|\Error|\Throwable|null $source
```
julienfalque: As `\Exception` and `\Error` both implement `\Throwable`, isn't `\Throwable|null` sufficient?

src/Error/Error.php
```diff
@@ -47,28 +47,43 @@
...
-    public function getType()
+    public function getFilePath()
     {
-        return $this->type;
+        return $this->filePath;
     }
 
     /**
-     * @return string
+     * @return \Exception|\Error|\Throwable|null
```
julienfalque: Same here.

src/Runner/Runner.php
```diff
@@ -250,16 +250,17 @@ private function fixFile(\SplFileInfo $file, LintingResultInterface $lintingResu
     /**
      * Process an exception that occurred.
      *
-     * @param string $name
+     * @param string                       $name
+     * @param \Exception|\Error|\Throwable $e
```
julienfalque: Same here.

julienfalque:
👍 Looks nice, could you add a screenshot of the result? Thanks :)

src/Error/Error.php
SpacePossum: Throwable was added in PHP7, this way of typehinting keeps SCA happy when checking on older (but still supported by the project) versions of PHP

SpacePossum:
lint failure example:
screenshot from 2017-02-27 12-10-44

broken fixer example:
screenshot from 2017-02-27 12-11-25

julienfalque:
I find the stacktrace hard to read. Wouldn't it be easier with functions and files on the same line (and thus without the _Function_/_File_ label column), maybe with different colors? What about showing `class::method()` instead of just `method`?

SpacePossum:
The problem is that it might be closures (and in theory functions) and we get the file and not the class, so making it `class::method()` will be hard if possible at all.

julienfalque:
Are your sure? [Documentation seems to say otherwise](http://php.net/manual/en/function.debug-backtrace.php#refsect1-function.debug-backtrace-returnvalues).

SpacePossum:
ah, indeed you are right;
https://3v4l.org/rTa9C

will update

src/Error/Error.php
keradus: I'm fine with `?\Throwable`

julienfalque: Didn't know such syntax was possible in PHPDoc, is it defined somewhere?

keradus: PHPDoc spec or PSR doc-proposal
`@param <type> [$name] [descr]`

`<type>` is valid PHP Type. `?\Throwable` is valid PHP7.1 type.

julienfalque: As far as I understand, [ABNF definition in proposed PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#abnf) does not seem to allow this syntax. Is it supported by common IDEs like PhpStorm?

keradus: it was not updated for php7.1, as it is (sadly) kinda abandoned.
You may use a type of php now, right? `?int` is a valid type now

keradus: not sure about that one given IDE, but I don't write code to works nicely under one IDE.

SpacePossum:
```
\Throwable|null
```
is fine by me also as we would've likely more IDE support. Agreed we don't write code for IDE's, but this feels like a good compromise to me (and it would be more inline what we've now in the project).

julienfalque: I agree we don't write code for IDEs too, but as `\Throwable|null` is valid and is well supported, there is no need to use an alternative syntax IMO.

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,118 @@
...
+            $process
+        )));
+
+        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
+        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
+        foreach ($errors as $i => $error) {
+            $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
+            if ($showDetails && null !== $e = $error->getSource()) {
+                $this->output->writeln(array(
+                    '      '.($this->isDecorated ? '<bg=yellow;fg=black;>Details</>' : 'Details'),
```
julienfalque: Is the `isDecorated` check required? That's a real question, I would expect the Symfony Console component to do it internally but I don't know if it does.

SpacePossum: thanks for checking, will update!
I think there are more major projects that would benefit from that optimization :)

SpacePossum:
updated output

output

julienfalque:
I'd like to make a few suggestions regarding the output.

The _Details_ section could be simplified by:
* displaying the exception class, message and code on a single line like Symfony does, e.g. `[FooException] Bar (123)`;
* removing the exception file location, which is redundant with the stacktrace.

I still find hard to read the stacktrace. I think it would be better to have each file/function on a single line like regular stacktraces, e.g. `Foo->bar /path/to/Foo.php:123`, with different colors for the code part and the location part. Lines will be longer but most people have large enough screens, I suppose.

Those changes would even remove the need to have section titles, empty line separation should be enough.

What do you think?

julienfalque:
> This would create long lines when we keep the namespace

True, what about using an empty line between class and message (if I remembed correctly, Symfony actually displays exceptions that way):
```
[Some\Namespaced\FooException]

Some message that we could wrap at 80 characters for example. (123)
```

> when ever I get more pixels I add more consoles rather than make the consoles wider

Well, same for me actually, but I still want to give it a try :)

SpacePossum:
fair points, however you example doesn't show the file where the exception is thrown, only class, message and line number :)

SpacePossum:
oh awesome, on HHVM the linenumbers are resolved with diff. values \o/

julienfalque:
I played a little with your branch, here's what I got:

One-line stacktrace

or

Multiline stacktrace

The exception part is a bit scary and maybe too big, but I like the stacktrace. I think it can still be improved though, for example with the `->` in white to easily distinguish class from method, and maybe replace `:` with `at line`.

What do you think?

SpacePossum:
I like the second version and agree on all points :)
Can you PR my PR with the changes or talk me through how you got that result?

julienfalque:
Cool! Well I made pure _quick & dirty_ changes just to get the result so I must clean a bit before opening a PR 😅. I can't do it right now but if you want to do it in the meantine, go ahead, I just changed some `writeln()` calls and used basic `<comment>` and `<info>` coloring.

julienfalque:
Thanks @SpacePossum :)

By the way, I checked how Symfony displays exceptions, just to make sure it wouldn't be easier to just reuse it. The display is closer to the first version I proposed above. But `Application::renderException()` does not seem to support `\Throwable` so we can't use it anyway.

SpacePossum:
thanks for checking @julienfalque :+1:  :)

keradus:
for that purpose any throwable could be converted into `FatalThrowableError` exception

julienfalque:
Ah you're right, didn't know that. Shall we use it then?

keradus:
if it would allow us to reuse the code instead of writing our own - why not ?

SpacePossum:
Capturing the output as generated by SF for each error/exception raised and than reformatting for our need is probably more work than writing our own logic for the output. 
As much as I (try to) like the stand-alone Console component of SF to apply DRY, this is another example that might not fit. Happy to be proven otherwise, till than I think we are on good direction to provide the output wanted.

keradus:
Why the output that Symfony provides us out of the box is not matching our requirements ?

SpacePossum:
maybe I misunderstood, what part of SF exception output should we consider to reuse?

keradus:
What I was answering to:

>> But `Application::renderException()` does not seem to support `\Throwable` so we can't use it anyway.
> 
> any throwable could be converted into `FatalThrowableError` exception (so we could reuse `renderException`

Why can't we reuse code from Symfony components and need to do it on our own ?

SpacePossum:
https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Console/Application.php#L641

- output of the trace is based on verbosity level of the output, not on application logic
- length of trace cannot be controlled
- prints out details about the command, making it not suitable to be called multiple times

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,134 @@
...
+        $this->output->writeln(array('', sprintf(
+            'Files that were not fixed due to errors reported during %s:',
+            $process
+        )));
+
+        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
+        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
+        foreach ($errors as $i => $error) {
+            $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
+            if ($showDetails && null !== $e = $error->getSource()) {
```
keradus: do not assign inside condition

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public function listErrors($process, array $errors)
+    {
+        $this->output->writeln(array('', sprintf(
+            'Files that were not fixed due to errors reported during %s:',
+            $process
+        )));
+
+        $showDetails = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
+        $showTrace = $this->output->getVerbosity() >= OutputInterface::VERBOSITY_VERY_VERBOSE;
```
keradus: I would say that trace is debug.
then, why not `details => vv, showtrace => vvv` instead ?

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,134 @@
...
+        foreach ($errors as $i => $error) {
+            $this->output->writeln(sprintf('%4d) %s', $i + 1, $error->getFilePath()));
+            if ($showDetails && null !== $e = $error->getSource()) {
+                $class = sprintf('[%s]', get_class($e));
+                $message = $e->getMessage();
+                if (0 !== $code = $e->getCode()) {
```
keradus: same here

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,134 @@
...
+
+                    $this->output->writeln(sprintf('      <error>  %s  </error>', $this->prepareOutput($line)));
+                }
+
+                if ($showTrace && !$e instanceof LintingException) { // stack trace of lint exception is of no interest
+                    $this->output->writeln('');
+                    $stackTrace = $e->getTrace();
+                    $stackTraceCount = count($stackTrace);
+                    foreach ($stackTrace as $trace) {
+                        if ($stackTraceCount > 4 && isset($trace['class'], $trace['function']) && 'Symfony\Component\Console\Command\Command' === $trace['class'] && 'run' === $trace['function']) {
```
keradus: having a dictionary:
```php
$isTooLong = $stackTraceCount > 4;
$isCommandRun = isset($trace['class'], $trace['function']) && 'Symfony\Component\Console\Command\Command' === $trace['class'] && 'run' === $trace['function']) ;
```

`$isTooLong || $isCommandRun` - display up to 4 traces, stop on symfony command run
`$isCommandRun` - stop on symfony command run, regardless it's 1 or 10 traces
`$isTooLong && $isCommandRun` - stop on symfony command run but only if we displayed 4 traces, if we would display less traces it would require to display traces from internal symfony code. what is the value here?

tests/Console/Output/ErrorOutputTest.php
```diff
@@ -0,0 +1,111 @@
...
+        $output = new StreamOutput(fopen('php://memory', 'w', false));
+        $output->setDecorated(false);
+        $output->setVerbosity($verbosityLevel);
+
+        $errorOutput = new ErrorOutput($output);
+        $errorOutput->listErrors($process, array($error));
+
+        rewind($output->getStream());
+        $displayed = stream_get_contents($output->getStream());
+        $displayed = str_replace(PHP_EOL, "\n", $displayed); // normalize line breaks
```
keradus: please describe why it's needed

tests/Console/Output/ErrorOutputTest.php
```diff
@@ -0,0 +1,111 @@
...
+    public function provideTestCases()
+    {
+        $lineNumber = __LINE__;
+        list($exceptionLineNumber, $error) = $this->getErrorAndLineNumber(); // note: keep call and __LINE__ separated with one line break
+        ++$lineNumber;
+
+        return array(
+            array($error, OutputInterface::VERBOSITY_NORMAL, $lineNumber, $exceptionLineNumber, 'VN'),
+            array($error, OutputInterface::VERBOSITY_VERBOSE, $lineNumber, $exceptionLineNumber, 'VV'),
+            array($error, OutputInterface::VERBOSITY_VERY_VERBOSE, $lineNumber, $exceptionLineNumber, 'VVV'),
```
keradus: verbosity debug is missing

keradus:
I like the proposed changes. Please resolve raised issues from my PR.
Also, changing base branch to master, as this is not a bugfix

src/Console/Output/ErrorOutput.php
SpacePossum: >then, why not details => vv, showtrace => vvv instead ?

this is the current situation, note that `vvv` is not `VERBOSITY_DEBUG` but `VERBOSITY_VERY_VERBOSE`, one cannot set the output to debug level from the command line

src/Console/Output/ErrorOutput.php
SpacePossum: I do not understand the dictionary part.
The code is;
If the exception is thrown somewhere inside of SF, than the trace is up to 4 entries, show those.
If the exception is thrown from the Fixer namespace (> 4 trace entries) don't bother the printing the first four as it is of no interest how the command executing within SF is done, only what goes wrong inside the fixer app itself.

tests/Console/Output/ErrorOutputTest.php
SpacePossum: :+1:  added

src/Console/Output/ErrorOutput.php
keradus: my dict was your condition simplified to helper vars.

keradus: > If the exception is thrown somewhere inside of SF, than the trace is up to 4 entries, show those.

what for? why 4? if the exception is thrown inside SF, show first place of SF and that's it. Even better - we know what would be the first SF that we dont care - run of our command.

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,138 @@
...
+
+                        $this->output->writeln(sprintf('      <error>  %s  </error>', $this->prepareOutput($line)));
+                    }
+
+                    if ($showTrace && !$e instanceof LintingException) { // stack trace of lint exception is of no interest
+                        $this->output->writeln('');
+                        $stackTrace = $e->getTrace();
+                        $stackTraceCount = count($stackTrace);
+                        foreach ($stackTrace as $trace) {
+                            if ($stackTraceCount > 4 && isset($trace['class'], $trace['function']) && 'Symfony\Component\Console\Command\Command' === $trace['class'] && 'run' === $trace['function']) {
```
keradus: I would remove `$stackTraceCount > 4 && ` part.
If error is inside our code, it's nice to have full stack.

src/Console/Output/ErrorOutput.php
keradus: http://symfony.com/doc/current/console/verbosity.html claims other way

src/Console/Output/ErrorOutput.php
SpacePossum: >what for?

because the trace might help finding the cause

>why 4?

because:

>we know what would be the first SF that we dont care 

which is @ 4, so that is why 4

src/Console/Output/ErrorOutput.php
SpacePossum: done

src/Console/Output/ErrorOutput.php
```diff
@@ -85,15 +85,7 @@ public function listErrors($process, array $errors)
...
-                        $stackTrace = $e->getTrace();
-                        $stackTraceCount = count($stackTrace);
-                        foreach ($stackTrace as $trace) {
-                            if ($stackTraceCount > 4 && isset($trace['class'], $trace['function']) && 'Symfony\Component\Console\Command\Command' === $trace['class'] && 'run' === $trace['function']) {
-                                $this->output->writeln('      [ ... ]');
-
-                                break;
-                            }
-
+                        foreach ($e->getTrace() as $trace) {
```
keradus: I said I would remove the `count > 4`, not breaking on Command::run

SpacePossum: :+1:  partial revert done

src/Console/Output/ErrorOutput.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        $this->output = $output;
+        $this->isDecorated = $output->isDecorated();
+    }
+
+    /**
+     * @param string  $process
+     * @param Error[] $errors
+     */
+    public function listErrors($process, array $errors)
```
keradus: This is public method so one could still pass here (during developing of PHP CS Fixer itself) an empty array.
Check here if `$errors` is not empty, and if it does then do the quick return.
That way also we won't need to check it when calling the method.

SpacePossum:
hmm I don't get git on this, on a rebase on 2.1 it markes all changes on 2.1 as in conflict with the commits here even if those files are not touched in the commits of this branch.
Shall I create a new branch? btw. I get the same on all my branches that started of from 2.1

keradus:
1. please don't create new branch (even if you would like to recreate the branch from scratch, you could still reuse original branch name to not have to open new PR)
2. your issue is because you try to rebase on 2.1, while base branch is... master. just rebase on master (this is not a bugfix, so base branch shall not be 2.1)

SpacePossum:
ah yes, I forgot the target is master because of the branch name :man_facepalming:

keradus:
ping @SpacePossum , could you rebase at Thursday please ?

keradus:
Thanks @SpacePossum and @julienfalque

SpacePossum:
thanks @julienfalque nice to see this in!

julienfalque:
Thanks to **you** @SpacePossum: you did the job, I just added some newlines in it :p

PR #2568, opened 2017-02-27
Require PHP 5.6+

keradus:
Current tool release: v2.1.0
Proposed target of this change: v2.3.0

~Purpose of this PR is not to clean up the old code that could be cleaned after min PHP version bump, but to discuss the change of required version itself.~

I would like to hear community voice.

-------------

TODO:
- [x] bump requirement
- [x] drop checks like `defined(T_POW)`
- [x] merge test methods like `provideCases54`
- [x] use short array syntax
- [x] use `?:`
- [x] use `...`


keradus:
cc @SpacePossum @localheinz @julienfalque @stof @sstok @GrahamCampbell @greg0ire @kalessil 
and everyone willing to comment !

kalessil:
Perhaps v3.0.0 would fit more:
- then it would be possible go directly to PHP7 (jumping from PHP 5.3 to 5.6/7.0 is not really makes difference)
- 2.x branch would still serve needs of slowly migrating legacy systems;

In general, the symfony strategy for migrating to PHP7 would work as well.

GrahamCampbell:
Kill it with fire.

julienfalque:
I tend to agree with @kalessil here: I think bumping PHP version requirement is not a breaking change on its own but as we stop testing versions that are currently supported, we might introduce breaking changes later. Composer will help prevent this but it would still be a problem for people that install the tool without Composer and expect it to always work on the platform they use, at least until next major version.

So 👍 for bumping PHP requirement to whatever version we decide, as long as we bump PHP-CS-Fixer to a new major version as well.

Regarding PHP version, I'd vote 7.0. Or at least 5.6 as it is the lowest supported version.

kalessil:
+1 onece more for PHP7, I'll gladly support migrating language level to PHP7 once the way back is cut.

jtojnar:
[selfoss](https://github.com/SSilence/selfoss) targets PHP 5.4 (still supported on [Debian](https://wiki.debian.org/PHP#Available_versions)) but since PHP-CS-Fixer is a dev dependency, only run on Travis and dev machines, it should not be a problem. We use [php lint](https://github.com/JakubOnderka/PHP-Parallel-Lint) for checking the syntax on older PHP versions anyway.

GrahamCampbell:
> We use php lint for checking the syntax on older PHP versions anyway.

PHP Lint doesn't do anything at all the same as PHP-CS-Fixer. PHP Lint just checks if the files are valid PHP files, and that's it.

GrahamCampbell:
Infact, on php 7, PHP-CS-Fixer actually allows files that fail php linting through without error in some cases. For example, cases where a class property is defined twice. PHP's native linter would flag that normally.

jtojnar:
@GrahamCampbell PHP-CS-Fixer can warn about some syntax features not supported by older versions. For catching wider range of issues, one should of course use both tools.

keradus:
btw, if 7, then 7.1, cause why not, huh?

the benefits of 7, like strict types or primitives used for typehinting could be automatically applied on codebase and we could release "version for php7"  as a phar file easily (without changing the code distributed by composer or git clone).
we already do some of it on SCA job at Travis, just not build result of it to phar file.

SpacePossum:
Bumping to 5.5.9 for 2.3 sounds good to me. People can lock to version 2.2.* if they can't meet the new PHP requirement. Only thing I wonder, would we keep supporting 2.2 with bug fixes along side the 2.2+ branch?

I'd vote the highest stable PHP major+minor version available when releasing 3.x.

(hmm.. another thing to think about is the the `selfupdate` command which doesn't check PHP version when updating, still :+1: for this though)

TomasVotruba:
I'd go with PHP 7 for next release :+1: , because [promoting 5.5- is](http://php.net/supported-versions.php) promoting unknown security issues.

This would show people it's normal to upgrade to newer PHP.

Also quoting @Seldaek from https://seld.be/notes/php-versions-stats-2016-2-edition :

update-reqs

jtojnar:
@TomasVotruba In this case, it is okay to drop PHP 5 because PHP-CS-Fixer is intended for developers; PHP 5.4 is unfortunately far from being dead. It is [available](https://wiki.debian.org/PHP#Available_versions) on Debian Wheezy which runs on many shared web hostings. Wheezy is supported until [May 2018](https://wiki.debian.org/LTS) and the package was last updated at the [beginning of February](https://packages.qa.debian.org/p/php5/news/20170207T155036Z.html).

Because of the cheap shared hostings many people cannot actually upgrade and we need to maintain backports like https://github.com/fossar/twitteroauth-php54

TomasVotruba:
@jtojnar I don't talk hosting, I talk about the technology - see the link provided.

Also, hosting argument is irrelevant argument, because **they are money oriented**. They will always focus on mainstream and take in account they upgrade costs.

**It's our responsibility to push them further** so common developer doesn't have to be stuck on deprecated technologies and so they don't slow down open-source maturity cyclus.

keradus:
Thanks for the feedback.
I bumped required version of PHP to 5.6+ for line 2.x.
Bumping to 7+ will be done during releasing 3.0.

TomasVotruba:
@keradus Thanks for bumping to PHP 5.6 :+1: 

Let me know when you want to upgrade to PHP 7. I did it in few packages now, getting quite handy.

keradus:
Sorry, I didn't get second part.
Any changes for PHP requirements will be noticed in changelog and composer requirements.

TomasVotruba:
@keradus I meant: Ping me, when you need to upgrade to PHP 7. I can do it.

keradus:
take a look at our travis SCA task, we even have some fixers that automates the work ;)

TomasVotruba:
SCA task? Could you share a link?

I have few sniffs that do the work :)

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.travis.yml#L14

if those sniffs are fixers of feel free to share them if they are not duplicating existing ones

TomasVotruba:
E.g. this one: https://github.com/slevomat/coding-standard#slevomatcodingstandardtypehintstypehintdeclaration-

I [used them to upgrade ApiGen](https://github.com/ApiGen/ApiGen/pull/811/commits/162e508f7ddf2aa480c7c0815640eafe6e9c0a42)

keradus:
nah, we don't like checkers, we like fixers

TomasVotruba:
It works the same on ECS.

keradus:
ECS Auto Ltd ?

TomasVotruba:
EasyCodingStandard

keradus:
so, not related

TomasVotruba:
It works on any code.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2568#issuecomment-287648073

TomasVotruba:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2568#issuecomment-287645428

kalessil:
A couple more PHP 5.6 tweaks can be applied (I'm on #2534, so no PR this time):

[PHP_5_6_tweaks.patch.txt](https://github.com/FriendsOfPHP/PHP-CS-Fixer/files/857805/PHP_5_6_tweaks.patch.txt)

keradus:
Hi, thanks for patch. Good start, but it doesn't cover all places of syntax sugar that could be used (like `::class` is nice, but should be applied globally)

kalessil:
Right, I didn't touch tests - it's plenty of `::class` cases + 4 Elvis operators there.

keradus:
please don't focus on sugar syntax for now, the issue you mentioned will be time spent better.

SpacePossum:
this looks good to me, lets get this rebase and merged, after that the syntax sugar will come in later :)

keradus:
As discussed, php engine was bumped to 5.6.
Bumping to 7.x will be made during release of version 3, as it will bring some BC breakers (eg due to params typehinting)

Issue #2561, opened 2017-02-26
NoUnreachableDefaultArgumentValueFixer - extend docs/mark as risky


keradus:
we should mention in docs the problematic case discussed in #2485 
I'm even fine with marking the fixer as risky for that reason

SpacePossum:
Should be easy enough I think. Marking the fixer `risky` is a BC break or a bug fix? I'd vote for the latter.

keradus:
yeah, later

Issue #2485, opened 2017-01-31
php-cs-fixer is removing variables defaulted to null.


trsteel88:
The PHP version you are using:
```
PHP 7.1.0
```

PHP CS Fixer version you are using:
```
PHP CS Fixer 2.0.0
```

The command you use to run the fixer (and the configuration file you are using if any).

```
php-cs-fixer fix
```

Config:

```php
return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setRules(array(
        '@Symfony' => true,
        '@Symfony:risky' => true,
        'array_syntax' => array('syntax' => 'long'),
        'combine_consecutive_unsets' => true,
        // one should use PHPUnit methods to set up expected exception instead of annotations
        'general_phpdoc_annotation_remove' => array('expectedException', 'expectedExceptionMessage', 'expectedExceptionMessageRegExp'),
        'heredoc_to_nowdoc' => true,
        'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block'),
        'no_unreachable_default_argument_value' => true,
        'no_useless_else' => true,
        'no_useless_return' => true,
        'ordered_class_elements' => true,
        'ordered_imports' => true,
        'php_unit_strict' => true,
        'phpdoc_add_missing_param_annotation' => true,
        'phpdoc_order' => true,
        'psr4' => true,
        'strict_comparison' => true,
        'strict_param' => true,
    ))
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->exclude('tests/Fixtures')
            ->in(__DIR__)
    )
;
```

A minimal sample of valid PHP code that is not fixed correctly (if applicable).

The following code is changed from:

```php

public function sidePanelAction(Request $request, $selected_category_id, $count_posts, $show_active_only)
```

The fixed version of that code after you run the fixer and the version you expected.

```php

public function sidePanelAction(Request $request, $selected_category_id = null, $count_posts, $show_active_only)
```

--------------------

php-cs-fixer is removed the "= null" for the variables selected_category_id.

This is only an issue when rendering esi within Symfony:

```twig
{{ render_esi(controller('AcmeBlogBundle:Category:sidePanel', {'selected_category_id': category.id|default(null), 'count_posts': false, 'show_active_only': true })) }}
```

[...]

PR #2572, opened 2017-02-28
NoUnreachableDefaultArgumentValueFixer - Mark as risky

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2561

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -426,8 +426,9 @@ Choose from the list of available rules:
    | *Rule is: configurable.*
 
 * **no_unreachable_default_argument_value**
-   | In method arguments there must not be arguments with default values
+   | In function arguments there must not be arguments with default values
    | before non-default ones.
+   | *Rule is: risky.*
 
 * **no_unused_imports** [@Symfony]
    | Unused use statements must be removed.
--- a/src/Fixer/FunctionNotation/NoUnreachableDefaultArgumentValueFixer.php
+++ b/src/Fixer/FunctionNotation/NoUnreachableDefaultArgumentValueFixer.php
@@ -25,11 +25,6 @@ use PhpCsFixer\Tokenizer\Tokens;
  */
 final class NoUnreachableDefaultArgumentValueFixer extends AbstractFixer
 {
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -426,8 +426,9 @@ Choose from the list of available rules:
    | *Rule is: configurable.*
 
 * **no_unreachable_default_argument_value**
-   | In method arguments there must not be arguments with default values
-   | before non-default ones.
+   | In function and method arguments there must not be arguments with
```
keradus: method is a subtype of function

src/Fixer/FunctionNotation/NoUnreachableDefaultArgumentValueFixer.php
```diff
@@ -53,18 +48,38 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
                     '<?php
 function example($foo = "two words", $bar) {}
 '
                 ),
-            )
+            ),
+            null,
+            null,
+            null,
+            'Modifies the signature of methods/functions, risky when using systems relying on those (for example through reflection).'
```
keradus: could we ref. to example from Symfony, as the statement is not obvious for all devs.

README.rst
SpacePossum: indeed, so you want `function` only here?

src/Fixer/FunctionNotation/NoUnreachableDefaultArgumentValueFixer.php
SpacePossum: not sure SF does this, do you made have an example? (I don't use SF that much, mainly Silex)

keradus: initial post of https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2485
also, when you process input parameters for API routes.

ok, let as leave it as is then ;)

keradus:
Thank you @SpacePossum.

Issue #2559, opened 2017-02-25
TernaryOperatorSpacesFixer - Add align option(s)


sebastianbergmann:
With this [`.php_cs`](https://github.com/sebastianbergmann/phpunit/blob/master/.php_cs) `php-cs-fixer` unaligns aligned ternary operator statements:

```diff
diff --git a/src/TextUI/TestRunner.php b/src/TextUI/TestRunner.php
index ca82375..39bfb06 100644
--- a/src/TextUI/TestRunner.php
+++ b/src/TextUI/TestRunner.php
@@ -729,8 +729,8 @@ protected function handleConfiguration(array &$arguments)
             );
         }
 
-        $arguments['debug']     = isset($arguments['debug'])     ? $arguments['debug']     : false;
-        $arguments['filter']    = isset($arguments['filter'])    ? $arguments['filter']    : false;
+        $arguments['debug']     = isset($arguments['debug']) ? $arguments['debug'] : false;
+        $arguments['filter']    = isset($arguments['filter']) ? $arguments['filter'] : false;
         $arguments['listeners'] = isset($arguments['listeners']) ? $arguments['listeners'] : [];
```

Is this a bug in one of the rules I am using? Am I missing something in my [`.php_cs`](https://github.com/sebastianbergmann/phpunit/blob/master/.php_cs) to achieve what I want?

Thanks!

localheinz:
Related to https://github.com/sebastianbergmann/phpunit/pull/2516

localheinz:
@sebastianbergmann 

As you can see in [`.php_cs`](https://github.com/sebastianbergmann/phpunit/blob/6052d68cae04df7f4baf5e823e7f214c40d60d1c/.php_cs#L64), the [`TernaryOperatorSpacesFixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Operator/TernaryOperatorSpacesFixer.php) is enabled, and this is the fixer sorting out spaces in ternary statements. 

If you don't want it, you should disable or remove it.

localheinz:
Other than that, maybe someone wants to take a stab at turning it into a _configurable_ fixer?

sebastianbergmann:
What then is the purpose of `ternary_operator_spaces` if it does not format the code as I have it right now? Does it simply ensure that there is a space around `?` and `:`?

sebastianbergmann:
[...]

Issue #1055, opened 2015-03-13
`align_double_arrow` fixer sometimes aligns with extra whitespace


jaredh159:
Before:

``` php
$foo = [
    "bar" => 1,
    "baz"   => 2,
];
```

After:

``` php
$foo = [
    "bar"   => 1,
    "baz"   => 2,
];
```

What I would expect:

``` php
$foo = [
    "bar" => 1,
    "baz" => 2,
];
```

Maybe this wasn't the intention of the fixer, so it's not an issue, but I often rename variables in ways that leave double arrows too far to the right, and I was expecting the fixer to use as little whitespace as possible to align them.  I created a failing test case and tried to fix it, but the code in this fixer is a little over my head.

keradus:
@dericofilho what do you think ?

GrahamCampbell:
I mentioned this issue before. If we have way too many spaces, the fixer won't strip them. It'll just align everything really far over.

dericofilho:
`align_double_arrow` is meant to align to the rightmost `T_DOUBLE_ARROW`. Back then, when we debated about that, I argued that this fixer should be somehow tied to another fixer (that, AFAIK, does not exist in PHP-CS-Fixer) that would do space normalization first.

In the case of php.tools, I do have a Pass (a.k.a Fixer, here) that normalize spaces between tokens before actually aligning `T_DOUBLE_ARROW` or `=`.

So it is a schrodinger bug. It is and it is not wrong.

nochso:
Currently working around this by fixing twice:
First with `unalign_double_arrow` and again with `align_double_arrow`. Thus stripping extra spaces before aligning with the minimum offset.

I've tried using both fixers at once, but that didn't work for me. Is there some other way without running php-cs-fixer twice?

SpacePossum:
if we make the prio of `unalign_double_arrow` < of `again with align_double_arrow` then running the fixer with both fixers might do the trick ;)

GrahamCampbell:
> I've tried using both fixers at once, but that didn't work for me. Is there some other way without running php-cs-fixer twice?

You MUST NOT enable those fixers at the same time. They are conflicting. In 2.0, we're working on a way for php-cs-fixer to actually stop people doing that.

keradus:
@GrahamCampbell is right

keradus:
Closes due to merge of #2573

Issue #2411, opened 2016-12-17
[BUG] BinaryOperatorSpacesFixer does not add whitespace around equal sign


buffcode:
Configuration:
```php
return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setCacheFile(__DIR__ . '/.php_cs.cache')
    ->setRules(array(
        '@Symfony'                => true,
        '@PHP70Migration'         => true,
        'ordered_imports'         => true,
        'phpdoc_order'            => true,
        'phpdoc_summary'          => false,
        'phpdoc_separation'       => false,
        'phpdoc_var_without_name' => false,
        'phpdoc_no_package'       => false,
        'concat_space'            => [
            'spacing' => 'one',
        ],
        'binary_operator_spaces' => [
            'align_double_arrow' => true,
            'align_equals'       => true,
        ],
    ))
    ->setFinder($finder)
;
```

Input:
```php
$templating='a';
$subject   = 'b';
$subject  ='c';
```

Output:
```php
$templating='a';
$subject   = 'b';
$subject   ='c';
```

Expected output (missing whitespace around equal sign):
```php
$templating = 'a';
$subject    = 'b';
$subject    = 'c';
```

Possible reason:
The logic for adding whitespace around the operater in BinaryOperatorSpacesFixer.php is at the end of `fixWhiteSpaceAroundOperator`, but when `align_equals` is enabled, the function returns in line 136 and forwards to `AlignEqualsFixerHelper`. This fixer won't inject additional whitespace around the equal sign.

NewFuture:
meet the same bug

kaiyulee:
also with `align_double_arrow`

keradus:
Closes due to merge of #2573

Issue #2355, opened 2016-12-03
Not possible to only align equals and not change other binary operators


GrahamCampbell:
This is due to the merge into BinaryOperatorSpacesFixer, but no config to turn the other parts of the fixer off.

SpacePossum:
I did that, and my reason was if someone cares about the CS of/around `BinaryOperatorSpaces` we allow for variation/configuration where there are open spots in the PSR's, but thats it. Cutting back repeating and conflicting logic is important for the tool.
If someone really wants messy spaces around binary operators but still want to align equals or arrows the user is out of luck atm. Feel free to open an PR to make more parts configurable, especially welcome before I start merging the `not_operator_with_space` and the `not_operator_with_successor_space` into `BinaryOperatorSpacesFixer` using config.

GrahamCampbell:
I will likely prep a PR then. I think it's important for people to be able to just run the alignment part on it's own and nothing else if they want.

keradus:
Closes due to merge of #2573

PR #2573, opened 2017-03-02
BinaryOperatorSpaces reworked

SpacePossum:
closes
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1055
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2355
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2411
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2811

replaces
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1566
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2555

closes part of:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2559

Fixing examples.

Default:
```diff
--- Original
+++ New
@@ @@
<?php
-    $a= 1  + $b^ $d !==  $e or   $f;
+    $a = 1 + $b ^ $d !== $e or $f;
```

Fixing with configuration: `['operators' => ['=' => 'align', 'xor' => NULL]]`.
```diff
--- Original
+++ New
@@ @@
 <?php
 $aa=  1;
-$b=2;
+$b =2;

$c = $d    xor    $e;
-$f    -=  1;
+$f -= 1;
```

Fixing with configuration: `['operators' => ['+=' => 'align_single_space']]`.
```diff
--- Original
+++ New
@@ @@
 <?php
-$a = $b+=$c;
-$d = $ee+=$f;
+$a = $b  += $c;
+$d = $ee += $f;
```

Fixing with configuration: `['operators' => ['===' => 'align_single_space_minimal']]`.
```diff
--- Original
+++ New
@@ @@
 <?php
-$a = $b===$c;
-$d = $f   ===  $g;
-$h = $i===  $j;
+$a = $b === $c;
+$d = $f === $g;
+$h = $i === $j;
```

```
Fixing examples:
   --- Original
   +++ New
   @@ @@
    <?php
   -    $a= 1  + $b^ $d !==  $e or   $f;
   +    $a = 1 + $b ^ $d !== $e or $f;

 * Example #2. Fixing with configuration: ['operators' => ['=' => 'align', 'xor' => null]].
   --- Original
   +++ New
   @@ @@
    <?php
    $aa=  1;
   -$b=2;
   +$b =2;

    $c = $d    xor    $e;
   -$f    -=  1;
   +$f -= 1;

 * Example #3. Fixing with configuration: ['operators' => ['+=' => 'align_single_space']].
   --- Original
   +++ New
   @@ @@
    <?php
   -$a = $b +=$c;
   -$d = $ee+=$f;
   +$a = $b  += $c;
   +$d = $ee += $f;

   -$g = $b     +=$c;
   -$h = $ee+=$f;
   +$g = $b     += $c;
   +$h = $ee    += $f;

 * Example #4. Fixing with configuration: ['operators' => ['===' => 'align_single_space_minimal']].
   --- Original
   +++ New
   @@ @@
    <?php
   -$a = $b===$c;
   -$d = $f   ===  $g;
   -$h = $i===  $j;
   +$a = $b === $c;
   +$d = $f === $g;
   +$h = $i === $j;
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -231,14 +231,18 @@ Choose from the list of available rules:
 
 * **binary_operator_spaces** [@Symfony]
 
-  Binary operators should be surrounded by at least one space.
+  Binary operators should be surrounded by space as configured.
 
   Configuration options:
 
-  - ``align_double_arrow`` (``false``, ``null``, ``true``): whether to apply, remove or
-    ignore double arrows alignment; defaults to ``false``
-  - ``align_equals`` (``false``, ``null``, ``true``): whether to apply, remove or ignore
-    equals alignment; defaults to ``false``
+  - ``align_double_arrow`` (``false``, ``null``, ``true``): (deprecated) Whether to apply,
+    remove or ignore double arrows alignment; defaults to ``false``
+  - ``align_equals`` (``false``, ``null``, ``true``): (deprecated) Whether to apply, remove
+    or ignore equals alignment; defaults to ``false``
+  - ``default`` (``'align'``, ``'align_single_space'``, ``'align_single_space_minimal'``,
+    ``'single_space'``, ``null``): default fix strategy; defaults to ``'single_space'``
+  - ``operators`` (``array``): dictionary of ``binary operator`` => ``fix strategy``
+    values that differ from the default strategy; defaults to ``[]``
 
 * **blank_line_after_namespace** [@PSR2, @Symfony]
 
--- a/src/AbstractAlignFixerHelper.php
+++ b/src/AbstractAlignFixerHelper.php
@@ -18,6 +18,8 @@ use PhpCsFixer\Tokenizer\Tokens;
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  *
  * @internal
+ *
+ * @deprecated
  */
 abstract class AbstractAlignFixerHelper
 {
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -28,6 +28,7 @@ use PhpCsFixer\RuleSet;
 use PhpCsFixer\StdinFileInfo;
 use PhpCsFixer\Tokenizer\Tokens;
 use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
@@ -160,7 +161,7 @@ final class DescribeCommand extends Command
             $output->writeln(sprintf('Fixer is configurable using following option%s:', 1 === count($options) ? '' : 's'));
 
             foreach ($options as $option) {
-                $line = '* <info>'.$option->getName().'</info>';
+                $line = '* <info>'.OutputFormatter::escape($option->getName()).'</info>';
 
                 $allowed = HelpCommand::getDisplayableAllowedValues($option);
                 if (null !== $allowed) {
@@ -175,7 +176,7 @@ final class DescribeCommand extends Command
                     $line .= ' (<comment>'.implode('</comment>, <comment>', $allowed).'</comment>)';
                 }
 
-                $description = preg_replace('/(`.+?`)/', '<info>$1</info>', $option->getDescription());
+                $description = preg_replace('/(`.+?`)/', '<info>$1</info>', OutputFormatter::escape($option->getDescription()));
                 $line .= ': '.lcfirst(preg_replace('/\.$/', '', $description)).'; ';
                 if ($option->hasDefault()) {
                     $line .= sprintf(
--- a/src/Fixer/Operator/AlignDoubleArrowFixerHelper.php
+++ b/src/Fixer/Operator/AlignDoubleArrowFixerHelper.php
@@ -21,6 +21,8 @@ use PhpCsFixer\Tokenizer\Tokens;
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  * @author Graham Campbell <graham@alt-three.com>
+ *
+ * @deprecated
  */
 final class AlignDoubleArrowFixerHelper extends AbstractAlignFixerHelper
 {
@@ -32,6 +34,17 @@ final class AlignDoubleArrowFixerHelper extends AbstractAlignFixerHelper
      * @var int
      */
     private $currentLevel = 0;
+
+    public function __construct()
+    {
+        @trigger_error(
+            sprintf(
+                'The "%s" class is deprecated. You should stop using it, as it will be removed in 3.0 version.',
+                __CLASS__
+            ),
+            E_USER_DEPRECATED
+        );
+    }
 
     /**
      * {@inheritdoc}
--- a/src/Fixer/Operator/AlignEqualsFixerHelper.php
+++ b/src/Fixer/Operator/AlignEqualsFixerHelper.php
@@ -20,9 +20,22 @@ use PhpCsFixer\Tokenizer\Tokens;
 /**
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  * @author Graham Campbell <graham@alt-three.com>
+ *
+ * @deprecated
  */
 final class AlignEqualsFixerHelper extends AbstractAlignFixerHelper
 {
+    public function __construct()
+    {
+        @trigger_error(
+            sprintf(
+                'The "%s" class is deprecated. You should stop using it, as it will be removed in 3.0 version.',
+                __CLASS__
+            ),
+            E_USER_DEPRECATED
+        );
+    }
+
     /**
      * {@inheritdoc}
      */
--- a/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
+++ b/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
@@ -14,6 +14,7 @@ namespace PhpCsFixer\Fixer\Operator;
 
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Console\Command\HelpCommand;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
@@ -23,6 +24,7 @@ use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use PhpCsFixer\Tokenizer\TokensAnalyzer;
+use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
@@ -140,13 +142,25 @@ final class BinaryOperatorSpacesFixer extends AbstractFixer implements Configura
     private $alignOperatorTokens = [];
 
     /**
+     * @var array<string, string>
+     */
+    private $operators = [];
+
+    /**
      * {@inheritdoc}
      */
     public function configure(array $configuration = null)
     {
+        if (
+            null !== $configuration &&
+            (array_key_exists('align_equals', $configuration) || array_key_exists('align_double_arrow', $configuration))
+        ) {
+            $configuration = $this->resolveOldConfig($configuration);
+        }
+
         parent::configure($configuration);
 
-        $this->resolveConfig(null === $configuration ? [] : $configuration);
+        $this->operators = $this->resolveOperatorsFromConfig();
     }
 
     /**
@@ -254,6 +268,32 @@ $h = $i===  $j;
                 ->getOption(),
             (new FixerOptionBuilder('operators', 'Dictionary of `binary operator` => `fix strategy` values that differ from the default strategy.'))
                 ->setAllowedTypes(['array'])
+                ->setAllowedValues([function ($option) {
+                    foreach ($option as $operator => $value) {
+                        if (!in_array($operator, self::$supportedOperators, true)) {
+                            throw new InvalidOptionsException(
+                                sprintf(
+                                    'Unexpected "operators" key, expected any of "%s", got "%s".',
+                                    implode('", "', self::$supportedOperators),
+                                    is_object($operator) ? get_class($operator) : gettype($operator).'#'.$operator
+                                )
+                            );
+                        }
+
+                        if (!in_array($value, self::$allowedValues, true)) {
+                            throw new InvalidOptionsException(
+                                sprintf(
+                                    'Unexpected value for operator "%s", expected any of "%s", got "%s".',
+                                    $operator,
+                                    implode('", "', self::$allowedValues),
+                                    is_object($value) ? get_class($value) : (null === $value ? 'null' : gettype($value).'#'.$value)
+                                )
+                            );
+                        }
+                    }
+
+                    return true;
+                }])
                 ->setDefault([])
                 ->getOption(),
             // add deprecated options as BC layer
@@ -276,26 +316,26 @@ $h = $i===  $j;
     {
         $tokenContent = strtolower($tokens[$index]->getContent());
 
-        if (!array_key_exists($tokenContent, $this->configuration)) {
+        if (!array_key_exists($tokenContent, $this->operators)) {
             return; // not configured to be changed
         }
 
-        if (self::SINGLE_SPACE === $this->configuration[$tokenContent]) {
+        if (self::SINGLE_SPACE === $this->operators[$tokenContent]) {
             $this->fixWhiteSpaceAroundOperatorToSingleSpace($tokens, $index);
 
             return;
         }
 
         // schedule for alignment
-        $this->alignOperatorTokens[$tokenContent] = $this->configuration[$tokenContent];
+        $this->alignOperatorTokens[$tokenContent] = $this->operators[$tokenContent];
 
-        if (self::ALIGN === $this->configuration[$tokenContent]) {
+        if (self::ALIGN === $this->operators[$tokenContent]) {
             return;
         }
 
         // fix white space after operator
         if ($tokens[$index + 1]->isWhitespace()) {
-            if (self::ALIGN_SINGLE_SPACE_MINIMAL === $this->configuration[$tokenContent]) {
+            if (self::ALIGN_SINGLE_SPACE_MINIMAL === $this->operators[$tokenContent]) {
                 $tokens[$index + 1] = new Token([T_WHITESPACE, ' ']);
             }
 
@@ -354,121 +394,90 @@ $h = $i===  $j;
         return false;
     }
 
-    private function resolveConfig(array $configuration)
+    /**
+     * @return array<string, string>
+     */
+    private function resolveOperatorsFromConfig()
     {
-        $this->configuration = [];
+        $operators = [];
 
-        if (0 === count($configuration)) {
-            $this->resolveTokenDefaults(self::SINGLE_SPACE);
-
-            return;
-        }
-
-        $this->resolveTokenDefaults(array_key_exists('default', $configuration) ? $configuration['default'] : self::SINGLE_SPACE);
-
-        if (array_key_exists('align_equals', $configuration) || array_key_exists('align_double_arrow', $configuration)) {
-            $this->resolveOldConfig($configuration);
-
-            return;
-        }
-
-        if (!array_key_exists('operators', $configuration)) {
-            return;
-        }
-
-        foreach ($configuration['operators'] as $operator => $value) {
-            if (!in_array($operator, self::$supportedOperators, true)) {
-                throw new InvalidFixerConfigurationException(
-                    $this->getName(),
-                    sprintf(
-                        'Unexpected "operators" key, expected any of "%s", got "%s".',
-                        implode('", "', self::$supportedOperators),
-                        is_object($operator) ? get_class($operator) : gettype($operator).'#'.$operator
-                    )
-                );
+        if (null !== $this->configuration['default']) {
+            foreach (self::$supportedOperators as $operator) {
+                $operators[$operator] = $this->configuration['default'];
             }
+        }
 
-            if (!in_array($value, self::$allowedValues, true)) {
-                throw new InvalidFixerConfigurationException(
-                    $this->getName(),
-                    sprintf(
-                        'Unexpected value for operator "%s", expected any of "%s", got "%s".',
-                        $operator,
-                        implode('", "', self::$allowedValues),
-                        is_object($value) ? get_class($value) : (null === $value ? 'null' : gettype($value).'#'.$value)
-                    )
-                );
-            }
-
+        foreach ($this->configuration['operators'] as $operator => $value) {
             if (null === $value) {
-                unset($this->configuration[$operator]);
+                unset($operators[$operator]);
             } else {
-                $this->configuration[$operator] = $value;
+                $operators[$operator] = $value;
             }
         }
 
         if (!defined('T_SPACESHIP')) {
-            unset($this->configuration['<=>']);
+            unset($operators['<=>']);
         }
 
         if (!defined('T_COALESCE')) {
-            unset($this->configuration['??']);
+            unset($operators['??']);
         }
+
+        return $operators;
     }
 
+    /**
+     * @param array $configuration
+     *
+     * @return array
+     */
     private function resolveOldConfig(array $configuration)
     {
-        $message = 'Given configuration is deprecated and will be removed in 3.0.';
+        $newConfig = [
+            'operators' => [],
+        ];
 
         foreach ($configuration as $name => $setting) {
-            if ('align_equals' === $name) {
+            if ('align_double_arrow' === $name) {
                 if (true === $configuration[$name]) {
-                    $this->configuration['='] = self::ALIGN;
-                    $message .= sprintf(" Use configuration: ['operators' => ['=' => '%s']] as replacement for ['%s' => true].", self::ALIGN, $name);
+                    $newConfig['operators']['=>'] = self::ALIGN;
                 } elseif (false === $configuration[$name]) {
-                    $this->configuration['='] = self::SINGLE_SPACE;
-                    $message .= sprintf(" Use configuration: ['operators' => ['=' => '%s']] as replacement for ['%s' => false].", self::SINGLE_SPACE, $name);
-                } else { // === null
-                    unset($this->configuration['=']);
+                    $newConfig['operators']['=>'] = self::SINGLE_SPACE;
+                } elseif (null !== $configuration[$name]) {
+                    throw new InvalidFixerConfigurationException(
+                        $this->getName(),
+                        sprintf(
+                            'Invalid configuration: The option "align_double_arrow" with value %s is invalid. Accepted values are: true, false, null.',
+                            $configuration[$name]
+                        )
+                    );
                 }
-            } elseif ('align_double_arrow' === $name) {
+            } elseif ('align_equals' === $name) {
                 if (true === $configuration[$name]) {
-                    $this->configuration['=>'] = self::ALIGN;
-                    $message .= sprintf(" Use configuration: ['operators' => ['=>' => '%s']] as replacement for ['%s' => true].", self::ALIGN, $name);
+                    $newConfig['operators']['='] = self::ALIGN;
                 } elseif (false === $configuration[$name]) {
-                    $this->configuration['=>'] = self::SINGLE_SPACE;
-                    $message .= sprintf(" Use configuration: ['operators' => ['=>' => '%s']] as replacement for ['%s' => false].", self::SINGLE_SPACE, $name);
-                } else { // === null
-                    unset($this->configuration['=>']);
+                    $newConfig['operators']['='] = self::SINGLE_SPACE;
+                } elseif (null !== $configuration[$name]) {
+                    throw new InvalidFixerConfigurationException(
+                        $this->getName(),
+                        sprintf(
+                            'Invalid configuration: The option "align_equals" with value %s is invalid. Accepted values are: true, false, null.',
+                            $configuration[$name]
+                        )
+                    );
                 }
             } else {
                 throw new InvalidFixerConfigurationException($this->getName(), 'Mixing old configuration with new configuration is not allowed.');
             }
         }
 
-        @trigger_error($message, E_USER_DEPRECATED);
-    }
+        @trigger_error(sprintf(
+            'Given configuration is deprecated and will be removed in 3.0. Use configuration %s as replacement for %s.',
+            HelpCommand::toString($newConfig),
+            HelpCommand::toString($configuration)
+        ), E_USER_DEPRECATED);
 
-    /**
-     * @param null|string $default
-     */
-    private function resolveTokenDefaults($default = null)
-    {
-        if (null === $default) {
-            return;
-        }
-
-        foreach (self::$supportedOperators as $operator) {
-            $this->configuration[$operator] = $default;
-        }
-
-        if (!defined('T_SPACESHIP')) {
-            unset($this->configuration['<=>']);
-        }
-
-        if (!defined('T_COALESCE')) {
-            unset($this->configuration['??']);
-        }
+        return $newConfig;
     }
 
     // Alignment logic related methods
--- a/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
+++ b/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
@@ -12,13 +12,14 @@
 
 namespace PhpCsFixer\Fixer\Operator;
 
-use PhpCsFixer\AbstractAlignFixerHelper;
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use PhpCsFixer\Tokenizer\TokensAnalyzer;
@@ -30,9 +31,123 @@ use PhpCsFixer\Tokenizer\TokensAnalyzer;
 final class BinaryOperatorSpacesFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
-     * @var AbstractAlignFixerHelper[]
+     * @internal
      */
-    private $alignFixerHelpers = [];
+    const SINGLE_SPACE = 'single_space';
+
+    /**
+     * @internal
+     */
+    const ALIGN = 'align';
+
+    /**
+     * @internal
+     */
+    const ALIGN_SINGLE_SPACE = 'align_single_space';
+
+    /**
+     * @internal
+     */
+    const ALIGN_SINGLE_SPACE_MINIMAL = 'align_single_space_minimal';
+
+    /**
+     * @internal
+     * @const Placeholder used as anchor for right alignment.
+     */
+    const ALIGN_PLACEHOLDER = "\x2 ALIGNABLE%d \x3";
+
+    /**
+     * Keep track of the deepest level ever achieved while
+     * parsing the code. Used later to replace alignment
+     * placeholders with spaces.
+     *
+     * @var int
+     */
+    private $deepestLevel;
+
+    /**
+     * Level counter of the current nest level.
+     * So one level alignments are not mixed with
+     * other level ones.
+     *
+     * @var int
+     */
+    private $currentLevel;
+
+    private static $allowedValues = [
+        self::ALIGN,
+        self::ALIGN_SINGLE_SPACE,
+        self::ALIGN_SINGLE_SPACE_MINIMAL,
+        self::SINGLE_SPACE,
+        null,
+    ];
+
+    /**
+     * @var string[]
+     */
+    private static $supportedOperators = [
+        '=',
+        '*',
+        '/',
+        '%',
+        '<',
+        '>',
+        '|',
+        '^',
+        '+',
+        '-',
+        '&',
+        '&=',
+        '&&',
+        '||',
+        '.=',
+        '/=',
+        '=>',
+        '==',
+        '>=',
+        '===',
+        '!=',
+        '<>',
+        '!==',
+        '<=',
+        'and',
+        'or',
+        'xor',
+        '-=',
+        '%=',
+        '*=',
+        '|=',
+        '+=',
+        '<<',
+        '<<=',
+        '>>',
+        '>>=',
+        '^=',
+        '**',
+        '**=',
+        '<=>',
+        '??',
+    ];
+
+    /**
+     * @var TokensAnalyzer
+     */
+    private $tokensAnalyzer;
+
+    /**
+     * @var array<string, string>
+     */
+    private $alignOperatorTokens = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->resolveConfig(null === $configuration ? [] : $configuration);
+    }
 
     /**
      * {@inheritdoc}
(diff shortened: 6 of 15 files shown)


src/AbstractAlignFixerHelper.php
```diff
@@ -35,6 +50,22 @@
...
+     * @var string
+     */
+    private $alignmentStrategy = self::ALIGN;
+
+    /**
+     * @param string $set
+     */
+    public function setAlignmentStrategy($set)
+    {
+        if (self::ALIGN !== $set && self::ALIGN_SINGLE_SPACE !== $set && self::ALIGN_SINGLE_SPACE_MINIMAL !== $set) {
```
SpacePossum: throw exception, rename `$set`

src/AbstractAlignFixerHelper.php
```diff
@@ -70,7 +101,6 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
     protected function replacePlaceholder(Tokens $tokens)
     {
         $tmpCode = $tokens->generateCode();
-
```
SpacePossum: revert

src/Fixer/Operator/AlignEqualsFixerHelper.php
```diff
@@ -22,6 +22,13 @@
  */
 final class AlignEqualsFixerHelper extends AbstractAlignFixerHelper
 {
+    private $content;
+
+    public function setTokenContent($content)
```
SpacePossum: rename method

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -81,30 +156,36 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
+            $this->runAlignFixer($file, $tokens);
+        }
     }
 
     /**
      * {@inheritdoc}
      */
     public function getDefinition()
     {
+        // TODO
```
SpacePossum: fix

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
```diff
@@ -343,7 +456,7 @@ function b(
         );
     }
 
-    public function testWrongConfigItem()
+    public function fixme_testWrongConfigItem()
```
SpacePossum: todo

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
```diff
@@ -353,7 +466,7 @@ public function testWrongConfigItem()
         $this->fixer->configure(array('foo' => true));
     }
 
-    public function testWrongConfigValue()
+    public function fixme_testWrongConfigValue()
```
SpacePossum: todo

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -106,66 +186,64 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
 ',
-                    array('align_double_arrow' => true)
+                    array(
+                        '===' => 'align_single_space_minimal',
+                    )
                 ),
             ),
             null,
-            'Aligns or unaligns `=` in consecutive assignments, or `=>` in array initializations',
+            'Align (but with (minimal) one space), un aligns `=` in consecutive assignments, or `=>` in array initializations',
```
SpacePossum: needs better copy

crhallberg:
I would expect the correct changes for 'align_single_space_minimal' to look more like this, leaving multiple spaces intact, meaning `$d` technically goes unchanged.

Fixing with configuration: ['===' => 'align_single_space_minimal'].
```diff
--- Original
+++ New
@@ @@
 <?php
-$a = $b===$c;
-$d = $f   ===  $g;
-$h = $i===  $j;
+$a = $b === $c;
+$d = $f   ===  $g;
+$h = $i ===  $j;
```

SpacePossum:
Your example result is not aligned for `===`, but if I understand you correctly you want to be able to configure minimal single space around an operator, like;
```diff
--- Original
+++ New
@@ @@
<?php
-$a = $b+=           $c;
+$a = $b +=           $c;
-$d = $e   -=$f;
+$d = $e   -= $f;
$g = $h        +    $i; // not touched
```
That is missing currently, I can take a look, shouldn't be hard to add I think.

For aligning you can use `align_single_space` for cases like;
```diff
--- Original
+++ New
@@ @@
<?php
-$a = $b    =>           $c;
+$a = $b    => $c;
-$d = $e    =>$f;
+$d = $e    => $f;
$g = $h    => $i;

$z=$x=>$y;

```

or use `align` if you want no white space to be added if it is not needed to align,
or `align_single_space_minimal`; ensures single space around the operator and use minimal white space to align items of a group

crhallberg:
That's correct. My mistake. You're right that `===` wouldn't be affected by `align_single_space_minimal`.

keradus:
what is the state here @SpacePossum  ?

SpacePossum:
on my list of PR's to finish :|

keradus:
I have changed base branch to master

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
```diff
@@ -22,20 +22,105 @@
  *
  * @internal
  *
- * @covers \PhpCsFixer\AbstractAlignFixerHelper
- * @covers \PhpCsFixer\Fixer\Operator\AlignDoubleArrowFixerHelper
- * @covers \PhpCsFixer\Fixer\Operator\AlignEqualsFixerHelper
```
SpacePossum: these 3 classes can be deprecated and removed on 3.0

README.rst
```diff
@@ -230,14 +230,92 @@ Choose from the list of available rules:
...
-  Binary operators should be surrounded by at least one space.
+  Binary operators should be surrounded by space as configured.
 
   Configuration options:
 
-  - ``align_double_arrow`` (``false``, ``null``, ``true``): whether to apply, remove or
-    ignore double arrows alignment; defaults to ``false``
-  - ``align_equals`` (``false``, ``null``, ``true``): whether to apply, remove or ignore
-    equals alignment; defaults to ``false``
+  - ``!=`` (``'align'``, ``'align_single_space'``, ``'align_single_space_minimal'``,
```
SpacePossum: well this has pretty much wrecked the readme :|

julienfalque: What about grouping all of these inside a single option that accepts an array with operators as keys and spacing to apply as value?

SpacePossum:
readme test will fail because of https://github.com/symfony/symfony/issues/18481

SpacePossum:
readme failures are fixed over the diff. SF versions,
not sure we want those fixes in diff. PR targeting 2.2, lemme know

SpacePossum:
Ready for review, would be nice to work towards closing.
Only thing to discuss is deprecation of no longer used helper classes (or deleting those all together).

Aerendir:
Maybe adding to the README some of the examples in this  DISCUSSION ABOUT THE DIFFERENT BEHAVIORS OF "align", "align_single_space", ecc. is useful to better understand the behaviors without having to try the rule. Currently I'm not at computer, but I will test the new behavior once I will. Thank you for your work @SpacePossum!

keradus:
sadly, there is no place in readme for providing examples, but we do it when you execute `php-cs-fixer describe rule_name`

Aerendir:
I've tested the PR with this configuration and it seems to work well for me (at least for my use cases).

The configuration I've used is this:

```php
'binary_operator_spaces' => [
    '=' => 'align_single_space_minimal',
    '=>' => 'align_single_space_minimal'
],
```

The fixes are those:

```php
- $store   = $this->getDoctrine()->getRepository('AppBundle:Store')->findAll()[0];
+ $store = $this->getDoctrine()->getRepository('AppBundle:Store')->findAll()[0];
```

```php
- public function thankyouAction(Feedback $feedback =  null, Request $request): Response
// Remove two spaces after $feedback
+ public function thankyouAction(Feedback $feedback = null, Request $request): Response
```

```php
- $subscribedFeatures =  $this->getContainer()->get(...
// Remove two spaces after =
+ $subscribedFeatures = $this->getContainer()->get(...
```

```php
return [
-            StoreCreateEvent::CREATE              => 'onStoreCreate',
-            StoreCreateFromEmailEvent::CREATE     => 'onStoreCreateFromEmail',
-            StoreReclaimWithEmailEvent::RECLAIM   => 'onStoreReclaimWithEmail',
-            StorePlanUpdateEvent::UPDATE          => 'onStorePlanUpdate',
-            'workflow.store.entered.suitable'     => 'onStoreEnteredSuitable',
        ];

return [
+            StoreCreateEvent::CREATE            => 'onStoreCreate',
+            StoreCreateFromEmailEvent::CREATE   => 'onStoreCreateFromEmail',
+            StoreReclaimWithEmailEvent::RECLAIM => 'onStoreReclaimWithEmail',
+            StorePlanUpdateEvent::UPDATE        => 'onStorePlanUpdate',
+            'workflow.store.entered.suitable'   => 'onStoreEnteredSuitable',
        ];
```

```php
-    $this->entityManager        = $entityManager;
-    $this->token                = $token->getToken();

+ $this->entityManager = $entityManager;
+ $this->token         = $token->getToken();
```

This is wrong, instead
```php
- $this->newName
-            = $this->path
-            = $this->randomName
-            = $this->remoteFile
-            = $this->tmpContent
-            = null;

this->newName
+ = $this->path
+ = $this->randomName
+ = $this->remoteFile
+ = $this->tmpContent
+ = null;
```

The last example is the indentation of a fluent interface and it is aligned wrongly in my opinion.

What do you think?

README.rst
julienfalque: ping @SpacePossum

src/AlignmentFixerHelper.php
```diff
@@ -0,0 +1,341 @@
...
+        for ($j = 0; $j <= $this->deepestLevel; ++$j) {
+            $placeholder = sprintf(self::ALIGNABLE_PLACEHOLDER, $j);
+
+            if (false === strpos($tmpCode, $placeholder)) {
+                continue;
+            }
+
+            $lines = explode("\n", $tmpCode);
+            $linesWithPlaceholder = [];
+            $blockSize = 0;
```
julienfalque: `$blockSize` is a very misleading name IMO. What about `$groupIndex`?

src/AlignmentFixerHelper.php
```diff
@@ -0,0 +1,341 @@
...
+    private $currentLevel = 0;
+
+    /**
+     * @param Tokens                $tokens
+     * @param array<string, string> $toAlign
+     */
+    public function fixAlignment(Tokens $tokens, array $toAlign)
+    {
+        foreach ($toAlign as $tokenContent => $alignStrategy) {
+            if (self::ALIGN !== $alignStrategy && self::ALIGN_SINGLE_SPACE !== $alignStrategy && self::ALIGN_SINGLE_SPACE_MINIMAL !== $alignStrategy) {
```
SpacePossum: For a simple check against two values I see no need for wrapping into `in_array` by the cost of both creating an array and a function call as both are not needed and less explicit.

src/AlignmentFixerHelper.php
```diff
@@ -0,0 +1,341 @@
...
+
+            if ('=>' === $tokenContent) {
+                $this->injectAlignmentPlaceholdersForArrow($tokensClone, 0, count($tokens));
+            } else {
+                $this->injectAlignmentPlaceholders($tokensClone, 0, count($tokens), $tokenContent);
+            }
+
+            // for all tokens that should be aligned but do not have anything to align with, fix spacing if needed
+            if (self::ALIGN_SINGLE_SPACE === $alignStrategy || self::ALIGN_SINGLE_SPACE_MINIMAL === $alignStrategy) {
+                if ('=>' === $tokenContent) {
```
SpacePossum: not rely, one case checks the content (after the type is already checked), the other only the type (as casing for any `=>` operator is of no concern).

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -30,65 +31,149 @@
...
+
+        foreach ($configuration as $name => $value) {
+            if ('align_equals' === $name || 'align_double_arrow' === $name) {
+                $this->configuration = $this->resolveOldConfig($configuration);
+
+                break;
+            }
+
+            if ('<=>' === $name && !defined('T_SPACESHIP')) {
+                unset($this->configuration[$name]);
```
SpacePossum: That would be less convenient for project supporting multiple PHP versions as these cannot have one CS configuration. We do this at other places as well (`ReturnTypeDeclarationFixer` can be configured to on PHP < 7) and at other place we dont (`VisibilityRequiredFixer` `const` config for example). Since we do not do this the same everywhere I wonder what the best direction would be.

src/AlignmentFixerHelper.php
```diff
@@ -0,0 +1,341 @@
...
+
+                continue;
+            }
+
+            if ($token->isGivenKind(T_ARRAY)) { // don't use "$tokens->isArray()" here, short arrays are handled in the next case
+                $from = $tokens->getNextMeaningfulToken($index);
+                $until = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $from);
+                $index = $until;
+
+                $this->injectArrayAlignmentPlaceholders($tokens, $from, $until);
```
SpacePossum: I only moved this logic, I'll check what the deal is here.

src/AlignmentFixerHelper.php
```diff
@@ -0,0 +1,341 @@
...
+    }
+
+    private function injectAlignmentPlaceholdersForArrow(Tokens $tokens, $startAt, $endAt)
+    {
+        for ($index = $startAt; $index < $endAt; ++$index) {
+            $token = $tokens[$index];
+
+            if ($token->isGivenKind([T_FOREACH, T_FOR, T_WHILE, T_IF, T_SWITCH])) {
+                $index = $tokens->getNextMeaningfulToken($index);
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $index);
```
SpacePossum: would be an existing bug I think, I'll have a look :+1:

SpacePossum:
@Aerendir thanks for testing and reporting back, I'll have a look at your last sample :+1:

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
julienfalque: I would be in favor of always throwing exceptions but I'm fine with both solutions. Though we should probably open a new PR to standardize this.

src/AlignmentFixerHelper.php
julienfalque: IMO the two `foreach` can be merged to simplify code and improve readability. The second one already performs `strtolower()` calls for operators that don't need it anyway.

SpacePossum: I give it some more thoughts, I'll fix the other issues first and see in the end if it still makes sense. I just dislike doing tests within a loop if it is already known what the outcome will be.

SpacePossum:
expecting test to fail

Aerendir:
Any news about this?

SpacePossum:
One case is still failing and needs to be resolved, after that the usually rounds of review will happen again I guess.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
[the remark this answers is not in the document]
SpacePossum: I think it is better this way, we can set the `space ship op` default configuration without users having to deal with disabling the option in their > PHP X projects.

SpacePossum:
@Aerendir I've made a fix (see last two commits for it + some tests) for the case you've found.

New round of reviews is welcome from everyone :)

Aerendir:
@SpacePossum , just tried: it works. The bug is fixed.

The failing fixing is not failing anymore: it leave the code untouched, without removing the formatting required tabulation spaces.

src/Console/Command/DescribeCommand.php
```diff
@@ -160,7 +161,7 @@ private function describeRule(OutputInterface $output, $name)
             $output->writeln(sprintf('Fixer is configurable using following option%s:', 1 === count($options) ? '' : 's'));
 
             foreach ($options as $option) {
-                $line = '* <info>'.$option->getName().'</info>';
+                $line = '* <info>'.OutputFormatter::escape($option->getName()).'</info>';
```
keradus: is it needed?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -30,65 +31,161 @@
...
+     * @internal
      */
-    private $alignFixerHelpers = [];
+    const SINGLE_SPACE = 'single_space';
+
+    /**
+     * @var array
+     */
+    private static $defaultConfiguration = [
+        '=' => self::SINGLE_SPACE,
```
keradus: put them all in nested key like `operators`, so one could be able to add next option in future,

Also, I would add a `default` (to not have a need to override all operators one by one)
Then,
```php
'default' => self::SINGLE_SPACE,
'operators' => [
    '=' => null,
    '*' => null,
    // ...
],
```

and then, when I want to change config, I update default or particular operators. when I want to resolve, I take first operator, then default if operator is not configured.

src/Console/Command/DescribeCommand.php
SpacePossum: Yes, I believe so. Even if SF gets patched we would've to black list the bugged versions otherwise. Beside the fixes is most likely going to be changing the escape sequence from `<<<` to some non printable char, however if we use that char we would end up with the same problem.
Escaping here will resolve all those issues.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: Interesting idea.
However I'll wait with updating us the current configuration setup has been used since the start of the PR so maybe people like it better, lets hear from others as well :)

keradus: they will be able to use "current" way via just adding `operators` envelope and don't give a shit about `default` key.
yet, with "current" way it's not possible to further extend the configuration. Do you remember we had rootless configs that we deprecated and replaced with enveloped ones? That's exactly this scenario!

src/Console/Command/DescribeCommand.php
keradus: why non-printable char would be an issue for us...
ok, let it stay like it is, but I still extra confused about it...
maybe we could at least add some test that would fail without escaping on broken symfony/console version ?

SpacePossum: I don't understand, the tests did fail with SF console 3.0.4, that is why I changed it so the tests are green again

keradus: oh, OK. I was not able to reproduce test failure localy

SpacePossum: about "\0" issue might, be resolved here; https://github.com/symfony/symfony/pull/23730#discussion_r135804540

anyway, escaping the output still seems like a good thing to do to prevent other regressions and issues (and we don't have to exclude known SF console releases with the issue)

SpacePossum: To reproduce locally:

```
[12:55 PM]-[possum@zoo-playground-1]-[~/lala] 
$ cat test.php 
<?php
require_once __DIR__.'/vendor/autoload.php';
$output = new Symfony\Component\Console\Output\ConsoleOutput();
$output->writeln('[<<]');
die;
[12:55 PM]-[possum@zoo-playground-1]-[~/lala] 
$ composer require symfony/console 3.0.0
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating symfony/console (v3.0.4 => v3.0.0): Loading from cache
Writing lock file
Generating autoload files

[12:55 PM]-[possum@zoo-playground-1]-[~/lala] 
$ php test.php 
[<<]

[12:55 PM]-[possum@zoo-playground-1]-[~/lala] 
$ composer require symfony/console 3.0.4
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 0 installs, 1 update, 0 removals
  - Updating symfony/console (v3.0.0 => v3.0.4): Loading from cache
Writing lock file
Generating autoload files

[12:56 PM]-[possum@zoo-playground-1]-[~/lala] 
$ composer info -D|grep symfony/console
1:symfony/console v3.0.4 Symfony Console Component

[12:56 PM]-[possum@zoo-playground-1]-[~/lala] 
$ php test.php 
[\]
```

keradus: btw, symfony/symfony#23730 just got merged

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -159,10 +172,7 @@ public function getDefinition()
 $c = $d    xor    $e;
 $f    -=  1;
 ',
-                    [
-                        '=' => 'align',
-                        'xor' => null,
-                    ]
+                    ['operators' => ['=' => 'align', 'xor' => null]]
```
keradus: no constant ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -105,35 +106,47 @@ public function configure(array $configuration = null)
...
+        if (isset($configuration['default'])) {
+            $this->resolveTokenDefaults($configuration['default']);
+        } else {
+            $this->resolveTokenDefaults(self::SINGLE_SPACE);
+        }
 
-                break;
-            }
+        if (isset($configuration['align_equals']) || isset($configuration['align_double_arrow'])) {
+            $this->resolveOldConfig($configuration);
```
keradus: is there a deprecation triggered when one uses old config ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -242,40 +248,36 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
-        $options[] =
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('default', 'default fix strategy.'))
+                ->setDefault(self::SINGLE_SPACE)
+                ->setAllowedValues($allowedValues)
+                ->getOption(),
+            (new FixerOptionBuilder('operators', 'List of binary operator to fix.'))
+                ->setAllowedTypes(['array'])
+//                ->setAllowedValues([
+//                    (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf($a), FIXME
```
keradus: fixme ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: yes, 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573/files#diff-fdd74d10fe54a085bcd3669c08cb37d9R399
tested here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573/files#diff-09e3cb60a7a2729fabc2c9950e007302R665

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: yeah, ping @julienfalque how to setup allowed values for associative array? couldn't find where we use something like this ATM

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
keradus: :+1:
sorry, github linked me in notification only to latest commit, in which i couldn't find it.
thanks

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
keradus: the format will be like:
```
operators => {
    '=' => string|null,
    '*' => string|null,
    ...
}
```

right ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: you mean for `operators` or for all options, is it worth it?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: indeed, with the `string` value being one of 4 values

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
keradus: sorry, for `align`

src/AbstractAlignFixerHelper.php
```diff
@@ -18,6 +18,8 @@
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  *
  * @internal
+ *
+ * @deprecated
  */
 abstract class AbstractAlignFixerHelper
```
SpacePossum: should we trigger a deprecation error for this class as well or is this enough together with the errors triggered by the two children?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
julienfalque: At the moment there is no validator for options that are associative arrays. A solution could be to add a normalizer to validate it manually, maybe using a custom `OptionsResolver` in it?

SpacePossum: Thanks I give it a try! : )
TBH I do 't like `normalizers` to do validation as to have a clear separation of concerns. As such I think I'll do validation with the `configure` method.

SpacePossum: all updated

SpacePossum:
TODO: test on mulitple repo's to see if all good

SpacePossum:
0 hits on SF 2.7

SpacePossum:
On PHPUnit `master` `./src` with;
```
        'binary_operator_spaces' => [
            'operators' => [
                '+' => 'align_single_space',
                '=' => 'align_single_space',
                '=>' => 'align_single_space',
            ],
        ]
```

one hit:

```
            if (false === \getenv($name)) {
                \putenv("{$name} ={$value}");
            }
```

to

```
            if (false === \getenv($name)) {
                \putenv("{$name} = {$value}");
            }
```

src/Fixer/Operator/AlignmentFixerHelper.php
```diff
@@ -107,6 +113,28 @@ public function fixAlignment(Tokens $tokens, array $toAlign)
     }
 
     /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return false|int
+     */
+    private function isEqualPartOfDeclareStatement(Tokens $tokens, $index)
```
SpacePossum: duplicate of the one in the fixer, not sure about the composition we want though...

keradus:
```diff
-\putenv("{$name} ={$value}");
+\putenv("{$name} = {$value}");
```

wait wait wait, change inside a string ?

keradus:
ping @SpacePossum

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -135,11 +279,20 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     protected function createConfigurationDefinition()
     {
         return new FixerConfigurationResolver([
-            (new FixerOptionBuilder('align_double_arrow', 'Whether to apply, remove or ignore double arrows alignment.'))
+            (new FixerOptionBuilder('default', 'default fix strategy.'))
+                ->setDefault(self::SINGLE_SPACE)
+                ->setAllowedValues(self::$allowedValues)
+                ->getOption(),
+            (new FixerOptionBuilder('operators', 'List of binary operator to fix.'))
```
keradus: description is wrong.
we need to say that this is assoc key with key being ... and value being ...

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -97,9 +229,18 @@ public function getDefinition()
...
+        return -1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function isCandidate(Tokens $tokens)
     {
-        return true;
+        return true; // count($tokens) > 4; php open, 2 args + operator but not worth the check
```
keradus: yep, not worth, drop the comment

README.rst
```diff
@@ -1291,8 +1294,8 @@ Choose from the list of available rules:
     ``true``
   - ``identical`` (``bool``, ``null``): change identical (``===``, ``!==``) statements;
     defaults to ``true``
-  - ``less_and_greater`` (``bool``, ``null``): change less and greater than (``<``, ``<=``,
-    ``>``, ``>=``) statements; defaults to ``null``
+  - ``less_and_greater`` (``bool``, ``null``): change less and greater than (``<``,
+    ``<=``, ``>``, ``>=``) statements; defaults to ``null``
```
SpacePossum: WTH?

README.rst
```diff
@@ -231,14 +231,18 @@ Choose from the list of available rules:
...
-    ignore double arrows alignment; defaults to ``false``
-  - ``align_equals`` (``false``, ``null``, ``true``): whether to apply, remove or ignore
-    equals alignment; defaults to ``false``
+  - ``align_double_arrow`` (``false``, ``null``, ``true``): (deprecated) Whether to apply,
+    remove or ignore double arrows alignment; defaults to ``false``
+  - ``align_equals`` (``false``, ``null``, ``true``): (deprecated) Whether to apply, remove
+    or ignore equals alignment; defaults to ``false``
+  - ``default`` (``'align'``, ``'align_single_space'``, ``'align_single_space_minimal'``,
+    ``'single_space'``, ``null``): default fix strategy; defaults to ``'single_space'``
+  - ``operators`` (``array``): list of binary ``operator`` ``fix strategy`` pairs;
```
keradus: `that are exception for default strategy` perhaps ?

README.rst
```diff
@@ -231,14 +231,18 @@ Choose from the list of available rules:
...
+    defaults to ``[]``
```
keradus: i got confused, example would be:
```php
[
    ['+', 'align'], // a pair inside list !
]
```

or

```php
[
    '+' => 'align',
]
(a dictionary)
```


?

SpacePossum: not sure what you want me to put in as copy, it works as requested a few comments back.

keradus: exactly how you updated it :+1: @SpacePossum

README.rst
```diff
@@ -1291,8 +1295,8 @@ Choose from the list of available rules:
     ``true``
   - ``identical`` (``bool``, ``null``): change identical (``===``, ``!==``) statements;
     defaults to ``true``
-  - ``less_and_greater`` (``bool``, ``null``): change less and greater than (``<``, ``<=``,
-    ``>``, ``>=``) statements; defaults to ``null``
+  - ``less_and_greater`` (``bool``, ``null``): change less and greater than (``<``,
+    ``<=``, ``>``, ``>=``) statements; defaults to ``null``
```
SpacePossum: This is the result of escaping before doing the word wrap. So after the escaping the line it gets longer, than it gets word wrapped and then unescaped later on.
It took me way to long to figure out why this happened and wonder if it worth fixing it.

src/Fixer/Operator/AlignDoubleArrowFixerHelper.php
```diff
@@ -21,6 +21,10 @@
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  * @author Graham Campbell <graham@alt-three.com>
+ *
+ * @internal
```
keradus: adding `@internal` tag is bc breaker.
while I agree it would be a good move, we shall not. yet, as we deprecated it, it's better in long-term :+1:

src/Fixer/Operator/AlignEqualsFixerHelper.php
```diff
@@ -20,9 +20,24 @@
 /**
  * @author Carlos Cirello <carlos.cirello.nl@gmail.com>
  * @author Graham Campbell <graham@alt-three.com>
+ *
+ * @internal
```
keradus: same here

README.rst
keradus: it's not worth spending time to fix it. I don't consider it a bug. both outcomes are fine :+1:

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -29,66 +29,147 @@
  */
 final class BinaryOperatorSpacesFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    use AlignmentFixerHelperTrait;
```
keradus: ah, finaly as a trait ;)
but actually... we have previous helper extracted as they were used in multiple fixers, now we manage to have one fixer to rule all cases.
would this trait really be used somewhere else ?
if not, there is no need to have it extracted (which encourage to use it somewhere else).

When we want to extract logic, we make separated **class**. it's good.
When we want to extract behavior (feature=trait), we extract it to **trait** - like "LoggerTrait" adds logging into any class it's injected. But that reusable logic.
Here, I'm not sure about alignmnet trait.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -29,66 +29,147 @@
...
+    private static $allowedValues = [
+        self::ALIGN,
+        self::ALIGN_SINGLE_SPACE,
+        self::ALIGN_SINGLE_SPACE_MINIMAL,
+        self::SINGLE_SPACE,
+        null,
+    ];
+
+    /**
+     * @var array
```
keradus: `string[]`

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -29,66 +29,147 @@
...
+     * @var TokensAnalyzer
+     */
+    private $tokensAnalyzer;
+
     /**
-     * @var AbstractAlignFixerHelper[]
+     * @var array<string, string>
      */
-    private $alignFixerHelpers = [];
+    private $alignOperatorTokens = [];
```
keradus: what is the value? string / token ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -219,15 +337,120 @@ private function isDeclareStatement(Tokens $tokens, $index)
...
     }
 
-    private function runHelperFixers(\SplFileInfo $file, Tokens $tokens)
+    private function resolveConfig(array $configuration)
     {
-        /** @var AbstractAlignFixerHelper $helper */
-        foreach ($this->alignFixerHelpers as $helper) {
-            if ($tokens->isChanged()) {
-                $tokens->clearEmptyTokens();
+        $this->configuration = [];
```
keradus: `resolveConfig` is called in `configure`, which already assing value to `$this->configuration`. why we totally override it here? I don't get it.

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
```diff
@@ -22,20 +22,359 @@
...
+\t\$foo         = 1;
+\t\$looooongVar = 2;
+\t\$middleVar   = 1;
+}",
+                "<?php function myFunction() {
+\t\$foo= \t1;
+\t\$looooongVar\t  = 2;
+\t\$middleVar\t= 1;
+}",
+                ['operators' => ['=' => 'align_single_space_minimal']],
```
keradus: no constants for strategy ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: ` @var array<string, string>`
`token type, alignment strategy`

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: I couldn't fit in the BC layer for the old configuration in the current configuration resolving logic, if anyone knows a better way please send a PR to my branch.

SpacePossum:
All green after refactor.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: >in such case, does it make sense to still call parent::configure ?

I think so; as the parent does check of the configuration that the fixer itself doesn't.
Utest should be close to all paths, but a test with if `<=>` configured by default on PHP > 7 is lacking for example, there might be more missing tests.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -219,15 +354,423 @@ private function isDeclareStatement(Tokens $tokens, $index)
...
+    {
+        // fix white space after operator is not needed as BinaryOperatorSpacesFixer took care of this (if strategy is _not_ ALIGN)
+        if ($tokens[$index - 1]->isWhitespace()) {
+            if (self::ALIGN_SINGLE_SPACE_MINIMAL === $alignStrategy) {
+                $content = $tokens[$index - 1]->getContent();
+                if (' ' !== $content && false === strpos($content, "\n") && !$tokens[$tokens->getPrevNonWhitespace($index - 1)]->isComment()) {
+                    $tokens[$index - 1] = new Token([T_WHITESPACE, ' ']);
+                }
+            }
+        } else {
```
julienfalque: Please use an early return.

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -219,15 +354,423 @@ private function isDeclareStatement(Tokens $tokens, $index)
...
+            $placeholder = sprintf(self::ALIGN_PLACEHOLDER, $j);
+
+            if (false === strpos($tmpCode, $placeholder)) {
+                continue;
             }
 
-            $helper->fix($tokens);
+            $lines = explode("\n", $tmpCode);
+            $groupIndex = [];
+            $blockSize = 0;
```
julienfalque: I think you renamed the wrong variable after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573#discussion_r134124989.

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
```diff
@@ -22,20 +23,358 @@
  *
  * @internal
  *
- * @covers \PhpCsFixer\AbstractAlignFixerHelper
```
julienfalque: Are those classes still covered by other tests? Even if deprecated, they are supposed to work until v3.0... but is it worth it?

julienfalque:
I'm not sure merging this PR will allow us to close #2559: does it fix the unalignment of ternary operator like the example in the first post there? If so, can you please add a test case?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: if $blockSize -> $groupIndex, then curren $groupIndex -> ?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -219,15 +354,423 @@ private function isDeclareStatement(Tokens $tokens, $index)
         return false;
     }
 
-    private function runHelperFixers(\SplFileInfo $file, Tokens $tokens)
+    private function resolveConfig(array $configuration)
+    {
+        $this->configuration = [];
```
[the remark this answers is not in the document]
SpacePossum: Later this should all be easy I think, as currently this is setup for some BC layer of the old configuration items.
I tried other ways but didn't work out. So if the configuration should be handled in another way please PR it because I don't know how the current-configuration-PR-way fits better in the new configuration system (yes, this PR pre-dates that X| )

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -219,15 +354,423 @@ private function isDeclareStatement(Tokens $tokens, $index)
...
+    private function injectAlignmentPlaceholders(Tokens $tokens, $startAt, $endAt, $tokenContent)
+    {
+        for ($index = $startAt; $index < $endAt; ++$index) {
+            $token = $tokens[$index];
+
+            $content = $token->getContent();
+            if (
+                strtolower($content) === $tokenContent
+                && $this->tokensAnalyzer->isBinaryOperator($index)
+                && ('=' === $content ? !$this->isEqualPartOfDeclareStatement($tokens, $index) : true)
```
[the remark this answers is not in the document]
SpacePossum: done

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: done

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
SpacePossum: For that you would need the current Bin.Op.Fixer and its test back, you want to maintain those as well as part of the test suit?

SpacePossum:
PR description updated; this PR does _not_ fix the ternary operator part of the issue

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
julienfalque: Wasn't current `$groupIndex` named `$groups` before?

tests/Fixer/Operator/BinaryOperatorSpacesFixerTest.php
julienfalque: Not really :p

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
SpacePossum: :+1: done

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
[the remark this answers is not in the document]
julienfalque: The way you handle the configuration is fine, I don't think there is a better way to handle cases like deprecated options at the moment anyway. But thinking about it again, the `$configuration` property is internal so current state is fine. If one needs to change that to add a new option, it should not be that hard actually so feel free to ignore my previous comment :)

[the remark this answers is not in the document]
keradus: I raised concern for exactly same line as well: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2573#discussion_r139283693 .
For me, we shall not change `configuration` property definition structure defined in parent class. Dev not aware of this fixer won't know that this property is handled differently than how it's defined by parent class.

keradus:
:+1: after CIs are green.
Could you do some squashing ?

SpacePossum:
rebased and squashed to 1 commit for each author

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks for all the help, reviews and ideas everyone here :)

keradus:
It even has more comments than BracesFixer rework, #472

README.rst
xdoctordog: fv

PR #2574, opened 2017-03-02
Add xml.xsd

keradus:
Continuation of #2562 , this time for xml

### DIFF
--- a/doc/xml.xsd
+++ b/doc/xml.xsd
@@ -1,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+    <xs:element name="applied_fixer">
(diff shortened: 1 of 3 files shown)


doc/xml.xsd
```diff
@@ -0,0 +1,83 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
+
+    <xs:element name="applied_fixer">
+        <xs:complexType>
+            <xs:simpleContent>
+                <xs:extension base="xs:string">
+                    <xs:attribute type="xs:string" name="name" use="optional"/>
```
SpacePossum: the element `applied_fixer` is optional, but if it is in it must have a name attribute I think?

keradus: good! done!

Issue #1212, opened 2015-05-21
Fixer for logical operators?


Anahkiasen:
Would it be possible to add a fixer that changes `and` and `or` to `&&` and `||`? Thought that was already in contrib but seems I was mistaken. Might have time for a PR later if the idea is ok?

GrahamCampbell:
They'd be quite risky as they could totally change the code behaviour.

gharlan:
http://php.net/manual/en/language.operators.precedence.php

gharlan:
Example:

```php
if ($a = foo() and bar()) {
}

// it is the same as:
if (($a = foo()) && bar()) {
}

// and not:
if ($a = foo() && bar()) {
}

Anahkiasen:
> They'd be quite risky as they could totally change the code behaviour.

Well no, if it'd be a fixer the other way around yes but not a fixer to && ||

gharlan:
> Well no, if it'd be a fixer the other way around yes but not a fixer to && ||

Hmm, look at my example..

Anahkiasen:
No I understand your example, but my point is most of the time when people use `and` they don't know about the differences and fixing it to && would most of the time make the code work as the developer originally intended. The fact that the fixer would be breaking shouldn't matter, we have a lot of breaking fixers already (strict, strict_params etc)

GrahamCampbell:
> but my point is most of the time when people use and they don't know about the differences and fixing it to && would most of the time make the code work as the developer originally intended.

Waaaaa. I disagree. If people don't know the difference, they need to get a grip. We can't work on the assumption that everyone else is stupid!

Anahkiasen:
I feel like I'm not making myself clear here, this would be a completely _optional_ fixer, in `contrib`. This is PHP *CS* Fixer, a lot of CS define whether to use and/or or &&/|| so it seemed logical to me to have a fixer in that regard. We don't have to make any assumptions about the user base or anything because it doesn't matter, since it's optional

gharlan:
I'm not against the fixer, but it should be marked as risky and in contrib level.
And I wouldn't use this fixer. ;)

SpacePossum:
It would be tricky and could possible lead to troubles. 
However I agree with @Anahkiasen; the new fixer would as risky to use as the `strict somparison`-fixer. I would use the fixer, but I wouldn't blindly rely on the changes it made ;)

Soullivaneuh:
[...]

Issue #2201, opened 2016-09-24
Enforce single-style logical operators


dkarlovi:
We should be able to specify preference for `&&`, `||` instead of `and`, `or`, maybe not even allow named variants [because they act differently](http://stackoverflow.com/a/34848592/672885).

    if ($a or $b) {
        // ...
    }

becomes

    if ($a || $b) {
        // ...
    }

gharlan:
-> #1212

dkarlovi:
Yeah, I've even searched for this issue before submitting it. :(

Duplicate of #1212, closing.

PR #2577, opened 2017-03-06
Add LogicalOperatorsFixer

hkdobrev:
Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1212

This fixer adds the ability to ~~toggle between using keywords for logical operators - `and` and `or`, and using symbols - `&&` and `||`~~ convert logical operators using keywords `and`, `or` to their counterparts with higher precedence `&&`, `||`.

~~The fixer is configurable so you can choose either one. The default is using characters - `&&`, `||`.!~

As discussed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1212, the fixer is risky. ~~It also doesn't try to add parenthesis around conditions when converting from not using keywords to using keywords (which have lower precedence than the assignment operation).~~ People using it need to double check if there was a case where using keywords would have produced different logic and it was intentional.

---

This is my first fixer and my first dive into PHP CS Fixer. Please try to review it thoroughly and provide feedback. Thank you!

### DIFF
--- a/src/Fixer/Operator/LogicalOperatorsFixer.php
+++ b/src/Fixer/Operator/LogicalOperatorsFixer.php
@@ -1,0 +1,76 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Operator;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Haralan Dobrev <hkdobrev@gmail.com>
+ */
+final class LogicalOperatorsFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Use `&&` and `||` logical operators instead of `and` and `or`.',
+            [
+                new CodeSample(
+'<?php
+
+if ($a == "foo" and ($b == "bar" or $c == "baz")) {
+}
+'
+                ),
+            ],
+            null,
+            'Risky, because you must double-check if using and/or with lower precedence was intentional.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_LOGICAL_AND, T_LOGICAL_OR]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if ($token->isGivenKind(T_LOGICAL_AND)) {
+                $tokens[$index] = new Token([T_BOOLEAN_AND, '&&']);
+            } elseif ($token->isGivenKind(T_LOGICAL_OR)) {
+                $tokens[$index] = new Token([T_BOOLEAN_OR, '||']);
+            }
+        }
+    }
+}
--- a/tests/Fixer/Operator/LogicalOperatorsFixerTest.php
+++ b/tests/Fixer/Operator/LogicalOperatorsFixerTest.php
@@ -1,0 +1,74 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\Operator;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Haralan Dobrev <hkdobrev@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Operator\LogicalOperatorsFixer
+ */
+final class LogicalOperatorsFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param string $expected
+     * @param string $input
+     */
+    public function testFix($expected, $input)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
+            [
+                '<?php if ($a == "foo" && $b == "bar") {}',
+                '<?php if ($a == "foo" and $b == "bar") {}',
+            ],
+            [
+                '<?php if ($a == "foo" || $b == "bar") {}',
+                '<?php if ($a == "foo" or $b == "bar") {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" and ($b == "bar" or $c == "baz")) {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" and ($b == "bar" || $c == "baz")) {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" && ($b == "bar" or $c == "baz")) {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" AND ($b == "bar" OR $c == "baz")) {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" and ($b == "bar" OR $c == "baz")) {}',
+            ],
+            [
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" AND ($b == "bar" or $c == "baz")) {}',
+            ],
+        ];
+    }
+}
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -39,6 +39,7 @@ $config = PhpCsFixer\Config::create()
         'header_comment' => ['header' => $header],
         'heredoc_to_nowdoc' => true,
         'list_syntax' => ['syntax' => 'long'],
+        'logical_operators' => true,
         'method_argument_space' => ['ensure_fully_multiline' => true],
         'method_chaining_indentation' => true,
         'multiline_comment_opening_closing' => true,
--- a/README.rst
+++ b/README.rst
@@ -721,6 +721,12 @@ Choose from the list of available rules:
   - ``syntax`` (``'long'``, ``'short'``): whether to use the ``long`` or ``short`` ``list``
     syntax; defaults to ``'long'``
 
+* **logical_operators**
+
+  Use ``&&`` and ``||`` logical operators instead of ``and`` and ``or``.
+
+  *Risky rule: risky, because you must double-check if using and/or with lower precedence was intentional.*
+
 * **lowercase_cast** [@Symfony]
 
   Cast should be written in lower case.


keradus:
Hi. Thank you for opening a PR !

1. I do not see much value in allowing for `&& => and` transformation
2. Please analyse the context of logic term and modify code only when safe / add missing parenthesis, this fixer may not be risky when one will not use naive approach.

tests/Fixer/Operator/LogicalOperatorsFixerTest.php
```diff
@@ -0,0 +1,92 @@
...
+ * @internal
+ */
+final class LogicalOperatorsFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed $expected
+     * @param mixed $input
+     * @param mixed $configuration
```
keradus: string, string, array

hkdobrev:
@keradus Thanks for your quick feedback!

> 1. I do not see much value in allowing for && => and transformation

I agree. I've added it, because some coding standards would actually require the keywords approach. E.g. Kohana :( Anyway, if others don't see value in it I'll be happy to remove this and make the fixer non-configurable.

> 2. Please analyse the context of logic term and modify code only when safe / add missing parenthesis, this fixer may not be risky when one will not use naive approach.

If we remove conversion from `&&` to `and` then the fixer is quite safe. It should be still be marked as risky as it is theoretically possible to change behaviour, but it's the opposite conversion which is riskier.

Example:
```php
$a = $b and $c;
```
In the code above, the `or $c` part is executed only when the `$b` condition evaluates to true and it doesn't affect the value of `$a`. It's most probably a bug and most certainly was not the original intent.

After running the fixer we get:
```php
$a = $b && $c;
```
Now we have `$a` which is always a boolean depending on both `$b` and `$c` conditions.

When using the keywords approach people should normally write:
```php
$a = ($b and $c);
```

Converting this we get:
```php
$a = ($b && $c);
```
The parentheses here are obsolete, but if we have the conversion only in this direction, then we can make it quite safer without adding more complexity to it.

What do you think?

tests/Fixer/Operator/LogicalOperatorsFixerTest.php
hkdobrev: @keradus I've updated it to `string`, `string`, `array<string, bool>` which is even more specific and I've seen the `array<>` annotation in the PHP CS Fixer codebase.

keradus:
Quick note:
please avoid squashing on your side, it's easier to review changes when they are not squashed all the time (I noticed this PR has been pushed with force at least 2 times).
Then, I will squash your commits during merging.

keradus:
Reasoning makes sense to me.
I know this feature is problematic, but it was already requested few times and ppl want to use it, even if they have to be careful about it - definitely not a rule to blindly run on existing legacy code (`$db = connect() or die;`), but a guard to not use `and` in nice, new code base.

Regardless of implementation - what do you think @SpacePossum ?

hkdobrev:
I think this fixer would most likely be used by people who understand the difference and are not using the low precedence of `and` and `or`, but just want to unify their codebase.

@Anahkiasen dkarlovi I'd be happy to get your feedback on this as you have previously opened #1212 and #2201. Thanks!

julienfalque:
Allowing conversion from `&&`/`||` to `and`/`or` has no much added value to me too but, if someone may use it, why not? It doesn't make the implementation much complicated.

Anyway, I vote 👍 for making the fixer aware of the precedence of the operators in the expressions it tries to fix, and remove the need to mark it risky.

SpacePossum:
@hkdobrev thanks for picking this up and sending a PR :+1: 

On concept level I like the idea of having a fixer that changes `and` to `&&`, `or` to `||`, while dealing with precedence issues by adding braces when needed.
I know this leaves risky cases, but I think there are enough valid fixable cases. Marking the fixer `riskly` with good documentation would help.
I think there is enough community support for this fixer (which helps when maintaining it after development and possibly additional development on it later). 
I don't think there is a lot of community support for fixing the other way around TBH. 

@julienfalque 
>if someone may use it, why not? It doesn't make the implementation much complicated.

It adds quite a bit of complexity to the fixer (i.e. making it configurable for example, but also possibly more complex priority cases and such), and if we know of no people who might want to use and help out to maintain it I think it is better left out.

dkarlovi:
@hkdobrev my use case is only to enforce consistency across the entire team, all the enumerated use cases where `and` is actually the correct solution look like huge gotchas to me. I'm all for `and => &&`, not the other way around (would not use it).

Thank you for making this contribution!

hkdobrev:
OK, I'll remove the configuration from the fixer. Yay! 🎉 

@julienfalque 
> Anyway, I vote 👍 for making the fixer aware of the precedence of the operators in the expressions it tries to fix, and remove the need to mark it risky.

I don't think there's a way to make the fixer no risky, though. Even when converting only `and->&&`, `or->||`, there are still cases where the logic of the code would change and there is no easy way to fix it safely. And even if there was an easy way, it's not just about fixing it, because the programmer needs to check whether the initial version was a mistake or intentional. Therefore, it would always stay a risky fixer which needs to be double-checked.

keradus:
> Allowing conversion from `&&`/`||` to `and`/`or` has no much added value to me too but, if someone may use it, why not?

to not spread bad standards

hkdobrev:
I've updated the fixer and the PR description. ✅

julienfalque:
> there are still cases where the logic of the code would change and there is no easy way to fix it safely

Which cases?

> the programmer needs to check whether the initial version was a mistake or intentional

As the purpose of this tool is to fix code style, if the behavior of the code doesn't change, then it's perfectly fine IMO. It should not care if the code contains bugs or not.

The programmer should check his code even if he doesn't run the tool at all. Adding parentheses where he doesn't expect it might even help spot a bug.

keradus:
> As the purpose of this tool is to fix code style, if the behavior of the code doesn't change, then it's perfectly fine IMO. It should not care if the code contains bugs or not.

Not only. It also fix trivial issues of non-experienced devs. Like enforcing strict comparisons (which obviously change code behaviour).

julienfalque:
@keradus Yes but `strict_comparison` fixer is _expected_ to change code behavior. I think the one we're talking about here can be implemented in a way that never changes code behavior, then marking it as risky does not make sense to me.

SpacePossum:
I think we all agree having this fixer not be risky would be the best, lets aim for that and see if it is possible :)
thanks for taking the config option out @hkdobrev

src/Fixer/Operator/LogicalOperatorsFixer.php
```diff
@@ -0,0 +1,80 @@
...
+            'Risky, because you must double-check if using and/or with lower precedence was intentional'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_LOGICAL_AND) || $tokens->isTokenKindFound(T_LOGICAL_OR);
```
SpacePossum: `$tokens->isAnyTokenKindsFound(array(T_LOGICAL_AND, T_LOGICAL_OR))`

src/Fixer/Operator/LogicalOperatorsFixer.php
```diff
@@ -0,0 +1,80 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_LOGICAL_AND) && !$token->isGivenKind(T_LOGICAL_OR)) {
```
SpacePossum: this check is not needed, you check for the candidates below

tests/Fixer/Operator/LogicalOperatorsFixerTest.php
```diff
@@ -0,0 +1,60 @@
...
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" and ($b == "bar" or $c == "baz")) {}',
+            ),
+            array(
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" and ($b == "bar" || $c == "baz")) {}',
+            ),
+            array(
+                '<?php if ($a == "foo" && ($b == "bar" || $c == "baz")) {}',
+                '<?php if ($a == "foo" && ($b == "bar" or $c == "baz")) {}',
```
SpacePossum: please make a test with diff. casing of `or` and `and` so these are covert as well

hkdobrev: Done. ✅

hkdobrev:
Do you guys think it's worth it to detect whether there was an assignment combined with `and`, `or` keywords?

```php
// Lower precedence is used intentionally here
$connection = $db->connect() or die();
// Parentheses need to be around the assignment
($connection = $db->connect()) or die();

// Wrong previous logic
$result = $this->isActive() and this->isEnabled();
// Parentheses around assignments would keep wrong logic
($result = $this->isActive()) and this->isEnabled();
```

I just want to get a consensus, before I spend more time on it. Also ideas about how to approach this are very welcome! I don't have much experience with AST and tokens.

julienfalque:
That's exactly what I (and @keradus too, I think) suggested 😅. This would allow the fixer to always replace `and`/`or` operators without changing code behavior. Note that you would have to do the same for all operators that have more precedence than `and`/`or` and less than `&&`/`||`, so that would be `??`, `?:`, `=`, `+=`,  `-=`,  `*=`,  `**=`,  `/=`,  `.=`,  `%=`,  `&=`,  `|=`,  `^=`,  `<<=` and `>>=`.

hkdobrev:
@julienfalque But I don't want to keep the wrong logic in place with parentheses. Look at my second example. I think having a risky fixer which doesn't put parentheses and you need to look through the changes would be more useful to the people who would want this replacement.

julienfalque:
IMO having the fixer add the parentheses anyway would actually help me spot the issue as it would appear in `fix --diff` output, in `git diff` and even in commits if I still didn't fix the issue. If I see parentheses like in your second example, I would find it weird and check why it happens.

On the other hand, if the fixer does not touch the wrong line, it won't attract my attention and I won't see it, just like I didn't see it when I actually wrote the code or merged it from someone else.

SpacePossum:
I see the point for not wanting to keep code we prop. think of as not good, however the scope of the tool is to fix CS and/or optimize code where possible while trying to not effect functional execution of the code. 
This scopes out doing SCA/sniffing for cases we cannot fix and also that we should try to not change code behavior.
Fixing code to functional the same code can be done here and is therefor preferred.
There are various ways the user can take note that the code is changed (or needs changing) (exit code, CLI options, file changes, etc.).

hkdobrev:
OK, so in order to do this I understand we need to do the following:
- If the `and`/`or` keywords are on the right side of an assignment operation, we should put a single set of parentheses surrounding the left side of the assignment, the equals sign, the first part on the right side of the assignment before the first `and`/`or` keyword.
- If we find `and`/`or` keywords without assignment operators, we wouldn't put parentheses.

Are my assumptions correct? Am I missing some cases? How would I go about in detecting that? Going through the tokens manually? Could you point me to some fixers changing similar things? Thanks!

SpacePossum:
@julienfalque can you confirm the the assumptions of @hkdobrev ?

julienfalque:
@hkdobrev your assumptions are correct. Also, when there is an assignment operator, the fixer should check if there are parentheses before adding them.

And yes, you will have to go through the tokens manually but methods from `Tokens` may help you. I don't think there already are similar fixers.

julienfalque:
@hkdobrev, do you think you would have time to finish this PR?

SpacePossum:
ping @hkdobrev  :)

SpacePossum:
friendly reminder @hkdobrev :)

kubawerlos:
Maybe I'm bit pessimistic but I think making it non-risky would be really hard - with many, many cases to consider - would need to handle nested operators with needed parentheses, also sequences of mixed comparison, assignments with ternary operators. Like this:

```php
$hereBeDragons1 or $hereBeDragons2 || $hereBeDragons3 and $foo = $hereBeDragons4 ? $hereBeDragons5 and $hereBeDragons6 && $hereBeDragons7 : $hereBeDragons8 xor $hereBeDragons9
```

where each of `$hereBeDragons` is at least as complex as:
```php
in_array($baz->getValue(15, function ($x) { return @Crazy\Package::someService($x); }), array_map(SOME_CONST_WITH_ARRAY));
```

Point is we won't be able to make it non-risky at reasonable cost of time, so maybe let's stick with having it risky?

hkdobrev:
> Point is we won't be able to make it non-risky at reasonable cost of time, so maybe let's stick with having it risky?

👍 I've tried to take this up a couple of times, but my skills with the AST are not good enough to do handle all cases and I couldn't continue.

hkdobrev:
I was able to test it in our codebase and the review wasn't that hard. A couple of regex searches validated we didn't have many assignment with `and/or` to go through.

julienfalque:
Yeah I think @kubawerlos is right. We could start with a risky version of the fixer and improve it later.

hkdobrev:
Well, the current version of this branch is a risky fixer. Has something updated in the fixers API that needs to be reflected? Should I rebase on top of latest master?

SpacePossum:
After re-re-reading all of this and looking at what this fixer is doing, I'm :+1: for the current state (after rebase ;) ).

hkdobrev:
@keradus Thanks for picking this up! I've updated the code style so the build could pass.

keradus:
please stop any changes here, i'm in middle of merging this one.
i need to drop your comment, it's will cause Travis to fail

keradus:
So, dropped.
You followed the hint of `fabbot.io` CI. Sadly, for 2.11+ branches fabbot goes crazy and nobody knows why, it ask us to follow some rules, but we don't know which and why :(
eg, since 2.3, we use short array syntax, while fabbot is asking to use long syntax again.
cc @fabpot

keradus:
It was quite a thing here. Happy to finally merge this - over a year in fridge!
Thanks @hkdobrev !

fabpot:
@keradus I'm on it for fabbot (should be fixed later today).

hkdobrev:
Thanks for getting this over the finish line! 🎉 I hope it helps some people.

fabpot:
@keradus Got it! We are executing the `.php_cs` file in a sandbox and we didn't allow using `getenv` (when we were unable to execute the `.php_cs`, we were using a default configuration, which added to the confusion). This is fixed now. So, now, if there is something wrong with the configuration file, we send an error message (to make debugging easier next time it happens). In the meantime, I will manually re-run fabbot on red PRs now to get the good result.

keradus:
big thanks for fixing that @fabpot !

> I will manually re-run fabbot on red PRs now to get the good result.

IMO, if it's manual work, then no need to waste time for it, at least on PHP CS Fixer repo ;)

PR #2580, opened 2017-03-07
NoSpacesAfterFunctionNameFixer - Fix after dynamic call.

SpacePossum:
Fix these cases as well
```diff
--- Original
+++ New
@@ @@
<?php
-$b->$a  (); $c  ();
+$b->$a(); $c();
```

I don't know if this is a special case which should handled diff., but I think its OK.


keradus:
> (maybe we should fix echo $b[0] (); as well?

:+1:

src/Fixer/FunctionNotation/NoSpacesAfterFunctionNameFixer.php
```diff
@@ -65,6 +63,15 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
                     $this->fixFunctionCall($tokens, $index);
                 }
+            } elseif ($tokens[$lastTokenIndex]->equalsAny($braceTypes)) {
+                $type = Tokens::detectBlockType($tokens[$lastTokenIndex]);
+                switch ($type['type']) {
+                    case Tokens::BLOCK_TYPE_ARRAY_INDEX_CURLY_BRACE:
+                    case Tokens::BLOCK_TYPE_DYNAMIC_VAR_BRACE:
+                    case Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE:
+                    case Tokens::BLOCK_TYPE_PARENTHESIS_BRACE:
+                        $this->fixFunctionCall($tokens, $index);
```
keradus: huh? why not simply `if` ?

src/Fixer/FunctionNotation/NoSpacesAfterFunctionNameFixer.php
```diff
@@ -65,6 +63,15 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
                 if (!$tokens[$possibleDefinitionIndex]->isGivenKind(T_FUNCTION)) {
                     $this->fixFunctionCall($tokens, $index);
                 }
+            } elseif ($tokens[$lastTokenIndex]->equalsAny($braceTypes)) {
+                $type = Tokens::detectBlockType($tokens[$lastTokenIndex]);
+                switch ($type['type']) {
```
keradus: `$block['type']`

src/Fixer/FunctionNotation/NoSpacesAfterFunctionNameFixer.php
SpacePossum: to test against 4 possible values? dunno I like this style better I suppose

keradus: swich with only one case is just an if statement. replace `case` with `&&` ;)
remember that consecutive `case` statements without body (or break statement) may be misleading

keradus: What's more, it's like `==` vs `===` :
https://3v4l.org/pH2jC

src/Fixer/FunctionNotation/NoSpacesAfterFunctionNameFixer.php
```diff
@@ -65,6 +63,17 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
                 if (!$tokens[$possibleDefinitionIndex]->isGivenKind(T_FUNCTION)) {
                     $this->fixFunctionCall($tokens, $index);
                 }
+            } elseif ($tokens[$lastTokenIndex]->equalsAny($braceTypes)) {
+                $block = Tokens::detectBlockType($tokens[$lastTokenIndex]);
+                if (
```
julienfalque: Altough this is not reported by fabbio.io or Travis, [PSR-2 does not allow a newline after the `(` or before the `)` of the `if` structure](http://www.php-fig.org/psr/psr-2/#overview):

> Opening parentheses for control structures MUST NOT have a space after them, and closing parentheses for control structures MUST NOT have a space before.

You should use `in_array()`, which would allow you to use such formatting in the array.

By the way, is there a fixer for this?

keradus: fixer for converting into `in_array` ? no

keradus: > PSR-2 does not allow a newline after the ( or before the ) of the if structure:

could you quote the rule ?

julienfalque: No, a fixer to remove the forbidden spaces here.

julienfalque: > could you quote the rule ?

I did.

SpacePossum: The rule doesn't state anything about how multi lines, so I value it as incomplete.
That is my problem with PSR12 in general, it adds some guides which could be amended on PSR1/2 if FIG would allow it, than PSR12 could introduce new rules for new PHP syntax that was not around the time PSR1/2 were made. But lets not focus on this to much, happy to change to whatever style that gets this PR merged.

keradus:
Thank you @SpacePossum.

PR #2582, opened 2017-03-07
Improve AppVeyor and Travis CI build time

julienfalque:
Let's try to make AppVeyor take less time to finish the builds.

Note: I renamed the config file to `.appveyor.yml` for consistency with Travis, now both filenames start with a dot.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-build: off
+build: false
 version: dev-{build}
 shallow_clone: false
 clone_folder: C:\projects\php-cs-fixer
@@ -6,14 +6,19 @@ clone_folder: C:\projects\php-cs-fixer
 environment:
     matrix:
         - php_ver: 5.6.17
+          SKIP_LINT_TEST_CASES: 1
         - php_ver: 7.0.9
 
 cache:
     - '%APPDATA%\Composer'
+    - '%LOCALAPPDATA%\Composer'
+    - C:\tools\php -> .appveyor.yml
+
+init:
+    - SET PATH=C:\tools\php;%PATH%
 
 install:
-    - choco install -my --allow-empty-checksums php -version %php_ver%
-    - SET PATH=C:\tools\php;%PATH%
+    - IF NOT EXIST C:\tools\php (choco install --yes --allow-empty-checksums php --version %php_ver%)
     - cd C:\tools\php
     - copy php.ini-production php.ini
     - echo date.timezone="UTC" >> php.ini
@@ -21,16 +26,16 @@ install:
     - echo extension_dir=ext >> php.ini
     - echo extension=php_curl.dll >> php.ini
     - echo extension=php_openssl.dll >> php.ini
-    - cd C:\tools
+    - cd C:\projects\php-cs-fixer
     - appveyor DownloadFile https://getcomposer.org/composer.phar
-    - appveyor DownloadFile https://phar.phpunit.de/phpunit-5.7.phar
     - git config --global github.accesstoken 5e7538aa415005c606ea68de2bbbade0409b4b8c
+    - php composer.phar global show -ND 2>1 | findstr "hirak/prestissimo" || php composer.phar global require hirak/prestissimo
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --no-interaction --no-progress --optimize-autoloader --prefer-stable --no-ansi
+    - php composer.phar update --no-interaction --no-progress --optimize-autoloader --prefer-stable --no-ansi
 
 test_script:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\phpunit-5.7.phar --verbose
+    - vendor\bin\phpunit --verbose
     - php php-cs-fixer --diff --dry-run -v fix
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,7 +45,7 @@ matrix:
 
 cache:
     directories:
-        - $HOME/.composer/cache
+        - $HOME/.composer
 
 before_install:
     # check phpdbg
@@ -63,7 +63,7 @@ before_install:
     - if [ $TRAVIS_TAG ]; then export TASK_CS=0; fi
 
     # Composer: boost installation
-    - travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
+    - composer global show -ND 2>&1 | grep "hirak/prestissimo" || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
 
     # Composer: enforce given Symfony components version
     - 'if [ "$SYMFONY_VERSION" != "" ]; then sed -i "s/\"symfony\/\([^\"]*\)\": \"^2[^\"]*\"/\"symfony\/\1\": \"$SYMFONY_VERSION\"/g" composer.json; fi'


appveyor.yml
```diff
@@ -9,7 +9,7 @@ environment:
         - php_ver: 7.0.9
 
 cache:
-    - '%APPDATA%\Composer'
+    - '%LOCALAPPDATA%\Composer\files'
```
keradus: on unix we cache `~/.composer/cache`
why only `files` here ?

julienfalque: I just saw it on another project where Composer cache seems to work, but just `'%LOCALAPPDATA%\Composer'` would be fine to me too. I'll remove it once the current build will be finished so we know whether this really improves something.

keradus:
now it's failing :D

julienfalque:
Well that's not a huge success for now 😅

keradus:
I wonder why 5.x job tooks 20+ minutes ...

julienfalque:
@keradus If I ever succeed at improving AppVeyor build time, it will only be one or two minutes. The longest step is the PHPUnit test suite itself so it looks like PHP-CS-Fixer has a real performance issue on Windows.

julienfalque:
I really don't understand why AppVeyor won't restore Composer cache. Is it something that must be enabled on AppVeyor side?

keradus:
The only reason I could see is that your file is not in use.
AppVeyor default config file is `appveyor.yml`. You have renamed it.
One may use custom filename, but first he has to set it up in settings on web. We do not configure anything in CI services directly (keeping whole configuration in file).
Try to un-rename the file.

keradus:
Thats it

in different PR, we have two builds (2 different PHP versions): https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-5016

in this PR, we have only one: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-5022

julienfalque:
@keradus This is because I commented out the PHP 5.x test for now, just to make the whole build faster :) This proves the new filename is taken into account though.

keradus:
:+1:

5.x is indeed that slower on Windows, Idid just run it.
On unix it's 2x slower, on Windows.. argh. Slow part is linting. Please set up `SKIP_LINT_TEST_CASES=1` for PHP 5.x

keradus:
I don't see any config for cache on AppVeyor account settings. Please play with it.
also, you could install `hirak/prestissimo` globally like we do in Travis, it should increase speed a bit.

julienfalque:
Looks like we have no choice but merging the PR to see if the cache works: http://help.appveyor.com/discussions/problems/4048-build-cache-not-restored#comment_39042560.

julienfalque:
Wow, definitely faster with `SKIP_LINT_TEST_CASES=1`!

.appveyor.yml
```diff
@@ -1,36 +1,38 @@
-build: off
-version: dev-{build}
+build: false
```
keradus: what is the difference between `off` and `false` ?

.appveyor.yml
```diff
@@ -1,36 +1,38 @@
-build: off
-version: dev-{build}
```
keradus: why dropping it ?

.appveyor.yml
julienfalque: None, just consistency with `shallow_clone`.

.appveyor.yml
julienfalque: Just wanted to see if it has something to do with cache not working. It doesn't so I'll restore it.

.appveyor.yml
```diff
@@ -1,36 +1,38 @@
...
 
 environment:
     matrix:
         - php_ver: 5.6.17
+          SKIP_LINT_TEST_CASES: 1
         - php_ver: 7.0.9
 
 cache:
-    - '%APPDATA%\Composer'
+    - '%LOCALAPPDATA%\Composer\files'
```
keradus: `files` are cached libraries, but `repo` is which libraries and how to get them. without `repo` the `repo` will be first generated from scratch, and then reuse `files`. why not to cache `repo` as well? (basically, cache whole `Composer` dir)

julienfalque:
I don't see any real improvement with `hirak/prestissimo` but downloading it does take time, not sure it's worth it.

.appveyor.yml
```diff
@@ -1,36 +1,39 @@
...
     - echo date.timezone="UTC" >> php.ini
     - echo memory_limit=512M >> php.ini
     - echo extension_dir=ext >> php.ini
     - echo extension=php_curl.dll >> php.ini
     - echo extension=php_openssl.dll >> php.ini
-    - cd C:\tools
-    - appveyor DownloadFile https://getcomposer.org/composer.phar
-    - appveyor DownloadFile https://phar.phpunit.de/phpunit-5.7.phar
+    - cd C:\projects\php-cs-fixer
+    - php -r "readfile('http://getcomposer.org/installer');" | php
```
keradus: why changing the way of composer installation ?
I would either install it via choco (so we use AppVeyor choco cache) or keep phar download (so we could cache the phar file as well)

.appveyor.yml
```diff
@@ -1,36 +1,39 @@
...
+    - php composer.phar global require hirak/prestissimo
     - git config --global github.accesstoken 5e7538aa415005c606ea68de2bbbade0409b4b8c
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --no-interaction --no-progress --optimize-autoloader --prefer-stable --no-ansi
+    - php composer.phar update --no-interaction --no-progress
```
keradus: why dropping the flags ?

.appveyor.yml
julienfalque: I've tried a lot of things to make this cache work 😅

appveyor.yml
keradus: as you said, we won't until this will be merged

keradus:
just before prestissimo: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-5023
just after: https://ci.appveyor.com/project/keradus/php-cs-fixer/build/dev-5024

about 14s boost per job

julienfalque:
@keradus Installing `hirak/prestissimo` takes more than 14s. But if the cache works, it should not on subsequent builds so 👍

.appveyor.yml
```diff
@@ -1,36 +1,39 @@
...
     - cd C:\tools\php
     - copy php.ini-production php.ini
     - echo date.timezone="UTC" >> php.ini
     - echo memory_limit=512M >> php.ini
     - echo extension_dir=ext >> php.ini
     - echo extension=php_curl.dll >> php.ini
     - echo extension=php_openssl.dll >> php.ini
-    - cd C:\tools
+    - cd C:\projects\php-cs-fixer
     - appveyor DownloadFile https://getcomposer.org/composer.phar
```
keradus: add `composer.phar` to cache and then:
`if not exist composer.phar appveyor DownloadFile https://getcomposer.org/composer.phar`

the only thing is to set up the folder where we want to download it.

julienfalque: Are you sure we want to cache Composer itself? There would be no way to invalidate this cache I think.

keradus:
> Installing `hirak/prestissimo` takes more than 14s.
And yet, overall time of whole build is 14s faster, and that includes time of installing prestissimo. Prestissimo parallelize installing of packages, and we ask for installing 44 packages...

julienfalque:
I would also like to try to add `C:\tools\php -> .appveyor.yml` to the cache items and install PHP only when the directory does not exist. But I'm not sure whether this will speed up builds or just break everything 😀. Shall I do it?

.appveyor.yml
keradus: why not, no need for installing same app 50 times per day.
you could set up `->` syntax to make phar file cache dependent on yml config.

keradus:
what about 5.x vs 7.x ?
currently, they are installed in same dir.
Won't it mix the cache? or cache is stored per job, not per build ?

keradus:
(btw, don't cache whole `C:\tools\`, there were other autoinstalled apps there as well)

julienfalque:
I couldn't find information about whether cache is per job or per build. I hope it's per job, otherwise we obviously can't use it for PHP, and we might even face issues with Composer. So what do we do?

keradus:
If you are willing to play with it on some private playground repo it would be great.
If not, let us not include `php` dir into cache - as a safe approach.

julienfalque:
Sure, I'll see what I can do with my fork of the project.

.appveyor.yml
julienfalque: This would work for PHP installation because versions are hardcoded in the file, so it will never change until we update the file. On the other hand, we always download the latest version of Composer, which can change at any time and is not related to this config file. A new minor version has even been released between your comment and my response here 😁.

Anyway, I just made a few tests with my fork and downloading the file only takes 2 seconds, so it really isn't worth it.

keradus: 2 seconds here, 5 seconds there...
we don't rely on edge composer, we could have outdated one.

ok, let it stay like it is

julienfalque:
Good news: cache works fine and is job dependent so caching PHP installation is not an issue, see [builds on my fork](https://ci.appveyor.com/project/julienfalque/php-cs-fixer/history).

I made more tests with and without `hirak/prestissimo` and it actually makes the build take more time when Composer cache is available:
* in [this job](https://ci.appveyor.com/project/julienfalque/php-cs-fixer/build/dev-5/job/jshdoyaspd8e3q59): installing the plugin took 16 seconds, updating the project dependencies took 22 seconds: total is **38 seconds**;
* in [this job](https://ci.appveyor.com/project/julienfalque/php-cs-fixer/build/dev-6/job/upwxnrrr6vv4953t): no plugin installation, updating the project dependencies took **25 seconds**.

Do you still want to use the plugin anyway?

keradus:
Great to hear that you found some time to play with it ;)

 > Do you still want to use the plugin anyway?

What if we would cache both:
> '%APPDATA%\Composer'
> '%LOCALAPPDATA%\Composer'
?

julienfalque:
It won't change anything: on master `%APPDATA%\Composer` is used and Composer never retrieves vendors from cache so I think the directory is never used at all and AppVeyor just stores an empty cache for it.

keradus:
I let myself modify a bit config for debugging.
I restored the Roaming caching and see the cache - sadly, as you pointed, it's never used so it's empty as we didn't install anything globally on master yet.
But my point is we could follow the approach you made with `php` package. `composer g show -ND` will tell us if there is already plugin installed or not, if not - we could install it, if yet, we won't.

julienfalque:
Sure, will try that.

julienfalque:
I think we're good now :)

`%LOCALAPPDATA%\Composer%` is used for cache while `%APPDATA%\Composer` is used for everything else so both are required.

.appveyor.yml
```diff
@@ -1,36 +1,45 @@
...
-    - appveyor DownloadFile https://phar.phpunit.de/phpunit-5.7.phar
     - git config --global github.accesstoken 5e7538aa415005c606ea68de2bbbade0409b4b8c
+    - IF EXIST %APPDATA%\Composer\composer.json (
+          php composer.phar global show -ND | findstr "hirak/prestissimo" || php composer.phar global require hirak/prestissimo
+      ) ELSE (
+          php composer.phar global require hirak/prestissimo
+      )
```
keradus: all 5 lines could be replaced with:
`php composer.phar global show -ND 2>1 | findstr "hirak/prestissimo" || php composer.phar global require hirak/prestissimo`

(`2>1` is added)

keradus:
could we apply same approach for composer plugin for Travis ?

keradus:
Great work here, awesome to see AppVeyor fast finally!

julienfalque:
Thanks for your suggestions and help!

PR #2585, opened 2017-03-08
Update readme to make risky easier to set

Rockstar04:
I had to dig through code and commits to figure out what this wanted to be set to, was it a list of risky fixers I wanted or true/false?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -183,7 +183,7 @@ A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
 
 The ``--allow-risky`` option allows you to set whether risky rules may run. Default value is taken from config file.
-Risky rule is a rule, which could change code behaviour. By default no risky rules are run.
+Risky rule is a rule, which could change code behaviour. By default no risky rules are run. Allowed input for this flag is ``yes`` or ``no``
 
 The ``--stop-on-violation`` flag stops execution upon first file that needs to be fixed.
 


julienfalque:
Actually I would have expected this option to not accept value at all and always mean _yes_ when passed, and _no_ when not passed. Wouldn't it be better to fix the option definition instead?

keradus:
~then, we would have 4 possible cases:~
~- not passed~
~- passed without value~
~- passed with yes (legacy, to be removed at 3.x)~
~- passe with no (legacy, to be removed at 3.x)~

How to revoke risky allowance via CLI when it's granted in config file?
Remember that we take CLI value, if there is none we fall back to config file.

julienfalque:
Right, I forgot about the config file part ;)

SpacePossum:
I like better the better docs, however the description is not completely correct;

>  $ php php-cs-fixer fix --dry-run --allow-risky=aa

is also allowed. Maybe besides the docs it would be nice to check if `yes` or `no` is passed, otherwise throw a config exception stating what values are allowed? (some as when no value is passed, would be nice to list the possible options in the exception message)

julienfalque:
What about `--allow-risky` and `--disallow-risky`?

keradus:
What is the benefit of having two flags instead of one option with 2 values ?

Rockstar04:
```
php-cs-fixer --version
PHP CS Fixer version 2.1.0 by Fabien Potencier and Dariusz Ruminski
```
```
php-cs-fixer fix --allow-risky=true
php-cs-fixer fix --allow-risky=aa
```
Both of the options above return an error, only `yes` allows the risky test to run.

SpacePossum:
Can you post the error?

I think it is about the finder, does the following work?

> $ php-cs-fixer fix --allow-risky=aa .

Rockstar04:
Very interesting.

`php-cs-fixer fix --allow-risky=aa .` DID work, but the following:

screenshot_2017-03-08_16-16-02
Command from the screenshot:
 ` php-cs-fixer fix --allow-risky=aa --rules=@Symfony,@Symfony:risky,ordered_imports module/`

SpacePossum:
The added dot at the end of the command tells the fixer to fix files in the current directory, so you won't get the error about the finder anymore.
The value `aa` for `--allow-risky` is resolved as `false`, which triggers the exception you got as there are risky fixers in the set.
I opt. for changing this to: only allowing `yes` or `no` as value for `--allow-risky` and update the code that resolves the value and update the exception messages accordingly.

julienfalque:
> What is the benefit of having two flags instead of one option with 2 values ?

With current option, one must know how to tell _yes_ or _no_, hence the current PR. One could expect other values to work as weel: _true_, _false_, _Y_, _N_ and so on. Using only boolean options means no value to know about, passing just one option or the other is enough, and they are listed in the command help so no need to check the README file.

Furthermore, you would let Symfony Console process them completely instead of validating the value manually. The only case that would still require manual check is when both options are passed, which should trigger an error.

SpacePossum:
I think one option would be more easy, it would remove the need for testing for conflicting flags.
When nothing or something else than "yes" or "no" is passed we can show the exception message:
```
`--allow-risky` requires value `yes` or `no`, got `X`.
```
Would be nice to roll out for `--using-cache` as well.

keradus:
> Furthermore, you would let Symfony Console process them completely instead of validating the value manually. The only case that would still require manual check is when both options are passed, which should trigger an error.

It's the same, checking if there are 2 flags at the same time or if there is unknown value.

keradus:
ping

keradus:
ping @Rockstar04

Rockstar04:
What do you need from me @keradus ? From the tests and results above it appears there is some inconsistencies with either the underlying CLI library or some other issue. This isn't a blocker for me, I know now that I have to use "yes". All I wanted to accomplish with the readme tweak was to make it easier on anyone else who didn't find it obvious.

SpacePossum:
There are two things left to do:
- rebase
- the README.rst is generated, so the code that generates it needs updating as well;  https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.1/src/Console/Command/FixCommandHelp.php#L83

after you change the command help code, you can regenerate the README using
`php php-cs-fixer readme > README.rst`

Rockstar04:
If changes are going to be made to how "allow risky" is toggled, then there the documentation can be updated as part of that change.

PR #2587, opened 2017-03-09
NoClosingTagFixerTest - handle file without operations

keradus:


keradus:
Thank you @SpacePossum.

Issue #1408, opened 2015-09-23
Progress Bar


DavidGarciaCat:
Hey there,

Usually I run these commands on my projects:

```
$ bin/php-cs-fixer fix features
$ bin/php-cs-fixer fix --fixers=-visibility spec
$ bin/php-cs-fixer fix src
```

IMO display a progress bar, displaying how many files have been processed, or maybe the %, will be great. At this moment I am working with thousand of files, and I don't know how long need to wait before run other commands.

What do you think abut this?

Best,

GrahamCampbell:
Hmm, well we iterate files without knowing the "count" if you will, so we don't know the progress.

DavidGarciaCat:
Maybe we can add a `count()` at the beginning, just to know that, and keep the developer up to date about the % completed.

If there is another developer that knows better than me where can be applied, IMO this update can help. If there are no other developers to handle this, I can review the code, but will take more time, just because I will need to review the code first.

funivan:
% at any point can change in different ranges. You can see 30% and then 15% so :-1: 
How many files have been processed is not so important IMHO.

GrahamCampbell:
Yeh, :-1: for having to load the entire list of files into memory at once.

GrahamCampbell:
Well, or double iterating...

DavidGarciaCat:
OK if you think this is not so important, closing this issue now

keradus:
TBH I would like to have that feature as alternative to "dot" progress indicator.

DavidGarciaCat:
IMO I think this can be a very useful feature (maybe can display this if we pass an argument).
Any other comment about this?

flip111:
You can:

1. always count all the files up ahead
2. optionally always count all the files up ahead (with command line flag)
3. only count all the files up ahead once, then after cache this number. On the next run your estimate might be a bit off if you added or deleted a file.

SpacePossum:
closing as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2591 has been merged, the new options for the process view should be enough

PR #2591, opened 2017-03-12
Improve process output

julienfalque:
This PR changes the process output to look like PHPUnit's one:
* limits line length to 80 characters;
* displays the progression percentage before each newline so you know whether you have time to get a coffee 😉;

Output

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -187,6 +187,18 @@ Risky rule is a rule, which could change code behaviour. By default no risky rul
 
 The ``--stop-on-violation`` flag stops execution upon first file that needs to be fixed.
 
+The ``--show-progress`` option allows you to choose the way process progress is rendered:
+
+* ``none``: disables progress output;
+* ``run-in``: simple single-line progress output;
+* ``evaluating``: multiline progress output with number of files and percentage on each line. Note that with this option, the files list is evaluated before processing to get the total number of files and then kept in memory to avoid using the file iterator twice. This has an impact on memory usage so using this option is not recommended on very large projects.
+
+If the option is not provided, it defaults to ``run-in`` unless a config file that disables output is used, in which case it defaults to ``none``. This option has no effect if the verbosity of the command is less than ``verbose``.
+
+.. code-block:: bash
+
+    $ php php-cs-fixer.phar fix --verbose --show-progress=evaluating
+
 The command can also read from standard input, in which case it won't
 automatically fix anything:
 
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "doctrine/annotations": "^1.2",
         "ext-tokenizer": "*",
         "sebastian/diff": "^1.1",
-        "symfony/console": "^2.3 || ^3.0",
+        "symfony/console": "^2.4 || ^3.0",
         "symfony/event-dispatcher": "^2.1 || ^3.0",
         "symfony/filesystem": "^2.4 || ^3.0",
         "symfony/finder": "^2.2 || ^3.0",
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -103,6 +103,7 @@ final class FixCommand extends Command
                     new InputOption('diff', '', InputOption::VALUE_NONE, 'Also produce diff for each file.'),
                     new InputOption('format', '', InputOption::VALUE_REQUIRED, 'To output results in other formats.'),
                     new InputOption('stop-on-violation', '', InputOption::VALUE_NONE, 'Stop execution on first violation.'),
+                    new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in, or estimating).'),
                 )
             )
             ->setDescription('Fixes a directory or a file.')
@@ -135,6 +136,7 @@ final class FixCommand extends Command
                 'diff' => $input->getOption('diff'),
                 'stop-on-violation' => $input->getOption('stop-on-violation'),
                 'verbosity' => $verbosity,
+                'show-progress' => $input->getOption('show-progress'),
             ),
             getcwd()
         );
@@ -169,12 +171,23 @@ final class FixCommand extends Command
             }
         }
 
-        $showProgress = $resolver->getProgress();
+        $progressType = $resolver->getProgress();
+        $finder = $resolver->getFinder();
+
+        if ('none' === $progressType || null === $stdErr) {
+            $progressOutput = new NullOutput();
+        } elseif ('run-in' === $progressType) {
+            $progressOutput = new ProcessOutput($stdErr, $this->eventDispatcher, null);
+        } else {
+            $finder = new \ArrayIterator(iterator_to_array($finder));
+            $progressOutput = new ProcessOutput($stdErr, $this->eventDispatcher, count($finder));
+        }
+
         $runner = new Runner(
-            $resolver->getFinder(),
+            $finder,
             $resolver->getFixers(),
             $resolver->getDiffer(),
-            $showProgress ? $this->eventDispatcher : null,
+            'none' !== $progressType ? $this->eventDispatcher : null,
             $this->errorsManager,
             $resolver->getLinter(),
             $resolver->isDryRun(),
@@ -182,11 +195,6 @@ final class FixCommand extends Command
             $resolver->getDirectory(),
             $resolver->shouldStopOnViolation()
         );
-
-        $progressOutput = $showProgress && $stdErr
-            ? new ProcessOutput($stdErr, $this->eventDispatcher)
-            : new NullOutput()
-        ;
 
         $this->stopwatch->start('fixFiles');
         $changed = $runner->fix();
--- a/src/Console/Output/ProcessOutput.php
+++ b/src/Console/Output/ProcessOutput.php
@@ -52,11 +52,40 @@ final class ProcessOutput implements ProcessOutputInterface
      */
     private $output;
 
-    public function __construct(OutputInterface $output, EventDispatcher $dispatcher)
+    /**
+     * @var int
+     */
+    private $files;
+
+    /**
+     * @var int
+     */
+    private $processedFiles = 0;
+
+    /**
+     * @var int
+     */
+    private $symbolsPerLine;
+
+    /**
+     * @param OutputInterface $output
+     * @param EventDispatcher $dispatcher
+     * @param int|null        $nbFiles
+     */
+    public function __construct(OutputInterface $output, EventDispatcher $dispatcher, $nbFiles)
     {
         $this->output = $output;
         $this->eventDispatcher = $dispatcher;
         $this->eventDispatcher->addListener(FixerFileProcessedEvent::NAME, array($this, 'onFixerFileProcessed'));
+
+        if (null !== $nbFiles) {
+            $this->files = $nbFiles;
+
+            //   80               (lines max length)
+            // - total length x 2 (e.g. "  1 / 123" => 6 digits and padding spaces)
+            // - 11               (extra spaces, parentheses and percentage characters, e.g. " x / x (100%)")
+            $this->symbolsPerLine = 80 - strlen((string) $this->files) * 2 - 11;
+        }
     }
 
     public function __destruct()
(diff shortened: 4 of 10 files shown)


composer.json
```diff
@@ -17,7 +17,7 @@
         "php": "^5.3.6 || >=7.0 <7.2",
         "ext-tokenizer": "*",
         "sebastian/diff": "^1.1",
-        "symfony/console": "^2.3 || ^3.0",
+        "symfony/console": "^2.4 || ^3.0",
```
julienfalque: The `BufferedOutput` class used in test was introduced in 2.4.

src/Console/Command/FixCommand.php
```diff
@@ -184,7 +184,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         );
 
         $progressOutput = $showProgress && $stdErr
-            ? new ProcessOutput($stdErr, $this->eventDispatcher)
+            ? new ProcessOutput($stdErr, $this->eventDispatcher, count($resolver->getFinder()))
```
GrahamCampbell: This `count` is very expensive. It forces the iterator to go all the way through.

src/Console/Output/ProcessOutput.php
```diff
@@ -68,6 +85,20 @@ public function onFixerFileProcessed(FixerFileProcessedEvent $event)
...
         $status = self::$eventStatusMap[$event->getStatus()];
         $this->output->write($this->output->isDecorated() ? sprintf($status['format'], $status['symbol']) : $status['symbol']);
+
+        ++$this->processedFiles;
+
+        $symbolsOnCurrentLine = $this->processedFiles % $this->symbolsPerLine;
+        $end = $this->processedFiles === $this->files;
+        if (0 === $symbolsOnCurrentLine || $end) {
+            $this->output->writeln(sprintf(
+                '%s %'.strlen((string) $this->files).'d / %d (%3d%%)',
```
GrahamCampbell: Bit of a hack to cast this to a string, in order to then use it later in sprintf, uncast. Abstractly, we can't be sure the `%d` gives something of the same length as your string cast did.

src/Console/Command/FixCommand.php
julienfalque: On the project (445 files), it takes ~0.05s (basic `microtime()` test). Is it really that expensive?

I tried to store the files in an array and use it for both counting and running the fixers. Don't know if it reduces execution time but it consumes 4M memory.

src/Console/Output/ProcessOutput.php
julienfalque: > Abstractly, we can't be sure the %d gives something of the same length as your string cast did.

What values would give such difference?

src/Console/Command/FixCommand.php
SpacePossum: The cost depends on the Finder implementation used and on what set it is used, so the numbers provided give some indication but not much IMHO.
It is the same discussion as here https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1408
I'm not in favor (but not a strong opinion) to always count ahead, but as an opt-in I think this looks nice (or provide an opt-out back to the current way of dots).
Happy to hear more opinions though!

src/Console/Output/ProcessOutput.php
SpacePossum: only downside I see is that is done in the constructor (76) already. I would remove it from the constructor and reuse the `strlen((string) $this->files)` in a var.

composer.json
SpacePossum: 2.3 is supported to May this year (http://symfony.com/roadmap?version=2.3#checker), would be nice if we could keep it (especially if the up is only for test). However if this is a lot of work, than up'ping is :+1:  with me

keradus: bumping is fine

sstok: Note that 2.4 is EOL so you'd properly want 2.7 (if it's still active) or 2.8 😃

keradus: As I do always answer to you @sstok, if we could work properly on both 2.8 and 2.4 no need for dropping 2.4.
I saw a lot of legacy, monolithic projects that updates their codebase to have better CS and this kind of change would disallow them to install the tool as dev-dependency and require them to install it as phar file.

src/Console/Command/FixCommand.php
keradus: @julienfalque , imagine finder that checks for content, like all files that do not contain `ignore CS`.
but I do agree that this is very edge case.

storing outcome of finder in memory may be difficult if the project is extra large. we could try approach like `if ($count < SOME_LIMIT) { $finder = ArrayIterator(iterator_to_array($finder)); }`
The, all the rest work out of the box, count is cheap and we don't iterate over disc for searching for the files twice if there are not that much files.

keradus:
what about 2 reporters - `txt` and `simple_txt` (current one) ?

src/Console/Command/FixCommand.php
SpacePossum: To would enforce that the finders used must be rewindable for when after first iteration the limit is reached.

keradus: if it's not rewindable we could convert it to array anyway (even if it would be big memory-consuming array)
next argument to have 2 `ProcessOutput` implementations

composer.json
julienfalque: As @SpacePossum said, the class is only used for tests and is not very complex, maybe we can just write our own?

keradus: if we want to use feature of 2.4, let us bump to 2.4.

src/Console/Output/ProcessOutput.php
julienfalque: Why in the destructor?

SpacePossum: Because it is already there?

julienfalque: There is nothing in the destructor except the removal of the event listener. Did you mean `onFixerFileProcessed`?

SpacePossum: oh sorry! I misread the GH interface, on the callback sounds good to me :+1:

GrahamCampbell:
> Do we want to have that kind of progress indicator

Maybe. Would be cool to have it as optional.

julienfalque:
What about adding an option like `--show-progress` that would enable showing the number of files and the progress percentage? By default lines length would still be limited to 80 characters, without further informations.

keradus:
IMO, we should:
1. add new option for command
- `new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Specify progress indicator.')` (no default value !)
- possible values are: `none, run-in, estimating` (means none, current way, new way); feel free to propose other naming; would be good to document the difference between them
- if option is not provided in CLI command run, it shall fallback to `ConfigInterface::getHideProgress : bool`, which means `run-in` or `none`
- resolving which `ProgressOutput` shall be use needs to be done via `ConfigurationResolver::getProgress`, which will return `ProcessOutputInterface` now instead of boolean flag

2. Update config
- `Config::getProgress() : ProcessOutputInterface` shall be added, but without touching interface
- in `ConfigurationResolver` we should detect if the method is there, if so we shall use it, if not we shall use `ConfigurationResolver::getProgress`

3. new `ProgressOutput` class
- as finder may be any `iterable`, it may not be rewindable, having that in mind, if we gonna use new progress class in given run I would conditionally transform finder to regular array, so it will become countable (not all `iterables` are `countable` !). I know it will increase memory usage for big projects, we should mention it in user documentation of progress types.

julienfalque:
Sounds good to me. Should `ConfigInterface::getHideProgress` be deprecated?

keradus:
if you want to deprecate sth you need to say that this is abandoned or shall be done other way. here, it's the second case. but to say "do it in another way", you need to provide that way. new interface perhaps

julienfalque:
`Config::getProgress()` could be part of a new interface that extends `ConfigInterface`.

julienfalque:
New option added, but I didn't implement it exactly like you suggested: the `ProcessOutputInterface` instance is still created in the `FixCommand`.

Also, with `run-in` option output is still limited to 80 characters per line but without the numbers so it has no impact on memory.

keradus:
`run-in` name had sense when it would represent current output.

but you did changed it (thus the name may not be the best as you suggested, as the format is different) - not sure is it good to change current output instead of providing different progress type.

name should follow the representation. I am very against `full`, as it's full in what way? use full screen, `run-in` use only 40% of full type, or what? and then, even if one would be able to provide reasoning why `full` may be an OK name for current types, then I will add one-liner progress `bar`, which works completly different than `full`/`estimating`

julienfalque:
Updated README file and reverted `run-in` to original single-line rendering.

> I am very against full, as it's full in what way?

Sure, doesn't really make sense, let's keep current options 👍

README.rst
```diff
@@ -187,6 +187,18 @@ Risky rule is a rule, which could change code behaviour. By default no risky rul
 
 The ``--stop-on-violation`` flag stops execution upon first file that needs to be fixed.
 
+The ``--show-progress`` option allows you to choose the way process progress is rendered:
+
+* ``none``: disables progress output;
+* ``run-in``: simple single-line progress output;
+* ``evaluating``: multiline progress output with number of files and percentage on each line.
```
keradus: please mention memory issue with `evaluating` for big projects

src/Console/Output/ProcessOutput.php
```diff
@@ -52,11 +52,31 @@
...
+     * @var int
+     */
+    private $processedFiles = 0;
+
+    /**
+     * @var int
+     */
+    private $symbolsPerLine;
+
+    public function __construct(OutputInterface $output, EventDispatcher $dispatcher, $nbFiles = null)
```
keradus: typehinting of last param is missing, you need to add a docblock

src/Console/Output/ProcessOutput.php
```diff
@@ -52,11 +52,31 @@
...
     {
         $this->output = $output;
         $this->eventDispatcher = $dispatcher;
         $this->eventDispatcher->addListener(FixerFileProcessedEvent::NAME, array($this, 'onFixerFileProcessed'));
+
+        if (null !== $nbFiles) {
+            $this->files = $nbFiles;
+            $this->symbolsPerLine = 80 - strlen((string) $this->files) * 2 - 11;
```
keradus: for me those are magic values, 80, 11, 2 ;)

src/Console/Output/ProcessOutput.php
```diff
@@ -68,6 +88,27 @@ public function onFixerFileProcessed(FixerFileProcessedEvent $event)
     {
         $status = self::$eventStatusMap[$event->getStatus()];
         $this->output->write($this->output->isDecorated() ? sprintf($status['format'], $status['symbol']) : $status['symbol']);
+
+        ++$this->processedFiles;
+
+        if (null !== $this->files) {
+            $symbolsOnCurrentLine = $this->processedFiles % $this->symbolsPerLine;
+            $end = $this->processedFiles === $this->files;
```
keradus: boolean vars should be named with prefix like `is` or `has`

src/Console/Output/ProcessOutput.php
keradus: also, please remove default value

src/Console/Output/ProcessOutput.php
```diff
@@ -68,6 +88,27 @@ public function onFixerFileProcessed(FixerFileProcessedEvent $event)
     {
         $status = self::$eventStatusMap[$event->getStatus()];
         $this->output->write($this->output->isDecorated() ? sprintf($status['format'], $status['symbol']) : $status['symbol']);
+
+        ++$this->processedFiles;
```
keradus: could be done only when `this->files` is not null

src/Console/Output/ProcessOutput.php
```diff
@@ -52,11 +52,40 @@
      */
     private $output;
 
-    public function __construct(OutputInterface $output, EventDispatcher $dispatcher)
+    /**
+     * @var int
```
keradus: `?int`

keradus:
Thank you @julienfalque.

PR #2594, opened 2017-03-14
Travis - generate coverage report at 7.1 and clean up build matrix

keradus:
ref https://bugs.php.net/bug.php?id=74137 :/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,8 @@ env:
 matrix:
     fast_finish: true
     include:
-        - php: 7.0
-          env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: 7.1
-          env: TASK_SCA=1
+          env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: nightly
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -16,10 +16,8 @@ env:
 matrix:
     fast_finish: true
     include:
-        - php: 7.0
-          env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: 7.1
-          env: TASK_SCA=1
```
keradus: it's enough to run SCA only on `nightly`, as we treat it as non-experimental job

.travis.yml
```diff
@@ -16,10 +16,8 @@ env:
...
+          env: DEPLOY=yes TASK_TESTS_COVERAGE=1
```
keradus: coverage at 7.1 is more fully

.travis.yml
```diff
@@ -29,7 +27,7 @@ matrix:
         - php: 5.5
           env: SKIP_LINT_TEST_CASES=1
         - php: 5.6
-        - php: 7.1
+        - php: 7.0
           env: SYMFONY_VERSION="^2.8"
```
keradus: no need for having two 7.1 jobs, we could check Symfony 2.8 at PHP 7.0

PR #2595, opened 2017-03-14
NoUnusedImportsFixer - Fix import usage detection with properties

julienfalque:
When an imported name is similar to a property in the file, it is always considered as used.

### DIFF
--- a/tests/Fixer/Import/NoUnusedImportsFixerTest.php
+++ b/tests/Fixer/Import/NoUnusedImportsFixerTest.php
@@ -381,6 +381,26 @@ EOF;
         $this->doTest($expected, $input);
     }
 
+    public function testPropertyName()
+    {
+        $expected = <<<'EOF'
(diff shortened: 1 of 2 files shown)


tests/Fixer/Import/NoUnusedImportsFixerTest.php
```diff
@@ -381,6 +381,26 @@ public function testVariableName()
...
+
+$foo->bar = null;
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
+use Foo\Bar;
+
+$foo->bar = null;
```
keradus: what about `$bar = null` ? (or do we have this case already?)

julienfalque: This case is already covered by a test.

keradus:
Thank you @julienfalque.

PR #2601, opened 2017-03-16
PhpUnitDedicateAssertFixer - fix typos in is_numeric and is_resource

ivan1986:
fix is_*e2808b*

PR #2603, opened 2017-03-16
Add InvisibleSymbols Fixer

ivan1986:
fixer for #2601

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -477,10 +477,6 @@ Choose from the list of available rules:
 
   Code MUST use configured indentation type.
 
-* **invisible_symbols**
-   | Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other
-   | unicode symbols.
-
 * **is_null** [@Symfony:risky]
 
   Replaces is_null(parameter) expression with ``null === parameter``.
@@ -719,6 +715,11 @@ Choose from the list of available rules:
 * **no_whitespace_in_blank_line** [@Symfony]
 
   Remove trailing whitespace at the end of blank lines.
+
+* **non_printable_character**
+   | Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other
+   | invisible unicode symbols.
+   | *Rule is: risky.*
 
 * **normalize_index_brace** [@Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -717,9 +717,11 @@ Choose from the list of available rules:
   Remove trailing whitespace at the end of blank lines.
 
 * **non_printable_character**
-   | Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other
-   | invisible unicode symbols.
-   | *Rule is: risky.*
+
+  Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other
+  invisible unicode symbols.
+
+  *Risky rule: risky when strings contain intended invisible characters.*
 
 * **normalize_index_brace** [@Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -478,7 +478,8 @@ Choose from the list of available rules:
   Code MUST use configured indentation type.
 
 * **invisible_symbols**
-   | Remove ZWSP, NBSP and other unicode symbols.
+   | Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other
+   | unicode symbols.
 
 * **is_null** [@Symfony:risky]
 
--- a/README.rst
+++ b/README.rst
@@ -477,6 +477,9 @@ Choose from the list of available rules:
 
   Code MUST use configured indentation type.
 
+* **invisible_symbols**
+   | Remove ZWSP, NBSP and other unicode symbols.
+
 * **is_null** [@Symfony:risky]
 
   Replaces is_null(parameter) expression with ``null === parameter``.
--- a/src/Fixer/Basic/InvisibleSymbolsFixer.php
+++ b/src/Fixer/Basic/InvisibleSymbolsFixer.php
@@ -54,7 +54,7 @@ final class InvisibleSymbolsFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Remove ZWSP, NBSP and other unicode symbols.',
+            'Remove Zero-width space (ZWSP), Non-breaking space (NBSP) and other unicode symbols.',
             array(
                 new CodeSample(
 '<?php
--- a/src/Fixer/Basic/InvisibleSymbolsFixer.php
+++ b/src/Fixer/Basic/InvisibleSymbolsFixer.php
@@ -1,0 +1,85 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Basic;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer ZWSP, NBSP and other unicode symbols.
+ *
+ * @author Ivan Boprzenkov <ivan.borzenkov@gmail.com>
+ */
+final class InvisibleSymbolsFixer extends AbstractFixer
+{
+    private $symbolsReplace;
+
+    public function __construct()
+    {
+        parent::__construct();
+        $this->symbolsReplace = array(
+            pack('CCC', 0xe2, 0x80, 0x8b) => '',
+            pack('CCC', 0xe2, 0x80, 0x87) => ' ',
+            pack('CCC', 0xe2, 0x80, 0xaf) => ' ',
+            pack('CCC', 0xe2, 0x81, 0xa0) => '',
+            pack('CC', 0xc2, 0xa0) => ' ',
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            $token->setContent(strtr($token->getContent(), $this->symbolsReplace));
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Remove ZWSP, NBSP and other unicode symbols.',
+            array(
+                new CodeSample(
+'<?php
+
+echo "'.pack('CCC', 0xe2, 0x80, 0x8b).'Hello'.pack('CCC', 0xe2, 0x80, 0x87).'World'.pack('CC', 0xc2, 0xa0).'!";
+'
+                ),
+            )
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run first (at least before Fixers that using Tokens) - for speed reason of whole fixing process
+        return 90;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
+    }
+}
--- a/src/Fixer/Basic/InvisibleSymbolsFixer.php
+++ b/src/Fixer/Basic/InvisibleSymbolsFixer.php
@@ -18,7 +18,7 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Tokens;
 
 /**
- * Fixer ZWSP, NBSP and other unicode symbols.
+ * Fixer Zero-width space (ZWSP), Non-breaking space (NBSP) and other unicode symbols.
  *
  * @author Ivan Boprzenkov <ivan.borzenkov@gmail.com>
  */
--- a/src/Fixer/Basic/NonPrintableCharacterFixer.php
+++ b/src/Fixer/Basic/NonPrintableCharacterFixer.php
@@ -30,11 +30,11 @@ final class NonPrintableCharacterFixer extends AbstractFixer
     {
         parent::__construct();
         $this->symbolsReplace = array(
-            pack('CCC', 0xe2, 0x80, 0x8b) => '',
-            pack('CCC', 0xe2, 0x80, 0x87) => ' ',
-            pack('CCC', 0xe2, 0x80, 0xaf) => ' ',
-            pack('CCC', 0xe2, 0x81, 0xa0) => '',
-            pack('CC', 0xc2, 0xa0) => ' ',
+            pack('CCC', 0xe2, 0x80, 0x8b) => '', // ZWSP
+            pack('CCC', 0xe2, 0x80, 0x87) => ' ', // FIGURE SPACE
+            pack('CCC', 0xe2, 0x80, 0xaf) => ' ', // NBSP
+            pack('CCC', 0xe2, 0x81, 0xa0) => '', // WORD JOINER
+            pack('CC', 0xc2, 0xa0) => ' ', // NO-BREAK SPACE
         );
     }
 
--- a/src/Fixer/Basic/NonPrintableCharacterFixer.php
+++ b/src/Fixer/Basic/NonPrintableCharacterFixer.php
@@ -73,18 +73,18 @@ echo "'.pack('CCC', 0xe2, 0x80, 0x8b).'Hello'.pack('CCC', 0xe2, 0x80, 0x87).'Wor
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isAnyTokenKindsFound(array(
-            T_STRING,
-            T_STRING_VARNAME,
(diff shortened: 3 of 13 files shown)


README.rst
```diff
@@ -295,6 +295,9 @@ Choose from the list of available rules:
 * **indentation_type** [@PSR2, @Symfony]
    | Code MUST use configured indentation type.
 
+* **invisible_symbols**
+   | Remove ZWSP, NBSP and other unicode symbols.
```
keradus: shortcuts need to be resolved

keradus:
Hi !
Thanks for raising it that fast !
It's a great start for a fixer !

IMO, we should make it configurable, allowing to fix any of 3 types:
- comments (remove odd whitespaces inside comments)
- whitespaces (remove odd whitespaces from other whitespaces)
- strings (so case you found initial issue)

that way one may want to fix only whitespaces or whitespaces+comments, but not strings (he could desire odd chars in them in edge cases)

also, be aware that if token would be empty after dropping odd spaces - we should remove whole token (could happen only for `T_WHITESPACE` token)

ivan1986:
@keradus IMO does not need to configure it - if we found this symbols in the code it's a copy-paste mistake or we do some real strange.

src/Fixer/Basic/InvisibleSymbolsFixer.php
```diff
@@ -0,0 +1,85 @@
...
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            $token->setContent(strtr($token->getContent(), $this->symbolsReplace));
```
keradus: no need for translation in all tokens, please check the tokens first (eg, no need for fixing `T_PRIVATE`)

tests/Fixer/Basic/InvisibleSymbolsFixerTest.php
```diff
@@ -0,0 +1,54 @@
...
+    /**
+     * @param string       $expected
+     * @param string       $input
+     * @param \SplFileInfo $file
+     *
+     * @dataProvider provideExamples
+     */
+    public function testFix($expected, $input, $file)
+    {
+        $this->doTest($expected, $input, $file);
```
keradus: is there a need for passing a file ?

src/Fixer/Basic/InvisibleSymbolsFixer.php
```diff
@@ -0,0 +1,85 @@
...
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer Zero-width space (ZWSP), Non-breaking space (NBSP) and other unicode symbols.
+ *
+ * @author Ivan Boprzenkov <ivan.borzenkov@gmail.com>
+ */
+final class InvisibleSymbolsFixer extends AbstractFixer
```
keradus: This rule may change content of vars, thus it's risky (I agree that in most cases it would be a proper fix, but sadly it may be desired to have it in string values in some edge cases).
For that, please mark rule as risky

src/Fixer/Basic/InvisibleSymbolsFixer.php
```diff
@@ -0,0 +1,85 @@
...
+
+namespace PhpCsFixer\Fixer\Basic;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer Zero-width space (ZWSP), Non-breaking space (NBSP) and other unicode symbols.
```
keradus: this is unicode symbol: `☯`
how does the rule deal with it ?

SpacePossum:
Nice fixer idea, love to see this in.
I was thinking about proposed configuration and just thinking out loud here, maybe we could make generic 'NonPrintableCharacterFixer' (name can change of course) ?
We could than merge the 'line_ending' fixer in this as well.

Than we could add configuration of character mapping, for example:
* `\r\n` => `\n`
* `\r` => ''
* `0xe2` => ' '

We allow configuration for 'T_ENCAPSED_AND_WHITESPACE, T_END_HEREDOC, T_OPEN_TAG, T_WHITESPACE, T_COMMENT, T_DOC_COMMENT, T_START_HEREDOC, any thing else' to give control about what should be fixed (making the fixer risky by configuration).

WDYT?

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -54,21 +54,33 @@ public function fix(\SplFileInfo $file, Tokens $tokens)
...
 
 echo "'.pack('CCC', 0xe2, 0x80, 0x8b).'Hello'.pack('CCC', 0xe2, 0x80, 0x87).'World'.pack('CC', 0xc2, 0xa0).'!";
 '
                 ),
-            )
+            ),
+            null,
+            null,
+            null,
+            'Risky change strings data.'
```
keradus: This is not a valid sentence.

tests/Fixer/Basic/NonPrintableCharacterFixerTest.php
```diff
@@ -22,33 +22,26 @@
...
+     * @param string      $expected
+     * @param null|string $input
      *
      * @dataProvider provideExamples
      */
-    public function testFix($expected, $input, $file)
+    public function testFix($expected, $input = null)
     {
-        $this->doTest($expected, $input, $file);
+        $this->doTest($expected, $input);
```
keradus: FYI: even when the content was read from file, it's not obligatory to pass the file to `doTest`.

ivan1986:
@SpacePossum change name, about tokens - a good solution would be to process all text in one strtr, but I do not know how to do it.

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,97 @@
...
+'<?php
+
+echo "'.pack('CCC', 0xe2, 0x80, 0x8b).'Hello'.pack('CCC', 0xe2, 0x80, 0x87).'World'.pack('CC', 0xc2, 0xa0).'!";
+'
+                ),
+            ),
+            null,
+            null,
+            null,
+            'Risky if invisible unicode contains in data strings.'
```
julienfalque: What about: `Risky when strings contain intended invisible characters.`?

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,97 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run first (at least before Fixers that using Tokens) - for speed reason of whole fixing process
+        return 90;
```
julienfalque: Is this really required? How does it improve process speed?

keradus:
> a good solution would be to process all text in one strtr, but I do not know how to do it.

it's not a good solution. it's slow solution that will have to revisit all content even if we don't want to modify most of it.

tests/Fixer/Basic/NonPrintableCharacterFixerTest.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param null|string $input
+     *
+     * @dataProvider provideExamples
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideExamples()
```
julienfalque: `provideFixCases`?

tests/Fixer/Basic/NonPrintableCharacterFixerTest.php
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return array(
```
julienfalque: Could you please add more test cases, with invisible characters in comments for example?

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,97 @@
...
+
+namespace PhpCsFixer\Fixer\Basic;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer Zero-width space (ZWSP), Non-breaking space (NBSP) and other invisible unicode symbols.
```
julienfalque:
```diff
-Fixer
+Removes
```

Also, what about adding links to some doc about these symbols? Would that be useful?

keradus:
@SpacePossum , about your idea - it may indeed be nice to have next `General***` fixer that could be used inside different fixers, like we already do eg with `GeneralPhpdocAnnotationRemoveFixer`.
Then, this one, tabs to spaces, line endings, could be rewritten to use that `General***`.
But let us keep it outside of scope of this PR.

tests/Fixer/Basic/NonPrintableCharacterFixerTest.php
```diff
@@ -0,0 +1,55 @@
...
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                '<?php echo "Hello World !";',
```
keradus: not needed, it's autogenerated from next case

ivan1986: fixed

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,88 @@
...
+{
+    private $symbolsReplace;
+
+    public function __construct()
+    {
+        parent::__construct();
+        $this->symbolsReplace = array(
+            pack('CCC', 0xe2, 0x80, 0x8b) => '',
+            pack('CCC', 0xe2, 0x80, 0x87) => ' ',
+            pack('CCC', 0xe2, 0x80, 0xaf) => ' ',
```
keradus: it is not tested

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,88 @@
...
+ */
+final class NonPrintableCharacterFixer extends AbstractFixer
+{
+    private $symbolsReplace;
+
+    public function __construct()
+    {
+        parent::__construct();
+        $this->symbolsReplace = array(
+            pack('CCC', 0xe2, 0x80, 0x8b) => '',
```
keradus: would be nice to provide names of that symbols

ivan1986: hmm, no ideas, add coments

src/Fixer/Basic/NonPrintableCharacterFixer.php
ivan1986: added

ivan1986:
added remove empty tokens
added isAnyTokenKindsFound
I not sure about type tokens list in isAnyTokenKindsFound, please check it and if it wrong tell me correct list for fix

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            $token->setContent(strtr($token->getContent(), $this->symbolsReplace));
+            if ($token->isEmpty()) {
```
SpacePossum: the check and remove is not needed, it's done by the token itself already; https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.1/src/Tokenizer/Token.php#L477  (the `remove` method does not exists anyway I think?)

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,102 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(array(
+            T_STRING,
+            T_STRING_VARNAME,
+            T_INLINE_HTML,
+            T_WHITESPACE,
```
SpacePossum: for a candidate check I would return `count($tokens) > 1` as that would result in the same but a bit faster

src/Fixer/Basic/NonPrintableCharacterFixer.php
ivan1986: removed empty

add test for repeat symbols - all ok

keradus: good catch here

src/Fixer/Basic/NonPrintableCharacterFixer.php
keradus: it's not the same. 
and `count($tokens) > 1` is already checked **before** calling `isCandidate` ;)

cc @ivan1986

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,88 @@
...
+        return count($tokens) > 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            $token->setContent(strtr($token->getContent(), $this->symbolsReplace));
```
keradus: still, https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2603#discussion_r106534435
for me, lack of resolving this one comment is last blocker for this PR, all the rest are just minors

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        return count($tokens) > 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
```
keradus: index is not used

keradus:
please fix deprecations

src/Fixer/Basic/NonPrintableCharacterFixer.php
ivan1986: fixed

src/Fixer/Basic/NonPrintableCharacterFixer.php
ivan1986: added list tokens and add if

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,102 @@
...
+        T_STRING,
+        T_STRING_VARNAME,
+        T_INLINE_HTML,
+        T_WHITESPACE,
+        T_COMMENT,
+        T_ENCAPSED_AND_WHITESPACE,
+        T_CONSTANT_ENCAPSED_STRING,
+        T_END_HEREDOC,
+        T_OPEN_TAG,
+        T_DOC_COMMENT,
```
keradus: if I would remove `T_DOC_COMMENT` test for this fixer would still pass.
it is not good. probably you need to provide more test cases

keradus: I see you added test for `T_DOC_COMMENT`. that's nice, but following are still missing:
```
        T_STRING,
        T_STRING_VARNAME,
        T_INLINE_HTML,
        T_WHITESPACE,
        T_ENCAPSED_AND_WHITESPACE,
        T_END_HEREDOC,
        T_OPEN_TAG,
        T_START_HEREDOC,
```

ivan1986:
broken only in php 7.2 because <7.2 in composer

keradus:
That's not true. Read the error log.
Failing because of trailing whitespaces in:
> ./tests/Fixer/Basic/NonPrintableCharacterFixerTest.php:62: ./tests/Fixer/Basic/NonPrintableCharacterFixerTest.php:66: function f(string $p) ./tests/Fixer/Basic/NonPrintableCharacterFixerTest.php:71: ./tests/Fixer/Basic/NonPrintableCharacterFixerTest.php:75: function f(string $p)

julienfalque:
@ivan1986 Actually Travis build fails because there are trailing whitespaces like [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2603/files#diff-ae2304b53b736acede7e556747d7a9f0R72) (quite ironic 😁)

src/Fixer/Basic/NonPrintableCharacterFixer.php
```diff
@@ -0,0 +1,100 @@
...
+ * Removes Zero-width space (ZWSP), Non-breaking space (NBSP) and other invisible unicode symbols.
+ *
+ * @author Ivan Boprzenkov <ivan.borzenkov@gmail.com>
+ */
+final class NonPrintableCharacterFixer extends AbstractFixer
+{
+    private $symbolsReplace;
+
+    private static $tokens = array(
+        T_NUM_STRING,
```
keradus: `T_NUM_STRING` and `T_WHITESPACE` are not covered by tests

keradus:
Thank you @ivan1986.

PR #2610, opened 2017-03-17
Fix early bracket close for do-while loop inside an if without brackets

### LINKED ISSUE #2609: Early bracket close for do-while loop inside an if without brackets
Given the example file `test.php`

```php
<?php

if (true)
    do {
        echo 'Fail to fix :/'.PHP_EOL;
    } while (false);
```
I get this code when execute `php-cs-fixer fix test.php`

```php
if (true) {
    do {
        echo 'Fail to fix :/'.PHP_EOL;
    }
} while (false);
```

As is shown the `if` statement ends before the `while` which causes code to break.
I found code like this in an old library for connecting to S3 at a project I'm working on.

My `php` and and `PHP CS Fixer` versions:
```
$ php -v
PHP 7.1.2 (cli) (built: Feb 17 2017 10:52:17) ( NTS )
    with Xdebug v2.5.1, Copyright (c) 2002-2017, by Derick Rethans
$ php-cs-fixer -V
PHP CS Fixer 2.2.0-DEV by Fabien Potencier and Dariusz Ruminski    
```

felixgomez:
Fix #2609

### DIFF
--- a/src/Console/Command/FixCommandHelp.php
+++ b/src/Console/Command/FixCommandHelp.php
@@ -83,7 +83,7 @@ display a summary of proposed fixes, leaving your files unchanged.
 The <comment>--allow-risky</comment> option allows you to set whether risky rules may run. Default value is taken from config file.
 Risky rule is a rule, which could change code behaviour. By default no risky rules are run.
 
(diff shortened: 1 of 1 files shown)


keradus:
Hi
Thanks for contributing !

You forgot to commit test changes

PR #2614, opened 2017-03-18
Ensure that '*Fixer::fix()' won't crash when running on non-candidate collection

keradus:


SpacePossum:
don't you need the PHP version test and token count > 0 test to fix all cases of calling the fix method while not a candidate?

keradus:
I do, should be done globally.

For now, we assume that one will always call `isCandidate`, which may not happen.
We need to always check restrictions in `fix`, while `isCandidate` should be only quick check that is helpfull, but is optional.
IMO, we should make `isCandidate` removed from the interface at v3 and call them directly in AbstractFixer.

keradus:
what do you think @SpacePossum about always call `isCandidate` in `fix` method ?
Then, we could remove `isCandidate` at v3 or maybe keep it for custom runners (like integrations into IDEs)

SpacePossum:
Maybe keeping the `isCandidate` method is the best route and indeed call it from the `fix` method.
For performance we should not no longer call `isCandidate` than when fixing, but just `fix`.

keradus:
@SpacePossum , reworked

src/AbstractFixer.php
```diff
@@ -45,6 +46,13 @@ public function __construct()
         }
     }
 
+    public function fix(\SplFileInfo $file, Tokens $tokens)
```
Slamdunk: Since now Fixers are obliged to implement `applyFix` method, it would be better to make this method final, so Fixer creators are not confused implementing both.
It would be a BC break, but maybe on 2.3 would be ok

julienfalque: Introducing a BC break requires a new major version and is not allowed on minor versions as per SemVer.

Slamdunk: Requiring a method to be final is as breaking as requiring a new (protected) method (`applyFix`).
Users are not affected, maintainers of side-projects (that require PHP-CS-Fixer) are affected.

julienfalque: Correct, the new abstract method is already a BC break. I think it should not be abstract and have an empty body instead. Side-projects that extend this class will still override `fix` so it should not break.

Slamdunk: I am a side-project maintainer: I ended up having business logic on `fix`, but had to implement `applyFix`: what should have I done?
If `fix` would be final, I would got two parse error, and the rename would be easy.

keradus: No, this is not a BC breaker.
Side-projects shall use `FixerInterface`, which has only `fix` method (no `applyFix`).
`AbstractFixer` is marked as **internal** and must not be used outside of this package.

Slamdunk: If it is internal and [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/COOKBOOK-FIXERS.md#step-1---creating-files), https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2670

PR #2615, opened 2017-03-18
Add DescribeCommand test

julienfalque:


composer.json
```diff
@@ -17,7 +17,7 @@
         "php": "^5.3.6 || >=7.0 <7.2",
         "ext-tokenizer": "*",
         "sebastian/diff": "^1.1",
-        "symfony/console": "^2.3 || ^3.0",
+        "symfony/console": "^2.4 || ^3.0",
```
julienfalque: `Symfony\Component\Console\Tester\CommandTester::getStatusCode()` was introduced in 2.4.

GrahamCampbell: 👎 You should instead add this as a dev-dependency, and don't change the version constraint on the original.

julienfalque: Version 2.3 won't be covered by tests using `--prefer-lower` Composer option anymore, is that ok?

GrahamCampbell: Hmm.

GrahamCampbell: Maybe just skip that test if symfony version is too low.

julienfalque: Added a `method_exists()` check.

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -0,0 +1,155 @@
...
+
+
+EOT;
+
+        $this->assertSame(
+            str_replace(array('%spaces%', "\n"), array('   ', PHP_EOL), $expected),
+            $commandTester->getDisplay()
+        );
+
+        if (method_exists($commandTester, 'getStatusCode')) {
```
GrahamCampbell: Would be better to make this test it's own test method, so we can actually mark it as skipped properly, instead of silently not testing it.

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+            'command' => $command->getName(),
+        ));
+    }
+
+    /**
+     * @return CommandTester
+     */
+    private function execute()
+    {
+        $fixer = $this->getMockBuilder('PhpCsFixer\Tests\Console\Command\DefinedAndConfigurableFixerInterface')->getMock();
```
keradus: Use Prophecy instead

tests/Console/Command/DefinedAndConfigurableFixerInterface.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace PhpCsFixer\Tests\Console\Command;
+
+use PhpCsFixer\Fixer\ConfigurableFixerInterface;
+use PhpCsFixer\Fixer\DefinedFixerInterface;
+
+/**
+ * @internal
+ */
+interface DefinedAndConfigurableFixerInterface extends ConfigurableFixerInterface, DefinedFixerInterface
```
keradus: this is no go

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+        $this->assertSame(
+            str_replace(array('%spaces%', "\n"), array('   ', PHP_EOL), $expected),
+            $this->execute()->getDisplay()
+        );
+    }
+
+    public function testExecuteStatusCode()
+    {
+        if (!method_exists('Symfony\Component\Console\Tester\CommandTester', 'getStatusCode')) {
```
keradus: please remember to drop it with #2591

tests/Console/Command/DescribeCommandTest.php
julienfalque: Done.

tests/Console/Command/DefinedAndConfigurableFixerInterface.php
julienfalque: Removed as Prophecy allows to create stubs that implement several interfaces.

keradus:
Thank you @julienfalque.

PR #2616, opened 2017-03-19
Exclude more tests in phar version

keradus:


SpacePossum:
Won't this remove `src/Test/IntegrationCase.php` for example, that is a valid API/endpoint right?

GrahamCampbell:
I guess that class should have been marked as `@internal`.

SpacePossum:
don't think so, it a useful class for authors of 3rd party fixers

keradus:
that's my goal @SpacePossum 
this class is still visible when one install package as library as dependency via composer, but when one install project as CLI app, he doesn't need to have it.

SpacePossum:
alright than :)

Issue #2000, opened 2016-06-17
Fix indentation in OOP Nesting


citosid:
Is there a way to fix this:

``` php
$object->method()
       ->another()
       ->oneMore();
```

And turn it into this:

``` php
$object->method()
    ->another()
    ->oneMore();
```

The difference is that, according to [This Link](http://pear.php.net/manual/en/rfc.cs-enhancements.splitfunccall.php) when nesting, we should have only 4 spaces in the line under, not in the same column the first `->` was added.

keradus:
Wow, PEAR standards... someone is still using them :D

One way or another, I like it the second way, a fixer would be nice for it.
Could you work on it please?

citosid:
@keradus we are not using them fully, we are using PSR-2 and PSR-4, but this standard from PEAR is nice, the code is more readable, and I could not find anything in PSR.

I could, do you have any suggestion on how to do it?

keradus:
if the first token of line is `->` then check indentation of prev line, add one level of deep to it and fix current line indentation, if needed

Also, please check README and CONTRIBUTING files

keradus:
Closing as #3063 bringing this feature is already released for ages.
We simply forgot to close this feature request ;)

PR #2617, opened 2017-03-19
Method chaining indentation fixer [WIP]

boliev:
Closes: #2000


src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $this->ident = $this->whitespacesConfig->getIndent();
+        foreach ($tokens as $index => $token) {
+            $content = $token->getContent();
+            if ($content === '->' && $index > 0) {
```
keradus: - use `equals` instead
- change `foreach` into `for` and you will have one condition less for each loop

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+            array(new CodeSample("\$user->setEmail('voff.web@gmail.com')\n    ->setPassword('233434');"),
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return 50;
```
keradus: reasoning is missing

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
```
keradus: you are fixing content around `->`, check for it instead

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,62 @@
...
+            array(
+                '<?php
+
+    $user->setEmail(\'voff.web@gmail.com\')
+        ->setPassword(\'233434\')
+        ->setEmailConfirmed(false)
+        ->setEmailConfirmationCode(\'123456\')
+        ->setHashsalt(\'1234\')
+        ->setTncAccepted(true);
+', // This is expected output
```
keradus: please drop the comment

keradus:
I like the idea. Sadly, tests are failing

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $this->ident = $this->whitespacesConfig->getIndent();
+        foreach ($tokens as $index => $token) {
+            $content = $token->getContent();
+            if ($content === '->' && $index > 0) {
+                $prev = $tokens[$index - 1];
+                $prevContent = $prev->getContent();
+                $matches = array();
+                if (preg_match('/([\n\r|\n])(\s*)/uis', $prevContent, $matches)) {
```
keradus: utf? caseless? dot-all?
this regex cannot gain from that modifiers

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,62 @@
...
+                '<?php
+
+    $user->setEmail(\'voff.web@gmail.com\')
+
+     ->setPassword(\'233434\')
+        ->setEmailConfirmed(false)
+->setEmailConfirmationCode(\'123456\')
+
+                ->setHashsalt(\'1234\')
+  ->setTncAccepted(true);
```
keradus: add utest with Windows line endings and tabs

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,62 @@
...
+        ->setHashsalt(\'1234\')
+        ->setTncAccepted(true);
+', // This is expected output
+                '<?php
+
+    $user->setEmail(\'voff.web@gmail.com\')
+
+     ->setPassword(\'233434\')
+        ->setEmailConfirmed(false)
+->setEmailConfirmationCode(\'123456\')
```
keradus: add utest like
```php
$foo
         ->bar1() // comment
      ->bar2() /*
comment
*/
  ->bar3()
    // comment
        ->bar4();
```

julienfalque:
Sometimes it makes sense to indent more than once some of the chained function calls. A good example is the Symfony Config component, which has a builder with a fluent interface to define bundle configuration. Each nested level of configuration has an extra indent, see the [examples in the documentation](https://symfony.com/doc/current/components/config/definition.html). Does this fixer handle such case?

Also, I use to put the first method call and the semicolon on their own line too, e.g.:

```php
$foo
    ->bar()
    ->baz()
;
```

instead of

```php
$foo->bar()
    ->baz();
```

I think it would be nice if the fixer can handle that.

keradus:
1. No, as it cannot knows that in this case those methods should be on extra indent level or not
2. different rule, different fixer

julienfalque:
> No, as it cannot knows that in this case those methods should be on extra indent level or not

Of course, but it could at least detect that used indentation is valid and leave it untouched.

> different rule, different fixer

👍

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Method chaining MUST indented with one tab.',
```
julienfalque: Should be _"Method chaining MUST **be** indented with one tab."_; not sure about using the word _tab_ though.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
boliev: I can't use `$tokens[$index]->equals('->')`, because '->' token is an array
`class PhpCsFixer\Tokenizer\Token#137 (4) {
  private $content =>
  string(2) "->"


  private $id =>
  int(366)
  private $isArray =>
  bool(true)
  private $changed =>
  bool(false)
}
`
and there is a condition in equals function
`if ($this->isArray !== is_array($otherPrototype)) {
            return false;
        }`

SpacePossum: please try
```php
$tokens[$index]->isGivenKind(T_OBJECT_OPERATOR)
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
SpacePossum: i.e.
```php
return tokens->isTokenKindFound(T_OBJECT_OPERATOR);
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
keradus: `isGivenKind` is better here indeed, but `equals` would also work - `equals([T_OBJECT_OPERATOR])`

boliev: It doesn't work too. T_OBJECT_OPERATOR have value 361
```define ('T_OBJECT_OPERATOR', 361);```
but token's id is 366
```
class PhpCsFixer\Tokenizer\Token#131 (4) {
  private $content =>
  string(2) "->"
  private $id =>
  int(366)
  private $isArray =>
  bool(true)
  private $changed =>
  bool(false)
}


public function isGivenKind($possibleKind)
{
    return $this->isArray && (is_array($possibleKind) ? in_array($this->id, $possibleKind, true) : $this->id === $possibleKind);
}
```

Comment (keradus) [inline src/Fixer/Whitespace/MethodChainingIndentationFixer.php]: @boliev 
It works, as I did prove it with latest commit on your branch.
(The only failure is about not-updated readme file, you probably need to rebase a branch)

File: src/Fixer/Whitespace/MethodChainingIndentationFixer.php [hunk 9]
```diff
@@ -93,8 +94,8 @@ public function isCandidate(Tokens $tokens)
     private function getRightIdents($index, Tokens $tokens)
     {
         for ($i = $index; $i >= 0; --$i) {
-            if (preg_match('/[\n\r|\n](\s*)/uis', $tokens[$i]->getContent(), $matches)) {
-                if ($tokens[$i + 1]->getContent() === '->') {
+            if (preg_match('/[\n\r|\n](\s*)/i', $tokens[$i]->getContent(), $matches)) {
```
keradus: why `i` ?

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,106 @@
...
+     * @var string
+     */
+    private $ident;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $this->ident = $this->whitespacesConfig->getIndent();
```
keradus: no need for that property, it is used only in one place.

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+                ->setHashsalt(\'1234\')
+  ->setTncAccepted(true);
+',
+            ),
+            array(
+                "<?php\r\n\$user->setEmail('voff.web@gmail.com')\r\n    ->setPassword('233434')\r\n    ->setEmailConfirmed(false)\r\n    ->setEmailConfirmationCode('123456')\r\n    ->setHashsalt('1234')\r\n    ->setTncAccepted(true);",
+                "<?php\r\n\$user->setEmail('voff.web@gmail.com')\r\n\r\n     ->setPassword('233434')\r\n\t->setEmailConfirmed(false)\r\n\t\t  ->setEmailConfirmationCode('123456')\r\n->setHashsalt('1234')\r\n->setTncAccepted(true);",
```
keradus: Fixer is implementing `WhitespacesAwareFixerInterface`.
For that, you need to write test cases that are using different configurations.
Take a look at other fixers implementing this interface

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
boliev:
```
$foo->aaa()
->bbb(
     )->ccc()
        ->ddd()->
  eee();
```
I'm not sure how it's should be after fix

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,152 @@
...
+final class MethodChainingIndentationFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $indent = $this->whitespacesConfig->getIndent();
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+        for ($index = 1; $index < count($tokens); ++$index) {
```
keradus: please put some empty lines for readability here
in general, you could put them before all control structures (if, for, ...) and after `}`

keradus:
WIP flag was removed, sorry, but nobody saw that

@julienfalque , could you make a review please ?

README.rst
```diff
@@ -539,6 +539,10 @@ Choose from the list of available rules:
   - ``keep_multiple_spaces_after_comma`` (``bool``): whether keep multiple spaces
     after comma; defaults to ``false``
 
+* **method_chaining_indentation**
+
+  Method chaining MUST be indented with one tab.
```
julienfalque: I would not use the word *tab* as code can be indented with spaces (and that's what most PHP projects do nowadays). What about:

```
Method chaining MUST be properly indented.
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer for rules defined in PSR2 ¶2.4.
```
julienfalque: This fixer is not PSR-2 specific as it supports indentation with tabs as well.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+            array(new CodeSample("<?php\n\$user->setEmail('voff.web@gmail.com')\n         ->setPassword('233434');"),
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // fixer is idempotent
```
julienfalque: Like all fixers should be ;)

You are using the default priority here, you can drop the method entirely as it's the same in parent class.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    /**
+     * @param int    $index
+     * @param Tokens $tokens
+     * @param mixed  $indent
```
julienfalque: Expected type is `string`. But I would remove the parameter as it can be retrieved from `$this->whitespacesConfig`.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @return string
+     */
+    private function getRightIndents($index, Tokens $tokens, $indent)
+    {
+        for ($i = $index; $i >= 0; --$i) {
+            $currentWhitespaces = $this->isLineBreak($tokens[$i]);
+
+            if ($currentWhitespaces !== false) {
```
julienfalque: Yoda style.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+
+                return $currentWhitespaces.$indent;
+            }
+        }
+
+        return $indent;
+    }
+
+    /**
+     * @param $index
```
julienfalque: Missing type.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+     *
+     * @return bool
+     */
+    private function needLineBreak($index, Tokens $tokens)
+    {
+        $prevMeaningful = $tokens->getPrevMeaningfulToken($index);
+        $isComment = false;
+
+        for ($i = $index; $i > $prevMeaningful; --$i) {
+            if ($tokens[$i]->equals(array(T_OBJECT_OPERATOR)) || $this->isLineBreak($tokens[$i]) !== false) {
```
julienfalque: Yoda style.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+                if ($isComment) {
```
julienfalque: Can be replaced with `return $isComment;`.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,160 @@
...
+     * @return string|bool
+     */
+    private function isLineBreak(Token $token)
+    {
+        $matches = array();
+        $content = $token->getContent();
+
+        if (preg_match('/\R(\s*)/', $content, $matches)) {
+            if (!isset($matches[1])) {
+                false;
```
julienfalque: This `if` does not do anything.

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,120 @@
...
+ * @author Vladimir Boliev <voff.web@gmail.com>
+ *
+ * @internal
+ */
+final class MethodChainingIndentationFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed      $expected
```
julienfalque: Expected type is `string`.

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,120 @@
...
+     * @param null|mixed $input
```
julienfalque: `null|string`.

boliev:
> I still think it would be nice to support cases with several indent levels by not forcing them to have only one indent level
Not sure about it. This means that for example code:
`
$foo
    ->foo()
        ->bar()
            ->baz()
        ->bar()
            ->foo()
                ->baz()
    ->qux()
`
will not fixed?

SpacePossum:
```php
$foo
    ->foo()
        ->bar()
            ->baz()
        ->bar()
            ->foo()
                ->baz()
    ->qux()
```

I don't see logic on which the idents should be preserved, @julienfalque what kind of detection/rules for this do you propose (pseudo logic is good enough)?

julienfalque:
I would use the following rules:
* indentation on each line must be valid per the configured indentation type:
    * for tabs, any number of tabs is valid;
    * for spaces, number of spaces must be a multiple of the configured number of spaces for 1 indent level (e.g. with 4 spaces: 4, 8, 12, ...);
* each method can add one and only one indent level;
* each method can remove any indent level, except the first one.

With these rules, this is valid:
```php
$foo
    ->foo()
        ->bar()
            ->baz()
        ->bar()
            ->foo()
                ->baz()
    ->qux()
```
This:
```php
$foo
    ->foo()
            ->bar()
        ->baz()
->qux()
```
would become:
```php
$foo
    ->foo()
        ->bar()
        ->baz()
    ->qux()
```

SpacePossum:
sounds complex; besides
```
$foo
    ->foo()
            ->bar()
        ->baz()
->qux()
```
how we know;
```
$foo
    ->foo()
        ->bar()
            ->baz()
->qux()
```

was not intended?

In summary I'm -1 on having another indent configuration option (next to the `WhitespacesFixerConfig`).
Having the fixer to fix to the same indention for all calls sounds good to me and enough for first implementation. If some one want to add more options to it it can be done later.

julienfalque:
I didn't say this needs a new option, `WhitespacesFixerConfig` is enough. But yes, I agree this can be done in a separate PR.

julienfalque:
> how we know;
> ```
> $foo
>     ->foo()
>         ->bar()
>             ->baz()
> ->qux()
> ```
> was not intended?

That's my point: there is no way to know whether extra indentation levels are intended. This fixer assumes they never are and always removes them despite there are valid cases where they can be intended.

Loosening the fixer about this would require a new configuration option if done in a separate PR, while it should be the default behavior IMO.

boliev:
@keradus @julienfalque Guys, are we closing this pr?

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,141 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param Token $token
+     *
+     * @return string|bool
+     */
+    private function isLineBreak(Token $token)
```
julienfalque: IMO, `is*()` methods should not mix boolean and non-boolean return values. Can we change this method so it returns `string|null` instead?

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $matches = array();
```
julienfalque: Initializing this variable is not required as the `preg_match()` call will do it.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,141 @@
...
+        $content = $token->getContent();
```
julienfalque: This variable is unnecessary as it is used only once.

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,126 @@
...
+$foo
+->bar1()
+->bar2();',
+            ),
+            array(
+                '<?php $foo
+    ->bar();',
+                '<?php $foo
+->bar();',
+            ),
```
julienfalque: Please add a case with method chaining without whitespaces, e.g. `$foo->bar()->baz()->qux()`.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private function getRightIndents($index, Tokens $tokens)
+    {
+        $indent = $this->whitespacesConfig->getIndent();
+
+        for ($i = $index; $i >= 0; --$i) {
+            $currentWhitespaces = $this->isLineBreak($tokens[$i]);
+
+            if (false !== $currentWhitespaces) {
+                if ($tokens[$i + 1]->equals(array(T_OBJECT_OPERATOR))) {
```
julienfalque: I would replace `equals()` with `isGivenKind()`.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,138 @@
...
+     *
+     * @return string|bool
+     */
+    private function getLineBreak(Token $token)
+    {
+        if (preg_match('/\R(\s*)/', $token->getContent(), $matches)) {
+            return $matches[1];
+        }
+
+        return false;
```
julienfalque: I would use `null` here.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,138 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * @param int    $index
+     * @param Tokens $tokens
+     *
+     * @return string|bool
```
julienfalque: Unless I'm missing something, this method returns `string` values only.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+
+        for ($index = 1; $index < count($tokens); ++$index) {
```
SpacePossum: please don't `count` each iteration;
```
for ($index = 1, $count = count($tokens);  $index < $count; ++$index) {
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,138 @@
...
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * @param Token $token
+     *
+     * @return string|bool
```
julienfalque: `string|null`

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -110,7 +110,7 @@ private function needLineBreak($index, Tokens $tokens)
         $isComment = false;
 
         for ($i = $index; $i > $prevMeaningful; --$i) {
-            if ($tokens[$i]->isGivenKind(T_OBJECT_OPERATOR) || false !== $this->getLineBreak($tokens[$i])) {
+            if ($tokens[$i]->isGivenKind(T_OBJECT_OPERATOR) || !is_null($this->getLineBreak($tokens[$i]))) {
```
keradus: `!is_null` => `null !==`

keradus:
could you add this rule to local `.php_cs.dist` to show it works nicely ?

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,133 @@
...
+namespace PhpCsFixer\Tests\Fixer\Whitespace;
+
+use PhpCsFixer\Test\AbstractFixerTestCase;
+use PhpCsFixer\WhitespacesFixerConfig;
+
+/**
+ * @author Vladimir Boliev <voff.web@gmail.com>
+ *
+ * @internal
+ * @coversNothing
```
julienfalque: Should be `@covers \PhpCsFixer\Fixer\Whitespace\MethodChainingIndentationFixer`.

README.rst
```diff
@@ -548,6 +548,10 @@ Choose from the list of available rules:
   - ``keep_multiple_spaces_after_comma`` (``bool``): whether keep multiple spaces
     after comma; defaults to ``false``
 
+* **method_chaining_indentation**
+
+  Method chaining MUST be properly indented.
```
keradus: could you add this rule to `.php_cs.dist` file to show that it works nicely and do not change properly intended code ?

.php_cs.dist
```diff
@@ -37,6 +37,7 @@ return PhpCsFixer\Config::create()
         'semicolon_after_instruction' => true,
         'strict_comparison' => true,
         'strict_param' => true,
+        'method_chaining_indentation' => true,
```
julienfalque: Please maintain this list sorted alphabetically: the rule should be inserted at line 26.

keradus:
I can see a lot of crashes in code of that new fixer :fearful:

keradus:
way better ! great you have figure out what was it that fast!
but that shows me I will need to make some regression tests for this fixer before merging it

boliev:
Hey @keradus 
Can I do anything else for this task?

keradus:
@julienfalque you have refused this PR, please review it again

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,173 @@
...
+    /**
+     * @param int    $start
+     * @param int    $end
+     * @param Tokens $tokens
+     *
+     * @return bool
+     */
+    private function isMultiLineMethod($start, $end, Tokens $tokens)
+    {
+        $tokenCleanContent = trim($tokens[$end]->getContent());
```
julienfalque: This variable is used only once, it can be passed directly where used.

.php_cs.dist
julienfalque: Your last commit moved this rule back to the end of the list. Please move it after `list_syntax` again.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,172 @@
...
+     * @param int    $start
+     * @param int    $end
+     * @param Tokens $tokens
+     *
+     * @return bool
+     */
+    private function isMultiLineMethod($start, $end, Tokens $tokens)
+    {
+        if (')' === trim($tokens[$end]->getContent())) {
+            if (CT::T_BRACE_CLASS_INSTANTIATION_CLOSE === $tokens[$end]->getId()) {
```
keradus: `getId` is not to compare types. use `isGivenKind`

keradus:
BTW, I don't think it would ever go to Symfony ruleset:
```diff
         $node
             ->children()
-                ->scalarNode('class')->isRequired()->cannotBeEmpty()->end()
-                ->scalarNode('property')->defaultNull()->end()
-                ->scalarNode('manager_name')->defaultNull()->end()
+            ->scalarNode('class')->isRequired()->cannotBeEmpty()->end()
+            ->scalarNode('property')->defaultNull()->end()
+            ->scalarNode('manager_name')->defaultNull()->end()
             ->end()
```

keradus:
```diff
$foo = $this->foo(array(
         1,
     ))
-        ->createView();
+    ->createView();
```

not sure is it a good change, any opinions here ? @julienfalque @SpacePossum

julienfalque:
> BTW, I don't think it would ever go to Symfony ruleset

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2617#issuecomment-287690551 ;)

I think the fixer can be merged as is, I plan to add a configuration option to support nested indentation levels later (unless someone else is willing to work on it as well).

> not sure is it a good change

It looks good to me, the issue is that lines 1 and 2 have too many indentation levels, but I think it's out of the scope of the fixer here.

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * @param int    $start
+     * @param int    $end
+     * @param Tokens $tokens
+     *
+     * @return bool
+     */
+    private function isMultiLineMethod($start, $end, Tokens $tokens)
+    {
+        if (')' === trim($tokens[$end]->getContent())) {
```
SpacePossum:
```php
$tokens[$end]->equals(')')
```

I think, as there cannot be tokens of which the content is `)` whit whitespace around it

gharlan:
> I plan to add a configuration option to support nested indentation levels later

What do you think about a configuration by method name patterns. This way we could **force** deeper indentation levels for some methods.
Something like:

```php
'method_chaining_indentation' => [
    'increment' => ['children', 'prototype', '/^.*Node$/'],
    'decrement' => ['end'],
],
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
boliev: I can't use equals here, because sometimes token is not an array
selection_007
But sometimes is 
selection_008
It is because of **BraceClassInstantiationTransformer**

julienfalque:
@gharlan Not sure it's worth it. Also, it would not work in all situations. For example, what if you use two different method chaining APIs that have methods with the same name but don't imply the same indentation levels?

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
SpacePossum: ah right, indeed!
please only remove the `trim` than :)

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * @param int    $start
+     * @param int    $end
+     * @param Tokens $tokens
+     *
+     * @return bool
+     */
+    private function isMultiLineMethod($start, $end, Tokens $tokens)
+    {
+        if (')' === $tokens[$end]->getContent()) {
```
keradus: `getContent` shall not be used for comparision

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /**
+     * @param string $content
+     *
+     * @return string|null
+     */
+    private function getLineBreak($content)
+    {
+        if (preg_match('/\R{1}([ \t]*)$/', $content, $matches)) {
```
keradus: `1 ==` would be nice, to have real boolean value (especially since output is `false|int`)

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
boliev: @keradus So, what the right way to do that?
I can't use equals here, because sometimes token is not an array
selection_007
But sometimes is 
selection_008
It is because of **BraceClassInstantiationTransformer**

julienfalque: You can use `equalsAny()`:
```php
if ($tokens[$end]->equalsAny([')', [CT::T_BRACE_CLASS_INSTANTIATION_CLOSE]])) {
    // ...
}
```

keradus: wait wait wait wait wait
you are handling `T_BRACE_CLASS_INSTANTIATION_CLOSE` here ? why ??

boliev: Testcase
```
<?php
return new Foo("param", [
    (new Bar("param1", "param2"))
        ->Foo([
            (new Bar())->foo(),
        ])
    ]);
```
The parentheses before `new Bar("param1", "param2")` is `T_BRACE_CLASS_INSTANTIATION_OPEN `
The parentheses after `new Bar("param1", "param2")` is `T_BRACE_CLASS_INSTANTIATION_CLOSE `

keradus:
As you granted us rights to modify your branch, I let myself fix deprecations you faced after merging with master branch

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
```diff
@@ -0,0 +1,192 @@
...
+            return $files;
+        }
+
+        $finder = Finder::create()
+            ->files()
+        ;',
+            ],
+            [
+                '<?php
+        $finder = Finder::create()
```
keradus: we need example with `::` (not `->`) that will also got fixed

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,174 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param string $content
+     *
+     * @return string|null
+     */
+    private function getLineBreak($content)
```
keradus: this is returning just spaces. spaces are not line breakers

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
keradus: example of `var_dump($tokens->generatePartialCode($start, $end));` that goes into this method:


```
string(42) "
    $user->setEmail('voff.web@gmail.com')"
```
```
string(42) "
    $user->setEmail('voff.web@gmail.com')"
```
```
string(10) "<?php
$foo"
```
```
string(10) "<?php
$foo"
```
```
string(10) "<?php
$foo"
```
```
string(10) "<?php
$foo"
```
```
string(8) "



$foo"
```
```
string(36) "

        $finder = Finder::create()"
```
```
string(34) "        $finder = Finder::create()"
```
```
string(37) "        $replacements = $replacements"
```
```
string(15) "
            })"
```
```
string(20) "        return foo()"
```
```
string(14) "
            )"
```


you are calling `isMultiLineMethod` method for input that is not method at all ...

keradus:
ping @boliev

boliev:
@keradus Of course you can modify the branch. Do I need to change something in the PR settings?

keradus:
no, as you can see:
> I let myself fix deprecations you faced after merging with master branch

already 5 days ago.

nevertheless, newly raised comments and failing Travis are waiting for you

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
boliev: Yes, and if a code is not method then it returns false. It check if code is method and it's multiline.
isMultiLineMethod returns true for code like 
```
foo(
    $bar1,
    $bar2
)
```

tests/Fixer/Whitespace/MethodChainingIndentationFixerTest.php
boliev: Don't understand this one

keradus: in utests there are cases when
```php
$foo
             ->bar()
```
got fixed, but no case like:


```php
$foo
             ::bar()
```

src/Fixer/Whitespace/MethodChainingIndentationFixer.php
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * @param int    $start
+     * @param int    $end
+     * @param Tokens $tokens
+     *
+     * @return bool
+     */
+    private function isMultiLineMethod($start, $end, Tokens $tokens)
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2617#discussion_r122853242

it is not checking is it a method at all.
it's checking is there some parenthesis matching (even outside of `start-end` scope!), which doesn't mean it's a method. what about `array(       )` ?

it's not checking is it multiline at all

keradus:
ping @boliev

keradus:
ping @boliev

boliev:
@keradus 
I think you can reject this PR, because I have no idea what to reply: 'I have no idea what is this method checking. And for that, it's hard to maintain it.'

keradus:
explaining the concerns just before the statement you qoute would be nice start

keradus:
Closed by #3063

PR #2618, opened 2017-03-20
Update README.rst

mhitza:
Composer global install sets executables under `~/.config/composer/vendor/bin` instead of `~/.composer/vendor/bin`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -56,14 +56,13 @@ Then, just run ``php-cs-fixer``.
 Globally (Composer)
 ~~~~~~~~~~~~~~~~~~~
 
-To install PHP CS Fixer, install Composer and issue the following command:
+To install PHP CS Fixer, `install Composer <https://getcomposer.org/download/>`_ and issue the following command:
 
 .. code-block:: bash
 
-    $ ./composer.phar global require friendsofphp/php-cs-fixer
+    $ composer global require friendsofphp/php-cs-fixer
 
-Then make sure you have ``~/.composer/vendor/bin`` in your ``PATH`` and
-you're good to go:
+Then make sure you have the global Composer binaries directory in your ``PATH``. This directory is platform-dependent, see `Composer documentation <https://getcomposer.org/doc/03-cli.md#composer-home>`_ for details. Example for some Unix systems:
 
 .. code-block:: bash
 
--- a/src/Console/Command/ReadmeCommand.php
+++ b/src/Console/Command/ReadmeCommand.php
@@ -99,14 +99,13 @@ Then, just run ``php-cs-fixer``.
 Globally (Composer)
 ~~~~~~~~~~~~~~~~~~~
 
-To install PHP CS Fixer, install Composer and issue the following command:
+To install PHP CS Fixer, `install Composer <https://getcomposer.org/download/>`_ and issue the following command:
 
 .. code-block:: bash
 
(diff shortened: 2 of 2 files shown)


SpacePossum:
Hi!
Thanks for the PR, looks like a good improvement. Please note that the README file is generated. Therefore please update the [readme command](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.1/src/Console/Command/ReadmeCommand.php#L113).

README.rst
```diff
@@ -67,7 +67,7 @@ you're good to go:
 
 .. code-block:: bash
 
-    $ export PATH="$PATH:$HOME/.composer/vendor/bin"
+    $ export PATH="$PATH:$HOME/.config/composer/vendor/bin"
```
keradus: original path is correct on my ubuntu

julienfalque: According to [Composer doc](https://getcomposer.org/doc/03-cli.md#composer-home), the default config directory varies depending on the Unix system you are using and both values are valid actually. IMO it would be better to say this is just an example and add this doc link to the README.

Anyway, if we update the `code-block`, we should also update the sentence above as it mentions the path too.

SpacePossum: if both are correct, than I think we don't need to change it

mhitza:
Should I change the commit with @julienfalque 's suggestion or just close the PR as @SpacePossum suggested?

SpacePossum:
I would be happy with a rephrasing of this instruction.
Anyone want to make a suggestion?

julienfalque:
What about:

```
To install PHP-CS-Fixer, `install Composer <https://getcomposer.org/download/>`_ and issue the following command:

.. code-block:: bash

    $ composer global require friendsofphp/php-cs-fixer

Then make sure you have the global Composer binaries directory in your ``PATH``. This directory is platform-dependent, see `Composer documentation <https://getcomposer.org/doc/03-cli.md#composer-home>`_ for details. Example for some Unix systems:

.. code-block:: bash

    $ export PATH="$PATH:$HOME/composer/vendor/bin"
```

keradus:
ping @mhitza, could you follow @julienfalque suggestion ?

mhitza:
@keradus updated the Readme command with @julienfalque 's suggestion

src/Console/Command/ReadmeCommand.php
```diff
@@ -99,14 +99,13 @@ protected function execute(InputInterface $input, OutputInterface $output)
 Globally (Composer)
 ~~~~~~~~~~~~~~~~~~~
 
-To install PHP CS Fixer, install Composer and issue the following command:
+To install PHP-CS-Fixer, `install Composer <https://getcomposer.org/download/>`_ and issue the following command:
```
keradus: name of the tool is `PHP CS Fixer`
name of bin file or composer package is `php-cs-fixer`.

`PHP-CS-Fixer` is some weird combination of those.

mhitza: Removed the dashes from the name.

julienfalque: Wrong suggestion here, my bad :)

julienfalque:
@mhitza You still need to update the README file itself:

```
$ php php-cs-fixer readme > README.rst
```

mhitza:
@julienfalque regenerated README.rst with the command you gave me

keradus:
Thank you @mhitza.

PR #2619, opened 2017-03-20
PSR-5 @inheritDoc support

julienfalque:
This PR adds support for [PSR-5 explicit inheritance tag](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#61-making-inheritance-explicit-using-the-inheritdoc-tag): the tag can be a regular tag instead of an inline one, and uses a capital "D".


SpacePossum:
:+1: for the idea, please fix the tests :)

README.rst
```diff
@@ -713,6 +713,12 @@ Choose from the list of available rules:
 
   Fix phpdoc inline tags, make inheritdoc always inline.
 
+  Configuration options:
+
+  - ``psr5`` (``bool``): whether the fixer should be PSR-5 compliant (uses
```
keradus: option name is too generic

julienfalque: Mhh... `psr5_inheritdoc`?

SpacePossum: hmm I wonder, IIRC the only reason why we always put the tag with braces ATM is because SF needed that as a work around. Maybe worth checking if this is still the case.

keradus: work around for what ?

julienfalque: Wasn't `@inheritdoc` an inline tag only before PSR-5 anyway?

GrahamCampbell: PSR-5 specifically is not a code standard spec (it explicitly says this in the spec), so defining some code style options as being "psr5" perhaps doesn't make sense.

SpacePossum: >work around for what ?

can't recall, IIRC @stof told me one time, but I cannot find his comment back :|

stof: AFAIK, https://github.com/FriendsOfPHP/Sami does not support ``@inheritdoc`` but only the inline tag, because PSR-5 was not yet drafted at this time (and phpdocumentor had not yet implemented a non-inline tag for that, as they added it later to support the Symfony use-case in a cleaner way according to their phpdoc semantic).

SpacePossum:
maybe it would be better to have a fixer for the casing of PHPDoc annotations and one fixer (this one) to deal with inline tags?

julienfalque:
It would be better, but I think that would still require an option in `phpdoc_inline_tag` to tweak its behavior, so maybe not worth it? Maybe better to rename the current fixer instead?

SpacePossum:
After the split up you would only need the configuration option here to (not) always make the inheritdoc inline. Like to hear more opinions!

keradus:
What about:

1. new fixer `GeneralPhpdocTagRenameFixer`
Fixer that will rename inline-tag or annotation-tag based on configuration:
```php
[
    'fix_annotation' => boolean,
    'fix_inline' => boolean,
    'ignore_casing' => boolean,
    'replacements' => [ 'inheritdocS' => 'inheritdoc', ...]
]
```

2. new fixer `PhpdocTagCasingFixer`
that will be proxy fixer to `GeneralPhpdocTagRenameFixer` that will fix most common tags

3. new fixer `PhpdocInlineTagNormalizerFixer`
that will:
```diff
- @{{   tag description?      }}}}}}
+ {@tag description?}

- {{   @tag description?      }}}}}}
+ {@tag description?}
```

4. current `PhpdocInlineTagFixer`
- shall reuse fixers 1 and 3 without extra logic
- shall be deprecated


-----

That way, you could achieve initial goal in very generic way: have `@inheridDoc` annotation tag

SpacePossum:
sounds good to me :+1:

julienfalque:
Sounds like a good idea. Could you elaborate what the options would do exactly?

keradus:
> ```php
> [
>     'fix_annotation' => boolean,
>     'fix_inline' => boolean,
>     'ignore_casing' => boolean,
>     'replacements' => [ 'inheritdocS' => 'inheritdoc', ...]
> ]
> ```

`replacements` - providing a replacement map
`ignore_casing` - if true the replaced tag may be written with different casing than provided in `replacements`, eg `iNhErItDoCs` would also be modified
`fix_inline` - rename inline tags, like in `* Foo @{inheritdocs} because... `
`fix_annotation` - rename annotations, like in `* @inheritdocs because....`

keradus:
ping @julienfalque

julienfalque:
PR reworked from scratch.

README.rst
```diff
@@ -979,9 +993,20 @@ Choose from the list of available rules:
...
+  instead).
 
-  Fix phpdoc inline tags, make inheritdoc always inline.
+* **phpdoc_inline_tag_normalizer** [@Symfony]
+
+  Fixes PHPDoc inline tags.
+
+  Configuration options:
+
+  - ``inheritdoc_always_inline`` (``bool``): whether to force ``@inheritDoc`` tag to
```
julienfalque: I'm going to rename this option to `force_inheritdoc_type` and make it accept three values:
- `inline`: always make the `@inheritDoc` tag inline;
- `annotation`: always make the `@inheritDoc` tag an annotation;
- `none` (default): don't force a type for `@inheritDoc` tags (but still fix the inline ones).

WDYT?

README.rst
```diff
@@ -541,6 +541,20 @@ Choose from the list of available rules:
...
+
+  Configuration options:
+
+  - ``case_sensitive`` (``bool``): whether tags should be replaced only if they have
+    exact same casing; defaults to ``false``
+  - ``fix_annotation`` (``bool``): whether annotation tags should be fixed; defaults
+    to ``true``
+  - ``fix_inline`` (``bool``): whether inline tags should be fixed; defaults to ``true``
+  - ``replacements`` (``array``): a map of tags to replace; defaults to
+    ``['inheritDocs' => 'inheritDoc']``
```
keradus: `general_*` rule shall have empty default config (meaning they shall not do anything until configured) + must not be part of any ruleset.
`general_*` rules are sth to be configured by end user to match his case, when added to ruleset, they will get overwritten by usage directly added in config file. for that reason, ruleset always use dedicated proxy fixer pointing to configured `general_*` rule instance

src/Fixer/Phpdoc/PhpdocInlineTagNormalizerFixer.php
```diff
@@ -0,0 +1,107 @@
...
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            // Move `@` inside tag, for example @{tag} -> {@tag}, replace multiple curly brackets,
+            // remove spaces between '{' and '@', remove white space between end
+            // of text and closing bracket and between the tag and inline comment.
+            $content = preg_replace_callback(
+                '#(?:@{+|{+[ \t]*@)[ \t]*(example|id|internal|inheritdoc|link|source|toc|tutorial)s?([^}]*)(?:}+)#i',
```
keradus: maybe we shall support any tag here, not only this list ?

src/Fixer/Phpdoc/PhpdocInlineTagNormalizerFixer.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('inheritdoc_always_inline', 'Whether to force `@inheritDoc` tag to be inline.'))
```
keradus: this option is not tested

README.rst
keradus:
```php
[
    'tags' => [
        'inheritdoc' => 'inline' | 'annotation',
        'any_other_tag' => 'inline' | 'annotation',
    ],
]
```

maybe it shall be yet another fixer?
so fixer to manage given tags to be converted to inline or annotation,
and fixer to normalize inline tags ?

keradus:
This PR is touching 25 files, including already existing rules and few generic files. For that, it's hard to focus on end-change we want to achieve here. Having all changes in one commit is not helping as well.
I think we shall split this PR. As first step I see extracting ability for ProxyFixer to proxy multiple fixers at once. WDYT ?

README.rst
julienfalque: Default updated.

src/Fixer/Phpdoc/PhpdocInlineTagNormalizerFixer.php
julienfalque: I added a `tags` option.

src/Fixer/Phpdoc/PhpdocInlineTagNormalizerFixer.php
julienfalque: Test added.

SpacePossum:
@julienfalque can you split us the PR? I think a new fixer like the `PhpdocTagCasingFixer` can be reviewed on its own for example?

julienfalque:
I extracted support for multiple proxied fixers in `AbstractProxyFixer`: #3149.

keradus:
extracted PR got merged

julienfalque:
Rebased.

keradus:
I like the idea of @SpacePossum to split this PR down to ease with review and overall process. 
Do you think there is still some nice thing that could be extracted?

julienfalque:
Maybe, but I'd like to address https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2619#discussion_r140091476 first.

julienfalque:
This PR should be ready for a new review iteration. Not sure there is more that can be extracted into a separate PR though.

README.rst
keradus: wdyt @julienfalque ?

julienfalque: Actually I think I implemented your suggestion and just forgot to answer here.

GrahamCampbell:
What's the status of this PR?

julienfalque:
This PR still needs some review. It's not a priority to me anymore as I tend to remove `@inherticDocs` at all now but I'd still be happy to get it merged :)

SpacePossum:
Thank you @julienfalque.

PR #2620, opened 2017-03-20
Finder - Remove `*.twig` as default

SpacePossum:
Twig files do not (or should) not contain PHP code. Therefore it does not make sense to me why we should try to fix those by default.

### DIFF
--- a/src/Finder.php
+++ b/src/Finder.php
@@ -28,7 +28,6 @@ class Finder extends BaseFinder
             ->files()
             ->name('*.php')
             ->name('*.phpt')
-            ->name('*.twig')
             ->ignoreDotFiles(true)
             ->ignoreVCS(true)
             ->exclude('vendor')


keradus:
:-1:
BC breaker

GrahamCampbell:
👍 By definition, all bug fixes are breaking changes, and this is a bug fix.

keradus:
> Twig files do not (or should) not contain PHP code.

they should not or must not contain PHP code?
If `should not` - this is change of behaviour, thus BC breaker
If `must not` - this is fix

Is it possible by design to put PHP code into twig file?

SpacePossum:
>Is it possible by design to put PHP code into twig file?

no, Twig will not execute or process PHP code in side a Twig template.
This is by design, hiding PHP details for designers and such.
However, since Twig templates can contain any text you can put PHP code in there, but the only function for that would be to render that code, not to execute it.

keradus:
OK, then :+1:

btw, something wrong happen with hooks...

keradus:
I don't treat it as a bug for user, as he will not suffer any wrong behaviour without that change

keradus:
Thank you @SpacePossum.

PR #2621, opened 2017-03-20
Tokenizer - fix edge cases with empty code, registered found tokens and code hash

SpacePossum:
Currently both the `Tokens` class and some `Fixers` depend on behavior of the `FileFilterIterator` to filter out empty files and by that empty strings.
This breaks the possibility to use the `Tokens` and `Fixers` outside of this context.

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -135,15 +135,15 @@ class Tokens extends \SplFixedArray
             foreach ($array as $key => $val) {
                 $tokens[$key] = $val;
             }
+        } else {
+            $index = 0;
 
-            return $tokens;
+            foreach ($array as $val) {
+                $tokens[$index++] = $val;
+            }
         }
 
-        $index = 0;
-
-        foreach ($array as $val) {
-            $tokens[$index++] = $val;
-        }
+        $tokens->generateCode(); // regenerate code to calculate code hash
 
         return $tokens;
     }
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -69,7 +69,7 @@ class Tokens extends \SplFixedArray
      *
      * @var bool[]
      */
-    private $foundTokenKinds;
+    private $foundTokenKinds = array();
 
     /**
      * Clone tokens collection.
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -803,11 +803,12 @@ PHP;
     {
         $code = '<?php echo 1;';
 
-        $tokens = Tokens::fromCode($code);
-        $this->assertTrue($tokens->isTokenKindFound(T_OPEN_TAG));
+        $tokens1 = Tokens::fromCode($code);
+        $tokens2 = Tokens::fromArray($tokens1->toArray());
 
-        $tokens2 = Tokens::fromArray($tokens->toArray());
+        $this->assertTrue($tokens1->isTokenKindFound(T_OPEN_TAG));
         $this->assertTrue($tokens2->isTokenKindFound(T_OPEN_TAG));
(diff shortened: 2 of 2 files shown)


keradus:
That check should be enabled for all rules, not only that few.
I assume you have fixed part when it's failing without that extra check, but we should add them globally for performance as well.

I would propose a solution after initial discussion it #2614 :

globally:
```diff
-public function fix(...)
+protected function fixCollection(...)
```

then, in `AbstractFixer`:
```php
public function fix(...)
{
    if (0 < $tokens->count() && $this->isCandidate(...)) {
        $this->fix(...);
    }
}
```

then, also, we could drop other usages of `->isCandidate` in our source

SpacePossum:
That makes total sense. I've removed the `Fixers` leaving the `Tokens` update to handle empty string correctly.

keradus:
then, I applied mentioned solution in #2614

SpacePossum:
#2614 deals with `fix` being depended on `candidate` method, this PR fixes `Tokens::fromCode('')` case only.

keradus:
I know, that's why I sent that PR to fix what we agreed on already. I still need to investigate that `fromArray` case

src/Tokenizer/Tokens.php
```diff
@@ -156,19 +156,21 @@ public static function fromArray($array, $saveIndexes = null)
      */
     public static function fromCode($code)
     {
-        $codeHash = self::calculateCodeHash($code);
+        if ('' !== $code) {
```
keradus: this change is not tested anywhere.
----------

what is the user case - why we need this change?
```php
$a = Tokens::fromCode('');
$b = Tokens::fromCode('');
```

on master, `$a === $b`, on your branch, they don't. what is a benefit of this change? It's an exception of how the collection is reused.

also, if one wants to have a new, empty collection - there is perfect way to do it - `new Tokens()`

keradus:
Currently, this PR does not fix any bug during cloning.

Added test that prove cloning works is nice, adding `=array()` initialization for property (which is needed for new test) is nice, but I don't see any bug fix here.
3rd thing done here is change of handling empty string passed to `fromCode`, I commented it separately as code comment.

--------------------------------

> Btw.: `fromArray` does not run the transformers nor set the tokens types found like `fromCode` does, maybe we should make this the same or update the docs?

- if `fromArray` would accept result of `token_get_all`, it shall call transformers, but it accepts only `Token[]`, which shall be ready for use instances, so already after transforming
- during investigating your question I discovered that `fromArray` is wrongly not calling `$this->registerFoundToken($token);` nor `->changeCodeHash(...)`

src/Tokenizer/Tokens.php
SpacePossum: IIRC if you run the tokens with input code `''` and than set other other code things went wrong, I'll hunt this down

keradus:
ping @SpacePossum

keradus:
ping @SpacePossum

keradus:

SpacePossum:
This PR:
- proves bug when using `Tokens::fromCode('')` by utest
- fixes bug (see above)

Add tests for the following cases currently not covered:
- creating multiple Tokens by `Tokens::fromCode('')`
- Tokens cloning
- `Tokens::fromArray` combined with `isTokenKindFound`
- `Tokens::fromArray(array());`

This leave the case @keradus found:
>during investigating your question I discovered that fromArray is wrongly not calling [...] ->changeCodeHash(...)

I'm not sure if this should but part of this PR. It will take me some time to reproduce and fix that case. Any help is welcome on this.

keradus:
@SpacePossum , please check out last commit in this PR

SpacePossum:
thanks @keradus looks good! I think this is ready now :)

keradus:
Thank you @SpacePossum.

SpacePossum:
Thank you @keradus another bunch of rare cases fixed/covered :D

PR #2630, opened 2017-03-22
HeaderCommentFixer - Fix trailing whitespace not removed after <?php

julienfalque:


keradus:
Thank you @julienfalque.

PR #2633, opened 2017-03-23
Add snake to camel case fixer.

vmelnik-ukraine:
To avoid/fix snake case variables in the code base.

### DIFF
--- a/src/Fixer/Naming/SnakeCaseToCamelCaseFixer.php
+++ b/src/Fixer/Naming/SnakeCaseToCamelCaseFixer.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Naming;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Victor Melnik <melnikvictorl@gmail.com>
+ */
+final class SnakeCaseToCamelCaseFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Convert `$snake_case` variables to `$camelCase`.',
+            array(
+                new CodeSample(
+                    "<?php\n\$snake_case = '';"
+                ),
+            )
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_VARIABLE);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_VARIABLE)) {
+                continue;
+            }
+
+            if (preg_match('/_/', $token->getContent())) {
+                $token->setContent('$'.$this->camelize(substr($token->getContent(), 1)));
+            }
+        }
+    }
+
+    /**
+     * @param string $name variable name to camelize
+     *
+     * @return string
+     */
+    private function camelize($name)
+    {
+        return lcfirst(str_replace(' ', '', ucwords(str_replace('_', ' ', $name))));
+    }
+}
--- a/tests/Fixer/Naming/SnakeCaseToCamelCaseFixerTest.php
+++ b/tests/Fixer/Naming/SnakeCaseToCamelCaseFixerTest.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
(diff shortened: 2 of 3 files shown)


tests/Fixer/Naming/SnakeCaseToCamelCaseFixerTest.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function provideFixCases()
+    {
+        return array(
+            // should not fix
+            array('<?php $variable = 1;'),
+            // should fix
+            array(
+                '<?php $camelCase = 1;',
+                '<?php $camel_case = 1;',
```
keradus:
```diff
 <?php
-$camel_case = 1;
-$camelCase = 1;
```

this changes code behaviour.

------

this does not
```diff
 <?php
 function foo ()
 {
-    $camel_case = 1;
+    $camelCase = 1;
 }
```

keradus:
should be configurable:
- what: vars, properties, functions, methods, classes
- how: safe, all

like it's safe to change private methods/properties/non-global vars, but it's risky to change all, like global vars (test cases you have provided), protected methods, classes, ...

SpacePossum:
>like it's safe to change private methods/properties/non-global vars

sadly its not, you can still use the name in a dynamic context;

```php
public function callPrivateMethod($snakeNameCase = 'a_b_c') {
    $this->$snakeNameCase;
}

private function a_b_c()
{
}

```

julienfalque:
Changing the name of a private property/method can even break code that uses reflection.

src/Fixer/Naming/SnakeCaseToCamelCaseFixer.php
```diff
@@ -0,0 +1,73 @@
...
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_VARIABLE)) {
+                continue;
+            }
+
+            if (preg_match('/_/', $token->getContent())) {
```
julienfalque: Not sure but maybe `false !== strpos($token->getContent(), '_')` would be more efficient.

tests/Fixer/Naming/SnakeCaseToCamelCaseFixerTest.php
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @internal
+ */
+final class SnakeCaseToCamelCaseFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param string $expected
+     * @param null   $input
```
julienfalque: `null|string`

julienfalque:
We need more test cases: like `"$variable inside string"` for example.

keradus:
> Also, some people prefer to avoid consecutive uppercase letters, e.g.: `getOrmInstance()` instead of `getORMInstance()`, should the fixer support this case?

Proper `camelCase` transformation for `to_JSON` is `toJson`, as `toJSON` would mean `to_j_s_o_n`. Discussed dozen of times on devs communities

vmelnik-ukraine:
> Changing the name of a private property/method can also break code that uses reflection.

@julienfalque, seems like we'll have to mark the whole fixer as risky then.

> Like @keradus said, this fixer should also cover properties, methods, ... and be configurable.

@julienfalque, yes, I agree.

> What about supporting other casings like with constants, e.g. SOME_CONSTANT_VALUE instead of SomeConstantValue?

Shouldn't we have a dedicated fixer for that? I can't remember any other cases where I'd use this type of naming except with constants, so seems it's not directly related to snake vs. camel problem. Wdyt?

> Also, some people prefer to avoid consecutive uppercase letters, e.g.: getOrmInstance() instead of getORMInstance(), should the fixer support this case?

> Proper camelCase transformation for to_JSON is toJson, as toJSON would mean to_j_s_o_n. 

@keradus, @julienfalque if it's a common approach, then let's add a test for that

keradus:
as a POC, when you will be happy about proposed solution include it in `.php_cs.dist` file as well, then we will see how it works with quite a piece of code.

GrahamCampbell:
I'm 👎 on the idea. There are so many ways this can break code.

keradus:
That's why I asked to add it to local config file. If author will manage to propose a fixer that would not broke PHP CS Fixer codebase, it may be good enought

julienfalque:
> Shouldn't we have a dedicated fixer for that? I can't remember any other cases where I'd use this type of naming except with constants, so seems it's not directly related to snake vs. camel problem. Wdyt?

It's not a snake vs. camel case but it's still a naming style case that could be fixed. I won't be against extending the scope of the fixer here, maybe with configuration like:

```php
[
    'naming_styles' => [
        'constants' => 'uppercase_underscore',
        'variables' => 'camel_case',
        'functions' => 'snake_case',
        'methods' => 'camel_case',
        // ...
    ],
]
```

keradus:
`'classes' => 'PascalCase'`

GrahamCampbell:
Be careful with that. That may conflict with PSR0/4 fixers.

GrahamCampbell:
Infact, it should only be allowed to be run if those fixers are off.

keradus:
> That's why I asked to add it to local config file. If author will manage to propose a fixer that would not broke PHP CS Fixer codebase, it may be good enough

@vmelnik-ukraine, you have made nice POC for starting the work. We are aware that this fixer is not obvious and extra tricky. Are you taking the challenge to finish it making the state of fixer good enough to be run over codebase of PHP CS Fixer itself?

vmelnik-ukraine:
@keradus yes, I'd like to give it a try.

keradus:
Sadly, no changes here, thus I'm closing this PR.
Feel free to open it once more after you will try to apply the changes

PR #2636, opened 2017-03-23
Allow phpdoc for language constructs, fixes #2634

### LINKED ISSUE #2634: phpdoc_to_comment inconsistent treatment of PHPDoc tags
`phpdoc_to_comment` fixer does not consistently detect phpDocumenter comments. Consider the following two examples.

```php
<?php

class A
{
    public function b()
    {
        /** @var int $c */
        print($c = 0);
    }
}
```

This example will be changed as follows:

```diff
--- Original
+++ New
@@ @@
     {
-        /** @var int $c */
+        /* @var int $c */
         print($c = 0);
     }
 }
```

However, exactly the same code formatted in a slightly different way will not be changed by the fixer, as shown below.

```php
<?php

class A
{
    public function b()
    {
        
        print(
            /** @var int $c */
            $c = 0
        );
    }
}
```

I think this fixer should either:

1. Be changed to treat any occurrence of `@foo` (i.e. `\b@\w+\b`) as valid docblock content, or
2. Include a *simple validation* mode that behaves the way described in 1.

Currently this fixer seems to be trying to be "too clever" to determine if a comment comprises valid docblock content which is causing more harm than good in many usage scenarios.

ceeram:


SpacePossum:
if `list` is supported, shouldn't its shorthand counter part on 7.1 be as well? For example:

```php
/** @var int $a */
[$a] = $array;
```

ceeram:
Possibly

Op 24 mrt. 2017 08:19 schreef "SpacePossum" <notifications@github.com>:

keradus:
ping @ceeram

ceeram:
whats up?

keradus:
could you handle case provided by SpacePossum please ?

ceeram:
I can do that. Not sure when I get to doing so.

Op 30 mrt. 2017 22:56 schreef "Dariusz Rumiński" <notifications@github.com>:

could you handle case provided by SpacePossum please ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2636#issuecomment-290542511>,
or mute the thread

keradus:
Closing as picked up by SpacePossum in #2657
cc @ceeram

PR #2637, opened 2017-03-24
ToolInfo - use static dir check for composer discovery

Slamdunk:
In a Composer installation `php-cs-fixer` binary is present in
* `vendor/friendsofphp/php-cs-fixer/php-cs-fixer`
* `vendor/bin/php-cs-fixer` (symlink)

but someone may want to re-symlink the symlink in a custom/private bin dir.
If so, the script currently fails to detect whether is installed through composer or not.

This is due to the `$_SERVER['SCRIPT_NAME']` variable, that could vary a lot.
But composer installations can have only 2 result path, so just check them statically.

### DIFF
--- a/src/ToolInfo.php
+++ b/src/ToolInfo.php
@@ -23,7 +23,6 @@ use PhpCsFixer\Console\Application;
  */
 final class ToolInfo
 {
-    const COMPOSER_INSTALLED_FILE = '/../../composer/installed.json';
     const COMPOSER_PACKAGE_NAME = 'friendsofphp/php-cs-fixer';
 
     public static function getComposerVersion()
@@ -35,7 +34,7 @@ final class ToolInfo
         }
 
         if (null === $result) {
-            $composerInstalled = json_decode(file_get_contents(self::getScriptDir().self::COMPOSER_INSTALLED_FILE), true);
+            $composerInstalled = json_decode(file_get_contents(self::getComposerInstalledFile()), true);
 
             foreach ($composerInstalled as $package) {
                 if (self::COMPOSER_PACKAGE_NAME === $package['name']) {
@@ -73,33 +72,14 @@ final class ToolInfo
         static $result;
 
         if (null === $result) {
-            $result = !self::isInstalledAsPhar() && file_exists(self::getScriptDir().self::COMPOSER_INSTALLED_FILE);
+            $result = !self::isInstalledAsPhar() && file_exists(self::getComposerInstalledFile());
         }
 
         return $result;
     }
 
-    private static function getScriptDir()
+    private static function getComposerInstalledFile()
     {
-        static $result;
-
-        if (null === $result) {
-            $script = $_SERVER['SCRIPT_NAME'];
-
-            if (is_link($script)) {
-                $linkTarget = readlink($script);
-
-                // If the link target is relative to the link
-                if (false === realpath($linkTarget)) {
-                    $linkTarget = dirname($script).'/'.$linkTarget;
-                }
-
-                $script = $linkTarget;
-            }
-
-            $result = dirname($script);
-        }
-
-        return $result;
+        return __DIR__.'/../../../composer/installed.json';
     }
 }
--- a/tests/ToolInfoTest.php
+++ b/tests/ToolInfoTest.php
@@ -30,4 +30,16 @@ final class ToolInfoTest extends \PHPUnit_Framework_TestCase
     {
         $this->assertFalse(ToolInfo::isInstalledAsPhar());
     }
+
+    public function testIsInstalledByComposer()
+    {
+        $this->assertFalse(ToolInfo::isInstalledByComposer());
+    }
+
+    public function testGetComposerVersionThrowsExceptionIfOutsideComposerScope()
+    {
+        $this->setExpectedException('LogicException');
+
+        ToolInfo::getComposerVersion();
+    }
 }


tests/ToolInfoTest.php
```diff
@@ -23,11 +23,61 @@
...
+     */
+    public function testIsInstalledByComposer($scriptName)
+    {
+        $phpunitBin = $this->getRootDir().'/vendor/bin/phpunit';
+        if (!is_file($phpunitBin)) {
+            $this->markTestSkipped(sprintf('Binary "%s" is necessary to emulate PHP-CS-Fixer installed through composer', $phpunitBin));
+        }
+
+        if (!is_link($scriptName)) {
+            $this->markTestSkipped(sprintf('OS "%s" doesn\'t support the way the "%s" symlink is written', PHP_OS, $scriptName));
```
Slamdunk: I don't like very much the need to skip Windows, but neither I like to create symlink at runtime.
As of yet the two symlink are created under `tests/Fixtures` and committed.

src/ToolInfo.php
```diff
@@ -26,67 +26,89 @@
...
+     * @var null|bool
+     */
+    private $isInstalledByComposer;
+
+    /**
+     * @var null|string
+     */
+    private $scriptDir;
+
+    public function __construct($script = null)
```
keradus: please add typehinting as `?string`, no need for default value IMO

tests/ToolInfoTest.php
```diff
@@ -23,11 +23,61 @@
...
     {
-        $this->assertFalse(ToolInfo::isInstalledAsPhar());
+        $toolInfo = new ToolInfo();
+        $this->assertFalse($toolInfo->isInstalledAsPhar());
+    }
+
+    /**
+     * @dataProvider provideScriptName
+     *
+     * @param mixed $scriptName
```
keradus: please don't use `mixed`

keradus:
I like the idea.
Anyone willing to review this PR ?

src/ToolInfo.php
Slamdunk: So the code that uses `ToolInfo` should specify everytime `$_SERVER['SCRIPT_NAME']`?
Morover, it's a bit early to use 7.1-only feature, isn't it?

keradus: with `?string`, one could still use `new ToolInfo(null)`, which would means exactly the same as `new ToolInfo()` with null in constructor prototype.

let me rephrase, please put typehinting for parameter in docblock, so one won't provide here an instance of `SplFile` or sth...

src/ToolInfo.php
```diff
@@ -26,67 +26,92 @@
...
+    /**
+     * @var null|string
+     */
+    private $scriptDir;
+
+    /**
+     * @param ?string $script
+     */
+    public function __construct($script = null)
     {
```
Slamdunk: @keradus is this ok?

keradus: oh, ok, let it be. I hoped for non-optional parameter

tests/ToolInfoTest.php
keradus: argh, I totally cannot work with two created symlink files on my OS (Win10)

src/ToolInfo.php
Slamdunk: I prefer non-optional parameter too, but only in a well-organized dependency injection context.
As of yet, ToolInfo needs to be instantiade where it is used.

keradus: still, one could make it like `new ToolInfo($path)` or `new ToolInfo(null)`, right ?

src/ToolInfo.php
```diff
@@ -26,67 +26,92 @@
...
+    /**
+     * @var null|string
+     */
+    private $scriptDir;
+
+    /**
+     * @param string $script
+     */
+    public function __construct($script)
     {
```
Slamdunk: The perfect final form of this constructor, imho, would it to be `$this->script = $script;` and somewhere else a factory that injects the real script name, and factories for classes that use it that inject `ToolInfo` through its factory.
In the meantime...

src/ToolInfo.php
```diff
@@ -26,67 +26,92 @@
...
+     */
+    private $isInstalledByComposer;
+
+    /**
+     * @var null|string
+     */
+    private $scriptDir;
+
+    /**
+     * @param string $script
```
keradus: from the code I see it's nullable

src/ToolInfo.php
keradus: :+1: but https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2637/files#r108142119

src/ToolInfo.php
Slamdunk: I'm not used to phpdoc: `?string` or `null|string`?

SpacePossum: the later
:+1:

keradus: :+1: for earlier
both are acceptable

keradus:
@julienfalque , could you make a review please?

tests/ToolInfoTest.php
```diff
@@ -23,11 +23,61 @@
...
+        return array(
+            array($this->getRootDir().'/vendor/bin/phpunit'),
+            array(__DIR__.'/Fixtures/toolinfo-phpunit-symlink'),
+            array(__DIR__.'/Fixtures/toolinfo-phpunit-double-symlink'),
+        );
+    }
+
+    public function testGetComposerVersionThrowExceptionIfOutsideComposerScope()
+    {
+        $toolInfo = new ToolInfo(uniqid('non_existent_script_'));
```
julienfalque: `uniqid()` does not seem to be necessary here.

tests/ToolInfoTest.php
SpacePossum: can we create and remove the symlink(s) in the setup/teardown of the tests or would that complicate the tests to much?

src/ToolInfo.php
```diff
@@ -26,67 +26,92 @@
...
+     */
+    private $scriptDir;
+
+    /**
+     * @param null|string $script
+     */
+    public function __construct($script)
     {
-        static $result;
+        $this->script = (null !== $script ? $script : $_SERVER['SCRIPT_NAME']);
```
Slamdunk: IMHO this:
```php
$this->script = null !== $script ? $script : $_SERVER['SCRIPT_NAME'];
```
Is unreadable, [operator precedence](http://php.net/manual/en/language.operators.precedence.php) can be a nightmare without parenthesis.

Slamdunk:
> Could you also write a test for isInstalledByComposer() returning false when not installed through Composer (if possible)?

Not feasable, https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.1.2/src/ToolInfo.php#L76 will always be true in the test environment and can't be mocked without writing production code just for the purpose of mocking.

julienfalque:
Unless I'm missing something, I think it can work with `new ToolInfo('script_outside_vendor_dir')` like you did in the last test.

src/ToolInfo.php
julienfalque: It's pretty obvious to me that ternary operator will be evaluated before the assignment as the opposite would almost defeat the purpose of the ternary operator, but I agree parentheses still help for readability. What do others think?

tests/ToolInfoTest.php
```diff
@@ -21,13 +21,87 @@
  */
 final class ToolInfoTest extends \PHPUnit_Framework_TestCase
 {
+    const PHPUNIT_BIN = '/../vendor/bin/phpunit';
+
+    protected function setUp()
+    {
+        foreach ($this->provideScriptName() as $testArguments) {
+            $symlink = current($testArguments);
+            if ($symlink === self::PHPUNIT_BIN or is_link(__DIR__.$symlink)) {
```
keradus: oh damn, `or`...

Slamdunk: Will change. As of yet I'm trying to figure out how to run all on Windows.

OT: I cannot understand the love of the community for `||` and `&&`, code should be readable as fast as possible. Words will always be faster than symbols.

tests/Fixtures/symlinks/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
```
julienfalque: I would add the directory to the root .gitignore file instead, with a comment to explain its purpose.

tests/ToolInfoTest.php
```diff
@@ -21,13 +21,87 @@
  */
 final class ToolInfoTest extends \PHPUnit_Framework_TestCase
 {
+    const PHPUNIT_BIN = '/../vendor/bin/phpunit';
+
+    protected function setUp()
```
julienfalque: I'm not sure using `setUp`/`tearDown` is relevant: the created symlinks are only used in `testIsInstalledByComposer`, maybe it would be easier to handle them in the test directly?

tests/ToolInfoTest.php
```diff
@@ -21,13 +21,87 @@
...
+            array('/Fixtures/symlinks/toolinfo-phpunit-symlink'),
+            array('/Fixtures/symlinks/toolinfo-phpunit-double-symlink'),
+        );
+    }
+
+    public function testGetComposerVersionThrowExceptionIfOutsideComposerScope()
+    {
+        $toolInfo = new ToolInfo('non_existent_script');
+
+        $this->assertFalse($toolInfo->isInstalledByComposer());
```
julienfalque: This assertion should be moved to a new test.

tests/ToolInfoTest.php
keradus: For the sake of homogeneity and safety.
First, as the project currently never uses `or`.
Second, as most devs cannot properly understand that difference - even if you would, then new contributor probably wont when he will be touching this code.

Slamdunk: Never intented to troll or change the status quo.
For sure consistency and homogeneity come first: will change in next commits.
My OT was more generic, outside this project.

tests/ToolInfoTest.php
Slamdunk: If an assertion fails, the test stop, but `tearDown` is still called

tests/ToolInfoTest.php
keradus: I would rename the method to `testIsInstalledAsPhar` and that's it

julienfalque: This method already exists, and it would not follow PHPUnit conventions IMO.

Slamdunk:
Ok, tests now run both on Windows and Linux.
The main difference is that composer creates binaries under `vendor/bin` in different ways: on Linux is just a symlink, and on Windows it's a special executable, not a symlink, so we need to uniform the method to find `composer.json` according to this distinction.

src/ToolInfo.php
```diff
@@ -23,83 +23,114 @@
...
-        if (null === $result) {
-            $script = $_SERVER['SCRIPT_NAME'];
+        if (null === $this->composerInstalledFile) {
+            $script = $this->script;
+
+            while (is_link($script)) {
+                // On Windows, composer creates under vendor/bin executables that
+                // aren't symlinks, so we need to manually stop at a common point
+                // equal from Windows to Linux
+                if ('vendor'.DIRECTORY_SEPARATOR.'bin' === substr(dirname($script), -10)) {
```
julienfalque: AFAIK, `DIRECTORY_SEPARATOR` isn't required as paths using `/` work on Windows with PHP.

Also, this will fail when Composer is configured to install dependencies in a different directory than `vendor`.

keradus: it's required as you compare it to result of dirname, which return data with OS respecting dir separator.

tests/ToolInfoTest.php
keradus: then what about moving this one assertion to that existing method?

tests/ToolInfoTest.php
```diff
@@ -21,13 +21,90 @@
...
+        $this->setExpectedException('LogicException');
+
+        $toolInfo->getComposerVersion();
+    }
+
+    private static function getPhpunitBin()
+    {
+        static $phpunitBin;
+        if (null === $phpunitBin) {
+            $phpunitBin = DIRECTORY_SEPARATOR.'..'.DIRECTORY_SEPARATOR.'vendor'.DIRECTORY_SEPARATOR.'bin'.DIRECTORY_SEPARATOR.'phpunit';
```
keradus: no need for cache in global state, this is just concatenation

src/ToolInfo.php
Slamdunk: The custom `bin-dir` is a good point, but:
1. if you are under Windows
1. and you use a custom `bin-dir`
there is no way to properly detect a composer installation since bins created by composer on Windows are not symlinks.

A step behide: is this test really necessary on Windows?
On Linux I usually create symlink in my `$PATH`, but on Windows linking is hard and users/softwares prefer to add another path to `$PATH` AFAIK.

Soooo, what should I do?

tests/ToolInfoTest.php
Slamdunk: `testIsInstalledAsPhar` is a bit misleading: currently it tests that it is NOT installed by Phar, because `'phar://' === substr(__DIR__, 0, 7)` is not mockable.

The test would look like:
```php
public function testIsInstallAsPhar()
{
    $toolInfo = new ToolInfo(null);
    $this->assertFalse($toolInfo->isInstalledAsPhar());
    $this->assertFalse($toolInfo->isInstalledByComposer());
}
```

src/ToolInfo.php
julienfalque: The test works now so there is no reason to remove it :)

For the custom `bin-dir`, maybe we can check if the filename ends with `.bat` instead?

src/ToolInfo.php
```diff
@@ -23,83 +23,114 @@
...
 
-            $result = dirname($script);
+            $composerInstalledJson = DIRECTORY_SEPARATOR.'composer'.DIRECTORY_SEPARATOR.'installed.json';
+            // Windows real file is under `vendor/bin` because composer
+            // doesn't use symlinks on that OS
+            // On Linux instead the executabile is under `vendor/maintainer/package`
+            $this->composerInstalledFile = dirname(dirname($script)).$composerInstalledJson;
+            if (!file_exists($this->composerInstalledFile)) {
+                $this->composerInstalledFile = dirname(dirname(dirname($script))).$composerInstalledJson;
+            }
```
Slamdunk: @julienfalque [bin-dir](https://getcomposer.org/doc/06-config.md#bin-dir) could be in any directory. The only way to know the exact position of `vendor` (and therefore installed.json) would be to read it, but this would lead to a strict coupling to composer package and its implementations.

Let me bring your attention to the fact that all thing code is here only to get the composer version (and it is installed through composer for cache).

What about removing everything and use https://github.com/Ocramius/PackageVersions ? @keradus ?

julienfalque: Looks like a much better solution 👍

Slamdunk: https://github.com/Ocramius/PackageVersions requires `~7.0` 😞

Slamdunk: Ok, like Ocramius/PackageVersions I would to for `__DIR__` in replacement of `$_SERVER['SCRIPT_NAME']`, commits soon

julienfalque:
Is it really important to rely on Composer to detect PHP-CS-Fixer version? Wouldn't it be easier to put it in a constant?

Slamdunk:
Yes, if we want to expose `dev-master` commit

Slamdunk:
What about now?

julienfalque:
Now code is more simple but Composer detection will always fail for custom `bin-dir`.

Slamdunk:
Why? The code now only looks for `ToolInfo.php` parents, no matter where you want your binaries.

src/ToolInfo.php
```diff
@@ -73,31 +72,31 @@ public static function isInstalledByComposer()
         static $result;
 
         if (null === $result) {
-            $result = !self::isInstalledAsPhar() && file_exists(self::getScriptDir().self::COMPOSER_INSTALLED_FILE);
+            $result = !self::isInstalledAsPhar() && file_exists(self::getComposerInstalledFile());
```
julienfalque: `file_exists()` can be replaced with `null !==` as the file is already checked in `getComposerInstalledFile()`.

src/ToolInfo.php
```diff
@@ -73,31 +72,31 @@ public static function isInstalledByComposer()
...
-            if (is_link($script)) {
-                $linkTarget = readlink($script);
-
-                // If the link target is relative to the link
-                if (false === realpath($linkTarget)) {
-                    $linkTarget = dirname($script).'/'.$linkTarget;
+            $composerInstalledFile = '/composer/installed.json';
+            $checkedPaths = array(
+                // this project itself at ./src
+                __DIR__.'/../vendor'.$composerInstalledFile,
```
julienfalque: Not sure we should check this one actually: the project does not require itself via Composer so I think the version won't be available in the json file.

src/ToolInfo.php
```diff
@@ -73,31 +72,31 @@ public static function isInstalledByComposer()
...
+                // the project installed through composer at ./vendor/friendsofphp/php-cs-fixer/src
+                __DIR__.'/../../..'.$composerInstalledFile,
+            );
+
+            foreach ($checkedPaths as $path) {
+                $result = $path;
```
julienfalque: This should be done in the `if`, otherwise the last checked file will be returned even if it doesn't exist.

src/ToolInfo.php
Slamdunk: What about now?

julienfalque:
👍

src/ToolInfo.php
```diff
@@ -73,33 +72,14 @@ public static function isInstalledByComposer()
...
-                }
-
-                $script = $linkTarget;
-            }
-
-            $result = dirname($script);
-        }
-
-        return $result;
+        return __DIR__.'/../../../composer/installed.json';
```
keradus: I got lost. You check only this path, but from your initial comment:
> But composer installations can have only 2 result path, so just check them statically.

julienfalque: See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2637#discussion_r108693134

keradus: thanks, makes sense

keradus:
Thank you @Slamdunk.

keradus:
info: merged into 2.1 instead of 2.2

Issue #2570, opened 2017-02-28
Fixer for magic constants casing


jtojnar:
Similarly to `NativeFunctionCasingFixer`, there should be a fixer for [magic constants](https://secure.php.net/manual/en/language.constants.predefined.php) like `__DIR__`.

ntzm:
I'm having a go at this

PR #2642, opened 2017-03-24
Add MagicConstantCasingFixer

ntzm:
Implements #2570.

Please let me know if I've done anything wrong.

### DIFF
--- a/src/Fixer/Casing/MagicConstantCasingFixer.php
+++ b/src/Fixer/Casing/MagicConstantCasingFixer.php
@@ -26,21 +26,6 @@ final class MagicConstantCasingFixer extends AbstractFixer
     /**
      * {@inheritdoc}
      */
-    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
-    {
-        $magicConstants = $this->getMagicConstants();
-        $magicConstantTokens = $this->getMagicConstantTokens();
-
-        foreach ($tokens as $token) {
-            if ($token->isGivenKind($magicConstantTokens)) {
-                $token->setContent($magicConstants[$token->getId()]);
-            }
-        }
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function getDefinition()
     {
         return new FixerDefinition(
@@ -55,6 +40,21 @@ final class MagicConstantCasingFixer extends AbstractFixer
     public function isCandidate(Tokens $tokens)
     {
         return $tokens->isAnyTokenKindsFound($this->getMagicConstantTokens());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $magicConstants = $this->getMagicConstants();
+        $magicConstantTokens = $this->getMagicConstantTokens();
+
+        foreach ($tokens as $token) {
+            if ($token->isGivenKind($magicConstantTokens)) {
+                $token->setContent($magicConstants[$token->getId()]);
+            }
+        }
     }
 
     /**
--- a/src/Fixer/Casing/MagicConstantCasingFixer.php
+++ b/src/Fixer/Casing/MagicConstantCasingFixer.php
@@ -26,7 +26,7 @@ final class MagicConstantCasingFixer extends AbstractFixer
     /**
      * {@inheritdoc}
      */
-    public function applyFix(\SplFileInfo $file, Tokens $tokens)
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
         $magicConstants = $this->getMagicConstants();
         $magicConstantTokens = $this->getMagicConstantTokens();
--- a/src/Fixer/Casing/MagicConstantCasingFixer.php
+++ b/src/Fixer/Casing/MagicConstantCasingFixer.php
@@ -28,13 +28,14 @@ final class MagicConstantCasingFixer extends AbstractFixer
     public function fix(\SplFileInfo $file, Tokens $tokens)
     {
         $magicConstants = $this->getMagicConstants();
+        $magicConstantTokens = $this->getMagicConstantTokens();
 
         for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
-            if (!$tokens[$index]->isGivenKind(array_keys($magicConstants))) {
+            if (!$tokens[$index]->isGivenKind($magicConstantTokens)) {
                 continue;
             }
 
-            if (!array_key_exists($tokens[$index]->getId(), $magicConstants)) {
+            if (!isset($magicConstants[$tokens[$index]->getId()])) {
                 continue;
             }
 
@@ -58,7 +59,7 @@ final class MagicConstantCasingFixer extends AbstractFixer
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isAnyTokenKindsFound(array_keys($this->getMagicConstants()));
+        return $tokens->isAnyTokenKindsFound($this->getMagicConstantTokens());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


ntzm:
Fixing 5.* now

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,85 @@
...
+    {
+        return $tokens->isAnyTokenKindsFound(array_keys($this->getMagicConstants()));
+    }
+
+    /**
+     * @return array<int, string>
+     */
+    private function getMagicConstants()
+    {
+        $magicConstants = array(
```
keradus: could be static.
it;s not const array yet, so no need to recreate it each time method is called

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,85 @@
...
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $magicConstants = $this->getMagicConstants();
+
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(array_keys($magicConstants))) {
+                continue;
+            }
+
+            if (!array_key_exists($tokens[$index]->getId(), $magicConstants)) {
```
keradus: `isset` is way faster

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $magicConstants = $this->getMagicConstants();
+
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind(array_keys($magicConstants))) {
```
keradus: no need to call `array_keys` for each iteration
similar with `isCandidate`

keradus:
what if someone imported his own `__file__` constant ?

ntzm:
Thanks @keradus, I will take those into consideration!

ntzm:
I've updated my PR with your suggestions.

In terms of importing a magic constant, I don't think it's possible (correct me if I'm wrong), see:
https://3v4l.org/uuGn4
https://3v4l.org/ndn9b

keradus:
you have used wrong syntax to import a const, but indeed it's not possible to import const named like keyword https://3v4l.org/nug2Z

ntzm:
Oops 🤦‍♂️

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,104 @@
...
+    {
+        $magicConstants = $this->getMagicConstants();
+        $magicConstantTokens = $this->getMagicConstantTokens();
+
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
+            if (!$tokens[$index]->isGivenKind($magicConstantTokens)) {
+                continue;
+            }
+
+            if (!isset($magicConstants[$tokens[$index]->getId()])) {
```
SpacePossum: is it ever possible that this passes after the `if` on line 34?

ntzm: You're right 🤦‍♂️ will remove

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,100 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function fix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $magicConstants = $this->getMagicConstants();
+        $magicConstantTokens = $this->getMagicConstantTokens();
+
+        for ($index = 0, $count = $tokens->count(); $index < $count; ++$index) {
```
SpacePossum: this can be simplified even further now :)
```php
foreach($tokens as $token) {
    if ($token->isGivenKind($magicConstantTokens)) {
        $token->setContent($magicConstants[$token->getId()]);
    }
}
```

tests/Fixer/Casing/MagicConstantCasingFixerTest.php
```diff
@@ -0,0 +1,119 @@
...
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return array(
+            array(
+                '<?php echo __LINE__;',
+                '<?php echo __line__;',
```
SpacePossum: please mix up some of the cases, now you test all-lower and all-upper, having one variation would be nice (for example `__lInE__`)

julienfalque:
What about supporting `::class` as well?

keradus:
valid scenario, but then it's not a magic constant

julienfalque:
According to [PHP Manual](http://php.net/manual/en/language.constants.predefined.php), it is.

keradus:
wow, :+1:

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -0,0 +1,100 @@
...
+                T_LINE => '__LINE__',
+                T_FILE => '__FILE__',
+                T_DIR => '__DIR__',
+                T_FUNC_C => '__FUNCTION__',
+                T_CLASS_C => '__CLASS__',
+                T_METHOD_C => '__METHOD__',
+                T_NS_C => '__NAMESPACE__',
+            );
+
+            if (PHP_VERSION_ID >= 50400) {
```
keradus: please replace with `defined('T_TRAIT_C')`, as we do in other places.

keradus:
`Foo::class` are 3 tokens, you fix last one, so you fix single token as well.

ntzm:
Yep, but `T_CLASS` can also refer to when you declare a class, so I'll have to do some checks for `T_DOUBLE_COLON` before it

julienfalque:
> Also anyone else having trouble viewing the travis logs? They aren't loading for me

Yep, same here :/

keradus:
You are not interested in `T_CLASS`, you are in `CT::T_CLASS_CONSTANT` (thanks to `ClassConstantTransformer`).

SpacePossum:
Dealing with `T_CLASS_CONSTANT` looks good :)
Please rebase, btw. the last Travis job for this PR has logs again \o/

ntzm:
What am I running to rebase sorry? I tried `git rebase -i master` and squashing everything but my first commit but I got conflicts 😢

SpacePossum:
don't worry about squashing, as long as you are the only author the squashing can be done simple at the end before it is merged

SpacePossum:
LGTM :)

src/Fixer/Casing/MagicConstantCasingFixer.php
```diff
@@ -77,6 +78,10 @@ private function getMagicConstants()
             if (defined('T_TRAIT_C')) {
                 $magicConstants[T_TRAIT_C] = '__TRAIT__';
             }
+
+            if (PHP_VERSION_ID >= 50500) {
```
keradus: again, no need for checking php version

ntzm: How should I check if we should be fixing `::class` then?

keradus: we should fix it, no need for checking it

SpacePossum: it cheaper to the test here one time and when not hit not add it to the array, that way you won't compare all the tokens against a type you already know is never available.

keradus:
please add to symfony ruleset

keradus:
Thank you @ntzm.

SpacePossum:
nice, thanks so much @ntzm :D

Issue #1492, opened 2015-11-08
Items of the @property tag should be aligned vertically


ffsantos92:
I'm using the fixer:
- phpdoc_align (all items of the `@param`, `@throws`, `@return`, `@var`, and `@type` phpdoc tags must be aligned vertically).

But why not include also all items of the `@property` tag?

drAlberT:
:+1:  `@property` should be aligned too

keradus:
feel free to collaborate ;)

julienfalque:
Closing as #2649 was merged.

PR #2643, opened 2017-03-26
Align @property PHPDoc tags

ntzm:
See #1492

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -809,8 +809,8 @@ Choose from the list of available rules:
 
 * **phpdoc_align** [@Symfony]
 
-  All items of the @param, @throws, @return, @var, and @type phpdoc tags
-  must be aligned vertically.
+  All items of the @param, @throws, @return, @var, @type and @property
+  phpdoc tags must be aligned vertically.
 
 * **phpdoc_annotation_without_dot** [@Symfony]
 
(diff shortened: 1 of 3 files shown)


keradus:
I'm afraid this changes the behaviour and for that shall be optional

ntzm:
I'd say it fixes behaviour, this is how I'd expect it to work

julienfalque:
`@property` is not listed in the current description and it was never documented as supported by the fixer so it's a new feature, not a bugfix. But I agree it would make sense to enable it by default anyway. Is there a specific reason it was not implemented before?

ntzm:
Maybe we could make this fixer configurable and allow the user to decide which tags they want to align?

SpacePossum:
>Maybe we could make this fixer configurable and allow the user to decide which tags they want to align?

I think that is what @keradus meant :)  

> Is there a specific reason it was not implemented before?

AFAIK no, prop, no one find time to make it

ntzm:
I'll have a go at making it configurable tonight. In terms of backwards compatibility, if no options are passed should it behave like it currently does (all but `@property`)?

keradus:
Default config should be like current behaviour. It can be changed in future on 3.x line.
Just please, don't make `include_property` or sth like this config key. Keys should be as much generic as possible, like `[tags => return, var, ...]` or sth

ntzm:
Gonna close this and open up a new one tonight 👍

keradus:
ping @ntzm

ntzm:
Got a bit stuck on stuff last night, is there a place I can ask noob questions?

SpacePossum:
You can find us on gitter; fixer channel:
https://gitter.im/FriendsOfPHP/PHP-CS-Fixer
but you can send also direct messages to people.
I try to be there a bit today :)

PR #2645, opened 2017-03-27
HHVM: handle T_HH_ERROR

keradus:


src/Tokenizer/Tokens.php
```diff
@@ -961,6 +961,10 @@ public function setCode($code)
             $this->registerFoundToken($token);
         }
 
+        if (defined('T_HH_ERROR') && $this->isTokenKindFound(T_HH_ERROR)) {
+            throw new \ParseError('Parsing error, encountered "T_HH_ERROR".');
```
GrahamCampbell: Not sure about this error message. If we do keep the token name in the error message, it would be more consistent (I think), to omit the quotes.

SpacePossum:
:+1:  (and :+1: for the branch name ;) )

src/Tokenizer/Tokens.php
keradus: I don't find any way to make this consistent with PHP error message:
`"syntax error, unexpected 'asd' (T_STRING), expecting ...`.
under HHVM, you just got T_HH_ERROR for almost all tokens if parsing fails.

GrahamCampbell:
So many dependencies.

keradus:
so?

PR #2648, opened 2017-03-28
Move non-code covering tests to AutoReview subnamespace

keradus:


tests/AutoReview/FixerFactoryTest.php
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ *
+ * @coversNothing
+ * @group auto-review
+ */
+final class FixerFactoryTest extends \PHPUnit_Framework_TestCase
```
julienfalque: I think this can be merged in `FixerTest` (like before) as both classes do general tests on all fixers. Otherwise class name should be changed as it does no test `FixerFactory` specifically.

tests/AutoReview/FixerTest.php
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ *
+ * @coversNothing
+ * @group auto-review
+ */
+final class FixerTest extends \PHPUnit_Framework_TestCase
```
julienfalque: I would rename this class to `FixersTest` just to be more explicit.

tests/AutoReview/FixerFactoryTest.php
keradus: my reasoning:
`FixerTest` checks fixer atomically one by one.
`FixerFactoryTest` checks relations between fixers (which is used only inside FixerFactory, that's why it was put here in the first place)

SpacePossum: -1 for the merge. I agree some test here are not about the `FixerFactory` and should be moved to a better test some day, but the `FixerTest` is not the right spot either IMHO. Anyway that move does not need to be part of this PR I think.

tests/AutoReview/FixerTest.php
SpacePossum: I'm -1 on this, as the test is about the `Fixer` interface (and/or the known implementing classes of it) and not about the `Fixers` interface. The 1-on-1 naming has my preference.

SpacePossum:
:+1:

tests/AutoReview/FixerTest.php
julienfalque: The 1-on-1 naming is not valid here as namespaces are different. IMO this convention is only important when the test actually covers a specific class. But I get the point.

keradus:
Thanks for reviewing the concept.
I agree that naming could be improved. Let us make it later if one will have better idea, and merge it for now as it's easy to have conflicts on this PR, thus I don't want to have this PR open for too long

PR #2651, opened 2017-03-29
Fix UPGRADE.md table syntax so it works in GitHub

ntzm:
GitHub doesn't like when you miss out the starting pipe and have an empty cell first

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -50,7 +50,7 @@ Exit codes
 Exit codes for `fix` command have been changed and are build using the following bit flags:
 
 | 1.x bit | 2.0 bit | Description                                                 | Note                                                             |
-| ------- | ------- | ----------------------------------------------------------- | ---------------------------------------------------------------- |
+| -------:| -------:| ----------------------------------------------------------- | ---------------------------------------------------------------- |
 | 0       | 0       | OK                                                          |                                                                  |
 | 1       | 1       | General error (or PHP/HHVM minimal requirement not matched) | no longer used for other states, never combined with other flags |
 |         | 4       | Some files have invalid syntax                              | flag was added, works only in dry-run mode                       |
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -22,8 +22,8 @@ CLI options
 
 1.x             | 2.0             | Description                                                               | Note
 --------------- | --------------- | ------------------------------------------------------------------------- | ----
-                | --allow-risky   | Are risky fixers allowed                                                  |
-                | --cache-file    | The path to the cache file                                                | option was added
+&nbsp;          | --allow-risky   | Are risky fixers allowed                                                  |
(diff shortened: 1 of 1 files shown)


SpacePossum:
hmm odd we never spotted that before, good catch!
is there a way we can fix the layout without the `&nbsp;`, it feels odd to mix HTML with MD.

ntzm:
You can surround the whole table with pipes:

```
| heading | foo |
| ------- | --- |
| foo     | bar |
```

SpacePossum:
worth I try I would say :)

SpacePossum:
looking good :)
thanks!

keradus:
opening pipe on left side, but not closing pipe on right side ?

ntzm:
Shall I convert the other tables as well?

julienfalque:
@ntzm I'd say yes, for consistency.

UPGRADE.md
```diff
@@ -49,15 +49,15 @@ Exit codes
...
-1.x bit | 2.0 bit | Description                                                 | Note
--------:| -------:| ----------------------------------------------------------- | ----
-0       | 0       | OK                                                          |
-1       | 1       | General error (or PHP/HHVM minimal requirement not matched) | no longer used for other states, never combined with other flags
-        | 4       | Some files have invalid syntax                              | flag was added, works only in dry-run mode
-        | 8       | Some files need fixing                                      | flag was added, works only in dry-run mode
-16      | 16      | Configuration error of the application                      |
-32      | 32      | Configuration error of a Fixer                              |
-        | 64      | Exception within the application                            | flag was added
+| 1.x bit | 2.0 bit | Description                                                 | Note
```
julienfalque: You forgot the last `|` here.

UPGRADE.md
```diff
@@ -49,15 +49,15 @@ Exit codes
...
+| -------:| -------:| ----------------------------------------------------------- | ---------------------------------------------------------------- |
```
julienfalque: Can you also replace the `:` here? I don't think they were intended.

keradus:
ping @ntzm

ntzm:
Oops, I'll fix those in an hour or so

UPGRADE.md
```diff
@@ -49,15 +49,15 @@ Exit codes
 
 Exit codes for `fix` command have been changed and are build using the following bit flags:
 
-1.x bit | 2.0 bit | Description                                                 | Note
--------:| -------:| ----------------------------------------------------------- | ----
```
keradus: from what I see @julienfalque asked you to remove the `:` here

they are supposed to align columns with integer to right side. I would keep that alignment.

julienfalque: Oh I didn't know that, sorry.

keradus: changed

keradus:
Thank you @ntzm , really great you spot the issue and be breave to fix it yourself !

PR #2656, opened 2017-03-30
Code grooming

keradus:


src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -111,6 +111,7 @@ public function configure(array $configuration = null)
                     break;
                 case 'square_brace_block':
                     $this->tokenKindCallbackMap[CT::T_ARRAY_SQUARE_BRACE_OPEN] = 'fixStructureOpenCloseIfMultiLine'; // typeless '[' tokens should not be fixed (too rare)
+                    break;
```
GrahamCampbell: Meh. I guess this makes sense to prevent errors from adding new cases. I somehow doubt php optimizes this break away though.

GrahamCampbell:
Looks good. 👍

src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
keradus: indeed, it's not about performance but being error prone

PR #2657, opened 2017-03-31
PhpdocToCommentFixer - Allow phpdoc for language constructs

### LINKED ISSUE #2634: phpdoc_to_comment inconsistent treatment of PHPDoc tags

SpacePossum:
closes #2634 
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2636

rebase the work of @ceeram and add supported for short array destructing syntax

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocToCommentFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocToCommentFixer.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Fixer\Phpdoc;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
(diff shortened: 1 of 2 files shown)


ceeram:
Nice. Thanks for finishing this.

Op 31 mrt. 2017 8:19 a.m. schreef "SpacePossum" <notifications@github.com>:

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks @ceeram :)

PR #2660, opened 2017-03-31
Fixed typo in suggest for ext-mbstring

pascal-hofmann:


keradus:
wow, we just released this, thanks !

GrahamCampbell:
👍

GrahamCampbell:
> fabbot.io — The rules contain unknown fixers (non_printable_character).

Does this mean fabbot.io is executing our .php_cs file? If so, that's definitely a bad idea, since someone can send a PR and execute arbitrary code on their servers.

keradus:
it's ok as fabbot runs it in virtualized sandbox
in the end, it's not up to me or you, but up to fabbot.

GrahamCampbell:
What sort of information could be leaked from the sandbox is the real question I guess.

keradus:
It's up to the infrastructure and fabbot.io decided they are prepared enough to do this.
If you want, raise a concern directly to them, not here.

keradus:
Thank you @pascal-hofmann.

PR #2663, opened 2017-04-01
Use colors for keywords in commands output

julienfalque:
Before:
Before

After:
After


README.rst
```diff
@@ -214,8 +214,8 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``syntax`` (``'long'``, ``'short'``): whether to use the ``long`` or ``short`` array
-    syntax; defaults to ``'long'``
+  - ``syntax`` (``'long'``, ``'short'``): whether to use the ``long`` or ``short`` array syntax;
```
julienfalque: I couldn't find any reference to such limit. Anyway, changes in this file are only due to `` ` `` characters being replaced by coloring in the `fix --help` output and thus not taken into account in the line length calculation there.

SpacePossum: http://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#line-length

I guess we follow Silex/Twig/SF guidelines, however since we don't use sphinx (I think?) for generating docs we could drop the limit. So maybe a more generic question for the community; should we stay with `.rst` format or update to `.md` ?

julienfalque: IMO we should have a concise README (Markdown) and move all details here to doc/ (reStructuredText), but that's out of the scope of this PR.

SpacePossum: If changing the format is out of scope then the changes here should be minimized.
I.e. there is no need to change the line length here and so should be reverted.

julienfalque: I didn't change any line length. I expected the README file to not change at all but actually got surprised the related test failed when I ran the test suite.

In `fix --help` output, as `` ` `` characters are replaced by color tags and keywords become shorter so wordwrapping is impacted and sometimes allows a little more text before a newline. As the README file uses the newlines from the `fix --help` output without changing them, it is impacted as well. Not sure we can do something about it or if it's even worth it.

keradus: 1. rst format must be kept
2. i don't mind line length in this file

SpacePossum: alright, lets leave this changes as is.

src/Console/Command/CommandHelp.php
```diff
@@ -420,7 +430,8 @@ function (FixerOptionInterface $optionA, FixerOptionInterface $optionB) {
             }
         }
 
-        return $help;
+        // prevent "\</foo>" from being rendered as an escaped literal style tag
+        return preg_replace('#\\\\(</.*?>)#', '<<$1', $help);
```
SpacePossum: Can you create a test scenario for this changes? I cannot create a scenario where when removing this the output changes.

SpacePossum:
Looks like a nice addition to me, however there is a changes in the output that is less nice IMO.
Current master
```
/-[/home/possum/work/PHP-CS-Fixer]-[git master] 
$ php php-cs-fixer describe array_syntax --no-ansi
Description of array_syntax rule.
PHP arrays should be declared using the configured syntax (requires PHP >= 5.4 for short syntax).

* syntax ('long', 'short'): whether to use the `long` or `short` array syntax; defaults to 'long'
```

on this branch
```
/-[/home/possum/work/PHP-CS-Fixer]-[git help-keywords-colors] 
$  php php-cs-fixer describe array_syntax --no-ansi
Description of array_syntax rule.
PHP arrays should be declared using the configured syntax (requires PHP >= 5.4 for short syntax).

* syntax ('long', 'short'): whether to use the long or short array syntax; defaults to 'long'
```

note the diff in the last line, from "`long` or `short` array syntax;" to "long or short array syntax;", i.e. the back ticks are removed on no color outputs.

src/Console/Command/CommandHelp.php
julienfalque: This happens with `native_function_invocation` description: _Add leading `\` before function invocation [...]._ You may make it more obvious by replacing `<info>` tags with `<error>` tags.

keradus:
yep, I think we shall keep `'` around optoin value

julienfalque:
I ended up leaving one backtick (instead of two) around all keywords, even when colors are supported as we have no access to the output instance in `Command::configure()` thus we can't detect color support at that point.

keradus:
could you make new `before/after` screens please?

julienfalque:
PR description updated.

src/Console/Command/ReadmeCommand.php
```diff
@@ -213,7 +213,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $help = $command->getHelp();
         $help = str_replace('%command.full_name%', 'php-cs-fixer.phar '.$command->getName(), $help);
         $help = str_replace('%command.name%', $command->getName(), $help);
+        $help = preg_replace('#<<(</.*?>)#', '\\\\$1', $help);
```
keradus: why is this needed ? when I drop it, all works...

src/Console/Command/ReadmeCommand.php
```diff
@@ -213,7 +213,9 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $help = preg_replace('#</?(comment|info)>#', '``', $help);
+        $help = preg_replace('#`(``.+?``)`#', '$1', $help);
```
keradus: this one is supposed to replace triple backtick with double backtick. how we created triple backtick in first place ?

src/Console/Command/ReadmeCommand.php
julienfalque: I added it because of [this change](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2663/files#diff-a8259d3156d95c86bb9eb9f59fa60b4bR434). But actually Symfony already does this replacement when writing output so yes, this change here is probably unnecessary.

src/Console/Command/ReadmeCommand.php
julienfalque: The triple backticks are introduced by the line just before this one, but I can probably merge both regular expressions.

keradus: ok, no need for. thanks for info

julienfalque:
Thanks :)

keradus:
Thank you @julienfalque.

PR #2667, opened 2017-04-02
Add NoBreakCommentFixer

julienfalque:
Fixer for the following [PSR-2 rule](http://www.php-fig.org/psr/psr-2/#switch-case):

> There MUST be a comment such as `// no break` when fall-through is intentional in a non-empty case body.

It always uses `// no break` for now but the rule does not seem to prevent other comments, should it be configurable?

Also, I was thinking about removing unnecessary `break` statements (when they are in the last `case` body for example) but maybe it's out of scope here.


SpacePossum:
I'm -1 on concept.
I so no problem developers adding comments so having a rule discouraging those seems wrong to me. I've no idea why this made to PSR2 btw.

SpacePossum:
>Also, I was thinking about removing unnecessary break statements (when they are in the last case body for example) but maybe it's out of scope here.

That should be another fixer IMO, however I would not be in fav. of that one.

julienfalque:
> having a rule discouraging those seems wrong to me.

It does encourage them: the fixer adds the `// no break` comment when missing, and removes it when it does not make sense (when there is an actual `break` for example).

SpacePossum:
>and removes it when it does not make sense 

the detection of this (`make sense`) is missing I guess, i.e. it always removes the comment?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+        return 0;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $commentPosition
+     */
+    private function removeComment(Tokens $tokens, $commentPosition)
+    {
+        $tokens[$commentPosition]->clear();
```
SpacePossum: [$tokens->clearTokenAndMergeSurroundingWhitespace($commentPosition)](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.1/src/Tokenizer/Tokens.php#L1124)
and than you don't need line 156-159

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+ */
+final class NoBreakCommentFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'There must be a comment when fall-through is intentional in a non-empty case body.',
```
SpacePossum: the description is incomplete IMHO, it should state that it removes comments as well

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+
+            if ($tokens[$i]->isGivenKind(array(T_BREAK, T_CONTINUE, T_RETURN))) {
+                if (null !== $commentPosition) {
+                    $this->removeComment($tokens, $commentPosition);
+                }
+
+                break;
+            }
+
+            if ($tokens[$i]->isGivenKind(array(T_COMMENT)) && '// no break' === $tokens[$i]->getContent()) {
```
SpacePossum: why not allow `/* no break*/` and `# no break` and variations; `# we do not break here by design`?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+                $commentPosition = $i;
+                continue;
```
SpacePossum: missing line break above `continue`

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+
+        $position = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $position);
+
+        if ($initialToken->isGivenKind(array(T_DO))) {
+            $position = $tokens->findBlockEnd(
+                Tokens::BLOCK_TYPE_PARENTHESIS_BRACE,
+                $tokens->getNextTokenOfKind($position, array('('))
+            );
+
+            return $tokens->getNextTokenOfKind($position, array(';'));
```
SpacePossum: please add a test for a structure that ends with `?>` and not with `;`

tests/Fixer/ControlStructure/NoBreakCommentFixerTest.php
```diff
@@ -0,0 +1,504 @@
...
+    case 1:
+        // no break
+    case 2:
+        bar();
+        // no break
+        break;
+    default:
+        baz();
+}',
+            ),
```
SpacePossum: please add a test with at least 3 `switch` constructions all with missing comments and make all of those on one line

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer for rule defined in PSR2 ¶5.2.
+ */
+final class NoBreakCommentFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
```
SpacePossum: priority is missing, must go after PHPdoc to comment fixer

julienfalque: `phpdoc_to_comment` priority is 25, so this one already runs after it as its priority is 0 (default). Do you want me to make the priority explicit anyway?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Using the suggested replacement breaks some tests, are you sure it's the same behavior?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Added as full description below.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Handling different comment types makes sense. But only one sentence should be treated as the "no break" comment IMO, otherwise we might confuse with an unrelated comment. Should the sentence be configurable then?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Added it.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: we should have it tested in the FixerFactoryTest and pref. with a fixture test in the priority folder to prevent regression when the priority of any of the two fixers changes.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: hmm maybe make the delete optional, I don't see why `# we do not break here by design` should be removed. Maybe just check 'break' is in the content of the comment? Would like more opinions on this though.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: btw. there might be more priority issues but I didn't check, but braces as well

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: > Maybe just check 'break' is in the content of the comment?

Could be nice but could still lead to unexpected results IMO, I'd like to hear more opinions on that too :)

> if you allow/insert comment, why call the fixer no_break_comment and not break_comment fixer?

I wanted the name to be _"No break" Comment Fixer_, but yeah without quotes, it's confusing. What about _Fall-through Case Comment Fixer_ (`fall_through_case_comment`)? I would only name it `break_comment` if we include the removal of unecessary `break`statements in the scope of this fixer.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,218 @@
...
+            'Adds a "no break" comment before fall-through cases, and removes it if there is no fall-though.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_CASE);
```
SpacePossum: or `default` as a `switch` with only a default with `break` and comment should also be fixed?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,216 @@
...
+                break;
+            }
+
+            if ($tokens[$i]->isGivenKind(array(T_COMMENT)) && '// no break' === $tokens[$i]->getContent()) {
+                $commentPosition = $i;
+                continue;
+            }
+
+            if ($tokens[$i]->isGivenKind(array(T_CASE, T_DEFAULT))) {
+                if (!$empty) {
```
julienfalque: It would require to write:

```php
if ($empty) {
    if (null !== $commentPosition) {
        $this->removeComment($tokens, $commentPosition);
    }
} elseif (null === $commentPosition) {
    $this->insertCommentAt($tokens, $i);
} else {
    $this->ensureNewLineAt($tokens, $commentPosition);
}
```
and having a single `if` inside another `if` looks weird to me.

tests/Fixer/ControlStructure/NoBreakCommentFixerTest.php
```diff
@@ -0,0 +1,504 @@
...
+        );
```
SpacePossum: please add a alternative syntax case as well

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: It not the same, but good enough. I think the fixer should not remove the whitespace but leave that to other fixers.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: LGTM

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Added a test and made it pass.

tests/Fixer/ControlStructure/NoBreakCommentFixerTest.php
```diff
@@ -0,0 +1,561 @@
...
+            }
+        };
+        // no break
+    case 2;
+        bar();
+}',
+                '<?php
+switch ($foo) {
+    case 1;
+        $foo = new class {
```
keradus: this case requires php7

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: I mean the `require` write you show LGTM and the nested if is just what it is ^_^

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: I finally used `clearTokenAndMergeSurroundingWhitespace` as it was required for some new test cases I added.

> I think the fixer should not remove the whitespace but leave that to other fixers.

I tend to agree but I think it's better if the fixer is able to leave code as clean as possible on its own instead of assuming other fixers will clean it after.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: The fixer now checks any type of comment (including `/** ... */`) and trims the text. It still checks if the text is `no break` though, should we move to just detecting "break" like you suggested?

tests/Fixer/ControlStructure/NoBreakCommentFixerTest.php
julienfalque: Test added.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Is the priority issue with `phpdoc_to_comment` still relevant now that the fixer checks all type of comments? What would be the issue with `braces`?

tests/Fixer/ControlStructure/NoBreakCommentFixerTest.php
julienfalque: Case moved to a dedicated test.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    /**
+     * @param Token $token
+     *
+     * @return bool
+     */
+    private function isNoBreakComment(Token $token)
+    {
+        if (!$token->isGivenKind(array(T_COMMENT, T_DOC_COMMENT))) {
```
SpacePossum: please use the short hand [!$token->isComment()](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/Tokenizer/Token.php#L352)

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: With all types of comments checked the prio issues should be over. I'll check again when this PR is close to merging :+1:

julienfalque:
This one should be ready for a new review iteration :)

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,309 @@
...
+    private function insertCommentAt(Tokens $tokens, $casePosition)
+    {
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+
+        $newlinePosition = $this->ensureNewLineAt($tokens, $casePosition);
+
+        $newlineToken = $tokens[$newlinePosition];
+
+        $nbNewlines = substr_count($newlineToken->getContent(), $lineEnding);
+        if ($tokens[$newlinePosition - 1]->isGivenKind(T_OPEN_TAG) && false !== strpos($tokens[$newlinePosition - 1]->getContent(), $lineEnding)) {
```
SpacePossum: are you even inserting a comment at a place where the previous token is `T_OPEN_TAG`, I think that is impossible as there cannot be `T_OPEN_TAG` tokens inside a `switch` IIRC.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,309 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $casePosition
+     */
+    private function fixCase(Tokens $tokens, $casePosition)
+    {
+        $empty = true;
+        $fallThrough = true;
+        $commentPosition = null;
+        for ($i = $tokens->getNextTokenOfKind($casePosition, [':', ';']) + 1, $max = count($tokens); $i < $max; ++$i) {
```
julienfalque: No it's not: `$casePosition` is an integer parameter of the function, and comes from the foreach in `applyFix()`.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: [It's actually possible](https://3v4l.org/cHseT). I added test cases to cover this.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,309 @@
...
+                if (!$empty && $fallThrough) {
+                    if (null !== $commentPosition && $tokens->getPrevNonWhitespace($i) !== $commentPosition) {
+                        $this->removeComment($tokens, $commentPosition);
+                        $commentPosition = null;
+                    }
+
+                    if (null === $commentPosition) {
+                        $this->insertCommentAt($tokens, $i);
+                    } else {
+                        $this->ensureNewLineAt($tokens, $commentPosition);
```
julienfalque: The current behavior of this fixer is to make sure the comment is on its own line, hence this `ensureNewLineAt()` call. It does not add, change or remove any other new line so it should not conflict with other fixers.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: ah right, I misread the param name, :+1:

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: oh man, PHP haha

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,309 @@
...
+     * @param Tokens $tokens
+     * @param int    $casePosition
+     */
+    private function fixCase(Tokens $tokens, $casePosition)
+    {
+        $empty = true;
+        $fallThrough = true;
+        $commentPosition = null;
+        for ($i = $tokens->getNextTokenOfKind($casePosition, [':', ';']) + 1, $max = count($tokens); $i < $max; ++$i) {
+            if ($tokens[$i]->isGivenKind([T_SWITCH, T_IF, T_ELSE, T_ELSEIF, T_FOR, T_FOREACH, T_WHILE, T_DO, T_FUNCTION, T_CLASS])) {
```
keradus: not sure...
array packing (https://blog.blackfire.io/php-7-performance-improvements-packed-arrays.html) + immutable array (https://blog.blackfire.io/php-7-performance-improvements-immutable-arrays.html) would work better than static array IMO (as it would register some static memory that would not be freezed until end of whole tool execution)

I would leave it as-is

src/Fixer/ControlStructure/NoBreakCommentFixer.php
keradus: feel free to drop whole logic for handling open tags and fix file only for monolitic code

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: If there is no real performance gain, I'll leave it as-is too, as moving the array to a variable would decrease readability IMO.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Well, it works fine so I think it's ok to keep it.

IMO we should implement a new token transformer to extract the whitespace from `T_OPEN_TAG`, this would simplify logic here a lot. But it would be a BC break so not going to happen soon.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
keradus: exactly ;) and I would keep code to run nicer in newer, not older compier.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
keradus: discussed already in other threads and you know it's not gonna happen soon or easy.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: alright than lets agree to disagree (on readability) and leave as is :)

src/Fixer/ControlStructure/NoBreakCommentFixer.php
SpacePossum: alright, if you want to keep/add/detect the indent than please add priority tests for at least `braces` and `indentation_type`

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -50,6 +53,19 @@ public function isCandidate(Tokens $tokens)
     /**
      * {@inheritdoc}
      */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
```
julienfalque: @keradus I'm thinking about working on a PR to write all configuration definitions this way (method chaining only) as this is now possible on 2.3. Would you accept such PR?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -76,7 +92,7 @@ private function fixCase(Tokens $tokens, $casePosition)
                 continue;
             }
 
-            if ($tokens[$i]->isGivenKind([T_BREAK, T_CONTINUE, T_RETURN])) {
+            if ($tokens[$i]->isGivenKind([T_BREAK, T_CONTINUE, T_RETURN, T_EXIT, T_THROW])) {
```
SpacePossum: `T_GOTO` is missing >: )

julienfalque: Damn...

keradus: let us save the world, `unset(T_GOTO)` ;P

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -0,0 +1,321 @@
...
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $position => $token) {
```
keradus: <?php

$arr = [1, 2, 3];

foreach ($arr as $a) {
    if (1 === $a) {
        $arr[] = 4;
    }

    var_dump($a);
}

keradus: thus, this require a fix

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -68,9 +68,10 @@ protected function createConfigurationDefinition()
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        foreach ($tokens as $position => $token) {
-            if ($token->isGivenKind([T_CASE, T_DEFAULT])) {
+        for ($position = 0, $max = count($tokens); $position < $max; ++$position) {
+            if ($tokens[$position]->isGivenKind([T_CASE, T_DEFAULT])) {
                 $this->fixCase($tokens, $position);
+                $max = count($tokens);
```
keradus: would it make sense to iterate backwards instead ?

julienfalque: Oh yeah definitely, will do that 👍

keradus:
Thank you @julienfalque.

GrahamCampbell:
> There MUST be a comment such as // no break

Not sure this belongs in the PSR-2 rules tbh.

GrahamCampbell:
I'd also be interested to see what the symfony core team make of having this fixer applied to symfony/symfony.

keradus:
good to have direct link to PSR2 spec claiming that in initial message

GrahamCampbell:
Yeh, I was quoting PSR-2. I says there must be some kind of comment, but it doesn't say what comment. I don't think this is something that should be auto-fixed.

keradus:
oh no, there must be sth, so fixer is putting sth. there must be _sth_, co you could configure what.

PR #2668, opened 2017-04-02
add covers annotation

keradus:


keradus:
> Coverage decreased (-6.0%) to 83.14%

:(

GrahamCampbell:
These look incorrect:

image

Missing covers annotations on classes that actually test those.

keradus:
only one fixer uses that 2 helpers.
i add the annotations for it, but in the end it would be nice to have separated, dedicated tests for them

SpacePossum:
>-4.35 src/Console/Command/ReadmeCommand.php

the test for this is in;
```
 * @coversNothing
 * @group auto-review
```

is that the cause of the drop in coverage report?

keradus:
could be, but do we really test `ReadmeCommand` in this file?
it was introduced as only comparing current readme generated by command with readme stored in file

SpacePossum:
We test the output of the command against what we expect; unless the REAME _had_ to change, so most of the time we do at least test that the output generated is as expected. However I see your point ;)

keradus:
ok, on the other side - if we would create unit test for ReadmeCommand, what would it test besides generated output? maybe indeed we should move readme test outside auto-review group

SpacePossum:
moving it as suggested move perfect sense to me :+1:

keradus:
moved

SpacePossum:
:+1:  however my guess is https://github.com/keradus/PHP-CS-Fixer/blob/ab97b807614b78c100946493d0c84085cae19066/tests/Console/Command/ReadmeCommandTest.php#L40 needs to be updated

keradus:
you are right

PR #2669, opened 2017-04-03
TokensTest - grooming

SpacePossum:
should fix https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/217815194

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -705,7 +705,7 @@ class Tokens extends \SplFixedArray
     /**
      * Find a sequence of meaningful tokens and returns the array of their locations.
      *
-     * @param array      $sequence      an array of tokens (same format used by getNextTokenOfKind)
+     * @param array      $sequence      an array of tokens (kinds) (same format used by getNextTokenOfKind)
      * @param int        $start         start index, defaulting to the start of the file
      * @param int        $end           end index, defaulting to the end of the file
      * @param bool|array $caseSensitive global case sensitiveness or an array of booleans, whose keys should match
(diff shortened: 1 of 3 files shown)


src/Tokenizer/Tokens.php
```diff
@@ -705,7 +705,7 @@ public function getPrevMeaningfulToken($index)
     /**
      * Find a sequence of meaningful tokens and returns the array of their locations.
      *
-     * @param array      $sequence      an array of tokens (same format used by getNextTokenOfKind)
+     * @param array      $sequence      an array of tokens (kinds) (same format used by getNextTokenOfKind)
```
julienfalque: Why putting parentheses around "kinds"?

SpacePossum: because you can pass an array of Tokens or an array of tokens kinds;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.1/src/Tokenizer/Tokens.php#L734

keradus:
could you point the logic change please? i see only reformatting

julienfalque:
There is no logic change, only breaking down the `$params` array into decidated arguments to avoid the PHPUnit 5.7.18 regression case, which occurs when a test argument is an array that mixes objects with scalars.

This increases the number of test arguments but removes the need to use `call_user_func_array`, which is an improvement to me.

keradus:
ok, good for me, but it's not a fix, just small refactoring/cleaning ;)
:+1: but please adjust title

keradus:

SpacePossum:
all green
granted after the fix on PHPUnit all value left in this PR is:
- comment copy
- PHPDoc type hint tweaks
- easier to read and maintain utest

not sure if enough to validate all the changes in the PR, I think it is, but I'm OK to drop as well

keradus:
Thank you @SpacePossum.

PR #2670, opened 2017-04-03
AbstractFixer: use applyFix instead of fix

Slamdunk:
Stated that:
1. `AbstractFixer` is `@internal`
1. `applyFix` is mandatory
1. Tests pass
1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.0/doc/COOKBOOK-FIXERS.md#step-1---creating-files

`fix` should be final

### DIFF
--- a/src/AbstractFixer.php
+++ b/src/AbstractFixer.php
@@ -61,7 +61,7 @@ abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
         }
     }
 
-    public function fix(\SplFileInfo $file, Tokens $tokens)
+    final public function fix(\SplFileInfo $file, Tokens $tokens)
     {
         if (0 < $tokens->count() && $this->isCandidate($tokens) && $this->supports($file)) {
             $this->applyFix($file, $tokens);


Slamdunk:
https://www.martinfowler.com/bliki/CallSuper.html

SpacePossum:
>https://www.martinfowler.com/bliki/CallSuper.html

not sure how related, we don't force anyone through our public API to call `super::` at some point I think?

SpacePossum:
LGTM :+1:

Slamdunk:
```java
public class AlphaTestCase  extends TestCase
{
  final protected void setUp() throws Exception {
    alphaProjectSetup();
    doSetUp();
  }    
  protected void doSetUp() throws Exception {        
  }
}
```
> While this works, it runs into the problem of confusing anyone who is familiar JUnit. Every project they've been on, every book they've read says they should override setUp, not this new-fangled doSetUp. This is a good case to use final, since there's such a high chance that people will get confused. But even with final, the confusion a different setup method causes is very painful.

It can be also read as: it is confusing to have two methods, but you if need to, at lest you enforce consistency between abstract classes end implementations.

keradus:
Thank you @Slamdunk.

PR #2672, opened 2017-04-03
Bump symfony/* deps

keradus:
requires #2568


keradus:
cc @sstok

SpacePossum:
yeah, lets go! :)    :+1:

julienfalque:
Wouldn't `^2.8|^3.0` be better? As Symfony 3.0 is just Symfony 2.8 with deprecated code removed, if your code is compatible with 3.0, then it is automatically compatible with 2.8.

SpacePossum:
The removal of deprecated code in 3.0 is a good reason to jump to it :)
It makes sure our code doesn't use deprecated SF code, but also that users of the code in this repo are.

>Should the code related to old versions of Symfony be dropped in this PR as well?

where needed yes, everything else can be done all in time

(note to self; https://github.com/symfony/symfony/pull/17255 can now be used in the tests)

julienfalque:
The test suite runs on 3.x so deprecated code from 2.x can't be used anyway. Removing the `^2.8` part of the constraint would force every side project to update to 3.x even if their code is compatible with 2.8. And I don't think keeping 2.8 would be a maintenance burden.

glensc:
@julienfalque this means such projects are forced to use .phar version of the tool, or composer global. so i agree `^2.8|^3.0` would had been better.

keradus:
That's exactly why (deps issues) we do have `phar` releases as well.
If you for some reason have to stack with old deps and cannot upgrade them, you may not integrate PHP CS Fixer directly in project composer file or just stack with PHP CS Fixer like with other deps - keeping it on 2.2.x level

PR #2674, opened 2017-04-05
SemicolonAfterInstructionFixer - Fix case where block ends with an opening curly brace

ntzm:
`<?php if ($x) { ?>` was turned into `<?php if ($x) {; ?>`


julienfalque:
👍 (AppVeyor failures unrelated, probably due to recent Symfony releases).

SpacePossum:
Nice catch;
however this should still be fixed:

```php
<?php function A(){  } ?>
```

so I think the code needs to be updated.

ntzm:
You're right @SpacePossum

keradus:
Thank you @ntzm.

SpacePossum:
Thanks @ntzm solid fix!

PR #2677, opened 2017-04-07
README: Correct progressbar option support

buffcode:
Is incorrectly `evaluating` instead of `estimating`


keradus:
Thank you @buffcode.

Issue #2650, opened 2017-03-29
indentation_type fixer kills switch spacing


rbayliss:
Running the indentation_type fixer with an indentation of '  ' (two spaces) yields the following results:

**Input**
```php
<?php 

switch($myvar) {
  case 1:
    echo 'Got case 1';
    break;
  default:
    echo 'Got default case';
}
```

**Output**
```php
switch($myvar) {
  case 1:
  echo 'Got case 1';
  break;
  default:
  echo 'Got default case';
}
```
Is this expected?  If so, is there another fixer that can correct this spacing?  I'd like it to look just like the input.

SpacePossum:
Hi!
Thanks for the report. Could you please provide the PHP version you are using together with version of the Fixer and the configuration of the Fixer?
This information will help us to answer your question, thanks!

rbayliss:
Yep, sorry for omitting that.  PHP 7.0.12, and the bleeding edge of php-cs-fixer: 9e3302c8fa9c0c622f839fa5f9233f428521011c

**.php_cs:**
```php
<?php

$finder = PhpCsFixer\Finder::create()
  ->in(__DIR__.'/scrap')
  ->files('*');

$config = \PhpCsFixer\Config::create()
  ->setFinder($finder)
  ->setIndent('  ')
  ->setUsingCache(FALSE)
  ->setRules([
    'indentation_type' => TRUE,
  ]);

return $config;
```

ntzm:
I believe this is because the indentation fixer only expects the indentation to be either 4 spaces, tabs, or a mixture of spaces and tabs ([reference](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Whitespace/IndentationTypeFixer.php#L78)).

rbayliss:
Yeah, that's what I'm seeing too... this feels like a bug, right?  I'm willing to submit a PR, but I'm not really sure where to start with this one.

medgoprot:
Someone should close this as the fix was already merged.

keradus:
sure thing, thanks for pointing this out!

PR #2679, opened 2017-04-07
[WIP] First attempt at solving double indent issue #2650

ntzm:
Issue: #2650

This is an interesting one, because we can't match just double spaces, because that would break files with four spaces for indentation. I decided to go with the whitespace config value, but I don't think this is a good idea because if a user wanted to change from double space to tabs or four spaces, it wouldn't pick it up correctly.

One solution that could work making indentation_type configurable, and allowing the user to pass in the "from" indentation (e.g. if a user wanted to convert from 2 spaces to 4 spaces: `indentation_type => ['from' => '  ']],` or something like that. Is that a good route to go down? This option would also work for 8 space or more indent (😱)

Thoughts & feedback would be greatly appreciated


ntzm:
Closing because I don't have the time to do this right now

PR #2682, opened 2017-04-10
DoctrineAnnotationIndentationFixer - fix handling nested annotations

### LINKED ISSUE #2681: Doctrine annotation indenter incorrectly adds indents after annotations containing other annotations
The Doctrine annotation indenter behaves unexpectedly for this input, which I've added as a new test in my fork: https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/master...edhgoose:fix-doctrine-annotation-indenter

```
/**
 * @Foo(x={
 * @Bar
 * })
 * @Foo\z
 */
```

I'd expect the `@Bar` to be indented and the rest to be left as is.

This is the output, note `@Bar` is indented correctly but the following annotation `@Foo\z` becomes intended by 4 spaces.

```
/**
 * @Foo(x={
 *     @Bar
 * })
 *     @Foo\z
 */
```

It looks like the annotator doesn't "reset" after closing the `@Bar` annotation.

PHP-CS-Fixer version is master (0a19b49934ad624845b296dc2a584444d6e44787).

PHP version:

```
PHP 7.0.11 (cli) (built: Sep 30 2016 12:31:36) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Xdebug v2.4.1, Copyright (c) 2002-2016, by Derick Rethans
```

I've had a go at fixing it, but I'm struggling to find the cause of the issue. Any assistance would be great and I'll gladly submit a fix.

julienfalque:
Fixes #2681.

@edhgoose I cherry-picked your commit, thanks for the test!

### DIFF
--- a/src/Fixer/DoctrineAnnotation/DoctrineAnnotationIndentationFixer.php
+++ b/src/Fixer/DoctrineAnnotation/DoctrineAnnotationIndentationFixer.php
@@ -41,19 +41,29 @@ final class DoctrineAnnotationIndentationFixer extends AbstractDoctrineAnnotatio
      */
     protected function fixAnnotations(Tokens $tokens)
     {
-        $previousLineBracesDelta = 0;
-        $indentLevel = 0;
-        foreach ($tokens as $index => $token) {
+        $annotationPositions = array();
+        for ($index = 0, $max = count($tokens); $index < $max; ++$index) {
+            if (!$tokens[$index]->isType(DocLexer::T_AT)) {
+                continue;
+            }
+
             $annotationEndIndex = $tokens->getAnnotationEnd($index);
             if (null === $annotationEndIndex) {
                 return;
             }
 
+            $annotationPositions[] = array($index, $annotationEndIndex);
+            $index = $annotationEndIndex;
+        }
+
+        $previousLineBracesDelta = 0;
+        $indentLevel = 0;
+        foreach ($tokens as $index => $token) {
             if (!$token->isType(DocLexer::T_NONE) || false === strpos($token->getContent(), "\n")) {
                 continue;
             }
 
-            if (!$this->indentationCanBeFixed($tokens, $index)) {
+            if (!$this->indentationCanBeFixed($tokens, $index, $annotationPositions)) {
                 continue;
             }
 
@@ -106,16 +116,18 @@ final class DoctrineAnnotationIndentationFixer extends AbstractDoctrineAnnotatio
     }
 
     /**
-     * @param Tokens $tokens
-     * @param int    $newLineTokenIndex
+     * @param Tokens            $tokens
+     * @param int               $newLineTokenIndex
+     * @param array<array<int>> $annotationPositions Pairs of begin and end indexes of main annotations
      *
      * @return bool
      */
-    private function indentationCanBeFixed(Tokens $tokens, $newLineTokenIndex)
+    private function indentationCanBeFixed(Tokens $tokens, $newLineTokenIndex, array $annotationPositions)
     {
-        $atIndex = $tokens->getPreviousTokenOfType(DocLexer::T_AT, $newLineTokenIndex);
-        if (null !== $atIndex && $tokens->getAnnotationEnd($atIndex) > $newLineTokenIndex) {
-            return true;
+        foreach ($annotationPositions as $position) {
+            if ($newLineTokenIndex >= $position[0] && $newLineTokenIndex <= $position[1]) {
+                return true;
+            }
         }
 
         for ($index = $newLineTokenIndex + 1, $max = count($tokens); $index < $max; ++$index) {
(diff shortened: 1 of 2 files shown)


edhgoose:
Sweet :) Thank you for fixing it!

src/Fixer/DoctrineAnnotation/DoctrineAnnotationIndentationFixer.php
```diff
@@ -24,6 +24,11 @@
 final class DoctrineAnnotationIndentationFixer extends AbstractDoctrineAnnotationFixer
 {
     /**
+     * @var array
```
keradus: not true until `fixAnnotations` is called.
I would remove the property and pass it directly as method parameter into `indentationCanBeFixed`

julienfalque: Done.

src/Fixer/DoctrineAnnotation/DoctrineAnnotationIndentationFixer.php
```diff
@@ -108,14 +118,16 @@ private function getLineBracesDelta(Tokens $tokens, $index)
     /**
      * @param Tokens $tokens
      * @param int    $newLineTokenIndex
+     * @param array  $annotationPositions
```
keradus: what is the structure?
pairs of start/end indices ?

keradus:
@edhgoose, I understand that you confirm the fix works ?

src/Fixer/DoctrineAnnotation/DoctrineAnnotationIndentationFixer.php
julienfalque: Yep, `array<array<int>>`?

keradus: yep.
sadly, it's not easy to document non-structured elements like array of pairs :( please provide that info in param description

edhgoose:
Perfectly. Thank you @julienfalque and @keradus!!

edhgoose:
Actually, @julienfalque and @keradus there might be a case where this doesn't work as expected. I don't know if this is a different issue or not, but here's a couple of test cases which demonstrate the problem:

```
            array('
/**
 * @Foo(x={
 *     @Bar
 * })
 */', '
/**
 * @Foo(x={
 * @Bar})
 */'),
            array('
/**
 * @Foo(x={
 *     @Bar,
 *     @Charlie
 * })
 */', '
/**
 * @Foo(x={
 * @Bar,
 * @Charlie})
 */'),
            array('
/**
 * @Foo(
 *     x="1",
 *     y="2"
 * )
 */', '
/**
 * @Foo(
 * x="1",
 * y="2")
 */'),
```

I'm not exactly sure what correct behaviour should be, but it appears that the problem is that if the closing bracket of the annotation is on the same line as any other content, then that content is not indented. I'd argue personally that the end of the annotation should be put on the next line, but I don't know how easy that is to do?

Here's the output from PHPUnit for each of the cases:

```
 /**
  * @Foo(x={
- *     @Bar
- * })
+ * @Bar}) // @Bar should be indented here, and ideally the bracket on the next line
  */
 class FooClass
 {
 }

  /**
  * @Foo(x={
  *     @Bar,
- *     @Charlie
- * })
+ * @Charlie}) // @Charlie should be indented here, and ideally the bracket on the next line
  */

 /**
  * @Foo(
  *     x="1",
- *     y="2"
- * )
+ * y="2") // y should be indented here, and ideally the bracket on the next line
  */

```

julienfalque:
@edhgoose These results are actually the expected behavior: the fixer currently only changes the indentation, nothing else. Your examples all have data along with closing parentheses/braces, separating them must be done manually for the moment. Having a fixer for that would be great though.

The only change for the indentation fixer I can imagine is a new option to choose what to considerer when updating the indentation of such line (e.g. `mixed_lines` => `indent|unindent`).

edhgoose:
Ok, let me have a look at that. Would the intention be that any annotation that has something inside it (i.e. anything beyond `@Foo()`) would be split across multiple lines?

julienfalque:
> Would the intention be that any annotation that has something inside it (i.e. anything beyond `@Foo()`) would be split across multiple lines?

If you're talking about the new fixer: I don't really know. Should the arguments be always accross multiple lines? One per line like method arguments rules from PSR-2? Only if they already contain a line break? Or if they exceed a length limit? What about nested annotations? I think this deserves a dedicated new issue for discussion.

For the indentation one, I would add a new option to change the behavior regarding mixed lines (lines that contain closing parentheses/braces *and* other things). Current behavior is to remove one indentation level for those lines as if they only had the closing tokens. This option would allow you to configure the fixer so that it keeps the same indentation level instead. So your first example would be fixed like this:
```diff
 /**
  * @Foo(x={
- * @Bar})
+ *     @Bar})
  */,
```

keradus:
Thank you @julienfalque.

PR #2684, opened 2017-04-11
BracesFixer - new options for braces position after control structures and anonymous constructs

aidantwoods:
Personally, I prefer my braces to be consistent across classes, functions, and control structures. This adds an extra option to apply those fixes which is not enabled by default.

> Whether the opening brace should be placed on "next" or "same" line after control structures not covered by position_after_functions_and_oop_constructs.


src/Fixer/Basic/BracesFixer.php
```diff
@@ -169,9 +169,17 @@ protected function createConfigurationDefinition()
             ->getOption()
         ;
 
+        $positionAfterControlStructures = new FixerOptionBuilder('position_after_control_structures', 'whether the opening brace should be placed on "next" or "same" line after control structures not covered by position_after_functions_and_oop_constructs.');
```
julienfalque: I would remove the _"not covered by position_after_functions_and_oop_constructs"_ part.

README.rst
```diff
@@ -250,6 +250,10 @@ Choose from the list of available rules:
 
   - ``allow_single_line_closure`` (``bool``): whether single line lambda notation
     should be allowed; defaults to ``false``
+  - ``position_after_control_structures`` (``'same'``): whether the opening brace
```
julienfalque: Looks like there an issue with README helper: the option has two allowed values but only one is listed here. This is not related to this PR as the `position_after_functions_and_oop_constructs` option has the same issue. I'll investigate that.

julienfalque:
Code looks good, please add test cases that cover new option.

src/Fixer/Basic/BracesFixer.php
aidantwoods: In theory that's what it'll do (since I'm applying to tokens returned by `getControlTokens`),  should exclude things covered by position_after_functions_and_oop_constructs as far as I can tell?

julienfalque: Yes, IMO option names are explicit enough about that. But having description explicit as well won't hurt after all, so leave the sentence as is if you prefer :)

README.rst
keradus: I have subtracted it to separated issue: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2685
just to not forgot about it

aidantwoods:
I'm having a little trouble getting tests to pass:
https://github.com/aidantwoods/PHP-CS-Fixer/blob/6834d70dde748a89a250613cd1a5ab2f062c8732/src/Fixer/Basic/BracesFixer.php#L497

the function `fixMissingControlBraces` attempts to correct whitespace on existing braces, causing whitespace to be changed, and then subsequently changed back by my new option. However, this causes a test failure because `$tokens` has now been marked as changed (even though the text ends up being the same).

> `Tokens collection built on expected code must not be marked as changed after fixing.`

The whitespace correction done by `fixMissingControlBraces` should be moved to `fixSpaceAroundToken` as far as I can tell?

aidantwoods:
In one of the new test cases, fixMissingControlBraces causes the following change which causes the failure

```diff
--- Original
+++ New
@@ @@
 <?php
-if (true)
-{
+if (true) {
     $a = 1;
 }
-else
-{
+else {
     $b = 2;
 }
```

keradus:
> Well that was much easier than expected

> In one of the new test cases, fixMissingControlBraces causes the following change which causes the failure

We are all awaiting for green Travis status ;)

aidantwoods:
> We are all awaiting for green Travis status ;)

Hehe, it seems I spoke a little too soon with
> Well that was much easier than expected

Could someone advise on the best way to proceed with: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2684#issuecomment-293303840

Being a little new to the repo 😉:

I couldn't really see what the intent of the brace shift was (as-in why it was done in a function that is apparently for adding control braces, and not in one responsible for correcting placement), or where the offending line in that function should be moved to (i.e. moving the line speaks to intent of why it was put there in the first place).

aidantwoods:
I mean, I could just special case the line that's causing trouble so that it doesn't run when my new configuration option is enabled.

Just it tripped me up since the reason for the error wasn't readily apparent – since later functions (actually responsible for correcting brace *position*) were reverting the change to position that this function made.

aidantwoods:
@keradus All green on travis :D

Have rebased on master and just special cased that offending line (as mentioned above)

I'm not sure what's up with the fabbit.io check

keradus:
> I'm not sure what's up with the fabbit.io check

ignore it, they need to update PHP CS Fixer version they are using

aidantwoods:
Okay, so just to give a little bit more info on the intent of this feature (so any bugs can be identified later).

The aim of this setting is to make all control structures behave with the same rules that we have for functions (in PSR2), mostly this is just placing the brace on the next line e.g.

screen shot 2017-04-27 at 21 59 23

Though (as with functions) if the "body of arguments" spans multiple lines, then both the closing brace to the body of arguments and the opening brace for the main body are placed on the next line, separated by a single space:
e.g. with the `if` statement here

screen shot 2017-04-27 at 22 04 22

Anything that doesn't conform to the rules for function opening brace placement when applied across all control structures should be considered a bug.

At present, I don't think that PSR dictates a rule for styling functions with return type hints and a multiline "body of arguments", the already existing behaviour appears to keep the main function body opening brace on the same line as the type hint and the closing brace for the body of arguments. E.g.

```php
<?php

function foo(
    $bar,
    $baz,
    $boo
) : int {
}
```

I haven't (intentionally) modified the behaviour for this, though I think in a case like this, the placing the opening brace to the main body on the next line may be more in sync with the "next line" look

```php
<?php

function foo(
    $bar,
    $baz,
    $boo
) : VeryLongInterfaceName
{
}
```

Though I can't see anywhere obvious to make that change, and it's quite a subtle difference anyway when compared with the changes already made by this option (and arguably this isn't really within this option's scope anyway, so should probably be a separate config setting).

Given all that, I think I'm happy with everything this option does now (as long as there are no bugs anyone can see), or let me know if you'd like more test cases or something like that.

julienfalque:
> At present, I don't think that PSR dictates a rule for styling functions with return type hints and a multiline "body of arguments"

[Proposed PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments) currently mentions this case and does not seem to strictly forbid the opening brace to be on a new line.

> I think in a case like this, the placing the opening brace to the main body on the next line may be more in sync with the "next line" look

Like you said: this would require a new dedicated option as it changes the default behavior. It should be done in a separate PR, but if you're willing to work on it, go ahead :)

aidantwoods:
Just while I'm poking around that test file, looks like `provideFixCommentBeforeBraceCases` has a few typos (latter tests have four items in the array, so I think the config variable is never actually used).

keradus:
great catch. boom!
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2731

aidantwoods:
@julienfalque right, https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2684/commits/4f8fe4c44fd2ba45a0431cd420785b370308b941 should complete the requested changes

keradus:
> Good shout 

@julienfalque , don't shout at people ;)

tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -1515,6 +1651,23 @@ function & lambda()
...
+    };
+}',
+                '<?php
+
+function & lambda()
+{
+    return function () {
+    };
+}',
+                self::$configurationOopPositionSameLine + self::$configurationCtrlStructPositionNextLine,
```
aidantwoods: This test is probably a good visualisation of what exactly this new option applies to – it covers everything that isn't covered by `position_after_functions_and_oop_constructs`. So this test takes something that would be unchanged by default `@PSR2` and applying the non-default configs of both this new option and `position_after_functions_and_oop_constructs` acts to invert the brace behaviour:
Because `position_after_functions_and_oop_constructs` specifically selects (non-anonymous) classes and functions, and `position_after_control_structures` will very specifically select the anonymous ones (as well as other control structures)

aidantwoods:
@julienfalque 

> Like you said: this would require a new dedicated option as it changes the default behavior. It should be done in a separate PR, but if you're willing to work on it, go ahead :)

Right, I've gone ahead and implemented this too – the PR is targeted at this branch, just so we don't need to needlessly resolve conflicts (much of the same lines of code are being edited).

Currently here: https://github.com/aidantwoods/PHP-CS-Fixer/pull/1

But I'll create a separate PR targeting this repo's master once this branch is merged

aidantwoods:
@julienfalque, @keradus anything else that needs doing before merge?

aidantwoods:
@julienfalque 
> Anonymous functions and classes are not control structures.

Indeed ;-)

I'd always assumed that PSR2 considered them control structures (given there was no [voting option](http://www.php-fig.org/psr/psr-2/#a3-survey-results) for anonymous functions, and they follow the same brace rules as control structures wrt. space after the `function` keyword, `function (` (much like `if (`, `for (` etc...) and brace on the same line etc...).

I'll split it into two options for clarity though :)

aidantwoods:
@julienfalque

Anonymous and control structures are decoupled!

tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -3910,7 +4124,67 @@ public function bar() {
...
+    });
+}',
+                '<?php
+if (1) {
+  $message = (new class() extends Foo
+  {
+    public function bar() { echo 1; }
+  });
+}',
+                self::$configurationOopPositionSameLine + self::$configurationCtrlStructPositionNextLine + self::$configurationAnonymousPositionNextLine,
```
aidantwoods: This test encompasses the effect of all the brace moving options being applied, for an `inverted` look on what you'd expect from a PSR2 document.

aidantwoods:
@julienfalque any more changes that need doing before merge? 😉

src/Fixer/Basic/BracesFixer.php
```diff
@@ -169,9 +169,25 @@ protected function createConfigurationDefinition()
...
         ;
 
+        $positionAfterControlStructures = new FixerOptionBuilder('position_after_control_structures', 'whether the opening brace should be placed on "next" or "same" line after control structures.');
+        $positionAfterControlStructures = $positionAfterControlStructures
+            ->setAllowedValues([self::LINE_NEXT, self::LINE_SAME])
+            ->setDefault(self::LINE_SAME)
+            ->getOption()
+        ;
+
+        $positionAfterAnonymousStructures = new FixerOptionBuilder('position_after_anonymous_structures', 'whether the opening brace should be placed on "next" or "same" line after anonymous constructs (anonymous classes and lambda functions).');
```
aidantwoods: Done!

aidantwoods:
Well that made for some very careful rebasing 😜 

3 conflicts for this o.0

screen shot 2017-05-08 at 18 50 34

julienfalque:
Woops :D

aidantwoods:
What's the best thing to do about https://github.com/aidantwoods/PHP-CS-Fixer/pull/1?

It's kinda queued behind this one (to avoid merge conflicts – files are changed in many of the same places).

Shall I submit a separate PR as is, or wait out this one's merge? (Ideally this next one would go through in the same release 😉)

keradus:
This PR will be merged as soon as Travis will be green.
Please do not push anything to this PR now.
New PR for sure, doesn't matter would you wait for this one to be merged or not (and then rebase when this PR will be merged).

aidantwoods:
> This PR will be merged as soon as Travis will be green.

Great! :D I'll just wait it out then!

keradus:
20 minutes for php 5.6 and counting...

aidantwoods:
Yeah... that's been happening a lot recently: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2893#issuecomment-312325843

Weird, as it's so much faster on the other builds...

keradus:
Thank you @aidantwoods.

SpacePossum:
thanks @aidantwoods , nice new options :D

Issue #2680, opened 2017-04-07
single_import_per_statement config being ignored


pilgerone:
The PHP version you are using:
`PHP 7.1.1 (cli) (built: Jan 18 2017 18:51:14) ( ZTS MSVC14 (Visual C++ 2015) x86 )`

PHP CS Fixer version you are using:
`PHP CS Fixer version 2.2.0 by Fabien Potencier and Dariusz Ruminski (d6f1742)`

The command you use to run the fixer (and the configuration file you are using if any).
`php \php-cs-fixer-v2.phar fix file.php --verbose --config=path\to\.php_cs.dist`

Config:
```
<?php
return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setRules(array(
        '@PSR2' => true,
        '@Symfony' => true,
        '@Symfony:risky' => true,
        'array_syntax' => array('syntax' => 'short'),
        'combine_consecutive_unsets' => true,
        // one should use PHPUnit methods to set up expected exception instead of annotations
        'general_phpdoc_annotation_remove' => array('expectedException', 'expectedExceptionMessage', 'expectedExceptionMessageRegExp'),
        'heredoc_to_nowdoc' => true,
        'linebreak_after_opening_tag' => true,
        'no_extra_consecutive_blank_lines' => array('break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block'),
        'no_short_echo_tag' => true,
        'no_unreachable_default_argument_value' => true,
        'no_useless_else' => true,
        'no_useless_return' => true,
        'ordered_class_elements' => true,
        'ordered_imports' => true,
        'php_unit_strict' => true,
        'phpdoc_add_missing_param_annotation' => true,
        'phpdoc_order' => true,
        'semicolon_after_instruction' => true,
        'strict_comparison' => true,
        'strict_param' => true,
        'single_import_per_statement' => false,
    ))
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->exclude('tests/Fixtures')
            ->in(__DIR__)
    )
;
```

A minimal sample of valid PHP code that is not fixed correctly (if applicable).
```
use some\namespace\{
    Class1,
    Class2,
    Class3
};
```

The fixed version of that code after you run the fixer:
```
use some\namespace\Class1;
use some\namespace\Class2;
use some\namespace\Class3;
```

And the version you expected:
```
[...]

PR #2687, opened 2017-04-12
Fix rule sets overwriting rules defined independently

pilgerone:
Fixing the issue described in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2543 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2680.
Rewritten the resolveSet() method to be more concise by using recursion.

### DIFF
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -271,36 +271,25 @@ final class RuleSet implements RuleSetInterface
     private function resolveSet()
     {
         $rules = $this->set;
-        $hasSet = null;
 
-        // expand sets
-        do {
-            $hasSet = false;
+        $resolveRulesRecursively = function ($ruleSet, $parentValue) use (&$resolveRulesRecursively) {
+            $parsedRules = [];
 
-            $tmpRules = $rules;
-            $rules = array();
+            foreach ($ruleSet as $name => $value) {
+                $value = $parentValue ? $value : false;
 
-            foreach ($tmpRules as $name => $value) {
-                if (!$hasSet && '@' === $name[0]) {
-                    $hasSet = true;
-                    $set = $this->getSetDefinition($name);
-
-                    foreach ($set as $nestedName => $nestedValue) {
-                        // if set value is false then disable all fixers in set, if not then get value from set item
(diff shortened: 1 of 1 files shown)


SpacePossum:
Thanks for your PR!
However this is duplicate to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2545 and same issue as https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2543. Both are still discussed so lets see where the discussion ends before we start reviewing solutions.

pilgerone:
You're right. Sorry about this. Kinda new to this collaboration thing. :confused:

I've pinpointed the problem in the code. It's a bit hard to explain exactly what its doing, but it's definitely overwriting the rules set further below on the configuration file in some cases.

I'll read the other PR thread thoroughly and see if I can come up with a test to show what I mean.

keradus:
I think we got what your issue is and I confirm that it is a bug.
The thing is that we need to decide how we want to fix more general bug, having your case in mind.

pilgerone:
If I understood correctly, the problem is how to treat overwriting rules that contain arrays (e.g.: `rule = ['a','b','c']`) being overwritten (e.g.: by something like `rule = ['a']`).  Is that right?

keradus:
no, it's about disabling/enabling rule, changing config is slightly other thing. let us not spread discussion about 1 thing into 2 threads

keradus:
@pilgerone , @SpacePossum 
is there something left here after merge of #2545 ?

SpacePossum:
I think @pilgerone and I thought the same thing as this PR and mine (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2545) seem to work in a similar way :)
However, I would very much like to see this confirmed.

pilgerone:
Yeap. All good after the latest release.

I guess we can safely discard this one in favor to @SpacePossum 's.  :+1:

SpacePossum:
Thanks @pilgerone for this PR, your thoughts (and comments) on my PR and testing/reporting back on this one. 
Happy to see this is resolved, thanks!

PR #2691, opened 2017-04-13
Add Behat annotation to ignored annotations for Doctrine fixers

stof:
This avoids breaking the Behat annotation when fixing Doctrine annotations

### DIFF
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -177,7 +177,7 @@ abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements 
                 'BeforeScenario',
                 'AfterScenario',
                 'BeforeStep',
-                'AfterStep'
+                'AfterStep',
 
                  // PHPCheckStyle
                  'SuppressWarnings',
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -165,6 +165,20 @@ abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements 
                  'ticket',
                  'uses',
 
+                // Behat
+                'Given',
+                'When',
+                'Then',
+                'Transform',
+                'BeforeSuite',
+                'AfterSuite',
+                'BeforeFeature',
+                'AfterFeature',
+                'BeforeScenario',
+                'AfterScenario',
+                'BeforeStep',
+                'AfterStep'
+
                  // PHPCheckStyle
                  'SuppressWarnings',
 


julienfalque:
Could you please update the ignored tags cases in the [Doctrine Annotation fixers tests](https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master/tests/Fixer/DoctrineAnnotation) to include these new ones?

I'm not sure we should have those in the default list though.

stof:
Well, if we don't have them in the default, we break Behat contexts by default when running the fixer. Why is it better than breaking PHPUnit exception assertions ?
And *adding* new ignored annotations is a pain, because it requires copying all the default ignored ones too. We cannot easily merge the array (and there is no way to access the default array easily).

julienfalque:
In the beginning I only wanted to have the same ignored tags as Doctrine Annotation lib itself, but PHP CS Fixer uses PHPUnit, probably just like most of the projects that uses this tool so I added PHPUnit tags to the list for convenience.

> And adding new ignored annotations is a pain, because it requires copying all the default ignored ones too.

I agree it would be nice to be able to merge new values with default ones.

stof:
> I agree it would be nice to be able to merge new values with default ones.

I just though about a way to do it: expose the default array through a static method (a static property would not fit it as the default array would be mutable). This way, the config file can use ``array_merge(AbstractDoctrineAnnotationFixer::getDefaultIgnoredAnnotation(), array('Transform'))``

julienfalque:
Would work for this fixer, but something more generic that could be used for any array option of any fixer would be better IMO.

stof:
Well, this would require stopping to use the OptionResolver of Symfony to do the processing of options, as it does not allow merging an array value with the default value.

keradus:
normalizer could change value of passed option, eg add default values to passed array

stof:
@keradus but then you loose the ability to unignore some. So this should not be added blindly to all array configurations

keradus:
@julienfalque , @stof 
guys, please finish this PR ;)

julienfalque:
@stof Would this PR remain relevant if #2700 is merged?

stof:
@julienfalque probably not. To be tried again once the other PR is merged.

keradus:
#2700 is merged. please check it out @julienfalque , @stof

keradus:
ping @julienfalque , @stof

keradus:
ping @julienfalque , @stof

keradus:
ping @julienfalque , @stof

julienfalque:
Let's close. The initial issue has been fixed so adding Behat annotations to ignored list may not be required anymore anyway.

keradus:
Cool.
@stof, in any case, feel free to reopen adding more information

PR #2700, opened 2017-04-16
Fix Doctrine Annotation end detection

### LINKED ISSUE #2692: Doctrine annotation space fixer is breaking things
The fixer is clearing tokens of type `T_NONE`. But it does not check first that these tokens are actually containing spaces only.
This is the main reason why Behat annotations are required to be ignored by this fixer (see #2691): the fixer is clearing tokens containing `^` or `+` when encountering a Step regex with a brace in it, thus breaking the regex.

julienfalque:
Fixes #2692.


keradus:
Thank you @julienfalque.

Issue #2697, opened 2017-04-14
no_extra_consecutive_blank_lines removes line between cases when it throws


stof:
When the last line of a `case` is using `throw` rather than `return`, `break` or `continue`, the `no_extra_consecutive_blank_lines` removes the empty line between this case and the next one. But it does not do it for other constructs. This looks like a bug to me. there is no reason to forbid empty lines between cases only when throwing.

SpacePossum:
Hmm sounds odd indeed, do you have sample?

stof:
Existing code (part of a bigger file defining a class):

```php
    /**
     * @return Node
     */
    private function parsePrimaryExpression()
    {
        $token = $this->stream->getCurrent();

        switch ($token->getType()) {
            case Token::NAME_TYPE:
                $this->stream->next();

                if ('(' === $this->stream->getCurrent()->getValue()) {
                    if (!isset($this->functions[strtolower($token->getValue())])) {
                        $errorKey = count($this->functions) ? 'formula.unrecognized_function' : 'formula.no_registered_functions';

                        throw new SyntaxError($errorKey, array(
                            '%name%' => strtoupper($token->getValue()),
                            '%functions_list%' => implode(', ', array_map('strtoupper', array_keys($this->functions))),
                        ));
                    }

                    $arguments = $this->parseArguments();
                    $node = new FunctionNode($token->getValue(), $arguments);

                    //Throws a Syntax Error if the node is not validated
                    $this->functions[$node->getAttribute('name')]->validateNode($arguments, $this->context);

                    return $node;
                }

                throw new SyntaxError('formula.missing_arguments_opening_parenthesis');

            case Token::NUMBER_TYPE:
            case Token::STRING_TYPE:
                $this->stream->next();

                return new ConstantNode($token->getValue());

            default:
                throw new SyntaxError('formula.unrecognized_characters');
        }
    }
```

and here is the generated diff:

```diff
                 }
 
                 throw new SyntaxError('formula.missing_arguments_opening_parenthesis');
-
             case Token::NUMBER_TYPE:
             case Token::STRING_TYPE:
                 $this->stream->next();
```

It removes the line after `throw`, but not after `return` for the next case

SpacePossum:
The removal of the line after the `throw` and `return` is done by configuration and I consider it therefor not a bug.
[...]

Issue #2694, opened 2017-04-13
Indention fixing - When removing a useless else, lines are not unindented properly


stof:
When the code inside the `else` spans multiple lines, the lines are not unindented properly.

Current diff:

```diff
-            } else {
-                return $this->redirectToRoute('space_challenge_manager_teams_pr
ize_sequence_details', array(
+            }
+
+            return $this->redirectToRoute('space_challenge_manager_teams_prize_sequence_details', array(
                     'space' => $space->getId(),
                     'team' => $team->getId(),
                     'level' => $level->getId(),
                     'sequence' => $prizeData['sequence']->getId(),
                 ));
-            }
```

Expected diff:

```diff
-            } else {
-                return $this->redirectToRoute('space_challenge_manager_teams_pr
ize_sequence_details', array(
-                    'space' => $space->getId(),
-                    'team' => $team->getId(),
-                    'level' => $level->getId(),
-                    'sequence' => $prizeData['sequence']->getId(),
-                ));
             }
+
+            return $this->redirectToRoute('space_challenge_manager_teams_prize_sequence_details', array(
+                'space' => $space->getId(),
+                'team' => $team->getId(),
+                'level' => $level->getId(),
+                'sequence' => $prizeData['sequence']->getId(),
+            ));
```

SpacePossum:
The `NoUselessElseFixer` doesn't do this by design as I didn't want to copy-paste the logic of the `braces` and `indentation_type` (DRY/single responsibility principle). 
However the priority (`useless else` must run before indentation fixers) is guaranteed.
So either;
- the priorities are messed up (would make the issue a bug)
- the indentation fixers are not run at all by configuration (would make the issue invalid)
- the indentation fixers do not fix the code after `NoUselessElseFixer` is done (would make the issue a FR request)

I expect the 3rd.

stof:
indentation fixers are running in my case. This is my config:

```php

return PhpCsFixer\Config::create()
    ->setRules(array(
        '@Symfony' => true,
        '@Symfony:risky' => false,
        'strict_param' => false,
        'array_syntax' => array('syntax' => 'long'),
        'doctrine_annotation_braces' => true,
        'doctrine_annotation_indentation' => true,
        'doctrine_annotation_spaces' => true,
        'heredoc_to_nowdoc' => true,
        'no_useless_else' => true,
        'ordered_class_elements' => true,
        'ordered_imports' => true,
    ))
[...]

PR #2701, opened 2017-04-18
NoExtraConsecutiveBlankLinesFixer - Add more configuration options related to switch statements

SpacePossum:
following https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2697

### DIFF
--- a/tests/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixerTest.php
+++ b/tests/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixerTest.php
@@ -933,6 +933,135 @@ class Foo
         ];
     }
 
+    /**
+     * @param array       $config
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideSwitchCases
+     */
+    public function testInSwitchStatement(array $config, $expected, $input = null)
+    {
+        $this->fixer->configure(['tokens' => $config]);
+        $this->doTest($expected, $input);
+    }
+
+    public function provideSwitchCases()
+    {
+        return [
+            [
+                [
+                    'break',
+                    'continue',
+                    'extra',
+                    'return',
+                    'throw',
+                ],
+                '<?php
+                    /** a  */
+                    switch ($a) {
+                        case 1:
+                            break;
+                        case 2:
+                            continue;
+                        case 3:
+                            return 1;
+                        case 4:
+                            throw $e;
+                        case 5:
+                            throw new \Exception();
+                        case Token::STRING_TYPE:
+                            echo 123;
+
+                            return new ConstantNode($token->getValue());
+                        case 7:
+                            return new ConstantNode($token->getValue());
+                        case 8:
+                            return 8;
+                        default:
+                            echo 1;
+                    }',
+                '<?php
+                    /** a  */
+                    switch ($a) {
+                        case 1:
+                            break;
+
+                        case 2:
+                            continue;
+
+                        case 3:
+                            return 1;
+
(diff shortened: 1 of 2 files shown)


tests/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -952,6 +952,135 @@ public function provideMessyWhitespacesCases()
...
+    
+                            return new ConstantNode($token->getValue());
+                        case 7:
+                            return new ConstantNode($token->getValue());
+                        case 8:
+                            return 8;
+                        default:
+                            echo 1;
+                    }',
+                '<?php
```
SpacePossum: **@stof I've tried to recreate the issue you reported (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2697) here, however it does not fail for me. Can you confirm you have both `throws` and `return` configured on the fixer?**

stof:
@SpacePossum my config is at https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2694#issuecomment-293953049 and I'm using the 2.2.1 release

SpacePossum:
Your configuration for the Fixer resolves to the `@Symfony` set definition;
```
'curly_brace_block',
'extra',
'parenthesis_brace_block',
'square_brace_block',
'throw',
'use'
```
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/RuleSet.php#L82)

Which explains `but line is not removed after return in this case.`, as `return` is not configured.
The reason why `throw` is in as default but `return` is not, is because we asked @ the symfony repo what they would like as default;
https://github.com/symfony/symfony/pull/19167

We never got around making it possible to add the config options specific for switch statements, so I picked the safe option (i.e. do not fix).
I don't see this feature to be added any time soon TBH, but maybe someone wants to follow op on it.

stof:
Well, I said `-1 for return in a switch` in my comment, but I would have added `-1 for throw in a switch` if I had thought about it too. What I was -1 about was the fact of removing lines between cases due to the last statement of the previous case

SpacePossum:
Yeah, looking back on the SF PR things may not have been as clear as they could've been.
For SF project it might be better to configure the fixer to `extra` only in their config file, so cases within `switch` statements are not touched. At least until we've the options to handle `switch` statements in another way.
To sum up the issue; this is a configuration issue and feature request for dealing with `switch` statements in another matter.
This PR is still relevant IMO as it adds new configuration options that I think are good, although these are not the requested options.

julienfalque:
👍

keradus:
:+1: but:
- please rebase
- please say, is it closing #2697 ?

SpacePossum:
* does not close https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2697
* will rebase soon :+1:

tests/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixerTest.php
```diff
@@ -1054,16 +1050,14 @@ public function provideSwitchCases()
                     default: return 1;}
 
 
-                    // above stays empty'
-                ,
+                    // above stays empty',
```
SpacePossum: fixer is bugged :(

SpacePossum:
tests are green

SpacePossum:
with :+1: from @keradus and approve of @julienfalque this on is RTM right?

keradus:
Thank you @SpacePossum.

Issue #2490, opened 2017-02-01
wrong files included


theofidry:
Hey,

For a project, I have the following configuration:

```php
// .php_cs.dist
<?php

declare(strict_types=1);

use PhpCsFixer\Config;
use PhpCsFixer\Finder;

$finder = Finder::create()
    ->in([
        __DIR__.'/src',
        __DIR__.'/tests'
    ])
;

return Config::create()
    ->setRiskyAllowed(true)
    ->setRules([...])
    ->setFinder($finder)
;
```

Yet, when I run it, I can see that the file `var/SymfonyRequirements.php` is affected. Is it the expected behaviour? (I wouldn't expect other directories than the one specified to be appended).

```
$ php -v
PHP 7.1.0-dev (cli) (built: Jan  3 2017 09:51:30) ( NTS )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.1.0-dev, Copyright (c) 1998-2016 Zend Technologies
```

PHP CS Fixer version you are using:
```
$ php-cs-fixer -V
PHP CS Fixer 2.0.0
```

GrahamCampbell:
What exact command are you using to execute php-cs-fixer please?

theofidry:
`$ bin/php-cs-fixer fix`

GrahamCampbell:
Hmm, how odd!

keradus:
`bin/php-cs-fixer fix -vvv` ?

theofidry:
```
$ bin/php-cs-fixer fix -vvv
string(55) "/path/to/project/.php_cs does not exist"
string(50) "pick /path/to/project/.php_cs.dist"
Loaded config default from "/path/to/project/.php_cs.dist".
Using cache file ".php_cs.cache".
[...]

PR #2705, opened 2017-04-19
Configuration resolvement - Do not fix the same path multiple times.

SpacePossum:
Detect if the same path has been passed multiple times, either by passing the same value multiple times are using a mix of absolute and relative paths.

We can deal with https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2490 here as well.
There are two calls to make on the PR and than it can be finished.

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -212,6 +212,7 @@ final class ConfigurationResolver
                     continue;
                 }
 
+                $configFile = realpath($configFile);
                 $config = include $configFile;
 
                 // verify that the config has an instance of Config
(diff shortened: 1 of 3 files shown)


tests/Console/ConfigurationResolverTest.php
```diff
@@ -379,6 +379,47 @@ public function testResolvePathRelativeB()
...
+        $resolver = new ConfigurationResolver(
+            $this->config,
+            array(
+                'path' => $paths,
+                'config' => __DIR__.'/../Fixtures/ConfigurationResolverConfigFile/case_4/my.php_cs',
+            ),
+            dirname(__DIR__)
+        );
+
+        $this->assertCount($expectedCount, $resolver->getPath());
```
julienfalque: Wouldn't it be better to check the exact result (i.e. `assertSame()`)? I think having only the count assertion makes the test weak.

SpacePossum: done

keradus:
Hi. Still failing on Windows

SpacePossum:
all green :)
@keradus like to hear you opinion on the two TODO's in the code

src/Console/ConfigurationResolver.php
```diff
@@ -358,23 +359,32 @@ public function getPath()
...
-                        if (!file_exists($absolutePath)) {
-                            throw new InvalidConfigurationException(sprintf(
-                                'The path "%s" is not readable.',
-                                $path
-                            ));
-                        }
+                $this->path = array();
+                foreach ($this->options['path'] as $path) {
+                    if ('' === $path) {
+                        // TODO  trigger deprecate/warning?
```
keradus: when it could happen currently ?

src/Console/ConfigurationResolver.php
```diff
@@ -358,23 +359,32 @@ public function getPath()
...
+                    $path = @realpath($absolutePath);
+                    if (false === $path) {
+                        throw new InvalidConfigurationException(sprintf(
+                            'The path "%s" is not readable.',
+                            $absolutePath
+                        ));
+                    }
+
+                    if (isset($this->path[$path])) {
+                        continue; // TODO trigger warning/deprecate?
```
keradus: this happens for duplicated path, right ?

julienfalque: Yep.

keradus: I would not break the flow if someone passed same file twice (now or later, so for sure not a deprecation).
It may be nice to show info on console (like we do with xdebug enabled), but nothing crucial, if we would not show anything and just fix file once - good for me as well.

src/Console/ConfigurationResolver.php
SpacePossum: >php-cs-fixer fix ""
>php-cs-fixer fix ''
php-cs-fixer fix "" '' --config=.php_cs

keradus: feel free to raise warning/deprecation/error

SpacePossum: :+1: I've added a deprecation warning.

keradus:
Code looks nice.
Yet, I want to play with outcome of code on your branch. (ergo manual tests)

SpacePossum:
yes, please, lets see this one goes better

SpacePossum:
ping @keradus

keradus:
sorry @SpacePossum , I will try to play with it next week :(

SpacePossum:
edge case not worth the risk

Issue #2690, opened 2017-04-13
Update README regarding risky tests


julienfalque:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2689#issuecomment-293840492

Also, the documentation does not mention the `setRiskyAllowed` method at all, it would be a nice addition.

SpacePossum:
The doc clearly states an instance of `PhpCsFixer\ConfigInterface` must be returned, which is a pointer to all its methods including `setRiskyAllowed`. I see no update needed on this topic.

However `Please be aware that this feature is experimental.` is something we can drop, as this is done and stable?

keradus:
it's stable

PR #2706, opened 2017-04-19
Update README

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2690

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1113,7 +1113,7 @@ fixed but without actually modifying them:
 
 Instead of using command line options to customize the rule, you can save the
 project configuration in a ``.php_cs.dist`` file in the root directory of your project.
-The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.2/src/ConfigInterface.php>`_
+The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.3/src/ConfigInterface.php>`_
 which lets you configure the rules, the files and directories that
 need to be analyzed. You may also create ``.php_cs`` file, which is
 the local configuration that will be used instead of the project configuration. It
--- a/README.rst
+++ b/README.rst
@@ -1112,8 +1112,8 @@ fixed but without actually modifying them:
     $ php php-cs-fixer.phar fix /path/to/code --dry-run
 
 Instead of using command line options to customize the rule, you can save the
-project configuration in a ``.php_cs.dist`` file in the root directory
-of your project. The file must return an instance of ``PhpCsFixer\ConfigInterface``,
+project configuration in a ``.php_cs.dist`` file in the root directory of your project.
+The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.2/src/ConfigInterface.php>`_
 which lets you configure the rules, the files and directories that
 need to be analyzed. You may also create ``.php_cs`` file, which is
 the local configuration that will be used instead of the project configuration. It
@@ -1168,8 +1168,7 @@ The following example shows how to use all ``Symfony`` rules but the ``full_open
     ;
 
 You may want to use non-linux whitespaces in your project. Then you need to
-configure them in your config file. Please be aware that this feature is
-experimental.
+configure them in your config file.
 
 .. code-block:: php
 
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -85,6 +85,16 @@ final class FixCommand extends Command
 
     /**
      * {@inheritdoc}
+     *
+     * Override here to only generate the help copy when used.
+     */
+    public function getHelp()
+    {
+        return HelpCommand::getHelpCopy();
+    }
+
+    /**
+     * {@inheritdoc}
      */
     protected function configure()
     {
@@ -107,7 +117,6 @@ final class FixCommand extends Command
                 )
             )
             ->setDescription('Fixes a directory or a file.')
-            ->setHelp(CommandHelp::getHelpCopy())
         ;
     }
 
--- a/src/Console/Application.php
+++ b/src/Console/Application.php
@@ -14,9 +14,11 @@ namespace PhpCsFixer\Console;
 
 use PhpCsFixer\Console\Command\DescribeCommand;
 use PhpCsFixer\Console\Command\FixCommand;
+use PhpCsFixer\Console\Command\HelpCommand;
 use PhpCsFixer\Console\Command\ReadmeCommand;
 use PhpCsFixer\Console\Command\SelfUpdateCommand;
 use Symfony\Component\Console\Application as BaseApplication;
+use Symfony\Component\Console\Command\ListCommand;
 
 /**
  * @author Fabien Potencier <fabien@symfony.com>
@@ -28,9 +30,6 @@ final class Application extends BaseApplication
 {
     const VERSION = '2.2.4-DEV';
 
-    /**
-     * Constructor.
-     */
     public function __construct()
     {
         error_reporting(-1);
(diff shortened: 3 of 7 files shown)


src/Console/Command/CommandHelp.php
```diff
@@ -113,7 +113,7 @@ public static function getHelpCopy()
 
 Instead of using command line options to customize the rule, you can save the
 project configuration in a <comment>.php_cs.dist</comment> file in the root directory
-of your project. The file must return an instance of ``PhpCsFixer\ConfigInterface``,
+of your project. The file must return an instance of `PhpCsFixer\\\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.1/src/ConfigInterface.php>`_,
```
julienfalque: Can we use `PhpCsFixer\Console\Application::VERSION` so that we don't need to update the link manually?

SpacePossum: No, as master points to 2.3.[...] and that is not on GH.
We only change the interface on a major anyway and than we can adjust this

keradus: but then a dev will start looking at codebase on GH that is for given tag 2.2.1, even if we would already release 2.9

keradus: you could extract `getLatestTag` from `ReadmeCommand::getLatestDownloadUrl` and then use that tag here instead, so it will automatically update readme file during releasing, as we do in other places.

README.rst
```diff
@@ -1113,7 +1113,7 @@ fixed but without actually modifying them:
 
 Instead of using command line options to customize the rule, you can save the
 project configuration in a ``.php_cs.dist`` file in the root directory
-of your project. The file must return an instance of ``PhpCsFixer\ConfigInterface``,
+of your project. The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.1/src/ConfigInterface.php>`_,
```
SpacePossum: I should restore the other pair of back ticks.

src/Console/Command/FixCommand.php
```diff
@@ -85,6 +85,16 @@ public function __construct()
 
     /**
      * {@inheritdoc}
+     *
+     * Override here to only generate the help copy when used.
```
SpacePossum: turns out this saves more time during fixing than I thought :)

src/Console/Command/CommandHelp.php
```diff
@@ -312,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
+
+        if (null !== $version) {
+            return $version;
+        }
+
+        $currentMajor = (int) Application::VERSION;
+        $changelogFile = __DIR__.'/../../../CHANGELOG.md';
+
+        if (!is_file($changelogFile)) {
+            return null;
```
julienfalque: Shouldn't we fallback to `$this->getApplication()->getVersion()` here?

src/Console/Command/CommandHelp.php
```diff
@@ -312,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
 
+    public static function getLatestReleaseVersionFromChangeLog()
+    {
+        static $version = null;
+
+        if (null !== $version) {
+            return $version;
+        }
+
+        $currentMajor = (int) Application::VERSION;
```
julienfalque: I would still use `$this->getApplication()->getVersion()`.

src/Console/Command/CommandHelp.php
```diff
@@ -313,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
         return $allowed;
     }
 
+    public static function getLatestReleaseVersionFromChangeLog()
```
keradus: please add typehinting (so all tools will know it is supposed to return....)

src/Console/Command/CommandHelp.php
```diff
@@ -313,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
+
+        if (null !== $version) {
+            return $version;
+        }
+
+        $currentMajor = (int) Application::VERSION;
+        $changelogFile = __DIR__.'/../../../CHANGELOG.md';
+
+        if (!is_file($changelogFile)) {
+            return null;
```
keradus: exception ?

src/Console/Command/CommandHelp.php
```diff
@@ -312,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
+        }
+
+        $changelog = file_get_contents($changelogFile);
```
julienfalque: Don't know if this may happen but I would either check if `false !== $changelog = @file_get_contents(...)` or add a `is_readable()` check before.

src/Console/Command/CommandHelp.php
```diff
@@ -312,6 +319,39 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
+
+        for ($i = $currentMajor; $i > 0; --$i) {
+            preg_match('/Changelog for v('.$i.'.\d+.\d+)/', $changelog, $matches);
+
+            if (2 === count($matches)) {
```
julienfalque: `preg_match()` can be used directly in the `if` so you don't need to count `$matches`.

src/Console/Command/CommandHelp.php
```diff
@@ -242,6 +243,12 @@ public static function getHelpCopy()
 EOF
         ;
 
+        $template = str_replace(
+            '%%CONFIG_INTERFACE_URL%%',
```
julienfalque: Can you use three `%` for consistency with the next placeholder?

src/Console/Command/CommandHelp.php
SpacePossum: That would meaning passing the version along, as it won't be static anymore. Not much gain I think as the version of the app will never change during runtime

src/Console/Command/CommandHelp.php
keradus: no, as `getVersion` may provide you dev version that is not yet tagged, so all links with that version would not work.
all versions in changelog are tagged.

src/Console/Command/CommandHelp.php
```diff
@@ -314,6 +320,51 @@ public static function getDisplayableAllowedValues(FixerOptionInterface $option)
...
+
+        $currentMajor = (int) Application::VERSION;
+        $changelogFile = __DIR__.'/../../../CHANGELOG.md';
+        $changelog = @file_get_contents($changelogFile);
+
+        if (false === $changelog) {
+            $error = error_get_last();
+
+            throw new \RuntimeException(sprintf(
+                'Failed to read content of the change log file "%s".%s',
```
julienfalque: _change log_ => _changelog_?

SpacePossum:
What is left is fixing the display of the RST formatted URL's in the commandline output.
I'm planning on making the result of `CommandHelp::getHelpCopy` commandline friendly output only and fix to RST where needed in the readme command. Current the result is a mix of commandline stuff (with `<comment>` tags and such) and RST formatted links, we should return in one format IMO.

SpacePossum:
new

old (top) vs new

README.rst
```diff
@@ -1113,8 +1113,8 @@ fixed but without actually modifying them:
     $ php php-cs-fixer.phar fix /path/to/code --dry-run
 
 Instead of using command line options to customize the rule, you can save the
-project configuration in a ``.php_cs.dist`` file in the root directory
-of your project. The file must return an instance of ``PhpCsFixer\ConfigInterface``,
+project configuration in a ``.php_cs.dist`` file in the root directory of your project.
+The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.2/src/ConfigInterface.php>`_
```
TomasVotruba: Link should be probably general, not locker to the branch, so it up to date with master branch and dead link in case of change is obvious.

`/src/ConfigInterface.php` should do the trick.

SpacePossum: the link is generated for each version, so should never be dead (as long as GH is around)

TomasVotruba: Oh, now I see it's generated. I thought it's `README.md` file :+1:

keradus:
Thank you @SpacePossum.

PR #2707, opened 2017-04-19
Deprecate other than "yes" or "no" for input options

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2585

Deprecate values other than "yes" and "no" for `--allow-risky` and `--using-cache`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -181,7 +181,7 @@ Complete configuration for rules can be supplied using a ``json`` formatted stri
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
 
-The ``--allow-risky`` option (pass `yes` or `no`) allows you to set whether risky rules may run. Default value is taken from config file.
+The ``--allow-risky`` option (pass ``yes`` or ``no``) allows you to set whether risky rules may run. Default value is taken from config file.
 Risky rule is a rule, which could change code behaviour. By default no risky rules are run.
 
 The ``--stop-on-violation`` flag stops execution upon first file that needs to be fixed.
@@ -1180,7 +1180,7 @@ experimental.
         ->setLineEnding("\r\n")
     ;
 
-By using ``--using-cache`` option with `yes` or `no` you can set if the caching
+By using ``--using-cache`` option with ``yes`` or ``no`` you can set if the caching
 mechanism should be used.
 
 Caching
--- a/README.rst
+++ b/README.rst
@@ -181,7 +181,7 @@ Complete configuration for rules can be supplied using a ``json`` formatted stri
 A combination of ``--dry-run`` and ``--diff`` will
 display a summary of proposed fixes, leaving your files unchanged.
 
-The ``--allow-risky`` option allows you to set whether risky rules may run. Default value is taken from config file.
+The ``--allow-risky`` option (pass `yes` or `no`) allows you to set whether risky rules may run. Default value is taken from config file.
 Risky rule is a rule, which could change code behaviour. By default no risky rules are run.
 
 The ``--stop-on-violation`` flag stops execution upon first file that needs to be fixed.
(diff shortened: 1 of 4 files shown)


tests/Console/ConfigurationResolverTest.php
```diff
@@ -1038,6 +1000,37 @@ public function testResolveConfigFileOverridesDefault()
...
+        $resolver = new ConfigurationResolver(
+            $this->config,
+            array('allow-risky' => 'yes please'),
+            ''
+        );
+
+        $this->assertFalse($resolver->getRiskyAllowed());
+    }
+
+    public function getResolveBooleanOptions()
```
SpacePossum: I don't see much value in that TBH

README.rst
```diff
@@ -1180,7 +1180,7 @@ experimental.
         ->setLineEnding("\r\n")
     ;
 
-By using ``--using-cache`` option with yes or no you can set if the caching
+By using ``--using-cache`` option with `yes` or `no` you can set if the caching
```
SpacePossum: done

src/Console/ConfigurationResolver.php
```diff
@@ -784,4 +784,39 @@ private function setOption($name, $value)
...
+        }
+
+        if ('no' === $value) {
+            return false;
+        }
+
+        @trigger_error(
+            sprintf(
+                'Expected "yes" or "no" for option "%s", other values are deprecated and support will be removed in 3.0. Got "%s", this implicitly set the option to "false".',
+                $optionName, $value
```
SpacePossum: fixed ;)

julienfalque:
👍

src/Console/ConfigurationResolver.php
```diff
@@ -784,4 +784,36 @@ private function setOption($name, $value)
...
+    /**
+     * @param string $optionName
+     *
+     * @return bool
+     */
+    private function resolveOptionBooleanValue($optionName)
+    {
+        $value = $this->options[$optionName];
+        if (is_bool($value)) {
+            return $value;
```
keradus: is it ever happening, as we pass value of InputOption which is `yes`/`no` string ?

tests/Console/ConfigurationResolverTest.php
```diff
@@ -1038,6 +1000,37 @@ public function testResolveConfigFileOverridesDefault()
...
+    {
+        return array(
+            array(true, true, 'yes'),
+            array(true, true, true),
```
keradus: how could user pass `true` to `ConfigurationResolver` from CLI?

src/Console/ConfigurationResolver.php
SpacePossum: it is happening in the existing test(case) of the configuration resolving

tests/Console/ConfigurationResolverTest.php
SpacePossum: I don't think that is possible, you think it is OK to drop support for it?

src/Console/ConfigurationResolver.php
keradus: from what I can see it is happening on your branch, but not on 2.2/master

tests/Console/ConfigurationResolverTest.php
keradus: not important for sure.
on the other side, we follow that in basically all tests.

src/Console/ConfigurationResolver.php
SpacePossum: here is one example;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/tests/Console/ConfigurationResolverTest.php#L1006

from the CLI we get `null` or a `string` but never booleans AFAICS

keradus: damn, then we need to keep BC

keradus:
Thank you @SpacePossum.

PR #2708, opened 2017-04-19
Add PhpUnitTestClassRequiresCoversFixer

keradus:
I originally used slightly modified version to add `@covers \Foo` annotations, but it wasn't generic enough to work for all cases. Here, generic version to ensure that any of those annotation is present

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -79,7 +79,7 @@ final class MyTest extends \PHPUnit_Framework_TestCase
             $indent = $tokens[$index - 1]->isGivenKind(T_WHITESPACE)
                 ? (
                     false !== strpos($tokens[$index - 1]->getContent(), "\n")
-                    ? preg_replace('/^.*\v*/', '', $tokens[$index - 1]->getContent())
+                    ? preg_replace('/^.*\R*/', '', $tokens[$index - 1]->getContent())
                     : $tokens[$index - 1]->getContent()
                 )
                 : '';
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -24,8 +24,6 @@ use PhpCsFixer\Tokenizer\Tokens;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
- *
- * @tofo Before revealing feature let wait for https://github.com/sebastianbergmann/phpunit/pull/2560
  */
 final class PhpUnitTestClassRequiresCoversFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
 {
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -74,7 +74,7 @@ final class MyTest extends \PHPUnit_Framework_TestCase
             }
 
             $prevIndex = $tokens->getPrevMeaningfulToken($index);
-            $index = $tokens[$prevIndex]->isGivenKind(array(T_FINAL, T_ABSTRACT)) ? $prevIndex : $index;
+            $index = $tokens[$prevIndex]->isGivenKind(T_FINAL) ? $prevIndex : $index;
 
             $indent = $tokens[$index - 1]->isGivenKind(T_WHITESPACE)
                 ? (
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -1,0 +1,152 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @tofo Before revealing feature let wait for https://github.com/sebastianbergmann/phpunit/pull/2560
+ */
+final class PhpUnitTestClassRequiresCoversFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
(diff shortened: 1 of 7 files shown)


keradus:
let us ignore old PHP failoures and merge this one only after #2568

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $phpUnitIndicator = new PhpUnitIndicator();
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_CLASS)) {
```
SpacePossum: maybe move this check in `isPhpUnitClass` method, as now that method doesn't check if the token at given index is T_CLASS

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -0,0 +1,150 @@
...
+                continue;
+            }
+
+            if (!$phpUnitIndicator->isPhpUnitClass($tokens, $index)) {
+                continue;
+            }
+
+            $prevIndex = $tokens->getPrevMeaningfulToken($index);
+            $index = $tokens[$prevIndex]->isGivenKind(array(T_FINAL, T_ABSTRACT)) ? $prevIndex : $index;
```
SpacePossum: no need for testing T_ABSTRACT, as `isPhpUnitClass` will return false for this case

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -0,0 +1,150 @@
...
+
+            $indent = $tokens[$index - 1]->isGivenKind(T_WHITESPACE)
+                ? (
+                    false !== strpos($tokens[$index - 1]->getContent(), "\n")
```
julienfalque: Not sure but I think this inner ternary expression is not necessary: the regular expression won't match anything if the token does not contain newlines anyway (maybe this requires to replace `\v*` with `\v+` though).

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -0,0 +1,150 @@
...
+                    ? preg_replace('/^.*\v*/', '', $tokens[$index - 1]->getContent())
```
julienfalque: Isn't `\R` better than `\v`?

src/Indicator/PhpUnitIndicator.php
```diff
@@ -0,0 +1,47 @@
...
+namespace PhpCsFixer\Indicator;
+
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @internal
+ */
+final class PhpUnitIndicator
+{
+    public function isPhpUnitClass(Tokens $tokens, $index)
```
julienfalque: Why not adding this to `TokensAnalyzer`?

tests/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixerTest.php
```diff
@@ -0,0 +1,244 @@
...
+                     * @coversNothing
+                     */
+                    class FooTest extends \PHPUnit_Framework_TestCase {}
+                ',
+                '<?php
+
+                    class FooTest extends \PHPUnit_Framework_TestCase {}
+                ',
+            ),
+            'without docblock #2' => array(
```
julienfalque: This case seems to be identical to the previous one.

tests/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixerTest.php
```diff
@@ -0,0 +1,244 @@
...
+            ),
+            'already with annotation: @coversDefaultClass' => array(
+                '<?php
+                    /**
+                     * @coversDefaultClass
+                     */
+                    class FooTest extends \PHPUnit_Framework_TestCase {}
+                ',
+            ),
+            'already with annotation: @coversNothing' => array(
```
julienfalque: This case is already covered with the expected part of the next case.

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -0,0 +1,150 @@
...
+ */
+final class PhpUnitTestClassRequiresCoversFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'PHPUnit test class requires any of `covers*` annotation.',
```
julienfalque: This description is not explicit enough IMO, I would write something like _"Adds a default `@coversNothing` annotation to PHPUnit test classes that have no `@covers*` annotation."_

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: why ?

src/Indicator/PhpUnitIndicator.php
keradus: I would rather decouple `TokensAnalyzer`

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
julienfalque: `\v` is for any vertical whitespaces, which probably matches more characters than just newlines (which is what you want to match if I understand correctly). `\R` on the other hand only matches `\n`, `\r\n` and `\r`.

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: done

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: done

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: while I see definition of `\r`, I cant see it for `\R`. isn't `\R` be everytihng except `\r` ? then, it would match way more than `\v`

julienfalque: From http://php.net/manual/en/regexp.reference.escape.php:

> \R: line break: matches \n, \r and \r\n

keradus: :+1:

tests/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixerTest.php
keradus: :+1:

tests/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixerTest.php
keradus: :+1:

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: nice catch !

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
keradus: in `TokensAnalyser::isSth` we check if it is given token, if not we do raise an exception.
now, I apply that pattern here as well.

PR #2709, opened 2017-04-20
code grooming

keradus:


src/Doctrine/Annotation/Tokens.php
```diff
@@ -301,7 +301,7 @@ public function offsetSet($index, $token)
     public function offsetUnset($index)
     {
         if (!isset($this[$index])) {
-            throw new \OutOfBoundsException('Index %s is invalid or does not exist.');
+            throw new \OutOfBoundsException(sprintf('Index %s is invalid or does not exist.', $index));
```
keradus: buahaha

tests/Cache/FileHandlerTest.php
```diff
@@ -138,7 +138,10 @@ public function testWriteWritesCache()
         $handler->write($cache);
 
         $this->assertFileExists($file);
-        $this->assertSame($cache->toJson(), file_get_contents($file));
```
julienfalque: What's the point of this change?

tests/Cache/FileHandlerTest.php
```diff
@@ -138,7 +138,10 @@ public function testWriteWritesCache()
...
+
+        $actualCacheJson = file_get_contents($file);
+
+        $this->assertSame($cache->toJson(), $actualCacheJson);
```
SpacePossum: can't we use [assertStringEqualsFile](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertJsonFileEqualsJsonFile) or [assertStringEqualsFile](https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertStringEqualsFile)

tests/Cache/FileHandlerTest.php
keradus: some of the SCA tools asked to use file-dedicated assertion, but that assertion take content of that file as expected value, while for us it;s actual value we compare to expected json value.

for not having complain in the end

julienfalque: I'd suggest you open issues for those tools 😋

GrahamCampbell: It's a bug that the tool doesn't pick it up when you assign to a variable first, since that assignment is live when we reach the assertion.

PR #2711, opened 2017-04-20
FixCommand - fix diff optional value handling

keradus:
regression fix for #2554:
```php
new InputOption('diff', '', InputOption::VALUE_OPTIONAL, 'Also produce diff for each file.', 'sbd'),
```
that means:
- 'diff' option may or may not be passed
- when it's passed, value for it is optional
- 'sbd' is default value of `diff` option regardless it is passed or not (!!!)

I guess assumption during creating 2254 was that `sbd` would be only a value for `diff` if one would pass `--diff` (without value). Apparently, it is a value even if one would not pass that option at all, leading to generate diff even without using `--diff` option.

ref https://github.com/symfony/symfony/issues/11572


SpacePossum:
oh sorry about that one :(
thanks for picking it up!

Issue #2712, opened 2017-04-20
ordered_imports ignores indentation


pilgerone:
With `ordered_imports = true` but `single_import_per_statement = false`, the fixer messes up the indentation on PHP 7.1. For example:
Original:
```
<?php

namespace Foo\Bar;

use Foo\Bar\Baz\{
    C,
    B,
    D,
    A
};
```
Expected:
```
<?php

namespace Foo\Bar;

use Foo\Bar\Baz\{
    A,
    B,
    C,
    D
};
```
Actual:
```
<?php

namespace Foo\Bar;

use Foo\Bar\Baz\{A, B, C, D};
```
Might seem harmless but becomes a real problem with big namespaces and class names.

I'm willing to take a swing at this one. Should I do a PR with some tests and attempt a change to the fixer?

SpacePossum:
>Should I do a PR with some tests and attempt a change to the fixer?

Would be awesome. The Fixer is a bit tricky (or at least an odd duck) when it comes to fixing the code. Feel free to give it a shot!

PR #2715, opened 2017-04-21
OrderedImportsFixer - handle indented groups

pilgerone:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2712

I've tried to keep it small and elegant. It's passing the tests alright, but I hope it meets the standards.

### DIFF
--- a/src/Fixer/Import/OrderedImportsFixer.php
+++ b/src/Fixer/Import/OrderedImportsFixer.php
@@ -14,6 +14,7 @@ namespace PhpCsFixer\Fixer\Import;
 
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
@@ -31,8 +32,9 @@ use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  * @author SpacePossum
  * @author Darius Matulionis <darius@matulionis.lt>
+ * @author Adriano Pilger <adriano.pilger@gmail.com>
  */
-final class OrderedImportsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+final class OrderedImportsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
     const IMPORT_TYPE_CLASS = 'class';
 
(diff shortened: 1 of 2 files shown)


src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -368,6 +375,15 @@ private function getNewOrder(array $uses, Tokens $tokens)
 
                                     continue;
                                 }
+                                //if there is any line ending inside the group import, it should be indented properly
```
julienfalque: Please add an empty line before this comment.

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -368,6 +375,15 @@ private function getNewOrder(array $uses, Tokens $tokens)
...
+                                if ('' === $firstIndent &&
```
julienfalque: Please add an empty line after the opening brace.

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -663,16 +663,36 @@ public function provide70Cases()
...
 use function some\a\{fn_a, fn_b, fn_c};
 use A\B;
+use const some\b\{
+    ConstC,
+    ConstA,
+    ConstB
+};
 use const some\a\{ConstA, ConstB, ConstC};
 use some\b\{
-    ClassF,
```
julienfalque: Why did you change the order here?

SpacePossum: its an extra test :+1:

julienfalque: Then it should be a new case instead of updating an existing one.

SpacePossum: It is a tweak, so it tests more than before. Nothing is lost, as the `expected` part of this change is also tested. Making a new one would duplicate a lot, so why not just fix this?

pilgerone:
Lesson learnt: don't ever rebase before setting the right target.

All sorted out now I guess. Target and base.

tests/Fixer/Import/OrderedImportsFixerTest.php
pilgerone: It's actually a reminiscent of the early stage of my experiments. The tests I've added to the bottom should cover this situation as well. It could be rolled back. But, yes, it doesn't break and it covers a bit more.

At first glance this particular test seems to focus on ordering namespaces vs. functions vs. constants vs. classes but there was already some imports out of order inside the one of the groups. So I'm not really sure of the organization logic there. 

I guess we could refactor the tests as to leave each testing exactly one behavior. I'm not sure this is necessary or desirable, though.

I'm up to whatever you guys decide.

SpacePossum: I would merge this change in the test in a heartbeat, but that is just my vote ;)

julienfalque: I actually misread the changes, I'm fine with it :)

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -369,6 +376,17 @@ private function getNewOrder(array $uses, Tokens $tokens)
                                     continue;
                                 }
 
+                                //if there is any line ending inside the group import, it should be indented properly
```
julienfalque: Please add a space after `//`.

keradus:
Thank you @pilgerone.

PR #2717, opened 2017-04-22
Remove old polyfills

GrahamCampbell:
Maybe I'm missing something, but do we still need these?


SpacePossum:
Indeed, these can go. Not only on master but on 2.3 as well :)

keradus:
good catch

keradus:
thank you, replaced by #2742

PR #2718, opened 2017-04-22
Remove old Symfony exception message expectation

julienfalque:


SpacePossum:
Tests still pass so looks like a good clean up to me. Is this possible because the SF deps. version were bumped?

keradus:
yep, dropping old symfony is the reason why we could clean this up

keradus:
Thank you @julienfalque.

PR #2720, opened 2017-04-22
Proof of concept for fabbot issue

GrahamCampbell:
Let's see what this does...


GrahamCampbell:
Wow:

image

GrahamCampbell:
Meh.

SpacePossum:
You don't want to know how deep the rabbit hole goes?
Not sure about how big of a risk `fabbot.io` is in because of this, therefor pinging some know involved people.
@fabpot @stof

GrahamCampbell:
I did already report this earlier as an obvious problem. Was dismissed by @keradus already.

GrahamCampbell:
We are somewhat limited as to what data can be stolen from the server, since we'd have to chunk it into 80 chars, or whatever github status api truncates the output too. As for what we can execute on the server, I didn't wanna try that out.

SpacePossum:
I _expect_ no critical data is on fabbot.io. However trying out security issues on that domain in public using this repo without consent of both the owner of fabbot.io and this repo is out of line IMHO.
The owners of fabbot.io should get a fair chance to deal with security issues before going public.
If they do not want to deal with those issues there should still be on OK given by the owners of this repo before it is used to test the issues are test further.
So, yes, the security of fabbot.io is not of our concern, but that does not mean we should not report those concerns let alone try out the issues in public using this repo.
I respect you  as contributor (even while we had our fair share of disagreements here ;) ) and as such I believe you had no wrong intent to try this out. However since this is public I would like to urge you not proceed trying out this further ATM.

GrahamCampbell:
> I expect no critical data is on fabbot.io. However trying out security issues on that domain in public using this repo without consent of both the owner of fabbot.io and this repo is out of line IMHO.

That's why I've not done anything more than what's seen on this PR. Exactly what I have done is also documented by the commit history.

> The owners of fabbot.io should get a fair chance to deal with security issues before going public.

To be fair, I've not done anything that isn't already glaringly obvious.

> However since this is public I would like to urge you not proceed trying out this further ATM.

I have no plans to. That's why I closed it.

SpacePossum:
thanks @GrahamCampbell , I know you all is good
now lets see what we get back from the owners of fabbot.io :)

keradus:
> Was dismissed by @keradus already.

Nope. I said it's not an dirrect issue of PHP CS Fixer but service that use, among others, PHP CS Fixer.
It's up to fabbot.io how they put the code they executed into sandbox. I'm not claiming it's good or bad. 

I'm saying you report your concern in wrong place...
owner of the service will rather not see this report if it's reported in wrong place...

GrahamCampbell:
Fair. :P

fabpot:
Indeed, reporting such issues by email would have been better.

We are using a sandbox for fabbot, but we missed a change in PHP-CS-Fixer (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2244/files#diff-62b374855f5961454b817b144c531e27R167 to be precise) that has the bad consequence to bypass our sandbox.

Anyway, we fixed it now, and everything should be under control. Thanks for the report.

GrahamCampbell:
👍

keradus:
fabpot, I bet you mean `$config = include $configFile;`. same possibility was there even before cleanup ;)

SpacePossum:
good to hear it is fixed, but I think indeed no new vulnerabilities were added since about launch ;)

Issue #2512, opened 2017-02-06
Please remove required for php7.2 or add 7.2.0-dev


chacal88:
- Installation request for fabpot/php-cs-fixer 1.12.x-dev -> satisfiable by fabpot/php-cs-fixer[1.12.x-dev].
    - fabpot/php-cs-fixer 1.12.x-dev requires php ^5.3.6 || >=7.0 <7.2 -> your PHP version (7.2.0-dev) does not satisfy that requirement.

SpacePossum:
We don't support 7.2 yet, removing the limit in the composer file would suggest otherwise.
Please help out supporting all [7.1 features](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2090) and we can move on for 7.2 :D

curry684:
I don't quite agree with that statement - nightlies are never supported by anyone, and that's all there is for PHP 7.2.

But meanwhile this does make it impossible to add PHP-CS-Fixer to your dev dependencies (where it should be) and test against nightly PHP builds anywhere, just because composer won't install automatically.

I can work around this in tests with `--ignore-platform-reqs` but it's just silly to enforce this in a library. PHPunit doesn't officially *support* it either but that's a galaxy away from actively *blocking* it.

SpacePossum:
I didn't say anything about nightlies, if PHP7.2 was released now as stable I wouldn't update the restriction in composer either. The other way around as well, so if the fixer would work correctly on PHP7.2 while that still being a nightly; I would remove the restriction in composer in new releases.

The only exception I would consider is removing the restriction in the 'nighty' releases of the tool itself and the branch where it was build from, i.e. master. But for 1.12.x it makes no sense, I think no one should expect the 1.12.x line to support PHP7.2 at any time as development of that line has stopped.

SpacePossum:
That is an other problem than original reported;
> Installation request for fabpot/php-cs-fixer 1.12.x-dev -> satisfiable by fabpot/php-cs-fixer[1.12.x-dev].

For 1.12.x I stated support for PHP 7.2 is not likely to ever happen. 

For your issue;
> v2.0.1 requires php ^5.3.6 || >=7.0 <7.2
> I just want to ensure the library can run with stable software on the upcoming PHP version.

IMHO my point sticks, as of now we cannot ensure you the library is stable for the upcoming PHP version.
[...]

PR #2722, opened 2017-04-24
Remove needlessly strict PHP requirement

kelunik:
There's no reason to restrict the upper version of PHP, this only makes testing on newer PHP versions harder, see e.g. https://travis-ci.org/amphp/amp/jobs/225272947#L854.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || >=7.0 <7.2",
+        "php": ">=5.6",
         "doctrine/annotations": "^1.2",
         "ext-tokenizer": "*",
         "sebastian/diff": "^1.4",


SpacePossum:
Hi!
The upper limit is done because currently we cannot state that we support 7.2.
(and strictly speaking your change would indicate we support PHP 6 as well ;) )
For more background and discussion I recommend reading this thread where a similar request as yours was made
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2512

Of course we would very much like to state that we support PHP 7.2, so we are always looking for people to help out with that :)

GrahamCampbell:
👎 For merging this.

keradus:
Exactly, if we cannot ensure we works on 7.2 or so, we don't claim it. We are not random library that may work on any 5.x | 7.x. We are using internal, embedded into php `tokenizer` module, that basically comes with bc breaker with ever PHP minor (feature) release.

For that, as @SpacePossum stated, it would be great if you could help bring us closer to support 7.2.

keradus:
Closing due to discussion

PR #2724, opened 2017-04-25
Revert #2554 Add short diff. output format

keradus:
Sadly, #2554 introduced BC breaker of CLI interface. We need to rollback it.


SpacePossum:
sadly needed, but indeed very much needed
:+1: 
thanks for picking this up

PR #2727, opened 2017-04-26
PHPUnit - use speedtrap

keradus:


julienfalque:
Looks like most slow tests are related to Doctrine Annotation fixers :(

keradus:
yep. feel free to try improve their performance ;)

julienfalque:
Yep.

PR #2729, opened 2017-04-27
DescribeCommand - add decorated output test

julienfalque:


keradus:
could it be targetted for 2.2?

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -57,21 +56,64 @@ public function testExecuteOutput()
...
+   \033[32m+<?php echo 'good stuff and good thing';\033[39m
+\033[33m   ----------- end diff -----------\033[39m
+
+
+EOT;
+
+        $actual = $this->execute(true)->getDisplay();
+
+        if (false !== strpos($actual, "\033[0m")) {
+            $expected = str_replace("\033[39m", "\033[0m", $expected);
```
julienfalque: This can be removed when merging up to 2.3 as this does not occur anymore with `symfony/console:^3.0`.

src/Differ/DiffConsoleFormatter.php
```diff
@@ -79,7 +79,7 @@ function ($string) use ($isDecorated, $lineTemplate) {
                     "#\n\r|\n#",
                     $isDecorated
                         ? OutputFormatter::escape(rtrim($diff))
-                        : $diff
+                        : rtrim($diff)
```
julienfalque: This fixes a small inconsistency I noticed while fixing the test: without colored output, there is an extra line with three spaces after each diff. This does not occur with colored output.

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -57,21 +56,64 @@ public function testExecuteOutput()
...
-%spaces%
    ----------- end diff -----------
 
 
 EOT;
 
         $this->assertSame(
-            str_replace(array('%spaces%', "\n"), array('   ', PHP_EOL), $expected),
-            $this->execute()->getDisplay()
+            str_replace(array("\n"), array(PHP_EOL), $expected),
```
julienfalque: Note to self: arrays not required here.

src/Differ/DiffConsoleFormatter.php
julienfalque: This fixes a small inconsistency I noticed while fixing the test: without colored output, there is an extra line with three spaces after each diff. This does not occur with colored output.

Edit:

**before**
before

**after**
after

keradus: nice, I remember I was looking for the reason of that issue

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -57,21 +56,64 @@ public function testExecuteOutput()
...
+EOT;
+
+        $actual = $this->execute(true)->getDisplay();
+
+        if (false !== strpos($actual, "\033[0m")) {
+            $expected = str_replace("\033[39m", "\033[0m", $expected);
+        }
+
+        $this->assertSame(
+            str_replace("\n", PHP_EOL, $expected),
```
keradus: just pass true to `getDisplay` instead

keradus:
Thank you @julienfalque.

PR #2733, opened 2017-04-29
Also fix .php_cs.dist

GrahamCampbell:
Noticed that the long array syntax was still used in this file.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -42,5 +42,6 @@ return PhpCsFixer\Config::create()
         PhpCsFixer\Finder::create()
             ->exclude('tests/Fixtures')
             ->in(__DIR__)
+            ->name('.php_cs.dist')
     )
 ;
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -40,6 +40,7 @@ return PhpCsFixer\Config::create()
     ))
     ->setFinder(
         PhpCsFixer\Finder::create()
+            ->ignoreDotFiles(false)
             ->exclude('tests/Fixtures')
             ->in(__DIR__)
             ->name('.php_cs.dist')


GrahamCampbell:
Oh, I guess this doesn't work because "dot files" are excluded. :/

keradus:
:-1: for this change.
People are using our config file as example of it, they try to mimic it and adjust for their needs. I don't wanna spread fixing dot-files as good practice, as it is not.

GrahamCampbell:
> People are using our config file as example of it, they try to mimic it and adjust for their needs.

Well, that's not a good reason to not fix our own code.

> I don't wanna spread fixing dot-files as good practice, as it is not.

I don't think that makes sense as a blanket statement, however it makes sense as a good default.

GrahamCampbell:
> I don't think that makes sense as a blanket statement

In fact, this repository is evidence that fixing dot files is needed.

SpacePossum:
why does the finder ignore files even when added explicitly using `name()`, looks odd to me

GrahamCampbell:
> why does the finder ignore files even when added explicitly using name(), looks odd to me

Because files starting with a dot were explicitly blacklisted, and the blacklist overrides any whitelist.

SpacePossum:
>Because files starting with a dot were explicitly blacklisted, and the blacklist overrides any whitelist.

to bad :|
I agree with @keradus here (fixing the file is a good idea, something maybe we can add/tweak in the travis config, but as default it makes sense to me to skip dot files).

GrahamCampbell:
We are skipping dot files by default. Anyone that just copies the config file for this repo is naive at best. They should read the readme.

keradus:
I don't trust devs that much to believe that all of them would read it.
image is worth a thousand words.... example? docs...

keradus:
thank you, replaced by #2742

PR #2738, opened 2017-04-29
Upgrade tests to use new, namespaced PHPUnit TestCase class

keradus:


GrahamCampbell:
Might be worth adding a unit test to check tests extend the new class. Would have to rely on tokenizing the files to work out what they extend though (I can't think of any other way to check this).

GrahamCampbell:
👎 On doing this on 2.2.x. It's not a bug fix.

keradus:
ultimatelly, on newer branch, we will just drop support for old php that provides that class.

it's not a feature as well.

keradus:
> Might be worth adding a unit test to check tests extend the new class.

btw, check out last commit ;)

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -82,7 +83,7 @@ public function testThatSrcClassHaveTestClass($className)
         }
 
         $this->assertTrue(class_exists($testClassName), sprintf('Expected test class "%s" for "%s" not found.', $testClassName, $className));
-        $this->assertTrue(is_subclass_of($testClassName, '\PHPUnit_Framework_TestCase'), sprintf('Expected test class "%s" to be a subclass of "\PHPUnit_Framework_TestCase".', $testClassName));
+        $this->assertTrue(is_subclass_of($testClassName, '\PHPUnit\Framework\TestCase'), sprintf('Expected test class "%s" to be a subclass of "\PHPUnit\Framework\TestCase".', $testClassName));
```
GrahamCampbell: This test works both ways, because both names are bound to the same class I think?

keradus: no.

https://github.com/sebastianbergmann/phpunit/blob/4.8/src/ForwardCompatibility/TestCase.php

is something is subclass of `PHPUnit_Framework_TestCase`, then it may not be subclass of `PHPUnit\Framework\TestCase`. we expect later, so we check for later. if one would implement test using first class, this assertion would fail.

GrahamCampbell: Oh I see. I assumed it was a `class_alias`.

PR #2740, opened 2017-05-01
Add VoidReturnFixer

mrmark:
Hello, first time contributor here!

This fixer adds the PHP7.1 void return type on functions that have no return statements or all the return statements are empty.

For reviewers, I'd like to call out the following:
* Please ensure `getDefinition` looks right.
* `applyFix` could likely be optimized by not using a `foreach` and instead jumping over function body definitions when a function is skipped.  Let me know if this should be done.
* In `fixFunctionDefinition`, I'm unsure if `new Token([T_STRING, 'void']),` is correct.

I ran this fixer on one of my personal projects and it worked well, except that it applies the fix to methods that implement interfaces, abstract classes, etc that belong to 3rd party vendors.  This doesn't seem to break tests, but also doesn't seem 100% correct.  Unsure if there is a better way to fix this or because it's not 100% correct, then this fixer shouldn't be included.  Maybe this is a fixer that you just run occasionally instead of all the time?  Anyways, feedback on this is welcome.

Cheers and thanks!

### DIFF
--- a/src/Fixer/FunctionNotation/VoidReturnFixer.php
+++ b/src/Fixer/FunctionNotation/VoidReturnFixer.php
@@ -98,7 +98,7 @@ final class VoidReturnFixer extends AbstractFixer
                 continue;
             }
 
-            if (!$tokens[$startIndex]->equals('{')) {
+            if ($tokens[$startIndex]->equals(';')) {
                 // No function body defined, fallback to PHPDoc.
                 if ($this->hasVoidReturnAnnotation($tokens, $index)) {
                     $this->fixFunctionDefinition($tokens, $startIndex);
@@ -120,7 +120,7 @@ final class VoidReturnFixer extends AbstractFixer
     }
 
     /**
-     * Determine if there is a non-void return annotation in the function's PHPDoc comment.
+     * Determine whether there is a non-void return annotation in the function's PHPDoc comment.
      *
      * @param Tokens $tokens
      * @param int    $index  The index of the function token
@@ -130,8 +130,7 @@ final class VoidReturnFixer extends AbstractFixer
     private function hasReturnAnnotation(Tokens $tokens, $index)
     {
         foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
-            $types = $return->getTypes();
-            if (count($types) > 1 || 'void' !== $types[0]) {
+            if (['void'] !== $return->getTypes()) {
                 return true;
             }
         }
@@ -140,7 +139,7 @@ final class VoidReturnFixer extends AbstractFixer
     }
 
     /**
-     * Determine if there is a void return annotation in the function's PHPDoc comment.
+     * Determine whether there is a void return annotation in the function's PHPDoc comment.
      *
      * @param Tokens $tokens
      * @param int    $index  The index of the function token
@@ -150,8 +149,7 @@ final class VoidReturnFixer extends AbstractFixer
     private function hasVoidReturnAnnotation(Tokens $tokens, $index)
     {
         foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
-            $types = $return->getTypes();
-            if (1 === count($types) && 'void' === $types[0]) {
+            if (['void'] === $return->getTypes()) {
                 return true;
             }
         }
@@ -160,7 +158,7 @@ final class VoidReturnFixer extends AbstractFixer
     }
 
     /**
-     * Determine the function already has a return type hint.
+     * Determine whether the function already has a return type hint.
      *
      * @param Tokens $tokens
      * @param int    $index  The index of the end of the function definition line, EG at { or ;
@@ -176,7 +174,7 @@ final class VoidReturnFixer extends AbstractFixer
     }
 
     /**
-     * Determine if the function has a void return.
+     * Determine whether the function has a void return.
      *
      * @param Tokens $tokens
      * @param int    $startIndex Start of function body
@@ -188,15 +186,16 @@ final class VoidReturnFixer extends AbstractFixer
     {
         for ($i = $startIndex; $i < $endIndex; ++$i) {
             if ($tokens[$i]->isGivenKind(T_YIELD)) {
-                return false; // Do not apply fix as generators cannot return void.
+                return false; // Generators cannot return void.
             }
+
             if (!$tokens[$i]->isGivenKind(T_RETURN)) {
                 continue;
             }
 
             $nextToken = $tokens->getNextMeaningfulToken($i);
             if (!$tokens[$nextToken]->equals(';')) {
-                return false; // Do not apply fix, non-empty return statement found.
+                return false;
             }
         }
 
--- a/src/Fixer/FunctionNotation/VoidReturnFixer.php
+++ b/src/Fixer/FunctionNotation/VoidReturnFixer.php
@@ -1,0 +1,254 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\FunctionNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\Annotation;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Mark Nielsen
+ */
+final class VoidReturnFixer extends AbstractFixer
+{
+    /**
+     * @internal
+     */
+    const VOID_RETURN_PATTERN = '/@return\s+void(?!\|)/';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Add void return type to functions with missing or empty return statements, but priority is given to `@return` annotations. Requires PHP >= 7.1.',
+            [
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(\$a) {};",
+                    new VersionSpecification(70100)
+                ),
+            ],
+            'Rule is applied only in a PHP 7.1+ environment.',
+            'Modifies the signature of functions.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run before ReturnTypeDeclarationFixer and PhpdocNoEmptyReturnFixer
+        return 15;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return PHP_VERSION_ID >= 70100 && $tokens->isTokenKindFound(T_FUNCTION);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // These cause syntax errors.
+        static $blacklistFuncNames = [
+            [T_STRING, '__construct'],
+            [T_STRING, '__destruct'],
+            [T_STRING, '__clone'],
+        ];
+
+        for ($index = $tokens->count() - 1; 0 <= $index; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $funcName = $tokens->getNextMeaningfulToken($index);
+            if ($tokens[$funcName]->equalsAny($blacklistFuncNames, false)) {
+                continue;
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+
+            if ($this->hasReturnTypeHint($tokens, $startIndex)) {
+                continue;
+            }
+
+            if (!$tokens[$startIndex]->equals('{')) {
+                // No function body defined, fallback to PHPDoc.
+                if ($this->hasVoidReturnAnnotation($tokens, $index)) {
+                    $this->fixFunctionDefinition($tokens, $startIndex);
+                }
+
+                continue;
+            }
+
+            if ($this->hasReturnAnnotation($tokens, $index)) {
+                continue;
+            }
+
+            $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $startIndex);
+
+            if ($this->hasVoidReturn($tokens, $startIndex, $endIndex)) {
+                $this->fixFunctionDefinition($tokens, $startIndex);
+            }
+        }
+    }
+
+    /**
+     * Determine if there is a non-void return annotation in the function's PHPDoc comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasReturnAnnotation(Tokens $tokens, $index)
+    {
+        foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
+            if (0 === preg_match(self::VOID_RETURN_PATTERN, $return->getContent())) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Determine if there is a void return annotation in the function's PHPDoc comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasVoidReturnAnnotation(Tokens $tokens, $index)
+    {
+        foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
+            if (1 === preg_match(self::VOID_RETURN_PATTERN, $return->getContent())) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Determine the function already has a return type hint.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the end of the function definition line, EG at { or ;
+     *
+     * @return bool
+     */
+    private function hasReturnTypeHint(Tokens $tokens, $index)
+    {
+        $endFuncIndex = $tokens->getPrevTokenOfKind($index, [')']);
+        $nextIndex = $tokens->getNextMeaningfulToken($endFuncIndex);
+
+        return $tokens[$nextIndex]->equals([CT::T_TYPE_COLON, ':']);
+    }
+
+    /**
+     * Determine if the function has a void return.
+     *
+     * @param Tokens $tokens
+     * @param int    $startIndex Start of function body
+     * @param int    $endIndex   End of function body
+     *
+     * @return bool
+     */
+    private function hasVoidReturn(Tokens $tokens, $startIndex, $endIndex)
+    {
+        for ($i = $startIndex; $i < $endIndex; ++$i) {
+            if ($tokens[$i]->isGivenKind(T_YIELD)) {
+                return false; // Do not apply fix as generators cannot return void.
+            }
+            if (!$tokens[$i]->isGivenKind(T_RETURN)) {
+                continue;
+            }
+
+            $nextToken = $tokens->getNextMeaningfulToken($i);
+            if (!$tokens[$nextToken]->equals(';')) {
+                return false; // Do not apply fix, non-empty return statement found.
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * Apply the fix to the function definition.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the end of the function definition line, EG at { or ;
+     */
+    private function fixFunctionDefinition(Tokens $tokens, $index)
+    {
+        $endFuncIndex = $tokens->getPrevTokenOfKind($index, [')']);
+        $tokens->insertAt($endFuncIndex + 1, [
+            new Token([CT::T_TYPE_COLON, ':']),
+            new Token([T_WHITESPACE, ' ']),
+            new Token([T_STRING, 'void']),
+        ]);
+    }
+
+    /**
+     * Find all the return annotations in the function's PHPDoc comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return Annotation[]
+     */
+    private function findReturnAnnotations(Tokens $tokens, $index)
+    {
+        do {
+            $index = $tokens->getPrevNonWhitespace($index);
+        } while ($tokens[$index]->isGivenKind([
+            T_ABSTRACT,
+            T_FINAL,
+            T_PRIVATE,
+            T_PROTECTED,
+            T_PUBLIC,
+            T_STATIC,
+        ]));
+
+        if (!$tokens[$index]->isGivenKind(T_DOC_COMMENT)) {
+            return [];
+        }
+
+        $doc = new DocBlock($tokens[$index]->getContent());
+
+        return $doc->getAnnotationsOfType('return');
+    }
+}
--- a/src/Fixer/FunctionNotation/VoidReturnFixer.php
+++ b/src/Fixer/FunctionNotation/VoidReturnFixer.php
@@ -28,11 +28,6 @@ use PhpCsFixer\Tokenizer\Tokens;
 final class VoidReturnFixer extends AbstractFixer
 {
     /**
-     * @internal
-     */
-    const VOID_RETURN_PATTERN = '/@return\s+void(?!\|)/';
-
-    /**
      * {@inheritdoc}
      */
     public function getDefinition()
@@ -135,7 +130,8 @@ final class VoidReturnFixer extends AbstractFixer
     private function hasReturnAnnotation(Tokens $tokens, $index)
     {
         foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
-            if (0 === preg_match(self::VOID_RETURN_PATTERN, $return->getContent())) {
+            $types = $return->getTypes();
+            if (count($types) > 1 || 'void' !== $types[0]) {
                 return true;
             }
         }
@@ -154,7 +150,8 @@ final class VoidReturnFixer extends AbstractFixer
     private function hasVoidReturnAnnotation(Tokens $tokens, $index)
     {
         foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
-            if (1 === preg_match(self::VOID_RETURN_PATTERN, $return->getContent())) {
+            $types = $return->getTypes();
+            if (1 === count($types) && 'void' === $types[0]) {
                 return true;
             }
         }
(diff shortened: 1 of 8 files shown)


mrmark:
Thanks for the review!  Just pushed up requested changes.  FYI, this checks off the first checkbox in #2090.

mrmark:
Any tips on how to fix the `fabbot.io` check?

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
```diff
@@ -0,0 +1,82 @@
...
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
```
GrahamCampbell: 👎 For that. It would indicate a programmer error more likely in the function body, particularly if they are implementing an interface with return types.

keradus:
Ignore fabbot.io, it's not marked as `Required` hook.
(under the hood - fabbot has not been upgraded yet to use newest PHP CS Fixer, that's why it is unaware of rules we are using)

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,151 @@
...
+ */
+final class VoidReturnFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Add void return type to functions with missing or empty return statements.',
```
keradus: PHP requirement should be here, like in all other fixers. This is what is used for readme file.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,151 @@
...
+            [
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(\$a) {};",
+                    new VersionSpecification(70100)
+                ),
+            ],
+            'Rule is applied only in a PHP 7.1+ environment.',
+            'Modifies the signature of functions'
```
keradus: missing `.`

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,151 @@
...
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 1; 0 <= $index; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $funcName = $tokens->getNextMeaningfulToken($index);
+            if ('__construct' === $tokens[$funcName]->getContent()) {
+                continue; // Cannot declare return type on constructors.
```
keradus: what about other magic methods ?

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,151 @@
...
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 1; 0 <= $index; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $funcName = $tokens->getNextMeaningfulToken($index);
+            if ('__construct' === $tokens[$funcName]->getContent()) {
```
keradus: `->equals`

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,151 @@
...
+                continue; // Cannot declare return type on constructors.
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+            if ('{' !== $tokens[$startIndex]->getContent()) {
```
keradus: equals

src/Fixer/FunctionNotation/VoidReturnFixer.php
GrahamCampbell: Missing test to check that descriptions have a fullstop. :P

keradus: then what you are waiting for ?

src/Fixer/FunctionNotation/VoidReturnFixer.php
GrahamCampbell: The others seem fine, just construct for some bs reason.

mrmark:
Thanks for the review! Just pushed up requested changes.  Some notes though:
* Found that `__destruct` and `__clone` produce syntax errors if void return type is added, so good finds!
* I added several other magic methods to the blacklist as they should return something even though adding void return type doesn't cause a syntax error.  Let me know if I shouldn't do this.
* I added tests that include all magic methods just so everyone is clear on expected behavior.
* While converting to `\PhpCsFixer\Tokenizer\Token::equals`, I used case insensitive when comparing to the void token in `fixFunctionDefinition`.  My reasoning was incase someone was using something like `Void` instead of `void`.  Let me know if this is wrong and it should always be lowercase.

GrahamCampbell:
> I added several other magic methods to the blacklist as they should return something even though adding void return type doesn't cause a syntax error. Let me know if I shouldn't do this.

👎 Only add the ones that don't work. Interesting that destruct doesn't work.

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
julienfalque: How would it be an error?

GrahamCampbell: Oh, commented on the wrong comment. Anyways, this was with reference to correcting "incorrect" return types:

https://3v4l.org/KjQbp

GrahamCampbell: Will fail on file include, instead of at runtime.

mrmark:
Looks like Travis build broke because `\PhpCsFixer\AbstractFixer::createConfigurationDefinition` doesn't return anything because it just throws errors, expecting to be overridden.  So, it is getting void return type, but none of the overrides get it.  Any suggestions?

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
mrmark: That almost seems better because you failed to implement the interface.  If you have an empty `baz()` as in your example, you get a fatal error once it is called.  Unsure if partially written code should count as a valid use case.

julienfalque:
What about checking `@return` annotation?

julienfalque:
> 👎 Only add the ones that don't work.

Why? If the intended behavior of the fixer is to not touch those methods, it should be covered as well.

---
For the methods that only throw exceptions without any `return`/`yield` statement, maybe the fixer should just ignore them?

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
julienfalque: I think fixers should assume the processed code is correct. If a method like in this example is fixed, it will help the developer notice the error as he will see an unexpected `void` in the diff (or his test suite will fail). The fixer is risky so users are aware they should double check their codebase after a PHP CS Fixer run.

GrahamCampbell:
> If the intended behavior of the fixer is to not touch those methods, it should be covered as well.

"If". I don't see why it shouldn't fix methods that can be fixed.

julienfalque:
Actually the methods that are blacklisted because _"void return type does not make sense"_ (`__get`, `__isset`, `__sleep`, `__toString`, `__set_state` and `__debugInfo`) are all supposed to return something. That means they should contain a `return` statement and thus will be ignored by the fixer anyway so yes, 👍 for not blacklisting them and not having specific test cases.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $blacklistFuncNames = [
```
GrahamCampbell: can be "static"

GrahamCampbell:
Ok, looks good. 👍

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
keradus: if fixer make error visible on compile level instead of run level, even better.

mrmark:
> For the methods that only throw exceptions without any return/yield statement, maybe the fixer should just ignore them?

> What about checking @return annotation?

I like both of these suggestions, but I went with the `@return` annotation to give developers a way to override this fixer.

Changes for this new review:
* Added `static` on `$blacklistFuncNames`
* Removed tests and blacklist for all magic methods except for the three that cause syntax errors.
* Updated description about annotations.
* I raised the priority to 15 to get ahead of the PhpdocNoEmptyReturnFixer so we can use those annotations prior to their removal.
* Now functions are skipped if they have a return annotation other than `@return void`.
* Now functions with no bodies (EG: interface/abstract), but have `@return void` annotation will receive the void return fix.

Please closely review `findReturnAnnotations`, I mostly grabbed that from another fixer and I'm not 100% sure I got it right.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,243 @@
...
+                continue;
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+            if (!$tokens[$startIndex]->equals('{')) {
+                // No function body defined, fallback to PHPDoc.
+                if ($this->hasVoidReturnAnnotation($tokens, $index)) {
+                    $this->fixFunctionDefinition($tokens, $startIndex);
+                }
+                continue;
```
julienfalque: Please add an empty line before `continue`.

mrmark:
Ran this on my test project and _I think_ I found a valid use case to @GrahamCampbell's argument about not modifying the return type:

```php
interface InterfaceA {
    public function foo(): string
}

class Test implements InterfaceA {
    public function foo(): string
    {
        throw new LogicException('Not implemented yet');
    }
}
```
Currently the fixer will change `: string` to `: void` even though `foo` method's signature is correct and the `LogicException` is thrown properly without a fatal error.  With the current fixer code though, this can be fixed by adding an annotation.

Should we expect the developer to add the annotation or do we want to remove this feature of the fixer?

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,243 @@
...
+     * Determine if there is a void return annotation in the function's PHPDoc comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasVoidReturnAnnotation(Tokens $tokens, $index)
+    {
+        $returns = $this->findReturnAnnotations($tokens, $index);
```
julienfalque: `$returns` is used only once, you can pass its value directly to the `foreach`.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,243 @@
...
+
+    /**
+     * Determine if there is a non-void return annotation in the function's PHPDoc comment.
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasReturnAnnotation(Tokens $tokens, $index)
```
julienfalque: This method is the exact opposite of `hasVoidReturnAnnotation()` method. Any call to it can be replaced with `!$this->hasVoidReturnAnnotation(...)`.

mrmark:
I guess a third option would to also implement @julienfalque's 2nd suggestion and skip methods that only throw exceptions, but I'm not entirely sure how to do that safely (EG: if the throw is in an if statement, etc).  Suggestions or code samples are welcomed.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,243 @@
...
+                continue;
+            }
+
+            $funcName = $tokens->getNextMeaningfulToken($index);
+            if ($tokens[$funcName]->equalsAny($blacklistFuncNames, false)) {
+                continue;
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+            if (!$tokens[$startIndex]->equals('{')) {
```
julienfalque: At this point, you will either call `hasVoidReturnAnnotation()` or `hasReturnAnnotation()`. If one returns `true`, then the other returns `false`, so you can replace the two calls with
```
$hasVoidReturnAnnotation = $this->hasVoidReturnAnnotation(...);
```
and use the variable instead.

julienfalque:
Yes I think it's better to ignore this case as it only throws an exception. Maybe you can look only for `throw` statements that are *not* inside a control structure as you cannot be sure those will be executed.

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: They are subtly different, `hasReturnAnnotation` is looking for any non-void return annotation while `hasVoidReturnAnnotation` is looking specifically for the existence of a void return annotation.  So `!$this->hasVoidReturnAnnotation(...)` would be a false positive when there are no PHPDocs at all.

mrmark:
So, I'm having a hard time coming up with a way to check for unconditional throws in a method with no returns/yields.  I could make a dumb check to see if the first statement in the function body is a throw.

At this point though, I'm more in favor of not editing existing return type hints.  What are the reviewers thoughts?  Or does anyone want to jump into the code to help wrap this up?

keradus:
>  I'm more in favor of not editing existing return type hints

if there is already a typehint for output, don't modify it.
responsibility of this fixer is to add void typehint, not to modify typehint

mrmark:
Ready for review again, the return type check is now in the main loop so it will skip functions with return type hints earlier now.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,254 @@
...
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run before ReturnTypeDeclarationFixer and PhpdocNoEmptyReturnFixer
+        return 15;
```
keradus: you forgot to update https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/AutoReview/FixerFactoryTest.php#L47

mrmark:
Updated

keradus:
damn, you squashed commits :( now I have no easy way to only look at code changed since my last comment :(

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,254 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Mark Nielsen
+ */
+final class VoidReturnFixer extends AbstractFixer
+{
+    /**
+     * @internal
```
keradus: :+1:

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,254 @@
...
+    {
+        return new FixerDefinition(
+            'Add void return type to functions with missing or empty return statements, but priority is given to `@return` annotations. Requires PHP >= 7.1.',
+            [
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction foo(\$a) {};",
+                    new VersionSpecification(70100)
+                ),
+            ],
+            'Rule is applied only in a PHP 7.1+ environment.',
```
keradus: no need to duplicate it from main description

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,254 @@
...
+     *
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasReturnAnnotation(Tokens $tokens, $index)
+    {
+        foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
+            if (0 === preg_match(self::VOID_RETURN_PATTERN, $return->getContent())) {
```
keradus: this will catch false positives here like
```php
/**
 * Is there a @return void annotation ?
 *
 * @return bool
 */
```

Use DocBlock -> getAnnotation('return')

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,254 @@
...
+     * @param int    $index  The index of the end of the function definition line, EG at { or ;
+     *
+     * @return bool
+     */
+    private function hasReturnTypeHint(Tokens $tokens, $index)
+    {
+        $endFuncIndex = $tokens->getPrevTokenOfKind($index, [')']);
+        $nextIndex = $tokens->getNextMeaningfulToken($endFuncIndex);
+
+        return $tokens[$nextIndex]->equals([CT::T_TYPE_COLON, ':']);
```
keradus: isGivenKind, as this token could only has `:` as value

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -165,6 +165,8 @@ public function getFixersPriorityCases()
             [$fixers['function_to_constant'], $fixers['no_whitespace_in_blank_line']], // tested also in: function_to_constant,no_whitespace_in_blank_line.test
             [$fixers['list_syntax'], $fixers['binary_operator_spaces']], // tested also in: list_syntax,binary_operator_spaces.test
             [$fixers['list_syntax'], $fixers['ternary_operator_spaces']], // tested also in: list_syntax,ternary_operator_spaces.test
+            [$fixers['void_return'], $fixers['return_type_declaration']], // tested also in: void_return,return_type_declaration.test
+            [$fixers['void_return'], $fixers['phpdoc_no_empty_return']], // tested also in: void_return,phpdoc_no_empty_return.test
```
keradus: where are those files ?

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: I'm using `\PhpCsFixer\DocBlock\DocBlock::getAnnotationsOfType` in `\PhpCsFixer\Fixer\FunctionNotation\VoidReturnFixer::findReturnAnnotations` - think that takes care of it.

tests/AutoReview/FixerFactoryTest.php
mrmark: Sorry, I have no idea what those comments mean.  Should I remove the comments or is there something else I need to do?

mrmark: Ah, found the files - I'll try to figure it out...

mrmark:
Latest round of updates pushed, not squished ;)

src/Fixer/FunctionNotation/VoidReturnFixer.php
keradus: then why using regex here instead of getTypes ?

mrmark: Didn't know about that API and saw another fixer using a regex.  Updated to use the API.

keradus: in that case feel free to fix that another fixer as well, maybe we miss fixing it during DocBlock introduction

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
```diff
@@ -55,6 +55,20 @@ public function provideFixCases()
                  */
                 function foo($param) {}
             '],
+            ['<?php
+                /**
+                 * @return void|array
```
keradus: is it legal ?
how could one make function like that?
`null|array` is possible in 7.1, but I don't think that one could mix void with sth

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: I'm having hard enough time with this one ;) - it's in `\PhpCsFixer\Fixer\Phpdoc\PhpdocNoEmptyReturnFixer::fixAnnotation` if you are curious.

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
mrmark: I wanted to test `void|something` to make sure my new conditions were working.  Guess it should have been `something|void`.  I can remove if you want.  You can do it, you just cannot type hint on it (EG: conditionally return an array or nothing).

keradus: > conditionally return an array or nothing

```php
<?php

function foo() {
    if (false) {
        return [];
    }
}
var_dump(foo());
```

will output NULL, not void. you cannot mix void with sth

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,251 @@
...
+     * @param Tokens $tokens
+     * @param int    $index  The index of the function token
+     *
+     * @return bool
+     */
+    private function hasVoidReturnAnnotation(Tokens $tokens, $index)
+    {
+        foreach ($this->findReturnAnnotations($tokens, $index) as $return) {
+            $types = $return->getTypes();
+            if (1 === count($types) && 'void' === $types[0]) {
```
julienfalque: I would replace this with `if (['void'] === $return->getTypes())`.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,251 @@
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Determine the function already has a return type hint.
```
julienfalque: "Determine _if_".

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,251 @@
...
+     *
+     * @return bool
+     */
+    private function hasVoidReturn(Tokens $tokens, $startIndex, $endIndex)
+    {
+        for ($i = $startIndex; $i < $endIndex; ++$i) {
+            if ($tokens[$i]->isGivenKind(T_YIELD)) {
+                return false; // Do not apply fix as generators cannot return void.
+            }
+            if (!$tokens[$i]->isGivenKind(T_RETURN)) {
```
julienfalque: Please add an empty line before this `if`.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,251 @@
...
+                continue;
+            }
+
+            $nextToken = $tokens->getNextMeaningfulToken($i);
+            if (!$tokens[$nextToken]->equals(';')) {
+                return false; // Do not apply fix, non-empty return statement found.
```
julienfalque: The comment is not relevant here IMO. You should move it to the main tokens iteration or remove it.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,251 @@
...
+                continue;
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+
+            if ($this->hasReturnTypeHint($tokens, $startIndex)) {
+                continue;
+            }
+
+            if (!$tokens[$startIndex]->equals('{')) {
```
julienfalque: Please replace with `if ($tokens[$startIndex]->equals(';'))`.

src/Fixer/FunctionNotation/VoidReturnFixer.php
keradus: I would not, never liked that construction.
but no strong opinion, up to you @mrmark

src/Fixer/FunctionNotation/VoidReturnFixer.php
keradus: "Determine whether"

julienfalque: All PHPDoc descriptions in this class use _Determine if_ so I would use it here as well... or update all the others :)

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: Never tried that, according to docs, it does exactly what we like, plus it should be easier to read, http://php.net/manual/en/language.operators.array.php

Going to go for it.

tests/Fixer/FunctionNotation/VoidReturnFixerTest.php
mrmark: `void` functions still return null:
```
php > $a = function(): void {};
php > var_dump($a());
NULL
```
I'm removing this though since it's a point of contention and I'm no longer worried about this use case due to the updated if statement.

mrmark:
Updated PR, thanks!

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,250 @@
...
+            if ($tokens[$startIndex]->equals(';')) {
+                // No function body defined, fallback to PHPDoc.
+                if ($this->hasVoidReturnAnnotation($tokens, $index)) {
+                    $this->fixFunctionDefinition($tokens, $startIndex);
+                }
+
+                continue;
+            }
+
+            if ($this->hasReturnAnnotation($tokens, $index)) {
```
keradus: if method has body, we are executing `findReturnAnnotations` twice.

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -0,0 +1,250 @@
...
+        for ($i = $startIndex; $i < $endIndex; ++$i) {
+            if ($tokens[$i]->isGivenKind(T_YIELD)) {
+                return false; // Generators cannot return void.
+            }
+
+            if (!$tokens[$i]->isGivenKind(T_RETURN)) {
+                continue;
+            }
+
+            $nextToken = $tokens->getNextMeaningfulToken($i);
```
keradus:
```diff
-$nextToken = $tokens->getNextMeaningfulToken($i);
+$i = $tokens->getNextMeaningfulToken($i);
```

keradus:
:+1: but minors

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: Fixed, thanks

src/Fixer/FunctionNotation/VoidReturnFixer.php
mrmark: Only `hasVoidReturnAnnotation` and `hasReturnAnnotation` call that method, and each of those are only called once (for functions with body and for functions without body).

mrmark:
Rebased and pushed - note that after the rebase, the README regenerated slightly different.

mrmark:
Woot, thanks for the approval and thanks for iterating on this so many times with me.

Anything else I need to do to get this merged?

julienfalque:
@mrmark Nope, everything looks good IMO :)

keradus:
@mrmark , thanks for great work here ;)
this of this PR as finished even if not merged, I will merge it later.

mrmark:
Great, thanks all!

keradus:
Thank you @mrmark.

PR #2742, opened 2017-05-02
Code cleanup

keradus:


GrahamCampbell:
I still strongly feel we should fix our own config file though.

keradus:
Thank you @GrahamCampbell.

PR #2744, opened 2017-05-04
AbstractDoctrineAnnotationFixerTestCase - split fixers test cases

julienfalque:
Currently, each Doctrine Annotation fixers test runs with a DocBlock comment applied to four elements successively: a class, a method, a property and an element that is not supposed to have a DocBlock, like a regular variable.

This patch makes the test data providers return the comments already applied to those elements as actual separate test cases. This means there are four times more test cases now but all are faster than before so they should not be detected as too slow with #2727.


keradus:
Thank you @julienfalque.

PR #2745, opened 2017-05-04
Fix Doctrine Annotation newlines

julienfalque:
~Includes #2744 which should be merged first.~

### DIFF
--- a/src/Doctrine/Annotation/Tokens.php
+++ b/src/Doctrine/Annotation/Tokens.php
@@ -197,7 +197,7 @@ final class Tokens extends \SplFixedArray
                 isset($this[$index + 3])
                 && $this[$index + 2]->isType(DocLexer::T_NONE)
                 && $this[$index + 3]->isType(DocLexer::T_OPEN_PARENTHESIS)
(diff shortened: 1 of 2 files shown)


src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
```diff
@@ -159,7 +161,7 @@ private function fixSpacesAroundParentheses(Tokens $tokens)
             }
 
             if ($token->isType(DocLexer::T_NONE)) {
-                if (false !== strpos($token->getContent(), "\n")) {
+                if (false !== strpos($token->getContent(), $lineEnding)) {
```
keradus: This fixer is **not** adding any newline (nor indent)/
What is the reason to make it `WhitespacesAware` ?

All tests works without that line change.

julienfalque: I wrote tests with `setWhitespacesConfig()` calls in the first place so implementing the interface was required. But indeed, defining the whitespaces config isn't actually required at all.

keradus:
Thank you @julienfalque.

PR #2754, opened 2017-05-06
composer: prepare for new PHPUnit, drop 4.x

TomasVotruba:
PHPUnit 4.x is not supported anymore and [was aimed for PHP 5.3, 5.4 and 5.5](https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.8.0#requirements)

With bump to PHP 5.6, I think we could slowly move to PHPUnit 6.0 starting with this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "gecko-packages/gecko-php-unit": "^2.0",
         "johnkary/phpunit-speedtrap": "^1.1",
         "justinrainbow/json-schema": "^5.0",
-        "phpunit/phpunit": "^4.8.35 || ^5.4",
+        "phpunit/phpunit": "^5.4",
         "satooshi/php-coveralls": "^1.0",
         "symfony/phpunit-bridge": "^3.2.2"
     },


keradus:
ref #2738

julienfalque:
@TomasVotruba Can you please rebase your branch and fix the conflict?

TomasVotruba:
@julienfalque Sure, I've done it via Github GUI right now.

TomasVotruba:
@keradus It looks like you ajdusted the codebase to be compatible with PHPUnit 5.* in https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/3a0ae89fdbcd4ad6c606a70889398dfb2906b1b3

Just to remove old PHPUnit is missing.

If that's not it, what exactly do you mean?

keradus:
To update from PHPUnit X to PHPUnit Y we shall also stop using deprecated features from version X and rely on new features introduced in Y. That's how last update was executed (using future-compat layer)
It shall not be applied by hand, but by **introducing new fixer**.

To drop PHPUnit 4.x we shall:
- stop using annotations (`@expectedException @expectedExceptionCode @expectedExceptionMessage @expectedExceptionMessageRegExp`) - use dedicated method instead
- stop using `setExpectedException` - use `expectExcepted` instead

That is allowing us to nicely drop 4.x and using 5.2 (or 5.4, as we also rely on compat-layer). It's not that we forgot to cut it off as you suggest, and it's way far to 6.x.
Just cutting off available dep version is no gain at all.

TomasVotruba:
I understand the process and if test would fail, I would focus on it.
I didn't know there were some deprecations, since tests are passing.

Also I use `@expectedException` in PHPUnit 6 with no problem.

I see you have much bigger plans than my intentions were, so I can close this PR with peace.

Thanks for explaining your intentions in more detailed way. It's much more pleasant for me to read and understand than :-1:  or ref. I realize it takes some extra effor, so thanks for taking the time :+1:

keradus:
> Also I use `@expectedException` in PHPUnit 6 with no problem.

it works, but best practices for expecting exception (here - to not rely on comment annotation but real code) was changed in PHPUnit 5 already

TomasVotruba:
I've heard about it back and forth, but never came across official source.
Could you please share a link please?

keradus:
https://thephp.cc/news/2016/02/questioning-phpunit-best-practices

TomasVotruba:
Thank you, will read

keradus:
replaced by #2840

PR #2758, opened 2017-05-06
Readme - improve CI integration guidelines

keradus:


hkdobrev:
On a related note, I think the diff filter could be replaced with just lowercase d, which includes everything except for deleted files. It's also more future proof.

keradus:
I would keep it as is, lowercased filter was introduced in git v2

PR #2759, opened 2017-05-08
Update gitattributes

SpacePossum:
exclude some files not needed in export


keradus:
Thank you @SpacePossum.

PR #2763, opened 2017-05-10
Fix describe command with PSR-0 fixer

### LINKED ISSUE #2761: describe command fails for psr0
If you run this command:
`php-cs-fixer describe psr0`

Here's the output:

```
Description of psr0 rule.
Classes must be in a path that matches their namespace, be at least one namespace deep and the class name should match the file name.

This fixer may change your class name, which will break the code that is depended on old name.

* dir (string): the directory where the project code is placed; required

Fixing examples:


  [PhpCsFixer\ConfigurationException\RequiredFixerConfigurationException]
  [psr0] Missing required configuration: The required option "dir" is missing.


describe <name>
```

As you can see, the examples section is broken.

My PHP version: PHP 7.1.4
My PHP-CS-Fixer version: 2.3.1 (d5257f7)

julienfalque:
Fixes #2761.

Also adds a test that runs the `describe` command with all fixers and checks that the exit code is `0`.

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -225,7 +225,13 @@ final class DescribeCommand extends Command
                 $old = $codeSample->getCode();
                 $tokens = Tokens::fromCode($old);
                 if ($fixer instanceof ConfigurableFixerInterface) {
-                    $fixer->configure($codeSample->getConfiguration());
+                    $configuration = $codeSample->getConfiguration();
+
+                    if (null === $configuration) {
+                        $configuration = array();
+                    }
+
+                    $fixer->configure($configuration);
                 }
 
                 $file = $codeSample instanceof FileSpecificCodeSampleInterface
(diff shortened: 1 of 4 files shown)


keradus:
looks cool

src/Console/Command/DescribeCommand.php
```diff
@@ -225,7 +225,13 @@ private function describeRule(OutputInterface $output, $name)
                 $old = $codeSample->getCode();
                 $tokens = Tokens::fromCode($old);
                 if ($fixer instanceof ConfigurableFixerInterface) {
-                    $fixer->configure($codeSample->getConfiguration());
+                    $configuration = $codeSample->getConfiguration();
+
+                    if (null === $configuration) {
```
SpacePossum: isn't this an error in the configuration of the sample itself? i.e. shouldn't we just throw an exception here?

julienfalque: No, that just means the code sample uses the default configuration of the fixer.

The changes here prevent calling deprecated code in such case by passing an empty array instead of `null`, which is deprecated. We didn't notice the deprecated code usage before because that was not covered by tests yet.

keradus: no, it's the interface of `CodeSample` for non-configurable fixers and **not** passing any configuration to the fixer (when you pass `{"fixer": true}` in config)

SpacePossum:
:+1:

:ship:

keradus:
Thank you @julienfalque.

PR #2764, opened 2017-05-11
Add missing PHP71Migration:risky ruleset

teohhanhui:


julienfalque:
Thanks for this contribution. However the changes proposed here are already introduced in #2740 which will probably be merged soon, let's wait and see what happens there first.

Anyway, I don't considerer this a bugfix but a new feature so this PR should target master branch.

keradus:
Closing as no value is bring here.
in such case, you could create 7.0.1, 7.0.2, 7.2 in near future, and all rulesets would be the same.
we create a new ruleset when sth new is gonna to be there, not to just alias rulesets.

PR #2766, opened 2017-05-12
Fix default option of psr0 fixer

### LINKED ISSUE #2761: describe command fails for psr0

mlocati:
Fix #2761

### DIFF
--- a/src/Fixer/Basic/Psr0Fixer.php
+++ b/src/Fixer/Basic/Psr0Fixer.php
@@ -98,7 +98,7 @@ class InvalidName {}
             $path = str_replace('\\', '/', $file->getRealPath());
             $dir = dirname($path);
 
-            if (isset($this->configuration['dir'])) {
+            if ('' !== $this->configuration['dir']) {
                 $dir = substr($dir, strlen(realpath($this->configuration['dir'])) + 1);
 
                 if (false === $dir) {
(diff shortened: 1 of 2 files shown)


src/Fixer/Basic/Psr0Fixer.php
```diff
@@ -158,6 +158,7 @@ protected function createConfigurationDefinition()
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('dir', 'The directory where the project code is placed.'))
                 ->setAllowedTypes(['string'])
+                ->setDefault('')
```
keradus: this is wrong

keradus:
Hi. Thanks for taking look into that issue.
Sadly, I find changing default value as a small bc breaker.
For that and fact that #2761 was closed by alternate fix PR - I'm closing this PR.
Thank you nevertheless

src/Fixer/Basic/Psr0Fixer.php
mlocati: So, should we remove [this example](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/c606b75a87baf017178b95f93406942f7816277f/src/Fixer/Basic/Psr0Fixer.php#L39-L45) from the fixer?

PR #2768, opened 2017-05-12
Tokens::ensureWhitespaceAtIndex - clean up comment check, add check for T_OPEN

SpacePossum:
The current logic assumes comment tokens can have a trailing line break, however the transformers make sure this is never the case anymore.
Second change is that an T_OPEN tag can still have a trailing line break (i.e. not transformed to own whitespace token).
I theory this is also true for close tags, but inserting an white space token after a close tag is not valid in the first place.

### DIFF
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -869,6 +869,85 @@ PHP;
     }
 
     /**
+     * @param string $expected   valid PHP code
+     * @param string $input      valid PHP code
+     * @param int    $index      token index
+     * @param int    $offset
+     * @param string $whiteSpace white space
+     *
+     * @dataProvider provideEnsureWhitespaceAtIndexCases
+     */
+    public function testEnsureWhitespaceAtIndex($expected, $input, $index, $offset, $whiteSpace)
+    {
+        $tokens = Tokens::fromCode($input);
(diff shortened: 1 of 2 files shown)


tests/Tokenizer/TokensTest.php
```diff
@@ -869,6 +869,85 @@ public function assertFindBlockEnd($expectedIndex, $source, $type, $searchIndex)
...
+     * @param string $whiteSpace white space
+     *
+     * @dataProvider provideEnsureWhitespaceAtIndexCases
+     */
+    public function testEnsureWhitespaceAtIndex($expected, $input, $index, $offset, $whiteSpace)
+    {
+        $tokens = Tokens::fromCode($input);
+        $tokens->ensureWhitespaceAtIndex($index, $offset, $whiteSpace);
+
+        $this->assertSame($expected, $tokens->generateCode());
```
SpacePossum: no access to `AbstractFixerTestCase::assertTokens`  :(

keradus:
Thank you @SpacePossum.

tests/Tokenizer/TokensTest.php
keradus: I think it's not working correctly.
Could you please take over the https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2771
and fix 2.2 line as well if needed ?

PR #2769, opened 2017-05-13
Psr0Fixer - remove duplicated example

julienfalque:
In #2763 I fixed the broken example but it was actually identical to the other one so I should have removed it instead.


keradus:
Thank you @julienfalque.

Issue #2693, opened 2017-04-13
The doctrine annotation space fixer should not add spaces before colons for assignments


stof:
while adding spaces on both sides of `=` may make sense (but why space around equals in arrays and no spaces around assignments ? this is a weird rule), I think that `:` in arrays should have a space only after it, not before, to match common JSON conventions (the syntax using `:` is inspired by JSON after all)

julienfalque:
> why space around equals in arrays and no spaces around assignments ?

I couldn't find official conventions about this so I looked at [Doctrine Annotation documentation](http://docs.doctrine-project.org/projects/doctrine-common/en/latest/reference/annotations.html) and [Symfony documention](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/routing.html). Most of the time, they have no space for argument assignments but spaces before and after array assignment. I applied the same rule for array assignments using a colon

> I think that : in arrays should have a space only after it, not before, to match common JSON conventions

This behavior could be tweaked by introducing new configuration options to the fixer.

stof:
Then OK for spaces around `=` in arrays. But none of these documentation pages are using `:` (because it was added only after they were written). And the PR adding support for it was precisely about making it more similar to JSON.

julienfalque:
I wasn't aware of that, makes perfect sense.

julienfalque:
Closing as #2770 was merged.

PR #2664, opened 2017-04-01
Add DoctrineAnnotationArrayAssignmentFixer

julienfalque:
In Doctrine Annotations, you can use `=` or `:` in arrays. This fixer forces the usage of the configured operator:

```diff
 /**
- * @Foo(foo="bar", {bar:"baz"})
+ * @Foo(foo="bar", {bar="baz"})
  */
```

PR #2770, opened 2017-05-13
DoctrineAnnotationSpaces - split assignments options

julienfalque:
This PR addresses #2693 by splitting the opinionated options `around_argument_assignments` and `around_array_assignments` into new specific options to allow more control over how spaces are updated around assignments.

Do you think it would make sense to split `around_parentheses` and `around_commas` as well?

/cc @stof

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -440,14 +440,29 @@ Choose from the list of available rules:
 
   Configuration options:
 
+  - ``after_argument_assignments`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after argument assignment operator
+  - ``after_array_assignments_colon`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after array assignment ``:`` operator
+  - ``after_array_assignments_equals`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after array assignment ``=`` operator
   - ``around_argument_assignments`` (``bool``): whether to fix spaces around
-    argument assignment operator; defaults to ``true``
+    argument assignment operator (deprecated, use
+    ``before_argument_assignments`` and ``after_argument_assignments`` options
+    instead); defaults to ``true``
   - ``around_array_assignments`` (``bool``): whether to fix spaces around array
-    assignment operators; defaults to ``true``
+    assignment operators (deprecated, use ``before_array_assignments_*`` and
+    ``after_array_assignments_*`` options instead); defaults to ``true``
   - ``around_commas`` (``bool``): whether to fix spaces around commas; defaults to
     ``true``
   - ``around_parentheses`` (``bool``): whether to fix spaces around parentheses;
     defaults to ``true``
+  - ``before_argument_assignments`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces before argument assignment operator
+  - ``before_array_assignments_colon`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces before array ``:`` assignment operator
+  - ``before_array_assignments_equals`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces before array ``=`` assignment operator
   - ``ignored_tags`` (``array``): list of tags that must not be treated as Doctrine
     Annotations; defaults to ``['abstract', 'access', 'code', 'deprec',
     'encode', 'exception', 'final', 'ingroup', 'inheritdoc', 'inheritDoc',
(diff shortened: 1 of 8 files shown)


keradus:
@stof, could you take a look please ?

keradus:
@stof ?

keradus:
from #2664 : 
> Note to self: this new fixer must run before doctrine_annotation_spaces once #2770 is merged.

please apply priority in this PR

README.rst
```diff
@@ -374,6 +374,12 @@ Choose from the list of available rules:
...
   Configuration options:
 
+  - ``after_argument_assignments`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after argument assignment operator
+  - ``after_array_assignments_colon`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after array assignment ``:`` operator
+  - ``after_array_assignments_equals`` (``null``, ``bool``): whether to add, remove or
+    ignore spaces after array assignment ``=`` operator
   - ``around_argument_assignments`` (``bool``): whether to fix spaces around
     argument assignment operator; defaults to ``true``
```
stof: IMO, the doc should contain the info that it is deprecated

julienfalque: Done.

julienfalque:
Added an integration test for the priority requirement.

keradus:
ping @julienfalque 
what is the state here?

ping @stof 
any fixer logic concerns here ?

julienfalque:
Rebased to fix conflicts. I think this PR still needs some reviews though.

src/Console/Command/DescribeCommand.php
```diff
@@ -174,14 +174,17 @@ private function describeRule(OutputInterface $output, $name)
                 }
 
                 $description = preg_replace('/(`.+?`)/', '<info>$1</info>', $option->getDescription());
-                $line .= ': '.lcfirst(preg_replace('/\.$/', '', $description)).'; ';
```
keradus: why `;` was moved ?

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
```diff
@@ -76,14 +73,50 @@ protected function createConfigurationDefinition()
...
-                (new FixerOptionBuilder('around_argument_assignments', 'Whether to fix spaces around argument assignment operator.'))
+                (new FixerOptionBuilder('around_argument_assignments', 'Whether to fix spaces around argument assignment operator (deprecated, use `before_argument_assignments` and `after_argument_assignments` options instead).'))
                     ->setAllowedTypes(['bool'])
                     ->setDefault(true)
                     ->getOption(),
-                (new FixerOptionBuilder('around_array_assignments', 'Whether to fix spaces around array assignment operators.'))
+                (new FixerOptionBuilder('before_argument_assignments', 'Whether to add, remove or ignore spaces before argument assignment operator.'))
+                    ->setAllowedTypes(['null', 'bool'])
+                    ->setDefault(function (Options $options) {
+                        return $options['around_argument_assignments'] ? false : null;
```
keradus: why this option depends on another option? :|
it shouldn't

src/Console/Command/DescribeCommand.php
GrahamCampbell: Because it was moved to the parts below.

keradus: that's what my question is about.

julienfalque: I moved it because now there is one case where `;` is not needed: when the default value is computed at runtime with a callable.

src/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixer.php
julienfalque: This is the BC part: using the deprecated options will not directly tweak fixer's behavior anymore but will set the default value of the new options, which then tweak the fixer's behavior.

keradus: I don't like it.
get default value as default for 3.x line, and then, if one passed deprecated value, adjust value of new option outside of definition.

julienfalque: Done.

keradus:
Thank you @julienfalque.

PR #2773, opened 2017-05-15
Travis - use stages

keradus:
Main benefits:
- deployment is done only when whole matrix of previous step is green [using pure matrix without stages, build is executed on each job that matches the conditions (here it was matching for env var set up for one job), but without checking other jobs (so it was possible that 7.1 was green so deployment happen even if 5.6 was red)]
- less conditions (like checking `TASK_*` vars which were local implementation of stages idea)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,104 +5,112 @@ sudo: false
 git:
     depth: 1
 
-env:
-    global:
-        - DEFAULT_COMPOSER_FLAGS="--no-interaction --no-progress --optimize-autoloader"
-        - TASK_TESTS=1
-        - TASK_TESTS_COVERAGE=0
-        - TASK_CS=1
-        - TASK_SCA=0
-
-matrix:
-    fast_finish: true
-    include:
-        - php: 7.1
-          env: DEPLOY=yes TASK_TESTS_COVERAGE=1
-        - php: nightly
-          env: TASK_SCA=1 COMPOSER_FLAGS="--ignore-platform-reqs" SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_IGNORE_ENV=1
-        - php: 5.6
-          env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
-        - php: 7.0
-          env: SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1
-
 cache:
     directories:
         - $HOME/.composer
 
-before_install:
-    # check phpdbg
-    - phpdbg --version 2> /dev/null || { echo 'No phpdbg'; export TASK_TESTS_COVERAGE=0; }
+env:
+    global:
+        - DEFAULT_COMPOSER_FLAGS="--no-interaction --no-progress"
+        - COMPOSER_FLAGS=""
 
+before_install:
     # turn off XDebug
     - phpenv config-rm xdebug.ini || return 0
-
-    # validate tasks configuration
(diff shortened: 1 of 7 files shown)


keradus:
Note: still having some troubles of running extended PHPMD not from project-based vendor dir

phpmd.xml
```diff
@@ -18,10 +18,10 @@
 
     <rule ref="rulesets/naming.xml/ConstantNamingConventions" />
 
-    <rule ref="vendor/mi-schi/phpmd-extension/rulesets/cleancode.xml/DataStructureMethods" />
-    <rule ref="vendor/mi-schi/phpmd-extension/rulesets/cleancode.xml/SwitchStatement" />
+    <rule ref="../../../../../mi-schi/phpmd-extension/rulesets/cleancode.xml/DataStructureMethods" />
```
julienfalque: These relative paths feel weird and make the file unusable without Travis. Should we move it somewhere else?

keradus: it works without travis, it just requires tool to be installed globally

julienfalque: Will it always work with `../../../../../` wherever the project is?

keradus: it's relative path not to cwd, but between installation of phpmd-ext and phpmd

keradus: > instead of moving it, we could rename it phpmd.xml.dist and add phpmd.xml to .gitignore, just like with phpunit.xml.

Totally not get that part... we don't have phpmd dep in composer.json file anymore with this PR.
With `phpunit.xml.dist` cs `phpunit.xml` is the thing that config is automatically pick via phpunit and non-dist config is user-version, while dist is repo version. why would one create other `phpmd.xml` file? phpmd cannot automatically match used config file, user needs to chose one

.travis.yml
```diff
@@ -5,109 +5,119 @@ sudo: false
...
-        tags: true
-        condition: $DEPLOY = yes
+jobs:
+  include:
+    -
+        stage: Static Code Analysis
+        php: 7.1
+        env: COMPOSER_FLAGS="--no-dev --prefer-stable"
+        install:
+            - travis_retry composer global require $DEFAULT_COMPOSER_FLAGS phpmd/phpmd:^2.4.3 mi-schi/phpmd-extension:^4.2 maglnet/composer-require-checker:^0.1.4
```
keradus: this could be moved and stored in dedicated `dev-tools/composer.json` instead, wdyf ?

phpmd.xml
julienfalque: To me, `.dist` means "sane defaults". That would make explicit that the file may not match all cases and that one may use different settings. Anyway, if paths are relative to phpmd installtion, then it should always work.

.travis.yml
SpacePossum: as a file to the repo? 
hmmm I like minimal files repos and we've a lot of "meta" files around already, have a large travis config might not be ideal, but than again the files are typically not reused.
I like the current state of this PR :+1:

keradus: yeah, as file in repo
having the file, one could easily reproduce Travis steps locally (which is what I do with every 2.x->2.y merge)

SpacePossum: makes sense, maybe we can move more dev tooling config to one dir, like `phpmd.xml` for example or would that be impractival?

keradus: The, we could even make composer script to run them for whole project.
We could move phpmd and composer checker only IMO.

julienfalque:
Can we remove the `--optimize-autoloader` flag from command in .appveyor.yml file as well?

keradus:
good catch

SpacePossum:
nice, thanks :)

PR #2771, opened 2017-05-14
[WIP] AssertTokens Trait

keradus:
```
D:\GIT\fork\PHP-CS-Fixer (2.x_assertTokens)
λ vendor\bin\phpunit tests\Tokenizer\TokensTest.php
PHPUnit 5.7.19 by Sebastian Bergmann and contributors.

Testing PhpCsFixer\Tests\Tokenizer\TokensTest
................................................................. 65 / 87 ( 74%)
.........S.....FFF.FFF                                            87 / 87 (100%)

Time: 514 ms, Memory: 4.00MB

There were 6 failures:

1) PhpCsFixer\Tests\Tokenizer\TokensTest::testEnsureWhitespaceAtIndex with data set #0 ('<?php ', '<?php', 0, 1, ' ')
The token at index 0 must be:
{
    "id": 379,
    "name": "T_OPEN_TAG",
    "content": "<?php ",
    "isArray": true,
    "changed": false
},
got:
{
    "id": 321,
    "name": "T_INLINE_HTML",
    "content": "<?php",
    "isArray": true,
    "changed": false
}.
Failed asserting that false is true.

D:\GIT\fork\PHP-CS-Fixer\src\Test\Assert\AssertTokensTrait.php:33
D:\GIT\fork\PHP-CS-Fixer\tests\Tokenizer\TokensTest.php:879

2) PhpCsFixer\Tests\Tokenizer\TokensTest::testEnsureWhitespaceAtIndex with data set #1 ('<?php\n', '<?php', 0, 1, '\n')
The token at index 0 must be:
{
    "id": 379,
    "name": "T_OPEN_TAG",
    "content": "<?php\n",
    "isArray": true,
    "changed": false
},
got:
{
    "id": 321,
    "name": "T_INLINE_HTML",
    "content": "<?php",
    "isArray": true,
    "changed": false
}.
Failed asserting that false is true.

D:\GIT\fork\PHP-CS-Fixer\src\Test\Assert\AssertTokensTrait.php:33
D:\GIT\fork\PHP-CS-Fixer\tests\Tokenizer\TokensTest.php:879

3) PhpCsFixer\Tests\Tokenizer\TokensTest::testEnsureWhitespaceAtIndex with data set #2 ('<?php  ', '<?php', 0, 1, '     ')
The token at index 0 must be:
{
    "id": 379,
    "name": "T_OPEN_TAG",
    "content": "<?php\t",
    "isArray": true,
    "changed": false
},
got:
{
    "id": 321,
[...]

PR #2774, opened 2017-05-15
AssertTokens Trait

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2771

### DIFF
--- a/tests/AutoReview/ProjectCodeTest.php
+++ b/tests/AutoReview/ProjectCodeTest.php
@@ -65,6 +65,7 @@ final class ProjectCodeTest extends TestCase
         \PhpCsFixer\Runner\FileFilterIterator::class,
         \PhpCsFixer\Runner\FileLintingIterator::class,
         \PhpCsFixer\StdinFileInfo::class,
+        \PhpCsFixer\Test\Assert\AssertTokensTrait::class,
         \PhpCsFixer\Test\IntegrationCaseFactory::class,
         \PhpCsFixer\Tokenizer\Transformers::class,
     ];
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -12,6 +12,7 @@
 
 namespace PhpCsFixer\Tests\Tokenizer;
 
+use PhpCsFixer\Test\Assert\AssertTokensTrait;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use PHPUnit\Framework\TestCase;
@@ -26,6 +27,8 @@ use PHPUnit\Framework\TestCase;
  */
 final class TokensTest extends TestCase
 {
+    use AssertTokensTrait;
+
     public function testReadFromCacheAfterClearing()
     {
         $code = '<?php echo 1;';
@@ -873,7 +876,7 @@ PHP;
         $tokens = Tokens::fromCode($input);
         $tokens->ensureWhitespaceAtIndex($index, $offset, $whiteSpace);
 
-        $this->assertSame($expected, $tokens->generateCode());
+        $this->assertTokens(Tokens::fromCode($expected), $tokens);
     }
 
     public function provideEnsureWhitespaceAtIndexCases()
--- a/src/Test/AbstractFixerTestCase.php
+++ b/src/Test/AbstractFixerTestCase.php
@@ -18,6 +18,7 @@ use PhpCsFixer\FixerFactory;
 use PhpCsFixer\Linter\Linter;
 use PhpCsFixer\Linter\LinterInterface;
 use PhpCsFixer\RuleSet;
+use PhpCsFixer\Test\Assert\AssertTokensTrait;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use PhpCsFixer\Utils;
@@ -29,6 +30,8 @@ use Prophecy\Argument;
  */
 abstract class AbstractFixerTestCase extends TestCase
 {
+    use AssertTokensTrait;
+
     /**
      * @var LinterInterface
      */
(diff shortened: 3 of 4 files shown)


SpacePossum:
Follow up on (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2771#issuecomment-301386231)

It turns out we never dealt with `<?php` correctly (note that this must be the only content of the file, so no white space in it what so ever).
This PR fixes assuring with space after a open tag that already has trailing white space (i.e. T_OPEN_TAG followed by T_WHITESPACE).

Fixing the rare case of a while with only `<?php` will require a new token transformer.
(https://3v4l.org/LvoW5 shows the case with short tags enabled, so not helpful here)

SpacePossum:
btw. this method doesn't play well with cleared tokens I think, but lets schedule that tweak for a next PR.

keradus:
@SpacePossum , is it ready? is there a need to push part of this fix into 2.2 line ?

SpacePossum:
the test still fail :|  (src is good, just not the test)

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -65,6 +65,7 @@
         \PhpCsFixer\Runner\FileFilterIterator::class,
         \PhpCsFixer\Runner\FileLintingIterator::class,
         \PhpCsFixer\StdinFileInfo::class,
+        \PhpCsFixer\Test\Assert\AssertTokensTrait::class,
```
julienfalque: Shouldn't the test ignore traits instead?

tests/Tokenizer/TokensTest.php
```diff
@@ -873,44 +876,100 @@ public function testEnsureWhitespaceAtIndex($expected, $input, $index, $offset,
...
+     * @param int    $index      token index
+     * @param int    $offset
+     * @param string $whiteSpace white space
+     *
+     * @dataProvider provideEnsureWhitespaceAtIndexCasesWithOpenTag
+     */
+    public function testEnsureWhitespaceAtIndexWithOpenTag($expected, $input, $index, $offset, $whiteSpace)
+    {
+        if (ini_get('short_open_tag')) {
+            $this->markTestSkipped('The short_open_tag option is required to be disabled.');
```
julienfalque: Please add a note about why this is required (either in the skip message or as a comment).

tests/AutoReview/ProjectCodeTest.php
SpacePossum: wouldn't be nice to trait as well?

tests/Tokenizer/TokensTest.php
SpacePossum: isn't it obvious?

tests/AutoReview/ProjectCodeTest.php
julienfalque: I don't think this is even possible.

tests/Tokenizer/TokensTest.php
julienfalque: Not to me, but maybe I'm missing something.

tests/AutoReview/ProjectCodeTest.php
SpacePossum: we can create a PHPUnit class that imports the Trait and test it?

julienfalque: I'm not a big fan of testing "internal" things, like checking private properties or calling private methods. IMO we should only test public things. If a trait has a public method, then we can test it on the class that use it directly.

SpacePossum: testing the public methods of a trait directly by using it is better than testing it through some usage, I see no diff. with testing public methods of a class directly by creating an instance of the class rather than testing the class through some other class.

SpacePossum:
btw. @julienfalque is right, the fix in Tokens should end up in the 2.x line before we release a new 2.x version

tests/AutoReview/ProjectCodeTest.php
keradus: underlaying reason of creating `trait` is to provide reusable feature that is not in context of abstract entity. thus, that feature could be tested.
after said that, we introduced this trait to test other bugfix, let us not focus in this PR on assertion2trait extraction as we need to update assertion as well (to handle cleared tokens)

keradus:
That's why I asked https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2774#issuecomment-301492813 ;)

please backport src/Tokenizer/Tokens.php and related tests (without trait usage) into 2.2 line

SpacePossum:
@keradus do you want a new PR for the 2.x line?

keradus:
In original PR targeted for 2.2 you complained you cannot use `assertTokens`.
At master branch you could use `assertTokens`, but it reveals bug in original solution on 2.2 line.
Thus, one PR fixing bug shall be targetted on 2.2 (without using of `assertTokens`), while on master there should be only `assertTokens` applied (as test will be already there due to 2.2 bugfix)

keradus:
ping @SpacePossum

keradus:
what about
> Thus, one PR fixing bug shall be targetted on 2.2 (without using of `assertTokens`), while on master there should be only `assertTokens` applied (as test will be already there due to 2.2 bugfix)


?

SpacePossum:
>Thus, one PR fixing bug shall be targetted on 2.2 (without using of assertTokens),

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2783

>while on master there should be only assertTokens applied (as test will be already there due to 2.2 bugfix)

sure, but the fixed test needs to be on master as well

keradus:
tests will be on master, when you will rebase your PR after I will merge new PR to 2.2 line

SpacePossum:
>tests will be on master, when you will rebase your PR after I will merge new PR to 2.2 line

>Thus, one PR fixing bug shall be targetted on 2.2 (without using of assertTokens),
>while on master there should be only assertTokens applied (as test will be already there due to 2.2 bugfix)

if on 2.2 there is only the bug fix of the method in Tokens and on master only the change from content check vs assertTokens check where do the fixed test cases come from?

keradus:
bugfix in src deserves tests update as well, otherwise if wouldn't pass, right?
2.2

keradus:
#2783 is merged, please rebase and squash all to one commit of yours

SpacePossum:
All green, good to go.
All that is left is your commit @keradus so you could reopen https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2771 and merge just your commit without me a `commiter`
:+1:

keradus:
Thank you @SpacePossum.

PR #2776, opened 2017-05-15
Fixer indep. of tokens supplier.

SpacePossum:
I'm feeding the fixers tokens using a diff. source, this makes the fixer not depended on the count check of the Runner/FixCommand on the amount of tokens > 0.

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -105,7 +105,7 @@ echo 1;
      */
     public function isCandidate(Tokens $tokens)
     {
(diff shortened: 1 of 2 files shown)


keradus:
1. why only those fixers
2. this check is actually in `AbstractFixer`, not the `Runner`/`FixCommand` - https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/AbstractFixer.php#L66 , so check is already executed. extra check in `Runner`/`FixCommand` is there only for legacy and is not executed for Fixers extending `AbstractFixer`, as we didn't doing it one day in `AbstractFixer`

SpacePossum:
hmm I wonder how much behind I am atm, these changes come from a stash of changes I've got locally on the fixer in a project where I use it.
Anyway, can go indeed :)

PR #2777, opened 2017-05-15
SelfUpdateCommand - more details on download/copy failure.

SpacePossum:


src/Console/Command/SelfUpdateCommand.php
```diff
@@ -109,14 +110,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $tempFilename = basename($localFilename, '.phar').'-tmp.phar';
         $remoteFilename = sprintf('https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/%s/php-cs-fixer.phar', $remoteTag);
 
+        $details = null;
         try {
-            $copyResult = @copy($remoteFilename, $tempFilename);
-            if (false === $copyResult) {
-                $output->writeln(sprintf('<error>Unable to download new version %s from the server.</error>', $remoteTag));
+            $copyResult = copy($remoteFilename, $tempFilename);
```
keradus: `@` was silenting the warnings you got when sth goes wrong that couldn't be catched

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -109,14 +110,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $tempFilename = basename($localFilename, '.phar').'-tmp.phar';
         $remoteFilename = sprintf('https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/download/%s/php-cs-fixer.phar', $remoteTag);
 
+        $details = null;
         try {
```
julienfalque: I think this `try/catch` is now useless as `copy()` does not throw exceptions.

keradus: would be nice to double check it, maybe it does on 7.x

julienfalque: It only triggers warnings and returns `false` (see https://3v4l.org/SatCS).

SpacePossum: we don't transform warnings/errors to exceptions?

keradus: as discussed, we don't.

SpacePossum:
closing after discussion, cannot be done without BC break

PR #2775, opened 2017-05-15
NoExtraConsecutiveBlankLinesFixer - remove duplicate code sample.

SpacePossum:

PR #2778, opened 2017-05-15
AutoReview - watch that code samples are unique

keradus:
ref #2769, #2775

### DIFF
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Tests\AutoReview;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerDefinition\CodeSampleInterface;
 use PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface;
 use PhpCsFixer\FixerDefinition\VersionSpecificCodeSampleInterface;
 use PhpCsFixer\FixerFactory;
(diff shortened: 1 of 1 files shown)


tests/AutoReview/FixerTest.php
```diff
@@ -77,6 +76,16 @@ public function testFixerDefinitions(FixerInterface $fixer)
                 $tokens
             );
             $this->assertTrue($tokens->isChanged(), sprintf('[%s] Sample #%d is not changed during fixing.', $fixer->getName(), $sampleCounter));
+
+            $duplicatedCodeSample = array_search(
+                $sample,
+                array_slice($samples, 0, $sampleCounter),
+                false
+            );
+            $this->assertEmpty(
```
SpacePossum: maybe `assertFalse` is more strict?

SpacePossum:
:+1: 
please rebase

tests/AutoReview/FixerTest.php
julienfalque: Actually it *must* be used. Currently, the branch contains two duplicated code examples (from the referenced PRs, not included here). The test only detects one because in `psr0`, the duplicated example is found at index 0, so `assertEmpty` passes while `assertFalse` will not.

Also, this could be simplified using:

```php
$this->assertNotContains(
    $sample,
    array_slice($samples, 0, $sampleCounter),
    sprintf('[%s] Duplicated code examples detected.', $fixer->getName()),
    false,
    false
);
```

but we would lose the indexes of duplicated examples, so current code is ok as well :)

julienfalque:
I wanted to add this test too but you were faster than me ;)

However I agree with @SpacePossum, this should target 2.2.

keradus:
done

tests/AutoReview/FixerTest.php
```diff
@@ -49,10 +50,8 @@ public function testFixerDefinitions(FixerInterface $fixer)
         $this->assertNotEmpty($samples, sprintf('[%s] Code samples are required.', $fixer->getName()));
 
         $dummyFileInfo = new StdinFileInfo();
-        $sampleCounter = 0;
-        foreach ($samples as $sample) {
-            ++$sampleCounter;
-            $this->assertInstanceOf('PhpCsFixer\FixerDefinition\CodeSampleInterface', $sample, sprintf('[%s] Sample #%d', $fixer->getName(), $sampleCounter));
+        foreach ($samples as $sampleCounter => $sample) {
+            $this->assertInstanceOf(CodeSampleInterface::class, $sample, sprintf('[%s] Sample #%d', $fixer->getName(), $sampleCounter));
```
julienfalque: You should revert the `::class` notation usage now that you target 2.2.

keradus: aaaaaaaa

PR #2779, opened 2017-05-16
Apply fully multiline rules to the codebase

greg0ire:
From http://www.php-fig.org/psr/psr-2/ :
Argument lists MAY be split across multiple lines, where each subsequent
line is indented once. When doing so, the first item in the list MUST be
on the next line, and there MUST be only one argument per line.


julienfalque:
As discussed in #2440, we won't merge this PR for now.

We might be interested in merging it in lower branch though, as the fixer will be merged in a new minor version. Let's postpone this PR until #2440, we'll decide then what to do here.

PR #2785, opened 2017-05-18
Allow searching for configuration files in parent directories

mlocati:
What about adding a way to look for configuration files in the directories containing what's being fixed?

Before:
```
php-cs-fixer fix --config=/path/to/project/.php-cs /path/to/project/src/File.php
php-cs-fixer fix --config=/path/to/project/.php-cs /path/to/project/src/Dir/File.php
```

After:
```
php-cs-fixer fix --config=- /path/to/project/src/File.php
php-cs-fixer fix --config=- /path/to/project/src/Dir/File.php
```


keradus:
just run the command from root project location, in which you have your project-specific config

.php_cs.cache
```diff
@@ -0,0 +1 @@
...es_casting":true,"no_alias_functions":true,"non_printable_character":true,"php_unit_construct":true,"php_unit_dedicate_assert":true,"psr4":true,"silenced_deprecation_error":true,"array_syntax":{"syntax":"short"},"combine_consecutive_unsets":true,"general_phpdoc_annotation_remove":["expectedException","expectedExceptionMessage","expectedExceptionMessageRegExp"],"header_comment":{"header":"This file is part of PHP CS Fixer.\n\n(c) Fabien Potencier <fabien@symfony.com>\n    Dariusz Rumi\u0144ski <dariusz.ruminski@gmail.com>\n\nThis source file is subject to the MIT license that is bundled\nwith this source code in the file LICENSE."},"heredoc_to_nowdoc":true,"list_syntax":{"syntax":"long"},"no_short_echo_tag":true,"no_unreachable_default_argument_value":true,"no_useless_else":true,"no_useless_return":true,"ordered_class_elements":true,"ordered_imports":true,"php_unit_strict":true,"php_unit_test_class_requires_covers":true,"phpdoc_add_missing_param_annotation":true,"phpdoc_order":true,"semicolon_after_instruction":true,"strict_comparison":true,"strict_param":true},"hashes":{"src\\Console\\ConfigurationResolver.php":-1258164583,"tests\\Console\\ConfigurationResolverTest.php":518865729}}
```
keradus: that's how you end up when wrongly touch config-file handling

mlocati:
> just run the command from root project location

It's not always so easy: I have configured my IDE to run php-cs-fixer on the currently editing file simply by hitting a toolbar icon, but it's quite hard to set the currently working directory since I work with many projects, each of them with its own configuration file

.php_cs.cache
mlocati: That's how I end up when I'm on a hurry 😉

keradus: no. that's the thing. file shall never be generated in this repo

mlocati: Sure! But I run the php-cs-fixer in this repo while developing this PR against the files I modified, and I mistakenly added the cache file to the commit

julienfalque:
I'm 👎 on this: using `-` to search in parent folders looks weird and passing an absolute path should be the way to go IMO.

mlocati:
>  an absolute path should be the way to go IMO.

I explained way I'd need this feature. Could you explain why we should use only absolute paths?

julienfalque:
Can't your IDE provide the root path of the project of the file your are editing?

mlocati:
> Can't your IDE provide the root path of the project of the file your are editing?

Sure it can, but I have some IDE projects that contain multiple PHP projects, so the "root" for the IDE can't be the same as the "root" for php-cs-fixer.

keradus:
Then your IDE should allow to have multiple projects opened at same time having multiple roots available.
I think this change has very narrow usage for non-standard issue, while it brings more problems and maintaining to the repo.

mlocati:
> Then your IDE should allow to have multiple projects opened at same time having multiple roots available.

It does so, but the project structure is not under my control.

> I think this change has very narrow usage for non-standard issue, while it brings more problems and maintaining to the repo.

I think it's much easier to let php-cs-fixer find the right configuration file instead of asking users to `cd` into the right directory and/or requiring the full path of the configuration file. Obviously the final decision is not up to me, btw :wink:

SpacePossum:
Thank you for the suggestion and use-case for it.
However searching up is not a good idea IMHO as we don't no where to stop.

For example I don't want to the fixer to pick some "random" config file I've in a higher level (like `/home/possum/work` when I fix `/home/possum/work/a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/[...]/`)

Therefor `I think it's much easier to let php-cs-fixer find the right configuration file` is not as easy as it sounds. The current way of finding the config file is complex enough and I don't think we need another.

Therefor I'm closing this after discussion.

mlocati:
> For example I don't want to the fixer to pick some "random" config file I've in a higher level

Then don't use `-` as a configuration file...

> Therefor I'm closing this after discussion

No problem for me, I'll continue to use my script wrapped around the call to PCF. I only thought others could have found this feature useful

keradus:
btw, not nice to investigate content of host env from project dir

PR #2787, opened 2017-05-21
Add warnings about missing dom ext and require json ext

keradus:


src/Report/JunitReporter.php
```diff
@@ -34,6 +34,10 @@ public function getFormat()
      */
     public function generate(ReportSummary $reportSummary)
     {
+        if (!extension_loaded('dom')) {
```
SpacePossum: Can we check this when parsing the config? Than the user will know the report cannot be generated before fixing, maybe better UX?

keradus: early check would be nice indeed, but since this is a public method, we need to check it here nevertheless

keradus: let us keep this PR as a quickfix, adding nice ahead-of time warning would be great improvement, but it shall deserve a dedicated PR to provide it in nice version (instead of version provided fast)

Issue #2760, opened 2017-05-10
Short diff option


julienfalque:
Short diff support was introduced by #2554 but reverted by #2724 because of a BC break in the option handling. We can enable this feature again by introducing a new option like `diff-type`.

SpacePossum:
-1 for adding a format that is not defined.
Free formats like the `short diff` one are horror for BC promise as no one can tell if the output is correct or bugged because the format is not defined. This leaves us with BC supporting bugged-output even if everyone would agree it is bug; as we didn't promise anything defined through definition (like `diff 1.4` vs. `diff 2.0`) we cannot touch it.

I think what would be best is to move away from the `half-way-udiff format` we got now towards defined formats only. Till than it is, IMHO, best not to another undefined format.

+1 for closing this FR

robfrawley:
Can't we follow the [GNUs definition of udiff](http://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html)? What defined format would you like to see implemented instead?

SpacePossum:
>Wouldn't it then be a logical conclusion to say our short diff implementation need remain compatible with the GNU's patch utility

That is the point, however... both `sebastian/diff` versions `1.4` and `2.x` even with all there options are _not_ GNU  `udiff` and _not_ GNU `patch` compat.

So I'm still against adding the short diff as proposed in this issue through the linked PR, as it would add an short diff option (whatever that may be) that is not defined by a standard like GNU `udiff`.

keradus:
does `sebastian/diff` v2 supports `udiff` format?
if not, is it rejected `udiff` request? could you provide link for it?

also, could you please compare difference between `sebastian/diff` v2 and udiff?

SpacePossum:
>does sebastian/diff v2 supports udiff format?

Not OOTB, but you can write your own output builder

>if not, is it rejected udiff request? could you provide link for it?

No, it was never offered IIRC. Besides `diff` package never promised to be GNU `udiff` AFAIK.

>also, could you please show difference between sebastian/diff v2 and udiff?

https://github.com/sebastianbergmann/diff/blob/2.0.1/tests/DifferTestTest.php#L47
Here a bunch of tests cases are filtered out because these are known to be non-compat with `patch`.
These tests come from here;
https://github.com/sebastianbergmann/diff/blob/2.0.1/tests/DifferTest.php#L716
where you can see the `expected` part being non-GNU-`udiff`.

[...]

PR #2788, opened 2017-05-22
Add short differ

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2760


tests/RuleSetTest.php
```diff
@@ -494,14 +492,14 @@ private function sort(array &$data)
     }
 
     /**
-     * @param array $values
+     * @param array $data
      *
      * @return bool
      */
-    private function allInteger(array $values)
+    private function allIntegerKeys(array &$data)
```
SpacePossum: prevents reference mismatch

src/Console/ConfigurationResolver.php
```diff
@@ -253,7 +255,21 @@ public function getConfigFile()
     public function getDiffer()
     {
         if (null === $this->differ) {
-            $this->differ = false === $this->options['diff'] ? new NullDiffer() : new SebastianBergmannDiffer();
+            if (false !== $this->options['differ']) {
+                // 'diff' value does not matter here
```
SpacePossum: ?

src/Console/ConfigurationResolver.php
```diff
@@ -253,7 +255,21 @@ public function getConfigFile()
...
+                if ('sbd' === $this->options['differ']) {
```
SpacePossum: what does `long` mean, of which format?

SpacePossum:
>I think this PR should target master (new feature, even if it was already introduced in 2.2 first).

I like more votes on this, but like you said, it was on 2.2 and I see nothing changed that require master now.

>Also, this brings some unrelated cleanups that should be part of a new PR (which could target 2.2).

there is no need for it IMHO

src/Console/ConfigurationResolver.php
julienfalque: If `--differ` option is passed a value then the tool behaves as if `--diff` was passed as well, even if it wasn't.

src/Console/ConfigurationResolver.php
julienfalque: Yeah not very good too... `full-files`?

src/Console/ConfigurationResolver.php
SpacePossum: yes, that is in the README as `default`

src/Console/ConfigurationResolver.php
SpacePossum: you won't _always_ get the `full` file diff even when using the `long` format. 
You get an extended diff in Unified Format (well, on 1.14.1 I hope, before that the format is a bit more loose, i.e. the diff is in a format that can be described as the `sbd` diff. format)

src/Console/ConfigurationResolver.php
julienfalque: This _default_ means that `--diff` is equivalent to `--diff --differ=sbd`, not that `--differ=sbd` is equivalent to `--diff --differ=sbd`. One could expect that changing the diff type does not implicitly enables diff output.

SpacePossum: >One could expect that changing the diff type does not implicitly enables diff output.

I doubt that, why change or set a format of something you don't want?

I'm not going back and forward on copy and branch change requests that were not a problem before unless someone points out what relevant changes have happened between than and now.

keradus:
Nice to pick it up ! :)

1. Passing differ on 2.2 lie works for a while (while it was breaking sth else, so it was reverted), thus I'm fine to introduce it again for 2.2 line.

2. On 2.3 line code for resolving differ was modified, please try to reuse it to minimize conflicts.

3. `php-cs-fixer fix --diff=$SHOW_DIFF --differ=sbd` would not work at all...

SpacePossum:
>php-cs-fixer fix --diff=$SHOW_DIFF --differ=sbd would not work at all...

good thing it doesn't :)

keradus:
that's the thing. why not? if one want to efficiently manage `--diff` option via boolean flag of his, why to override it ?
either:
- `--diff` is source of true shall diff be generated or not, if yes - then pick differ from `--differ` param
- `--diff` or `--differ` could be use, but not both at the same time.

now, it's confusing that one option is silently changing another one.

SpacePossum:
`--diff` never was a boolean flag, I opt removing it on 3.x and use `differ` only

keradus:
Then my bad. In that case let `--differ` override `--diff` and let's deprecate `--diff`

keradus:
ping @SpacePossum

src/Console/Command/FixCommand.php
```diff
@@ -134,18 +135,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-                'using-cache' => $input->getOption('using-cache'),
-                'cache-file' => $input->getOption('cache-file'),
-                'format' => $input->getOption('format'),
-                'diff' => $input->getOption('diff'),
+                'rules' => $passedRules,
+                'show-progress' => $input->getOption('show-progress'),
                 'stop-on-violation' => $input->getOption('stop-on-violation'),
+                'using-cache' => $input->getOption('using-cache'),
                 'verbosity' => $verbosity,
-                'show-progress' => $input->getOption('show-progress'),
```
localheinz: 👍 https://plugins.jetbrains.com/plugin/5919-lines-sorter

SpacePossum: yeah I got that one, picked it up from another comment of you :)

SpacePossum:
one more PR (https://github.com/sebastianbergmann/diff/pull/60) and I can continue on this :)

SpacePossum:
needs to be done in another way

PR #2792, opened 2017-05-22
Use composer-require-checker

keradus:


glensc:
@keradus didn't bother to write down what this is for. so here it is:

https://packagist.org/packages/maglnet/composer-require-checker

A CLI tool to analyze composer dependencies and verify that no unknown symbols are used in the sources of a package. This will prevent you from using "soft" dependencies that are not defined within your composer.json require section.

keradus:
wow, commenting on few-months old PR :)
yeah, obviously no need for copy-paste description available on given project ;)

.composer-require-checker.json
```diff
@@ -0,0 +1,16 @@
+{
+    "symbol-whitelist" : [
+        "HHVM_VERSION_ID", "T_HH_ERROR",
+        "PHPUnit\\Framework\\TestCase", "Prophecy\\Argument",
+        "GeckoPackages\\PHPUnit\\Constraints\\SameStringsConstraint",
```
SpacePossum: With `gecko-packages/gecko-php-unit` in the require (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/composer.json#L21) part if `composer.json` this should no longer be needed.
Just a note to myself to check on this.

PR #2794, opened 2017-05-22
Drop HHVM support

keradus:
1. HHVM don't care about php incompatibilities we raised for years.

2. https://seld.be/notes/php-versions-stats-2017-1-edition
> HHVM usage is not included above in the graph it is at 0.36%

### DIFF
--- a/src/Linter/ProcessLinter.php
+++ b/src/Linter/ProcessLinter.php
@@ -158,14 +158,10 @@ final class ProcessLinter implements LinterInterface
      */
     private function prepareCommand($path)
     {
-        $executable = ProcessUtils::escapeArgument($this->executable);
-
-        if (defined('HHVM_VERSION')) {
-            $executable .= ' --php';
-        }
-
-        $path = ProcessUtils::escapeArgument($path);
-
-        return sprintf('%s -l %s', $executable, $path);
+        return sprintf(
+            '%s -l %s',
+            ProcessUtils::escapeArgument($this->executable),
+            ProcessUtils::escapeArgument($path)
+        );
     }
 }
--- a/tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
+++ b/tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
@@ -494,39 +494,24 @@ final class NoEmptyStatementFixerTest extends AbstractFixerTestCase
      */
     public function testCasesWithShortOpenTag($expected, $input = null)
     {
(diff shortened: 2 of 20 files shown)


tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
```diff
@@ -494,37 +494,24 @@ public function log($msg)
...
+                '<? ',
+                '<? ;',
+            ],
+            [
+                '<?= ',
+                '<?= ;',
+            ],
+        ];
 
         return $cases;
```
julienfalque: `return [ ... ];` directly?

src/Linter/ProcessLinter.php
```diff
@@ -160,10 +160,6 @@ private function prepareCommand($path)
     {
         $executable = ProcessUtils::escapeArgument($this->executable);
 
-        if (defined('HHVM_VERSION')) {
-            $executable .= ' --php';
-        }
-
         $path = ProcessUtils::escapeArgument($path);
 
         return sprintf('%s -l %s', $executable, $path);
```
julienfalque: I would move `ProcessUtils::escapeArgument()` calls here instead of using variables.

tests/Tokenizer/TokensTest.php
```diff
@@ -380,15 +380,6 @@ public function provideMonolithicPhpDetection()
      */
     public function testShortOpenTagMonolithicPhpDetection($source, $monolithic)
     {
-        /*
-         * short_open_tag setting is ignored by HHVM
-         * @see https://github.com/facebook/hhvm/issues/4758
-         */
-        if (!ini_get('short_open_tag') && !defined('HHVM_VERSION')) {
```
julienfalque: The tests fail because of this `if` removal. You should restore it and only remove the `!defined('HHVM_VERSION')` check and the comment above.

keradus: good catch

keradus: done

tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
keradus: done

src/Linter/ProcessLinter.php
keradus: done

keradus:
Thank you @julienfalque.

Issue #2798, opened 2017-05-23
no_unneeded_control_parentheses have no effect


nazarhussain:
PHP version `5.6.30`, CS Fixer version `2.3.1` 

I used command  `./vendor/bin/php-cs-fixer fix -v --diff --dry-run <file path>`

I have set `'no_unneeded_control_parentheses' => ['break', 'clone', 'continue', 'echo_print', 'return', 'switch_case', 'yield'],` in my configuration. 

Here is the sample code: 

```
switch ($status) {
    case 2: {
       // Single line 
    }
    break;

    case 3: {
       // Multline code 
       // Multline code 
       // Multline code 
    }
    break;

}
```

CS Fixer don't detect any error here, as parentheses are not needed at least in one line code block. 

Can any one help out in it.

SpacePossum:
Hi!
Thanks for the detailed report. Your case is known and a feature request to remove such braces has been done here https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1540 , sadly no one found the time to finish a good PR for the feature. As we would like to have it, please let me know if you've time to help out! :)

nazarhussain:
@SpacePossum If you can guide me how to stat with it, I will try to get it done.

SpacePossum:
sure, I started on this one time, it might be a nice base to get going;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1883

in order to get ready:
* the fixer needs to be modernized to the new standards
* it must be able to deal with the following cases currently not covered by the tests:

```php
<?php

$c=  1;
$b = 'c';
$a = 1;

switch($a) {
    case (1 === $a ? true : ${$b}) : {
        echo 456;
        break;
    }

    case ${$b}: {{
        echo 123;
        break;
    }}
}
```

SpacePossum:
other way could be to modify the `no_unneeded_control_parentheses` fixer, which might be easier now I think about it

nazarhussain:
@SpacePossum I believe you reached to solution. As you have more insights to this library you can fix it quickly. That will be great help for community. Because this fix will help to improve code as per `PSR2`

keradus:
Closed by #2929

Issue #1540, opened 2015-11-24
Feature request: `case` without curly braces


gharlan:
To avoid something like this:

``` php
switch ($foo) {
    case 'bar': {
        // ...
    }
}
```

SpacePossum:
:+1: 
(I would like have a fixer to do the reverse, so I could base it on the one you want ;)  )

keradus:
wow, any framework uses braces for case?
just curious

SpacePossum:
don't know of any major framework, I just keep using those ^-^

keradus:
ok, I'm good with it ;)
btw, you need to break a case before closing `}` ?

SpacePossum:
>btw, you need to break a case before closing } ?

not needed, this;
```php
<?php
$a = 0;
switch($a) {
    case 0 : {
        echo '0';
    }
    break;
    default : {
        echo 'test';
    }
}
```
acts the same as:
```php
<?php
$a = 0;
switch($a) {
    case 0 : {
        echo '0';
        break;
    }
    default : {
        echo 'test';
    }
}
```

keradus:
so you still need the break, if you could ommit it (if closing `}` acts as break) it would be cool ;)

julienfalque:
Actually, these extra braces can appear anywhere, not only around `case` body ([example](https://3v4l.org/G2gZK)). This is why having the `break` statement inside or outside the braces has the same result.

AFAIK, these braces have no special meaning (no variable scope or something similar) and I think the fixer should be aware of that and remove them all.

keradus:
just be aware of `$array{$index}` :D

keradus:
Closed by #2929

PR #2799, opened 2017-05-24
NoUnneededControlParenthesesFixer - Remove useless braces after case and default

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2798
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1540

not sure we want to add a configuration option to enable this behavior?

ping @nazarhussain

### DIFF
--- a/src/Fixer/ControlStructure/NoUnneededControlParenthesesFixer.php
+++ b/src/Fixer/ControlStructure/NoUnneededControlParenthesesFixer.php
@@ -55,7 +55,7 @@ final class NoUnneededControlParenthesesFixer extends AbstractFixer implements C
      */
     public function isCandidate(Tokens $tokens)
     {
-        $types = [];
+        $types = [[T_DEFAULT]];
 
         foreach (self::$loops as $loop) {
             $types[] = (array) $loop['lookupTokens'];
@@ -122,6 +122,8 @@ yield(2);
 
         foreach ($tokens as $index => $token) {
             if (!$token->equals('(')) {
+                $this->fixCaseDefault($tokens, $index);
+
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


julienfalque:
I think this behavior was not expected from initial version of the fixer, so I would consider this a new feature which would require a configuration option. The option may be enabled by default in 3.0 though.

Would it be hard to support removing these unneeded braces anywhere in the code (instead of in `case`s only) like mentioned in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1540#issuecomment-300417299 ?

nazarhussain:
@SpacePossum @juliendufresne May be I am wrong, but the configuration option already exists `no_unneeded_control_parentheses`. Is it not good enough for this feature?

SpacePossum:
@nazarhussain the current option is used for `case (1): ` to `case 1:` fixing. We could reuse the option for the new fixing as well. 

>I think this behavior was not expected from initial version of the fixer, so I would consider this a new feature which would require a configuration option. The option may be enabled by default in 3.0 though.

fine and fine by me

>Would it be hard to support removing these unneeded braces anywhere in the code (instead of in cases only) like mentioned in #1540 (comment) ?

that is a bit of work because;
- `{` can mean a lot of things in PHP (like `$a{1}`, `$a{$a}`, `if (){}`, etc.)
- not all useless `{` should be removed, for example in the following code the `{` pair is useless but a lot people like them anyway ;)
  ```php
      if ($a){echo 2;} 
      echo 1;`
  ```

julienfalque:
> not all useless { should be removed, for example in the following code the { pair is useless but a lot people like them anyway ;)
> ```php
>    if ($a){echo 2;} 
>    echo 1;`
> ```

Of course :p

Maybe detecting whether the `{` is preceded by a control structure (or part of a variable variable, e.g. `${$a}`) should be enough? If not done in this PR, would this be considered a bugfix or a new feature?

SpacePossum:
I would say new feature, but not a BC breaking one so on 2.x line ;)

keradus:
1. fixer is suppose to remove useless parenthesis. those are `(`, not `{`. why to remove later?
2. this is behaviour change of existing fixer, for that it's bc breaker.

either it's a new fixer or new, disabled by default, option for existing fixer.
imo, new fixer.

SpacePossum:
new fixer could work;
so we want a black or white list for tokens before `{`?

Blacklist is prop. easiest;
* `{`
* `;`
* `<?php`

not sure this will be safe if the code being fixed is used as an include though

keradus:
> not sure this will be safe if the code being fixed is used as an include though

what does it change? result of php file is same regardless it's executed directly or included.

keradus:
@SpacePossum ? was that intentional close ?

Issue #2802, opened 2017-05-24
NoUselessElseFixer - bug: fixer brace logic


shmax:
Hi there, great project!

I recently ran the fixer on a large code base, and wound up with some subtle bugs. I tracked it down to this original code from my project (I've simplified it to the point where it's nonsensical, but still demonstrates the bug):
```php
function create_joins($value)
{
	$ret = '';

	if (stripos($value,'JOIN ') === false)
	{
		if (count($value))
		{
			$ret = "foo";
		}
		else
			throw new Exception("Value should not be empty");
	}
	else
		$ret .= $value;

	return $ret;
}
```
After running the fixer, I get:
```php
<?php

function create_joins($value)
{
    $ret = '';

    if (stripos($value, 'JOIN ') === false) {
        if (count($value)) {
            $ret = 'foo';
        } else {
            throw new Exception('Value should not be empty');
        }
    }

    $ret .= $value;

    return $ret;
}

```

You can see that the logic has changed; `$ret .= $value` now executes on every call of the create_joins function, whereas before it only executed when the `if` condition failed.

This is with:
```
PHP: 7.0.19
PHP-CS-Fixer: 2.3.1
```

And with the following `.php_cs.dist`:
```php
$finder = new PhpCsFixer\Finder();
$config = new PhpCsFixer\Config('json-schema', 'json-schema style guide');
$finder->in(__DIR__);
/* Based on ^2.1 of php-cs-fixer */
$config
    ->setRules(array(
        // default
        '@PSR2' => true,
        '@Symfony' => true,
        // additionally
        'array_syntax' => array('syntax' => 'short'),
        'binary_operator_spaces' => false,
        'concat_space' => array('spacing' => 'one'),
        'no_unused_imports' => false,
        'no_useless_else' => true,
        'no_useless_return' => true,
        'ordered_imports' => true,
        'phpdoc_no_package' => false,
        'phpdoc_order' => true,
        'phpdoc_summary' => false,
        'pre_increment' => false,
        'trailing_comma_in_multiline_array' => false,
        'simplified_null_return' => false,
    ))
    ->setFinder($finder)
;
return $config;
```
Please advise.

SpacePossum:
Hi, thanks for reporting!
By looking at the samples and digging in my mind I'm pretty sure this is done by the `NoUselessElseFixer and it is a bug.
Till the bug is fixed I advice to switch off the `no_useless_else` rule in your configuration.

[...]

PR #2807, opened 2017-05-26
NoUselessElseFixer - Fix detection of conditional block

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2802

I've modified the fixer to skip all `if`, `elseif`, `else` cases if these miss the `{}` body, this means we loose a bunch of opportunities. However these are edge cases and hard to fix.
So lets bug fix first and I've someone want to fix those edges as well than that can be done in another PR.


shmax:
👍 for the quick PR!

keradus:
did you drop or move some test cases ?

SpacePossum:
yes, I moved these cases:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2807/files#diff-36575c3d3501696daa2da706f9912c92L297

as these are no longer fixed, these are no part of;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2807/files#diff-36575c3d3501696daa2da706f9912c92R776

GrahamCampbell:
Looks pretty good. 👍

keradus:
Thank you @SpacePossum.

PR #2808, opened 2017-05-26
Fixing code style automagically by pre-commit hook

PatchRanger:
## What
I've created a command to be placed inside `composer.json`. It is intended to copy bash-script as pre-commit git hook. Pre-commit git hook checks changes introduced by the commit - and automagically fixes them, according to defined code style standarts. In case if code style standarts are not defined (project missing `.php_cs`) - then a simple notification is shown.

## How
In order to get the power of auto-fixing for your project:
- Create `.php_cs` file, defining your code style rules.
- Require `php-cs-fixer` in `composer.json`.
- Run `composer install`.
=> When you try to commit to your project, your commit will be fixed to correspond code style on-the-fly without aborting commit.

The PR is based on similar experience from https://github.com/mautic/mautic/pull/2462 & https://github.com/mautic/mautic/pull/3157 . I've decided to push this functionality forward to incorporating it into `PHP-CS-Fixer` itself in order to make it re-usable by other projects.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.php_cs
 /composer.lock
 /phpunit.xml
 /vendor/
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,46 @@
+<?php
+
+$header = <<<'EOF'
+This file is part of PHP CS Fixer.
+
+(c) Fabien Potencier <fabien@symfony.com>
+    Dariusz Rumiński <dariusz.ruminski@gmail.com>
+
+This source file is subject to the MIT license that is bundled
+with this source code in the file LICENSE.
+EOF;
+
+return PhpCsFixer\Config::create()
+    ->setRiskyAllowed(true)
+    ->setRules([
+        '@PHP56Migration' => true,
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'array_syntax' => ['syntax' => 'short'],
+        'combine_consecutive_unsets' => true,
+        // one should use PHPUnit methods to set up expected exception instead of annotations
+        'general_phpdoc_annotation_remove' => ['expectedException', 'expectedExceptionMessage', 'expectedExceptionMessageRegExp'],
+        'header_comment' => ['header' => $header],
+        'heredoc_to_nowdoc' => true,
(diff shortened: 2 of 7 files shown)


keradus:
*Let us **not** go into implementation details yet. It's always better to discuss about idea **before** implementing it.*

@juliendufresne , @SpacePossum 
Do we want to have hook on our side? I don't find it that much convenience, yet I see a lot of issues with it. Git vcs could have only one pre-commit hook, with this approach we will either fail to hook-in or override already existing hook. I would rather go in direction to suggest hook manager, like https://github.com/phpro/grumphp

juliendufresne:
hello @keradus I'm thinking you are willing to ping @julienfalque instead of me :)

PatchRanger:
> Let us not go into implementation details yet. It's always better to discuss about idea before implementing it.

As I've mentioned, it is already implemented (by me too) in another project - and I've decided to put this piece of code into correct place. Looks like there could be "even more correct" places (such as the project you pointed to). ))

> Git vcs could have only one pre-commit hook, with this approach we will either fail to hook-in or override already existing hook.

I see your point, it definitely makes sense.

> I would rather go in direction to suggest hook manager, like https://github.com/phpro/grumphp

Interesting, could be the way to go.

juliendufresne:
IMHO, php-cs-fixer can't provide this kind of hook mechanism because it requires PHP and git to be located on the same environment which is not the case for virtual environment such as vagrant or docker.

To take my case as an example, I don't want my host to have PHP installed. I'd rather rely on vagrant or docker. I can't commit within that environment (vagrant or docker) because they are generic and used by every developers of my teams.
In such a case, you cannot use a commit hook unless you know how to run PHP.

keradus:
Sorry @juliendufresne for accidental pinging, no clue why GitHub suggested you would be better fit for that ping :D anyway, input is appreciate !

SpacePossum:
@PatchRanger thanks for PR! I like the code and the idea, but I feel like this might not be the place to add it, or at least not in its current form. Maybe we could create the code as example in the docs?

keradus:
or point to grumphp, as it already do the job pritty well ?

Issue #2805, opened 2017-05-25
ReadmeCommand crashes on phar


keradus:
Affected: v2.2.4+ (shall be fixed for 2.2 line)

Readme command is crashing due to lack of `CHANGELOG` file in phar release.
Previously, we did a file-existence check and use app version as fallback: https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/2427334856f8fd8de63f13f794b3933e334d6f24#diff-80572d2a8ca8b3463f1e6eb9b184d5f2L260

This check was wrongly removed and should be recovered.

PR #2809, opened 2017-05-26
Phar release - fix readme generation

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2805

### DIFF
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -385,13 +385,8 @@ EOF
     private static function getChangeLogFile()
     {
         $changelogFile = __DIR__.'/../../../CHANGELOG.md';
-        if (!is_file($changelogFile)) {
-            return null;
-        }
 
-        $changelogFile = realpath($changelogFile);
-
-        return false === $changelogFile ? null : $changelogFile;
+        return is_file($changelogFile) ? $changelogFile : null;
     }
 
     /**
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -338,10 +338,14 @@ EOF
             return $version;
         }
 
-        $currentMajor = (int) Application::VERSION;
-        $changelogFile = __DIR__.'/../../../CHANGELOG.md';
-        $changelog = @file_get_contents($changelogFile);
+        $changelogFile = self::getChangeLogFile();
+        if (null === $changelogFile) {
+            $version = Application::VERSION;
 
+            return $version;
+        }
+
+        $changelog = @file_get_contents($changelogFile);
         if (false === $changelog) {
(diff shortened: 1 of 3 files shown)


src/Console/Command/HelpCommand.php
```diff
@@ -376,6 +380,21 @@ protected function initialize(InputInterface $input, OutputInterface $output)
...
+     * @return string|null
+     */
+    private static function getChangeLogFile()
+    {
+        $changelogFile = __DIR__.'/../../../CHANGELOG.md';
+        if (!is_file($changelogFile)) {
+            return null;
+        }
+
+        $changelogFile = realpath($changelogFile);
```
keradus: why realpath is needed ?

GrahamCampbell: In order to tell if it exists. It gives `false` if the path doesn't resolve, which is used on the next line.

keradus: we already checked that it exists 4 lines before, and we don't care about linkshere...

GrahamCampbell: the behaviour of realpath inside a phar is different

SpacePossum: That is what I want to try out here as the file should always exists (both phar and non phar). For non-phar it should always be readable, for phar it is not (that is reported). Using `realpath` should result OK for non-phar and `false` for phar so the fallback would be used. This way we won't hide a failed readable check for non-phar.
Lets see what happens ;)

SpacePossum: btw. how can I create a phar locally to test?

GrahamCampbell: I don't think so. It just returns `false` without error. :)

keradus: > That is what I want to try out here as the file should always exists

no, it's not (and it shouldn't)

SpacePossum: all green without it, but I didn't test the phar because I don't know how

keradus:
works !

```
D:\GIT\fork\PHP-CS-Fixer ((detached from SpacePossum/2_2_missing_changelog_handling))
λ php php-cs-fixer readme | head -4
PHP Coding Standards Fixer
==========================

The PHP Coding Standards Fixer tool fixes *most* issues in your code when you

D:\GIT\fork\PHP-CS-Fixer ((detached from SpacePossum/2_2_missing_changelog_handling))
λ php php-cs-fixer.phar readme | head -4
PHP Coding Standards Fixer
==========================

The PHP Coding Standards Fixer tool fixes *most* issues in your code when you
```

keradus:
Thank you @SpacePossum.

PR #2810, opened 2017-05-26
Add Transformers test

SpacePossum:


tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $transformers = $accessible->items;
+
+        $this->assertInternalType('array', $transformers);
+        $this->assertContainsOnlyInstancesOf('PhpCsFixer\Tokenizer\TransformerInterface', $transformers);
+
+        $classes = [];
+        /** @var \SplFileInfo $file */
+        foreach (Finder::create()->files()->in(__DIR__.'/../../src/Tokenizer/Transformer') as $file) {
+            $class = '\\PhpCsFixer\\Tokenizer\\Transformer\\'.$file->getBasename('.php');
+            $this->assertTrue(class_exists($class, false), sprintf('Failed transformer "%s" is loaded by Transformers', $class));
```
julienfalque: This assertion is weak IMO: it will pass even if the class was loaded via some unrelated code, which does not mean an actual instance of the class is in the `items` array. I think it would be better to check that the array contains an instance of the class instead.

SpacePossum: @see line https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2810/files/953a14281ac3290c202885a84bd781b35ab5ca2d#diff-cb57e42e2185428b89f838f3192e2d1bR42

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,58 @@
...
+        $transformers = Transformers::create();
+        $accessible = new AccessibleObject($transformers);
+        $transformers = $accessible->items;
+
+        $this->assertInternalType('array', $transformers);
+        $this->assertContainsOnlyInstancesOf('PhpCsFixer\Tokenizer\TransformerInterface', $transformers);
+
+        $classes = [];
+        /** @var \SplFileInfo $file */
+        foreach (Finder::create()->files()->in(__DIR__.'/../../src/Tokenizer/Transformer') as $file) {
```
keradus: isn't it just copy-paste of transformer logic ?
the generation of tested and expected values shouldn't be the same.
use hardcoded expected value (`[Transformer\Foo::Class, TransformerBar::class, ...]` or sth)

keradus:
ping @SpacePossum

keradus:
?

SpacePossum:
auto closed because I switched to 2.2 with a forced commit back to tip of head of 2.2, will be back soon

tests/Tokenizer/TransformersTest.php
SpacePossum: Additional assert added later on the flow. So this one is still done (as to prevent abstract fails if the class is not loaded) but the possible false positive is dealt with later.

tests/Tokenizer/TransformersTest.php
SpacePossum: done :)

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,76 @@
...
+ * @internal
+ *
+ * @covers \PhpCsFixer\Tokenizer\Transformers
+ */
+final class TransformersTest extends TestCase
+{
+    public function testCreateOneTime()
+    {
+        $transformers = Transformers::create();
+        $this->assertSame($transformers, Transformers::create());
```
julienfalque: I would add an assertion (maybe in a dedicated test method) that `$transformers` is an instance of `Transformers`.

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,76 @@
...
+            \PhpCsFixer\Tokenizer\Transformer\ArrayTypehintTransformer::class,
+            \PhpCsFixer\Tokenizer\Transformer\SquareBraceTransformer::class,
+            \PhpCsFixer\Tokenizer\Transformer\NamespaceOperatorTransformer::class,
+            \PhpCsFixer\Tokenizer\Transformer\ClassConstantTransformer::class,
+            \PhpCsFixer\Tokenizer\Transformer\ReturnRefTransformer::class,
+        );
+
+        $classes = array();
+        foreach ($transformersClasses as $class) {
+            $this->assertTrue(class_exists($class, false), sprintf('Failed transformer "%s" is loaded by Transformers.', $class));
```
julienfalque: This assertion remains weak and I think the other ones you added below allow you to drop it entirely.

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,76 @@
...
+            $transformer = new $class();
+            if (PHP_VERSION_ID >= $transformer->getRequiredPhpVersionId()) {
+                $classes[$class] = true;
+            }
+        }
+
+        $accessible = new AccessibleObject($transformers);
+        /** @var array $transformers */
```
julienfalque: Is this comment useful?

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+        $this->assertInternalType('array', $transformers);
+        $this->assertCount(count($classes), $transformers);
+        $this->assertContainsOnlyInstancesOf('PhpCsFixer\Tokenizer\TransformerInterface', $transformers);
+
+        foreach ($transformers as $transformer) {
+            unset($classes[get_class($transformer)]);
+        }
+
+        $this->assertCount(0, $classes);
```
julienfalque: I would add an assertion message to make its purpose explicit.

tests/Tokenizer/TransformersTest.php
SpacePossum: I like the fail message over the generic PHP crash and burn one on class does not exists

tests/Tokenizer/TransformersTest.php
SpacePossum: yes, it provides type hinting for IDE's (so it knows the iterating (as done later) of the var is OK)

tests/Tokenizer/TransformersTest.php
julienfalque: If the class does not exist when trying to instantiate it on the line below, it means there is an error in the test itself, not in the tested code: one forgot to update the list when removing a Transformer class. It's quite unlikely and easy to fix anyway, not worth an assertion for that IMO.

SpacePossum: > If the class does not exist when trying to instantiate it on the line below, it means there is an error in the test itself

no, its the job of the unit being tested (Transformers) to (lazy) load the class

SpacePossum:
AppVeyor fail is unrelated.

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,77 @@
...
+        $classes = array();
+        foreach ($transformersClasses as $class) {
+            $this->assertTrue(class_exists($class, false), sprintf('Failed transformer "%s" is loaded by Transformers.', $class));
+            $transformer = new $class();
+            if (PHP_VERSION_ID >= $transformer->getRequiredPhpVersionId()) {
+                $classes[$class] = true;
+            }
+        }
+
+        $accessible = new AccessibleObject($transformers);
```
keradus: don't like digging into internals of tested object here :(

SpacePossum: maybe the class is to closed to be tested properly without digging in?

julienfalque: Don't like too but indeed we currently have no choice if we want to test this.

SpacePossum:
closing as there is not much we can test without looking into the internals

keradus:
hey, don't feel sad about it. nice try, but I agree that proper testing of this class is not an easy task

PR #2812, opened 2017-05-30
ProcessLinter - compatibility with Symfony 3.3

keradus:
Sadly, 3.3 introduced new deprecation (#21474).

One can not operate with `Process` directly to have same code on 3.2 and 3.3 that is not triggering the deprecation on 3.3 (as to avoid triggering deprecation one must to pass array to Process, which is not possible on 3.2). For that, I go behind `ProcessBuilder`, which has same interface between 3.2 and 3.3...

### DIFF
--- a/src/Linter/ProcessLinter.php
+++ b/src/Linter/ProcessLinter.php
@@ -16,7 +16,7 @@ use PhpCsFixer\FileRemoval;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
-use Symfony\Component\Process\ProcessUtils;
+use Symfony\Component\Process\ProcessBuilder;
 
 /**
  * Handle PHP code linting process.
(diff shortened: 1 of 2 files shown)


src/Linter/ProcessLinter.php
```diff
@@ -154,22 +154,17 @@ private function createProcessForSource($source)
...
      */
-    private function prepareCommand($path)
+    private function prepareProcess($path)
     {
-        $executable = ProcessUtils::escapeArgument($this->executable);
-
+        $arguments = array('-l', $path);
         if (defined('HHVM_VERSION')) {
-            $executable .= ' --php';
+            array_unshift($arguemtns, '--php');
```
SpacePossum: `$arguemtns` is a typo I think :)

keradus: :+1:

keradus:
and now the windows is failing (code is fine, we need to adjust tests)

SpacePossum:
awesome :+1: 
thanks :)

PR #2816, opened 2017-06-01
Tokenizer - better docs and validation

keradus:


src/Tokenizer/Token.php
```diff
@@ -217,6 +222,8 @@ public function getPrototype()
     /**
      * Get token's content.
      *
+     * It shall be used only for getting the content of token, not for checking it against excepted value.
```
GrahamCampbell: Very long line. Maybe wrap at 80 chars?

keradus: big :-1: for 80chars, we do not live in 800x600 screens age

GrahamCampbell: We do when it comes to reading github diffs...

image

keradus: poor for github.
we don't follow 80chars in code at all

src/Tokenizer/Token.php
```diff
@@ -61,9 +61,14 @@ public function __construct($token)
             $this->isArray = true;
             $this->id = $token[0];
             $this->content = $token[1];
-        } else {
+        } elseif (is_string($token)) {
             $this->isArray = false;
             $this->content = $token;
+        } else {
+            throw new \InvalidArgumentException(sprintf(
```
SpacePossum: please add utest for this

keradus: that's what happen when you extract not-related, side-off changes from another, big PR...

SpacePossum: the downside of nice reviewable PR's and clean commit history ;)

keradus: tests added

SpacePossum: :+1: thanks :)

tests/Tokenizer/TokenTest.php
```diff
@@ -285,24 +285,34 @@ public function provideIsWhitespaceCases()
         );
     }
 
-    public function testPropertiesOfArrayToken()
+    /**
+     * @param array|string|mixed $prototype
```
julienfalque: I would replace with `mixed` only.

keradus: I was wondering how to nicely write it, `array|string` is valid subset of types, while `mixed` is unvalid subset. together it's `array|string|mixed`.

julienfalque: `array|string|mixed` sounds like _"can be anything, or an array or a string"_ which feels weird and is still an invalid subset.

keradus: done

PR #2817, opened 2017-06-01
Tokenizer - use future-compatible interface

keradus:


src/Fixer/ArrayNotation/NoTrailingCommaInSinglelineArrayFixer.php
```diff
@@ -84,7 +84,7 @@ private function fixArray(Tokens $tokens, $index)
 
         if ($beforeEndToken->equals(',')) {
             $tokens->removeTrailingWhitespace($beforeEndIndex);
-            $beforeEndToken->clear();
```
julienfalque: `$beforeEndToken` variable is now used only once, so you can pass its value directly here instead.

src/Fixer/ArrayNotation/TrimArraySpacesFixer.php
```diff
@@ -71,12 +71,14 @@ private static function fixArray(Tokens $tokens, $index)
             $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_ARRAY_SQUARE_BRACE, $startIndex);
         }
 
-        $nextToken = $tokens[$startIndex + 1];
+        $nextIndex = $startIndex + 1;
+        $nextToken = $tokens[$nextIndex];
```
julienfalque: `$nextToken` variable is now used only once, so you can pass its value directly instead.

src/Fixer/ArrayNotation/TrimArraySpacesFixer.php
```diff
@@ -71,12 +71,14 @@ private static function fixArray(Tokens $tokens, $index)
...
         $nextNonWhitespaceIndex = $tokens->getNextNonWhitespace($startIndex);
         $nextNonWhitespaceToken = $tokens[$nextNonWhitespaceIndex];
         $tokenAfterNextNonWhitespaceToken = $tokens[$nextNonWhitespaceIndex + 1];
 
-        $prevToken = $tokens[$endIndex - 1];
+        $prevIndex = $endIndex - 1;
+        $prevToken = $tokens[$prevIndex];
```
julienfalque: `$prevToken` variable is now used only once, so you can pass its value directly instead.

src/Fixer/FunctionNotation/ReturnTypeDeclarationFixer.php
```diff
@@ -76,14 +77,15 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 continue;
             }
 
-            $previousToken = $tokens[$index - 1];
+            $previousIndex = $index - 1;
+            $previousToken = $tokens[$previousIndex];
```
julienfalque: `$previousToken` variable is now used only once, so you can pass its value directly instead.

src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -94,38 +94,6 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     }
 
     /**
-     * @param Tokens $tokens
-     * @param int    $endIndex
-     */
-    private function fixWhiteSpaceAroundSequence(Tokens $tokens, $endIndex)
```
julienfalque: 🤘

src/Tokenizer/Tokens.php
```diff
@@ -858,6 +860,16 @@ public function isChanged()
         return false;
     }
 
+    public function isEmptyAt($index)
```
julienfalque: Not sure this read-only method really adds value.

src/Fixer/ArrayNotation/NoTrailingCommaInSinglelineArrayFixer.php
keradus: it doesn't matter for performance
and we often do things like here - have both `$fooIndex` and `$fooToken` vars

src/Tokenizer/Tokens.php
keradus: actually, it's a left-over from a idea that was not decided.
I'm wondering about dropping (on 2.x line deprecating) `Token::isEmpty` in favour of `Tokens:isEmptyAt`.
WDYT ?

SpacePossum: what does `empty` mean here,` no content`, `''` or cleared token?

keradus: no data attached to token. currently, it means `new Token('')` (here and on 2.2-master line)
we cannot make `$tokens[$index] = null` or `unset($tokens[$index])` easy, as (beside BC breaker) it would require change for all collection looping...

SpacePossum: than the change from Token to Tokens as proposed here makes sense to me :)

PR #2818, opened 2017-06-01
Token become immutable, performance optimizations

keradus:
TODO:
- [x] rebase on master

```
Executing on Symfony 2.7 with 45files that reports violations

before
> Checked all files in 77.743 seconds, 20.000 MB memory used
> Checked all files in 78.761 seconds, 20.000 MB memory used

after improvement
> Checked all files in 70.417 seconds, 20.000 MB memory used
> Checked all files in 71.144 seconds, 20.000 MB memory used

About 10% improvement
```

Deps:
- [x] #2819
- [x] #2816
- [x] #2817
- [x] #2823

### DIFF
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -59,6 +59,10 @@ class Token
             $this->isArray = true;
             $this->id = $token[0];
             $this->content = $token[1];
+
+            if ($token[0] && '' === $token[1]) {
+                throw new \InvalidArgumentException('Cannot set empty content for id-based Token.');
+            }
         } elseif (is_string($token)) {
             $this->isArray = false;
             $this->content = $token;
@@ -96,9 +100,14 @@ class Token
      * Clear token at given index.
      *
      * Clearing means override token by empty string.
+     *
+     * @deprecated since 2.4
      */
     public function clear()
     {
+        @trigger_error(__METHOD__.' is deprecated and will be removed in 3.0.', E_USER_DEPRECATED);
+        Tokens::setLegacyMode(true);
+
         $this->content = '';
         $this->id = null;
         $this->isArray = false;
@@ -109,6 +118,9 @@ class Token
      */
     public function clearChanged()
     {
+        @trigger_error(__METHOD__.' is deprecated and will be removed in 3.0.', E_USER_DEPRECATED);
+        Tokens::setLegacyMode(true);
+
         $this->changed = false;
     }
 
@@ -359,9 +371,13 @@ class Token
      * Check if token is empty, e.g. because of clearing.
      *
      * @return bool
+     *
+     * @deprecated since 2.4
      */
     public function isEmpty()
     {
+        @trigger_error(__METHOD__.' is deprecated and will be removed in 3.0.', E_USER_DEPRECATED);
+
         return null === $this->id && ('' === $this->content || null === $this->content);
     }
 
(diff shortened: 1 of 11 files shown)


src/Tokenizer/Tokens.php
```diff
@@ -1085,18 +1175,26 @@ public function isMonolithicPhp()
...
+
+            return true;
         }
 
-        return true;
+        if ($this->isTokenKindFound(T_INLINE_HTML)) {
+            return false;
+        }
+
+        return 1 >= ($this->countTokenKind(T_OPEN_TAG) + $this->countTokenKind(T_OPEN_TAG_WITH_ECHO));
```
julienfalque: Yoda style in expressions using `<` or `>` is confusing IMO, I would swap the operands.

src/Tokenizer/Tokens.php
```diff
@@ -1218,11 +1314,41 @@ private function changeCodeHash($codeHash)
...
+     */
+    private function extractTokenKind($token)
+    {
+        return $token instanceof Token
             ? ($token->isArray() ? $token->getId() : $token->getContent())
             : (is_array($token) ? $token[0] : $token)
-        ;
-
-        $this->foundTokenKinds[$tokenKind] = true;
+            ;
```
julienfalque: This semicolon should be less indented.

src/Tokenizer/Tokens.php
```diff
@@ -874,9 +945,14 @@ public function clearAt($index)
      *
      * @param int                $index
      * @param Token|array|string $token token prototype
+     *
+     * @deprecated since 2.4, use offsetSet instead
      */
     public function overrideAt($index, $token)
     {
+        @trigger_error(__METHOD__.' is deprecated and will be removed in 3.0, use offsetSet instead.', E_USER_DEPRECATED);
+        self::$isLegacyMode = true;
```
keradus: no, it would change meaning of code.

src/Tokenizer/Tokens.php
keradus: we use yoda across the project in such cases nevertheless

src/Tokenizer/Tokens.php
keradus: done

src/Tokenizer/Tokens.php
```diff
@@ -68,11 +68,20 @@ class Tokens extends \SplFixedArray
...
      *
-     * @var bool[]
+     * @var array<string|int, int>
      */
     private $foundTokenKinds = array();
 
     /**
+     * @var bool
+     *
+     * @internal
```
keradus: done

src/Tokenizer/Token.php
```diff
@@ -61,6 +61,10 @@ public function __construct($token)
             $this->isArray = true;
             $this->id = $token[0];
             $this->content = $token[1];
+
+            if ($token[0] && '' === $token[1]) {
+                throw new \InvalidArgumentException('Cannot set empty content for id-based Token.');
```
keradus: missing validation

src/Tokenizer/Token.php
```diff
@@ -495,6 +517,7 @@ public function setContent($content)
 
         // setting empty content is clearing the token
         if ('' === $content) {
+            @trigger_error(__METHOD__.' shall not be use to clear token, use Tokens::clearAt instead.', E_USER_DEPRECATED);
```
keradus: > and will throw an exception in 3.0

it won't, as the method won't be there at all

julienfalque: Then isn't this deprecation notice useless as we already have the one for the method deprecation itself?

keradus: not fully, as when set with empty content, we raise even bigger issue to dev

keradus:
> Please add tests that cover `Tokens` changes regarding "immutable mode".

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2818/commits/9b63057380de9a557b7bbc8affcca64a8bc1b92e

src/Tokenizer/Token.php
julienfalque: 👍 but please fix the typo: _shall not be use**d**_.

keradus: done

PR #2820, opened 2017-06-02
MagicConstantCasingFixer - Remove defined check

SpacePossum:
This test is no longer needed on 2.3 and up.

https://3v4l.org/ufX57


keradus:
Thank you @SpacePossum.

Issue #2806, opened 2017-05-26
Simplify property declarations


petr-buchin:
Can I use php-cs-fixer to simplify such property declarations:

```php
private $myProperty = null;
private $yetOneProperty = null;
```

as follows ?

```php
private $myProperty;
private $yetOneProperty;
```

Thanks in advance!

SpacePossum:
Hi!
Currently we've got a Fixer for this, however it shouldn't be a hard one to build.
Please let me know if you can help out, happy to help out :)

petr-buchin:
@SpacePossum , it would be very nice to have this feature.

Unfortunately, for now I cannot help with this, but I will happy to help after first stable release of [MongoDB ODM](https://github.com/tequila/mongodb-odm), which is planned to the June, 2017. 
For now ODM is using [zend-code](https://github.com/zendframework/zend-code) lib, which generates pretty ugly code, for example always generates property like `private $property = null;`, i.e. with `null` default value, so I'm wondering whether there is a quick solution for this for the ODM users, before I release normal library for code generation :)

SpacePossum:
Alright, no rush!
Lets see who can start on this first, you or I, and than we can get this fixer in.
Good luck with MongoDB ODM, looks promising :)

petr-buchin:
@ntzm , @SpacePossum , @keradus, @julienfalque  - thank you guys!

SpacePossum:
you're welcome @petr-buchin , enjoy the next release :)

PR #2822, opened 2017-06-04
Add NoNullPropertyInitializationFixer

ntzm:
Implements #2806

Should this be configurable? i.e. if somebody wanted to have `= null;` on all of them?

Thanks for taking the time to review!


src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,79 @@
...
+            $equalsTokenIndex = $tokens->getNextMeaningfulToken($varTokenIndex);
+
+            if (!$tokens[$equalsTokenIndex]->equals('=')) {
+                continue;
+            }
+
+            $valueTokenIndex = $tokens->getNextMeaningfulToken($equalsTokenIndex);
+            $valueToken = $tokens[$valueTokenIndex];
+
+            if ($valueToken->isGivenKind(T_STRING) && $valueToken->getContent() === 'null') {
```
ntzm: I feel like there's a better way to detect `null`?

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,79 @@
...
+            $equalsTokenIndex = $tokens->getNextMeaningfulToken($varTokenIndex);
+
+            if (!$tokens[$equalsTokenIndex]->equals('=')) {
+                continue;
+            }
+
+            $valueTokenIndex = $tokens->getNextMeaningfulToken($equalsTokenIndex);
+            $valueToken = $tokens[$valueTokenIndex];
+
+            if ($valueToken->isNativeConstant() && strtolower($valueToken->getContent()) === 'null') {
```
julienfalque: Please use yoda style for the second condition.

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,79 @@
...
+
+            $varTokenIndex = $tokens->getNextMeaningfulToken($index);
+
+            if (!$tokens[$varTokenIndex]->isGivenKind(T_VARIABLE)) {
+                continue;
+            }
+
+            $equalsTokenIndex = $tokens->getNextMeaningfulToken($varTokenIndex);
+
+            if (!$tokens[$equalsTokenIndex]->equals('=')) {
```
julienfalque: I think support for grouped properties is missing, e.g.:
```diff
-<?php class Foo { public $foo = null, $bar = null; }
+<?php class Foo { public $foo, $bar; }
```

julienfalque:
> Should this be configurable? i.e. if somebody wanted to have `= null;` on all of them?

Not sure it would be widely used so maybe not worth it. I would suggest to focus on current behavior and see later for the configuration option.

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,79 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
```
keradus: use regular `for` loop, and then you could make a jump of `$index = $valueTokenIndex;`, so you won't iterate tokens between those twice

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
keradus: do not use `getContent` for comparision purposes.
use `->equals(..., false)` instead (false for case-insensitive)

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
```diff
@@ -0,0 +1,79 @@
...
+            ],
+            [
+                '<?php class Foo { public $bar; protected $baz; }',
+                '<?php class Foo { public $bar = null; protected $baz = null; }',
+            ],
+            [
+                '<?php class Foo { public $bar = \'null\'; }',
+            ],
+            [
+                '<?php class Foo { public $bar; }',
```
keradus: not needed, this case is autogenerated from 1st case in this provider

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
```diff
@@ -0,0 +1,79 @@
...
+     * @dataProvider provideCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideCases()
+    {
+        return [
```
keradus: please add case with property declared as final

keradus:
I see no reason for configuring this fixer to be honset

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
ntzm: It's not possible to declare a property as final in PHP

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
ntzm: Working on this now

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
keradus: too much java lately ;)

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
```diff
@@ -20,6 +20,8 @@
  * @internal
  *
  * @covers \PhpCsFixer\Fixer\ClassNotation\NullPropertyDeclarationFixer
+ *
+ * @group new
```
keradus: i bet this is temporary. you could achieve same with `--filter` param without modifying the code itself

ntzm: Yep, forgot to remove! Thanks for the tip!

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
ntzm: Having a bit of a nightmare getting this working lol, gonna give it a go when I have the chance in the week

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
```diff
@@ -0,0 +1,76 @@
...
+            [
+                '<?php class Foo { public $bar; protected $baz; }',
+                '<?php class Foo { public $bar = null; protected $baz = null; }',
+            ],
+            [
+                '<?php class Foo { public $bar = \'null\'; }',
+            ],
+            [
+                '<?php class Foo { public function bar() { return null; } }',
+            ],
```
SpacePossum: please add the tests from here as well;
https://github.com/SpacePossum/PHP-CS-Fixer/commit/756cd62c5b91383079de0491faadd5d7de110b72#diff-f58632716eb903df26e8da18f82e6016R42
(at least the one that cover a case not list here ;) )

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: What is causing you trouble? Can we help you?

keradus: feel free to check out visibility fixer, which have to play with multiple properties as well

ntzm: Just trying to detect when it's a multiple properties, thanks @keradus I will check that out tonight!

keradus:
ping @ntzm

ntzm:
Sorry I haven't got the time to complete this, anyone is free to use all my code without crediting they want it to finish off the feature! Sorry about that

julienfalque:
No worries. Too bad you didn't have time to complete this PR but I think that what you already did is great and should be kept so I'm reopening it. If you're ok with that, I'd like to push commits direcly here to add the missing part as soon as I can.

ntzm:
Oh yeah that's fine sorry!!! Thanks for completing the rest of it 😆

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
SpacePossum: missing anonymous class, traits etc. @ see https://github.com/SpacePossum/PHP-CS-Fixer/commit/756cd62c5b91383079de0491faadd5d7de110b72#diff-f58632716eb903df26e8da18f82e6016R42

julienfalque: Tests added.

SpacePossum: Still missing traits, multiple nested classes and multiple non nested classes in one file. Pease check the link for more test cases.

julienfalque: Added more tests. I think coverage is fine now.

SpacePossum: almost ;)  please update at least one test with diff. casing for `var` (like `VAR`) and one for scoping (like `PuliC`),
other missing is `public const $a = null` on PHP 7.1. and a test where the properties are defined after or in between methods.
And an un-godly comment one like;
```php
class A {
#
private#
#
$a#
#
=#
#
null#
,#
$b#
=#
null#
#
;#
#
}
```

After that I'm out of ideas

julienfalque: Added all cases except the one for `public const $a = null`, the assignment part is required when declaring a constant, even with `null`.

keradus: then please add `public const` case to prove it doesn't wrongly remove initial value

julienfalque: Of course, what was I thinking 😑. Test cases added.

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,84 @@
...
+ * @author ntzm
+ */
+final class NullPropertyDeclarationFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_PUBLIC, T_PROTECTED, T_PRIVATE, T_VAR]);
```
SpacePossum: T_CLASS or T_TRAIT must be found together with the current check

keradus: no need for, you cannot have `T_PUBLIC` etc without classy token, as it would be invalid code

SpacePossum: you can have T_PUBLIC in an interface which has no properties so is never a candidate

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -66,10 +66,14 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 $index = $tokens->getNextMeaningfulToken($index);
 
                 if ($tokens[$index]->equals('=')) {
-                    $valueTokenIndex = $index = $tokens->getNextMeaningfulToken($index);
+                    $index = $tokens->getNextMeaningfulToken($index);
 
                     if ($tokens[$index]->equals([T_STRING, 'null'], false)) {
-                        $tokens->clearRange($varTokenIndex + 1, $valueTokenIndex);
+                        for ($i = $varTokenIndex + 1; $i <= $index; ++$i) {
+                            if (!$tokens[$i]->isComment() && !($tokens[$i]->isWhitespace() && false !== strpos($tokens[$i]->getContent(), "\n"))) {
```
julienfalque: I would swap the whitespace and the comment checks: whitespaces are more likely to be found here than comments, this should make this `if` slightly faster in such case.

Also, I would remove whitespaces even when they contain a newline (except when preceded by a single-line comment), WDYT?

SpacePossum: :+1: for the first (can you do it?),
second not so much, we have better fixers to deal with line breaks

julienfalque: Done.

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
```diff
@@ -0,0 +1,190 @@
...
+                '<?php class Foo { private $bar; }',
+                '<?php class Foo { private $bar = null; }',
+            ],
+            [
+                '<?php class Foo { var $bar; }',
+                '<?php class Foo { var $bar = null; }',
+            ],
+            [
+                '<?php class Foo { VAR $bar; }',
+                '<?php class Foo { VAR $bar = null; }',
```
keradus: we need a test with `= \null`, which won't be fixed with current implementation

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,91 @@
...
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author ntzm
+ */
+final class NullPropertyDeclarationFixer extends AbstractFixer
```
keradus: One day we normalized names to actually provide info what will be fixed version.
Here, we don't know

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,91 @@
...
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author ntzm
```
keradus: do you want to add you here guys ? @julienfalque @SpacePossum

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
```diff
@@ -0,0 +1,91 @@
...
+ */
+final class NullPropertyDeclarationFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Simplifies null class property declarations.',
```
keradus: simplifies... how ?

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: I never add myself as `@author` so no, thanks :)

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: `NoNullPropertyFixer`?

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: What about _"Removes class properties initialization when the value is `null`."_?

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
keradus: whole property would be removed? (even with that, code would work, so...)
`NoNullPropoertyInitializationFixer` ?

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
julienfalque: Good catch!

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
keradus: I bet it would follow decision from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2822#discussion_r123598808 concern

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: > even with that, code would work, so...

Haha, PHP...

👍 for `NoNullPropertyInitializationFixer`.

src/Fixer/ClassNotation/NoNullPropertyInitializationFixer.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+final class NoNullPropertyInitializationFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Removes class properties initialization when the value is `null`.',
```
keradus: following most (but not all) form of description, sth like:
> Properties MUST not be explicitly initialized with `null`.

julienfalque: Description updated.

src/Fixer/ClassNotation/NullPropertyDeclarationFixer.php
julienfalque: Fixer renamed.

tests/Fixer/ClassNotation/NullPropertyDeclarationFixerTest.php
julienfalque: Fixed.

keradus:
Thank you @ntzm.

SpacePossum:
thanks all authors :)  :+1:

PR #2825, opened 2017-06-07
Add PhpdocTypesOrderFixer

julienfalque:
A new fixer that sorts the types in PHPDoc annotations like `@param`. For now, it only moves `null` to the first position:
```diff
 /**
- * @return string|null
+ * @return null|string
  */
```
An option allows to move it to the last position instead:
```diff
 /**
- * @return null|string
+ * @return string|null
  */
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1041,7 +1041,7 @@ Choose from the list of available rules:
     position of ``null`` (overrides ``sort_algorithm``); defaults to
     ``'always_first'``
   - ``sort_algorithm`` (``'alpha'``, ``'none'``): the sorting algorithm to apply;
-    defaults to ``'none'``
+    defaults to ``'alpha'``
 
 * **phpdoc_var_without_name** [@Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -1037,8 +1037,9 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``null_position`` (``'always_first'``, ``'always_last'``, ``'default'``): forces the
-    position of ``null``; defaults to ``'always_first'``
+  - ``null_adjustment`` (``'always_first'``, ``'always_last'``, ``'none'``): forces the
+    position of ``null`` (overrides ``sort_algorithm``); defaults to
+    ``'always_first'``
   - ``sort_algorithm`` (``'alpha'``, ``'none'``): the sorting algorithm to apply;
     defaults to ``'none'``
 
--- a/README.rst
+++ b/README.rst
@@ -1031,6 +1031,17 @@ Choose from the list of available rules:
 
   The correct case must be used for standard PHP types in phpdoc.
 
+* **phpdoc_types_order**
+
+  Sorts PHPDoc types.
+
+  Configuration options:
+
+  - ``null_position`` (``'always_first'``, ``'always_last'``, ``'default'``): forces the
+    position of ``null``; defaults to ``'always_first'``
+  - ``sort_algorithm`` (``'alpha'``, ``'none'``): the sorting algorithm to apply;
+    defaults to ``'none'``
+
 * **phpdoc_var_without_name** [@Symfony]
 
   @var and @type annotations should not contain the variable name.
--- a/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
@@ -94,7 +94,7 @@ final class PhpdocTypesOrderFixer extends AbstractFixer implements Configuration
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('sort_algorithm', 'The sorting algorithm to apply.'))
                 ->setAllowedValues(['alpha', 'none'])
-                ->setDefault('none')
+                ->setDefault('alpha')
                 ->getOption(),
             (new FixerOptionBuilder('null_adjustment', 'Forces the position of `null` (overrides `sort_algorithm`).'))
                 ->setAllowedValues(['always_first', 'always_last', 'none'])
--- a/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
@@ -1,0 +1,206 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Phpdoc;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\Annotation;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Utils;
+
+final class PhpdocTypesOrderFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Sorts PHPDoc types.',
+            [
+                new CodeSample(
+                    '<?php
+/**
+ * @param string|null $bar
+ */'
+                ),
+                new CodeSample(
+                    '<?php
+/**
+ * @param null|string $bar
+ */',
+                    ['null_position' => 'always_last']
+                ),
+                new CodeSample(
+                    '<?php
+/**
+ * @param null|string|int|\Foo $bar
+ */',
+                    ['sort_algorithm' => 'alpha']
+                ),
+                new CodeSample(
+                    '<?php
+/**
+ * @param null|string|int|\Foo $bar
+ */',
+                    [
+                        'sort_algorithm' => 'alpha',
+                        'null_position' => 'always_last',
+                    ]
+                ),
+                new CodeSample(
+                    '<?php
+/**
+ * @param null|string|int|\Foo $bar
+ */',
+                    [
+                        'sort_algorithm' => 'alpha',
+                        'null_position' => 'default',
+                    ]
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('sort_algorithm', 'The sorting algorithm to apply.'))
+                ->setAllowedValues(['alpha', 'none'])
+                ->setDefault('none')
+                ->getOption(),
+            (new FixerOptionBuilder('null_position', 'Forces the position of `null`.'))
+                ->setAllowedValues(['always_first', 'always_last', 'default'])
+                ->setDefault('always_first')
+                ->getOption(),
+        ]);
+    }
+
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $doc = new DocBlock($token->getContent());
+            $annotations = $doc->getAnnotationsOfType(Annotation::getTagsWithTypes());
+
+            if (!count($annotations)) {
+                continue;
+            }
+
+            foreach ($annotations as $annotation) {
+                $types = $annotation->getTypes();
+
+                // fix main types
+                $annotation->setTypes($this->sortTypes($types));
+
+                // fix @method parameters types
+                $line = $doc->getLine($annotation->getStart());
+                $line->setContent(preg_replace_callback('/(@method\s+.+?\s+\w+\()(.*)\)/', function (array $matches) {
+                    $sorted = preg_replace_callback('/((?:^|,)\s*)([^\s]+)/', function (array $matches) {
+                        return $matches[1].$this->sortJoinedTypes($matches[2]);
+                    }, $matches[2]);
+
+                    return $matches[1].$sorted.')';
+                }, $line->getContent()));
+            }
+
+            $tokens[$index] = new Token([T_DOC_COMMENT, $doc->getContent()]);
+        }
+    }
+
+    /**
+     * @param string[] $types
+     *
+     * @return string[]
+     */
+    private function sortTypes(array $types)
+    {
+        foreach ($types as $index => $type) {
+            $types[$index] = preg_replace_callback('/^([^<]+)<(?:(.+?)(,\s*))?(.*)>$/', function (array $matches) {
+                return $matches[1].'<'.$this->sortJoinedTypes($matches[2]).$matches[3].$this->sortJoinedTypes($matches[4]).'>';
+            }, $type);
+        }
+
+        if ('alpha' === $this->configuration['sort_algorithm']) {
+            $types = Utils::stableSort(
+                $types,
+                function ($type) { return $type; },
+                function ($typeA, $typeB) {
+                    $regexp = '/^\\??\\\?/';
+
+                    return strcasecmp(
+                        preg_replace($regexp, '', $typeA),
+                        preg_replace($regexp, '', $typeB)
+                    );
+                }
+            );
+        }
+
+        if ('default' !== $this->configuration['null_position']) {
+            $nulls = [];
+            foreach ($types as $index => $type) {
+                if (preg_match('/^\\\?null$/i', $type)) {
+                    $nulls[$index] = $type;
+                    unset($types[$index]);
+                }
+            }
+
+            if (count($nulls)) {
+                if ('always_last' === $this->configuration['null_position']) {
+                    array_push($types, ...$nulls);
+                } else {
+                    array_unshift($types, ...$nulls);
+                }
+            }
+        }
+
+        return $types;
+    }
+
+    /**
+     * @param string $types
+     *
+     * @return string
+     */
+    private function sortJoinedTypes($types)
+    {
+        $types = array_filter(
+            preg_split('/([^|<]+(?:<.*>)?)/', $types, -1, PREG_SPLIT_DELIM_CAPTURE | PREG_SPLIT_NO_EMPTY),
+            function ($value) {
+                return '|' !== $value;
+            }
+        );
+
+        return implode('|', $this->sortTypes($types));
+    }
+}
--- a/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
@@ -45,7 +45,7 @@ final class PhpdocTypesOrderFixer extends AbstractFixer implements Configuration
 /**
  * @param null|string $bar
  */',
-                    ['null_position' => 'always_last']
+                    ['null_adjustment' => 'always_last']
                 ),
                 new CodeSample(
                     '<?php
@@ -61,7 +61,7 @@ final class PhpdocTypesOrderFixer extends AbstractFixer implements Configuration
  */',
                     [
                         'sort_algorithm' => 'alpha',
-                        'null_position' => 'always_last',
+                        'null_adjustment' => 'always_last',
                     ]
                 ),
                 new CodeSample(
@@ -71,7 +71,7 @@ final class PhpdocTypesOrderFixer extends AbstractFixer implements Configuration
  */',
                     [
                         'sort_algorithm' => 'alpha',
-                        'null_position' => 'default',
+                        'null_adjustment' => 'none',
                     ]
                 ),
             ]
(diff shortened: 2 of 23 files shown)


localheinz:
@julienfalque 

Great idea, like it a lot!

GrahamCampbell:
> For now, it only moves null to the first position:

👎

julienfalque:
@GrahamCampbell Can you be more specific?

GrahamCampbell:
I don't like the idea of moving null first. Should be alphabetic, or null last imo.

julienfalque:
I personnaly prefer having `null` always last, hence the option to enable this behavior. The default value is set to match PHP CS Fixer's own style.

keradus:
@GrahamCampbell, it's so great that we built a tool that is not enforcing everyone to use same ruleset ;)

:+1: for rule and configuration, yet it shall also handle cases like `null|int|bool`
later, it could be enhanced to more general rule like `int|bool|null` -> `null|int|bool` -> `null|bool|int`

src/Console/Output/ProcessOutput.php
```diff
@@ -59,14 +59,14 @@
...
-     * @var int|null
+     * @var null|int
      */
     private $symbolsPerLine;
 
     /**
      * @param OutputInterface $output
      * @param EventDispatcher $dispatcher
-     * @param int|null        $nbFiles
+     * @param null|int        $nbFiles
```
keradus: please extract self-CS changes into 2.2 line

src/DocBlock/Annotation.php
```diff
@@ -205,10 +205,9 @@ private function getTypesContent()
                 throw new \RuntimeException('This tag does not support types.');
             }
 
-            $tagSplit = preg_split('/\s*\@'.$name.'\s*/', $this->lines[0]->getContent(), 2);
-            $spaceSplit = preg_split('/\s/', $tagSplit[1], 2);
+            preg_match('/\@'.$name.'\s+([^\s|<]+(?:<.*?>)?(?:\|[^\s|<]+(?:<.*?>)?)*)/', $this->lines[0]->getContent(), $matches);
```
julienfalque: This is a bugfix, extracted into #2841 which should be merged first.

src/Console/Output/ProcessOutput.php
julienfalque: Done: #2842.

I tried applying the fixer to 2.3 branch as well: only one fix. Is it worth a PR for it?

keradus: nope

README.rst
```diff
@@ -941,6 +941,15 @@ Choose from the list of available rules:
 
   The correct case must be used for standard PHP types in phpdoc.
 
+* **phpdoc_types_order**
+
+  Sorts PHPDoc types.
+
+  Configuration options:
+
+  - ``null_last`` (``bool``): whether ``null`` should be the last type; defaults to
```
keradus: from option name, i wouldn't say that `false` would cause null to be first

README.rst
```diff
@@ -941,6 +941,15 @@ Choose from the list of available rules:
 
   The correct case must be used for standard PHP types in phpdoc.
 
+* **phpdoc_types_order**
```
keradus: is it sorting everything or only `null`? if later, name is too generic

julienfalque: That's on purpose because I want the fixer to be extended in the future to have more advanced sorting options, which is something you seem to agree with:

> later, it could be enhanced to more general rule like int|bool|null -> null|int|bool -> null|bool|int

Extending it may require to deprecate the current option depending on what will be added, but at least we won't have to deprecate the fixer itself in favor of a new one.

README.rst
julienfalque: Indeed.

What about having two options instead, e.g. `first_types` and `last_types` (there are probably better names) that both accept a list of _type categories_? For now, we would only have the `null` category but we could easily extend that list like (e.g. `['objects', 'scalars', 'null']`) without [changing the options in the future](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2825#discussion_r121333649).

keradus: not good idea IMO, as then one would `first_types=null` and `last_types=null` :D
also, I wouldn't go with that big granularity

what about:
- sort_types: bool (sort all types alphabetically)
- null: [opt] `first|last`
?

julienfalque: Nice start but not customizable enough IMO because on my projects I would like to sort by category of type instead of alphabetically. Another suggestion:

- `sort_algorithm`:
  - `alphabetically` (ignoring leading `\` if any)
  - `by_category` (from "heaviest" to "lightest", i.e.: classes, `object`, `callable`, `resource`, `array<...>`, `...[]`, `array`, scalars, `mixed`, `null`)
  - `none`
- `sort_order`:
  - `ascending`
  - `descending`
- `null_position` (takes precedence over other options):
  - `always_first`
  - `always_last`
  - `default`

Given this discussion, maybe it's better to implement whatever we agree on in this PR directly?

keradus: sorting alphabetically in descending order is weird... imo not needed

what's the difference between not setting up `null_position` and setting it with `default` value ?

`by_category` algorithm is a bit overkill imo, we should rather ask user to have one (maybe nullable) type instead of mixed types ;)

-------------

> Given this discussion, maybe it's better to implement whatever we agree on in this PR directly?

imo:
- sort_algorithm: 'alpha' (same, short value as in `ordered_imports` rule for consistency)
- nuill_position: 'always_first, always_last'

is enoguh for this PR. later, we could add other algorithms if needed

julienfalque: > what's the difference between not setting up null_position and setting it with default value ?

Not setting it would be the same as setting it to `always_first`, as it would be the default value of the option.

> imo:
> - sort_algorithm [opt]: 'alpha' (same, short value as in `ordered_imports` rule for consistency)
> - nuill_position [opt]: 'always_first, always_last'
> is enoguh for this PR. later, we could add other algorithms if needed

👍 (but still with `default` as possible value for `null_position`)

keradus: ok, cool here :+1:

keradus:
rebased.

keradus:
is it still WIP @julienfalque  ?

julienfalque:
Yes, I still have to implement https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2825#discussion_r121370276.

README.rst
julienfalque: Done.

tests/Fixer/Phpdoc/PhpdocTypesOrderFixerTest.php
```diff
@@ -0,0 +1,409 @@
...
+            [
+                '<?php /** @method \Bar|Foo foo(\Bar|Foo $foo, \Bar|Foo $bar) */',
+                '<?php /** @method Foo|\Bar foo(Foo|\Bar $foo, Foo|\Bar $bar) */',
+            ],
+            [
+                '<?php /** @var array|Bar\Baz|bool[]|false|Foo|int|object|resource|string|string[]|null */',
+                '<?php /** @var string[]|resource|false|object|null|Foo|Bar\Baz|bool[]|string|array|int */',
+            ],
+            [
+                '<?php /** @var array<int, string>|null Foo */',
```
SpacePossum: missing nested cases;
```
array<array<int, array<int, Foo|null>>, array<int,NULL|float>
```

julienfalque: There is one below this one. Did you mean to sort collection key/value types as well, e.g.:
```diff
-/** @var array<Foo|null> */
+/** @var array<null|Foo> */
```

keradus: it would be nice indeed ;)

tests/Fixer/Phpdoc/PhpdocTypesOrderFixerTest.php
```diff
@@ -0,0 +1,409 @@
...
+                '<?php /** @var ?Foo|Bar */',
+            ],
+            [
+                '<?php /** @var ?\Bar|Foo */',
+                '<?php /** @var Foo|?\Bar */',
+            ],
+            [
+                '<?php /** @var Bar|?\Foo */',
+                '<?php /** @var ?\Foo|Bar */',
+            ],
```
SpacePossum: please add tests for:
```
@var a null|\null
@var b null|null
@var c \null|null
@var d \null|\null
```

non-should be fixed, non should mark the Tokens collection as being marked as changed.

Please add a test for
```php
@var a \null|int
@var b array<\null|int>
@var c array<array<array<array<\null|int>\null|int>\null|int>\null|int>>>>
```

tests/Fixer/Phpdoc/PhpdocTypesOrderFixerTest.php
julienfalque: Done.

tests/Fixer/Phpdoc/PhpdocTypesOrderFixerTest.php
julienfalque: Added all these test cases but:
```php
<?php /** @var array<array<array<array<\null|int>\null|int>\null|int>\null|int>>>> */
```
which looks invalid. Is that on purpose?

SpacePossum:
```php
    /**
     * @var array<int, array<int, array<int, array<int, array<OutputInterface|null>>>>>
     */
```

julienfalque: Done.

src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
```diff
@@ -0,0 +1,213 @@
...
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $doc = new DocBlock($token->getContent());
+            $annotations = $doc->getAnnotationsOfType(Annotation::getTagsWithTypes());
+
+            if (empty($annotations)) {
```
SpacePossum: `!count`

src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
```diff
@@ -0,0 +1,213 @@
...
+    private function sortTypes(array $types)
+    {
+        foreach ($types as $index => $type) {
+            $types[$index] = preg_replace_callback('/^([^<]+)<(?:(.+?)(,\s*))?(.*)>$/', function (array $matches) {
+                return $matches[1].'<'.$this->sortJoinedTypes($matches[2]).$matches[3].$this->sortJoinedTypes($matches[4]).'>';
+            }, $type);
+        }
+
+        if ('alpha' === $this->configuration['sort_algorithm']) {
+            if (PHP_VERSION_ID < 70000) {
```
SpacePossum: why do you need this?

julienfalque: Because of the way [sorting order of equal elements has changed with PHP 7](http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.other.sort-order):
> The internal sorting algorithm has been improved, what may result in different sort order of elements, which compare as equal, than before.

On PHP 7, equal elements will remain in the same order while in PHP 5 the order is unpredictable. I'll add a comment with the link above.

src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
julienfalque: Done.

src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
SpacePossum: ah check, they recommend your workaround for PHP7 as well I think;
```
Note:
Don't rely on the order of elements which compare as equal; it might change anytime.
```

julienfalque: I never saw order changes for equal elements on PHP 7 (tests pass) so I assumed current algorithm does maintain correct order and that the note is about future versions of PHP. Shall I apply to "fix" to PHP 7+ anyway?

SpacePossum: I would, as they state it might change anytime, would be a shame to go back and fix this on some other PHP version later on.

julienfalque: Done.

src/Fixer/Phpdoc/PhpdocTypesOrderFixer.php
```diff
@@ -0,0 +1,210 @@
...
+        foreach ($types as $index => $type) {
+            $types[$index] = preg_replace_callback('/^([^<]+)<(?:(.+?)(,\s*))?(.*)>$/', function (array $matches) {
+                return $matches[1].'<'.$this->sortJoinedTypes($matches[2]).$matches[3].$this->sortJoinedTypes($matches[4]).'>';
+            }, $type);
+        }
+
+        if ('alpha' === $this->configuration['sort_algorithm']) {
+            // make sure order of equal elements stays the same
+            foreach ($types as $index => $type) {
+                $types[$index] .= '#'.$index;
```
keradus: implementing stable sorting via `#` is hacky. extract stable sorting to accept any input. as a side note, you are doing Schwartzian transform here. as we do it as well in `FixerFactory::sortFixers`, it's next things that could be extracted.

src/Utils.php
```diff
@@ -126,4 +126,36 @@ public static function calculateTrailingWhitespaceIndent(Token $token)
...
+     *
+     * @param array    $elements
+     * @param callable $getComparedValue a callable that takes a single element and returns the value to compare
+     * @param callable $compareValues    a callable that compares two values
+     *
+     * @return array
+     */
+    public static function stableSort(array $elements, callable $getComparedValue, callable $compareValues)
+    {
+        array_walk($elements, function ($element, $key) {});
```
keradus: ?

src/Utils.php
```diff
@@ -126,4 +126,35 @@ public static function calculateTrailingWhitespaceIndent(Token $token)
...
+     * Sorts the given values using a Schwartzian transform and a custom
+     * comparison function. Order of equal elements is maintained.
+     *
+     * @param array    $elements
+     * @param callable $getComparedValue a callable that takes a single element and returns the value to compare
+     * @param callable $compareValues    a callable that compares two values
+     *
+     * @return array
+     */
+    public static function stableSort(array $elements, callable $getComparedValue, callable $compareValues)
```
keradus: input array is modified.

src/Utils.php
julienfalque: Damn, you reviewed faster than I expected :D

src/Utils.php
julienfalque: It's not, but I could change the signature to do that.

SpacePossum: ? you do modify the input (ok with me though)

src/Utils.php
```diff
@@ -126,4 +126,35 @@ public static function calculateTrailingWhitespaceIndent(Token $token)
...
+        });
+
+        usort($elements, function ($a, $b) use (&$compareValues) {
+            $comparison = $compareValues($a[2], $b[2]);
+
+            if (0 !== $comparison) {
+                return $comparison;
+            }
+
+            return self::cmpInt($a[1], $b[1]);
```
SpacePossum: for input `$a = ['z' => 1, 'b' => 1, 'c' => 3];` `$a[1]` etc. will not be an int

julienfalque: Yes, the method only supports integer keys for now. It's internal and we only pass it arrays with integer keys only.

README.rst
```diff
@@ -1002,6 +1002,17 @@ Choose from the list of available rules:
...
+
+  Sorts PHPDoc types.
+
+  Configuration options:
+
+  - ``null_position`` (``'always_first'``, ``'always_last'``, ``'default'``): forces the
+    position of ``null``; defaults to ``'always_first'``
+  - ``sort_algorithm`` (``'alpha'``, ``'none'``): the sorting algorithm to apply;
```
GrahamCampbell: How about primitives first/last, in order, then objects, in order, either before/after primitives?

GrahamCampbell:
Priority integration test missing. This fixer needs to run after the fixer that changes the case of `null`.

GrahamCampbell:
Are we missing tests that check that only `null`, with that exact casing, is treated as that type?

README.rst
julienfalque: Yeah I wanted to implement something like that, but now I'd rather get this one merged and implement the new sort algorithm in a separate PR as it would probably lead to long discussions :)

julienfalque:
Handling of `null` is case-insensitive so I don't think there is a priority issue here.

GrahamCampbell:
> Handling of null is case-insensitive so I don't think there is a priority issue here.

That is a bug. PHPDoc is treated case sensitive in the rest of php-cs-fixer.

GrahamCampbell:
Apart from in the case of the import fixers checking for references, where we are more conservative.

src/Utils.php
keradus: no, if it is not modifying, it's great

src/Utils.php
```diff
@@ -126,4 +126,36 @@ public static function calculateTrailingWhitespaceIndent(Token $token)
...
+     *
+     * @return mixed[]
+     */
+    public static function stableSort(array $elements, callable $getComparedValue, callable $compareValues)
+    {
+        array_walk($elements, function (&$element, $index) use ($getComparedValue) {
+            $element = [$element, $index, $getComparedValue($element)];
+        });
+
+        usort($elements, function ($a, $b) use (&$compareValues) {
```
keradus: why reference ?

julienfalque: Removed.

src/Utils.php
SpacePossum: thanks for the doc updates :) :+1:

README.rst
```diff
@@ -1013,6 +1013,17 @@ Choose from the list of available rules:
 
   The correct case must be used for standard PHP types in phpdoc.
 
+* **phpdoc_types_order**
+
+  Sorts PHPDoc types.
+
+  Configuration options:
+
+  - ``null_position`` (``'always_first'``, ``'always_last'``, ``'default'``): forces the
```
keradus: what is `null_position => default` ?

README.rst
```diff
@@ -1013,6 +1013,17 @@ Choose from the list of available rules:
...
+    position of ``null``; defaults to ``'always_first'``
+  - ``sort_algorithm`` (``'alpha'``, ``'none'``): the sorting algorithm to apply;
+    defaults to ``'none'``
```
keradus: hm... what about `defaults to 'alpha'` ?
cc @SpacePossum @juliendufresne

README.rst
julienfalque: `default` keeps the position of `null` as defined by the sort algorithm, unlike `always_first` and `always_last` which override it.

README.rst
julienfalque: I'm not sure most people would use `alpha`, so it wouldn't be a sane default.

I'm not a big fan of alphabetical sorting TBH, I would rather use something like https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2825#discussion_r124588919.

README.rst
SpacePossum: and what if the sort algorithm is configured `none`, would this be a case of misconfiguration?

README.rst
SpacePossum: I'm :+1: for `alpha`
Btw. how does `alpha` handle pending `\` and casing , like `\int|AbstractTest|\AbstractElseNode|BackUp` ?

README.rst
julienfalque: It's not. We could add validation for that but we do have fixers that don't complain when they are configured to do nothing.

README.rst
julienfalque: Leading `\` and `?` (e.g. `?Foo`) are ignored and sorting in case-insensitive.

So, `alpha` as default? Do we apply it to our codebase as well?

README.rst
SpacePossum: Yeah I know, I'm in fav. of moving away from that. Configuring fixers typically means you want them to do something ;) 
Besides it is also a performance impact as the fixer would still be constructed, configured and tested for candidates, something the fixer is even run as well (like here, the candidate check could see the fixer is configured to do nothing and return `false`)
Throwing the misconfigured exception would be better UX/DX IMHO, however for the current ones it could be a BC break depending on your view of feature vs. bug.

julienfalque: To me it's a validation that we should always do so lack of it is a bug. I totally agree it would be a UX improvement.

If we decide to add this validation, I think we should do it in a separate PR to apply it everywhere at once.

keradus: if the fixer is configured to do nothing, then `isCandidate` shall return `false`.
i already started applying this approach.

we could raise a warning on CLI output that some fixer is configured to do nothing, but I am :-1 for raising error.

--------------

@juliendufresne , if `default` means to not move it on top of sorting algorithm, then name is misleading. for me, `default` is meaningful, like `by default we move the null to...`.

README.rst
keradus: yep, I would keep it as that. if one want to have not-standardized order, he could configure the fixer or not use it. idea of fixing is to keep code as similar across whole project of given user as possible. for that, configuring to not update the code is another direction.

README.rst
julienfalque: What would you suggest instead of `default`?

keradus: what about `null_adjustment=always_first|always_last|none` ?

julienfalque: Option updated.

README.rst
julienfalque: Defaults updated and applied.

julienfalque:
I added a few test cases that show how the fixer behaves when sorting types that end with `[]` or contain `\` or `_`.

src/Utils.php
```diff
@@ -126,4 +126,36 @@ public static function calculateTrailingWhitespaceIndent(Token $token)
 
         return ltrim($str, "\n");
     }
+
+    /**
+     * Sorts the given values using a Schwartzian transform and a custom
```
keradus: what about replacing summary with:

```
Perform stable sorting using provided comparison function.

Stability is ensured by using Schwartzian transform.
```

?

julienfalque: Done.

keradus:
Happy to finally have it ;)

keradus:
Thank you @julienfalque.

PR #2828, opened 2017-06-07
Enhancement: Reference phpunit.xsd from phpunit.xml.dist

localheinz:
This PR

* [x] references `phpunit.xsd` from `phpunit.xml.dist`
* [x] drops an unsupported attribute `syntaxCheck`

💁‍♂️ Helps with validation and auto-completion in IDEs that support it!

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,15 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
-<phpunit backupGlobals="false"
-         backupStaticAttributes="false"
-         bootstrap="./vendor/autoload.php"
-         colors="true"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
-         processIsolation="false"
-         stopOnFailure="false"
-         syntaxCheck="false"
+<phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.8/phpunit.xsd"
+    backupGlobals="false"
+    backupStaticAttributes="false"
+    bootstrap="./vendor/autoload.php"
+    colors="true"
+    convertErrorsToExceptions="true"
+    convertNoticesToExceptions="true"
+    convertWarningsToExceptions="true"
+    processIsolation="false"
+    stopOnFailure="false"
+    syntaxCheck="false"
 >
     <testsuites>
         <testsuite>
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -12,7 +12,6 @@
     convertWarningsToExceptions="true"
     processIsolation="false"
     stopOnFailure="false"
-    syntaxCheck="false"
 >
     <testsuites>
         <testsuite>


phpunit.xml.dist
```diff
@@ -1,15 +1,18 @@
...
-         colors="true"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
-         processIsolation="false"
-         stopOnFailure="false"
-         syntaxCheck="false"
+<phpunit
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
```
localheinz: Not sure how you feel about the indentation here, but let me know!

phpunit.xml.dist
```diff
@@ -1,15 +1,18 @@
...
+    backupGlobals="false"
+    backupStaticAttributes="false"
+    bootstrap="./vendor/autoload.php"
+    colors="true"
+    convertErrorsToExceptions="true"
+    convertNoticesToExceptions="true"
+    convertWarningsToExceptions="true"
+    processIsolation="false"
+    stopOnFailure="false"
+    syntaxCheck="false"
```
localheinz: screen shot 2017-06-07 at 16 59 04

```
$ ./vendor/bin/phpunit --version

PHPUnit 5.7.20 by Sebastian Bergmann and contributors.
```

phpunit.xml.dist
julienfalque: Yeah I personally prefer this style 👍 

However I'm not sure that referencing the XSD from vendor directory is a good idea: we use several versions of PHPUnit which might have different XSD files. I would rather use the online version, e.g. https://schema.phpunit.de/5.4/phpunit.xsd. WDYT?

phpunit.xml.dist
julienfalque: I couldn't find any documentation about this option. Shall we drop it?

phpunit.xml.dist
localheinz: Figured if we used the relative location, it would always be current, but to make sure we can point to the online version suggested by the minimum requirement in [`composer.json`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json#L38), what do you think?

phpunit.xml.dist
localheinz: Let's drop it!

phpunit.xml.dist
localheinz: Related, why are we requiring

```json
{
    require: {
        "phpunit/phpunit": "^4.8.35 || ^5.4.3",
    }
}
```

instead of only `^5.4.3` or even higher?

After all, we require at least PHP 5.6, could surely use something more current for `phpunit/phpunit`!

SpacePossum: I would go with `http://schema.phpunit.de/4.8/phpunit.xsd` and we can update that when we update the phpunit required version.

Changes are looking good, thanks :)

julienfalque:
@localheinz Code is looking good, but I think this can be merged into 2.2 instead of master, can you please update the PR?

keradus:
I can do it during merging

keradus:
Thank you @localheinz.

localheinz:

PR #2829, opened 2017-06-07
Fix: Use phpunit/phpunit:^5.4.3 only

localheinz:
This PR

* [x] requires `phpunit/phpunit:^5.4.3` only

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2828#discussion_r120751017.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
         "justinrainbow/json-schema": "^5.0",
         "mi-schi/phpmd-extension": "^4.2",
         "phpmd/phpmd": "^2.4.3",
-        "phpunit/phpunit": "^4.8.35 || ^5.4.3",
+        "phpunit/phpunit": "^5.4.3",
         "satooshi/php-coveralls": "^1.0",
         "symfony/phpunit-bridge": "^3.2.2"
     },


composer.json
```diff
@@ -35,7 +35,7 @@
         "justinrainbow/json-schema": "^5.0",
         "mi-schi/phpmd-extension": "^4.2",
         "phpmd/phpmd": "^2.4.3",
-        "phpunit/phpunit": "^4.8.35 || ^5.4.3",
```
localheinz: Not sure what the idea here is, am I missing something?

julienfalque: This is because we don't want to update the requirement without updating all the code that uses it. A similar PR (#2754) has already been rejected for the same reason.

localheinz: Ha, nice! Thank you!

keradus:
Thanks for understanding @localheinz

localheinz:
I really like https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2754#issuecomment-301604574, @keradus!

keradus:
Slowly but surely @localheinz ;)

keradus:
replaced by #2840

PR #2826, opened 2017-06-07
Exceptions - provide utests

localheinz:
This PR

* [x] covers exceptions missing tests

PR #2830, opened 2017-06-08
Differs - add tests

localheinz:
This PR

* [x] covers differs

Follows #2826.


tests/Differ/SebastianBergmannShortDifferTest.php
```diff
@@ -0,0 +1,79 @@
...
+    public function testComposedDifferIsConfigured()
+    {
+        $differ = new SebastianBergmannShortDiffer();
+
+        $reflection = new \ReflectionObject($differ);
+
+        $property = $reflection->getProperty('differ');
+        $property->setAccessible(true);
+
+        $composedDiffer = $property->getValue($differ);
```
localheinz: :trollface:

src/Differ/SebastianBergmannDiffer.php
```diff
@@ -24,9 +24,9 @@
      */
     private $differ;
 
-    public function __construct()
+    public function __construct(Differ $differ = null)
```
keradus: bc breaker is not acceptable here

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $this->assertSame($diff, $differ->diff($old, $new));
+    }
+
+    /**
+     * @return \PHPUnit_Framework_MockObject_MockObject|Differ
+     */
+    private function createDifferMock()
+    {
+        return $this->getMock('SebastianBergmann\Diff\Differ');
```
keradus: btw, across whole project we are not using old phpunit3-style mocking, but prophecy (which was integrated in 4.5 already and is described as best practice)

tests/Differ/SebastianBergmannShortDifferTest.php
keradus: `PhpCsFixer\Test\AccessibleObject`... ups :D

src/Differ/SebastianBergmannDiffer.php
SpacePossum: is it? the class is final and the functionality doesn't change I think

tests/Differ/SebastianBergmannDifferTest.php
localheinz: Adjusting!

src/Differ/SebastianBergmannDiffer.php
keradus: it's final? cool !

tests/Differ/SebastianBergmannShortDifferTest.php
localheinz: Thank you, @keradus!

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -0,0 +1,64 @@
...
+
+        $differ = new SebastianBergmannDiffer($composedDiffer);
+
+        $this->assertSame($diff, $differ->diff($old, $new));
+    }
+
+    /**
+     * @return \Prophecy\Prophecy\ObjectProphecy
+     */
+    private function createDifferMock()
```
keradus: no need for proxy method

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -0,0 +1,64 @@
...
+final class SebastianBergmannDifferTest extends Framework\TestCase
+{
+    public function testIsDiffer()
+    {
+        $differ = new SebastianBergmannDiffer();
+
+        $this->assertInstanceOf('PhpCsFixer\Differ\DifferInterface', $differ);
+    }
+
+    public function testDiffReturnsDiff()
```
keradus: not sure here.

it looks like here you are not testing the functionallity of differ, but it's implementation details.

why not to use that real Differ instead of mocked one and compare inputs against real, expected diff blob ?

currently, this test check if external SB Differ is called and proxied by our differ adapter class.
if one would replace external implementation with different one that provides different format, this test would still work after.

Instead of testing "is this class properly do the proxying", I would test "Is the public output for public params is matching expected one", regardless how it's generated under the hood.

keradus: actually, exactly the same for next test class, too much of testing the internal implementation details instead of public interface

keradus:
@localheinz , please don't squash your commits, we will do it during merging automatically ;)
without squashing, it's easier to see the difference instead of taking a look at whole diff again ;)

src/Differ/SebastianBergmannShortDiffer.php
```diff
@@ -24,9 +24,9 @@
      */
     private $differ;
 
-    public function __construct()
+    public function __construct(Differ $differ = null)
```
keradus: With injecting you killed the difference between `SebastianBergmannDiffer` and `SebastianBergmannShortDiffer`, as one could inject anything violating the idea behind the class. They are suppose to provide diff in given format, not just to proxy things ;)

keradus:
I faintly guess you agree with my reasoning ?

tests/Differ/SebastianBergmannDifferTest.php
localheinz: 👍

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        $old = 'foo';
+        $new = 'bar';
+
+        $diff = <<<'TXT'
+--- Original
++++ New
+@@ @@
+-foo
++bar
```
keradus: I think we shall have more complicated diffs here as well. this one is really trivial

keradus:
@localheinz , please fix travis

keradus:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/241961977#L330
:(

tests/Differ/NullDifferTest.php
```diff
@@ -0,0 +1,43 @@
...
+use PHPUnit\Framework;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Differ\NullDiffer
+ */
+final class NullDifferTest extends Framework\TestCase
```
julienfalque: Please import FQCN (same for other files).

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -77,7 +77,7 @@ public function __toString()
 @@ @@
 -<?php namespace Foo\Bar;
 +<?php
-
```
localheinz: @GrahamCampbell 

The trailing space is provided by the output from `diff()`.

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -92,7 +92,7 @@ public function __toString()
          $this->foo = $foo;
          $this->bar = $bar;
      }
-
```
localheinz: @GrahamCampbell 

The trailing space is provided by the output from `diff()`.

tests/Differ/SebastianBergmannDifferTest.php
julienfalque: If you need these trailing spaces in the strings, then you can write them like this:
```php
$str = 'Foo
    '.'
Bar';
```
Then the tests will pass but the trailing spaces won't eventually get removed by an IDE that is configured to.

GrahamCampbell: You can write the code using concatenation of strings then, to avoid this, like @keradus suggested for me in my last PR.

julienfalque: Is it the expected behavior of the differ though? /cc @SpacePossum

keradus: it is
(we don't require that, but it how SB differ works)

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -0,0 +1,77 @@
...
+use PHPUnit\Framework;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Differ\SebastianBergmannDiffer
+ */
+final class SebastianBergmannDifferTest extends Framework\TestCase
```
julienfalque: Please import FQCN.

tests/Differ/SebastianBergmannShortDifferTest.php
```diff
@@ -0,0 +1,70 @@
...
+use PHPUnit\Framework;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Differ\SebastianBergmannShortDiffer
+ */
+final class SebastianBergmannShortDifferTest extends Framework\TestCase
```
julienfalque: Please import FQCN.

keradus:
Thank you @localheinz.

localheinz:
Thank you, @GrahamCampbell, @julienfalque, @keradus, and @SpacePossum!

SpacePossum:
Thanks @localheinz : D

keradus:
with your approach @localheinz , you will shortly cover all untested classes ! ;D

PR #2833, opened 2017-06-08
Doctrine\Annotation\Token - provide utests

localheinz:
This PR

* [x] covers token

Follows #2826.

### DIFF
--- a/tests/Doctrine/Annotation/TokenTest.php
+++ b/tests/Doctrine/Annotation/TokenTest.php
@@ -118,6 +118,10 @@ final class TokenTest extends TestCase
                 42,
                 '42',
             ),
+            'different' => array(
+                42,
+                9001,
+            ),
             'array-containing-equal-but-not-same' => array(
                 42,
                 array(
--- a/tests/Doctrine/Annotation/TokenTest.php
+++ b/tests/Doctrine/Annotation/TokenTest.php
@@ -61,8 +61,8 @@ final class TokenTest extends TestCase
     /**
      * @dataProvider providerIsType
      *
-     * @param int       $type
-     * @param int|int[] $types
+     * @param mixed $type
+     * @param mixed $types
      */
     public function testIsTypeReturnsTrue($type, $types)
     {
@@ -96,8 +96,8 @@ final class TokenTest extends TestCase
     /**
      * @dataProvider providerIsNotType
      *
-     * @param int       $type
-     * @param int|int[] $types
+     * @param mixed $type
+     * @param mixed $types
      */
     public function testIsTypeReturnsFalse($types, $type)
     {
--- a/tests/Doctrine/Annotation/TokenTest.php
+++ b/tests/Doctrine/Annotation/TokenTest.php
@@ -1,0 +1,153 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Doctrine\Annotation;
+
+use Doctrine\Common\Annotations\DocLexer;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Annotation/TokenTest.php
```diff
@@ -0,0 +1,153 @@
...
+        $this->assertFalse($token->isType($types));
+    }
+
+    /**
+     * @return array
+     */
+    public function providerIsNotType()
+    {
+        return array(
+            'equal-but-not-same' => array(
```
julienfalque: This cases are violating the test and the tested methods API as they both expect `int|int[]`. I would either add validation in the tested method (e.g. throwing an exception when a string is passed instead of an integer) or remove these test cases.

Anyway, can you please add cases with different integers?

tests/Doctrine/Annotation/TokenTest.php
```diff
@@ -0,0 +1,157 @@
...
+        $this->assertSame('', $token->getContent());
+    }
+
+    public function testConstructorSetsValues()
+    {
+        $type = 42;
+        $content = 'questionable';
+
+        $token = new Token(
+            $type,
```
localheinz: Ha, constructor doesn't reject non-integer values, should I add a test that asserts that the type can be a string?

tests/Doctrine/Annotation/TokenTest.php
localheinz: Not sure what you mean, neither one of

* [`Token::__construct()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/a897664741ae53abd51e95c9797a79aa11af2b64/src/Doctrine/Annotation/Token.php#L40-L44)
* [`Token::setType()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/a897664741ae53abd51e95c9797a79aa11af2b64/src/Doctrine/Annotation/Token.php#L61-L64)
* [`Token::isType()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/a897664741ae53abd51e95c9797a79aa11af2b64/src/Doctrine/Annotation/Token.php#L93-L100)

rejects invalid values.

Can you elaborate, please?

julienfalque: No they don't, that's my point. What I mean is that, as the methods are supposed to only accept  `int|int[]`, how can you be sure that they must return `false` when you pass them a string? I think in such case the methods should throw an exception instead, which would require to update the test.

localheinz: Totally agree, would you like me to add that behaviour in this PR?

julienfalque: Actually no, I edited my previous comment (too late sorry :p).

tests/Doctrine/Annotation/TokenTest.php
```diff
@@ -0,0 +1,157 @@
...
+
+        $token->setType($type);
+
+        $this->assertSame($type, $token->getType());
+    }
+
+    /**
+     * @dataProvider providerIsType
+     *
+     * @param mixed $type
```
julienfalque: Please update with correct types (`int` and `int|int[]`) like for the tested method, and remove invalid test cases instead.

tests/Doctrine/Annotation/TokenTest.php
```diff
@@ -0,0 +1,152 @@
...
+        $this->assertSame($content, $token->getContent());
+    }
+
+    public function testCanClearContent()
+    {
+        $content = 'questionable';
+
+        $token = new Token();
+
+        $token->setContent($content);
```
SpacePossum: question on concept level @keradus and @julienfalque ; are we going to deprecate this for ` PhpCsFixer\Doctrine\Annotation\Token;` as well?

keradus: annotation tokenizer has no `change` tracking, which was root reason to drop it at php tokenizer

keradus:
Thank you @localheinz.

SpacePossum:
thanks @localheinz much appreciated :)

localheinz:

PR #2837, opened 2017-06-08
Fix for whitespace after opening tag issue in the braces fixer

### LINKED ISSUE #2836: InvalidArgumentException: Cannot set empty content for id-based Token.
Master 5d509fbed1c716ee24ea904c20eb18c08bc9f77d.

Error at `src/Tokenizer/Token.php:64`. Called by:

* `PhpCsFixer\Tokenizer\Token->__construct(Array)`,
* `PhpCsFixer\Tokenizer\Tokens->ensureWhitespaceAtIndex(36, 1, '')`,
* `PhpCsFixer\Fixer\Basic\BracesFixer->fixIndents(Object(PhpCsFixer\Tokenizer\Tokens))`

More information to follow very soon.

GrahamCampbell:
Closes #2836.

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -352,7 +352,11 @@ class Tokens extends \SplFixedArray
 
         if ($this[$index]->isWhitespace()) {
             $whitespace = $removeLastCommentLine($this, $index - 1, $indexOffset, $whitespace);
-            $this[$index] = new Token(array(T_WHITESPACE, $whitespace));
+            if ('' === $whitespace) {
+                $this->clearAt($index);
+            } else {
+                $this[$index] = new Token(array(T_WHITESPACE, $whitespace));
+            }
 
             return false;
         }
--- a/tests/Fixer/Basic/BracesFixerTest.php
+++ b/tests/Fixer/Basic/BracesFixerTest.php
@@ -3648,4 +3648,12 @@ if (true)
     {
         $this->doTest($expected, $input);
     }
+
+    public function testStructureSeparatedByTags()
+    {
+        $this->doTest(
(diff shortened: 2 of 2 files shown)


GrahamCampbell:
The test fails on 2.2, 2.3, and master. After my fix, it passes on those branches.

GrahamCampbell:
One thing to note, is that my change to the tokens class is showing as uncovered, even though the test I have is really designed to cover it. D'no if we want to do anything about that?

image

SpacePossum:
If the `$removeLastCommentLine` call returns `''` it means no additional white space token has to be added (it should only if the previous token has pending white space).
Your test is now green because you no longer add the invalid token with content `''`,
however I'm not sure you need to clear the token at the `$index` in the first place.
To test you want to call `ensure white space at` for some index just behind a `<?php` index _without_ a line break.
Sorry can't help out atm, no dev machine ;)

GrahamCampbell:
@SpacePossum Good point. It's probably out of scope of this fixer to remove the whitespace at the end of the opening `<?php` even though there's a closing brace on the next line. Maybe I need to add another test case to another fixer, that's meant to do that?

SpacePossum:
maybe add test cases here;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/tests/Tokenizer/TokensTest.php#L946
with;

```
            array(
                '<?php  ',
                '<?php  ',
                0,
                1,
                '  ',
```

and

```
            array(
                '<?php  ',
                '<?php  ',
                1,
                1,
                '  ',
```

that would test two cases currently missing and might proven the bug or at least exclude the Tokens method as cause.

SpacePossum:
closing in fav. of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2851 which has your fix (and authored as such) + tests covering the fix, thanks!

PR #2839, opened 2017-06-09
Use PHP 7.2 polyfill instead of xml one

keradus:
ref https://github.com/symfony/polyfill/pull/87

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "symfony/polyfill-php54": "^1.0",
         "symfony/polyfill-php55": "^1.3",
         "symfony/polyfill-php70": "^1.0",
-        "symfony/polyfill-xml": "^1.3",
+        "symfony/polyfill-php72": "^1.4",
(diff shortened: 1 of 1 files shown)


SpacePossum:
thanks @keradus  :)

composer.json
```diff
@@ -41,7 +41,6 @@
     },
     "suggest": {
         "ext-mbstring": "For handling non-UTF8 characters in cache signature.",
-        "ext-xml": "For better performance.",
```
GrahamCampbell: Probably worth keeping this, since php 7.2 is not out yet.

keradus: in whole project we go with approach to be future-ready instead of legacy-ready.
if keep sth, i would rather keep `suggest: php7.2` :D but that's pointless, so I removed it

GrahamCampbell: Well, until php 7.2 is out, and this project supports it, surely we should still recommend the xml extension for better performance?

PR #1557, opened 2015-11-29
PhpdocSeparationFixerTest - Add nested separation fixer tests

GrahamCampbell:
Not passing yet. Needs fixing.

PR #2841, opened 2017-06-10
Fix handling of PHPDoc collection types

julienfalque:
Considering this PHPDoc:
```
/**
 * @param array<int, string>|null $foo
 */
```
`Annotation::getTypes()` will return `['array<int,']`. This PR fixes the method to return `['array<int, string>', 'null']` instead.

Not sure if the bug really affects any existing fixer, but it does affect the fixer introduced in #2825.

### DIFF
--- a/src/DocBlock/Annotation.php
+++ b/src/DocBlock/Annotation.php
@@ -205,10 +205,9 @@ class Annotation
                 throw new \RuntimeException('This tag does not support types.');
             }
 
-            $tagSplit = preg_split('/\s*\@'.$name.'\s*/', $this->lines[0]->getContent(), 2);
-            $spaceSplit = preg_split('/\s/', $tagSplit[1], 2);
+            preg_match('/\@'.$name.'\s+(([^\s|<]+(?:<.*?>)?)(?:\|(?2))*)/', $this->lines[0]->getContent(), $matches);
 
-            $this->typesContent = $spaceSplit[0];
+            $this->typesContent = $matches[1];
         }
 
         return $this->typesContent;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Does this handle nested instances?

julienfalque:
Good point, I don't think so. Checking.

GrahamCampbell:
This is a known bug. Unfortunately, I think @keradus closed it because I hadn't fixed it. Can't remember the issue number. :/

GrahamCampbell:
> Good point, I don't think so. Checking.

Yeh, it's not something we can fix with regular expressions properly, but something that could be hacked together, to recognise a finite level of nesting.

SpacePossum:
this one https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1557? still open ;)

julienfalque:
I see, not really the same problem actually. I thought you were talking about something like `array<int, array<int, string>>` but this does not work either so thanks anyway :p

julienfalque:
Last test case is failing. I still have to figure out how to fix it (probably with [recursivity](http://php.net/manual/en/regexp.reference.recursive.php) in the pattern).

GrahamCampbell:
Well, regular expressions are only regular, so as I said, we can only accept a finite number of matching brackets.

keradus:
like recursive patterns ? (`?R`)

GrahamCampbell:
I think those recursive patterns only go to a nesting level of 256?

GrahamCampbell:
Meh, practically, I guess that's fine. :P

keradus:
@julienfalque , is there anything left in this PR after merge of #2858 ?

julienfalque:
Nope, all the tests I added here pass with your changes, so we can close this one, thanks for taking care of it 👍

PR #2842, opened 2017-06-10
Move null to first position in PHPDoc types

julienfalque:
Applies fixes with the `phpdoc_types_order` rule introduced in #2825.


keradus:
Thank you @julienfalque.

keradus:
@julienfalque , please check if there is anything to apply on 2.3 line

julienfalque:
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2825#discussion_r121256050.

Issue #2834, opened 2017-06-08
Process output - Use all columns


julienfalque:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2832#issuecomment-307062182

SpacePossum:
closing as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2843 is merged
(please reopen if anything was left to do here ;) )

PR #2843, opened 2017-06-10
Add estimating-max progress output type

julienfalque:
Implements #2834.

Note that with these changes, `ProcessOutput` is able to handle any number of columns. Maybe it would be better to add a new option to define the number of columns like PHPUnit does, i.e. `--columns=<int|"max">`, then we could use it for `run-in` progress output as well. WDYT?

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -27,6 +27,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\ConsoleOutputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Terminal;
 use Symfony\Component\EventDispatcher\EventDispatcher;
 use Symfony\Component\Stopwatch\Stopwatch;
 
@@ -105,7 +106,7 @@ final class FixCommand extends Command
                     new InputOption('diff', '', InputOption::VALUE_NONE, 'Also produce diff for each file.'),
                     new InputOption('format', '', InputOption::VALUE_REQUIRED, 'To output results in other formats.'),
                     new InputOption('stop-on-violation', '', InputOption::VALUE_NONE, 'Stop execution on first violation.'),
-                    new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in, or estimating).'),
+                    new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in, estimating or estimating-max).'),
                 ]
             )
             ->setDescription('Fixes a directory or a file.')
@@ -178,10 +179,15 @@ final class FixCommand extends Command
         if ('none' === $progressType || null === $stdErr) {
             $progressOutput = new NullOutput();
         } elseif ('run-in' === $progressType) {
-            $progressOutput = new ProcessOutput($stdErr, $this->eventDispatcher, null);
+            $progressOutput = new ProcessOutput($stdErr, $this->eventDispatcher, null, null);
         } else {
             $finder = new \ArrayIterator(iterator_to_array($finder));
-            $progressOutput = new ProcessOutput($stdErr, $this->eventDispatcher, count($finder));
+            $progressOutput = new ProcessOutput(
+                $stdErr,
+                $this->eventDispatcher,
+                'estimating-max' === $progressType ? $this->getMaxColumns() : null,
+                count($finder)
+            );
         }
 
         $runner = new Runner(
@@ -274,5 +280,15 @@ final class FixCommand extends Command
         }
 
         return $exitStatus;
+    }
+
+    private function getMaxColumns()
+    {
+        // Terminal was introduced in symfony/console 3.2
+        if (!class_exists(Terminal::class)) {
+            return $this->getApplication()->getTerminalDimensions()[0];
+        }
+
+        return (new Terminal())->getWidth();
     }
 }
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -185,7 +185,7 @@ final class FixCommand extends Command
             $progressOutput = new ProcessOutput(
                 $stdErr,
                 $this->eventDispatcher,
-                'estimating-max' === $progressType ? $this->getMaxColumns() : null,
+                'estimating-max' === $progressType ? (new Terminal())->getWidth() : null,
                 count($finder)
             );
         }
@@ -280,15 +280,5 @@ final class FixCommand extends Command
         }
 
         return $exitStatus;
-    }
-
-    private function getMaxColumns()
-    {
-        // Terminal was introduced in symfony/console 3.2
-        if (!class_exists(Terminal::class)) {
-            return $this->getApplication()->getTerminalDimensions()[0];
-        }
-
-        return (new Terminal())->getWidth();
     }
 }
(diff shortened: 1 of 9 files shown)


SpacePossum:
>--columns=<int|"max">

:+1: it would be more generic and can be applied for all process output'ers and maybe later for other output as well

src/Console/Command/FixCommand.php
```diff
@@ -105,7 +106,8 @@ protected function configure()
                     new InputOption('diff', '', InputOption::VALUE_NONE, 'Also produce diff for each file.'),
                     new InputOption('format', '', InputOption::VALUE_REQUIRED, 'To output results in other formats.'),
                     new InputOption('stop-on-violation', '', InputOption::VALUE_NONE, 'Stop execution on first violation.'),
-                    new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in, or estimating).'),
+                    new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in or estimating).'),
+                    new InputOption('columns', '', InputOption::VALUE_REQUIRED, 'Number of columns to use for progress output (integer or "max").'),
```
keradus: :-1: for having 2 CLI option for managing progress indicator.
what is the real usage of later option? why would anyone set columns to 25, 87 or 142 ?

src/Console/Command/FixCommand.php
```diff
@@ -275,4 +288,23 @@ private function calculateExitStatus($isDryRun, $hasChangedFiles, $hasInvalidErr
...
+            return null;
+        }
+
+        if ('max' === $columns) {
+            // Terminal was introduced in symfony/console 3.2
+            if (!class_exists(Terminal::class)) {
+                return $this->getApplication()->getTerminalDimensions()[0];
+            }
+
+            return (new Terminal())->getWidth();
```
keradus: are `getTerminalDimensions.0` and `getWidth` always providing same value ?

src/Console/Command/FixCommand.php
julienfalque: Could be useful when executed in a context where the number of columns does not exist (e.g. output is a file instead of a terminal) or cannot be detected, and the default 80 is not suitable.

SpacePossum: > -1 for having 2 CLI option for managing progress indicator.
, what is the real usage of later option? 

so we won't have to double the two current outputs;
* run-in
* estimating
* run-in-max
* estimating-max

etc. Secondly the diff and report formatters could also respect the second param later on.

>why would anyone set columns to 25, 87 or 142 ?

dunno, if it is a flag (i.e. boolean for max or non-max i.e. 80 chars. fixed) I'm fine with that as well.

src/Console/Command/FixCommand.php
julienfalque: `getTerminalDimensions()` has been deprecated and the notice suggests to use `Terminal` instead so it would be surprising that they don't work the same way.

keradus: ah, great, so we cannot use only old interface as it's deprecated and we cannot use only new interface as it's not backported as compatibility layer...

src/Console/Command/FixCommand.php
keradus: let us have
- none
- run-in
- estimating
- estimating-max

and remove `run-in` on 3.x line

SpacePossum: removing `run-in` means no fallback for `estimating` when no estimating can/should be done (i.e. non-rewindable file iterators)

keradus: we don't need that fallback, as we don't rewind

SpacePossum:
what is the roadmap for this PR?

keradus:
I would say https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2843#discussion_r123894800

src/Console/Command/FixCommand.php
SpacePossum: :+1: lets do that

keradus: hm... probably `estimating` could be removed as well on 3.0

keradus:
ping @julienfalque

src/Console/Command/FixCommand.php
```diff
@@ -178,10 +179,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
         if ('none' === $progressType || null === $stdErr) {
             $progressOutput = new NullOutput();
         } elseif ('run-in' === $progressType) {
```
keradus: please add a TODO to remove `run-in` and `estimating` on 3.0

src/Console/Command/FixCommand.php
```diff
@@ -275,4 +281,14 @@ private function calculateExitStatus($isDryRun, $hasChangedFiles, $hasInvalidErr
 
         return $exitStatus;
     }
+
+    private function getMaxColumns()
+    {
+        // Terminal was introduced in symfony/console 3.2
```
keradus: bump the dep

src/Console/Output/ProcessOutput.php
```diff
@@ -91,11 +93,21 @@ public function __destruct()
 
     public function onFixerFileProcessed(FixerFileProcessedEvent $event)
     {
+        if (
+            null === $this->files
+            && null !== $this->symbolsPerLine
+            && 0 === $this->processedFiles % $this->symbolsPerLine
+            && 0 !== $this->processedFiles
+        ) {
+            $this->output->writeln('');
```
keradus: why this is needed ?

src/Console/Command/FixCommand.php
julienfalque: Are you sure we want to do that? `run-in` is the only option that allows to have some progress output without first iterating all files.

src/Console/Output/ProcessOutput.php
julienfalque: This prints a newline when max column width is defined and reached and we don't know the total number of files.

This would be a `run-in-max` progress output option but it doesn't exist so technically this is not used. But now that the `ProcessOutput` class can be configured this way, I preferred to implement it.

src/Console/Command/FixCommand.php
julienfalque: Done.

src/Console/Command/FixCommand.php
keradus: as already described in one of outdated comments, i don't see having that iterating as an issue.
for small collection of files - it's fast.
for extra big collection of files - it won't be that slower, and not having % how far you are when you fix 10_000 files is not nice

src/Console/Output/ProcessOutput.php
keradus: :|
please add comment to params to be obligatory at 3.0 and this condition to be removed then

src/Console/Command/FixCommand.php
julienfalque: TODO added.

src/Console/Output/ProcessOutput.php
julienfalque: The class is `@internal` so if you don't want this, I can simply remove it.

keradus: I want to have as close to v3 version as possible, yet having all v2 progress types available to achieve.
if you say that it's possible to (internally) call the service with parameters that would lead to this case, i would either allow that (current state) with TODO (as requested), or disallow that by throwing exception

julienfalque: TODO added.

keradus:
Thank you @julienfalque.

Issue #2844, opened 2017-06-11
Bug when running braces fixer along side phpdoc_indent


GrahamCampbell:
On current master branch, consider the following input:

```php
<?php

function foo()
{
    $foo->bar(/** oops */$baz);
}
```

When, either the `braces` fixer or the `phpdoc_indent` fixer are run separately, by themselves, no fixes are made, however, when they are run together, we crash:

```
UnexpectedValueException Missing block start. 
    .../Tokenizer/Tokens.php:469 PhpCsFixer\Tokenizer\Tokens::findBlockEnd
    .../Fixer/Basic/BracesFixer.php:345 PhpCsFixer\Fixer\Basic\BracesFixer::fixIndents
    .../Fixer/Basic/BracesFixer.php:147 PhpCsFixer\Fixer\Basic\BracesFixer::applyFix
```

![image](https://user-images.githubusercontent.com/2829600/27010418-d12a151c-4e9b-11e7-849c-617109d11c7c.png)

![image](https://user-images.githubusercontent.com/2829600/27010424-e2e684d4-4e9b-11e7-89c6-65f17e69551d.png)

GrahamCampbell:
Here's a PR that demonstrates this issue: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2845.

PR #2845, opened 2017-06-11
Added failing integration test

GrahamCampbell:
Also failing on the current 2.2 and 2.3 branches.

### DIFF
--- a/tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
+++ b/tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
@@ -1,0 +1,14 @@
+--TEST--
+Integration of fixers: braces,phpdoc_indent.
+--RULESET--
+{
+    "braces": true,
+    "phpdoc_indent": true
+}
+--EXPECT--
+<?php
+
+function foo()
+{
+    $foo->bar(/** oops */$baz);
+}


GrahamCampbell:
If I add this on to the v2.3.2 code, then it **passes**.

GrahamCampbell:
On the 2.3 branch, it is broken after edb500910845946200d0b273b12782167d87285e, but working at ae8f46b7f1e4678d48fba0676130339bf4488c54. Essentially looks like https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2817 broke things.

GrahamCampbell:
Unhelpful that travis is broken for a different reason atm:

image

What we should see is:

```
There was 1 failure:

1) PhpCsFixer\Tests\IntegrationTest::testIntegration with data set #0 (PhpCsFixer\Test\IntegrationCase Object (...))
Errors reported during fixing of file "misc/braces,phpdoc_indent.test": 2: /Users/graham/GitHub/PHP-CS-Fixer/tests/Fixtures/Integration/.tmp.php

Failed asserting that an array is empty.

/Users/graham/GitHub/PHP-CS-Fixer/src/Test/AbstractIntegrationTestCase.php:218
/Users/graham/GitHub/PHP-CS-Fixer/src/Test/AbstractIntegrationTestCase.php:122
```

keradus:
from failure, i would say it happens with simplified code sample and only one used rule. thus utest instead of itest
if happens on 2.2 as well, it shall target 2.2

GrahamCampbell:
> from failure, i would say it happens with simplified code sample and only one used rule. thus utest instead of itest

Nope. It only happens when both fixers are enabled. It does not happen when only one is enabled.

GrahamCampbell:
As I said in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2844.

GrahamCampbell:
> if happens on 2.2 as well, it shall target 2.2

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2847.

PR #2847, opened 2017-06-11
Indent fixes

GrahamCampbell:
Coses #2844.

### DIFF
--- a/tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
+++ b/tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
@@ -1,0 +1,14 @@
+--TEST--
+Integration of fixers: braces,phpdoc_indent.
+--RULESET--
+{
+    "braces": true,
+    "phpdoc_indent": true
+}
+--EXPECT--
+<?php
+
+function foo()
+{
+    $foo->bar(/** oops */$baz);
+}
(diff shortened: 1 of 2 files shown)


tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
```diff
@@ -0,0 +1,14 @@
+--TEST--
+Integration of fixers: braces,phpdoc_indent.
```
keradus: instead of integration test write proper utest. `braces` has nothing to do here

GrahamCampbell: it doesn't fail when we just run one fixer, only when we run them both, and its the braces fixer that crashes!

keradus: it fails, but not crash, when run over one fixer.
any fixer that would run after phpdoc_indent could crash, as phpdoc_intent leaves collection in invalid state. as told in other pr and here, this has nothing to do with that particular rules combination.

keradus:
replaced by #2848

tests/Fixtures/Integration/misc/braces,phpdoc_indent.test
GrahamCampbell: > it fails, but not crash, when run over one fixer.

Not when I run the tool itself. It finishes with OK for both fixers separately.

keradus: yet one of them leave the collection in invalid state. that's why i said fail, not crash

PR #2611, opened 2017-03-17
Fix issue FriendsOfPHP/PHP-CS-Fixer#2176: Indentation issue with comment

owebia:
Fix indentation of double slash comments AND indentation of code following comments (the first line of code following a comment was not well indented as it is not preceded by a ";" or a "}").

Issue #2539, opened 2017-02-14
Wrong indentation for class functions with comments


joseluisq:
When reporting an issue (bug) please provide the following information:

__The PHP version you are using:__
v7.0.15

__PHP CS Fixer version you are using:__
 v2.1.0

__The command you use to run the fixer (and the configuration file you are using if any).__
```sh
php-cs-fixer fix CategoryController.php --rules=@PSR2 
```

__A minimal sample of valid PHP code that is not fixed correctly (if applicable).__

__Before:__
```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;

class CategoryController extends Controller
{
    const MODEL = 'App\Category';

    use RESTActions;

    /**
     * Show the default view.
     *
     * @return \Illuminate\Http\Response
     */
public function index(){return view('category');}
}
```

__After:__
```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;

class CategoryController extends Controller
{
    const MODEL = 'App\Category';

    use RESTActions;

    /**
     * Show the default view.
     *
     * @return \Illuminate\Http\Response
     */
public function index()
{
    return view('category');
}
}
```

__The fixed version of that code after you run the fixer and the version you expected.__

```php
<?php

namespace App\Http\Controllers;

use Illuminate\Http\Request;

class CategoryController extends Controller
{
    const MODEL = 'App\Category';

    use RESTActions;

    /**
     * Show the default view.
     *
     * @return \Illuminate\Http\Response
     */
    public function index()
    {
        return view('category');
    }
}
```

When I remove the comment then indentation issue is fixed.

keradus:
Hi. Thanks for your report.
I classify this as lack of feature (we didn't break the code, we didn't make it uglier either).
Actually, this is a duplicate, I would really want to see it solved.

Would you like to send a PR to make it happen ?

joseluisq:
In fact I did search some related issue before but I did find it..
Sure, I could send a PR for this in near future. I'll need to explore a bit the codebase before. :sweat_smile:

keradus:
probably `braces` fixer is missing proper comment handling

joseluisq:
Ok, I'll fork the codebase..

MarekLew:
This problem is duplicated and closed,closed and duplicated. Indentation after comment not work.

keradus:
Fixed, will be released soon. thanks to @julienfalque !

Issue #1371, opened 2015-09-01
Alignment get messed up when docblock is present


abhimanyu003:
Alignment get messed up when docblock is present

**Code before cs-fixer**

``` php
    /**
     * @param $param
     * @return mixed
     */
        public function length($string)
    {
        return strlen($string);
    }
```

**Code after cs-fixer**

``` php
    /**
     * @param $param
     * @return mixed
     */
        public function length($string)
        {
            return strlen($string);
        }
```

command used `php-cs-fixer fix --level=psr2 File.php`

GrahamCampbell:
I don't think that's a bug? That looks like the correct behaviour. The indentation of the function was matched with the definition.

abhimanyusharma003:
May be I didn't explain it properly.

Please check these two files 

* Before cs-fix https://gist.github.com/anonymous/21b3eee6f6d3e54390fa
* After cs-fix https://gist.github.com/anonymous/c411cd0a2e509b2728a7

`methodFirst` is properly indented in both the files.

Please check what happen to `methodSecond` and `methodThird` after you run cs-fix, they are not aligned with the with the `methodFirst`

GrahamCampbell:
You did explain it right, and I don't think this is a bug.

abhimanyusharma003:
Is there any way align all the methods of class when the doc block is present. They only align whey doc is not there.

GrahamCampbell:
I agree our fixers don't fix all cases, but that's not really a bug.

GrahamCampbell:
We just need more fixers.

svenax:
Of course it is a bug. The indentation fixer should *fix* incorrect indentation, no?

GrahamCampbell:
No, It just matches the function body with that of the definition.

GrahamCampbell:
That is "fixing" it.

svenax:
OK. I still think it is a bug. When there are no comments present, the indentation is normalized a expected:
```php
class test
{
public function one()
{
}

          public function two()
            {
            }
}
```
is formatted as
```php
class test
{
    public function one()
    {
    }

    public function two()
    {
    }
}
```
When there are comments though, the indentation is not normalized, e.g.:
```php
class test
{
// Comment
public function one()
{
}

          // Comment
          public function two()
            {
            }
}
```
is formatted as
```php
class test
{
    // Comment
public function one()
{
}

          // Comment
          public function two()
          {
          }
}
```

[...]

Issue #2608, opened 2017-03-17
PSR-2 indentation not applied fully


mbrodala:
Given the following sample file `test.php`:

```php
<?php

use Foo\BarTrait;

/**
 * Class docblock
 */
class Foo {
  use BarTrait;

  /**
   * Description for method
   *
   * @return void
   */
  public function bar() {

    echo 'Test';
  }
}
```

When I run the command `bin/php-cs-fixer fix test.php` I get this:

```php
<?php

use Foo\BarTrait;

/**
 * Class docblock
 */
class Foo
{
    use BarTrait;

  /**
   * Description for method
   *
   * @return void
   */
  public function bar()
  {
      echo 'Test';
  }
}
```

As you can see the 1st level of indentation within the class has not been applied consistently.

Apparently this is related to the method docblock since I get the following when removing it:

```php
<?php

use Foo\BarTrait;

/**
 * Class docblock
 */
class Foo
{
    use BarTrait;

    public function bar()
    {
        echo 'Test';
    }
}
```

A few more details about my environment:

```
$ php -v
PHP 5.6.30-7+deb.sury.org~trusty+1 (cli)
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.5.0, Copyright (c) 2002-2016, by Derick Rethans
    with blackfire v1.15.0~linux-x64-non_zts56, https://blackfire.io, by Blackfireio Inc.
$ php-cs-fixer -V
PHP CS Fixer version 2.1.2 by Fabien Potencier and Dariusz Ruminski
```

keradus:
hi. its a valid case. do you want to propose a fix ?

mbrodala:
It seems like this is a general issue with indentation near comments AFAIS (#1196, #1371)

keradus:
indeed.

julienfalque:
Closing as #2849 was merged.

PR #2849, opened 2017-06-12
BracesFixer - Fix indentation issues with comments

### LINKED ISSUE #797: Constructor Not Being Indented Correctly
I just ran php-cs-fixer on symfony's 2.6, branch, and found it corrected the indentation on only some functions:

https://github.com/GrahamCampbell/symfony/commit/b25bc38c9898fdf0dec69158938e6ea19a0e26fe#diff-59433ee1eeb1e828c36bb517135f09e5R44

The constructor is incorrectly indented like the other functions, but php-cs-fixer didn't change the indentation.

### LINKED ISSUE #1196: wrong indentation after comment in if statement
All leading whitespaces are tabs.
Fixer is run with psr2 level.
As a result echo '1'; is not indented properly.
If I run fixer with config (listed below) then tab in echo line is not converted and line is not properly indented.

Also fixer doesn't cut to minimum indentation level although I'm not sure if this is a bug or intended behaviour.

.php_cs

``` php
<?php

    ->setUsingCache(true)
    ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
    ->fixers(array(
        'encoding',
        'short_tag',
        'braces',
        'eof_ending',
        'linefeed',
        'parenthesis',
        'php_closing_tag',
        'whitespacy_lines',
        'trailing_spaces',
        'indentation'
    ))
;
```

test.php

``` php
<?php

    if (false) {
    // line after comment is not indented properly
    echo '1';

        $id = 10;
    }
```

### LINKED ISSUE #1344: Indentation fixer -  Wrong indentation
The issue was reported in StyleCI/StyleCI#534.

Since [StyleCI](https://github.com/StyleCI/StyleCI/) uses PHP-CS-Fixer it's better to reference the issue here.

The problem is when the class' code has an indent of 4 spaces and one method has 5 spaces. The `indentation` fixer just fails to fix these lines with 5 spaces.

### LINKED ISSUE #2176: Indentation issue with comment
Hey guys I'm seeing an issue with indentation when an inline comment is present... this has been formated by php-cs-fixer already

```
class fp_charts extends WP_Widget
{
    public function __construct()
    {
        $widget_details = array(
            'classname' => 'fp_charts',
            'description' => 'Create charts using supplied data'
        );
        parent::__construct('fp_charts', 'FlowPress Charts', $widget_details);
    }

    //Backend form
  public function form($instance)
  {
```

julienfalque:
Fixes #797, #1196, #1371 and #2539.
Fixes #2176 at least partially as some comments don't provide input code so I couldn't reproduce them.
May fix #1344, to confirm.
Replaces #2611.


SpacePossum:
can you add the utests from 2611 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2611/files) ?

julienfalque:
Tests added. Note that I added them to a different provider method as I don't think they were related to `testPreserveLineAfterControlBrace`. All passed without further changes.

I also added test cases (and a fix) for #2608 which is now fixed by this PR as well. Without this last change, this one was only fixed partially: everything was correcly indented, except multiline content of PHPDoc comments.

GrahamCampbell:
Looks good as far as I can see. 👍

julienfalque:
@GrahamCampbell you reported #797, would it be possible for you to somehow check whether this PR fixes your issue?

GrahamCampbell:
My mac is currently broken, so I am unable to run code easily for now. Taking it in for repair today, so will check on this soon. :)

GrahamCampbell:
Got myself a working Ubuntu environment in the meanwhile. I can confirm that this PR behaves correctly on the issue that I reported. :+1:

julienfalque:
Thanks for checking @GrahamCampbell!

SpacePossum:
```php
    public function provideFixControlContinuationBracesCases()
    {
        return array(
            array(
                '<?php
class A
{
    public function B()
    {
        /*
                */
      $a = 1;
    }
}
',
                '<?php
class A {
    public function B()
    {/*
        */
      $a = 1;
    }
}
',
```

Adding the case shows a problem; after fixing and refixing the code changes.

SpacePossum:
```php
    public function provideFixControlContinuationBracesCases()
    {
        return array(
            array(
                '<?php
class B
{
    public function B()
    {
        /*
            *//**/
        $a = 1;
    }
}
',
                '<?php
class B {
    public function B()
    {
    /*
        *//**/
       $a = 1;
    }
}
',
            ),
            array(
                '<?php
class C
{
    public function C()
    {
        /* */#
        $a = 1;
    }
}
',
                '<?php
class C {
    public function C()
    {
    /* */#
       $a = 1;
    }
}
',
            ),
```

two more cases, not sure if you want to pick up these as well

src/Fixer/Basic/BracesFixer.php
```diff
@@ -429,7 +439,11 @@ function ($item) {
                     || !($nextToken->isWhitespace() && $nextToken->isWhitespace(" \t"))
```
SpacePossum: while you're at it ;) can you please wrap the `($nextToken [...] && [...] "\n") // preserve blank lines` 
within `(` `)`

and update `// set indent only if it is not a case, when comment is following { in same line` to `on the same line`

SpacePossum:
if you've time, please take a look at the code coverage by the tests as well

src/Fixer/Basic/BracesFixer.php
julienfalque: Updated the comment.

I added braces while fixing the first test case, is it what you wanted to do?

julienfalque:
Test cases added and fixed.

src/Fixer/Basic/BracesFixer.php
SpacePossum: yeah, thanks man :)

mooror:
I'm currently using the latest PHP-CS-Fixer via its atom package and experiencing comment indenting problems myself. I am excited to see a possible fix well under way. 

Keep up the good work,
Mooror

src/Fixer/Basic/BracesFixer.php
SpacePossum: last changes undid the request;
I like when a condition in the form;
```php

if ($a || $b && $c) {
    echo 1;
}
```

is written out explicit as;
```php

if ($a || ($b && $c)) {
    echo 1;
}
```

maybe one day I make a fixer for it ;)

src/Fixer/Basic/BracesFixer.php
```diff
@@ -354,13 +354,35 @@ function ($item) {
...
+                            }
+                        }
+
+                        if ($i === $startBraceIndex) {
+                            $nextLineCanBeIndented = true;
+                        }
+                    }
+                }
+
+                if ($nextLineCanBeIndented) {
```
SpacePossum: this check can be moved at the end inside the block `if (1 === $nextlevel)` (i.e. remove line `}` at ln 374 and close later)

SpacePossum:
please add;
```php
    public function provideFixControlContinuationBracesCases()
    {
        return array(
            array(
'<?php
if ($a){/*
*/
echo 1;
}
',
```

note, this one is fixed correctly

```php
<?php
if ($a) { /*
*/
echo 1;
}
```

i.e. issue is `{` followed by multi line comment directly

src/Fixer/Basic/BracesFixer.php
julienfalque: Done.

julienfalque:
Rebased onto 2.2 and added the test case.

SpacePossum:
```php
    public function provideFixControlContinuationBracesCases()
    {
        return array(
            array(
'<?php
if ($a) { /**/ /*
*/
    echo 1;
    echo 2;
}
',
                '<?php
if ($a){ /**/ /*
*/
echo 1;
echo 2;
}
',
```

sorry, gave you the wrong case :(

julienfalque:
Tests updated and fixed.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -354,13 +354,47 @@ function ($item) {
...
+                            if (!$tokens[$i]->isWhitespace() && !$tokens[$i]->isComment()) {
+                                break;
+                            }
+                        }
+
+                        if ($i === $startBraceIndex) {
+                            $nextLineCanBeIndented = true;
+                        }
+
+                        if ($nextLineCanBeIndented) {
```
SpacePossum: can be merged with 375; `if ($i === $startBraceIndex || nextLineCanBeIndented)`

src/Fixer/Basic/BracesFixer.php
SpacePossum:
```php
                if (
                    !$nextNonWhitespaceToken->isComment()
                    || (
                        !($nextToken->isWhitespace() && $nextToken->isWhitespace(" \t"))
                        && ($nextToken->isWhitespace() && 1 === substr_count($nextToken->getContent(), "\n")) // preserve blank lines
                    )
                ) {
```

this test still confuses me haha

src/Fixer/Basic/BracesFixer.php
julienfalque: Done.

SpacePossum: :+1:

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2849#discussion_r125471150
and please throw in some comments to make it easier to follow the code ;)

src/Fixer/Basic/BracesFixer.php
julienfalque: Actually the condition could be simplified, it should be easier to read now :)

julienfalque:
Fixed the first example you provided. I'm not sure about the second one though: how can you tell the indentation of the second comment line is intended and should not be changed?

SpacePossum:
the second comment line is not of concern of the fixer is it? i.e. it is not related to any braces?

julienfalque:
It is: the fixer makes sure the braces are properly placed and that their content is properly indented. Here is the fixer's description:
> The body of each structure MUST be enclosed by braces. Braces should be properly placed. Body of braces should be properly indented.

In the example above, the second comment was not moved because of the various _"previous token is not a comment"_ checks that prevented anything that follows a comment to be properly indented, which is what this PR tries to fix.

keradus:
the second case I provided shall not be modified (and it's not modified on `2.2@stable`)

julienfalque:
> it's not modified on `2.2@stable`

Like I said, anything that follows the first comment is not modified on `2.2@stable`. If you try this:
```php
function foo()
{
    $bar = 1;                   // multiline ...
                                $foo = 3;
    $baz  = 2;                  // next comment
}
```
you'll notice that the `$foo` line isn't modified. With the changes in this PR, it will be, but I see no reason to _not_ fix this line when it is a comment.

SpacePossum:
what about skipping any line that is line break + white space* + comment?

julienfalque:
Then following case would not be changed:
```php
function foo()
{
        // foo
    return $foo;
}
```
which looks wrong to me.

keradus:
the thing is that a lot of users align "single-line" comments to make them be kind-of multiline. breaking it is no-go

keradus:
Extra work made here, glory to you @julienfalque !

julienfalque:
Thank you guys for the help you provided ;)

SpacePossum:
awesome to see this fixed, one of the oldest long standing hard to fix bugs is finally gone :D
thanks @julienfalque !

mooror:
Job well done!

MarekLew:
don`t work corectly for one line comment


```
abstract class XAbcObject
{
    protected $dbc;

    protected $fields; // aa

  protected $emptyFieldsMessages; // bbb
  protected $pk; // cc
  protected $table; // dd
  protected $values = array(); // ee
  protected $errorMsg;  
    protected $errorField;
....
```

julienfalque:
@MarekLew Thanks for your report, can you please open a dedicated issue?

Issue #2813, opened 2017-05-30
NoLeadingImportSlashFixer is not working when namespace is missing


kubawerlos:
Due to [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Import/NoLeadingImportSlashFixer.php#L60) condition when namespace is missing NoLeadingImportSlashFixer won't work on file. According to [manual](http://php.net/manual/en/language.namespaces.importing.php):

> the leading backslash is unnecessary and not recommended, as import names must be fully qualified, and are not processed relative to the current namespace.

Can we update the fixer to always remove leading slash?

SpacePossum:
Hi!

Do you mean changing something like
```
<?php

use \DateTime;
```

to 
```
<?php

use DateTime;
```

? I would not recommend removing the base namespace slash here

julienfalque:
Why not? This looks like a bug to me, although there is an [explicit test case for this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/Fixer/Import/NoLeadingImportSlashFixerTest.php#L52).

SpacePossum:
>Why not? This looks like a bug to me, although there is an explicit test case for this.

because if you have `DateTime` in the same namespace as the code with the `use` import than changing from `use \DateTime` to `use DateTime` means you import a diff. class altogether.

julienfalque:
This is not true because, as quoted from [manual](http://php.net/manual/en/language.namespaces.importing.php) by @kubawerlos:
> import names must be fully qualified, and are not processed relative to the current namespace.

so `use DateTime` is equivalent to `use \DateTime`, whatever the namespace this is declared in.

kubawerlos:
@SpacePossum - it's exactly as @julienfalque said - there is no difference at all.

This is being problematic with OrderedImportsFixer and sorting alphabetically as all imports with leading slashes are moved before the ones without it. I'll make PR, but as enhancement or bugfix?

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/706#issuecomment-62686893
it seems this was done for a reason (so not a bug), didn't test it yet on all PHP versions we support

julienfalque:
The warning in linked comment seems to occur both [with](https://3v4l.org/b5QC6) and [without](https://3v4l.org/LeOiF) the leading slash.

kubawerlos:
[...]

PR #2854, opened 2017-06-15
NoLeadingImportSlashFixer - Removing leading slash from import even when in global space

kubawerlos:
For #2813


keradus:
master is reserved for features, please target your PR to 2.2

kubawerlos:
done, I was mislead by the label of issue

keradus:
Thank you @kubawerlos.

PR #2855, opened 2017-06-15
Changing README extension from "rst" to "md"

kubawerlos:
It's due to visual reason for code in italic text:

In current "rst" format at https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/README.rst when you search for "Risky rule: risky when the function" you can see double backticks.

At https://github.com/kubawerlos/PHP-CS-Fixer/blob/readme-from-rst-to-md/README.md when searching for the very same "Risky rule: risky when the function" the code is formatted properly.


keradus:
Hi. Thank you for your help, yet I must say that changing file format for readme is not acceptable due to way we autogenerate website from that file.
BTW, if changing format, it would be nice to use specific file syntax instead of only changing ext. I know that GH has special handling of md files that could accept most of rst syntax, yet it's official md syntax is not supporting it.
If you want to fix handling that italic text, please adjust ReadmeCommand

kubawerlos:
Do you mean http://cs.sensiolabs.org? There is also the issue with double backticks. Would it be difficult to change "rst" to "md" there?

By special handling do you mean making double backticks to code, not only single ones? I've verified that Parsedown is accepting doubles as well and it can be change in separate PR.

keradus:
yes, that one. as said, changing format is not an option for now.

I mean that we don't write readme on our own, but generate it via ReadmeCommand. So we do generate those ticks. We could change a way we do generate it if we want

keradus:
website releases are not connected to major version releasing of app

kubawerlos:
then anytime - this pull request will be here ;)

julienfalque:
Could it be a bug in the way reStructuredText is parsed? Isn't there a workaround that does not require moving to Markdown?

kubawerlos:
Not according to [Common markup for Markdown and reStructuredText](https://gist.github.com/dupuy/1855764):

> Markdown allows monospaced text within bold or italic sections, but not vice versa - reStructuredText allows neither.

Issue #2451, opened 2017-01-06
Fixing annotation block


TheLuckyOne:
Hi all

$ php -v
PHP 7.1.0 (cli) (built: Dec  2 2016 11:32:42) ( NTS )

$ /usr/local/bin/php-cs-fixer -V
PHP CS Fixer 2.0.0 by Fabien Potencier and Dariusz Ruminski

The command you use to run the fixer (and the configuration file you are using if any).
/usr/local/bin/php-cs-fixer fix test.php
```
You are running php-cs-fixer with xdebug enabled. This has a major impact on runtime performance.
Loaded config default.

Fixed all files in 0.026 seconds, 6.000 MB memory used
```

A minimal sample of valid PHP code that is not fixed correctly (if applicable).
```
<?php
/**
   * AdvBanners.
    */
class AdvBanners
{
    /**
         * @var int
         */
    private $id;
}
```
The fixed version of that code after you run the fixer and the version you expected.
After run:
```
<?php
/**
   * AdvBanners.
    */
class AdvBanners
{
    /**
         * @var int
         */
    private $id;
}
```

Expected:
```
<?php
/**
 * AdvBanners.
 */
class AdvBanners
{
    /**
     * @var int
     */
    private $id;
}
```

Is it a bug or?

keradus:
it is not a bug. bug is when the code was modified in wrong way.

here, you miss a feature.

keradus:
do you want to work on it and send a PR introducing new feature ?

TheLuckyOne:
I'll try :-)

SpacePossum:
this is fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2857 and will be part of the next minor release

PR #2857, opened 2017-06-16
Add AlignMultilineCommentFixer

Slamdunk:
Hi, this fixer is intended to align multiline comments and docblock:

```diff
     /* or /**
      * Multiline comment
-*
+     *
  Lines not prefixed with asterisk are left untouched
-       *
-   */
+     *
+     */
```

Single line comments, either `//` or `#`, are unaffected.

### Tabs

Tabs are now counted as 1 space while counting the doc head position, and removed inside the comment in favour of spaces.

Should we handle them differently?
Any suggestion?


julienfalque:
Love the idea!

> Tabs are now counted as 1 space while counting the doc head position, and removed inside the comment in favour of spaces.

I wonder if it would be better to implement `WhitespacesAwareFixerInterface` to leverage `WhitespacesFixerConfig`. If the user wants to indent with tabs, you may use tabs to indent inner lines of the comments, e.g.:

```
[tab]/*
[tab] * Foo
[tab] */
```

WDYT?

Also, in case the comment starts with `/**`, I think the fixer should add missing `*` on each line as required by [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#3-definitions):

> In case a DocComment spans multiple lines then every line MUST start with an asterisk (*) that SHOULD be aligned with the first asterisk of the opening clause.

_Note: this PR implements #2451._

Slamdunk:
### Tabs
Tabs expose some issues in the way this fixer is written at this time.
Now it:

1. Saves the last new-line position
1. Counts the number of chars till the `/*` opening "*tag*"
1. Prepend to each line as many spaces as the chars count.

It is an alignment **relative** to the `/*` opening "*tag*" position.
This algorithm is easy and straightforward to implement because it ignores the context of possible PHP language structure (ex. the method the docblock is written for), an no indentation count is done, i.e. you can put the comment on any position and not in a multiple of, for example, four.

Therefore, how should we implement Tab handling if the "*position*" is not a multiple of a default value?
May the user have to configure how many spaces a tab weights?
And if so, should we by the way fill with spaces if the numbers give us a rest?
```php
$a = 123;  /*
*/

// 11 chars before comment, how to indent?
$a = 123;  /*
>TAB>TAB...*/

// Or
$a = 123;  /*
>TAB>TAB>TAB.*/

// And if the user has set 8 spaces for a tab in the IDE?
$a = 123;  /*
>TAB...*/
```
### PSR-5
Regarding PSR-5, this can be an (excellent) option, I'm going to implement it (next week)

README.rst
```diff
@@ -207,6 +207,10 @@ automatically fix anything:
 
 Choose from the list of available rules:
 
+* **align_multiline_comment**
```
keradus: add it to locak `.php_cs.dist` file

Slamdunk: Done

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -780,7 +780,7 @@ public function provide70Cases()
     |--------------------------------------------------------------------------
     | Test sorting by length
     |--------------------------------------------------------------------------
-    */
+     */
```
keradus: this looks wrong. it shall either fix intend for everything or for nothing here.
please add this to new fixer test cases

Slamdunk: Often the multiline comment is used to comment code out temporarly.
As such, we can't indent arbitrary lines; in theory even lines starting with `*` could be something other than a comment, but `*` is an accepted standard, enforced in code and conventions by DocBlocks.

Look at this example;
```php
/*
$string = ' Hello World,
    this is a commented out code,
  | with a line starting with a pipe.
';
 */
$string = 'Hello World!';
```
We could let the user specify his/her favourite "*comment starting char*", but honestly I find it overkill.

I suggest to convert that specific comment to asterisks, and keep the current implementation.

julienfalque: > `*` is an accepted standard, enforced in code and conventions by DocBlocks

I would say that it only applies to DocBlocks, which start with `/**`. Maybe the fixer should be less strict with regular multiline `/*` comments? or maybe not touch them at all?

julienfalque:
> ```php
> $a = 123; /*
> */
>
> // ...
> ```

Aligning multiline comments that start on the same line as other statements looks weird to me. I think people would either write the multiline comment before the statement or use a single-line comment (i.e. `//`). Do you have examples of actual usage of this?

Slamdunk:
No example for this, I discovered the case in a legacy code and wanted to be sure it was in the tests.

We can skip the case, but the Tab issue remains: what if the user configured tab usage, but the fixer counts indivisible chars number, i.e. a mix of tabs and spaces?

tests/Fixer/Import/OrderedImportsFixerTest.php
Slamdunk: Multiline comments in docblock style is widely used, maybe we can split it in a second fixer, but I would keep it in some way

keradus: then, again, add it as testcase and prove it's not modified.

src/Fixer/Comment/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $intendationChars = 0;
+        foreach ($tokens as $index => $token) {
+            $newlinePosition = mb_strrpos($token->getContent(), "\n");
```
keradus: is there any real need for using mb string function to detect `\n` ?

src/Fixer/Comment/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,76 @@
...
+
+            if ($token->isComment() && '/*' === substr($token->getContent(), 0, 2) && false !== $newlinePosition) {
+                $content = $token->getContent();
+                $content = preg_replace('/^([ \t]+)?\*/m', str_repeat(' ', $intendationChars).'*', $content);
```
keradus: this is wrong. you need to insert whitespace type requested by user, thus tabs or spaces

tests/Fixer/Import/OrderedImportsFixerTest.php
Slamdunk: Sorry I don't understand what should I test.
Should or shouldn't multiline comments be fixed?
If yes, the closing tag MUST be fixed according to opening tag, and `*` must be aligned.

Maybe you didn't see the complete diff:
```diff
     /*
     |--------------------------------------------------------------------------
     | Test sorting by length
     |--------------------------------------------------------------------------
-    */
+     */
```
Here that pipes are wrong aligned, not the `*`

src/Fixer/Comment/AlignMultilineCommentFixer.php
Slamdunk: No real need, I'm just used to it; I'll change this.

src/Fixer/Comment/AlignMultilineCommentFixer.php
Slamdunk: First someone needs to reply to the `Tab` issue I explained in

1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2857#issuecomment-309051132
1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2857#issuecomment-309279975

julienfalque: The tab issue is the reason I suggested to not fix multiline comments that start on the same line as another statement. I don't see a solution for those but using spaces only.

If the fixer only cares about comments that start on their own line, it will be able to:
- fix indentation of comments start;
- align subsequent lines to this start using configured indentation style.

But, like you said, this kind of comment is often used to disable parts of the code. I don't think all users will expect those to be changed by the fixer, and you have no way to detect whether the comment is expected to be aligned.

So IMO, by default the fixer should only care about multiline comments that:
- are `/**` DocBlocks;
- start on their own line.

Maybe fixing regular `/*` comments and/or comments that start on the same line as another statement could be enabled using options that default to `false`.

tests/Fixer/Import/OrderedImportsFixerTest.php
keradus: I would align the stars only if whole comment block is based on stars, like
```
/*
 *
   *
*
         */
```,
not mixed up with different things, like `|`.

WDYF?

Comment (Slamdunk) [inline tests/Fixer/Import/OrderedImportsFixerTest.php]: OT: You mean `WDYT`, don't you? Because http://www.urbandictionary.com/define.php?term=wdyf 😀 

I'm rewriting everything with @julienfalque suggestions: multiline comments will be fixed like now, but with an option `false` by default.
The fixer by default will only fix PHPDocs, and so here a new question: by default it will be a PHPDoc related fixer, but an option will fix non-PHPDoc aka multiline comments, should be:

1. Keep it under `Comment` namespace
1. Move it to `Phpdoc` namespace
1. Split it

?

Comment (keradus) [inline tests/Fixer/Import/OrderedImportsFixerTest.php]: oh, so by default it gona fix only docblock? then cool, let fix case you mention with option=true as well.

as main target would be docblock, i would move it to phpdoc group

we don't split fixers if they are the same except config.

**Comment by Slamdunk** [comment]
So now we have:

1. Default config fixes PHPDoc only, enforcing `*` on every line
1. With `psr5_multiline_start` to `false` lines not starting with `*` are unchanged
1. With `fix_multiline_comment` to `true` multiline comments are also fixed, but lines not starting with `*` are **never** changed
1. PHPDoc / Multiline comments not starting on a new line are never changed

`WhitespacesAwareFixerInterface` is used only for line ending; there is no need to check/use indentation set by the user, because it is copied from the `/**` position.

To me there is only one thing left: what could be an appropriate name for this fixer and its options?

**Comment by julienfalque** [comment]
> With `psr5_multiline_start` to `false` lines not starting with `*` are unchanged

I would drop this option. Let's promote (_de-facto_) standards and always enable this behavior for DocComments.

> With `fix_multiline_comment` to `true` multiline comments are also fixed

What about `fix_all_multiline_comments`? Current name suggests it only fixes single-line comments by default.

> but lines not starting with * are never changed

I like [@keradus' suggestion](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2857#discussion_r122657200) about fixing those comments only when all their lines start with a `*`. Maybe this would allow you to drop this option as well (sorry 😀).

**Comment by Slamdunk** [comment]
Ok, what about:

1. Drop `psr5_multiline_start`: DocComment MUST follow PSR-5
1. Rename `fix_multiline_comment` to `multiline_comment_blocks` which defaults to `false` (`fix` verb should be implicit in options context)
1. Add `multiline_comment_blocks_mixed_content` which defaults to `false` to fix also multiline comments with mixed content

?

**Comment by Slamdunk** [comment]
`multiline_comment_blocks` would be mandatory in order to get `multiline_comment_blocks_mixed_content` work; code samples will be provided of course.

**Comment by julienfalque** [comment]
I still think `multiline_comment_blocks_mixed_content` option would be overkill but if you really want to implement it I'm fine so 👍 for your suggestions.

**Comment by Slamdunk** [comment]
Done

File: src/Fixer/Phpdoc/AlignMultilineCommentFixer.php [hunk 5]
```diff
@@ -0,0 +1,165 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_COMMENT, T_DOC_COMMENT]);
```
julienfalque: Is it possible to use `$this->tokenKinds`?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+        $lineEndingLength = mb_strlen($lineEnding);
```
julienfalque: Is `mb_*` functions usage needed?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,165 @@
...
+            if (false !== $indentation && $newlinePosition && $token->isGivenKind($this->tokenKinds)) {
+                $lines = explode($lineEnding, $content);
+                $skipCurrentCommentToken = false;
+                foreach ($lines as $lineNumber => $line) {
+                    if ($lineNumber === 0) {
+                        continue;
+                    }
+
+                    $line = ltrim($line);
+                    if ($token->getId() === T_COMMENT && (!isset($line[0]) || $line[0] !== '*')) {
```
julienfalque: Please use `$token->isGivenKind(T_COMMENT)` + Yoda style.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,165 @@
...
+                $indentation = mb_substr($content, $newlinePosition + $lineEndingLength);
+                continue;
+            }
+
+            if (false !== $indentation && isset($content[0]) && trim($content) === '') {
+                $indentation = preg_replace('/[ \t]*[\r\n]+/m', '', $content);
+                continue;
+            }
+
+            if (false !== $indentation && $newlinePosition && $token->isGivenKind($this->tokenKinds)) {
```
julienfalque: If I understand correctly, all the code above in the `foreach` tries to find the indentation of the comment, testing each token one by one. Wouldn't it be easier to iterate until you find a comment token and then check the previous one only?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,165 @@
...
+                $lines = explode($lineEnding, $content);
+                $skipCurrentCommentToken = false;
+                foreach ($lines as $lineNumber => $line) {
+                    if ($lineNumber === 0) {
```
julienfalque: Please use Yoda style.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,165 @@
...
+                        if (false === $this->configuration['multiline_comment_blocks_mixed_content']) {
+                            $skipCurrentCommentToken = true;
+                            break;
+                        }
+                        continue;
+                    }
+
+                    if (!isset($line[0])) {
+                        $line = '*'.$line;
+                    } elseif ($line[0] !== '*') {
```
julienfalque: Yoda style.

README.rst
```diff
@@ -207,6 +207,19 @@ automatically fix anything:
...
+
+  Multiline doc comment: enforce asterisk start on each line [PSR-5] and
+  align them.
+
+  Configuration options:
+
+  - ``multiline_comment_blocks`` (``bool``): also align doc-like multiline comments
+    if all lines start with an asteristk ``*``; defaults to ``false``
+  - ``multiline_comment_blocks_mixed_content`` (``bool``): also align multiline
```
Slamdunk: Of course 👍

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
Slamdunk: I thought it at the time of writing and I would say no: `isCandidate` is a contract of the public `FixerInterface` with no declared bond with configuration setup.
Maybe the software implementation at the act of `php-cs-fixer fix` execution could be functioning, but we can't make this assumption in the present nor in the future.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
Slamdunk: It depends on what you consider *easier*: handling the case of PHPDoc started on a code line requires a backward loop, inside the main loop, with the same checks. The code would be a bit faster and a bit less readable/maintanable.
The current implementation does no less than 1 and no more than 2 operations per line, `strpos` (super fast) and `trim` (medium fast), with less code.

Without an evidence of a performance issue, I vote for less code. WDYT?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
Slamdunk: If we know (as we do) that new lines can be only `\r` and `\n`, I think no.
In my business I handle everyday non-ASCII chars that become part of PHPDoc, and already run into `str*` issues for that, so by default I use `mb_*` like the two function used below this line, and for consistency I use them everywhere.

Said that, I'm not going to search a test case to prove my point and so I will not argue a change request in this way.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
keradus: actually, we do this. isCandidate is bounded to given configuration.
we use configuration inside `isCandidate` in multiple fixers. please do it here as well.

we call `isCandidate` directly in `fix` method:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/AbstractFixer.php#L64

for custom rules, we call it in runner:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Runner/Runner.php#L179

with custom runner, it's responsibility of dev to call it properly.

-------

Original issue was that one was able to call `fix` without calling `isCandidate/supports`. now, fix call them out of the box for our fixers and still require a manual run **before** fixing/**after** configuring for custom ones.
I do agree that if called before configuring only, it may provide wrong output. 
`FixerInterface::isCandidate` and `FixerInterface::supports` are implementation details that shall not be part of interface (yet, part of `AbstractFixer`)

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
SpacePossum: finding a case for this should be easy, we run into a similar issue in this before, +1 staying with mb* here

Slamdunk: https://3v4l.org/sm3h1

Dunno how to interpret there results.
Do you know any programmer using `UCS-2` for source code?

Slamdunk:
Changes done.
Default is `phpdocs_only`, but we can also decide for `phpdocs_like`.

Let me know

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
julienfalque: Why would you need a backward loop? Checking if the previous token is a whitespace with a newline should be enough.

On a semantical point of view, your code says _"check all tokens to detect indentation, and if a comment happens to be found, apply this indentation to it"_. I would refactor the code so it says _"when finding a comment token, detect indentation and apply it to the token"_ instead.

I tried to rewrite `applyFix()`:
```php
    {
        $lineEnding = $this->whitespacesConfig->getLineEnding();
        foreach ($tokens as $index => $token) {
            if (!$token->isGivenKind($this->tokenKinds)) {
                continue;
            }

            $whitespace = '';
            $previousIndex = $index - 1;
            if ($tokens[$previousIndex]->isWhitespace()) {
                $whitespace = $tokens[$previousIndex]->getContent();
                --$previousIndex;
            }
            if ($tokens[$previousIndex]->isGivenKind(T_OPEN_TAG)) {
                $whitespace = preg_replace('/[^\s]/', '', $tokens[$previousIndex]->getContent()).$whitespace;
            }

            if (!preg_match('/\R([ \t]*)$/', $whitespace, $matches)) {
                continue;
            }

            $indentation = $matches[1];
            $lines = preg_split('/\R/', $token->getContent());
            $replaceCurrentCommentToken = true;
            foreach ($lines as $lineNumber => $line) {
                if (0 === $lineNumber) {
                    continue;
                }

                $line = ltrim($line);
                if ($token->isGivenKind(T_COMMENT) && (!isset($line[0]) || '*' !== $line[0])) {
                    if ('all_multiline' !== $this->configuration['comment_type']) {
                        $replaceCurrentCommentToken = false;
                        break;
                    }
                    continue;
                }

                if (!isset($line[0])) {
                    $line = '*'.$line;
                } elseif ('*' !== $line[0]) {
                    $line = '* '.$line;
                }

                $lines[$lineNumber] = $indentation.' '.$line;
            }

            if ($replaceCurrentCommentToken) {
                $tokens[$index] = new Token([$token->getId(), implode($lineEnding, $lines)]);
            }
        }
    }
```
As you can see, it only checks if the previous token is a whitespace (or an open tag, due to the way PHP tokenizes it). This also removes the need to use functions like `(mb_)strpos`. All tests pass. WDYT?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
keradus: none

Slamdunk:
@julienfalque 👍 

I've also added you as an `@author` and added two optimized, the `\S` and `continue 2`.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,155 @@
...
+            $lines = preg_split('/\R/', $token->getContent());
+            foreach ($lines as $lineNumber => $line) {
+                if (0 === $lineNumber) {
+                    continue;
+                }
+
+                $line = ltrim($line);
+                if ($token->isGivenKind(T_COMMENT) && (!isset($line[0]) || '*' !== $line[0])) {
+                    if ('all_multiline' !== $this->configuration['comment_type']) {
+                        continue 2;
```
keradus: big :-1: for `continue X`, it's basically a [goto](http://php.net/manual/en/control-structures.goto.php)
goto

Slamdunk: I disagree: yes `continue` and `break` are `goto`s, but `continue X` doesn't differ from `continue`.
With this `continue 2` I had the chance to remove code without adding new one, resulting in less code, one less variable and one less `if` with the same tests, that still pass.
Moreover, the skip for the current cicle is done where the condition happens and not in a step further.

julienfalque:
> I've also added you as an `@author`

This was not required as I never do it myself but thanks :)

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
keradus: less code doesn't mean it's easier to maintain

Slamdunk: Not that would be a valid argument, but just to get know of what others implements:
current projects that use `continue X`:

1. [composer](https://github.com/composer/composer/blob/1.4.2/src/Composer/DependencyResolver/Solver.php#L786)
1. doctrine [ORM](https://github.com/doctrine/doctrine2/blob/v2.5.6/lib/Doctrine/ORM/UnitOfWork.php#L2622) & [DBAL](https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Platforms/SqlitePlatform.php#L1034)
1. [monolog](https://github.com/Seldaek/monolog/blob/1.23.0/src/Monolog/Processor/IntrospectionProcessor.php#L77)
1. [symfony](https://github.com/symfony/symfony/blob/v3.3.2/src/Symfony/Component/Process/Pipes/WindowsPipes.php#L62)
1. [zend framework](https://github.com/zendframework/zend-mail/blob/release-2.8.0/src/Storage/Maildir.php#L73)
1. [phpunit](https://github.com/sebastianbergmann/php-code-coverage/blob/5.2.1/src/Node/File.php#L244)

keradus: analysed only first sample.
example in method of 200 lines written 6 (!!!) years ago.
It doesn't prove it's nice and good approach to follow, just nobody had a need to touch it in ages.

GrahamCampbell: > yes continue and break are gotos

No they are not. They are structured control statements. Goto is not structured.

Slamdunk: @keradus all first example are commit with more than 2 years and this is actually the point: `continue X` is no more no less maintainable than other statement.
The [last example](https://github.com/sebastianbergmann/php-code-coverage/commit/ac04311721b75aa3a1536d3143c9fae2fdb902f1#diff-fd0ca4d34d5a26ca6c7435bf9b3e9b3aR244) is written by Sebastian Bergmann 6 months ago.

@GrahamCampbell ok, so? Can you prove a statement pro or against `control` that can't be proven for `control X` too?

Slamdunk:
@julienfalque I see your change request is still alive, are there any ther changes pending that I don't notice?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Multiline doc comment: enforce asterisk start on each line [PSR-5] and align them.',
```
julienfalque: I would update the description: _Each line of multi-line DocComments must have an asterisk and must be aligned with the one_.

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('comment_type', 'Whether to align doc-like multiline comments if all lines start with an asterisk [`phpdocs_like`] or all multile comments  with mixed content on lines that start with an asteristk [`all_multiline`]'))
```
julienfalque: I would update the description: _Whether to fix PHPDoc comments only (`phpdocs_only`), any multi-line comment whose lines all start with an asterisk (`phpdocs_like`) or any multi-line comment (`all_multiline`)._

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
julienfalque: While I'm not against using `continue 2`, I think it can be replaced with something easier to read. As it's here to check whether all lines of the comment start with an `*`, maybe you could do the same with a regular expression before iterating the lines instead?

Slamdunk: Yes, we can do it:
```php
if ($token->isGivenKind(T_COMMENT) && 'all_multiline' !== $this->configuration['comment_type'] && 1 === preg_match('/\R(?:\R|\s*[^\s\*])/', $token->getContent())) {
    continue;
}
```
Do you find it easier to read, understand and maintain?
@keradus do you?

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
Slamdunk: Updated

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
Slamdunk: Updated

julienfalque: Sorry, just noticed there is a missing word in the text I suggested: _... aligned with the **first** one._

Slamdunk: Done :)

src/Fixer/Phpdoc/AlignMultilineCommentFixer.php
```diff
@@ -0,0 +1,156 @@
...
+
+            $indentation = $matches[1];
+            $lines = preg_split('/\R/', $token->getContent());
+            foreach ($lines as $lineNumber => $line) {
+                if (0 === $lineNumber) {
+                    continue;
+                }
+
+                $line = ltrim($line);
+                if ($token->isGivenKind(T_COMMENT) && (!isset($line[0]) || '*' !== $line[0])) {
```
julienfalque: Is this `if` still relevant with the regexp check you added above?

Slamdunk: Yes, when `comment_type === all_multiline`

Slamdunk: See, you are a reviewer and didn't spot the fact that the code is duplicated but mandatory, how about a future maintainer?
I still vote for the `continue 2` :|

julienfalque: The conditions are slightly different actually: one checks that the comment contains a least one line without `*` while the other one checks that the current line has an `*`.

Slamdunk: As far as i can see
```
'/\R(?:\R|\s*[^\s\*])/'
```
is the same as
```php
foreach ($lines as $line) {
    $line = ltrim($line);
    if (!isset($line[0]) || '*' !== $line[0]) {
        // code
    }
}
```

julienfalque: Yep, but the first check is done only once, not on each line. Maybe you could reuse the regular expression in some way that would not require to iterate each line. But I'm fine with current code TBH.

Slamdunk: If `// code` is a `continue 2`, they are the same.

Better not raise complexity even further, is you are ok now I'm ok too.

keradus:
Thank you @Slamdunk.

SpacePossum:
thanks @Slamdunk nice to have in upcoming release :)

PR #2858, opened 2017-06-16
Support generic types

keradus:


keradus:
ref #2841

tests for annotation itself for now are working, yet there are few more cases when utests are failing for some fixers

src/DocBlock/Annotation.php
```diff
@@ -16,10 +16,16 @@
...
  *
  * @author Graham Campbell <graham@alt-three.com>
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
 class Annotation
 {
     /**
+     * @internal
+     */
+    const REGEX_TYPES = '(?<types>(?<type>(?:[\\\\@$]?\w+(?:\[\])?)|(?<generic>[\\\\@$]?\w+<(?:[\\\\@$]?\w+,\s*)?(?:(?&types)|(?&generic))>))(?:\|(?:(?&type)|(?&generic)))*)';
```
keradus: i thought I saw `@this` somewhere, but it was `$this` :D indeed, could be simplified

src/DocBlock/Annotation.php
```diff
@@ -147,7 +160,25 @@ public function getTag()
      */
     public function getTypes()
     {
-        return explode('|', $this->getTypesContent());
+        if (null === $this->types) {
+            $this->types = explode('|', $this->getTypesContent());
```
keradus: nice catch, it's from old implementation when i was compering old behaviour vs new one

src/DocBlock/Annotation.php
keradus: astually, we have `@this` in `PhpdocReturnSelfReferenceFixerTest`.
i will add test to `AnnotationTest` as well

src/DocBlock/Annotation.php
```diff
@@ -16,10 +16,39 @@
...
+    # <simple> is any non-array, non-generic, non-alternated type, eg `int` or `\Foo`
+    # <generic> is generic collection type, like `array<string, int>`, `Collection<Item>` and more complex like `Collection<int, \null|SubCollection<string>>`
+    # <type> is <simple>, <simple>[] or <generic> type, like `int`, `bool[]` or `Collection<ItemKey, ItemVal>`
+    # <types> is one or more types alternated via `|`, like `int|bool[]|Collection<ItemKey, ItemVal>`
+    (?<types>
+        (?<type>
+            (?:(?&simple)(?:\[\])?)
+            |
+            (?<generic>
+                (?<simple>
```
julienfalque: I would swap the `simple` matching group here with the first reference to it a few lines above.

src/DocBlock/Annotation.php
```diff
@@ -16,10 +16,39 @@
...
+                    [\\\\@$]?[\\\\\w]+
```
julienfalque: No need to match `\` in the first character class as it will be matched by the second.

keradus: good catch, originally I was checking only `\null`, then realised it could be `Foo\Bar`

src/DocBlock/Annotation.php
keradus: done

src/DocBlock/Annotation.php
```diff
@@ -16,10 +16,46 @@
...
+    # <generic> is generic collection type, like `array<string, int>`, `Collection<Item>` and more complex like `Collection<int, \null|SubCollection<string>>`
+    # <type> is <simple>, <array> or <generic> type, like `int`, `bool[]` or `Collection<ItemKey, ItemVal>`
+    # <types> is one or more types alternated via `|`, like `int|bool[]|Collection<ItemKey, ItemVal>`
+    (?<types>
+        (?<type>
+            (?<array>
+                (?&scalar)\[\]
+            )
+            |
+            (?<scalar>
```
julienfalque: Now I would swap `scalar` matching group with the first reference to it a few lines above 😁.

src/DocBlock/Annotation.php
```diff
@@ -16,10 +16,46 @@
...
  */
 class Annotation
 {
     /**
+     * Regex to match any types, shall be used with `x` modifier.
+     *
+     * @internal
+     */
+    const REGEX_TYPES = '
+    # <scalar> is any non-array, non-generic, non-alternated type, eg `int` or `\Foo`
```
julienfalque: `\Foo` isn't a scalar.

keradus: it is. scalar, not primitvie

src/DocBlock/Annotation.php
keradus: i find this cleaner, as defined separatelly, nested defs like `<generic>` are not well readable

src/DocBlock/Annotation.php
julienfalque: I think you are confusing scalar and primitive:

>  PHP supports ten primitive types.
>
> - Four scalar types: boolean, integer, float (floating-point number, aka double), string.
> - Four compound types: array, object, callable, iterable.
> - And finally two special types: resource, NULL.

http://php.net/manual/en/language.types.intro.php

keradus: then in php terminology it's neither scalar nor primitive...

julienfalque: What about _&lt;simple&gt;_ then? Or _&lt;single&gt;_?

keradus: `<type>` is kind of `<single>`, as there is no alternation. single could mean item of collection, like types vs type.
simple :+1:

SpacePossum: :+1: for simple, does this apply to simple values as well like `false` in `@return false|array<int, false>` ?

tests/DocBlock/AnnotationTest.php
```diff
@@ -239,7 +331,7 @@ public function provideTypesCases()
         return array(
             array(array('Foo', 'null'), array('Bar[]'), '     * @param Foo|null $foo', '     * @param Bar[] $foo'),
             array(array('false'), array('bool'), '*   @return            false', '*   @return            bool'),
-            array(array('RUNTIMEEEEeXCEPTION'), array('Throwable'), "\t@throws\t  \t RUNTIMEEEEeXCEPTION\t\t\t\t\t\t\t\n\n\n", "\t@throws\t  \t Throwable\t\t\t\t\t\t\t\n\n\n"),
+            array(array('RUNTIMEEEEeXCEPTION'), array('Throwable'), "* \t@throws\t  \t RUNTIMEEEEeXCEPTION\t\t\t\t\t\t\t\n\n\n", "* \t@throws\t  \t Throwable\t\t\t\t\t\t\t\n\n\n"),
```
SpacePossum:
```
*\t@throws
*@throws
```

would still work? I'm not sure why this change is needed beside maybe adding the `*`

src/DocBlock/Annotation.php
keradus: yes. anything directly typehinted with php7 I would say

tests/DocBlock/AnnotationTest.php
keradus: yes, there was a missing `*`.
I will add tests for those.

keradus: added

keradus:
nullable types supported

PR #2851, opened 2017-06-14
Tokens - fix ensureWhitespaceAtIndex

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2837
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2836

ping @GrahamCampbell 

the reason the test pass on 2.2 is because we allow;
> $whitespace = '';
>$this[$index] = new Token(array(T_WHITESPACE, $whitespace));

on master we don't.
UTest covers both path's

PR #2861, opened 2017-06-16
Ported fix onto master to show that it doesn't work

GrahamCampbell:
Demonstrates that https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2851 doesn't work on master.


SpacePossum:
thanks for checking, looking at it I think this might a issue with removing cleared tokens before testing the tokens match the expected tokens.

tests/Tokenizer/TokensTest.php
```diff
@@ -888,6 +888,27 @@ public function provideEnsureWhitespaceAtIndexCases()
     {
         return [
             [
+                '<?php echo 1;',
```
SpacePossum: Please add this at line 883
```php
        $this->assertTrue($tokens->isChanged(), 'Tokens collection must be marked as changed after fixing.');
        $tokens->clearEmptyTokens();
```
after that the tests should be green

GrahamCampbell: Done.

SpacePossum: thanks :)
>$this->assertTrue($tokens->isChanged()

this was incorrect of me, as not all tests make a change, sorry about that :(

SpacePossum:
replaced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2870

PR #2863, opened 2017-06-17
FileSpecificCodeSample - add tests

localheinz:
This PR

* [x] covers `FileSpecificCodeSample`

Follows #2830.

### DIFF
--- a/tests/FixerDefinition/FileSpecificCodeSampleTest.php
+++ b/tests/FixerDefinition/FileSpecificCodeSampleTest.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\FixerDefinition;
+
+use PhpCsFixer\FixerDefinition\FileSpecificCodeSample;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\FixerDefinition\FileSpecificCodeSample
+ */
+final class FileSpecificCodeSampleTest extends TestCase
+{
+    public function testImplementsFileSpecificCodeSampleInterface()
+    {
+        $sample = new FileSpecificCodeSample(
+            file_get_contents(__FILE__),
+            new \SplFileInfo(__FILE__)
+        );
+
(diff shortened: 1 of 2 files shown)


tests/FixerDefinition/FileSpecificCodeSampleTest.php
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    public function testDefaults()
+    {
+        $sample = new FileSpecificCodeSample(
+            file_get_contents(__FILE__),
+            new \SplFileInfo(__FILE__)
+        );
+
+        $this->assertNull($sample->getConfiguration());
```
julienfalque: I would add assertions for `getCode()` and `getSplFileInfo()` as well.

localheinz: Doing this below, this is just to demonstrate that `$configuration` is an optional parameter and that its value defaults to `null` unless provided.

julienfalque: I know, but not testing them here is assuming they always work as expected when `$configuration` is not provided.

tests/FixerDefinition/FileSpecificCodeSampleTest.php
```diff
@@ -0,0 +1,71 @@
...
+final class FileSpecificCodeSampleTest extends TestCase
+{
+    public function testImplementsFileSpecificCodeSampleInterface()
+    {
+        $sample = new FileSpecificCodeSample(
+            file_get_contents(__FILE__),
+            new \SplFileInfo(__FILE__)
+        );
+
+        $this->assertInstanceOf('PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface', $sample);
```
SpacePossum: can we use
```
FileSpecificCodeSampleInterface::class
```
?

localheinz: But what about [`.travis.yml`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/.travis.yml#L23-L26)?

julienfalque: Oh right, forgot this PR targets 2.2!

localheinz: Definitely up for the `class` keyword, though - not being able to use it feels like traveling back in time about 3.5 years.

SpacePossum: oh damn, sorry for the suggestion :(

localheinz: Ha, it's nice that most of the time all of you are keeping an eye out! 👍

keradus: will be done on 2.3 line

keradus:
Thank you @localheinz.

localheinz:

PR #2864, opened 2017-06-19
WhitespacesAwareFixerInterface clean up

Slamdunk:
Tests pass and `$this->whitespacesConfig` is never called

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocSeparationFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocSeparationFixer.php
@@ -16,7 +16,6 @@ use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\DocBlock\Annotation;
 use PhpCsFixer\DocBlock\DocBlock;
 use PhpCsFixer\DocBlock\TagComparator;
-use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
It's a bug that it's not called. It should be.

Slamdunk:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/DocBlock/Line.php#L134

julienfalque:
Looks good to me but this should target 2.2.

Slamdunk:
Target changed to `2.2` and rebased

SpacePossum:
@GrahamCampbell can you elaborate on your comment, I wonder about this as well.

Slamdunk:
@SpacePossum it's easy: the whitespace configuration was used, but now the only functionalities that interact with new-lines are:

1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/Fixer/Phpdoc/PhpdocSeparationFixer.php#L104
1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/Fixer/Phpdoc/PhpdocSeparationFixer.php#L170

A method that find itself the newline style:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/DocBlock/Line.php#L134

So everything is ok

SpacePossum:
thanks @Slamdunk :)

keradus:
Thank you @Slamdunk.

Issue #96, opened 2012-06-25
weird issue with quotes in string in "include" fixer


jokeyrhyme:
This is a bit annoying, but my request validation code keeps getting borked up by the "include" fixer.

This:

```
<?php
exit('POST must include "file"');
```

becomes:

``` <?php
exit('POST must include 'file'');
```

For some reason, the word "file" (with double-quotes) is changed to have single-quotes. This, of course, makes the line invalid (the first single-quote string is now terminated prior to the word "file" and the word "file" is just dangling out in the open.

PR #2865, opened 2017-06-20
AutoReview\FixerTest - test configuration samples

SpacePossum:
Test that all build in fixers:
- provide a sample with `default` configuration if applicable (i.e. all configuration have a default)
- if default config sample is available it must be the first sample
- there should not be two samples with default configuration (readability)

### DIFF
--- a/src/FixerDefinition/FixerDefinitionInterface.php
+++ b/src/FixerDefinition/FixerDefinitionInterface.php
@@ -35,7 +35,7 @@ interface FixerDefinitionInterface
     public function getConfigurationDescription();
 
     /**
-     * @return null|array $defaultConfiguration     null for non-configurable fixer
+     * @return null|array $defaultConfiguration null for non-configurable fixer
      *
      * @deprecated will be removed in 3.0
      */
--- a/src/FixerDefinition/FixerDefinitionInterface.php
+++ b/src/FixerDefinition/FixerDefinitionInterface.php
@@ -28,14 +28,14 @@ interface FixerDefinitionInterface
     public function getDescription();
 
     /**
-     * @return null|string $configurationDescription null for non-configurable fixer
+     * @return null|string null for non-configurable fixer
      *
      * @deprecated will be removed in 3.0
      */
     public function getConfigurationDescription();
 
     /**
-     * @return null|array $defaultConfiguration null for non-configurable fixer
+     * @return null|array null for non-configurable fixer
      *
      * @deprecated will be removed in 3.0
      */
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -72,10 +72,7 @@ final class FixerTest extends TestCase
                 $this->assertTrue($fixerIsConfigurable, sprintf('[%s] Sample #%d has configuration, but the fixer is not configurable.', $fixerName, $sampleCounter));
                 $this->assertInternalType('array', $config, sprintf('[%s] Sample #%d configuration must be an array or null.', $fixerName, $sampleCounter));
 
-                $sortedConfig = $this->sortConfiguration($config);
-                $this->assertSame($config, $sortedConfig, sprintf('[%s] Please sort the configuration of sample #%d.', $fixerName, $sampleCounter));
-
-                $configSamplesProvided[$sampleCounter] = $sortedConfig;
+                $configSamplesProvided[$sampleCounter] = $config;
             } elseif ($fixerIsConfigurable) {
                 if (!$sample instanceof VersionSpecificCodeSampleInterface) {
                     $this->assertArrayNotHasKey('default', $configSamplesProvided, sprintf('[%s] Multiple non-versioned samples with default configuration.', $fixerName));
@@ -257,30 +254,6 @@ final class FixerTest extends TestCase
         return array_map(function (FixerInterface $fixer) {
             return array($fixer);
         }, $fixers);
-    }
-
-    private function sortConfiguration(array $configuration)
-    {
-        reset($configuration);
-        $key = key($configuration);
-        if (is_string($key)) {
-            $key = strtolower($key);
-            if (false !== strpos(strtolower($key), 'order') || false !== strpos(strtolower($key), 'sort')) {
-                return $configuration;
-            }
-        }
-
-        is_int($key)
-            ? sort($configuration)
-            : ksort($configuration);
-
-        foreach ($configuration as $key => $value) {
-            if (is_array($value)) {
-                $configuration[$key] = $this->sortConfiguration($value);
-            }
-        }
-
-        return $configuration;
     }
 
     private function getAllFixers()
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -37,6 +37,7 @@ final class FixerTest extends TestCase
         'header_comment' => array('header' => true),
     );
 
+    // do not modify this structure without prior discussion
     private $allowedFixersWithoutDefaultCodeSample = array(
         'general_phpdoc_annotation_remove' => true,
     );
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -12,9 +12,10 @@
 
 namespace PhpCsFixer\Tests\AutoReview;
 
-use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\DefinedFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerOption;
 use PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface;
 use PhpCsFixer\FixerDefinition\VersionSpecificCodeSampleInterface;
 use PhpCsFixer\FixerFactory;
@@ -24,6 +25,7 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ * @author SpacePossum
  *
  * @internal
  *
@@ -32,6 +34,11 @@ use PHPUnit\Framework\TestCase;
  */
 final class FixerTest extends TestCase
 {
+    // do not modify this structure without prior discussion
+    private $allowedRequiredOptions = array(
+        'header_comment' => array('header' => true),
+    );
+
     /**
      * @param FixerInterface $fixer
      *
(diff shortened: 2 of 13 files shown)


src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
```diff
@@ -105,7 +104,6 @@ protected function createConfigurationDefinition()
         $annotations = new FixerOptionBuilder('annotations', 'List of annotations to remove, e.g. `["author"]`.');
         $annotations = $annotations
             ->setAllowedTypes(array('array'))
-            ->setDefault(array())
```
keradus: BC breaker, revert it !

SpacePossum: sure not a bug? configuring this as `array()` will result the fixer doing nothing

keradus: yes, as any of General fixer behave. we already have few non-general fixers that approves empty configs and skip the fixing.
the only rule that **requires** configuration is header, as we cannot guess what header comment shall be there.

SpacePossum: PSR0 has a hard requirement on the `dir` as well I think. But sure I can revert this and adjust the test if needed.

keradus: looks like it could work without:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/Basic/Psr0Fixer.php#L95
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/Fixer/Basic/Psr0FixerTest.php#L59


actually, only `header_comment` throws `RequiredFixerConfigurationException` intentionally, so only this fixer require config. in `psr0` there is indeed missing of default value for `dir` config key, but if one won't pass any configuration (`psr0 => true`), then configs are not resolved and everything works properly. sure, we shall add that default empty value for clarity

tests/AutoReview/FixerTest.php
```diff
@@ -74,23 +97,78 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+
             $this->assertFalse(
                 $duplicatedCodeSample,
-                sprintf('[%s] Code sample #%d duplicates #%d.', $fixer->getName(), $sampleCounter, $duplicatedCodeSample)
+                sprintf('[%s] Code sample #%d duplicates #%d.', $fixerName, $sampleCounter, $duplicatedCodeSample)
             );
         }
 
+        if ($fixerIsConfigurable) {
+            $options = $fixer->getConfigurationDefinition()->getOptions();
```
julienfalque: I would remove the two `/** @var FixerOption $option */` comments below and add one here:
```php
/** @var FixerOption[] $options */
$options = $fixer->getConfigurationDefinition()->getOptions();
```

tests/AutoReview/FixerTest.php
```diff
@@ -74,23 +97,78 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+
+            // test that all configuration options are used at least one time in the samples
+
+            $optionsNames = array();
+            /** @var FixerOption $option */
+            foreach ($options as $option) {
+                $optionsNames[$option->getName()] = $option->hasDefault() ? 1 : 0;
+            }
+
+            if (count($optionsNames)) {
```
julienfalque: You could `count($options)` here and include the loop above in this `if`.

SpacePossum: don't think so, `options` will always be set here (with a count), however we want to know if after filtering there are still options left

tests/AutoReview/FixerTest.php
```diff
@@ -83,14 +108,55 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
             );
         }
 
+        if ($fixerIsConfigurable) {
+            if (isset($configSamplesProvided['default'])) {
+                reset($configSamplesProvided);
+                $this->assertSame('default', key($configSamplesProvided), 'First code sample must be for the default configuration.');
+            } else {
+                if (!isset($this->allowedRequiredOptions[$fixerName])) {
+                    $this->fail(sprintf('[%s] Has no sample for default configuration.', $fixerName));
```
keradus: i wounder could we do that `assertSame + fail` with one assertion... i like to have assertions in tests, not `fail` directly, imo later shall be only inside assertion definition ;)

tests/AutoReview/FixerTest.php
julienfalque: What filtering?

tests/AutoReview/FixerTest.php
```diff
@@ -74,23 +96,64 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+                $this->assertSame('default', key($configSamplesProvided), sprintf('[%s] First sample must be for the default configuration.', $fixerName));
+            } else {
+                $this->assertArrayHasKey($fixerName, $this->allowedRequiredOptions, sprintf('[%s] Has no sample for default configuration.', $fixerName));
+            }
+
+            // test that all configuration options are used at least one time in the samples
+
+            /** @var FixerOption[] $options */
+            $options = $fixer->getConfigurationDefinition()->getOptions();
+            $optionsNames = array();
```
julienfalque: This variable can be moved inside the `if` below.

tests/AutoReview/FixerTest.php
```diff
@@ -74,23 +96,64 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+                        foreach ($config as $name => $value) {
+                            ++$optionsNames[$name];
+                        }
+                    }
+                }
+
+                $optionsNames = array_filter(
+                    $optionsNames,
+                    function ($count) {
+                        return 1 > $count;
```
julienfalque: I would write `0 === $count`.

tests/AutoReview/FixerTest.php
```diff
@@ -74,23 +96,64 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+
+            // test that all configuration options are used at least one time in the samples
+
+            /** @var FixerOption[] $options */
+            $options = $fixer->getConfigurationDefinition()->getOptions();
+            $optionsNames = array();
+
+            if (count($options)) {
+                foreach ($options as $option) {
+                    $optionsNames[$option->getName()] = 0; //$option->hasDefault() ? 1 : 0;
```
julienfalque: Why did you replace `$option->hasDefault() ? 1 : 0` with `0`?

SpacePossum: after the PR of @keradus we are sure at this point in the code a fixer has a code sample with default configuration. Here I start counting that _besides_ that code sample with default config there is at least one other code sample using the option, therefor we always have a sample with the option _other_ than the default.
(we still might have duplicates though, I address that later I hope, but configuration arrays are not easy to compare)

SpacePossum:
expect tests to fail

src/FixerDefinition/FixerDefinitionInterface.php
```diff
@@ -35,7 +35,7 @@ public function getDescription();
     public function getConfigurationDescription();
 
     /**
-     * @return null|array $defaultConfiguration     null for non-configurable fixer
```
SpacePossum: missing fixer for this, or bugged fixer

GrahamCampbell: Depends. PHP allows methods to not actually declare their parameters. It is possible that someone might write doc like this, and then access the bound parameters using func_get_args.

SpacePossum: agreed :+1: ,
I made comment for myself to check on later for another case; the multiple spaces between:
```$defaultConfiguration``` and ```null```
which I think is better to be a single space as there is nothing else in the PHPDoc to align with.

**Comment by SpacePossum** [comment]
@julienfalque 

since you are the author of the following fixers I would like your input on the following incomplete sample cases:

```
2) PhpCsFixer\Tests\AutoReview\FixerTest::testFixerDefinitions with data set #94 (PhpCsFixer\Fixer\DoctrineAnnotation\DoctrineAnnotationBracesFixer Object (...))
[doctrine_annotation_braces] No sample with configuration options "ignored_tags".
Failed asserting that actual size 1 matches expected size 0.

/home/possum/work/PHP-CS-Fixer/tests/AutoReview/FixerTest.php:167

3) PhpCsFixer\Tests\AutoReview\FixerTest::testFixerDefinitions with data set #95 (PhpCsFixer\Fixer\DoctrineAnnotation\DoctrineAnnotationIndentationFixer Object (...))
[doctrine_annotation_indentation] No sample with configuration options "ignored_tags".
Failed asserting that actual size 1 matches expected size 0.

/home/possum/work/PHP-CS-Fixer/tests/AutoReview/FixerTest.php:167

4) PhpCsFixer\Tests\AutoReview\FixerTest::testFixerDefinitions with data set #96 (PhpCsFixer\Fixer\DoctrineAnnotation\DoctrineAnnotationSpacesFixer Object (...))
[doctrine_annotation_spaces] No sample with configuration options "ignored_tags", "around_parentheses", "around_commas", "around_argument_assignments", "around_array_assignments".
Failed asserting that actual size 5 matches expected size 0.

/home/possum/work/PHP-CS-Fixer/tests/AutoReview/FixerTest.php:167
```

if you've no time or don't find this interesting I can puzzle something together so don't feel pushed ;)

File: tests/AutoReview/FixerTest.php [hunk 9]
```diff
@@ -182,6 +259,30 @@ public function provideFixerConfigurationDefinitionsCases()
...
+        reset($configuration);
+        $key = key($configuration);
+        if (is_string($key)) {
+            $key = strtolower($key);
+            if (false !== strpos(strtolower($key), 'order') || false !== strpos(strtolower($key), 'sort')) {
+                return $configuration;
+            }
+        }
+
+        is_int($key)
```
keradus: side effect. use regular condition.

tests/AutoReview/FixerTest.php
```diff
@@ -182,6 +259,30 @@ public function provideFixerConfigurationDefinitionsCases()
         }, $fixers);
     }
 
+    private function sortConfiguration(array $configuration)
+    {
+        reset($configuration);
+        $key = key($configuration);
+        if (is_string($key)) {
+            $key = strtolower($key);
+            if (false !== strpos(strtolower($key), 'order') || false !== strpos(strtolower($key), 'sort')) {
```
keradus: why such condition ?

julienfalque:
I would add the following samples:

**DoctrineAnnotationBracesFixer**

```php
new CodeSample(
    "<?php\n/**\n * @Foo()\n * @Bar()\n */\nclass Bar {}",
    array('ignored_tags' => array('Bar'))
),
```

**DoctrineAnnotationIndentationFixer**

```php
new CodeSample(
    "<?php\n/**\n *  @Foo(\n *   foo=\"foo\"\n *  )\n *  @Bar(\n *   bar=\"bar\"\n *  )\n */\nclass Bar {}",
    array('ignored_tags' => array('Bar'))
),
```

**DoctrineAnnotationSpacesFixer**

```php
new CodeSample(
    "<?php\n/**\n * @Foo ( foo  = \"foo\" )\n */\nclass Bar {}",
    array('around_parentheses' => false)
),
new CodeSample(
    "<?php\n/**\n * @Foo ( foo = \"foo\"  ,bar = \"bar\" )\n */\nclass Bar {}",
    array('around_commas' => false)
),
new CodeSample(
    "<?php\n/**\n * @Foo ( foo  = \"foo\" )\n */\nclass Bar {}",
    array('around_argument_assignments' => false)
),
new CodeSample(
    "<?php\n/**\n * @Foo ( {\"foo\"  : \"foo\"} )\n */\nclass Bar {}",
    array('around_array_assignments' => false)
),
new CodeSample(
    "<?php\n/**\n * @Foo ( )\n * @Bar ( )\n */\nclass Bar {}",
    array('ignored_tags' => array('Bar'))
),
```

---
Thinking about it: some options have been deprecated on master. Adding tests that require to have one sample for each option will lead test builds to always fail once merged on master because the samples will either be missing or use deprecated code. We should anticipate that, not sure how though. Maybe a second PR that targets master and updates the test to ignore deprecated option? WDYT?

SpacePossum:
I've removed the tests that cannot be completed in a sane way, i.e. we cannot compare configuration values to detect double samples, we cannot test that each configuration option is used at least one time as the test cannot see if the option is deprecated.
I propose to merge as is or to close, as the amount of time and review rounds on this is getting out of hand.

src/FixerDefinition/FixerDefinitionInterface.php
SpacePossum: looking again, the issue is that `@return <type> <var>` does not make sense

SpacePossum:
I think this one is ready to go?

keradus:
Thank you @SpacePossum.

PR #2866, opened 2017-06-20
Add SingleLineCommentStyleFixer, deprecate HashToSlashCommentFixer

Slamdunk:
```diff
 <?php
-/* a comment */
+// a comment
 $a = 1;
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -495,6 +495,7 @@ Choose from the list of available rules:
 * **hash_to_slash_comment** [@Symfony]
 
   Single line comments should use double slashes ``//`` and not hash ``#``.
+  DEPRECATED: Use "single_line_comment_style" instead.
 
 * **header_comment**
 
@@ -1095,6 +1096,16 @@ Choose from the list of available rules:
   Each namespace use MUST go on its own line and there MUST be one blank
   line after the use statements block.
 
+* **single_line_comment_style**
+
+  Converts multi-line comments that have only one line of actual content
+  into single-line comments, and hash comments to slash.
+
+  Configuration options:
+
+  - ``comment_type`` (``'all'``, ``'hash'``, ``'star'``): fix star comment ``/* */`` [``star``],
+    hash comment ``#`` [``hash``], or both [``all``]; defaults to ``'all'``
+
 * **single_quote** [@Symfony]
 
   Convert double quotes to single quotes for simple strings.
@@ -1106,11 +1117,6 @@ Choose from the list of available rules:
 * **standardize_not_equals** [@Symfony]
 
   Replace all ``<>`` with ``!=``.
-
-* **star_to_slash_comment**
-
-  Converts multi-line comments that have only one line of actual content
-  into single-line comments.
 
 * **strict_comparison**
 
--- a/README.rst
+++ b/README.rst
@@ -1104,7 +1104,7 @@ Choose from the list of available rules:
   Configuration options:
 
   - ``comment_types`` (``array``): list of comment types to fix; defaults to
-    ``['star', 'hash']``
+    ``['asterisk', 'hash']``
 
 * **single_quote** [@Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -1107,6 +1107,11 @@ Choose from the list of available rules:
 
   Replace all ``<>`` with ``!=``.
 
+* **star_to_slash_comment**
+
+  Converts multi-line comments that have only one line of actual content
+  into single-line comments.
+
 * **strict_comparison**
 
   Comparisons should be strict.
--- a/README.rst
+++ b/README.rst
@@ -1098,13 +1098,13 @@ Choose from the list of available rules:
 
 * **single_line_comment_style**
 
-  Converts multi-line comments that have only one line of actual content
-  into single-line comments, and hash comments to slash.
+  Single-line comments and multi-line comments with only one line of
+  actual content should use the ``//`` syntax.
 
   Configuration options:
 
-  - ``comment_type`` (``'all'``, ``'hash'``, ``'star'``): fix star comment ``/* */`` [``star``],
-    hash comment ``#`` [``hash``], or both [``all``]; defaults to ``'all'``
+  - ``comment_types`` (``array``): list of comment types to fix; defaults to
+    ``['star', 'hash']``
 
 * **single_quote** [@Symfony]
 
--- a/src/AbstractProxyFixer.php
+++ b/src/AbstractProxyFixer.php
@@ -45,9 +45,33 @@ abstract class AbstractProxyFixer extends AbstractFixer
     /**
      * {@inheritdoc}
      */
+    public function isRisky()
+    {
+        return $this->proxyFixer->isRisky();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getPriority()
     {
         return $this->proxyFixer->getPriority();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supports(\SplFileInfo $file)
+    {
+        return $this->proxyFixer->supports($file);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setWhitespacesConfig(WhitespacesFixerConfig $config)
+    {
+        return $this->proxyFixer->setWhitespacesConfig($config);
     }
 
     /**
--- a/src/AbstractProxyFixer.php
+++ b/src/AbstractProxyFixer.php
@@ -71,7 +71,8 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     public function setWhitespacesConfig(WhitespacesFixerConfig $config)
     {
-        return $this->proxyFixer->setWhitespacesConfig($config);
+        parent::setWhitespacesConfig($config);
+        $this->proxyFixer->setWhitespacesConfig($config);
     }
 
     /**
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -661,7 +661,7 @@ final class ConfigurationResolver
             return true;
         }, $rules));
 
-        /** @var string[] $availableFixers */
+        /** @var string[] $configuredFixers */
         $configuredFixers = array_keys($ruleSet->getRules());
 
         /** @var string[] $availableFixers */
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -661,10 +661,10 @@ final class ConfigurationResolver
             return true;
         }, $rules));
 
-        /* @var string[] $availableFixers */
+        /** @var string[] $availableFixers */
         $configuredFixers = array_keys($ruleSet->getRules());
 
-        /* @var string[] $availableFixers */
+        /** @var string[] $availableFixers */
         $availableFixers = array_map(function (FixerInterface $fixer) {
             return $fixer->getName();
         }, $this->createFixerFactory()->getFixers());
--- a/src/Fixer/Comment/HashToSlashCommentFixer.php
+++ b/src/Fixer/Comment/HashToSlashCommentFixer.php
@@ -12,18 +12,19 @@
 
 namespace PhpCsFixer\Fixer\Comment;
 
-use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\AbstractProxyFixer;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Tokenizer\Token;
-use PhpCsFixer\Tokenizer\Tokens;
 
 /**
  * Changes single comments prefixes '#' with '//'.
  *
  * @author SpacePossum
+ *
+ * @deprecated in 2.4, proxy to SingleLineCommentStyleFixer
  */
-final class HashToSlashCommentFixer extends AbstractFixer
+final class HashToSlashCommentFixer extends AbstractProxyFixer implements WhitespacesAwareFixerInterface
 {
     /**
      * {@inheritdoc}
@@ -31,7 +32,7 @@ final class HashToSlashCommentFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Single line comments should use double slashes `//` and not hash `#`.',
+            sprintf('Single line comments should use double slashes `//` and not hash `#`. DEPRECATED: Use "%s" instead.', $this->proxyFixer->getName()),
             [new CodeSample('<?php # comment')]
         );
     }
@@ -39,21 +40,11 @@ final class HashToSlashCommentFixer extends AbstractFixer
     /**
      * {@inheritdoc}
      */
-    public function isCandidate(Tokens $tokens)
+    protected function createProxyFixer()
     {
-        return $tokens->isTokenKindFound(T_COMMENT);
-    }
+        $fixer = new SingleLineCommentStyleFixer();
+        $fixer->configure(['comment_type' => 'hash']);
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
-    {
-        for ($i = 1, $count = count($tokens); $i < $count; ++$i) {
-            $originalContent = $tokens[$i]->isGivenKind(T_COMMENT) ? $tokens[$i]->getContent() : null;
-            if (null !== $originalContent && '#' === $originalContent[0]) {
-                $tokens[$i] = new Token([T_COMMENT, '//'.substr($originalContent, 1)]);
-            }
-        }
+        return $fixer;
     }
 }
--- a/src/Fixer/Comment/HashToSlashCommentFixer.php
+++ b/src/Fixer/Comment/HashToSlashCommentFixer.php
@@ -13,7 +13,6 @@
 namespace PhpCsFixer\Fixer\Comment;
 
 use PhpCsFixer\AbstractProxyFixer;
-use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 
@@ -24,7 +23,7 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
  *
  * @deprecated in 2.4, proxy to SingleLineCommentStyleFixer
  */
-final class HashToSlashCommentFixer extends AbstractProxyFixer implements WhitespacesAwareFixerInterface
+final class HashToSlashCommentFixer extends AbstractProxyFixer
 {
     /**
      * {@inheritdoc}
@@ -43,7 +42,7 @@ final class HashToSlashCommentFixer extends AbstractProxyFixer implements Whites
     protected function createProxyFixer()
     {
         $fixer = new SingleLineCommentStyleFixer();
-        $fixer->configure(['comment_type' => 'hash']);
+        $fixer->configure(['comment_types' => ['hash']]);
 
         return $fixer;
     }
--- a/src/Fixer/Comment/SingleLineCommentStyleFixer.php
+++ b/src/Fixer/Comment/SingleLineCommentStyleFixer.php
@@ -14,9 +14,9 @@ namespace PhpCsFixer\Fixer\Comment;
 
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
-use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerConfiguration\FixerOptionValidatorGenerator;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Token;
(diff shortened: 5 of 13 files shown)


Slamdunk:
I was thinking to also add this:
```diff
 <?php
-/*
- * One liner multilne comment
- */
+// One liner multilne comment
 $a = 1;
```
WDYT?

keradus:
> ```php
>  <?php
> -/* a comment */
> +// a comment
>  $a = 1;
> ```

vs

> MultilineToSingleCommentFixer

where is that multiline ?

README.rst
```diff
@@ -595,6 +595,10 @@ Choose from the list of available rules:
 
   *Risky rule: risky if any of the functions ``intval``, ``floatval``, ``doubleval``, ``strval`` or ``boolval`` are overridden.*
 
+* **multiline_to_single_comment**
+
+  Convert one-liner multiline comment to single line comment.
```
keradus: there is no such thing as `one-liner multi-liner`

Slamdunk:
> There are only two hard things in Computer Science: cache invalidation and naming things.
> -- Phil Karlton

`/* */` allow a comment to span on multiple lines, `//` doesn't.

`One-liner` refers to the actual weight of the comment, `multiline comment` refers to the comment type used.

Any suggestion is welcome, I'm really bad at naming things.

julienfalque:
What about _"Converts multi-line comments that have only one line of actual content into single-line comments."_?

Slamdunk:
@julienfalque 👍 

And what about the fixer name?

julienfalque:
I have no better suggestion for the fixer name for now :)

keradus:
you have 3 types of comments in php - based on stars, slashes and hashes

Slamdunk:
So, what would you suggest?
Something like `StarToSlashCommentFixer`?

SpacePossum:
maybe one configurable fixer to update all single line comments to one style (which effectively would deprecate `hash_to_slash_comment`)?

README.rst
GrahamCampbell: > one-liner multiline

That makes no sense, lol.

Slamdunk: @GrahamCampbell thank you for your constructive contribution 👍

Slamdunk:
I've renamed the fixer to `StarToSlashCommentFixer` and now it acts like this:
```diff
 <?php
-/* first comment */
+// first comment
 $a = 1;
-/*
- * second comment
- */
+// second comment
 $b = 2;
 /*
  * third
  * comment
  */
 $c = 3;
```
I would appreciate very much a review on the tests to be sure no case is missing.

@SpacePossum don't know the policy for deprecations here, as of yet the two fixers do (slightly) different things so I would keep them both.

julienfalque:
I like @SpacePossum's suggestion. Then the fixer name could be `SingleLineCommentStyleFixer` or something similar (yes, I find `StarToSlashCommentFixer` kind of weird 😛).

Slamdunk:
The idea of merging two similar fixers is good, but it have to be planned.
Now we have:

- `hash_to_slash_comment` that **always** moves `#` to `//`
- `star_to_slash_comment` that **sometimes** moves `/* */` to `//`

They are orthogonal and can coexists without repeated code.
IMHO we can keep them as they are now.

@keradus WDYT?
In case of deprecaton and merging, what's the correct process to follow?

julienfalque:
To deprecate a fixer, you should:
- mark it as `@deprecated since 2.x [...]`;
- add a `trigger_error('...', E_USER_DEPRECATED)` call in its constructor;
- update its description to mention that it is deprecated (and suggest the new fixer, if any);
- mark tests that cover the fixer as `@legacy` (with assertions on the deprecation notice when relevant).

Slamdunk:
@julienfalque 👍 

I'll start writing the deprecation as soon as @keradus give us a feedback

keradus:
1. `E_USER_DEPRECATED` in fixer constructor is not good idea. welcome to deprecation hell of fixer factory that will automatically register (via instance creation) of all fixers @julienfalque 
2. I do not know for now how much of the code will be similar, yet idea is similar for sure.
If both @SpacePossum and @julienfalque are for unified fixer, I would go that way as well.
3. also, I would allow to change multiline `/*..*/` comments to multiline `//` as well (with an option), but not `/**...*/` docblock comments
4. currently existing fixer you gonna deprecate shall became proxy fixer in your PR.

Slamdunk:
> also, I would allow to change multiline `/*..*/` comments to multiline `//` as well (with an option), but not `/**...*/` docblock comments

DocBlock comments are not part of this Fixer

> currently existing fixer you gonna deprecate shall became proxy fixer in your PR.

What happens if someone config both the new fixer and the proxy?
Or better, what should happen?

keradus:
both of them will be executed, but old fixer will perform no changes

julienfalque:
> `E_USER_DEPRECATED` in fixer constructor is not good idea. welcome to deprecation hell of fixer factory that will automatically register (via instance creation) of all fixers

👍 this should be implemented in `FixerFactory` in a separate PR. @Slamdunk please just add the `@deprecated` tag and update the description of the deprecated fixer.

Slamdunk:
So no code-enforced warning like `trigger_error`? Ok

keradus:
if you want to, go with separate PR after this one got merged

Slamdunk:
Pushed changes:

1. renamed `StarToSlashCommentFixer` to `SingleLineCommentStyleFixer`
1. made it configurable to do `HashToSlashCommentFixer` job too
1. changed `HashToSlashCommentFixer` to be a proxy for `SingleLineCommentStyleFixer`, with **less** proprity
1. signed `HashToSlashCommentFixer` deprecated, `HashToSlashCommentFixerTest` legacy
1. added deprecation description to `HashToSlashCommentFixer`

#### Notes

To implement correctly the [proxy pattern](https://en.wikipedia.org/wiki/Proxy_pattern) `HashToSlashCommentFixer` doesn't extends anymore `AbstractFixer` and implements directly the interfaces needed.
As such, in order to avoid copy-paste of the trivial code of `getName` method, I moved it to `Utils`, seizing the opportunity to add a test to it.

keradus:
> > shall became proxy fixer
> 
> To implement correctly the proxy pattern ...

Use https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/AbstractProxyFixer.php instead

src/Fixer/Comment/HashToSlashCommentFixer.php
```diff
@@ -12,48 +12,95 @@
...
-        return $tokens->isTokenKindFound(T_COMMENT);
+        return Utils::formatFixerName(get_called_class());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return $this->realFixer->getPriority() - 1;
```
keradus: actually, i don't see much value in defining different priority.
if there is any, add integration test to show it

Slamdunk:
@keradus Didn't see `AbstractProxyFixer` and I'll use it, but there could be an improvement: `setWhitespacesConfig` is proxied, I'm going to proxy it too.

keradus:
please doublecheck if there are currently any proxy fixers that are manually implementing `setWhitespacesConfig`

Slamdunk:
Edits done. I've also proxied `isRisky` and `supports` that previously were missing.

tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
```diff
@@ -0,0 +1,306 @@
...
+                '<?php /* start-end */',
+            ],
+        ];
+    }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testAllCases()
```
julienfalque: Please extract the expected and input strings into a data provider, then write a new test that uses it but does not call `$this->fixer->configure(...)`.

README.rst
```diff
@@ -1095,6 +1096,16 @@ Choose from the list of available rules:
...
   line after the use statements block.
 
+* **single_line_comment_style**
+
+  Converts multi-line comments that have only one line of actual content
+  into single-line comments, and hash comments to slash.
+
+  Configuration options:
+
+  - ``comment_type`` (``'all'``, ``'hash'``, ``'star'``): fix star comment ``/* */`` [``star``],
```
keradus: what about
```
- `hash` (boolean)
- `star` (boolean)
```

?

src/AbstractProxyFixer.php
```diff
@@ -53,6 +61,22 @@ public function getPriority()
...
+    {
+        return $this->proxyFixer->supports($file);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setWhitespacesConfig(WhitespacesFixerConfig $config)
+    {
+        return $this->proxyFixer->setWhitespacesConfig($config);
```
keradus: what if fixer is not whitespace aware ?

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -0,0 +1,108 @@
...
+                $tokens[$index] = new Token([$token->getId(), '//'.substr($content, 1)]);
+                continue;
+            }
+            if ('hash' === $config || '/*' !== substr($content, 0, 2) || preg_match('/[^\s\*].*\R.*[^\s\*]/s', $commentContent)) {
+                continue;
+            }
+            $nextTokenIndex = $index + 1;
+            if (isset($tokens[$nextTokenIndex])) {
+                $nextToken = $tokens[$nextTokenIndex];
+                if (false === strpos($nextToken->getContent(), $lineEnding)) {
```
julienfalque: This assumes the next token is a whitespace while it can be something else. I added the following case:
```php
<?php
   /* first *//*
   second
   */

```
which should not be changed but:
```diff
--- Expected
+++ Actual
@@ @@
 <?php
-   /* first *//*
-   second
-   */
+   // first// second
```
---
Also, I would replace `false === strpos($nextToken->getContent(), $lineEnding)` with `!preg_match('/\R/', $nextToken->getContent())` for two reasons:
- `strpos("\n\n\n\n", "\r\n") // false`: the content has newlines, but not the expected ones;
- this removes the need to use `$lineEnding` and thus you don't need to implement `WhitespacesAwareFixerInterface` at all, as it is used here only.

README.rst
Slamdunk: In another PR @julienfalque [asked me](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2857#discussion_r122779498) to switch from booleans to one option with multiple values (for a valid reason) and I supposed this to be the preferred way.
I have no preference, if you ask for it again I'll change

src/AbstractProxyFixer.php
Slamdunk: Thought it but was missing the right way to implement it.
Now found: I'll call the parent and the proxyfixer, so the config is available for both, and the check is done for both.

README.rst
julienfalque: IMO this option should be an array that can contain a subset of `['star', 'hash']`. `'all'` shouldn't be allowed as it would enable new behavior added in the future by default (very unlikely, but still).

But I think boolean options would be fine as well.

julienfalque: IMO this option should be an array that can contain a subset of `['star', 'hash']`. `'all'` shouldn't be allowed as it would enable new behavior added in the future by default (very unlikely, but still).

But I think boolean options would be fine as well.

src/Fixer/Comment/SingleLineCommentStyleFixer.php
Slamdunk: 👍 

The example you provided should be changed:
```diff
 <?php
-   /* first *//*
-   second
-   */
+   /* first */// second
```
Added to the tests and fixed, push soon.

Slamdunk:
Feedbacks implemented

julienfalque:
I just tried running the new fixer on the project itself and it changed this (in `ConfigurationResolver`):
```diff
-/* @var string[] $availableFixers */
+// @var string[] $availableFixers
```

IMO the behavior of the fixer is correct because this comment should have been a DocComment instead of a regular C-style comment, this should be fixer manually.

Could this be a bug in `phpdoc_to_comment` though? Or a conflict between the two fixers?

Anyway can you please add the new fixer to project's `.php_cs.dist`?

Slamdunk:
Those comment were introduced ex-novo by @localheinz in https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/393e0cc9c96275aa40ab370cdc0788bb5c69c99d#diff-62b374855f5961454b817b144c531e27R624 (no previous commit) so yes, they are just a missed-DocComment.

Fixer added to `.php_cs.dist`; those two cases manually moved to DocComments.

src/Console/ConfigurationResolver.php
```diff
@@ -661,10 +661,10 @@ private function validateRules(array $rules)
             return true;
         }, $rules));
 
-        /* @var string[] $availableFixers */
+        /** @var string[] $availableFixers */
```
SpacePossum: :+1: for fixing, but should be;
```
/** @var string[] $configuredFixers */
```

Slamdunk: Done

Slamdunk:
@keradus sorry for previous squashes, but I'm used to them because I find they make `git blame` and history easier to read, even though I agree should not be used in a WIP.
So I would appreciate a squash before merge, or if you prefer or find hurdles, I'll rebase and squash it for you.

keradus:
my policy:
- never squash/ammend in PR
- always squash during merging PR (except edge cases with multiple authors...)

we started doing the squashing automatically before github started to offer that ;)

src/AbstractProxyFixer.php
```diff
@@ -53,6 +61,23 @@ public function getPriority()
...
+    {
+        return $this->proxyFixer->supports($file);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setWhitespacesConfig(WhitespacesFixerConfig $config)
+    {
+        parent::setWhitespacesConfig($config);
```
keradus: still, parent may not be whitespace aware

Slamdunk: I'm sorry I don't understand this.
If the real fixer needs a WhitespaceConfig (aka WC):

1. And the proxy isnt configuring it manually, real fixer setWC will raise an Exception
1. And the proxy configures it manually, everything is OK
1. And the proxy expects setWC to be called automatically and it doesn't implements WCAwareInterface, real fixer setWC will raise an Exception
1. And the proxy expects setWC to be called automatically and it implements WCAwareInterface, everything is OK
1. If in the tests someone calls proxy->setWC manually to make the tests pass without implementing WCAwareInterface, parent::setWC will raise an Exception

What is the case you talking of?

tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
```diff
@@ -0,0 +1,321 @@
...
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideAllCases
+     */
+    public function testAllCases($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
```
Slamdunk: I'm sorry I don't get it.
`configure` is called at Fixer contruction, isn't it?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.3.2/src/AbstractFixer.php#L53
What case exactly should I cover?

julienfalque: Sorry I wasn't clear. By default, the constructor calls `configure([])`, which sets the default configuration. The test I want you to add should call `configure(['hash', 'star'])`, which should have the exact same effect.

Slamdunk: May I ask you what the benefit would be?

julienfalque: To prove the fixer works the same with both implicit and explicit default config.

Slamdunk: This should be tested in the code that handles allowed values + config default, shouldn't it?
Do you have records of a fixer that misbehaved that way, since the time new config system was introduced?

julienfalque: Not having failure records of a scenario is not a valid reason to not test that scenario IMO.

However such failure here should only happen if you do something really weird when overriding `configure` method, which is very unlikely so yes, let's forget this. Although I would say having more tests can't hurt :)

Slamdunk: > Not having failure records of a scenario is not a valid reason to not test that scenario IMO.

I disagree. A test should (pick one or more):

1. Explain an intent
1. Demonstrate an intent or a functionality
1. Explore bad/unhappy paths
1. Show separation of responsabilities
1. Prove a bug

Which all sum up to have a purpose.

The work done in v2 for configuration is astonishing in term of OOP.
Now we can trust configuration bonds are well checked and enforced because of the new structure.
If and when a bug will be found in a `configure`, tests should be addressed in the context they belong.

The test you suggest would have two flaws:

1. It wouldn't have a purpose
1. It would test the right thing in the wrong place

> Although I would say having more tests can't hurt :)

Actually more tests can hurt: imagine a day when you write legit code, that make the test you suggest fails. If you don't know why the test was written, how do you pick a thoughtful choice?

tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
```diff
@@ -0,0 +1,321 @@
...
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Comment\SingleLineCommentStyleFixer
+ */
+final class SingleLineCommentStyleFixerTest extends AbstractFixerTestCase
+{
+    public function testInvalidConfig()
+    {
+        $this->setExpectedException(\PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException::class);
```
julienfalque: Please import the class name.

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -0,0 +1,123 @@
...
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('comment_types', 'List of comment types to fix'))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([
+                    (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf(['star', 'hash']),
+                ])
+                ->setDefault(['star', 'hash'])
```
julienfalque: _Asterisk_ would probably be more accurate than _star_.

GrahamCampbell: I think `star` is ok, and it's what I think of when I see a `*. Kleene star, etc. :P

keradus: we use asterisk in other places

Slamdunk:
*star* to *asterisk* done

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -0,0 +1,123 @@
...
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class SingleLineCommentStyleFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    private $starEnabled;
```
julienfalque: Please update the property name to match the option.

tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
```diff
@@ -0,0 +1,322 @@
...
+        $this->fixer->configure(['abc']);
+    }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideStarCases
+     */
+    public function testStar($expected, $input = null)
```
julienfalque: Please update the test and data provider name to match the option.

Slamdunk:
*star* to *asterisk* done (2)

src/AbstractProxyFixer.php
keradus: right, i forgot that parent::setWC is checking the interface and raise exception, if not available
:+1:

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -0,0 +1,123 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class SingleLineCommentStyleFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    private $asteriskEnabled;
+    private $hashEnabled;
```
keradus: typehinting is missing

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -0,0 +1,123 @@
...
+                $nextToken = $tokens[$nextTokenIndex];
+                if (!$nextToken->isWhitespace() || !preg_match('/\R/', $nextToken->getContent())) {
+                    continue;
+                }
+
+                $tokens[$nextTokenIndex] = new Token([$nextToken->getId(), ltrim($nextToken->getContent(), " \t")]);
+            }
+
+            $content = '//';
+            if (preg_match('/[^\s\*]/', $commentContent)) {
```
keradus: `1 ===`, be strict.

keradus:
Thank you @Slamdunk.

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -78,28 +112,30 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
                 }
 
                 $tokens[$nextTokenIndex] = new Token([$nextToken->getId(), ltrim($nextToken->getContent(), " \t")]);
             }
 
             $content = '//';
-            if (preg_match('/[^\s\*]/', $commentContent)) {
-                $content = '// '.preg_replace('/[\s\*]*([^\s\*](.+[^\s\*])?)[\s\*]*/', '\1', $commentContent);
+            if (1 === preg_match('/[^\s\*]/', $commentContent)) {
+                $content = '// '.preg_replace('/[\s\*]*([^\s\*](?:.+[^\s\*])?)[\s\*]*/', '\1', $commentContent);
```
Slamdunk: I always thought raising the complexity of a regex would also slow it down, in this case is the opposite, thank you for the TIL
```php
$patterns = [
    'without-?:'    => '/[\s\*]*([^\s\*](.+[^\s\*])?)[\s\*]*/',
    'with-?:'       => '/[\s\*]*([^\s\*](?:.+[^\s\*])?)[\s\*]*/',
];

foreach ($patterns as $type => $pattern) {
    $start = microtime(true);
    for ($i = 0; $i < 10000000; ++$i) {
        preg_replace($pattern, '\1', ' * * a b c * * ');
    }
    $elapsed = microtime(true) - $start;

    var_dump(sprintf('%10s => %s', $type, $elapsed));
}

// string(29) "without-?: => 2.6824641227722"
// string(29) "   with-?: => 2.6746520996094"
```

keradus: I came from perl ;P (the one that P in PCRE stands for)

PR #2867, opened 2017-06-20
VersionSpecification - Fix copy-paste typo

SpacePossum:


src/FixerDefinition/VersionSpecification.php
```diff
@@ -36,19 +36,21 @@
...
-        }
+        if (null !== $maximum) {
+            if (!is_int($maximum) || 1 > $maximum) {
+                throw new \InvalidArgumentException('Maximum needs to be either <null> or an integer greater than 0.');
+            }
 
-        if (null !== $maximum && null !== $minimum && $maximum < $minimum) {
-            throw new \InvalidArgumentException('Maximum should not be less than the minimum');
+            if (null !== $minimum && $maximum < $minimum) {
+                throw new \InvalidArgumentException('Maximum should not be less than the minimum.');
```
julienfalque: "_...should not be **lower** than..._"?

src/FixerDefinition/VersionSpecification.php
```diff
@@ -36,19 +36,21 @@
     public function __construct($minimum = null, $maximum = null)
     {
         if (null === $minimum && null === $maximum) {
-            throw new \InvalidArgumentException('Either minimum or maximum need to be specified');
+            throw new \InvalidArgumentException('Minimum and/or maximum need to be specified.');
         }
 
         if (null !== $minimum && (!is_int($minimum) || 1 > $minimum)) {
-            throw new \InvalidArgumentException('Minimum needs to be either null or an integer greater than 0');
+            throw new \InvalidArgumentException('Minimum needs to be either <null> or an integer greater than 0.');
```
keradus: I see this null change in 2nd PR. why that change ?

SpacePossum: for clarity, but happy to revert

keradus: let us do it globally or not at all.

SpacePossum: :+1:  I'll revert than

keradus:
@SpacePossum , as it was raised by your concern, could you take a look ?

src/FixerDefinition/VersionSpecification.php
```diff
@@ -36,19 +36,21 @@
     public function __construct($minimum = null, $maximum = null)
     {
         if (null === $minimum && null === $maximum) {
-            throw new \InvalidArgumentException('Either minimum or maximum need to be specified');
+            throw new \InvalidArgumentException('Minimum and/or maximum need to be specified.');
```
keradus: or === and|xor

SpacePossum: >Either minimum or maximum

means either one of them, so not both?

>Minimum or maximum 

mean one of the two but not both?

I dunno, maybe bin the PR?

julienfalque: I would use _and/or_ too but I often use the equivalent in french as well so not sure it's actually valid in english :)

keradus:
```
noun: OR; noun: or; plural noun: ors
1.
a logical operation which gives the value one if at least one operand has the value one, and otherwise gives a value of zero.
```

by google

SpacePossum:
```
Used to link alternatives.
‘a cup of tea or coffee’
‘are you coming or not?’
‘I either take taxis or walk everywhere’
‘it doesn't matter whether the theory is right or wrong’
```

by the oxford dictionaries (https://en.oxforddictionaries.com/definition/or)
2nd sample implies; you cannot come and both not come

keradus:
Thank you @SpacePossum.

GrahamCampbell:
Did you forget to close this?

keradus:
no, I got up from the desk

GrahamCampbell:
Ha, lol. I assumed you had a tool that committed and closed. ;)

PR #2869, opened 2017-06-20
Fix handling required configuration

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2865#discussion_r122938913

### DIFF
--- a/src/AbstractFixer.php
+++ b/src/AbstractFixer.php
@@ -63,6 +63,10 @@ abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
 
     final public function fix(\SplFileInfo $file, Tokens $tokens)
     {
+        if ($this instanceof ConfigurableFixerInterface && null === $this->configuration) {
+            throw new RequiredFixerConfigurationException($this->getName(), 'Configuration is required.');
+        }
+
         if (0 < $tokens->count() && $this->isCandidate($tokens) && $this->supports($file)) {
             $this->applyFix($file, $tokens);
         }
--- a/src/Fixer/Basic/Psr0Fixer.php
+++ b/src/Fixer/Basic/Psr0Fixer.php
@@ -99,7 +99,7 @@ class InvalidName {}
             $path = str_replace('\\', '/', $file->getRealPath());
             $dir = dirname($path);
 
-            if ($this->configuration['dir']) {
+            if ('' !== $this->configuration['dir']) {
                 $dir = substr($dir, strlen(realpath($this->configuration['dir'])) + 1);
 
                 if (false === $dir) {
--- a/src/Fixer/Basic/Psr0Fixer.php
+++ b/src/Fixer/Basic/Psr0Fixer.php
@@ -42,6 +42,13 @@ final class Psr0Fixer extends AbstractPsrAutoloadingFixer implements Configurati
 namespace PhpCsFixer\FIXER\Basic;
 class InvalidName {}
 ',
+                    new \SplFileInfo(__FILE__)
+                ),
+                new FileSpecificCodeSample(
+                    '<?php
+namespace PhpCsFixer\FIXER\Basic;
+class InvalidName {}
+',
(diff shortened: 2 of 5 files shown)


src/AbstractFixer.php
```diff
@@ -63,6 +63,10 @@ public function __construct()
 
     final public function fix(\SplFileInfo $file, Tokens $tokens)
     {
+        if ($this instanceof ConfigurableFixerInterface && null === $this->configuration) {
```
SpacePossum: this only test top level, i.e. options of a configuration can be required as well, shouldn't we test those as well?

julienfalque: These checks are already done with Symfony's OptionsResolver so `$this->configuration !== null` means the configuration has been set and is valid.

SpacePossum: guess not https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2865#discussion_r122938913
that was the point that started this, 
```
but if one won't pass any configuration (psr0 => true), then configs are not resolved and everything works properly
```

keradus: first part true, second not

`configuration === null` means that `->configure(null)` was called only in constructor and then ignored (https://github.com/keradus/PHP-CS-Fixer/blob/8387d21b6979a0a1f30900c5bbb847236bac1fa3/src/AbstractFixer.php#L55), and the `->configure(config)` was not called at all (as if it would, then it would throw exception once more, but this time not silenced)

so `configuration===null` means configuration was not set at all, if it would it either be non-empty or throw an exception

src/Fixer/Basic/Psr0Fixer.php
```diff
@@ -92,7 +99,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
             $path = str_replace('\\', '/', $file->getRealPath());
             $dir = dirname($path);
 
-            if (isset($this->configuration['dir'])) {
+            if ($this->configuration['dir']) {
```
keradus: `0` is not string and would not pass option resolving
`"0"` would pass and it's config issue. anyway, here we shall have valid values, any validation shall be done during config resolving

tests/AutoReview/FixerTest.php
```diff
@@ -145,12 +145,29 @@ public function provideFixerDefinitionsCases()
...
         foreach ($configurationDefinition->getOptions() as $option) {
             $this->assertNotEmpty($option->getDescription());
+
+            $this->assertSame(
+                !isset($allowedRequiredOptions[$fixer->getName()][$option->getName()]),
+                $option->hasDefault(),
+                sprintf(
+                    $option->hasDefault()
+                        ? 'Option `%s` of fixer `%s` is wrongly listed in `$allowedRequiredOptions` structure, as it is not required. If you just changed that option to not be required anymore, please adjust mentioned structure.'
+                        : 'Option `%s` of fixer `%s` shall not be required. If you want to introduce new required option please adjust mentioned `$allowedRequiredOptions`.',
```
keradus: :+1:, it was in previous form of this msg

src/AbstractFixer.php
julienfalque: Because the `null !== $this->configuration` check was missing so you could run the fixer without calling `configure()` first. Now you must call it, which will then give the configuration to `OptionsResolver` which will then validate it.

src/Fixer/Basic/Psr0Fixer.php
julienfalque: Yes I meant `'0'`.

Indeed, if we consider `'0'` an invalid configuration value, this should be checked while resolving configuration. But why would we consider it invalid?

keradus: right, it's valid. now i got your point ;)

keradus:
@SpacePossum , as it was raised by your concern, could you take a look ?

PR #2872, opened 2017-06-22
Just calling php-cs-fixer didn't work for me

mudhoney:


SpacePossum:
Hi @mudhoney

Thanks for the PR, I like the suggested change :)
There are 2 details to fix;
- this effects all versions we support, so your PR better fits the 2.2 branch as target
- the README.rst is generated, so the [generator](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/Console/Command/ReadmeCommand.php#L97) needs to be updated as well

mudhoney:
Ok I will try to fix other issues :)

julienfalque:
Closing in favor of #2873.

PR #2876, opened 2017-06-23
DescribeCommand - Show which sample uses the default configuration

SpacePossum:
old
```
$ php php-cs-fixer describe no_extra_consecutive_blank_lines
Description of no_extra_consecutive_blank_lines rule.
Removes extra blank lines and/or blank lines following configuration.

* tokens (array): list of tokens to fix; defaults to ['extra']

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ @@
    <?php

    $foo = array("foo");

   -
    $bar = "bar";
```

new

```
$ php php-cs-fixer describe no_extra_consecutive_blank_lines
Description of no_extra_consecutive_blank_lines rule.
Removes extra blank lines and/or blank lines following configuration.

* tokens (array): list of tokens to fix; defaults to ['extra']

Fixing examples:
 * Example #1. Fixing with the default configuration
   --- Original
   +++ New
   @@ @@
    <?php

    $foo = array("foo");

   -
    $bar = "bar";
```

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -25,6 +25,7 @@ use PhpCsFixer\FixerDefinition\FixerDefinitionInterface;
 use PhpCsFixer\FixerDefinition\VersionSpecificCodeSampleInterface;
 use PhpCsFixer\FixerFactory;
 use PhpCsFixer\RuleSet;
+use PhpCsFixer\RuleSetInterface;
 use PhpCsFixer\StdinFileInfo;
 use PhpCsFixer\Tokenizer\Tokens;
 use Symfony\Component\Console\Command\Command;
@@ -56,9 +57,14 @@ final class DescribeCommand extends Command
     private $fixers;
 
     /**
+     * @var RuleSet
+     */
+    private $ruleSet;
+
+    /**
      * @param null|FixerFactory $fixerFactory
      */
-    public function __construct(FixerFactory $fixerFactory = null)
+    public function __construct(FixerFactory $fixerFactory = null, RuleSetInterface $ruleSet = null)
     {
         parent::__construct();
 
@@ -67,6 +73,7 @@ final class DescribeCommand extends Command
             $fixerFactory->registerBuiltInFixers();
         }
 
+        $this->ruleSet = null === $ruleSet ? new RuleSet() : $ruleSet;
         $this->fixerFactory = $fixerFactory;
     }
 
@@ -105,7 +112,7 @@ final class DescribeCommand extends Command
             $this->describeList($output, $e->getType());
 
             throw new \InvalidArgumentException(sprintf(
-                '%s %s not found.%s',
+                '%s "%s" not found.%s',
                 ucfirst($e->getType()), $name, null === $alternative ? '' : ' Did you mean "'.$alternative.'"?'
             ));
         }
@@ -265,7 +272,8 @@ final class DescribeCommand extends Command
             throw new DescribeNameNotFoundException($name, 'set');
         }
 
-        $ruleSet = new RuleSet(array($name => true));
+        $ruleSetClass = get_class($this->ruleSet);
+        $ruleSet = new $ruleSetClass(array($name => true));
         $rules = $ruleSet->getRules();
         ksort($rules);
 
@@ -320,8 +328,7 @@ final class DescribeCommand extends Command
             return $this->setNames;
         }
 
-        $set = new RuleSet();
-        $this->setNames = $set->getSetDefinitionNames();
+        $this->setNames = $this->ruleSet->getSetDefinitionNames();
         sort($this->setNames);
 
         return $this->setNames;
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -149,11 +149,12 @@ final class DescribeCommand extends Command
         }
 
         if ($fixer instanceof ConfigurationDefinitionFixerInterface) {
-            $output->writeln('Fixer is configurable using following options:');
-
             $configurationDefinition = $fixer->getConfigurationDefinition();
+            $options = $configurationDefinition->getOptions();
 
-            foreach ($configurationDefinition->getOptions() as $option) {
+            $output->writeln(sprintf('Fixer is configurable using following option%s:', 1 === count($options) ? '' : 's'));
+
+            foreach ($options as $option) {
                 $line = '* <info>'.$option->getName().'</info>';
 
                 $allowed = HelpCommand::getDisplayableAllowedValues($option);
@@ -177,7 +178,7 @@ final class DescribeCommand extends Command
                         HelpCommand::toString($option->getDefault())
                     );
                 } else {
-                    $line .= 'required';
+                    $line .= '<comment>required</comment>';
                 }
 
                 $output->writeln($line);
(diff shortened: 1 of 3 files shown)


tests/AutoReview/DescribeCommandTest.php
```diff
@@ -1,64 +0,0 @@
...
-use PHPUnit\Framework\TestCase;
-use Symfony\Component\Console\Tester\CommandTester;
-
-/**
- * @internal
- *
- * @coversNothing
- * @group auto-review
- */
-final class DescribeCommandTest extends TestCase
```
julienfalque: Why moving those tests?

GrahamCampbell:
:+1:

SpacePossum:
>However I wonder if this should target master. This kind of looks like a new feature rather than a bugfix 

why not upping docs on the branch these docs relate to?

>Also: would it be better to show what the default configuration is? 

is always `[`]` ? The options and their defaults are listed above the samples already.

tests/AutoReview/DescribeCommandTest.php
SpacePossum: I was confused about the two test both called `DescribeCommandTest.php` which both test the same class, what is the reason to have two?

julienfalque: - `PhpCsFixer\Tests\Console\Command\DescribeCommandTest`: generic tests with fake data;
- `PhpCsFixer\Tests\AutoReview\DescribeCommandTest`: tests that check the command runs fine with all real fixers.

The separation was made on purpose but I don't remember where we discussed that. Though I agree this is a but confusing so I'm fine with a single class.

julienfalque:
> why not upping docs on the branch these docs relate to?

👍 

> is always `[]` ? The options and their defaults are listed above the samples already.

I meant to compute the default values in such case. I would find that convenient but indeed, maybe it's just overkill.

tests/AutoReview/DescribeCommandTest.php
keradus: it's on purpose indeed. checking execution for all real fixers (not mociking one) is not unit test anymore

SpacePossum: I see, reverted

SpacePossum:
I would say overkill indeed,
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2865
will make sure the code sample with default is always shown first, which is directly after the list of configuration options and their defaults.

src/Console/Command/DescribeCommand.php
```diff
@@ -149,11 +149,12 @@ private function describeRule(OutputInterface $output, $name)
         }
 
         if ($fixer instanceof ConfigurationDefinitionFixerInterface) {
-            $output->writeln('Fixer is configurable using following options:');
-
             $configurationDefinition = $fixer->getConfigurationDefinition();
+            $options = $configurationDefinition->getOptions();
+
+            $output->writeln(sprintf('Fixer is configurable using following option%s:', 2 > count($options) ? '' : 's'));
```
julienfalque: I would write `1 === count($options)`.

keradus: done

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -73,11 +73,11 @@ public function testExecuteOutputWithDecoration()
 Fixes stuff.
 Replaces bad stuff with good stuff.
 
-Fixer is configurable using following options:
+Fixer is configurable using following option:
```
keradus: ha, `options` version is not tested

SpacePossum: yeah, prop. together with this line https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2876/files#diff-88cad9ab17d8e3b415e648e3fd31f3d2L180

SpacePossum: I've update the tests to test a whole bunch of stuff previously not tested.
There are still a lot of cases not tested, but I would like to move on with this PR and not try to cover all things that were not covered previously, I think that can be done in another PR.

src/Console/Command/DescribeCommand.php
```diff
@@ -265,7 +272,8 @@ private function describeSet(OutputInterface $output, $name)
             throw new DescribeNameNotFoundException($name, 'set');
         }
 
-        $ruleSet = new RuleSet(array($name => true));
+        $ruleSetClass = get_class($this->ruleSet);
+        $ruleSet = new $ruleSetClass(array($name => true));
```
keradus: I don't fully get this change. You injected `RuleSet` to constructor, and then you recreate it here. it's messy.
and more, it's wrong, as constructor accept any interface, so given class could have private constructor

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -32,18 +34,31 @@
...
 final class DescribeCommandTest extends TestCase
 {
+    /**
+     * @var Application
+     */
+    private $application;
+
+    protected function setUp()
+    {
+        $this->application = new Application();
```
julienfalque: I don't see much value in creating the application in `setUp()` so I would keep doing it in each test instead. But if you prefer this way, then please clear the property in `tearDown()` to:
- prevent side effects between tests;
- free memory after all tests;

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -209,22 +295,64 @@ private function execute($decorated)
...
 
         return $commandTester;
     }
 }
+
+/**
+ * @internal
+ * @coversNothing
+ */
+final class TestRuleSet implements RuleSetInterface
```
keradus: shall be prophesized

tests/Console/Command/DescribeCommandTest.php
SpacePossum: > * prevent side effects between tests;

its a private property, side effects can only be occur on purpose

> free memory after all tests;

PHPUnit should release the test class itself, clearing its properties as well. There is no other way of releasing the memory (i.e. setting it to `null` does not help)

src/Console/Command/DescribeCommand.php
SpacePossum: >and more, it's wrong, as constructor accept any interface, so given class could have private constructor

The interface https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.3/src/RuleSetInterface.php#L24 forces the constructor to be public

What _is_ messy; is the `RuleSet` being both an instance of a collection of rules and a collection of _all_ sets, which makes this needed.
I.e we do _not_ allow someone to create a `RuleSet` of their own and add it to the set of all `RuleSet`'s as the later does not exists.

tests/Console/Command/DescribeCommandTest.php
```diff
@@ -144,17 +175,70 @@ public function testExecuteWithoutName()
...
      * @return CommandTester
      */
-    private function execute($decorated)
+    private function execute($name, $decorated, $verbosityLevel = OutputInterface::VERBOSITY_NORMAL)
     {
         $fixer = $this->prophesize();
         $fixer->willImplement('PhpCsFixer\Fixer\DefinedFixerInterface');
         $fixer->willImplement('PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface');
 
+        $fixer->configure(array('functions' => array('foo' => 'bar')))->willReturn(null);
```
SpacePossum: Its not wrong, its a test that it is returned, since we cannot test for the correct pseudo type `void` this is the best we can do.

SpacePossum:
I'll rework this when I've the time, thanks all!

tests/Console/Command/DescribeCommandTest.php
julienfalque: > its a private property, side effects can only be occur on purpose

By side effect I mean: the instance is used in some way in a test that may affect another test as they share the same instance.

> PHPUnit should release the test class itself, clearing its properties as well. There is no other way of releasing the memory (i.e. setting it to null does not help)

Not sure it's PHPUnit's responsability to do that, could make sense. But I'm pretty sure it doesn't currently. Why setting the property to `null` wouldn't help?

SpacePossum: >By side effect I mean: the instance is used in some way in a test that may affect another test as they share the same instance.

please provide a POC for this

>Not sure it's PHPUnit's responsability to do that, could make sense. But I'm pretty sure it doesn't currently. Why setting the property to null wouldn't help?

it is, we never set a test class to null

>Why setting the property to null wouldn't help?

why would it?

julienfalque: Setting the property to `null` would remove the only reference to the `Application` instance so PHP should be able to free memory, even if the test class instance itself still exists.

keradus:
reopen until that as a feature request

tests/Console/Command/DescribeCommandTest.php
keradus: between tests - then the `setUp` is called again overwriting last value of the property.

to `null` - we only do it in `tests/Console/ConfigurationResolverTest.php`, it's not needed indeed, it shall be handled by phpunit after whole test file is executed

julienfalque: > between tests - then the setUp is called again overwriting last value of the property.

Totally missed this obvious point. Sorry for being stubborn 😅

SpacePossum:
testing as GH does not show new commits

PR #2877, opened 2017-06-23
Fix PHPMD violations

julienfalque:
I just ran PHPMD on the codebase and fixed reported issue. This will prevent PHPMD errors for the next PR that will change the file.

### DIFF
--- a/src/Fixer/FunctionNotation/VoidReturnFixer.php
+++ b/src/Fixer/FunctionNotation/VoidReturnFixer.php
@@ -203,8 +203,6 @@ final class VoidReturnFixer extends AbstractFixer
     }
 
     /**
-     * Apply the fix to the function definition.
-     *
      * @param Tokens $tokens
      * @param int    $index  The index of the end of the function definition line, EG at { or ;
      */


GrahamCampbell:
Why do we get phpmd to only run on the changed files? Just get it to check everything?

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -203,8 +203,6 @@ private function hasVoidReturn(Tokens $tokens, $startIndex, $endIndex)
     }
 
     /**
-     * Apply the fix to the function definition.
```
GrahamCampbell: Instead of removing this, we could make it more informative?

SpacePossum:
>Why do we get phpmd to only run on the changed files? Just get it to check everything?

performance, there is no need to run it on everything all the time. We run in this case because a PHPMD violation was merged to master, however the should no longer happen when 2.4 because the lowest maintained version (IIRC).

src/Fixer/FunctionNotation/VoidReturnFixer.php
SpacePossum: the method name says it all?

GrahamCampbell:
> performance, there is no need to run it on everything all the time

Possible to implement a cache I guess. Sounds like a job for a different CI service, such as scrutinizer.

SpacePossum:
could be done, but would take considerable time and now this little thingy is blocking PR's so maybe in another follow up PR?

src/Fixer/FunctionNotation/VoidReturnFixer.php
GrahamCampbell: The same is true for many of the other methods in this class, which phpmd has not picked up though too.

SpacePossum: yeah, I would delete them all, but than I again I know I'm a bit of a minimalist on these things

julienfalque: Indeed, but the purpose of this PR is mostly to
> prevent PHPMD errors for the next PR that will change the file.

keradus:
Thank you @julienfalque.

PR #2880, opened 2017-06-24
NoBreakCommentFixer - fix edge case

keradus:


GrahamCampbell:
Perhapse a better way to write this fixer is to find the end of switch blocks, then backward iterate them?

GrahamCampbell:
Unrelated to this PR ish, but what about this test case:

```php
<?php

$foo = 1;

switch ($foo) {
    case 1:
?>
<?php
    default:
        echo 'lol';
}
```

keradus:
don't know, tbh don't care, we don't test such cases at all.
if you care and if it's failing - open a PR.
from my perspective - we could skip files containing multiple php code mixed with html

julienfalque:
Actually the test case suggested by @GrahamCampbell was failing so I added it.

keradus:
Thank you @julienfalque.

PR #2882, opened 2017-06-26
DocBlock\Annotation - Fix parsing of collections with multiple key types

julienfalque:
This fixes parsing collections that have key types like `foo|bar`. Parsing still fails with key types like `foo|bar<baz>` but I'm not sure we do want to support that.


keradus:
Thank you @julienfalque.

PR #2883, opened 2017-06-26
DescribeCommand - Show which sample uses the default configuration 

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2876
which cannot be reopend for whatever reason


keradus:
Thank you @SpacePossum.

PR #2884, opened 2017-06-27
Add custom rulesets

sseidelmann:
Hi everybody,

We are using the Fixer in company for different projects with different technologies. We have a default Styleguide and some concrete styleguides for those frameworks.
So we would have a company wide ruleset with will be extended by these frameworks.

With this pull request you are able to add a new RuleSetDefinition from the configuration file to the whole fix process, eg:
```
<?php
    ->setCustomRuleSets([
        '@MyCompany' => [
            '@PSR2' => true,
        ],
        '@MyCompanyFramework1' => [
            '@MyCompany' => true,
            'no_useless_else' => false,
        ],
        '@MyCompanyFramework2' => [
            '@MyCompany' => true,
            'phpdoc_order' => false,
        ]
    ])
    ->setRules([
        '@MyCompanyFramework2' => true
    ]);
```

Please leave a note about that improvement.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -32,6 +32,7 @@ class Config implements ConfigInterface
     private $name;
     private $phpExecutable;
     private $rules = ['@PSR2' => true];
+    private $customRuleSets = [];
     private $usingCache = true;
 
     public function __construct($name = 'default')
@@ -61,6 +62,14 @@ class Config implements ConfigInterface
     public function getCustomFixers()
     {
         return $this->customFixers;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getCustomRuleSets()
+    {
+        return $this->customRuleSets;
     }
 
     /**
@@ -172,6 +181,16 @@ class Config implements ConfigInterface
     public function setCacheFile($cacheFile)
     {
         $this->cacheFile = $cacheFile;
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setCustomRuleSets(array $ruleSets)
+    {
+        $this->customRuleSets = $ruleSets;
 
         return $this;
     }
--- a/src/Config.php
+++ b/src/Config.php
@@ -23,6 +23,7 @@ class Config implements ConfigInterface
 {
     private $cacheFile = '.php_cs.cache';
     private $customFixers = [];
+    private $customRuleSets = [];
     private $finder;
     private $format = 'txt';
     private $hideProgress = false;
@@ -32,7 +33,6 @@ class Config implements ConfigInterface
     private $name;
     private $phpExecutable;
     private $rules = ['@PSR2' => true];
-    private $customRuleSets = [];
     private $usingCache = true;
 
     public function __construct($name = 'default')
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -35,6 +35,15 @@ interface ConfigInterface
     public function getCustomFixers();
 
     /**
+     * Get rule sets.
+     *
+     * Keys of array are names of ruleset, values are the rules.
+     *
+     * @return array
+     */
+    public function getCustomRuleSets();
+
+    /**
      * Returns files to scan.
      *
      * @return iterable|\Traversable|string[]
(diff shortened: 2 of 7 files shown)


src/Config.php
```diff
@@ -179,6 +188,16 @@ public function setCacheFile($cacheFile)
     /**
      * {@inheritdoc}
      */
+    public function setCustomRuleSets(array $ruleSets)
```
keradus: As we have `registerCustomFixers`, I would rather do `registerCustomRuleSets`

src/ConfigInterface.php
```diff
@@ -35,6 +35,15 @@ public function getCacheFile();
     public function getCustomFixers();
 
     /**
+     * Get rule sets.
+     *
+     * Keys of array are names of ruleset, values are the rules.
+     *
+     * @return array
+     */
+    public function getCustomRuleSets();
```
GrahamCampbell: Major breaking change to add a new method to an interface.

GrahamCampbell:
:-1: on adding this feature. I don't think it's needed, and adds bloat.

src/ConfigInterface.php
keradus: hm... BC breaker

GrahamCampbell: Yeh, I said that 30 seconds earlier. :P

src/RuleSet.php
```diff
@@ -203,7 +210,10 @@
      */
     private $rules;
 
-    public function __construct(array $set = [])
+    /**
+     * {@inheritdoc}
+     */
+    public function __construct(array $set = [], array $customSetDefinitions = [])
```
keradus: it could be nice moment do decouple definitions from ruleset that is suppose to be used during fixing

src/RuleSet.php
```diff
@@ -257,7 +273,7 @@ public function getRules()
      */
     public function getSetDefinitionNames()
     {
-        return array_keys($this->setDefinitions);
+        return array_keys(array_replace($this->customSetDefinitions, $this->setDefinitions));
```
keradus: `array_replace` on assoc array amy not be the best option

src/RuleSetInterface.php
```diff
@@ -21,9 +21,23 @@
...
-    public static function create(array $set = []);
+    /**
+     * Static creation for the rule set.
+     *
+     * @param array $set                  the set of rules
+     * @param array $customSetDefinitions the custom set definitions
+     *
+     * @return mixed
+     */
+    public static function create(array $set = [], array $customSetDefinitions = []);
```
keradus: BC breaker

SpacePossum:
I like the idea :)  so :+1: for that, however we've to figure out how to do this without BC breaks.

keradus:
Hi @sseidelmann , thank you for picking up this topic. I do appreciate it and the feature is welcome.

ping @SpacePossum , you may be interested in this PR.

1. In future, please make a prior discussion before doing the work, we would save time and effort this way
2. no BC breakers are allowed except for `@internal` classes
3. 3 src classes have beed modified, yet only one test class.
4. travis is sad :(

sseidelmann:
@keradus,

Thanks for your Feedback.
"In future, please make a prior discussion before doing the work, we would save time and effort this way" - Thats right...

keradus:
ping @localheinz , I remember you wanted this feature, could you take a look ?

keradus:
ping @sseidelmann

sseidelmann:
@keradus Should i do anything here?

keradus:
points 2-4 of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2884#issuecomment-311323660 for start ;)

keradus:
ping @sseidelmann

keradus:
ping @sseidelmann

keradus:
ping @sseidelmann

keradus:
PR has been abandoned. Closing

salarmehr:
I need this feature too. Any update about another PR after 4 years?

kubawerlos:
@salarmehr what do you need it for? I cannot understand what this feature adds that cannot be achieved by simply having set of rules somewhere and share it between projects.

Issue #2695, opened 2017-04-13
useless else fixer should also convert elseif to if when possible


stof:
When the `else` part of a `elseif` is useless, the fixer should turn it into a `if`

stof:
rule to determine it is useless:

- all previous `if` of the chain broke the execution path (`return`, `continue`, `throw`, `break`, ...)
- it is the last path in the chain (no `else` or `elseif` after it).

Note that the second part of the condition needs to be checked again for previous conditions of the if/elseif chain once a useless else is fixed (as the previous `elseif` becomes the last one)

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2885 IIUC , please reopen if not so (ping @stof cc @julienfalque )

PR #2885, opened 2017-06-27
Add NoSuperfluousElseifFixer

julienfalque:
Implements #2695.

I'm pretty sure we can add more test cases, I'm open to suggestions :)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -29,6 +29,7 @@ $config = PhpCsFixer\Config::create()
         'no_extra_consecutive_blank_lines' => ['tokens' => ['break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block']],
         'no_null_property_initialization' => true,
         'no_short_echo_tag' => true,
+        'no_superfluous_elseif' => true,
         'no_unreachable_default_argument_value' => true,
         'no_useless_else' => true,
         'no_useless_return' => true,
--- a/src/AbstractNoUselessElseFixer.php
+++ b/src/AbstractNoUselessElseFixer.php
@@ -1,0 +1,214 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+abstract class AbstractNoUselessElseFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before NoWhitespaceInBlankLineFixer, NoExtraConsecutiveBlankLinesFixer, BracesFixer and after NoEmptyStatementFixer.
+        return 25;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    protected function elseIsSuperfluous(Tokens $tokens, $index)
+    {
+        $previousBlockStart = $index;
+        do {
+            // Check if all 'if', 'else if ' and 'elseif' blocks above this 'else' always end,
+            // if so this 'else' is overcomplete.
+            list($previousBlockStart, $previousBlockEnd) = $this->getPreviousBlock($tokens, $previousBlockStart);
+
+            // short 'if' detection
+            $previous = $previousBlockEnd;
+            if ($tokens[$previous]->equals('}')) {
+                $previous = $tokens->getPrevMeaningfulToken($previous);
+            }
+
+            if (
+                !$tokens[$previous]->equals(';') ||                              // 'if' block doesn't end with semicolon, keep 'else'
+                $tokens[$tokens->getPrevMeaningfulToken($previous)]->equals('{') // empty 'if' block, keep 'else'
+            ) {
+                return false;
+            }
+
+            $candidateIndex = $tokens->getPrevTokenOfKind(
+                $previous,
+                [
+                    ';',
+                    [T_BREAK],
+                    [T_CLOSE_TAG],
+                    [T_CONTINUE],
+                    [T_EXIT],
+                    [T_GOTO],
+                    [T_IF],
+                    [T_RETURN],
+                    [T_THROW],
+                ]
+            );
+
+            if (
+                null === $candidateIndex
+                || $tokens[$candidateIndex]->equalsAny([';', [T_CLOSE_TAG], [T_IF]])
+                || $this->isInConditional($tokens, $candidateIndex, $previousBlockStart)
+                || $this->isInConditionWithoutBraces($tokens, $candidateIndex, $previousBlockStart)
+            ) {
+                return false;
+            }
+
+            // implicit continue, i.e. delete candidate
+        } while (!$tokens[$previousBlockStart]->isGivenKind(T_IF));
+
+        return true;
+    }
+
+    /**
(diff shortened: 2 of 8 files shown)


.php_cs.dist
```diff
@@ -28,7 +28,7 @@ return PhpCsFixer\Config::create()
         'no_null_property_initialization' => true,
         'no_short_echo_tag' => true,
         'no_unreachable_default_argument_value' => true,
-        'no_useless_else' => true,
+        'no_useless_else' => ['convert_elseif_to_if' => true],
```
keradus: not sure about the option... imo, it's just a next case we didn't spot before, shall be fixed always.
WDYT @julienfalque @SpacePossum

julienfalque: So we would consider this a bugfix? I'm ok with that, though I'll have to target 2.2 then :p

src/Fixer/ControlStructure/NoUselessElseFixer.php
```diff
@@ -198,6 +229,35 @@ private function fixEmptyElse(Tokens $tokens, $index)
      */
     private function clearElse(Tokens $tokens, $index)
```
SpacePossum: This method name is misleading now, as it does not just clear the `else` statement anymore.
Please refactor the logic into a new private method and leave this one as is.

tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
```diff
@@ -954,6 +954,164 @@ public function provideIsInConditionWithoutBracesCases()
...
+                        '.'
+                        break;
+                    } else/* bar */if (7)
+                        return 2 + 3;
+                    else {# baz
+                        die(\'foo\');
+                    }
+                }',
+            ],
+        ];
```
SpacePossum: please add;
```php
<?php
            if ($a === false)
            {
                if ($v) { $ret = "foo"; }
                elseif($a)
                    die;
            }
            elseif($a)
                $ret .= $value;

            return $ret;
```

```php
<?php

if ($a)
    echo 1;
else if ($b)
    die;
else {
    echo 2;
}
```

```php
<?php
if ($a) {
    echo 1;
} else if ($b)
    die;
else {
    echo 2;
}
```

```php
<?php
if ($a) {
    echo 1;
} else if ($b) {
    die;
} else {
    echo 2;
}
```

src/Fixer/ControlStructure/NoUselessElseFixer.php
julienfalque: I thought about it: converting a `elseif` into a `if` is basically a `else` removal so I decided to keep on method. But all the new logic is in a specific `if` so using a new method instead makes sense too.

tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
julienfalque: Done.

SpacePossum:
for the record; I disagree this being a bug fix as I left it out by design. I think adding it to the fixer is a break on the single responsibility principle, same as why I always said `no` to request of letting the fixer as `fix` the identification afterwards.

That being sad I would like to hear what cases this PR is supposed to fixed;
```php
<?php
if ($a) {
    echo 1;
} elseif ($b) {
    echo 2;
}
```

```php
<?php
if ($a) {
    die;
} elseif ($b) {
    echo 2;
}
```

```php
<?php
if ($a) {
    die;
} elseif ($b) {
    echo 2;
} else {
    echo 3;
}
```

```php
<?php
if ($a) {
    die;
} elseif ($b) {
    die;
} else {
    echo 3;
}
```

```php
<?php
if ($a) {
    echo 1;
} elseif ($b) {
    die;
} else {
    echo 3;
}
```

<?php
if ($a) {
    die;
} elseif ($b) {
    die;
} elseif ($c) {
    echo 1;
} else {
    echo 3;
}

julienfalque:
> For the record; I disagree this being a bug fix as I left it out by design.

Ok so let's keep it as new feature.

> I think adding it to the fixer is a break on the single responsibility principle, same as why I always said no to request of letting the fixer also fix the identification after removing the else.
i.e. I think this should be a new fixer.

IMO adding this behavior to the current fixer makes sense. As most of the logic is shared, I'd like to hear more opinions about this before refactoring.

---
About the cases you listed, here are the changes the fixer should apply:

```diff
 <?php
 if ($a) {
     echo 1;
 } elseif ($b) {
     echo 2;
 }
```

```diff
 <?php
 if ($a) {
     die;
-} elseif ($b) {
+}
+if ($b) {
     echo 2;
 }
```

```diff
 <?php
 if ($a) {
     die;
-} elseif ($b) {
+}
+if ($b) {
     echo 2;
 } else {
     echo 3;
 }
```

```diff
<?php
 if ($a) {
     die;
-} elseif ($b) {
+}
+if ($b) {
     die;
-} else {
+
     echo 3;
-}
+
```

```diff
 <?php
 if ($a) {
     echo 1;
 } elseif ($b) {
     die;
 } else {
     echo 3;
 }
```

I think those cases are already covered though.

GrahamCampbell:
```
+}
+if ($b) {
```

I would rather it inserted a newline inbetween those.

SpacePossum:
IMHO there is a big diff. between the fixer and that is why they should be separated;
the `use else fixer` starts at an `T_ELSE` and only analyzes up to see if all paths lead to a stop.

Your fixer should, IMO, also look down. I.e. not fix:

```php
<?php
if ($a) {
    die;
} elseif ($b) {
    echo 2;
} else {
    echo 3;
}
```

here the `else` statement is not `useless`, so should not be touched by the `useless else fixer`.
If people want to change these cases that is fine by me, but it would consider it a feature loss if we force people to also change `elseif` statements.

julienfalque:
> @GrahamCampbell I would rather it inserted a newline inbetween those.

What you want is #2384.

Adding this empty line would be an assumption that most people would add it, which is not true IMO. I guess we could say the same about moving the `if` to the next line but I don't think most people write something like
```php
if ($a) {
    // ...
} if ($b) {
    // ...
}
```
---
> @SpacePossum here the else statement is not useless, so should not be touched by the useless else fixer.

In your example the `else` would not be removed. The fixer would only convert the `elseif` into an `if`, which looks valid to me, unless I am missing something.

> If people want to change these cases that is fine by me, but it would consider it a feature loss if we force people to also change elseif statements.

Current implementation requires to enable an option to fix your example, nothing will change for people that don't want to fix it.

SpacePossum:
>In your example the else would not be removed. The fixer would only convert the elseif into an if, which looks valid to me, unless I am missing something.

I'm saying nothing should be touched in the sample by the `no-useless_else`-fixer at all, so no convert of `elseif` as well no matter what.I.e. the fixer is not called `no_useless_else_and_elseif_to_if_splitup_fixer`

>Current implementation requires to enable an option to fix your example, nothing will change for people that don't want to fix it.

With above in mind I'm -1 on adding this to the fixer, including using configuration.
So I'm -1 on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2885#discussion_r124260965 where a RFC is done on the PR and agreed on by for this PR to _always_ apply the new `elseif` fixing, even more the comment is incorrect as `just a next case we didn't spot before` is simply not true, it was left out by design.

GrahamCampbell:
> Adding this empty line would be an assumption that most people would add it, which is not true IMO. 

I think that most people would, actually.

gharlan:
@SpacePossum For me the `elseif` contains an useless `else` in your sample. So it should be removed by a `NoUselessElseFixer`.

src/AbstractNoUselessElseFixer.php
```diff
@@ -0,0 +1,214 @@
...
+        return 25;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    protected function elseIsSuperfluous(Tokens $tokens, $index)
```
keradus: this is method, not property
```diff
-elseIsSuperfluous
+isElseSuperfluous
```

julienfalque: Method renamed to `isSuperfluousElse`.

keradus:
Thank you @julienfalque.

Issue #2062, opened 2016-07-09
DeclareStrictTypesFixer split


keradus:
So far, we have a (risky) fixer that adds/relocate `declare(strict_types=1)`.

I think we should split it.
1. fixer to add `declare` statement if missing (risky)
2. fixer to relocate declaration (non-risky) - here, with option what should be the position (same line with opening tag or next one).

We will be more generic and allow to have better user experience, like allowing to move declaration to same place for all files but not enforcing adding it to every single file. Or allow to place the declaration to place defined by `Extended Coding Style Guide` proposal.

Also, a ref: https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#3-declare-statements-namespace-and-use-declarations

> Each declare statement (e.g. declare(ticks=);) MUST be on its own line.
> When the opening <?php tag is on the first line of the file, it MUST be on its own line with no other statements

keradus:
cc @Seldaek @SpacePossum discussion/help needed

SpacePossum:
Hmmm....
We have been discussing the placement of both the header, strict type declaration and namespace location in various PR's and issues. PSR does help a bit, but also seems incomplete on new features.

Maybe we could take a diff route:
- create a new fixer, something like `start_of_file`-fixer (terrible name, but for the sake of discussion)
- remove the relocate part of the strict type fixer
- remove the relocate part of the header fixer and the "spacing above" part

The new fixer can be configured to order the first parts of a PHP file;
- header
- declare strict (note: must always be the first statement)
- namespace

Together with:
- empty line after open tag
- empty line after/before header
- empty line after/before namespace

This will result in no more conflicts and should give everyone a way to configure what they want.

(for example https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1993#issuecomment-231410650 would be fixed as well)

keradus:
I like the idea. But it looks like a lot of work.
So, one way or another, `DeclareStrictTypesFixer` needs to be split first.

sstok:
Can we at least for now place the `declare(strict_types=1);` on it's own line?
That's also what is proposed in the psr document.

I'm facing a conflict between StyleCI and php-cs-fixer, because php-cs-fixer wants to keep it on the opening tag line, but StyleCI insists on placing it on it's own line.

@GrahamCampbell is StyleCI using a custom fixer for this? because it should not give a different result.

SpacePossum:
[...]

Issue #2846, opened 2017-06-11
ProcessLinterTest fails on Mac


GrahamCampbell:
```
There were 2 failures:

1) PhpCsFixer\Tests\Linter\ProcessLinterTest::testPrepareCommandOnPhpOnWindows with data set #0 ('php', 'foo.php', 'php -l foo.php')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-php -l foo.php
+'php' '-l' 'foo.php'

/Users/graham/GitHub/PHP-CS-Fixer/tests/Linter/ProcessLinterTest.php:63

2) PhpCsFixer\Tests\Linter\ProcessLinterTest::testPrepareCommandOnPhpOnWindows with data set #1 ('C:\Program Files\php\php.exe', 'foo bar\baz.php', '"C:\Program Files\php\php.exe...z.php"')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-"C:\Program Files\php\php.exe" -l "foo bar\baz.php"
+'C:\Program Files\php\php.exe' '-l' 'foo bar\baz.php'

/Users/graham/GitHub/PHP-CS-Fixer/tests/Linter/ProcessLinterTest.php:63
```

GrahamCampbell:
```
vendor/bin/phpunit --version
PHPUnit 5.7.20 by Sebastian Bergmann and contributors.
```

GrahamCampbell:
Well, obviously, the windows tests weren't meant to run, but they did. :/

GrahamCampbell:
Related: https://github.com/sebastianbergmann/phpunit/issues/2697, https://github.com/sebastianbergmann/phpunit/issues/2704.

keradus:
poor mac. we don't support it

GrahamCampbell:
> poor mac. we don't support it

What, you can't be serious. Loads of PHP developers use PHP-CS-Fixer on Mac. More than use Windows these days, I'm pretty sure!

SpacePossum:
we support Mac, we don't support running the internal tests on Mac with a bugged PHPUnit version ;)
anyway, must be fixed @ PHPUnit if I understand correctly.

keradus:
we don't run any tests on mac, for that there is no official support. it execution works, it's good. but as we don't test it, there's no official support

ConnorVG:
[...]

PR #2891, opened 2017-06-30
feature: Introduce `declare_strict_types_position` fixer

aidantwoods:
A start on #2062


SpacePossum:
Thanks for picking this one up, it has been an issue a long time but is not trivial to solve, so help on it is very welcome.
Your PR takes a diff. route than @keradus suggested in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2062, there it was suggested to split up the fixer. I think that suggestion would be more flexible and more powerful.
To summarize the suggested route:
- create a new fixer called something like `declare_strict_type_location`
- `declare_strict_type_location` will have the configuration options similar to the ones you've added here to `declare_strict_type`
- remove the (re)location logic from `declare_strict_type`, it should only insert the statement
- make sure the priorities are correct; `declare_strict_type` goes before `declare_strict_type_location`

In the end this makes the current `declare_strict_type` more simple and the solution still brings the functionality you want. It will also bring a nice way for users to enforce the location of existing declarations without forcing new ones everywhere.

I think your PR is a good start and by moving the new bits around the suggestion of @keradus could be done. WDYT?

aidantwoods:
> [...] I think your PR is a good start and by moving the new bits around the suggestion of @keradus could be done. WDYT?

Sounds good – I wasn't sure if we meant split into config options, or into separate fixers ;p I'll split up that new functionality into separate fixers in that case!

>  It will also bring a nice way for users to enforce the location of existing declarations without forcing new ones everywhere.

It does this already (via default `false` option on the `'relocate_to'`) though it's probably clearer to do this via a separate fixer 😉 

---

Any idea what is going on with the tests – half the CI builds seems to be failing without giving a reason

SpacePossum:
>It does this already (via default false option on the 'relocate_to') though it's probably clearer to do this via a separate fixer

ah yes, you already made it into config :+1:  (I need a coffee I think ;) )

About the test, half fail because only half runs, which makes sense as you're fixer only works on PHP 7 (+) ;)
The utests themselves do not fail either, however Travis is sad because PHPUnit detected the fixer uses deprecated code. 
The deprecated code used is `Token::setContent`, recently we made tokens immutable so this method has been deprecated, but not yet removed as to not break our BC promise.
The alternative is to not modify a token, but to create a new token and replace the old token in the collection. This can be done by using `Tokens::overrideAt` or directly by array access support; `$tokens[$toOverrideIndex] = $myNewToken`.

aidantwoods:
> About the test, half fail because only half runs, which makes sense as you're fixer only works on PHP 7 (+) ;)

Ah yes, this makes sense now 😉 

Is there a proper alternative for `Token::isEmpty`? I've just changed it to a check `Token::getContent` for now, not sure if that was the envisioned work-around.

aidantwoods:
> > It does this already (via default false option on the 'relocate_to') though it's probably clearer to do this via a separate fixer
> 
> ah yes, you already made it into config 👍 (I need a coffee I think ;) )

Are we still wanting separate fixers?

SpacePossum:
I'm :+1: for two fixers, like to hear from @keradus and @julienfalque

aidantwoods:
> I'm 👍 for two fixers, like to hear from @keradus and @julienfalque

I shall await further feedback in that case – I don't really mind either way, current method just had less inertia 😉 

--- 

Cheers for pointing out the deprecation notices on those tests – tests started getting very noisy when I rebased on `master` and ran into #2846 causing a couple failures locally that had nothing to do with my changes!

TomasVotruba:
@aidantwoods Great job with this fixer. I was looking for this feature for a long time myself :+1:

julienfalque:
Sorry for not answering earlier.

I do like the current state of this PR (i.e. a single configurable fixer). However, like @keradus said in #2062, part of what it does is risky and part of it is non-risky. For that reason, I would split it into two fixers too.

aidantwoods:
Ah yes, I hadn't considered that the location fixer was in fact non-risky (and so would be unnecessarily marked as risky).
Splitting into separate fixers makes sense to me! Will try and get it done soon.

aidantwoods:
Fixers are split, have rebased on master.

(you guys work fast here 😜 – a whole class was deprecated and started throwing notices between my commits last week and now!)

src/Fixer/Strict/DeclareStrictTypesPositionFixer.php
```diff
@@ -0,0 +1,194 @@
...
+                new Token('='),
+                new Token([T_LNUMBER, '1']),
+                new Token(')'),
+            ];
+        }
+
+        return $sequence;
+    }
+
+    private function insertSequence(Tokens $tokens, array $sequence = null)
```
aidantwoods: Originally I tried to make this method `protected` in `DeclareStrictTypesFixer`, and then just extend the class so I could reuse this method. Looks like fixers have to be final classes though...

Open to suggestions, as I don't really like this duplication. I could stick it in a trait? Let me know where to put the trait file if so though.

src/Fixer/Strict/DeclareStrictTypesPositionFixer.php
```diff
@@ -0,0 +1,194 @@
...
+    {
+        return new FixerDefinition(
+            'Move the strict type declaration to the configured location. Requires PHP >= 7.0.',
+            [
+                new VersionSpecificCodeSample(
+                    '<?php declare(strict_types=1);',
+                    new VersionSpecification(70000)
+                ),
+            ],
+            null
```
SpacePossum: `null` can be dropped.
Please add a sample with diff. configuration than the default.

src/Fixer/Strict/DeclareStrictTypesPositionFixer.php
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // check if the declaration is already done
+        $searchIndex = $tokens->getNextMeaningfulToken(0);
+        $sequence = $this->getDeclareStrictTypeSequence();
+        $sequenceLocation = $tokens->findSequence(array_merge($sequence, [new Token(';')]), $searchIndex, null, false);
```
SpacePossum: Please find the sequence without the `;` and check for the `;` when the sequence is found (if needed).
Doing the search only once will be a big performance gain.

src/Fixer/Strict/DeclareStrictTypesPositionFixer.php
SpacePossum: I like the idea to reuse the code, but wonder if it worth it. As always it is trade off for DRY vs. code that code maintained (i.e. changed) independently. I'm OK with the duplication here, wonder what others think.
If we leave it here, please change the optional declare of `$sequence` to required (i.e. remove `= null`).
Please add the type hint `@param array<int, Token> $sequence`.

src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -135,36 +131,30 @@ private function fixStrictTypesCasing(Tokens $tokens, array $sequence)
         }
     }
 
-    private function insertSequence(Tokens $tokens)
+    private function insertSequence(Tokens $tokens, array $sequence = null)
```
aidantwoods: I'll await feedback on code reuse for this one.

src/Fixer/Strict/DeclareStrictTypesPositionFixer.php
aidantwoods: I too don't mind the duplication here that much (since it's quite minimal). That said, it's a great candidate for reuse (since the code is doing basically the same thing in both places) – and any bugfixes will have to also be duplicated to both locations at present.

src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -76,21 +77,16 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
-
         $sequence = $this->getDeclareStrictTypeSequence();
         $sequenceLocation = $tokens->findSequence($sequence, $searchIndex, null, false);
-        if (null === $sequenceLocation) {
-            $this->insertSequence($tokens); // declaration not found, insert one
 
-            return;
+        if (null === $searchIndex) {
+            $this->insertSequence($tokens); // declaration not found, insert one
+        } elseif (null === $sequenceLocation) {
```
aidantwoods: D'oh! 😆

tests/Fixtures/Integration/priority/declare_strict_types,single_blank_line_before_namespace.test
```diff
@@ -5,7 +5,8 @@ Integration of fixers: declare_strict_types,single_blank_line_before_namespace.
 --REQUIREMENTS--
 {"php": 70000}
 --EXPECT--
-<?php declare(strict_types=1);
+<?php
+declare(strict_types=1);
```
keradus: why this change ?

README.rst
```diff
@@ -339,6 +339,17 @@ Choose from the list of available rules:
...
   *Risky rule: forcing strict types will stop non strict code from working.*
 
+* **declare_strict_types_position**
+
+  Move the strict type declaration to the configured location. Requires
+  PHP >= 7.0.
+
+  Configuration options:
+
+  - ``relocate_to`` (``'next'``, ``'same'``): whether ```declare(strict_types=1)````` should
```
keradus: too many backticks

aidantwoods: Not my bug?
screen shot 2017-07-29 at 12 52 22

keradus: use single backticks, like https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/f7241f507111ff78204590402ba0c05aa47b250b/src/Fixer/LanguageConstruct/DirConstantFixer.php#L32

tests/Fixtures/Integration/priority/declare_strict_types,single_blank_line_before_namespace.test
aidantwoods: Matches (proposed) PSR-12 recommendation.
https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#25-keywords-and-types

>  [...]If present, each of the blocks below MUST be separated by a single blank line, and MUST NOT contain a blank line.[...]
> * [...]
> * One or more declare statements.

README.rst
aidantwoods: Alright, should we not fix the issue/add a test to ban sequential backticks in descriptions though? ;p

keradus: go ahead and sent a fix. in the meantime, fix to single backticks to not crap the readme

aidantwoods: Roger that

aidantwoods:
For visibility: I just rebased on current master so tests are up to date on this branch.

tests/Fixer/Strict/DeclareStrictTypesFixerTest.php
```diff
@@ -141,11 +155,11 @@ public function provideMessyWhitespacesCases()
     {
         return [
             [
-                "<?php declare(strict_types=1);\r\nphpinfo();",
+                "<?php\r\ndeclare(strict_types=1);\r\n\tphpinfo();",
```
keradus: do not change existing tests that are verifying if the feature works.
only add new tests.

aidantwoods: Is the objection to introducing a newline, or to adding the tab?

aidantwoods: > do not change existing tests that are verifying if the feature works.
> only add new tests.

This test needs changing either way.
1. If we want to match (proposed) PSR-12 by default, we need the add the newline by default.
2. The feature and its tests are broken in release version, the full test case is
   ```php
   [
       "<?php\r\ndeclare(strict_types=1);\r\n\tphpinfo();",
       "<?php\r\n\tphpinfo();",
   ],
   ```
   The `\t` should have never been removed.

aidantwoods: Previous version of the fixer doing tab removal is probably related to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2058

keradus: > If we want to match (proposed) PSR-12 by default, we need the add the newline by default.

who told you that? you must not change default behaviour of fixer, it will fail on any integration usage that are using this rule.
if you want to change behaviour of fixer, the only way is to introduce option for that leaving default behaviour the same

aidantwoods: > you must not change default behaviour of fixer, it will fail on any integration usage that are using this rule.

Oh, I didn't realise that. Should I bug this then?
https://travis-ci.org/Parsemd/Parsemd/builds/249843237
vs.
https://travis-ci.org/Parsemd/Parsemd/builds/250188863
(update killed the cron)

I'll change the default on the position fixer to `same` in that case.

aidantwoods: > it will fail on any integration usage that are using this rule.

Actually, having a think about it: it shouldn't fail with integration usage. Not unless the integration is using the fixer on code that has `declare(strict_types=1)` absent. On code that has already been fixed (or that where `declare(strict_types=1)` is already present), the `DeclareStrictTypesFixer` won't touch it. The repositioning fixer doesn't run by default anyway, so can probably leave the default as is?

Only thing that has changed RE this is that iff code lacks `declare(strict_types=1)` then it'll insert it on the next line (instead of on the same one).

keradus: > Oh, I didn't realise that. Should I bug this then?

no, i have no idea what is `dev-feature/braces/position_after_return_type_hint` version you had installed.
other thing - assume it was regular 2.x release. If we were suppose to fix sth, but for some reason we didn't fix it, and with newer release we catch it and finally fix it - it's expected bugfix. if we claim to fix in one way (eg "same line"), and then out of the blue redefine the rule to fix code another way (eg "new line"), then this is not bugfix but different rule under same name. this is not ok unless configured.

> I'll change the default on the position fixer to same in that case.

Position fixer is new, so it could have any default, as long as it's not conflicting with another rule.

----------------------

thing here. existing fixer, with default config, shall work same way as on master. that's it.

aidantwoods: > if we claim to fix in one way (eg "same line"), and then out of the blue redefine the rule to fix code another way (eg "new line")

The only thing the current fixer says is
> Force strict types declaration in all files.

Not sure that really makes any guarantees about position, but nevertheless I'm happy to change to insert on same line – I was simply unaware of the compatibility requirement and figured this was a better option for working with the proposed PSR standard concerning it.

I shall change the insertion fixer to insert on a new line, and the positioning fixer will stay as is.

---

> > Should I bug this then?
> 
> no, i have no idea what is `dev-feature/braces/position_after_return_type_hint` version you had installed.

`dev-feature/braces/position_after_return_type_hint` is my branch based on `master` with two unrelated commits. The cron failed when I rebased my commits on that branch (so something at your end changed).
As for whether it's still like that, I'd have to check. I'd also have to check what it did before. Looks as if it is a BC-break if this is indeed a change from a release version, per what has been laid out here (in-fact maybe even worse so because it modifies code that has already gone through the fixer).

Happy to look into it at some point, just trying to give a heads up on a possible bug here really, since this seems to be an important consideration?

keradus: > The only thing the current fixer says is
> > Force strict types declaration in all files.

No, fixer does also provide examples (available only on CLI level). So for all the ppl who were into given style, changing it out of the box is not acceptable.
That's why we have tests. if you need to modify tests instead of only adding new tests, you change specification. Sometimes it's a bugfix, sometimes it's BC breaker.

------------

> I shall change the insertion fixer to insert on the same line

if you want to change behaviour, simply add option for that.

> the positioning fixer will stay as is.

does new fixer change sth that is also changing by current fixer? (not saying it's bad, just asking to say how to handle it if needed)

---------------

> Happy to look into it at some point, just trying to give a heads up on a possible bug here really, since this seems to be an important consideration?

if you want to, do it in separated issue, not this PR. what is needed is version before, version after, minimal code sample and rule that changed output.

TomasVotruba:
@aidantwoods Was there something that stopped you from finishing this?

aidantwoods:
@TomasVotruba 

> Was there something that stopped you from finishing this?

If I recall I think I was waiting for feedback on some code duplication i.e. whether or not to keep it, or move into a trait (if the latter, where should the trait go in the folder structure – doesn't seem to be precedent for traits in this repo as I remember), and classes don't seem to be allowed to be non-final (so cannot inherit). Duplication isn't too bad, but if there is a bug in the code you'd need to remember to fix both code locations. ([duplication site here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2891#discussion_r126226004))

There's some more I can do without that feedback though, so I'll try pick up on this again soon regardless.

TomasVotruba:
@aidantwoods Ok, thank your for finding time to this reply.

garbast:
Can we have this merged please?

I would love to be able to enforce the position and format of declare in case the declare is present and not have them added to all php files.

README.rst
SpacePossum: test added here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4844/files#diff-0ae0077906c99aaf35085e05c46cefd3R352

aidantwoods:
I'm happy to do some more work on this. It's (clearly) been a while since these changes where proposed and I'm haven't yet had a look at the current codebase for comparison. Perhaps someone more involved with this project could lend a hand in migrating this PR if necessary, or perhaps have a quick glance to offer some initial pointers?

shakaran:
@aidantwoods would you mind rebase this PR against master, so @keradus and @SpacePossum could check the remaining things to get this  merge in next release? Thanks for all your effort

aidantwoods:
@shakaran unfortunately I think rebasing is probably going to be non-trivial (given more than 4 years have passed since the initial PR). If anyone who currently works on the project is willing to lend a hand I'm happy to help too, but it seems interest has waned.

kubawerlos:
@aidantwoods what is the problem? If only `README.rst` then remove it and run `php dev-tools/doc.php` to generate fresh docs.

shakaran:
> @shakaran unfortunately I think rebasing is probably going to be non-trivial (given more than 4 years have passed since the initial PR). If anyone who currently works on the project is willing to lend a hand I'm happy to help too, but it seems interest has waned.

Since I cannot commit to your branch in this PR, I fork and make the changes in my own account, it is under

https://github.com/shakaran/PHP-CS-Fixer/tree/feature/DeclareStrictTypesFixer-split

If you can cherry pick my commits:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/d318f5fc5ba3c3ae8d58161564aafb2e1bada410

https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/380b040f11aaee372a2889f7a6b4ef471883d60b

https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/380b040f11aaee372a2889f7a6b4ef471883d60b

https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/777444074e36c174cdd09172c710dfeace7d499f

https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/51033fce507a695e5a0502b379fcc4ccaef08ef5

That update the docs as @keradus  requested, and the intefaces, but the logic should be checked since not sure if all is ok, but at least you have a base to update the remaning changes and hopefully we get this merged soon ;)

Basically are only 4 files to update:

src/Fixer/Strict/DeclareStrictTypesFixer.php
tests/Fixer/Strict/DeclareStrictTypesFixerTest.php
tests/Fixtures/Integration/priority/declare_strict_types,declare_equal_normalize.test
tests/Fixtures/Integration/priority/declare_strict_types,single_blank_line_before_namespace.test

Let me know if I can help more, I am just new with the fixers, but I am trying to learn and do my best here.

PR #2893, opened 2017-06-30
Support OSX

### LINKED ISSUE #2892: CI - Add running CI tests on Mac
* Travis supports testing on Mac: https://docs.travis-ci.com/user/osx-ci-environment/

* Only two tests don't pass: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2846

* These will probably be fixed upstream: https://github.com/sebastianbergmann/phpunit/issues/2697

* No Mac support is 😢

aidantwoods:
Add support for OSX by testing it in Travis CI (merge probably depends on an update to phpunit, see: https://github.com/sebastianbergmann/phpunit/issues/2697)

If anyone can see a nicer way of doing this let me know – testing various configs in travis isn't my favourite activity ;p

---

Closes #2892

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,10 +53,41 @@ jobs:
                 - vendor/bin/phpunit --verbose || travis_terminate 1
                 - php php-cs-fixer --diff --dry-run -v fix
 
+        - &OSX_TEST_JOB
+            stage: Test
+            os: osx
+            language: generic
+            before_install:
+                # Set up OSX environment
+                - travis_retry brew update
+                - travis_retry brew tap homebrew/homebrew-php
+                - travis_retry brew install $phpver || brew outdated $phpver || brew upgrade $phpver
+                - brew link $phpver
+                - travis_retry wget https://raw.githubusercontent.com/composer/getcomposer.org/master/web/installer -O - -q | php -- --install-dir=/usr/local/bin && mv /usr/local/bin/composer.phar /usr/local/bin/composer
+                - travis_retry composer self-update
+
+                # turn off XDebug
+                - phpenv config-rm xdebug.ini || return 0
+
+                # Composer: boost installation
+                - composer global show -ND 2>&1 | grep "hirak/prestissimo" || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
+            install:
+                - travis_retry composer update $DEFAULT_COMPOSER_FLAGS $COMPOSER_FLAGS
+                - composer info -D | sort
+            script:
+                - vendor/bin/phpunit --verbose || travis_terminate 1
+                - php php-cs-fixer --diff --dry-run -v fix
+
+            env: phpver=php70
+
         -
             <<: *STANDARD_TEST_JOB
             php: 5.6
             env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+
+        -
+            <<: *OSX_TEST_JOB
+            env: phpver=php56 COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
 
         -
             <<: *STANDARD_TEST_JOB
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -52,12 +53,43 @@ jobs:
...
+            os: osx
+            language: generic
+            before_install:
+                # Set up OSX environment
+                - travis_retry brew update
+                - travis_retry brew tap homebrew/homebrew-php
+                - travis_retry brew install $phpver || brew outdated $phpver || brew upgrade $phpver
+                - brew link $phpver
+                - travis_retry wget https://raw.githubusercontent.com/composer/getcomposer.org/master/web/installer -O - -q | php -- --install-dir=/usr/local/bin && mv /usr/local/bin/composer.phar /usr/local/bin/composer
+                - travis_retry composer self-update
```
GrahamCampbell: Why is this needed?

keradus:
Supported env is not a feature. header adjusted.

.travis.yml
```diff
@@ -52,12 +53,43 @@ jobs:
...
+
+                # turn off XDebug
+                - phpenv config-rm xdebug.ini || return 0
+
+                # Composer: boost installation
+                - composer global show -ND 2>&1 | grep "hirak/prestissimo" || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
```
GrahamCampbell: Is this done for the linux builds?

SpacePossum: it is, check the rest of the file ;)

GrahamCampbell: 👍

.travis.yml
aidantwoods: Travis don't include anything useful for PHP dev in their OSX builds (in-fact specifying `language: php` flat out breaks things, so idk) – let me know if there's a better solution though!

keradus:
big :-1: for extending matrix that much, we don't need 3 new jobs just to test osx for sure.

.travis.yml
GrahamCampbell: I am specifically referring to the self-update. You already installed the latest version.

SpacePossum:
Looks promising :)
One the two expected tests to fail, an explicit test on `PHP_OS` or `PHP_OS_FAMILY` and skip when needed would be OK IMHO. It's not ideal, but it only effects two tests so worth it to get this going.

aidantwoods:
> big 👎 for extending matrix that much, we don't need 3 new jobs just to test osx for sure.

How many tests would you like? Just tried to mirror the linux tests so far

keradus:
it increase workload a lot for Travis...
what about taking other apporach?
current state is Travis for Linux, AppVeyor for for Windows... maybe CircleCI for OSX ?

GrahamCampbell:
Maybe circle for everything tbh. Travis is very slow compared to circle.

aidantwoods:
> it increase workload a lot for Travis...
> what about taking other apporach?
> current state is Travis for Linux, AppVeyor for for Windows... maybe CircleCI for OSX ?

Perhaps, yes. A separate build stage might be another option too?

RE the extra workload, shouldn't be too noticeable with concurrent builds (I hope). I think Travis may be having a slow period – up to 20mins on this linux job... https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/248866481

OSX hasn't even started yet.

keradus:
separate stage would be executed after current testing stage, making all executes longer.
concurrency works inside same stage, not across different stages.

laso, consurrency works up to 4 jobs for free plans

julienfalque:
I would say the current slowliness is more a performance issue in the codebase than Travis being slow. This is a pure guess and has to be verified though.

Anyway, I think it would be better to target 2.2.

.travis.yml
aidantwoods: Ah, yes you're right there – I'll remove that!

keradus:
the failing of utests is probably on higher branch, is that right @aidantwoods  ?

aidantwoods:
> separate stage would be executed after current testing stage, making all executes longer.
concurrency works inside same stage, not across different stages.

Indeed, so we'd be looking at an extra 2-3mins if all the OSX builds start together on a separate stage (though compared to 20mins for that `php56` build, not sure how significant that is?)

aidantwoods:
> the failing of utests is probably on higher branch, is that right @aidantwoods ?

The `ProcessLinterTest` failures are present in `2.2` too

aidantwoods:
https://github.com/aidantwoods/PHP-CS-Fixer/commit/287a853b52a7639af621fa67ec752bb0f4808b99 adds two OSX builds onto `2.2`'s Travis, incase we wanted to rebase this onto `2.2`.

keradus:
@aidantwoods, could you execute tests of 2.2 branch on mac and send a fix as a separate PR please?

keradus:
replaced by #2932

PR #2895, opened 2017-07-02
ProjectCodeTest - check that classes in no-tests exception exist

keradus:


tests/AutoReview/ProjectCodeTest.php
```diff
@@ -56,6 +56,12 @@
         'PhpCsFixer\Tokenizer\Transformers',
     );
 
+    public function testThasClassesWithoutTestsVarIsProper()
```
SpacePossum: :+1: for idea, little typo in method name

keradus: done

PR #2896, opened 2017-07-02
Move testing related classes from src to tests

keradus:
tests only classes shall not be exposed in src dir, that way, due to that issue, currently we do have phpunit-related classes as non-dev dependency


SpacePossum:
Agreed in concept, just wondering how we are going to deal with exposing the integration test cases which we do ship as API. If we move all test related deps. to `dev deps` will the `phar` release still contain usable integration tests logic as we exclude the `dev.` and the `tests` folder from the release?

keradus:
currently (on 2.2/2.3/master branches), we do not expose `src/Test` folder (that was moved to `tests/Test` and have deprecated (on 2.4) aliases in `src/Test` now) in phar file.

if installed via composer, `src/Test` will still be visible, yet deprecated on 2.4

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -339,6 +339,7 @@ private function getTestClasses()
             ->in(__DIR__.'/..')
             ->exclude(array(
                 'Fixtures',
+                'Test',
```
julienfalque: Why excluding this directory?

SpacePossum:
`AccessibleObject` is a nice re-usable class so it might make sense to keep it as support end point

keradus:
> Why not moving `AccessibleObject` as well?

I was thinking about it for a while, but ultimately I do want to remove it, not just hide it.
Adding deprecation is nice, but will be separate PR, as we need to solve deprecations first, so stop using it.

tests/AutoReview/ProjectCodeTest.php
keradus: this is used for data providing to determine that `tests/` classes are internal and final. it's not the case for `tests/Test`, as we expose it.
Actually, we also have `tests/Test/SthSthTest.php`, which shall be internal. I removed `Test` from excluding and enforce file to have `Test` suffix few lines above

julienfalque: Isn't the point of moving those files from `src` to `tests` to *not* expose them anymore, and thus mark them all as `@internal`?

keradus: hm... I'm not sure do we want to hide it, some devs relys on them, and they are written in very generic form.

julienfalque: I believe they were in `src` for that reason. If you still want to support them for external usage, then moving them outside `src` does not make sense to me.

keradus: having them in `src` requires you to have whole dev-deps in `require` instead of `dev-require`. we face it with GeckoPackages and PHPUnit packages. that's what composer requirements checker complains about. so current pathing is broken

rybakit:
@keradus Looks like this change breaks my builds. I install packages with the `--prefer-dist` composer option and get 
```
Could not scan for classes inside ".../php-cs-fixer/tests/Test/Assert/AssertTokensTrait.php" 
which does not appear to be a file nor a folder
```
but apparently there is no `tests` directory because of the `.gitattributes` rules.

julienfalque:
Should be fixed with #2899.

TomasVotruba:
How should I test own fixers now?

SpacePossum:
>How should I test own fixers now?

The tests are still available, so you can keep on using those for now.
The difference with before this PR are:
- the tests are in a different namespace (so you need to update the `use` statement in your own tests)
- you need the `dev` dependencies of this project in your project

We're are not sure yet if are going to support the (re)use of the tests in v3.x, although I'm in favor of that. The discussion is about how we could do this without forcing `dev` dependencies (like PHPUnit and GeckoPHPUnit) on the non-dev releases.

TomasVotruba:
> you need the dev dependencies of this project in your project

How can I do that for only this package? 

This would complicate using own fixers a lot. It took me while to figure out, that I need to add "GeckoPHPUnit" manually.

Is that intention?

SpacePossum:
>How can I do that for only this package? I don't want to install dev dependencies of all my dependencies.

not sure if composer support that in a nice way, why would you need the `tests` logic in a non-dev environment?

>What would be advantage for end-user of this?

The typical end-user of the project are people that don't use the test cases (either through `phar` or by `composer` install), the advantage for them is not having to install the dependencies of the tests.

TomasVotruba:
> not sure if composer support that in a nice way, why would you need the tests logic in a non-dev environment?

I don't need tests logic of my dependencies in any environment. I need only `/src`.

> The typical end-user of the project are people that don't use the test cases (either through phar or by composer install), the advantage for them is not having to install the dependencies of the tests.

Having or not having such dependencies doesn't cause any problems. It's nice to have. This force me to write own class which would test Fixer, so much pain for anyone who wants to write own fixer. This class will never will be as good as original.

That contradics option to write own fixer. This is ambiguous claim - "you can add own fixer, but you can't test it".

keradus:
1. Removing (now, deprecating) classes from src that are directly relying on test deps is must have.
2. if you have installed 2.1.0 you have `Abstract...TestCase`, but you cannot use it unless you installed phpunit+gecko manually, that's the issue
3. it doesn't force you to do anything, tests classes are still there, only deprecated
4. it's not claiming you can't test custom, own rule. to use it under PHP CS Fixer you need to follow provided interface, but you can test it however you want to
5. no need for bolding
6. new package possibility was discussed already 10 days ago, it may happen

Once more, to focus on what's important, provided test-related classes were buggy. This is first step to fix the situation, and that step itself is not breaking anything.

TomasVotruba:
> it doesn't force you to do anything, tests classes are still there, only deprecated

It will and there is no suggested solution. Deprecating without suggestion a solution will break things randomly. I won't remember this in couple of months when I run `composer update` and my tests will break.

I care because I met few random breaks in this package during last weeks and it took unncesseary hours of my life.

keradus:
> I won't remember this in couple of months when I run composer update and my tests will break.

it could happen only with major release, which... is major, as it breaks BC, right? always read upgrade/changelog while upgrading majors.

> I had no issue with them at all and I use them quite often.

bug was described in this pr and raised in other, dedicated issues.

TomasVotruba:
We are heading in unrelated topics. The point is: there is no replacement. That's all I wanted to say.

keradus:
Cool. stick to current solution while waiting for new one.

still valid:  
> > I care because I met few random breaks in this package during last weeks and it took unncesseary hours of my life.

> then simply report the break.
> deprecation notice is not a break.
> what were a random breaks?

PR #2900, opened 2017-07-04
VoidReturnFixer - handle functions containing anonymous functions/classes

bendavies:
VoidReturnFixer does not work if the function contains a closure.
Not had the chance to look at fixing it yet, so here's a failing a test in case someone wants to tackle it.

cc @mrmark


julienfalque:
What do you mean _it does not work_? Does it crash? Or does it make unexpected changes in your files? Could you provide more details like input code, your config, etc.?

keradus:
Thanks for report @bendavies !
Any particular reason why to use not-released version ?

bendavies:
@keradus no problem. because i wanted to use `VoidReturnFixer` in my project :)

keradus:
@bendavies , as you allowed us to push to your branch, I pushed the fix directly to it.

bendavies:
@keradus that appears to be the default 😃. Thanks for the fix

keradus:
Thank you @bendavies.

bendavies:
@keradus thank you!

PR #2903, opened 2017-07-04
[WIP] BracesFixer: Add position_after_return_type_hint option

aidantwoods:
Add option to determine whether the opening brace following a return type hint should be placed on "next" or "same" line.

See text in latter part of this comment for what this implements: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2684#issuecomment-297846431

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -282,6 +282,9 @@ Choose from the list of available rules:
     the opening brace should be placed on "next" or "same" line after
     classy constructs (non-anonymous classes, interfaces, traits, methods
     and non-lambda functions); defaults to ``'next'``
+  - ``position_after_return_type_hint`` (``'next'``, ``'same'``): whether the opening
+    brace following a return type hint should be placed on "next" or "same"
+    line; defaults to ``'same'``
 
 * **cast_spaces** [@Symfony]
 
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -485,7 +485,7 @@ class Foo
                         $tokens->ensureWhitespaceAtIndex(
                             $startBraceIndex - 1,
                             1,
-                            self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && ':' === $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->getContent() ?
+                            self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->equals([CT::T_TYPE_COLON, ':']) ?
                                 $this->whitespacesConfig->getLineEnding().$indent
                                 : ' '
                         );
@@ -498,7 +498,7 @@ class Foo
                             || $token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)
                         )
                         && !$tokens[$tokens->getPrevNonWhitespace($startBraceIndex)]->isComment()
-                        && !(self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && ':' === $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->getContent())
+                        && !(self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->equals([CT::T_TYPE_COLON, ':']))
                     ) {
                         $ensuredWhitespace = ' ';
                     } else {
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -172,6 +172,10 @@ class Foo
                 ->setAllowedValues([self::LINE_NEXT, self::LINE_SAME])
                 ->setDefault(self::LINE_SAME)
                 ->getOption(),
+            (new FixerOptionBuilder('position_after_return_type_hint', 'whether the opening brace following a return type hint should be placed on "next" or "same" line.'))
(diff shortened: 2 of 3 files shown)


src/Fixer/Basic/BracesFixer.php
```diff
@@ -172,6 +172,10 @@ protected function createConfigurationDefinition()
                 ->setAllowedValues([self::LINE_NEXT, self::LINE_SAME])
                 ->setDefault(self::LINE_SAME)
                 ->getOption(),
+            (new FixerOptionBuilder('position_after_multiline_argument_function_return_type_hint', 'whether the return type hint should be placed on "next" or "same" line after a function that has a multiline body of arguments.'))
```
aidantwoods: Open to suggestions on this option name 😉

keradus:
Travis is sad

aidantwoods:
> Travis is sad

hehe, my fault on not flagging the 7.0 tests – one sec

SpacePossum:
please rebase :)

Could you add a view more test cases for each configuration option with code containing "messy" comments?
I would like to see how the fixer would behave, something like;
```php
<?php
        function foo(
            $bar,
            $boo
        ): void# A
# B
{# C
# D
}# E
# F
```

aidantwoods:
> please rebase :)

Onto current master I assume? Or where? ;p

> Could you add a view more test cases for each configuration option with code containing "messy" comments?

Sure! I wont get a chance until later today, but I'll add a few when I get some time!

SpacePossum:
on master indeed ;)
(btw. we should check priority with the ReturnTypeDeclarationFixer)

src/Fixer/Basic/BracesFixer.php
```diff
@@ -488,6 +498,7 @@ function ($item) {
                             || $token->isGivenKind($classyTokens) && !$tokensAnalyzer->isAnonymousClass($index)
                         )
                         && !$tokens[$tokens->getPrevNonWhitespace($startBraceIndex)]->isComment()
+                        && !(self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && ':' === $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->getContent())
```
julienfalque: Please use `->equals(':')` instead of `->getContent() === ':'`.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -478,7 +482,13 @@ function ($item) {
 
                 if (!$isAnonymousClass && $prevToken->isWhitespace() && false !== strpos($prevToken->getContent(), "\n")) {
                     if (!$tokens[$startBraceIndex - 2]->isComment()) {
-                        $tokens->ensureWhitespaceAtIndex($startBraceIndex - 1, 1, ' ');
+                        $tokens->ensureWhitespaceAtIndex(
+                            $startBraceIndex - 1,
+                            1,
+                            self::LINE_NEXT === $this->configuration['position_after_return_type_hint'] && $token->isGivenKind(T_FUNCTION) && ':' === $tokens[$tokens->getNextNonWhitespace($closingParenthesisIndex)]->getContent() ?
```
julienfalque: Please use `->equals(':')` instead of `->getContent() === ':'`.

tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -4671,4 +4672,245 @@ public function testDoWhileLoopInsideAnIfWithoutBrackets($expected, $input = nul
...
+
+    public function provideBraceNextLineReturnTypeHint()
+    {
+        return [
+            [
+                '<?php
+        function foo($bar): void
+        {
+        }',
+                null,
```
julienfalque: Please replace `null` with actual input code that requires changes (i.e. with opening brace on same line as `void`). The result code will then be checked again to make sure it is not changed (like current test case).

If possible, add input code to other test cases that don't have one as well.

README.rst
```diff
@@ -270,6 +270,9 @@ Choose from the list of available rules:
     the opening brace should be placed on "next" or "same" line after
     classy constructs (non-anonymous classes, interfaces, traits, methods
     and non-lambda functions); defaults to ``'next'``
+  - ``position_after_return_type_hint`` (``'next'``, ``'same'``): whether the opening
```
julienfalque: Does `position_after_return_type_hint` option always override `position_after_functions_and_oop_constructs`? Or when arguments list is multiline only?

src/Fixer/Basic/BracesFixer.php
aidantwoods: I'll get the hang of all the standard methods eventually 😉

tests/Fixer/Basic/BracesFixerTest.php
aidantwoods: > The result code will then be checked again to make sure it is not changed (like current test case).

Noted! I'll add a few variations in that case.

README.rst
aidantwoods: I plan to adjust to fixer to override when argument list is multiline only, and honour the `position_after_functions_and_oop_constructs` setting otherwise (I think that makes the most sense). The override behaviour at present is asymmetric and I'm not happy with it.

src/Fixer/Basic/BracesFixer.php
aidantwoods: `->equals(':')` doesn't appear to be doing the same thing... possibly something to do with `->isArray` setting?

keradus: then you encountered custom token.
PHP's native `token_name` is not aware about custom tokens.

We do transform of some tokens to different them here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/master/src/Tokenizer/Transformer
like what is `class` token? is it about defining new class  (`class Foo {}`) or getting it's name (`Foo::class`) ? you don't know, that's why we introduced new, custom token for later.

Use `$token->getName()` to get name of token (regardless it's custom or native), or `$token->toArray()`/`$token->toJson()` to dump the token properties

julienfalque:
This PR didn't get much activity in several years, closing.

PR #2910, opened 2017-07-08
PhpdocAnnotationWithoutDotFixer - Restrict lowercasing

### LINKED ISSUE #2906: phpdoc_annotation_without_dot incorrectly lowercases the first letter of an acronym
For instance:

```
    * @param string $ipAddress IPv4 or IPv6 address as a string
```
is turned into:

```
    * @param string $ipAddress iPv4 or IPv6 address as a string
```

Note the "iPv4".

oschwald:
In particular, if the first two letters of the comment are uppercase, we
no longer lowercase the first.

Fixes #2906.

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php
@@ -90,9 +90,14 @@ function foo ($bar) {}
                 $optionalTypeRegEx = $annotation->supportTypes()
                     ? sprintf('(?:%s\s+(?:\$\w+\s+)?)?', preg_quote(implode('|', $annotation->getTypes())))
                     : '';
-                $content = preg_replace_callback('/^(\s*\*\s*@\w+\s+'.$optionalTypeRegEx.')(.*)$/', function (array $matches) {
-                    return $matches[1].lcfirst($matches[2]);
-                }, $startLine->getContent(), 1);
+                $content = preg_replace_callback(
+                    '/^(\s*\*\s*@\w+\s+'.$optionalTypeRegEx.')(\p{Lu}?(?=\P{Lu}))(.*)$/',
+                    function (array $matches) {
+                        return $matches[1].strtolower($matches[2]).$matches[3];
+                    },
+                    $startLine->getContent(),
+                    1
+                );
                 $startLine->setContent($content);
             }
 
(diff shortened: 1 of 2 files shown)


tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
```diff
@@ -45,6 +45,9 @@ public function provideFixCases()
      * Description.
      *
      * @param string|null $str   some string
+     * @param string $ip         IPv4 is not lowercased
```
keradus: I wouldn't lowercase non-words, simple check would be that 2 first chars are letters, so no lowercasing for `T34` and `S§`. Please add tests

oschwald: Thanks for the feedback. I updated it.

src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php
```diff
@@ -90,9 +90,14 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 $optionalTypeRegEx = $annotation->supportTypes()
                     ? sprintf('(?:%s\s+(?:\$\w+\s+)?)?', preg_quote(implode('|', $annotation->getTypes())))
                     : '';
-                $content = preg_replace_callback('/^(\s*\*\s*@\w+\s+'.$optionalTypeRegEx.')(.*)$/', function (array $matches) {
-                    return $matches[1].lcfirst($matches[2]);
-                }, $startLine->getContent(), 1);
+                $content = preg_replace_callback(
+                    '/^(\s*\*\s*@\w+\s+'.$optionalTypeRegEx.')(\p{Lu}?(?=\p{Ll}|\p{Zs}))(.*)$/',
+                    function (array $matches) {
+                        return $matches[1].strtolower($matches[2]).$matches[3];
```
SpacePossum: maybe it is better to test if `mb_strtolower` is available and use it if it is? (not sure so like to hear other opinions on this and I know it is a bit of feature creep in this PR)

oschwald: That makes sense, although, to do it right, we would likely want to detect the encoding of the file. Grepping the source, it doesn't appear this is done outside of the cache. It would probably make the most sense for this to be done once per file outside of any particular fixer.

keradus:
Thank you @oschwald.

SpacePossum:
thanks @oschwald :+1:   :)

PR #2916, opened 2017-07-12
FixerFactory - drop return in sortFixers(), never used

TomasVotruba:


SpacePossum:
good catch :)   :+1:

keradus:
Thank you @TomasVotruba.

TomasVotruba:
You're welcome

Issue #2923, opened 2017-07-19
$token->isChanged question


TomasVotruba:
PHP 7.1.6

I [detected changed line to report it](https://github.com/Symplify/Symplify/blob/41dc8391d855f12cbc2acc93a7236488fcb5825c/packages/EasyCodingStandard/packages/FixerRunner/src/Application/FixerFileProcessor.php#L121-L132
) to user.

### 2.3.2

This works:

```php
private function detectChangedLineFromTokens(Tokens $tokens): int
{
    $line = 0;
    foreach ($tokens as $token) {
        $line += substr_count($token->getContent(), PHP_EOL);
        if ($token->isChanged()) {
            return ++$line;
        }
    }
    return $line;
}
```

*[See Travis Line](https://travis-ci.org/Symplify/Symplify/builds/255250882#L377)*

### 2.3.3

No token is `$token->isChanged()` true anymore. Not sure why.

*[See Travis line](https://travis-ci.org/Symplify/Symplify/builds/255252177#L396)*

Somewhere between [2.3.2 and 2.3.3 a Token.php behaviors was changed](https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/v2.3.2...v2.3.3#diff-b853b43b9aa6d26771db195ab8a079c8).

I spent 30 minutes looking for it but I'm lost and not sure where it came from.

keradus:
> $token->isChanged() broke between 2.3.2 and 2.3.3

`$token->isChanged()` was not broken. it works like it should.
this method is providing information if given token instance was somehow changed.
it's `detectChangedLineFromTokens` is failing in some cases. if one is not changing a token, but replacing a token (with `$tokens[$index] = new Token(...)`), then your method will fail to perform detection, as token instance under `$index` was never modified, but whole collection was.

in 2.3.2->2.3.3 release we changed `$tokens[$index]->override...` to `$tokens[$index] = new Token...` in few places. but later form was used before 2.3.3 in some places as well, thus your method was not working all the time properly in lower version as well

julienfalque:
Should we deprecate the `isChanged()` method?

keradus:
it wasn't originally deprecated as for `->setContent` we do have future-ready interface (tokens[index] = ...)
for `isChanged` itself, it gonna just be removed. but yeah, probably we could.

TomasVotruba:
I understand changes. What is the solution to this then?

SpacePossum:
[...]

PR #2924, opened 2017-07-19
Add missing Token deprecations

julienfalque:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2923#issuecomment-316428641.


keradus:
please add `@deprecated` annotation where it's missing (clearChanged, isChanged, setContent)

julienfalque:
Done.

Now I'm wondering. Introducing deprecations requires a minor version bump, but this one should have been introduced in 2.4 with other deprecations in the same class. Should I change the target?

SpacePossum:
>requires a minor version bump

Strictly it doesn't, although this might be how SF implements semver.
Master target now will mean this will in 2.4.x I guess, so all good to me :)

julienfalque:
Bumping minor version when introducing a deprecation *is* a [Semver requirement](http://semver.org/#spec-item-7):
> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated.

SpacePossum:
sorry about that :/

but what why? so if you forgot a deprecation you're not allowed to patch it? semver is weird sometimes

julienfalque:
That's why I am wondering: should we follow semver strictly and target 2.5 or merge it as a fix for a 2.4 oversight? I would go for 2.4.

src/Tokenizer/Token.php
```diff
@@ -339,9 +341,13 @@ public function isCast()
      * Check if token was changed.
      *
      * @return bool
+     *
+     * @deprecated since 2.5
```
keradus: 2.4

keradus:
2.4 bugfix

src/Tokenizer/Token.php
julienfalque: Done.

SpacePossum:
we don't want to trigger warning on all those methods?

keradus:
is it missing somewhere ?

SpacePossum:
GH UI is playing mind tricks :+1:

SpacePossum:
please rebase @julienfalque

keradus:
@fabpot, we are experiencing weird behaviour of fabbot.io service.

In `.php_cs` file, we are asking for using `'array_syntax' => array('syntax' => 'short'),` ruleset, yet we get complains that we shall use long form. Could you take a look please?

julienfalque:
Debug tests reverted. Relevant config file that is not correctly taken into account by fabbot.io can be viewed here: https://github.com/julienfalque/PHP-CS-Fixer/blob/e0f3841c5663652e6a371da3acf2adc68b75a42c/.php_cs

keradus:
or directly in PR history commit:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2924/commits/e0f3841c5663652e6a371da3acf2adc68b75a42c

together with fabbot report:
http://fabbot.io/report/FriendsOfPHP/PHP-CS-Fixer/2924/e0f3841c5663652e6a371da3acf2adc68b75a42c

keradus:
Thank you @julienfalque.

PR #2925, opened 2017-07-20
Improve CI integration suggestion

### LINKED ISSUE #2908: Wrong CI integration on config change
Currently, suggested integration looks like:

```
$ IFS=$'\n'; COMMIT_SCA_FILES=($(git diff --name-only --diff-filter=ACMRTUXB "${COMMIT_RANGE}")); unset IFS
$ vendor/bin/php-cs-fixer fix --config=.php_cs.dist -v --dry-run --stop-on-violation --using-cache=no --path-mode=intersection -- "${COMMIT_SCA_FILES[@]}"
```

but with that, if one change `.php_cs.dist` file, we shall execute fixing on all files (as there could be new rule configured)
probably we need to change `COMMIT_SCA_FILES` to dot if `.php_cs.dist` was modified.

julienfalque:
The new suggestion fixes only changed files except when `.php_cs.dist` or `composer.lock` is part of the list. In such case, all files defined by the config are fixed because the list of rules to apply or the way they are applied may have changed.

I tested it manually and it worked fine, especially with filenames that contain spaces. I'm not a bash expert though, suggestions and corrections are welcome :)

Closes #2908.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1226,7 +1226,7 @@ Then, add the following command to your CI:
 .. code-block:: bash
 
     $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
-    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs\.dist|composer\.lock)$"; then IFS=$'\n' EXTRA_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]}); fi
+    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs(\.dist)?|composer\.lock)$"; then IFS=$'\n' EXTRA_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]}); fi
     $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist -v --dry-run --stop-on-violation --using-cache=no "${EXTRA_ARGS[@]}"
 
 Where ``$COMMIT_RANGE`` is your range of commits, eg ``$TRAVIS_COMMIT_RANGE`` or ``HEAD~..HEAD``.
--- a/README.rst
+++ b/README.rst
@@ -1225,8 +1225,11 @@ Then, add the following command to your CI:
 
 .. code-block:: bash
 
-    $ IFS=$'\n'; CHANGED_FILES=($(git diff --name-only --diff-filter=ACMRTUXB "${COMMIT_RANGE}")); unset IFS
-    $ EXTRA_ARGS=('--path-mode=intersection' '--' "${CHANGED_FILES[@]}")
+    $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
+    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs\.dist|composer\.lock)$"
+    $ then
+    $     IFS=$'\n' EXTRA_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]})
+    $ fi
     $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist -v --dry-run --stop-on-violation --using-cache=no "${EXTRA_ARGS[@]}"
 
 Where ``$COMMIT_RANGE`` is your range of commits, eg ``$TRAVIS_COMMIT_RANGE`` or ``HEAD~..HEAD``.
--- a/README.rst
+++ b/README.rst
@@ -1226,10 +1226,7 @@ Then, add the following command to your CI:
 .. code-block:: bash
 
     $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
-    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs\.dist|composer\.lock)$"
-    $ then
-    $     IFS=$'\n' EXTRA_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]})
-    $ fi
+    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs\.dist|composer\.lock)$"; then IFS=$'\n' EXTRA_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]}); fi
     $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist -v --dry-run --stop-on-violation --using-cache=no "${EXTRA_ARGS[@]}"
 
 Where ``$COMMIT_RANGE`` is your range of commits, eg ``$TRAVIS_COMMIT_RANGE`` or ``HEAD~..HEAD``.
--- a/README.rst
+++ b/README.rst
@@ -1225,8 +1225,8 @@ Then, add the following command to your CI:
 
 .. code-block:: bash
 
-    $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -1225,8 +1225,12 @@ Then, add the following command to your CI:
 
 .. code-block:: bash
 
-    $ IFS=$'\n'; COMMIT_SCA_FILES=($(git diff --name-only --diff-filter=ACMRTUXB "${COMMIT_RANGE}")); unset IFS
-    $ vendor/bin/php-cs-fixer fix --config=.php_cs.dist -v --dry-run --stop-on-violation --using-cache=no --path-mode=intersection -- "${COMMIT_SCA_FILES[@]}"
+    $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
+    $ if ! echo "$CHANGED_FILES" | grep -qE "^(\.php_cs\.dist|composer\.lock)$"
+    $ then
+    $    IFS=$'\n' FIX_ARGS=('--path-mode=intersection' '--' ${CHANGED_FILES[@]})
```
keradus: in original solution, there were `unset IFS` also. is it not needed for sure ?

keradus:
While I do like concept to fix this issue, i have no idea would your solution works or not. 
For that, let us test it.
Having that in mind, I added tests for currently released version (#2947), please let us focus on that PR first, then we will come back to your PR as we will have possibility to add tests

README.rst
julienfalque: Nope: when a variable definition is prepended to a command, it only applies to this command.

keradus:
#2947 got merged finally.
please rebase and add tests to ensure new solution is proper ;)

julienfalque:
Done.

tests/CiIntegrationTest.php
```diff
@@ -168,6 +163,80 @@ public function testIntegration()
...
     }
 
+    public function testIntegrationWithChangesToConfigFile()
+    {
+        static::executeCommand(implode(' && ', array(
+            'git checkout -b changes-to-config-file -q',
+            'echo "" >> dir\ b/file\ b.php',
+            'echo "echo 1;" >> dir\ b/file\ b.php',
+            // `sed -i ...` is not handled the same on Linux and macOS
+            'sed -e \'s/@Symfony/@PSR2/\' .php_cs.dist > .php_cs.dist.new',
```
keradus: as you exclude 3 different files in `src`, you shall test all 3 cases in `tests` (probably with data provider)

tests/CiIntegrationTest.php
```diff
@@ -80,39 +80,34 @@ public function tearDown()
...
         )));
 
         $integrationScript = explode("\n", str_replace('vendor/bin/', './../../../', file_get_contents(__DIR__.'/../dev-tools/ci-integration.sh')));
         $steps = array(
-            'COMMIT_RANGE="master..case1"',
+            'COMMIT_RANGE="master..random-changes"',
             $integrationScript[3],
-            $integrationScript[4],
-            $integrationScript[5],
+            "$integrationScript[4]; $integrationScript[5] $integrationScript[6]; $integrationScript[7]",
```
keradus: why you are joining 4-7 instead of providing them as-is ?

tests/CiIntegrationTest.php
julienfalque: Not sure I understand, which files are you talking about?

tests/CiIntegrationTest.php
julienfalque: I did this for readability of the suggestion, but as it's meant for CI integration, it doesn't really matter so I joined the lines in the suggestion directly.

tests/CiIntegrationTest.php
keradus: you have test with `.php_cs.dist` modified.
you shall add test cases about `.php_cs` and `composer.lock`

tests/CiIntegrationTest.php
```diff
@@ -103,16 +103,11 @@ public function testIntegration()
...
-            'echo "${#CHANGED_FILES[@]}"',
-            'echo "${CHANGED_FILES[@]}"',
-            'echo "${CHANGED_FILES[0]}"',
-            'echo "${CHANGED_FILES[1]}"',
+            'echo "$CHANGED_FILES"',
         ));
 
         $this->assertSame(
             array(
-                '2',
```
keradus: please keep number of files

tests/CiIntegrationTest.php
```diff
@@ -168,6 +163,80 @@ public function testIntegration()
...
+            'echo "echo 1;" >> dir\ b/file\ b.php',
+            // `sed -i ...` is not handled the same on Linux and macOS
+            'sed -e \'s/@Symfony/@PSR2/\' .php_cs.dist > .php_cs.dist.new',
+            'mv .php_cs.dist.new .php_cs.dist',
+            'git add .',
+            'git commit -m "Random changes including config file" -q',
+        )));
+
+        $integrationScript = explode("\n", str_replace('vendor/bin/', './../../../', file_get_contents(__DIR__.'/../dev-tools/ci-integration.sh')));
+        $steps = array(
```
keradus: logic of this test is same as prev one.
you just need data provider for branch-case and expected outputs

tests/CiIntegrationTest.php
julienfalque: `$CHANGED_FILES` is not an array anymore, it would not really make sense.

tests/CiIntegrationTest.php
julienfalque: Done.

tests/CiIntegrationTest.php
julienfalque: Done.

README.rst
```diff
@@ -1225,8 +1225,8 @@ Then, add the following command to your CI:
 
 .. code-block:: bash
 
-    $ IFS=$'\n'; CHANGED_FILES=($(git diff --name-only --diff-filter=ACMRTUXB "${COMMIT_RANGE}")); unset IFS
-    $ EXTRA_ARGS=('--path-mode=intersection' '--' "${CHANGED_FILES[@]}")
+    $ CHANGED_FILES=`git diff --name-only --diff-filter=ACMRTUXB $COMMIT_RANGE`
```
keradus: bash linters will ask you to use `${FOO}` instead of `$FOO`

src/Console/Command/HelpCommand.php
```diff
@@ -250,30 +250,17 @@ public static function getHelpCopy()
...
-            implode("\n", array_map(function ($line) { return '    $ '.$line; }, $ciIntegration)),
-            $template
-        );
-
-        return str_replace(
-           '%%%FIXERS_DETAILS%%%',
-            self::getFixersHelp(),
-            $template
-        );
+        return strtr($template, array(
```
keradus: use str_replace instead, we don't need recursive replacing done by strtr

tests/CiIntegrationTest.php
```diff
@@ -80,21 +80,32 @@ public function tearDown()
...
-            'rm -r dir\ c',
-            'echo "" >> dir\ b/file\ b.php',
-            'echo "echo 1;" >> dir\ b/file\ b.php',
-            'git add .',
-            'git commit -m "case1" -q',
+    /**
+     * @param string   $branchName,
+     * @param string[] $caseCommands,
+     * @param string[] $expectedResult1Lines,
+     * @param string[] $expectedResult2Lines,
```
keradus: extra commas

julienfalque: Nice catch, fixed.

README.rst
julienfalque: Done (also applied some suggestions from http://www.shellcheck.net)

keradus:
Thank you @julienfalque.

PR #2928, opened 2017-07-20
TokensAnalyzer::getClassyElements - Anonymous class support

SpacePossum:


tests/Tokenizer/TokensAnalyzerTest.php
```diff
@@ -68,16 +68,51 @@ class Foo2
...
+        $this->assertSame('property', $elements[1]['type'], '@{1}');
+        $this->assertSame('property', $elements[2]['type'], '@{2}');
+        $this->assertSame('property', $elements[3]['type'], '@{3}');
+        $this->assertSame('property', $elements[4]['type'], '@{4}');
+        $this->assertSame('const', $elements[5]['type'], '@{5}');
+        $this->assertSame('method', $elements[6]['type'], '@{6}');
+        $this->assertSame('method', $elements[7]['type'], '@{7}');
+        $this->assertSame('method', $elements[8]['type'], '@{8}');
+        $this->assertSame('class', $elements[9]['type'], '@{9}');
+        $this->assertSame('const', $elements[10]['type'], '@{10}');
```
keradus: what are those `@{\d+}` ?

SpacePossum:
hmm looking at this again there are some BC break in here.
More on higher level the values returned by the method or not really helpful as it is (still) not clear which class components belong to which class.

tests/Tokenizer/TokensAnalyzerTest.php
SpacePossum: indexes like ` $elements[10]['type'], '@{10}');` index 10 -> @{10}

keradus: ah, so this is actually raw-printed as literally `@{10}` ? not a template or sth...

SpacePossum: yeah, afraid so

keradus: maybe `Elemend #i` or `Elemend under #i index` ?

julienfalque: Why not replacing with a single
```php
$this->assertSame(array(/* the full expected result */), $elements);
```
?

SpacePossum: can do :)
maybe we should think about the bigger questions first;
currently when using the method you've no idea to which class the returned elements belong, which is sub optimal.
Fixing that would be a BC break (like I did here and I only did it half way).

keradus: > currently when using the method you've no idea to which class the returned elements belong,

but that also happens when you have multiple (non-anonymous) classes in one file.

not this PR related imo

julienfalque:
This is a BC break indeed: _classy elements_ are class constants, properties and methods, not anonymous classes.

SpacePossum:
Well that is not the BC breaker IMHO, someone ask for _all_ classy elements of a file so anonymous elements are part of that. Same as current when 2 non-anonymous class definitions are in the same file.
The BC breaker part is introducing `class` as property which should be reverted.

keradus:
> The BC breaker part is introducing class as property which should be reverted.

I understand @julienfalque to have exactly this part in mind, not multiple classes (anon or not) per file

SpacePossum:
I stand corrected as I indeed misread the comment :+1:

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -63,6 +63,12 @@ public function getClassyElements()
                 continue;
             }
 
+            if ($token->isClassy()) { // anonymous class in class
+                $classCurlyBraces = 0;
```
keradus: if you have anonymous class in regular class, won't this line break detection of end of regular class ?

SpacePossum: hmm good point, this is not covered by test, will update.

keradus:
Thank you @SpacePossum.

PR #2929, opened 2017-07-20
Add NoUnneededCurlyBracesFixer

### LINKED ISSUE #1540: Feature request: `case` without curly braces

### LINKED ISSUE #2798: no_unneeded_control_parentheses have no effect

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2799

closes #2798
closes #1540

$ php php-cs-fixer describe no_unneeded_curly_braces
Description of no_unneeded_curly_braces rule.
Removes unneeded curly braces.

Fixing examples:
 * Example #1.
```diff
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ @@
   -<?php { echo 1; }
   +<?php  echo 1;
   ----------- end diff -----------
```

tested on this repo and SF

### DIFF
--- a/src/Fixer/ControlStructure/NoUnneededCurlyBracesFixer.php
+++ b/src/Fixer/ControlStructure/NoUnneededCurlyBracesFixer.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ControlStructure;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class NoUnneededCurlyBracesFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Removes unneeded curly braces that are superfluous and aren\'t part of a control structure\'s body.',
+            [
+                new CodeSample(
+                    '<?php {
+    echo 1;
+}
+
+switch ($b) {
+    case 1: {
+        break;
+    }
+}
+'
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must be run before NoUselessElseFixer and NoUselessReturnFixer.
+        return 26;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound('}');
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($this->findCurlyBraceOpen($tokens) as $index) {
+            if ($this->isOverComplete($tokens, $index)) {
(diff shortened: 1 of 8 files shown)


GrahamCampbell:
For the record, 👎 on this fixer existing.

GrahamCampbell:
Oh, wait. It doesn't do what I thought... maybe it needs renaming?

julienfalque:
@GrahamCampbell What confused you exactly? The fixer's name looks good to me.

keradus:
as he didn't provided any details, i bet he was thinking about braces in:
```php
if (true) {
    echo "foo";
}
```

which is not needed in fact, but we want to keep it (and this fixer is not removing them).

more precise name, yet more ugly, would be sth like `NoBracesOfStatementGroupWithoutControlStructureFixer`

SpacePossum:
>NoBracesOfStatementGroupWithoutControlStructureFixer

how about;

>NoCurlyBracesWithoutControlStructureFixer

?  not sure about the namespace of the fixer though

julienfalque:
I'm fine with current name, maybe it's better to make the description more explicit about what is an _unneeded curly brace_ instead? Like _Removes curly braces that are superfluous and don't wrap a control structure's body._ WDYT?

SpacePossum:
Can we please remove fabbot.io as service? It provides no meaning as we tests the CS of the code better.
I know it is some way to see what version of the fixer fabbot.io uses. However that is not of interest of most contributors and all are effected by the failed check as PR's show as failed, both in on the person PR over as well in the PR overview of the repo.

keradus:
I would rather detect that our cofig is too future-aware and skip fabbot in given situation. Yet, nothing this PR is related to.

SpacePossum:
Can we than remove fabbot.io as being required to pass?
It is related to this PR, as I feel not acting on fail status of this PR because I assume false negative of fabbot.io while there might be a legit failure of travis.

keradus:
> Can we than remove fabbot.io as being required to pass?

it is not required, required are only AppVeyor and Travis
your concern is about github ui directly, as they display overall status on top of all integrations, not only required ones.

keradus:
done: #2935

tests/Fixer/ControlStructure/NoUnneededCurlyBracesFixerTest.php
```diff
@@ -0,0 +1,152 @@
...
+    public function testNoFix7($expected)
+    {
+        $this->doTest($expected);
+    }
+
+    public function provideNoFixCases7()
+    {
+        return [
+            [
+                '<?php
```
julienfalque: Why do the two first test cases here require PHP 7?

SpacePossum: they don't, I just kept on writing tests :P

SpacePossum: fixed :)

src/Fixer/ControlStructure/NoUnneededCurlyBracesFixer.php
```diff
@@ -0,0 +1,110 @@
...
+ */
+final class NoUnneededCurlyBracesFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Removes unneeded curly braces that are superfluous and aren\'t of a control structure\'s body.',
```
SpacePossum: done :+1:

keradus:
Thank you @SpacePossum.

PR #2935, opened 2017-07-23
special handling of fabbot.io service if it's using too old PHP CS Fixer version

keradus:


julienfalque:
This `try/catch` has a drawback: when running the tool internally, if the config is invalid, it will fallback to an empty ruleset even if not run by fabbot.io. This means that a legitimate error is silenced and we have no way to detect it, even with CI.

This can be reproduced easily by adding a non-existant rule in the config of your own fork and running `./php-cs-fixer fix`. The only clue is that it runs much faster ;)

@fabpot May I ask you if there is a reliable way that would help us detect when the config file is loaded by fabbot.io?

keradus:
best solution IMHO - if fabbot.io would export global env var, like `CI_FABBOT` or sth

fabpot:
I've set `FABBOT_IO` to `1` when running PHP-CS-Fixer from fabbot.io.

PR #2933, opened 2017-07-22
ProcessLinterTest - Ensure Windows test only runs on Windows, add a Mac test execution

aidantwoods:
As requested here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2893#issuecomment-317202736

Not sure on methodology though 😉 

They pass:
screen shot 2017-07-22 at 20 26 56

cc: @keradus 

---

We should probably open a bug to remind us to remove this hotfix when phpunit is working properly.

PR #2936, opened 2017-07-23
CircleCI - use PHP 7.1

aidantwoods:


circle.yml
```diff
@@ -1,3 +1,7 @@
+machine:
+  php:
+    version: 5.6.14
```
keradus: i dont find this config to work on osx

aidantwoods: Hmm... yeah I think you're right actually – only seeing it mentioning ubuntu now looking at it.

I'll try pull it from homebrew in that case.

circle.yml
```diff
@@ -1,6 +1,11 @@
 machine:
-  php:
-    version: 5.6.14
+  xcode:
+    version: 8.2
```
keradus: do we need to configure xcode ?

circle.yml
```diff
@@ -1,6 +1,11 @@
...
+
+dependencies:
+  pre:
+    - brew tap homebrew/homebrew-php
+    - brew install php56
```
keradus: could we jump to 7.1 ?

circle.yml
aidantwoods: Circle docs indicate xcode is bundled at `7.x`, latest homebrew requires `>=8`

circle.yml
aidantwoods: Yup, you're happy for no tests on `5.x` for Mac then?

keradus: yep, as long as we could have only one compiler version let it be newest for biggest code execution

aidantwoods:
Circle UI is very overwhelming ;p can you spot where that `brew install` is? `phpunit` reads the php version as `5.5` :/

aidantwoods:
Two remaining failures are expected until #2933

circle.yml
```diff
@@ -1,11 +1,18 @@
+machine:
+  xcode:
```
keradus: all the other has 4 spaces indent

keradus:
Thank you @aidantwoods

PR #2938, opened 2017-07-24
Tests fix configuration of project

SpacePossum:
- test own configuration
- add getter for the fixer name of an invalid fixer configuration exception
- add tests for new getter
- fix swapped tests names for the configuration exceptions

### DIFF
--- a/tests/AutoReview/ProjectFixerConfigurationTest.php
+++ b/tests/AutoReview/ProjectFixerConfigurationTest.php
@@ -10,7 +10,7 @@
  * with this source code in the file LICENSE.
  */
 
-namespace PhpCsFixer\Tests;
+namespace PhpCsFixer\Tests\AutoReview;
 
 use PhpCsFixer\Console\ConfigurationResolver;
 use PHPUnit\Framework\TestCase;
@@ -20,7 +20,8 @@ use PHPUnit\Framework\TestCase;
  *
  * @internal
  *
- * @covers .php_cs.dist
+ * @coversNothing
+ * @group auto-review
  */
 final class ProjectFixerConfigurationTest extends TestCase
 {
--- a/tests/AutoReview/ProjectFixerConfigurationTest.php
+++ b/tests/AutoReview/ProjectFixerConfigurationTest.php
@@ -31,7 +31,7 @@ final class ProjectFixerConfigurationTest extends TestCase
 
         $this->assertFileExists($setLocation, 'Failed to find ".php_cs.dist".');
         $this->assertTrue(is_readable($setLocation), sprintf('Config file "%s" is not readable.', $setLocation));
-    
+
         /** @var \PhpCsFixer\Config $config */
         $config = require $setLocation;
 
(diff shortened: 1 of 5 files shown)


keradus:
> filter out rules unknown to fabbot
> filter out rules with unknown configuration to fabbot

for now, `.php_cs.dist` is pretty nice example. while I do understand your changes, I'm not sure are the overwhelm in config file is worth it. a lot of logic makes our file not to be an example. IMO when we want to run the rules that are too new for fabbot.io, we could skip all rules (still, letting fabbot perform other proofreading like spell checking). We do check the rules on our side anyway, thus imo simple callback "do PHP CS Fixer based checks if possible or skip them completely is enough.

SpacePossum:
makes a lot of sense, maybe a little on your PR to only catch `PhpCsFixer\ConfigurationException\InvalidConfigurationException` ?
eveything else means something is really wrong I think ;)

I'll clean out this PR soon

keradus:
yeah, probably we don't need to catch everythink...

SpacePossum:
you should only need
`PhpCsFixer\ConfigurationException\InvalidConfigurationException`
fixer config exception extends from it so should be OK

tests/ProjectFixerConfigurationTest.php
```diff
@@ -0,0 +1,37 @@
...
+namespace PhpCsFixer\Tests;
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ *
+ * @covers .php_cs.dist
```
keradus: wow. probably it;s not working

tests/ProjectFixerConfigurationTest.php
```diff
@@ -0,0 +1,37 @@
...
+        $setLocation = __DIR__.'/../.php_cs.dist';
+
+        /** @var \PhpCsFixer\Config $config */
+        $config = require $setLocation;
+
+        $this->assertInstanceOf('PhpCsFixer\Config', $config);
+        $this->assertEmpty($config->getCustomFixers());
+        $this->assertNotEmpty($config->getRules());
+    }
+}
```
keradus: it's not testing our package as application/library, it's checking it's own config file that is not used when executing code as app/lib on user side.
imo, AutoReview here

SpacePossum:
expect utest to fail until https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2935

tests/ProjectFixerConfigurationTest.php
SpacePossum: ^_^
is there a correct way to hint a tests covers a file (so not a class)?

keradus: probably not, but if https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2938#discussion_r129013202 , then it will be coversNothing.

tests/AutoReview/ProjectFixerConfigurationTest.php
```diff
@@ -0,0 +1,46 @@
...
+{
+    public function testCreate()
+    {
+        $setLocation = __DIR__.'/../../.php_cs.dist';
+
+        $this->assertFileExists($setLocation, 'Failed to find ".php_cs.dist".');
+        $this->assertTrue(is_readable($setLocation), sprintf('Config file "%s" is not readable.', $setLocation));
+
+        /** @var \PhpCsFixer\Config $config */
+        $config = require $setLocation;
```
keradus: prior 2 assertions are pointless, as if file is not there/ is not readable, without having those assertions this test will fail as well

keradus:
Thank you @SpacePossum.

PR #2940, opened 2017-07-24
Add NoHomoglyphNamesFixer

mcfedr:
This is useful for programmers who often have multiple active
keyboard layouts and might accidentally name things with homoglyphs
in without realising

Homoglyphs are unicode characters that look very similar to basic ascii
characters.

You might start typing using cyrillic keyboard layout, backspace when you
see your mistake, but leave a С character, which looks just like a C, but is actually
a cyrillic letter for 'S sound'. 

The problem is compounded by IDE autocomplete, with works just fine with this
badly named class/variable/etc. And even PSR-0/4 autoloading, which works fine,
until you hit a filesystem that doens't work well with unicode characters and everything
breaks.

### DIFF
--- a/src/Fixer/Naming/HomoglyphNamesFixer.php
+++ b/src/Fixer/Naming/HomoglyphNamesFixer.php
@@ -1,0 +1,236 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Naming;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Fred Cox <mcfedr@gmail.com>
+ */
+final class HomoglyphNamesFixer extends AbstractFixer
+{
+    /**
+     * Used the program https://github.com/mcfedr/homoglyph-download
+     * to generate this list from
+     * http://homoglyphs.net/?text=abcdefghijklmnopqrstuvwxyz&lang=en&exc7=1&exc8=1&exc13=1&exc14=1.
+     *
+     * Symbols replaced include
+     * - Latin homoglyphs
+     * - IPA extensions
+     * - Greek and Coptic
+     * - Cyrillic
+     * - Cyrillic Supplement
+     * - Letter­like Symbols
+     * - Latin Numbers
+     * - Fullwidth Latin
+     *
+     * This is not the complete list of unicode homographs, but limited
+     * to those you are more likely to have typed/copied by accident
+     *
+     * @var array
+     */
+    private static $replacements = [
+        'O' => '0',
+        '０' => '0',
+        'I' => '1',
+        '１' => '1',
+        '２' => '2',
+        '３' => '3',
+        '４' => '4',
+        '５' => '5',
+        '６' => '6',
+        '７' => '7',
+        '８' => '8',
+        '９' => '9',
+        'Α' => 'A',
+        'А' => 'A',
+        'Ａ' => 'A',
+        'ʙ' => 'B',
+        'Β' => 'B',
+        'В' => 'B',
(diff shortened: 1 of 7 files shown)


src/Fixer/Naming/HomoglyphNamesFixer.php
```diff
@@ -0,0 +1,250 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Fred Cox <mcfedr@gmail.com>
+ */
+final class HomoglyphNamesFixer extends AbstractFixer
+{
+    /**
+     * Used the site https://github.com/mcfedr/homoglyph-download.
```
keradus: 1. why the fork? (https://github.com/mcfedr/homoglyph)
2. are the list of "types" (ipa, greek, cyrillic, ...) is just a full list of homoglyphs provided by that repo, or just a subset ?
3. why copy paste the code?
4. at https://github.com/mcfedr/homoglyph/blob/master/php/homoglyph.php#L6 there are 1560 homoglyphs, in your PR way less.
5. dataset of homoglyphs is not the code, should not be in code but in separated file storage

src/Fixer/Naming/HomoglyphNamesFixer.php
```diff
@@ -0,0 +1,250 @@
...
+     * @return bool
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_VARIABLE, T_STRING]);
+    }
+
+    public static function match($a)
+    {
+        return static::$replacements[$a];
```
keradus: no need for proxy function

src/Fixer/Naming/HomoglyphNamesFixer.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens->findGivenKind([T_VARIABLE, T_STRING]) as $tokensByKind) {
```
keradus: why only those tokens, and not others, like `T_CONSTANT_ENCAPSED_STRING` ?

src/Fixer/Naming/HomoglyphNamesFixer.php
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    /**
+     * Check if the fixer is a candidate for given Tokens collection.
+     *
+     * Fixer is a candidate when the collection contains tokens that may be fixed
+     * during fixer work. This could be considered as some kind of bloom filter.
+     * When this method returns true then to the Tokens collection may or may not
+     * need a fixing, but when this method returns false then the Tokens collection
+     * need no fixing for sure.
```
keradus: no need for copy-pasting the docs

keradus:
I think that idea itself is nice ;)
what about `space` homoglyphs ? there are few spaces in UTF ;d

src/Fixer/Naming/HomoglyphNamesFixer.php
mcfedr: 1/2/3/4. I came to this because of issues with cyrillic letters in my company, I went though a number of datasets, and looked at what the options are. At first I though I'd use the set from https://github.com/codebox/homoglyph - I forked it to add a php exporter.

But since then I came across the set at http://homoglyphs.net and decided it was more appropriate - its much smaller, and I felt focused on letters that a) look identical, not just similar b) are more likely to have been used accidentally - the full set contains symbols such as `①` which you are unlikely to type on your keyboard and unlikely to confuse with `1`

5. I put it in a seperate `Helper` class at first, but not sure where best in the project to put it. I think it makes sense for it to be PHP.

src/Fixer/Naming/HomoglyphNamesFixer.php
mcfedr: As far as I could tell, these are all the tokens that are literal names of classes etc. Its a valid use case to have cyrillic or other symbols in strings used by code.

mcfedr:
I'm was focusing on homoglyphs that you might have in code, so class names or variable names. I think it would be inappropriate to start editing literal strings in code.

The point was to change symbols you might not notice. Given that spaces are not normally part of a name these would be noticeable.

mcfedr:
Just having a look, you can put weird spaces in a name, this is valid, but its pretty noticeable.

```php
$replac ements = ...;
```
Weird spaces in general, instead of normal spaces, and php cannot parse the file, so I would say its not an issue.

keradus:
Thank you @mcfedr.

mcfedr:
@keradus Thank you for merging me

PR #2942, opened 2017-07-27
Use just the PHP Version without PHP_EXTRA_VERSION

rquadling:
By using a more generalised version number for PHP, running php-cs-fixer in a container (`PHP 7.0.20-2~ubuntu16.04.1+deb.sury.org+1`) vs a host (`PHP 7.0.20`) means the cache file is acceptable in both instances.

This extends to a slightly more optimized approach for using PHP-CS-FIXER in a pipeline.

Adding the following lines to your `.gitattributes` file ...
```
# Always use local .php_cs.cache and don't overwrite it in a merge.
.php_cs.cache merge=ours
```
in conjunction with the appropriate tooling to update the repo when code is fixed, will provide for a far faster analysis of the code.

If the pipeline can differentiate by branch, then you can add a --no-cache option to the master branch, forcing a full scan on changes to master.

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -184,7 +184,7 @@ final class ConfigurationResolver
                 $this->cacheManager = new FileCacheManager(
                     new FileHandler($this->getCacheFile()),
                     new Signature(
-                        PHP_VERSION,
+                        PHP_VERSION_ID,
                         ToolInfo::getVersion(),
                         $this->getRules()
                     ),


rquadling:
confused.com regarding `fabbot.io — Some changes should be done to comply with our standards.` as all I did was add `_ID` to 1 line! Ha!

SpacePossum:
Hi and thanks for the PR :)
Before reviewing I just wanted to let you know you can ignore `fabbot.io` for now, we are working on it ;)

keradus:
@rquadling , this change would invalidate all current caches, right ?

rquadling:
Ah. Yes.

I suppose a better option would be to take PHP_VERSION and strip/remove PHP_EXTRA_VERSION.

But using PHP_VERSION_ID is a consistent value and comparable if needed within the cache layer.

But, if you swap environments for what is ostensible the same version of PHP, you are already invalidating the cache every time.

So, for example, run php-cs-fix on host twice. First time is long, second time fast. No go to your container which is using the same code base (i.e. just mapped in drive), run php-cs-fix twice, slow and fast.

I'm happy to go with anything that would be least impactful. And that could be just stripping off the PHP_EXTRA_VERSION value.

`PHP_EXTRA_VERSION ? substr(PHP_VERSION, 0, - strlen(PHP_EXTRA_VERSION)) : PHP_VERSION`

Would that fit better? For those without PHP_EXTRA_VERSION, there should be no change to the cache file.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1649#issuecomment-168550347

SpacePossum:
Catching diff. RC's is very nice, but RC1 on Windows should be RC1 on Ubuntu/CentOS/Debian for example right?
Not sure we can match on RC's but ignore OS details though.

robfrawley:
@keradus A few comments above the one you referenced he mentions perhaps only using the major and minor version, as well, and the majority consensus in that conversation seems to be against using the current implementation, no? Including the extra string (with vendors) _is_ particularly inconvenient. I don't see a major benefit in _not_ invalidating between RCs anyway, but if that is desirable and a real world break exists otherwise, can't we glean that information from the extra section if a beta or rc version string exists, and otherwise, stick to major.minor.patch?

SpacePossum:
> can't we glean that information from the extra section if a beta or rc version string exists, and otherwise, stick to major.minor.patch?

that would be awesome :)

GrahamCampbell:
> By using a more generalised version number for PHP, running php-cs-fixer in a container (PHP 7.0.20-2~ubuntu16.04.1+deb.sury.org+1) vs a host (PHP 7.0.20) means the cache file is acceptable in both instances.

What about other patch releases that do not increment the patch version, we will want these to flush the cache, so this pr breaks this, such as the php releases on ubuntu 14.04 for example.

robfrawley:
Why?

rquadling:
I'm going to ask a dumb/naive question here, but surely the cache is just holding the crc of the file's contents and its name?

In looking at my cache file, the only thing I see in it is a copy of the rules and the file hashes.

What makes any of that information specific to a PHP release?

As a comparison, I just ran the same installation of php-cs using PHP 5.6.

Comparing the cache files, the only difference is the PHP version. All file CRCs are the same.

GrahamCampbell:
> Why? In what real-world use-case does this cause issues?

I just gave you one. Ubuntu release new versions of php without changing the version, and the whole point of invalidating the cache, is because different versions of php may behave differently, in fact, will behave differently, by definition (however, this does not imply it will necessarily affect php-cs-fixer, but we will have to assume it might).

robfrawley:
@rquadling I'm not sure what the difference between cache files is either, but there must be cases where differences occur for this to exist, I'd hope. ;-)

> Ubuntu release new versions of php without changing the version

@GrahamCampbell That is a crazy oversimplification: They do not "release new versions of PHP without changing the version." They may release "updates to previous builds" but those builds are the version reported by PHP, regardless of the "package build number" and are expected to behave exactly the same. Again, a "real-world use-case" means something where we can point to a referenceable situation where not including the vendor would result in incorrect behavior. Do we have one?

SpacePossum:
>What makes any of that information specific to a PHP release?

A simple example why the PHP version is used in the cache.
Lets say I fix the following on PHP5.5 using `pow_to_exponentiation`:

```php
<?php
echo pow(1,2);
```

Fixer does not fix the file and stores some meta data about it in the cache.
Now running the same configuration and fixer version on PHP7 with the same cache file will result in the same, i.e. no fixes while now these should've been done.

For that the cache should be at least based on the PHP version and the fixer version.

>Ubuntu release new versions of php without changing the version

That should only be done for patch version updates of PHP which are of no concern for the cache.

The only "real world cases" I see for using the complete PHP version info is RC versions. As these might have the same major/minor/patch values while acting differently. Other info like build number and OS are not relevant for the cache AFAIK.

rquadling:
I may be completely missing the point here and I really do apologise for that.

In my mind, I see php-cs-fixer as, in the bluntest of terms, a text manipulator. It doesn't "understand" the code as such, beyond having a knowledge of keywords and syntax. It isn't a static analysis tool. It's a code styling tool.

As such, it doesn't matter if my OSX PHP 7.0.20 (70020) _executes_ my code ever so slightly differently to my docker container PHP 7.0.20+xyz (70020) as that wouldn't be identifiable within my code. For me, my assumption is if I'm running my code on PHP 7.0.20, then I'm running it on PHP 7.0.20. I suspect most people running live code would expect the same. The specifics of the providers patches/enhancements should NOT have any bearing on the layout/styling of my code. Even if there is a subtle difference in the execution of my code. The styling of my code would certainly have no need to know the patch level.

With regard to what data is needed to be stored for php-cs-fixer in the cache, does it matter what version of PHP is being used to coerce the code to the standard required? Surely php-cs-fixer isn't actually running the code? I don't even see the need to hold the PHP version at all as it surely should have no effect on what operations php-cs-fixer performs. If the file has changed based upon the hash value of the contents of the file, it gets processed. Simples.

As far as I can tell, I should be able to use ANY version of PHP to coerce my code to the standard I have described in the configuration. And even any other non PHP tool that can read the list of crcs to determine what files have been changed.

Like I said right at the top, I am probably missing something. Opinionated as I am, I am willing to listen so understand what I'm missing.

SpacePossum:
>The specifics of the providers patches/enhancements should NOT have any bearing on the layout/styling of my code.

agreed, the exception being RC IMHO

>I don't even see the need to hold the PHP version at all as it surely should have no effect on what operations php-cs-fixer performs. 

Maybe my example wasn't clear? 
PHP-CS_Fixer takes the PHP version it runs on into consideration as to not break your code by modernizing it to, lets say, PHP7 syntax when you run the tool on PHP5.x.

So again; if you run the fixer on lets say PHP 4 and tell it to change your code to use short array syntax it won't do it. If you would take the cache from that run and execute again on PHP 7 your code get changed. If you would take out the PHP version check it wouldn't change the code on PHP7 while it should.

It doesn't execute your code, but it does basic SCA for good reasons, it is not a simple regex based search replace tool ;)

rquadling:
Aha. I see. So one "solution" could be to have php-cs-fixer require a target PHP version and then coerce the code to that version (so short arrays would be rejected as an option if target is PHP 4 for example). That would probably be a significant development and not easily done.

So, we want the version of PHP as defined by PHP and not defined by a provider who may or may not actually provide a change to the version number.

Would a whitelist of acceptable PHP_EXTRA_VERSIONs could be a relatively low impact way to handle this?

Or an option for php-cs-fixer to indicate if the PHP_EXTRA_VERSION should be incorporated into the cache. Default would be yes to maintain BC. And the option would simply remove the PHP_EXTRA_VERSION off PHP_VERSION before storing it in the cache.

Functionally, this would provide me the solution I would need where I'm just working on the same version in all places I'm running php-cs-fixer. (By same I mean PHP_VERSION_ID matches).

SpacePossum:
I would like to see if we could try to determine RC versions (like storing `PHP7.2.0RC1vendor-build-number` as 7.2.0.1) with an acceptable reliability. This would suit your use case I think and that of others as well, without adding new options.

robfrawley:
Obviously, the real implementation would use `PHP_VERSION_EXTRA` and not an array of random extra strings, but something like this should work.

```php
function getPhpVersion($v) {
        $version = PHP_VERSION_ID;

        if (1 === preg_match('{^-(?<extra>(beta|rc)-?([0-9]+)?(-dev)?)}i', $v, $matches)) {
                $version .= ':'.$matches['extra'];
        }       

        return $version;
}

$phpExtraVersions = [
        '',
        '-1+ubuntu17.04.1+deb.sury.org+1',
        '-beta1-dev',
        '-rc10',
        '-rc2-dev',
        '-beta-2-dev',
        '-beta2',
        '-beta-9'
];

foreach ($phpExtraVersions as $v) {
        echo getPhpVersion($v).PHP_EOL;
}
```
Outputs:
```
70107
70107
70107:beta1-dev
70107:rc10
70107:rc2-dev
70107:beta-2-dev
70107:beta2
70107:beta-9
```

SpacePossum:
Is there something like `PHP_VERSION_EXTRA`? (I'm just curious as goolge'ing didn't bring up anything).

Your sample code looks promising :)
If there is nothing "out there" we could utilize on than maybe it would be nice to see this sample integrated as a PR. However please note this is just my opinion and maybe it is best to hear from others as well on this topic. Thanks for the effort so far.

robfrawley:
Yeah; I misordered the working of the constant. ;-)

keradus:
`'-2~ubuntu16.04.1+deb.sury.org+1'`
shall also leave `-2`, as
```
php -r "echo PHP_VERSION;"
7.1.1-1
```

is completely valid and different than 7.1.1

there is a lot of edge cases here and you won't be able to find all.
it's better to not use the cache if we are not sure than use it if we wrongly think it's valid.

and in the end, exchanging cache across different env is not popular (big :-1: to do so anyway)

robfrawley:
We are positive that cached between different deb package releases don't differ in what is semantically valid in the language itself though. The only changes between these are security patches.

keradus:
what we care is syntax and if code is valid or not. security patches don't count, beta/rc channels does.
so, in `7.1.1-1` the `-1` is from debian package manager, not from php versioning ?

rquadling:
Having reworked our pipelines I no longer need this feature or have an issue. So if it is problematic (and it seems to be), then I'm happy to close this PR.

SpacePossum:
Thanks for reporting back.
I'm going to close this PR than as it seems no one is working on it and the PR is not ready for merging as is now.
If someone want to resolve/work more on this than please discuss here further before opening a new PR.
To sum up, we are open to the improvement to deal with same PHP version with different vendor info attached, however the implementation must be good enough to see the diff. in PHP version (up to patch number) _and_ RC version.

PR #2943, opened 2017-07-27
FunctionToConstantFixer - test with diff. arguments than fixable

SpacePossum:
While prep'ing for a nice suggestion from @kalessil here https://github.com/kalessil/phpinspectionsea/commit/79ca0d9e5036156980bfe31387d2fa5728485992#commitcomment-23268362 I noticed we didn't test for some cases we shouldn't fix.
Good news is that we don't touch the cases, but it is nice to have tests to prove it :)


keradus:
Thank you @SpacePossum.

PR #2944, opened 2017-07-28
FunctionToConstantFixer - handle get_class() -> __CLASS__ as well

SpacePossum:
ping @kalessil :)

### DIFF
--- a/src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
+++ b/src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
@@ -30,17 +30,26 @@ final class FunctionToConstantFixer extends AbstractFixer implements Configurati
     /**
      * @var string[]
      */
-    private static $availableFunctions = [
-        'get_class' => '__CLASS__',
-        'php_sapi_name' => 'PHP_SAPI',
-        'phpversion' => 'PHP_VERSION',
-        'pi' => 'M_PI',
-    ];
+    private static $availableFunctions;
 
     /**
-     * @var array<string, string>
+     * @var array<string, Token>
      */
     private $functionsFixMap;
+
+    public function __construct()
+    {
+        if (null === self::$availableFunctions) {
+            self::$availableFunctions = [
+                'get_class' => new Token([T_CLASS_C, '__CLASS__']),
+                'php_sapi_name' => new Token([T_STRING, 'PHP_SAPI']),
+                'phpversion' => new Token([T_STRING, 'PHP_VERSION']),
+                'pi' => new Token([T_STRING, 'M_PI']),
+            ];
+        }
+
+        parent::__construct();
+    }
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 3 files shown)


src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -31,8 +31,9 @@
      * @var string[]
      */
     private static $availableFunctions = [
-        'phpversion' => 'PHP_VERSION',
+        'get_class' => 'T_CLASS',
```
julienfalque: I think the constant you meant is `__CLASS__`?

tests/Fixer/LanguageConstruct/FunctionToConstantFixerTest.php
```diff
@@ -126,6 +126,30 @@ public function provideTestCases()
                 '<?php echo phpversion(); echo php_sapi_name(); echo M_PI;',
                 ['functions' => ['pi', 'phpversion']],
             ],
+            'get_class => T_CLASS' => [
```
julienfalque: Please add a test case with a trait instead of a class and one with a call to `get_class()` from outside a class. I think the code should not be changed for the latter as the result is not the same (`false` + warning vs. `""`).

kalessil:
Awesome, thanks @SpacePossum !

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
SpacePossum: rotf, yeah thanks :)

tests/Fixer/LanguageConstruct/FunctionToConstantFixerTest.php
kalessil: Trait context can be fixed as well without any side effects: https://3v4l.org/i6M96

Checking that the call happening in a method makes sense (we don't do this, but I'll fix this), @SpacePossum is it possible or needs significant efforts?

SpacePossum: https://3v4l.org/pm1jC lemme check

SpacePossum: TBH I think it is an odd thing that this wasn't patched in PHP to act the same when they added the warning.
Not sure checking the calling `get_class` outside class `""` vs. `false` is worth the effort. The reason for that is that it makes no sense anyway to do it, as all it is is a very expensive way to get `false` or `''`.

kalessil: LGTM, skipping https://github.com/kalessil/phpinspectionsea/issues/425 then as well.

julienfalque: 👍 regarding `get_class()` from outside a class. I would still add a test case with a trait though.

keradus:
Thank you @SpacePossum.

PR #2946, opened 2017-07-29
Detect extra old installations

keradus:
When instlaled via composer, there are about:
- 2500 daily installation with old vendor name (regardless of version)
- 3000 daily installation of v1 (with new vendor name)

For that, we have >5k daily downloads of version we do not support.
Let us try to reduce that numer.

### DIFF
--- a/Symfony/CS/Console/Application.php
+++ b/Symfony/CS/Console/Application.php
@@ -57,8 +57,7 @@ class Application extends BaseApplication
      */
     public function doRun(InputInterface $input, OutputInterface $output)
     {
-        // setup output
-        $stdErr = false// $output instanceof ConsoleOutputInterface
+        $stdErr = $output instanceof ConsoleOutputInterface
             ? $output->getErrorOutput()
             : ($input->hasParameterOption('--format', true) && 'txt' !== $input->getParameterOption('--format', null, true) ? null : $output)
         ;
--- a/Symfony/CS/Console/Application.php
+++ b/Symfony/CS/Console/Application.php
@@ -13,6 +13,9 @@
 namespace Symfony\CS\Console;
 
 use Symfony\Component\Console\Application as BaseApplication;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\CS\Console\Command\FixCommand;
 use Symfony\CS\Console\Command\ReadmeCommand;
 use Symfony\CS\Console\Command\SelfUpdateCommand;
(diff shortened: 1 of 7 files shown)


keradus:
@localheinz , maybe you could take a look on this and review it please ?

Symfony/CS/ToolInfo.php
```diff
@@ -19,23 +19,38 @@
...
 
         if (null === $result) {
-            $composerInstalled = json_decode(file_get_contents(self::getScriptDir().self::COMPOSER_INSTALLED_FILE), true);
+            $composerInstalled = json_decode(file_get_contents(self::getComposerInstalledFile()), true);
 
             foreach ($composerInstalled as $package) {
-                if (self::COMPOSER_PACKAGE_NAME === $package['name']) {
-                    $result = $package['version'].'#'.$package['dist']['reference'];
+                if (
+                    self::COMPOSER_PACKAGE_NAME === $package['name'] ||
```
julienfalque: `in_array()`?

Symfony/CS/Console/Application.php
```diff
@@ -48,4 +51,31 @@ public function getLongVersion()
...
+            $warningsDetector = new WarningsDetector();
+            $warningsDetector->detectOldVendor();
+            $warningsDetector->detectOldMajor();
+
+            if (FixCommand::COMMAND_NAME === $this->getCommandName($input)) {
+                $warningsDetector->detectXdebug();
+            }
+
+            foreach ($warningsDetector->getWarnings() as $warning) {
+                $stdErr->writeln(sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', $warning));
```
keradus: it's moved 1-1 from `FixCommand`, no idea

julienfalque: Just tried, I confirm that Symfony internally strips style tags when needed. However there are other places where this check could be dropped so, if you agree, I'll do this in a separate PR.

keradus: let's go with separated PR indeed, input from person who did this check originally would be nice as well

Symfony/CS/ToolInfo.php
keradus: done

keradus:
@localheinz , maybe you could take a look on this and review it please ?

keradus:
https://repo-stats.github.io
stats

Almost 30% is still using v1. Let us hope with this change it will decrease

Symfony/CS/Console/Application.php
SpacePossum: `stdErr` is an instance of `OutputInterface` and we don't check the formatter, so I think we cannot be sure escaping is done here always? If we assume we can rely on this on other places than lets remove the check here as well in a new PR, otherwise lets leave as is ;)

localheinz:
@keradus 

Ha, apologies for the delay!

keradus:
no issue, I assumed you were not interested in this ;)

PR #2947, opened 2017-07-29
Test suggested CI integration

keradus:
even if bash minimal integration command is small, it's tricky and have few edge-cases. instead of checking them all manually with every modification of suggested integration, let us start testing them with automate tests.

---------------

BTW, I tried bats.
Sadly, I don't find it nice.
One think is that repo is almost dead, last release 2 years ago.
Other thing is functionallity. As long as all assertions come fine, it's ok. but when sth fail, you want to know as much details as possible.
This is example output:
```
 ✗ check changed files
   (in test file functional/foo.bats, line 21)
     `[ "$output" = "asd" ]' failed
```

no way to check what was value of `$output`


tests/Fixtures/ci-integration/setUp.sh
```diff
@@ -0,0 +1,19 @@
...
+
+cd $( cd -P "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+
+git init -q
+git config user.name test
+git config user.email test
+git add .
+git commit -m "init" -q
+
+git checkout -b case1 -q
```
keradus: by adding more and more branches, it's possible to create next and next test cases

tests/CiIntegrationTest.php
```diff
@@ -0,0 +1,181 @@
...
+use Symfony\Component\Process\Process;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ *
+ * @coversNothing
+ */
+final class CiIntegrationTest extends TestCase
```
keradus: probably will need to be executed only on linux

dev-tools/ci-integration/step2-extra_args.sh
```diff
@@ -0,0 +1 @@
+EXTRA_ARGS=('--path-mode=intersection' '--' "${CHANGED_FILES[@]}")
```
keradus: I have split original first step into 2 steps, so it will be easier to adjust #2925 to testing script

keradus:
cc @julienfalque 
could you take a look please?

tests/CiIntegrationTest.php
```diff
@@ -0,0 +1,187 @@
...
+
+        $this->assertRegExp(
+            '/^('.preg_quote($optionalIncompatibilityWarning, '/').')?'.preg_quote($executionDetails, '/').'$/',
+            trim($result3['stderr'])
+        );
+        $this->assertRegExp(
+            '/^Checked all files in \d+\.\d+ seconds, \d+\.\d+ MB memory used$/',
+            trim($result3['output'])
+        );
+        $this->assertSame(0, $result3['code']);
```
julienfalque: This assertion is not needed because the related `static::executeScript()` call will throw an exception if the exit code of the command is not 0.

I would drop the `$crashOnError` parameter from `executeScript()`, make it always pass `$crashOnError = false` to `executeCommand()` and check the exit code of each command *before* checking output and stderr.

tests/CiIntegrationTest.php
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    private static function executeCommand($command, $crashOnError)
+    {
+        $process = new Process($command, static::$fixtureDir);
+        $process->run();
+
+        $result = array(
```
julienfalque: Why not simply return `$process`?

dev-tools/ci-integration/step1-changed_files.sh
```diff
@@ -0,0 +1 @@
+IFS=$'\n'; CHANGED_FILES=($(git diff --name-only --diff-filter=ACMRTUXB "${COMMIT_RANGE}")); unset IFS
```
julienfalque: Splitting the script into single-line files looks overkill to me. Wouldn't it be easier to have a single file instead?

tests/CiIntegrationTest.php
```diff
@@ -0,0 +1,187 @@
...
+                $result['output']
+            ));
+        }
+
+        return $result;
+    }
+
+    private static function executeScript(array $scriptParts, $crashOnError)
+    {
+        file_put_contents(static::$tmpFilePath, implode("\n", array_merge(array('#!/usr/bin/env bash', 'set -e', ''), $scriptParts)));
```
julienfalque: What is the purpose of the `set -e` line?

tests/Fixtures/ci-integration/setUp.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
```
keradus: this file was created when i was trying to test this whole behaviour in bash only (with `bats` - tests runner), it requires a setup to be dedicated file
with current php-based `executeCommand` method we could probably run setup with that method instead of dedicated file.
That's what you mean?

tests/CiIntegrationTest.php
keradus: as we indeed always pass `true`, I would indeed drop the param, but still raise condition and rely on exception class details which is exposing command-line error details

good catch

tests/CiIntegrationTest.php
keradus: method doesn't need to expose all process details

tests/Fixtures/ci-integration/setUp.sh
julienfalque: Yep :)

tests/CiIntegrationTest.php
keradus: to crash the script when any command of it is failing, not when last one will fail
we already do it in other `.sh` scripts we do have

dev-tools/ci-integration/step1-changed_files.sh
keradus: right
artifact from `bats` approach when it requested each testing command to be separated file

tests/CiIntegrationTest.php
julienfalque: Well this is for internal purposes only: the method is private and its result not meant to be used outside the class, so it's not that important. IMO wrapping data inside an array complicates things as you have to know the structure of the array. Using the instance would be easy with IDE autocompletion. Anyway it's a detail, if you prefer using an array, just leave it as is :)

keradus: you got me with strong typehinting.

tests/Fixtures/ci-integration/setUp.sh
keradus: done

tests/CiIntegrationTest.php
keradus: done

tests/CiIntegrationTest.php
keradus: and then I sit down to this today.
it may happen that we will extract this `bash-code-runner` to separate class. then, tightening interface with external dependency.

dev-tools/ci-integration/step1-changed_files.sh
keradus: done

keradus: done

PR #2952, opened 2017-07-30
BlankLineBeforeStatementFixer - Reference fixer instead of test class

localheinz:
This PR

* [x] updates a comment to reference fixer instead of the test class

Follows #2384.


keradus:
Thank you @localheinz.

localheinz:
Thank you, @keradus!

PR #2953, opened 2017-07-31
BlankLineBeforeStatementFixer - Add more statements

localheinz:
This PR

* [x] adds more statements to the `BlankLineBeforeStatement` fixer
    * [x] `die()`
    * [x] `exit()`
    * [x] `goto`
    * [x] `yield`

Follows #2384.

💁‍♂️ Not sure what I missed out, can you think of any other statements we could/should add?

### DIFF
--- a/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
+++ b/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
@@ -474,7 +474,6 @@ if ($foo === $bar) {
             ],
         ];
     }
-
     /**
      * @dataProvider providerFixWithFor
      *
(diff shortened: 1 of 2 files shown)


julienfalque:
What about `die`?

localheinz:
How about `goto` as well?

julienfalque:
Not sure about `goto`. We think using it is a bad practice and we don't really want to support it.

localheinz:
@julienfalque 

>Not sure about `goto`. We think using it is a bad practice and we don't really want to support it.

Totally agree, and have no intention of using it, either!

I don't mind removing it, let me know!

keradus:
don't mind about `goto`. if it would not be there, I wouldn't ask for it. But if it's already there, I don't mind keeping it.

keradus:
I love our test set :D

julienfalque:
Let's keep `goto` then.

keradus:
having said that, please make tests happy ;)

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
```diff
@@ -525,25 +525,33 @@ public function providerFixWithGoto()
...
+a:
+
 if ($foo === $bar) {
     echo $baz;
     goto a;
 }',
             ],
             [
                 '<?php
+a:
```
keradus: `a: goto a;` :D:D:D goto bomb :D

keradus:
One of 200k assertions is still failing

localheinz:
@keradus 

Fixing it, give me a moment.

keradus:
Thank you @localheinz.

localheinz:

PR #2954, opened 2017-07-31
NoBreakCommentFixer - Disable case sensitivity

### LINKED ISSUE #2948: NoBreakCommentFixer should ignore case of first character
Currently, if there's a `// No break` comment, PHP-CS-Fixer adds another `// no break` comment. It should either fix the case of the comment that's already there or instead just ignore the case and do nothing.

julienfalque:
Fixes #2948.


keradus:
Thank you @julienfalque.

PR #2959, opened 2017-08-02
MethodArgumentSpaceFixer - Skip body of fixed function

### LINKED ISSUE #2958: ensure_fully_multiline format more than needed
I've found that method_argument_space with ensure_fully_multiline as true (as is set in PSR-2) is messing with other code not related.

- PHP versions: 7.0 & 7.1 (other not checked)
- PHP CS Fixer 2.4.0 Silver Gingerbread by Fabien Potencier and Dariusz Ruminski

php-cs-fixer fix --using-cache=no --rules='{"method_argument_space": {"ensure_fully_multiline": true}}' example.php

```php
<?php
exit(call_user_func(function () {
    echo "a", "b",
        "c", "\n";
    return 0;
}));
```

Result (**why is this considering that the function must be in a new line?**):
```php
<?php
exit(call_user_func(
    function () {
    echo "a",
    "b",
        "c",
    "\n";
    return 0;
}
));
```

Expected (**repair only the echo line, don't touch anything else**):
```php
<?php
exit(call_user_func(function () {
    echo "a",
    "b",
        "c",
    "\n";
    return 0;
}));
```

greg0ire:
This fixer is supposed to fix arguments, not the body of the function
itself, so just skip it, be it when parsing or when fixing.
Fixes #2958

### DIFF
--- a/tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
+++ b/tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
@@ -648,6 +648,39 @@ UNAFFECTED
                 null,
                 ['ensure_fully_multiline' => true],
             ],
+            'test function argument with multiline echo in it' => [
+            <<<'UNAFFECTED'
+<?php
+call_user_func(function ($arguments) {
+    echo 'a',
+      'b';
+}, $argv);
+UNAFFECTED
+            ,
+                null,
+                ['ensure_fully_multiline' => true],
+            ],
+            'test function argument with oneline echo in it' => [
+            <<<'EXPECTED'
+<?php
+call_user_func(
+    function ($arguments) {
+    echo 'a', 'b';
+},
+$argv
+);
+EXPECTED
+            ,
+            <<<'INPUT'
+<?php
+call_user_func(function ($arguments) {
+    echo 'a', 'b';
+},
+$argv);
+INPUT
+            ,
+                ['ensure_fully_multiline' => true],
+            ],
         ];
     }
 
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -163,6 +163,12 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
                 continue;
             }
 
+            if ($token->equals('}')) {
+                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index, false);
+
+                continue;
+            }
+
             if ($token->equals(',')) {
                 $this->fixSpace2($tokens, $index);
                 if (!$isMultiline && $this->isNewline($tokens[$index + 1])) {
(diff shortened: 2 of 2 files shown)


greg0ire:
@eclipxe13 please review

eclipxe13:
It works great. I add a double check in test just to make sure that a fix inside an anonymous function has been done and no change was made to the upper call, share to you.
```php
            'test fix call inside function does not change upper call' => [
                <<<'EXPECTED'
<?php
call_user_func(function ($arguments) {
    xecho(
        'a',
    'b'
    );
}, $argv);
EXPECTED
                ,
                <<<'INPUT'
<?php
call_user_func(function ($arguments) {
    xecho('a',
    'b');
}, $argv);
INPUT
                ,
            ],
```
PHPUnit pass. I also run the fix on the code base where I found the issue and it works as expected.

I hope it reach the master branch soon.

greg0ire:
Thanks for your thorough testing and bug reports, much appreciated!

keradus:
Please change target branch to 2.2

keradus:
@greg0ire , probably 2.4 instead of 2.2, as on 2.2 there were no `ensure_fully_multiline` option.
or, even without this option, some fix has to gone to 2.2 ?

greg0ire:
No it should be 2.4 indeed I think.

greg0ire:
Should be fine now, sorry for the mixup.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -648,6 +648,39 @@ class FooClass
...
                 ['ensure_fully_multiline' => true],
             ],
+            'test function argument with multiline echo in it' => [
+            <<<'UNAFFECTED'
+<?php
+call_user_func(function ($arguments) {
+    echo 'a',
+      'b';
+}, $argv);
+UNAFFECTED
```
keradus: does `ensure_fully_multiline:true` means following fix shall be made?
```diff
-foo($a, $b);
+foo(
    $a,
    $b
);
```

eclipxe13: No. `foo($a, $b);` is fine. But if you are using 
```
foo($a,
    $b);
```
By PSR2 or Symfony then you MUST write it as:
```
foo(
    $a,
    $b
);
```
See http://www.php-fig.org/psr/psr-2/#46-method-and-function-calls

greg0ire: No it does not, since the original function call is not event partially multiline.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -648,6 +648,39 @@ class FooClass
...
+            ,
+                null,
+                ['ensure_fully_multiline' => true],
+            ],
+            'test function argument with oneline echo in it' => [
+            <<<'EXPECTED'
+<?php
+call_user_func(
+    function ($arguments) {
```
greg0ire: I contributed nothing regarding indention in my feature. We established it was out of scope and that all was provided was a best effort to make the code PSR2 compliant, which I believe this is.

julienfalque: True, I remember now 👍

keradus:
Thank you @greg0ire.

hkdobrev:
Thanks for this! It helped me with a `[$a, $b, $c] = explode();` call being put on multiple lines when in the body of a closure used as an argument.

Could you please release a patch version with this? Thank you!

keradus:
Hi. That's great to hear !
For next patches releses (2.2.x, 2.4.x) I want to have https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2987 resolved first.

PR #2962, opened 2017-08-04
Allow sebastian/diff 2.0

garak:


keradus:
:-1:
there is some more thing to do here to adopt v2 than just allow it in composer.
@SpacePossum knows the agenda, as he brought v2 to live.

garak:
Well, the build is green.
Also, it looks like the bigger change in new version is dropping support for older php.

keradus:
no. biggest change for v2 is unified diff format

garak:
Are you telling that such change is impacting here? If so, there's some problems with tests

keradus:
as already said, SpacePossum knows the agenda

derrabus:
Is there something, a potential contributor could help you with? Right now, for projects that fetch dev tools like CS Fixer and PHPUnit via Composer (which is a common practise), this issue is blocking the PHPUnit upgrade.

orlangur:
@derrabus PHPUnit upgrade to which version? 6.2 is the last version with new features which seems to be still compatible (https://github.com/magento/magento2/blob/develop/composer.json#L77), the next release with new featurs is 6.4 to be released in October.

garak:
```
$ composer why-not phpunit/phpunit:6.3.0
phpunit/phpunit  6.3.0       requires          sebastian/comparator (^2.0.2)                  
myproject        dev-master  does not require  sebastian/comparator (but 2.0.0 is installed)  
phpunit/phpunit  6.3.0       requires          sebastian/diff (^2.0)                          
myproject        dev-master  does not require  sebastian/diff (but 1.4.3 is installed)   
```

derrabus:
@orlangur I'm talking about 6.3. https://github.com/sebastianbergmann/phpunit/blob/6.3.0/composer.json#L40

orlangur:
So? Why do you need 6.3 if it does not contain any new features?

derrabus:
@orlangur I'm not saying, I need it. I'm saying the upgrade is blocked.

keradus:
@derrabus, there were sth to be done.
SpacePossum in general is bringing big feature of better diff, for that he done all the work in sebastianbergman/differ, phpunit/phpunit and there is agenda what needs to be done in our package.
Sadly, only SpacePossum is aware of all actions needed for our repo, and he is offline for few days.

keradus:
ping @SpacePossum

SpacePossum:
Sorry I was away for a while.

Since `diff 2` will give us what we need (line numbers + option to pass own formatter) we want to move to there. However `diff 2` requires PHP 7 while this project supports PHP 5.6.
Therefor I created a fork of `diff 2` and I'm adding support for PHP 5.6 there.
After that I'll add the formatters we need/want.

This means this PR will not be merged, however when the work is done the problem of not being able to require both PHPUnit 6.3 and this project should be resolved.

To respond on some other things on this PR:

>Also, it looks like the bigger change in new version is dropping support for older php.

the biggest changes were:
- fix a whole bunch of bugs
- create diff-output much closer to the udiff format
- allow for custom formatters
- allow adding linenumbers to the output
- and yes, PHP7

>Are you telling that such change is impacting here? If so, there's some problems with tests

yes, version 1.4 of `diff` has an impact to us which are resolved on 2.0. However the tests do not reflect that as such test would be green on 1.4 and red on 2.0 because of a bug in 1.4, which makes no sense to have here.

derrabus:
So, diff 2 has been created for a feature this project needs and now this project forks it to actually be able to use it? This issue keeps confusing me. 😉 

Anyway, thanks for the update, @SpacePossum.

SpacePossum:
I started PR-ing `diff` to add the features we want here. However this could not be done without BC breaks on `diff`. Therefor it had to be bumped a major to `v2`, maintainer was happy with that if PHP version would be bumped to PHP 7 as well.
So yeah, I kind of shot myself in the foot ;) 
anyway when all is done we should've a nice solution and the fork can go when this project bumps to PHP 7 : P

enumag:
What's wrong with bumping this project to PHP 7 now?

SpacePossum:
>What's wrong with bumping this project to PHP 7 now?

There are still a lot users that don't use PHP7, bumping now would not be nice to them.
Besides we like to have a user base that is bigger than PHP7 users, as this also means more contributors (I hope ;)).

kubawerlos:
I believe after Symfony 4 will released (which require PHP 7) a logical step after 2.5 will be to aim to version 3 with PHP 7 support only.

keradus:
for last month, almost 40% users installing PHP CS Fixer via composer are using 2.0-2.2:
https://repo-stats.github.io/
previously, the update between minors were pretty fast, the fact that a lot of users stucked at 2.2 is IMO caused by PHP 5.6 bump we did with 2.3. If our users cannot yet update to 5.6, they won't do it to 7.x.

We prefer to keep lower PHP version lower in favour of not having next tool version to long-term support.

Due to great reasoning of @SpacePossum , I'm closing this PR.
Yet, feel free to continue discussion

dunglas:
Maybe it's just time to remove PHP 7 support from PHP CS Fixer? Symfony 4 itself will only support PHP 7.1.

kubawerlos:
@keradus and @SpacePossum I cannot see what wrong would make merging this. Yes, diff 2 has some new features, but seems we won't have them here for a while.

For people using PHP 5.6 (even 5.3) diff 1.4 would be installed and they will be happy as the are using PHPUnit 5. For some of us having PHP 7 it would allow install latest PHPUnit, which might have interesting features quite soon.

keradus:
as @SpacePossum (who did the work and create most of the code for diff v2!) wrote, it's in the progress. agenda is there, work is started, please be patient.

PR #2969, opened 2017-08-08
Shrink down AccessibleObject usage

keradus:


tests/Linter/ProcessLinterProcessBuilderTest.php
```diff
@@ -0,0 +1,72 @@
...
+     *
+     * @testWith ["php", "foo.php", "php -l foo.php"]
+     *           ["C:\\Program Files\\php\\php.exe", "foo bar\\baz.php", "\"C:\\Program Files\\php\\php.exe\" -l \"foo bar\\baz.php\""]
+     * @requires OS ^Win
+     */
+    public function testPrepareCommandOnPhpOnWindows($executable, $file, $expected)
+    {
+        $this->assertSame(
+            $expected,
+            (new ProcessLinterProcessBuilder($executable))->build($file)->getCommandLine()
```
keradus: 5.3 syntax

src/Linter/ProcessLinterProcessBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    /**
+     * @param string $path
+     *
+     * @return Process
+     */
+    public function build($path)
+    {
+        // @TODO change at 2.4 to:
```
GrahamCampbell: Annotations aren't valid in `//` style comments I don't think?

src/Linter/ProcessLinterProcessBuilder.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+final class ProcessLinterProcessBuilder
+{
+    /**
+     * @var string
+     */
+    private $executable;
+
+    /**
+     * @param string $executable PHP executable, null for autodetection
```
julienfalque: You removed `null` from original allowed types but forgot to update the description.

src/Linter/ProcessLinterProcessBuilder.php
julienfalque: The point is to have those listed by IDEs (which are pretty lenient with this) and ultimately removed, so I'd say this is perfectly fine.

keradus: this is not an annotation. it's fine.

src/Linter/ProcessLinterProcessBuilder.php
keradus: nice catch

PR #2975, opened 2017-08-12
[TrailingCommaInMultilineArrayFixer] simplify isMultilineArray condition

TomasVotruba:
Just a trial. This might speed up things a bit.


keradus:
please prove its actually fasten anything.
nest scenario: 1 jump into function less, 1 trivial object (no work in constructor) instantation less
I wouldn't expect this PR to gain any visible performance improvement ,it's extra small microoptimization.

TomasVotruba:
Sorry, I didn't mean speed up in performance, but in reading the code.

julienfalque:
I also doubt the performance gain will ever be noticeable. But indeed this simplifies the code so 👍 .

This should target 2.2 though. @TomasVotruba if you have some time, could you please rebase onto 2.2?

TomasVotruba:
@juliendufresne Thanks for reply. I've a little time so I've rather prepared standalone PR #2982

PR #2982, opened 2017-08-17
TrailingCommaInMultilineArrayFixer - simplify isMultilineArray condition

TomasVotruba:
Replaces #2975 with 2.2 branch as target


src/Fixer/ArrayNotation/TrailingCommaInMultilineArrayFixer.php
```diff
@@ -65,12 +65,6 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
      */
     private function fixArray(Tokens $tokens, $index)
```
keradus: if you assume now that the array is multiline and don't want to check that statement inside of the method (as it's private and you could assume that), please at least make the function name indicates that

TomasVotruba: Sure, how exactly?
`fixMultilineArray`?

keradus:
Thank you @TomasVotruba.

TomasVotruba:
Glad to help

Issue #2967, opened 2017-08-08
AlignMultilineCommentFixer - cyrillic symbol replaces to undefined unicode symbol


ivaaaan:
**Before running:**
```
    /**
     * @SWG\Get(
     *     path="/api/v0/cards",
     *     operationId="listCards",
     *     tags={"Банковские карты"},
     *     summary="Возвращает список банковских карт."
     *  )
     */
    public function indexAction()
    {

    }
```

**After:**
```
    /** 
     * @SWG\Get(
     *     path="/api/v0/cards",
     *     operationId="listCards",
     *     tags={"Банковские карты"},
     *     summary="Возвращает список банковски�
     * карт.",
     */
    public function indexAction()
    {

    }
```

**Excepted:**
```
    /**
     * @SWG\Get(
     *     path="/api/v0/cards",
     *     operationId="listCards",
     *     tags={"Банковские карты"},
     *     summary="Возвращает список банковских карт."
     *  )
     */
    public function indexAction()
    {

    }
```

Problem only with cyrillic char "х".

**PHP version:**
```
$ php -v
PHP 7.1.8-2+ubuntu16.04.1+deb.sury.org+4 (cli) (built: Aug  4 2017 13:04:12) ( NTS )
```
**PHP-CS-FIXER version:**
```
$ php-cs-fixer -v
PHP CS Fixer 2.4.0 Silver Gingerbread by Fabien Potencier and Dariusz Ruminski (63661f3)
```
**Command:**
`$ php-cs-fixer fix --allow-risky=yes`

**PHP-CS-FIXER configuration:**
```
<?php

$finder = PhpCsFixer\Finder::create()
    ->in(__DIR__."/src")
;

return PhpCsFixer\Config::create()
    ->setRules([
    '@Symfony' => true,
    '@Symfony:risky' => true,
    '@PHP71Migration' => true,
    'array_syntax' => ['syntax' => 'short'],
    'dir_constant' => true,
    'heredoc_to_nowdoc' => true,
    'linebreak_after_opening_tag' => true,
    'modernize_types_casting' => true,
    'no_multiline_whitespace_before_semicolons' => true,
    'no_unreachable_default_argument_value' => true,
    'no_useless_else' => true,
    'no_useless_return' => true,
    'ordered_class_elements' => true,
    'ordered_imports' => true,
    'phpdoc_add_missing_param_annotation' => ['only_untyped' => false],
    'phpdoc_order' => true,
[...]

PR #2984, opened 2017-08-19
AlignMultilineCommentFixer - handle uni code

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2967

target 2.4 as fixer does not exist on lower branches


SpacePossum:
not sure about if the other regex'es need the flag as well, cannot create a failing case for those

keradus:
Thank you @SpacePossum.

PR #2987, opened 2017-08-19
BracesFixer - Fix incorrect indentation of comments

### LINKED ISSUE #2986: `braces` fixer mess with inline comment indentation
```console
robert@robert-laptop:~/Poligon/php-cs-fixer$ php -v
PHP 7.1.8-2+ubuntu16.04.1+deb.sury.org+4 (cli) (built: Aug  4 2017 13:04:12) ( NTS )
    with Zend OPcache v7.1.8-2+ubuntu16.04.1+deb.sury.org+4, Copyright (c) 1999-2017, by Zend Technologies

robert@robert-laptop:~/Poligon/php-cs-fixer$ vendor/bin/php-cs-fixer -V
PHP CS Fixer 2.2.5 by Fabien Potencier and Dariusz Ruminski
```
Config:

```php
    ->setRiskyAllowed(true)
    ->setRules([
        'braces' => true,
    ])
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->in(__DIR__)
    );

return $config;
```

Command:

```console
robert@robert-laptop:~/Poligon/php-cs-fixer$ vendor/bin/php-cs-fixer fix
```

Diff:

```diff
diff --git a/test1.php b/test1.php
index 7ab983e..f0174cd 100644
--- a/test1.php
+++ b/test1.php
@@ -1,17 +1,17 @@
 <?php
 
 class A
 {
     public function b()
     {
         if (true) {
             $this->doSomethingElse();
 
-//            if ($extension === 'composer') {
+            //            if ($extension === 'composer') {
 //                return [];
 //            }
         } elseif (false) {
             $this->doSomethingElse();
         }
     }
 }
```

Expected result is to not changing anything - such comment does not meant to be indented in any way. But I'm not sure why `braces` fixer touch this comment - it looks completely unrelated.

rob006:
Fixes #2986

### DIFF
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -846,6 +846,11 @@ class Foo
 
         $nextToken = $tokens[$nextTokenIndex];
         if ($nextToken->isComment()) {
+            $previousToken = $tokens[$nextTokenIndex - 1];
+            // do not indent comments starting at the beginning of line - non-indented comments should not change indentation
+            if ($previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())) {
+                return;
+            }
             $tokens[$nextTokenIndex] = new Token(array(
                 $nextToken->getId(),
                 preg_replace(
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -847,8 +847,11 @@ class Foo
         $nextToken = $tokens[$nextTokenIndex];
         if ($nextToken->isComment()) {
             $previousToken = $tokens[$nextTokenIndex - 1];
-            // do not indent comments starting at the beginning of line - non-indented comments should not change indentation
-            if ($previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())) {
+            // do not indent inline comments used to comment out unused code
+            if (
+                0 === strpos($nextToken->getContent(), '//' . $this->whitespacesConfig->getIndent())
+                && $previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())
+            ) {
                 return;
             }
             $tokens[$nextTokenIndex] = new Token(array(
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -364,7 +364,7 @@ class Foo
                     $nextLineCanBeIndented = false;
                     if ($nestToken->equalsAny(array(';', '}'))) {
                         $nextLineCanBeIndented = true;
-                    } elseif ($this->isCommentWithFixableIndendation($tokens, $nestIndex)) {
+                    } elseif ($this->isCommentWithFixableIndentation($tokens, $nestIndex)) {
                         for ($i = $nestIndex; $i > $startBraceIndex; --$i) {
                             if ($tokens[$i]->equalsAny(array(';', '}'))) {
                                 $nextLineCanBeIndented = true;
@@ -890,7 +890,7 @@ class Foo
      *
      * @return bool
      */
-    private function isCommentWithFixableIndendation(Tokens $tokens, $index)
+    private function isCommentWithFixableIndentation(Tokens $tokens, $index)
     {
         if (!$tokens[$index]->isComment()) {
             return false;
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -849,7 +849,7 @@ class Foo
             $previousToken = $tokens[$nextTokenIndex - 1];
             // do not indent inline comments used to comment out unused code
             if (
-                0 === strpos($nextToken->getContent(), '//' . $this->whitespacesConfig->getIndent())
+                0 === strpos($nextToken->getContent(), '//'.$this->whitespacesConfig->getIndent())
                 && $previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())
             ) {
                 return;
--- a/tests/Fixer/Basic/BracesFixerTest.php
+++ b/tests/Fixer/Basic/BracesFixerTest.php
@@ -2225,6 +2225,95 @@ function D() /**
 }',
                 self::$configurationOopPositionSameLine,
             ),
+            array(
+                '<?php
+if ($foo) {
+    foo();
+
+//    if ($bar === \'bar\') {
+//        return [];
+//    }
+} else {
+    bar();
+}
+',
+            ),
+            array(
+                '<?php
+if ($foo) {
+    foo();
+
+//    if ($bar === \'bar\') {
+    //        return [];
+//    }
+} else {
+    bar();
+}
+',
+            ),
+            array(
+                '<?php
+if ($foo) {
+    foo();
+
+//    if ($bar === \'bar\') {
+//        return [];
+//    }
+    
+    $bar = \'bar\';
+} else {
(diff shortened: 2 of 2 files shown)


tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -2259,8 +2259,8 @@ function D() /**
 //    if ($bar === \'bar\') {
 //        return [];
 //    }
-    
-    $bar = \'bar\';
+    '."\n".
```
keradus: you could just do:
```diff
-'."\n".
+'.'
+
```

keradus: but why you need trailing spaces here? they are relevant for test ?

rob006: They help me learn that I should use `1 === preg_match('/\R$/', $previousToken->getContent())` instead of `$previousToken->isWhitespace("\n")`, so I keep them. :D
But I can remove them - not sure if they give any value or only noise.

keradus: if they are crucial for some edge case - keep them for sure

src/Fixer/Basic/BracesFixer.php
```diff
@@ -364,7 +364,7 @@ function ($item) {
                     $nextLineCanBeIndented = false;
                     if ($nestToken->equalsAny(array(';', '}'))) {
                         $nextLineCanBeIndented = true;
-                    } elseif ($this->isCommentWithFixableIndendation($tokens, $nestIndex)) {
+                    } elseif ($this->isCommentWithFixableIndentation($tokens, $nestIndex)) {
```
julienfalque: Nice catch! Even with this explicit diff I had to read it three times to spot what was wrong

src/Fixer/Basic/BracesFixer.php
```diff
@@ -846,6 +846,11 @@ private function ensureWhitespaceAtIndexAndIndentMultilineComment(Tokens $tokens
 
         $nextToken = $tokens[$nextTokenIndex];
         if ($nextToken->isComment()) {
+            $previousToken = $tokens[$nextTokenIndex - 1];
+            // do not indent comments starting at the beginning of line - non-indented comments should not change indentation
```
keradus:
```php
function foo()
{
    $a = 1;
// we will return sth
    return $a;
}
```

here, it shall be fixed.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -846,6 +846,11 @@ private function ensureWhitespaceAtIndexAndIndentMultilineComment(Tokens $tokens
...
+            if ($previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())) {
+                return;
```
keradus:
```php
function foo()
{
    $a = 1;
// we will return sth
    return $a;
}
```

here, it shall be fixed.

rob006: Why? Such comments are often used for temporary comment of unused code:

```php
function foo()
{
    $a = 1;
// var_dump($a);
    return $a;
}
```

Indenting these in practically always unwanted.

julienfalque: Actually I'm not sure what to do here. Absence of indentation in cases like @keradus's example should be fixed IMO, but there is no way to know if the comment is a real comment or just "removed" code. Though I think this case is pretty rare and unindented comments will most of the time be "removed" code.

On the other hand, when using a version control system like Git, one could say that commenting outcode instead of completely removing it is a bad practice. So maybe adding a new option here would be better?

SpacePossum: This case of the comment at the location is not related to braces is it? Even if it should be indented should it be done by the `braces` fixer?
IMHO the fixer should not touch this comment (same for (to much) indented comments).

julienfalque: Current behavior of the fixer *is* to fix the indentation of anything inside braces. We might extract that into a separate fixer but it's out of the scope of this PR.

keradus: I see no point of having that config.
Commented code is temporary code, either remove it before commiting or indent it properly.

I have understood #2986 bug that only first line of comment is aligned and not all of them, why you treat is as no alignment shall be done.
With later, you won't fix real comments in favor of not touching temporary commented code that shall not be part of the repo in first place.

If intention of this PR is to do the later (not touch comments without indentation), I'm big :-1: for it, due to reasoning above.

rob006: @keradus For me not breaking code is more important than fixing it, and right now php-cs-fixer makes my code worse. I would understand this if such behavior will be result of separate fixer for indentation - then I can just disable it. But this is `braces` fixer, which is part of `PSR2` ruleset. For me, as a end user, this is completely unclear why `braces` fixer reindent my comments.

keradus: `braces` is reindenting the code since beginning...

rob006: This does not explain **why** it doing it. It clearly breaks SRP here.

keradus: This is one of biggest fixers we have. It was introduced ages before this repo was able to properly manage dependencies between different fixers, thus to not have conflicts between them or run them in random order, it was prepared as single rule.
If you wish, go ahead! **Do** show how it shall be done and decouple it ;)

rob006: Sorry, but I don't have enough time and knowledge to refactor this fixer. I also need to fix #2986 in 2.2.x line and I highly doubt that such decoupling will be merged into this line.
Can't we just agree that fixing indentation should not be directly connected with fixing braces position? And then consider this PR as reducing damages of bad design of this fixer? Or at least introduce config option to disable indentation fixing? I really need fixer for braces, but don't care about fixing of indentation (especially that at this point it is far from perfect).

rob006: @keradus Take a look at last commit - right now it does not indent only comments that looks like commented code. Your example is covered.

src/Fixer/Basic/BracesFixer.php
```diff
@@ -846,6 +846,14 @@ private function ensureWhitespaceAtIndexAndIndentMultilineComment(Tokens $tokens
 
         $nextToken = $tokens[$nextTokenIndex];
         if ($nextToken->isComment()) {
+            $previousToken = $tokens[$nextTokenIndex - 1];
+            // do not indent inline comments used to comment out unused code
+            if (
+                0 === strpos($nextToken->getContent(), '//'.$this->whitespacesConfig->getIndent())
+                && $previousToken->isWhitespace() && 1 === preg_match('/\R$/', $previousToken->getContent())
```
keradus: could you please describe what condition you want to cover here?

keradus: For me, while using PHP Storm, commented code looks exactly like this:
```diff
 if (1) {
-    // echo 2;
+    // echo 2;
 }
```

if I get you right, you also check that there is no single-line spaces before `//`, right ?

rob006: > For me, while using PHP Storm, commented code looks exactly like this

This is configurable:
ec515ddd

> if I get you right, you also check that there is no single-line spaces before //, right ?

Only comments proceeded by EOL pass this condition, so basically this PR affects only non-indented comments. If someone will use your PhpStorm configuration, this will not change anything for him (unless he's using crazy single-space indentation).

keradus:
Thank you @rob006.

PR #2994, opened 2017-08-22
ProtectedToPrivateFixer: minor optimization

Slamdunk:
Following https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2972#discussion_r133924837 I tried to make this fixer faster, but the [trait check](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.4/src/Fixer/ClassNotation/ProtectedToPrivateFixer.php#L135) doesn't let me skip getting the class close before the main loop.

Any other idea is welcome


keradus:
Thank you @Slamdunk.

keradus:
merged into 2.2

Issue #2993, opened 2017-08-22
Any plans for sebastian/diff ^2.0 compatibility?


alcohol:
As the title suggests; `sebastian/diff` has been out for a few weeks now. `phpunit` 6.3 and higher requires it, making a side-by-side install through `require-dev` impossible at the moment.

SpacePossum:
Hi and thanks for reporting!

This issue has been noted and is a duplicate of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2962.
We are in the process of resolving this issue ASAP, please bear with us for a bit longer :)

alcohol:
Oh sorry, I could not find that one. But I was not looking for closed issue, I guess that explains it :-)

SpacePossum:
Sorry for the wait @alcohol and all the upvoters of this issue.
It took me longer than I wanted, but here we are, free of the `diff 1.4 dep`. 

For now only on master, will become part of the next minor release of the project.

alcohol:
No need to apologize. Appreciate all the amazing work everyone on this project puts in. Thanks!

Issue #1620, opened 2015-12-20
Displaying line numbers


helpfulrobot:
Following on from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1370

Could this be solved by extending the fixers to provide a list of erroneous line numbers after fix has been called? The temporal coupling isn't great, but it won't break the API... 

Edit: This might also lead to faster fix times as we could check the array length on that array instead of comparing entire source strings...

Edit: [There's a WIP branch...](https://github.com/helpfulrobot/PHP-CS-Fixer/commits/show-line-numbers)

SpacePossum:
closed by the merge of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3162

PR #2998, opened 2017-08-23
Diff - add line numbers

SpacePossum:
Small patch to add linenumbers to the `diff` output.

closes;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2993
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1620
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2529

(2529 also reports on columns, but I don't care about those ATM)

I fix the `diff` output to strict `unified diff` output in another PR some time, including adding the files names and multiple headers etc etc... however that is all none blocking

### DIFF
--- a/tests/Differ/SebastianBergmannDifferTest.php
+++ b/tests/Differ/SebastianBergmannDifferTest.php
@@ -28,16 +28,12 @@ final class SebastianBergmannDifferTest extends AbstractDifferTestCase
         $diff = <<<'TXT'
 --- Original
 +++ New
-@@ @@
+@@ -1,3 +1,4 @@
  <?php
  class Foo extends Bar {
 -    function __construct($foo, $bar) {
 +    public function __construct($foo, $bar)
 +    {
-         $this->foo = $foo;
-         $this->bar = $bar;
-     }
- }
 
 TXT;
 
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "ext-tokenizer": "*",
         "doctrine/annotations": "^1.2",
         "gecko-packages/gecko-php-unit": "^2.0",
-        "sebastian/diff": "^1.4",
+        "php-cs-fixer/diff": "dev-master",
         "symfony/console": "^3.2",
         "symfony/event-dispatcher": "^3.0",
         "symfony/filesystem": "^3.0",
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -12,8 +12,9 @@
 
 namespace PhpCsFixer\Console\Command;
 
+use PhpCsFixer\Diff\Differ;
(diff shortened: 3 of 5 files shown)


keradus:
yeah, ignore the columns for ever

tests/Differ/SebastianBergmannDifferTest.php
```diff
@@ -28,16 +28,12 @@ public function testDiffReturnsDiff()
...
 --- Original
 +++ New
-@@ @@
+@@ -1,3 +1,4 @@
  <?php
  class Foo extends Bar {
 -    function __construct($foo, $bar) {
 +    public function __construct($foo, $bar)
 +    {
-         $this->foo = $foo;
```
julienfalque: Shouldn't the diff show a few lines after the last changed one?

SpacePossum: I hope not ;)
I changed the output builder to not add same lines before and after a diff chunk, within a chunk this OK. 
Side note;
The amount of not change lines between a different lines before creating a new chunk is not configurable within `diff 2`, it will be in a new output builder I'm writing. I could an option to display "context" lines before and after each chunk if there is a need (or when I find the time).

SpacePossum: btw. the old way of adding these trailing lines was why you committed https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/3204a0703f711ea791bd9b486e5d132c174d959e (which was reverted later on, but still ;))

julienfalque: I didn't mean to show all the lines after the changes (which was indeed the reason of my commit you mentioned and should not happen), but only a few, like e.g.:

File A.txt:
```
A
B
C
D
E
F
G
H
I
```

File B.txt:
```
A
B
C
D

F
G
H
I
```

Diff:
```
 $ diff -u A.txt B.txt
--- A.txt	2017-08-23 19:34:19.194610864 +0200
+++ B.txt	2017-08-23 19:34:11.984558683 +0200
@@ -2,7 +2,7 @@
 B
 C
 D
-E
+
 F
 G
 H
```

This diff shows 3 unchanged lines before the changed one (skipping the first one) and 3 after (skipping the last one). Isn't that possible with the new differ?

SpacePossum: The current output builders provided by the `diff` package don't have options for this,
however I'm finishing up one that will allow this.

TomasVotruba:
This would be great! Thanks for opening this

SpacePossum:
this is to early, we need to do so other stuff first as not to BC break a bunch of stuff

TomasVotruba:
Sure, thanks for the news

Issue #2912, opened 2017-07-10
Info when overriding path from config


keradus:
If path is fully configured in config file and one calls CLI with path arg, eg
`php php-cs-fixer fix -v .` ,
we shall raise comment that path was overridden

ref #2911 

v2.2

PR #2999, opened 2017-08-23
Notice when paths from config file are overridden by command arguments

julienfalque:
Implements #2912.

I didn't add test as the command isn't tested at all, I plan to do it properly in a separate PR.

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -176,7 +176,7 @@ final class FixCommand extends Command
 
         if (null !== $stdErr && $resolver->configFinderIsOverridden()) {
             $stdErr->writeln(
-                sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'Paths from configuration file overridden by paths provided as command arguments.')
+                sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'Paths from configuration file have been overridden by paths provided as command arguments.')
             );
         }
 
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -175,7 +175,9 @@ final class FixCommand extends Command
         $finder = $resolver->getFinder();
 
         if (null !== $stdErr && $resolver->configFinderIsOverridden()) {
-            $stdErr->writeln('<bg=yellow;fg=black;>Paths from configuration file overridden by paths provided as command arguments.</>');
+            $stdErr->writeln(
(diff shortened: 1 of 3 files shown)


src/Console/ConfigurationResolver.php
```diff
@@ -493,6 +498,14 @@ public function shouldStopOnViolation()
     }
 
     /**
+     * @return null|bool
+     */
+    public function configFinderIsOverridden()
+    {
+        return $this->configFinderOverridden;
```
keradus: this shall never return null. 
all properties here are nullable, but getters are not providing nulls. null means here only "lazy computed"

julienfalque: Updated.

keradus:
please dont amend

src/Console/ConfigurationResolver.php
```diff
@@ -493,6 +498,14 @@ public function shouldStopOnViolation()
     }
 
     /**
+     * @return bool
+     */
+    public function configFinderIsOverridden()
+    {
+        return $this->configFinderOverridden;
```
keradus: if one wont call `getFinder` first, this will wrongly be always `false`

keradus: hm... lack of tests for it ;P

src/Console/ConfigurationResolver.php
```diff
@@ -99,6 +99,11 @@
     private $fixers;
 
     /**
+     * @var bool
+     */
+    private $configFinderOverridden = false;
```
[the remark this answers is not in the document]
julienfalque: Updated.

src/Console/ConfigurationResolver.php
julienfalque: Test added.

src/Console/Command/FixCommand.php
```diff
@@ -174,6 +174,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $progressType = $resolver->getProgress();
         $finder = $resolver->getFinder();
 
+        if (null !== $stdErr && $resolver->configFinderIsOverridden()) {
+            $stdErr->writeln(
+                sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'Paths from configuration file have been overridden by paths provided as command arguments.')
```
keradus: added `->isDecorated` for now for consistency.
we need to analyse do we still need to check `->isDecorated` on our side (and maybe drop it globally, not only for this one message)

keradus:
Thank you @julienfalque.

julienfalque:
Thanks :)

PR #3001, opened 2017-08-24
[2.5] Bugfix braces

mnabialek:
In new 2.5 version is problem (comparing to 2.3 version, maybe same was in 2.4) - when we have PHP file with content:

```
<?php

function test()
{
//    $closure = function ($callback) use ($query) {
//        doSomething();
//
//        return true;
//    };

    $a = 3;
}

```

BracesFixer will add indentation to line between **doSomething();** and **return true;** like this:

```
//        doSomething();
    //
//        return true;
```

what doesn't make sense in my opinion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,19 @@ CHANGELOG for PHP CS Fixer
 
 This file contains changelogs for stable releases only.
 
+Changelog for v2.5.0
+--------------------
+
+* feature #2770 DoctrineAnnotationSpaces - split assignments options (julienfalque)
+* feature #2843 Add estimating-max progress output type (julienfalque)
+* feature #2885 Add NoSuperfluousElseifFixer (julienfalque)
+* feature #2929 Add NoUnneededCurlyBracesFixer (SpacePossum)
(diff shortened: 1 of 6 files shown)


mnabialek:
I see the build failed, but those additional spaces were added just to make sure in case of additional spaces everything will work the same

keradus:
affected line: 2.2
ref #2987

ping @rob006 , please take a look here

rob006:
Fix looks valid to me.

@mnabialek Try put `'.'` after these trailing whitespaces.

mnabialek:
@rob006 It's done

rob006:
@mnabialek 
Like this:

```php
//    $closure = function ($callback) use ($query) {
//        doSomething();
//        '.'
//        return true;
//    };
```

This also needs to be merged into 2.2 line.

mnabialek:
@rob006 So you want me to create one more PR into 2.2 branch?

SpacePossum:
@mnabialek can retarget this PR to 2.2? If this change applies to 2.2 nicely it can be done during merging.

keradus:
no it cant. 
@mnabialek target 2.2 please

julienfalque:
Closing as replaced by #3002.

PR #3003, opened 2017-08-25
[2.4] SingleLineCommentStyleFixer - don't convert PHP annotations

mnabialek:
Now, when you add PHP annotations to help your IDE do resolve variable type like this:

```
<?php

/* @var User $user */
$user->doSomething();'
```

SingleLineCommentStyleFixer will change this into 

```
<?php

// @var User $user
$user->doSomething();'
```

what will make this annotation useless.

### DIFF
--- a/src/Fixer/Comment/SingleLineCommentStyleFixer.php
+++ b/src/Fixer/Comment/SingleLineCommentStyleFixer.php
@@ -120,7 +120,7 @@ $c = 3;',
 
                 continue;
             }
-            if (!$this->asteriskEnabled || '/*' !== substr($content, 0, 2) || 1 === preg_match('/[^\s\*].*\R.*[^\s\*]/s', $commentContent)) {
+            if (!$this->asteriskEnabled || '/*' !== substr($content, 0, 2) || 1 === preg_match('/[^\s\*].*\R.*[^\s\*]/s', $commentContent) || 1 === preg_match('/@var.*/', $commentContent)) {
                 continue;
             }
 
--- a/tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
+++ b/tests/Fixer/Comment/SingleLineCommentStyleFixerTest.php
@@ -207,6 +207,13 @@ second line*/',
             [
                 '<?php # test',
             ],
+            [
+                '<?php
+
+/* @var User $user */
+/** @var User $user */
+$user->doSomething();',
+            ],
         ];
     }
 


mnabialek:
Probably without PHP it will be difficult to make this build pass :)

SpacePossum:
Out of curiosity why don't you use the PHPDoc/PSR5 version of this?  
(like `/** @var User $user */`)

mnabialek:
@SpacePossum To be honest I do, but I looked quickly on StackOverflow today and saw also single asterisk notation and showed it here, but this PR fix them both as in example in test

julienfalque:
Like @SpacePossum said, PHPDoc, PSR-5 and even Doctrine Annotations require to use DocComments, not regular comments (so `/** @foo */` instead of `/* @foo */`). This fixer is supposed to ignore DocComments so 👎

mnabialek:
@julienfalque  As I've already written it solves both cases - just look here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3003/files#diff-20f2bd91db5054a181355eae17331e3f

SpacePossum:
If there is popular IDE support for this I tend to +1 this. 
I do however feel the new match is to relax, for example `/* @vara*/`, `/* @var a*/` should still be converted. IMHI the match should be something like `/*[optional white space but no line break]@var [type] [$var][optional white space and comment]*/`

mnabialek:
@SpacePossum To be honest it's hard to say. I think in normal comments you don't write `@var` at all and if you do I think it's safer to leave it not converted than break someone block. Also are we sure those doc blocks have always exact same format? I'm looking at PhpStorm IDE and for example such comment:

```
/** Some dummy info @var \App\Models\User $user Some dummy info */
$user->doSomething();
```

or

```
/** Some dummy info @var \App\Models\User|\Another\Class $user Some dummy info */
$user->doSomething();
```

would also help IDE but I think there is no point to get so crazy. I think the `@var` assumption at the beginning is good enough to cover all standard usages.

GrahamCampbell:
:-1:

GrahamCampbell:
> DocComments, not regular comments (so /** @foo */ instead of /* @foo */).

The single star is not phpdoc, it is just a **comment**, so whatever's inside is just free text, and is meant garbage to any IDE.

mnabialek:
@GrahamCampbell  Not true, just look at below screen to see it's helping IDE (PhpStorm 2017.2.1 EAP) to know what's variable type:

php annotations

As you see PhpStorm works well with `/**` and /*` annotation. You can also look at https://stackoverflow.com/a/14115662/3593996 to see people are using annotations this way

GrahamCampbell:
> As you see PhpStorm works well with /** and /* annotation.

Yeh, PhpStorm has not implemented it correctly. They've only done it to keep people happy who also don't do phpdocc right themselves. :/

mnabialek:
@GrahamCampbell So you think if someone is using annotation "in wrong way", this fixer should convert  his comment into single line, right? Should it do the same when annotation is correct? (now it's also converting annotation to single line comment what makes this single fixer useless at least in my projects)

GrahamCampbell:
> "in wrong way"

Well, this is the definition provided by the php language itself. We have `T_DOC_COMMENT` and `T_COMMENT`, being different things.

> it's also converting annotation to single line comment what makes this single fixer useless

Yes, anything in `T_COMMENT` is just free text, not an annotation, even if it looks like one, if you see what I mean.

GrahamCampbell:
So, an annotation in `/*` is not an annotation, it is just free text in a comment.

GrahamCampbell:
And free text is safe to be converted to the other style of comment.

keradus:
:-1: here as well.
IDE aims to match as much different syntaxes user could came up to help him
Fixer aims to match only proper syntax.

bycosta:
How about this?

```php
<?php

/** @var $bar \Foo\Bar */
$bar->doSomething();
```

This is being converted to:

```php
<?php

// @var $bar \Foo\Bar
$bar->doSomething();
```

To me this looks like a bug since now a correct PHPDoc that the IDEs read won't work.

mnabialek:
@bycosta Yes, this is what I'm saying. This PR fixed this but is fixing:
```
/* @var $bar \Foo\Bar */
```

too and that's why it was not approved. I haven't had time to look at it and create different fix

SpacePossum:
@bycosta this fixer should never touch PHPDoc's, and looking at the current implementation (so even without this PR) it doesn't. So your case might be valid bug report but not related to this fixer.
My guess for now is that the PHPDoc to comment fixer transfers the PHPDoc to comment.
Could you open a new issue for your case?

mnabialek:
I'm closing this one and investigate this further.

Issue #2907, opened 2017-07-07
[Meta] PHP 7.2 features 


SpacePossum:
PHP 7.2.0 Alpha 3 is [released](http://php.net/index.php#id2017-07-06-2) , lets list the things we need to take care of or could support.

- [ ] "object" (in any case) can no longer be used as a class name. [RFC](https://wiki.php.net/rfc/object-typehint)
- [ ] It is now possible to remove argument type annotations when overriding an inherited method. [RFC](https://wiki.php.net/rfc/parameter-no-type-variance)
- [ ] It is now allowed to override an abstract method with another abstract method in a child class. [RFC](https://wiki.php.net/rfc/allow-abstract-function-override)
- [x] A trailing comma in group use statements is now allowed. [RFC](https://wiki.php.net/rfc/list-syntax-trailing-commas) [PR](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2941)
- [ ] The "object" type annotation is now supported. [RFC](https://wiki.php.net/rfc/object-typehint)
- [ ] The `(unset)` cast has been deprecated. This does not affect the `unset($var)` language construct.
- [ ]  Deprecations for PHP 7.2 [RFC](https://wiki.php.net/rfc/deprecations_php_7_2)

Sources:
https://github.com/php/php-src/blob/php-7.2.0RC2/NEWS
https://github.com/php/php-src/blob/php-7.2.0RC2/UPGRADING

Note: I've only listed the first bunch of items I spotted that might effect us, i.e. because we made assumptions on PHP behavor and/or syntax, and items we could write fixers for.

Please comment if I forgot anything (@ maintainers, feel free to update this issue with more details when needed)

~~Ps. Jun 18 should be the feature freeze; https://wiki.php.net/todo/php72 so in theory new items might be in Beta 1 (that are not listed here yet)~~

glensc:
the freeze date is actually `Jul 18 2017 | PHP 7.2 branched and feature freeze` on the linked page

SpacePossum:
thanks :)  I either copy-pasted wrongly or the doc linked was updated, anyway I'll update the ticket :)

glensc:
[PHP 7.2.0 Beta 1 Released](http://php.net/archive/2017.php#id2017-07-20-1) :)

PHP 7.2.0 Beta 1 builds on previous releases with:

- The much anticipated [Sodium extension](https://github.com/php/php-src/pull/2560)
[...]

PR #3006, opened 2017-08-28
[2.5] Add PHP 7.2 to Travis build matrix

Jean85:
This is needed to test any development of #2907, since `php: nightly` on Travis now points to 7.3, and requiring 7.2 goes to BETA3 instead.

PR #3007, opened 2017-08-28
Add PHP 7.2 to Travis build matrix

Jean85:
This is the same as #3006 but for the 2.2 branch, as requested.

> This is needed to test any development of #2907, since php: nightly on Travis now points to 7.3, and requiring 7.2 goes to BETA3 instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
           env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: nightly
           env: TASK_SCA=1 COMPOSER_FLAGS="--ignore-platform-reqs" SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_IGNORE_ENV=1
+        - php: 7.2
+          env: TASK_SCA=1 COMPOSER_FLAGS="--ignore-platform-reqs" SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_IGNORE_ENV=1
         - php: 5.3
           env: SKIP_LINT_TEST_CASES=1 COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
           dist: precise


keradus:
I actually asked you to change base branch of already existing PR instead of creating new PR, but it's fine as well ;)
closing other PR, as only this one may be merged

keradus:
LGTM ;)

Jean85:
No problem ;) I created a new one also because changing the base branch from the web interface was bringing the whole 2.5 branch with it onto 2.2!

keradus:
it's because you change the branch to lower one.
if you take a look, change a branch is... doing a rebase. `git rebase` of branch is also needed when you rebase the PR targed

.travis.yml
```diff
@@ -20,6 +20,8 @@ matrix:
           env: DEPLOY=yes TASK_TESTS_COVERAGE=1
         - php: nightly
           env: TASK_SCA=1 COMPOSER_FLAGS="--ignore-platform-reqs" SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_IGNORE_ENV=1
+        - php: 7.2
+          env: TASK_SCA=1 COMPOSER_FLAGS="--ignore-platform-reqs" SYMFONY_DEPRECATIONS_HELPER=weak PHP_CS_FIXER_IGNORE_ENV=1
```
julienfalque: Tests pass so IMO we should also:
* update the `php` constraint in composer.json to allow PHP 7.2;
* update handling of `PHP_CS_FIXER_IGNORE_ENV` to always allow PHP 7.2;
* drop specific nightly flags here (`--ignore-platform-reqs`, `SYMFONY_DEPRECATIONS_HELPER` and `PHP_CS_FIXER_IGNORE_ENV`).

Also, test coverage is always done on the highest tested PHP version so it should be done on 7.2 now.

Jean85: I don't think we should move the coverage to a non-stable release of PHP.
Also, it seems that adding 7.2 to the allowed composer versions are blocked by #2907.

keradus: No official support for 7.2 until we fix #2907 as @Jean85 claims :+1:
Fact that tests passed is caused by... not having tests for 7.2 syntax...

julienfalque: What is there in #2907 that still prevents us for claiming PHP 7.2 support?

keradus: please discuss it in proper thread
this PR claims to extend matrix to use both 7.2 and 7.3 for beta-tests, not about officially supporting 7.2

keradus:
Thank you @Jean85.

PR #3008, opened 2017-08-28
Update composer.json

EvKoh:
Compatible with symfony 3.3


GrahamCampbell:
This won't work, as shown by the test suite.x

julienfalque:
This change was already proposed and rejected so I'm closing. See #2962 for details.

PR #3009, opened 2017-08-30
CiIntegrationTest - run local

SpacePossum:
Allow local run with `xdebug` enabled and non tracked `composer.lock`


tests/CiIntegrationTest.php
```diff
@@ -219,22 +229,31 @@ public function getIntegrationCases()
...
                 '...',
             ),
+        );
+    }
+
+    public function testWithComposerLock()
+    {
+        $process = new Process('git check-ignore composer.lock', static::$fixtureDir);
+        if (0 === $process->run()) {
+            $this->markTestSkipped('The file "composer.lock" is ignored by git.');
```
keradus: what about creating `.gitignore` file for our fixture repo that will unignore lock file ?
if not works, then I would change `git add .` to `git add . && git add composer.lock --force`

SpacePossum: :+1:  updated

tests/Fixtures/ci-integration/.gitignore
```diff
@@ -0,0 +1 @@
+!/composer.lock
```
julienfalque: What about using `!/**`? If I'm not wrong, this should include all files, so we won't have the same issue with other files.

SpacePossum: I tend to develop towards as `close` as possible and `open` when needed. In this case we don't need to open up the `ignore` fully so I would not do it. However, I'm open for other opinions.

keradus: actually, as `close` as possible here would mean `as independent` as possible, as it's integration test.
for that, we want to be affected by global config as less as possible, isn't it right?

SpacePossum: It is a two way as most of the time, in my POV;
- If I open this up to `all/wildcards` and in the future something is added to a test or files or directories within in this directory that should be ignored; than this needs to be adjusted. 
- If above does not happen, than the current state of the PR is as good as the suggested open one.

However as I just want the tests to pass locally; I do not feel strongly about the solution taken as both will do.

keradus: that's exactly the thing. we want to have git sandbox. for that, we shall not inherit from project/global git config

SpacePossum: updated

tests/Fixtures/ci-integration/.gitignore
```diff
@@ -0,0 +1,2 @@
+# reset all ignore rules to create sandbox for integration test
+!/**
```
GrahamCampbell: Missing newline at EOF.

keradus:
Thank you @SpacePossum.

PR #3010, opened 2017-08-30
Fix handling of Github releases

### LINKED ISSUE #2997: SelfUpdateCommand - Update to newest version, not latest published
Currently the `SelfUpdateCommand` updates based on the last version released, which may not be newest version within the given major version. We should fix this logic.

Think of it, this might be nice candidate for a new package in a new repo as a lot of this code can be reused by other packages, WDYT?

(cc @julienfalque)

julienfalque:
Fixes #2997.


src/Console/Command/NewVersionChecker.php
```diff
@@ -0,0 +1,169 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Console\Command;
+
+/**
+ * @internal
+ */
+class NewVersionChecker
```
SpacePossum: can be `final` I think? and feel free to `@author` yourself ;)

src/Console/Command/NewVersionChecker.php
```diff
@@ -0,0 +1,169 @@
...
+        if (null !== $this->availableVersions) {
+            return;
+        }
+
+        $url = 'https://api.github.com/repos/FriendsOfPHP/PHP-CS-Fixer/tags';
+
+        $result = @file_get_contents(
+            $url,
+            false,
+            stream_context_set_default(array(
```
SpacePossum: This is process wide right? we should set back .. although edge case..
Would a move to Guzzle or `curl` be hard or have serious side effects for us? Asking because both would provide more details and control when something does not work as expected, like getting the response code, dealing with timeouts etc.

src/Console/Command/NewVersionChecker.php
julienfalque: Class made `final`.

src/Console/Command/NewVersionChecker.php
julienfalque: Replaced `stream_context_set_default()` with `stream_context_create()`, I think this is the only place where we use this kind of context.

I'm 👍 for moving to curl or Guzzle but I think it's out of the scope of this PR.

keradus: :-1: for guzzle, it's extra big library we don't need

src/Console/Command/NewVersionChecker.php
```diff
@@ -0,0 +1,167 @@
...
+
+        return 0;
+    }
+
+    /**
+     * @param string $tag
+     *
+     * @return null|int[]
+     */
+    private function parseVersion($tag)
```
keradus: why it returns `null` for `v2.0.0-alpha` ?

src/Console/Command/NewVersionChecker.php
```diff
@@ -0,0 +1,167 @@
...
+
+        foreach ($result as $tag) {
+            $version = $this->parseVersion($tag['name']);
+            if (null !== $version) {
+                $this->availableVersions[] = $version;
+            }
+        }
+
+        // sort from newest version to oldest
+        usort($this->availableVersions, array($this, 'compareVersions'));
```
keradus: don't use string to refer to function, this call won't be detected by SCA tools

src/Console/Command/NewVersionChecker.php
keradus: assuming there are releases:
- v2.7.0
- v2.7.1
- v2.8.0-RC

and user has downloaded phar of v2.8.0-RC
then, what will happen when user trigger self-update?

keradus: hint: it will fail/crash, as it will try to use `null` as array :D

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -69,36 +69,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
             return 1;
         }
 
-        $remoteTag = $this->getLatestTag();
-        if (null === $remoteTag) {
-            $output->writeln('<error>Unable to determine newest version.</error>');
+        $checker = new NewVersionChecker('v1.1.0');
```
keradus: I assume this is WIP

src/Console/Command/NewVersionChecker.php
```diff
@@ -0,0 +1,167 @@
...
+     * @var null|array<int[]>
+     */
+    private $availableVersions;
+
+    /**
+     * @param string $currentVersion
+     */
+    public function __construct($currentVersion)
+    {
+        $this->currentVersion = $this->parseVersion($currentVersion);
```
keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3010/files#r136204306

`currentVersion` might be `2.0.0-RC`, here, you will assign null to `currentVersion` which claims to be `int[]`, and then you will provide null to array typehinted functions. not gonna work

you have to add releases like `2.0.0-RC` to tests

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -69,36 +69,31 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-            return 0;
-        }
-
-        $currentVersion = 'v'.$this->getApplication()->getVersion();
-        if ($currentVersion === $remoteTag) {
+        $latestVersion = $checker->getLatestVersion();
```
keradus: previously, there were nice error msg if fetching from github fails (and it may happen!)
> $output->writeln('<error>Unable to determine newest version.</error>');

now, we just have app crashed

src/Console/Command/NewVersionChecker.php
julienfalque: Replaced with `composer/semver` usage.

src/Console/Command/NewVersionChecker.php
julienfalque: Done.

src/Console/Command/SelfUpdateCommand.php
julienfalque: :D

src/Console/Command/NewVersionChecker.php
julienfalque: Replaced with `composer/semver` usage.

src/Console/Command/SelfUpdateCommand.php
julienfalque: Error handling restored.

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -69,17 +70,24 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         }
 
-        $checker = new NewVersionChecker('v1.1.0');
+        $checker = new NewVersionChecker($this->getApplication()->getVersion());
+
+        try {
+            $latestVersion = $checker->getLatestVersion();
+            $latestVersionOfCurrentMajor = $checker->getLatestVersionOfCurrentMajor();
+        } catch (\Exception $exception) {
+            $output->writeln('<error>Unable to determine newest version.</error>');
```
SpacePossum: maybe print the exception message to the output for details as well (could be done based on verbosity level maybe)?

composer.json
```diff
@@ -15,6 +15,7 @@
     ],
     "require": {
         "php": "^5.3.6 || >=7.0 <7.2",
+        "composer/semver": "^1.4",
```
SpacePossum: really like this move :+1:  :)

src/Console/Command/SelfUpdateCommand.php
julienfalque: Exception message added but not based on verbosity level: IMO you don't want to run the command again just to know why it failed. Also this would require you to be aware that this error message will be shown when increasing verbosity.

SpacePossum: makes total sense, thanks man :)

composer.json
```diff
@@ -15,9 +15,9 @@
     ],
     "require": {
         "php": "^5.3.6 || >=7.0 <7.2",
-        "composer/semver": "^1.4",
         "ext-json": "*",
         "ext-tokenizer": "*",
+        "composer/semver": "^1.4",
```
keradus: admit it, you did add this dep originally by editing the file, not via command, didn't you ?

SpacePossum: > admit it, you did add this dep originally by editing the file, not via command, didn't you ?

asking cause ` "sort-packages": true` should've prevented the insert at the previous location? ;)
anyway, I edit `composer.json` by hand most of the time as well : P

tests/Console/SelfUpdate/NewVersionCheckerTest.php
```diff
@@ -0,0 +1,361 @@
...
+            array('v2.5.0', null),
+            array('v3.0.0', null),
+            array('v3.2.0', null),
+            array('v3.2.5', null),
+        );
+    }
+
+    private function createGithubClientStub()
+    {
+        $githubClient = $this->prophesize('PhpCsFixer\Console\SelfUpdate\GithubClient');
```
keradus: you shall prophesize only interfaces.
why this internal class is not final? :P

composer.json
julienfalque: ;)

src/Console/SelfUpdate/NewVersionChecker.php
```diff
@@ -0,0 +1,121 @@
...
+    private $currentVersion;
+
+    /**
+     * @var null|string[]
+     */
+    private $availableVersions;
+
+    /**
+     * @param string       $currentVersion
+     * @param GithubClient $githubClient
```
keradus: not true

src/Console/SelfUpdate/NewVersionChecker.php
```diff
@@ -0,0 +1,121 @@
...
+     */
+    public function __construct($currentVersion, GithubClient $githubClient = null)
```
keradus: avoid optional deps

tests/Console/SelfUpdate/NewVersionCheckerTest.php
julienfalque: Interface added.

src/Console/SelfUpdate/NewVersionChecker.php
julienfalque: Updated.

src/Console/SelfUpdate/NewVersionChecker.php
julienfalque: It is now :)

src/Console/SelfUpdate/NewVersionChecker.php
```diff
@@ -0,0 +1,117 @@
...
+namespace PhpCsFixer\Console\SelfUpdate;
+
+use Composer\Semver\Comparator;
+use Composer\Semver\Semver;
+use Composer\Semver\VersionParser;
+
+/**
+ * @internal
+ */
+final class NewVersionChecker
```
keradus: imo this class does too much things.
it shall:
- provide newest version
- provide newest version for given MAJOR
- expose compare function for 2 provided versions (returning -1, 0, 1)

julienfalque: I don't understand your comment: you say it does too much things and then you suggest that it should do more things that it already does (it already provides the newest version and the newest version for current MAJOR).

keradus: that's the thing - it doesnt. it does **if** they are newer. methods are not atomic. either provide value or compare them, you do both at once.
so if I will want to get newest version, i cannot do it nicely as I need to provide `0.0.0` as base version used for comparision

julienfalque: Class updated to have atomic methods.

keradus:
I have rebased this PR to fix hidden conflict (had to rename utest providers to match new AutoReview tests).

keradus:
Thank you @julienfalque.

julienfalque:
Thanks :)

PR #3013, opened 2017-09-02
Adjust phpunit configuration

localheinz:
This PR

* [x] references `phpunit.xsd` using a relative path
* [x] adjusts configuration to be as strict as possible
* [x] increases verbosity

Follows https://github.com/php-coveralls/php-coveralls/pull/242#discussion_r136697081.


phpunit.xml.dist
```diff
@@ -13,6 +18,7 @@
     convertWarningsToExceptions="true"
     processIsolation="false"
     stopOnFailure="false"
+    verbose="true"
```
keradus: with this, do we still need to add `-v` when we run tool under Travis/AppVeyor/CircleCI ?

localheinz: I don't think so, let me remove it!

keradus:
Thank you @localheinz.

keradus:
@localheinz 
now, we have 150 risky tests on 2.2 line like:
```
150) PhpCsFixer\Tests\Linter\TokenizerLinterTest::testLintSource with data set #0 ('<?php echo 123;')
This test did not perform any assertions
```

could you fix it please ?

localheinz:

Will take a look st the risky tests tomorrow!

PR #3015, opened 2017-09-03
Fix exception arguments

julienfalque:
The three classes have the same constructor which has only three arguments, the `$previous` exception being the last one.


keradus:
Thank you @julienfalque.

PR #3021, opened 2017-09-05
Risky rules cleanup

kubawerlos:
`PowToExponentiationFixer` is a risky fixer therefore it should be in `@PHP56Migration:risky` group, not `@PHP56Migration`.

### DIFF
--- a/src/RuleSet.php
+++ b/src/RuleSet.php
@@ -151,19 +151,20 @@ final class RuleSet implements RuleSetInterface
             'psr4' => true,
             'silenced_deprecation_error' => true,
         ),
-        '@PHP56Migration' => array(
+        '@PHP56Migration' => array(),
+        '@PHP56Migration:risky' => array(
             'pow_to_exponentiation' => true,
         ),
         '@PHP70Migration' => array(
-            '@PHP56Migration' => true,
+            'ternary_to_null_coalescing' => true,
+        ),
+        '@PHP70Migration:risky' => array(
+            '@PHP56Migration:risky' => true,
+            'declare_strict_types' => true,
             'random_api_migration' => array('replacements' => array(
                 'mt_rand' => 'random_int',
                 'rand' => 'random_int',
             )),
-            'ternary_to_null_coalescing' => true,
-        ),
-        '@PHP70Migration:risky' => array(
-            'declare_strict_types' => true,
         ),
         '@PHP71Migration' => array(
             '@PHP70Migration' => true,
--- a/tests/RuleSetTest.php
+++ b/tests/RuleSetTest.php
@@ -306,6 +306,8 @@ final class RuleSetTest extends TestCase
             array(array('@PSR1' => true), true),
             array(array('@PSR2' => true), true),
             array(array('@Symfony' => true), true),
+            array(array('@Symfony:risky' => true), false),
+            array(array('@PHP56Migration:risky' => true), false),
             array(
                 array(
                     '@Symfony:risky' => true,
@@ -313,13 +315,31 @@ final class RuleSetTest extends TestCase
                 ),
                 false,
             ),
-            array(
-                array(
-                    '@Symfony:risky' => true,
-                ),
-                false,
-            ),
         );
+    }
+
+    /**
+     * @requires PHP 7.0
+     */
+    public function testPHP70MigrationSet()
+    {
+        $this->testRiskyRulesInSet(array('@PHP70Migration' => true), true);
+    }
+
+    /**
+     * @requires PHP 7.0
+     */
+    public function testPHP70MigrationRiskySet()
+    {
+        $this->testRiskyRulesInSet(array('@PHP70Migration:risky' => true), false);
+    }
+
+    /**
+     * @requires PHP 7.1
+     */
+    public function testPHP71MigrationSet()
+    {
+        $this->testRiskyRulesInSet(array('@PHP71Migration' => true), true);
     }
 
     public function testInvalidConfigNestedSets()
--- a/tests/RuleSetTest.php
+++ b/tests/RuleSetTest.php
@@ -319,27 +319,42 @@ final class RuleSetTest extends TestCase
     }
 
     /**
+     * @param array $set
+     * @param bool  $safe
+     *
+     * @dataProvider providePhp70SafeSets
      * @requires PHP 7.0
      */
-    public function testPHP70MigrationSet()
+    public function testPhp70MigrationSet(array $set, $safe)
     {
-        $this->testRiskyRulesInSet(array('@PHP70Migration' => true), true);
+        $this->testRiskyRulesInSet($set, $safe);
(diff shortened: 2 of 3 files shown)


SpacePossum:
Looks like a good fix.
However I wonder, there is a [utest](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/tests/RuleSetTest.php#L280) for this preventing risky rules in non-risky tests, so somehow that test does not what it should.

julienfalque:
@SpacePossum The test is correct, but the `@PHP56Migration` ruleset is not returned by the data provider it uses, which explains why we didn't get any failure.

julienfalque:
@kubawerlos As this is a bugfix, can you please update your PR to target 2.2 branch?

kubawerlos:
Test was fine, simply didn't test migration sets. Now it does, which also discovered `@PHP70Migration` contained 2 risky fixers.

src/RuleSet.php
```diff
@@ -165,22 +165,22 @@
...
         ],
         '@PHP70Migration' => [
-            '@PHP56Migration' => true,
+            'ternary_to_null_coalescing' => true,
+        ],
+        '@PHP70Migration:risky' => [
+            '@PHP56Migration:risky' => true,
+            'declare_strict_types' => true,
             'non_printable_character' => [
                 'use_escape_sequences_in_strings' => true,
```
julienfalque: Not that with this option the `non_printable_character` fixer is not risky so having it in the `@PHP70Migration` ruleset is valid, though this will make the test fail as the `isRisky()` method always returns `true`, so not sure what to do here.

julienfalque:
@kubawerlos Indeed, you should create an empty `@PHP56Migration` ruleset to avoid breaking configs that use it (which would also happen on master anyway).

src/RuleSet.php
kubawerlos: I don't understand how - looking at [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/tests/Fixer/Basic/NonPrintableCharacterFixerTest.php#L151) test case both string are different and someone might relate to comparing with such a string and fixer will change it (it as bad as overriding native functions).

julienfalque: Those strings are not different (on PHP 7+), see https://3v4l.org/MMMlM.

tests/RuleSetTest.php
```diff
@@ -302,24 +302,31 @@ public function testRiskyRulesInSet(array $set, $safe)
...
             ),
-            array(
-                array(
-                    '@Symfony:risky' => true,
-                ),
-                false,
-            ),
         );
+
+        if (PHP_MAJOR_VERSION >= 7) {
```
keradus: no, this will not mark tests as skipped on php5

kubawerlos: I know, but I can't skip only some tests in provider. Or can I?

src/RuleSet.php
kubawerlos: Oh, yeah, right, I made mistake in mt test - silly me... So what about it? I'd leave it here - in README it's described as risky, so I'd expect it to belong to only risky groups.

tests/RuleSetTest.php
keradus: you need to make 2nd test method with 2nd provider, that will (only one of them) contain php version requirement

keradus:
please avoid squashing, you just changed sth, but I don't know what as there is only one ticket :( for that, I need to read everything again

tests/RuleSetTest.php
```diff
@@ -306,22 +306,42 @@ public function provideSafeSets()
...
+    {
+        $this->testRiskyRulesInSet(array('@PHP70Migration' => true), true);
+    }
+
+    /**
+     * @requires PHP 7.0
+     */
+    public function testPHP70MigrationRiskySet()
+    {
+        $this->testRiskyRulesInSet(array('@PHP70Migration:risky' => true), false);
```
keradus: please create data provider.

tests/RuleSetTest.php
kubawerlos: I wanted to avoid copying test method, so extracted cases... well, simply look at the updated PR, made it few seconds before your comment.

tests/RuleSetTest.php
kubawerlos: Ok

tests/RuleSetTest.php
keradus: @kubawerlos 
I do know it would be nice to achieve that, yet, for us it's better to be aware that some tests were skipped as php is not new enough.

kubawerlos:
I have created provider for PHP 7.1 as well as the `PHP71Migration:risky` will be there in newer branches.

tests/RuleSetTest.php
```diff
@@ -306,19 +306,54 @@ public function provideSafeSets()
...
+    /**
+     * @param array $set
+     * @param bool  $safe
+     *
+     * @dataProvider providePhp71SafeSets
+     * @requires PHP 7.1
+     */
+    public function testPhp71MigrationSet(array $set, $safe)
+    {
+        $this->testRiskyRulesInSet($set, $safe);
```
keradus: to be honest I'm not sure if calling real test method from inside of other test method is a good idea

kubawerlos: I wasn't either, but then I've found 6 cases of this in tests e.g. [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/tests/Fixer/DoctrineAnnotation/DoctrineAnnotationSpacesFixerTest.php#L461). Want me to change to doTest-solution?

One more question: can I create `@PHP71Migration:risky` group (which will only has `@PHP70Migration:risky` in 2.2 branch)? It's not against YAGNI as we know we are going to need it.

SpacePossum: I've no strong opinion about point 1.

About point two; maybe we can have a meta test that checks that [all know set-names](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/RuleSet.php#L270) are being tested by checking the test providers? Than missing `@PHP71Migration:risky` test on higher branch will show up during merging. It would also protect us in the future when adding a new set, WDYT?

keradus: :+1: for 1st of Kuba's paragraph
:+1: for 2nd paragraph of @SpacePossum 
(thus :-1: for 2nd paragraph from Kuba, no need for a ruleset that would be the same as other one - same way we don't have ruleset for 7.2)

keradus:
Thank you @kubawerlos.

keradus:
thanks to your generic approach @kubawerlos , we fixed
```diff
-non_printable_character [@PHP70Migration]
+non_printable_character [@PHP70Migration:risky]
```
on 2.4 line during merging of this PR, thanks !

Issue #3019, opened 2017-09-04
ConfigurationResolver - handle `rules=`  passed empty string 


SpacePossum:
```
$ php php-cs-fixer fix test.php --dry-run -vvv --rules=
You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.
If you need help while solving warnings, ask at https://gitter.im/FriendsOfPHP/PHP-CS-Fixer, we will help you!
Loaded config default.
Using cache file ".php_cs.cache".

                                  
  [ErrorException]                
  Uninitialized string offset: 0  
                                  

Exception trace:
 () at /home/possum/work/PHP-CS-Fixer/src/Console/ConfigurationResolver.php:622
```

tool should exit using the correct exit code for misconfig

PR #3024, opened 2017-09-06
ConfigurationResolver - Handle empty "rules" value

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3019

### DIFF
--- a/tests/Console/Command/FixCommandTest.php
+++ b/tests/Console/Command/FixCommandTest.php
@@ -12,7 +12,6 @@
 
 namespace PhpCsFixer\Tests\Console\Command;
 
-use PhpCsFixer\ConfigurationException\InvalidConfigurationException;
 use PhpCsFixer\Console\Application;
 use PhpCsFixer\Console\Command\FixCommand;
 use PHPUnit\Framework\TestCase;
@@ -41,7 +40,7 @@ final class FixCommandTest extends TestCase
         $this->doTestExecute(
             array('--rules' => ''),
             array(
-                'class' => InvalidConfigurationException::class,
+                'class' => 'PhpCsFixer\ConfigurationException\InvalidConfigurationException',
                 'regex' => '#^Empty rules value is not allowed\.$#',
             )
         );
@@ -60,7 +59,7 @@ final class FixCommandTest extends TestCase
             )
         );
 
-        $this->assertSame(0, $cmdTester->getStatusCode(), 'Expected exit code mismatch.');
+        $this->assertSame(0, $cmdTester->getStatusCode(), "Expected exit code mismatch. Output:\n".$cmdTester->getDisplay());
     }
 
     /**
--- a/tests/Console/Command/FixCommandTest.php
+++ b/tests/Console/Command/FixCommandTest.php
@@ -1,0 +1,110 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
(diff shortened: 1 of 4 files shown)


tests/Console/Command/FixCommandTest.php
```diff
@@ -0,0 +1,105 @@
...
+use PhpCsFixer\Console\Application;
+use PhpCsFixer\Console\Command\FixCommand;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Tester\CommandTester;
+
+/**
+ * @internal
+ *
+ * @covers \PhpCsFixer\Console\Command\FixCommand
```
SpacePossum: not sure about this

tests/Console/Command/FixCommandTest.php
```diff
@@ -0,0 +1,105 @@
...
+        $this->application = new Application();
+    }
+
+    public function testEmptyRulesValue()
+    {
+        $this->doTestExecute(
+            array('--rules' => ''),
+            array(
+                'class' => InvalidConfigurationException::class,
+                'regex' => '#^Empty rules value is not allowed\.$#',
```
keradus: not sure about having same utests for 2 layers - currently, it's for command and resolver. but the logic is the same, and command is passing values to resolver.

SpacePossum: Thought about it, but I went for this route.
When passing "php-cs-fixer --rule=&nbsp;&nbsp;&nbsp;" than the value of `--rule=` will be set to "" (and not "&nbsp;&nbsp;&nbsp;").
This behavior seems to be bound to SF/console and I thought it best to cover it in a test as well.
So in theory the application will never passes "&nbsp;&nbsp;&nbsp;" as value for `rules` to the configuration resolver, however people may use the configuration outside of the application so may pass it.

keradus:
Travis needs fixing
I fixed one of the warnings, leving 2nd for you @SpacePossum

SpacePossum:
thanks @keradus 
test update and rebased on 2.2

tests/Console/Command/FixCommandTest.php
```diff
@@ -52,7 +52,7 @@ public function testEmptyRulesValue()
      */
     public function testEmptyFormatValue()
     {
-        $cmdTester = $this->doTestExecute(array('--using-cache' => 'not today'));
+        $cmdTester = $this->doTestExecute(array('--using-cache' => 'not today', '--rules=switch_case_semicolon_to_colon'));
```
keradus: btw, why passing `--rules` fixing the issue?

SpacePossum: the test check the result of fixing this file, which might have diff. with the default set on diff. PHP versions

keradus: well, that change didn't fix travis.

keradus:
not sure what is the issue, works locally:
```
~/github/PHP-CS-Fixer$ docker run -it --rm --name my-running-script -v "$PWD":/usr/src/myapp -w /usr/src/myapp php:7.2.0RC1-cli php vendor/phpunit/phpunit/phpunit tests/Console/Command/FixCommandTest.php
PHPUnit 5.7.21 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.0RC1
Configuration: /usr/src/myapp/phpunit.xml.dist

Testing PhpCsFixer\Tests\Console\Command\FixCommandTest
..                                                                                                                              2 / 2 (100%)

Time: 109 ms, Memory: 10.00MB

OK (2 tests, 4 assertions)
```

SpacePossum:
fixed now, commit history is a bit of a mess, I'll clean up later

keradus:
Thank you @SpacePossum.

Issue #3028, opened 2017-09-07
Incompatible with PHPUnit 6.3.0


inoas:
When requiring `PHPUnit 6.3.0` it will require `sebastian/diff ^2.0` whereas even the latest `PHP-CS-Fixer` still requires ` "sebastian/diff": "^1.4",`

SpacePossum:
duplicate of https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2993

PR #3029, opened 2017-09-07
Allow sebastian/diff ^2.0

inoas:
Solves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3028


SpacePossum:
Hi and thanks for the PR.

However this cannot be merged because it would case a BC break within our project as `diff 2` generates a different out than `diff 1`.
We are in the process of resolving this issue using a different approach.

Please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2998 and the referred PR's and issues if you want to know more about the background of this issue.

PR #3031, opened 2017-09-07
IndentationTypeFixer - fix handling tabs in indented comments

### LINKED ISSUE #3030: Last tabs are changed to 4 spaces in nested arrays in multi-line comments
The PHP version you are using: `7.1.8`

PHP CS Fixer version you are using: `2.5.0`

The command you use to run the fixer (and the configuration file you are using if any).

`vendor/bin/php-cs-fixer fix`

A minimal sample of valid PHP code that is not fixed correctly (if applicable).

(indent tabs)

```
<?php

return [
	[
		/*
		 * Nice
		 */
		'hi',
	],
];
```

Config:

```
<?php

use PhpCsFixer\Config;
use PhpCsFixer\Finder;

return Config::create()
	->setIndent("\t")
	->setRules([
		'@Symfony' => true,
		'ordered_imports' => true,
	])
	->setFinder(
		Finder::create()
			->in(__DIR__)
	);
```

The fixed version of that code after you run the fixer and the version you expected.

Fixed: (might want to view whitespace)

```
<?php

return [
	[
		/*
	     * Nice
	     */
		'hi',
	],
];
```

Expected not to change.

keradus:
closes #3030

### DIFF
--- a/src/Fixer/Whitespace/IndentationTypeFixer.php
+++ b/src/Fixer/Whitespace/IndentationTypeFixer.php
@@ -60,6 +60,8 @@ final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwa
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
+        $expectedIndent = $this->whitespacesConfig->getIndent();
+
         foreach ($tokens as $index => $token) {
             if ($token->isComment()) {
                 $content = preg_replace('/^(?:(?<! ) {1,3})?\t/m', '\1    ', $token->getContent(), -1, $count);
@@ -70,7 +72,9 @@ final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwa
                 }
 
                 // change indent to expected one
-                $content = preg_replace('/^    /m', $this->whitespacesConfig->getIndent(), $content);
+                $content = preg_replace_callback('/^(?:    )+/m', function ($matches) use ($expectedIndent) {
+                    return str_replace('    ', $expectedIndent, $matches[0]);
+                }, $content);
 
                 $tokens[$index] = new Token(array($token->getId(), $content));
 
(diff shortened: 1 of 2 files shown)


src/Fixer/Whitespace/IndentationTypeFixer.php
```diff
@@ -70,7 +72,9 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 }
 
                 // change indent to expected one
-                $content = preg_replace('/^    /m', $this->whitespacesConfig->getIndent(), $content);
+                $content = preg_replace_callback('/^(?:    )+/m', function ($matches) use ($expectedIndent) {
```
SpacePossum: Does this play nicely with the double space indentation option?
Like doesn't this fix 2x4 spaces to 2x2 spaces and on second run the 2x2 spaces to 2 spaces?

keradus: no clue, I didn't touch 2x spaces here, not sure do we have tests for those.
maybe let us drop 2x spaces support in general? I remember I added tests with tabs for all whitespace-touching fixers and done integration tests for them, but probably noone has run it for 2 spaces

SpacePossum: I recalled this https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2650 which seems related to this issue as well.

keradus: let us not fix everything at once. until we will have at least integration test of 2 spaces, I can't say anything about quality of supporting it.

SpacePossum: all right, I'll only check if you fix causes regression with two-space config than, :+1:

keradus: green ?

SpacePossum: I didn't find any new issue with the double spaces, so :+1:  from me

PR #3036, opened 2017-09-10
Drop not existing Standalone group from PHPUnit configuration and duplicated internal tags

keradus:


src/ToolInfo.php
```diff
@@ -25,14 +25,8 @@
 {
     const COMPOSER_PACKAGE_NAME = 'friendsofphp/php-cs-fixer';
 
-    /**
-     * @internal
-     */
```
keradus: already on class-level

PR #3037, opened 2017-09-11
Update SF rule set

SpacePossum:


tests/Fixtures/Integration/set/@Symfony.test-out.php
```diff
@@ -101,7 +101,18 @@ class FooBarTest extends \PHPUnit_Framework_TestCase implements Test1Interface,
...
 }
+
+final class finalClass
+{
+    public function finalMethod()
+    {
+    }
+}
+
+echo 1;
```
keradus: changes shall be done also in `Symfony_whitespaces*` files (basically, files are the same except whitespaces)

keradus:
If one cannot fix one case of rule violation of one rule and on top of that claims project is not following that rule, we shall probably remove `protected_to_private`
https://github.com/symfony/symfony/pull/24136

keradus:
also, probably good time to inject:
```
        'no_unreachable_default_argument_value' => false,
        'braces' => array('allow_single_line_closure' => true),
        'heredoc_to_nowdoc' => false,
```

SpacePossum:
I thing `protected_to_private` is an issue on the SF codebase only, not in the whole eco-culture of SF projects (like Silex, Twig, etc.). So it might be good to keep promoting it.

SpacePossum:
``
        'no_unreachable_default_argument_value' => false,
        'braces' => array('allow_single_line_closure' => true),
        'heredoc_to_nowdoc' => false,
```

IIRC `heredoc_to_nowdoc` was reverted on request by SF community, or rejected.

`no_unreachable_default_argument_value` didn't work out because of reflection (argument guessing on routing IRIC)

:+1: for `braces` config tweaking

File: tests/Fixtures/Integration/set/@Symfony.test-out.php (hunk 1 above)
Comment (SpacePossum) [reply tests/Fixtures/Integration/set/@Symfony.test-out.php]: :+1: overlooked those

**Comment by keradus** [comment]
thus they shall fix `protected_to_private` case on their repo ;P ok, let keep it for now.


`no_unreachable_default_argument_value` is preventing `ParamConverter`s to work, which a lot of SF apps rely on, I would drop it from ruleset.
`heredoc_to_nowdoc`, as rejected, shall also be dropped from ruleset.

**Comment by SpacePossum** [comment]
```
no_unreachable_default_argument_value is preventing ParamConverters to work, which a lot of SF apps rely on, I would drop it from ruleset.
heredoc_to_nowdoc, as rejected, shall also be dropped from ruleset.
```

these are not in the set, so no update needed

>thus they shall fix protected_to_private case on their repo ;P ok, let keep it for now.

no update needed

>'braces' => array('allow_single_line_closure' => true),

already in the set, no update needed

tests/Fixtures/Integration/set/@Symfony.test-out.php
SpacePossum: updated

keradus:
Thank you @SpacePossum.

PR #3038, opened 2017-09-11
Support PHP 7.2

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2907

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.3.6 || >=7.0 <7.2",
+        "php": "^5.3.6 || >=7.0 <7.3",
         "ext-json": "*",
         "ext-tokenizer": "*",
         "composer/semver": "^1.4",
--- a/tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
+++ b/tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
@@ -202,7 +202,7 @@ final class NoLeadingImportSlashFixerTest extends AbstractFixerTestCase
      * @param string      $expected
      * @param null|string $input
      *
-     * @dataProvider provideFix54Cases
+     * @dataProvider provideFix56Cases
      * @requires PHP 5.6
      */
     public function testFix56($expected, $input = null)
--- a/tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
+++ b/tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
@@ -59,6 +59,14 @@ final class NoLeadingImportSlashFixerTest extends AbstractFixerTestCase
             ),
             array(
                 '<?php
+                use/*1*/A\B;
+                ',
+                '<?php
+                use/*1*/\A\B;
+                ',
+            ),
+            array(
+                '<?php
                 $a = function(\B\C $a) use ($b){
 
                 };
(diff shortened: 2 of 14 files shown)


tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
```diff
@@ -47,6 +47,14 @@ public function provideFixCases()
             ],
             [
                 '<?php
+                use/*1*/A\B;
```
SpacePossum: fixes could/should be backport to lower branches

composer.json
```diff
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.3.6 || >=7.0 <7.2",
+        "php": "^5.3.6 || >=7.0 <7.3",
```
keradus: just because of this bump, do we know that integration test of 7.2 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/tests/Fixtures/Integration/misc/PHP7_2.test) contains all syntax changes added in 7.2 ?

SpacePossum: I couldn't find any more new syntax changes, but a second look wouldn't hurt.
There are more opportunities for fixing towards PHP7.2, but I couldn't find more breaking issues.

keradus:
ref https://github.com/SpacePossum/PHP-CS-Fixer/pull/19

composer.json
keradus: ref https://github.com/SpacePossum/PHP-CS-Fixer/pull/20

keradus:
@SpacePossum , thanks for picking up PR from me
you still have todo in opening comment. Is it still valid after my config grooming ?

SpacePossum:
@keradus thanks for the PR. This resolved the TODO item, I've removed it.

keradus:
:+1:


@julienfalque 
as we do have me and @SpacePossum as authors, could you double check this PR please ?

tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
```diff
@@ -189,4 +197,70 @@ class Bar {
             ),
         );
     }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFix54Cases
```
julienfalque: Wrong data provider.

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -1721,17 +1721,31 @@ public function provideFix70TypesOrderAndAlphabetCases()
     /**
      * @param string      $expected
      * @param null|string $input
+     * @param array       $config
```
julienfalque: `null|array`

tests/Fixer/Import/SingleLineAfterImportsFixerTest.php
```diff
@@ -425,6 +425,18 @@ class Test {}
...
 ',
             ),
+            array(
+                ' <?php
+use some\a\ClassA;
+use function some\a\fn;
+use const some\c;
+
+',
+
```
julienfalque: Extra blank line.

tests/Fixer/Import/NoLeadingImportSlashFixerTest.php
keradus: done
btw, we need to improve data provider matching

tests/Fixer/Import/OrderedImportsFixerTest.php
keradus: done

tests/Fixer/Import/SingleLineAfterImportsFixerTest.php
keradus: done

tests/Fixer/Import/OrderedImportsFixerTest.php
```diff
@@ -1721,7 +1721,7 @@ public function provideFix70TypesOrderAndAlphabetCases()
     /**
      * @param string      $expected
      * @param null|string $input
-     * @param array       $config
+     * @param null}array  $config
```
julienfalque: Nope :p

keradus: done

keradus:
Thank you @SpacePossum.

keradus:
really great work here, longly awaited @SpacePossum 1

tests/Fixer/Import/NoUnusedImportsFixerTest.php
```diff
@@ -683,4 +683,47 @@ public function testWithSameNamespaceImportAndUnusedImportAfterNamespaceStatemen
...
+     */
+    public function testFix72($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFix72Cases()
+    {
+        return array(
+            array( // TODO test shows lot of cases where imports are not removed while could be
```
keradus: somehow we didn't spot it.
@SpacePossum could you take a look at it ?

SpacePossum: This is done on purpose; so we know this case (and other cases like group import) are not fixed by the fixer while it could.
I still left the case in to prove what happens on PHP7.2, i.e. the fixer doesn't crash nor makes incorrect changes. This way we prevent regression.
However, when someone wants to patch up the fixer to support this still should fail and the TODO message will tell why.
I patched up the other fixers, but this one was kind of to big to fix by me ATM.

keradus: OK, good enough.
I notice the `todo` during merging 2.2->up, I didn't know the background here, that's why asking.

PR #3039, opened 2017-09-11
YodaStyleFixer - Fix echo case

SpacePossum:
.


keradus:
Thank you @SpacePossum.

PR #3040, opened 2017-09-11
Update composer.json

Bukashk0zzz:
Update sebastian/diff dependency version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "composer/semver": "^1.4",
         "doctrine/annotations": "^1.2",
         "gecko-packages/gecko-php-unit": "^2.0",
-        "sebastian/diff": "^1.4",
+        "sebastian/diff": "^1.4|^2.0",
         "symfony/console": "^3.2",
         "symfony/event-dispatcher": "^3.0",
         "symfony/filesystem": "^3.0",


SpacePossum:
Thanks for reporting.
Please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2993 and others.

Bukashk0zzz:
@SpacePossum So close this PR? How I understand for now there are no any working solution. Maybe for now just merge this?

SpacePossum:
sadly we cannot merge this, as the output between diff 2 and 1 slightly differs, beside diff 2 requires PHP7 and we ship with dependencies lower than that.

Bukashk0zzz:
@SpacePossum But this not problem. You can choose what version you need. If you haven't php 7 then diff 1 will be installed.

Also tests are passed that's mean that there no any break problems.

keradus:
v1 || v2 is not and issue for us for php5 vs php7 indeed.
issue is bc break of format produced by differ. yes, there is a bc break. yes, tests are not covering that.

SpacePossum:
when you download the `phar` release you cannot change the `diff` version we shipped with it (in a sane way).

Bukashk0zzz:
Thanks for explanation.

PR #1030, opened 2015-02-26
Added NewlineAfterOpenTagFixer

ceeram:
Adding new fixer: NewlineAfterOpenTag. refs #641

Issue #939, opened 2015-01-02
Join replaced by implode...


RobinDev:
... until here, no problem.

But when you have a function named `join`, PHP-CS-Fixer changes it name.

Eg. : 
```php
class Tagada
{
    public function join()
    {
        //...
    }
}
```
Will be transformed in : 
```php
class Tagada
{
    public function implode()
    {
        //...
    }
}
```

RobinDev:
Oups... I watch the code after posting this issue and it's resolved in v1.4.

Sorry.

PR #3047, opened 2017-09-12
feature: Add a `compare` command to show fixers' usage in current config

### LINKED ISSUE #2996: Command to list unspecified fixers
It is difficult to keep tracking of which fixers you already wrote in the `.php_cs` and which one you didn't, expecially if you use rulesets or a new version of the tool is released.

I propose a new command that gets `.php_cs` config, makes a diff with all the fixers available and shows which ones are not specified.

With a flag we can also print a `var_export` with all these unspecified fixers set to `false` that can be copy-pasted in the `.php_cs`, so the user can progressively set them to true, or leave them to false to be able to re-run the command in the future.

Suggestions on the name of this command and flags are welcome.

### LINKED ISSUE #3045: How to compare available fixers and the ones I actually use
PHPCS Fixer provides a lot of fixers.
The configuration can be also nested and customized.
New fixes also are added constantly.

So it becomes really hard to understand which fixers are actually really used and which ones, instead, aren't.

But there are some fixers that can be really useful.

Unfortunately discover those fixers is really hard as someone has to go fixer by fixer to check if it is used by some of the used groups or if it set in in the custom `.php_cs`.

So a command that provides a list of all available fixers comparing them with the ones actually used should be really useful, doesn't it?

Aerendir:
Fix #2996
Fix #3045

---

This command compares the builtin Fixers and compares them with the ones actually configured.

The result is the following table:

schermata 2017-09-12 alle 11 06 05


localheinz:
@Aerendir 

Nice! 

How about keeping rules sorted by name, though?

Slamdunk:
Nice WIP.

As I wrote in #2996, there should be a third column (or something similar) that points out whether a rule is specified or not, that is different from "in use":

#### Case 1: `new_shiny_rule` is enabled
```php
$config->setRules([
    'old_known_rule' => true,
    'new_shiny_rule' => true,
]);
```
It means: "I know the rule and I want it"

#### Case 2: `new_shiny_rule` is disabled
```php
$config->setRules([
    'old_known_rule' => true,
    'new_shiny_rule' => false,
]);
```
It means: "I know the rule and I don't want it"

#### Case 3: `new_shiny_rule` is unknown
```php
$config->setRules([
    'old_known_rule' => true,
]);
```
It means: "I didn't know about the rule"

The current implementation can't distinguish Case 2 from Case 3.

julienfalque:
I was not fan of the idea but now that I see it, I actually like it :)

Aerendir:
@localheinz Done: now the fixers are sorted alphabetically!

But I have a problem o.O

To make the comparison I use `in_array()` without strict comparison.

But the current configuration has `'strict_param' => true,` and this changes the comparison this way:

    - in_array($fixer, $configured) ? "<fg=green;>\xE2\x9C\x94</>" : "<fg=red;>\xE2\x9C\x96</>",
    + in_array($fixer, $configured, true) ? "<fg=green;>\xE2\x9C\x94</>" : "<fg=red;>\xE2\x9C\x96</>",

This change breaks the code: any ideas about how to solve?

@Slamdunk I understand your point and I think it maybe useful. But currently I have no idea of how to implement it :) Give me some time to think deeper to the problem.

@localheinz I don't understand your last comment: can you explain me better?

@julienfalque Happy to read this from you :)

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,117 @@
...
+        $table->setHeaders([
+            [new TableCell(sprintf('Found <fg=yellow;>%s built-in</> fixers. Of those, <fg=yellow;>%s are configured</> to actually be used.', count($builtIn), count($configured)), ['colspan' => 3])],
+            ['Fixer', 'In use', 'Is risky'],
+        ]);
+
+        $body = [];
+        foreach ($builtIn as $fixer) {
+            $body[] = [
+                $fixer->getName(),
+                in_array($fixer, $configured) ? "<fg=green;>\xE2\x9C\x94</>" : "<fg=red;>\xE2\x9C\x96</>",
```
keradus: answering your question,

we don't find comparing object instances with `==` as good practice.
change it to `in_array($fixer->getName(), $configuredFixerNames, true)`

Slamdunk:
@Aerendir for the issue I underlined, you should use:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/5642a36a60c11cdd01488d192541a89bb44a4abf/src/Console/ConfigurationResolver.php#L454

And compare it to the names list of the rules found in the FixerFactory

src/Console/Command/CompareCommand.php
Aerendir: Yes, I don't see it as a good practice, too.

But in this particular case, a non strict comparison is the only way I know.

Comparing strictly don't work.

But using the strict comparison, the check fails as the two objects (the one in the configured fixers and the one in the fixers list) aren't the same object.

    // From fixers list
    PhpCsFixer\Fixer\Phpdoc\AlignMultilineCommentFixer {#592
        -tokenKinds: array:1 [
            0 => 378
        ]
      #configuration: array:1 [
        "comment_type" => "phpdocs_only"
      ]
      #whitespacesConfig: PhpCsFixer\WhitespacesFixerConfig {#35
        -indent: "    "
        -lineEnding: "\n"
      }
      -configurationDefinition: PhpCsFixer\FixerConfiguration\FixerConfigurationResolver {#593
        -options: array:1 [
            0 => PhpCsFixer\FixerConfiguration\FixerOption {#595
            -name: "comment_type"
            -description: "Whether to fix PHPDoc comments only (`phpdocs_only`), any multi-line comment whose lines all start with an asterisk (`phpdocs_like`) or any multi-line comment (`all_multiline`)."
            -default: "phpdocs_only"
            -isRequired: false
            -allowedTypes: null
            -allowedValues: array:3 [
                0 => "phpdocs_only"
              1 => "phpdocs_like"
              2 => "all_multiline"
            ]
            -normalizer: null
          }
        ]
        -registeredNames: array:1 [
            0 => "comment_type"
        ]
      }
    }

    // From configured fixers
    PhpCsFixer\Fixer\Phpdoc\AlignMultilineCommentFixer {#939
        -tokenKinds: array:1 [
            0 => 378
        ]
      #configuration: array:1 [
        "comment_type" => "phpdocs_only"
      ]
      #whitespacesConfig: PhpCsFixer\WhitespacesFixerConfig {#1030
        -indent: "    "
        -lineEnding: "\n"
      }
      -configurationDefinition: PhpCsFixer\FixerConfiguration\FixerConfigurationResolver {#940
        -options: array:1 [
            0 => PhpCsFixer\FixerConfiguration\FixerOption {#942
            -name: "comment_type"
            -description: "Whether to fix PHPDoc comments only (`phpdocs_only`), any multi-line comment whose lines all start with an asterisk (`phpdocs_like`) or any multi-line comment (`all_multiline`)."
            -default: "phpdocs_only"
            -isRequired: false
            -allowedTypes: null
            -allowedValues: array:3 [
                0 => "phpdocs_only"
              1 => "phpdocs_like"
              2 => "all_multiline"
            ]
            -normalizer: null
          }
        ]
        -registeredNames: array:1 [
            0 => "comment_type"
        ]
      }
    }

Any ideas about how to solve this?

Is worth the time to use a workaround?

keradus: As stated in my comment:
```
change it to in_array($fixer->getName(), $configuredFixerNames, true)
```

so, use **name** instead of full **object**

keradus: @Aerendir or use https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/RuleSet.php#L240

Aerendir: Ok, done.

Aerendir:
I've added a command to hide the already configured fixers.

Using

    php-cs-fixer compare --hide-configured

the result is this

schermata 2017-09-12 alle 13 13 49

@Slamdunk did you mean to add the `['old_known_rule' => true, 'new_shiny_rule' => false,]` configs to the `.php_cs` file?

localheinz:
@Aerendir 

Ha, I'm trying to tell you that in the referenced tests, I try to compare the available built-in fixers with the ones actually configured. For a custom configuration I want to make sure I have all of the available built-in fixers configured (even if I don't use them), just to make sure I am aware of all built-in fixers. 

On second thought, maybe it makes also sense to group fixers 

- in a first block, show all built-in fixers in use
- in a second block, show all built-in fixers not in use

But, just an idea, not sure if it makes things easier for anyone.

Aerendir:
@localheinz , let me recap to be sure I'm understanding.

You are saying that if you want to use only, say, 50 fixers you want to anyway configure the remaining 104 (setting them in a way they basically o nothing)?

Your goal which is? If I'm understanding well, is to not miss new fixers.

So, if currently there are 154 built-in fixers and tomorrow other 3 are added, you want to know which are those newly added 3 fixers?

Am I understanding well?

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,157 @@
...
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
+                    new InputOption('show-risky', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.', false),
```
Slamdunk: Description of `show-risky` should be fixed

Aerendir: Done!

Aerendir:
@Slamdunk , your issue is similar to the @localheinz 's one? Or is it different?

Becuase reading again the entire discussion, I'm a little bit confused about who are asking what!

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,157 @@
...
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
+                    new InputOption('show-risky', '', InputOption::VALUE_REQUIRED, 'Shows also the riky fixers.', false),
```
Slamdunk: _riky_ typo.
I suggest a different description: `Shows additional flag about rules riskiness`

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,157 @@
...
+                    new InputOption('hide-configured', '', InputOption::VALUE_OPTIONAL, 'Hides all the rules currently configured to highlight only the ones not already in use.', false),
```
Slamdunk: It seems to me that both `show-risky` and `hide-configured` should be `InputOption::VALUE_NONE`.
Only if done this way, you can safely call them on boolean checks like
```php
if ($input->getOption('hide-configured')) { /* ... */ }
```
instead of
```php
$hideConfigured = null === $input->getOption('hide-configured') ? true : false;
```

Aerendir: Good advice @Slamdunk ... Done!

src/Console/Command/CompareCommand.php
Aerendir: Ok, I'll fix the typo.

About your description, it is misleading: if set, the flag will show also the risky fixers. If not set, the risky fixers don't compare in the list at all.

So, describing as you suggest, makes me think that if I don't pass the flag, anyway I see also the risky fixers, but this is not true.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,150 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
```
keradus: If I won't provide the config, I expect that command would automatically pick same config as `FixCommand` would.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,150 @@
...
+                    new InputOption('show-risky', '', InputOption::VALUE_NONE, 'Shows also the riky fixers.'),
```
keradus: they shall not be hidden by default.

Aerendir: Are you speaking about risky tests? I'm following the logic of the `FixCommand` where you have to explicitly allow the use of them.

src/Console/Command/CompareCommand.php
Aerendir: Ok, I'll give this a look.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+                : implode('\' => false,'."\n".'\'', $dump);
+
+            $output->writeln(
+                empty($dump)
+                    ? $line
+                    : "\nCopy and paste the following rules in your .php_cs file:\n\n"
+                    .'\\\\ Below the rules I don\'t want to use'."\n"
+                    .'\''.$line.'\' => false'."\n"
+                    .'\\\\ END Rules to never use'."\n"
+            );
```
Slamdunk: I appreciate the new `dump` flag, but do not generate manually the code please.
Instead, fill a new PHP array with `$dump` values as keys, `false` as values and then use the native [var_export](https://secure.php.net/manual/en/function.var-export.php) function; this would be much less error-prone.

Aerendir: Ok, I'll give this too a look. Thanks for the advice!

src/Console/Command/CompareCommand.php
keradus: allowing usage needs dedicated confirmation, every single risky rule is changing the behaviour of code.
no need to hide it be default in list

keradus:
@Aerendir this PR deserves discussion what we want to achieve. If you don't want to spent a lot of time with changing implementation again and again, let us first discuss what we want to achieve and how, then let us implement it once.

Slamdunk:
@keradus it seems to me that this topic _has_ been discussed.
#2996 is three weeks now without replies; @julienfalque and @localheinz requested here some changes that denote this PR intentions are good, and you did something similar, without refusing or hindering the result.

@Aerendir did a good starting point, is improving fast, and [Cunningham's Law](https://en.wikipedia.org/wiki/Ward_Cunningham#Cunningham.27s_Law) proves that this feature can reach a final point soon if @Aerendir is willing to continue his commitment.

keradus:
> #2996 is three weeks now without replies

it indicate not that topic is discussed fully, but lack of big interest about it


If you prefer going around, sure :+1:
be aware I only looked at technical issue you raised about adding `, true` to `in_array`, not about whole concept and created output. not yet.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
+                    new InputOption('show-risky', '', InputOption::VALUE_NONE, 'Shows also the riky fixers.'),
+                    new InputOption('hide-in-use', '', InputOption::VALUE_NONE, 'Hides all the rules currently configured (and so, used) to highlight only the ones not already in use.'),
+                    new InputOption('dump', '', InputOption::VALUE_NONE, 'Dumps the comparing result in a copy-and-pastable format ready for the .php_cs file.'),
+                ]
+            )
+            ->setDescription('Compares existent features with the ones actually configured.')
```
keradus: features -> rules

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
+                    new InputOption('show-risky', '', InputOption::VALUE_NONE, 'Shows also the riky fixers.'),
+                    new InputOption('hide-in-use', '', InputOption::VALUE_NONE, 'Hides all the rules currently configured (and so, used) to highlight only the ones not already in use.'),
```
keradus: `show-risky` but `hide-in-use`

src/Console/Command/CompareCommand.php
keradus: features -> rules

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+                'config' => $passedConfig,
+            ],
+            getcwd()
+        );
+
+        $inUse = $resolver->getRules();
+        $configured = $resolver->getConfig()->getRules();
+        $builtIn = $this->fixerFactory->getFixers();
+
+        usort($builtIn, function ($a, $b) {
```
keradus: type-hinting is missing

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+            [
+                'allow-risky' => true,
+                'config' => $passedConfig,
+            ],
+            getcwd()
+        );
+
+        $inUse = $resolver->getRules();
+        $configured = $resolver->getConfig()->getRules();
+        $builtIn = $this->fixerFactory->getFixers();
```
keradus: please use nouns in names

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+        foreach ($builtIn as $fixer) {
+            ++$builtInCount;
+
+            if ($fixer->isRisky() && false === $input->getOption('show-risky')) {
+                // Don't show risky fixers if required
+                continue;
+            }
+
+            $isUsed = array_key_exists($fixer->getName(), $inUse);
+            $isConfigured = array_key_exists($fixer->getName(), $configured);
```
keradus: isn't `isset` way faster?

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
+                    new InputOption('show-risky', '', InputOption::VALUE_NONE, 'Shows also the riky fixers.'),
+                    new InputOption('hide-in-use', '', InputOption::VALUE_NONE, 'Hides all the rules currently configured (and so, used) to highlight only the ones not already in use.'),
+                    new InputOption('dump', '', InputOption::VALUE_NONE, 'Dumps the comparing result in a copy-and-pastable format ready for the .php_cs file.'),
```
keradus: :-1: this looks wrong

code shall filter by sth, here, by `isUsed` and `isConfigured`, which will be general solution to filter what we want to display together with replacement for current `hide-in-use` and `show-risky` options.

then, code shall provide different output formats, like a table by default and json as an possibility

tests/Console/Command/CompareCommandTest.php
```diff
@@ -0,0 +1,70 @@
...
+    protected function setUp()
+    {
+        $this->application = new Application();
+    }
+
+    public function testCompareCommand()
+    {
+        $cmdTester = $this->doTestExecute();
+
+        $this->assertSame(0, $cmdTester->getStatusCode(), "Expected exit code mismatch. Output:\n".$cmdTester->getDisplay());
```
keradus: cool test.
nice start for sure

yet, whole content of `execute` is not tested at all

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+
+        $rows = [];
+        $dump = [];
+        $builtInCount = 0;
+        foreach ($builtIn as $fixer) {
+            ++$builtInCount;
+
+            if ($fixer->isRisky() && false === $input->getOption('show-risky')) {
+                // Don't show risky fixers if required
+                continue;
```
keradus: getting info, filtering, getting info, filtering, getting info, filtering. this jumping is not nice

i would rather collect all needed data, it's cheep, and then filter whatever is interesting for end-user in one place

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            $isUsed = array_key_exists($fixer->getName(), $inUse);
```
keradus: if user has ruleset `[@Symfony => true, other_rules => ...]` and then see that he is using `braces` rule, he has no clue why.
we shall display where the rule "in use" comes from, like it is active as user requested it directly or it's part of set he is using

src/Console/Command/CompareCommand.php
Slamdunk: > then, code shall provide different output formats, like a table by default and json as an possibility

This whole purpose of this command is to have an evidence of current PHP-CS-Fixer version and the user config. What would be the added value of dumping in json? I see no case where this command is related to anything outside PHP world, and I've never seen a `.php_cs` file written in json or with json-related data.

The table format is the default, the `dump` version is to enable a copy-pasteble code to be inserted in the user `.php_cs`

> code shall filter by sth

@keradus is right, the `dump` flag must mirror the output of the command executed without the flag

keradus: what I mean is to pick up a format and output in it, not randomly dump php code at the end ;)

src/Console/Command/CompareCommand.php
Aerendir: Yes, this is a very good point. I were thinking at it yesterday, but I have no idea where to get this info: can you give me some advice?

PS I'll ignore all other points as I'm completely refactoring the command. Once I'll commit the modifications, we'll can start to improve it...

Aerendir:
Ok, I've completely rewritten the command.

After the "experiments" of yesterday and all the feedbacks you gave me, I've refactored the code to be more readable and more maintainable (at least, I hope!).

Anyway, now the default behavior is to show all the `Fixer`s:

    php-cs-fixer compare

schermata 2017-09-13 alle 17 19 46

Now the table shows 4 columns (and not only 3 as yesterday did):

1. The `Fixer`'s name;
2. If it is configured (`true` if it is explicitly set in the `.php_cs`);
3. If it is effectively used (`false` if it is explicitly set in the `.php_cs` as `'fixer_name' => false`);
4. If it is risky.

Then, it is possible to not show some fixers using the appropriate tags:

    php-cs-fixer compare --hide-configured --hide-enabled --hide-risky

schermata 2017-09-13 alle 17 24 01

Those are fixers that exist but that are never used, nor by some of the `RuleSet`s I've configured, nor by directly enabling including them in the `.php_cs`.

At this point, as @keradus pointed out, it will be useful to know why some fixers are enabled: they are enabled because of the inclusion in some of the `RuleSet`s I've configured, but which?

**Currently I have no idea about how to get this information: any advice will be really appreciated!**

Having this information will make us able to get another column `is_inherited` and also another one `inheritance path` where the full inheritance path is exposed (if this may be useful).

I'm not sure this is really useful, as once I've discovered a rule is enabled, I'm only intrested in deciding if I want it be enabled or not: if not I simply include it in the `.php_cs` setting it to `false` (or configuring it in a way more suitable for my needs).

Recapping:

1. All `Fixer`s are shown by default;
2. It is possible to hide some `Fixer`s using `--hide-configured`, `--hide-enabled` and `--hide-risky`;
3. Lets discuss about the opportunity of displaying also the `RuleSet`s that makes a `Fixer` configured.

For the moment, I'm coming back to work! :P

Aerendir:
PS

Next thing I'll implement is the ability to filter by negations: `--hide-not-configured`, `--hide-not-enabled` and `--hide-not-risky`: this will make us able to further filter the shown `Fixer`s.

Slamdunk:
> At this point, as keradus pointed out, it will be useful to know why some fixers are enabled: they are enabled because of the inclusion in some of the RuleSets I've configured, but which?
> Currently I have no idea about how to get this information: any advice will be really appreciated!

I think there is no real need to know if the _user_ set or not a ruleset; it would be enough to know whether a rule is inside a ruleset. In this case we can do the same done by `FixCommand` and its helper:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/5ab1424fc372a97ee37e521239896b693cb5dd6c/src/Console/Command/HelpCommand.php#L420-L435

Aerendir:
@Slamdunk , great! I'll give it a look ASAP. Thank you!

Aerendir:
Done! Now the command supports inheritance.

schermata 2017-09-14 alle 13 46 58

And also the hiding of inherited commands

schermata 2017-09-14 alle 13 47 19

src/Console/Command/CompareCommand.php
```diff
@@ -198,24 +204,16 @@ private function buildTable(OutputInterface $output)
...
-            $this->hideConfigured ? self::YES : self::NO,
-            $this->hideEnabled ? self::YES : self::NO,
-            $this->hideRisky ? self::YES : self::NO
-        );
+        $columns = [
+            sprintf('Fixer (%s)', count($this->builtInFixers)),
+            sprintf("Configured (%s)\nAre hided: %s", count($this->configuredFixers), $this->hideConfigured ? self::YES : self::NO),
+            sprintf("Enabled (%s)\nAre hided: %s", count($this->enabledFixers), $this->hideEnabled ? self::YES : self::NO),
+            sprintf("Risky (%s)\nAre hided: %s", count($this->riskyFixers), $this->hideRisky ? self::YES : self::NO),
+            sprintf("Inherited (%s)\nAre hided: %s", $this->countInherited, $this->hideInherited ? self::YES : self::NO),
```
jenkoian: `hided` isn't really a word, should be `Are hidden`

Aerendir: You are right... Fixed!

src/Console/Command/CompareCommand.php
Aerendir: Solved: risky tests are shown by default.

src/Console/Command/CompareCommand.php
Aerendir: Solved

src/Console/Command/CompareCommand.php
Aerendir: Solved: the `--dump` flag dumps exactly what is shown in the table and uses `var_export` to generate a valid array that is copy-and-pastable.

src/Console/Command/CompareCommand.php
Aerendir: Fixed.

src/Console/Command/CompareCommand.php
Aerendir: Done.

src/Console/Command/CompareCommand.php
Aerendir: Fixed.

src/Console/Command/CompareCommand.php
Aerendir: Fixed.

src/Console/Command/CompareCommand.php
Aerendir: Done.

src/Console/Command/CompareCommand.php
Aerendir: I have never tested their speed, but reading the code, also other commands makes use of this solution intead of using `isset`. Anyway, I don't think is something that deserves a lot of attention.

src/Console/Command/CompareCommand.php
Aerendir: Done.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+    }
+
+    /**
+     * @param FixerInterface $fixer
+     *
+     * @return bool
+     */
+    private function isFixerConfigured(FixerInterface $fixer)
+    {
+        return array_key_exists($fixer->getName(), $this->configuredFixers);
```
keradus: depends on scenario, `array_key_exists` is 4-10k (yes, k!) times slower. don't use it if you don't need to

Aerendir: Just because I'm curious: have you any test to prove this? Anyway I trust you, so I'm going to change this.

tests/Console/Command/CompareCommandTest.php
```diff
@@ -0,0 +1,70 @@
...
+ * @internal
+ *
+ * @covers \PhpCsFixer\Console\Command\CompareCommand
+ */
+final class CompareCommandTest extends TestCase
+{
+    /**
+     * @var Application
+     */
+    private $application;
```
julienfalque: No need for a property here: it's used only once in `doTestExecute()`, you could use a regular variable there instead.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+    /**
+     * @param OutputInterface $output
+     *
+     * @return Table
+     */
+    private function buildTable(OutputInterface $output)
+    {
+        $table = new Table($output);
+
+        $columns = [
```
julienfalque: The table is starting to get quite large. Maybe we could reduce the number of columns.

I would suggest to replace _Configured_ and _Enabled_ (and maybe _Inherited_ too?) columns with a simple symbol before the fixer name, e.g.:
```
✓ fixer_enabled_explicitly
✖ fixer_not_enabled
? fixer_enabled_implicitly_by_a_ruleset
```
(no idea what would be the symbol for the third case).

WDYT?

src/Console/Command/CompareCommand.php
keradus: DIY: https://www.google.pl/search?q=isset+vs+array_key_exists+performance&oq=isset+vs+array_key_exists+performance&aqs=chrome..69i57j69i60l3.6926j0j7&sourceid=chrome&ie=UTF-8

a lot of analysis were executed. you need to get outcome of them with caution - a performance difference depends very much on given test, dataset and user php version.
pick one and execute yourself, eg on 3v4l.
one thing is constant - `isset` is **always** faster

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+         * @var RuleSet $set
+         */
+        foreach ($this->ruleSets as $name => $set) {
+            /**
+             * @var string
+             * @var bool   $value
+             */
+            foreach ($set->getRules() as $rule => $value) {
+                if (false === isset($this->list[$rule]['in_set'])) {
+                    $this->list[$rule]['in_set'] = [];
```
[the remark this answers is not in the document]
keradus: please keep initialization.
autovivification is bad practice, could raise warnings as well (depends on php settings)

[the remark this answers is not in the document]
julienfalque: Which settings would make PHP trigger warnings in such case?

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Adamo Aerendir Crespi <hello@aerendir.me>
+ */
+final class CompareCommand extends Command
+{
+    const COMMAND_NAME = 'compare';
```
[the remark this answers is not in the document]
Aerendir: I think that `list-fixers` doesn't makes the sense of the command: it actually compares, not limiting itself to list. It doesn't only list the built in fixers, but actually makes you able to compare the list of built-in fixers with the list of the ones you have configured to be enabled or not, making you able to "debug" your configuration. `list-fixers` is so limiting as name :(

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+            ->setDescription('Compares existent Fixers with the ones actually configured or enabled by inheritance.')
+        ;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $this->initialize($input, $output);
```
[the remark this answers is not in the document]
Aerendir: Yes, you are right: I'm going to remove this call...

src/Console/Command/CompareCommand.php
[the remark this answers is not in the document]
Aerendir: Yes, it may trigger warnings. Anyway, makes the code clearer. I prefer to keep this snippet as is.

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,323 @@
...
+            }
+        }
+
+        /** @var FixerInterface $fixer */
+        foreach ($this->builtInFixers as $fixer) {
+            if ($isInherited = !empty($this->list[$fixer->getName()]['in_set'])) {
+                ++$this->countInherited;
+            }
+
+            $this->list[$fixer->getName()]['name'] = $fixer->getName();
```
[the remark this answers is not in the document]
Aerendir: Yes, maybe: I like the suggestion... I called the method `processFixer` (because it actually doesn't simply create a row, but makes the checks to get the info about the `Fixer`) and added also another method `processRuleSet`: the `execute()` method is now cleaner.

src/Console/Command/CompareCommand.php
Aerendir: Yes, it is going to be quite large.

The problem is that the symbols are UNICODE and there are not a lot of them.

Here is a list of the available one: http://www.utf8-chartable.de/unicode-utf8-table.pl?start=9984&number=128&names=-&utf8=string-literal

Can you find one that is suitable to represent an inherited `Fixer`?

In the mean time a solution to consider is to put the inheritance path directly under the Fixer name (as is the path that makes the table really large)...

I'll make some variations to see how it will look like...

tests/Console/Command/CompareCommandTest.php
Aerendir: Not consider this: is just a draft to not get the red cross, it is not implemented yet :)

src/Console/Command/CompareCommand.php
julienfalque: Maybe we can always use `✓` when the fixer is enabled, and choose a different color depending on how it is enabled, e.g. green for explicitly configured fixers and yellow for inherited ones?

We could also replace symbols with actual words when verbosity level is increased? This would allow color blind people to still be able to get all the details.

Aerendir: Yes, I've done exactly this: green for Fixers configured in `.php_cs`, yellow for inherited, red for not enabled.

See the comment I'm going to write (be patient! :) )

Aerendir:
After the suggestions of @julienfalque , I've refactored a bit the code to build a thiner table.

This is the result:

schermata 2017-09-15 alle 10 49 52

schermata 2017-09-15 alle 10 50 11

As you can see, I've moved the inheritance path directly under the Fixer name.

The Fixer name is also in different colors:

- green if it is directly configured;
- yellow if it inherits;
- red if it is not enabled (independently by the fact it is configured or inherited)

THE PROBLEMs WITH COLORS

Colors are a great and fast way to highlight difference.
But there are two main drawbacks:

1. People with color blindness will not recognize them;
2. People with different color schemes set in their editor may have problems, too.

I don't know if this is a great problem or, instead, a simple trifle, something that we don't have to take care about: anyway, there are the columns.

What do you think about?

src/Console/Command/CompareCommand.php
Aerendir: I've just tried to run the command without a `.php_cs` file and it works.
What do you expect exactly? I think I'm not understanding your point here.

keradus: with using `ConfigurationResolver`, it's fine

src/Console/Command/CompareCommand.php
```diff
@@ -0,0 +1,355 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function configure()
+    {
+        $this
+            ->setName(self::COMMAND_NAME)
+            ->setDefinition(
+                [
+                    new InputOption('config', '', InputOption::VALUE_REQUIRED, 'The path to a .php_cs file.'),
```
keradus:
```diff
-.php_cs
+config
```

there could be multiple configs named differently

keradus:
colors are helper for 2nd colour, even if all would have same colour, all info are there.
don't treat it as an issue for sure.

keradus:
Proposal looks interesting, I need to play with it a bit to see how it's implemented and what's the output, yet...
You are connecting here rules with ruleset (which is great :+1:), yet we have in plan to rewrite the rulesets. 

For that, I would like to request once more writing utests for this great new command, so nobody will break it in future.

Aerendir:
@keradus , ok for colors...

About rulesets, so what should have I to do now? Have I to wait? For what?

In the mean time I've realigned the branch with the `master` branch, so it can be merged when ready...

Can you list me the next steps?

What should happen for the PR being merged into the `master` branch?

Aerendir:
PS
ci/circleci ever gives errors on the PR, but it seems is a problem of exhausted memory on the server part. What should have I to do to fix this issue (if something I can do on myself)...

Aerendir:
In the mean time I've also added icons to fixers' names, to distinguish them also visually and not only by colors: this way is way more clear the "nature" of each fixer.

schermata 2017-09-20 alle 16 33 42
schermata 2017-09-20 alle 16 33 53

Aerendir:
Ok, @keradus , let's test... but... I need some guidance :)

1. What should have I to test? I've given a look at existent tests of the other commands and they are very specific.
As this command creates a list that is dynamic as depends on the given configuration, I cannot set a fixed expected test: so what should have I to test? Any ideas from which I can start? What do you think may be some useful tests to be sure the command will not be broken in future?

2. You said I've increased the consumed memory by 30%... Ok, but how I did this?
I simply copied and pasted the test of FixCommandTest, removed all the code and using it as a "fake" test for this command. Can you give me some guidance about what increased the memory usage?

keradus:
1. created mocked input of `available rules` and provided `config` (probably multiple cases to check different things). Then, probably decouple the command to two parts - data preparation and data presentation. After that, you can verify if data prepared on top of mocked available_rules/config is matching expectations, and then if generated display output on top of given data matches expectations.

2. I investigated why CircleCI crashed, it's because huge memory consumption bump. I didn't investigate why it's bumped, you probably need to use some tools like xdebug, phpdbg, blackfire.io, sth...

julienfalque:
I think the memory usage increase is due to the fact that you call `$fixerFactory->registerBuiltInFixers();` in the command constructor. Also, the command stores a lot of things in its properties, maybe this uses too much memory too?

keradus:
ping @Aerendir

Aerendir:
@keradus I've read. I need some time to find a way to profile the code. I've never used blackfire.io nor profiled my code (something on the todo list) and this requires me some time to study... Be patient :)

keradus:
ok, sure thing !
In that case, could you please first focus on providing full utests, and the performance issue later?
(that way, we will also be confident that performance fix won't break feature itself)

Aerendir:
Ok @keradus , I will focus on them first... I need to better understand how to implement them, reading again and again your previous comments because I still have no idea of how to implement them... But I have to say also that I didn't think at them too much in the last days...

keradus:
ping @Aerendir  ;) just a friendly reminder :cake:

Aerendir:
Hi @keradus , excuse me for the late reply :( As you may have imagined I've had no time to study all the things required to write a test (performance issues and how to concretely test the command)... I'm sorry :(

keradus:
@Aerendir thanks for info !

@SpacePossum , @julienfalque , are you able to proceed with @Aerendir about this PR to have aligned and finished console output ? Then, we could handle tests on our side internally.

julienfalque:
I'd be happy to continue this PR when I have time to :)

keradus:
actually my idea is a bit different (If I get you correct @julienfalque)

let us finish review of current work in this PR, ask for changes if they are needed, yet let us ignore the tests part. then, after we will have what nice state of what we want to have in generated output, then we will add tests, probably as separated PR.

julienfalque:
Ok but I'd prefer to have tests in the same PR, it often helps spotting edge case bugs we might miss during review process.

keradus:
ok, sure. yet first, let us try to finalize what output we want to achieve, then we will cover it with tests, to not block the process, like we sadly just did for weeks :(

PowerKiKi:
I cannot find recent requests for changes regarding the output. Can we consider that everybody agree on the output and that we can proceed with unit tests ?

SpacePossum:
I would change the name from `compare` to `describe-config`.
The commands would than be;
```
  describe        Describe rule / ruleset.
  describe-config Describe configuration.
  fix             Fixes a directory or a file.
  help            Displays help for a command
  list            Lists commands
  readme          Generates the README content, based on the fix command help.
  self-update     [selfupdate] Update php-cs-fixer.phar to the latest stable version.
```

happy to hear opinions about this though, I think `compare` is kind of generic and made me think about comparing something related with files and CS.

I would hide the second row:
```
|                                                                         | Are hidden: ✖   | Are hidden: ✖ | Are hidden: ✖ | Are hidden: ✖   |

```

and only show it when something is filtered and only put `(filtered)` in the field if it is filtered and leave it empty if not.
I would hide the count number and only show when a higher verbosity flag is used.

I think it would be nice to show if a fixer:
- has been deprecated
- is a custom fixer

```
$ php php-cs-fixer compare -h
```

shows a few default options it gets from the SF default command which I think we don't need here;

```
  -q, --quiet            Do not output any message
  -V, --version          Display this application version
  -n, --no-interaction   Do not ask any interactive question
```

(same for `describe` command, I'll get that in another PR though)

I'm not sure about the `--dump` option TBH, I'll play around with it and come back to that.

So, I didn't want to go in by committing changes to my liking without discussion first. 
Please start by rebasing the branch. Than everyone that can help out please let us know here so we could coordinate, like helping with the utest or with something else etc.
Lets get this one going again and finish it :)
All in all, lot of very nice work!

PowerKiKi:
I would like to suggest `show` as command name for the following reasons:

- I feel it "shows" the current configuration more than it "describes" it
- it might be more familiar for composer users, so more naturally discoverable
- it avoids possible confusion between similar sounding describe commands
- it happens to be much shorter to type

scuben:
@Aerendir Would love to see this PR progressing. Do you need any help?

Aerendir:
@scuben at time, I had problems with testing.

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3047#issuecomment-331007978

My tests consume too much memory and I don't know how to find the bottleneck...

More, I have no idea of what to test concretely (currently there is only a simple "smoke" test)...

scuben:
What would be the best way to help with Tests or other adjustments to the code? 

Would you mind give write access to me so i can give it a shot (and revase first)?

Aerendir:
@scuben , I've given write access to my fork...

Unfortunately I don't know which is the best way to help with tests. Maybe the first step is to make them not consume too much memory as they currently do (causing circleci to fail).

scuben:
Thank you @Aerendir 

For the first part I renamed the command to `show`.

Following changes had been proposed to the functionality of the PR:

1. Hide the second row (information about what is hidden by command option)
2. Removing default command options `--quiet`, `--version` and `--no-interaction`
3. Add information about a fixer being deprecated or custom

Please give feedback about this three proposed changes. I personally would like to have Nr. 1 and 3 implemented. Nr. 2 seems like over-engineering also because it does not seems simple to achieve. Also `--version` will print the library version like on all other commands, `--no-interaction` will make sure there is no interaction (i know, it's always no-interaction, but this wont break anything having it around). And `--quite` will be quiet output like expected even no one would ever use it.

Next step for me would be:

- [ ] Gather your feedback about the proposed functionality
- [ ] Rebase on top of master
- [ ] Check and resolve any integration tests (fabbot, appveyor, circleci, travis-ci)

Aerendir:
@scuben , first thank you for the work you are doing on this PR: I'm very glad :)

Coming to your questions:

1. I've added it for clarity, but maybe that, given the current state of the output, it is non useful anymore;
2. I don't think that having those flags around is a big issue: maybe they can removed later (or we can simply intercept them in the command and throw an exception if they are explicitly passed);
3. Completely agree on this: it would be a great feature for the command!

During development I also thought at another feature, but it is quite complex I suspect: perform a diff between the old already existing `.php_cs` and the new one.

Practically, when updating PHPCSFixer there are often new fixers that no one knows are introduced (unless you read the `CHANGELOG`).

The command may output the new added fixers on the first run after a command update.

Obviously, to run the diff it is required that the `.php_cs` generated before the update still exists as it is required to perform the diff...

I didn't investigate further this feature and so I don't know practically how complex it is, but it is something very useful (and this is also the reason why the command is called `compare` and not `show`: the original goal is exactly this: show which fixers are added, which are removed, and from which rule sets they come).

Aerendir:
Maybe a better solution is to call the command `show` (as you did) and then add a `--compare` flag to perform the diff if possible (maybe on the first run after the update of the tool).

So, to be clear, doing something like this:

```console
$ vendor/bin/php-cs-fixer show --compare
```

scuben:
@Aerendir you're welcome!

I like to heave a possibility to dump new fixers or compare with the current .php_cs. But I would suggest removing the dump/compare from this PR and create a new one with this idea (maybe separate command or as an improvement for this one). WDYT?

Aerendir:
I think that to perform the comparison we need a lot of code that is in this command.

Practically the missed part is the loading of the current `.php_cs` file: all the rest of the code is the same.

I stopped on this because the tests didn't passed and I didn't know how to fix, then other things popped out on my daily work day and...

More, there was the possibility that the loading of the file changed during the development cycle and so this last missed feature never saw the light.

@keradus some months ago pointed out that [there was a plan to change the handling of the rule sets](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3047#issuecomment-329896494): I don't know how this impact the command (for this reason he also proposed to deeply test the command to avoid BC changes to other parts of the tool).

So, the first thing to do, before any consideration, is to understand how to load the `.php_cs` file.

Then we should think about the diff between the content of the `.php_cs` file and the rules loaded by this command. IMHO.

keradus:
It's great to see some movement here :heart:

Test is must-have indeed.

I wouldn't include performing the diff between 2 configs in this PR. If we want the diff, let us start with diff between currently loaded rules and all possible rules.

Aerendir:
@keradus , no, not between two configs, I wrote the wrong thing -.-'.

I think at this: I run the command each time before I commit.
Each run updates/creates the `.php_cs.cache` file that contains the list of files processed and the rules applied.

Now, given these rules, it is possible to compare them with the new one as modified by the update of the tool: maybe some rules are added, some others are removed, some others are added/removed by some rulesets...

This way is easier to understand what's going on...

PS
I'm going to update other my comments to fix the error about the file I'm intending to use...

keradus:
> Each run updates/creates the `.php_cs.cache` file that contains the list of files processed and the rules applied.

But exactly same rules are already in `.php_cs`, no need to take them from cache file...
But then, if config is `@Symfony`, then upgrade may add new rules to it, keeping config the same, but not the used rules...
That way we are hitting problem of not versioned rulesets, and for that let us fix it proper way - add versioning to rulesets.
But that's super different thing now.

IMO, let us focus in this PR on minimal feature and add more and more later, if not - we won't merge it anytime, as it won't be ready, ever. It's already half year old, isn't it ?
For that, if we want to have `compare` already in this PR, let it be `compare` between config file and "all rules" IMHO.

Later, we can add whatever feature for command we want, but let us not postpone the command indefinitely.

Aerendir:
@keradus 

> That way we are hitting problem of not versioned rulesets, and for that let us fix it proper way - add versioning to rulesets.

For me its ok, if this is the background, then it's better to release faster than waiting for the new compare feature.

After, it can always be added later (in PHPCSFixer itself or in this command)...

This told, are there any modification that should be made to the command? Or it is ready for testing and then merging?

scuben:
Good, so I'm going to remove the dump functionality then to have the features finished soon. @Aerendir I'll also squash the commits after we have all the features in place that we want. So if you want to save some code from the dump functionality you can save them now. Please give me a short info if you saved the data or if you don't want to save it.

The following list ist whats to be done next:

- [ ] Rebase or merge master
- [ ] Hide the second row (information about what is hidden by command option)
- [ ] Update command output to resolve big column length due to the long list of groups (e.g. php_unit_dedicate_assert]
- [ ] Add information about a fixer being deprecated or custom

After that I would call another review round.

scuben:
While fixing the path-output I introduced a new option to disable the output of the path with `--hide-path`.

scuben:
Multiple fixes and improvements:
- renamed `--hide-path` to `--hide-inheritance`
- fixed fixer sorting (now alphabetical as intended)
- added first line with config name and path (same as in fix command)
- new way to count (configured, enabled, risky, ...). The count number now respects the `--hide-*` filters. 
- **changed fixer's name color/icon:**
-- yellow/plus: not defined fixers (not configured && not enabled)
-- green/tick: enabled (regardless of configured or inherited)
-- red/cross: disabled (configured && not enabled)

scuben:
## Feature complete!

I'd like anyone to invite to review (or re-review) this PR.
ping: @jenkoian @Slamdunk @julienfalque @keradus 

After the review process I'd like to start with the testing. Thus I'm not the big phpunit expert I'd like to get some hints on how to start or structure the tests and what kind if test I should not miss.

scuben:
@Aerendir I didn't want to upset you with the removal of the dump functionality. It was intended to get this PR merged rather sooner than later with the possibility after that to add the dump functionality in a later PR. If you'd like to have the dump functionality still in this PR then I (or you?) can work on it. Looking forward to see your feedback on the current state.

Aerendir:
@scuben , I'm just exploring the command as it is now.

Why do you think it is not ready now? (I've not seen its current output).

scuben:
@Aerendir 

> Why do you think it is not ready now? (I've not seen its current output).

You mean the dump functionality? To be honest: I did not check it in depth. Will add it tonight and have a deeper look at it.

Aerendir:
@scuben , here are my thoughts:

1. Inheritance: for me is more clear the old format of inheritance: `Rule set 1 > Rule set 2 > Rule set 3`: please, can you reformat it this way? Listing the inheritance one rule set per line makes the output really long in some cases! (and this was discussed and something not desired)
2. When a Fixer is enabled by inheritance, we may add the same `>` before its name, after the thick. Something like `✔ > blank_line_after_namespace`: this way is immediately clear simply scrolling the list that that fixer is inherited. It was something proposed by @julienfalque in a review of the PR and I reserved myself to think deeply at it. I never implemented it as the problems with the tests arose, then time passed, this PR started to get dust... You know the story :P
3. When a `--hide-*` option is passed, the corresponding column should be not printed. The number of columns was a complaint, for this reason I tried to put as many information as possible all in the first column. Hiding the not wanted column completely from the output makes the resulting table thiner and easier to read.

ABOUT THE DUMP FUNCTIONALITY
It did nothing fancy: simply the `$this->dump` property stores the result of the `array_filter` in `::filterFixers()`.

Then, if the `--dump` option is passed, it uses `var_export` to print the list of the current table output in the format `fixer_name => false`.

This way is possible to copy and paste it in the `.php-cs` file to explicitly disable the listed rules.

To have full control over the configuration, in fact, it is possible to call the command this way:

    php-cs-fixer show --hide-configured --hide-custom --hide-deprecated --dump

The dumped result can be copy and pasted in the `php_cs` file and then I can enable only the fixers I really want.

scuben:
@Aerendir I implemented number 2.) and 3.) of your list. Number 1 will look very bad because of the long inheritance of e.g. `php_unit_no_expectation_annotation`. Can you confirm that?

I reintroduces the dump functionality but changed it in favour of a `--compare` input option. With this option activated it will dump the diff between available and configured fixers. Is that what it should be about in your opinion?

scuben:
### This is how it looks now

````
php-cs-fixer show
````
image

````
php-cs-fixer --hide-inherited
````
image

````
php-cs-fixer --compare
````
image

**Comment by Aerendir** [comment]
@scuben , great work!

Ok for the #1: it effectively become a bit long to show on screen: listing them vertically is a better solution, I agree.

About the `>` I think that it could be better to show it in front of the name, like this:

schermata 2018-03-13 alle 10 13 55

While, about the `--compare` option I agree: its a good balance.

Maybe we can also set the entire table to show only the fixers that are intercepted by the compare.

In my case, for example, this is the result of `php-cs-fixer show --compare`:

schermata 2018-03-13 alle 10 21 57

schermata 2018-03-13 alle 10 22 43

As you can see, it shows me all the 175 fixers, but of those only 57 are available for explicit configuration.

Maybe showing only the ones dumped is better. WDYT?

So, this is the summary:

1. Ok, better to display the list vertically;
2. Better to move the inheritance symbol `>` at the beginning of the name of the fixer, without parenthesis
3. Better to show only compared fixers in the table when `--compare` is passed.

After those modifications I think the command is really ready and we should think at how to properly test it (maybe @keradus can give us useful tips and suggestions).

WDYT?

**Comment by Aerendir** [comment]
I'm thinking deeply about ht e`--compare` option, checking the code, too.

Currently you have implemented a dedicated method to make the comparison, a dedicated property to save it, and then there is the dump method.

But the `--compare` option basically is equivalent to this:

    $: php-cs-fixer show --hide-enabled --hide-configured

So, we are overdoing things as, simply using what already exists in the command we can achieve the same result, without doing extra work.

I think that we should remove the extra logic for configuration.

This is also the reason for which I told to use the `--dump` option to simply dump the current table output.

If we like the `--compare` we may use it to internally set the `--hide-enabled` and `--hide-configured` options.

But doing the calculation again for me is overkill... And also complicates the command logic and makes the table and the dumped result disconnected (see my point 3 in my previous comment).

**Comment by julienfalque** [comment]
I finally took some time to have a look at this PR. It still looks promising but I think we should change a few things:
* I agree with @Aerendir about how inheritance is shown: it was more clear with previous format (single line below the fixer name). I'd prefer something like:
  * `enabled explicitly in configuration file`
  * `inherited from @Foo > @Bar` (maybe with a line length limit?)
* Previous colors for fixer names were better too IMO, at least using yellow for disabled rules feels weird
* The first column already allows you to know whether a rule is enabled and how. I think columns _Configured_, _Enabled_ and _Inherited_ can be dropped as they don't add any information.

WDYT?

File: src/Console/Command/ShowCommand.php [hunk 26]
```diff
@@ -0,0 +1,507 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+
+/**
+ * @author Adamo Aerendir Crespi <hello@aerendir.me>
+ * @author Patrick Landolt <landolt@gmail.com>
+ */
+final class ShowCommand extends Command
+{
+    const COMMAND_NAME = 'show';
+    const THICK = "\xE2\x9C\x94";
```
julienfalque: This constant and the ones below should be marked `@internal`.

src/Console/Command/ShowCommand.php
```diff
@@ -0,0 +1,507 @@
...
+                    new InputOption('hide-enabled', '', InputOption::VALUE_NONE, 'Hide fixers that are currently enabled (the ones that are not disabled with [\'fixer_name\' => false]).'),
+                    new InputOption('hide-risky', '', InputOption::VALUE_NONE, 'Hide fixers that are marked as risky.'),
+                    new InputOption('hide-inherited', '', InputOption::VALUE_NONE, 'Hide fixers that inherited from RuleSets.'),
+                    new InputOption('hide-deprecated', '', InputOption::VALUE_NONE, 'Hide fixers that are deprecated.'),
+                    new InputOption('hide-custom', '', InputOption::VALUE_NONE, 'Hide fixers that are custom.'),
+                    new InputOption('hide-inheritance', '', InputOption::VALUE_NONE, 'Hide the addition inheritance information.'),
+                    new InputOption('compare', '', InputOption::VALUE_NONE, 'Dumps the comparing result between your config and all available fixers in a copy-and-pastable format ready for the .php_cs file.'),
+                ]
+            )
+            ->setDescription('Shows existent Fixers with the ones actually configured or enabled by inheritance.')
```
julienfalque: I would write _"Lists all available fixers and shows which ones are enabled, inherited or disabled."_.

src/Console/Command/ShowCommand.php
```diff
@@ -0,0 +1,507 @@
...
+    {
+        return isset($this->enabledFixers[$fixer->getName()]);
+    }
+
+    /**
+     * @param FixerInterface $fixer
+     *
+     * @return bool
+     */
+    private function isFixerEnabledTroughInheritance(FixerInterface $fixer)
```
ntzm: Need to correct all `Trough` => `Through`

src/Console/Command/ShowCommand.php
```diff
@@ -0,0 +1,507 @@
...
+        $this->undefinedFixers = array_diff_key($this->fixerList, $this->configuredFixers, $this->enabledFixers);
+    }
+
+    /**
+     * @param OutputInterface $output
+     */
+    private function dumpComparison(OutputInterface $output)
+    {
+        if (empty($this->undefinedFixers)) {
+            $output->writeln("\nYou are aware of all exsisting rules! Yeah!");
```
ntzm: *existing

src/Console/Command/ShowCommand.php
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    /**
+     * @param string           $name
+     * @param RuleSetInterface $set
+     */
+    private function processRuleSet($name, RuleSetInterface $set)
+    {
+        /**
+         * @var string
```
ntzm: Should be `* @var string $rule`

localheinz:
@Aerendir 

Anything I can do to help get this shipped?

Aerendir:
@localheinz reading again the comments, three main problems emerged:

1. Which format to use for the inherited fixers (it seems we all agree on the "one-line" format, below the fixer's name)
2. How to format the `dump`
3. Versioning of fixers: we were waiting for its implementation to make the command more precise.

Unfortunately, I didn't follow the development of PHPCSFixer in the last months, so I don't know if the versioning of fixers is now implemented or not.

This is the main pain point... All the others I think can be addressed relatively easily.

Then, there are the tests that I'm not able to write (mine fails due to memory consumption and I don't know how to use less).

localheinz:
@Aerendir 

I'm happy to take over this PR, or propose a new based on the work of this PR - what do you think?

Aerendir:
@localheinz , I've done all in my possibilities. This PR is waiting for someone who can write tests for it as I'm not able to write them nor to find how to improve them.

If you like to contribute to this PR I would be very happy.

If you prefer to simply get the code and create a new PR, I cannot do anything about: this is the open source!

If you'll continue to work on it, then the choice is all your!
Obviously, I've spent a lot of time on this code, getting a mention will please me a lot :)

localheinz:
@Aerendir 

Absolutely, I’ll see to it!

src/Console/Command/ShowCommand.php
Aerendir: Done!

Aerendir:
I've updated the PR to reflect the current changes in `master` and to fix suggestions proposed in various reviews and comments.

All checks now pass and so I think the PR is ready to be merged.

I also renamed the command to `list` as it seems to me more appropriate.

WDYT? Is this PR ready to be merged?

src/Console/Command/ShowCommand.php
Aerendir: Done

src/Console/Command/ShowCommand.php
Aerendir: Done

src/Console/Command/ShowCommand.php
Aerendir: Done

src/Console/Command/ShowCommand.php
Aerendir: Done

src/Console/Command/CompareCommand.php
Aerendir: Done

src/Console/Command/CompareCommand.php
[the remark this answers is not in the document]
Aerendir: @julienfalque In the end I renamed the command to `list` as this is more descriptive, as you suggested.

kubawerlos:
> Is this PR ready to be merged?

@Aerendir with failing checks? Not even ready for review.

Aerendir:
@kubawerlos They are related to the changed name: I've fixed them now.

Aerendir:
ping @julienfalque , @keradus , @kubawerlos , @localheinz , @Slamdunk

Slamdunk:
> This PR is waiting for someone who can write tests for it as I'm not able to write them nor to find how to improve them.

I really appreciate the idea, I would needed this command a lot in the past, but the lack of tests is totally a no-go in a maintainer perspective.

Please note that, like you, here no one gets paid for maintaining this project, and what you are asking with the code written like this is: "_Please take my 500 LOC code that it isn't proved it works, clean it up for me and maintain it for free for the next years for the million users that expect a high-quality, reliable and consistent tool_".

We can't do anything do change the second part of this phrase, but it is up to you to ease it improving the first part, i.e. proving with tests what are your intents, how you've code it and how they behave as expected.

src/Console/Command/ListFixersCommand.php
```diff
@@ -0,0 +1,524 @@
...
+    private function filterFixers()
+    {
+        $hideConfigured = $this->hideConfigured;
+        $hideEnabled = $this->hideEnabled;
+        $hideRisky = $this->hideRisky;
+        $hideInherited = $this->hideInherited;
+        $hideDeprecated = $this->hideDeprecated;
+        $hideCustom = $this->hideCustom;
+
+        $rows = array_filter($this->fixerList, function (array $fixer) use ($hideConfigured, $hideEnabled, $hideRisky, $hideInherited, $hideDeprecated, $hideCustom) {
```
Wirone, line 322: No need to introduce variables and pass them through `use` to the closure - `$this` can be used inside as long as closure is not defined as `static`.

Wirone (changes requested):
While the general idea is interesting, I find the implementation not good enough (I'm going to leave one example comment) and of course lack of tests is a no-go, as already mentioned before. @Aerendir You should provide fixtures with example config files and run the command against them in various configuration, checking if the output matches expected one. Then we can see it works properly, no one is going to test it manually 🙂

Wirone:
@Aerendir in terms of command testing you may want to look at #6957 🙂.

Aerendir:
@Wirone , thank you for the hint... I'll look at it...

Aerendir:
@Wirone , thanks!

Wirone:
@Aerendir I've rebased branch and fixed the issues (almost), could you re-fork the project and bring the branch back, so I can force-push changes and refresh the PR?

Wirone:
Hard to tell, I don't know GitHub's internals. I am AFK, will look at it later.

Aerendir:
Ok, anyway, I'm working on the command... At most, I will create a new PR linking this one...

Wirone:
Because of technical limitations the feature will be continued in #7254.

PR #3052, opened 2017-09-12
Fix false positive warning about paths overridden by provided as command arguments

### LINKED ISSUE #3044: False positive warning about paths overridden by provided as command arguments
In latest 2.6.0 tag for a config:
```php
<?php

return PhpCsFixer\Config::create();
```

when running:
```bash
vendor/bin/php-cs-fixer fix --config=config-from-above.php src
```

I'm getting "Paths from configuration file have been overridden by paths provided as command arguments" warning.

kubawerlos:
Fixes #3044

### DIFF
--- a/src/Finder.php
+++ b/src/Finder.php
@@ -33,4 +33,14 @@ class Finder extends BaseFinder
             ->exclude('vendor')
         ;
     }
+
+    /**
+     * Check if this finder wasn't extended.
+     *
+     * return bool
+     */
+    public function isOriginal()
+    {
+        return serialize(new self()) === serialize($this);
+    }
 }
--- a/src/Finder.php
+++ b/src/Finder.php
@@ -33,14 +33,4 @@ class Finder extends BaseFinder
             ->exclude('vendor')
         ;
     }
-
-    /**
-     * Check if this finder wasn't extended.
-     *
-     * return bool
-     */
-    public function isOriginal()
-    {
-        return serialize(new self()) === serialize($this);
-    }
 }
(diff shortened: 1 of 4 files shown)


src/Finder.php
```diff
@@ -33,4 +33,14 @@ public function __construct()
...
     }
+
+    /**
+     * Check if this finder wasn't extended.
+     *
+     * return bool
+     */
+    public function isOriginal()
+    {
+        return serialize(new self()) === serialize($this);
```
keradus: nope
when you will change a property like filtering, yet still not configuring source where to find files, this will return false, wihch would be false positive.

keradus: here is how we detect that finder from config file is fully defined:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/Console/ConfigurationResolver.php#L777

kubawerlos: When I will change a property like filtering I will change finder, so false would be correct.

I'm not aiming to check if finder is fully defined, but to check if maybe someone extended Finder like [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/.php_cs.dist#L50).

I think we are not interested if Finder is fully defined, only if default one is overridden.

I'm open to suggestion how solve this differently as this is nicest solution I have found.

keradus: if one has a not fully defined finder in his config and provided a path, we shall not display a warning, as he is not overwriting the finder.

only fully-defined finder could be overwritten with cli `path` argument.
if finder in config is not fully defined, cli `path` argument will fulfill it: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.5/src/Console/ConfigurationResolver.php#L807

keradus:
Thank you @kubawerlos.

Issue #3057, opened 2017-09-14
Missing space with is_null in condition


roukmoute:
The PHP version you are using:
```
▶ php -v
PHP 7.1.7-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Jul  7 2017 09:41:45) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.7-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

PHP CS Fixer version you are using: `2.6-dev`

Example:
```php
<?php if (is_null($x) && $y) echo "foo";
```

Expected:
```php
<?php if (null === $x && $y) echo "foo";
```

Actual:
```php
<?php if (null === null&& $y) echo "foo";
```

SpacePossum:
Hi @roukmoute 

Can you share the command and/or configuration file you use? It will help to reproduce this issue.
Since you are on `2.6-dev`; do you have the commit hash of that version?

roukmoute:
I did P.R. directly @SpacePossum  :)

SpacePossum:
awesome, thanks so much man :)
when I've time I'll have a look at the PR :+1:

PR #3058, opened 2017-09-14
IsNullFixer - fix whitespace handling

roukmoute:
Fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3057

### DIFF
--- a/tests/Fixer/LanguageConstruct/IsNullFixerTest.php
+++ b/tests/Fixer/LanguageConstruct/IsNullFixerTest.php
@@ -76,6 +76,7 @@ final class IsNullFixerTest extends AbstractFixerTestCase
             '<?php $b = a(a(a(b() === null) === null) === null) === null;',
             '<?php $b = \is_null(a(\is_null(a(\is_null(a(\is_null(b())))))));'
         );
+        $this->doTest('<?php if ($x === null && $y) echo "foo";', '<?php if (is_null($x) && $y) echo "foo";');
     }
 
     public function provideYodaFixCases()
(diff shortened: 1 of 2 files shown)


tests/Fixer/LanguageConstruct/IsNullFixerTest.php
```diff
@@ -204,6 +205,36 @@ public function provideYodaFixCases()
                 '<?php $result = ((null === $a) <> false); ?>',
                 '<?php $result = (is_null($a) <> false); ?>',
             ],
+
+            // check with logical operator
```
julienfalque: The next test case does not contain any logical operator, shouldn't the comment be after it?

src/Tokenizer/Token.php
```diff
@@ -474,6 +474,16 @@ public function isWhitespace($whitespaces = " \t\n\r\0\x0B")
     }
 
     /**
+     * Check if token is a logical operator.
+     *
+     * @return bool
+     */
+    public function isLogicalOperator()
```
julienfalque: Technically, adding this method is a new feature which would require a minor version increment. But this PR is a bugfix that should target a lower branch. Could you please remove it?

keradus:
if 2.2.8, then please change target branch ;)

tests/Fixer/LanguageConstruct/IsNullFixerTest.php
```diff
@@ -76,6 +76,7 @@ public function testNonYodaFix()
             '<?php $b = a(a(a(b() === null) === null) === null) === null;',
             '<?php $b = \is_null(a(\is_null(a(\is_null(a(\is_null(b())))))));'
         );
+        $this->doTest('<?php if ($x === null && $y) echo "foo";', '<?php if (is_null($x) && $y) echo "foo";');
```
SpacePossum: This is currently the only test that fails without your proposed fixes.

The test passes without your fixes, when you disable this line;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.7/src/Fixer/LanguageConstruct/IsNullFixer.php#L106

This makes me think that the `IsNullFixer` should only be changed there by adding a test before the call to `removeTrailingWhitespace`. 

The other test cases are all nice, so :+1: for adding :)

roukmoute:
@juliendufresne @SpacePossum updates done.

I've change branch.
I only remove the use of `removeTrailingWhitespace()`.
I kept all new tests.

julienfalque:
> I only remove the use of `removeTrailingWhitespace()`.

I think you actually forgot to commit that :)

roukmoute:
@juliendufresne I don't understand.
I've commited: `Remove the use of the removeTrailingWhitespace() method`.

Have I miss something ?

SpacePossum:
The target is good now :+1: 
The test file looks good :+1:
Changes in `Token.php` are reverted :+1: 
Sadly, now there is no fix proposal left in the fixer file itself.

roukmoute:
@SpacePossum I have followed your commentary `The test passes without your fixes, when you disable this line;` and this [comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3058#discussion_r138987448).
Do you prefer I add a check like before?
Oh perhaps I add a condition before the `removeTrailingWhitespace()` method?

roukmoute:
Oh!
I've just seen that I've completely forgotten to remove `removeTrailingWhitespace()` :-D

I better understand your comments now :)

SpacePossum:
:)
thanks for updating!

>Oh perhaps I add a condition before the removeTrailingWhitespace() method?

I would've to take a better look at before making that call. What I did notice however is that removing all calls to `removeTrailingWhitespace` in the fixer still makes the tests pass. So I wonder, maybe the tests are incomplete or the fixer is doing unneeded things.

roukmoute:
If we try with:
```php
array(
    '<?php if (null === $u ) echo "foo"; ?>',
    '<?php if (is_null($u) ) echo "foo"; ?>',
),
```
we keep a space.
But I don't think it's the job of `IsNullFixer` to remove this space.

Am I right, or I have to improve this P.R.?

julienfalque:
No it's not its job. Some people even like to have spaces inside braces: `if ( is_null($foo) ) { ... }`, the fixer should not remove them, another fixer already takes care of that, if enabled.

SteveJobzniak:
Thanks for this PR. We'll start using this is_null fixer as soon as this fix is merged. :-)

keradus:
Thank you @roukmoute.

Issue #3035, opened 2017-09-09
New fixer: ErrorSuppressionFixer


kubawerlos:
What about fixer with this description:

* **no_errors_suppression**

  There should not be ``@`` before functions.

  Configuration options:

  - ``exclude`` (``array``): list of functions to ignore; defaults to ``[]``

keradus:
Hi.
This rule would be conflicting with `silenced_deprecation_error` rule, which enforces `@` for `trigger_error` (for `E_USER_DEPRECATE`).
Please propose a solution to get non-conflicting rules.

kubawerlos:
Then maybe something like this:

* **error_suppression**

  Removes or ensures suppression of error messages.

  Configuration options:

  - ``allow`` (``array``): list of functions allowed to have error suppresion; defaults to ``[]``
  - ``ensure`` (``array``): list of functions required to have error suppresion; defaults to ``['trigger_error']``

That way all functions not in ``allow`` or ``ensure`` lists would have error suppresion removed. This will make fixer ``silenced_deprecation_error`` deprecated.

keradus:
good direction, yet not enough. in that scenario, how to achieve behaviour of newly deprecated rule with new fixer?

using `ensure=>['trigger_error']` as config for new fixer would:
- put `@` for all `trigger_error` usage, not only user deprecations
- remove `@` from other cases

both things are not done by `silenced_deprecation_error` rule

kubawerlos:
To achieve `silenced_deprecation_error ` with the generalised  `allow`/`ensure` approach, fixer would need 2-dimensional array with first parameter of function as 2nd dimension - this even sounds like a horrible idea.

Maybe a fixer with 2 flags:
- `silence_deprecation_error` with `bool, defaults to `true`
- `noise_other_cases` with `bool`, defaults to `false`

Not a brilliant either, but could replace and extend `silenced_deprecation_error`. I think that also would cover most real life scenarios, when the goal is not to use `@`, except deprecation errors.

keradus:
not most generic approach, but probably cleanest one indeed.
option names would benefit a little grooming, also option value shall accept `null` (to leave it as-is, eg when one want to ensure `@` for user deprecations, yet don't touch anything else)

Looks nice. Do you want to make new PR for this new fixer and deprecating existing one ?

kubawerlos:
@keradus I have answered your comments in PR.

PR #3060, opened 2017-09-14
Add ErrorSuppressionFixer

kubawerlos:
#3035

Initial commit - fixer and tests are based on `SilencedDeprecationErrorFixer` thus @HeahDude is added as author.

Option allows only boolean value meaning to do what is should or not.

Having also `null` to ignore option would make `false` doing the opposite to `true`: 
- for `silence_deprecation_error` it would sense - to remove `@` from deprecation notices
- but for `noise_remaining_usages` what would that mean? Add `@` everywhere?

Feel free to propose better naming for options, descriptions and interesting test cases.

My doubt is line 106 in `ErrorSuppressionFixer.php` - I thought it should be something like:

```php
$indexes[$prev] = (true === $this->configuration['noise_remaining_usages']) ? false : null;
```

but that failed some tests. Anyway it's just a hint where to look for flaws.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -575,6 +575,22 @@ Choose from the list of available rules:
 
   *Risky rule: risky if the ``ereg`` function is overridden.*
 
+* **error_suppression** [@Symfony:risky]
+
+  Error control operator should be added to deprecation notices and/or
+  removed from other cases.
+
+  *Risky rule: risky because adding/removing ``@`` might cause changes to code behaviour or if ``trigger_error`` function is overridden.*
+
+  Configuration options:
+
+  - ``mute_deprecation_error`` (``bool``): whether to add ``@`` in deprecation
+    notices; defaults to ``true``
+  - ``noise_remaining_usages`` (``bool``): whether to remove ``@`` in remaining
+    usages; defaults to ``false``
+  - ``noise_remaining_usages_exclude`` (``array``): list of global functions to
+    exclude from removing ``@``; defaults to ``[]``
+
 * **escape_implicit_backslashes**
 
   Escape implicit backslashes in strings and heredocs to ease the
@@ -1462,9 +1478,10 @@ Choose from the list of available rules:
   ``(int)``, ``(double)`` and ``(real)`` as ``(float)``, ``(binary)`` as
   ``(string)``.
 
-* **silenced_deprecation_error** [@Symfony:risky]
+* **silenced_deprecation_error**
 
-  Ensures deprecation notices are silenced.
+  Ensures deprecation notices are silenced. DEPRECATED: use
+  ``error_suppression`` instead.
 
   *Risky rule: silencing of deprecation errors might cause changes to code behaviour.*
 
--- a/src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
+++ b/src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
@@ -1,0 +1,171 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\LanguageConstruct;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Jules Pietri <jules@heahprod.com>
+ * @author Kuba Werłos <werlos@gmail.com>
+ */
+final class ErrorSuppressionFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    const OPTION_MUTE_DEPRECATION_ERROR = 'mute_deprecation_error';
+    const OPTION_NOISE_REMAINING_USAGES = 'noise_remaining_usages';
+    const OPTION_NOISE_REMAINING_USAGES_EXCLUDE = 'noise_remaining_usages_exclude';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Error control operator should be added to deprecation notices and/or removed from other cases.',
+            [
+                new CodeSample("<?php\ntrigger_error('Warning.', E_USER_DEPRECATED);\n"),
+                new CodeSample(
+                    "<?php\n@mkdir(\$dir);\n@unlink(\$path);\n",
+                    [self::OPTION_NOISE_REMAINING_USAGES => true]
+                ),
+                new CodeSample(
+                    "<?php\n@mkdir(\$dir);\n@unlink(\$path);\n",
+                    [
+                        self::OPTION_NOISE_REMAINING_USAGES => true,
+                        self::OPTION_NOISE_REMAINING_USAGES_EXCLUDE => ['unlink'],
+                    ]
+                ),
+            ],
+            null,
+            'Risky because adding/removing `@` might cause changes to code behaviour or if `trigger_error` function is overridden.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(['@', T_STRING]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder(self::OPTION_MUTE_DEPRECATION_ERROR, 'Whether to add `@` in deprecation notices.'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(true)
+                ->getOption(),
+            (new FixerOptionBuilder(self::OPTION_NOISE_REMAINING_USAGES, 'Whether to remove `@` in remaining usages.'))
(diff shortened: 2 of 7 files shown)


src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Jules Pietri <jules@heahprod.com>
+ * @author Kuba Werłos <werlos@gmail.com>
```
keradus: damn, I anticipated `ś` at the end ;)

keradus:
@TomasVotruba , could you please help our community once more ?
This PR needs a decent review for quite a long already. Not only the code needs it, but also the idea itself.
Could you please take care of it ?

kubawerlos:
@keradus it needs update - I see test are not covering all cases

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+            'Adding or removing ``@`` might cause changes to code behaviour.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_STRING);
```
TomasVotruba: I'd check for these too:

- ','
- `T_CONSTANT` (not sure if that covers E_USER_DEPRECATED),

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $indexes = [];
+
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+
+            if ($token->equals('@') && true === $this->configuration['noise_remaining_usages']) {
```
TomasVotruba: First, there should be check for token, as we don't care about any other

```php
if (! $token->isGivenKind(T_STRING)) {
    continue;
}
```

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+            if ($tokens[$prev]->isGivenKind(T_STRING) || $tokens[$prev]->equals('@')) {
+                $indexes[$prev] = null;
+
+                continue;
+            }
+
+            $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $tokens->getNextTokenOfKind($index, [T_STRING, '(']));
+            if ($tokens[$tokens->getPrevMeaningfulToken($end)]->equals([T_STRING, 'E_USER_DEPRECATED'])) {
+                $indexes[$start] = (true === $this->configuration['silence_deprecation_error']) ? true : null;
+            }
```
TomasVotruba: Codeblock from `if (!$token->equals([T_STRING, 'trigger_error'], false)) {` to here should be decoupled to own method to improve readability and maintainability.

E.g. `$this->fixTriggerError($index, $tokens)`

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+
+            if ($tokens[$prev]->isGivenKind(T_STRING) || $tokens[$prev]->equals('@')) {
+                $indexes[$prev] = null;
+
+                continue;
+            }
+
+            $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $tokens->getNextTokenOfKind($index, [T_STRING, '(']));
+            if ($tokens[$tokens->getPrevMeaningfulToken($end)]->equals([T_STRING, 'E_USER_DEPRECATED'])) {
+                $indexes[$start] = (true === $this->configuration['silence_deprecation_error']) ? true : null;
```
TomasVotruba: Ternary operator with composed conditions is hard to read, yoda with moreover.
What does this do?

Could you split it to or simplify? Well, comparison bool to `true` could shortened:

```php
$indexes[$start] = $this->configuration['silence_deprecation_error'] ? true : null;
```

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        foreach ($indexes as $key => $index) {
+            if (true === $index) {
+                $tokens->insertAt($key, new Token('@'));
+            } elseif (false === $index) {
+                $tokens->clearAt($key);
+            }
+        }
```
TomasVotruba: What does this do? Could you maybe decouple to own method with descriptive name?

TomasVotruba:
@keradus Sure I'll give it some time, when you asked so nicely :+1: 

Personally I'd split it to fixers, since:

- one feature is `trigger_error` `E_USER_DEPRECATED` with `@`
- second is remove (almost) all `@` 
    - here should be option to whitelist functions to skip with `trigger_error` as default; also many people use `@mkdir` for checking already existing directory etc.

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,125 @@
...
+
+            if ($token->equals('@') && true === $this->configuration['noise_remaining_usages']) {
+                if (!array_key_exists($index, $indexes)) {
+                    $indexes[$index] = false;
+
+                    continue;
+                }
+            }
+
+            if (!$token->equals([T_STRING, 'trigger_error'], false)) {
```
TomasVotruba: Here would be best place to check against list of allowed functions

TomasVotruba:
I've just read original issue and discussion.
If I understand that correctly, this should cover only the 2nd fixer is suggested.

So this fixer should only take care of  `@` removal with `trigger_error` function as default.

keradus:
that's the thing, we can not have 2 fixers, one adding `@` (even only in some cases), one removing `@` , as they will be conflicting, and whole PHP CS Fixer project is written in a way to avoid conflicting rules.

TomasVotruba:
The other case is to keep this fixer as it is and accept it will containt this hidden reference to other fixer - "we skip trigger_error here, because it is covered by @see AnotherFixer".

That would definitelly made my code review less wtf.

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
kubawerlos: For files with no constants we still might want to remove `@`.

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
kubawerlos: `@` has no kind :( only we can do is `->equals('@')`.

kubawerlos:
@TomasVotruba thanks for the review, I've reworked fixer, it should be much easier to read it's code. Also older version was missing some cases.

README.rst
```diff
@@ -553,6 +553,20 @@ Choose from the list of available rules:
...
+  Error control operator should be added to deprecation notices and/or
+  removed from other cases.
+
+  *Risky rule: adding and/or removing ``@`` might cause changes to code behaviour.*
+
+  Configuration options:
+
+  - ``mute_deprecation_error`` (``false``, ``true``): whether to add ``@`` in deprecation
+    notices; defaults to ``true``
+  - ``noise_remaining_usages`` (``false``, ``true``): whether to remove ``@`` in remaining
```
keradus: If I would only read option, then I understand `noise_remaining_usages=yes` to actually add `@`.
Is it really adding them ?

kubawerlos: How come? `@` is known a silence operator, so "noising" would naturally mean removing it?

I've chosen the names "mute" and "noise" as they are opposite words and to have an order (which is alphabetical by option name) where the "remaining" option is second.

keradus: my bad, ignore :+1:

julienfalque:
Is there a specific reason not to implement some `exclude` option as proposed in #3035? IMO it's a must have, as some `@` usages are legit, e.g.:
```php
// failure is handled so one doesn't care about the E_WARNING
if (false === $file = @fopen(...)) {
    // throw some exception
}
```

keradus:
ping @kubawerlos 
what is the state of this PR ?

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder(self::OPTION_MUTE_DEPRECATION_ERROR, 'Whether to add `@` in deprecation notices.'))
+                ->setAllowedValues([true, false])
```
ntzm: Should be `->setAllowedTypes(['bool'])` instead

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,169 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $functionsAnalyzer = new FunctionsAnalyzer();
+        $excludedFunctions = array_map('strtolower', $this->configuration[self::OPTION_NOISE_REMAINING_USAGES_EXCLUDE]);
```
ntzm: Please use

```php
array_map(function ($function) {
    return strtolower($function);
}, $this->configuration[self::OPTION_NOISE_REMAINING_USAGES_EXCLUDE]);
```

instead, see #3769

src/Fixer/LanguageConstruct/ErrorSuppressionFixer.php
```diff
@@ -0,0 +1,169 @@
...
+                new CodeSample(
+                    "<?php\n@mkdir(\$dir);\n@unlink(\$path);\n",
+                    [
+                        self::OPTION_NOISE_REMAINING_USAGES => true,
+                        self::OPTION_NOISE_REMAINING_USAGES_EXCLUDE => ['unlink'],
+                    ]
+                ),
+            ],
+            null,
+            'Adding and/or removing `@` might cause changes to code behaviour.'
```
ntzm: Is also risky if `trigger_error` is overwritten

keradus:
Thank you @kubawerlos.

PR #3059, opened 2017-09-14
Fix couple of typos

KKSzymanowski:
Fixed three typos in the Readme.

PR #3064, opened 2017-09-15
Fix couple of typos

KKSzymanowski:
Follow up on #3059

Fixed three typos in the Readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -957,7 +957,7 @@ Choose from the list of available rules:
 
   Class names should match the file name.
 
-  *Risky rule: this fixer may change you class name, which will break the code that is depended on old name.*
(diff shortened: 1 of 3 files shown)


julienfalque:
Note that some of these changes can be applied to 2.2 as well. @keradus, do you want to handle this while merging? Or shall we open a separate PR?

keradus:
I cannot downgrade sth during merging. If pr is for 2.2 line, i will catch it up to 2.6.
but if sth is for 2.6 line, i won't apply changes back to 2.2 during merging

SpacePossum:
It would be nice to have some way to see on which version a rule was added so when you want fix multiple you don't need to go through all the release notes.

keradus:
Thank you @KKSzymanowski.

Issue #3056, opened 2017-09-14
IncrementStyleFixer


kubawerlos:
Some say pre incrementation/decrementation should be used if possible, other say this micro optimization comes at the cost of readability. Sumarizing: I'd like to propose a fixer to replace `pre_increment` one with description:

* **increment_style**

  Pre or post increment and decrement operators should be used.

  Configuration options:

  - ``style`` (``'pre'``, ``'post'``): whether to use the ``pre`` or ``post``increment and decrement operators; defaults to ``'pre'``

Obviously it won't be new fixer, more a modification of `pre_increment`, but configuration above looks cleaner than using `pre_increment` with `false` to achieve post increment/decrement operators.

keradus:
+/- 0
+0.5 if implementation won't create too big changes (so won't introduce more code to maintain)

PR #3065, opened 2017-09-15
Add IncrementStyleFixer

kubawerlos:
As I saw no objection in #3056 this is my plan:

- [x] create PostIncrementFixer
- [x] merge with PretIncrementFixer into IncrementStyleFixer 
- [x] remove PostIncrementFixer
- [x] get feedback and add improvements
- [x] deprecate PretIncrementFixer


tests/Fixer/Operator/PostIncrementFixerTest.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        $preIncrementFixerTest = new PreIncrementFixerTest();
+
+        return array_map(
+            'array_reverse',
```
keradus: that kind of invoking a function is undetectable by SCA tools, call it directly, not by name

keradus:
> As I saw no objection

easy, one day, especially during end of working week, it not too much time given

keradus:
regardless of real change proposed in this PR, looking at your steps, you understand our cycle pretty well :+1:

tests/Fixer/Operator/PostIncrementFixerTest.php
kubawerlos: done, what about changing calls in whole code?

keradus: :+1:, but as separated PR.
are there more calls like that? I tried to catch all of them :(

keradus:
I think there is no need to review `PostIncrementFixer`. Please make `IncrementStyleFixer` directly and then we could review, as it's the combined fixer that will deserve a review.

src/Fixer/Operator/IncrementStyleFixer.php
```diff
@@ -0,0 +1,191 @@
...
+                }
+
+                $endIndex = $this->findEnd($tokens, $index);
+
+                $nextToken = $tokens[$tokens->getNextMeaningfulToken($endIndex)];
+                if ($nextToken->equalsAny([';', ')'])) {
+                    $tokens->clearAt($index);
+                    $tokens->insertAt($tokens->getNextNonWhitespace($endIndex), clone $token);
+                }
+            } elseif ('pre' === $this->configuration['style'] && $tokensAnalyzer->isUnarySuccessorOperator($index)) {
```
keradus: const for `pre` is missing

tests/Fixer/Operator/IncrementStyleFixerTest.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public function provideFixPostIncrementCases()
+    {
+        return array_map(function (array $case) {
+            return array_reverse($case);
+        }, $this->provideFixPreIncrementCases());
+    }
+
+    public function provideFixPreIncrementCases()
```
keradus: it's not only about incrementation

tests/Fixer/Operator/PostIncrementFixerTest.php
kubawerlos: Only 4 cases -> https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3080

src/Fixer/Operator/IncrementStyleFixer.php
kubawerlos: do you mean to extract `pre` and `post` to class constants?

tests/Fixer/Operator/IncrementStyleFixerTest.php
kubawerlos: As `pre_increment` fixer is not only about incrementation. Looks better than `provideFixPreIncrementAndDecrementCases`. Following this the fixer should be named `IncrementAndDecrementStyleFixer`, shouldn't it?

src/Fixer/Operator/IncrementStyleFixer.php
keradus: yes

src/Fixer/Operator/IncrementStyleFixer.php
```diff
@@ -28,13 +28,16 @@
  */
 final class IncrementStyleFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    const STYLE_PRE = 'pre';
+    const STYLE_POST = 'post';
```
keradus: yep, now add
```
/**
 * @internal
 */
```

to them so they won't be part of bc promise we have to keep

kubawerlos: done

keradus:
Your understanding of the PHP CS Fixer is pretty nice @kubawerlos !
To be honest not much here that I could add as review concerns.

Please deprecate original fixer ;)

kubawerlos:
@keradus done and rebased.

src/Fixer/Operator/IncrementStyleFixer.php
```diff
@@ -0,0 +1,201 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Pre- or post-increment and decrement operators should be used if possible.',
+            [new CodeSample("<?php\n\$a++;\n\$b--;")]
```
keradus: add sample with non-default config

src/Fixer/Operator/IncrementStyleFixer.php
```diff
@@ -45,7 +45,13 @@ public function getDefinition()
     {
         return new FixerDefinition(
             'Pre- or post-increment and decrement operators should be used if possible.',
-            [new CodeSample("<?php\n\$a++;\n\$b--;")]
+            [
+                new CodeSample("<?php\n\$a++;\n\$b--;"),
+                new CodeSample(
+                    "<?php\n++\$a;\n--\$b;",
+                    ['style' => 'post']
```
keradus: const

kubawerlos: yeah, forgot about it, fixed

src/Fixer/Operator/PreIncrementFixer.php
```diff
@@ -12,107 +12,36 @@
...
+use PhpCsFixer\AbstractProxyFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Tokenizer\Tokens;
-use PhpCsFixer\Tokenizer\TokensAnalyzer;
 
 /**
  * @author Gregor Harlan <gharlan@web.de>
+ *
+ * @deprecated in 2.8, proxy to IncrementStyleFixer
```
keradus: So. When we deprecate an option, method, class, we trigger `E_USER_DEPRECATION`.
Here, it's more tricky, as whole fixer got deprecated.

Yet, it could be nice to trigger `E_USER_DEPRECATION` as well, even if it won't be just few lines addition (we need to handle those deprecations!).
Why? It would prevent having both, `pre_increment` and `increment_style` in `@Symfony` ruleset, as now we need to remember (which is always hard and we will forget, like this time) to replace deprecated rule in all rulesets.

kubawerlos: I based deprecation on `HashToSlashCommentFixer` - there is no `E_USER_DEPRECATION`. I'd say it's good to keep both now in `@Symfony`, so when people see in README they see both belongs to `@Symfony` (some may be thinking why old isn't in there and the suggested replacement is).

I was thinking that removing whole fixer is much more easier than removing option - you remove fixer class and make sure all tests are passing (which implicates removing tests for the fixer and rulesets entries).

About not forgetting I see simple solution: there is milestone for 3.0, why not creating an issue to remove deprecated fixers now?

To add `E_USER_DEPRECATION` wouldn't it be only to trigger it in `createProxyFixer`?

keradus: > I'd say it's good to keep both now in `@Symfony`

No, we do not keep usage of deprecated things, regardless they are methods, classes, rules.
Also, with having both of them, you end up running 2 different fixers for applying same rule, this is a CPU waste.

> so when people see in README they see both belongs to `@Symfony` (some may be thinking why old isn't in there and the suggested replacement is).

I don't believe it's a case. If sth is deprecated, old is deprecated, what does it mean? that it shall not be in use. If we ask end-user to not use it, showing we are still using it is confusing.

> I was thinking that removing whole fixer is much more easier than removing option (...)

it is, yet I don't find it related to triggering deprecation for end-user

> About not forgetting I see simple solution: there is milestone for 3.0, why not creating an issue to remove deprecated fixers now?

you are talking here about not forgetting of removing it at line 3.0, I'm saying about stop using it still on 2.x

> To add `E_USER_DEPRECATION` wouldn't it be only to trigger it in `createProxyFixer` ?

Or constructor directly.

kubawerlos: > No, we do not keep usage of deprecated things

Yeah, that make more sense, fixed.

> Or constructor directly.

Done, but I need a [hint](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/282452381).

keradus: moving it to `applyFix` is actually not helping in initial concern I raised - forbidding to use deprecated fixers in rulesets.

kubawerlos: Do you mean like an adding:

```php
    // tests/RuleSetTest.php
    /**
     * @group legacy
     * @expectedDeprecation Fixer "pre_increment" is deprecated and will be removed in 3.0. Use "increment_style" instead.
     */
    public function testUsingDeprecatedTriggerError()
    {
        RuleSet::create([
            '@PSR2' => true,
            'line_ending' => true,
            'strict_comparison' => true,
        ]);
    }
```
?

This won't pass with `E_USER_DEPRECATED` triggered in constructor either.

I could move it from `applyFix` to `isCandidate` so it would be tiggered when fixer would be in use even without actual fixing.

keradus: sth like this, yet you have no `pre_increment` in creater ruleset ;)

i think throwing deprecation from random places to find best one is not a way we shall go. 
For fixer, deprecation shall be triggered in constructor or nowhere.

kubawerlos: Can it be nowhere in this PR?

If I add deprecation error in constructor all tests that are using `FixerFactory::registerBuiltInFixers` (7 files) will have to be modified and also that won't be enough as still many deprecation errors will be thrown:

```
Fixer "pre_increment" is deprecated and will be removed in 3.0. Use "increment_style" instead: 203x
    153x in DescribeCommandTest::testDescribeCommand from PhpCsFixer\Tests\AutoReview
    16x in TextDiffTest::testDiffReportingDecorated from PhpCsFixer\Tests
    7x in DescribeCommandTest::setUp from PhpCsFixer\Tests\Console\Command
    6x in ConfigurationResolverTest::testConfigFinderIsOverridden from PhpCsFixer\Tests\Console
    2x in ProjectFixerConfigurationTest::testCreate from PhpCsFixer\Tests\AutoReview
    2x in ConfigTest::testConfigRulesUsingSeparateMethod from PhpCsFixer\Tests
    2x in ReadmeCommandTest::testIfReadmeFileIsCorrect from PhpCsFixer\Tests\Console\Command
    2x in ConfigTest::testCustomConfig from PhpCsFixer\Tests
    2x in ConfigTest::testConfigRulesUsingJsonMethod from PhpCsFixer\Tests
    2x in FixCommandTest::testEmptyRulesValue from PhpCsFixer\Tests\Console\Command
    1x in ApplicationTest::testApplication from PhpCsFixer\Tests\Console
    1x in FixCommandTest::setUp from PhpCsFixer\Tests\Console\Command
    1x in CommandTest::provideCommandHasNameConstCases from PhpCsFixer\Tests\AutoReview
    1x in ConfigurationResolverTest::testResolveRulesWithConfig from PhpCsFixer\Tests\Console
    1x in ConfigurationResolverTest::testResolveRulesWithOption from PhpCsFixer\Tests\Console
    1x in ConfigurationResolverTest::testResolveRulesWithUnknownRules from PhpCsFixer\Tests\Console
    1x in ConfigurationResolverTest::testResolveRulesWithConfigAndOption from PhpCsFixer\Tests\Console
    1x in ConfigurationResolverTest::testResolveCommandLineInputOverridesDefault from PhpCsFixer\Tests\Console
    1x in ConfigurationResolverTest::testResolveConfigFileOverridesDefault from PhpCsFixer\Tests\Console
```

I believe we should handle it in separate PR.

keradus: very well, let it be

kubawerlos: great, I have updated PR

tests/Fixer/Operator/PreIncrementFixerTest.php
```diff
@@ -27,6 +27,7 @@
      * @param string      $expected
      * @param null|string $input
      *
+     * @group legacy
```
keradus: btw, we use it only with combination of deprecated code being tested

kubawerlos: of course, fixed

keradus:
Thank you @kubawerlos.

PR #3068, opened 2017-09-16
Symfony:risky ruleset - add no_homoglyph_names

keradus:
I hoped to add it to PSR ruleset, but I cannot find it in spec :(
Symfony follows it.


julienfalque:
> Symfony follows it.

Where did you see that?

keradus:
codebase, golden rule - mimic recent code, here, mimic whole code.

Issue #3067, opened 2017-09-16
Disappearing function argument


nullref:
PHP-CS-Fixer 5642a36 with default settings running on PHP 7.1.9 exhibits a bug where a function argument is dropped if there is no space after the previous argument's comma and the next argument is on the next line together with the closing parenthesis.

```diff
% vendor/bin/php-cs-fixer fix --dry-run --diff test.php 
Loaded config default.
Using cache file ".php_cs.cache".
   1) test.php
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
 <?php

-f(1,2,
-3);
+f(
+    1,
+    ,
+3
+);
      ----------- end diff -----------
```

The argument "2" was removed.

keradus:
Very interesting case indeed and I totally confirm it's sth extra important to fix.
Yet, we can't fix it ! Why ? You refused to provide all info we asked you for :(

Please, to make it possible to fix the issue, fill the template:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.github/ISSUE_TEMPLATE

SpacePossum:
O. o which fixer does this?

keradus:
`method_argument_space`

keradus:
broken since 2.4.x, to be fixed with current stable 2.6

nullref:
I've taken a look at the code and isolated the behavior to the ensureFunctionFullyMultiline method in the MethodArgumentSpaceFixer class, more specifically the code run by fixNewline when $override is set to true.

As for an actual fix, I'm not sure.  Perhaps the actual problem is in fixSpace2 which states it adds spaces after commas, which appears not to have been done in this case.

keradus:
utest added: #3069
thanks for hint @nullref 
I will try to play with it tomorrow, so if you want to take even deeper look before me - go ahead ;)

at 2.4.0 we added new option (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2440), and that option is causing the issue.

PR #3069, opened 2017-09-17
MethodArgumentSpaceFixer - new test case

keradus:
closes: #3067

### DIFF
--- a/tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
+++ b/tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
@@ -315,6 +315,18 @@ INPUT
             ,
                 ['ensure_fully_multiline' => true],
             ],
+            'test wrongly formatted half-multiline function becomes fully-multiline' => [
+                '<?php
+f(
+    1,
+    2,
+3
+);',
+                '<?php
+f(1,2,
+3);',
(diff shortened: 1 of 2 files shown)


keradus:
@nullref could you please confirm the fix ?

nullref:
I tested it with my minimal test case and the original code I generated it from, both were processed without generating a syntax error.

tests/Fixer/FunctionNotation/MethodArgumentSpaceFixerTest.php
```diff
@@ -315,6 +315,18 @@ functionCall(
             ,
                 ['ensure_fully_multiline' => true],
             ],
+            'test wrongly formatted half-multiline function becomes fully-multiline' => [
+                '<?php
+f(
+    1,
+    2,
+3
```
julienfalque: Shouldn't this line be indented as well?

keradus: this rule is not changing indentation for arguments that are already on separated line, see `test ungodly code` case.
if we want to add it, it's ouf ot the scope of this fix

julienfalque: Damn it, I always forget that for this fixer! :D

SpacePossum:
thanks @keradus :+1:

PR #3070, opened 2017-09-17
YodaStyleFixer - Clarify configuration parameters

SteveJobzniak:
@SpacePossum Great job on this fixer. However, the configuration description is unclear... 

```
yoda_style [@Symfony]

Write conditions in Yoda style or not based on configuration.

Configuration options:

equal (bool, null): change equal (==, !=) statements; defaults to true
identical (bool, null): change identical (===, !==) statements; defaults to true
less_and_greater (bool, null): change less and greater than (<, <=, >, >=) statements; defaults to null
```

They all say: "Change X (if true) or not (if false)". That's not right and doesn't document the actual (great) possibilities of this fixer. It actually does:

true = Force to yoda style.
false = Force to non-yoda style.
null = Do nothing (ignore these statements).

This pull request tries to clarify the values. I hope the new descriptions are okay @keradus. I did not see any other fixers document params in this way. Because no other multi-option fixers used values in this weird way. ;-) It would have been better if the values had been a string: `'yoda'` or `'non-yoda'` or `'ignore'`. But since it has been released as 2.6, that can't be changed now without breaking backwards compatibility with 2.x of PHP-CS-Fixer. So improving the docs to explain true/false/null is the best we can do.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1283,16 +1283,17 @@ Choose from the list of available rules:
 
 * **yoda_style** [@Symfony]
 
-  Write conditions in Yoda style or not based on configuration.
+  Write conditions in Yoda style (``true``), non-Yoda style (``false``) or
+  ignore those conditions (``null``) based on configuration.
 
   Configuration options:
 
-  - ``equal`` (``bool``, ``null``): change equal (``==``, ``!=``) statements; defaults to
+  - ``equal`` (``bool``, ``null``): style for equal (``==``, ``!=``) statements; defaults to
     ``true``
-  - ``identical`` (``bool``, ``null``): change identical (``===``, ``!==``) statements;
+  - ``identical`` (``bool``, ``null``): style for identical (``===``, ``!==``) statements;
     defaults to ``true``
-  - ``less_and_greater`` (``bool``, ``null``): change less and greater than (``<``, ``<=``,
-    ``>``, ``>=``) statements; defaults to ``null``
+  - ``less_and_greater`` (``bool``, ``null``): style for less and greater than (``<``,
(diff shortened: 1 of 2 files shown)


SpacePossum:
Did you modify the REAME.rst by hand or regenerate it? as now the tests fail..
Beside that; I've no strong opinion on the copy used to describe the config, so I let the community deside ;)

(yet I still think using `true/false/null` as value was a good call over some magic string value, but that is part of the PR description but not a proposed change :) )

SteveJobzniak:
@SpacePossum Ohhh, I did it by hand. I didn't know you had an auto-generator (I see it now in contributing.md). I'll update the pull request in a moment!

SteveJobzniak:
@SpacePossum I've auto-generated the README now and made sure that the FixerDefinition matches it completely.

And sure, true/false/null is okay (and even if we wanted to, we can't change it now without breaking backwards-compatibility anyway...). We just need to document those 3 values, which is what I'm trying to do. Let's see what you guys think of this new update now!

README.rst
```diff
@@ -1283,16 +1283,17 @@ Choose from the list of available rules:
 
 * **yoda_style** [@Symfony]
 
-  Write conditions in Yoda style or not based on configuration.
+  Write conditions in Yoda style (``true``) or non-Yoda style (``false``) or
```
keradus: `foo, bar or baz` over `foo or bar or baz`

SteveJobzniak: Good catch!

SpacePossum:
:+1: if test are green :)

keradus:
Thank you @SteveJobzniak.

SteveJobzniak:
@keradus Thank you too! :)

PR #3073, opened 2017-09-17
SpaceAfterSemicolonFixer - Add option to remove space in empty for expressions

julienfalque:
This new option allows to apply the following change:
```diff
 <?php
-for ($i = 0; ; ++$i) {
+for ($i = 0;; ++$i) {
 }
```

I added it to the `@Symfony` rule set as the framework currently follows this rule for consistency (see https://github.com/symfony/symfony/pull/22931#discussion_r121027128).

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -45,6 +45,7 @@ $config = PhpCsFixer\Config::create()
         'phpdoc_types_order' => true,
         'semicolon_after_instruction' => true,
         'single_line_comment_style' => true,
+        'space_after_semicolon' => true,
         'strict_comparison' => true,
         'strict_param' => true,
         'yoda_style' => true,
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -45,7 +45,6 @@ $config = PhpCsFixer\Config::create()
         'phpdoc_types_order' => true,
         'semicolon_after_instruction' => true,
         'single_line_comment_style' => true,
-        'space_after_semicolon' => true,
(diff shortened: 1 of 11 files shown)


SpacePossum:
:+1: for idea

However, here SF team asks to not remove the space: 
https://github.com/symfony/symfony/pull/24123#issuecomment-327770614
so I'm not sure adding it to SF is a good idea, so I'm +0 on that change

keradus:
damn, linked fabbot report is dead
so, SF doesn't know which way they want to go? maybe go both ways locally, see amount of hit for having space and not having space, then we could propose to unify?

julienfalque:
@nicolas-grekas @fabpot Could you please confirm whether this new rule should be part of the `@Symfony` ruleset?

nicolas-grekas:
:+1: this is a common false positive right now

fabpot:
👍

.php_cs.dist
```diff
@@ -45,6 +45,7 @@ $config = PhpCsFixer\Config::create()
         'phpdoc_types_order' => true,
         'semicolon_after_instruction' => true,
         'single_line_comment_style' => true,
+        'space_after_semicolon' => true,
```
keradus: we could follow `@Symfony` ruleset here, no need for exception

julienfalque: Updated.

keradus:
Thank you @julienfalque.

PR #3074, opened 2017-09-18
[IO] Replace Diff with fork version

SpacePossum:
No BC breaks this time.
This will provide an upgrade path to producing diff's with linenumbers.

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2993

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "composer/semver": "^1.4",
         "doctrine/annotations": "^1.2",
         "gecko-packages/gecko-php-unit": "^2.0",
-        "sebastian/diff": "^1.4",
+        "php-cs-fixer/diff": "^1.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -20,7 +20,7 @@
         "composer/semver": "^1.4",
         "doctrine/annotations": "^1.2",
         "gecko-packages/gecko-php-unit": "^2.0",
-        "sebastian/diff": "^1.4",
+        "php-cs-fixer/diff": "dev-master",
```
keradus: before merging this PR, we shall make a release of `diff` to not enforce to using non-constant dev-master, with release version, composer will have way better installing option

keradus:
Thank you @SpacePossum.

PR #3076, opened 2017-09-19
Add ExplicitStringVariableFixer

Slamdunk:
```diff
-$a = "My name is $name!";
+$a = "My name is ${name}!";
```
Some notes:

1. `SingleQuoteFixer` is triggered by `T_CONSTANT_ENCAPSED_STRING` while this one is triggered by pure variable strings like `"$var"` or by `T_ENCAPSED_AND_WHITESPACE`, so no conflict between the two should happen
1. It seems that the fixer can be built only with tokenization, without involving regex, so I marked it as **non-risky**
1. Only [Simple sintax](https://secure.php.net/manual/en/language.types.string.php#language.types.string.parsing.simple) is involved, [Complex (curly) sintax](https://secure.php.net/manual/en/language.types.string.php#language.types.string.parsing.complex) (e.g. `"{$var}"`) is unaffected
1. Thanks to tokenization, current test cases seem the minimal set needed, but to me it's too good to be true, help in finding edge cases is welcome

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -545,7 +545,7 @@ Choose from the list of available rules:
 
 * **explicit_string_variable**
 
-  Convert implicit variables into explicit ones in double-quoted strings.
+  Converts implicit variables into explicit ones in double-quoted strings.
 
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -543,6 +543,10 @@ Choose from the list of available rules:
   Add curly braces to indirect variables to make them clear to understand.
   Requires PHP >= 7.0.
 
+* **explicit_string_variable**
+
+  Convert implicit variables into explicit ones in double quoted strings.
+
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
   PHP code must use the long ``<?php`` tags or short-echo ``<?=`` tags and not
--- a/README.rst
+++ b/README.rst
@@ -545,7 +545,8 @@ Choose from the list of available rules:
 
 * **explicit_string_variable**
 
-  Converts implicit variables into explicit ones in double-quoted strings.
+  Converts implicit variables into explicit ones in double-quoted strings
+  or heredoc syntax.
 
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -545,7 +545,7 @@ Choose from the list of available rules:
 
 * **explicit_string_variable**
 
-  Convert implicit variables into explicit ones in double quoted strings.
+  Convert implicit variables into explicit ones in double-quoted strings.
 
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -60,7 +60,8 @@ final class ExplicitStringVariableFixer extends AbstractFixer
 
             $prevToken = $tokens[$index - 1];
             if (
-                   $prevToken->isGivenKind(T_ENCAPSED_AND_WHITESPACE)
+                $prevToken->isGivenKind(T_ENCAPSED_AND_WHITESPACE)
+                || $prevToken->isGivenKind(T_START_HEREDOC)
                 || '"' === $prevToken->getContent()
             ) {
                 $tokens->overrideRange($index, $index, [
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -30,8 +30,15 @@ final class ExplicitStringVariableFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Convert implicit variables into explicit ones in double quoted strings.',
-            [new CodeSample('<?php $a = "My name is $name!";')]
+            'Convert implicit variables into explicit ones in double-quoted strings.',
+            [new CodeSample('<?php $a = "My name is $name!";')],
+            implode(PHP_EOL, [
+                'The reasoning behind this rule are the following:',
+                '- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow',
+                '- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred',
+                '- Explicit syntax allows word concatenation inside strings, e.g. "${var}IsAVar", implicit doesn\'t',
+                '- Explicit syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read',
+            ])
         );
     }
 
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -30,7 +30,7 @@ final class ExplicitStringVariableFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Convert implicit variables into explicit ones in double-quoted strings.',
+            'Converts implicit variables into explicit ones in double-quoted strings.',
             [new CodeSample('<?php $a = "My name is $name!";')],
             'The reasoning behind this rule is the following:'
                 ."\n".'- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow'
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -32,10 +32,10 @@ final class ExplicitStringVariableFixer extends AbstractFixer
         return new FixerDefinition(
             'Convert implicit variables into explicit ones in double-quoted strings.',
             [new CodeSample('<?php $a = "My name is $name!";')],
-            'The reasoning behind this rule are the following:'
+            'The reasoning behind this rule is the following:'
                 ."\n".'- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow'
-                ."\n".'- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred'
-                ."\n".'- Explicit syntax allows word concatenation inside strings, e.g. "${var}IsAVar", implicit doesn\'t'
+                ."\n".'- PHP manual marks `"$var"` syntax as implicit and `"${var}"` syntax as explicit: explicit code should always be preferred'
+                ."\n".'- Explicit syntax allows word concatenation inside strings, e.g. `"${var}IsAVar"`, implicit doesn\'t'
                 ."\n".'- Explicit syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read'
         );
     }
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -30,7 +30,7 @@ final class ExplicitStringVariableFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Converts implicit variables into explicit ones in double-quoted strings.',
+            'Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.',
             [new CodeSample('<?php $a = "My name is $name!";')],
             'The reasoning behind this rule is the following:'
                 ."\n".'- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow'
--- a/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
+++ b/src/Fixer/StringNotation/ExplicitStringVariableFixer.php
@@ -1,0 +1,69 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\StringNotation;
+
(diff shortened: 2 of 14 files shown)


keradus:
Hi. Could you please provide reasoning behind this rule?
Also, maybe there is some community following the rule ?

Slamdunk:
> Could you please provide reasoning behind this rule?

1. When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow (I have tons of legacy code with both used)
1. [Manual](https://secure.php.net/manual/en/language.types.string.php#language.types.string.parsing.simple) marks `"$var"` sintax as _implicit_ and `"${var}"` syntax as _explicit_: I think explicit code should always be preferred
1. _Explicit_ syntax allows word concatenation, e.g. `"${var}IsAVar"`, _implicit_ doesn't; if someone wants to use _implicit_, he/she has to use string concatenation `$var . "IsAVar"`, otherwise _explicit_ must be used. So, edge cases need _explicit_ syntax, and if there are cases without choises, I think all the other cases must follow them.
1. (Least important) _explicit_ syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read

> Also, maybe there is some community following the rule ?

Dunno ¯\\\_\(ツ\)\_\/¯

keradus:
Thanks for reasoning, I think we could put it to Fixer definition. (we have short summary and long description field)

In that case, let us wait a bit for community output is there a willing of ppl to use such a rule, maybe somehow changed/groomed/configured.

kubawerlos:
Ad 3. It's possible with _implicit_ syntax and without concatenation: `"{$var}IsAVar"` - hacky, but possible.

Anyway, what about instead of `bool` value making configuration option `syntax` with values `implicit`, `explicit` and `sprintf`? If I'd see variable inside double quoted string, in most cases I'd suggest to replace it with `sprintf` function.

Would be tricky (if possible) to convert from each 3 version to others, so maybe treating those as ordered solutions with ``implicit` being lowest and `sprintf` highest and the option could define what to upgrade?

~~Going further, having those 3 we could have 4th one: `sprintf_only_in_exceptions` and making it default so the fixer would be part of `@Symfony`.~~ - that's too far, not for the initial version for sure.

Slamdunk:
> Ad 3. It's possible with implicit syntax and without concatenation: "{$var}IsAVar" - hacky, but possible.

This is not implicit nor hacky, is [Complex (curly) sintax](https://secure.php.net/manual/en/language.types.string.php#language.types.string.parsing.complex), and as stated by this PR opening comment, it is not affected nor evaluated

> Anyway, what about instead of bool value making configuration option syntax with values implicit, explicit and sprintf? If I'd see variable inside double quoted string, in most cases I'd suggest to replace it with sprintf function.

Even though `sprintf` is the solution I mandate in my businesses, I don't like going that far because can reduce readibility:
```php
$a = "I am ${name} ${surname}, and live in ${city}, ${state}, ${street} ${streetNumber}";
$b = sprintf('I am %s %s, and live in %s, %s, %s %s', $name, $surname, $city, $state, $street, $streetNumber);
```

I thought about _implicit_ as an option, but I dislike it because:

1. How do you turn `"${var}IsAVar"` into _implicit_ syntax? `$var . "IsAVar"`? No thanks: if the developer wants everything inside a string (without concatenation) this fixer shouldn't change it.
1. If `"${var}IsAVar"` is changed to `$var . "IsAVar"`, how do you turn this back? Reverting `$var . "IsAVar"` to `"${var}IsAVar"` means also every concatenation becomes a single string, and we can't do that
1. PHP-CS-Fixer tries to promote good practices; _implicit_ variables are not a good practice

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,76 @@
...
+        return new FixerDefinition(
+            'Convert implicit variables into explicit ones in double-quoted strings.',
+            [new CodeSample('<?php $a = "My name is $name!";')],
+            implode(PHP_EOL, [
+                'The reasoning behind this rule are the following:',
+                '- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow',
+                '- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred',
+                '- Explicit syntax allows word concatenation inside strings, e.g. "${var}IsAVar", implicit doesn\'t',
+                '- Explicit syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read',
+            ])
```
Slamdunk: @keradus I don't know how you're used to create descriptions; hope you are ok with `implode(PHP_EOL, [])` syntax which allows indentation and less error-prone long strings writing.

Phrases were revisited; master of thesaurus @julienfalque please help me 🙏

kubawerlos:
> I don't like going that far because can reduce readibility

It wouldn't reduce it, it was already poor. It's like having a function with too many parameters, someone made it and no fixer can make it right for him.

I'd use the `sprintf` option in CI tools to point to the developer that it's not right to put variable in string in any way. PHP-CS-Fixer would suggest a solution, so with 1 or 2 variables it would be the final one. With your example it needs more work anyway.


> I thought about implicit as an option, but I dislike it (...)

Haven't thought about it this way - you're right, having it would make more harm than good.

Slamdunk:
> It wouldn't reduce it, it was already poor.

I disagree:
```php
$b = sprintf('I am %s %s, and live in %s, %s, %s %s', $name, $surname, $city, $state, $street, $streetNumber);
```
is mush less readable than this
```php
$a = "I am ${name} ${surname}, and live in ${city}, ${state}, ${street} ${streetNumber}";
```

> I'd use the sprintf option in CI tools to point to the developer that it's not right to put variable in string in any way. PHP-CS-Fixer would suggest a solution, so with 1 or 2 variables it would be the final one. With your example it needs more work anyway.

We can write such fixer, but can't be faced here. This fixer only takes one particular string type (double-quoted with implicit variables) and converts it to another particular string type (double-quoted with explicit variable).

What you are asking would have more sense for all string types: single-quoted, double-quoted, implicit syntax, explicit syntax, complex syntax etc.

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
keradus: we use `\n` everywhere

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,74 @@
...
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Convert implicit variables into explicit ones in double-quoted strings.',
+            [new CodeSample('<?php $a = "My name is $name!";')],
+            'The reasoning behind this rule are the following:'
+                ."\n".'- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow'
+                ."\n".'- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred'
+                ."\n".'- Explicit syntax allows word concatenation inside strings, e.g. "${var}IsAVar", implicit doesn\'t'
+                ."\n".'- Explicit syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read'
```
Slamdunk: @keradus is this way ok?

keradus:
> I'd use the sprintf option in CI tools to point to the developer that it's not right to put variable in string in any way.

PHP 5? sprintf over concatenation/interpolation

PHP 7? strings interpolation !
https://blog.blackfire.io/php-7-performance-improvements-encapsed-strings-optimization.html

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
julienfalque: Why not using HEREDOC or a single string, e.g.:
```php
<<<DESC
Foo:
- bar
- baz
DESC
```
or
```php
'Foo:
- bar
- baz'
```
?

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,74 @@
...
+ */
+final class ExplicitStringVariableFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Convert implicit variables into explicit ones in double-quoted strings.',
```
julienfalque: _"Convert**s**"_?

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,74 @@
...
+            [new CodeSample('<?php $a = "My name is $name!";')],
+            'The reasoning behind this rule are the following:'
```
julienfalque: _are_ -> _is_

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,74 @@
...
+                ."\n".'- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow'
+                ."\n".'- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred'
```
julienfalque:
```diff
-- PHP manual marks "$var" sintax as implicit and "${var}" syntax as explicit: explicit code should always be preferred
+- PHP manual marks `"$var"` syntax as implicit and `"${var}"` syntax as explicit: explicit code should always be preferred
```

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
```diff
@@ -0,0 +1,74 @@
...
+                ."\n".'- Explicit syntax allows word concatenation inside strings, e.g. "${var}IsAVar", implicit doesn\'t'
```
julienfalque: Please add backticks (_`_) around _"${var}IsAVar"_.

tests/Fixer/StringNotation/ExplicitStringVariableFixerTest.php
```diff
@@ -0,0 +1,53 @@
...
+            [
+                '<?php $a = "My name is ${name}!";',
+                '<?php $a = "My name is $name!";',
+            ],
+            [
+                '<?php $a = "${b}";',
+                '<?php $a = "$b";',
+            ],
+            ['<?php $a = \'My name is $name!\';'],
+            ['<?php $a = "My name is " . $name;'],
```
julienfalque: Please add the following test case:
```php
['<?php $a = "My name is {$name}!";'],
```

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
Slamdunk: No essential reason, just dislike to break indentation.
If you prefer, I'll pick one of the two you suggest

src/Fixer/StringNotation/ExplicitStringVariableFixer.php
Slamdunk: Ok, but:
```
$ grep -rni Convert\  src/Fixer/
src/Fixer/StringNotation/HeredocToNowdocFixer.php:32:            'Convert `heredoc` to `nowdoc` where possible.',
src/Fixer/StringNotation/SingleQuoteFixer.php:32:            'Convert double quotes to single quotes for simple strings.',
src/Fixer/ClassNotation/NoPhp4ConstructorFixer.php:33:            'Convert PHP4-style constructors to `__construct`.',
```

julienfalque: Those should be fixed as well IMO (not in this PR of course).

Slamdunk:
@julienfalque done everything you asked

Slamdunk:
@julienfalque Does this PR deserve the 2.8.0 milestone?

SpacePossum:
please add a negative test for `nowdoc` and a test case with many (15+; please use some sort of generating the input and output) positive fix cases within one "PHP file".

Slamdunk:
> a test case with many (15+; please use some sort of generating the input and output) positive fix cases within one "PHP file"

@SpacePossum sorry I don't understand this part, can you elaborate a bit more the idea?

SpacePossum:
@Slamdunk sorry to be vaguely about the second test I asked, let me try again :)

Your fixer fixes something like;
`<?php $a = "My name is $name!";`

What I would like to see as test (mostly to prevent regression btw.) is something like a test of;
input:
```
<?php
$a00 = "My name is $name!"
$a01 = "My name is $name!"
[...]
$a15 = "My name is $name!"
```
with a `expected` counter part.
With such a test we are sure if the fixer has a lot "stuff" to fix, it fixes all correctly.

Slamdunk:
I'm sorry I don't get how could that be useful.
Two consequent strings would be enough, wouldn't they?

SpacePossum:
Two would be nice, however if you've two in loop than more shouldn't be issue I think?
I was thinking of a construct something like https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3183/files#diff-542de4b103ea2736810b39d22a3c22b4R39 .
The reason I push for more tests like this is because we had some issues with multiple candidates before, so it is not directly related to this PR alone.

keradus:
@Slamdunk the point of that test is to prove fixer works properly when adding new tokens. imagine there is 1000 tokens, and token `-10` (so 990) needs a fixing. 
Then, during iteration, there is a lot of things to fix, thus fixer inserts 100 new tokens, collection size is not 1100, and last token that needs a fix is 1090. yet, last index you are currently visiting is still 1000, so last case is not fixed.
https://3v4l.org/YGtWf

README.rst
```diff
@@ -543,6 +543,11 @@ Choose from the list of available rules:
   Add curly braces to indirect variables to make them clear to understand.
   Requires PHP >= 7.0.
 
+* **explicit_string_variable**
```
keradus: I would add it to local `.php_cs.dist` file and then merge PR, wdyt @Slamdunk ?

Slamdunk: Done

Slamdunk: Done

keradus:
Thank you @Slamdunk.

gharlan:
> When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow (I have tons of legacy code with both used)

Wouldn't it be better to use `"{$foo}"` then, because it is more flexible? (can be used with array access, properties, methods)
So why not always using `"{$foo...}"`?

PR #3077, opened 2017-09-19
Let configuration files know the ConfigurationResolver

mlocati:
What about letting configuration files know the configuration resolver?

I'd need that because our project uses different rule sets for different files, and letting the configuration know what it's going to be used for would solve this.

I already answer why we need different rules:
1. we use by default the short array syntax and the short echo tags, but we have entry-point files that check the current PHP version and tell the user that they need a newer PHP version (if these files use newer syntax, the old PHP version will fail - see for instance https://3v4l.org/drhYT with PHP 5.3)
2. we use by default the *correct* indentation, but that messes up files with mixed PHP and HTML, so we need to disable indentation fixes for those files
3. we use PSR-4 by default to give the name of the files (file name === base class name), but we have some classes that don't follow that rule, so we sometimes need to disable PSR-4

In order to accomplish that, I currently use `debug_backtrace` to determine the `ConfigurationResolver` instance, but that's a lot hacky...

If you are interested in the current implementation of the above: [here it is](https://github.com/concrete5/concrete5/pull/5971).

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -215,6 +215,8 @@ final class ConfigurationResolver
                     continue;
                 }
 
+                $configurationResolver = $this;
+
                 $config = include $configFile;
 
                 // verify that the config has an instance of Config


julienfalque:
Allowing the tool to apply different rule sets depending on the files being fixed would be a nice feature IMO. But this solution is exposing the `ConfigurationResolver` instance which is supposed to be internal only. I think this is fragile and hacky.

mlocati:
>  I think this is fragile and hacky.

I know, but it's the only solution I can see with the current code...

SpacePossum:
I do however see a good use case for a way to configure different rules for different directories/files. It is a returning thing for SF but the case I always feel justifies something like this is the PHPUnit related fixers. I know if which folder I keep my tests, so no need to apply those fixers on other files.
Other scenario's can be thought off of course.

I don't won't to blow up your PR into a big meta discussion about this topic, however I agree with @julienfalque on this. The resolver should deal with configuration, the other way around is fragile.

mlocati:
So I'll continue to use `debug_backtrace`, which is already fragile and hacky :wink:

PR #3083, opened 2017-09-20
SingleBlankLineBeforeNamespaceFixer - Fix handling namespace right after opening tag

### LINKED ISSUE #3081: single_blank_line_before_namespace doesn't add a blank line if on first line
> The PHP version you are using:

`PHP 7.1.6 (cli) (built: Jun  8 2017 02:06:32) ( ZTS MSVC14 (Visual C++ 2015) x86 )`

> PHP CS Fixer version you are using:

`PHP CS Fixer 2.7.0-DEV by Fabien Potencier and Dariusz Ruminski`

> The command you use to run the fixer:

`php-cs-fixer fix --dry-run --diff --rules=single_blank_line_before_namespace test.php`

> A minimal sample of valid PHP code that is not fixed correctly:

```php
<?php
namespace X;
```

> The fixed version of that code after you run the fixer and the version you expected.

Current output:
```php
<?php
namespace X;
```
Expected output:
```php
<?php

namespace X;
```

I think that the root cause is that `token_get_all` returns a single token for the opening tag (eg the first token is `[T_OPEN_TAG, "<?php\n", 1]`) - see https://3v4l.org/38K9F.

mlocati:
Fix #3081

### DIFF
--- a/src/AbstractLinesBeforeNamespaceFixer.php
+++ b/src/AbstractLinesBeforeNamespaceFixer.php
@@ -46,7 +46,7 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
                 if ($token->isGivenKind(T_OPEN_TAG)) {
                     $openingToken = $token;
                     $openingTokenIndex = $index - $i;
-                    $newlineInOpening = strpos($token->getContent(), "\n") !== false;
+                    $newlineInOpening = false !== strpos($token->getContent(), "\n");
                     if ($newlineInOpening) {
                         ++$precedingNewlines;
                     }
--- a/src/AbstractLinesBeforeNamespaceFixer.php
+++ b/src/AbstractLinesBeforeNamespaceFixer.php
@@ -45,6 +45,7 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
                 $token = $tokens[$index - $i];
                 if ($token->isGivenKind(T_OPEN_TAG)) {
                     $openingToken = $token;
+                    $openingTokenIndex = $index - $i;
                     $precedingNewlinesInOpening = substr_count($token->getContent(), "\n");
                     $precedingNewlinesTotal += $precedingNewlinesInOpening;
                     break;
@@ -70,7 +71,7 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
             }
             // Remove new lines in opening token
             if (0 < $precedingNewlinesInOpening) {
-                $openingToken->setContent(rtrim($openingToken->getContent()).' ');
+                $tokens[$openingTokenIndex] = new Token(array(T_OPEN_TAG, rtrim($openingToken->getContent()).' '));
             }
 
             return;
@@ -81,12 +82,13 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
             $content = rtrim($openingToken->getContent());
             if (0 === $precedingNewlinesInOpening) {
                 // We have an opening tag without new lines: add a new line there
-                $openingToken->setContent($content.$lineEnding);
+                $newContent = $content.$lineEnding;
                 ++$precedingNewlinesInOpening;
             } else {
                 // Use the configured line ending for the PHP opening tag
-                $openingToken->setContent($content.str_repeat($lineEnding, $precedingNewlinesInOpening));
+                $newContent = $content.str_repeat($lineEnding, $precedingNewlinesInOpening);
             }
+            $tokens[$openingTokenIndex] = new Token(array(T_OPEN_TAG, $newContent));
         }
         $newlinesForWhitespaceToken = $expected - $precedingNewlinesInOpening;
         if ($previous->isWhitespace()) {
@@ -96,7 +98,7 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
                 $tokens->clearAt($previousIndex);
             } else {
                 // Fix the previous whitespace token
-                $previous->setContent(str_repeat($lineEnding, $newlinesForWhitespaceToken));
+                $tokens[$previousIndex] = new Token(array(T_WHITESPACE, str_repeat($lineEnding, $newlinesForWhitespaceToken)));
             }
         } elseif (0 < $newlinesForWhitespaceToken) {
             // Add a new whitespace token
--- a/src/AbstractLinesBeforeNamespaceFixer.php
+++ b/src/AbstractLinesBeforeNamespaceFixer.php
@@ -88,18 +88,22 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer implement
             $tokens[$openingTokenIndex] = new Token(array(T_OPEN_TAG, $newContent));
             --$newlinesForWhitespaceToken;
         }
-        if ($previous->isWhitespace()) {
-            if (0 === $newlinesForWhitespaceToken) {
-                // We have all the needed new lines in the opening tag
+        if (0 === $newlinesForWhitespaceToken) {
+            // We have all the needed new lines in the opening tag
+            if ($previous->isWhitespace()) {
                 // Let's remove the previous token containing extra new lines
                 $tokens->clearAt($previousIndex);
-            } else {
-                // Fix the previous whitespace token
-                $tokens[$previousIndex] = new Token(array(T_WHITESPACE, str_repeat($lineEnding, $newlinesForWhitespaceToken)));
             }
-        } elseif (0 < $newlinesForWhitespaceToken) {
+
+            return;
+        }
+        $newWhitespaceToken = new Token(array(T_WHITESPACE, str_repeat($lineEnding, $newlinesForWhitespaceToken)));
+        if ($previous->isWhitespace()) {
+            // Fix the previous whitespace token
+            $tokens[$previousIndex] = $newWhitespaceToken;
+        } else {
             // Add a new whitespace token
-            $tokens->insertAt($index, new Token(array(T_WHITESPACE, str_repeat($lineEnding, $newlinesForWhitespaceToken))));
+            $tokens->insertAt($index, $newWhitespaceToken);
         }
     }
 }
--- a/src/AbstractLinesBeforeNamespaceFixer.php
+++ b/src/AbstractLinesBeforeNamespaceFixer.php
@@ -48,6 +48,7 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer
                 }
             }
         }
+
         $previousIndex = $index - 1;
         $previous = $tokens[$previousIndex];
         if ($previous->isWhitespace()) {
--- a/src/AbstractLinesBeforeNamespaceFixer.php
+++ b/src/AbstractLinesBeforeNamespaceFixer.php
@@ -74,14 +74,16 @@ abstract class AbstractLinesBeforeNamespaceFixer extends AbstractFixer
                     ++$precedingNewlinesInOpening;
                 }
                 $newlinesForWhitespaceToken = $expected - $precedingNewlinesInOpening;
-                if (0 === $newlinesForWhitespaceToken) {
-                    // We have all the needed new lines in the opening tag
-                    // Let's remove the previous token containing extra new lines
-                    $tokens->clearAt($previousIndex);
-                } elseif ($previous->isWhitespace()) {
-                    // Fix the previous whitespace token
-                    $previous->setContent(str_repeat("\n", $newlinesForWhitespaceToken));
-                } else {
+                if ($previous->isWhitespace()) {
+                    if (0 === $newlinesForWhitespaceToken) {
+                        // We have all the needed new lines in the opening tag
+                        // Let's remove the previous token containing extra new lines
+                        $tokens->clearAt($previousIndex);
+                    } else {
+                        // Fix the previous whitespace token
+                        $previous->setContent(str_repeat("\n", $newlinesForWhitespaceToken));
+                    }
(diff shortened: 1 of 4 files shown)


src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -36,13 +36,17 @@ protected function fixLinesBeforeNamespace(Tokens $tokens, $index, $expected)
...
-            $opening = $tokens[$index - 2];
-            if ($opening->isGivenKind(T_OPEN_TAG)) {
-                $expected -= substr_count($opening->getContent(), "\n");
+        for ($i = 1; $i <= 2; ++$i) {
+            if (isset($tokens[$index - $i])) {
+                $opening = $tokens[$index - $i];
+                if ($opening->isGivenKind(T_OPEN_TAG)) {
+                    $expected -= substr_count($opening->getContent(), "\n");
+                    break;
+                } elseif (false === $opening->isGivenKind(T_WHITESPACE)) {
```
julienfalque: This `elseif` can be replaced with an `if`.

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -51,6 +55,8 @@ protected function fixLinesBeforeNamespace(Tokens $tokens, $index, $expected)
             } elseif (substr_count($previous->getContent(), "\n") !== $expected) {
                 $tokens[$previousIndex] = new Token(array(T_WHITESPACE, str_repeat("\n", $expected)));
             }
+        } elseif (0 < $expected && $previous->isGivenKind(T_OPEN_TAG)) {
```
julienfalque: I would not use Yoda style for this kind of comparison.

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -36,13 +36,17 @@ protected function fixLinesBeforeNamespace(Tokens $tokens, $index, $expected)
...
         // contains from the expected number in the following whitespace
-        if (isset($tokens[$index - 2])) {
-            $opening = $tokens[$index - 2];
-            if ($opening->isGivenKind(T_OPEN_TAG)) {
-                $expected -= substr_count($opening->getContent(), "\n");
+        for ($i = 1; $i <= 2; ++$i) {
+            if (isset($tokens[$index - $i])) {
+                $opening = $tokens[$index - $i];
+                if ($opening->isGivenKind(T_OPEN_TAG)) {
+                    $expected -= substr_count($opening->getContent(), "\n");
```
julienfalque: This is assuming that the open tag always ends with `"\n"`, which is not true.

Please add this test case:
```php
array("<?php\n\nnamespace X;", '<?php namespace X;'),
```

src/AbstractLinesBeforeNamespaceFixer.php
keradus: we do it in rest of the repo, let it stay

keradus:
ping @mlocati 
could you please address the review comments ?

src/AbstractLinesBeforeNamespaceFixer.php
mlocati: So, the fixer should also strip the trailing white space in `'<?php '`?

SpacePossum: >So, the fixer should also strip the trailing white space in '<?php '?

that is not needed IMHO, but the test case is to point something else I think, try;

```php
array("<?php \n\nnamespace X;", '<?php namespace X;'),
```

src/AbstractLinesBeforeNamespaceFixer.php
SpacePossum: we don't enforce it in our `.php_cs.dist`, should I enable the option of YodaFixer to start checking for it?

src/AbstractLinesBeforeNamespaceFixer.php
mlocati: > `array("<?php \n\nnamespace X;", '<?php namespace X;'),`

This passes the test

SpacePossum: good to hear :) than I ping @julienfalque for his thoughts on it

julienfalque: > So, the fixer should also strip the trailing white space in `'<?php '`?

Yes that's what I meant, sorry for being unclear.

I fixed a similar case in `header_comment` in #2630 which got merged so I think we should do it here too.

src/AbstractLinesBeforeNamespaceFixer.php
keradus: sure

src/AbstractLinesBeforeNamespaceFixer.php
mlocati: Done

mlocati:
I assumed `"\n"` for new lines because it was already so, but shouldn't we use `$this->whitespacesConfig->getLineEnding()`?

julienfalque:
Indeed, it would be nice to add the same new test cases but with `\r\n` as line ending.

mlocati:
Another question: to count new lines we count the numbers of `"\n"`. That's good when new lines are `"\n"` or `"\r\n"`, but not when new lines are `"\r"`.
What about adding this method to the `Token` class?
```php

    /**
     * Get the number of new lines in the content.
     *
     * @return int
     */
    public function getNumberOfNewLines()
    {
        if ($this->content) {
            $normalizedContent = str_replace("\r", "\n", str_replace("\r\n", "\n", $this->content));
            $result = substr_count($normalizedContent, "\n");
        } else {
            $result = 0;
        }
        return $result;
    }
```

mlocati:
> Indeed, it would be nice to add the same new test cases but with \r\n as line ending.

Done

julienfalque:
> Another question: to count new lines we count the numbers of "\n". That's good when new lines are "\n" or "\r\n", but not when new lines are "\r".

We don't officially support `\r` line endings as they are only used on some old systems, e.g. classic Mac OS (latest release in 2001). I would say it's not worth it.

tests/Test/AbstractFixerTestCase.php
```diff
@@ -110,11 +111,12 @@ protected function getTestFile($filename = __FILE__)
...
-     * @param string            $expected The expected fixer output
-     * @param null|string       $input    The fixer input, or null if it should intentionally be equal to the output
-     * @param null|\SplFileInfo $file     The file to fix, or null if unneeded
+     * @param string                      $expected    The expected fixer output
+     * @param null|string                 $input       The fixer input, or null if it should intentionally be equal to the output
+     * @param null|\SplFileInfo           $file        The file to fix, or null if unneeded
+     * @param null|WhitespacesFixerConfig $whitespaces The white space configuration, or null to use the default
      */
-    protected function doTest($expected, $input = null, \SplFileInfo $file = null)
+    protected function doTest($expected, $input = null, \SplFileInfo $file = null, WhitespacesFixerConfig $whitespaces = null)
```
julienfalque: No need to add this argument IMO. The tested fixer is available via the `fixer` property of the class, so a test can call `$this->fixer->setWhitespacesConfig(...)` before `$this->doTest(...)`.

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -34,22 +35,66 @@
...
-                $tokens->clearAt($previousIndex);
-            } elseif (substr_count($previous->getContent(), "\n") !== $expected) {
-                $tokens[$previousIndex] = new Token(array(T_WHITESPACE, str_repeat("\n", $expected)));
+                // Remove all the previous new lines
+                if ($previous->isWhitespace()) {
+                    $tokens->clearAt($previousIndex);
+                }
+                // Remove new lines in opening token
+                if (0 < $precedingNewlinesInOpening) {
+                    $openingToken->setContent(rtrim($openingToken->getContent()).' ');
```
julienfalque: Don't use `$token->setContent(...)` as it's been deprecated on higher branches. Please replace it with `$tokens[$index] = new Token(...)`.

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -34,22 +35,66 @@
...
+                        $openingToken->setContent($content.str_repeat($lineEnding, $precedingNewlinesInOpening));
+                    }
+                }
+                $newlinesForWhitespaceToken = $expected - $precedingNewlinesInOpening;
+                if ($previous->isWhitespace()) {
+                    if (0 === $newlinesForWhitespaceToken) {
+                        // We have all the needed new lines in the opening tag
+                        // Let's remove the previous token containing extra new lines
+                        $tokens->clearAt($previousIndex);
+                    } else {
```
julienfalque: There are too much nested `if/else` IMO. Isn't it possible to add some early returns?

mlocati:
PR updated

tests/Fixer/NamespaceNotation/NoBlankLinesBeforeNamespaceFixerTest.php
```diff
@@ -27,12 +28,16 @@
...
+     * @param null|WhitespacesFixerConfig $whitespaces
      */
-    public function testFix($expected, $input = null)
+    public function testFix($expected, $input = null, WhitespacesFixerConfig $whitespaces = null)
     {
-        $this->doTest($expected, $input);
+        if (null !== $whitespaces) {
+            $this->fixer->setWhitespacesConfig($whitespaces);
+        }
+        $this->doTest($expected, $input, null, $whitespaces);
```
julienfalque: No need to pass `$whitespaces` here.

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -34,23 +35,74 @@
...
+        $lineEnding = $this->whitespacesConfig->getLineEnding();
+        if (null !== $openingToken) {
+            $content = rtrim($openingToken->getContent());
+            if (0 === $precedingNewlinesInOpening) {
+                // We have an opening tag without new lines: add a new line there
+                $newContent = $content.$lineEnding;
+                ++$precedingNewlinesInOpening;
+            } else {
+                // Use the configured line ending for the PHP opening tag
+                $newContent = $content.str_repeat($lineEnding, $precedingNewlinesInOpening);
```
julienfalque: Not sure you need to use `str_repeat()` as the open tag never contains more than one newline.

mlocati: > Not sure you need to use str_repeat() as the open tag never contains more than one newline.

I agree: I refactored the method to store a boolean value to say if there's a PHP opening tag with a new line or not.

tests/Fixer/NamespaceNotation/NoBlankLinesBeforeNamespaceFixerTest.php
```diff
@@ -44,7 +49,8 @@ public function provideFixCases()
             array("<?php\nnamespace X;"),
             array("<?php\nnamespace X;", "<?php\n\n\n\nnamespace X;"),
             array("<?php\r\nnamespace X;"),
-            array("<?php\r\nnamespace X;", "<?php\r\n\r\n\r\n\r\nnamespace X;"),
+            array("<?php\nnamespace X;", "<?php\r\n\r\n\r\n\r\nnamespace X;"),
+            array("<?php\r\nnamespace X;", "<?php\r\n\r\n\r\n\r\nnamespace X;", new WhitespacesFixerConfig('    ', "\r\n")),
```
julienfalque: Maybe it would be nice to also add the test case with namespace right after the opening tag here, i.e.:
```php
array('<?php namespace X;'),
```

src/AbstractLinesBeforeNamespaceFixer.php
```diff
@@ -46,7 +46,7 @@ protected function fixLinesBeforeNamespace(Tokens $tokens, $index, $expected)
                 if ($token->isGivenKind(T_OPEN_TAG)) {
                     $openingToken = $token;
                     $openingTokenIndex = $index - $i;
-                    $newlineInOpening = strpos($token->getContent(), "\n") !== false;
+                    $newlineInOpening = false !== strpos($token->getContent(), "\n");
                     if ($newlineInOpening) {
                         ++$precedingNewlines;
                     }
```
SpacePossum: commit message made my day :D

mlocati: 😜

tests/Fixer/NamespaceNotation/NoBlankLinesBeforeNamespaceFixerTest.php
mlocati: Do you mean this?

```php
array('<?php\nnamespace X;', '<?php namespace X;'),
```

mlocati: If `NoBlankLinesBeforeNamespaceFixer` should keep `<?php namespace X;` untouched, the `$expected` parameter in 
`fixLinesBeforeNamespace` isn't enough: we need a way to differentiate *I want exactly 1 new line before namespace* and *I want up to 1 new line before namespace*

julienfalque: Ah sorry, I thought the fixer would not change anything in such case. Then you can ignore my previous comment.

src/AbstractLinesBeforeNamespaceFixer.php
keradus: nah ;P sad, you make him ;P
> Write like the Grand Master of the Jedi Order would do, let us.

keradus:
Thank you @mlocati.

SpacePossum:
thanks @mlocati :+1:  :)

PR #3084, opened 2017-09-20
Allow suppressing the xdebug warning

mlocati:
What about suppressing the xdebug warning if a `PHP_CS_FIXER_DISABLE_XDEBUG_WARN` environment variable is set and not falsy?

### DIFF
--- a/src/Console/WarningsDetector.php
+++ b/src/Console/WarningsDetector.php
@@ -48,7 +48,7 @@ final class WarningsDetector
 
     public function detectXdebug()
     {
-        if (extension_loaded('xdebug')) {
+        if (extension_loaded('xdebug') && !getenv('PHP_CS_FIXER_DISABLE_XDEBUG_WARN')) {
             $this->warnings[] = 'You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.';
         }
     }


julienfalque:
:+1: for me but it's a new feature so it must target master branch. Also it would be nice to add a test :)

mlocati:
> it's a new feature so it must target master branch

From [CONTRIBUTING.md](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/CONTRIBUTING.md):
> Default branch when adding new features.

And the [default branch](https://github.com/FriendsOfPHP/PHP-CS-Fixer/branches) is `2.6`: am I missing something or does `CONTRIBUTING.md` need to be fixed?

bendavies:
Have you considered handling this as composer does?
https://github.com/composer/composer/blob/master/src/Composer/XdebugHandler.php

mlocati:
> Have you considered handling this as composer does?

If you mean restarting php-cs-fixer without xdebug: that would be great

bendavies:
exactly

julienfalque:
> If you mean restarting php-cs-fixer without xdebug: that would be great

I'm not in favor of this for two reasons:
- the Composer implementation brought various issues, especially it does not support ini options passed at runtime with `php -d<option>`. Maybe it's fixed now though, didn't check.
- I personnaly think this is hacky, the correct fix should be to disable Xdebug by default and enable it on demand (I published [a little script](https://github.com/julienfalque/xdebug) that can help for that).

keradus:
Personally I also don't like hiding the warning, it's there for a reason.
Executing tool with xdebug enabled makes it dozen of times slower. As we spend a lot of time for performance improvements, big :-1: from me to hide this warning.

About restarting the tool - it's not easy thing with quite a bunch issues. I don't think it's worth the effort.
Instead of proposing a workaround, fix initial issue:
- stop running the tool with XDebug or
- disable XDebug by default or
- remove XDebug in favour phpdbg

If you don't want to do any of them for some reasons, it's just a warning in the end, overall output / exit status is the same.

mlocati:
I don't understand why this is a big problem hiding this warning: it's an opt-in feature... This can be useful for people that don't care about speed (for instance because php-cs-fixer is executed on a single small file), and that intercept the stderr to show *real* errors and not this warning.

keradus:
if output is supposed to be catch by sth and handle then
- rely on exit status
- switch from txt to json/xml format, which we do have BC promise

PR #3088, opened 2017-09-21
Drop dedicated Phar stub

keradus:


keradus:
> we are dropping HHVM and PHP version checks
> we are dropping registration of an error handler

We are **not** dropping them.
On original source, we do have 2 files: `php-cs-fixer` and `src/Resources/phar-stub.php` which are almost the same - difference is registering autoloader from composer or phar map.

This PR unifies file and drop one of them, so phar release will use the same entry point as composer-installed app. in that unified entry point, there are still checks for php versions and error handler registration.

localheinz:
@keradus 

Even better! 👍

php-cs-fixer
```diff
@@ -41,14 +41,22 @@ set_error_handler(function ($severity, $message, $file, $line) {
...
+    // Maybe this file is used as phar-stub? Let's try!
+    Phar::mapPhar('php-cs-fixer.phar');
+    require_once 'phar://php-cs-fixer.phar/vendor/autoload.php';
+} catch (PharException $e) {
+    // OK, it's not, let give Composer autoloader a try!
+    if (file_exists($a = __DIR__.'/../../autoload.php')) {
+        require_once $a;
+    } else {
+        require_once __DIR__.'/vendor/autoload.php';
+    }
```
julienfalque: This empty line should be after the closing brace below.

keradus: done

PR #3089, opened 2017-09-22
NoAliasFunctionsFixer - add imap aliases

Slamdunk:
This is not a bugfix nor a new complete feature: is `master` the right branch?

References:
1. [imap_create](https://secure.php.net/manual/en/function.imap-create.php)
1. [imap_fetchtext](https://secure.php.net/manual/en/function.imap-fetchtext.php)
1. [imap_header](https://secure.php.net/manual/en/function.imap-header.php)
1. [imap_listmailbox](https://secure.php.net/manual/en/function.imap-listmailbox.php)
1. [imap_listsubscribed](https://secure.php.net/manual/en/function.imap-listsubscribed.php)
1. [imap_rename](https://secure.php.net/manual/en/function.imap-rename.php)
1. [imap_scan](https://secure.php.net/manual/en/function.imap-scan.php)
1. [imap_scanmailbox](https://secure.php.net/manual/en/function.imap-scanmailbox.php)

PR #3093, opened 2017-09-23
NoUnneededFinalMethodFixer - Remove final keyword from private methods

localheinz:
This PR

* [x] adjusts the `NoUnneededFinalMethodFixer` to also remove an unnecessary `final` keyword from `private` methods

### DIFF
--- a/src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
+++ b/src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
@@ -36,6 +36,12 @@ final class Foo {
     final protected function bar() {}
     final private function baz() {}
 }'),
+                new CodeSample(
+'<?php
+class Foo {
+    final private function baz() {}
+}'
+                ),
             ]
         );
     }
@@ -61,24 +67,33 @@ final class Foo {
 
             $classOpen = $tokens->getNextTokenOfKind($index, ['{']);
             $prevToken = $tokens[$tokens->getPrevMeaningfulToken($index)];
-            if (!$prevToken->isGivenKind(T_FINAL)) {
-                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $classOpen);
+            $classIsFinal = $prevToken->isGivenKind(T_FINAL);
 
-                continue;
-            }
-
-            $index = $this->fixClass($tokens, $classOpen);
+            $index = $this->fixClass($tokens, $classOpen, $classIsFinal);
         }
     }
 
     /**
      * @param Tokens $tokens
      * @param int    $classOpenIndex
+     * @param bool   $classIsFinal
      *
      * @return int
      */
-    private function fixClass(Tokens $tokens, $classOpenIndex)
+    private function fixClass(Tokens $tokens, $classOpenIndex, bool $classIsFinal)
     {
+        $methodVisibilities = [
+            T_PUBLIC,
+            T_PROTECTED,
+            T_PRIVATE,
+        ];
+
+        if (!$classIsFinal) {
+            $methodVisibilities = [
+                T_PRIVATE,
+            ];
+        }
+
         $tokensCount = count($tokens);
         for ($index = $classOpenIndex + 1; $index < $tokensCount; ++$index) {
             // Class end
@@ -94,6 +109,12 @@ final class Foo {
             }
 
             if (!$tokens[$index]->isGivenKind(T_FINAL)) {
+                continue;
+            }
+
+            $nextMeaningfulToken = $tokens[$tokens->getNextMeaningfulToken($index)];
+
+            if (!$nextMeaningfulToken->isGivenKind($methodVisibilities)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


tests/Fixer/ClassNotation/NoUnneededFinalMethodFixerTest.php
```diff
@@ -51,6 +51,16 @@ private function baz() {}
...
 }',
             ],
+            'private-method' => [
+                '<?php
+class Foo {
+    private function baz() {}
+}',
+                '<?php
+class Foo {
+    final private function baz() {}
```
keradus: `final private` method. kudos for PHP

localheinz: Hehe, actually had an example of this lying around somewhere in **my** code, that's how I figured I should take a look.

keradus:
tests are failing :(

julienfalque:
Please update the PR to target branch 2.6 :)

keradus:
it's a feature imo

keradus:
@julienfalque 

this PR makes one fixer fix more cases, you call it bugfix (targeting 2.6)

#3089 makes one fixer fix mor cases, you call it feature (targeting master)
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3089#pullrequestreview-64574596

julienfalque:
@keradus I consider this one a bugfix because IMO it's something that should have been part of the initial implementation and is missing, while for #3089 it's not the case. Though I'm fine with considering this one as a new feature as well, your call :)

src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
```diff
@@ -61,24 +67,33 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
     {
+        $methodVisibilities = [
+            T_PUBLIC,
+            T_PROTECTED,
+            T_PRIVATE,
+        ];
+
+        if (!$classIsFinal) {
+            $methodVisibilities = [
+                T_PRIVATE,
```
julienfalque: `T_PRIVATE` is always part of the array, I would reverse the condition to initialize it with this value only, and add the others when the class is final.

src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
```diff
@@ -97,6 +112,12 @@ private function fixClass(Tokens $tokens, $classOpenIndex)
                 continue;
             }
 
+            $nextMeaningfulToken = $tokens[$tokens->getNextMeaningfulToken($index)];
+
+            if (!$nextMeaningfulToken->isGivenKind($methodVisibilities)) {
```
julienfalque: This does not cover the case where the visibility keyword is before `final`.

tests/Fixer/ClassNotation/NoUnneededFinalMethodFixerTest.php
```diff
@@ -51,6 +51,16 @@ private function baz() {}
     final private function baz() {}
 }',
             ],
+            'private-method' => [
```
julienfalque: Please add the following test case:

```diff
 class Foo {
-    private final function bar() {}
+    private function bar() {}
 }
```

localheinz:
Picked https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/2.6 as target branch now.

julienfalque:
@localheinz That was actually not needed as @keradus changed the target milestone to `2.7` (current master), which means we consider this a new feature. Sorry for the trouble.

localheinz:
@julienfalque 

No problem!

keradus:
Please rebase and make Travis happy

src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
```diff
@@ -61,34 +67,31 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
             $prevToken = $tokens[$tokens->getPrevMeaningfulToken($index)];
-            if (!$prevToken->isGivenKind(T_FINAL)) {
-                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $classOpen);
+            $classIsFinal = $prevToken->isGivenKind(T_FINAL);
 
-                continue;
-            }
-
-            $index = $this->fixClass($tokens, $classOpen);
+            $this->fixClass($tokens, $classOpen, $classIsFinal);
```
keradus: why this change? like, without `$index = ...` assignment we don't do the jump, so it's slower to finish the main loop. do we need to not do that jump ?

localheinz: If we jump to the end of the class and ignore the content of methods, we cannot fix anonymous class definitions inside methods. 

Do you have a suggestion, @keradus?

keradus: with that reasoning, keep your change :+1:

keradus:
@julienfalque , you added #3094, which got merged.
could you now verify this PR ?

keradus:
Thank you @localheinz.

localheinz:

SpacePossum:
thanks @localheinz and reviewers :D

PR #3094, opened 2017-09-24
NoUnneededFinalMethodFixer - Add test cases

julienfalque:
As spotted in #3093, `final` keyword may be after the visibility keyword. This PR also adds tests with final static methods.


julienfalque:
Added test cases without visibility keyword.

keradus:
Thank you @julienfalque.

Issue #2219, opened 2016-10-12
ClassAttributesSeparationFixer - PHPDoc options


localheinz:
Wondering, is there something similar to the `MethodSeparationFixer` (or has it been proposed), that separates fields _if_ they have a docblock?
### Before

``` php
<?php

namespace Foo;

class Example
{
    /**
     * @var string
     */
    private $foo;
    /**
     * @var int
     */
    private $bar;
}
```
### After

``` php
<?php

namespace Foo;

class Example
{
    /**
     * @var string
     */
    private $foo;

    /**
     * @var int
     */
    private $bar;
}
```

SpacePossum:
>Wondering, is there something similar to the MethodSeparationFixer (or has it been proposed)

AFAIK, no and no, but I would happily like to propose such a fixer together with you :)  
(kindly note: that doesn't mean I have the time to dev the fixer ATM :| )

localheinz:
@SpacePossum 

I'm not sure if I'll be able to build one, but I'm happy to try to do so!

keradus:
hm... i thought we have one. please check master branch

SpacePossum:
>at least for methods

there is, for attributes/properties/fields/const there isn't I think

TomasVotruba:
This is already resolved by #3098, isn't it?

SpacePossum:
it is indeed, at least in part :)

> that separates fields _if_ they have a docblock?

this option is missing. Having that option was requested in 3098 as well (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3098#discussion_r145252212) but I left it out (because of keeping the PR small and not having a lot of time adding the option(s)).

TomasVotruba:
@SpacePossum I miss that in examples.

@localheinz Do you still need this feature? IMO it's very rare to see code like this.

localheinz:
I don't know why or what happened, but the test cases I added passed, so I'm closing this issue.

Thank you!

PR #3098, opened 2017-09-26
MethodSeparationFixer - add class elements separation options

SpacePossum:
This PR adds options to the `MethodSeparationFixer`. New options are; `const` and `property` and `method` is the default.
The separation of this elements is the same as the methods are currently done with one exception;

```php
class Sample
{
    private $a; // this comment is allowed here

    private function doSomething() {
         //
    }  // this comment is still not allowed and moved down
}

```

Minor change is the fixer now allows comments after a method in an interface
Sample:

```php
interface SomeThing
{
    public function doSomething(); // this comment is now allowed
}
```

* There might be a BC break in the TokensAnalyzer, not sure.
* ~We might want to rename the MethodSeparationFixer in 3.0.~

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2219 (cc @localheinz)


src/Fixer/ClassNotation/MethodSeparationFixer.php
```diff
@@ -76,159 +113,187 @@ public function getPriority()
...
-        }
+        $types = ['const', 'method', 'property'];
+
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('elements', sprintf('List of classy elements; \'%s\'.', implode("', '", $types))))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([
+                    (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf($types),
+                ])
+                ->setDefault(['method'])
```
keradus: please add `TODO` for:
> We might want to rename the MethodSeparationFixer in 3.0.

and to switch default behaviour to fix all types

SpacePossum:
`fabbot.io` is false negative because we updated the CS rules of SF set but this not yet part of `fabbot.io`

README.rst
```diff
@@ -645,7 +645,13 @@ Choose from the list of available rules:
 
 * **method_separation** [@Symfony]
 
-  Methods must be separated with one blank line.
```
keradus: we should rename the fixer

src/Fixer/ClassNotation/MethodSeparationFixer.php
```diff
@@ -22,20 +26,30 @@
...
  *
  * @author SpacePossum
  */
-final class MethodSeparationFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+final class MethodSeparationFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
 {
     /**
+     * @var array<string, true>
+     */
+    private $classElementTypes = [];
```
keradus: do we need this property? can't we rely just on `$this->configuration['elements']` directly ?

src/Fixer/ClassNotation/MethodSeparationFixer.php
```diff
@@ -76,159 +113,187 @@ public function getPriority()
...
-        }
+        $types = ['const', 'method', 'property'];
+
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('elements', sprintf('List of classy elements; \'%s\'.', implode("', '", $types))))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([
+                    (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf($types),
+                ])
+                ->setDefault(['method']) // TODO Rename the MethodSeparationFixer in 3.0. Set default to ['const', 'method', 'property']
```
keradus: I would make it already on 2.x line to decrease amount of work needed to migrate 2.x->3.x
that way, on 2.x we will have future-compat layer already.

rename the fixer already and then create `MethodSeparationFixer` once more that will be only a proxy fixer with unmodified behaviour, no configuration options, and will be deprecated.

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -535,8 +535,8 @@ private function findClassyElements(Tokens $tokens, $start)
                 continue;
             }
 
-            if (!$inClass) {
-                $inClass = $token->isClassy();
+            if (false === $inClass) {
+                $inClass = $token->isClassy() ? $index : false;
```
keradus: concern (a little out of the blue):
does this method supports anonymous classes?

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -580,15 +580,27 @@ private function findClassyElements(Tokens $tokens, $start)
             }
 
             if (0 === $bracesLevel && $token->isGivenKind(T_VARIABLE)) {
-                $elements[$index] = ['token' => $token, 'type' => 'property'];
+                $elements[$index] = [
+                    'token' => $token,
+                    'type' => 'property',
+                    'class' => $inClass,
```
keradus: what is here? boolean ?

src/Fixer/ClassNotation/MethodSeparationFixer.php
```diff
@@ -76,159 +113,187 @@ public function getPriority()
...
+        return $tokens->isAnyTokenKindsFound(Token::getClassyTokenKinds());
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function applyFix(SplFileInfo $file, Tokens $tokens)
     {
         $tokensAnalyzer = new TokensAnalyzer($tokens);
+        $class = $classStart = $classEnd = false;
```
keradus: what is `$class` here if not start nor end ?

SpacePossum: `$class` value doesn't matter here, it is to keep SCA happy. Whenever `$class` is used as value it is the index of a classy (`T_CLASS`, `T_INTERFACE`, `T_TRAIT`). It cannot be otherwise unless invalid Tokens collection is used (i.e. tokenized invalid PHP).

src/Tokenizer/TokensAnalyzer.php
SpacePossum: `int`, index in the Tokens collection of the `classy` token

src/Tokenizer/TokensAnalyzer.php
SpacePossum: yes it does, anonymous classes use the same `T_CLASS` token as regular classes;
```bash
$ php test.php 

    0 | T_OPEN_TAG                              | <?php

    1 | T_WHITESPACE                            | 

    2 | T_VARIABLE                              | $a
    3 | T_WHITESPACE                            |  
    4 |                                         | =
    5 | T_WHITESPACE                            |  
    6 | T_NEW                                   | new
    7 | T_WHITESPACE                            |  
    8 | T_CLASS                                 | class
    9 | T_WHITESPACE                            |  
   10 |                                         | {
   11 | T_PUBLIC                                | public
```

src/Fixer/ClassNotation/MethodSeparationFixer.php
keradus: then what's the difference between `$class` and `$classStart` ?

src/Tokenizer/TokensAnalyzer.php
keradus: it's not question in terms of `isClassy` implementation details, but fixer implementation and coverage by tests.

src/Tokenizer/TokensAnalyzer.php
keradus: please add `index` or sth...

src/Fixer/ClassNotation/MethodSeparationFixer.php
SpacePossum: >then what's the difference between $class and $classStart ?

`$class` is provided by the analyzer, `$classStart` is the index of the opening `{` of the class.

src/Tokenizer/TokensAnalyzer.php
SpacePossum: this is the analyzer code, so I don't understand, it is covered by the analyzer tests

src/Fixer/ClassNotation/MethodSeparationFixer.php
SpacePossum: done and removed after additional request

README.rst
SpacePossum: done

src/Fixer/ClassNotation/MethodSeparationFixer.php
SpacePossum:
```
array(3) {
  [0]=>
  string(5) "const"
  [1]=>
  string(6) "method"
  [2]=>
  string(8) "property"
}
```

```
array(3) {
  ["const"]=>
  bool(true)
  ["method"]=>
  bool(true)
  ["property"]=>
  bool(true)
}
```

i.e. direct usage would be `in_array($type, $this->config['elements'], true)` vs. `isset($classElementTypes[$type])`, the last one is faster

src/Tokenizer/TokensAnalyzer.php
SpacePossum: done

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -604,6 +604,8 @@ private function findClassyElements(Tokens $tokens, $start)
             }
         }
 
+        ksort($elements);
```
SpacePossum: TODO: send this bug fix to 2.2 line as well

src/Fixer/ClassNotation/ClassAttributesSeparationFixer.php
```diff
@@ -0,0 +1,395 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        $types = ['const', 'method', 'property'];
+
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('elements', sprintf('List of classy elements; \'%s\'.', implode("', '", $types))))
```
julienfalque: I'm not a big fan of this option, it's not flexible enough IMO. I would replace it with something like:
```php
[
    'const' => 'no_blank_line | single_blank_line | single_blank_line_if_phpdoc',
    'property' => 'no_blank_line | single_blank_line | single_blank_line_if_phpdoc',
    'method' => 'no_blank_line | single_blank_line | single_blank_line_if_phpdoc',
]
```
WDYT?

tests/Fixtures/Integration/priority/ordered_class_elements,class_attributes_separation.test
```diff
@@ -1,7 +1,7 @@
 --TEST--
-Integration of fixers: ordered_class_elements,method_separation.
```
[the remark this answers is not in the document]
SpacePossum: do we keep those for the other deprecated fixers? seems like a lot of copy-paste for not much value to me...

src/Fixer/ClassNotation/ClassAttributesSeparationFixer.php
SpacePossum: I prefer to add one feature in one PR, here allow not _only_ methods but also other class elements to be separated by a single line. When this is done I'm happy to think about more options if there is a demand from a community for it ;)

tests/Fixtures/Integration/priority/ordered_class_elements,class_attributes_separation.test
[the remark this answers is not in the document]
julienfalque: This has value: deprecating a fixer does not mean we allow it to break. If we removed tests for other deprecated fixers, that's a mistake IMO.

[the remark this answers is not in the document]
SpacePossum: Having the fixture test `A+B` and `C+B` where `A === C` has little value, as we know the `A === C` because we test this as `A` is now proxied to `C`, no?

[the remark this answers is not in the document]
julienfalque: `A` being proxied to `C` is an implementation detail. The test is here to proove `A` continues to work as expected, no matter how it is implemented. Testing `C` does not mean `A` uses it correctly.

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -58,8 +58,8 @@ public function provideFixersPriorityCases()
 
         $cases = [
             [$fixers['elseif'], $fixers['braces']],
-            [$fixers['method_separation'], $fixers['braces']],
-            [$fixers['method_separation'], $fixers['indentation_type']],
```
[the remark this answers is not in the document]
keradus: we should

tests/Fixtures/Integration/priority/ordered_class_elements,class_attributes_separation.test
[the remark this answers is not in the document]
keradus: I agree with @julienfalque here
it's extra easy to drop tests for deprecated fixer when it got removed. we keep tests for other deprecated fixers. 
imagine that in future one will change ProxyFixer implementation, or A is proxied to B with configuration. we need to know it works the same as it was, so we change implementation, but keep tests.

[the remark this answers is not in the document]
SpacePossum: so IIUC, I should restore the old fixtures-files and than copy-paste those using the new fixer name and configuration so we've both?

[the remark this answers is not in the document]
julienfalque: Yes.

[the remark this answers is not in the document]
SpacePossum: and how to deal with deprecation notices raised in the integration tests than?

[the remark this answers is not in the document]
julienfalque: Tests with deprecated fixers should be moved to a specific method in the test class with a `@group legacy` annotation. But I think fixers don't trigger deprecation notices so you might not have to change anything. Though we should take care of it in a separate PR.

src/Tokenizer/TokensAnalyzer.php
SpacePossum: > fixer implementation and coverage by tests.

tests added for `anonymous classes` for the fixer implementation

SpacePossum:
interesting 7.2 failure https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/289951049#L697

keradus:
too many deprecations perhaps ?

SpacePossum:
`phpdbg` with `phpunit` coverage runs out of memory, this is a known issue.

keradus:
obviously phpunit with phpdbg/xdebug demands more memory, yet on master we don't demand that much it crashes the build

SpacePossum:
https://github.com/sebastianbergmann/phpunit/issues/2019
and
https://stackoverflow.com/questions/46001411/phpdbg-phpunit-throws-error-even-when-memory-limit-is-1

If I run the `phpdbg phpunit` setup local it crashes as well, running with `xdebug phpunit` setup there is no issue. Not sure how to proceed for now.

keradus:
if we don't get memory issue with phpdbg on master branch and we do on this PR, then I would say this PR requires too much memory. could we check it and shrink it down perhaps ?

SpacePossum:
I don't think the issue is within this PR, I think this PR just pushes the total amount of memory used over the total limit.

src/Tokenizer/TokensAnalyzer.php
keradus: what is the bug actually ?

keradus:
about memory issue.
memory needed by PHPUnit on master: 108MB, on this branch: 108MB

src/Tokenizer/TokensAnalyzer.php
SpacePossum: When there are anonymous classes within a class which also has methods before (or after) the anonymous class, the returned elements are not sorted.
However most fixers that use this method assume the returned `[index, [...]]` array is sorted on the index; low-to-high. When out of order some edge cases will throw the fixers off when they iterate over the returned value and insert/remove tokens.

tests/Fixer/ClassNotation/ClassAttributesSeparationFixerTest.php
```diff
@@ -957,24 +957,44 @@ class A
...
         $this->doTest($expected, $input);
     }
 
-    public function provideFix70Cases()
+    public function provideFix70BigCases()
     {
         $to = $from = '<?php ';
+
         for ($i = 0; $i < 15; ++$i) {
             $from .= sprintf('class A%d{public function AA%d(){return new class {public function BB%d(){}};}public function otherFunction%d(){}}', $i, $i, $i, $i);
```
SpacePossum: note to self: `%1$d` should be used and not multiple `$i` as args.

tests/Fixer/ClassNotation/ClassAttributesSeparationFixerTest.php
```diff
@@ -0,0 +1,1094 @@
...
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFix70BigCases()
+    {
+        $to = $from = '<?php ';
+
+        for ($i = 0; $i < 15; ++$i) {
+            $from .= sprintf('class A%d{public function AA%d(){return new class {public function BB%d(){}};}public function otherFunction%d(){}}', $i, $i, $i, $i);
```
SpacePossum: note to self: %1$d should be used and not multiple $i as args.
but more important, this blows up the memory usage of the fixer to extreme values

keradus: actually, string interpolation is better than sprintf on php 7

keradus: 3199 is supposed to fix it? if so, please cherry-pick it to this PR and revert my "big test" so we will see would it work indeed ;)

keradus:
Thank you @SpacePossum.

TomasVotruba:
Great job @SpacePossum :heart:  , finally can drop my property and constant fixer

SpacePossum:
Thanks for the ping back @TomasVotruba , always makes me happy to hear people like something I PR'ed :)  :+1:

keradus:
hehe, hint for next time @TomasVotruba : help with a review, especially when you already played with familiar concept, and feature would be finished sooner ;)

TomasVotruba:
@keradus I don't follow this repo, but feel free to ping me if you'd like my review ;)

keradus:
so dead PR ;P
my comment was connected to yours
> finally can drop my property and constant fixer

we have no clue what are those, if you have your custom fixer that you don't want to contribute, yet want to have it done by someone else to main repo, review is helpful.

TomasVotruba:
Sure, just ping me if need my review

PR #3100, opened 2017-09-27
NativeFunctionInvocationFixer - Fix test if previous token is already namespace separator

SpacePossum:


keradus:
Thank you @SpacePossum.

PR #3107, opened 2017-09-28
ConfigurationResolver - improve error message when rule is not found

SpacePossum:


tests/Console/ConfigurationResolverTest.php
```diff
@@ -986,12 +986,12 @@ public function testResolveRulesWithUnknownRules()
     {
         $this->setExpectedException(
             \PhpCsFixer\ConfigurationException\InvalidConfigurationException::class,
-            'The rules contain unknown fixers (bar).'
+            'The rules contain unknown fixers ("bar", "binary_operator_space" maybe use "binary_operator_spaces").'
```
keradus: Maybe:
```diff
-'The rules contain unknown fixers ("bar", "binary_operator_space" maybe use "binary_operator_spaces", "baz").'
+'The rules contain unknown fixers ("bar", "binary_operator_space" (did you mean "binary_operator_spaces"?), "baz").'
```

src/Console/ConfigurationResolver.php
```diff
@@ -702,13 +702,32 @@ private function validateRules(array $rules)
...
+                    $unknownFixer,
+                    null === $alternative ? '' : ' maybe use "'.$alternative.'"'
+                );
+            }
+
+            throw new InvalidConfigurationException(substr($message, 0, -2).').');
         }
     }
 
+    private function getAlternativeRule($unknownFixer, array $availableFixers)
```
keradus: DRY is violated

src/Console/ConfigurationResolver.php
```diff
@@ -702,13 +702,32 @@ private function validateRules(array $rules)
...
+    {
+        foreach ($availableFixers as $availableFixer) {
+            $distance = levenshtein($unknownFixer, $availableFixer);
+            if ($distance < 3) {
+                return $availableFixer;
```
keradus: what if there are multiple candidates matching `< 3` ? we shall pick most similar

SpacePossum: Together with the DRY comment; I followed the version of the `describe` command.
I can cook something up that fixes this at both places, but I wanted to prevent to introduce a new `util` class kind of thing.

tests/Console/ConfigurationResolverTest.php
SpacePossum: I played around with the copy and picked the current one as it is more compact when multiple unknown rules are used. However your suggestion looks better for when one or two unknown rules are passed.

keradus: for me it was like:
>  "xxx" maybe use "yyy"

no! I want to use XXX, why would I need to use YYY?

> "xxx" (did you mean "yyy"?)

oh, I made a typo !

src/Console/ConfigurationResolver.php
keradus:
```php
final class LevenshteinMatcher // go ahead and propose better
{
    /**
     * @var string[]
     */
    private $dict;

    public __construct(array $dict)
    {
        $this->dict = $dict;
    }

    public function match(string $value): string
    {
        // magic here
    }
}
```

```php
$matcher = new LevenshteinMatcher($availableFixers);
$bestMatch = $matcher->match($unknownFixer);
```

tests/Console/ConfigurationResolverTest.php
julienfalque: :+1: for @keradus' suggestion, I would even remove the main parentheses:
```diff
-The rules contain unknown fixers ("bar", "binary_operator_space" maybe use "binary_operator_spaces", "baz").
+The rules contain unknown fixers: "bar", "binary_operator_space" (did you mean "binary_operator_spaces"?), "baz".
```

src/Console/ConfigurationResolver.php
keradus: maybe https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3113 ?

src/Console/ConfigurationResolver.php
keradus: Please drop this method and use `WordMatcher` instead.
(branch rebase is needed)

SpacePossum:
reworked

SpacePossum:
targeting 2.2 now, so AppVeyor fails because it cannot deal with retargets

julienfalque:
Pushing again (e.g. with a commit amend) should fix AppVeyor failure.

keradus:
Thank you @SpacePossum.

Issue #3075, opened 2017-09-18
Yoda-style is inconsistent when using comparison operators


nicolas-grekas:
I just spotted that `count($groups) === 0`
is turned into `0 === count($groups)` when yoda style is enabled,
but `count($groups) > 0` is left unchanged,
whereas I would have expected `0 < count($groups)`.
That's a very minor issue so if it's not worth it to you, feel free to close :)

SpacePossum:
I left it in the `don't care so "null"` for SF the project (and default `SF ruleset` ) because of some discussion where people stated that Yoda is to prevent accidental assigned and that is never the case with `<` for example.

However, I did throw in an option. So from the docs:
```
yoda_style [@Symfony]

Write conditions in Yoda style or not based on configuration.

Configuration options:

equal (bool, null): change equal (==, !=) statements; defaults to true
identical (bool, null): change identical (===, !==) statements; defaults to true
less_and_greater (bool, null): change less and greater than (<, <=, >, >=) statements; defaults to null
```

key here is the `less_and_greater` option. 
So if you this changed, feel free to update the configuration of your project :)
Or if you are sure this is what should be the case for all SF projects we could adjust the `SF ruleset`.

SpacePossum:
Closing as I think this one is answered, i.e. the inconsistency can be resolved using different configuration.
Please @nicolas-grekas let me know if I missed something.

PR #3108, opened 2017-09-28
[CS] Yoda `less and greater`

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3083#discussion_r141382247

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -47,7 +47,11 @@ $config = PhpCsFixer\Config::create()
         'single_line_comment_style' => true,
         'strict_comparison' => true,
         'strict_param' => true,
-        'yoda_style' => true,
+        'yoda_style' => [
+            'equal' => true,
+            'identical' => true,
+            'less_and_greater' => true,
+        ],
     ])
     ->setFinder(
         PhpCsFixer\Finder::create()
--- a/src/AbstractAlignFixerHelper.php
+++ b/src/AbstractAlignFixerHelper.php
@@ -87,7 +87,7 @@ abstract class AbstractAlignFixerHelper
             $linesWithPlaceholder[$blockSize] = [];
 
             foreach ($lines as $index => $line) {
-                if (substr_count($line, $placeholder) > 0) {
+                if (0 < substr_count($line, $placeholder)) {
                     $linesWithPlaceholder[$blockSize][] = $index;
                 } else {
                     ++$blockSize;
@@ -96,7 +96,7 @@ abstract class AbstractAlignFixerHelper
             }
 
             foreach ($linesWithPlaceholder as $group) {
-                if (count($group) < 1) {
+                if (1 > count($group)) {
                     continue;
                 }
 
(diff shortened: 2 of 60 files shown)


.php_cs.dist
```diff
@@ -47,7 +47,11 @@ $config = PhpCsFixer\Config::create()
         'single_line_comment_style' => true,
         'strict_comparison' => true,
         'strict_param' => true,
-        'yoda_style' => true,
+        'yoda_style' => [
+            'equal' => true,
+            'identical' => true,
+            'less_and_greater' => true,
+        ],
```
keradus: does it mean we are more strict than `@Symfony` ruleset is ?

SpacePossum: yes, I didn't propose these options, only the default.

keradus: ok, so it's more strict not because SF want to use default option, but because SF didn't make a decision what they want to use. thanks for clarification ;)

SpacePossum: yeah, that sounds about right. I never told them about the other options ;)

keradus: then, maybe best approach would be to notify them, see if they want to actually adjust ruleset, and then use it as well?
to not have "our own way"

SpacePossum: @nicolas-grekas do you want, or know if SF community/maintainers want, to fix `greater`/`less` compares as well? (like the case you described @ https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3075)

nicolas-grekas: I would, dunno about others.

SpacePossum: alrighty, thanks :)
against my own principles I'm going to ping some others in this case, @stof @fabpot if you've time and care about this topic please let us know, no rush and sorry for the ping, thanks!

fabpot: Yoda conditions are only useful for equality comparisons, where `==` could be inadvertently replaced by `=`. As conditions do not have this issue and because it makes the code less readable, I would not use Yoda conditions for comparisons.

SpacePossum: OK, I'll leave SF (the repo and the RuleSet) alone than.
I agree with Fabien here and would not apply the rule to this repo either (and thus not add it to the RuleSet of it). However that call is up to @keradus (@julienfalque do you've a preference?).

keradus: IIRC, for each options we have `true|false|null`, like `yoda, non-yoda, ignore`.

In context of Symfony - if we don't want to enforce `yoda` for `less_and_greater`, maybe we want to enforce `non-yoda` for them?

nicolas-grekas: please don't enforce non-yoda, there is no reason to do so

julienfalque: IMO comparisons with less/greater operator are harder to read with Yoda style so I would not apply it for such comparisons. Though I also agree with @nicolas-grekas, I would not force them to be non-Yoda either. Basically, I like the default options of the fixer :)

SpacePossum:
Closing after discussion, thanks all! :+1:  :)

.php_cs.dist
keradus: so you said you find yoda for `>` hard to read, but you don't want to have non-yoda for them as well

nicolas-grekas: `if (123 < $a && $a < 256) {`
I'd like to be able to write this, yes.

PR #3110, opened 2017-09-28
StrictMethodsFixer

veewee:
Hi,

Since PHP 7 it doesnt make much sense anymore to have "@param" and "@return" tags in the docblocks. Lots of people are dropping those tags in favour for strict types. It would be great if we could use a fixer that is capable of adding strict types to the method and in the same time remove the annotation.

e.g:

From:
```php
/**
 * @param SomeClass $classes
 * @return SomeClass
 */
function x($class)
{
}
```

To:
```php
function x(SomeClass $class): SomeClass
{
}
```

I've started working on something, but there still are some issues to be resolved (it's a bit slow according to the tests, namespaces/alias support, edge cases ...)

If you think this could be of any value, feel free to give me some feedback and I'll try to work something out within my limited timeframe. If not: no problem, I had a lot of fun digging into the code of this project :)

### DIFF
--- a/src/Fixer/Strict/StrictMethodsFixer.php
+++ b/src/Fixer/Strict/StrictMethodsFixer.php
@@ -1,0 +1,261 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Strict;
+
+use PhpCsFixer\AbstractFunctionReferenceFixer;
+use PhpCsFixer\DocBlock\Annotation;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author VeeWee <toonverwerft@gmail.com>
+ */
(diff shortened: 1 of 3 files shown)


src/Fixer/Strict/StrictMethodsFixer.php
```diff
@@ -0,0 +1,297 @@
...
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $this->useAnnotatedTypesAsStrictTypes($tokens, $index);
```
keradus: this will extend the collection you are iterating over, making the loop not visiting all items.

if possible, iterate backwards.
if not (due to strong performance/simplicity reasons), switch to regular `for` loop and manually adjust `$limit` to `$tokens->count()` after calling this method

SpacePossum:
Hi @veewee ,

Thanks for the PR, I really like the idea! 
Currently we (the maintainers) are a bit busy so I'm sorry we didn't got back on your PR sooner.
You did mention there are a feel edge-cases that needs resolving, could you pick those up?
Let me know if you've more detailed questions about the project which we can help you with, like to see this finished :)

veewee:
Thanks for the feedback @SpacePossum. Currently I also lack the time to finish the PR. Lets keep it open and I'll come back at it once I find more time.

julienfalque:
I'm thinking it would be nice to split the fixer introduced here into two separate ones:
* a fixer that adds parameters and return values type-hints from PHPDoc
* a fixer that removes redundant PHPDoc that don't add useful information (same type as the actual type-hints and no desription of parameters or returns values)

WDYT?

keradus:
@veewee , after #3276 is done, I believe we could focus on this one now :D

veewee:
@keradus  yeah ... :)
@juliendufresne: Good argument. That's probably the way to go since it makes it possible to upgrade the target codebase in easy to review steps.

julienfalque:
I'm closing this PR as all desired functionnalities have been implemented as the following rules:
- `phpdoc_to_param_type`
- `phpdoc_to_return_type`
- `no_superfluous_phpdoc_tags`

Thanks @veewee for the time you invested on this anyway.

PR #3104, opened 2017-09-27
DoctrineAnnotationIndentationFixer - Fix str_repeat() error

julienfalque:
Fixes #3102.

@tsufeki Could you confirm that the fix works on your codebase please?

PR #3111, opened 2017-09-28
DoctrineAnnotationIndentationFixer - Restore test case

julienfalque:
Test case added in #3104 and merged in branch 2.2 is missing in branch 2.6.


keradus:
Thank you @julienfalque.

PR #3113, opened 2017-09-29
Add WordMatcher

keradus:


src/WordMatcher.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @param string $needle
+     *
+     * @return null|string
+     */
+    public function match($needle)
+    {
+        $word = null;
+        $distance = 7;
```
julienfalque: `levenshtein('braces', 'elseif') // 6`

Maybe this value should be computed from the length of the compared words?

keradus: I tried to adjust it with way how you proposed
I have feeling that this value will be magic to determine ;)

src/WordMatcher.php
```diff
@@ -0,0 +1,56 @@
...
+ * @author SpacePossum
+ *
+ * @internal
+ */
+final class WordMatcher
+{
+    /**
+     * @var string[]
+     */
+    private $dict;
```
julienfalque: `$dict` is a bit confusing IMO, why not simply `$candidates`?

keradus: both good, changing

src/WordMatcher.php
```diff
@@ -0,0 +1,59 @@
...
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * Base class for function reference fixers.
```
julienfalque: Description looks wrong.

tests/WordMatcherTest.php
```diff
@@ -0,0 +1,79 @@
...
+final class WordMatcherTest extends TestCase
+{
+    /**
+     * @param null|string $expected
+     * @param string      $needle
+     * @param array       $dict
+     *
+     * @dataProvider provideMatchCases
+     */
+    public function testMatch($expected, $needle, array $dict)
```
julienfalque: I would rename `$dict` here as well.

PR #3114, opened 2017-10-01
SelfAccessorFixer - Fix type declarations replacement

### LINKED ISSUE #3079: SelfAccessorFixer doesn't replace the class name with self when used in type hints
PHP version: 7.1.9
PHP CS Fixer version: 2.6.0

The `SelfAccessorFixer` doesn't replace the class name with the `self` keyword when used in type hints.

Actual result:
```php
class Foo
{
    public function bar(Foo $foo): Foo
    {
        // ...
    }
}
```

Expected result:
```php
class Foo
{
    public function bar(self $foo): self
    {
        // ...
    }
}
```

julienfalque:
Fixes #3079.

### DIFF
--- a/src/Cache/SignatureInterface.php
+++ b/src/Cache/SignatureInterface.php
@@ -39,5 +39,5 @@ interface SignatureInterface
      *
      * @return mixed
      */
-    public function equals(SignatureInterface $signature);
+    public function equals(self $signature);
 }
--- a/src/Fixer/ClassNotation/SelfAccessorFixer.php
+++ b/src/Fixer/ClassNotation/SelfAccessorFixer.php
@@ -140,7 +140,7 @@ class Sample
                 || (
                     null !== $insideMethodSignatureUntil
                     && $i < $insideMethodSignatureUntil
-                    && $prevToken->equalsAny(['(', ',', [CT::T_TYPE_COLON]])
+                    && $prevToken->equalsAny(['(', ',', [CT::T_TYPE_COLON], [CT::T_NULLABLE_TYPE]])
                 )
             ) {
                 $tokens[$i] = new Token([T_STRING, 'self']);
--- a/src/Fixer/ClassNotation/SelfAccessorFixer.php
+++ b/src/Fixer/ClassNotation/SelfAccessorFixer.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Fixer\ClassNotation;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use PhpCsFixer\Tokenizer\TokensAnalyzer;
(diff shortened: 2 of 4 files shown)


GrahamCampbell:
👍

keradus:
why not 2.2 ?

julienfalque:
Because of [PHP bug #60573](https://bugs.php.net/bug.php?id=60573) which was fixed in PHP 5.4.1. As 2.2 supports PHP 5.3, it would be affected by the bug because of the changes applied by the fixer. An alternative would be to disable the fixer on the project so we can target 2.2.

SpacePossum:
:+1: for disabling the fixer when running on PHP 5.4.1 (and remove support for PHP 5.4.1 in composer.json)

keradus:
2.2 shall not bump php requirement

julienfalque:
We cannot disable the fixer conditionally as we can't apply the change without breaking the tool on PHP 5.3: CI builds would fail on higher versions, where the fixer would still be enabled and try to apply the change again.

SpacePossum:
>2.2 shall not bump php requirement

no one suggested that, I just suggest we black list `5.4.1`, so no bump

>We cannot disable the fixer conditionally

we can check in the `isCandidate` is running on PHP 5.4.1 no, and on CI we can pick any other PHP 5.4.x version but 5.4.1?

keradus:
bug was not present only in `5.4.1`, it was present in `5.4.1` and lower, won't it ?

SpacePossum:
hmm indeed, 

`5.1.6 - 5.4.0, hhvm-3.10.1 - 3.22.0`
https://3v4l.org/3dLV3

do not raise the warning during linting

julienfalque:
The issue is [the change in `SignatureInterface`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3114/files#diff-331e87320e6c637bbdb0c39cd6f1b64eR42) that triggers the error on PHP lower than 5.4.1.

SpacePossum:
@Korikulum this PR should resolve your issue, do you maybe have time to review or test as well?

Korikulum:
@SpacePossum Hi, I've tested the changes & it seems that it still doesn't work when the type declarations are nullable, eg:

```php
class Foo
{
    public function bar(?Foo $foo): ?Foo {}
}
```

Other then that, everything is working fine.

Korikulum:
@SpacePossum And I just realized I was testing on php 7, not 7.1, so I have to correct myself again, `?self` does work in PHP, so my initial statement was correct, the fix doesn't work when the type declarations are nullable:

```php
class Foo
{
    public function bar(?Foo $foo): ?Foo {}
}
```

Sorry about the confusion.

SpacePossum:
ping @julienfalque :)

julienfalque:
Is there anything left to do here? `?self` [seems to work fine](https://3v4l.org/RLvaV).

SpacePossum:
I think, IIUC, that this test is missing:

```php
            [
                 '<?php interface Foo { public function bar(): ?self; }',
                 '<?php interface Foo { public function bar(): ?Foo; }',
             ],
```

julienfalque:
Test added (and fixed as it was actually failing).

SpacePossum:
Looks solid to me.
Can I bother you; the last fix has a `class` variant for fixing proving the fixer works correctly, can we've a `interface` with comment- and whitespace-mess as well? (something like `<?php interface D{public function a()\t/**/?:/**/Foo;`})
after that and test are green :+1: lets :ship:

julienfalque:
Test added.

keradus:
Thank you @julienfalque.

PR #3115, opened 2017-10-02
NoUnneededFinalMethodFixer - fix edge case

Slamdunk:
~Hope to find the fix soon~ Fixed.
#3093 introduced a bug that weren't there before.

### DIFF
--- a/src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
+++ b/src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
@@ -120,7 +120,7 @@ class Foo {
      */
     private function isPrivateMethod(Tokens $tokens, $index, $classOpenIndex)
     {
-        $index = max($classOpenIndex + 1, $tokens->getPrevTokenOfKind($index, ['{', '}']));
+        $index = max($classOpenIndex + 1, $tokens->getPrevTokenOfKind($index, [';', '{', '}']));
 
         while (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
             if ($tokens[$index]->isGivenKind(T_PRIVATE)) {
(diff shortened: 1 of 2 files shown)


Slamdunk:
@keradus hope you can push a subrelease soon: pretty every Template Method Pattern usage with constants, traits or variables are affected.

tests/Fixer/ClassNotation/NoUnneededFinalMethodFixerTest.php
```diff
@@ -211,6 +211,9 @@ class Foo {
             'abstract-class' => [
                 '<php abstract class Foo { final public function foo() {} }',
             ],
+            'abstract-construct' => [
+                '<?php abstract class Foo { private static $var; final public function foo() {} }',
```
localheinz: This actually affects all these cases where we have a `final` method that follows a `private` property - shall we rename the case, then?

src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
```diff
@@ -120,7 +120,7 @@ private function fixClass(Tokens $tokens, $classOpenIndex, $classIsFinal)
      */
     private function isPrivateMethod(Tokens $tokens, $index, $classOpenIndex)
     {
-        $index = max($classOpenIndex + 1, $tokens->getPrevTokenOfKind($index, ['{', '}']));
+        $index = max($classOpenIndex + 1, $tokens->getPrevTokenOfKind($index, [';', '{', '}']));
```
localheinz: Apologies for the problems, @Slamdunk, it was my mistake here not to think of examples where this can go wrong.

tests/Fixer/ClassNotation/NoUnneededFinalMethodFixerTest.php
Slamdunk: Done

src/Fixer/ClassNotation/NoUnneededFinalMethodFixer.php
Slamdunk: No need to apologize, as the creator of this Fixer I reviewed your code and didn't notice it until I applied the fixer to the codebase

localheinz: Great fixer, by the way!

keradus:
Thank you @Slamdunk.

PR #3118, opened 2017-10-02
Fix CS in cookbook

ntzm:


keradus:
it was there on purpose with later request to call fixer itself to fix all the work

ntzm:
Ah, that makes sense. Sorry!

SpacePossum:
@ntzm thanks for PR'ing, I thought the same until @keradus pointed it out, no need for sorries here :D

keradus:
remember, CS we use are changing during time, no need to always update sample code, if we could just ask to run the tool and it will fix not only empty skeleton of fixer, but also whole user-added logic

Issue #3122, opened 2017-10-04
Cache - Permission issues


dominikzogg:
When reporting an issue (bug) please provide the following information:

The PHP version you are using:
PHP 7.0.22-0ubuntu0.16.04.1 (cli) ( NTS )

PHP CS Fixer version you are using:
PHP CS Fixer 2.7.0 Sandy Pool by Fabien Potencier and Dariusz Ruminski (e4e93a1

The command you use to run the fixer (and the configuration file you are using if any).
```
$ php-cs-fixer fix --rules=@Symfony src
You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.
If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
Loaded config default.
Using cache file ".php_cs.cache".

Fixed all files in 2.525 seconds, 10.000 MB memory used

                                                                
  [Symfony\Component\Filesystem\Exception\IOException]          
  Failed to write file ".php_cs.cache", "no reason available".  
                                                                

fix [--path-mode PATH-MODE] [--allow-risky ALLOW-RISKY] [--config CONFIG] [--dry-run] [--rules RULES] [--using-cache USING-CACHE] [--cache-file CACHE-FILE] [--diff] [--format FORMAT] [--stop-on-violation] [--show-progress SHOW-PROGRESS] [--] [<path>]...
```

SpacePossum:
Hi and thanks for reporting.

The PHP CS Fixer is, most likely, not allowed to write a cache file which triggers this exception.
I wonder if we can improve the exception message, or is there something else you would like to see changed?

dominikzogg:
@SpacePossum i understand the message, but not the reason. the file get touched by the command, but not filled with data.

```
default /vagrant/chubbyphp-deserialization vagrant $ ls -la
total 108
drwxrwxr-x  7 vagrant vagrant  4096 Sep 26 14:08 .
drwxrwxr-x 30 vagrant vagrant  4096 Sep 18 18:30 ..
-rw-rw-r--  1 vagrant vagrant   743 Sep 26 14:02 composer.json
-rw-rw-r--  1 vagrant vagrant 52917 Mai 14 18:09 composer.lock
drwxrwxr-x  8 vagrant vagrant  4096 Okt  4 07:56 .git
-rw-rw-r--  1 vagrant vagrant    53 Sep 26 14:02 .gitignore
[...]

PR #3123, opened 2017-10-04
Cache - File permissions

SpacePossum:
lets try on `master` first and backport if effective

### DIFF
--- a/src/Cache/FileHandler.php
+++ b/src/Cache/FileHandler.php
@@ -60,13 +60,36 @@ final class FileHandler implements FileHandlerInterface
     {
         $content = $cache->toJson();
 
+        if (file_exists($this->file)) {
+            if (is_dir($this->file)) {
+                throw new IOException(
+                    sprintf('Cannot write cache file "%s" as the location exists as directory.', realpath($this->file)),
+                    0,
+                    null,
+                    $this->file
+                );
+            }
+
+            if (!is_writable($this->file)) {
+                throw new IOException(
+                    sprintf('Cannot write to file "%s" as it is not writable.', realpath($this->file)),
+                    0,
+                    null,
+                    $this->file
+                );
+            }
+        } else {
+            @touch($this->file);
+            @chmod($this->file, 0666);
+        }
+
         $bytesWritten = @file_put_contents($this->file, $content, LOCK_EX);
 
         if (false === $bytesWritten) {
             $error = error_get_last();
 
             throw new IOException(
-                sprintf('Failed to write file "%s", "%s".', $this->file, $error ? $error['message'] : 'no reason available'),
+                sprintf('Failed to write file "%s", "%s".', $this->file, isset($error['message']) ? $error['message'] : 'no reason available'),
                 0,
                 null,
                 $this->file
--- a/tests/Cache/FileHandlerTest.php
+++ b/tests/Cache/FileHandlerTest.php
@@ -147,6 +147,78 @@ final class FileHandlerTest extends TestCase
         $this->assertSame($cache->toJson(), $actualCacheJson);
     }
 
+    public function testWriteCacheToDirectory()
+    {
+        $dir = __DIR__.'/../Fixtures/cache-file-handler';
+
+        $handler = new FileHandler($dir);
+
+        $this->setExpectedExceptionRegExp(
+            'Symfony\Component\Filesystem\Exception\IOException',
+            sprintf('#^%s$#', preg_quote('Cannot write cache file "'.realpath($dir).'" as the location exists as directory.', '#')
+        ));
+
+        $handler->write(new Cache(new Signature(
+            PHP_VERSION,
+            '2.0',
+            array(
+                'foo',
+                'bar',
+            )
+        )));
+    }
+
+    public function testWriteCacheToNonWriteableFile()
+    {
+        $file = __DIR__.'/../Fixtures/cache-file-handler/cache-file';
+        if (is_writable($file)) {
+            $this->markTestSkipped(sprintf('File "%s" must be not writeable for this tests.', realpath($file)));
+
+            return;
+        }
+
+        $handler = new FileHandler($file);
+
+        $this->setExpectedExceptionRegExp(
+            'Symfony\Component\Filesystem\Exception\IOException',
+            sprintf('#^%s$#', preg_quote('Cannot write to file "'.realpath($file).'" as it is not writable.', '#')
+        ));
+
+        $handler->write(new Cache(new Signature(
+            PHP_VERSION,
+            '2.0',
+            array(
+                'foo',
+                'bar',
+            )
+        )));
+    }
+
+    public function testWriteCacheFilePermissions()
+    {
+        $file = __DIR__.'/../Fixtures/cache-file-handler/rw_cache.test';
+        @unlink($file);
+
+        $this->assertFileNotExists($file);
+
+        $handler = new FileHandler($file);
+        $handler->write(new Cache(new Signature(
+            PHP_VERSION,
+            '2.0',
+            array(
+                'foo',
+                'bar',
+            )
+        )));
+
+        $this->assertFileExists($file);
+        $this->assertTrue(@is_file($file), sprintf('Failed cache "%s" `is_file`.', $file));
+        $this->assertTrue(@is_writable($file), sprintf('Failed cache "%s" `is_writable`.', $file));
+        $this->assertTrue(@is_readable($file), sprintf('Failed cache "%s" `is_readable`.', $file));
+
+        @unlink($file);
+    }
+
     /**
      * @return string
      */


keradus:
why? what's the downside of having it on 2.2 directly ?

src/Cache/FileHandler.php
```diff
@@ -60,6 +60,15 @@ public function write(CacheInterface $cache)
     {
         $content = $cache->toJson();
 
+        if (file_exists($this->file) && !is_writable($this->file)) {
```
keradus: `file_exists` will return true for dir as well, use `is_file`

src/Cache/FileHandler.php
```diff
@@ -72,5 +81,7 @@ public function write(CacheInterface $cache)
                 $this->file
             );
         }
+
+        @chmod($this->file, 0666 & ~umask());
```
keradus: why with umask and not just 0666 ?

keradus: why setting the permission of file after we already write to it ?

SpacePossum: When you run as `root` with very strict settings the `cache` file might be created with write permissions for `root` only. When you than rerun as a regular user the `cache` cannot be modified.

keradus: but 666 gave permission to all users, root or not

SpacePossum: We never set the permissions to `0666` I think? Using `0666` with `& ~umask()` should give `0666` or lower when configured by user.

SpacePossum:
reworked on 2.2 and updated following feedback

src/Cache/FileHandler.php
keradus: I'm not sure what will be outcome of umask here TBH, never had to use it.
IMO 666 for cache is ok. it's only cache

SpacePossum: yeah, `0666` will work, I've updated the PR to it.

src/Cache/FileHandler.php
```diff
@@ -60,17 +60,39 @@ public function write(CacheInterface $cache)
...
             throw new IOException(
-                sprintf('Failed to write file "%s", "%s".', $this->file, $error ? $error['message'] : 'no reason available'),
+                sprintf('Failed to write file "%s", "%s".', $this->file, isset($error['message']) ? $error['message'] : 'no reason available'),
                 0,
                 null,
                 $this->file
             );
         }
+
+        @chmod($this->file, 0666);
```
keradus: I raised 2 concerns here, one about umask, which you answered, and 2nd one:
> why setting the permission of file after we already write to it ?

SpacePossum: >why setting the permission of file after we already write to it ?

If I run the fixer using `sudo` on Ubuntu box with default `umask` the cache file is created without write permissions for users. Like reported in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3122
If you run the fixer again without `sudo` it will fail to write to the previous created cache file because of lack of permission to do so.

keradus: it needs to be done only after file creation, not after every single writing to file.

SpacePossum: :+1: done

src/Cache/FileHandler.php
```diff
@@ -60,13 +60,36 @@ public function write(CacheInterface $cache)
     {
         $content = $cache->toJson();
 
+        if (file_exists($this->file)) {
+            if (is_dir($this->file)) {
```
keradus: so. now the sad part.

I think it could be somehow groomed and simplified.
I could even try to do it.
Yet, after my changes, I have no clue would it work or not.
Why we need to manually check `is_dir` to raise error instead of relying on error after `file_put_contents` ? Maybe we don't?
we shall cover the logic here with tests.

julienfalque: Maybe [vfsStream](https://github.com/mikey179/vfsStream) could help writing such tests?

SpacePossum: Than lets bin this PR until someone has the time to write a fix and tests for this.

SpacePossum:
tests added

tests/Cache/FileHandlerTest.php
```diff
@@ -26,6 +27,8 @@
  */
 final class FileHandlerTest extends TestCase
 {
+    use FileSystemAssertTrait;
```
keradus: not on 5.3,
which functionality of this trait is used in this file ?

tests/Cache/FileHandlerTest.php
```diff
@@ -145,6 +148,76 @@ public function testWriteWritesCache()
...
+            array(
+                'foo',
+                'bar',
+            )
+        )));
+    }
+
+    public function testWriteCacheToNonWriteableFile()
+    {
+        $file = __DIR__.'/../Fixtures/cache-file-handler/cache-file';
```
keradus: shouldn't this file be created/destroyed by test suite itself ?

tests/Cache/FileHandlerTest.php
```diff
@@ -145,6 +148,76 @@ public function testWriteWritesCache()
...
+                'foo',
+                'bar',
+            )
+        )));
+    }
+
+    public function testWriteCacheFilePermissions()
+    {
+        $file = __DIR__.'/../Fixtures/cache-file-handler/rw_cache.test';
+        @unlink($file);
```
keradus: why? this file must not exist before test execution

tests/Cache/FileHandlerTest.php
```diff
@@ -145,6 +148,76 @@ public function testWriteWritesCache()
...
+            array(
+                'foo',
+                'bar',
+            )
+        )));
+
+        $this->assertFileExists($file);
+        $this->assertFilePermissionMask(0666, $file);
+
+        @unlink($file);
```
keradus: This is not enough, that's probably why you had to try drop this file also in top of the method.
File shall be registered to be removed even if php will crash and not execute this line at all, use `FileRemoval` to do it

src/Cache/FileHandler.php
```diff
@@ -60,13 +60,36 @@ public function write(CacheInterface $cache)
...
+        }
+
         $bytesWritten = @file_put_contents($this->file, $content, LOCK_EX);
 
         if (false === $bytesWritten) {
             $error = error_get_last();
 
             throw new IOException(
-                sprintf('Failed to write file "%s", "%s".', $this->file, $error ? $error['message'] : 'no reason available'),
+                sprintf('Failed to write file "%s", "%s".', $this->file, isset($error['message']) ? $error['message'] : 'no reason available'),
```
keradus: you have added dedicated tests for 2 newly introduced `IOException` (big :+1: for that!),
is there a test for this `IOException` already?

tests/Cache/FileHandlerTest.php
keradus: with proper preconfiguration and cleanup, we don't need to mark this test as skipped

src/Cache/FileHandler.php
keradus: ok, there is.
I do believe that after solving my comments all of 3 test methods about `Symfony\Component\Filesystem\Exception\IOException` could be combined into 1 method with data provider

keradus:
looks promising ;)

tests/Cache/FileHandlerTest.php
SpacePossum: `$this->assertFilePermissionMask(0666, $file);`

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3123/files/cda7bf908fcbb0aa41e26dfe2dd5ecdedc8e61bc#diff-953496b5de42e8f87de0524f42d9cbcaR216

I can refactor to use the constraint directly to support 5.3

tests/Cache/FileHandlerTest.php
SpacePossum: The problem is that the tests runs in one PHP process.
Creating a file within that process and than make it impossible to write to using permissions is (near) impossible on *ux systems. 
The alternative would be to setup the fixture file outside the PHP process. This should than be done for each CI and will fragile and require additional action when running the test locally.

tests/Cache/FileHandlerTest.php
SpacePossum: to save people the headache when they manually stopped the test before the file was removed

tests/Cache/FileHandlerTest.php
SpacePossum: `FileRemoval` works based on `register_shutdown_function`, therefor `removed even if php will crash` is not true. When PHP exists on a fatal like out-op-memory the `register_shutdown_function` are not quarantined to be executed.

tests/Cache/FileHandlerTest.php
SpacePossum: please see other comments about why I used this setup.

tests/Cache/FileHandlerTest.php
keradus: Please do, we shall fix it on 2.2 without raising php engine

tests/Cache/FileHandlerTest.php
keradus: huh? you mean one cannot set permissions during execution? or I don't understand you

manual setup outside phpunit is not desired (except setup done like in `CiIntegrationTest`)

tests/Cache/FileHandlerTest.php
keradus: `@unlink` will also not handle `out-op-memory` cases, yet, `FileRemoval` will handle more cases than just `@unlink` ?

SpacePossum:
```
@unlink will also not handle out-op-memory cases, yet, FileRemoval will handle more cases than just @unlink ?
```

Don't think so, my setup always throws away the file if it exists before testing and tries to throw it away at the end of it.

tests/Cache/FileHandlerTest.php
SpacePossum: I'm saying it more trouble than its worth to go through all edge cases based on OS deployments and user account used. This is a fixture file like any other fixture we've in the repo, I don't see why the test or CI env. the test is executed in has to create the file.

src/Cache/FileHandler.php
SpacePossum: I don't think so, the cases need different setups to trigger the exception. Merging those will give little gain and more work to maintain.

tests/Cache/FileHandlerTest.php
keradus: change
```diff
+ throw new \Exception();
  @unlink($file);
```

is file removed? no ;)
with `FileRemoval`, it would. that's my point

SpacePossum: Using `FileRemoval` might help for an exception, it doesn't for uncontrolled stopping of the process. My point is that my version doesn't care about any of that.

keradus: but then you have weird unlinking in the beginning of test ;)

SpacePossum: From your POV ;)
I like to _setup_ my test at the _start_,
so in this test I don't want to have that file so I throw it away.

Issue #3048, opened 2017-09-12
Add a rule to root-namespace special compiled PHP functions


nicolas-grekas:
Since PHP 7.0, some functions are replaced by opcodes, producing much faster code.
Yet, for this to work, these functions need to be referenced in the root namespace at compile time:
Either there is no namespace, or they are prefixed by a `\`.
php-cs-fixer should have a rule to list those functions and enforce adding this `\` when used inside a namespaced file/class.

Here is the exact list of functions that have this behavior:
https://github.com/php/php-src/blob/f2db305fa4e9bd7d04d567822687ec714aedcdb5/Zend/zend_compile.c#L3872

bendavies:
I was just about to raise this. you beat me to it :)

bendavies:
btw, this is already covered by the `native_function_invocation` rule:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php

but as you said, it could have extra configuration to only replace the function calls you link to.

keradus:
looks like a nice scenario
please open a PR to improve configuration of the fixer @nicolas-grekas :+1:

SpacePossum:
```
"array_slice"
"assert"
"boolval"
"call_user_func"
"call_user_func_array"
"chr"
"count"
"defined"
"doubleval"
"floatval"
"func_get_args"
"func_num_args"
"get_called_class"
"get_class"
"gettype"
"in_array"
"intval"
"is_array"
"is_bool"
"is_double"
"is_float"
"is_int"
"is_integer"
"is_long"
"is_null"
"is_object"
"is_real"
"is_resource"
"is_string"
"ord"
"strlen"
"strval"
```

IIUC

nicolas-grekas:
@SpacePossum correct, with some subtelties if we want t deal with them (just discovered that eg `in_array` is optimized only if its 3rd arg is set)

keradus:
haha, next scenario when it's always better to use strict comparisons ;)

keradus:
@nicolas-grekas 

`native_function_invocation` adds `\` to each native function invocation.
this makes them to be handled faster, as some of them will be converted to opcodes, and the rest will be resolved immediately as root function instead of having a need to check if the function is also registered in current namespace.

is there anything left for this fixer ?

nicolas-grekas:
[...]

PR #3127, opened 2017-10-04
Add NativeConstantInvocationFixer

Slamdunk:
Following [NativeFunctionInvocationFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2462) here I propose:
```diff
-echo M_PI;
+echo \M_PI;
```
VLD before: https://3v4l.org/43Sci/vld
VLD after: https://3v4l.org/FXkg2/vld


src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,150 @@
...
+                continue;
+            }
+
+            $next = $tokens->getNextMeaningfulToken($index);
+            if ($tokens[$next]->equals('(')) {
+                continue;
+            }
+
+            $constantNamePrefix = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$constantNamePrefix]->isGivenKind([CT::T_USE_TRAIT, T_CLASS, T_CONST, T_DOUBLE_COLON, T_EXTENDS, T_FUNCTION, T_IMPLEMENTS, T_INTERFACE, T_NAMESPACE, T_NEW, T_NS_SEPARATOR, T_OBJECT_OPERATOR, T_TRAIT, T_USE])) {
```
Slamdunk: I don't like this whitelist very much, it's too brittle: suggestions welcome

Slamdunk:
```
There was 1 failure:

1) PhpCsFixer\Tests\Fixer\ConstantNotation\NativeConstantInvocationFixerTest::testFixWithDefaultConfiguration with data set #3 ('<?php namespace M_PI; use M_P...n; } }')
Failed asserting that 'PHP Fatal error:  Cannot declare class M_PI\M_PI because the name is already in use in /home/travis/build/FriendsOfPHP/PHP-CS-Fixer/cs_fixer_tmp_R5f8K8 on line 1

Source:
<?php namespace M_PI; use M_PI; class M_PI extends M_PI implements M_PI { const M_PI = 1; use M_PI; public $M_PI = 1; public function M_PI($M_PI) { $M_PI = M_PI(); return; } }' is null.

/home/travis/build/FriendsOfPHP/PHP-CS-Fixer/tests/Test/AbstractFixerTestCase.php:175
/home/travis/build/FriendsOfPHP/PHP-CS-Fixer/tests/Fixer/ConstantNotation/NativeConstantInvocationFixerTest.php:109
```
I'll fix this tomorrow, splitting all the stuff in separate cases.

keradus:
cc @nicolas-grekas

nicolas-grekas:
Can we set a list for the ones we'd like to always namespace?
I don't thing we will want to add a `\` to all of them in Symfony, but only to a selected few that could contribute to OPcache dead code elimination (eg `PHP_VERSION_ID`)

keradus:
so, you would say there is no need to add `\` for most of the constants in terms of performance ?

nicolas-grekas:
I think I will not advocate that the *maybe* perf benefit is worth the noise when writing code yes.

keradus:
Then maybe try to add a combined behavior described in other ticket you raised @nicolas-grekas 
- option to add `\` to preset constants where it makes sense (like that `PHP_VERSION_ID`, need to find all)
- option to change all constants to be applied only before phar building

?

Slamdunk:
> I think I will not advocate that the maybe perf benefit is worth the noise when writing code yes.

I have to say I don't understand this statement. In my humble opinion:

1. The extra `\` is **always** a pain in the *** in therm of noise when writing code: since the introduction of `NativeFunctionInvocationFixer` I have never seen a single programmer adding it manually, the automatism of this tool is here for this very purpose. Since the introduction of this PHP-CS-Fixer I have ended ordering `use` import manually, taking care of newlines, etc: the code deserves only my creativity, not my attention to this smallnesses.
1. Stated the first point, since all Symfony code (afaik) is all namespaced there will always be a performance improvement.
1. To me it will be very confusing to have half constants escaped and half not (`exclude` list apart): I am against this approach, if you don't like it don't use the fixer.

As far as I can see, there is no difference, in theory and in practice, between this fixer and `NativeFunctionInvocationFixer`, and shouldn't be.

keradus:
ref #3048
for functions: there is a difference between adding \` everywhere and for subset of functions that are boosted by compiler to opcode directly. from what I understand from @nicolas-grekas it's somehow similar here

Slamdunk:
I admit I know nothing about php source code and opcode sources.
But here are my tests:
```
$ php71 -r 'namespace A\B\C; for ($i=0;$i<100000000;++$i) {M_PI;}'

real    0m1.545s
user    0m1.541s
sys     0m0.002s

$ php71 -r 'namespace A\B\C; for ($i=0;$i<100000000;++$i) {\M_PI;}'

real    0m0.905s
user    0m0.903s
sys     0m0.000s
```
Repeated tests always show noticeable improvements around 40%, and with all constants.

Slamdunk:
Tests fo the same with opcache present, opcache.enable=1, opcache.enable_cli=1

nicolas-grekas:
Writing is one thing, reading another. Reading a code full of `\` is  no go to me.
I'm very thankful to the PHP internal team that they didn't make the `\` a requirement, PHP would become a joke to read.
The benchmark isn't representative of anything. I know no app that access PI 100M times.
A bench on a real app would be the only way to prove this could be worth the huge readability downgrade.

Slamdunk:
I understand your point, and I agree this fixer should be adopted only where it makes sense.

This library access T_* constants millions of times, I would be very happy to apply my fixers in half the time.

Slamdunk:
@nicolas-grekas continuing what written in #3222 we could use the `$categorize` parameter in [get_defined_constants](https://secure.php.net/manual/en/function.get-defined-constants.php) to get this fixer easy to configure and to escape only the certain constants:

```php
print_r(array_keys(get_defined_constants(true)));
/*
Array
(
    [0] => Core
    [1] => date
    [2] => libxml
    [3] => pcre
    [4] => sqlite3
    [5] => dom
    [6] => fileinfo
    [7] => filter
    [8] => hash
    [9] => iconv
    [10] => json
    [11] => posix
    [12] => session
    [13] => standard
    [14] => tokenizer
    [15] => xml
    [16] => openssl
    [17] => curl
    [18] => gd
    [19] => imap
    [20] => mbstring
    [21] => mysqli
    [22] => zlib
)
*/
```
By default only `Core` would be fixed, but the user could specify a different subset, or all of them.

WDYT?

nicolas-grekas:
I can't tell for everyone else, but on Symfony, we might just cherry-pick a few constants only (and not by "group"), less than 10 for sure (DIRECTORY_SEPARATOR, PHP_VERSION_ID, PHP_SAPI, maybe no more). I don't know how to help much more sorry.

Slamdunk:
Symfony is a leader project not only for its adoption, but also for the good standards it spreads and all the edge cases it faces.

I am willing to make this fixer configurable this way:

1. Escape all the constants
1. Escape a custom subset of constants
1. Escape the Symfony sybset of constants

@nicolas-grekas would you be so kind to research what would be the Symfony needs and list an "*official*" Symfony list of the constants to be escaped?

Slamdunk:
If can be of any help, at current commit
https://github.com/symfony/symfony/tree/61b753443d6903d8a9644f62caf020fdfca3ef6f
The constant with 10 or more occurrences in all symfony are:

| Constant | Count |
| --- | ---: |
| DIRECTORY_SEPARATOR | 481 |
| PHP_EOL | 78 |
| E_USER_DEPRECATED | 67 |
| DATE_RFC2822 | 39 |
| PHP_INT_MAX | 32 |
| UPLOAD_ERR_OK | 25 |
| E_DEPRECATED | 21 |
| PHP_SAPI | 20 |
| E_ERROR | 16 |
| UPLOAD_ERR_NO_FILE | 15 |
| E_USER_WARNING | 15 |
| JSON_PRETTY_PRINT | 15 |
| E_USER_NOTICE | 15 |
| STDIN | 15 |
| PHP_BINARY | 14 |
| LC_NUMERIC | 13 |
| INF | 13 |
| PATHINFO_EXTENSION | 12 |
| E_USER_ERROR | 12 |
| UPLOAD_ERR_INI_SIZE | 12 |
| E_NOTICE | 10 |
| E_RECOVERABLE_ERROR | 10 |
| E_ALL | 10 |
| PHP_VERSION | 10 |
| E_PARSE | 10 |

Excluding the `/Tests/` folders, the same count gives:

| Constant | Count |
| --- | ---: |
| DIRECTORY_SEPARATOR | 94 |
| E_USER_DEPRECATED | 22 |
| PHP_INT_MAX | 18 |
| PHP_SAPI | 14 |
| PHP_EOL | 13 |
| E_ERROR | 12 |
| PATHINFO_EXTENSION | 11 |
| STDIN | 10 |

Of course a constants may occur only one time but can be called a lot of times, but still this may help.

nicolas-grekas:
Cool, thanks for the analysis. PHP_VERSION_ID is missing, maybe because you did that on master? What about 3.4?
Of those constants, we should select only those involved in conditional branching.
That means DIRECTORY_SEPARATOR, PHP_SAPI and PHP_VERSION_ID only.

Slamdunk:
Yes the previous statistic was for master, where `PHP_VERSION_ID` appears only 7 times.
On 3.4 we have (still excluding `/Tests/`):

| Constant | Count |
| --- | ---: |
| E_USER_DEPRECATED | 352 |
| DIRECTORY_SEPARATOR | 115 |
| PHP_VERSION_ID | 62 |
| PHP_INT_MAX | 18 |
| PHP_SAPI | 17 |
| E_ERROR | 15 |
| PHP_EOL | 14 |
| PHP_INT_SIZE | 13 |
| PATHINFO_EXTENSION | 12 |
| STDIN | 10 |

nicolas-grekas:
OK, so I confirm: only `DIRECTORY_SEPARATOR`, `PHP_SAPI` and `PHP_VERSION_ID`.

Slamdunk:
I've thought about what you said "*we should select only those involved in conditional branching*" and in fact I've found various projects where *conditional branching* vary a lot from the Symfony one.

Thus I've added the `include` parameter to allow arbitrary custom whitelists.

keradus:
`@Symfony` ruleset shall be adjusted

Slamdunk:
I've added the rule in the `@Symfony:risky` set and applied it.

keradus:
that's great. please regenerate readme, as it has to update rule now belongs to ruleset

Slamdunk:
God, after years of contributing to this project, I still forget to rebuild README.rst every time :rage4: 

Done.

keradus:
Actually, me too ;P So we have automated tests for it :)

README.rst
```diff
@@ -694,6 +694,19 @@ Choose from the list of available rules:
...
+
+  Add leading ``\`` before constant invocation of internal constant to speed
+  up resolving. Constant name match is case-sensitive.
+
+  *Risky rule: risky when any of the constants are namespaced or overridden.*
+
+  Configuration options:
+
+  - ``exclude`` (``array``): list of constants to ignore; defaults to ``[]``
+  - ``include`` (``array``): list of constants to fix. Default is to fix all
```
keradus: I must admit it confuses me.
> Default is to fix all constants returned by ``get_defined_constants``; defaults to ``[]``

If I wouldn't know what we actually do in the fixer, I wouldn't know what is default.

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * @return string[]
+     */
+    private function getConstantNames()
+    {
+        $definedConstants = $this->configuration['include'];
+        if (empty($definedConstants)) {
+            $definedConstants = \get_defined_constants();
+            unset($definedConstants['NULL'], $definedConstants['FALSE'], $definedConstants['TRUE']);
```
keradus: for a phar release, I would like to `\` all native constants, including null/false/true. With which configuration I could achieve it ?

tests/Fixer/ConstantNotation/NativeConstantInvocationFixerTest.php
```diff
@@ -0,0 +1,247 @@
...
+     * @return array
+     */
+    public function provideFixWithDefaultConfigurationCases()
+    {
+        return [
+            ['<?php var_dump(NULL, FALSE, TRUE, 1);'],
+            ['<?php echo CUSTOM_DEFINED_CONSTANT_123;'],
+            ['<?php echo m_pi; // Constant are case sensitive'],
+            ['<?php namespace M_PI;'],
+            ['<?php namespace Foo; use M_PI;'],
```
keradus: please add `echo M_PI;` at the end of this case

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,169 @@
...
+            $tokens->insertAt($index, new Token([T_NS_SEPARATOR, '\\']));
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        $constantChecker = function ($value) {
```
keradus: please make it static function

README.rst
keradus: IMO, this proves it's one option to cover two scenarios.
I would split it into `fix_built_in: boolean` to decide whether they should be fixed or not, and on top of that `include: array` to add sth on top of that.
then, together with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3127/files#r150393437 , default config would be `fix_built_in: true, exclude: ['true', 'false', 'null']`

tests/Fixer/ConstantNotation/NativeConstantInvocationFixerTest.php
Slamdunk: This would result in:
```diff
-<?php namespace Foo; use M_PI; echo M_PI;
+<?php namespace Foo; use M_PI; echo \M_PI;
```
While with `<?php namespace Foo; use M_PI;` I want to check only trait case.

How would the addition be useful?

keradus: So please add such case as new case (so we would have both of them).

To have behavior of fixer specified in utest for case when global constant is imported.

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -136,15 +136,19 @@ protected function createConfigurationDefinition()
...
+            (new FixerOptionBuilder('include', 'List of additional constants to fix.'))
                 ->setAllowedTypes(['array'])
                 ->setAllowedValues([$constantChecker])
                 ->setDefault([])
                 ->getOption(),
             (new FixerOptionBuilder('exclude', 'List of constants to ignore.'))
                 ->setAllowedTypes(['array'])
                 ->setAllowedValues([$constantChecker])
-                ->setDefault([])
+                ->setDefault(['NULL', 'FALSE', 'TRUE'])
```
keradus: everywhere else, we use lowercased version

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -136,15 +136,19 @@ protected function createConfigurationDefinition()
         };
 
         return new FixerConfigurationResolver([
-            (new FixerOptionBuilder('include', 'List of constants to fix. Default is to fix all constants returned by `get_defined_constants`.'))
+            (new FixerOptionBuilder('fix_built_in', 'Whether to fix constants returned by `get_defined_constants`.'))
+                ->setAllowedTypes(['boolean'])
```
keradus: s/boolean/bool/

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,171 @@
...
+     */
+    private function getConstantNames()
+    {
+        $definedConstants = $this->configuration['include'];
+        if (true === $this->configuration['fix_built_in']) {
+            $definedConstants = array_merge($definedConstants, \array_keys(\get_defined_constants()));
+        }
+
+        return \array_diff(
+            $definedConstants,
```
keradus: as we call `array_unique` on `exclude`, we could call `array_unique` on `$definedConstants` as well (as I could include sth twice (same as I exclude sth twice and you handle it), and even include sth that is included by `fix_build_in`)

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,171 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $constantNames = $this->getConstantNames();
```
keradus: let us override `configure` and store `constantNames` in property. that way, it will be calculated once instead of per file

keradus: actually, i would convert it even to dict `['M_PI' => true, ...]`, so lookup if given const is supposed to be slashed would be `O(1)` (`isset($constantNames[$foo])`) instead of `O(n)` (`in_array($foo, $constantNames, true)`)


https://3v4l.org/qjKCq

WDYT?

Slamdunk: I know the isset/in_array thing, I started it this way because of casing.

So let me ask you one thing, linked to NULL/TRUE/FALSE comment: how should I handle casing?
Because `M_PI !== m_pi`: https://3v4l.org/cmQCG

keradus: both `isset` and `in_array` are case sensitive.
I don't see that you do casing manipulation.

what is the issue ?

Slamdunk:
I mean, `get_defined_constants` returns NULL, so excluding null lowercase will result in fixing null but not NULL. The behaviour is instead different for othet constants. If I take care of casing, I would have to exclude NULL, null, Null etc.

I would like to do a case-insensitive match (describing it in the readme) and taunt developers who use the same name on two constant, with different cases.

keradus:
I must admit I would keep it case sensitive, with special handling of true/false/null insensitiveness, so same as PHP treat the constants.

tests/Fixer/ConstantNotation/NativeConstantInvocationFixerTest.php
Slamdunk: Done

Slamdunk:
Oh my god, after push the latest changes I noticed this in the [define](https://secure.php.net/manual/en/function.define.php) man page:
> bool define ( string $name , mixed $value [, bool $case_insensitive = false ] )
> > case_insensitive
>      If set to TRUE, the constant will be defined case-insensitive. The default behavior is case-sensitive; i.e. CONSTANT and Constant represent different values.

The recent commits consider only NULL, FALSE and TRUE to be case-insensitive.

I'm afraid to ask it: should we give the ability to the user to choose casing for each custom constant?

Tecnically it would be easy, but honestly I would consider the following code a bad practice:
```php
define('FOO', 'bar ', true);
echo FOO . Foo . fOO . FoO;
// bar bar bar bar
```
And bad practices should be discouraged.

keradus:
> should we give the ability to the user to choose casing for each custom constant?

I would stick to case-sensitiveness for all constants except true/false/null.
For built-in constants, let us follow behaviour how they are declared, for custom `include` list, it's up to end-user what he will put here.

staabm:
whats open to get this feature landed?

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,211 @@
...
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $uniqueConfiguredExclude = \array_unique($this->configuration['exclude']);
+
+        // Case sensitive constants handling
+        $constantsToEscape = array_values($this->configuration['include']);
+        if (true === $this->configuration['fix_built_in']) {
+            $constantsToEscape = array_merge($constantsToEscape, \array_keys(\get_defined_constants()));
```
stof: `get_defined_constants` also includes userland constants (which would be constants define by the run of the PHP-CS-Fixer in such case). This could cause issues (and contradicts the fixer name saying `native`)
A solution would be to use the categorized output of `get_defined_constants`, to remove constants belonging to the `user` category.

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,211 @@
...
+            }
+
+            $constantNamePrefix = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$constantNamePrefix]->isGivenKind([CT::T_USE_TRAIT, T_CLASS, T_CONST, T_DOUBLE_COLON, T_EXTENDS, T_FUNCTION, T_IMPLEMENTS, T_INTERFACE, T_NAMESPACE, T_NEW, T_NS_SEPARATOR, T_OBJECT_OPERATOR, T_TRAIT, T_USE])) {
+                continue;
+            }
+
+            if (!isset($this->constantsToEscape[$tokenContent]) && !isset($this->caseInsensitiveConstantsToEscape[strtolower($tokenContent)])) {
+                continue;
+            }
```
stof: This fixer should also identify the constants imported with `use const` to avoid prefixing them. It would make the fixer much less risky (and it is necessary to support use statements if you want to later make this fixer configurable to allow injecting such use statements instead of injecting a `\`).

Slamdunk:
@stof you are right on both your statements. By the way this feature seems controversial, and it's almost 6 months is stalled by the maintainers; I'm not going to develop any further improvement without a clear direction by them.

If you are willing to fix them, I'll be happy to review & merge them as a PR upon my branch in my fork :+1:

keradus:
sadly, whole `\` thing got spread across multiple issues and PRs, each of them with conflicting ideas. I believe we got some conclusion in the end and @SpacePossum did implement part of them.
I believe this is now good time to revive the PR.

Can you, please:
- merge/rebase the PR with master
- point out what directions are still missing, so we could make a call on them

Slamdunk:
> point out what directions are still missing, so we could make a call on them

1. Strip user defined constants returned by `get_defined_constants`: easy pick, I'll face this soon
1. Do not fix constants imported with `use` statement: is there an helper class to retrieve imports?

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
Slamdunk: `user` category now skipped

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
Slamdunk: Done.

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $useDeclarations = (new NamespaceUsesAnalyzer())->getDeclarationsFromTokens($tokens);
+        $useConstantDeclarations = [];
+        foreach ($useDeclarations as $use) {
+            if ($use->isConstant()) {
```
Slamdunk: @keradus in order to have this clean API, I extended the `NamespaceUsesAnalyzer` behaviour, which is backward compatible

Slamdunk:
> 1. Strip user defined constants returned by get_defined_constants: easy pick, I'll face this soon
> 1. Do not fix constants imported with use statement: is there an helper class to retrieve imports?

Done both.

tests/Tokenizer/Analyzer/NamespaceUsesAnalyzerTest.php
```diff
@@ -93,24 +98,51 @@ public function provideNamespaceUsesCases()
...
+                    NamespaceUseAnalysis::TYPE_CONSTANT
+                ),
+            ], [1]],
 
             // TODO: How to support these:
 
-            // Function and constant imports
-            // ['<?php use function My\count;', [], []],
-            // ['<?php use function My\count as myCount;', [], []],
-            // ['<?php use const My\Full\CONSTANT;', [], []],
```
keradus: nice side effect of this PR :+1:

keradus:
Thank you @Slamdunk.

PR #3133, opened 2017-10-06
Unify Reporter tests

keradus:


tests/Report/AbstractReporterTestCase.php
```diff
@@ -0,0 +1,208 @@
...
+    {
+        $actualReport = $this->reporter->generate($reportSummary);
+
+        $this->assertFormat($expectedReport, $actualReport);
+    }
+
+    /**
+     * @return array
+     */
+    final public function provideGenerateCases()
```
keradus: as we want to test exactly same data against different report types

tests/Report/JunitReporterTest.php
```diff
@@ -42,47 +38,31 @@
 
     protected function setUp()
     {
-        $this->reporter = new JunitReporter();
+        parent::setUp();
+
         $this->xsd = file_get_contents(__DIR__.'/../../doc/junit-10.xsd');
```
julienfalque: This is used in `assertFormat()` only. I would replace the property (and the need to override `setUp()`) with a simple variable in `assertFormat()`.

tests/Report/AbstractReporterTestCase.php
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * @return ReporterInterface
+     */
+    abstract protected function createReporter();
+
+    /**
+     * @return string
+     */
+    abstract protected function createExpectedFormat();
```
julienfalque: The name is misleading IMO, why not simply `getFormat()`?

tests/Report/AbstractReporterTestCase.php
```diff
@@ -0,0 +1,203 @@
...
+abstract class AbstractReporterTestCase extends TestCase
+{
+    /**
+     * @var ReporterInterface
+     */
+    protected $reporter;
+
+    protected function setUp()
+    {
+        $this->reporter = $this->createReporter();
```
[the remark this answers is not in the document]
keradus: what for? we don't do it in any of tests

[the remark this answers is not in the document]
julienfalque: To reduce memory usage. Even if here it should not be a big amount of memory used, I think it's a good practice to avoid using properties in tests as much as possible, and clean them otherwise.

tests/Report/AbstractReporterTestCase.php
keradus: done

tests/Report/AbstractReporterTestCase.php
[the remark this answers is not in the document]
keradus: just tried it with
```php
    protected function tearDown()
    {
        unset($this->reporter);
    }
```

no change in memory usage

also, if creating data not unset in `tearDown` affects memory, we shall clean it up globally (not by creating a lot of `setUp` concrete methods manually).
Let us discuss it separatelly

tests/Report/JunitReporterTest.php
keradus: actually, this data is constant across tests, no need to read it from IO each time per case

keradus: moved to static

tests/Report/AbstractReporterTestCase.php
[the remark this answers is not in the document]
julienfalque: Ok then. Maybe using properties this way doesn't always affect memory usage much.

PR #3137, opened 2017-10-08
PHPUnit - use common base class

keradus:


tests/TestCase.php
```diff
@@ -0,0 +1,24 @@
...
+use PHPUnit\Framework\TestCase as BaseTestCase;
+
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ */
+abstract class TestCase extends BaseTestCase
+{
+}
```
SpacePossum: what is the added value over not using `PHPUnit\Framework\TestCase` directly?

Slamdunk: I'm curious too.
As a Contributor if I see `extends \PHPUnit\Framework\TestCase` I'm already in a confortable zone where I know what is going on; if I notice a project-specific base test case I must open the abstract class to inspect eventual setUp/tearDown overrides etc: it's and added step, useless for now.
But maybe you have something in mind for the future.

keradus: to do this https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3285

PR #3141, opened 2017-10-09
Unify description of deprecated fixer

kubawerlos:
Right now 2 of 3 deprecated fixers have capitalised word "DEPRECATED" as new sentence - let's make third one looks the same.

### DIFF
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -168,6 +168,29 @@ final class FixerTest extends TestCase
         $this->assertInstanceOf(\PhpCsFixer\Fixer\DefinedFixerInterface::class, $fixer);
     }
 
+    /**
+     * @dataProvider provideFixerDefinitionsCases
+     */
+    public function testDeprecatedFixersHaveCorrectSummary(FixerInterface $fixer)
+    {
+        $reflection = new \ReflectionClass($fixer);
+        $comment = $reflection->getDocComment();
+
+        if (is_string($comment) && false !== strpos($comment, '@deprecated')) {
+            $this->assertRegExp(
+                '/\. DEPRECATED: use `[a-z_]+` instead\.$/',
+                $fixer->getDefinition()->getSummary(),
+                'Deprecated fixer must contain correct "DEPRECATED" note in summary'
+            );
+        } else {
+            $this->assertNotRegExp(
+                '/DEPRECATED/',
+                $fixer->getDefinition()->getSummary(),
+                'Non-deprecated fixer cannot contain word "DEPRECATED" in summary'
+            );
+        }
+    }
+
     public function provideFixerDefinitionsCases()
     {
         return array_map(function (FixerInterface $fixer) {
--- a/README.rst
+++ b/README.rst
@@ -255,8 +255,8 @@ Choose from the list of available rules:
 
 * **blank_line_before_return**
 
-  An empty line feed should precede a return statement (deprecated, use
-  ``blank_line_before_statement`` instead).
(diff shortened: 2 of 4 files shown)


julienfalque:
👍 

What about doing the same for fixers options?

kubawerlos:
Can't see why not - what about format: always uppercase `DEPRECATED`, then colon and description what to replace with (starting with capital letter or not?)?

julienfalque:
Not sure I would use a capital letter to start the description but 👍

keradus:
@kubawerlos , unification is good for sure. Yet, how you plan to ensure that fixers deprecated later gonna follow this unification ?

kubawerlos:
@keradus obliviously with a test

@julienfalque let's unify descriptions of options in separate PR after unifying summaries.

Any thoughts about RegExes? Should I add dot and space at the beginning of first one?

tests/AutoReview/FixerTest.php
```diff
@@ -168,6 +168,28 @@ public function testFixersAreDefined(FixerInterface $fixer)
...
+    /**
+     * @dataProvider provideFixerDefinitionsCases
+     */
+    public function testDeprecatedFixersHaveCorrectSummary(FixerInterface $fixer)
+    {
+        $reflection = new \ReflectionClass($fixer);
+
+        if (false !== strpos($reflection->getDocComment(), '@deprecated')) {
+            $this->assertRegExp(
+                '/DEPRECATED: [a-z]/',
```
keradus:
```
/Deprecated: use `[a-z_]+` instead.$/
```

kubawerlos: done, what about 2nd RegEx? I was thinking about making it case-insensitive, but summary of `ereg_to_preg` contains "deprecated" word and it should be there.

keradus: case-insensitive variant would be less strict, thus less unified. and you wanted to unified it ;)

keradus: 2nd re is fine imo

SpacePossum:
Nice idea!
Maybe it is time we start thinking about how we can deprecate fixers in a way users of the `phar` releases also get notified when using a deprecated fixer and/or option?

julienfalque:
Yep, maybe instead of testing that the deprecation description is correct, we could add it the fixers definitions to that the README can be updated accordingly and so we can trigger warnings when using them. WDYT?

SpacePossum:
sounds good to me, I think the biggest task would be to create a way fixers can report warnings and have those displayed to the user in a nice matter.

keradus:
> sounds good to me, I think the biggest task would be to create a way fixers can report warnings and have those displayed to the user in a nice matter.

let us not include this in this PR ;)

kubawerlos:
@SpacePossum do you mean deprecating them by triggering error in constructor like in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3146?

@julienfalque what about creating interface `DeprecatedFixerInterface` (same as proposed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3146) with method `getSuccessor` to return the name of fixer to use?

SpacePossum:
@kubawerlos don't mind me, I was thinking bigger than is useful for this PR, I think this PR is on a good track :+1:

kubawerlos:
@SpacePossum can you create an issue and describe with details what you aim to achieve?

@keradus @julienfalque what's to do in this PR? I matter of unifying names it's finished.

keradus:
Thank you @kubawerlos.

PR #3142, opened 2017-10-09
PreIncrement fixer summary fix

kubawerlos:
Follow up to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3141


keradus:
Hi.
Thanks, yet it's not needed. Let us finish #3141 first, then this change will be applied during merging 2.2 -> 2.7 -> master. (If not, it would made branch failing)

PR #3144, opened 2017-10-10
PhpUnitDedicateAssertFixer - Sort map and array by function name

localheinz:
This PR

* [x] sorts a map of functions and function names by function name


src/Fixer/PhpUnit/PhpUnitDedicateAssertFixer.php
```diff
@@ -137,18 +137,18 @@ protected function createConfigurationDefinition()
...
             'is_boolean',
             'is_callable',
             'is_double',
             'is_float',
+            'is_infinite',
             'is_int',
             'is_integer',
             'is_long',
+            'is_nan',
+            'is_null',
```
keradus: what about calling sort on it, to actually ensure values in readme are sorted, even if this list would not be ?

julienfalque: Can't we replace this hardcoded array with `array_keys(self::$fixMap)`?

keradus: no, next time someone will update possible options, he will not-willingly update default value as well

SpacePossum:
@keradus can you kick circleci again, I don't have the permission to do so :|
@localheinz can you rebase this PR, sorry to bug you about it?

localheinz:
@SpacePossum 

Rebased!

SpacePossum:
thanks @localheinz :)

keradus:
Thank you @localheinz.

localheinz:
Thank you, @julienfalque, @keradus and @SpacePossum!

PR #3145, opened 2017-10-10
documentation - fix typos

localheinz:
This PR

* [x] fixes a small typo / wording issue in a docblock

### DIFF
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -430,9 +430,9 @@ class Token
     }
 
     /**
-     * Check if token is a whitespace.
+     * Check if token is whitespace.
      *
-     * @param null|string $whitespaces whitespaces characters, default is " \t\n\r\0\x0B"
+     * @param null|string $whitespaces whitespace characters, default is " \t\n\r\0\x0B"
      *
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


localheinz:
@SpacePossum 

What do you think if I just push a few more small fixes into this branch before merging? Reviewing some code, if I stumble across something, it's not really worth opening up another PR, I think.

src/Tokenizer/Token.php
```diff
@@ -454,9 +454,9 @@ public function isMagicConstant()
     }
 
     /**
-     * Check if token is a whitespace.
+     * Check if token is whitespace.
      *
-     * @param null|string $whitespaces whitespaces characters, default is " \t\n\r\0\x0B"
+     * @param null|string $whitespaces whitespace characters, default is " \t\n\r\0\x0B"
```
keradus: not 2.2 ?

keradus:
long time I did not see misc/grooming PR here ;)
Usually, back in all days, I was collecting those kind of trivials (not even minors ;P) and raised PR once per 10-15 days.
But it was old days, maybe we don't have that much minors currently to fix ;)

src/Tokenizer/Token.php
localheinz: Adjusted!

keradus:
Thank you @localheinz.

PR #3146, opened 2017-10-10
[RFC] Add deprecation error for deprecated fixers

kubawerlos:
Follow up to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3065. There's only "small" problem:

```
Remaining deprecation notices (7049)

Fixer "hash_to_slash_comment" is deprecated and will be removed in 3.0. Use "single_line_comment_style" instead: 7049x
(...)
```

The source for all of them is `FixerFactory::registerBuiltInFixers` which is used in tests and I haven't figured out how to solve it nicely. I have tried to add parameter `$triggerError` set by default to `true` and in tests to `false`, but that doesn't seem right. Also I have tried to add dummy interface `DeprecatedFixerInterface` and in 2nd `foreach` add `if` for instance of it - better, but still a little bit hacky.

Any suggestions? If not I'd say we could go with the interface solution (with added checks in auto review tests). 

Other fixers will be done after the solution is agreed.

### DIFF
--- a/src/Fixer/Comment/HashToSlashCommentFixer.php
+++ b/src/Fixer/Comment/HashToSlashCommentFixer.php
@@ -26,6 +26,17 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
  */
 final class HashToSlashCommentFixer extends AbstractProxyFixer implements DeprecatedFixerInterface
 {
+    public function __construct()
+    {
+        parent::__construct();
+
+        @trigger_error(sprintf(
+            'Fixer "%s" is deprecated and will be removed in 3.0. Use "%s" instead.',
+            $this->getName(),
+            current($this->proxyFixers)->getName()
+        ), E_USER_DEPRECATED);
+    }
+
     /**
      * {@inheritdoc}
      */
--- a/src/FixerFactory.php
+++ b/src/FixerFactory.php
@@ -79,6 +79,8 @@ final class FixerFactory
      */
     public function getFixers()
     {
+        $this->fixers = array_map([$this, 'retrieveFixer'], $this->fixers);
+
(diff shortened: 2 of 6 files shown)


keradus:
please rebase

kubawerlos:
Rebased, but still no proper solution - I have added `QuestionableHelper` (it's names says all) to silence deprecation error which is triggered in many, many tests (haven't add it to all as got tired of it).

The problem still stays the same - if we decide to trigger deprecation error in fixer's constructor how to nicely handle it across all tests?

tests/AutoReview/ProjectFixerConfigurationTest.php
```diff
@@ -27,15 +28,19 @@
 {
     public function testCreate()
     {
-        /** @var \PhpCsFixer\Config $config */
+        // @var \PhpCsFixer\Config $config
```
keradus: why that change?
are you aware that `//` is regular comment and any annotation inside it is actually not an annotation ?

keradus:
please don't squash :(
no, `questionableHelper` needs to be reverted (good lack with squashed commits ;p)

tests/AutoReview/ProjectFixerConfigurationTest.php
kubawerlos: no idea why this happened, definitely not on purpose

kubawerlos:
I wasn't squashing, but starting all over from master (which I repeated today) as it was different approach and only `HashToSlashCommentFixer` were staying.

kubawerlos:
White flag for now.

keradus:
yeah, this is not easy issue for sure. thanks for the work anyway!

PR #3149, opened 2017-10-11
AbstractProxyFixer - Support multiple proxied fixers

julienfalque:
Extracted from #2619.

### DIFF
--- a/src/AbstractProxyFixer.php
+++ b/src/AbstractProxyFixer.php
@@ -13,6 +13,7 @@
 namespace PhpCsFixer;
 
 use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\Tokenizer\Tokens;
 
 /**
@@ -23,15 +24,17 @@ use PhpCsFixer\Tokenizer\Tokens;
 abstract class AbstractProxyFixer extends AbstractFixer
 {
     /**
-     * @var FixerInterface
+     * @var FixerInterface[]
      */
-    protected $proxyFixer;
+    protected $proxyFixers;
 
     public function __construct()
     {
-        parent::__construct();
+        foreach (Utils::sortFixers($this->createProxyFixers()) as $proxyFixer) {
+            $this->proxyFixers[$proxyFixer->getName()] = $proxyFixer;
+        }
 
-        $this->proxyFixer = $this->createProxyFixer();
+        parent::__construct();
     }
 
     /**
@@ -39,7 +42,13 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $this->proxyFixer->isCandidate($tokens);
+        foreach ($this->proxyFixers as $fixer) {
+            if ($fixer->isCandidate($tokens)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
@@ -47,7 +56,13 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     public function isRisky()
     {
-        return $this->proxyFixer->isRisky();
+        foreach ($this->proxyFixers as $fixer) {
+            if ($fixer->isRisky()) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
@@ -55,7 +70,11 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     public function getPriority()
     {
-        return $this->proxyFixer->getPriority();
+        if (count($this->proxyFixers) > 1) {
+            throw new \LogicException('You need to override this method to provide the priority of combined fixers.');
+        }
+
+        return reset($this->proxyFixers)->getPriority();
     }
 
     /**
@@ -63,7 +82,13 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     public function supports(\SplFileInfo $file)
     {
-        return $this->proxyFixer->supports($file);
+        foreach ($this->proxyFixers as $fixer) {
+            if ($fixer->supports($file)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
@@ -72,7 +97,12 @@ abstract class AbstractProxyFixer extends AbstractFixer
     public function setWhitespacesConfig(WhitespacesFixerConfig $config)
     {
         parent::setWhitespacesConfig($config);
-        $this->proxyFixer->setWhitespacesConfig($config);
+
+        foreach ($this->proxyFixers as $fixer) {
+            if ($fixer instanceof WhitespacesAwareFixerInterface) {
+                $fixer->setWhitespacesConfig($config);
+            }
+        }
     }
 
     /**
@@ -80,11 +110,13 @@ abstract class AbstractProxyFixer extends AbstractFixer
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        $this->proxyFixer->fix($file, $tokens);
+        foreach ($this->proxyFixers as $fixer) {
+            $fixer->fix($file, $tokens);
+        }
     }
 
     /**
-     * @return FixerInterface
+     * @return FixerInterface[]
      */
-    abstract protected function createProxyFixer();
+    abstract protected function createProxyFixers();
 }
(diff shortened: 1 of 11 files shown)


src/AbstractProxyFixer.php
```diff
@@ -23,47 +24,71 @@
...
+     * @var FixerInterface[]
      */
-    protected $proxyFixer;
+    protected $proxyFixers;
 
     public function __construct()
     {
-        parent::__construct();
+        foreach (Utils::sortFixers($this->createProxyFixers()) as $proxyFixer) {
+            $this->proxyFixers[$proxyFixer->getName()] = $proxyFixer;
```
keradus: for what I see, we don't need assoc array here

src/AbstractProxyFixer.php
```diff
@@ -23,47 +24,71 @@
...
      * {@inheritdoc}
      */
     public function getPriority()
     {
-        return $this->proxyFixer->getPriority();
+        if (count($this->proxyFixers) > 1) {
+            throw new \LogicException('You need to override this method to provide the priority of combined fixers.');
+        }
+
+        return reset($this->proxyFixers)->getPriority();
```
keradus: why `reset($foo)` and not just `$foo[0]` ?

src/AbstractProxyFixer.php
```diff
@@ -23,47 +24,71 @@
...
 
     /**
      * {@inheritdoc}
      */
     public function supports(\SplFileInfo $file)
     {
-        return $this->proxyFixer->supports($file);
+        foreach ($this->proxyFixers as $fixer) {
+            if ($fixer->supports($file)) {
+                return true;
```
keradus: all proxied fixers need to support file, here, you implement "at least one fixer" rule

keradus: no, let us do it other way. i got your idea here

keradus:
2.7 is bugfix only and will not be maintained after 2.8 will be released. I'm changing the target branch

julienfalque:
Rebased onto master.

src/AbstractProxyFixer.php
julienfalque: Proxied fixers are sorted by priority, using names rather than some not so pedictable integer indexes is more convenient for main fixer to access them. It's used [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3149/files#diff-194fee79f2296fcded24ff97b2bb231fR48) for example.

src/AbstractProxyFixer.php
julienfalque: Same reason as above.

src/AbstractProxyFixer.php
keradus: then it shall be typehinted as `array<string, FixerInterface>`

SpacePossum:
please fix the project itself, all CI's are red :(  thanks! :)

src/AbstractProxyFixer.php
```diff
@@ -72,19 +97,26 @@ public function supports(\SplFileInfo $file)
...
     }
 
     /**
      * {@inheritdoc}
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        $this->proxyFixer->fix($file, $tokens);
+        foreach ($this->proxyFixers as $fixer) {
+            $fixer->fix($file, $tokens);
```
SpacePossum: shouldn't the `supports(\SplFileInfo $file)` be done here for each fixer, as the previous check has an early return with `true`?

src/AbstractProxyFixer.php
julienfalque: Done.

src/AbstractProxyFixer.php
julienfalque: I'm not sure I understand your comment. Each proxied fixer is going to do this check internally when calling `fix()`, isn't?

src/AbstractProxyFixer.php
```diff
@@ -74,7 +74,7 @@ public function getPriority()
             throw new \LogicException('You need to override this method to provide the priority of combined fixers.');
         }
 
-        return reset($this->proxyFixers)->getPriority();
+        return current($this->proxyFixers)->getPriority();
```
julienfalque: Oops, I didn't expect this change to break something.

src/AbstractProxyFixer.php
keradus: I dropped on it as well @SpacePossum (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3149#discussion_r144108084)

`fix` is calling `tokens->count`, `supports` and `isCandidate`

src/AbstractProxyFixer.php
```diff
@@ -23,47 +24,71 @@
...
 
     public function __construct()
     {
-        parent::__construct();
+        foreach (Utils::sortFixers($this->createProxyFixers()) as $proxyFixer) {
+            $this->proxyFixers[$proxyFixer->getName()] = $proxyFixer;
+        }
 
-        $this->proxyFixer = $this->createProxyFixer();
+        parent::__construct();
```
keradus: btw, why constructor was moved?

src/Fixer/Operator/PreIncrementFixer.php
```diff
@@ -29,19 +29,19 @@
     public function getDefinition()
     {
         return new FixerDefinition(
-            sprintf('Pre incrementation/decrementation should be used if possible. DEPRECATED: Use "%s" instead.', $this->proxyFixer->getName()),
+            'Pre incrementation/decrementation should be used if possible. DEPRECATED: Use "increment_style" instead.',
```
keradus: why hardcoding the name ?

src/FixerFactory.php
```diff
@@ -79,9 +79,7 @@ public function setWhitespacesConfig(WhitespacesFixerConfig $config)
      */
     public function getFixers()
     {
-        $this->sortFixers();
-
-        return $this->fixers;
+        return Utils::sortFixers($this->fixers);
```
keradus: originally, sorted fixers were assigned to `this->fixers`, so if one call method again, sorting will be ultra fast (as already sorted)

src/Fixer/Operator/PreIncrementFixer.php
julienfalque: Because I would have to hardcode it to retrieve the proxied fixer anyway, i.e. `$this->proxyFixers['increment_style']->getName()`.

src/AbstractProxyFixer.php
julienfalque: Because parent constructor initializes the configuration with default values by calling `configure([])`. If the `configure` methods needs to access a proxied fixer, the proxied fixers must be available. See [#2619](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2619/files#diff-60306f9e47866c05bdcbedf9aa5dc03dR77) for an example.

src/Fixer/Operator/PreIncrementFixer.php
keradus: or `current($this->proxyFixers)->getName()`, how you just did in `HashToSlashCommentFixer`

julienfalque: Done.

src/FixerFactory.php
julienfalque: The (now removed) `sortFixers()` method didn't seem to have a mechanism to check whether fixers are already sorted so I'm not sure this is true and I think it should be done in a separate PR.

src/AbstractProxyFixer.php
SpacePossum: ah right, so the abstract fixer could you return `true` for `supports` and `isCandidate` without asking the proxied fixers?

keradus: in theory, yes. it would work fine.
yet, someone external could depends on `supports`/`isCandidate` as well, so we need to provide good data. this case was discussed and it's problematic to have them to be called by outsiders before calling `->fix`, so we call them in `->fix` as well, and they will be dropped in v3 hopefully

src/FixerFactory.php
keradus: I'm not talking about any extra mechanism on our side.
PHP native `sort` is using quicksort alghoritm.
Complexity of sort operation depends on order of data, if they are in reverse order, complexity is `n*n`, if they are already in proper order, complexity is `n * log(n)`. 

so, nothing to do here on our side, except remembering the sorted order
```diff
-return Utils::sortFixers($this->fixers);
+$this->fixers = Utils::sortFixers($this->fixers);
+return $this->fixers;
```

julienfalque: Done.

keradus:
Thank you @julienfalque.

PR #3155, opened 2017-10-14
Add EscapeImplicitBackslashesFixer

Slamdunk:
```diff
-$a = 'My\Prefix\\';
+$a = 'My\\Prefix\\';
```

- [x] Fix [single quoted](https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.single)
- [x] Fix [double quoted](https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.double)
- [x] Fix [heredoc syntax](https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc)
- [x] Run before `SingleQuoteFixer`
- [x] Run before `HeredocToNowdocFixer`

No need to fix [nowdoc syntax](https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -545,8 +545,9 @@ Choose from the list of available rules:
 
 * **escape_implicit_backslashes**
 
-  Escape implicit backslashes in strings and heredocs to increase code
-  readability.
+  Escape implicit backslashes in strings and heredocs to ease the
+  understanding of which are special chars interpreted by PHP and which
+  not.
 
   Configuration options:
 
--- a/README.rst
+++ b/README.rst
@@ -538,11 +538,6 @@ Choose from the list of available rules:
 
   *Risky rule: risky if the ``ereg`` function is overridden.*
 
-* **explicit_indirect_variable**
-
-  Add curly braces to indirect variables to make them clear to understand.
-  Requires PHP >= 7.0.
-
 * **escape_implicit_backslashes**
 
   Escape implicit backslashes in strings and heredocs to ease the
@@ -556,6 +551,11 @@ Choose from the list of available rules:
   - ``heredoc_syntax`` (``bool``): whether to fix heredoc syntax; defaults to ``true``
   - ``single_quoted`` (``bool``): whether to fix single-quoted strings; defaults to
     ``false``
+
+* **explicit_indirect_variable**
+
+  Add curly braces to indirect variables to make them clear to understand.
+  Requires PHP >= 7.0.
 
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -543,9 +543,10 @@ Choose from the list of available rules:
   Add curly braces to indirect variables to make them clear to understand.
   Requires PHP >= 7.0.
 
-* **explicit_escape**
+* **escape_implicit_backslashes**
 
-  Escape implicit backslashes.
+  Escape implicit backslashes in strings and heredocs to increase code
+  readability.
 
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
--- a/README.rst
+++ b/README.rst
@@ -543,6 +543,10 @@ Choose from the list of available rules:
   Add curly braces to indirect variables to make them clear to understand.
   Requires PHP >= 7.0.
 
+* **explicit_escape**
+
+  Escape implicit backslashes.
+
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
   PHP code must use the long ``<?php`` tags or short-echo ``<?=`` tags and not
--- a/README.rst
+++ b/README.rst
@@ -548,6 +548,14 @@ Choose from the list of available rules:
   Escape implicit backslashes in strings and heredocs to increase code
   readability.
 
+  Configuration options:
+
+  - ``double_quoted`` (``bool``): whether to fix double-quoted strings; defaults to
+    ``true``
+  - ``heredoc_syntax`` (``bool``): whether to fix heredoc syntax; defaults to ``true``
+  - ``single_quoted`` (``bool``): whether to fix single-quoted strings; defaults to
+    ``false``
+
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
   PHP code must use the long ``<?php`` tags or short-echo ``<?=`` tags and not
--- a/src/Console/Command/ReadmeCommand.php
+++ b/src/Console/Command/ReadmeCommand.php
@@ -236,15 +236,15 @@ EOF;
         $help = preg_replace("#^\n( +\\$ )#m", "\n.. code-block:: bash\n\n$1", $help);
         $help = preg_replace("#^\n( +<\\?php)#m", "\n.. code-block:: php\n\n$1", $help);
         $help = preg_replace_callback(
-            "#^\s*<\?(\w+).*?\?>#ms",
+            '#^\\s*<\\?(\\w+).*?\\?>#ms',
             function ($matches) {
-                $result = preg_replace("#^\.\. code-block:: bash\n\n#m", '', $matches[0]);
+                $result = preg_replace("#^\\.\\. code-block:: bash\n\n#m", '', $matches[0]);
 
                 if ('php' !== $matches[1]) {
-                    $result = preg_replace("#<\?{$matches[1]}\s*#", '', $result);
+                    $result = preg_replace("#<\\?{$matches[1]}\\s*#", '', $result);
                 }
 
-                $result = preg_replace("#\n\n +\?>#", '', $result);
+                $result = preg_replace("#\n\n +\\?>#", '', $result);
 
                 return $result;
             },
--- a/src/Console/Command/ReadmeCommand.php
+++ b/src/Console/Command/ReadmeCommand.php
@@ -236,7 +236,7 @@ EOF;
         $help = preg_replace("#^\n( +\\$ )#m", "\n.. code-block:: bash\n\n$1", $help);
         $help = preg_replace("#^\n( +<\\?php)#m", "\n.. code-block:: php\n\n$1", $help);
         $help = preg_replace_callback(
-            '#^\\s*<\\?(\\w+).*?\\?>#ms',
+            '#^\s*<\?(\w+).*?\?>#ms',
             function ($matches) {
                 $result = preg_replace("#^\\.\\. code-block:: bash\n\n#m", '', $matches[0]);
 
--- a/src/Fixer/ControlStructure/NoBreakCommentFixer.php
+++ b/src/Fixer/ControlStructure/NoBreakCommentFixer.php
@@ -183,7 +183,7 @@ switch ($foo) {
 
         $text = preg_quote($this->configuration['comment_text'], '~');
 
-        return preg_match("~^((//|#)\s*$text\s*)|(/\*\*?\s*$text\s*\*/)$~i", $token->getContent());
+        return preg_match("~^((//|#)\\s*$text\\s*)|(/\\*\\*?\\s*$text\\s*\\*/)$~i", $token->getContent());
     }
 
     /**
--- a/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
+++ b/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
@@ -21,7 +21,7 @@ use PhpCsFixer\Tokenizer\Tokens;
 /**
  * @author Filippo Tessarotto <zoeslam@gmail.com>
  */
-final class ExplicitEscapeFixer extends AbstractFixer
+final class EscapeImplicitBackslashesFixer extends AbstractFixer
 {
     /**
      * {@inheritdoc}
--- a/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
+++ b/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
@@ -93,7 +93,7 @@ EOF;
         $doubleQuoteOpened = false;
         foreach ($tokens as $index => $token) {
             $content = $token->getContent();
-            if (null === $token->getId() && '"' === $content) {
+            if ($token->equals('"')) {
                 $doubleQuoteOpened = !$doubleQuoteOpened;
             }
             if (!$token->isGivenKind([T_ENCAPSED_AND_WHITESPACE, T_CONSTANT_ENCAPSED_STRING]) || false === strpos($content, '\\')) {
--- a/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
+++ b/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
@@ -112,7 +112,7 @@ EOF;
             );
             $isHeredocSyntax = !$isSingleQuotedString && !$isDoubleQuotedString;
             if (
-               (false === $this->configuration['single_quoted'] && $isSingleQuotedString)
+                (false === $this->configuration['single_quoted'] && $isSingleQuotedString)
                 || (false === $this->configuration['double_quoted'] && $isDoubleQuotedString)
                 || (false === $this->configuration['heredoc_syntax'] && $isHeredocSyntax)
             ) {
--- a/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
+++ b/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
@@ -106,10 +106,10 @@ EOF;
             }
 
             $isSingleQuotedString = $token->isGivenKind(T_CONSTANT_ENCAPSED_STRING) && '\'' === $content[0];
-            $isDoubleQuotedString = (
+            $isDoubleQuotedString =
                 ($token->isGivenKind(T_CONSTANT_ENCAPSED_STRING) && '"' === $content[0])
                 || ($token->isGivenKind(T_ENCAPSED_AND_WHITESPACE) && $doubleQuoteOpened)
-            );
+            ;
             $isHeredocSyntax = !$isSingleQuotedString && !$isDoubleQuotedString;
             if (
                 (false === $this->configuration['single_quoted'] && $isSingleQuotedString)
--- a/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
+++ b/src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
@@ -75,6 +75,15 @@ EOF;
     /**
      * {@inheritdoc}
      */
+    public function getPriority()
+    {
+        // Should run before single_quote
+        return 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
         static $singleQuotedRegex = '/(?<!\\\\)\\\\(?![\\\'\\\\])/';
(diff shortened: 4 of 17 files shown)


src/Fixer/StringNotation/ExplicitEscapeFixer.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+final class ExplicitEscapeFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Escape implicit backslashes.',
```
keradus: please provide some reasoning
https://3v4l.org/GOOLv

Slamdunk:
@keradus TL,DR: ease of reading

### On single-quoted strings

In single-quoted strings only `'` must be escaped, and `\` is optional, but mandatory at the end of string. This can be a burden while writing code because
```php
$a = 'My\Prefix\Full';
```
is ok, but if you need to refactor the code for the prefix only, you need to remember to add the backslash:
```php
$a = 'My\Prefix\\';
```
Later you need to put the full prefix again, but forget to remove the escaped backslash, and you end up in a mixed result:
```php
$a = 'My\Prefix\\Full';
```
Moral of the story:
1. There are two ways of doing the same thing
1. One is mandatory in some contexts

So let's use that one everywhere to prevent excessive brain-cost in writing and reading.

### On double-quoted strings & heredoc syntax

[Escaped chars here is a mess to remember.](https://secure.php.net/manual/en/language.types.string.php#language.types.string.syntax.double)
Tell me in one second which ones are special chars and which one are a backslash with a following normal char:
```php
var_dump("\d" === "\\d");           // bool(true)
var_dump("\e" === "\\e");           // bool(false)
var_dump("\f" === "\\f");           // bool(false)
var_dump("\n" === "\\n");           // bool(false)
var_dump("\s" === "\\s");           // bool(true)
var_dump("\r" === "\\r");           // bool(false)
var_dump("\t" === "\\t");           // bool(false)
var_dump("\0" === "\\0");           // bool(false)
var_dump("\8" === "\\8");           // bool(true)
var_dump("\xA" === "\\xA");         // bool(false)
var_dump("\xZ" === "\\xZ");         // bool(true)
var_dump("\u" === "\\u");           // bool(true)
var_dump("\u{0}" === "\\u{0}");     // bool(false)
```
https://3v4l.org/tb4PQ

keradus:
I know.
I raised my comment at `FixerDefinition` definition. Please provide reasoning in `FixerDefinition`, not PR comment ;)
So ppl that do not know - will have it explained

src/Fixer/StringNotation/ExplicitEscapeFixer.php
Slamdunk: Done

Slamdunk:
After running this fixer against PHP-CS-Fixer, i've noticed that:

1. The algorithm is good
1. In some context like this the result reduces the readibility, expecially for single-quoted strings

So I'm gonna make it configurable, with double-quoted and heredocs enabled by default, and single-quoted disabled by default

tests/Fixer/StringNotation/EscapeImplicitBackslashesFixerTest.php
```diff
@@ -0,0 +1,194 @@
...
+
+EOF
+,
+            ],
+            [
+<<<'EOF'
+<?php
+$var = "\e\f\n\r\t\v \\ \$ \"";
+$var = <<<HEREDOC_SYNTAX
+\e\f\n\r\t\v \\ \$ \"
```
julienfalque: Double-quotes in HEREDOC syntax do not need to be escaped so the backslash before it has no special meaning and should be doubled.

Slamdunk: Thank you for the pinpoint, done

Slamdunk:
Hi, I've set this fixer to run before `SingleQuoteFixer` and `HeredocToNowdocFixer`, so these two fixers can move a cleared dobuel-quote or heredoc syntax to a simplier single-quote or nowdoc syntax.

I've added this fixer to `.php_cs.dist` and you can see the result in the last commit: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155/commits/c53f656df3c1416f7baacefe785c3bbe288d7f9a

I think the result is good, the heredoc syntax in `HelpCommand` was escaped by this new fixer and then `HeredocToNowdocFixer` moved it to a nowdoc syntax, making it easier to manage.

On the other side, a string like this:
```php
"<?php namespace A\B;  class C {}"
```

julienfalque:
> So the result has a redundant backslash, but this is a `SingleQuoteFixer` flaw to not strip it.

Wouldn't this fixer change
```php
'<?php namespace A\B;  class C {}'
```
to
```php
'<?php namespace A\\B;  class C {}'
```
anyway?

Slamdunk:
Yes but only with `single_quoted` enabled, which is disabled by default

src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
```diff
@@ -0,0 +1,156 @@
...
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        static $singleQuotedRegex = '/(?<!\\\\)\\\\(?![\\\'\\\\])/';
+        static $doubleQuotedRegex = '/(?<!\\\\)\\\\(?![efnrtv$"\\\\0-7]|x[0-9A-Fa-f]|u{)/';
+        static $heredocSyntaxRegex = '/(?<!\\\\)\\\\(?![efnrtv$\\\\0-7]|x[0-9A-Fa-f]|u{)/';
+
+        $doubleQuoteOpened = false;
+        foreach ($tokens as $index => $token) {
+            $content = $token->getContent();
+            if (null === $token->getId() && '"' === $content) {
```
julienfalque: `if ($token->equals('"'))`

Slamdunk:
Take a look at those fixers code:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ab2e189d94698178988f9732bc75bb4ce8d16f77/src/Fixer/StringNotation/SingleQuoteFixer.php#L60-L61
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ab2e189d94698178988f9732bc75bb4ce8d16f77/src/Fixer/StringNotation/SingleQuoteFixer.php#L60-L61
They both skip to fix when backslashes ares odd.
This fixer makes backslashes even (when possible), so those two fixer can handle them properly.

I wouldn't call it a bug: they just don't fix **all** fixable single-quotes/heredoc, but only a secure subset.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
```diff
@@ -169,7 +169,7 @@ private function isNoBreakComment(Token $token)
 
         $text = $this->configuration['comment_text'];
 
-        return preg_match("~^((//|#)\s*$text\s*)|(/\*\*?\s*$text\s*\*/)$~i", $token->getContent());
+        return preg_match("~^((//|#)\\s*$text\\s*)|(/\\*\\*?\\s*$text\\s*\\*/)$~i", $token->getContent());
```
keradus: shouldn't we use `'` for string notation here?

julienfalque: The string contains variables, I would keep using `"`.

src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
```diff
@@ -0,0 +1,156 @@
...
+
+$doubleQuoted = "Interpret my \n but not my \a";
+
+$hereDoc = <<<HEREDOC
+Interpret my \100 but not my \999
+HEREDOC;
+
+EOF;
+
+        return new FixerDefinition(
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#issuecomment-336856953 is still awaiting
When John Doe will look at rule, he will have no clue why he shall follow it.

src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
```diff
@@ -0,0 +1,156 @@
...
+                continue;
+            }
+
+            // Nowdoc syntax
+            if ($token->isGivenKind(T_ENCAPSED_AND_WHITESPACE) && '\'' === substr(rtrim($tokens[$index - 1]->getContent()), -1)) {
+                continue;
+            }
+
+            $isSingleQuotedString = $token->isGivenKind(T_CONSTANT_ENCAPSED_STRING) && '\'' === $content[0];
+            $isDoubleQuotedString = (
```
keradus:
```diff
-( $a || $b )
+$a || $b
```

README.rst
```diff
@@ -537,6 +537,19 @@ Choose from the list of available rules:
...
+
+  Escape implicit backslashes in strings and heredocs to increase code
+  readability.
+
+  Configuration options:
+
+  - ``double_quoted`` (``bool``): whether to fix double-quoted strings; defaults to
+    ``true``
+  - ``heredoc_syntax`` (``bool``): whether to fix heredoc syntax; defaults to ``true``
+  - ``single_quoted`` (``bool``): whether to fix single-quoted strings; defaults to
```
keradus: why one would like to use it, pointing to already pointed example of this config:
```diff
-echo '<?php namespace A\B;  class C {}'
+echo '<?php namespace A\\B;  class C {}'
```

`\` has no special meaning inside single_quoted text except of escaping the quote. so `'\n'` doesn't make new line even if `\` is not double backslashed

src/Console/Command/ReadmeCommand.php
```diff
@@ -236,15 +236,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $help = preg_replace("#^\n( +\\$ )#m", "\n.. code-block:: bash\n\n$1", $help);
         $help = preg_replace("#^\n( +<\\?php)#m", "\n.. code-block:: php\n\n$1", $help);
         $help = preg_replace_callback(
-            "#^\s*<\?(\w+).*?\?>#ms",
+            '#^\\s*<\\?(\\w+).*?\\?>#ms',
```
keradus: so, it was admitted that escaping in single quoted string doesn't make it better, why to escape ?

src/Fixer/ControlStructure/NoBreakCommentFixer.php
Slamdunk: I think it would be better to write it like:
```php
sprintf('/my-regex-%s/', preg_quote($text));
```
I'm not a regex ninja, but I always had trobules on legacy code that doesn't escape regex variables.

README.rst
Slamdunk: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#issuecomment-336855147 ?

src/Console/Command/ReadmeCommand.php
Slamdunk: This PR escaped the double-quoted string, then `SingleQuoteFixer` found an even number of backslashes and converted it into a single-quoted string.

src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
Slamdunk: Tell me if it's ok now.

src/Fixer/ControlStructure/NoBreakCommentFixer.php
julienfalque: Nice catch! `preg_quote` added in #3263, thanks!

src/Console/Command/ReadmeCommand.php
keradus: so, we could use single `'` notation now ?

README.rst
keradus: you see, it's not intuitive and extra edgy case ;P

src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
keradus: In PR, to advocate the feature, you had to put a lot of description.
Yet, in `FixerDefinition`, you use only short `summary` parameter (1st param). You know that it also has 3rd param for extended, detailed `description` ? :)

src/Console/Command/ReadmeCommand.php
Slamdunk: I don't get what you expect from this PR.

If as a final result you would like a single-quoted string without escaped backslashes, than I think SingleQuoteFixer should be in charge to discard (with a dedication option) the superfluous backslashes.

Or maybe you are simply requesting me to manually edit the particular file, don't you?

README.rst
Slamdunk: Please be more explicit on what you would like me to do.

I think https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#issuecomment-336855147 is a valid point, but I agree it's edgy and is in fact disabled by default.

If you want me to delete it, I'll delete it.
If you want me to add more description (on this flag, I already know I have to improve the general summary), I'll add more description.

As of yet, I have no idea what I have to do 🦆

src/Console/Command/ReadmeCommand.php
keradus: yeah, if we disabled escaping in single quoted strings by default, no need to escape it here because of old config of rule

README.rst
keradus: at least descr would be nice ;)

src/Console/Command/ReadmeCommand.php
Slamdunk: Still I don't get what you want.
Please reply with "Slamdunk please [verb] [object]".

keradus: sure thing!
@Slamdunk , please consider applying the following patch:
```diff
-'#^\\s*<\\?(\\w+).*?\\?>#ms',
+'#^\s*<\?(\w+).*?\?>#ms',
```
as you configured the fixer to not touch single-quoted strings by default.

Slamdunk: > please consider applying the following patch

Ok, done manually.

> as you configured the fixer to not touch single-quoted strings by default.

This statement seems to me senseless. A code like `"a\b\c"` with this fixer and SingleQuoteFixer would always result in `'a\\b\\c'`, the `single_quoted` configuration doesn't matter because is not involved.

keradus: in that case, `single_quote` shall run before `escape_implicit_backslashes`

tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test
```diff
@@ -0,0 +1,11 @@
+--TEST--
+Integration of fixers: escape_implicit_backslashes,single_quote.
+--RULESET--
+{"escape_implicit_backslashes": {"double_quoted": true}, "single_quote": true}
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#discussion_r151743010

src/Console/Command/ReadmeCommand.php
Slamdunk: I'm against this. Single quote string is less complex to write and understand, and fixers that work together toward this goal should be encouraged.
I understand though that the result may be verbose: I can propose a new fixer opposite to this one that removes useless backslashes in single quoted strings. That new fixer would run at the end, so we could gain:

1. This fixer straighten backslashes everywhere
1. SingleQuote/HeredocToNowdoc move everything to a less complex string type
1. The new fixer removes verbosity

julienfalque: IMO removing extra backslashes could be a new option in `single_quote` rather than a new fixer.

keradus:
```json
{"single_quote": true, "escape_implicit_backslashes": {"single_quoted": false}}
```

current result:
```diff
-"#^\s*<\?(\w+).*?\?>#ms",
+'#^\\s*<\\?(\\w+).*?\\?>#ms',
```

Wow, but we just configured that single_quoted strings shall not have `\` escaped!
With current order, it creates conflict between 2 rules.

----------------

now, imagine that the order is reversed:

```json
{"single_quote": true, "escape_implicit_backslashes": {"single_quoted": false}}
```

new result:```diff
-"#^\s*<\?(\w+).*?\?>#ms",
+'#^\\s*<\\?(\\w+).*?\\?>#ms',
```

**Comment by Slamdunk** [comment]
> `$ php php-cs-fixer describe escape_implicit_backslashes`
> 
> In PHP, double-quoted strings and heredocs some chars like `n`, `$` or `u` have special meanings if preceeded by a backslash, (and some are special only if followed by
other special chars) while a backslash preceding other chars are interpreted like a plain backslash. The precise list of those special chars is hard to remember and to
identify quickly: this fixer escapes backslashes that do not start a special interpretation with the char after them.
It is possible to fix also single-quoted strings: in this case there is no special chars apart from single-quote and backslash itself, so the fixer simply ensure that all backslashes are escaped. Both single and double backslashes are allowed in single-quoted strings, so the purpose in this context is mainly to have a uniformed way to have them written all over the codebase.

@keradus WDYT?

**Comment by keradus** [comment]
the only thing I would change is to put the comma after parenthesis, not before

File: tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test (hunk 9 above)
Comment (keradus) [reply tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test]: ping @Slamdunk 
if I see correctly, it's the last raised concern

Comment (Slamdunk) [reply tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test]: I think the extended description of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#issuecomment-347001043 can be enough: one single point to retrieve all the infos about how and why doing stuffs.

Comment (keradus) [reply tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test]: yet, in this integration test, guy asked to not fix single quotes. and ultimatelly, they got fixed.

if in this confix would be a request to fix single_quoted strings as well, then output would be great. yet there wasn't.

Comment (Slamdunk) [reply tests/Fixtures/Integration/priority/escape_implicit_backslashes,single_quote.test]: I think one of us is not getting the whole picture.

1. SingleQuoteFixer (which is not part of this PR) will always fix well-formatted double-quoted strings into single-quoted strings
1. EscapeImplicitBackslashesFixer, when configured with `double_quoted` enabled, will always fix non-well-formatted double-quoted strings to well-formatted double-quoted strings

So if we run SingleQuoteFixer before EscapeImplicitBackslashesFixer we would simply end up that, in a *second* run, SingleQuoteFixer will fix what EscapeImplicitBackslashesFixer fixed in the first run.

Hence, as they are logically consecutive, EscapeImplicitBackslashesFixer must run before SingleQuoteFixer.

> yet, in this integration test, guy asked to not fix single quotes. and ultimatelly, they got fixed.

No, single-quoted didn't get fixed. A double-quoted string got fixed into a single-quoted string by SingleQuoteFixer.

Take this example:

```php
$a = "foo \ bar";
```

With EscapeImplicitBackslashesFixer configured with `double_quoted: true` (and no matter whether EscapeImplicitBackslashesFixer's `single_quoted` config is false or true) will fix it into:

```php
$a = "foo \\ bar";
```

Only then, SingleQuoteFixer find the double-quoted string, sees it as well-formatted (all the escapes are even) and so it can move it into a single-quoted string:

```php
$a = 'foo \\ bar';
```

keradus: OK, now I get it. thanks for explanation and patience.

keradus:
Thank you @Slamdunk.

PR #3160, opened 2017-10-17
Add DeprecatedFixerInterface

kubawerlos:
As @julienfalque suggested instead of checking description let's generate README accodring to `DeprecatedFixerInterface`.

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -18,6 +18,7 @@ use PhpCsFixer\Differ\DiffConsoleFormatter;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\DefinedFixerInterface;
+use PhpCsFixer\Fixer\DeprecatedFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSampleInterface;
 use PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface;
@@ -28,6 +29,7 @@ use PhpCsFixer\FixerFactory;
 use PhpCsFixer\RuleSet;
 use PhpCsFixer\StdinFileInfo;
 use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Utils;
 use PhpCsFixer\WordMatcher;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Formatter\OutputFormatter;
@@ -144,8 +146,18 @@ final class DescribeCommand extends Command
             $definition = new FixerDefinition('Description is not available.', []);
         }
 
+        $description = $definition->getSummary();
+        if ($fixer instanceof DeprecatedFixerInterface) {
+            $successors = $fixer->getSuccessorsNames();
+            $message = null === $successors
+                ? 'not supported anymore'
+                : sprintf('use %s instead', Utils::naturalLanguageJoinWithBackticks($successors));
+            $message = preg_replace('/(`.+?`)/', '<info>$1</info>', $message);
+            $description .= sprintf(' <error>DEPRECATED</error>: %s.', $message);
+        }
+
         $output->writeln(sprintf('<info>Description of</info> %s <info>rule</info>.', $name));
-        $output->writeln($definition->getSummary());
+        $output->writeln($description);
         if ($definition->getDescription()) {
             $output->writeln($definition->getDescription());
         }
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -149,8 +149,8 @@ final class DescribeCommand extends Command
         $description = $definition->getSummary();
         if ($fixer instanceof DeprecatedFixerInterface) {
             $successors = $fixer->getSuccessorsNames();
-            $message = null === $successors
-                ? 'not supported anymore'
+            $message = [] === $successors
+                ? 'will be removed on next major version'
                 : sprintf('use %s instead', Utils::naturalLanguageJoinWithBackticks($successors));
             $message = preg_replace('/(`.+?`)/', '<info>$1</info>', $message);
             $description .= sprintf(' <error>DEPRECATED</error>: %s.', $message);
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -21,6 +21,7 @@ use PhpCsFixer\Fixer\FixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerOptionInterface;
 use PhpCsFixer\FixerFactory;
 use PhpCsFixer\RuleSet;
+use PhpCsFixer\Utils;
 use Symfony\Component\Console\Command\HelpCommand as BaseHelpCommand;
 use Symfony\Component\Console\Formatter\OutputFormatter;
 use Symfony\Component\Console\Formatter\OutputFormatterStyle;
@@ -451,7 +452,11 @@ EOF
             }
 
             if ($fixer instanceof DeprecatedFixerInterface) {
-                $description .= sprintf(' DEPRECATED: use `%s` instead.', $fixer->getSuccessor());
+                $successors = $fixer->getSuccessorsNames();
+                $message = null === $successors
+                    ? 'not supported anymore'
+                    : sprintf('use %s instead', Utils::naturalLanguageJoinWithBackticks($successors));
+                $description .= sprintf(' DEPRECATED: %s.', $message);
             }
 
             $description = implode("\n   | ", self::wordwrap(
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -453,8 +453,8 @@ EOF
 
             if ($fixer instanceof DeprecatedFixerInterface) {
                 $successors = $fixer->getSuccessorsNames();
-                $message = null === $successors
-                    ? 'not supported anymore'
+                $message = [] === $successors
+                    ? 'will be removed on next major version'
                     : sprintf('use %s instead', Utils::naturalLanguageJoinWithBackticks($successors));
                 $description .= sprintf(' DEPRECATED: %s.', $message);
             }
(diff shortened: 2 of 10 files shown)


tests/AutoReview/FixerTest.php
```diff
@@ -176,17 +177,17 @@ public function testDeprecatedFixersHaveCorrectSummary(FixerInterface $fixer)
...
-                '/\. DEPRECATED: use `[a-z_]+` instead\.$/',
-                $fixer->getDefinition()->getSummary(),
-                'Deprecated fixer must contain correct "DEPRECATED" note in summary'
-            );
-        } else {
-            $this->assertNotRegExp(
-                '/DEPRECATED/',
-                $fixer->getDefinition()->getSummary(),
-                'Non-deprecated fixer cannot contain word "DEPRECATED" in summary'
+            $this->assertInstanceOf(
```
julienfalque: I would reverse the condition: if a fixer implements the `DeprecatedFixerInterface`, it must have the `@deprecated` annotation, otherwise it must not.

src/Fixer/DeprecatedFixerInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ * @author Kuba Werłos <werlos@gmail.com>
+ */
+interface DeprecatedFixerInterface extends FixerInterface
+{
+    /**
+     * Returns name of the fixer to use instead.
+     *
+     * @return string
+     */
+    public function getSuccessor();
```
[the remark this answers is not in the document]
kubawerlos: Sound a YAGNI to me - do we consider any existing fixer to be removed without an alternative? Similarly, I can't think of any case we would split a fixer into smaller ones. We could make the interface internal, so in future we could easily implement this change if it would be necessary.

kubawerlos:
@julienfalque it was [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3141#issuecomment-335377036) comment that start me to think we can have this for something useful actually - not only as help for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3146.

src/Fixer/DeprecatedFixerInterface.php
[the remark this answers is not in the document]
SpacePossum: >Sound a YAGNI to me - do we consider any existing fixer to be removed without an alternative? 

PSR0 fixer is likely to be dropped in 3.0 without an alternative.

>Similarly, I can't think of any case we would split a fixer into smaller ones. We could make the interface internal, so in future we could easily implement this change if it would be necessary.

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2619/files#diff-88b99bb28683bd5b7e3a204826ead112R1025

this case is happening now/soon :)

src/Console/Command/HelpCommand.php
```diff
@@ -449,6 +450,10 @@ function (FixerInterface $a, FixerInterface $b) {
                 $description = '[n/a]';
             }
 
+            if ($fixer instanceof DeprecatedFixerInterface) {
```
keradus: this condition is needed also for `DescribeCommand`

src/Fixer/DeprecatedFixerInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
```
keradus: @SpacePossum @juliendufresne 
we don't want ppl to implement this interface, for that it shall be internal.
or, we want to allow ppl the programatically check if given concrete fixer follows this interface?

src/Fixer/DeprecatedFixerInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+interface DeprecatedFixerInterface extends FixerInterface
+{
+    /**
+     * Returns name of the fixer to use instead.
+     *
+     * @return string
```
keradus: also, if it's providing a name, then show it in method name as well.
for now, looking only at method name, I actually expected `FixerInterface` as output

src/Fixer/DeprecatedFixerInterface.php
julienfalque: I would keep it public, not `@internal`. This can be useful for people that use this project as a lib to deprecate their own fixers.

SpacePossum: I agree with @julienfalque here, although I've no strong opinion on it, but it could be nice for 3rd party fixer providers to deprecate their fixers as well using the mechanic.

src/Console/Command/HelpCommand.php
kubawerlos: @keradus done

src/Fixer/DeprecatedFixerInterface.php
keradus: thanks for opinion :+1: let it stay public

src/Console/Command/DescribeCommand.php
```diff
@@ -144,8 +146,18 @@ private function describeRule(OutputInterface $output, $name)
             $definition = new FixerDefinition('Description is not available.', []);
         }
 
+        $description = $definition->getSummary();
+        if ($fixer instanceof DeprecatedFixerInterface) {
+            $successors = $fixer->getSuccessorsNames();
+            $message = null === $successors
```
julienfalque: `getSuccessorsNames()` isn't supposed to return `null` but an empty array.

src/Console/Command/DescribeCommand.php
```diff
@@ -144,8 +146,18 @@ private function describeRule(OutputInterface $output, $name)
...
+                ? 'not supported anymore'
```
julienfalque: This is not true: a deprecated fixer is still supported. The message should rather say that the fixer will be removed on next major version.

src/Console/Command/HelpCommand.php
```diff
@@ -451,7 +452,11 @@ function (FixerInterface $a, FixerInterface $b) {
             }
 
             if ($fixer instanceof DeprecatedFixerInterface) {
-                $description .= sprintf(' DEPRECATED: use `%s` instead.', $fixer->getSuccessor());
+                $successors = $fixer->getSuccessorsNames();
+                $message = null === $successors
```
julienfalque: `getSuccessorsNames()` isn't supposed to return `null` but an empty array.

src/Console/Command/HelpCommand.php
```diff
@@ -451,7 +452,11 @@ function (FixerInterface $a, FixerInterface $b) {
...
+                    ? 'not supported anymore'
```
julienfalque: This is not true: a deprecated fixer is still supported. The message should rather say that the fixer will be removed on next major version.

src/Fixer/DeprecatedFixerInterface.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace PhpCsFixer\Fixer;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ */
+interface DeprecatedFixerInterface extends FixerInterface
+{
+    /**
+     * Returns names of fixers to use instead.
```
julienfalque: I would add _", if any"_ at the end of the sentence so that it is more explicit that the returned array can be empty.

tests/AutoReview/FixerTest.php
```diff
@@ -176,18 +177,16 @@ public function testDeprecatedFixersHaveCorrectSummary(FixerInterface $fixer)
...
-                $fixer->getDefinition()->getSummary(),
-                'Deprecated fixer must contain correct "DEPRECATED" note in summary'
-            );
-        } else {
-            $this->assertNotRegExp(
-                '/DEPRECATED/',
-                $fixer->getDefinition()->getSummary(),
-                'Non-deprecated fixer cannot contain word "DEPRECATED" in summary'
-            );
+        $this->assertNotRegExp(
```
julienfalque: `assertContains`?

src/Utils.php
```diff
@@ -181,4 +181,26 @@ function ($a, $b) {
...
+            return sprintf('`%s`', $name);
+        }, $names);
+
+        $last = array_pop($names);
+
+        if ($names) {
+            return implode(', ', $names).' and '.$last;
+        }
+
+        return $last;
```
julienfalque: I would return an empty string instead of `null` here. But if you prefer returning `null`, you need to update the `@return` PHPDoc tag.

tests/UtilsTest.php
```diff
@@ -226,6 +226,39 @@ public function testSortFixers()
         );
     }
 
+    /**
+     * @dataProvider provideNaturalLanguageJoinWithBackticksCases
+     *
+     * @param string $joined
```
julienfalque: Should be `null|string`.

src/Console/Command/DescribeCommand.php
```diff
@@ -144,8 +146,18 @@ private function describeRule(OutputInterface $output, $name)
             $definition = new FixerDefinition('Description is not available.', []);
         }
 
+        $description = $definition->getSummary();
+        if ($fixer instanceof DeprecatedFixerInterface) {
```
julienfalque: Please add a test in `DescribeCommandTest` to cover this case.

src/Utils.php
```diff
@@ -181,4 +181,26 @@ function ($a, $b) {
...
     }
+
+    /**
+     * Join names in natural language wrapped in backticks, e.g. `a`, `b` and `c`.
+     *
+     * @param string[] $names
+     *
+     * @return string
+     */
+    public static function naturalLanguageJoinWithBackticks(array $names)
```
keradus: empty array is invalid input

kubawerlos: what do you mean? do you want exception to be thrown for empty array?

julienfalque: IMO it's fine, just like `implode` does not trigger any error when an empty array is used.

keradus: we won't call it for empty array, right?
yet, currently there is a boilerplate code to support handling it. and with it, we shall check if `output !== ''` whenever we want to use this method

tests/UtilsTest.php
```diff
@@ -226,6 +226,9 @@ public function testSortFixers()
         );
     }
 
+    /**
+     * @requires PHPUnit 5.2
```
keradus: drop it and use `->setExpectedException`

SpacePossum:
maybe a nice enhancement would be made in the [RuleSetTest](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.7/tests/RuleSetTest.php) to make sure we've no fixers in there that implement the new `DeprecatedFixerInterface`?
(you typing and thinking at the same time here)

kubawerlos:
I'm planning to do this in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3146 - together with testing that fixers implementing `DeprecatedFixerInterface` will trigger deprecation error. Of course after this one is merged.

keradus:
Thank you @kubawerlos.

PR #3162, opened 2017-10-19
Add udiff as diff format

SpacePossum:
```
[03:52 PM]-[possum@zoo1]-[/home/possum/work/PHP-CS-Fixer]-[git master] 
$ php php-cs-fixer fix --dry-run --diff-format=unidiff --diff  -vvv
.......................................................................................................................................................................................................................................................................................................F..............................................................................F..................................................................................................................................................................................
   1) src/Console/ConfigurationResolver.php (phpdoc_order, phpdoc_separation, phpdoc_trim)
--- Original
+++ New
@@ -252,9 +252,9 @@
     }

     /**
+     * @throws InvalidConfigurationException
+     *
      * @return DifferInterface
-     *
-     * @throws InvalidConfigurationException
      */
     public function getDiffer()
     {

   2) src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php (method_argument_space)
--- Original
+++ New
@@ -201,8 +201,7 @@
     case 42:
         break;
 }
-'
-                    ,
+',
                     [
                         'statements' => ['switch'],
                     ]


Checked all files in 31.170 seconds, 12.000 MB memory used
```

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/299
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2529
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1620

### DIFF
--- a/src/Fixer/Whitespace/CompactNullableTypehintFixer.php
+++ b/src/Fixer/Whitespace/CompactNullableTypehintFixer.php
@@ -33,7 +33,7 @@ final class CompactNullableTypehintFixer extends AbstractFixer
             'Remove extra spaces in a nullable typehint.',
             [
                 new VersionSpecificCodeSample(
-                    "<?php\nfunction sample(? string \$str): ? string\n{}",
+                    "<?php\nfunction sample(? string \$str): ? string\n{}\n",
                     new VersionSpecification(70100)
                 ),
             ],


src/Fixer/Alias/EregToPregFixer.php
```diff
@@ -49,7 +49,7 @@ public function getDefinition()
     {
         return new FixerDefinition(
             'Replace deprecated `ereg` regular expression functions with preg.',
-            [new CodeSample('<?php $x = ereg(\'[A-Z]\');')],
+            [new CodeSample("<?php \$x = ereg('[A-Z]');\n")],
```
keradus: why are all those new lines needed?

julienfalque: I was wondering too. I guess this is required to avoid seeing a bunch of `\ No newline at end of file` in code samples.

src/Console/ConfigurationResolver.php
```diff
@@ -258,15 +260,20 @@ public function getDiffer()
             $mapper = [
                 'null' => function () { return new NullDiffer(); },
                 'sbd' => function () { return new SebastianBergmannDiffer(); },
+                'unidiff' => function () { return new UnifiedDiffer(); },
```
keradus: is `unidiff` a common name? so far, we used `udiff` in discussion

src/Fixer/Alias/EregToPregFixer.php
SpacePossum: @julienfalque is right here, the diff samples do not look good without the ending line break, still correct unified diff but with the `\No newline at end of file` notice

src/Console/ConfigurationResolver.php
SpacePossum: both are common use, but I agree, lets use `udiff`, will update :+1:

tests/Console/ConfigurationResolverTest.php
```diff
@@ -1060,7 +1061,7 @@ public function testResolveCommandLineInputOverridesDefault()
         $this->assertSame(['php_unit_construct' => true], $resolver->getRules());
         $this->assertFalse($resolver->getUsingCache());
         $this->assertNull($resolver->getCacheFile());
-        $this->assertInstanceOf(\PhpCsFixer\Differ\SebastianBergmannDiffer::class, $resolver->getDiffer());
+        $this->assertInstanceOf(\PhpCsFixer\Differ\UnifiedDiffer::class, $resolver->getDiffer());
```
julienfalque: Please add a test that covers various combinations of `diff` and `diff-format` options.

tests/AutoReview/FixerTest.php
```diff
@@ -70,6 +71,9 @@ public function testFixerDefinitions(FixerInterface $fixer)
 
             $code = $sample->getCode();
             $this->assertStringIsNotEmpty($code, sprintf('[%s] Sample #%d', $fixerName, $sampleCounter));
+            if (!($fixer instanceof SingleBlankLineAtEofFixer)) {
+                $this->assertSame("\n", substr($code, -1), sprintf('[%s] Sample must end with linebreak #%d', $fixerName, $sampleCounter));
```
julienfalque: _"... Sample **#%d** must..."_.

src/Differ/UnifiedDiffer.php
```diff
@@ -0,0 +1,43 @@
...
+{
+    /**
+     * @var Differ
+     */
+    private $differ;
+
+    public function __construct()
+    {
+        $this->differ = new Differ(new UnifiedDiffOutputBuilder([
+            'fromFile' => 'Original', // TODO
```
julienfalque: What is left to do here?

SpacePossum: ideal case would be that the `diff` contains the file name (and not 'Original')

src/Console/ConfigurationResolver.php
```diff
@@ -258,15 +260,20 @@ public function getDiffer()
...
             ];
 
-            $option = $this->options['diff'] ? 'sbd' : 'null';
-
-            if (!isset($mapper[$option])) {
-                throw new InvalidConfigurationException(sprintf(
-                    'Differ must be "sbd" or "null", got "%s".',
-                    $this->options['diff']
-                ));
+            if ($this->options['diff-format']) {
```
keradus: ``--diff-format=udiff` without passing `--diff` will still produce diff ?

src/Console/ConfigurationResolver.php
```diff
@@ -258,15 +260,20 @@ public function getDiffer()
...
+                $option = $this->options['diff-format'];
+                if (!isset($mapper[$option])) {
+                    throw new InvalidConfigurationException(sprintf(
+                        '"diff-format" must be any of "%s", got "%s".',
+                        implode('", "', array_keys($mapper)),
+                        $option
+                    ));
+                }
+            } else {
+                $option = $this->options['diff'] ? 'sbd' : 'null';
```
keradus: TODO: change to `udiff` on 3.0

tests/Console/ConfigurationResolverTest.php
```diff
@@ -1060,21 +1061,25 @@ public function testResolveCommandLineInputOverridesDefault()
...
         $this->assertSame('json', $resolver->getReporter()->getFormat());
     }
 
     /**
      * @param string      $expected
-     * @param bool|string $differConfig
+     * @param bool|string $diffConfig
```
keradus: from cases I see it's null-able as well

src/Console/ConfigurationResolver.php
SpacePossum: yes it will, there is a utest for it

keradus: weird

SpacePossum: yeah, well, I feel don't anything for one or the other, so happy to modify the PR in any way that makes it merge-able ;)

keradus: are we fine with dropping the differ option on 3.0, leaving only udiff ?

SpacePossum: Depends if the people asking for `diff only` are OK with having the `udiff` format as proposed by this PR. If not, than it would be nice to keep the `differ-format` option on 3.0.
I'm not very happy with introducing `diff-format` here TBH, it is only done so we keep our BC promise.

keradus: that/s exactly why I;m asking, to see what we want to have on 3,0 line.

julienfalque: I'm not sure other formats will have actual value once true udiff will be available so I'm 👍 for dropping them in 3.0 along with the option, unless there are other useful formats we could add in the meantime.

keradus:
I rebased PR, is there anything left here @SpacePossum ?

SpacePossum:
I think is one is ready, there is https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3162#discussion_r146838541 and the nightly failure though

keradus:
Thank you @SpacePossum.

Issue #2767, opened 2017-05-12
IndentationTypeFixer - fixes white space that is not part of indention


Arctic76:
Hello guys!

It seems I can't get the "binary_operator_spaces" rule working in my config.
When I activate this rule, it is working well the first time I save ("Fix on save" option on VSCode), but then if I try to fix the file again or to save it, the rule detects that the previous work is not enough and add indentation again... so it breaks everything :/..
In my mind i think because I have tabs enabled instead of spaces, so may be it makes conflicts?

My PHP version is:
```
PHP 5.6.30 (cli) (built: May 10 2017 14:28:57) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
```
My OS is `MAC OS Sierra 10.12.4 (16E195)`.
My editor is VSCode, using the PHP CS Fixer extension [link](https://marketplace.visualstudio.com/items?itemName=junstyle.php-cs-fixer).
My PHP CS fixer version is: `PHP CS Fixer 2.3.1 by Fabien Potencier and Dariusz Ruminski`.

My config for the rules:
```
return PhpCsFixer\Config::create()
    ->setRules([
        'header_comment' => ['header' => $header],
        "@PSR1"                                 => true,
        '@PSR2'                                 => true,
        'binary_operator_spaces'                => ['align_double_arrow' => true, 'align_equals' => true],
        'blank_line_after_opening_tag'          => true,
        'braces'                                => ['position_after_functions_and_oop_constructs' => 'same'],
        'concat_space'                          => ['spacing' => 'one'],
        'hash_to_slash_comment'                 => true,
        'linebreak_after_opening_tag'           => true,
        'method_separation'                     => true,
        'no_blank_lines_after_class_opening'    => true,
        'no_empty_comment'                      => true,
        'no_empty_phpdoc'                       => true,
        'semicolon_after_instruction'           => true,
        'indentation_type'                      => true,
    ])
    ->setIndent("\t")
    ->setLineEnding("\n")
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->exclude('tests/Fixtures')
            ->in(__DIR__)
    )
;
```
VSCode config:
```
{
    "php-cs-fixer.onsave": true,
    "php-cs-fixer.config": ".vscode/.php_cs",
    "php-cs-fixer.autoFixBySemicolon": true,
}
```

Code sample after written
```
[...]

PR #3172, opened 2017-10-25
IndentationTypeFixer - do not touch whitespace that is not indentation

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2767

### DIFF
--- a/src/Fixer/Whitespace/IndentationTypeFixer.php
+++ b/src/Fixer/Whitespace/IndentationTypeFixer.php
@@ -27,6 +27,11 @@ use PhpCsFixer\Tokenizer\Tokens;
 final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
 {
     /**
+     * @var string
+     */
+    private $indent;
+
+    /**
      * {@inheritdoc}
      */
     public function getDefinition()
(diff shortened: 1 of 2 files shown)


src/Fixer/Whitespace/IndentationTypeFixer.php
```diff
@@ -60,36 +65,80 @@ public function isCandidate(Tokens $tokens)
...
+     * @param int    $index
+     *
+     * @return Token
+     */
+    private function fixIndentToken(Tokens $tokens, $index)
+    {
+        $content = $tokens[$index]->getContent();
+        $previousTokenHasTrailingLinebreak = false;
+
+        // TODO on 3.x this can be removed when we have a transformer for "T_OPEN_TAG" to "T_OPEN_TAG + T_WHITESPACE"
```
keradus: I don't think we gonna split it, we took proof of concept about it once and it ended up to not be worth it

julienfalque: Really? I think there are quite a few `if`s out there just to check whether a whitespace has an open tag before it. This would allow to simplify those and remove the need to care about it (which is a pain in the ass IMO).

keradus: exactly, few. @SpacePossum why initial attempt to have that transformer failed?

SpacePossum: I stopped working on it because it creates a lot of changes, which I would've to maintain up to the point we've a RC branch for 3.0

keradus:
Thank you @SpacePossum.

tests/Fixer/Whitespace/IndentationTypeFixerTest.php
```diff
@@ -281,6 +281,50 @@ public function provideMessyWhitespacesCases()
...
+
+        $this->doTest($input, $expected);
+    }
+
+    public function provideMessyWhitespacesReversedCases()
+    {
+        $filteredCases = array();
+        $cases = $this->provideMessyWhitespacesCases();
+
+        foreach ($cases as $key => $case) { // TODO on 5.6 bump use array_filter with ARRAY_FILTER_USE_KEY
```
keradus: @SpacePossum please send PR to master

PR #3173, opened 2017-10-25
SimplifiedNullReturnFixer - handle nullable return types

Slamdunk:
Which is the right branch for this kind of improvement?

- [x] Don't fix nullable return types
- [x] Remove `risky` flag

### DIFF
--- a/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
+++ b/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
@@ -130,15 +130,15 @@ final class SimplifiedNullReturnFixer extends AbstractFixer
     {
         $functionIndex = $index;
         do {
-            $functionIndex = $tokens->getPrevTokenOfKind($functionIndex, [[T_FUNCTION]]);
+            $functionIndex = $tokens->getPrevTokenOfKind($functionIndex, array(array(T_FUNCTION)));
             if (null === $functionIndex) {
                 return false;
             }
-            $openingCurlyBraceIndex = $tokens->getNextTokenOfKind($functionIndex, ['{']);
+            $openingCurlyBraceIndex = $tokens->getNextTokenOfKind($functionIndex, array('{'));
             $closingCurlyBraceIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $openingCurlyBraceIndex);
         } while ($closingCurlyBraceIndex < $index);
 
-        $nullableTypeIndex = $tokens->getNextTokenOfKind($functionIndex, [[CT::T_NULLABLE_TYPE]]);
+        $nullableTypeIndex = $tokens->getNextTokenOfKind($functionIndex, array(array(CT::T_NULLABLE_TYPE)));
 
         return null !== $nullableTypeIndex && $nullableTypeIndex < $openingCurlyBraceIndex;
     }
--- a/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
+++ b/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
@@ -30,9 +30,7 @@ final class SimplifiedNullReturnFixer extends AbstractFixer
     {
         return new FixerDefinition(
             'A return statement wishing to return `void` should not return `null`.',
-            array(new CodeSample('<?php return null;')),
-            null,
-            'Risky since PHP 7.1 as `null` and `void` can be hinted as return type and have different meaning.'
+            array(new CodeSample('<?php return null;'))
         );
     }
 
@@ -51,14 +49,6 @@ final class SimplifiedNullReturnFixer extends AbstractFixer
     public function isCandidate(Tokens $tokens)
     {
         return $tokens->isTokenKindFound(T_RETURN);
-    }
-
-    /**
-     * {@inheritdoc}
-     */
-    public function isRisky()
-    {
-        return true;
     }
 
     /**
--- a/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
+++ b/src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
@@ -138,8 +138,7 @@ final class SimplifiedNullReturnFixer extends AbstractFixer
             $closingCurlyBraceIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $openingCurlyBraceIndex);
         } while ($closingCurlyBraceIndex < $index);
 
-        $functionLastBraceIndex = $tokens->getPrevTokenOfKind($openingCurlyBraceIndex, [')']);
-        $nullableTypeIndex = $tokens->getNextTokenOfKind($functionLastBraceIndex, [[CT::T_NULLABLE_TYPE]]);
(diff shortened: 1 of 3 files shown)


Slamdunk:
I would like to remove the `risky` flag, since the fixer wouldn't produce anymore erroneous code.

What do you think?

Slamdunk:
@keradus ping, at least for labels 🙏

keradus:
@Slamdunk , sorry for big delay, simply too many things recently and i hoped someone else will pick the thread up (I was active in threads I participated previously, but was trying to not dive in too much threads at same time :( )

keradus:
label/milestone:
it's debatable and I could advocate both approaches, but so far we treated "better support of newer language syntax" as bugfixes and sent those to lowest supported branch (currently: 2.2)

tests/Fixer/ReturnNotation/SimplifiedNullReturnFixerTest.php
```diff
@@ -55,4 +55,44 @@ public function provideFixCases()
...
+            [
+                '<?php function foo(): ?int { $bar = function() { return; }; return null; }',
+                '<?php function foo(): ?int { $bar = function() { return null; }; return null; }',
+            ],
+            [
+                '<?php function foo(): int { return; }',
+                '<?php function foo(): int { return null; }',
+            ],
+            [
+                '<?php function foo(): void { return; }',
```
keradus: please add a case with pure `: null` (not `:?...`) as well

tests/Fixer/ReturnNotation/SimplifiedNullReturnFixerTest.php
```diff
@@ -55,4 +55,44 @@ public function provideFixCases()
...
+            [
+                '<?php function foo() { return; } function bar(): ?A\B\C\D { return null; } function baz() { return; }',
+                '<?php function foo() { return null; } function bar(): ?A\B\C\D { return null; } function baz() { return null; }',
+            ],
+            [
+                '<?php function foo(): ?int { $bar = function() { return; }; return null; }',
+                '<?php function foo(): ?int { $bar = function() { return null; }; return null; }',
+            ],
+            [
+                '<?php function foo(): int { return; }',
```
keradus: > Fatal error: A function with return type must return a value

src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
```diff
@@ -101,21 +102,48 @@ private function clear(Tokens $tokens, $index)
...
+        $content = rtrim($content, ');');
 
         return 'null' === strtolower($content);
     }
 
     /**
+     * Is the return within a function with a nullable return type?
+     *
+     * @param Tokens $tokens
+     * @param int    $index
```
keradus: this is an index of ?

tests/Fixer/ReturnNotation/SimplifiedNullReturnFixerTest.php
Slamdunk:
```
function foo(): int { return null; }
// Fatal error: Uncaught TypeError: Return value of foo() must be of the type integer, null returned

function foo(): int { return; }
// Fatal error: A function with return type must return a value
```
Both before and after the fix there is a Fatal error.

The reason why I wrote this test is to prove **this** PR doesn't change the (erroneous) behaviour of the fixer.

src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
Slamdunk: Just copied from the other fixer methods. Btw I've added the  description now.

Slamdunk:
> label/milestone:
> it's debatable and I could advocate both approaches, but so far we treated "better support of newer language syntax" as bugfixes and sent those to lowest supported branch (currently: 2.2)
> (in the end, currently 2.2 removes the retrurn **null** where it shouldn't)
> 
> WDYT?

My request for label/milestone was just to set some reminders. I respect your great efforts and time spent in this project, I don't expect a rapid merge, just that without labels it's easy to lose the plot.

Rebased onto 2.2, I agree.

tests/Fixer/ReturnNotation/SimplifiedNullReturnFixerTest.php
keradus: I agree that initial code was wrong as well. IMO we shall not change one wrong code into another, but simply don't touch it.

Slamdunk: Done.

src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
```diff
@@ -101,20 +102,51 @@ private function clear(Tokens $tokens, $index)
...
+    {
+        $functionIndex = $returnIndex;
+        do {
+            $functionIndex = $tokens->getPrevTokenOfKind($functionIndex, array(array(T_FUNCTION)));
+            if (null === $functionIndex) {
+                return false;
+            }
+            $openingCurlyBraceIndex = $tokens->getNextTokenOfKind($functionIndex, array('{'));
+            $closingCurlyBraceIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $openingCurlyBraceIndex);
+        } while ($closingCurlyBraceIndex < $returnIndex);
```
keradus: ha, so that's how you skip lambdas ?

keradus: finding `closingCurlyBraceIndex` is expensive, we don't care about that information for lambdas, you use it here only to determine if a `T_FUNCTION` a lambda or not.
what about using `TokensAnalyzer::isLambda` instead?

keradus:
for me, now the fixer doesn't look risky indeed.

src/Fixer/ReturnNotation/SimplifiedNullReturnFixer.php
Slamdunk: The purpose here is different.
Keep in mind that this fixer starts its algorithm from return keyword: what method signature this return belongs to?
Now imagine:

* a class method with a nullable return type
* that contains an anonymous class with a non-nullable method
* that contains an anonymous class with a nullable method
* that contains an anonymous class with a non-nullable method

If in this Matryoshka doll each method also has a return statement, we need this loop to determine each correspondent method signature.

The second-last test case represent this situation.

keradus: whole method, yes.
But, isn't the `do... while` construction just to skip the content of internal lambda ?

Slamdunk: No, it's to find out the function's signature the return belongs to.
It goes backward, if the function ends before return skip it (aka keep going backward), if the function ends after return use it.
It doesn't matter which type of function it is.

Slamdunk:
Riskyness removed

Slamdunk:
Since the removal of riskyness, it wasn't clear enough what now this Fixer does. I've added a `VersionSpecificCodeSample` to be clear.

keradus:
Thank you @Slamdunk.

PR #3177, opened 2017-10-26
Remove @requires PHPUnit 5.2 from test

kubawerlos:
I've got [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3160#discussion_r146700434) instruction from @keradus and I based that `@requires` on this one. So let's change it to avoid similar situation in the future.


keradus:
Thanks for PR !
Yet, let me replace it with #3182

PR #3180, opened 2017-10-28
Fix usage of static variables

kubawerlos:


keradus:
Thank you @kubawerlos.

PR #3182, opened 2017-10-28
Add support for PHPUnit 6, drop PHPUnit 4

keradus:


composer.json
```diff
@@ -32,10 +32,10 @@
         "symfony/stopwatch": "^3.0 || ^4.0"
     },
     "require-dev": {
-        "johnkary/phpunit-speedtrap": "^1.1",
+        "johnkary/phpunit-speedtrap": "^1.1 || ^2.0@dev",
         "justinrainbow/json-schema": "^5.0",
         "php-cs-fixer/accessible-object": "^1.0",
-        "phpunit/phpunit": "^4.8.35 || ^5.4.3",
+        "phpunit/phpunit": "^5.7.23 || ^6.4.3",
```
kubawerlos: Sort dependencies by the way?

keradus: done

tests/Console/ConfigurationResolverTest.php
```diff
@@ -988,8 +981,10 @@ public function testResolveRulesWithOption()
 
     public function testResolveRulesWithUnknownRules()
     {
-        $this->setExpectedException(
-            \PhpCsFixer\ConfigurationException\InvalidConfigurationException::class,
+        $this->expectException(
+            \PhpCsFixer\ConfigurationException\InvalidConfigurationException::class
```
localheinz: Quick note: it appears that `PhpCsFixer\ConfigurationException\InvalidConfigurationException` has already been imported, so while this out of scope of the changes suggested in this PR, what do you think about simplifying the FQCN here, as well as putting the invocation on a single line?

keradus: doing it in this one place is pointless, we already have a future request to import all usages like here, that would fix issue globally

.appveyor.yml
```diff
@@ -5,7 +5,6 @@ clone_folder: C:\projects\php-cs-fixer
 
 environment:
     matrix:
-        - php_ver: 7.1.2
```
julienfalque: What was wrong with this version?

julienfalque:
Arf, already merged, but 👍 anyway :)

.appveyor.yml
keradus: it was freezing on phpunit execution

keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3188

keradus:
done by #3201

PR #3183, opened 2017-10-30
FinalInternalClassFixer introduction

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1389

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -572,6 +572,21 @@ Choose from the list of available rules:
   Converts implicit variables into explicit ones in double-quoted strings
   or heredoc syntax.
 
+* **final_internal_class**
+
+  Internal classes should be ``final``.
+
+  *Risky rule: changing classes to ``final`` might cause code execution to break.*
+
+  Configuration options:
+
+  - ``annotation-black-list`` (``array``): class level annotations tags that must be
+    omitted to fix the class, even if others are in white list. (case in
+    sensitive); defaults to ``['@final', '@Entity', '@ORM']``
+  - ``annotation-white-list`` (``array``): class level annotations tags that must be
+    set in order to fix the class. (case in sensitive); defaults to
+    ``['@internal']``
+
 * **full_opening_tag** [@PSR1, @PSR2, @Symfony]
 
   PHP code must use the long ``<?php`` tags or short-echo ``<?=`` tags and not
--- a/README.rst
+++ b/README.rst
@@ -581,8 +581,8 @@ Choose from the list of available rules:
   Configuration options:
 
   - ``annotation-black-list`` (``array``): class level annotations tags that must be
-    omitted to fix the class, even if others are in white list. (case in
-    sensitive); defaults to ``['@final', '@Entity', '@ORM']``
+    omitted to fix the class, even if all of the white list ones are used
+    as well. (case in sensitive); defaults to ``['@final', '@Entity', '@ORM']``
   - ``annotation-white-list`` (``array``): class level annotations tags that must be
     set in order to fix the class. (case in sensitive); defaults to
     ``['@internal']``
--- a/src/ConfigurationException/InvalidConfigurationException.php
+++ b/src/ConfigurationException/InvalidConfigurationException.php
@@ -20,6 +20,7 @@ use PhpCsFixer\Console\Command\FixCommandExitStatusCalculator;
  * @author SpacePossum
  *
  * @internal
+ * @final Only internal extending this class is supported
  */
 class InvalidConfigurationException extends \InvalidArgumentException
 {
--- a/src/ConfigurationException/InvalidFixerConfigurationException.php
+++ b/src/ConfigurationException/InvalidFixerConfigurationException.php
@@ -20,6 +20,7 @@ use PhpCsFixer\Console\Command\FixCommandExitStatusCalculator;
  * @author SpacePossum
  *
  * @internal
+ * @final Only internal extending this class is supported
  */
 class InvalidFixerConfigurationException extends InvalidConfigurationException
 {
--- a/src/Fixer/ClassNotation/FinalInternalClassFixer.php
+++ b/src/Fixer/ClassNotation/FinalInternalClassFixer.php
@@ -1,0 +1,168 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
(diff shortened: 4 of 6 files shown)


README.rst
```diff
@@ -515,6 +515,13 @@ Choose from the list of available rules:
 
   *Risky rule: risky if the ``ereg`` funcion is overridden.*
 
+* **final_internal_class**
```
keradus: please add it to `.php_cs.dist` as we follow the rule (and using the fixer will be nice POC it works how we want to)

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All internal classes should be final except abstract ones. Warning! This could change code behavior.',
+            [
+                new CodeSample("<?php\n/**\n * @internal\n */\nclass Sample\n{\n}\n"),
+            ],
+            null,
+            'Changes code behavior.'
```
keradus: I think it might be helpful to say why ;)

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -0,0 +1,102 @@
...
+ */
+final class FinalInternalClassFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All internal classes should be final except abstract ones. Warning! This could change code behavior.',
```
keradus: warning is no longer needed in v2, when we have proper `FixerDefinition` object

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -0,0 +1,102 @@
...
+            $docToken = $tokens[$tokens->getPrevNonWhitespace($index)];
+
+            // ignore class if it has no class-level docs
+            if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $doc = new DocBlock($docToken->getContent());
+
+            // ignore class if it is has no @internal annotation
```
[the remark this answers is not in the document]
keradus: not sure is it a good idea in the end to depends on `@internal`, as we want to make classes final even when they are not internal.
Sf makes classes final via `@final` when they couldn't have real `final` (due to BC promise)

actually, I would mark as final all classes and then add some options:
- `ignore_when_@final`
- `ignore_when_@ORM`

src/ConfigurationException/InvalidFixerConfigurationException.php
```diff
@@ -20,6 +20,7 @@
  * @author SpacePossum
  *
  * @internal
+ * @final TODO make class final on 3.0
```
keradus: please mark `TODO` as `@TODO` ;) (so here, you would add 2 new annotations
also, as class is internal, we could make it really final.it was internal since beginning: https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/87242c89f47a9662d2f413f99e8c4f51b7029785

README.rst
julienfalque: Won't this make mocking internal classes in tests harder?

keradus: no. you don't mock classes. you mock interfaces.

src/ConfigurationException/InvalidFixerConfigurationException.php
```diff
@@ -21,7 +21,7 @@
  *
  * @internal
  */
-class InvalidFixerConfigurationException extends InvalidConfigurationException
+final class InvalidFixerConfigurationException extends InvalidConfigurationException
```
keradus: won't work actually as `InvalidForEnvFixerConfigurationException` is extending this class.
initial `@final` annotation mislead me

src/ConfigurationException/InvalidConfigurationException.php
```diff
@@ -20,6 +20,7 @@
  * @author SpacePossum
  *
  * @internal
+ * @final Only internal extending this class is supported
```
keradus: i must admit this looks weird.

if it's not final we know we could extend it, and we do know it's possible only internally as `@internal` is there, so any contact with this class is internal

README.rst
julienfalque: Interfaces are used to depend on abstractions instead of concrete implementations. Creating an interface for any `@internal` class you need to mock for testing purposes looks a bit overkill to me.  Would you then mark such interfaces as `@internal`? Obviously they cannot be `final` so the problem still exists in the end.

keradus: I see no reason why not mark interface with `@internal` if it is not supposed to be used externally

also, keep in mind that you don't always want/need to mock in first place

keradus:
I got lost ! I see squash ...

SpacePossum:
the previous version contained a lot of debug statements, so the effective diff is not really big...

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -32,6 +33,23 @@
...
+    {
+        parent::configure($configuration);
+
+        $intersect = array_intersect_assoc(
+            $this->configuration['annotation-white-list'],
+            $this->configuration['annotation-black-list']
+        );
+
+        if (count($intersect)) {
+            throw new InvalidFixerConfigurationException($this->getName(), sprintf('Annotation cannot be used in both the white- and black list, got duplicates: "%s".', implode('", "', array_keys($intersect))));
```
keradus: shouldn't this be on config-definition level ?

tests/Fixer/ClassNotation/FinalInternalClassFixerTest.php
```diff
@@ -284,4 +284,17 @@ public function provideAnonymousClassesCases()
             ],
         ];
     }
+
+    public function testConfigureSameAnnotationInBothLists()
+    {
+        $this->expectException(\PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException::class);
+        $this->expectExceptionMessageRegExp(
```
keradus: so nice to see new interface used ;d

src/Fixer/ClassNotation/FinalInternalClassFixer.php
SpacePossum: when validating one configuration option you don't have access to the other options to check for conflicts, I think.

keradus: you could have
first param of normalizer option is
`The closure is invoked when {@link resolve()} is called. The closure has access to the resolved values of other options through the passed {@link Options} instance.`

keradus: grep for `->setNormalizer` in our project

src/ConfigurationException/InvalidConfigurationException.php
keradus: class is not final. annotation is wrong.

src/Fixer/ClassNotation/FinalInternalClassFixer.php
SpacePossum: I don't see how it would work.
This tests if the two normalized option values do not conflict. Within both the `setAllowedValues` and `setNormalizer` callbacks there is no access to the both the normalized values.
This could _only_ work if there was some way to add a hook that is called after the normalization of all options has been done, but I don't know of such hook.

src/ConfigurationException/InvalidConfigurationException.php
SpacePossum: yeah, it is basically this case https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1389#issuecomment-139918457 :)
Guess we've no good answer now. We could add some other tag to this class and put that in the black list though.

src/Fixer/ClassNotation/FinalInternalClassFixer.php
keradus: hook for `setNormalizer` got 2 params, first is all options, so we have access to option we want to cross-validate

keradus:
Thank you @SpacePossum.

PR #3185, opened 2017-10-31
IndentationTypeFixerTest - clean up

SpacePossum:
for ref.: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3172#discussion_r147987886

### DIFF
--- a/tests/Fixer/Whitespace/IndentationTypeFixerTest.php
+++ b/tests/Fixer/Whitespace/IndentationTypeFixerTest.php
@@ -316,15 +316,12 @@ function myFunction() {
 
     public function provideMessyWhitespacesReversedCases()
     {
-        $filteredCases = [];
-        $cases = $this->provideMessyWhitespacesCases();
-
-        foreach ($cases as $key => $case) { // TODO on 5.6 bump use array_filter with ARRAY_FILTER_USE_KEY
-            if (!is_string($key) || false === strpos($key, 'mix indentation')) {
-                $filteredCases[] = $case;
-            }
-        }
-
-        return $filteredCases;
+        return array_filter(
+            $this->provideMessyWhitespacesCases(),
+            static function($key) {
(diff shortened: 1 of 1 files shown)


tests/Fixer/Whitespace/IndentationTypeFixerTest.php
```diff
@@ -316,15 +316,12 @@ public function testMessyWhitespacesReversed($expected, $input = null)
...
-        foreach ($cases as $key => $case) { // TODO on 5.6 bump use array_filter with ARRAY_FILTER_USE_KEY
-            if (!is_string($key) || false === strpos($key, 'mix indentation')) {
-                $filteredCases[] = $case;
-            }
-        }
-
-        return $filteredCases;
+        return array_filter(
+            $this->provideMessyWhitespacesCases(),
+            static function($key) {
```
keradus: interesting... :+1:
do you know if it affects memory (like all statics are globals ?) if not, we could build a fixer to have all lambdas static

SpacePossum:
```
[03:09 PM]-[possum@zoo1]-[/home/possum/work/PHP-CS-Fixer]-[git master] 
$ g st
## master...origin/master
A  showTokens.php
A  test.php
 M tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
A  tmp/FileOpeningFixer.php
A  tmp/StaticLambdaFixer.php
A  tmp/StaticLambdaFixerTest.php
A  tmp/test.php
A  tmp/test2.php
A  tmp/test3.php
```

almost done ;)

in case of memory;
static anonymous functions in PHP do not get `$this` access (which is nice in case of scoping/leaking) which keeps the reference count to the object (`$this`) lower. However I'm not sure if this really makes a difference, I didn't it test it reliable enough yet.

keradus:
Thank you @SpacePossum.

PR #3186, opened 2017-10-31
PhpUnitIndicator - Add tests

localheinz:
This PR

* [ ] adds tests for `PhpUnitIndicator`

Follows https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2708.

🤓 Still needs a bit of work, have a few failing test cases I will push.

### DIFF
--- a/tests/AutoReview/ProjectCodeTest.php
+++ b/tests/AutoReview/ProjectCodeTest.php
@@ -49,7 +49,6 @@ final class ProjectCodeTest extends TestCase
         \PhpCsFixer\FixerConfiguration\FixerOptionValidatorGenerator::class,
         \PhpCsFixer\FixerConfiguration\InvalidOptionsForEnvException::class,
         \PhpCsFixer\FixerFileProcessedEvent::class,
-        \PhpCsFixer\Indicator\PhpUnitTestCaseIndicator::class,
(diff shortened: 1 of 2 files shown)


tests/Indicator/PhpUnitIndicatorTest.php
```diff
@@ -0,0 +1,266 @@
...
+    {
+        $cases = [
+            'nothing' => '<?php
+
+class Foo
+{}
+',
+            'nothing-extends-old' => '<?php
+
+class Foo extends \PHPUnit_Framework_TestCase
```
localheinz: I am not sure about this case.

While I personally would mark an abstract test case as `abstract`, maybe there are people who don't. Then this would indicate a false negative for them (also case below).

keradus:
@localheinz 

let us figure out the state here:
- is this still WIP?
- is this PR adds more value to currently existing tests/Indicator/PhpUnitTestCaseIndicatorTest.php ?

if so, I'm supper happy to have this PR rebased and finished :+1:

SpacePossum:
This test and other changes are really great; however it seems I kind of did something the same when changing the logic for `PhpUnitTestCaseIndicator`.

With regrets I'm going to close this one as it got out dated, but please @localheinz lemme know if any of your changes should be part of the code base after all, happy to :+1: if this got updated!

PR #3187, opened 2017-10-31
StaticLambdaFixer - introduction

SpacePossum:
Lambda's not (indirect) referencing `$this` must be declared `static`.

```
$ php php-cs-fixer describe static_lambda
Description of static_lambda rule.
Lambda's not (indirect) referencing `$this` must be declared `static`.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,2 +1,2 @@
    <?php
   -$a = function () use ($b)
   +$a = static function () use ($b)
```


ntzm:
`Lambda's` should be `Lambdas`

localheinz:
@SpacePossum 

Interesting! Was going to ask why that makes sense - then I found this: http://php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static

SpacePossum:
thanks @ntzm , I'm no native English speaker (and even within my native language I'm not very good with spelling and grammar), so corrections like yours are most welcome :)

SpacePossum:
@localheinz your link reminds me that I should mark this fixer after all,
as we've no way to check if somewhere a `bindTo` call is done on the callable... sadly

tests/Fixer/FunctionNotation/StaticLambdaFixerTest.php
```diff
@@ -0,0 +1,212 @@
...
+    {
+        return [
+            [
+                '<?php
+                    class A
+                    {
+                        public function B()
+                        {
+                            $a = function () {
+                                $b = \'this\';
```
GrahamCampbell: What about the use of `$this` inside an `eval`?

SpacePossum: good ol' eval... 
fair point though :+1: , I can update to check if `eval` is used within the lambda

Slamdunk: `eval` can't be handled: how would you detect `eval(' return ' . strrev('; siht$ '));`?

Mark the fixer risky and give up the eval.

keradus: if skipping `eval` in general would make fixer non-risky, then don't handle `eval` at all.
or even more - if we need some tricks to hande eval, let us not do it. we don't need to support crap

README.rst
```diff
@@ -1254,6 +1254,12 @@ Choose from the list of available rules:
 
   Replace all ``<>`` with ``!=``.
 
+* **static_lambda**
```
keradus: could we add it to `.php_cs.dist` ?

tests/Fixer/FunctionNotation/StaticLambdaFixerTest.php
SpacePossum: The fixer will skip the lambda if `eval` is used within the body of the lambda,
same as for example when `require_once` is used etc.
This makes the fixer far more save to use, but at the small cost that some cases are not fixed while it could be. Than again, if you use `eval` (in a lambda) than you won't find much help from other tooling like SCA, basically you're on your own with stuff like that ;)

Slamdunk: @SpacePossum I would find useful to add your last thoughts in the extended description.

README.rst
keradus: ping @SpacePossum

SpacePossum:
expect tests to fail

keradus:
Thank you @SpacePossum.

PR #3188, opened 2017-10-31
AppVeyor - add PHP 7.x

keradus:
it will fail most probably...
let us treat it as TODO


julienfalque:
Looks like AppVeyor works fine again. There is still one failure but we know it's related to a recent change in symfony/finder.

julienfalque:
Green now :)

This should be merged on 2.11 I think.

keradus:
Thank you @julienfalque.

PR #3189, opened 2017-10-31
Add PHP_CS_FIXER_FUTURE_MODE env

keradus:
Adds `PHP_CS_FIXER_FUTURE_MODE` env variable, which make fixer configuration to work like we want it to work on 3.0, like changing default progress and differ.

It doesn't check fixer default configuration, as it would change behaviour of given fixer depends on that option, making run of that fixer unstable (in terms that run of fixer will not provide same output depends on env variable, which will break cache and CI integration)

### DIFF
--- a/src/AbstractFixer.php
+++ b/src/AbstractFixer.php
@@ -117,7 +117,7 @@ abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
 
         if (null === $configuration) {
             if (getenv('PHP_CS_FIXER_FUTURE_MODE')) {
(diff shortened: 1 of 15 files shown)


src/AbstractFixer.php
```diff
@@ -116,6 +116,10 @@ public function configure(array $configuration = null)
         }
 
         if (null === $configuration) {
+            if (getenv('PHP_CS_FIXER_FUTURE_MODE')) {
+                throw new \InvalidArgumentException('Parameter must not be `null`.');
```
SpacePossum: this check looks off, `getenv` will retrun `false|string` so never `null` if IIRC?

PR #3193, opened 2017-11-01
TokensAnalyzer::getClassyElements - sort result before returning

SpacePossum:


keradus:
Thank you @SpacePossum.

PR #3195, opened 2017-11-01
Add self-update command test

julienfalque:


src/Console/Command/SelfUpdateCommand.php
```diff
@@ -123,36 +136,29 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $remoteFilename = $this->toolInfo->getPharDownloadUri($remoteTag);
 
-        try {
-            $copyResult = @copy($remoteFilename, $tempFilename);
-            if (false === $copyResult) {
-                $output->writeln(sprintf('<error>Unable to download new version %s from the server.</error>', $remoteTag));
-
-                return 1;
-            }
+        $copyResult = @copy($remoteFilename, $tempFilename);
```
SpacePossum: no need for variable introduction, it is only used one time in the `if` below

tests/TextDiffTest.php
```diff
@@ -36,7 +36,9 @@
      */
     public function testDiffReportingDecorated($expected, $format, $isDecorated)
     {
-        $command = new FixCommand();
+        $command = new FixCommand(
+            $this->prophesize('PhpCsFixer\ToolInfo')->reveal()
```
SpacePossum: does this test need a `prophesize` version of `ToolInfo`?

src/Console/SelfUpdate/NewVersionChecker.php
```diff
@@ -19,7 +19,7 @@
 /**
  * @internal
  */
-final class NewVersionChecker
+class NewVersionChecker
```
SpacePossum: this is done for `prophesize` onyl?

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -75,21 +90,19 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            $latestVersion = $this->versionChecker->getLatestVersion();
+            $latestVersionOfCurrentMajor = $this->versionChecker->getLatestVersionOfMajor($currentMajor);
         } catch (\Exception $exception) {
             $output->writeln(sprintf(
                 '<error>Unable to determine newest version: %s</error>',
                 $exception->getMessage()
             ));
 
-            return 0;
+            return 1;
```
SpacePossum: good catch :+1: maybe this should go lower versions as well?

tests/TextDiffTest.php
julienfalque: Not sure, I'll have a look.

src/Console/Command/SelfUpdateCommand.php
julienfalque: Target branch is already 2.2 ;)

src/Console/SelfUpdate/NewVersionChecker.php
julienfalque: Yes. This is exactly the reason of my comment on [#3183](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3183#discussion_r147670555).

keradus: please bring final and create interface.

src/Console/Command/SelfUpdateCommand.php
keradus: this bugfix deserves dedicated PR that gonna be merged sooner:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3196

src/PharChecker.php
```diff
@@ -0,0 +1,36 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * @internal
+ */
+class PharChecker
```
keradus: shall be final

keradus:
big one, I will need deeper look.
can we somehow split it ?

src/Console/SelfUpdate/NewVersionChecker.php
julienfalque: TBH I don't see the point in creating an interface for testing purposes only.

keradus: `final` prevents partial mocks, which are extra bad approach imo.
i like to be defensive.

julienfalque: What do you mean partial mocks? Mocks with some methods defined in the test and some not defined? If so, using an interface does not prevent this. Also, when using Prophecy, as soon as a mock has at least one defined method, all methods used in the tested code must be defined as well. Calling an undefined method will throw an exception, which should prevent the issue.

tests/TextDiffTest.php
julienfalque: Removed unneeded dummies.

src/Console/Command/SelfUpdateCommand.php
julienfalque: Variable removed.

src/Console/SelfUpdate/NewVersionChecker.php
keradus: with built-in mock system of PHPUnit, it's possible to mock real class with using mocked one method but not another. final class, will forbid that totally.

julienfalque: We don't use PHPUnit mocks but Prophecy anyway so this still looks pointless to me.

keradus: final or not is a general rule. using final is being defensive. partial mocks were example, not the only reason.

julienfalque:
Updated to keep final classes final and use interfaces for mocks instead.

@keradus I think some small parts could be extracted, but most of the changes are directly tied to adding the `SelfUpdateCommand` test, so not worth it IMO.

src/Console/Command/FixCommand.php
```diff
@@ -68,14 +69,17 @@
      */
     private $defaultConfig;
 
-    public function __construct()
+    private $toolInfo;
```
keradus: typehinting is missing

src/Console/Application.php
```diff
@@ -33,16 +37,24 @@
 {
     const VERSION = '2.2.10';
 
+    private $toolInfo;
```
keradus: typehinting is missing

src/Console/Command/SelfUpdateCommand.php
```diff
@@ -33,6 +33,24 @@
 {
     const COMMAND_NAME = 'self-update';
 
+    private $versionChecker;
+
+    private $toolInfo;
+
+    private $pharChecker;
```
keradus: typehintings are missing

src/Console/ConfigurationResolver.php
```diff
@@ -104,6 +104,8 @@
      */
     private $configFinderIsOverridden;
 
+    private $toolInfo;
```
keradus: same

keradus:
1.3k lines on diff, not sure if "not worth it" indeed

julienfalque:
See #3246.

keradus:
#3246 got merged. please merge/rebase with 2.2

julienfalque:
Rebased.

src/PharCheckerInterface.php
```diff
@@ -0,0 +1,24 @@
...
+
+namespace PhpCsFixer;
+
+/**
+ * @internal
+ */
+interface PharCheckerInterface
+{
+    /**
+     * @param string $filename
```
keradus: what is the output ?

julienfalque: Updated.

keradus:
Thank you @julienfalque.

PR #3135, opened 2017-10-07
Add ArrayIndentationFixer

julienfalque:
Closes #770 and #1438.

PR #3197, opened 2017-11-02
DX groom CS

keradus:


julienfalque:
Extracted from/applied with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3135?

keradus:
yes, yet not only.
that way diff will be smaller and amount of conflicts between 2.2 <> master will decrease

PR #3199, opened 2017-11-02
TokensAnalyzer - getClassyElements iterates too many times

SpacePossum:
When searching within an anonymous class stop at the end of that.

### DIFF
--- a/src/Tokenizer/TokensAnalyzer.php
+++ b/src/Tokenizer/TokensAnalyzer.php
@@ -527,10 +527,11 @@ final class TokensAnalyzer
     /**
      * @param Tokens $tokens
      * @param int    $start
+     * @param bool   $searchingOnFileLevel if searching on file level
      *
      * @return array
      */
-    private function findClassyElements(Tokens $tokens, $start)
+    private function findClassyElements(Tokens $tokens, $start, $searchingOnFileLevel = true)
     {
         $elements = array();
         $inClass = false;
@@ -551,7 +552,7 @@ final class TokensAnalyzer
             }
 
             if ($token->isClassy()) { // anonymous class in class
-                $elements += $this->findClassyElements($tokens, $index);
+                $elements += $this->findClassyElements($tokens, $index, false);
 
                 continue;
             }
@@ -578,6 +579,10 @@ final class TokensAnalyzer
                 --$curlyBracesLevel;
 
                 if (0 === $curlyBracesLevel) {
+                    if (!$searchingOnFileLevel) {
+                        break;
+                    }
+
                     $inClass = false;
                 }
 
--- a/src/Tokenizer/TokensAnalyzer.php
+++ b/src/Tokenizer/TokensAnalyzer.php
@@ -538,12 +538,12 @@ final class TokensAnalyzer
     /**
      * Find classy elements.
      *
-     * Searches in tokens from the classy index till the end of the classy.
-     * Returns an 2D array; (int) the index until the method has analysed, (array) found classy elements.
+     * Searches in tokens from the classy (start) index till the end (index) of the classy.
+     * Returns an array; first value is the index until the method has analysed (int), second the found classy elements (array).
      *
      * @param int $index classy index
      *
-     * @return array<int, array>
+     * @return array
      */
     private function findClassyElements($index)
     {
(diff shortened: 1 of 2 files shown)


SpacePossum:
messed up the backport from master to 2.2 :/

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -526,10 +526,11 @@ public function isWhilePartOfDoWhile($index)
     /**
      * @param Tokens $tokens
      * @param int    $start
+     * @param bool   $searchingOnFileLevel if searching on file level
```
keradus: not sure here. like if we have class A, inside of which we have anon class B, inside of which we have anon class C...
when we call function to investigate class B and it find out about class C so it is just about to call investigation for class C, this 3rd investigation is not willing to investigate file-level class A indeed, but also it is not supposed to investigate class B as well

SpacePossum: it is a naming thing, if C is in B which is in A, C won't investigate B (anymore), it is more like a flag since only on file level you might have multiple classes you want to check on the same level, inside a class you never want to look outside it

keradus: :+1:
i double check the code, it's like you say.
so now, we just need to find proper name for it.

julienfalque: What about `$stopAfterCurrentClass` (would require to switch the boolean values)?

SpacePossum: I reworked the methods to only check each token one time, WDYT @keradus @julienfalque ?

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -525,33 +536,32 @@ public function isWhilePartOfDoWhile($index)
     }
 
     /**
-     * @param Tokens $tokens
-     * @param int    $start
+     * Find classy elements.
      *
-     * @return array
+     * Searches in tokens from the classy index till the end of the classy.
```
julienfalque: I would write _"... from the classy **start** index **to its end index.**"_

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -525,33 +536,32 @@ public function isWhilePartOfDoWhile($index)
...
+     * Returns an 2D array; (int) the index until the method has analysed, (array) found classy elements.
```
julienfalque: I would write _"Returns an array with last analyzed indexes as keys and arrays of found tokens as values."_

src/Tokenizer/TokensAnalyzer.php
julienfalque: :+1: with minors.

src/Tokenizer/TokensAnalyzer.php
SpacePossum: it only returns one index (the last analyzed) and the elements found

julienfalque: Ah you're right, then I would write: _"Returns an array with the classy end index as first value and an array of found tokens as second value."_

src/Tokenizer/TokensAnalyzer.php
```diff
@@ -525,33 +536,32 @@ public function isWhilePartOfDoWhile($index)
...
+     *
+     * @param int $index classy index
+     *
+     * @return array<int, array>
```
julienfalque: This should be `array<int|array>` (or maybe `array` would be enough given the detailed description above).

SpacePossum: it always return an `array<int, array>` , never `array<int>` or `array<array>` therefore `array<array|int>` would not be correct

src/Tokenizer/TokensAnalyzer.php
keradus: wow, I just saw that. from first glance looks OKish, yet it's important part of code, I need to checkout branch and play with the change

SpacePossum: thanks man :+1: ,
if it is to much trouble I roll back the last commit

keradus: nah, code looks really better now. i just need to analyse it...

keradus:
Thank you @SpacePossum.

tests/Tokenizer/TokensAnalyzerTest.php
```diff
@@ -216,26 +251,62 @@ public function otherFunction1()
                 9 => array(
                     'token' => $tokens[9],
                     'type' => 'method',
+                    // 'classIndex' => 1, @TODO on master
```
keradus: @SpacePossum could you handle the todos please?

PR #3200, opened 2017-11-02
Skip slow test when Xdebug is loaded

julienfalque:
When loaded, Xdebug slows down the data provider a lot. As the data provider is executed even when the related test is excluded with e.g. `--filter` option, this makes debugging tests with Xdebug a bit painful.

Not sure this is the best solution, I'm open to suggestions :)

### DIFF
--- a/tests/AutoReview/ProjectCodeTest.php
+++ b/tests/AutoReview/ProjectCodeTest.php
@@ -331,6 +331,10 @@ final class ProjectCodeTest extends TestCase
 
     public function provideDataProviderMethodNameCases()
     {
+        if (extension_loaded('xdebug') && false === getenv('CI')) {
+            $this->markTestSkipped('Data provider too slow when Xdebug is loaded.');
+        }
+
         $data = array();
 
         $testClassNames = $this->getTestClasses();


keradus:
is there an option to crash PHPUnit when XDebug is enabled? :D

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -331,6 +331,10 @@ function ($item) {
 
     public function provideDataProviderMethodNameCases()
     {
+        if (extension_loaded('xdebug')) {
+            $this->markTestSkipped('Data provider too slow when Xdebug is loaded.');
```
SpacePossum: can we detect if we're on a CI here and than throw an exception? accidentally skipping the tests on CI is unwanted I think

julienfalque: I think none of our CI builds use Xdebug so is this really needed?

julienfalque: [Travis CI](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables), [AppVeyor](https://www.appveyor.com/docs/environment-variables/) and [CircleCI](https://circleci.com/docs/1.0/environment-variables/) all define a `CI` environment variable so I updated my PR to not skip the test when this variable exists.

SpacePossum: thanks :)  good to ship!

keradus:
Thank you @julienfalque.

PR #3201, opened 2017-11-02
Add PHPUnit Migration rulesets and fixers

### LINKED ISSUE #2501: PHPUnit 6 Migration fixer
[PHPUnit 6](https://github.com/sebastianbergmann/phpunit/blob/6.0/ChangeLog-6.0.md#600---2017-02-03) no longer supports `PHPUnit_Framework_TestCase` as the base class for test case classes and requires the use of `PHPUnit\Framework\TestCase` instead.

`PHPUnit\Framework\TestCase` can be used as the base class for test case classes since [PHPUnit 5.4](https://github.com/sebastianbergmann/phpunit/blob/5.4/ChangeLog-5.4.md#540---2016-06-03).

I think a fixer would be useful that migrates PHPUnit test case classes from `PHPUnit_Framework_TestCase` to `PHPUnit\Framework\TestCase`.

keradus:
Closes #2501

I started working on it ages ago, yet it was already reviewed internally.
If you find sth that could be improved, please open a PR, I'm glad to see it!
If you figure out that sth is missing, please raise a feature request.
This PR doesn't contain replacement for getMock call with multiple arguments, as there is no 1-1 replacing method


src/Fixer/PhpUnit/PhpUnitMockFixer.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+final class PhpUnitMockFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Usages of `->getMockWithoutInvokingTheOriginalConstructor` method MUST be replaced by `->createMock` method.',
```
julienfalque: Why not supporting `getMock()` as well?

keradus: see opening ticket

julienfalque: > This PR doesn't contain getMock replacement, as there is no 1-1 replacing method

`getMock()` could be replaced by `createMock()` when only one argument is provided. This is a common case (I think).

SpacePossum:
:+1:  :100:   @keradus  thanks!

PR #3208, opened 2017-11-03
BracesFixer - Fix for instantiation in control structures

### LINKED ISSUE #3202: Braces Fixer adds braces on foreach
When upgrading to version 2.7.2 of PHP CS Fixer, the Braces fixer adds braces in a foreach.

The PHP version you are using:
7.1.8 ... 7.1.10

PHP CS Fixer version you are using:
2.7.2

The command you use to run the fixer.
`vendor/bin/php-cs-fixer fix --dry-run -vvv`

Running the braces fixer generates the following diff.

```diff
diff --git a/src/Browscap/Data/Factory/DataCollectionFactory.php b/src/Browscap/Data/Factory/DataCollectionFactory.php
index c2bac4f..e766712 100644
--- a/src/Browscap/Data/Factory/DataCollectionFactory.php
+++ b/src/Browscap/Data/Factory/DataCollectionFactory.php
@@ -79,7 +79,8 @@ class DataCollectionFactory

         $deviceDirectory = $resourceFolder . '/devices';

-        foreach (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($deviceDirectory)) as $file) {
+        foreach {
+            (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($deviceDirectory)) as $file) {
             /** @var $file \SplFileInfo */
             if (!$file->isFile() || 'json' !== $file->getExtension()) {
                 continue;
@@ -89,9 +90,11 @@ class DataCollectionFactory
             $this->addDevicesFile($file->getPathname());
         }

-        $uaSourceDirectory = $resourceFolder . '/user-agents';
+            $uaSourceDirectory = $resourceFolder . '/user-agents';
+        }

-        foreach (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($uaSourceDirectory)) as $file) {
+        foreach {
+            (new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($uaSourceDirectory)) as $file) {
             /** @var $file \SplFileInfo */
             if (!$file->isFile() || 'json' !== $file->getExtension()) {
                 continue;
@@ -101,7 +104,8 @@ class DataCollectionFactory
             $this->addSourceFile($file->getPathname());
         }

-        return $this->collection;
+            return $this->collection;
+        }
     }
```

The generated code is invalid and leads to an error like
```php
PHP Parse error:  syntax error, unexpected '{', expecting '(' in ...
```

julienfalque:
Fixes #3202.

Not sure how I managed to miss these obvious cases...

### DIFF
--- a/src/Tokenizer/Transformer/BraceClassInstantiationTransformer.php
+++ b/src/Tokenizer/Transformer/BraceClassInstantiationTransformer.php
@@ -69,6 +69,12 @@ final class BraceClassInstantiationTransformer extends AbstractTransformer
             array(T_VARIABLE),
             array(T_CLASS),
             array(T_ARRAY),
+            array(T_IF),
+            array(T_ELSEIF),
+            array(T_SWITCH),
+            array(T_FOR),
+            array(T_FOREACH),
+            array(T_WHILE),
         ))) {
             return;
         }
--- a/src/Tokenizer/Transformer/BraceClassInstantiationTransformer.php
+++ b/src/Tokenizer/Transformer/BraceClassInstantiationTransformer.php
@@ -62,18 +62,19 @@ final class BraceClassInstantiationTransformer extends AbstractTransformer
         }
 
         if ($tokens[$tokens->getPrevMeaningfulToken($index)]->equalsAny(array(
-            array(T_STRING),
             ']',
-            array(CT::T_ARRAY_SQUARE_BRACE_CLOSE),
             array(CT::T_ARRAY_INDEX_CURLY_BRACE_CLOSE),
-            array(T_VARIABLE),
-            array(T_CLASS),
+            array(CT::T_ARRAY_SQUARE_BRACE_CLOSE),
             array(T_ARRAY),
-            array(T_IF),
+            array(T_CLASS),
             array(T_ELSEIF),
-            array(T_SWITCH),
             array(T_FOR),
             array(T_FOREACH),
+            array(T_IF),
+            array(T_STATIC),
+            array(T_STRING),
+            array(T_SWITCH),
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
👍

SpacePossum:
```php
<?php

switch(new \stdClass()) {
    case (new \stdClass()) : {
        echo 123;
    }
}
```

I think `case` and `default` are missing as well.

julienfalque:
In the first example I think we should not exclude the braces: `case` does not require braces at all so it makes sense to treat those as _"instantiation braces"_ IMO.

I'm not sure about `include_once` either: https://3v4l.org/p2Kiq.

SpacePossum:
fair enough, lets add all the (odd) cases in a utest anyway, just to check what happens with the fixers that cause the ticket (3202)

src/Tokenizer/Transformer/BraceClassInstantiationTransformer.php
```diff
@@ -62,18 +62,19 @@ public function process(Tokens $tokens, Token $token, $index)
         }
```
SpacePossum: oops, sorry @julienfalque didn't meant to push directly to your branch, but I don't want to force push it out again... feel free to revert!

SpacePossum: PHP 5.6 fails because my test code is to modern, PHP 7.1 is broken pipe... so all good... well almost ;)

keradus:
Thank you @julienfalque.

julienfalque:
And thanks @SpacePossum :)

PR #3211, opened 2017-11-04
Use udiff format in CI

julienfalque:


keradus:
what about setting `PHP_CS_FIXER_FUTURE_MODE=1` instead ?

SpacePossum:
we want non-future to be tested as well, so maybe for one job on maybe like PHP nightly that makes sense?

julienfalque:
Was about to say the same, wouldn't using `PHP_CS_FIXER_FUTURE_MODE` change other things we don't want to change in CI builds?

keradus:
huh? following that reasoning I could say "we want non-udiff differ be tested as well"
but this PR is changing options passed to executing fix command, not tests. non-udiff differ is still tested. same, setting FUTURE env for fix command won't affect tests.

julienfalque:
It's not a reasoning but a real question: I have no idea what setting `PHP_CS_FIXER_FUTURE_MODE` really does right now.

julienfalque:
Ok I checked and indeed it may be a better alternative, PR updated :)

keradus:
as in readme, it set udiff and estimating-max by default

.appveyor.yml
```diff
@@ -39,4 +39,4 @@ before_test:
 test_script:
     - cd C:\projects\php-cs-fixer
     - vendor\bin\phpunit
-    - php php-cs-fixer --diff --dry-run -v fix
+    - php php-cs-fixer --diff-format=udiff --dry-run -v fix
```
keradus: this one is still old

julienfalque: Using `SOME_VAR=value command` does not work on Windows so I kept using the option instead.

keradus: what about 
```
- set PHP_CS_FIXER_FUTURE_MODE=1
- php php-cs-fixer --diff --dry-run -v fix
```
?

.appveyor.yml
```diff
@@ -27,7 +28,6 @@ install:
     - echo extension_dir=ext >> php.ini
     - echo extension=php_curl.dll >> php.ini
     - echo extension=php_openssl.dll >> php.ini
-    - echo extension=php_mbstring.dll >> php.ini
```
keradus: is it by purpose ?

julienfalque: Nope :)

.appveyor.yml
julienfalque: Done.

keradus:
Thank you @julienfalque.

PR #3212, opened 2017-11-04
Handle rulesets unknown to fabbot.io

julienfalque:


SpacePossum:
:+1: 
maybe we should put `dedicated expection` - classes for these cases on the roadmap for 3.0?

julienfalque:
👍 for dédicaced exceptions.

keradus:
Thank you @julienfalque.

Issue #3204, opened 2017-11-03
Namespaces On Same Line As PHP Opening Tag Are Prefixed With A New Line


moebrowne:
Since #3083 (v2.7.2) was merged namespaces that are defined on the same line as the opening PHP tag are prefixed with a newline when using the `no_blank_lines_before_namespace` rule.

For example the following:

```
<?php namespace Some\Name\Space;
```

Is transformed into:

```
<?php
namespace Some\Name\Space;
```

moebrowne:
Thanks @SpacePossum !

SpacePossum:
you're welcome :)  thanks!

PR #3216, opened 2017-11-06
AbstractLinesBeforeNamespaceFixer - add min. and max. option, not only single target count

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3204


julienfalque:
Does the bug affect 2.2?

keradus:
Thank you @SpacePossum.

Issue #3205, opened 2017-11-03
Files that are claimed to contain errors (but don't) are still changed


sebastianbergmann:
I am using the latest stable release:
```
$ php-cs-fixer --version
PHP CS Fixer 2.8.0 Round Numbers by Fabien Potencier and Dariusz Ruminski (89e7b08)
```
When I invoke `php-cs-fixer` for PHPUnit (https://github.com/sebastianbergmann/phpunit/) then I get the following:
```
$ php-cs-fixer fix
Loaded config default from "/usr/local/src/phpunit/.php_cs".

Fixed all files in 4.629 seconds, 22.000 MB memory used

Files that were not fixed due to errors reported during fixing:
   1) /usr/local/src/phpunit/src/Util/TextTestListRenderer.php
   2) /usr/local/src/phpunit/src/Util/XmlTestListRenderer.php
```
I have no idea what errors `php-cs-fixer` thinks these files have. `php -l` thinks they're fine.
```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

nothing to commit, working tree clean
```
As you can see above, `php-cs-fixer` did not change any files.
```
$ php-cs-fixer fix
Loaded config default from "/usr/local/src/phpunit/.php_cs".
Using cache file ".php_cs.cache".
   1) src/Util/TextTestListRenderer.php
   2) src/Util/XmlTestListRenderer.php

Fixed all files in 0.026 seconds, 8.000 MB memory used
```
However, running `php-cs-fixer` again changes the files that were reported to contain errors (see above) before:
```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/Util/TextTestListRenderer.php
	modified:   src/Util/XmlTestListRenderer.php

no changes added to commit (use "git add" and/or "git commit -a")
```

```
$ git diff
diff --git a/src/Util/TextTestListRenderer.php b/src/Util/TextTestListRenderer.php
index 4ee6ac769..be1d2c3f7 100644
--- a/src/Util/TextTestListRenderer.php
+++ b/src/Util/TextTestListRenderer.php
@@ -20,7 +20,8 @@ public function render(TestSuite $suite): string
     {
         $buffer = 'Available test(s):' . PHP_EOL;
 
[...]

PR #3217, opened 2017-11-06
TokenizerLinter - fix lack of linting when code is cached

SpacePossum:
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3205

The issue seems that the linter doesn't check because of caching, i.e. it thinks the tokens are not changed, however these are.

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -1195,7 +1195,7 @@ class Tokens extends \SplFixedArray
      */
     private static function calculateCodeHash($code)
     {
-        return (string) crc32($code);
+        return CodeHasher::calculateCodeHash($code);
     }
 
     /**
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -159,7 +159,7 @@ class Tokens extends \SplFixedArray
      */
     public static function fromCode($code)
     {
-        $codeHash = self::calculateCodeHash($code);
+        $codeHash = CodeHasher::calculateCodeHash($code);
 
         if (self::hasCache($codeHash)) {
             $tokens = self::getCache($codeHash);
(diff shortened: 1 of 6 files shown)


keradus:
> it thinks the tokens are not changed, however these are

could you provide some more details?
why tokens are thinking there were not changed? and when we check it in context of linter?

keradus:
Damn, pretty nice investigation.

I let myself commit the changes. Please verify @SpacePossum

SpacePossum:
thanks @keradus , fast fixed, looking good :)
:+1:

src/Tokenizer/Tokens.php
```diff
@@ -1195,7 +1195,7 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
      */
     private static function calculateCodeHash($code)
```
julienfalque: What about dropping this private method altogether?

keradus: let us hotfix issue and decouple Tokens with a plan, not ad hoc

julienfalque: Not sure I understand your comment. The changes here make `Tokens::calculateCodeHash` useless as it's just a proxy to another static method, so we could call it directly.

SpacePossum: Ditched it and did some other minors.
I wonder though, could we somehow prevent that the `linters` need to know that they should clear the cache before checking `code`?  hmm...

GrahamCampbell:
Nice. 👍

src/Tokenizer/Tokens.php
keradus: I wanted to have single point of integration between Tokens and hasher.

yes, we can, we could extract whole caching out of Tokenizer and make caching decorator.

totally out of scope, even if nice to have

SpacePossum: agreed on all points :+1:  so let us put it on the roadmap :)
hope you are with the intermediate PR to deal with the current issue, we can deal with the rest later I think?

keradus: that's why I suggested few comments above to not try to refactor the code ...

SpacePossum: check, I'll revert, hold my beer

SpacePossum: done

keradus:
Thank you @SpacePossum.

PR #3222, opened 2017-11-07
Add the "opcache-only" option to the native function invocation fixer

### LINKED ISSUE #3048: Add a rule to root-namespace special compiled PHP functions

leofeyer:
This PR fixes #3048 by adding an "opcache-only" option to the native function invocation fixer. The option is set to `false` by default to maintain backwards compatibility.


src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -194,4 +220,55 @@ private function normalizeFunctionNames(array $functionNames)
...
+            'is_string',
+            'ord',
+            'strlen',
+            'strval',
+            'function_exists',
+            'is_callable',
+            'extension_loaded',
+            'dirname',
+            'constant',
+            'define',
```
localheinz: What do you think about keeping items in this list sorted - this makes it easier to a) find and b) decide where to add items?

leofeyer: Changed in 86c225c63c7ee80ee399574c6b4a895f2329f14a.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -136,6 +152,14 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 continue;
             }
 
+            // in_array() will only be optimized if the third argument is true
+            if ('in_array' === $lowerFunctionName && $this->configuration['opcache-only']) {
+                $end = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $next);
+                if ('true' !== $tokens[$tokens->getPrevMeaningfulToken($end)]->getContent()) {
```
localheinz: Not entirely sure if this makes sense, but what do you think about lower-casing the content, just in case someone does not use the corresponding fixer for native constants and decides to write code like this:

```php
if (in_array($needle, $haystack, TrUe) {

}
```

SpacePossum:
similar as prep'ed (but was not PR'ed yet) https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3223

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
leofeyer: Very good remark. Changed in 057b3dc30d33eb57371fdb0b04790173c5b80b2d.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -136,6 +152,14 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 continue;
             }
 
+            // in_array() will only be optimized if the third argument is true
```
nicolas-grekas: re-reading php-src, I think this can be removed, the optim occurs when the 3rd arg is missing, when it's `true` or `false` also.
Thanks for the PR BTW!

leofeyer: You are welcome. Changed in 3cfc483e79e92f3f6fa9dcfd7ac763bdda9fc682.

nicolas-grekas:
Would it be possible (and easy) to add a config option where we could list some constants that should be also prefixed? There are a few that would benefit from compile time resolution (eg DIRECTORY_SEPARATOR or PHP_VERSION_ID, which are the two that we use quite often and that OPcache could optimize by removing dead code)

leofeyer:
I think this is already been dealt with in #3127. Maybe @Slamdunk can add an `opcache-only` option as well?

Actually, the native constants fixer is new, so we should not add the option there but do it right from the beginning. 😄  The whole purpose of the `opcache-only` option is to remain backwards compatible, so we don't have to wait for PHP-CS-Fixer version 3 to correctly optimize native function calls.

nicolas-grekas:
#3127 looks good, I forgot about it :) no need for `opcache-only` there, as I'm not sure this would be well-defined at all :)

Slamdunk:
@nicolas-grekas I am open to make #3127 configurable, setting `opcache-only` by default. I only need someone to pinpoint me where to look at [php-src](https://github.com/php/php-src) for those special constants list.

nicolas-grekas:
> those special constants list

what I meant before: none are special

leofeyer:
@keradus @SpacePossum Is there anything I can do to get this PR merged?

The failing [Circle-CI tests](https://circleci.com/gh/FriendsOfPHP/PHP-CS-Fixer/1744?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) are not related to this PR:

```
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 69632 bytes) in /Users/distiller/PHP-CS-Fixer/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php on line 222
```

SpacePossum:
As https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3223 competes here and because of previous discussions on this topic in general my opinion/review could be biased, therefore I hope the other members pick this one up. That being said, thanks for all the work man, much appreciated :+1:

leofeyer:
Since this PR will only be merged in version 3, I have backported the optimization for PHP-CS-Fixer 2:

https://github.com/leofeyer/optimize-native-functions-fixer

julienfalque:
Maybe I'm missing something but why would this PR be merged in 3.x only?

leofeyer:
@keradus has commented [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3048#issuecomment-330176410) and he has targeted the PR against the master branch.

julienfalque:
IIUC, @keradus said that the *default* behavior of the fixer could be changed in version 3, but nothing prevents from adding new features in fixers of 2.x versions, as long as they are opt-in and don't change the default behavior. That means using an option with a default value that matches the current behavior of the fixer, which seems to be the case here :)

leofeyer:
@julienfalque That's how I look at it, too. 😄

SpacePossum:
This PR got auto closed because of the merge of another PR bringing similar functionality, so in the end this idea/concept and setup are much appreciated :)
Thanks for all work you put in this PR @leofeyer !

PR #3223, opened 2017-11-07
Master native function invocation fixer ns scope include sets

SpacePossum:


src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -24,10 +24,28 @@
...
+     * @internal
+     */
+    const SET_ALL = '@all';
+
+    /**
+     * Subset of SET_INTERNAL.
+     *
+     * @internal
+     */
+    const SET_COMPILER_OPT = '@compiler_opt';
```
ntzm: Any reason to use `opt` instead of `optimized`?

SpacePossum: not really, it is kind of a common abbreviation I think?

keradus: i got opt as optional :D

SpacePossum:
CS will fail, but that are two bugs in the current fixer
I think RTM here

SpacePossum:
is there anyone who can review or tell me how to proceed?

SpacePossum:
rebased and reworked, anyone?

SpacePossum:
well... no feedback for 3 months... merging ;P

README.rst
```diff
@@ -807,6 +807,12 @@ Choose from the list of available rules:
   Configuration options:
 
   - ``exclude`` (``array``): list of functions to ignore; defaults to ``[]``
+  - ``include``: list of function names or sets to fix. Defined sets are
+    ``@internal`` (all native functions), ``@all`` (all global functions) and
+    ``@compiler_optimized`` (functions that are specially optimized by Zend);
+    defaults to ``['@internal']``
+  - ``scope`` (``'global'``, ``'namespaced'``): fix function calls only if done in "user
```
kubawerlos: I'd replace `in  "user defined namespaces"` (as it may suggest there might be `not user defined namespaces` or something) with simply `inside a namespace`.

Also when I put "function calls done inside namespace" into google translate, translated it to some language, then back to English it came back with "made" instead of "done".

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -93,22 +150,89 @@ public function isRisky()
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        $functionNames = $this->getFunctionNames();
+        $functionFilter = $this->getFunctionFilter();
 
-        $indexes = [];
+        if ('namespaced' === $this->configuration['scope']) {
```
kubawerlos: We could prevent `foreach` inside of `if` with changing `'namespaced'` to `'global'` here.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -24,9 +24,27 @@
...
  */
 final class NativeFunctionInvocationFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    /**
+     * @internal
+     */
+    const SET_ALL = '@all';
+
+    /**
+     * Subset of SET_INTERNAL.
```
kubawerlos: What about adding a link to https://github.com/php/php-src/blob/master/Zend/zend_compile.c with the function name `zend_try_compile_special_func` to look at, or simply direct link https://github.com/php/php-src/blob/php-7.2.4/Zend/zend_compile.c#L3872? Someone more interested what are "these strange" `SET_COMPILER_OPTIMIZED` functions will probably check the fixer code first.

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -117,81 +241,192 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
     {
-        $definedFunctions = \get_defined_functions();
+        static $optFunctions = null;
 
-        return \array_diff(
-            $this->normalizeFunctionNames($definedFunctions['internal']),
-            \array_unique($this->normalizeFunctionNames($this->configuration['exclude']))
-        );
+        if (null === $optFunctions) {
+            $optFunctions = $this->normalizeFunctionNames([
```
kubawerlos: Re-checked a list on php-src - all good. Also 'opt' is here instead of 'optimized' as mentioned earlier.

README.rst
SpacePossum: :+1: updated

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: :+1: done

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: :+1:  added

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: renamed all the `opt` to `optimized`

SpacePossum:
>Do we want to have \ for functions in this fixer or not? I see 2 array_reverse calls - one with leading \ and other without.

Good catch!
The class is one of the few that has been fixed like this within the project, mainly because the fixer is not part the rules-set of the project. Fixed the CS again.

SpacePossum:
Travis fails because of SF update, @ see the fix here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3676

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: renamed

SpacePossum:
so how to get out of this stalemate?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -93,22 +154,91 @@ public function isRisky()
...
+                        }
+                    }
 
-            $tokenContent = $token->getContent();
+                    return true;
+                }])
+                ->setDefault([self::SET_INTERNAL])
+                ->getOption(),
+            (new FixerOptionBuilder('scope', 'Only fix function calls that are made within a namespace or fix globally.'))
+                ->setAllowedValues(['global', 'namespaced'])
```
keradus: I got a little confusion with `global` - there is sth like a `global namespace` in PHP, which is basically code out of any namespace.
Yet here, we are actually fixing everything.

Maybe rename it to sth less confusing, eg `all`, or change option to boolean `fix_only_namespaced: boolean = false` ?

I'm leaving decision up to you here, maybe it misleaded only me ;)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -93,22 +154,91 @@ public function isRisky()
...
+                    foreach ($value as $functionName) {
+                        if (!\is_string($functionName) || '' === \trim($functionName) || \trim($functionName) !== $functionName) {
+                            throw new InvalidOptionsException(\sprintf(
+                                'Each element must be a non-empty, trimmed string, got "%s" instead.',
+                                \is_object($functionName) ? \get_class($functionName) : \gettype($functionName)
+                            ));
+                        }
 
-        $indexes = [];
+                        static $sets = [
```
keradus: no need for static, php7 will optimize it anyway, and we don't need to store it in memory indefinitely

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -24,16 +24,38 @@
...
+    /**
+     * @internal
+     */
+    const SET_ALL = '@all';
+
+    /**
+     * Subset of SET_INTERNAL.
+     *
+     * Change function call to functions known to be optimized by the Zend engine.
+     * At see: https://github.com/php/php-src/blob/master/Zend/zend_compile.c "zend_try_compile_special_func"
```
keradus: `At see`?
`@see` ?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -24,16 +24,38 @@
...
+     * and https://github.com/php/php-src/blob/master/ext/opcache/Optimizer/pass1_5.c
```
keradus: please point directly to `sha`, as `master` will change in future and file can be moved
eg `https://github.com/php/php-src/blob/3a599dccb12395f9b9647dd5cd9f00322b9b644a/ext/opcache/Optimizer/pass1_5.c` (==php-7.2.6 tag)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: thanks, fixed :+1:

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: I'll it open for a bit of time to get more opinions on it (@julienfalque suggested this https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3054#discussion_r139291584 one) :)

keradus: `fix_only_namespaced` in fact is the same as `ignore_global_space` that Julien suggested ;)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -50,6 +50,15 @@
      */
     const SET_INTERNAL = '@internal';
 
+    private $functionFilter;
```
keradus: `string[]` ? (guessing...)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
```diff
@@ -24,16 +24,48 @@
...
+     * @internal
+     */
+    const SET_COMPILER_OPTIMIZED = '@compiler_optimized';
+
+    /**
+     * @internal
+     */
+    const SET_INTERNAL = '@internal';
+
+    private $functionFilter;
```
keradus: `string[]` ? (guessing...)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: I rather don't have a boolean value here. Having a `scope` option of type `string` allows us to extend on later if we want to define other scopes.
Renaming the current possible values `'global', 'namespaced'` is fine by me if someone has a better idea.

keradus: `'all', 'only_namespaced'` ?

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: `static function`, would that be `callable` ? (`gettype()` state `"object"`)

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
gharlan: > Maybe rename it to sth less confusing, eg `all`

`all` and `namespaced` 👍

SpacePossum: all done

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
keradus: yeah, callable ;) good I asked, I wouldn't guess.
for that, nice to document the input/output as well

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
keradus: big thanks

src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
SpacePossum: >nice to document the input/output as well

do you maybe have an example for that? not sure how to do this

leofeyer:
Is there any particular reason why you have made `@internal` the default? As @nicolas-grekas has explained in #3048, there is no point in prefixing all internal functions, so the default should be `@compiler_optimized` IMHO.

ntzm:
Setting it to `@compiler_optimized` would change the default behaviour of the fixer, which would be a BC break

leofeyer:
The change has been merged into the master branch, so it will be released in version 3 and new major versions can have breaking changes.

ntzm:
`master` is the branch for the next minor, there is a `3.0` branch for BC breaks

ntzm:
No problems!

Issue #2814, opened 2017-05-30
`NoUnusedImportsFixer` - Ignore references in comments


julienfalque:
`no_unused_imports` fixer searches for references to the imported elements almost everywhere in the code, including comments. This sometimes leads to not removing `use` statements that are unused because the same word was found in a comment, despite the comment being unrelated. See https://github.com/symfony/symfony/pull/22962#discussion_r119100703 for an example.

This check is done using a regular expression on the PHP code, which is the reason why the fixer can find false positives quite easily. IMO this is a bug and we should improve the regular expression.

But if we consider this a behavior change, then it would be nice to introduce a strict mode via a configuration option. When enabled, the fixer would only check usage of the imported elements in actual code and ignore comments completely (or look for PHPDocs/custom annotations only, e.g. `@param Foo` or `@Annotation()`).

SpacePossum:
so skip comments all together and analyze PHPDocs?

GrahamCampbell:
👍 Would be good to be able to turn on case sensitive only matching in phpdoc too.

julienfalque:
> so skip comments all together and analyze PHPDocs?

I would add an option that accepts multiple values to specify where the fixer should search for usages:
- `code`: actual PHP code usage of the imported elements;
- `phpdoc_types`: usages in PHPDoc types in annotations like `@param`, `@return` and so on;
- `phpdoc_descriptions`: references in PHPDoc summaries and descriptions;
- `annotations`: usages like `@Foo`, `@Bar\Baz`, ... in DocComments;
- `comments`: references in comments not covered by previous values.

By default it would have all these values to keep current behavior.

WDYT?

> Would be good to be able to turn on case sensitive only matching in phpdoc too.

Not sure about that. PHP itself isn't case sensitive so I think this fixer should remain case insensitive.

keradus:
:+1: for both options.
even if php doesn't give a shit about casing, there are plenty of devs who do. yet indeed, option disabled by default

GrahamCampbell:
> Not sure about that. PHP itself isn't case sensitive so I think this fixer should remain case insensitive.

> even if php doesn't give a shit about casing, there are plenty of devs who do. yet indeed, option disabled by default

The php autoloader is case sensitive if you are on a case sensitive filesystem though.

GrahamCampbell:
[...]

PR #3226, opened 2017-11-07
Remove unused imports

ntzm:
Picked up using PhpStorm's inspector. Obviously these should have been picked up by the `no_unused_imports ` fixer, but I'm not sure if this is desired behaviour?


julienfalque:
The current implementation of `no_unused_imports` is very defensive to avoid removing `use` statement that might be used in DocComments (e.g. Doctrine Annotation). This sometimes leads to `use` statements being kept while they could be removed. We have an open issue to improve this: #2814.

keradus:
Thank you @ntzm.

PR #3229, opened 2017-11-07
Fix 'Abstract' typo

ntzm:


SpacePossum:
:+1: 
Changing the copy while no tests fail indicates these are not covered by utests,
if you've the time could you update the tests maybe as well to cover this?

ntzm:
Putting all in one PR

PR #3231, opened 2017-11-08
Fix typos

ntzm:


tests/Fixer/AbstractFixerTest.php
```diff
@@ -0,0 +1,79 @@
...
+
+        $fixer->doSomethingWithCreateConfigDefinition();
+    }
+
+    protected function createFixer()
+    {
+    }
+}
+
+final class UnconfigurableFixer extends AbstractFixer
```
ntzm: Any better way of doing this?

julienfalque: This is not PSR-4 compliant, this class should be extracted to its own file.

SpacePossum: on 2.2 this looks good to me :+1: 
when we bump to PHP7 we could use an anonymous class as replacement for this test class

tests/FixerConfiguration/FixerConfigurationResolverTest.php
```diff
@@ -152,13 +152,9 @@ public function testResolveWithNormalizers()
             }),
         ));
 
-        $catched = null;
+        $this->expectException('Symfony\Component\OptionsResolver\Exception\InvalidOptionsException');
```
julienfalque: Reading the test again, using `assertSame()` was probably the best way actually. Sorry I misled you.

tests/Fixer/AbstractFixerTest.php
julienfalque: This is not PSR-4 compliant, this class should be extracted to its own file.

tests/FixerConfiguration/FixerConfigurationResolverTest.php
ntzm: No worries :)

tests/Fixer/AbstractFixerTest.php
ntzm: Which directory do you think it should go in?

keradus: tests / fixtures

ntzm: Done

tests/Fixer/AbstractFixerTest.php
```diff
@@ -0,0 +1,55 @@
...
+    public function testCreateConfigurationDefinitionUnconfigurable()
+    {
+        $fixer = new UnconfigurableFixer();
+
+        $this->setExpectedException('LogicException', 'Cannot create configuration definition using Abstract parent, child not implementing "PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface".');
+
+        $fixer->doSomethingWithCreateConfigDefinition();
+    }
+
+    protected function createFixer()
```
keradus: ?

ntzm: If I don't do this, I get a `UnexpectedValueException: Cannot determine fixer class, perhaps you forget to override `getFixerName` or `createFixerFactory` method?` because it's trying to find a fixer called `abstract`. I could make this test class just extend `TestCase` instead of `AbstractFixerTestCase`?

keradus: you don't use anything from `AbstractFixerTestCase`, no need for extending it here.

keradus:
Thank you @ntzm.

tests/Fixtures/Test/AbstractFixerTest/UnconfigurableFixer.php
```diff
@@ -0,0 +1,38 @@
+<?php
```
keradus: @ntzm 
this file is excluded from rules, as all fixtures are, but if not needed, don't use Windows line endings

ntzm: Sorry, forgot to convert it

tests/Fixer/AbstractFixerTest.php
```diff
@@ -0,0 +1,51 @@
...
+
+namespace PhpCsFixer\Tests\Fixer;
+
+use PhpCsFixer\Tests\Fixtures\Test\AbstractFixerTest\UnconfigurableFixer;
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\AbstractFixer
```
keradus: @ntzm, class does not exist, with newer PHPUnit it is crashing our builds (starting from 2.7 branch, on which we started using newer PHPUnit)
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/299588274

ntzm: Sorry, don't know why I didn't check that. I see you've fixed it. Thanks

PR #3234, opened 2017-11-08
Simplify Cache\Signature::equals

ntzm:
Makes it a lot more readable IMO


keradus:
+/- 0
looks better, works worse for coverage branches (as no branch is created for complex condition)

keradus:
Thank you @ntzm.

Issue #3232, opened 2017-11-08
Increment/Decrement fixer


ntzm:
```php
// from
$i += 1;
// to
++$i;

// from
$i = $i + 1;
// to
++$i;

// from
$i -= 1;
// to
--$i;

// from
$i = $i - 1;
// to
--$i;
```

keradus:
then I;m afraid one will want to have reverse fixer...

SpacePossum:
I like this one :)

KacerCZ:
@keradus  Fixer 'increment_style' already exists for this :)

keradus:
@KacerCZ  huh? `increment_style` is about post/pre increment, not about `+=` operator at all!

exactly as @SpacePossum said

Soullivaneuh:
And what about:

```php
// from
$i = $i + 42;
// to
$i += 42;
```

Or maybe it's the role of another fixer?

ntzm:
@Soullivaneuh
As discussed here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3235#issuecomment-349463668 and here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3235#issuecomment-361674308, I think the best course of action is two separate fixers

PR #3235, opened 2017-11-08
Implement StandardizeIncrementFixer

ntzm:


tests/Fixer/Operator/StandardizeIncrementFixerTest.php
```diff
@@ -0,0 +1,143 @@
...
+            [
+                '<?php for ($i = 0; $i < $n; ++$i) {};',
+                '<?php for ($i = 0; $i < $n; $i += 1) {};',
+            ],
+            [
+                '<?php foo_bar(++$i);',
+                '<?php foo_bar($i = $i + 1);',
+            ],
+            [
+                '<?php $foo->bar = $foo->bar + 1;',
```
ntzm: Any thoughts on this and the few below? If the `__set` or `__get` methods were implemented in a way that it would not set/get the same value then it would make this fixer risky

tests/Fixer/Operator/StandardizeIncrementFixerTest.php
```diff
@@ -0,0 +1,143 @@
...
+            ],
+            [
+                '<?php $foo->bar += 1;',
+            ],
+            [
+                '<?php $foo["bar"] = $foo["bar"] + 1;',
```
ntzm: Same as above but with `offsetSet` and `offsetExists`

tests/Fixer/Operator/StandardizeIncrementFixerTest.php
julienfalque: Seems to be fine: https://3v4l.org/HlOvJ.

ntzm: Yep, you're right. Dunno what I was thinking

ntzm: Can anyone give me any pointers on how I would detect if it should be shortened in cases like this?

`$foo->bar = $foo->bar + 1` should be `++$foo->bar`

`$foo->bar[baz()] = $foo->bar[baz()] + 1` should stay the same

`$foo->bar[baz()] += 1` should be `++$foo->bar[baz()]` because `baz` will only be called once in both circumstances

`$foo->bar[$baz->qux] = $foo->bar[$baz->qux] + 1` should stay the same because `qux` may changed

`$foo->bar[$baz->qux] += 1` should be `++$foo->bar[$baz->qux]` because `qux` will only be accessed once in both circumstances

src/Fixer/Operator/StandardizeIncrementFixer.php
```diff
@@ -0,0 +1,212 @@
...
+        $valueIndex = $tokens->getNextMeaningfulToken($equalsIndex);
+        $value = $tokens[$valueIndex];
+
+        if (!$value->isGivenKind(T_VARIABLE)) {
+            return;
+        }
+
+        $secondVariable = $this->getVariableTokensFrom($tokens, $valueIndex);
+
+        if ($secondVariable['tokens']->toJson() !== $variable->toJson()) {
```
ntzm: Need a better way of comparing two instances of `Tokens`

tests/Fixer/Operator/StandardizeIncrementFixerTest.php
```diff
@@ -0,0 +1,179 @@
...
+                '<?php --$i;',
+                '<?php $i -= 1;',
+            ],
+            [
+                '<?php --$i;',
+                '<?php $i-=1;',
+            ],
+            [
+                '<?php ++$i;',
+                '<?php $i /* foo */ = $i /* bar */ + 1;',
```
ntzm: Is this the desired behaviour?

SpacePossum: please don't throw away comments, no matter how badly placed, leave them somewhere near their original place.

src/Fixer/Operator/StandardizeIncrementFixer.php
```diff
@@ -0,0 +1,212 @@
...
+            $variable = $this->getVariableTokensFrom($tokens, $index);
+
+            $nextIndex = $tokens->getNextMeaningfulToken($variable['last_index']);
+            $next = $tokens[$nextIndex];
+
+            $insert = null;
+
+            if ($next->isGivenKind([T_PLUS_EQUAL, T_MINUS_EQUAL])) {
+                $insert = $this->fixPlusMinusEqual($tokens, $nextIndex);
+            } elseif ('=' === $next->getContent()) {
```
keradus: ->equals,
don't use ->getContent for comparision

ntzm:
Struggling to find the time to fix the test cases. Should I close this PR and re-open when I've found the time to fix?

julienfalque:
If you still plan to finish it (and/or want someone to help), you can leave it open. It would be easy to forget about it if it was closed :)

ntzm:
I do plan on finishing this at some point! Hopefully early December because I've got some time off work. I'll keep it open!

Soullivaneuh:
See also https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3232#issuecomment-349356164

tests/Fixer/Operator/StandardizeIncrementFixerTest.php
keradus: any news on that ?

keradus:
we have very similar logic already here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3235/files#diff-ca612f63029ba7b7c2e1b238fdd2cda9R98

```php
[
    '<?php ++$foo["bar"];',
    '<?php $foo["bar"] = $foo["bar"] + 1;',
]
```

What about splitting current fixer into 2:
1st fixer:
```diff
-<?php $foo["bar"] = $foo["bar"] + $x;
+<?php $foo["bar"] += $x;
```

2nd fixer
```diff
-<?php $foo["bar"] += 1;
+<?php ++$foo["bar"];
```

?

ntzm:
Ok I've finally got some time to spend on this, I think it's a good idea to split into two fixers. I will implement the `+= 1` fixer first

SpacePossum:
Thanks @ntzm for working on this again :+1:
Splitting the functionality over two fixers makes sense to me, it will also give focus to this PR to get one thing done first, good call!

src/Fixer/Operator/StandardizeIncrementFixer.php
```diff
@@ -0,0 +1,111 @@
...
+            }
+
+            while ($tokens[$tokens->getPrevMeaningfulToken($nextIndex)]->equals('$')) {
+                $nextIndex = $tokens->getPrevMeaningfulToken($nextIndex);
+            }
+
+            if ($tokens[$tokens->getPrevMeaningfulToken($nextIndex)]->isGivenKind(T_OBJECT_OPERATOR)) {
+                $nextIndex = $tokens->getPrevMeaningfulToken($nextIndex);
+
+                goto loop;
```
ntzm: this is temporary

src/Fixer/Operator/StandardizeIncrementFixer.php
```diff
@@ -0,0 +1,138 @@
...
+    const EXPRESSION_END_TOKENS = [
+        ';',
+        ')',
+        ']',
+        ',',
+        ':',
+        [CT::T_DYNAMIC_PROP_BRACE_CLOSE],
+        [CT::T_DYNAMIC_VAR_BRACE_CLOSE],
+        [T_CLOSE_TAG],
+    ];
```
ntzm: Any more here?

ntzm:
This is no longer WIP

tests/Fixer/Operator/StandardizeIncrementFixerTest.php
```diff
@@ -0,0 +1,469 @@
...
+                '<?php $i -= 1 || true;',
+            ],
+            [
+                '<?php $i -= 1 and true;',
+            ],
+            [
+                '<?php $i -= 1 or true;',
+            ],
+            [
+                '<?php $i -= 1 xor true;',
```
julienfalque: Maybe good to add similar cases with `==`  and `===`?

tests/Fixtures/Integration/priority/increment_style,standardize_increment.test
```diff
@@ -0,0 +1,11 @@
+--TEST--
```
julienfalque: Please add a test case to [`FixerFactoryTest::testFixersPriority()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.10.0/tests/AutoReview/FixerFactoryTest.php#L60).

ntzm: Ah didn't realise that had to be added! Would be helpful to have that in the fixer cookbook

julienfalque:
👍 LGTM, just a minor thing to fix: the `<=>` operator does not exist in PHP 5.6 so the related test cases must be skipped on that version.

tests/Fixtures/Integration/priority/increment_style,standardize_increment.test
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3518

keradus:
Thank you  @SpacePossum as well

SpacePossum:
thanks @ntzm , great new fixer :D
:dancer:

PR #3237, opened 2017-11-09
UnconfigurableFixer - use only LF

keradus:


julienfalque:
We don't have a fixer for that?

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3231#discussion_r149936553

Issue #3240, opened 2017-11-09
NoExtraConsecutiveBlankLinesFixer crashes on bad whitespace characters


GrahamCampbell:
$ php -v

`7.1.11`

$ php-cs-fixer -V

`2.8.0`

The command you use to run the fixer (and the configuration file you are using if any).

`NoExtraConsecutiveBlankLinesFixer`

A minimal sample of valid PHP code that is not fixed correctly (if applicable).

https://github.com/ByTIC/Common/blob/8e7382a9e9f24e9e3d77abea3e8d03a0c6229e47/src/Payments/Gateways/Providers/Euplatesc/Gateway.php

Was hard to edit the file because editors and git settings kept fixing the characters. The original file demonstrates the issue just fine though.

```
 InvalidArgumentException · Cannot set empty content for id-based Token.
```

Happens as a result of calling `$this->tokens[$index] = new Token([T_WHITESPACE, $content]);`.

SpacePossum:
thanks for reporting,
this fails on 2.7 as well

PR #3241, opened 2017-11-10
NoExtraConsecutiveBlankLinesFixer - do not crash on ^M LF only

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3240


julienfalque:
Does that mean we officially support `\r` line endings now?

keradus:
no

but I agree it's nice to not crash

keradus:
Thank you @SpacePossum.

PR #3242, opened 2017-11-11
PhpUnitNoExpectationAnnotationFixer - fix ' handling

keradus:
gladly, we don't crash the code currently, post-fixing linter catches the issue and don't save the file


GrahamCampbell:
🎉

PR #3246, opened 2017-11-11
Make ToolInfo methods non-static

julienfalque:
Extracted from #3195.

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -105,6 +105,11 @@ final class ConfigurationResolver
     private $configFinderIsOverridden;
 
     /**
+     * @var ToolInfo
+     */
+    private $toolInfo;
+
+    /**
      * @var array
      */
     private $options = array(
@@ -150,10 +155,12 @@ final class ConfigurationResolver
     public function __construct(
         ConfigInterface $config,
         array $options,
-        $cwd
+        $cwd,
+        ToolInfo $toolInfo
     ) {
         $this->cwd = $cwd;
         $this->defaultConfig = $config;
+        $this->toolInfo = $toolInfo;
 
         foreach ($options as $name => $value) {
             $this->setOption($name, $value);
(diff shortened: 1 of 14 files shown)


src/Console/ConfigurationResolver.php
```diff
@@ -150,10 +155,12 @@
     public function __construct(
         ConfigInterface $config,
         array $options,
-        $cwd
+        $cwd,
+        ToolInfo $toolInfo
```
keradus: add missing `@param` for constructor

src/ToolInfo.php
```diff
@@ -27,73 +27,62 @@
 
     const COMPOSER_LEGACY_PACKAGE_NAME = 'fabpot/php-cs-fixer';
 
-    public static function getComposerInstallationDetails()
-    {
-        static $result;
+    private $composerInstallationDetails;
+
+    private $isInstalledByComposer;
```
keradus: typehints are missing

src/Console/ConfigurationResolver.php
julienfalque: This is conflicting with the config of the project. Shall I ajust `phpdoc_add_missing_param_annotation` to set `only_untyped` to `false`?

keradus: is it conflicting? `$config` is in the docblock already

julienfalque: You asked me to add a missing _typed_ `@param` while in our config we don't want to automatically add those.

keradus: then I see conflict, as `@param $toolInfo` was not required, and `@param $config` is already there...

keradus:
In PR that is origin for this PR, we actually have `ToolInfoInterface`, we shall have it here as well IMO

keradus:
all the rest looks really nice :+1:

keradus:
let m e begin the merging hell...

keradus:
Thank you @julienfalque.

PR #3248, opened 2017-11-11
Added failing test for align equals inside callback

GrahamCampbell:
On PHP-CS-Fixer 2.8.1, with only the `binary_operator_spaces` fixer with config:

```php
['operators' => ['=' => 'align_single_space_minimal']]
```

---

Input:

```php
<?php

f(function () {
    $fooo = 123;
    $ba = 123;
});
```

---

Actual:

No change.

---

Expected:

```php
<?php

f(function () {
    $fooo = 123;
    $ba   = 123;
});
```

---

This bug is not present if we just have:

```php
<?php

$fooo = 123;
$ba = 123;
```

It gets correctly fixed to:

```php
<?php

$fooo = 123;
$ba   = 123;
```


keradus:
:+1: for catch. please send a fix.

keradus:
@GrahamCampbell 
As you see, sadly we don't have enough menpower to handle this issue. Could you please help solving it ?

GrahamCampbell:
I don't have any time either I'm afraid. If I do find any, I'll take a look, but otherwise, there's nothing more that I can do here.

SpacePossum:
work continues here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3762

PR #3250, opened 2017-11-13
Travis: fail early, spare resources, save the Earth

Slamdunk:
Hi, before digging into the solution, here's the problem:

1. a lot of PRs fail at first
1. 99% PRs fail in all PHP versions
1. Travis is free but the Earth is not, we shouldn't waste resources
1. Waiting Travis for 20 minutes is boring

Like the "*Static Code Analysis*" stage, I would like to skip PHP 5.6 slowness if the failing test if failing for all.

So here you are: I moved the PHP version without specific envs to a new, second stage.

WDYT?

- [x] Split fast and slow tests
- [ ] Remove example failing test before PR merge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,7 +42,7 @@ jobs:
 
 
         - &STANDARD_TEST_JOB
-            stage: Test
+            stage: Fast Test
             php: 7.0
             install:
                 # Composer: enforce given Symfony components version
@@ -57,32 +57,38 @@ jobs:
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 5.3
             env: SKIP_LINT_TEST_CASES=1 COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
             dist: precise
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 5.4
             env: SKIP_LINT_TEST_CASES=1
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 5.5
             env: SKIP_LINT_TEST_CASES=1
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 5.6
             env: SYMFONY_VERSION="^2.8"
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 7.1
             env: PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="^4.0" MIN_STABILITY=dev
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: 7.2
             env: COLLECT_COVERAGE=1 SYMFONY_DEPRECATIONS_HELPER=weak
             before_script:
@@ -99,6 +105,7 @@ jobs:
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: nightly
             env: COMPOSER_FLAGS="--ignore-platform-reqs" PHP_CS_FIXER_IGNORE_ENV=1 SYMFONY_DEPRECATIONS_HELPER=weak
             script:
@@ -109,6 +116,7 @@ jobs:
 
         -
             <<: *STANDARD_TEST_JOB
+            stage: Test
             php: hhvm
             env: SKIP_LINT_TEST_CASES=1
             sudo: required
--- a/tests/UtilsTest.php
+++ b/tests/UtilsTest.php
@@ -27,6 +27,11 @@ use PHPUnit\Framework\TestCase;
  */
 final class UtilsTest extends TestCase
 {
+    public function testFastTest()
+    {
+        $this->assertFalse(true);
+    }
+
     /**
      * @param string $expected Camel case string
      * @param string $input    Input string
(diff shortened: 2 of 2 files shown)


keradus:
after such a lovely entry, is it some kind of wider initiative?
best approach would be actually to have Travis feature to stop other jobs if one failed.

.travis.yml
```diff
@@ -45,9 +45,20 @@ jobs:
...
+                - composer info -D | sort
+            script:
+                - vendor/bin/phpunit || travis_terminate 1
+                - PHP_CS_FIXER_FUTURE_MODE=1 php php-cs-fixer --diff --dry-run -v fix
+
         - &STANDARD_TEST_JOB
             stage: Test
-            php: 7.0
+            php: 5.6
+            env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
```
julienfalque: Now all entries that extend `STANDARD_TEST_JOB` will have this environment variable by default.

.travis.yml
```diff
@@ -45,9 +45,20 @@ jobs:
                 - ./dev-tools/vendor/bin/composer-require-checker check composer.json --config-file=.composer-require-checker.json
 
 
+        -
+            stage: Fast Test
```
julienfalque: Why not reuse `STANDARD_TEST_JOB` and override `stage` only?

keradus:
@julienfalque raised exactly the comments I would like to raise myself. after done his suggestions, you could add order of stages if needed.

Julien, do you like the idea behind this change in first place ? (regardless of implementation)

julienfalque:
I totally support the goal of this PR but I'm not sure about the way to do it.

> best approach would be actually to have Travis feature to stop other jobs if one failed.

Not if all jobs fail for different reasons. You would have to fix the first failure and trigger a new build, fix the second failure, and so on until all jobs pass. Not sure that would actually reduce overall build time.

Currently this PR has the same issue: if 7.0 job fails, you have no idea whether other jobs would pass.

Slamdunk:
> Not if all jobs fail for different reasons. You would have to fix the first failure and trigger a new build, fix the second failure, and so on until all jobs pass. Not sure that would actually reduce overall build time.

Yup, Travis shouldn't implement it for this reason. The user have to use the stages wisely.

> Currently this PR has the same issue: if 7.0 job fails, you have no idea whether other jobs would pass.

This is exactly the purpose of the PR. First run the 1 minute test, only then the 15 minutes one.

.travis.yml
Slamdunk: I tried in the first commit:
1. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3250/commits/e565c1b2993c0dcb174a214071bc28049c086513
1. https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/301347812?utm_source=github_status&utm_medium=notification

.travis.yml
Slamdunk: Heh, sorry. Definitely need help for the final version of this PR.

Slamdunk: Btw what you say doesn't seem to be true:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/301348417/config

julienfalque: This is because currently all entries that extend `STANDARD_TEST_JOB` override `env`. But if we have an entry without `env` in the future we might not notice that it uses this default.

What do you need help for exactly?

keradus:
we have multiple jobs, only 5.6 is taking 15 minutes. yet, you extracted one of 2 minutes jobs, not the 15 minutes one

Slamdunk:
> we have multiple jobs, only 5.6 is taking 15 minutes. yet, you extracted one of 2 minutes jobs, not the 15 minutes one

That was my intention, 1 fast test first and then the others. Would you prefer other tests first, then 1 slow test?

My choise is driver by the fact that version specific or depdencency specific fails are uncommon, so 1 fast test first should be enough to save the maximum amount of time and resources retaining a statistically appectable feedback.

.travis.yml
julienfalque: Avoiding copy-pasting in YAML can be achieved using references. This is already done with `&STANDARD_TEST_JOB` label and `*STANDARD_TEST_JOB` references in this file.

Out of curiosity: what is wrong with YAML?

Slamdunk: Source: https://twitter.com/search?q=%40ocramius%20%23yaml&src=typd

1. https://github.com/crdoconnor/strictyaml/blob/master/FAQ.rst#what-is-wrong-with-implicit-typing
1. https://news.ycombinator.com/item?id=12423733

I'll try soon another anchor

Slamdunk:
@julienfalque I've reused the default .travis.yml, had to copy-paste the `stage` of course. Any other suggestions?

julienfalque:
Looks better, I'm fine with duplicated `stage`. By the way, I'd say this should be done on 2.2 branch.

julienfalque:
Looks like changing target branch messed up CI builds, could you push your code again to trigger new ones (how ironic 😕)?

Slamdunk:
Done.

> Travis: Ran for 2 min 58 sec

Love it.

julienfalque:
So what do we do now? Shall we merge it a see how it goes for some time?

keradus:
Thank you @Slamdunk.

PR #3251, opened 2017-11-13
README.rst - Create title for config file

IanEdington:
I spent an unusually long time looking for this section of the docs.

The phrase `config file` is used throughout the docs but a section describing the config file is not present. This will make it easier for people to find the docs regarding config files.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1122,6 +1122,9 @@ fixed but without actually modifying them:
 
     $ php php-cs-fixer.phar fix /path/to/code --dry-run
 
+Config file
+-----------
+
 Instead of using command line options to customize the rule, you can save the
 project configuration in a ``.php_cs.dist`` file in the root directory of your project.
 The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.10/src/ConfigInterface.php>`_
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -120,6 +120,9 @@ fixed but without actually modifying them:
 
     <info>$ php %command.full_name% /path/to/code --dry-run</info>
 
+Config file
+-----------
+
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -1396,6 +1396,9 @@ fixed but without actually modifying them:
 
     $ php php-cs-fixer.phar fix /path/to/code --dry-run
 
+Config File
```
keradus: file shall be lowercased

keradus:
Hi !
This is a nice change indeed :+1:

Please, change target branch for 2.2.
Also, as we kindly ask you in contributing guidelines, don't manually modify the readme file. It's autogenerated from code and you shall change code itself, then autogenerate the readme file.
I expect CIs to fail due to that.

Slamdunk:
@IanEdington the right place to apply the edit is here:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/e5c2abb7addf85bfc241d880d516c66c4203d7f2/src/Console/Command/HelpCommand.php#L116-L125

README.rst
julienfalque: Actually I'm not sure: https://en.wikipedia.org/wiki/Capitalization#Titles

keradus: it's not a title, it's header.
even in link you provided, there are `By context`, `By name of style`, `Sentence case` and so

IanEdington:
@keradus @Slamdunk @julienfalque Wow best experience on a commit I've had in ages!! Timely, polite, and with good feedback. I will make that change. 🥇

keradus:
Sadly, you still need to regenerate the readme and commit it. We found no way to not have readme file included in repo and yet visible in github repo page.
please, run `php php-cs-fixer readme > README.rst` and commit the changes :+1:

keradus:
Thank you @IanEdington.

README.rst
julienfalque: I thought this should be applied to headers as well, ok then :+1:

PR #3253, opened 2017-11-14
SimplifiedIfReturnFixer - introduction

Slamdunk:
```diff
-if ($foo) { return true; } return false;
+return (bool) ($foo);
```
- [x] Add SimplifiedIfReturnFixer
- [x] ~Run after [BracesFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.1/src/Fixer/Basic/BracesFixer.php)~ ~BracesFixer has the lowest priority, can't do this. The user must run PHP-CS-Fixer twice~ Fix cases without braces too, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3253#issuecomment-556971775
- [x] Support negative statements like `if ($foo) { return false; } return true;`, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3253#pullrequestreview-94819359
- [x] Run after [NoSuperfluousElseifFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.1/src/Fixer/ControlStructure/NoSuperfluousElseifFixer.php)
- [x] Run after [NoUselessElse](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.1/src/Fixer/ControlStructure/NoUselessElseFixer.php)
- [x] Run before [NoMultilineWhitespaceBeforeSemicolonsFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.1/src/Fixer/Semicolon/NoMultilineWhitespaceBeforeSemicolonsFixer.php)
- [x] Run before [NoSinglelineWhitespaceBeforeSemicolonsFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.1/src/Fixer/Semicolon/NoSinglelineWhitespaceBeforeSemicolonsFixer.php)


src/Fixer/ControlStructure/SimplifiedIfReturnFixer.php
```diff
@@ -0,0 +1,50 @@
...
+ */
+final class SimplifiedIfReturnFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Simplify if conditions that return the result of the condition itself.',
```
Slamdunk: > There are only two hard things in Computer Science: cache invalidation and naming things.
>     - Phil Karlton

Soullivaneuh:
What if my PHP code looks like this?

```php
$foo = 'This is it.';

if ($foo) {
    return true;
}

return false;
```

The bool return type will be a string with this fixer.

Slamdunk:
I'm intended to fix it like:
```php
return !!($foo);
```

ntzm:
I think most would prefer `return (bool) $foo`?

Slamdunk:
I tought some type couldn't be explicitly converted to boolean, but after research it seems all can, I'll go with the `(bool)`

keradus:
> BracesFixer has the lowest priority, can't do this. The user must run PHP-CS-Fixer twice

wow wow wow, what's going on here? it's not an option!

Slamdunk:
@keradus consider this code:
```php
if ($foo)
   return true;
return false;
```
To be fixed in:
```php
return (bool) $foo;
```
And the following fixers enabled:

* BracesFixer (priority: -25)
* SimplifiedIfReturnFixer (current priority: 1)
* NoMultilineWhitespaceBeforeSemicolonsFixer (priority: 0)

Since NoMultilineWhitespaceBeforeSemicolonsFixer runs before BracesFixer, and since SimplifiedIfReturnFixer must run NoMultilineWhitespaceBeforeSemicolonsFixer, how can you fix the code in one step only?

src/Fixer/ControlStructure/SimplifiedIfReturnFixer.php
julienfalque: What about:
```diff
-Simplify if conditions that return the result of the condition itself.
+Simplify `if` control structures that return the boolean result of their condition.
```
?

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
```diff
@@ -0,0 +1,130 @@
...
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
```
julienfalque: I would add test cases with `else if` and `elseif`.

tests/Fixtures/Integration/priority/no_superfluous_elseif,simplified_if_return.test
```diff
@@ -0,0 +1,22 @@
+--TEST--
```
julienfalque: If the fixer supports fixing conditions with `else` or `elseif`, I think there is no priority issue with the `no_useless_else` and `no_superfluous_elseif` fixers.

Slamdunk: * `no_superfluous_elseif`: you are right, but I would prefer to keep the integration test to explicitly tell what I expect from the two combined
* `no_useless_else`: it would be much more complex to handle in this fixer the `if`+`else` condition rather than let `no_useless_else` simplify it first; morover, as for `no_superfluous_elseif`, I would prefer to keep the integration test to explicitly tell what I expect from the two combined

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
Slamdunk: Done

keradus:
just a note:
> The user must run PHP-CS-Fixer twice

for me, this is a blocker for this PR.
we need to figure out how to deal with it.

tests/Fixtures/Integration/priority/no_superfluous_elseif,simplified_if_return.test
keradus: if there is no priority issue but we simply want to ensure both of them are playing nicely together (regardless of executing order), those tests shall be moved from `priority` to `misc` folder

keradus:
> how can you fix the code in one step only?

please, start with putting the issue down to integration test, we would know what is expected output (regardless of amount of executions), and then we can adjust the code to pass in single execution

julienfalque:
@Slamdunk this PR stalled, any chance you can work on it? Do you need/want some help?

Slamdunk:
Actually only the integration tests against `BraceFixer` are missing: re-reading this PR now I would go for fixing cases without braces too, so we don't have to integrate with `BraceFixer` at all.

I would love to finish the work, but due to family commitments I cannot guarantee any date soon.

Btw I'll rebase this branch now so if you want you can push a PR against my fork without opening a new PR :+1:

Slamdunk:
@julienfalque PR rebased, first-post task list updated: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3253#issue-152441735

Slamdunk:
Ok, everything done.

The fixer now is independent from `BraceFixer` and one run is enough for the final form.

A side note: the parenthesis of the `if` statement are preserved because:

1. There already is a separate fixer for those who don't want them, [NoUnneededControlParenthesesFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.1/src/Fixer/ControlStructure/NoUnneededControlParenthesesFixer.php)
1. `if ($foo && $bar) return true; return false;` fixed without parenthesis would result in `return (bool) $foo && $bar;` and so the `(bool)` typecast would be applied to the first argument instead the whole sentence; it would be harmless but not very clear
1. It allows us to fix the negative sentences with the `!` exclamation mark and without the `(bool)` typecast

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
```diff
@@ -0,0 +1,250 @@
...
+            ],
+            'simple-braceless' => [
+                '<?php return (bool) ($foo)    ;',
+                '<?php if ($foo) return true; return false;',
+            ],
+            'simple-braceless-negative' => [
+                '<?php return ! ($foo)    ;',
+                '<?php if ($foo) return false; return true;',
+            ],
+            'bug1' => [
```
julienfalque: What's "bug1"?

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
```diff
@@ -0,0 +1,250 @@
...
+// C11
+false
+# C12
+;
+/* C13 */
+EOT
+                ,
+            ],
+            'else-if' => [
+                '<?php if ($bar) { return $bar; } else return (bool) ($foo)      ;',
```
julienfalque: `else` can be dropped (as done with `elseif`).

tests/Fixtures/Integration/priority/no_superfluous_elseif,simplified_if_return.test
julienfalque: The priority dependency with `no_useless_else` and `no_superfluous_elseif` still exists, right? IMO it's fine for now. If we want to make the fixer more autonomous, that can happen later.

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
Slamdunk: I don't remember: I do remember this test case exposed a bug I found in an earlier implementation, and yesterday I added array keys to the data provide as a refactoring practise

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
Slamdunk: If this context, `elseif` and `else if` are slightly different: `elseif` is findable and handable like `if`; `else if` instead can also be present as `else { if` and therefor be more complex to handle correctly. Since integration with `NoUselessElse` is already granted, I don't see the perks of adding complexity to the fixer in order to handle this case.

julienfalque: I understand, [maybe later](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3253#discussion_r352009528) then.

tests/Fixer/ControlStructure/SimplifiedIfReturnFixerTest.php
julienfalque: Can we change the name to something more useful?

Slamdunk: Of course, but I'm not able to think any meaningful phrase, as I don't know what the test case is testing :\

julienfalque: Maybe `consecutive_ifs`? Looks like those are the only test cases covering this scenario.

Slamdunk: Done, thank you for the suggestion :+1:

src/Fixer/ControlStructure/SimplifiedIfReturnFixer.php
```diff
@@ -0,0 +1,134 @@
...
+    {
+        return $tokens->isAllTokenKindsFound([T_IF, T_RETURN, T_STRING]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $foundIndexes = $tokens->findGivenKind([T_IF, T_ELSEIF]);
```
SpacePossum: Currently the fixer scans all tokens for all `T_IF` / `T_ELSEIF` and than iterates the reverse of that. For each match it search for the end of the body of that if/else match and than goes back to the start of the body and scans again within the body up to four times for the sequences. 

Maybe we can start at the end of the tokens and iterate back using int based indexes and than check for `T_IF` / `T_ELSEIF` ? And than if there is a match check if there is a candidate with early detection there isn't any?

Slamdunk: I don't get your last sentence :confused: 
Yes performance could be improved, but I'd prefer much more to let `Tokens` class provide performing operations and keep fixers' code readable and understandable.

SpacePossum: Nearly all fixers just iterate over the tokens and check for whatever kind of token they are looking for, nearly none scan all the tokens first and than go fixing. So for readable and understandable, I think it best to not introduce another method.

Slamdunk: I'm sorry, so is `findGivenKind` call the performance degradator you are talking about? I thought is is the least of the problems...

SpacePossum: feel free to not look into it, I'm just testing the code and share my findings, it is your PR of course

Slamdunk: I admit I lost myself in what you are asking me. I want to be clear that it was never my intention to go against this library conventions or best practices.
So, given that I didn't understand much, I'll change the `findGivenKind` to a simple loop through the tokens; any other change request will be applied as well, as long as I understand them :stuck_out_tongue_closed_eyes:

Slamdunk:
Thank you @SpacePossum

SpacePossum:
Thank you @Slamdunk.

PR #3260, opened 2017-11-15
Add DateTimeImmutableFixer

kubawerlos:
Ready for code review.

### DIFF
--- a/tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
+++ b/tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
@@ -1,0 +1,151 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\ClassUsage;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\ClassUsage\DateTimeImmutableFixer
+ */
+final class DateTimeImmutableFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
+            [
+                '<?php new DateTimeImmutable();',
+                '<?php new DateTime();',
+            ],
+            [
+                '<?php new DateTimeImmutable();',
+                '<?php new DATETIME();',
+            ],
+            [
+                '<?php new \DateTimeImmutable();',
+                '<?php new \DateTime();',
+            ],
+            [
+                '<?php new Foo\DateTime();',
+            ],
+            [
+                '<?php namespace Foo; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; new \DateTimeImmutable();',
+                '<?php namespace Foo; new \DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime; new \DateTimeImmutable();',
+                '<?php namespace Foo; use DateTime; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime; new Bar\DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime\Bar; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use Bar\DateTime; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime\Bar; use DateTime; use Baz\DateTime as BazDateTime; new \DateTimeImmutable();',
+                '<?php namespace Foo; use DateTime\Bar; use DateTime; use Baz\DateTime as BazDateTime; new DateTime();',
+            ],
+            [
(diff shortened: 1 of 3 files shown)


Slamdunk:
:+1: for immutability conversion
:-1: for timezone one

src/Fixer/FunctionNotation/DateTimeFixer.php
```diff
@@ -0,0 +1,85 @@
...
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\FunctionNotation;
```
localheinz: Maybe `ClassUsage` makes more sense than `FunctionNotation`?

keradus: let us first discuss about idea itself, not implementation details ;)

kubawerlos:
@Slamdunk timezone at all or the default value? It won't be UTC. but `date_default_timezone_get` and I'd use this rule in project where users can choose their time zone to make sure nowhere in code I forgot to use it.

Slamdunk:
Immutability can easily be forgotten and is an extremely valuable good practice to enforce.

Timezone instead for a project can be easily set up in a config once-for-all, and not having it written in each call reduce the verbosity and ease the find of pieces of code where  timezone must differs from default one. Moreover, what if I decide to change it after the first run? Would you change them all? If yes, this can easily alter the app domain logic without being noticed in a continuous environment. Too deep change for a coding style tool like this one.

Slamdunk:
@kubawerlos TL,DR: both timezone at all and the default value.

keradus:
@Slamdunk , I must admit I would not rely on globaly set timezone.
I experienced situation, where project was using one timezone global config, and team managing a dependency decided, that it's easier to them to adjust global timezone for their needs, and as it is a global state, those two requirements were conflicting. Outcome? One part of app was working incorrectly (providing end-user wrong times), and it took a while to figure out that it is a stupid dependency which is overriding default timezone.

For me, it's better to rely on explicit timezone parameter instead of implicit timezone config, that could be altered during runtime.

> Too deep change for a coding style tool like this one.

actually, same could be told for fourth of our rules, like upgrading PHP or a library, enforcing strict comparisons, declaring strict types and so...
or, you upgrade `coding style` definition to not only mention whitespaces...

Slamdunk:
What I dislike is explicit string timezone.

I can accept something like
```php
new DateTimeImmutable('now', MyApp::defaultTimezone());
```
But a low level string like `'UTC'`, once written, becames hard to track, change and maintain.

keradus:
totally agree with that, it shall not be a magic value.
yet, if you forgot to put your own constant there, fixer has to inject "sth".

Slamdunk:
I raise another concern. Regarding the timezone parameter, the manual says:

```
Note:
The $timezone parameter and the current timezone are ignored when the $time parameter either is a UNIX timestamp (e.g. @946684800) or specifies a timezone (e.g. 2010-01-28T15:00:00+02:00).
```
So it would be confusing to have a mandatory parameter that is then ignored.

I agree to the problem, but still I don't see PHP-CS-Fixer as the appropriate tool to solve it.

kubawerlos:
In such case fixer won't  add the timezone parameter - it will be added only when omission might result with problem. Calling `new \DateTime()` is always kind of magic - there is timezone of PHP and probably timezone of database connection to consider.

Slamdunk:
> it will be added only when omission might result with problem

And how are you going to detect this? I can use variables or whatever complex code for first argument.

tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
```diff
@@ -0,0 +1,151 @@
...
+            [
+                '<?php namespace Foo; use Bar\DateTime; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime\Bar; use DateTime; use Baz\DateTime as BazDateTime; new \DateTimeImmutable();',
+                '<?php namespace Foo; use DateTime\Bar; use DateTime; use Baz\DateTime as BazDateTime; new DateTime();',
+            ],
+            [
+                '<?php $foo = DateTimeImmutable::ISO8601;',
+                '<?php $foo = DateTime::ISO8601;',
```
julienfalque: Not sure this use case is relevant. IMO this should be ignored or optional.

tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
```diff
@@ -0,0 +1,151 @@
...
+                '<?php namespace Foo; use DateTime; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime; new Bar\DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use DateTime\Bar; new DateTime();',
+            ],
+            [
+                '<?php namespace Foo; use Bar\DateTime; new DateTime();',
```
julienfalque: Please add a similar test case for `date_create*` function usage: if function is imported from e.g. namespace `Foo` instead of global scope, the code should not be changed.

tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
kubawerlos: Done.

tests/Fixer/ClassUsage/DateTimeImmutableFixerTest.php
julienfalque: Ok, then maybe we could merge this PR as is and open a new one later to actually support this case in all related fixers? Introducing some `ImportsAnalyzer` looks like a good idea.

/cc @FriendsOfPHP/php-cs-fixer WDYT?

keradus:
Thank you @kubawerlos.

Slamdunk:
To promote the usage of this fixer and the transition from `DateTime` to `DateTimeImmutable`, this PHPStan rule can help: https://github.com/Slamdunk/phpstan-extensions/pull/10

PR #3263, opened 2017-11-16
NoBreakCommentFixer - Fix handling comment text with PCRE characters

julienfalque:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155#discussion_r151517076.

### DIFF
--- a/src/Fixer/ControlStructure/NoBreakCommentFixer.php
+++ b/src/Fixer/ControlStructure/NoBreakCommentFixer.php
@@ -166,7 +166,7 @@ switch ($foo) {
             return false;
         }
 
-        $text = $this->configuration['comment_text'];
+        $text = preg_quote($this->configuration['comment_text'], '~');
 
         return preg_match("~^((//|#)\s*$text\s*)|(/\*\*?\s*$text\s*\*/)$~i", $token->getContent());
     }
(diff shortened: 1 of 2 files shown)


SpacePossum:
nice catch :+1: 

maybe we could also test if `$this->configuration['comment_text']` (through configuration definition setup) doesn't contain `\\` itself so we don't insert an invalid comment token that way?

Slamdunk:
What about other special chars like pipe?

julienfalque:
> maybe we could also test if `$this->configuration['comment_text']` (through configuration definition setup) doesn't contain `\n\s*[^\\]{2}` itself so we don't insert an invalid comment token that way?

Indeed, shall we throw an exception in such case or just strip unwanted characters?

> What about other special chars like pipe?

They are all handled by `preg_quote` already, the point of the new test is to prove this call was missing, not that *all* characters are escaped one by one.

Slamdunk:
> They are all handled by preg_quote already, the point of the new test is to prove this call was missing, not that all characters are escaped one by one.

My fault sorry, I read bad the manual. 👍

SpacePossum:
I would throw an exception if the configured-content would not create one valid comment token, just IMHO of course :)

julienfalque:
I added the exception when comment text contains newlines and `rtrim` to prevent inserting trailing whitespaces.

SpacePossum:
thanks! looks all great to me! :)

SpacePossum:
grrrr sorry :(
phone acting against me...

keradus:
Thank you @julienfalque.

PR #3266, opened 2017-11-17
PhpUnitConstructFixer - multiple asserts bug

kubawerlos:


keradus:
is the issue only if assertions are on same line ?

kubawerlos:
no, the problem is always when the assertion to fix is not the first one - tests are in one line because I'm lazy and this is easier for me to read.

keradus:
Thank you @kubawerlos.

Issue #3274, opened 2017-11-24
[Feature Request] PHP Unit @test annotation (or not)


BackEndTea:
PHP Unit accepts the following as tests.
```php
    /**
     * @test
     */
    public function seeIfItWorks() {...}
```
or 
```php
    public function testSeeIfItWorks() {...}
```
I'd like to see a rule that enforces one or the other, so either remove ```test``` from the function name and add ```@test``` annotation, or remove ```@test``` annotation, and add ```test``` in front of the function name. (And also make sure the function is properly camel cased again.)

keradus:
Hi.
I think it is a good idea indeed.
Please, could you raise a PR ?

BackEndTea:
I'm playing around with it now, and i'll try and make a PR out of this when its ready.

keradus:
great to hear ;)
be sure to also give https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.8/doc/COOKBOOK-FIXERS.md  a try

BackEndTea:
Since this was merged in #3275 i think this can be closed

PR #3275, opened 2017-11-24
Add PhpUnitTestAnnotationFixer

BackEndTea:
This PR turns php unit tests into either
```php
    /**
     * @test
     */
    public function seeIfItWorks() {...}
```
or 
```php
    public function testSeeIfItWorks() {...}
```
as mentioned in #3274 

Covered edge cases can be found in the tests.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1101,6 +1101,19 @@ Choose from the list of available rules:
     ``['assertAttributeEquals', 'assertAttributeNotEquals', 'assertEquals',
     'assertNotEquals']``
 
+* **php_unit_test_annotation**
+
+  Adds or removes @test annotations from tests, following configuration.
+
+  *Risky rule: this fixer may change the name of your tests, and could cause incompatibility with abstract classes or interfaces.*
+
+  Configuration options:
+
+  - ``case`` (``'camel'``, ``'snake'``): whether to camel or snake case when adding the
+    test prefix; defaults to ``'camel'``
+  - ``style`` (``'annotation'``, ``'prefix'``): whether to use the @test annotation or
+    not; defaults to ``'prefix'``
+
 * **php_unit_test_class_requires_covers**
 
   Adds a default ``@coversNothing`` annotation to PHPUnit test classes that
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -1,0 +1,587 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @author Gert de Pagter
+ */
+final class PhpUnitTestAnnotationFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Adds or removes @test annotations from tests, following configuration.',
+            [
+                new CodeSample('<?php
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {} }'.$this->whitespacesConfig->getLineEnding()),
+                new CodeSample('<?php
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEnding(), ['style' => 'annotation']),
+                ],
+            null,
+            'This fixer may change the name of your tests, and could cause incompatibility with'.
+            ' abstract classes or interfaces.'
+        );
+    }
+
+    public function getPriority()
+    {
+        // must be run before the PhpdocSeparationFixer and PhpdocOrderFixer
+        return 10;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_FUNCTION]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach (array_reverse($this->findPhpUnitClasses($tokens)) as $indexes) {
+            if ('annotation' === $this->configuration['style']) {
+                $this->applyTestAnnotation($tokens, $indexes[0], $indexes[1]);
+            } else {
+                $this->removeTestAnnotation($tokens, $indexes[0], $indexes[1]);
+            }
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('style', 'Whether to use the @test annotation or not.'))
+                ->setAllowedValues(['prefix', 'annotation'])
+                ->setDefault('prefix')
+                ->getOption(),
+            (new FixerOptionBuilder('case', 'Whether to camel or snake case when adding the test prefix'))
+                ->setAllowedValues(['camel', 'snake'])
+                ->setDefault('camel')
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * @param Tokens $tokens
+     *
+     * @return int[][] array of [start, end] indexes from sooner to later classes
+     */
+    private function findPhpUnitClasses(Tokens $tokens)
+    {
+        $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
+        $phpunitClasses = [];
+
+        for ($index = 0, $limit = $tokens->count() - 1; $index < $limit; ++$index) {
+            if ($tokens[$index]->isGivenKind(T_CLASS) && $phpUnitTestCaseIndicator->isPhpUnitClass($tokens, $index)) {
+                $index = $tokens->getNextTokenOfKind($index, ['{']);
+                $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
+                $phpunitClasses[] = [$index, $endIndex];
+                $index = $endIndex;
+            }
+        }
+
+        return $phpunitClasses;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     * @param int    $endIndex
+     */
+    private function applyTestAnnotation(Tokens $tokens, $startIndex, $endIndex)
+    {
+        for ($i = $endIndex - 1; $i > $startIndex; --$i) {
+            if (!$this->isFunctionTest($tokens, $i)) {
+                continue;
+            }
+
+            $functionNameIndex = $tokens->getNextMeaningfulToken($i);
+            $functionName = $tokens[$functionNameIndex]->getContent();
+
+            //if the function name stats with test, we remove that
+            if ($this->startsWith('test', $functionName)) {
+                $newFunctionName = $this->removeTestFromFunctionName($functionName);
+                $tokens[$functionNameIndex] = new Token([T_STRING, $newFunctionName]);
+            }
+
+            $docBlockIndex = $this->getDockBlockIndex($tokens, $i);
+
+            //Create a new docblock if it didn't have one before;
+            if (!$this->doesFunctionHaveDocBlock($tokens, $i)) {
+                $this->createDocBlock($tokens, $docBlockIndex);
+
+                continue;
+            }
+            $lines = $this->updateDocBlock($tokens, $docBlockIndex);
+
+            $lines = $this->addTestToDocBlockIfNeeded($lines, $tokens, $docBlockIndex);
+
+            $lines = implode($lines);
+            $tokens[$docBlockIndex] = new Token([T_DOC_COMMENT, $lines]);
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     * @param int    $endIndex
+     */
+    private function removeTestAnnotation(Tokens $tokens, $startIndex, $endIndex)
+    {
+        for ($i = $endIndex - 1; $i > $startIndex; --$i) {
+            // We explicitly check again if the function has a doc block to save some time.
+            if (!$this->isFunctionTest($tokens, $i) || !$this->doesFunctionHaveDocBlock($tokens, $i)) {
+                continue;
+            }
+
+            $docBlockIndex = $this->getDockBlockIndex($tokens, $i);
+
+            $lines = $this->updateDocBlock($tokens, $docBlockIndex);
+
+            $lines = implode($lines);
+            $tokens[$docBlockIndex] = new Token([T_DOC_COMMENT, $lines]);
+
+            $functionNameIndex = $tokens->getNextMeaningfulToken($i);
+            $functionName = $tokens[$functionNameIndex]->getContent();
+
+            //if the function already starts with test were done
+            if ($this->startsWith('test', $functionName)) {
+                continue;
+            }
+
+            $newFunctionName = $this->addTestToFunctionName($functionName);
+            $tokens[$functionNameIndex] = new Token([T_STRING, $newFunctionName]);
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int$index
+     *
+     * @return bool
+     */
+    private function isFunctionTest(Tokens $tokens, $index)
+    {
+        // Check if we are dealing with a (non abstract, non lambda) function
+        if (!$this->isFunction($tokens, $index)) {
+            return false;
+        }
+
+        // if the function name starts with test its a test
+        $functionNameIndex = $tokens->getNextMeaningfulToken($index);
+        $functionName = $tokens[$functionNameIndex]->getContent();
+
+        if ($this->startsWith('test', $functionName)) {
+            return true;
+        }
+        // If the function doesn't have test in its name, and no doc block, its not a test
+        if (!$this->doesFunctionHaveDocBlock($tokens, $index)) {
+            return false;
+        }
+
+        $docBlockIndex = $this->getDockBlockIndex($tokens, $index);
+        $doc = $tokens[$docBlockIndex]->getContent();
+        if (false === strpos($doc, '@test')) {
+            return false;
+        }
+
+        return true;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    private function isFunction(Tokens $tokens, $index)
+    {
+        $tokensAnalyzer = new TokensAnalyzer($tokens);
+
+        return $tokens[$index]->isGivenKind(T_FUNCTION) && !$tokensAnalyzer->isLambda($index);
+    }
+
+    /**
+     * @param string $needle
+     * @param string $haystack
+     *
+     * @return bool
+     */
+    private function startsWith($needle, $haystack)
+    {
+        $len = strlen($needle);
+
+        return substr($haystack, 0, $len) === $needle;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    private function doesFunctionHaveDocBlock(Tokens $tokens, $index)
+    {
+        $docBlockIndex = $this->getDockBlockIndex($tokens, $index);
+
+        return $tokens[$docBlockIndex]->isGivenKind(T_DOC_COMMENT);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return int
+     */
+    private function getDockBlockIndex(Tokens $tokens, $index)
+    {
+        do {
+            $index = $tokens->getPrevNonWhitespace($index);
+        } while ($tokens[$index]->isGivenKind([T_PUBLIC, T_PROTECTED, T_PRIVATE, T_FINAL, T_ABSTRACT, T_COMMENT]));
+
+        return $index;
+    }
+
+    /**
+     * @param string $functionName
+     *
+     * @return string
+     */
+    private function removeTestFromFunctionName($functionName)
+    {
+        if ($this->startsWith('test_', $functionName)) {
+            if (is_numeric(substr($functionName, 6, 1))) {
+                return $functionName;
+            }
+
+            return  substr($functionName, 5);
+        }
+        if (is_numeric(substr($functionName, 5, 1))) {
+            return $functionName;
+        }
+        $functionName = substr($functionName, 4);
+
+        return lcfirst($functionName);
+    }
+
+    /**
+     * @param string $functionName
+     *
+     * @return string
+     */
+    private function addTestToFunctionName($functionName)
+    {
+        if ('camel' !== $this->configuration['case']) {
+            return 'test_'.$functionName;
+        }
+
+        return'test'.ucfirst($functionName);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return string
+     */
+    private function detectIndent(Tokens $tokens, $index)
+    {
+        if (!$tokens[$index - 1]->isWhitespace()) {
+            return ''; // cannot detect indent
+        }
+
+        $explodedContent = explode($this->whitespacesConfig->getLineEnding(), $tokens[$index - 1]->getContent());
+
+        return end($explodedContent);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $docBlockIndex
+     */
+    private function createDocBlock(Tokens $tokens, $docBlockIndex)
+    {
+        $lineEnd = $this->whitespacesConfig->getLineEnding();
+        $originalIndent = $this->detectIndent($tokens, $tokens->getNextNonWhitespace($docBlockIndex));
+        $toInsert = [
+            new Token([T_DOC_COMMENT, '/**'.$lineEnd."${originalIndent} * @test".$lineEnd."${originalIndent} */"]),
+            new Token([T_WHITESPACE, $lineEnd.$originalIndent]),
+        ];
+        $index = $tokens->getNextMeaningfulToken($docBlockIndex);
+        $tokens->insertAt($index, $toInsert);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $docBlockIndex
+     *
+     * @return Line[]
+     */
+    private function updateDocBlock(Tokens $tokens, $docBlockIndex)
+    {
+        $doc = new DocBlock($tokens[$docBlockIndex]->getContent());
+        $lines = $doc->getLines();
+
+        return $this->updateLines($lines, $tokens, $docBlockIndex);
+    }
+
+    /**
+     * @param Line[] $lines
+     * @param Tokens $tokens
+     * @param int    $docBlockIndex
+     *
+     * @return Line[]
+     */
+    private function updateLines($lines, Tokens $tokens, $docBlockIndex)
+    {
+        $needsAnnotation = 'annotation' === $this->configuration['style'];
+
+        $doc = new DocBlock($tokens[$docBlockIndex]->getContent());
+        for ($i = 0; $i < \count($lines); ++$i) {
+            //If we need to add test annotation and it is a single line comment we need to deal with that separately
+            if ($needsAnnotation && ($lines[$i]->isTheStart() && $lines[$i]->isTheEnd())) {
+                if (!$this->doesDocBlockContainTest($doc)) {
+                    $lines = $this->splitUpDocBlock($lines, $tokens, $docBlockIndex);
+
+                    return $this->updateLines($lines, $tokens, $docBlockIndex);
+                }
+                //One we split it up, we run the function again, so we deal with other things in a proper way
+            }
+
+            if (!$needsAnnotation &&
+                false !== \strpos($lines[$i]->getContent(), ' @test') &&
+                false === \strpos($lines[$i]->getContent(), '@testWith') &&
+                false === \strpos($lines[$i]->getContent(), '@testdox')
+            ) {
+                //We remove @test from the doc block
+                $lines[$i] = new Line(str_replace(' @test', '', $lines[$i]->getContent()));
+            }
+            //ignore the line if it isnt @depends
+            if (false === strpos($lines[$i], '@depends')) {
+                continue;
+            }
+
+            $lines[$i] = $this->updateDepends($lines[$i]);
+        }
+
+        return $lines;
+    }
+
+    /**
+     * Take a one line doc block, and turn it into a multi line doc block.
+     *
+     * @param Line[] $lines
+     * @param Tokens $tokens
+     * @param int    $docBlockIndex
+     *
+     * @return Line[]
+     */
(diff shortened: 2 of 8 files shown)


keradus:
Hi, thanks for quickly picking it up ! Really great to see it here ;)
Remember, that this project is heavily maintained and for that we need to work with code that we are super sure about, thus, if you will get some picky comments to change sth or review process won't be that fast as all of us would like, please be understanding, and sorry for that ;)

Don't bother about `getPriority` here, it's indicator of "when" given fixer shall be executed in comparison to other fixers - before or after some of them. It's our way of configuring the order of conflicting fixers.
As I don't see any conflict here, you don't need to implement the method on your own (so, you will use default implementation that is in abstract parent class).

About definitions - you mean the failing test? I will point to it in comment of given line soon ;)

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,256 @@
...
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {} }\n", ['annotation' => false]),
+                new CodeSample("<?php
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+public function testItDoesSomething() {}}\n",['annotation' => true]),
```
keradus: the test is failing, as it expect to have `CodeSample` without `configuration` value, so one that uses default config.
this given `CodeSample` here is in fact the that one case, yet it has the config explicitly set up same as default one.

To solve issue fast, simply drop configuration from this `ConfigSample`.
To solve it nicely yet a bit more work, you could fix test to approve both, lack of explicit config and explicit config same as default one.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,256 @@
...
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_FUNCTION]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach (array_reverse($this->findPhpUnitClasses($tokens)) as $indexes) {
+            call_user_func([$this, $this->callback], $tokens, $indexes[0], $indexes[1]);
```
keradus: please avoid doing this. it's basically calling a function by string name, it's impossible to track that kind of usages by any Static Code Analysis tools. simply, make a condition and call different methods here explicitly

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,256 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('annotation', 'Whether to use the @test annotation or not.'))
```
keradus: It's great to have configuration. Tet, let us make it even more generic allowing ppl to add more possibilities in the future:
`style`, allowed values `['annotation', 'prefix']` with default as `prefix` (I do strongly believe from my experience that most ppl are actually using `prefix` over `annotation`, so please use it as default. yet, i'm also awared that some ppl like to use annotation, so this option will be great for them!)

keradus:
btw, please don't squash the commits.
like now, I see that sth has been modified, some of my comments are outdated now, but I have no way to see what was actually changed, so I need to start from beginning...

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,256 @@
...
+            ->setAllowedValues([true, false])
```
keradus: consequtive methods shall be one level nested comparing to line with constructor

BackEndTea:
@keradus 
Will do, and thanks for the feedback so far

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,114 @@
...
+ */
+final class PhpUnitTestAnnotationFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed $expected
+     * @param mixed $input
+     */
+    public function testFix($expected, $input, array $config)
```
keradus: usually, we do `array $config = []`, so you don't need to write `expected, null, []` but just `expected` in your first case

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->fixer->configure($config);
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * Case 1: Notation present, and wanted
+     * Case 2: Notation present, and not wanted
+     * Case 3: Notation not present, but wanted
+     * Case 4: Notation not present, but wanted, and there is already a notation
+     * Case 4: Notation not present, and not wanted
```
keradus: hint:
create data provider like
```php
return [
    'Notation present, and wanted' => [ ... ],
    'Notation present, and not wanted' => [ ... ],
];
```

and then your descriptions will be used by PHPUnit during reporting ;)

localheinz:
@BackEndTea 

In regard to priority, probably makes sense to run this before the `NoEmptyPhpdocFixer`, so if the only content of the docblock is the `@test` annotation, that it will eventually be removed.

👍 

By the way, great idea!

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,256 @@
...
+{
+public function testItDoesSomething() {}}\n"),
+
+                new CodeSample("<?php
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {} }\n", ['annotation' => false])
```
localheinz: @BackEndTea 

Just an idea (since you and I have seen this in at least one repository):

* some people use `camelCase` with `@test` annotation
* some people use `snake_case` with `@test` annotation
* some people use prefix `test` followed by `PascalCase`
* some maybe even use prefix `test` followed by `PascalCase` and also have a redundant `@test` annotation

Not sure if the fixer should take care of the last case (we could just add a separate fixer), but I think it would be awesome if the fixer could cater to all the 3 cases before, what do you think?

README.rst
```diff
@@ -1046,6 +1046,15 @@ Choose from the list of available rules:
     ``['assertAttributeEquals', 'assertAttributeNotEquals', 'assertEquals',
     'assertNotEquals']``
 
+* **php_unit_test_annotation**
+
+  Adds or removes @test annotations from tests, following configuration.
+
+  Configuration options:
+
+  - ``annotation`` (``false``, ``true``): whether to use the @test annotation or not;
```
localheinz: Quick heads-up, in case you wondered, `README.rst` will eventually need to be regenerated by running 

```
$ ./php-cs-fixer readme > README.rst
```

BackEndTea: Ah thanks, forgot about that

keradus:
so.
before:
```php
/**
 * @test
 */
public function aaa() {}

/**
 * @test
 * @depends aaa
 */
public function bbb() {}
```

after:
so.

```php
/**
 */
public function testAaa() {}

/**
 * @depends aaa
 */
public function testBbb() {}
```

`@depends` annotation was not modified. tests will crash

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('style', 'Whether to use the @test annotation or not.'))
+                ->setAllowedValues(['prefix', 'annotation'])
+                ->setDefault('prefix')
+                ->setAllowedTypes(['string'])
```
keradus: no need to specify type when we only allowed explicitly pointed values.

BackEndTea:
Thanks for all the feedback so far.

@keradus 
I added a check for the ```@depends``` tags, and added tests to cover that behaviour.

@localheinz 
What priority do i have to give it in order to make sure that it is placed in the correct order?

And for your comment 
> * some people use camelCase with @test annotation
> * some people use snake_case with @test annotation
> * some people use prefix test followed by PascalCase
> * some maybe even use prefix test followed by PascalCase and also have a redundant @test annotation

This should cover case 1 and 3, but i'm not sure about how i should add support for case 2.

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,158 @@
...
+            'Annotation is used, and should be' => ['<?php
+class Test extends \PhpUnit\FrameWork\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {}
+    }', null, ['style' => 'annotation'],
+            ],
+            'Annotation is used, and it should not be' => ['<?php
```
keradus: let us globally change 
```php
$arr = ['aaa
bbb', 'ccc
ddd'];
```
into
```php
$aaa = [
    'aaa bbb',
    'ccc ddd',
];
```

we find it more readable

localheinz:
@BackEndTea @keradus 

I was just thinking, maybe it makes sense to separate this into two fixers?

* one that either removes the `test` prefix and adds the `@test` annotation or adds the `test` prefix and removes the `@test` annotation
* one that normalizes the test method name and is configurable, for example, accepts a `style` option
    * `camelCase`
    * `snake_case`
    * `wtf case` (see http://mnapoli.fr/using-non-breakable-spaces-in-test-method-names/)

Would that make sense?

localheinz:
@BackEndTea 

> What priority do i have to give it in order to make sure that it is placed in the correct order?

Maybe @keradus can help?

BackEndTea:
@localheinz 

I think i should be able to make this fix know if it should use camel or snake case, i'll get working on that soon. I don't mind adding support for wtf case, either, since it shouldn't be that hard, but i'm not sure if that is wanted.

One edge case i can think of is when the test goes from using an annotation to none, and the name, without test is just one word, but the standard for the project is snake case.

A fixer that normalises the test naming convention would be a separate fixer though.

keradus:
@localheinz

> I was just thinking, maybe it makes sense to separate this into two fixers?

in this PR, let us not make fixer that would fully normalize the test method.
Yet, if there is `/** @test */ function sth`, we need to know shall it be converted to `function testSth` or `function test_sth`.

> In the light of the latter, the former would probably just have to detect whether the test method name contains _, then prefix with test_, otherwise prefix with test and ucfirst() the current test method name (or something like that)
> Would that make sense?

No. When method is named with single word, like `pull`, you cannot detect it.


> > What priority do i have to give it in order to make sure that it is placed in the correct order?
> 
> Maybe @keradus can help?

out of the box - no clue. let us first finish spec of this fixer, then we will handle relations to different rules.


-----------

@BackEndTea

> I don't mind adding support for wtf case

no. don't. ever.


> One edge case i can think of is when the test goes from using an annotation to none, and the name, without test, is just one word, it wouldn't be able to know for sure what to use.

yeah, exactly. yet, we need to know which notation to use (camelCase vs snake_case).


> (Then again a test with a one word name is bad design and shouldn't happen)

No, it is not. There is a lot of valid cases to have single word.
(And even if it would be bad, we would still need to properly handle that)

> A fixer that normalises the test naming convention would be a separate fixer though.

exactly. let us not put it in scope of this PR.

BackEndTea:
Thanks for the insight @keradus 

> No. When method is named with single word, like pull, you cannot detect it.

I'm not sure how to properly detect it. And just defaulting to either camel or snake case would be bad. 

> > I don't mind adding support for wtf case
>
> no. don't. ever.

Got it.

>> A fixer that normalises the test naming convention would be a separate fixer though.
>
> exactly. let us not put it in scope of this PR.

Got it.

keradus:
> I'm not sure how to properly detect it.

you can't. you need config option for that.

BackEndTea:
I updated the addition logic to either add ```test_``` or ```test``` and uppercase the first letter of the original function, depending on the configuration.

For adding annotations, and thus removing ```test``` from the function name, it simply removes ```test_``` if its present, and otherwise it removes ```test```.

No wtf case support though.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -127,8 +134,11 @@ private function applyTestAnnotation(Tokens $tokens, $startIndex, $endIndex)
             if (!$this->startsWith('test', $functionName)) {
                 continue;
             }
-
-            $functionName = preg_replace('{test}', '', $functionName, 1);
+            if ($this->startsWith('test_', $functionName)) {
+                $functionName = preg_replace('{test_}', '', $functionName, 1);
```
keradus: why we need re here ?

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,335 @@
...
+
+            $functionNameIndex = $tokens->getNextMeaningfulToken($functionIndex);
+            $functionName = $tokens[$functionNameIndex]->getContent();
+
+            //ignore functions that don't start with test
+            if (!$this->startsWith('test', $functionName)) {
+                continue;
+            }
+            if ($this->startsWith('test_', $functionName)) {
+                $functionName = preg_replace('{test_}', '', $functionName, 1);
```
keradus: why we need re here ?

BackEndTea:
Known issue right now:
```php
    public function blabla() {}

    public function again() {}
```
becomes
```php
    /**
     * @test
     */
    public function blabla() {}
    /**
     * @test
     */

    public function again() {}
```

I can't seem to get it to properly add a new docblock if there is an empty new line above the function.

keradus:
> I can't seem to get it to properly add a new docblock if there is an empty new line above the function.

you need to split `T_WHITESPACE` into 2 tokens and add docblock between them

BackEndTea:
@keradus 
I have fixed that, and made sure the tests reflect that.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
BackEndTea: I updated it to use```substr``` instead now.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @author Gert de Pagter
+ */
+final class PhpUnitTestAnnotationFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    private $annotated;
```
keradus: please either typehint properties, or drop them (as we have `$configuration`)

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+     * @return bool
+     */
+    private function doesFunctionHaveDocBlock(Tokens $tokens, $index)
+    {
+        $docBlockIndex = $this->getDockBlockIndex($tokens, $index);
+
+        return $tokens[$docBlockIndex]->isGivenKind(T_DOC_COMMENT);
+    }
+
+    private function getDockBlockIndex(Tokens $tokens, $index)
```
keradus: typehinting is missing

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+        return $index;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $index
+     *
+     * @return bool
+     */
+    private function shouldBeChecked(Tokens $tokens, $index)
```
keradus: should **what** be checked ?

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+        $tokens->insertAt($index, $toInsert);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $docBlockIndex
+     *
+     * @return Line[]
+     */
+    private function updateDocBlock(Tokens $tokens, $docBlockIndex)
```
keradus: this looks overcomplicated. what is the purpose of this method ?

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+    /**
+     * @depends testCamelCased
+     */
+    public function test_depends_on_someone () {}
+
+    //It even has a comment
+    public function a_helper_function () {}
+
+    /**
+     * @depends depends_on_someone
```
keradus: please add test case with more complicated docblock, like
```php
/**
 * Description.
 *
 * @param ...
 *
 * @return ...
 *
 * @depends ...
 */
```

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+{
+    /**
+     * @test
+     */
+    public function it_has_snake_case() {}
+
+    public function helper_function() {}
+}',
+                ['case' => 'snake'],
+                ],
```
keradus: wrong indent

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+{
+    public function test_works_fine () {}
+
+    /**
+     * @depends test_works_fine
+     */
+    public function test_works_fine_too() {}
+}',
+                    ['style' => 'annotation'],
+                    ],
```
keradus: wrong indent

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+    public function works_fine () {}
+
+    /**
+     * @depends works_fine
+     */
+    public function works_fine_too() {}
+}',
+                    ['case' => 'snake'],
+                ],
+            'Class has both camel and snake case, annotated functions and not, and wants to add annotations' => [
```
keradus: again...

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+}',
+                '<?php
+class Test extends \PhpUnit\FrameWork\TestCase
+{
+    public function test_snake_cased () {}
+
+    public function testCamelCased () {}
+
+    /**
+     * @depends testCamelCased
```
keradus: please add test case with more complicated docblock, like
```php
/**
 * Description.
 *
 * @param ...
 *
 * @return ...
 */
```

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
keradus: here, in input code, the method we depends on is still called `test_depends_on_someone`.
It shall remain `depends_on_someone` in expected code

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+ *
+ * @covers \PhpCsFixer\Fixer\PhpUnit\PhpUnitTestAnnotationFixer
+ */
+final class PhpUnitTestAnnotationFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed $expected
+     * @param mixed $input
```
keradus: it shouldn't be mixed

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
keradus: this is pretty nice test case I was thinking about, big :+1: you added it already :)

yet, let us extend a bit, as I found edge case that is not properly handled
https://github.com/BackEndTea/PHP-CS-Fixer/pull/1

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+        return new FixerDefinition(
+            'Adds or removes @test annotations from tests, following configuration.',
+            [
+                new CodeSample("<?php
+class Test extends \\PhpUnit\\FrameWork\\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {} }\n"),
```
keradus: please don't use `\n` inside `"` strings. just use real new line.
also, please put closing `}` of class on proper, dedicated line

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param $docBlockIndex
+     */
+    private function createDocBlock(Tokens $tokens, $docBlockIndex)
+    {
+        $originalIndent = $this->detectIndent($tokens, $tokens->getNextNonWhitespace($docBlockIndex));
+        $toInsert = [
+            new Token([T_DOC_COMMENT, "/**\n$originalIndent * @test\n$originalIndent */"]),
+            new Token([T_WHITESPACE, "\n".$originalIndent]),
```
keradus: you need to use whitespace config, and then use proper line ending / indentation.
for now, it will produce wrong code for ppl using tabs / windows line endings.

cover such cases in new test method called `testMessyWhitespaces` (grep for it to get idea what to do)

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -0,0 +1,427 @@
...
+        return [
+            'Annotation is used, and should be' => [
+                '<?php
+class Test extends \PhpUnit\FrameWork\TestCase
+{
+    /**
+     * @test
+     */
+    public function itDoesSomething() {}
+    }',
```
keradus: wrong indent

keradus:
could you please include https://github.com/BackEndTea/PHP-CS-Fixer/pull/2 ? ;)

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+            (new FixerOptionBuilder('case', 'Whether to camel or snake case when adding the test prefix'))
+                ->setAllowedValues(['camel', 'snake'])
+                ->setDefault('camel')
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $index
```
keradus: type is missing

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+        do {
+            $index = $tokens->getPrevNonWhitespace($index);
+        } while ($tokens[$index]->isGivenKind([T_PUBLIC, T_PROTECTED, T_PRIVATE, T_FINAL, T_ABSTRACT, T_COMMENT]));
+
+        return $index;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $index
```
keradus: type is missing

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+        } else {
+            $newFunctionName = 'test'.ucfirst($functionName);
+        }
+
+        return $newFunctionName;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $docBlockIndex
```
keradus: type is missing

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,375 @@
...
+            new Token([T_DOC_COMMENT, "/**\n$originalIndent * @test\n$originalIndent */"]),
+            new Token([T_WHITESPACE, "\n".$originalIndent]),
+        ];
+        $index = $tokens->getNextMeaningfulToken($docBlockIndex);
+        $tokens->insertAt($index, $toInsert);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $docBlockIndex
```
keradus: type is missing

keradus:
@BackEndTea, thanks for great work!
I raised some comments, mainly minors, yet one bug and lack of whitespaces configuration are the blockers

BackEndTea:
@keradus I fixed the minor issues, as well as the bug, and the whitespace configuration.

BackEndTea:
I noticed an error, which also isn't covered by any test cases. It currently breaks when trying to handle a test that was annotated like this ```/** @test */``` instead of
```php
/**
 * @test
 */
```

BackEndTea:
On that note, i figured out some more edge cases involving single line doc blocks that aren't covered yet.

keradus:
this one-liner -> multiliner annotation is done only when annotation is added, right?

about dedicated fixer to do it always, let us finish this PR first and then we will see ;)


Sounds really great.
Sadly, my time this week is very limited. Let me try to do next round next week, or perhaps someone else is willing to take a look as well ? :)

BackEndTea:
Yeah the one liner to multi liner is only done when adding the ```@test``` annotation.

I could also change it to convert ```/** @group Database */``` to ```/** @test @group Database */```, instead of
```
/**
 * @test
 * @group Database
 */
```

But thats up to the maintainers to decide

keradus:
nah, when adding annotation convert to multiliner, that makes sense :+1:

BackEndTea:
I'm a little bit stuck on something
```php
/**
 * asdf@testasdf
 */
```
Gets accepted by phpUnit just the same as
```php
/**
 * @test
 */
```

Should this fixer 'see' the ```@test``` annotation in the first example?

keradus:
> I'm a little bit stuck on something (...) Gets accepted by phpUnit (...)

Don't bother about it.
IMO, you shall open a bug report ;)

> Should this fixer 'see' the `@test` annotation in the first example?

I believe that our DocBlock parser is not see that annotation, and it's good if it does't.
If that's the case, you have nothing to do with it.

> Of course it would remove the test prefix, but should it add an explicit @test to the Doc block?

yes, it should.

BackEndTea:
Things I have no idea how to currently do/fix:

- How to handle abstract test functions
- How to handle interfaces with test functions
- How to handle traits with test functions

For example:
```php
trait ItsAUser 
{
    public function testItsAUser()
    {
        $this->assertTrue($this->user->isUser());
    }
}

class TestUser extends \PHPUnit\Framework\TestCase
{
    use ItsAUser;
    
    protected $user;
    
    protected function setUp()
    {
        $this->user = new \Foo\User();
        
    }
}
```

There is good way to check that the trait is actually a test, and its not just a normal function that starts with test.

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
```diff
@@ -931,14 +899,93 @@ class Test extends \PhpUnit\FrameWork\TestCase
...
+        ];
+    }
+
+    /**
+     * @dataProvider provideNoChangeFixCases
+     *
+     * @param string $input
+     * @param array  $config
+     */
+    public function testItDoesNotChange($input, array $config = [])
```
keradus: please move it back to one test method / data provider

BackEndTea: Okay

keradus:
> How to handle abstract test functions

like any other tets function

> How to handle interfaces with test functions
> How to handle traits with test functions

for now, ignore them. let us focus on finishing PR rather than extending it's scope, it could be done later if needed

BackEndTea:
@keradus 
My concern is this, I could have an abstract function lets say 
```php
    abstract public function itDoesThings();
```
Now, when implemented, it is done like this
```php
    /**
     * @test
     */
    public function itDoesThings() {}
```

In the abstract class, there is no way to guarantee that an abstract function is a test or not

keradus:
then in abstract class it shall be marked with `@test` as well.

look, when one use other way to mark method as a test - a prefix, then he mark that abstract method with prefix as well.


> But it could break tests if the fixer ran over this.

any change of function name can break code. very good point. note that all of our phpunit fixers are marked as risky.

yes, mark this fixer as risky as well.

BackEndTea:
>yes, mark this fixer as risky as well.

Will do

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -0,0 +1,572 @@
...
+
+        return $lines;
+    }
+
+    /**
+     * @param DocBlock $doc
+     *
+     * @return bool
+     */
+    private function doesDocBlockContainTest(DocBlock $doc)
```
keradus: can we just get annotation from docblock? it has all the logic already

keradus: ping @BackEndTea

BackEndTea: I'm sorry, i don't understand exactly what it is you are asking.

keradus:
```php
private function doesDocBlockContainTest(DocBlock $doc)
{
    return !empty($doc->getAnnotationsOfType('test'));
}
```

BackEndTea: Then we get some issues with functions annotated like this:
```php
/**
 * This is a @test
 */
```
Which is a valid phpunit test

BackEndTea: It causes the ```Annotation does not get added when it is already present in a weird place``` case to break.

If we don't want to support that, i'm okay with that too

keradus: as said in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3275#issuecomment-349796720 , we shouldn't handle that scenario

BackEndTea: 👍 I'll update it then

tests/Fixer/PhpUnit/PhpUnitTestAnnotationFixerTest.php
keradus: it's added

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
keradus: done

keradus:
I let myself fix last raised issue

keradus:
new concerns from final tests:

1. when we want to use `@test` annotation and current test method name is `test123foo`, fixer would rename method to `123foo`, which is PHP syntax issue, this has to be fixed (probably adding annotation, yet leaving test prefix)

2. fixer is mature enough, let us fix priority:
fixer must run before `phpdoc_trim` and `no_empty_phpdoc`

3. again for `@test` anontation over prefix.
`testGetName` is renamed to `getName`. it's the issue, as it's overwriting internal PHPUnit method (most probably with wrong params, as test could have no params, and internal one has a param). Not sure about best approach here.

BackEndTea:
I added a test, and a fix for issue 1 you raised.

BackEndTea:
The only way around issue 3 i can really see is having a list of all phpunit functions, and adding those to some kind of black list. But that is going to be a big list

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -65,6 +65,12 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
         );
     }
 
+    public function getPriority()
+    {
+        // must be run before the PhpdocSeparationFixer and PhpdocOrderFixer
```
keradus: you need to also add this as a test here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.8/tests/AutoReview/FixerFactoryTest.php#L182

in tests here, you also have a comments with filename of integration test. you will need to add 2 of them, take a look at other ones so you will get idea how to create them

keradus:
> But that is going to be a big list

and imposible to maintain whenever external phpunit class will change/add/remove it's methods

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
BackEndTea: I'm adding those now, as well as running the fixer again

keradus:
please freeze now

BackEndTea:
:+1

keradus:
damn, nightly was failing...

keradus:
Thank you @BackEndTea.

keradus:
really great work here !

BackEndTea:
Real big thanks on all the feedback and help @keradus & @localheinz

PR #3276, opened 2017-11-24
Transform Fully Qualified parameters and return types to short version

veewee:
This PR removes fully qualified parameters and return types to the short versions when they are being imported / in de same namespace:

Result:
```php
<?php

use Foo\Bar;
use Foo\Bar\Baz;

class SomeClass
{
    public function doSomething(Bar $foo): Baz
    {
    }
}
```

Source:
```php
<?php

use Foo\Bar;
use Foo\Bar\Baz;

class SomeClass
{
    public function doSomething(\Foo\Bar $foo): \Foo\Bar\Baz
    {
    }
}
```

In older codebases (early namespaces), this was frequently used. Besides that, it can be used to fix autogenerated code from zend-code (https://github.com/zendframework/zend-code/issues/76) and crappy docblocks in the https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3110 PR I am working on:


keradus:
what about going even forward and having input like:
```
<?php

class SomeClass
{
    public function doSomething(\Foo\Bar $foo): \Foo\Bar\Baz
    {
    }
}
```

so even add imports when they are missing

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,219 @@
+<?php
+
+declare(strict_types=1);
```
keradus: please drop it. we supports 5.6

veewee:
Going forward doesnt seems like a good idea since I am changing more tokens into less tokens right?
I was indeed thinking of adding the use statements, but the problem is that I don't know where the imports should be added at te moment. Even if you find the index of the last use statement: at that point you will be adding multiple tokens to the file which makes it hard to loop through the document. Any help is appreciated, but currently I try to focus on getting the basics ready :)

veewee:
@keradus : Is it possible to partially mark a fixer for a specific PHP version? Strict return types are only available from PHP 7.0 on. Should I target PHP >= 7.0 or maybe it's better to split the fixer in 2 different fixers. how do you see this?

Slamdunk:
@keradus I think it would raise a lot the complexity, since the fixer would also need to check already imported class to avoid conflicts, like:

```php
use Foo\Bar;

var_dump(\Other\Bar::class);
```

@veewee I suggest to try to fix also:

* Partial namespace imported:

```php
use Foo\Bar;

// From
var_dump(\Foo\Bar\Baz::class);
// To
var_dump(Bar\Baz::class);
```

* Partial namespace declared:

```php
namespace Foo\Bar;

// From
var_dump(\Foo\Bar\Baz::class);
// To
var_dump(Baz::class);
```

keradus:
very well, let us keep moving FQCN into imports as separated fixer (so feature request that would not be solved in your PR)

veewee:
@Slamdunk : implemented the partial use + namespaces

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,202 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return PHP_VERSION_ID >= 70000 && $tokens->isTokenKindFound(T_FUNCTION);
```
keradus: why don't fix params on 5.x , definition says it shall

veewee: See question above. I changed the code to make it as PHP 5.6 compatible for parameters only.

tests/Fixer/Import/FullyQualifiedStrictTypesFixerTest.php
```diff
@@ -0,0 +1,283 @@
...
+EOF;
+
+        $this->doTest($expected, $input);
+    }
+
+    public function testImportedStrictTypesFixWithReturn()
+    {
+        if (PHP_VERSION_ID < 70000) {
+            $this->markTestSkipped('The strict return type : operator is only avaiable from PHP 7.0.');
+        }
```
Slamdunk: Why not using the dedicated method DocBlock
```php
/**
 * @requires PHP 7.0
 */
```
?

veewee: Fixed

README.rst
```diff
@@ -548,6 +548,10 @@ Choose from the list of available rules:
   PHP code must use the long ``<?php`` tags or short-echo ``<?=`` tags and not
   other tag variations.
 
+* **fully_qualified_strict_types**
```
keradus: please add it to local `.php_cs.dist` file

tests/Tokenizer/Analyzer/ArgumentsAnalyzerTest.php
```diff
@@ -42,13 +42,98 @@ public function testArguments($code, $openIndex, $closeIndex, array $arguments)
...
+    public function provideArgumentsInfoCases()
+    {
+        return [
+            ['<?php function($a){};', 3, 3, [
+                'default' => '',
+                'name' => '$a',
+                'name_index' => 3,
+                'type' => '',
+                'type_index_start' => -1,
+                'type_index_end' => -1,
```
keradus: -1 is invalid index

tests/Tokenizer/Analyzer/ArgumentsAnalyzerTest.php
```diff
@@ -42,13 +42,98 @@ public function testArguments($code, $openIndex, $closeIndex, array $arguments)
...
+            ]],
+            ['<?php function($a, $b = array(1,2), $c = 3){};', 5, 15, [
+                'default' => 'array(1,2)',
+                'name' => '$b',
+                'name_index' => 6,
+                'type' => '',
```
keradus: please use null instead of empty string.

src/Tokenizer/Analyzer/ArgumentsAnalyzer.php
```diff
@@ -83,4 +83,50 @@ public function getArguments(Tokens $tokens, $openParenthesis, $closeParenthesis
 
         return $arguments;
     }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $argumentStart
+     * @param int    $argumentEnd
+     *
+     * @return array
```
keradus: this array looks complicated. I must admit that looking on implementation I have no idea what will be the output - like is `name` obligatory, or could be empty. when i will get -1 instead of proper index.
probably new DTO object would be good.

src/Tokenizer/Analyzer/NamespaceUsesAnalyzer.php
```diff
@@ -0,0 +1,82 @@
...
+
+namespace PhpCsFixer\Tokenizer\Analyzer;
+
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @internal
+ */
+class NamespaceUsesAnalyzer
```
keradus: why not final ?

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,219 @@
...
+            }
+        }
+
+        return $type;
+    }
+
+    /**
+     * @param string $shortType
+     *
+     * @return array
```
keradus: `Token[]` perhaps ?

tests/Fixer/Import/FullyQualifiedStrictTypesFixerTest.php
```diff
@@ -0,0 +1,280 @@
...
+/**
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Import\FullyQualifiedStrictTypesFixer
+ */
+final class FullyQualifiedStrictTypesFixerTest extends AbstractFixerTestCase
+{
+    public function testImportedStrictTypesFixWithoutReturn()
```
keradus: please use single test method with data provider.
info in names is nice indeed, yet you could as well put that info in data provider

tests/Tokenizer/Analyzer/FunctionsAnalyzerTest.php
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @covers \PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer
+ */
+final class FunctionsAnalyzerTest extends TestCase
+{
+    /**
+     * @param string $code
+     * @param int    $methodIndex
+     * @param array  $expected
+     * @dataProvider provideFunctionsWithArgumentsCases
```
keradus: please separate params form data provider with empty line

keradus:
code in PR is nice, yet reviewing it requires some time. I wonder, maybe you could extract sth from this PR to separated one (like extracting refactoring of analyzers) ?

veewee:
Hi @keradus,

I've applied the remarks. I would prefer to keep both the fixer and the analyzers as one PR. Otherwise it would take me too much time to split them up and bring them together afterwards again.
The code inside the analyzers is code that was taken from existing fixers to make it more reusable.

Take all the time you need to review :)

keradus:
OK, thanks for update. Please be aware that I will rather not be able to do it this week. Sorry for that :(
Or, perhaps, someone else is willing to take a look as well ? ;)

(CircleCI: we faced the limit when we recently added new tests on 2.2, CirlceCI still works, merged 2.2->master, too much tests, CircleCI crashed. If you are willing to adjust memory, please raise new PR, thanks !)

veewee:
@keradus : I don't understand why the test is failing. It worked before rebasing against master :)

SpacePossum:
Hi @veewee,
 the failing `CiIntegrationTest::testIntegration` with `Broken pipe` is not related to the changes you made in the PR, no worries about that one (it is an unstable test we haven't completely figured out yet)

SpacePossum:
Lot of work so far, thanks!
I've a question about the fixer, both the sample code and unit test work with class methods; would this fixer also work for `functions` and methods within `traits` and `interfaces` ?

veewee:
@SpacePossum : It should, but I did not test those cases either. I'll add some test cases for that as well.

tests/Tokenizer/Analyzer/NamespaceUsesAnalyzerTest.php
```diff
@@ -111,6 +111,21 @@ public function provideNamespaceUsesCases()
                     21
                 ),
             ], [1, 12]],
+
+            // TODO: How to support these:
```
veewee: These TODOs are inheritence of the NoUnusedImportsFixer. If it's OK to you, I would leave them as-is for now and add support in another PR.

keradus:
@veewee , sorry for lack of update on this one from my side recently.
Please rebase and double confirm it's ready for new review round (maybe last one?)

veewee:
@keradus  : rebased against master. Feel free to do a new review. To me, this PR is at a final state; Additional work could be done in another one.

keradus:
@veewee , please solve conflicts, I can't merge

veewee:
Resolved.

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+
+        $lastIndex = $tokens->count() - 1;
+        for ($index = $lastIndex; $index >= 0; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            // Return types are only available since PHP 7.0
+            if (PHP_VERSION_ID >= 70000) {
+                $this->fixFunctionReturnType($tokens, $index, $namespaces, $useMap);
```
keradus: just being defensive here, but one day, after some refactoring, John Doe will call `fixFunctionReturnType` without knowledge he has to check PHP version first. I would put the check into method directly, as it looks like atomic part of it

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+                $this->generateTokensForShortType($shortType)
+            );
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     * @param array  $namespaces
+     * @param array  $useMap
```
keradus: what is inside those arrays? can we document that better ?

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+            $returnType->getStartIndex(),
+            $returnType->getEndIndex(),
+            $this->generateTokensForShortType($shortType)
+        );
+    }
+
+    /**
+     * @param string $type
+     * @param array  $namespaces
+     * @param array  $useMap
```
keradus: same here

keradus: looking at code, `namespaces` is `array<string, string>` ?

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
```diff
@@ -0,0 +1,141 @@
...
+     * @param string      $name
+     * @param int         $nameIndex
+     * @param null|string $default
+     * @param null|string $type
+     * @param null|int    $typeIndexStart
+     * @param null|int    $typeIndexEnd
+     */
+    public function __construct($name, $nameIndex, $default, $type, $typeIndexStart, $typeIndexEnd)
+    {
+        $this->name = (string) $name;
```
keradus: what is the valid case of passing here non-string and cast it into string?
i wouldn't support that, we expected a string, when we will bump requirements to 7.0 we will have string typehint in constructor, making compiler crash on non-strings and no casing will be done ever.

please, remove casing or proof non-string type is valid input, but then change the typehinting in docblock.

same in all other places (not only this class).

-------------------------------

i do know that casing is kind-of defensive approach, but if you want to go defensive, raise InvalidArgumentException if passed non-php typehinted argument is violating typehint from docblock

src/Tokenizer/Analyzer/Analysis/NamespaceUseAnalysis.php
```diff
@@ -0,0 +1,112 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Analyzer\Analysis;
+
+/**
+ * @internal
+ */
+final class NamespaceUseAnalysis implements StartEndTokenAwareAnalysis
```
keradus: just wondering, maybe it shall have only 2 properties - `namespaceAnalysis` and `isAliased` ?

src/Tokenizer/Analyzer/Analysis/NamespaceUseAnalysis.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        return $this->shortName;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isAliased()
+    {
+        return $this->aliased;
```
keradus: boolean shall be named with `is` prefix, same as method to access it is already named

keradus:
@TomasVotruba , as you have offered your help whenever we need to review sth, I would like to use it once more. Could you please help with that one ?

TomasVotruba:
@keradus Well, not whenever ("as a service" ;)), but when I find time and joy, I'll help.
Will look into this one

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
veewee: Indeed: I've added it as a defensive approuch since typehints are not supported yet. Will remove it ...

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_FUNCTION);
```
TomasVotruba: Also probably: `T_STRING`, `T_USE` and namespace separator

This means that this is not matched?

```php
use \Foo\Bar;

$value = new \Foo\Bar;
```

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $namespaces = array_map(function (NamespaceAnalysis $info) {
+            return $info->getFullName();
+        }, (new NamespacesAnalyzer())->getDeclarations($tokens));
+        $useMap = array_map(function (NamespaceUseAnalysis $info) {
+            return $info->getFullName();
+        }, (new NamespaceUsesAnalyzer())->getDeclarationsFromTokens($tokens));
```
TomasVotruba: These 2 operations with many calls and variables could be decoupled into 2 methods: 

```php
$this->getNamespacesFromTokens($tokens);
$this->getUseMapFromTokens($tokens);
```

making it more readable.

src/Tokenizer/Analyzer/Analysis/NamespaceUseAnalysis.php
veewee: I would keep those 2 cases separate: use statements can also import functions and multiple classes at once. Currently that is not supported, but it would probably be easier not to mix this with the NamespaceAnalysis

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
TomasVotruba: Yea, the check for version should be in `fixFunctionReturnType()` itself

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+     * @param int    $index
+     * @param array  $namespaces
+     * @param array  $useMap
+     */
+    private function fixFunctionArguments(Tokens $tokens, $index, array $namespaces, array $useMap)
+    {
+        $arguments = (new FunctionsAnalyzer())->getFunctionArguments($tokens, $index);
+
+        foreach ($arguments as $argument) {
+            if (!$argument->hasType()) {
```
TomasVotruba: Does this cover scalar types too?


If not, there should be something like:

```php
if ($argument->isScalar()) {
    continue;
}
```

I got bad experience with differentiation single word types like: `Type` vs. `string`
Caused few bugs already :)

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,224 @@
...
+                continue;
+            }
+
+            $shortType = $this->detectShortType($argument->getType(), $namespaces, $useMap);
+            if ($shortType === $argument->getType()) {
```
TomasVotruba: Here comment would be nice, since short type is "the last word of imported namespace" and that might not be clear to other programmers, who will read this in following years.

E.g.

```php
// use Foo\Bar => "Bar"
```

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,226 @@
...
+     * @param Tokens                $tokens
+     * @param int                   $index
+     * @param array<string, string> $namespaces a list of all FQN namespaces in the file with the short name as key
+     * @param array<string, string> $useMap     a list of all FQN use statements in the file with the short name as key
+     */
+    private function fixFunctionReturnType(Tokens $tokens, $index, array $namespaces, array $useMap)
+    {
+        if (PHP_VERSION_ID < 70000) {
+            return;
+        }
```
TomasVotruba: Nice :clap:

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,226 @@
...
+        $shortType = $this->detectShortType($returnType->getType(), $namespaces, $useMap);
+        if ($shortType === $returnType->getType()) {
+            return;
+        }
+
+        $tokens->overrideRange(
+            $returnType->getStartIndex(),
+            $returnType->getEndIndex(),
+            $this->generateTokensForShortType($shortType)
+        );
```
TomasVotruba: Line 156 to 163 is duplicated to 124 - 134, right?

If so, decoupling method would be nice and more bug proof

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,226 @@
...
+        foreach ($useMap as $shortName => $fullName) {
+            $matches = [];
+            $regex = '/^\\\\?'.preg_quote($fullName, '/').'\\\\(?P<className>.+)$/';
+            if (preg_match($regex, $type, $matches)) {
+                return $shortName.'\\'.$matches['className'];
+            }
+        }
+
+        return $type;
+    }
```
TomasVotruba: This might deservce own class. It would be useable out of this fixer for other classes and also for writing own fixers.
I'd definitelly use it :+1:

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
```diff
@@ -0,0 +1,226 @@
...
+        $parts = explode('\\', $shortType);
+
+        foreach ($parts as $index => $part) {
+            $tokens[] = new Token([T_STRING, $part]);
+            if ($index !== count($parts) - 1) {
+                $tokens[] = new Token([T_NS_SEPARATOR, '\\']);
+            }
+        }
+
+        return $tokens;
```
TomasVotruba: Nice method decoupling! :+1:

src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -85,8 +89,8 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
             return;
         }
 
-        $useDeclarations = $this->getNamespaceUseDeclarations($tokens, $useDeclarationsIndexes);
-        $namespaceDeclarations = $this->getNamespaceDeclarations($tokens);
+        $useDeclarations = (new NamespaceUsesAnalyzer())->getDeclarationsFromTokens($tokens);
+        $namespaceDeclarations = (new NamespacesAnalyzer())->getDeclarations($tokens);
```
TomasVotruba: Great job, I'll can finally use these :+1:

src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -233,9 +153,9 @@ private function removeUnusedUseDeclarations(Tokens $tokens, array $useDeclarati
         }
     }
 
-    private function removeUseDeclaration(Tokens $tokens, array $useDeclaration)
+    private function removeUseDeclaration(Tokens $tokens, NamespaceUseAnalysis $useDeclaration)
     {
-        for ($index = $useDeclaration['end'] - 1; $index >= $useDeclaration['start']; --$index) {
+        for ($index = $useDeclaration->getEndIndex() - 1; $index >= $useDeclaration->getStartIndex(); --$index) {
```
TomasVotruba: I'm bit confused by the `NamespaceUseAnalysis` name. Analysis is kind of report on many cases with numbers, like how many use imports in the code, what is their order etc.

But here it looks like 1 group of tokens. Using `NamespaceUseTokenGroup` would be more clear

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
```diff
@@ -0,0 +1,141 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Analyzer\Analysis;
+
+/**
+ * @internal
+ */
+final class ArgumentAnalysis
```
TomasVotruba: Beautiful abstraction :+1:

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
```diff
@@ -0,0 +1,141 @@
...
+     * @var int
+     */
+    private $nameIndex;
+
+    /**
+     * The type of the argument.
+     *
+     * @var null|string
+     */
+    private $type;
```
TomasVotruba: Follow up to my scalar type coment above. Would it be worth it to add `isScalar()` method here?

I'm working with types on many places and I have to check this on every space manually. This method would save repeated code.

src/Tokenizer/Analyzer/Analysis/StartEndTokenAwareAnalysis.php
```diff
@@ -0,0 +1,30 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Analyzer\Analysis;
+
+interface StartEndTokenAwareAnalysis
```
TomasVotruba: I :heart:  this abstraction

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
veewee: Currently only the types inside function arguments and return types are being replaced. Maybe thats something for a future PR?
(I would like to get this one in before adding more logic)

TomasVotruba: I'm ok with that, just it's not clear from fixer description. It should be stated explicitly there

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
veewee: Agree. This would also avoid passing the namespaces and uses across the fixer.
Not sure where to place it though. Any suggestions?

src/Fixer/Import/NoUnusedImportsFixer.php
veewee: That's a matter of perception. I like NamespaceUseAnalysis better: it's an analysis on a use statement. Just like a ArgumentAnalysis is an analysis of one argument. I can name it anything you want, but in the end : it is what it is :)

veewee:
@TomasVotruba 
Thanks for the review! Your enthousiasm is amazing :)

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
TomasVotruba: Great! I'm not exactly sure, what Analyzers are for, but if it would cover this, why not there:

```php
PhpCsFixer\Tokenizer\Analyzer\ShortNameAnalyzer
```

or

```php
PhpCsFixer\Tokenizer\Resolver\Namespace\ShortNameResolver
```

src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
```diff
@@ -0,0 +1,87 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Analyzer\Analysis;
+
+class TypeAnalysis implements StartEndTokenAwareAnalysis
```
keradus: let us keep all analysis final+internal for the first iteration. let us keep them a bit and let them become mature before putting a BC promise on them

src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
veewee: Great, I'll go for the resolver since it is more then an analysis

veewee:
@TomasVotruba 
I moved everything that shouldn't be in the fixer to separate classes (Token Resolvers, Token Generators and Token Analyzers)
Code looks even better now :)

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * ArgumentAnalysis constructor.
+     *
+     * @param string            $name
+     * @param int               $nameIndex
+     * @param null|string       $default
+     * @param null|TypeAnalysis $type
```
keradus: why optional ?

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
```diff
@@ -0,0 +1,111 @@
...
+     * @var int
+     */
+    private $nameIndex;
+
+    /**
+     * The type of the argument.
+     *
+     * @var null|TypeAnalysis
+     */
+    private $type;
```
keradus: this is not `type`, name is misleading

src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php
veewee: A function argument can have an empty type:
```function ($a) {}```

File: src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php (hunk 34 above)
Comment (veewee) [reply src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php]: `$typeAnalysis` feels a bit too bloated don't you think?

Comment (keradus) [reply src/Tokenizer/Analyzer/Analysis/ArgumentAnalysis.php]: no, as one, by just looking at name, would expect to have real argument here, not sth that will provide him info about indexes and is it scalar or not, without way to actually get that type.

File: src/Tokenizer/Analyzer/Analysis/NamespaceUseAnalysis.php [hunk 35]
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * The end index of the namespace declaration in the analyzed Tokens.
+     *
+     * @var int
+     */
+    private $endIndex;
+
+    /**
+     * NamespaceAnalysis constructor.
```
keradus: please avoid info-less doc summaries

tests/Tokenizer/Analyzer/Analysis/TypeAnalysisTest.php
```diff
@@ -69,7 +69,9 @@ public function provideScalarCases()
             ['int', true],
             ['iteratable', true],
             ['float', true],
+            ['self', true],
```
keradus: http://php.net/manual/pl/function.is-scalar.php

src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
```diff
@@ -12,50 +12,53 @@
...
 {
+    private static $scalarTypes = [
+        'array',
+        'bool',
+        'callable',
+        'int',
+        'iteratable',
+        'float',
+        'string',
```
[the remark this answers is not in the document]
keradus: http://php.net/manual/pl/function.is-scalar.php

scalar in PHP has a meaning already, reusing scalar word with different definition what is a scalar type actually gonna be super misleading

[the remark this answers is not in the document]
TomasVotruba: This is rather to exclude PHP-reserved types to avoid bugs.

I got a fixer that changes varaible name to it's type.
And without `mixed`/`object` as scalar (or *PHP-reserved* types) it made mess like:

```diff
 /**
  * @var mixed
  */
-private $variable;
+private $mixed;
```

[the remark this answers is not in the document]
keradus: i'm not saying anything about **calling** the `_is_scalar`.
i'm saying that scalar term already have a meaning in PHP as pseudotype, and it doesn't cover objects. reusing the term but with different meaning is super misleading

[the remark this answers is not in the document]
TomasVotruba: What term would you suggest?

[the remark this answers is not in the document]
keradus: change used term to some non-established in php internals and document what it is.
for now, i must admit that i have no clue what is this info suppose to be.
reverse of collection? but then why not `resource` ? no clue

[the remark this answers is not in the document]
TomasVotruba: @veewee Nice, thanks for the references in comments!

veewee:
@keradus : I would love to finish this one up. What's the state of this PR from your point of view?

TomasVotruba:
LGTM. What needs to be done here?

keradus:
@veewee , I see amazing job of yours here and I do appreciate it fully ! It looks really great, and I know this fixer was highly awaited by community!
My point of view is always the same - community need to review, I don't insist it has to be me every single time ;)
Sadly, it's hard to me to keep an eye on every single PR, sorry for big delays here, I take them on my foot :(

@TomasVotruba , thanks for big help with reviewing the PR, it's highly appreciated as well!

keradus:
Finally, merging :D

TomasVotruba:
@veewee Great job, thank you for all the work.

@keradus Thanks for getting to this. I know you're not the superman, so much appreciated :+1:

veewee:
Thanks for all the feedback! Really appreciate it!

PR #3278, opened 2017-11-26
PHPUnit will return methods fixer

kingmad:
In my company we've found that fixer which will help to change all occurrences of `->will($this->returnValue('some_value'));` to `->willReturn('some_value');` could be helpful to make our code a bit shorter and cleaner so I decide to contribute.

Also I had in mind that there's couple similar cases which can be fixed the same way so I decided to turn it to configurable fixer and here it is, working for following cases as shown below:

Input:
```<?php
    $someMock->method('someMethod')->will($this->returnValue(10));
    $someMock->method('someMethod')->will($this->returnSelf());
    $someMock->method('someMethod')->will($this->returnArgument(2));
    $someMock->method('someMethod')->will($this->returnCallback('str_rot13'));
    $someMock->method('someMethod')->will($this->returnValueMap(['a', 'b', 'c', 'd']));
```

Output:
```<?php
    $someMock->method('someMethod')->willReturn(10);
    $someMock->method('someMethod')->willReturnSelf();
    $someMock->method('someMethod')->willReturnArgument(2);
    $someMock->method('someMethod')->willReturnCallback('str_rot13');
    $someMock->method('someMethod')->willReturnMap(['a', 'b', 'c', 'd']);
```
Configuration options available:
```
[
    'returnValue' => true|false,
    'returnSelf' => true|false,
    'returnArgument' => true|false,
    'returnCallback' => true|false,
    'returnValueMap' => true|false,
]
```
where key references to PHPUnit's method name which can be turned on or off along with configuration `bool` value.

Please review this change.


README.rst
```diff
@@ -1038,6 +1038,25 @@ Choose from the list of available rules:
...
+* **php_unit_short_will_return**
+
+  Usage of ``->will($this->returnValue(..))`` statements must be replaced by
+  it's shorter equivalent ``->willReturn(...)``.
+
+  *Risky rule: risky when PHPUnit classes are overridden or not accessible, or when project has PHPUnit incompatibilities.*
+
+  Configuration options:
+
+  - ``returnArgument`` (``false``, ``true``): fix ``returnArgument`` occurrence; defaults
```
keradus: why would one want to fix `returnArgument`, yet not `returnCallback` ?

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,195 @@
...
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Michał Adamski <michal.adamski@gmail.com>
+ */
+final class PhpUnitShortWillReturnFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    const METHOD_RETURN_SELF = 'returnSelf';
```
keradus: please mark with `@internal`

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,195 @@
...
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($this->getSequences() as $method => $sequence) {
+            $this->fixSequence($tokens, $sequence, $method);
+        }
+
+        $tokens->clearEmptyTokens();
+
+        return $tokens->generateCode();
```
keradus: woooow, what's going on here? method must not return anything

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,195 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($this->getSequences() as $method => $sequence) {
+            $this->fixSequence($tokens, $sequence, $method);
+        }
+
+        $tokens->clearEmptyTokens();
```
keradus: why?

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,195 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($this->getSequences() as $method => $sequence) {
+            $this->fixSequence($tokens, $sequence, $method);
```
keradus: basically, you iterate over whole collection 5 times. it;s slow.
iterate only once, by searching for sequence with `[T_STRING]` instead of `[T_STRING, $longerOccurrence]` and then see if given `T_STRING` matches method name you are interested in

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
```diff
@@ -0,0 +1,186 @@
...
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixCases()
+    {
+        return [
+            'will return with integer' => [
```
keradus: we need to add cases with multilines and with comments put in ugly places

keradus:
```php
$someMock
    ->method('someMethod')
        ->willReturn(10)
        ;
```

keradus:
```php
$someMock
    ->method('someMethod')
        ->will(
            // foo
            $this->returnValue(10)
            // bar
        );
```

```php
$someMock->method('someMethod')->/* a */will/* b */(/* c */ $this->returnValue(10) /* d */);
```

SpacePossum:
:+1: for the idea/concept (however the PR does need some tweaking based on the review, let us know if you've questions about it!)

keradus:
ping @kingmad

kingmad:
I'll fix it all as soon as I'll find some time. Thank you!

README.rst
kingmad: I assume that someone could have some of methods overwritten or used in some non-standard way that fixer will break the final output so with this configuration it will be possible to skip such cases. By default it's turned on so IMO there're no contradictions.

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
kingmad: done

README.rst
keradus: even if enabled by default, I still see it as adding a complexity to the fixer and could create same question for end-user.
is this a real case when user do such a things? I would rather call it big misuse of phpunit which we shall not support.

what do you think ?

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
```diff
@@ -0,0 +1,255 @@
...
+    ->method(\'someMethod\')
+        ->will(
+            $this->returnValue(10)
+        );',
+            ],
+            'will return with multi lines, messy indents and comments inside' => [
+                '<?php $someMock
+    ->method(\'someMethod\')
+        ->willReturn(10)
+            // bar
```
keradus: what happened to `// foo` ?

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
```diff
@@ -0,0 +1,255 @@
...
+                '<?php $someMock
+    ->method(\'someMethod\')
+        ->will(
+            // foo
+            $this->returnValue(10)
+            // bar
+        );',
+            ],
+            'will return with block comments in weird places' => [
+                '<?php $someMock->method(\'someMethod\')->/* a */willReturn/* b */(10) /* d */;',
```
keradus: `/* c */` disappeared :(

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
kingmad: That's the tricky part. I assumed that while this part is truncated, it should be removed along with comment. Otherwise there's problem where comment should be moved:
Should it be like:
```
<?php $someMock
    ->method('someMethod')
            // foo
        ->willReturn(10)
            // bar
        ;
```
or
```
<?php $someMock
    ->method('someMethod')
        ->willReturn
// foo
(10)
            // bar
        ;
```
or as I propose removed completly?

Another solution that comes to my mind is to skip all occurrences where comments appears.
Maybe some of you have better ideas how to solve it?

keradus: no, don't skip cases with comments
no, don't remove comments

yeah, it's fine to move comment a little when needed :+1:

keradus:
ping @kingmad

kingmad:
@keradus Sorry for delay... I tried to figure out the solution for comments put in strange places couple times however I am unable to merge different types of comments ie /* */ and // which appears right before/after themselves (mostly the whitespaces between them are problematic). The only possible way which comes to my mind could be some complicated handling edge cases algorithms which IMO are over engineering and I'd rather keep away from such solutions. That's why I put this task away for some time. I could share some examples if anyone want to join or suggest accepting current solution which makes some comments disappeared when placed in problematic place.

README.rst
kingmad: I just thought it could be nice to fix only selected methods not all bundled together - I will not stick to my configurable solution and can remove this part as you wish ;)

keradus:
no worry @kingmad !
Are you able to push those code examples as utests to this PR, even if src is not (yet) handling them ?

kingmad:
sure, I will do it soon

keradus:
ping @kingmad , any chances to finish this one? :)

kingmad:
Sorry to say but as I wrote it before - I'll be thankful if you'll point direction how to solve the problem which was described above. IMO comments put in "uncomfortable" places should be cut out permanently (as I developed it before) otherwise I have no idea how to filter them out without some sophisticated solutions.

SpacePossum:
Hi @kingmad 
I think this is close to being finished, some open things:
- we should utilize the [PHPUnit class detection](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Indicator/PhpUnitTestCaseIndicator.php)
- I _think_ nested calls will not be fixed
- we, @keradus and @julienfalque , should make a call on the options support (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3278#discussion_r154496725) (I would vote +1 for not having the options)
- the complex space and comments cases need to handled

I'm happy to help out with these points. What could be a way forward for the last point is creating the unit tests for these cases without a fix. Just let those fail and commit the test. That way we can help out if you want and tweak your awesome work and get it ready for shipping :)

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,199 @@
...
+
+    const SEQUENCE = [
+        [T_OBJECT_OPERATOR, '->'],
+        [T_STRING, 'will'],
+        '(',
+        [T_VARIABLE, '$this'],
+        [T_OBJECT_OPERATOR, '->'],
+        [T_STRING],
+        '(',
+    ];
```
ntzm: Please add

```
/**
 * @internal
 */
```

To all constants

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,199 @@
...
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Michał Adamski <michal.adamski@gmail.com>
+ *
+ * @internal
```
ntzm: Doesn't need to be marked internal

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,199 @@
...
+        '(',
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Usage of `->will($this->returnValue(..))` statements must be replaced by it\'s shorter equivalent `->willReturn(...)`.',
```
ntzm:
```diff
-by it's shorter
+by its shorter
```

src/Fixer/PhpUnit/PhpUnitShortWillReturnFixer.php
```diff
@@ -0,0 +1,199 @@
...
+                ->setDefault(true)
+                ->getOption(),
+            (new FixerOptionBuilder(static::METHOD_RETURN_CALLBACK, 'Fix `returnCallback` occurrence'))
+                ->setAllowedTypes(['bool'])
+                ->setAllowedValues([false, true])
+                ->setDefault(true)
+                ->getOption(),
+            (new FixerOptionBuilder(static::METHOD_RETURN_VALUE_MAP, 'Fix `returnValueMap` occurrence'))
+                ->setAllowedTypes(['bool'])
+                ->setAllowedValues([false, true])
```
ntzm: `->setAllowedValues([false, true])` doesn't do anything here, as you've already set the allowed types

tests/Fixer/PhpUnit/PhpUnitShortWillReturnFixerTest.php
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\PhpUnit\PhpUnitShortWillReturnFixer
+ */
+final class PhpUnitShortWillReturnFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @param string        $expected
+     * @param string | null $input
```
ntzm: `string|null` instead please

keradus:
@kingmad, please look at last @SpacePossum comment

README.rst
```diff
@@ -1214,6 +1214,25 @@ Choose from the list of available rules:
...
+    to ``true``
+  - ``returnCallback`` (``false``, ``true``): fix ``returnCallback`` occurrence; defaults
+    to ``true``
+  - ``returnSelf`` (``false``, ``true``): fix ``returnSelf`` occurrence; defaults to ``true``
+  - ``returnValue`` (``false``, ``true``): fix ``returnValue`` occurrence; defaults to
+    ``true``
+  - ``returnValueMap`` (``false``, ``true``): fix ``returnValueMap`` occurrence; defaults
```
ntzm: to match with the other options in the project, can you use `snake_case` instead of `camelCase`?

SpacePossum:
replaced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4167

PR #3281, opened 2017-11-27
SelfAccessorFixer - stop modifying traits

kubawerlos:
For trait `self` was referring to it until PHP  5.6.5: https://3v4l.org/8bTQK

It was bug: https://bugs.php.net/bug.php?id=65419
Release log with fix: http://www.php.net/ChangeLog-5.php#5.6.5


tests/Fixer/ClassNotation/SelfAccessorFixerTest.php
```diff
@@ -86,6 +86,10 @@ public function provideFixCases()
                 // PHP < 5.4 compatibility: "self" is not available in closures
                 '<?php class Foo { function bar() { function ($a = Foo::BAZ) { new Foo(); }; } }',
             ),
+            array(
+                // In trait "self" will reference the class it's used in, not the actual trait
```
keradus: "(...) so we can't replace Foo with self here" perhaps ?

keradus:
Thank you @kubawerlos.

PR #3289, opened 2017-12-03
[bug] NoUnusedImportsFixer - import not removed when class name is part of constant

kubawerlos:
When I remove dash it will work fine.

### DIFF
--- a/tests/Fixer/Import/NoUnusedImportsFixerTest.php
+++ b/tests/Fixer/Import/NoUnusedImportsFixerTest.php
@@ -684,6 +684,136 @@ EOF;
         $this->doTest($expected, $input);
     }
 
+    public function testWhenImportedClassNameIsPrefixWithDashOfConstant()
+    {
+        $expected = <<<'EOF'
+<?php
+
+
+class Dummy
+{
+    const C = 'bar-bados';
+}
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
+use Foo\Bar;
+
+class Dummy
+{
+    const C = 'bar-bados';
+}
+EOF;
+
+        $this->doTest($expected, $input);
+    }
+
+    public function testWhenImportedClassNameIsSuffixWithDashOfConstant()
+    {
+        $expected = <<<'EOF'
+<?php
+
+
+class Dummy
+{
+    const C = 'tool-bar';
+}
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
+use Foo\Bar;
+
+class Dummy
+{
+    const C = 'tool-bar';
+}
+EOF;
+
+        $this->doTest($expected, $input);
+    }
+
+    public function testWhenImportedClassNameIsInsideWithDashOfConstant()
+    {
+        $expected = <<<'EOF'
+<?php
+
+
+class Dummy
+{
+    const C = 'tool-bar-bados';
+}
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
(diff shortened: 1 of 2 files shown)


SpacePossum:
Hi,

could you add this to the test:

```php
<?php
use A\ABC;
$a = 5-ABC::Test;
```

kubawerlos:
@SpacePossum not sure why, but done.

SpacePossum:
Ah sorry, combined these pass.
Lemme try again;

```php
    /**
     * @param string $expected
     *
     * @dataProvider provideWhenImportedClassIsUsedForConstantsCases
     */
    public function testWhenImportedClassIsUsedForConstants($expected)
    {
        $this->doTest($expected);
    }

    public function provideWhenImportedClassIsUsedForConstantsCases()
    {
        return array(
            array(
                '<?php
use A\ABC;
$a = 5-ABC::Test;
$a = 5-ABC::Test-5;
$a = ABC::Test-5;
'
            ),
            array(
'<?php
use A\ABC;
$a = 5-ABC::Test;
$a = 5-ABC::Test-5;
'
            ),
            array(
                '<?php
use A\ABC;
$a = 5-ABC::Test;
'
            ),
            array('<?php
use A\ABC;
$a = ABC::Test-5;
'
            ),
            array('<?php
use A\ABC;
$a = 5-ABC::Test-5;
'
            ),
        );
    }
```

These result in 3 failures which I think needs looking into.

kubawerlos:
I have added yet one more case.

SpacePossum:
thanks :)

I start to wonder though, can this be reliable done? And if so, is the `-` exception the only one to consider (what about `$a = 'TEST+a;'`) ?

ping @julienfalque IIRC you had some ideas for this fixer as well.

julienfalque:
IMO using regular expressions **cannot** be reliable enough. This PR is good as it improves the fixer, but I still think we should rewrite it anyway. I'll try to take some time to implement https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2814#issuecomment-307545175 soon.

keradus:
so, @julienfalque , @SpacePossum , please decide is there anything left to do in this PR or it's RTM ?

SpacePossum:
I think the proposed regex solution is not going to be reliable, if others feel confident than I'm happy to re-review.

tests/Fixer/Import/NoUnusedImportsFixerTest.php
```diff
@@ -684,6 +684,136 @@ public function testWithSameNamespaceImportAndUnusedImportAfterNamespaceStatemen
...
+',
+            ),
+            array('<?php
+use A\ABC;
+$a = 5-ABC::Test-5;
+',
+            ),
+            array('<?php
+use A\ABC;
+$b = $a-->ABC::Test;
```
julienfalque: I'm surprised this is valid PHP, what does this line do exactly?

SpacePossum: `$b` will be a boolean, the value is resolved by `$a--` than `>` than `ABC::Test`,
or throwing in some stuff to make it more easy;
`$b = $a-->ABC::Test;`
===
`$b = ($a--) > ABC::Test;`

julienfalque: Thanks :)

julienfalque:
I'm closing this PR as it is included in #3445, thanks!

PR #3290, opened 2017-12-04
Add PhpdocOpeningClosingFixer

Slamdunk:
```diff
-/******
+/*
  * Multiline comment with arbitrary asterisks count
- ******/
+ */

-/**\
+/*\
  * Multiline comment that seems a DocBlock
  */

 /**
  * DocBlock with arbitrary asterisk count at the end
- **/
+ */
```

This fixer fixes both multiline comments and docblock: I've namespaced it under Phpdoc because the main purpose here is to ease the reading and understanding of comments that aren't docblock but can be misundestood as a docblock, like `/*** C */` and `/**\ C */`.

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -32,6 +32,7 @@ $config = PhpCsFixer\Config::create()
         'list_syntax' => ['syntax' => 'long'],
         'method_chaining_indentation' => true,
         'method_argument_space' => ['ensure_fully_multiline' => true],
+        'multiline_comment_opening_closing' => true,
         'no_extra_blank_lines' => ['tokens' => ['break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block']],
         'no_null_property_initialization' => true,
         'no_short_echo_tag' => true,
--- a/README.rst
+++ b/README.rst
@@ -1207,6 +1207,12 @@ Choose from the list of available rules:
 
   Classy that does not inherit must not have inheritdoc tags.
 
+* **phpdoc_opening_closing**
+
+  DocBlocks must start with two asterisks, multiline comments must start
+  with a single asterisk, after the opening slash. Both must end with a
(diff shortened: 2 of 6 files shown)


Slamdunk:
@SpacePossum the backslash in the code sample seems be broking the colouring:
```
$ php php-cs-fixer describe phpdoc_opening_closing
Description of phpdoc_opening_closing rule.
DocBlocks must start with two asterisks, multiline comments must start with a single asterisk, after the opening slash. Both must end with a single asterisk before the closing slash.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,13 +1,13 @@
    <?php

   -/******
   +/*
     * Multiline comment with arbitrary asterisks count
   - ******/
   + */

   -/**</fg=red>
   +/*</fg=green>
     * Multiline comment that seems a DocBlock
     */

    /**
     * DocBlock with arbitrary asterisk count at the end
   - **/
   + */
```

julienfalque:
:+1: 

> I've namespaced it under Phpdoc because the main purpose here is to ease the reading and understanding of comments that aren't docblock but can be misundestood as a docblock

I would still move this fixer to the `Comment` namespace as it does not only deal with PhpDoc comments.

Slamdunk:
> I would still move this fixer to the Comment namespace as it does not only deal with PhpDoc comments.

It deals also with PhpDoc comments:
```diff
 /**
  * DocBlock with arbitrary asterisk count at the end
- **/
+ */
```

julienfalque:
That's what I meant :p

Happy to hear more opinions about it though.

keradus:
shall be in `Comment` namespace indeed

Slamdunk:
- [x] `Phpdoc` -> `Comment`
- [x] `PhpdocOpeningClosingFixer` -> `MultilineCommentOpeningClosingFixer`

julienfalque:
Shall we add it to the project config?

keradus:
Thank you @Slamdunk.

PR #3292, opened 2017-12-05
PHPUnit - set memory limit

veewee:
This PR removes the circle-ci memory issue. (As discussed in #3276)

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -47,6 +47,7 @@
 
     <php>
         <ini name="zend.enable_gc" value="0"/>
+        <ini name="memory_limit" value="1G"/>
         <env name="SKIP_LINT_TEST_CASES" value="0"/>
     </php>
 </phpunit>


keradus:
would it work if we remove config from global env setting and put it in phpunit config file?

phpunit.xml.dist
```diff
@@ -47,6 +47,7 @@
 
     <php>
         <ini name="zend.enable_gc" value="0"/>
+        <ini name="memory_limit" value="-1"/>
```
keradus: I would except to see failure if our test would pass but used 20GB RAM.
so maybe some real value like 1GB or whatever is needed rather than no limit at all?

veewee: Personally, I don't care about memory usage for cli tools.

Maybe it's better to set a max memory usage per fixer and display an error message when the memory exceeds this threshold? This way you also directly know which test is consuming a lot of memory.

Currently this is already been done for slow tests:
`You should really fix these slow tests (>100ms)...`


This could be implemented in a different PR.

SpacePossum: Does such test listener exists? i.e. "memory usage listener" similar to https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.8/phpunit.xml.dist#L36 ?

I made a mistake a few PR's back were I blew up memory usage and this was only detected by the utest on Travis.
Therefore I agree with @keradus here, maybe better to put it on something realistic.

veewee: This listener exists, but I am not sure how good it will work:
https://github.com/atrapalo/phpunit-memory-and-time-usage-listener

I'll change the config for now

keradus:
Thank you @veewee.

PR #3293, opened 2017-12-05
Make alignment of multi-line descriptions configurable

dontub:
This change allows to configure the alignment of multi-line descriptions in the phpdoc_align fixer. Especially useful for `@param` annotations if type hint and variable name are relatively long to avoid very long lines or lots of description lines.

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocAlignFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocAlignFixer.php
@@ -35,6 +35,8 @@ final class PhpdocAlignFixer extends AbstractFixer implements ConfigurationDefin
 {
     private $regex;
     private $regexCommentLine;
+    private $descriptionAlign;
+    private $descriptionExtraIndent;
 
     private static $alignableTags = [
         'param',
@@ -78,6 +80,8 @@ final class PhpdocAlignFixer extends AbstractFixer implements ConfigurationDefin
 
         $this->regex = '/^'.$indent.' \* @(?:'.$tagsWithName.'|'.$tagsWithoutName.'|'.$tagsWithMethodSignature.')'.$desc.'\s*$/u';
         $this->regexCommentLine = '/^'.$indent.' \*(?! @)(?:\s+(?P<desc>\V+))(?<!\*\/)$/u';
+        $this->descriptionAlign = $this->configuration['description_align'];
+        $this->descriptionExtraIndent = $this->configuration['description_extra_indent'];
     }
 
     /**
@@ -158,7 +162,29 @@ final class PhpdocAlignFixer extends AbstractFixer implements ConfigurationDefin
             ])
         ;
 
-        return new FixerConfigurationResolver([$tags->getOption()]);
+        $descriptionAlign = new FixerOptionBuilder('description_align', 'The alignment of a description running over multiple lines.');
+        $descriptionAlign
+            ->setAllowedTypes(['string'])
+            ->setAllowedValues([
+                'tag',
+                'hint',
+                'name',
+                'description',
+            ])
+            ->setDefault('description')
+        ;
+
+        $descriptionExtraIndent = new FixerOptionBuilder('description_extra_indent', 'Extra indent for a description running over multiple lines.');
+        $descriptionExtraIndent
+            ->setAllowedTypes(['int'])
+            ->setDefault(0)
+        ;
+
+        return new FixerConfigurationResolver([
+            $tags->getOption(),
+            $descriptionAlign->getOption(),
+            $descriptionExtraIndent->getOption(),
+        ]);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SpacePossum:
Hi and thanks for sending this PR!
I've a few questions on concept level and there is a SCA thingy that the CI tripped on (we don't allow `switch` constructions in the code base).
About the new options, will these with their defaults behave the same as the fixer currently does?

keradus:
also, is there some well-established standard / community that would benefit from that option ?

dontub:
With the default options the fixer behaves the same as the fixer currently does.

> also, is there some well-established standard / community that would benefit from that option ?

At least indentation by 1, 2, or 4 spaces I've found in different style guides. though there is no alignment of names or type hints:
https://www.drupal.org/docs/develop/coding-standards/api-documentation-and-comment-standards
https://www.mediawiki.org/wiki/Manual:Coding_conventions/PHP
https://www.mediawiki.org/wiki/Manual:Coding_conventions/JavaScript#Documentation
https://google.github.io/styleguide/javascriptguide.xml?showone=Comments#Comments
https://google.github.io/styleguide/javaguide.html#s7-javadoc

At least here it is suggested to align everything and add add an extra space:
https://en.wikibooks.org/wiki/PHP_Programming/Coding_Standards#Sample_File

I'll replace the `switch` if you agree in general with this PR.

keradus:
fact that there are indeed established standards for using 2 spaces for typescript and so has nothing to do with PHP.
but ok, even in 3 PHP links you provided there are different way.

:+1: for idea in that case

dontub:
The `switch` construct is replaced.

The first links where just to show that there exists style guides with only space indent on multi-line (`description_align = tag, description_extra_indent = n`). This would combine alignment with indentation of `n` spaces on multi-line comments, so yes it's not exactly the same as in those guides.

The final example could be achieved (apart from the double space after the type hint) with `description_extra_indent = 1`.

dontub:
Is there something keeping you off from merging?

SpacePossum:
hi @dontub and sorry for the wait, we are trying to catch up with all the PR's and issues,
hope to review this one soon

src/Fixer/Phpdoc/PhpdocAlignFixer.php
```diff
@@ -231,16 +257,22 @@ private function fixDocBlock($content)
...
+                            $indent += $varMax + 1;
+                        }
                     }
 
                     $line =
                         $item['indent']
-                        .' *  '
-                        .str_repeat(' ', $tagMax + $hintMax + $varMax + $extraIndent)
+                        .' * '
+                        .str_repeat(' ', max(0, $indent))
```
julienfalque: `max()` should be applied to `$this->descriptionExtraIndent` a few lines above as it is the only variable that can be negative:
```php
$indent = max(0, $this->descriptionExtraIndent);
```
This will prevent weird cases like e.g.:
```
$this->descriptionExtraIndent (-4) + $tagMax (6) + 2 = 4 instead of 8
```

Maybe even better would be to validate the `description_extra_indent` option instead.

Please also add a test case that covers this.

tests/Fixer/Phpdoc/PhpdocAlignFixerTest.php
```diff
@@ -920,4 +920,119 @@ public function provideInvalidPhpdocCases()
...
+    }
+
+    public function testDescriptionAlignHint()
+    {
+        $this->fixer->configure(['description_align' => 'hint']);
+
+        $expected = <<<'EOF'
+<?php
+    /**
+     * @param int $a Desc
```
julienfalque: Please add cases with e.g. `@method`, `@return`, ... and some cases with mixed tags. Same for other tests below.

dontub:
Thanks for the feedback. I've found some time today to look at it again. I'm wondering how this example should be aligned by default:
```
/**
 * @param int $foobar Desc
 * ription
 * @return int Desc
 * ription
 */
```
This is what my code does at the moment:
```
/**
 * @param  int $foobar Desc
 *                     ription
 * @return int Desc
 *                     ription
 */
```

This is what PHP-CS-Fixer does at the moment (one space less in the last description line):
```
/**
 * @param  int $foobar Desc
 *                     ription
 * @return int Desc
 *                    ription
 */
```

This is what I would expect:
```
/**
 * @param  int $foobar Desc
 *                     ription
 * @return int         Desc
 *                     ription
 */
```

Shall I change this PR regarding to this issue?

src/Fixer/Phpdoc/PhpdocAlignFixer.php
dontub: Actually a negative extra indent makes no sense to me as well, but do we need to forbid it?

dontub:
@julienfalque have you seen my questions in the comments above?

dontub:
@SpacePossum @julienfalque I'd like to see this merged. Though I'm wondering if I should take the time to resolve the conflicts again because there was no response since the last changes...

shakaran:
@dontub please, could you reupdate and solve the conflicts? Could be a chance to be merged in future releases and maintainer will probably comment about your last suggestion. Thanks for your effort

dontub:
@shakaran: I would resolve conflicts once again, if a maintainer shows interest in merging this. Otherwise I think it's wasted time.

PR #3297, opened 2017-12-05
PhpCsFixer\Config::getFinder() can return PhpCsFixer\Finder

SignpostMarv:
since `PhpCsFixer\Config::getFinder()` will return an instance of `PhpCsFixer\Finder` by default, it'd be beneficial to type-hint in the docblock that this is the case :)

### DIFF
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -37,7 +37,7 @@ interface ConfigInterface
     /**
      * Returns files to scan.
      *
-     * @return iterable|string[]|\Traversable|Finder
+     * @return Finder|iterable|string[]|\Traversable
      */
     public function getFinder();
 
@@ -121,7 +121,7 @@ interface ConfigInterface
     public function setCacheFile($cacheFile);
 
     /**
-     * @param iterable|string[]|\Traversable|Finder $finder
+     * @param Finder|iterable|string[]|\Traversable $finder
      *
      * @return self
      */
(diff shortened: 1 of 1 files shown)


SpacePossum:
Hi and thank you for your PR!

~~The additional type hint you suggest here seems overcomplete to me.
As `Finder` extends [`\IteratorAggregate`](https://github.com/symfony/finder/blob/2.0/Finder.php#L29) which extends [`Traversable`](http://php.net/manual/en/class.iteratoraggregate.php) which is what is now hinted as return type. I think the current type hinting is correct as we don't want to rely on the `Finder` implementation~~ (, besides if should hint `Finder` we would've to do it as FQCN or import it).

I just read up on the issue over @ phpstan, isn't this resolved at their end with https://github.com/phpstan/phpstan/pull/650 ?

SignpostMarv:
@SpacePossum phpstan/phpstan#650 only resolves the issue when Finder is explicitly listed as a return/param type.

SignpostMarv:
p.s. this came up because of the [php-cs-fixer config package I use](https://github.com/SignpostMarv/PHP-CS-Fixer-Config/blob/e9d86b06fab9ad6af47ea223083192cf156ff976/src/Config.php#L101), where I'm calling `Config::setFinder()` after passing `$finder->ignoreUnreadableDirs()` through `array_reduce()`

SpacePossum:
alrighty than, I'm leaning to +1 here,
however the type hint should still ref. the namespace of Finder, therefore I would pref. importing the Finder with its namespace.

SignpostMarv:
closing this one in favour of #3301 since I'm using 2.8 myself.

PR #3299, opened 2017-12-06
SingleQuoteFixer - fix single quote char

Slamdunk:
- [x] Add configuration for fixing strings containing single-quotes, disabled by default
- [x] Create integration test with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3155

### DIFF
--- a/src/Fixer/StringNotation/SingleQuoteFixer.php
+++ b/src/Fixer/StringNotation/SingleQuoteFixer.php
@@ -73,7 +73,7 @@ EOF;
 
             if (
                 '"' === $content[0] &&
-                (false === strpos($content, "'") || true === $this->configuration['strings_containing_single_quote_chars']) &&
+                (true === $this->configuration['strings_containing_single_quote_chars'] || false === strpos($content, "'")) &&
                 // regex: odd number of backslashes, not followed by double quote or dollar
                 !preg_match('/(?<!\\\\)(?:\\\\{2})*\\\\(?!["$\\\\])/', $content)
             ) {
--- a/src/Fixer/StringNotation/SingleQuoteFixer.php
+++ b/src/Fixer/StringNotation/SingleQuoteFixer.php
@@ -13,6 +13,9 @@
 namespace PhpCsFixer\Fixer\StringNotation;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Token;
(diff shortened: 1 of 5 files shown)


src/Fixer/LanguageConstruct/ClassKeywordRemoveFixer.php
```diff
@@ -219,7 +219,7 @@ private function replaceClassKeyword(Tokens $tokens, $classIndex)
 
         $tokens->insertAt($classBeginIndex, new Token(array(
             T_CONSTANT_ENCAPSED_STRING,
-            "'".$this->makeClassFQN($classImport, $classString)."'",
+            '\''.$this->makeClassFQN($classImport, $classString).'\'',
```
keradus: :-1: for this changes

Slamdunk: You mean only `"'"` or the entire PR?

keradus: we follow Sf rule here:
use single-quoted string except of cases when using it would require extra escaping

Slamdunk: If so, I'm going to make this configurable and disabled by default.

keradus: for which fixer?

Slamdunk: `SingleQuoteFixer`: I'll rebase this PR on master instead of 2.2, and create a configuration whether to fix or not double-quoted strings that contain single-quotes; disabled by default for Symfony and BC compatibility

Slamdunk:
I've rebased to master since it's a new functionality, added the new configuration, made it backward compatible, created integration tests with #3155 

As often happens to me, I'm confident with the code, but the name of the new configuration it's a bit raw: `strings_containing_single_quote_chars`

src/Fixer/StringNotation/SingleQuoteFixer.php
```diff
@@ -21,16 +24,30 @@
...
  */
-final class SingleQuoteFixer extends AbstractFixer
+final class SingleQuoteFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
      * {@inheritdoc}
      */
     public function getDefinition()
     {
+        $codeSamble = <<<'EOF'
```
SpacePossum: little typo in the var name

src/Fixer/StringNotation/SingleQuoteFixer.php
```diff
@@ -56,14 +73,27 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
 
             if (
                 '"' === $content[0] &&
-                false === strpos($content, "'") &&
+                (false === strpos($content, "'") || true === $this->configuration['strings_containing_single_quote_chars']) &&
```
SpacePossum: please swap the tests, the config check is cheaper than the `strpos`

src/Fixer/StringNotation/SingleQuoteFixer.php
```diff
@@ -56,14 +73,27 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
                 // regex: odd number of backslashes, not followed by double quote or dollar
                 !preg_match('/(?<!\\\\)(?:\\\\{2})*\\\\(?!["$\\\\])/', $content)
             ) {
                 $content = substr($content, 1, -1);
-                $content = str_replace(['\\"', '\\$'], ['"', '$'], $content);
+                $content = str_replace(['\\"', '\\$', '\''], ['"', '$', '\\\''], $content);
```
SpacePossum: isn't this a BC break, as it always does this change no matter what has been configured?

Slamdunk: No because the code enters here only if the configuration is enabled.

src/Fixer/StringNotation/SingleQuoteFixer.php
Slamdunk: Thanks

src/Fixer/StringNotation/SingleQuoteFixer.php
Slamdunk: Done

keradus:
Thank you @Slamdunk.

PR #3301, opened 2017-12-06
companion to FriendsOfPHP/PHP-CS-Fixer#3297, type hinting that \PhpCs…

SignpostMarv:
companion to FriendsOfPHP/PHP-CS-Fixer#3297, type hinting that \PhpCsFixer\Config::getFinder() can return \PhpCsFixer\Fixer

PR #3305, opened 2017-12-07
type hinting that \PhpCsFixer\Config::getFinder() can return \PhpCsFi…

SignpostMarv:
type hinting that \PhpCsFixer\Config::getFinder() can return \PhpCsFixer\Fixer

### DIFF
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -37,7 +37,7 @@ interface ConfigInterface
     /**
      * Returns files to scan.
      *
-     * @return iterable|string[]|\Traversable
+     * @return iterable|\PhpCsFixer\Finder|string[]|\Traversable
      */
     public function getFinder();
 
@@ -121,7 +121,7 @@ interface ConfigInterface
     public function setCacheFile($cacheFile);
 
     /**
-     * @param iterable|string[]|\Traversable $finder
+     * @param iterable|\PhpCsFixer\Finder|string[]|\Traversable $finder
      *
      * @return self
      */


SpacePossum:
The SCA failure on Travis seems unrelated to this PR. Could you rebase your branch to the latest of 2.8 to see if it helps?

keradus:
it was for 2.8 : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3301 
I aksed for 2.2, as if we would accept this change, then only for 2.2 target branch

SignpostMarv:
Does this still need rebasing?

keradus:
Don't bother with rebases here.
What is needed here is not to fix docblock to match fuzzy logic of phpstan, but to actually investigate the code do we really support both, collection of SplFileInfo and strings equally or not, and if so why, maybe with possibility to improve it.

SignpostMarv:
@keradus tl:dr; "is finder overkill?" ?

keradus:
finder is wrapper for iterable and adding finder here is solving an issue just for fuzzy phpstan logic due to issues fully discussed in referred issue.

SplFileInfo vs string case could be a thing that needs investigation indeed, and side effect of it could be solving in-place confusion of phpstan as well.

SignpostMarv:
@keradus in faffing around with my package trying to get a workaround, have discovered that 2.9 doesn't support `string[]` as per documentation of `PhpCsFixer\ConfigInterface::setFinder()`, but does support `SplFileInfo[]`

keradus:
yet, `string[]` might be supported if one provides a pure array, not Finder instance.
as said, I need to take deeper look to know shall we merge your docs update or drop string in first place -
 it's one of oldest code in out project.

I must admit it's not a prio for me on end-of-the-year time. Please be patient.
no need for further actions from your side here (like merges/rebases)

SignpostMarv:
@keradus When providing the pure `string[]` array. `FileFilterIterator` line 56 assumes `SplFileInfo[]`, leading to a fatal error.

keradus:
Thanks for checking @SignpostMarv . Removal of wrong `string[]` in docs has been done in #3415, thus I'm closing this PR.

Issue #3304, opened 2017-12-07
Skip non-UTF8 files/tokens ?


Slamdunk:
After https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3272 was released a lot of my CI builds on legacy apps started failing because sources are non-UTF8 and preg functions and/or this project code raise exceptions on something like:
```php
/**
 * NON-UTF8 file ending with this char: à
 */
```
After https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3291 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3296 I'm expecting this issue to occur with more frequency.

The issue is not with PHP-CS-Fixer since it's ok to expect UTF8 sources, but neither I can't have CI build failing for valid code.

We can't change file's encoding.

May I open a PR that adds code like https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.3/src/Runner/FileFilterIterator.php#L70-L86 to skip in a non-blocking way code that can't be properly fixed (i.e. that aren't UTF8)?

keradus:
skipping whole file is no-go.
instead, focus on tracking down where is given exception you experience raised and we could think how to fix it instead of skipping it.

Slamdunk:
I'm PRing soon a bunch of test-cases.

I have no idea how we could fix sources in different charsets without rewriting a lot of code. I think we'll need something like `WhitespacesAwareFixerInterface`: the Fixer should ask/get-configured (when needed) in order to handle efficiently token contents in different charsets.

keradus:
starting with PR adding failing case would be a great start. having that, we will figure out how to handle it nicely.

keradus:
Closing by merge of #3446

PR #3306, opened 2017-12-07
Token - better input validation

keradus:
discovered during #3304

### DIFF
--- a/tests/Tokenizer/TokenTest.php
+++ b/tests/Tokenizer/TokenTest.php
@@ -46,6 +46,34 @@ final class TokenTest extends TestCase
         return $prototype;
     }
 
+    /**
+     * @param mixed $input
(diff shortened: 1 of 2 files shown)


tests/Tokenizer/TokenTest.php
```diff
@@ -46,6 +46,34 @@ public function getForeachTokenPrototype()
...
     }
 
+    /**
+     * @param mixed $input
+     *
+     * @dataProvider provideConstructorValidationCases
+     */
+    public function testConstructorValidation($input)
+    {
+        $this->setExpectedException(get_class(new \InvalidArgumentException()));
```
Slamdunk: Why not `\InvalidArgumentException::class`?
https://3v4l.org/9SShI

PR #3296, opened 2017-12-05
Line break fix in RegExes

kubawerlos:
ref #3291

PR #3272, opened 2017-11-20
PhpdocTrimFixer - unicode support

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3271

PR #3309, opened 2017-12-08
Add non-UTF8 test cases

Slamdunk:
Reference: #3304

- [ ] Add test cases for all charset-aware fixers in 2.2 (waiting #3296)
- [ ] Add test cases for all charset-aware fixers new in 2.8 (different PR required)
- [ ] None of the above if the charset-aware part will be a wrapping part of fixers

As written in #3304 I have tons of code in different charsets that can in theory be well managed and fixed by this library if it will became charset-aware.

Relying on `/u` modifier of preg functions doesn't solve the issue (#3296 #3272), even though it is now more likely to encounter UTF8 files than non-UTF8 files.

A big hurdle here is that files don't declare their charset.

Making fixers charset-aware would be a nightmare: too much code to test and be aware of.

The first solution that comes into my mind is to let [Runner\Runner::fixFile](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.8/src/Runner/Runner.php#L149) do the dirty job:

1. Guess file charset right after `file_get_contents`, or (better) we can give the possibility to the user to set the charset of all sources during configuration, instead of guessing it
1. If not UTF8 convert it into UTF8
1. Keep track of original charset
1. Generate tokens and run fixers in UTF8-only
1. Generate code in UTF8
1. Re-convert to original charset just before `file_put_contents`

Requiring `iconv` and/or `mbstring` will be mandatory 😐 

Well, we would also need to:

1. Wrap every preg function to ensure `/u` flag
1. Write an AutoReview that ensure every preg function is wrapped (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3296#issuecomment-349591603)
1. Move every non-byte safe function to multibyte-safe ([MbStrFunctionsFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.8.3/src/Fixer/Alias/MbStrFunctionsFixer.php))

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocTrimFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocTrimFixerTest.php
@@ -59,6 +59,17 @@ function deactivateCompleted()
     return 0;
 }',
             ),
+            array(
+                mb_convert_encoding('
+<?php
+
+/**
+ * Test à
+ */
+function foo(){}
+
+', 'Windows-1252', 'UTF-8'),
+            ),
         );
     }
 


keradus:
Thanks for raising the case.
Please don't start fixing it before we will discuss and align about solution.

Your proposal looks quite a big for quite small use case.

One of the steps you proposed is to convert to UTF-8. Why not build a fixer for doing it, actually?
In your concrete case, you just have legacy codebase in non-UTF, or it's by purpose? Would you be fine with storing files as UTF ?

Slamdunk:
> One of the steps you proposed is to convert to UTF-8. Why not build a fixer for doing it, actually?
> In your concrete case, you just have legacy codebase in non-UTF, or it's by purpose? Would you be fine with storing files as UTF ?

This can't absolutely be done:

```php
// File stored as Windows-1252
header('Content-Type: text/html; charset=Windows-1252');
echo '<h1> à </h1>';
```

If this file gets converted to UTF-8, the website gets broken.

> Please don't start fixing it before we will discuss and align about solution.

Of course

> Your proposal looks quite a big for quite small use case.

I know. Charsets-world is a kraken 😢

Slamdunk:
The work to be done to solve this issue ~is~ may be a lot, I would like to hear all main contributors thoughts.

Ping @SpacePossum @julienfalque @localheinz

keradus:
I think we went into trap of `json_serialize`. ~nobody ever checks if it failed or not.
here, we were running preg_... without checking if result was successful.

For me checking the charset and then converting it back and forth is a lot of cpu wasted for a very little edge case.

If we would properly check errors of preg_... functions, we would probably be enough. (even if not fixed the edge case of non-utf charset, then at least not crashing at it).

-------------

indeed, more feedback is demanded.

langj:
> Well, we would also need to:
>
> 1. Wrap every preg function to ensure /u flag
> 2. Write an AutoReview that ensure every preg function is wrapped (#3296 (comment))
> 3. Move every non-byte safe function to multibyte-safe (MbStrFunctionsFixer)

Thank you for your proposal. Those steps you mentioned are also necessary to support just utf8, right?

Slamdunk:
> we were running preg_... without checking if result was successful.

I think about

1. an AutoReview that asserts every preg call must be a `PhpCsFixer\PregWrap` (or something)
1. `PhpCsFixer\PregWrap` does nothing in production except proxing the calls
1. `PhpCsFixer\PregWrap` under SUT checks that regexps have `/u` modifier

Easy to implement and maintain, zero forgetfulness

> For me checking the charset and then converting it back and forth is a lot of cpu wasted for a very little edge case.

One time per file, and only if the user specify a charset different from UTF-8, that would be the default of this tool.

> If we would properly check errors of preg_... functions, we would probably be enough.

Well, opting out all non-UTF8 projects? :\

> Those steps you mentioned are also necessary to support just utf8, right?

Yup

keradus:
> One time per file, and only if the user specify a charset different from UTF-8, that would be the default of this tool.

checking the charset is expensive

> Well, opting out all non-UTF8 projects? :\

no, fixer would still work, just given rule would not applied one particular change.

in the end, yes, I would rather promote legacy codebases to change to UTF-8 than start officially supporting non-UTFs, it's endless well.

keradus:
replaced by already merged #3446
thank you @Slamdunk !

PR #3311, opened 2017-12-08
FinalInternalClassFixer - fix typo

localheinz:
This PR

* [x] fixes a typo

💁‍♂️ It should be *case insensitive*, rather than *case in sensitive*.


keradus:
Thank you @localheinz.

localheinz:
Thank you, @keradus!

PR #3312, opened 2017-12-09
PhpUnitTestAnnotationFixer: Only remove prefix if it really is a prefix

localheinz:
This PR

* [x] adds more test cases
* [x] modifies detection of test prefix
* [x] removes `test` prefix only if the next character is not a number, nor a lower-case letter
* [x] removes a few related comments

Follows #3275.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -314,19 +314,13 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
      */
     private function removeTestFromFunctionName($functionName)
     {
-        if ($this->startsWith('test_', $functionName)) {
-            if (is_numeric(substr($functionName, 6, 1))) {
-                return $functionName;
-            }
+        $remainder = preg_replace('/^test_?/', '', $functionName);
 
-            return  substr($functionName, 5);
-        }
-        if (is_numeric(substr($functionName, 5, 1))) {
+        if ('' === $remainder || is_numeric($remainder[0])) {
             return $functionName;
         }
-        $functionName = substr($functionName, 4);
 
-        return lcfirst($functionName);
+        return lcfirst($remainder);
     }
 
     /**
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -88,7 +88,7 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
             if ('annotation' === $this->configuration['style']) {
                 $this->applyTestAnnotation($tokens, $indexes[0], $indexes[1]);
             } else {
-                $this->removeTestAnnotation($tokens, $indexes[0], $indexes[1]);
+                $this->applyTestPrefix($tokens, $indexes[0], $indexes[1]);
             }
         }
     }
@@ -140,7 +140,7 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
     private function applyTestAnnotation(Tokens $tokens, $startIndex, $endIndex)
     {
         for ($i = $endIndex - 1; $i > $startIndex; --$i) {
-            if (!$this->isFunctionTest($tokens, $i)) {
+            if (!$this->isTestMethod($tokens, $i)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -250,6 +250,56 @@ private function isFunction(Tokens $tokens, $index)
...
+    /**
+     * @param string $functionName
+     *
+     * @return bool
+     */
+    private function canRemoveTestPrefix($functionName)
+    {
+        $length = strlen($functionName);
+
+        if ($this->startsWith('test_', $functionName) && $length > 5) {
```
BackEndTea: Maybe add a test case to prove a function just being named ```test```or ```test_``` does not get renamed either

localheinz: @BackEndTea 

Good point, will take a look later!

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -291,26 +291,36 @@ private function getDockBlockIndex(Tokens $tokens, $index)
...
+     */
+    private function hasTestPrefix($functionName)
+    {
+        if (!$this->startsWith('test', $functionName) || strlen($functionName) < 5) {
+            return false;
+        }
+
+        $nextCharacter = $functionName[4];
+
+        return is_numeric($nextCharacter) || $nextCharacter === strtoupper($nextCharacter);
```
localheinz: The second part covers also cases where we have `test_` as prefix, as `'_' === strotoupper('_')`.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -291,6 +291,22 @@ private function getDockBlockIndex(Tokens $tokens, $index)
...
+     */
+    private function hasTestPrefix($functionName)
+    {
+        if (!$this->startsWith('test', $functionName) || strlen($functionName) < 5) {
+            return false;
+        }
+
+        $nextCharacter = $functionName[4];
+
+        return $nextCharacter === strtoupper($nextCharacter);
```
localheinz: If we preferred to be explicit here, then this should be

```php
return is_numeric($nextCharacter) || '_' === $nextCharacter || $nextCharacter === strtoupper($nextCharacter);
```

instead.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -291,26 +291,36 @@ private function getDockBlockIndex(Tokens $tokens, $index)
...
-        if ($this->startsWith('test_', $functionName)) {
-            if (is_numeric(substr($functionName, 6, 1))) {
-                return $functionName;
-            }
+        $remainder = preg_replace('/^test_?/', '', $functionName);
 
-            return  substr($functionName, 5);
-        }
-        if (is_numeric(substr($functionName, 5, 1))) {
+        if ('' === $remainder || is_numeric($remainder[0])) {
```
localheinz: Can't remove the prefix if the remainder is an empty string, or the first character is a digit.

SpacePossum:
LGTM, sadly the test fails on a not related issue (`echo: write error: Broken pipe`)

localheinz:
Thank you, @SpacePossum! Noticed this locally as well, but focused only on `PhpUnitTestAnnotation` fixer here.

localheinz:
@SpacePossum 

Initially I had three methods

- `hasTestPrefix()`
- `canRemoveTestPrefix()`
- `removeTestPrefix()`

What do you think, would the code be easier to understand if there was a single method to handle all of that? Similarly, for adding a test prefix, if necessary? After all, there’s some temporal coupling involved, as we always invoke `hasTestPrefix()`, immediately followed by `removeTestPrefix()`.

keradus:
then `removeTestPrefix` would need to check `hasTestPrefix` internally to decouple the methods.
if there is a need to call one method but not another (even conditionally), then 2 methods are fine.
if we always call them in chain (even conditionally), I don't see value of splitting them

localheinz:
@keradus @SpacePossum 

On second thought, we use `hasTestPrefix()` in two different scenarios:

* when verifying that we need to remove
* when verifying that we need to add

So I guess it's ok to leave them separate.

How about a bit of renaming, though? Please hang on for the next commit.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -140,29 +140,29 @@ private function findPhpUnitClasses(Tokens $tokens)
...
             $functionNameIndex = $tokens->getNextMeaningfulToken($i);
             $functionName = $tokens[$functionNameIndex]->getContent();
 
             if ($this->hasTestPrefix($functionName)) {
-                $newFunctionName = $this->removeTestFromFunctionName($functionName);
+                $newFunctionName = $this->removeTestPrefix($functionName);
                 $tokens[$functionNameIndex] = new Token([T_STRING, $newFunctionName]);
             }
 
-            $docBlockIndex = $this->getDockBlockIndex($tokens, $i);
```
localheinz: Here's a small typo in the method name.

SpacePossum:
the current PR'ed composition of methods and their naming looks spot on to me :+1:

localheinz:
@BackEndTea @keradus @SpacePossum 

I added one more test case, but other than that, I can't think of anything else to add at the moment.

BackEndTea:
Thanks for improving my code a bit @localheinz

keradus:
Somehow I missed that.
@localheinz , please rebase so I could merge it

localheinz:
@keradus 

Rebased!

keradus:
Thank you @localheinz.

localheinz:
Thank you, @BackEndTea, @keradus, and @SpacePossum!

PR #3318, opened 2017-12-12
SingleLineCommentStyleFixer - fix closing tag inside comment causes an error

### LINKED ISSUE #3315: Closing tag inside comment causes cs fixer to throw error
I seem to have found a bug with handling of comments. The following file will cause an "errors reported during linting after fixing" error:

```php
<?php
function foo() {
    /*?>*/
    return 'bar';
}
```
The issue seems to be with comments starting with `/*` (ie a single asterisk after the slash). Changing the start to `/**` fixes the problem.

The error persists if the comment is changed to either of the following:
```php
/*
?>
*/

/*
 * ?>
 */
```

The `single_line_comment_style` option seems to cause the error because (I'm guessing) it changes the comment to `// ?>` which PHP interprets as a comment followed by a real closing tag.

I'm sorry I'm not familiar enough with cs fixer to fully track down the problem and suggest a fix.

`/C/php/php.exe /C/php/php-cs-fixer.phar fix -n --verbose --config=php_cs.conf test.php`

```php
<?php

    ->setRules([
        'single_line_comment_style' => ['comment_types' => ['asterisk']],
    ])
;
```

#### Actual result
Files that were not fixed due to errors reported during linting after fixing:
 ... test.php

#### Expected result
```php
<?php

function foo() 
{
    /*?>*/
    return 'bar';
}
```

kubawerlos:
closes #3315

### DIFF
--- a/src/Fixer/Comment/SingleLineCommentStyleFixer.php
+++ b/src/Fixer/Comment/SingleLineCommentStyleFixer.php
@@ -122,7 +122,13 @@ $c = 3;
 
                 continue;
             }
-            if (!$this->asteriskEnabled || '/*' !== substr($content, 0, 2) || 1 === preg_match('/[^\s\*].*\R.*[^\s\*]/s', $commentContent)) {
+
+            if (
+                !$this->asteriskEnabled
+                || false !== strpos($commentContent, '?>')
+                || '/*' !== substr($content, 0, 2)
+                || 1 === preg_match('/[^\s\*].*\R.*[^\s\*]/s', $commentContent)
+            ) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


keradus:
why not 2.2 ?

kubawerlos:
This fixers does not exist in 2.2

keradus:
@SpacePossum 
please help with review here ;)

src/Fixer/Comment/SingleLineCommentStyleFixer.php
```diff
@@ -122,6 +122,11 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
 
                 continue;
             }
+
+            if (false !== strpos($commentContent, '?>')) {
```
SpacePossum: I would merge this in the check below, after the "cheaper" checks (like `!$this->asteriskEnabled`),
for reference: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3320/files

GrahamCampbell:
👍

keradus:
Thank you @kubawerlos.

PR #3319, opened 2017-12-12
Do not remove imports from GLOBAL namespace

henriquemoody:
When importing functions or constants from the GLOBAL namespace, they
are not  considered as used, because the fixer consider ` function` and
` const` as part of the FQN of the use statement.

### DIFF
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -171,6 +171,10 @@ final class NoUnusedImportsFixer extends AbstractFixer
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


SpacePossum:
Hi and thanks for PR'ing!

As this is a bug we would like to provide your fix on the lowest maintained branch, so could you retarget your PR to 2.2?
I was wondering, does this bug also effect;

```php

<?php
// importing a constant (PHP 5.6+)
use const A\B\SomeThing;

echo SomeThing;

```

?

henriquemoody:
This bug only affects imports from the GLOBAL namespace.

PR #3320, opened 2017-12-13
SingleLineCommentStyleFixer - do not transform asterisk comment if it…

SpacePossum:
… contains a PHP closing tag.

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3315


keradus:
do we need it that comples? ref #3318

SpacePossum:
oh, somehow missed the other PR, must have started before it and never checked before I opened it, d'oh : D

PR #3321, opened 2017-12-13
Do not remove imports from GLOBAL namespace

henriquemoody:
When importing functions or constants from the GLOBAL namespace, they
are not  considered as used, because the fixer consider ` function` and
`const` as part of the FQN of the use statement.

***
Duplicated of #3319 - Travis got crazy

### DIFF
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -171,6 +171,10 @@ final class NoUnusedImportsFixer extends AbstractFixer
                 continue;
             }
 
+            if (1 === preg_match('/^\s+(function|const)\s+(.+)$/i', $declarationContent, $matches)) {
+                $declarationContent = $matches[2];
+            }
+
             $declarationParts = preg_split('/\s+as\s+/i', $declarationContent);
 
             if (1 === count($declarationParts)) {
--- a/tests/Fixer/Import/NoUnusedImportsFixerTest.php
+++ b/tests/Fixer/Import/NoUnusedImportsFixerTest.php
@@ -684,6 +684,92 @@ EOF;
         $this->doTest($expected, $input);
     }
 
+    public function testFunctionsInTheGlobalNamespaceShouldNotBeRemoved()
+    {
+        $expected = <<<'EOF'
+<?php
+
+namespace Foo;
+
+use function is_int;
+
+is_int(1);
+
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
+namespace Foo;
+
+use function is_int;
+use function is_float;
+
+is_int(1);
+
+EOF;
+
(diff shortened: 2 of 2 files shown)


src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -171,6 +171,10 @@ private function getNamespaceUseDeclarations(Tokens $tokens, array $useIndexes)
                 continue;
             }
 
+            if (1 === preg_match('/^ (function|const)(.+)$/', $declarationContent, $matches)) {
```
SpacePossum: I think there must be one or more whitespace after `function|const` to prevent matching `use functionABC;`
Not sure about the `$` here, for example this valid:

```php
<?php

use function
    is_int;
```

(
```php
<?php

use function
 /*  */   is_int;
```

as well :|
)

The first space to match, are we sure there is always one (and not zero or multiple or combined with linebreaks and such)?

tests/Fixer/Import/NoUnusedImportsFixerTest.php
```diff
@@ -679,6 +679,62 @@ public function testWithSameNamespaceImportAndUnusedImportAfterNamespaceStatemen
...
+
+EOF;
+
+        $input = <<<'EOF'
+<?php
+
+namespace Foo;
+
+use function PHP_INT_MAX;
+use function PHP_INT_MIN;
```
SpacePossum: I think you meant `use const PHP_INT_MAX;` ?  thanks for adding support for this :+1:

keradus:
> Duplicated of #3319 - Travis got crazy

After changing base branch travis may still think it has to be build against old target branch.
Usually, rebase on new base branch with force push fix the issue on travis side.

henriquemoody:
I've just updated the branch. Is there anything else I should do?

SpacePossum:
cases like;

```php
<?php

namespace Foo;use FUNCTION
is_int;
use function
    is_float;

is_int(1);
```

will still fail (because of the white spaces used and the alternative casing of the keyword `function`).
I hope to take a closer look into a solution soon.

henriquemoody:
I patched it up and I created a test case for:

```php
<?php

namespace Foo;use FUNCTION
is_int;
use function
    is_float;
use
    const
        PHP_INT_MIN;

is_int(1);
```

Seems to be fine now.

SpacePossum:
@henriquemoody 
Thanks for all your work!
I did another review and come to the conclusion this bug can only be fixed by analyzing the tokens.
The current regex solution and the additional one proposed in the PR do not work when comments are `creatively` placed within or around `use` statements.

I cooked up; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3367 this includes some of your tests as well. Let me know what you think!

henriquemoody:
Closed in favor of #3367

PR #3337, opened 2017-12-21
BracesFixer: nowdoc bug on template files

Slamdunk:
Currently the fixer breaks code if the php tag is closed after a nowdoc syntax:
```diff
 <?php
 if (true) {
     $var = <<<'JS'
-JS;
-?>
+JS; ?>
 <?php
 }
```

- [x] Prove the bug
- [x] Find a solution (I'm scared of BracesFixer source 😨)

PR #3340, opened 2017-12-22
Verbose LintingException after fixing

Slamdunk:
I've had to struggle a few hours finding the root cause of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3337 because:

1. If a lot of fixers are enabled, LintingException doesn't trace what fixers where used for that specific file
1. If a LintingException occurs after fixing, the file content is (correctly) never altered

But if so, you don't know what caused the parse error, and what is the parse error; the plain PHP parse error doesn't help if the source code is thousands lines and the fixer altered it so much that the line numbers changed drastically.

Don't know the right branch for this change, ping me if a rebase is needed.

Before:

before

After:

after

### DIFF
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -100,17 +100,20 @@ final class ErrorOutput
                 }
             }
 
-            if (Error::TYPE_LINT === $error->getType() && $e instanceof LintingException) {
+            if (Error::TYPE_LINT === $error->getType() && 0 < count($error->getAppliedFixers())) {
                 $this->output->writeln('');
-                $this->output->writeln(sprintf('      Applied fixers: <comment>%s</comment>', implode(', ', $e->getAppliedFixers())));
+                $this->output->writeln(sprintf('      Applied fixers: <comment>%s</comment>', implode(', ', $error->getAppliedFixers())));
 
-                $diffFormatter = new DiffConsoleFormatter($this->isDecorated, sprintf(
-                    '<comment>      ---------- begin diff ----------</comment>%s%%s%s<comment>      ----------- end diff -----------</comment>',
-                    PHP_EOL,
-                    PHP_EOL
-                ));
+                $diff = $error->getDiff();
+                if (!empty($diff)) {
+                    $diffFormatter = new DiffConsoleFormatter($this->isDecorated, sprintf(
+                        '<comment>      ---------- begin diff ----------</comment>%s%%s%s<comment>      ----------- end diff -----------</comment>',
+                        PHP_EOL,
+                        PHP_EOL
+                    ));
 
-                $this->output->writeln($diffFormatter->format($e->getDiff()));
+                    $this->output->writeln($diffFormatter->format($diff));
+                }
             }
         }
     }
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -12,6 +12,7 @@
 
 namespace PhpCsFixer\Console\Output;
 
+use PhpCsFixer\Differ\DiffConsoleFormatter;
 use PhpCsFixer\Error\Error;
 use PhpCsFixer\Linter\LintingException;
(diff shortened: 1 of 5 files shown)


src/Console/Output/ErrorOutput.php
```diff
@@ -55,49 +56,62 @@ public function listErrors($process, array $errors)
...
-                            $this->outputTrace($trace);
-                        }
+                        break;
                     }
+
+                    $this->outputTrace($trace);
                 }
             }
+
+            if (Error::TYPE_LINT === $error->getType() && $e instanceof LintingException) {
```
Slamdunk: In this file, this if is the real change.

Above there is only an indentation change due to this minor enhancement:
```diff
-if ($showDetails) {
-    $e = $error->getSource();
-    if (null !== $e) {
+$e = $error->getSource();
+if (!$showDetails || null === $e) {
+    continue;
```

src/Linter/LintingException.php
```diff
@@ -14,7 +14,50 @@
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
```
Slamdunk: Needed to skip `ProjectCodeTest::testThatSrcClassesNotAbuseInterfaces`.

I think there is no need for interfaces here.

src/Linter/LintingException.php
```diff
@@ -14,7 +14,50 @@
...
+
+    /**
+     * @var null|string
+     */
+    private $diff;
+
+    /**
+     * @param array $appliedFixers List of applied fixers, if Error::TYPE_LINT
+     */
+    public function setAppliedFixers(array $appliedFixers)
```
keradus: don't create mutable objects

src/Linter/LintingException.php
```diff
@@ -14,7 +14,50 @@
...
+     */
+    private $appliedFixers = [];
+
+    /**
+     * @var null|string
+     */
+    private $diff;
+
+    /**
+     * @param array $appliedFixers List of applied fixers, if Error::TYPE_LINT
```
keradus: applicable only for specific type? sounds like a subclass

src/Linter/LintingException.php
```diff
@@ -14,7 +14,50 @@
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
  */
-class LintingException extends \RuntimeException
+final class LintingException extends \RuntimeException
```
keradus: wow
adding `@internal` is bc breaker
adding `final` is bc breaker

Slamdunk:
@keradus sorry, recaftored it all:

1. `src/Linter/LintingException.php ` now unaltered
1. diff details now saved into additional constructor arguments of `src/Error/Error.php`, which is both invariable and internal, so no BC break, only new internal API

SpacePossum:
I like the idea of having more details on failure, so :+1: for that :) Thanks for the efforts so far.

I'm wondering, the diff and applied fixers up till the problem: is this info the `Error` class should know about, or maybe better if the error-manager and the error-output-builder know how to deal with this meta info? Not sure about the composition of the classes here.

Slamdunk:
> is this info the Error class should know about, or maybe better if the error-manager

I would definitely say the Error: it already knows the type, the filename and all the details about the error because it has the exception. Applied fixers and diff are part of the context of the error.

I am not saying that the implementation in this PR is the ideal solution because I do not have extensive knowledge about the framework behing PHP-CS-Fixer, but speaking by abstract concepts, the error's details should fall into an Error entity, not into the ErrorManager service.

Slamdunk:
Rebased after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3405

keradus:
@FriendsOfPHP/php-cs-fixer could one pick this up for a review please?

SpacePossum:
The new details are outputted no matter what the verbosity level is?

keradus:
Thank you @Slamdunk.

PR #3341, opened 2017-12-23
Add phpdoc scalar type conversion to list of scalar types

icanhazstring:
Add phpdoc scalar conversion into list of valid scalar types as mentioned in
https://github.com/symfony/symfony/pull/25585#issuecomment-353685125

This is done, due to the fact that PHPDoc does not list `scalar` as a valid type.
So any 3rd party tool (i.e. IntelliJ/PHPStorm, Eclipse) does not recognize `scalar` as a valid type.

https://docs.phpdoc.org/references/phpdoc/types.html

So currently every IDE will show an error even there is none. For now `scalar` will be converted into a list of valid types `int|string|bool|float` to represent the same behavior as well as getting rid of phpdoc warnings.

Currently I added a PR for phpDocumentor, but as said, 3rd party IDEs do not support this type by now but it could be removed in the future if they do.

https://github.com/phpDocumentor/phpDocumentor2/pull/1923


keradus:
this is big change of behaviour of current fixer.
it has to be configurable.

GrahamCampbell:
👎 actually. Not every IDE, and the bug is theirs.

keradus:
:+1: for idea of getting rid of `scalar` pseudotype, it's not strict and different ppl, sadly, understand it differently.
but that diversity is next reason why it shall be configurable, or maybe even a separated rule ?

icanhazstring:
If you can point me into the direction of how to make it configurable I will change it :)

GrahamCampbell:
https://wiki.php.net/rfc/scalar-pseudo-type

icanhazstring:
That would be a nice addition in the future :)

keradus:
rfc was just raised yesterday.
next reason why it shall be separated fixer.

icanhazstring:
So as I looked into the docs. This just has to be a separate class with a different priority than the one I changed? Will do this after Christmas days.

keradus:
no need for touching priority

README.rst
```diff
@@ -1198,6 +1198,11 @@ Choose from the list of available rules:
   Scalar types should always be written in the same form. ``int`` not
   ``integer``, ``bool`` not ``boolean``, ``float`` not ``real`` or ``double``.
 
+* **phpdoc_scalar_breakdown** [@Symfony]
+
+  The scalar phpdoc type will be broken down into a list of valid scalar
+  types.
```
keradus: there aren't that much of them, please list them explicitly

src/Fixer/Phpdoc/PhpdocScalarFixer.php
```diff
@@ -40,7 +40,8 @@
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Scalar types should always be written in the same form. `int` not `integer`, `bool` not `boolean`, `float` not `real` or `double`.',
+            'Scalar types should always be written in the same form. '.
+            '`int` not `integer`, `bool` not `boolean`, `float` not `real` or `double`.',
```
keradus: please revert

tests/Fixer/Phpdoc/PhpdocScalarBreakdownFixerTest.php
```diff
@@ -0,0 +1,293 @@
...
+    /**
+     * @var notabooleanthistime|int|string|float|bool|integerr
+     */
+
+EOF;
+
+        $input = <<<'EOF'
+<?php
+    /**
+     * @var notabooleanthistime|scalar|integerr
```
keradus: please add test with `?scalar` and `Collection<scalar>`

src/Fixer/Phpdoc/PhpdocTypesFixer.php
```diff
@@ -48,6 +48,7 @@
         'true',
         'void',
         '$this',
+        'scalar',
```
keradus: shouldn't we extend tests for this fixer to check scalar as well ?

icanhazstring: Don't know. Because there are not all types tested as well. Just a small selection of em.

tests/Fixer/Phpdoc/PhpdocScalarBreakdownFixerTest.php
icanhazstring: You mean `scalar[]`? AFAIK `Collection<>` is no valid PHPDoc type.

keradus: No, I do mean `@var ?scalar` and `@var Collection<scalar>`, as stated.
both are valid according to psr-5 proposal, which we are following across the project.

icanhazstring: Oh, ok didn't even notice that one. Will do then

icanhazstring: AFAIK `@var ?scalar` is not even valid. If you add type hinting with `?string` you will get a `@param null|string`. So why should there be a test for it? Or should this be replaced with `null|int|string|float|bool` then?

icanhazstring:
Another thing I found:

Since `scalar[]` is probably a valid value. The current `AbstractPhpdocTypesFixer` does not work. Because I need to know if the given type in the annotation was defined as an array. So I need to rework a bit of this class to actually now if the current type was defined as array.
 
I would suggest a change on the abstract `normalize($type)` with some kind of flag:
```
abstract protected function normalize($type, $isArray);
```

Where this would be `true` if the `AbstractPhpdocTypesFixer`

tests/Fixer/Phpdoc/PhpdocScalarBreakdownFixerTest.php
keradus: > I guess this should be done in the `PhpdocTypeFixer` to replace `?string` with `null|string`?

No, why ? How did you come up with this? I didn't asked for it, `?string` is perfectly valid, and since 7.1 it's better than `null|string` alternation, as it could by typehinted as-is in function prototype.

> `AFAIK @var ?scalar is not even valid`

It is. It is valid as much as `?string` is.

> If you add type hinting with `?string` you will get `@param null|string`

No, If I add typehinting `?foo` I will get `@param ?foo`. `?foo` is new way of saying type is nullable without saying what is default value, one could use it as-is in documentation.

> So why should there be a test for it? 

To prove it works properly.

> Or should this be replaced with `null|int|string|float|bool` then?

indeed

icanhazstring: > It is. It is valid as much as ?string is.
In which spec is this even listed?

keradus: phpdoc accepts any type that is valid for php itself and it's combinations, like `int|string` or `int[]`.
`?string` is valid PHP type (it is a type, it is not a string with null as default value - similar, yet different)

keradus: > Doesn't make sense to me.
> will be a shown as valid input. but it is not

typehint is not taken into consideration if it is only in typehint. for **any** type. 
not only for `?foo`. use it properly and you will see error
`Fatal error: Uncaught TypeError: Argument 1 passed to foo() must be of the type string or null, integer given`

https://3v4l.org/MbOgD

keradus: > The new declaration for psr-5 `(int|bool|float|string)[]` is currently not valid for phpdoc.

why you say it is not valid? it is valid according to spec. if some IDE is not supporting it - issue of IDE.
yet, sure, if you want to do `int[]|bool[]|float[]|string[]`, go ahead

icanhazstring: Then if we are going to full phpspec. The `PhpdocTypeFixer` and `PhpdocScalarFixer` won't work with `?STRING` or anything else. because it is not listed as type inside theses classes.

So we have to rewrite the whole thing and consider `?` and `[]` as type modifier which should be ignored. Currently `PhpdocTypeFixer` simply cuts `[]` off and appends it again at the end after doing some fixes.

So this change will take some time to extract the modifiers of the type and pass em down to any specific type fixing itself.

src/Fixer/Phpdoc/PhpdocTypesFixer.php
GrahamCampbell: We should probably add in some extra types that we missed, from https://github.com/phpDocumentor/TypeResolver/blob/master/src/TypeResolver.php#L47 too.

GrahamCampbell: Also, this list was sorted alphabetically, and now is not after adding `scalar` to it.

icanhazstring: Will add the missing types and resort them as well

GrahamCampbell: > Will add the missing types and resort them as well

I have already done this in another PR.

icanhazstring: Well then will remove the adding here to avoid conflicts

tests/Fixer/Phpdoc/PhpdocScalarBreakdownFixerTest.php
```diff
@@ -0,0 +1,293 @@
...
+ * @covers \PhpCsFixer\Fixer\Phpdoc\PhpdocScalarBreakdownFixer
+ */
+final class PhpdocScalarBreakdownFixerTest extends AbstractFixerTestCase
+{
+    public function testBasicFix()
+    {
+        $expected = <<<'EOF'
+<?php
+    /**
+     * @return int|string|float|bool
```
nicolas-grekas: I would suggest to turn that into `string|int|float|bool`, to group numbers together

keradus: let us use `bool|float|int|string` so we have numbers grouped and alpha order ;)

icanhazstring:
Yep. Sorry. Couldn't find the time in past days. Will give an update before weekend tho :)

icanhazstring:
Yep. Sorry. Couldn't find the time in past days. Will give an update before weekend tho :)

icanhazstring:
I have reworked the type fixing a bit.

**What changed**
- added support for `Collection<>` (like @GrahamCampbell suggested) any further support for other generics should be done in another PR (see below)

- Modifier like array, nullable and generics will be removed from the `$type` so every conversion works as expected. Previously if you have given a type an array modifier `integer[]`, the conversion didn't work to `int[]`

- `normalize($type)` can now return an array as well. This was needed to be aware that we should use PSR-5 array notation `(type)[]` without parsing types again and improve execution time a bit.

**What should be done further** (probably in another PR)
Currently I have done the modifier parsing inside the `AbstractPhpDocTypeFixer` which is not a perfect place. I noticed that the `Docblock\Annotation::getTypes()` already has a regex matching *array*, *simple* or *generic*. This method however only returns `string[]`. A new class `DocBlock\Type` should be introduced to reflect what type the annotations contain and if the type has any modifiers (like `isGeneric`) and what the modifier looks like. So we can then use some `TypeFormatter` inside the `AbstractPhpDocTypeFixer` to actually return a proper format.

icanhazstring:
So. With PSR-5 abandoned. How to we go with this?

keradus:
the same as if it wouldn't

julienfalque:
@icanhazstring Hi, do you still have interest in moving this forward?

icanhazstring:
Will have roncheck with latest IDE Versions. Give ne some dass. Otherwise i'll Close it 😉

Wirone:
Closing, since source branch has been deleted. Also PHPStorm and [PHPStan](https://phpstan.org/writing-php-code/phpdoc-types#basic-types) supports `scalar`. I believe this is not needed anymore.

image

Anyway, thank you @icanhazstring for taking your time and working on this 🍻.

PR #3342, opened 2017-12-23
PhpUnitDedicateAssertFixer - Remove unexistent method is_boolean 

carusogabriel:
With this PR, I want to propose to removal of `is_boolean` from `PhpUnitDedicateAssertFixer`. As far as I know, `is_bool` does not have an alias in PHP. `is_int` and `is_float` have, but `is_bool` doesn't. As the proposal of `PhpUnitDedicateAssertFixer` is to fix e.g.:
```php
$this->assertTrue(is_string('foo'));
```
to
```php
$this->assertInternalType('string', 'foo');
```
the fixer will never find a
```php
$this->assertTrue(is_boolean(true));
```
I know that `PHPUnit`'s `assertInternalType` method [does support `assertInternalType('boolean', true)`](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsType.php#L119), but the fixer will never find an `is_boolean`.


Slamdunk:
Yup: https://3v4l.org/ecNmf
```php
var_dump(function_exists('is_boolean'));
```

SpacePossum:
great catch :+1:  thanks!

(I've no idea why I've ever added that (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1792))

SpacePossum:
@carusogabriel can you retarget this fix to 2.2, I think it will be of use there are well :)

carusogabriel:
After seen #1792, and my recently [PR in Symfony](https://github.com/symfony/symfony/pull/25420), I guess we can create some Fixers :nerd_face:

keradus:
Thank you @carusogabriel.

PR #3343, opened 2017-12-26
Tokens - improve block end lookup

keradus:
Before:
> Fixed all files in 17.040 seconds, 14.000 MB memory used
> Fixed all files in 17.615 seconds, 14.000 MB memory used
> Fixed all files in 17.188 seconds, 14.000 MB memory used

After:
> Fixed all files in 16.581 seconds, 14.000 MB memory used
> Fixed all files in 16.579 seconds, 14.000 MB memory used
> Fixed all files in 16.456 seconds, 14.000 MB memory used

few % improved

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -47,6 +47,13 @@ class Tokens extends \SplFixedArray
     private static $cache = [];
 
     /**
+     * Cache of block edges. Any change in collection will invalidate it.
+     *
+     * @var array<int, int>
+     */
+    private $blockEndCache = [];
+
+    /**
      * crc32 hash of code string.
      *
      * @var string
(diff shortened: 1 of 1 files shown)


keradus:
@SpacePossum , we had a small talk about it ages ago. take a look ;)

SpacePossum:
awesome :)

SpacePossum:
One question,
if `$isLegacyMode` is `true` the cache might be outdated if tokens are changed using `Token::override` / `Tokens::overrideAt` I think, maybe best not to use the cache if in legacy mode?

keradus:
very good point

keradus:
done

SpacePossum:
thanks, looks good to me :+1:

PR #3346, opened 2017-12-26
StdinFileInfo - drop getContents

keradus:
method does not exist in parent class, it's not used as well.
class is internal, so we could drop it.


SpacePossum:
relies on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3345

PR #3349, opened 2017-12-27
Fix stdin handling and add tests for it

keradus:
fixing stdin is not working as filtering / linting processes are consuming the content of stdin, which is readable **once**

```
ker@dus:~/github/PHP-CS-Fixer$ vendor/bin/phpunit tests/Smoke/StdinTest.php 
PHPUnit 5.7.26 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.0-2+ubuntu17.04.1+deb.sury.org+2
Configuration: /home/keradus/github/PHP-CS-Fixer/phpunit.xml.dist

Testing PhpCsFixer\Tests\Smoke\StdinTest
F                                                                                                                                                                                                                                                                      1 / 1 (100%)

 1. 143ms to run PhpCsFixer\Tests\Smoke\StdinTest:testFixingStdin


Time: 165 ms, Memory: 4.00MB

There was 1 failure:

1) PhpCsFixer\Tests\Smoke\StdinTest::testFixingStdin
Failed asserting that Array &0 (
    'code' => 0
    'error' => 'Loaded config default from "/home/keradus/github/PHP-CS-Fixer/.php_cs.dist".
'
    'output' => '
Checked all files in 0.001 seconds, 8.000 MB memory used
'
) is identical to Array &0 (
    'code' => 8
    'error' => 'Loaded config default from "/home/keradus/github/PHP-CS-Fixer/.php_cs.dist".
'
    'output' => '   1) PHP-CS-Fixer/tests/Fixtures/Integration/set/@PSR2.test-in.php
--- Original
+++ New
@@ @@
-<?
+<?php
 namespace Vendor\Package;
-use FooInterface, FooInterfaceB;
+
+use FooInterface;
+use FooInterfaceB;
 use BarClass as Bar;
 use OtherVendor\OtherPackage\BazClass;
-class Foo extends Bar implements FooInterface{
-    var $aaa = 1, $bbb = 2;

+class Foo extends Bar implements FooInterface
+{
+    public $aaa = 1;
+    public $bbb = 2;
+
     public function sampleFunction($a, $b = null)
     {
         if ($a === $b) {
             bar();
-        } else if ($a > $b) {
+        } elseif ($a > $b) {
             $foo->bar($arg1);
         } else {
-                BazClass::bar($arg2, $arg3);
+            BazClass::bar($arg2, $arg3);
         }
     }

-    static public  final function bar() {
-    // method body
+    final public static function bar()
+    {
+        // method body
     }
 }

 class Aaa implements
-    Bbb, Ccc,
+    Bbb,
+    Ccc,
     Ddd
-    {
-    }
-?>
+{
+}


Checked all files in 0.009 seconds, 8.000 MB memory used
'
).

/home/keradus/github/PHP-CS-Fixer/tests/Smoke/StdinTest.php:52

FAILURES!
Tests: 1, Assertions: 1, Failures: 1.
```


Deps:
- [x] #3345
- [x] #3346

Broken since:
- #722

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -128,10 +128,7 @@ jobs:
             php: 7.1
             install: ./dev-tools/build.sh
             script:
-                - php php-cs-fixer.phar --version
-                - php php-cs-fixer.phar readme
-                - php php-cs-fixer.phar describe header_comment
-                - php php-cs-fixer.phar fix src/Config.php -vvv --dry-run --diff
+                - vendor/bin/phpunit tests/Smoke/PharTest.php
             deploy:
                 provider: releases
                 api_key:
--- a/src/FileReader.php
+++ b/src/FileReader.php
@@ -36,7 +36,7 @@ final class FileReader
     /**
      * @return self
      */
-    public static function createSingletone()
+    public static function createSingleton()
     {
         if (null === self::$instance) {
             self::$instance = new self();
--- a/src/FileReader.php
+++ b/src/FileReader.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * File reader that unify access to regular file and stdin-alike file.
+ *
+ * Regular file could be read multiple times with `file_get_contents`, but file provided on stdin can not.
+ * Consecutive try will provide empty content for stdin-alike file.
+ * This reader unifies access to them.
+ *
+ * @internal
+ */
+final class FileReader
+{
+    /**
+     * @var ?self
+     */
+    private static $instance;
+
+    /**
+     * @var ?string
+     */
+    private $stdinContent;
+
+    /**
+     * @return self
+     */
+    public static function createSingletone()
+    {
+        if (null === self::$instance) {
+            self::$instance = new self();
+        }
+
+        return self::$instance;
+    }
+
+    /**
+     * @param string $filePath
+     *
+     * @return string
+     */
+    public function read($filePath)
+    {
+        if ('php://stdin' === $filePath) {
+            if (null === $this->stdinContent) {
+                $this->stdinContent = $this->readRaw($filePath);
+            }
+
+            return $this->stdinContent;
+        }
+
+        return $this->readRaw($filePath);
+    }
+
+    /**
+     * @param string $realPath
+     *
+     * @return string
+     */
+    private function readRaw($realPath)
+    {
+        return file_get_contents($realPath);
+    }
+}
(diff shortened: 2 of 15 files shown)


src/FileReader.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * File reader that unify access to regular file and stdin-alike file.
+ *
+ * Regular file could be read multiple times with `file_get_contents`, but file provided on stdin can not.
+ * Consecutive try will provide empty content for stdin-alike file.
+ * This reader unifies access to them.
+ *
+ * @internal
+ */
+final class FileReader
```
keradus: This is the sad part of this PR.
Alternative would be to not rely on in-memory `stdin` stream, but store file to physical drive, which would slow down the process so I don't want it.

.travis.yml
```diff
@@ -128,10 +128,7 @@ jobs:
             php: 7.1
             install: ./dev-tools/build.sh
             script:
-                - php php-cs-fixer.phar --version
-                - php php-cs-fixer.phar readme
-                - php php-cs-fixer.phar describe header_comment
-                - php php-cs-fixer.phar fix src/Config.php -vvv --dry-run --diff
```
keradus: basically, those were testig only exit status. let us test some content as well

src/FileReader.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * @var ?string
+     */
+    private $stdinContent;
+
+    /**
+     * @return self
+     */
+    public static function createSingletone()
```
Slamdunk: It should be Singleton, without the `e` (https://en.wikipedia.org/wiki/Singleton_pattern).

How is that an excellent developer like you doesn't know how to properly spell such an important and fundamental design pattern? (just joking :P)

keradus: nice catch, will fix later for sure :+1:
so much used to DIC ;P I was angry for writing this shit down, but well, stdin stream has global state :( (and in my opinion singleton === global)

Slamdunk: I think you are too good.

STDIN support was introduced in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/310 and from then on I've never seen a use case that wouldn't be better handled with normal command argument.

Now (a lot of versions after v0.5.0, the first tag containing it) we have better verbosity and diff to know what is going on without altering the real code.

So, if it would be up to me, I would just drop it, not because the feature is totaly useless, but because it is becoming a burden more than a valuable feature.

Regarding the implementation:

1. Singletons should have private contructor
1. I would drop Singleton at all
1. I would rely only on the public static method, after making the class final with the contructor set to private; calling `createSingleton` every time doesn't bring value to me for such a simple class

keradus: There were actually multiple issues/questions about stdin even for already v2.
Even one with a guy having troubles using it, but we thought it was different issue. And in the end, it was possible to run stdin in very narrow configuration, like no cache and php7 linting (well... CI and tools integrations? :D)
Having in mind this and [that](https://youtrack.jetbrains.com/issue/WI-38820), I wouldn't drop the feature.

About implementation concerns. Let us go from... bottom.
> I would rely only on the public static method

nope ;) you can't test it really. That's exactly the case why I did created regular class, why i'm not using static property for data read from stdin (as it would be shared across all instances, causing tests to be problematic), why I don't hide constructor and so.
I want to have this class as "clean" as possible with one violation only - `createSingleton`, which gonna be removed as soon as we will switch to dependency injection and we will treat `FileReader` as a service. Then, the only change to `FileReader` would be to drop static `createSingleton`, not redesign of whole class (even when it's small one), which will help us in LTS branch merging

PR #3354, opened 2017-12-29
Added missing types to the PhpdocTypesFixer

GrahamCampbell:
I've picked from https://github.com/phpDocumentor/TypeResolver/blob/0.5.0/src/TypeResolver.php#L47 the types that we missed. I'll also need to send back to them the types they missed.


keradus:
tests are missing. I don't like features/bugfixes passing without touching tests

GrahamCampbell:
Sure, can add more tests. I don't think our existing tests cover that whole array as is anyway, but sure, can add more tests.

GrahamCampbell:
Will do tomorrow, will need to pull down the code to my laptop.

SpacePossum:
:+1: for the idea, love to see to test :)

keradus:
ping @GrahamCampbell

keradus:
Thank you @GrahamCampbell.

PR #3360, opened 2018-01-02
Adjust Symfony ruleset

keradus:


src/RuleSet.php
```diff
@@ -157,9 +168,11 @@
             'modernize_types_casting' => true,
             'no_alias_functions' => true,
             'no_homoglyph_names' => true,
-            'non_printable_character' => true,
+            'non_printable_character' => [
+                // @TODO: on 3.0 switch whole rule to `=> true`, currently we use custom config that will be default on 3.0
+                'use_escape_sequences_in_strings' => true,
+            ],
             'php_unit_construct' => true,
-            'php_unit_dedicate_assert' => true,
```
keradus: As by default it applies changes for PHPUnit 5.0, but Sf can apply on Sf only 4.8.
As this is part of migration ruleset, it shall not be part of style ruleset of `@Symfony` (and Sf is already using PHPUnit migration ruleset pinned to PHPUnit 4.8).

src/RuleSet.php
```diff
@@ -157,9 +168,11 @@
             'modernize_types_casting' => true,
             'no_alias_functions' => true,
             'no_homoglyph_names' => true,
-            'non_printable_character' => true,
+            'non_printable_character' => [
+                // @TODO: on 3.0 switch whole rule to `=> true`, currently we use custom config that will be default on 3.0
+                'use_escape_sequences_in_strings' => true,
```
keradus: Sf is following this new feature, which is disabled by default on 2.x and will be enabled by default on 3.x

src/RuleSet.php
```diff
@@ -109,7 +109,18 @@
...
+                // @TODO: on 3.0 switch whole rule to `=> true`, currently we use custom config that will be default on 3.0
+                'tags' => [
+                    'method',
+                    'param',
+                    'property',
+                    'return',
+                    'throws',
+                    'type',
+                    'var',
+                ],
```
keradus: Sf is following this new feature, which is disabled by default on 2.x and will be enabled by default on 3.x

src/RuleSet.php
SpacePossum: maybe add this reason as a comment to the code?

keradus: it is not changing to `'php_unit_dedicate_assert' => false`, it is dropping the line.
hard to add comment to line that does not exist anymore ;)

SpacePossum: haha fair enough :)
:ship:

SpacePossum:
looks good, however sadly the tests are failing

PR #3362, opened 2018-01-02
Add BacktickToShellExecFixer

Slamdunk:
```diff
-$plain = `ls -lah`;
+$plain = shell_exec("ls -lah");

-$withVar = `ls -lah $var1 ${var2} {$var3} {$var4[0]} {$var5->call()}`;
+$withVar = shell_exec("ls -lah $var1 ${var2} {$var3} {$var4[0]} {$var5->call()}");
```

Reference: https://secure.php.net/manual/en/language.operators.execution.php

### DIFF
--- a/tests/Fixer/Alias/BacktickToShellExecFixerTest.php
+++ b/tests/Fixer/Alias/BacktickToShellExecFixerTest.php
@@ -41,43 +41,31 @@ final class BacktickToShellExecFixerTest extends AbstractFixerTestCase
                 '<?php shell_exec("ls -lah");',
                 '<?php `ls -lah`;',
             ],
-            'with double quote' => [
-<<<'EOT'
-<?php
-shell_exec("ls -lah a\\\"m\\\\z");
-EOT
-,
-<<<'EOT'
-<?php
-`ls -lah a\"m\\\\z`;
-EOT
-,
-            ],
             'with variables' => [
                 '<?php shell_exec("$var1 ls ${var2} -lah {$var3} file1.txt {$var4[0]} file2.txt {$var5->call()}");',
                 '<?php `$var1 ls ${var2} -lah {$var3} file1.txt {$var4[0]} file2.txt {$var5->call()}`;',
             ],
-            'with double quote and variable' => [
+            'with single quote' => [
 <<<'EOT'
 <?php
-shell_exec("ls -lah a\\\"m\\\\z $var");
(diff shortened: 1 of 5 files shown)


tests/Fixer/Alias/BacktickToShellExecFixerTest.php
```diff
@@ -0,0 +1,74 @@
...
+shell_exec("ls -lah a\\\"m\\\\z $var");
+EOT
+,
+<<<'EOT'
+<?php
+`ls -lah a\"m\\\\z $var`;
+EOT
+,
+            ],
+        ];
```
kubawerlos: Can you add test case with `$input`: ``<?php `echo 'foo\`bar'`;``?

Slamdunk: Done, thanks

tests/Fixer/Alias/BacktickToShellExecFixerTest.php
```diff
@@ -0,0 +1,86 @@
...
+,
+            ],
+            'with variables' => [
+                '<?php shell_exec("$var1 ls ${var2} -lah {$var3} file1.txt {$var4[0]} file2.txt {$var5->call()}");',
+                '<?php `$var1 ls ${var2} -lah {$var3} file1.txt {$var4[0]} file2.txt {$var5->call()}`;',
+            ],
+            'with double quote and variable' => [
+<<<'EOT'
+<?php
+shell_exec("ls -lah a\\\"m\\\\z $var");
```
keradus: is it needed to have extra escaping here ?

Slamdunk:
After further investigations I decided to skip cases with special chars like single-quotes, double-quotes and backticks because the final escaped sequence depends on the context:
```php
// Single-quotes behave the same, but affect other special chars so better skip them too
var_dump(`echo a\'b` === shell_exec("echo a\\'b"));
var_dump(`echo 'ab'` === shell_exec("echo 'ab'"));

var_dump(`echo a\"b` === shell_exec("echo a\\\"b"));
var_dump(`echo 'a"b'` === shell_exec("echo 'a\"b'"));

var_dump(`echo 'a\`b'` === shell_exec("echo 'a`b'"));
var_dump(`echo a\\\`b` === shell_exec("echo a\\`b"));
```

keradus:
Thank you @Slamdunk.

PR #3364, opened 2018-01-03
Configuration - finder string[]

SpacePossum:
In the `ConfigInterface` we suggest it is OK to return `string[]` for the finder.
However both the `Runner` and the `ConfigurationResolver` do not like this.

Since we never suggested that the `Runner` accepts  the `string[]` I've only updated that one only to be more defensive.

The `ConfigurationResolver` now transforms the string array to an iterator providing `SplFileInfo` objects based on the strings in the array.

for ref.:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.13/src/ConfigInterface.php#L40
and:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3305#issuecomment-352557362

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -767,7 +767,9 @@ final class ConfigurationResolver
             array_walk(
                 $currentFinder,
                 function (&$item) {
-                    $item = new \SplFileInfo($item);
+                    if (!($item instanceof \SplFileInfo)) { // do not use `is_string` to prevent reference mismatch
+                        $item = new \SplFileInfo($item);
+                    }
                 }
             );
 
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -753,12 +753,25 @@ final class ConfigurationResolver
             $this->getPath()
         ));
 
+        $currentFinder = $this->getConfig()->getFinder();
+
         if (!count($paths)) {
             if ($isIntersectionPathMode) {
                 return new \ArrayIterator(array());
             }
 
-            return $this->iterableToTraversable($this->getConfig()->getFinder());
+            if (!is_array($currentFinder)) {
+                return $this->iterableToTraversable($currentFinder);
+            }
+
+            array_walk(
+                $currentFinder,
+                function (&$item) {
+                    $item = new \SplFileInfo($item);
+                }
+            );
+
+            return new \ArrayIterator($currentFinder);
         }
 
         $pathsByType = array(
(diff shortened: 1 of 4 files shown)


src/Runner/FileFilterIterator.php
```diff
@@ -61,7 +66,7 @@ public function accept()
 
         $this->visitedElements[$path] = true;
 
-        if ($file->isDir() || $file->isLink()) {
+        if (!$file->isFile() || $file->isLink()) {
```
SpacePossum: more defensive, there is a test to prove the need for this

src/Runner/FileFilterIterator.php
```diff
@@ -52,7 +52,12 @@ public function __construct(
 
     public function accept()
     {
+        /** @var \SplFileInfo $file */
         $file = $this->current();
+        if (!($file instanceof \SplFileInfo)) {
```
julienfalque: Parentheses around `instanceof` are superfluous (here and on other changed lines in this PR).

SpacePossum: I like the parentheses as I don't want to state 
>is `!$file` an instance of `\SplFileInfo`

even if the operator precedence changes the meaning of the sentence for me

src/Console/ConfigurationResolver.php
```diff
@@ -775,7 +790,7 @@ function ($path) {
         }
 
         $nestedFinder = null;
-        $currentFinder = $this->iterableToTraversable($this->getConfig()->getFinder());
+        $currentFinder = $this->iterableToTraversable($currentFinder);
```
julienfalque: This is the only place where `iterableToTraversable` is called now. I think we should drop it as it is private and trivial.

src/Runner/FileFilterIterator.php
keradus: `FileFilterIterator` has to iterate over `SplFileInfo`. if it doesn't, there is sth horribly wrong

SpacePossum: maybe change to throwing an exception than?

keradus: how one can pass string here in the first place?

keradus:
While I see your work in this PR nice, I, sadly, do believe the bug was only in our docs, not in the code. As we can see in your PR we obviously don't support `string[]`, and I must admit that there is no benefit of doing so, `SplFileInfo[]` is enough, while it brings extra compilations to code.
Please extract the cleanup/other fixes (isDir vs isFile) to separated PR and close this one.

keradus:
As we do have #3415, could this one be closed @SpacePossum ?

SpacePossum:
indeed @keradus :+1: 
closing in favor of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3415

PR #3365, opened 2018-01-03
COOKBOOK-FIXERS.md - update to provide definition instead of description

keradus:


src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -242,14 +242,6 @@ protected function createConfigurationDefinition()
         ));
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    protected function getDescription()
```
SpacePossum: nice catch :)

Issue #3291, opened 2017-12-05
Convert the string from utf8 to windows 1252.


icanc:
#### The PHP version 

```
PHP 7.2.0 (cli) (built: Dec  3 2017 21:46:44) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2017 Zend Technologies
```
#### PHP CS Fixer version

```
PHP CS Fixer 2.8.3 Round Numbers by Fabien Potencier and Dariusz Ruminski
```
 
#### The command 

```
php-cs-fixer fix ./Test.php --diff
```

#### Test.php (utf-8)

```
<?php

class Test
{
//酰
}

```

#### The result
 
```
Loaded config default.
Using cache file ".php_cs.cache".
   1) Test.php
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
 <?php

 class Test
 {
-//酰
+    //?    ?
 }
      ----------- end diff -----------
```

Slamdunk:
Hi, could you re-run the command with `--verbose` flag so it outpus also the fixer name that caused the unexpected change?

kubawerlos:
It's this [line](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/Fixer/Basic/BracesFixer.php#L864), because of `\R` does not work as `\n|\r|\r\n`: https://3v4l.org/fVYlQ

We can either replace `\R` with the `\n|\r|\r\n` or expect `mbstring` extension and use `mb_ereg_replace`. I'll try to to it today.

keradus:
cc @julienfalque 
could you please take a look ?

keradus:
https://3v4l.org/3n9gB
cc @kubawerlos

keradus:
Closing due to release of #3446 , please update version you are using

PR #3366, opened 2018-01-03
Add RegularExpressionUtf8ModifierFixer

kubawerlos:
This fixer will add `u` modifier to "preg_*" functions and replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3296 and solves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3291

Looking forward for some feedback.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1288,6 +1288,11 @@ Choose from the list of available rules:
     ones; defaults to ``['getrandmax' => 'mt_getrandmax', 'rand' =>
     'mt_rand', 'srand' => 'mt_srand']``
 
+* **regular_expression_utf8_modifier**
+
+  Modifier ``u`` (pattern and subject strings are treated as UTF-8) should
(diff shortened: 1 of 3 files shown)


keradus:
:-1: until we will find a way to have it enabled for our repo, and we do know that start using `/u` brought few issues to us

kubawerlos:
> until we will find a way to have it enabled for our repo

what do you mean? I have tested it by adding it to `.php_cs.dist` and with the test from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3296 and all went fine.

keradus:
because our tests are written in utf-8, so enforcing regexes to work in utf-8 doesn't change anything for us.
it matters when the code is not in utf-8, vide https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3309/files

PR #3370, opened 2018-01-04
AbstractFixer - FQCN in in exceptions

Slamdunk:
I just adapted the code to other LogicException thrown in AbstractFixer.

### DIFF
--- a/src/AbstractFixer.php
+++ b/src/AbstractFixer.php
@@ -166,7 +166,7 @@ abstract class AbstractFixer implements FixerInterface, DefinedFixerInterface
     public function setWhitespacesConfig(WhitespacesFixerConfig $config)
     {
         if (!$this instanceof WhitespacesAwareFixerInterface) {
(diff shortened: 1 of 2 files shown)


src/AbstractFixer.php
```diff
@@ -166,7 +166,7 @@ public function getConfigurationDefinition()
     public function setWhitespacesConfig(WhitespacesFixerConfig $config)
     {
         if (!$this instanceof WhitespacesAwareFixerInterface) {
-            throw new \LogicException('Cannot run method for class not implementing `WhitespacesAwareFixerInterface`.');
+            throw new \LogicException('Cannot run method for class not implementing "PhpCsFixer\Fixer\WhitespacesAwareFixerInterface".');
```
keradus: why removing backticks ?

SpacePossum: these are replaced to be the same as done @ line 156

keradus:
Thank you @Slamdunk.

PR #3371, opened 2018-01-04
IsNullFixer must be run before YodaStyleFixer

kubawerlos:
@SpacePossum what do you think?

### DIFF
--- a/tests/AutoReview/FixerFactoryTest.php
+++ b/tests/AutoReview/FixerFactoryTest.php
@@ -83,6 +83,7 @@ final class FixerFactoryTest extends TestCase
             [$fixers['function_to_constant'], $fixers['no_trailing_whitespace']], // tested also in: function_to_constant,no_trailing_whitespace.test
             [$fixers['function_to_constant'], $fixers['no_whitespace_in_blank_line']], // tested also in: function_to_constant,no_whitespace_in_blank_line.test
             [$fixers['indentation_type'], $fixers['phpdoc_indent']],
+            [$fixers['is_null'], $fixers['yoda_style']], // tested also in: is_null,yoda_style.test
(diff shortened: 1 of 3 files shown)


SpacePossum:
Looking good thanks!
Please add an entry here as well https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.13/tests/AutoReview/FixerFactoryTest.php#L59

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -83,6 +83,7 @@ public function provideFixersPriorityCases()
             [$fixers['function_to_constant'], $fixers['no_trailing_whitespace']], // tested also in: function_to_constant,no_trailing_whitespace.test
             [$fixers['function_to_constant'], $fixers['no_whitespace_in_blank_line']], // tested also in: function_to_constant,no_whitespace_in_blank_line.test
             [$fixers['indentation_type'], $fixers['phpdoc_indent']],
+            [$fixers['is_null'], $fixers['yoda_style']],
```
keradus: please add comment with a file as well (as when there is lack of comment, we know we don't have itest for given priority test yet)

kubawerlos: added and rebased

keradus:
I wonder, as long as this PR itself is good, maybe we shall drop `yoda` part of `IsNullFixer`, as yoda is handled fully by dedicated rule.

kubawerlos:
I thought about it too and created an issue with more details: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3399

keradus:
Thank you @kubawerlos.

PR #3382, opened 2018-01-04
DefinedFixerInterface - incorporate into FixerInterface

keradus:


julienfalque:
We should add a note about BC breaks in the `UPGRADE.md` file.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3389#issuecomment-355532770

TomasVotruba:
@julienfalque No troubles, I'm covering it in Rector: https://github.com/rectorphp/rector/blob/master/src/config/level/friends-of-php/php-cs-fixer30.yml

Already experimenting with upgrade to 3.0: https://github.com/Symplify/Symplify/pull/648 and this is must have

PR #3384, opened 2018-01-04
diff-format option: drop sbd diff, use udiffer by default, drop SebastianBergmannDiffer and SebastianBergmannShortDiffer classes

keradus:


src/Console/Command/HelpCommand.php
```diff
@@ -101,8 +101,8 @@ public static function getHelpCopy()
 
 The <comment>--diff-format</comment> option allows to specify in which format the fixer should output the changes it makes:
 
-* <comment>udiff</comment>: unified diff format;
-* <comment>sbd</comment>: Sebastianbergmann/diff format (default when using `--diff` without specifying `diff-format`).
+* <comment>null</comment>: no diff;
```
julienfalque: Does it make sense to keep the `--diff-format` option while only one format is supported? I mean the other possible value (`null`) disables the diff output, which is what `--diff` option already does.

keradus: perfect candidate for next PR, could you create one ?

keradus: extracted to https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3400

PR #3385, opened 2018-01-04
ConfigurableFixerInterface::configure - param is now not nullable and not optional

keradus:


julienfalque:
👍 looking good.

What about adding a note to the UPGRADE file? Users may need to update their config file.

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3389#issuecomment-355532770

PR #3387, opened 2018-01-04
FixCommand - forbid passing 'config' and 'rules' options together

keradus:


src/Console/Command/FixCommand.php
```diff
@@ -121,11 +122,13 @@ protected function configure()
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
-        $verbosity = $output->getVerbosity();
-
         $passedConfig = $input->getOption('config');
         $passedRules = $input->getOption('rules');
 
+        if (null !== $passedConfig && null !== $passedRules) {
+            throw new InvalidConfigurationException('Passing both `--config` and `--rules` options is not possible.');
```
SpacePossum: Technically passing the two is possible (that is how we end up here), it is however not valid ;)

PR #3392, opened 2018-01-05
[PhpdocScalarFixer] Nullable must be null

carusogabriel:
@GrahamCampbell Should we fix it? Sometimes I catch myself doing this :sweat_smile:

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocScalarFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocScalarFixer.php
@@ -30,6 +30,7 @@ final class PhpdocScalarFixer extends AbstractPhpdocTypesFixer
         'boolean' => 'bool',
         'double' => 'float',
         'integer' => 'int',
+        'nullable' => 'null',
         'real' => 'float',
         'str' => 'string',
     );


src/Fixer/Phpdoc/PhpdocScalarFixer.php
```diff
@@ -30,6 +30,7 @@
         'boolean' => 'bool',
         'double' => 'float',
         'integer' => 'int',
+        'nullable' => 'null',
```
keradus: what the hell is `nullable` type? there is no such type nor pseudotype. nullable is attriubte of type since 7.1, like nullable string (`?string)`. nullable without type is meaningless as a type

carusogabriel: So this change is right? As `PhpdocScalarFixer` is support to fix it :grimacing:

keradus: How did you get that conclusion?
No, this change is not right. there is no such type or pseudotype as `nullable`.

SpacePossum:
Thanks for the proposal @carusogabriel ! however I'm closing this one.
The way I see it is that `nullable` is not a type and also it is not the same as `null`; therefore we shouldn't fix towards that.
After checking some repo's I didn't find the mistaken usage of the `nullable` "type hint", however if there are more people here who like this PR go through please let us know with a comment here.

PR #3393, opened 2018-01-05
Method call typos

Slamdunk:
Running phpstan I found those typos.


SpacePossum:
now we wpuld get; 
`isNewLine` and `fixNewline`

maybe it is better to use one way of writing `Newline` or `NewLine`

Slamdunk:
I noticed the mismatch, but didn't want to change signatures.

Ping me again if you want a change.

SpacePossum:
please change on of the two signature to match the other :+1:

keradus:
Thank you @Slamdunk.

PR #3401, opened 2018-01-07
Use `strpos` instead of `substr` where possible

ntzm:
Some benefits:

- Less to maintain (only one value instead of two to change)
- Faster
- Less memory intensive
- Slightly less code 😉

### DIFF
--- a/src/Fixer/ArrayNotation/TrimArraySpacesFixer.php
+++ b/src/Fixer/ArrayNotation/TrimArraySpacesFixer.php
@@ -88,7 +88,7 @@ final class TrimArraySpacesFixer extends AbstractFixer
                 !$nextNonWhitespaceToken->isComment()
                 || $nextNonWhitespaceIndex === $prevNonWhitespaceIndex
                 || $tokenAfterNextNonWhitespaceToken->isWhitespace(" \t")
-                || '/*' === substr($nextNonWhitespaceToken->getContent(), 0, 2)
+                || 0 === strpos($nextNonWhitespaceToken->getContent(), '/*')
             )
         ) {
             $tokens->clearAt($nextIndex);
--- a/src/Fixer/ClassNotation/ClassDefinitionFixer.php
+++ b/src/Fixer/ClassNotation/ClassDefinitionFixer.php
@@ -369,9 +369,9 @@ interface Bar extends
 
                 if ($tokens[$prevNonWhite]->isComment() || $tokens[$nextNonWhite]->isComment()) {
                     $content = $tokens[$prevNonWhite]->getContent();
-                    if (!('#' === $content || '//' === substr($content, 0, 2))) {
+                    if (!('#' === $content || 0 === strpos($content, '//'))) {
                         $content = $tokens[$nextNonWhite]->getContent();
-                        if (!('#' === $content || '//' === substr($content, 0, 2))) {
+                        if (!('#' === $content || 0 === strpos($content, '//'))) {
                             $tokens[$i] = new Token(array(T_WHITESPACE, ' '));
                         }
                     }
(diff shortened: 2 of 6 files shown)


ntzm:
Not sure why fabbot is failing, running `fix` locally doesn't change anything

keradus:
Because you are contributing to LTS branch and fabbot is running on newest (or, sometimes, almost newest) version.
For PHP CS Fixer integration with fabbot is unique and it means for 2.2 branch we are running PHP CS Fixer 2.2 (via Travis) and 2.9 (via fabbot), and in very edge cases they are some bugs fixed on higher branch only.
in this case - ignore fabbot complains (it's exactly why we don't require it to pass)

keradus:
:-1:

------------

> Faster

Disagree. this is true only in positive case - when needle is indeed there, and even then the gain is minimal.
Yet, if the needle is not there, `substr` is faster multiple times, as `strpos` has to iterate over whole string, while `substr` doesn't.

```php
<?php

$tmp = null;

$tests = [
    'miss' => str_repeat('*!', 10000),
    'hit' => 'HIT'.str_repeat('*!', 10000),
];

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 'HIT' === substr($tests['hit'], 0, 3);
}
printf("substr when hit took: %d\n", (microtime(true) - $start) * 1000);
$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 0 === strpos($tests['hit'], 'HIT');
}
printf("strpos when hit took: %d\n", (microtime(true) - $start) * 1000);

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 'HIT' === substr($tests['miss'], 0, 3);
}
printf("substr when miss took: %d\n", (microtime(true) - $start) * 1000);

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 0 === strpos($tests['miss'], 'HIT');
}
printf("strpos when miss took: %d\n", (microtime(true) - $start) * 1000);
```

```
substr when hit took: 7
strpos when hit took: 5
substr when miss took: 6
strpos when miss took: 47
```

keradus:
* - if the needle is not a prefix, then we `strpos` is always slower

```php
<?php

$tmp = null;

$tests = [
    'miss' => str_repeat('*!', 10000).str_repeat('*!', 10000),
    'hit' => str_repeat('*!', 10000).'HIT'.str_repeat('*!', 10000),
];

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 'HIT' === substr($tests['hit'], 20000, 3);
}
printf("substr when hit took: %d\n", (microtime(true) - $start) * 1000);

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = false !== strpos($tests['hit'], 'HIT');
}
printf("strpos when hit took: %d\n", (microtime(true) - $start) * 1000);

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = 'HIT' === substr($tests['miss'], 0, 3);
}
printf("substr when miss took: %d\n", (microtime(true) - $start) * 1000);

$start = microtime(true);
for ($i = 0; $i < 100000; $i++) {
  $tmp = false !== strpos($tests['miss'], 'HIT');
}
printf("strpos when miss took: %d\n", (microtime(true) - $start) * 1000);
```

```
substr when hit took: 7
strpos when hit took: 48
substr when miss took: 5
strpos when miss took: 91
```

ntzm:
Ah ok that makes a lot of sense, will close 👍 Thanks for reviewing as always :)

PR #3367, opened 2018-01-04
NoUnusedImportsFixer - fix comment handling

SpacePossum:
`NoUnusedImportsFixer`

- [x]  crashes on comments containing the '/' character.
- [x]  removes comments placed in or around `use` statements
- [x]  removes _used_ imports if comments are placed in or around `use` statements
- [x] removes _used_ `function` and `const` imports (replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3321)

The other added delimiters are not strictly needed now, however it is good practice and more future prove (and saves time checking each time).

So kind [reminder](https://3v4l.org/mB2vq);
> preg_quote($a) !== preg_quote($a, '/')

PR #3402, opened 2018-01-07
Always provide delimiter to `preg_quote` calls

ntzm:
These aren't needed and probably won't be needed, but it's good to have them just in case IMO

### DIFF
--- a/src/DocBlock/Annotation.php
+++ b/src/DocBlock/Annotation.php
@@ -217,7 +217,7 @@ class Annotation
      */
     public function setTypes(array $types)
     {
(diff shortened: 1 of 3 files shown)


SpacePossum:
good idea!

kindly note; duplicates https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3367 :)

keradus:
So, I would say it other way.
Let us make a fixer for it.

SpacePossum:
>Let us make a fixer for it.

That is going to be a highly risky one, as it is impossible to determine the delimiter that should be used.
Could be an autoreview/sniff though.

keradus:
Thank you @ntzm.

Issue #3399, opened 2018-01-07
Removing "use_yoda_style" option in "is_null" fixer


kubawerlos:
As we have `yoda_style` to handle Yoda conditions we can easily drop handling it in `is_null` fixer.

That would be split into 2 pull request:
- deprecating option to master branch
- removing option to 3.0 branch

Pinging Jedi Master @SpacePossum

SpacePossum:
sounds like a good idea to me :)
maybe the force be with your PR's!

keradus:
Kuba, could you please propose PRs ? Thanks !

keradus:
closing as done

PR #3404, opened 2018-01-07
IsNullFixer - Deprecate "use_yoda_style"

kubawerlos:
First part of https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3399

### DIFF
--- a/src/Fixer/LanguageConstruct/IsNullFixer.php
+++ b/src/Fixer/LanguageConstruct/IsNullFixer.php
@@ -201,7 +201,7 @@ final class IsNullFixer extends AbstractFixer implements ConfigurationDefinition
      */
     protected function createConfigurationDefinition()
     {
-        // @todo 3.0 drop me
+        // @todo 3.0 drop `ConfigurationDefinitionFixerInterface`
         return new FixerConfigurationResolver([
             (new FixerOptionBuilder('use_yoda_style', '(deprecated) Whether Yoda style conditions should be used.'))
                 ->setAllowedTypes(['bool'])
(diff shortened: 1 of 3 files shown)


src/Fixer/LanguageConstruct/IsNullFixer.php
```diff
@@ -188,7 +202,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     protected function createConfigurationDefinition()
     {
         return new FixerConfigurationResolver([
-            (new FixerOptionBuilder('use_yoda_style', 'Whether Yoda style conditions should be used.'))
```
keradus: `// @todo 3.0 drop me`

tests/Fixer/LanguageConstruct/IsNullFixerTest.php
```diff
@@ -52,6 +60,9 @@ public function testCorrectConfiguration()
     }
 
     /**
+     * @group legacy
+     * @expectedDeprecation Using "use_yoda_style" is deprecated and will be removed in 3.0. Use "yoda_style" fixer instead.
```
keradus: this is the only test that will be left on 3.0. instead of adding legacy group, remove `$this->fixer->configure(['use_yoda_style' => true]);` line from it.

kubawerlos: I was planning to do it in the PR for 3.0, but I'm getting your point.

kubawerlos: And not this one, we will keep `testYodaFix`.

keradus: you are right, my bad :+1:

keradus:
Thank you @kubawerlos.

PR #3405, opened 2018-01-07
Fix `fopen` mode

ntzm:
From the [manual](http://php.net/manual/en/function.fopen.php):

> To use these flags, specify either 'b' or 't' as the last character of the mode parameter.


keradus:
hm... I never experience any issues with it. Did you ?

anyway, :+1:

ntzm:
Not sure, just some code analysis flagged it up

keradus:
Thank you @ntzm.

PR #3406, opened 2018-01-07
Fix for escaping in README

kubawerlos:


README.rst
```diff
@@ -6,7 +6,7 @@ whether you want to follow PHP coding standards as defined in the PSR-1, PSR-2,
 or other community driven ones like the Symfony one.
 You can **also** define your (teams) style through configuration.
 
-It can modernize your code (like converting the `pow` function to the `**` operator on PHP 5.6)
+It can modernize your code (like converting the ``pow`` function to the ``**`` operator on PHP 5.6)
```
SpacePossum: isn't this duplicating your PR here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3408/files#diff-88b99bb28683bd5b7e3a204826ead112L941 ?

kubawerlos: This was separate as it's not a fixer description. And the point of others was to remove `()` from function names. Here we could go with complex example i.e. `pow($foo, $bar)` and `$foo ** $bar`.

SpacePossum:
@kubawerlos sorry for keep this one waiting, please rebase and we should be good to go I think

keradus:
Thank you @kubawerlos.

PR #3410, opened 2018-01-07
FixerFactoryTest - test that priority pair fixers have itest

keradus:


keradus:
@localheinz could you please take a look on this AutoReview ?

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -199,4 +214,79 @@ function ($name) {
 
         return $cases;
     }
+
+    /**
+     * @dataProvider provideFixersPriorityParisHaveItestCases
+     */
+    public function testFixersPriorityParisHaveItest(FixerInterface $first, FixerInterface $second)
```
ntzm: Typo: paris -> pairs

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -260,12 +260,13 @@ public function testFixersPriorityPairsHaveIntegrationTest(FixerInterface $first
 
     public function provideFixersPriorityPairsHaveIntegrationTestCases()
     {
+        $self = $this;
```
localheinz: Not sure if there's a convention on how to do this on PHP 5.3, but I myself have used (and seen) others use

```php
$that = $this;
```

Maybe using `$self` is a bit confusing?

keradus: I saw both conventions, no hard feeling about it

julienfalque: `var that = this` is common in Javascript. Also, `$self` could be confused with `self`. I would go for `$that`.

keradus: > `var that = this` is common in Javascript

`that` is disrespected by Crockford, he asked you to use self ;P

we gonna drop it in very next branch...

SpacePossum: :+1: for `$self` , 
I get confused when `this` is referred by `that`

keradus: guys, let go anyway here, I'm fine with both, if any of you wants - let you just change it yourself ;)
Please, let us focus on real idea behind this PR instead ;)

keradus:
Thank you @SpacePossum.

keradus:
replaced by #3425

PR #3411, opened 2018-01-07
Better type hinting for AbstractFixerTestCase::$fixer

kubawerlos:
So IDE won't complain that `Method 'configure' not found in null|\PhpCsFixer\Fixer\FixerInterface`.


SpacePossum:
thanks,
I'll leave this one to @keradus as if we merge this it should not be done on the 3.x line IIUC

keradus:
Thank you @kubawerlos.

PR #3412, opened 2018-01-07
Convert `strtolower` inside `strpos` to just `stripos`

ntzm:
Possibility for a new fixer as well?


SpacePossum:
+0, is this really faster here?

ntzm:
I haven't compared but it's 1 function call vs 2, so I assume it's faster?

julienfalque:
Not sure it's faster but it's simpler so :+1: for me.

Though this an edge case, not sure it's worth creating a fixer.

keradus:
Thank you @ntzm.

PR #3413, opened 2018-01-08
DescribeCommand: add fixer class when verbose

Slamdunk:
Because:

1. Fixer's names are dinamically generated and cannot be directly copy-pasted to `grep` (even though the manual change is trivial)
1. Namespaces aren't shown anywhere

I always lose precious minutes when bugfixing existing fixers finding where is the class of a fixer name.
I propose to output the fixer class if the `describe` command is called with verbosity flag.

fixer_class


SpacePossum:
I'm +1 for this.

About point 2;
all fixers that are not in the repo should be prefixed with a `vendor/` (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.9/src/FixerNameValidator.php#L28).
This serves as an alternative as namespace.

Slamdunk:
> About point 2;
> all fixers that are not in the repo should be prefixed with a vendor/ (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.9/src/FixerNameValidator.php#L28).
> This serves as an alternative as namespace.

I don't get this. I meant that if I want to know what is the PHP-CS-Fixer namespace of `no_useless_return`, it is not shown anywhere except in the class itself.
What did you mean?

Slamdunk:
> Please target `3.0`

Done

src/Console/Command/DescribeCommand.php
```diff
@@ -156,6 +156,10 @@ private function describeRule(OutputInterface $output, $name)
             $description .= sprintf(' <error>DEPRECATED</error>: %s.', $message);
         }
 
+        if ($output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE) {
+            $output->writeln(sprintf('Fixer class: <comment>%s</comment>.', get_class($fixer)));
+        }
+
         $output->writeln(sprintf('<info>Description of</info> %s <info>rule</info>.', $name));
```
keradus: I would put class name after `rule` , as user was asking for `rule`, so let us display `rule` first, always (regardless verbosity)

keradus:
why 3.0? I don't believe we have and bc promise on describe output

Slamdunk:
> why 3.0? I don't believe we have and bc promise on describe output

So should I have to go back to master?

SpacePossum:
```
I don't get this. I meant that if I want to know what is the PHP-CS-Fixer namespace of `no_useless_return`, it is not shown anywhere except in the class itself.
What did you mean?
```

Since `no_useless_return` does _not_ start with `^[A-Z][a-zA-Z0-9]*\/` it cannot be a custom fixer, so the fixer must be from this repo, so the namespace of the fixer is the same as of this repo. That is all I tried to state.

SpacePossum:
>why 3.0? I don't believe we have and bc promise on describe output

I don't mind anyway btw.

Slamdunk:
> it cannot be a custom fixer, so the fixer must be from this repo, so the namespace of the fixer is the same as of this repo

Ok sorry, so let me explain better what I meant: for me the namespace of `no_useless_return` is `ReturnNotation`, its folder. That particular namespace of all non-custom fixers is hidden in the doc.

Slamdunk:
I have no idea what have I done, need to reopen this on a new PR 😢

keradus:
> That particular namespace of all non-custom fixers is hidden in the doc.

it is hidden in docs, as regular user don't care about it

PR #3414, opened 2018-01-08
DescribeCommand: add fixer class when verbose

Slamdunk:
Following https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3413 as I have no idea why Github doesn't allow changing the branch back to master.

Also implemented https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3413#discussion_r160126392


keradus:
Thank you @Slamdunk.

PR #3415, opened 2018-01-08
FileFilterIterator - input checks and utests

SpacePossum:
Tree shake of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3364 ,
removing `string[]` support even more will be done in a follow up PR.

### DIFF
--- a/tests/Runner/FileFilterIteratorTest.php
+++ b/tests/Runner/FileFilterIteratorTest.php
@@ -12,13 +12,14 @@
 
 namespace PhpCsFixer\Tests\Runner;
 
-use PhpCsFixer\Cache\CacheManagerInterface;
 use PhpCsFixer\FixerFileProcessedEvent;
 use PhpCsFixer\Runner\FileFilterIterator;
 use PHPUnit\Framework\TestCase;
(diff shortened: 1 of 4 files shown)


tests/Runner/FileFilterIteratorTest.php
```diff
@@ -0,0 +1,188 @@
...
+            FixerFileProcessedEvent::NAME,
+            function () {
+                throw new \Exception('No event expected.');
+            }
+        );
+
+        $filter = new FileFilterIterator(
+            new \ArrayIterator(array(
+                new \SplFileInfo(__DIR__),
+                new \SplFileInfo('__INVALID__'),
```
SpacePossum: not sure about these type of invalid `\SplFileInfo` objects, should we try to detect and throw an exception or silent ignore?

keradus: I wouldn't go that far, it will fail anyway on reading the content, right ?

SpacePossum: for now we would skip the entry because it is not a file, we used to crash on these

keradus:
Thank you @SpacePossum.

PR #3420, opened 2018-01-09
SingleLineCommentStyleFixer - fix 'strpos() expects parameter 1 to be string, boolean given'

keradus:
broken since #3318

### DIFF
--- a/src/Fixer/Comment/SingleLineCommentStyleFixer.php
+++ b/src/Fixer/Comment/SingleLineCommentStyleFixer.php
@@ -116,7 +116,8 @@ $c = 3;
             }
 
             $content = $token->getContent();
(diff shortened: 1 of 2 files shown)


keradus:
cc @kubawerlos

keradus:
I will keep your added test @SpacePossum , but keep in mind that we already did have this test in codebase - I caught this error as Travis nightly build crashed on it (nightly is not stable, different workers will bring different nightly version, that's why we don't see this crash usually)

keradus:
Thank you @SpacePossum.

SpacePossum:
oh I completely missed we already had an unstable test for the case! :)
thanks for the fix!

Issue #3422, opened 2018-01-10
FunctionToConstant - add "get_called_class"


SpacePossum:
```diff
-echo get_called_class();
+echo static::class;
```

```php
<?php

class A
{
    public function func()
    {
        var_dump(
            get_called_class(),
            static::class
        );
    }
}

class B extends A
{

}

$b = new B();
$b->func();

var_dump(
    get_called_class(),
    static::class
);

```

https://3v4l.org/AbMhs 

There is a small difference in behavior when calling "get_called_class" vs. using "static::class" outside of a class. Since this use case doesn't make sense and triggers a warning anyway I think it is not worth the trouble detecting if call is outside of a class.

Available since PHP 5.5

SpacePossum:
discussion can continue in the PR.

PR #3423, opened 2018-01-10
FunctionToConstantFixer - add fix "get_called_class" option.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3422


julienfalque:
Looks good 👍 but shouldn't the fixer check whether `get_called_class` is called from inside a class?

SpacePossum:
from the issue description;

```
There is a small difference in behavior when calling "get_called_class" vs. using "static::class" outside of a class. Since this use case doesn't make sense and triggers a warning anyway I think it is not worth the trouble detecting if call is outside of a class.
```

for me it is like using `__CLASS__` outside a class scope;
it doesn't work, triggers a warning, it is a mess, I don't think we've to reproduce that mess using different code or prevent/stop changing one mess into another.

src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -142,32 +148,38 @@ protected function createConfigurationDefinition()
                 ->setAllowedValues([
                     (new FixerOptionValidatorGenerator())->allowedValueIsSubsetOf($functionNames),
                 ])
-                ->setDefault($functionNames)
+                ->setDefault([
+                    'get_class',
+                    'php_sapi_name',
+                    'phpversion',
+                    'pi',
```
keradus: shouldn't we add `get_called_class` to local config and TODO for v3 to add it by default ?

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks for reviewing @julienfalque and merging @keradus :)

keradus:
@SpacePossum , could you resolve todo on 3.x line ?

PR #3424, opened 2018-01-10
AbstractFixerTest: fix expectException arguments

Slamdunk:
`expectException` accepts only one argument; the second one must be moved to `expectExceptionMessage`

### DIFF
--- a/tests/Fixer/AbstractFixerTest.php
+++ b/tests/Fixer/AbstractFixerTest.php
@@ -27,7 +27,8 @@ final class AbstractFixerTest extends TestCase
     {
         $fixer = new UnconfigurableFixer();
 
-        $this->expectException(\LogicException::class, 'Cannot configure using Abstract parent, child not implementing "PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface".');
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessage('Cannot configure using Abstract parent, child not implementing "PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface".');
 
         $fixer->configure(['foo' => 'bar']);
     }
@@ -36,7 +37,8 @@ final class AbstractFixerTest extends TestCase
     {
         $fixer = new UnconfigurableFixer();
 
-        $this->expectException(\LogicException::class, 'Cannot get configuration definition using Abstract parent, child not implementing "PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface".');
+        $this->expectException(\LogicException::class);
+        $this->expectExceptionMessage('Cannot get configuration definition using Abstract parent, child not implementing "PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface".');
 
         $fixer->getConfigurationDefinition();
     }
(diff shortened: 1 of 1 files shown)


tests/Fixer/AbstractFixerTest.php
```diff
@@ -54,7 +57,7 @@ public function testSetWhitespacesConfigUnconfigurable()
     {
         $fixer = new UnconfigurableFixer();
 
-        $this->expectException('LogicException');
+        $this->expectException(\LogicException::class);
```
keradus: was done already in #3421

keradus:
Good catch, thanks @Slamdunk.

Slamdunk:
Found thanks to [PHPStan](https://github.com/phpstan/phpstan). I suggest to integrate it for `3.0`

keradus:
why 3.0 and not 2.0? it will make maintainability between 2<>3 a hell.
I have experience a lot of false positives with phpstan :(

Slamdunk:
> why 3.0 and not 2.0? it will make maintainability between 2<>3 a hell.

Nothing special:

1. it requires PHP 7
1. I don't like using the phar distribution because is harder to extend and to use with other vendor's extensions
1. so I would go with `composer require`, but would conflict with build on older php versions

But all this can be solved:

1. we can create a dedicated stage with newest PHP version
1. `composer remove` in other builds phpstan and its dependencies
1. or `composer require` only in that stage the needed dependencies (but the dev couldn't run it on his/her pc)

It would not be a big deal: static analysis tools are not run as frequently as php-cs-fixer or phpunit, we could let run that check on travis only, if you don't want to change che `composer.json`

> I have experience a lot of false positives with phpstan :(

I had bad experiences with all static analysis tools out there, and a lot of false positives too: this is a little cost to accept with each and every of them, because you know, they don't run the code.
After struggling a lot, I chose PHPStan because:

1. Is pure PHP, no extensions needed
1. It loads class autoloaders, so knows the project
1. It's easy to handle false-positives
1. Is trivial to implement custom checks, the test suite is well written. As an example, I just made a rule that require every string that resambles a know class to be written with `::class` notation: here the [test](https://github.com/Slamdunk/phpstan-extensions/blob/master/tests/StringToClassRuleTest.php), the [test-asset](https://github.com/Slamdunk/phpstan-extensions/blob/master/tests/TestAsset/FooStringToClass.php) and the final [rule](https://github.com/Slamdunk/phpstan-extensions/blob/master/lib/StringToClassRule.php).
1. Has MIT licence

I can try a PR: what branch would you me to target? `2.2` or `2.9`?

keradus:
for most of your wondering, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.9/.travis.yml#L27

you know, for our case, we write fixers, like with your example of `::class`, you are more than aware about https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2918 ;) yet indeed, a lot of work is to be done to execute first context-aware fixer via PHP CS Fixer

if one is using any SCA tool, it's great. yet no one does everything, that's why i'm not afraid of using many of them. give it a try starting since 2.2, if some rules won't be able to execute on 5.3, we will enable it on master

PR #3431, opened 2018-01-11
Add missing tests

SpacePossum:


src/PharChecker.php
```diff
@@ -22,6 +22,15 @@
      */
     public function checkFileValidity($filename)
     {
+        if (!is_string($filename)) {
```
keradus: we don't check primitive types described in docblocks anywhere, and they will be strongly typehinted on 7+

tests/Console/Output/NullOutputTest.php
```diff
@@ -0,0 +1,34 @@
...
+ * @covers \PhpCsFixer\Console\Output\NullOutput
+ */
+final class NullOutputTest extends TestCase
+{
+    public function testNullOutput()
+    {
+        $output = new NullOutput();
+        $output->printLegend();
+        // no exceptions, no output, so OK
+        $this->addToAssertionCount(1);
```
keradus: oh my... looks extra hacky to bypass "no assertion in test" warnign

tests/FixerFileProcessedEventTest.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $status = FixerFileProcessedEvent::STATUS_NO_CHANGES;
+        $event = new FixerFileProcessedEvent($status);
+
+        $this->assertSame($status, $event->getStatus());
+        $this->assertFalse($event->isPropagationStopped());
+
+        $event->stopPropagation();
+
+        $this->assertTrue($event->isPropagationStopped());
```
keradus: it's testing logic of Sf, not ours

src/PharChecker.php
```diff
@@ -31,7 +40,7 @@ public function checkFileValidity($filename)
                 throw $e;
             }
 
-            return $e->getMessage();
+            return 'Failed to create phar. '.$e->getMessage();
```
keradus: it sounds like it `Failed to create phar file`, not `Failed to create Phar instance`

tests/Console/Output/NullOutputTest.php
SpacePossum: no, there is no `assertNoExceptionHasBeenThrown` and/or `assertNoEchoOrPrintHasBeenDoneToStdOutOrStdErr` AFAIK

src/PharChecker.php
SpacePossum: lets drop this when we are on PHP7 than

keradus: but we don't do any of those checks, anywhere...
either everywhere or nowhere, staying in the middle is weird

tests/Console/Output/NullOutputTest.php
keradus: maybe raise a concern on phpunit ?
either we will know the official way, or it will be a feature request

tests/Linter/ProcessLintingResultTest.php
```diff
@@ -0,0 +1,78 @@
...
+        ;
+
+        $process
+            ->isSuccessful()
+            ->willReturn(true)
+        ;
+
+        $result = new ProcessLintingResult($process->reveal());
+        $result->check();
+        $result->check(); // call twice
```
keradus: why we are calling it twice ?

tests/Linter/ProcessLintingResultTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+        $this->addToAssertionCount(1); // no exception, so we are good here
```
keradus: maybe set expectation on how much times prophesized methods shall be called ? then, we will have assertions and no need for this trick

tests/Linter/ProcessLintingResultTest.php
SpacePossum: `ProcessLintingResult` caches the result internal, so this tests checks that path as well (`if (null === $this->isSuccessful) {`)

tests/Console/Output/NullOutputTest.php
julienfalque: Maybe https://github.com/sebastianbergmann/phpunit/issues/2484#issuecomment-278233214

SpacePossum: `@doesNotPerformAssertions` seems to effect coverage :|

tests/Linter/ProcessLintingResultTest.php
keradus: you are executing the path twice, but you are not testing was the path different.

if one would drop internal cache of the class, your tests would still pass, which means that you are not testing internal cache.

src/PharChecker.php
SpacePossum: removed

tests/FixerFileProcessedEventTest.php
SpacePossum: removed

src/PharChecker.php
SpacePossum: updated

tests/Linter/ProcessLintingResultTest.php
SpacePossum: removed

tests/Linter/ProcessLintingResultTest.php
SpacePossum: -1, I don't want to test how often the (prophesized) methods are called internally.
PHPUnit really misses a default which I always went about like this. Maybe I reopen the discussion about this there some day.

keradus: @SpacePossum , you don't need to count them, use `->shouldBeCalled()`.
if you don't call them, no need to prophesize them, right ?

tests/FixerConfiguration/FixerOptionValidatorGeneratorTest.php
```diff
@@ -0,0 +1,52 @@
...
+ * @internal
+ *
+ * @covers \PhpCsFixer\FixerConfiguration\FixerOptionValidatorGenerator
+ */
+final class FixerOptionValidatorGeneratorTest extends TestCase
+{
+    /**
+     * @param bool  $expected
+     * @param array $input
+     * @param mixed $input
```
keradus: input input ;)

SpacePossum:
I think this batch is big enough now, so lets process this PR :)

tests/Console/Output/NullOutputTest.php
keradus: @SpacePossum , when I use the annotation, I see the coverage:
https://github.com/sebastianbergmann/phpunit/issues/2484#issuecomment-357230090

SpacePossum: updated to use the annotation :+1:

tests/Linter/TokenizerLintingResultTest.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    /**
+     * @doesNotPerformAssertions
+     */
+    public function testTokenizerLintingResultOK()
+    {
+        $result = new TokenizerLintingResult();
+        $result->check();
+
+        $this->addToAssertionCount(1);
```
keradus: is it still needed when we do have annotation now ?

SpacePossum: no, not sure why I only removed the comment :|

tests/Console/Output/NullOutputTest.php
```diff
@@ -0,0 +1,33 @@
...
+use PhpCsFixer\Console\Output\NullOutput;
+use PhpCsFixer\Tests\TestCase;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Console\Output\NullOutput
+ * @doesNotPerformAssertions
```
keradus: it doesn't work on class level, only on method-level, at least at phpunit 5.6

SpacePossum:
ping @keradus anything left here you want me to update?

keradus:
not sure, I was awaiting for someone to review it ;)

keradus:
Thank you @SpacePossum.

Issue #2538, opened 2017-02-13
PHPUnit setup and teardown methods should be protected


fabpot:
What about fixing PHPUnit `setup()` and `teardown()` methods to be `protected` when they are `public`? That's a common mistake and would be nice to have a fixed for that.

SpacePossum:
Would be nice to fixer. 

How would the detection if we are fixing a utest class look like (as both methods might be used in other classes as well);
- must be method of a utest class
- class is a likely to be a utest class if declared in a namespace that contains `test`; or
- class extends `\PHPUnit_Framework_TestCase`; or
- or class extends `PHPUnit\Framework\TestCase`; or
- class name ends with `Test`

Still risky fixer as all tests would not be fool proof, wonder if others have an idea to do the detection better :)

Bilge:
Not worth the effort. Only a perfectionist would care about the difference; it has no bearing on functionality whatsoever.

keradus:
actually we had same issue and had to fix it.
I would simplify check - `*Test(?:Case)?` in name or parent class.

sebastianbergmann:
While you're at it, rename `setup` to `setUp` and `teardown` to `tearDown`.

PR #3434, opened 2018-01-11
Add PhpUnitSetUpTearDownVisibilityFixer

BackEndTea:
This fixer changes the setup and teardown function of phpunit classes to protected, to match its parent.

As talked about in #2538

This is ready for review

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1125,6 +1125,11 @@ Choose from the list of available rules:
     ``'newest'``
   - ``use_class_const`` (``bool``): use ::class notation; defaults to ``true``
 
+* **php_unit_set_up_tear_down_visibility**
+
+  Changes the visibility of the setUp and tearDown functions of phpunit to
+  protected, to match its parent.
+
 * **php_unit_strict**
 
   PHPUnit methods like ``assertSame`` should be used instead of
--- a/README.rst
+++ b/README.rst
@@ -1128,7 +1128,7 @@ Choose from the list of available rules:
 * **php_unit_set_up_tear_down_visibility**
 
   Changes the visibility of the setUp and tearDown functions of phpunit to
-  protected, to match its parent.
+  protected, to match the PHPUnit TestCase.
 
 * **php_unit_strict**
 
--- a/README.rst
+++ b/README.rst
@@ -1130,6 +1130,8 @@ Choose from the list of available rules:
   Changes the visibility of the setUp and tearDown functions of phpunit to
   protected, to match the PHPUnit TestCase.
 
+  *Risky rule: this fixer may change functions named setUp or tearDown outside of PHPUnit tests, when a class is wrongly seen as a PHPUnit test.*
+
 * **php_unit_strict**
 
   PHPUnit methods like ``assertSame`` should be used instead of
--- a/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
@@ -50,8 +50,19 @@ final class MyTest extends \PHPUnit_Framework_TestCase
 }
 '
                 ),
-            ]
+            ],
+            null,
+            'This fixer may change functions named setUp or tearDown outside of PHPUnit tests, '.
+            'when a class is wrongly seen as a PHPUnit test.'
         );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
     }
 
     /**
--- a/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
@@ -67,7 +67,7 @@ final class MyTest extends \PHPUnit_Framework_TestCase
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        foreach (array_reverse($this->findPhpUnitClasses($tokens)) as $indexes) {
+        foreach ($this->findPhpUnitClasses($tokens) as $indexes) {
             $this->fixSetUpAndTearDown($tokens, $indexes[0], $indexes[1]);
         }
     }
@@ -75,23 +75,19 @@ final class MyTest extends \PHPUnit_Framework_TestCase
     /**
      * @param Tokens $tokens
      *
-     * @return int[][] array of [start, end] indexes from sooner to later classes
+     * @return \Generator
      */
     private function findPhpUnitClasses(Tokens $tokens)
     {
         $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
-        $phpunitClasses = [];
 
-        for ($index = 0, $limit = $tokens->count() - 1; $index < $limit; ++$index) {
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
             if ($tokens[$index]->isGivenKind(T_CLASS) && $phpUnitTestCaseIndicator->isPhpUnitClass($tokens, $index)) {
-                $index = $tokens->getNextTokenOfKind($index, ['{']);
-                $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
-                $phpunitClasses[] = [$index, $endIndex];
-                $index = $endIndex;
+                $start = $tokens->getNextTokenOfKind($index, ['{']);
+                $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $start);
+                yield [$start, $endIndex];
             }
         }
-
-        return $phpunitClasses;
     }
 
     /**
@@ -101,6 +97,7 @@ final class MyTest extends \PHPUnit_Framework_TestCase
      */
     private function fixSetUpAndTearDown(Tokens $tokens, $startIndex, $endIndex)
     {
+
         $counter = 0;
         $tokensAnalyzer = new TokensAnalyzer($tokens);
 
--- a/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
@@ -97,7 +97,6 @@ final class MyTest extends \PHPUnit_Framework_TestCase
      */
     private function fixSetUpAndTearDown(Tokens $tokens, $startIndex, $endIndex)
     {
-
         $counter = 0;
         $tokensAnalyzer = new TokensAnalyzer($tokens);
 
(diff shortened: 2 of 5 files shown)


SpacePossum:
Hi, thanks for the PR, good idea to get this done :)

README.rst
```diff
@@ -1125,6 +1125,11 @@ Choose from the list of available rules:
     ``'newest'``
   - ``use_class_const`` (``bool``): use ::class notation; defaults to ``true``
 
+* **php_unit_set_up_tear_down_visibility**
+
+  Changes the visibility of the setUp and tearDown functions of phpunit to
+  protected, to match its parent.
```
SpacePossum: I would reword a bit, as the new casing might not be the one from the direct parent, it should always be the one of the PHPUnit class I think?

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_FUNCTION]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach (array_reverse($this->findPhpUnitClasses($tokens)) as $indexes) {
```
SpacePossum: Maybe it is more efficient to search backwards and store the result, than the `array_reverse` is not needed.
Even more optimized code is possible by using `yield` in the `findPhpUnitClasses`, but this is not the most important optim here ;)

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     * @param int    $endIndex
+     */
+    private function fixSetUpAndTearDown(Tokens $tokens, $startIndex, $endIndex)
+    {
+        $tokensAnalyzer = new TokensAnalyzer($tokens);
```
SpacePossum: please keep two flags, when both methods are seen (and fixed or not doesn't matter); do an early return

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+    private function isMethodSetUpOrTearDown(Tokens $tokens, $index)
+    {
+        $functionNameIndex = $tokens->getNextMeaningfulToken($index);
+        $functionName = \strtolower($tokens[$functionNameIndex]->getContent());
+
+        if ('setup' !== $functionName && 'teardown' !== $functionName) {
+            return false;
+        }
+
+        return true;
```
SpacePossum: can be one line `return 'setup' !== $functionName && 'teardown' !== $functionName;`

tests/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixerTest.php
```diff
@@ -0,0 +1,246 @@
...
+class FixerTest extends \PhpUnit\FrameWork\TestCase
+{
+
+    public function setUp() {}
+
+    public function tearDown() {}
+}
+',
+            ],
+            'setUp and tearDown stay protected if they are' => [
```
SpacePossum: great idea, however you already test this as part of the `expected` result in the previous test case

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+            $visibility = $tokensAnalyzer->getMethodAttributes($i)['visibility'];
+
+            if (T_PROTECTED === $visibility) {
+                continue;
+            }
+
+            if (T_PUBLIC === $visibility) {
+                do {
+                    $i = $tokens->getPrevMeaningfulToken($i);
+                } while (!$tokens[$i]->isGivenKind(T_PUBLIC));
```
SpacePossum: you can use `$tokens->getPrevTokenOfKind` to find the `T_PUBLIC` as replacement of this loop

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     * @param int    $endIndex
+     */
+    private function fixSetUpAndTearDown(Tokens $tokens, $startIndex, $endIndex)
+    {
+        $tokensAnalyzer = new TokensAnalyzer($tokens);
+        for ($i = $endIndex - 1; $i > $startIndex; --$i) {
+            if (!$this->isMethod($tokens, $i)) {
```
SpacePossum: the `isMethod` method content can be moved into `isMethodSetUpOrTearDown`, but that is just my test

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -0,0 +1,165 @@
...
+                continue;
+            }
+            if (!$this->isMethodSetUpOrTearDown($tokens, $i)) {
+                continue;
+            }
+
+            $visibility = $tokensAnalyzer->getMethodAttributes($i)['visibility'];
+
+            if (T_PROTECTED === $visibility) {
```
SpacePossum: `|| T_PRIVATE`
or maybe more defensive would to act only `T_PUBLIC` or `null`

BackEndTea: No problem, but having a setUp or tearDown as a private function would throw an error on execution of the tests.

SpacePossum: indeed it would, however we might be fixing a non-PHPUnit class here that happens to look like one

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
BackEndTea: I'm not quite sure how to do that, so if possible, i'd like some help with that.

tests/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixerTest.php
BackEndTea: True, but this test explicitly makes sure nothing changes if they are protected.
So even if it may be redundant, i personally like having that extra check. If needed i can remove it though.

BackEndTea:
@SpacePossum This is once again ready for review

tests/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixerTest.php
SpacePossum: I understand, but you already have the extra test, when ever you pass;
```
        $this->doTest(
            $a,
            $b
        );
```

it will always also test;

```
        $this->doTest(
            $a,
            null
        );
```

so if you have two test with same value for `$a` than one repeats the other

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
SpacePossum: the easiest way would be to iterate backwards in `findPhpUnitClasses`,

```php
for ($index = $tokens->count() - 1; $index >= 0; --$index) {
```

there should be bunch of fixers that do this already which might help as an example. Maybe I can help you out later with this.

tests/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixerTest.php
```diff
@@ -0,0 +1,282 @@
...
+ *
+ * @covers \PhpCsFixer\Fixer\PhpUnit\PhpUnitSetUpTearDownVisibilityFixer
+ */
+final class PhpUnitSetUpTearDownVisibilityFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed      $expected
+     * @param null|mixed $input
```
SpacePossum: little left over; `mixed` should be `string` I think

src/Fixer/PhpUnit/PhpUnitSetUpTearDownVisibilityFixer.php
```diff
@@ -67,31 +67,27 @@ public function isCandidate(Tokens $tokens)
...
-        for ($index = 0, $limit = $tokens->count() - 1; $index < $limit; ++$index) {
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
             if ($tokens[$index]->isGivenKind(T_CLASS) && $phpUnitTestCaseIndicator->isPhpUnitClass($tokens, $index)) {
-                $index = $tokens->getNextTokenOfKind($index, ['{']);
-                $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
-                $phpunitClasses[] = [$index, $endIndex];
-                $index = $endIndex;
+                $start = $tokens->getNextTokenOfKind($index, ['{']);
+                $endIndex = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $start);
+                yield [$start, $endIndex];
```
SpacePossum: thanks for this change, code looks nice and fast : )
I'll test again soon

SpacePossum:
@BackEndTea if you've the time, could you rebase your branch and add this new fixer to the [.php_cs.dist](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/.php_cs.dist) of the project?

keradus:
> btw. is this fixer `risky`

wow, why it is not marked as risky ?

BackEndTea:
@keradus @SpacePossum Is this good for the next release, or are there still things that need to be fixed?

SpacePossum:
@BackEndTea please pull my commits in if you want to keep up working in this.
I think this looks good, but we've to wait for people to find the time to merge and release and such, please be patient with us, busy days! :)

BackEndTea:
No problem, just wanted to make sure that you weren't waiting for something from my side.

keradus:
@BackEndTea 
@SpacePossum , please finish review

SpacePossum:
I finished my review 19 days ago, we are waiting for a second OK

keradus:
ah, didn't saw your approval on PR ;)

keradus:
Thank you @BackEndTea.

BackEndTea:
Thank you @keradus @SpacePossum

PR #3438, opened 2018-01-12
PhpUnitTestAnnotationFixer: Do not prepend with test if method is test()

### LINKED ISSUE #3437: PhpUnitTestAnnotationFixer false positive since 2.10
It now duplicates `test()` name to `testTest()`. It didn't happen before 2.10.

What was changed?

image

See also Travis: https://travis-ci.org/Symplify/Symplify/jobs/327849604#L644

localheinz:
This PR

* [x] fixes the edge case

Fixes #3437.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -300,7 +300,7 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
             return false;
         }
 
-        if  (strlen($functionName) === 4) {
+        if (4 === strlen($functionName)) {
             return true;
         }
 
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -300,7 +300,7 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
             return false;
         }
 
-        if (4 === strlen($functionName)) {
+        if ('test' === $functionName) {
             return true;
         }
 
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -296,8 +296,12 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
      */
     private function hasTestPrefix($functionName)
     {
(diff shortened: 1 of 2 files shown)


src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -296,10 +296,14 @@ private function getDocBlockIndex(Tokens $tokens, $index)
      */
     private function hasTestPrefix($functionName)
     {
-        if (!$this->startsWith('test', $functionName) || strlen($functionName) < 5) {
+        if (!$this->startsWith('test', $functionName)) {
             return false;
         }
 
+        if (4 === strlen($functionName)) {
```
keradus: what we do for <4 now ?

SpacePossum: `abc` becomes `testAbc` IIUC

localheinz: We won’t get this far if `$functionName` doesn’t start with `test`, see condition above.

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -296,10 +296,14 @@ private function getDocBlockIndex(Tokens $tokens, $index)
      */
     private function hasTestPrefix($functionName)
     {
-        if (!$this->startsWith('test', $functionName) || strlen($functionName) < 5) {
+        if (!$this->startsWith('test', $functionName)) {
             return false;
         }
 
+        if ('test' === $functionName) {
```
localheinz: @keradus @SpacePossum 

Is this better?

SpacePossum: :+1: it is more explicit and faster, does this solution work for `TesT` etc. as well?

localheinz: I can adjust!

I think, though, that for PHPUnit it needs to be `test` (case-sensitive) to qualify as test.

localheinz: Also see [`PhpUnitTestAnnotationFixer::startsWith()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.9.1/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php#L257-L262), where we compare case-sensitively.

SpacePossum: ah that makes sense! never mind me :)

keradus:
Thank you @localheinz.

localheinz:

PR #3439, opened 2018-01-12
Bug MultilineCommentOpeningClosingFixer - do not remove PHPDoc

kubawerlos:
Ping @Slamdunk

### DIFF
--- a/src/Fixer/Comment/MultilineCommentOpeningClosingFixer.php
+++ b/src/Fixer/Comment/MultilineCommentOpeningClosingFixer.php
@@ -76,18 +76,24 @@ EOT
                 continue;
             }
 
+            $tokenId = $token->getId();
             $newContent = $originalContent;
 
             // Fix opening
             if ($token->isGivenKind(T_COMMENT)) {
-                $newContent = preg_replace('/^\\/\\*\\*+/', '/*', $newContent);
+                $replacement = '/*';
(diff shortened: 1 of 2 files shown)


Slamdunk:
This PR goes against the purpose of the fixer: https://3v4l.org/nYudj

`/*** foo */` is **not** a `T_DOC_COMMENT`, is a simple `T_COMMENT`.

The fixer aims to have only two way of writing multiline comments: one for `T_DOC_COMMENT` and one for `T_COMMENT`.
It shouldn't be up to `MultilineCommentOpeningClosingFixer` to fix the human error you found.

kubawerlos:
> it was never phpdoc

only in the meaning of PHP tokens, yet it is common mistake/practice to do [it](https://github.com/phingofficial/phing/blob/master/classes/phing/tasks/system/SwitchTask.php#L91) and PHPStorm also treats it as PHPDoc.

> turning comment into phpdoc is a serparated rule

which rule?

> It shouldn't be up to MultilineCommentOpeningClosingFixer to fix the human error you found

well, half of the fixers could be seen this way.

Slamdunk:
> which rule?

Would be a new rule that if a `T_COMMENT` would have a meaning to be a `T_DOC_COMMENT`, it should be fixed to that type.

Such rules doesn't exists yet, you can PR it.

kubawerlos:
Ok, that make sense.

PR #3436, opened 2018-01-11
Add a handful of integration tests

BackEndTea:

PR #3444, opened 2018-01-14
IntegrationTest - ensure tests in priority dir are priority tests indeed

keradus:
In current state, this PR is expected to fail.

Raised by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3436#discussion_r161375420 .

We do have integration tests. What for ?
Sometimes we are not sure how 2+ fixers would play together, so we do make an integration test between them.

Integration test runner is smart enough to double check priorities:
 - if priorities are the same, the expected outcome must be the same regardless of order that fixers would be applied,
- if priorities are different, we don't care.

Now the same, but from fixed code point of view:
- if `$fixedInputCode === $fixedInputCodeWithReversedFixers`, then priorities doesn't matter,
- if `$fixedInputCode !== $fixedInputCodeWithReversedFixers`, then priorities have to be different.

This is all what matters for end-user.
This is all that's needed to be stable, and whenever 2+ fixers becomes priority-related or stop being one, the tests are checking they are still running together.
This is all that is needed for devs writing custom rules.

-------------------------

Now, at this repo we went one step forward.
We do know by heart that some of the fixers have priority conflicts and we created auto review for them that checks explicit priority and fact that priority itest exist, as addition to itest itself (ref `\PhpCsFixer\Tests\AutoReview\FixerFactoryTest`).
It was super important at beginning (and still important, yet not that much), when we basically didn't have any itest at all. This explicit priority comparison was the only point when we were checking the order of fixers is what we assumed, even when it was not tested by itest.

Now, theoretically, we could drop distinguishes between `misc` and `priority` itests, then we shall reject this PR, or we could keep it and finish this PR.
There is a lot of boilerplate added in this PR that allows itests to be extended even more, yet don't focus on the implementation yet, first let us discuss the idea behind this PR.
To ease what is a real change, I assumed we are using `tests\test` as external package that have to be generic, and adjust project-specific tests in our `IntegrationTest` directly. I also added 2 notes to put focus on that few lines that are project-specific.

Also, keep in mind that apparently not all of `priority` tests are passing new check, what is proving that they (the few that are not passing) are not priority tests in the end...


tests/Test/InternalIntegrationCaseFactory.php
```diff
@@ -0,0 +1,35 @@
...
+final class InternalIntegrationCaseFactory extends IntegrationCaseFactory
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function determineSettings(SplFileInfo $file, $config)
+    {
+        $parsed = parent::determineSettings($file, $config);
+
+        $parsed['isExplicitPriorityCheck'] = in_array('priority', explode(DIRECTORY_SEPARATOR, $file->getRelativePathname()), true);
```
keradus: note 1

tests/IntegrationTest.php
```diff
@@ -40,4 +42,34 @@ protected static function getTempFile()
...
+    {
+        parent::assertRevertedOrderFixing($case, $fixedInputCode, $fixedInputCodeWithReversedFixers);
+
+        $settings = $case->getSettings();
+
+        if (!isset($settings['isExplicitPriorityCheck'])) {
+            static::markTestIncomplete('Missing `isExplicitPriorityCheck` extension setting.');
+        }
+
+        if ($settings['isExplicitPriorityCheck']) {
```
keradus: note 2

keradus:
cc @SpacePossum as author of original concern and team member
cc @julienfalque as team member
also, cc @localheinz for his great autoreview experience

SpacePossum:
Great work, love to see this! IIRC we used to test something like this, but it was removed because the implementation was not very good. 
I'm going to review / play around with this soon :)

keradus:
Grep for `shouldCheckPriority`, it is still there for compat reasons and was deprecated in #2483.
It was enabling the check to test that rules have different priorities.
It was deprecated to see that if output depends on order, the priorities are set.

Yet now, we want to do it from other side, if we are setting the priorities, output must depend on it.

keradus:
@SpacePossum , could you review please ?

tests/IntegrationTest.php
```diff
@@ -40,4 +42,34 @@ protected static function getTempFile()
...
+            static::assertSame(
+                $settings['isExplicitPriorityCheck'],
```
SpacePossum: this is always `true`-ish here, because of line 67

SpacePossum:
one minor and good to go, very nice stuff!

tests/IntegrationTest.php
keradus: done

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -216,6 +216,7 @@ public function testThatTestClassesAreAbstractOrFinal($className)
         $rc = new \ReflectionClass($className);
 
         $this->assertTrue(
+            $rc->isInterface() || // due to hhvm only, @TODO remove me whem hhvm support is dropped
```
keradus: hhvm is soooo stupid

SpacePossum:
thanks @keradus :D

PR #3446, opened 2018-01-14
Add PregWrapper

kubawerlos:
Solution for https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3309 and idea how to handle non-UTF8 strings.

Of course, as @Slamdunk already [mendioned](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3309#issuecomment-351030840), with `AutoReview` class to check if all `preg_*` functions are wrapped.

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocTrimFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTrimFixer.php
@@ -15,7 +15,7 @@ namespace PhpCsFixer\Fixer\Phpdoc;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\PregWrapper;
+use PhpCsFixer\Preg;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
@@ -92,7 +92,7 @@ final class Foo {}
      */
     private function fixStart($content)
     {
-        return PregWrapper::replace(
+        return Preg::replace(
             '~
                 (^/\*\*)                  # DocComment begin
                 (?:
@@ -115,7 +115,7 @@ final class Foo {}
      */
     private function fixEnd($content)
     {
-        return PregWrapper::replace(
+        return Preg::replace(
             '~
                 (\R[ \t]*(?:\*[ \t]*)?\S.*?) # last line with useful content
                 (?:
--- a/src/Fixer/Phpdoc/PhpdocTrimFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTrimFixer.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Fixer\Phpdoc;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\PregWrapper;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
@@ -91,7 +92,7 @@ final class Foo {}
      */
     private function fixStart($content)
     {
-        return preg_replace(
+        return PregWrapper::replace(
             '~
                 (^/\*\*)                  # DocComment begin
                 (?:
@@ -114,7 +115,7 @@ final class Foo {}
      */
     private function fixEnd($content)
     {
-        return preg_replace(
+        return PregWrapper::replace(
             '~
                 (\R[ \t]*(?:\*[ \t]*)?\S.*?) # last line with useful content
                 (?:
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -20,11 +20,11 @@ namespace PhpCsFixer;
 final class Preg
 {
     /**
-     * @param string   $pattern
-     * @param string   $subject
-     * @param string[] &$matches
-     * @param int      $flags
-     * @param int      $offset
+     * @param string        $pattern
+     * @param string        $subject
+     * @param null|string[] &$matches
+     * @param int           $flags
+     * @param int           $offset
      *
      * @throws \RuntimeException
      *
@@ -46,11 +46,11 @@ final class Preg
     }
 
     /**
-     * @param string   $pattern
-     * @param string   $subject
-     * @param string[] &$matches
-     * @param int      $flags
-     * @param int      $offset
+     * @param string        $pattern
+     * @param string        $subject
+     * @param null|string[] &$matches
+     * @param int           $flags
+     * @param int           $offset
      *
      * @throws \RuntimeException
      *
@@ -76,7 +76,7 @@ final class Preg
      * @param string|string[] $replacement
      * @param string|string[] $subject
      * @param int             $limit
-     * @param int             $count
+     * @param null|int        &$count
      *
      * @throws \RuntimeException
      *
@@ -102,7 +102,7 @@ final class Preg
      * @param callable        $callback
      * @param string|string[] $subject
      * @param int             $limit
-     * @param int             $count
+     * @param null|int        &$count
      *
      * @throws \RuntimeException
      *
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer;
 
 /**
+ * This class replaces preg_* functions to better handle UTF8 strings.
+ *
  * @author Kuba Werłos <werlos@gmail.com>
  *
  * @internal
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -40,7 +40,7 @@ final class Preg
             return $result;
         }
 
-        $result = preg_match(self::removeUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
+        $result = @preg_match(self::removeUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
         if (false !== $result) {
             return $result;
         }
@@ -66,7 +66,7 @@ final class Preg
             return $result;
         }
 
-        $result = preg_match_all(self::removeUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
+        $result = @preg_match_all(self::removeUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
         if (false !== $result) {
             return $result;
         }
@@ -92,7 +92,7 @@ final class Preg
             return $result;
         }
 
-        $result = preg_replace(self::removeUtf8Modifier($pattern), $replacement, $subject, $limit, $count);
+        $result = @preg_replace(self::removeUtf8Modifier($pattern), $replacement, $subject, $limit, $count);
         if (null !== $result) {
             return $result;
         }
@@ -118,7 +118,7 @@ final class Preg
             return $result;
         }
 
-        $result = preg_replace_callback(self::removeUtf8Modifier($pattern), $callback, $subject, $limit, $count);
+        $result = @preg_replace_callback(self::removeUtf8Modifier($pattern), $callback, $subject, $limit, $count);
         if (null !== $result) {
             return $result;
         }
@@ -132,6 +132,8 @@ final class Preg
      * @param int    $limit
      * @param int    $flags
      *
+     * @throws PregException
+     *
      * @return string[]
      */
     public static function split($pattern, $subject, $limit = -1, $flags = 0)
@@ -141,7 +143,12 @@ final class Preg
             return $result;
         }
 
-        return preg_split(self::removeUtf8Modifier($pattern), $subject, $limit, $flags);
+        $result = @preg_split(self::removeUtf8Modifier($pattern), $subject, $limit, $flags);
+        if (false !== $result) {
+            return $result;
+        }
+
+        throw new PregException('Error occurred when calling preg_split.', preg_last_error());
     }
 
     /**
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -176,7 +176,11 @@ final class Preg
             return array_map(__METHOD__, $pattern);
         }
 
-        $delimiter = $pattern[0];
+        if ('' === $pattern) {
+            return '';
+        }
+
+        $delimiter = substr($pattern, 0, 1);
 
         $endDelimiterPosition = strrpos($pattern, $delimiter);
 
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -13,7 +13,8 @@
 namespace PhpCsFixer;
 
 /**
- * This class replaces preg_* functions to better handle UTF8 strings.
+ * This class replaces preg_* functions to better handling UTF8 strings,
+ * ensuring no matter "u" modifier is present or absent subject will be handled correctly.
  *
  * @author Kuba Werłos <werlos@gmail.com>
  *
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -26,7 +26,7 @@ final class Preg
      * @param int           $flags
      * @param int           $offset
      *
-     * @throws \RuntimeException
+     * @throws PregException
      *
      * @return int
      */
@@ -42,7 +42,7 @@ final class Preg
             return $result;
         }
 
-        throw new \RuntimeException('Error occurred when calling preg_match.', preg_last_error());
+        throw new PregException('Error occurred when calling preg_match.', preg_last_error());
     }
 
     /**
@@ -52,7 +52,7 @@ final class Preg
      * @param int           $flags
      * @param int           $offset
      *
-     * @throws \RuntimeException
+     * @throws PregException
      *
      * @return int
      */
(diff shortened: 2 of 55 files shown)


src/PregWrapper.php
```diff
@@ -0,0 +1,29 @@
...
+
+namespace PhpCsFixer;
+
+class PregWrapper
+{
+    public static function replace($pattern, $replacement, $subject)
+    {
+        // TODO: check if "u" modifier is in the pattern and throw exception if so
+
+        $newSubject = preg_replace($pattern.'u', $replacement, $subject);
```
keradus: I must admit that I was exactly thinking about this solution - try with `u`, and if it crashes try without `u`, yet didn't have time to see it working. I think this is the solution we shall aim to, if it works

:+1: :heart:

src/PregWrapper.php
```diff
@@ -0,0 +1,29 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+class PregWrapper
```
keradus: final, internal

src/Fixer/Phpdoc/PhpdocTrimFixer.php
```diff
@@ -114,15 +115,15 @@ private function fixStart($content)
      */
     private function fixEnd($content)
     {
-        return preg_replace(
+        return PregWrapper::replace(
```
keradus: I do understand that we gonna replace all usage of `preg_*` ?

Slamdunk: Yes: a new AutoReview will ensure `preg_*` calls outside PregWrapper are forbidden and must use the proxy instead.

src/PregWrapper.php
Slamdunk: I like the double-try too

src/PregWrapper.php
```diff
@@ -0,0 +1,29 @@
...
+{
+    public static function replace($pattern, $replacement, $subject)
+    {
+        // TODO: check if "u" modifier is in the pattern and throw exception if so
```
Slamdunk: @kubawerlos I would do this check during PHP-CS-Fixer SUT only

keradus: Instead, I would gracefully add `u` and later gracefully remove `u`, so it wouldn't matter if dev provided `u` in input or not

Slamdunk: What I meant is that I would dislike having some calls with the flag and some without. All developers should be aware of the `UTF-8` topic and should know that thay have to let the wrapper take care of this, so no `u` flag around the project should be allowed.
IMHO.

keradus: if you want all dev to be aware, don't differentiate behaviour during SUT

src/PregWrapper.php
```diff
@@ -0,0 +1,29 @@
...
+
+        $newSubject = preg_replace($pattern.'u', $replacement, $subject);
+
+        if (null === $newSubject && PREG_BAD_UTF8_ERROR === preg_last_error()) {
+            $newSubject = preg_replace($pattern, $replacement, $subject);
+        }
+
+        return $newSubject;
```
keradus: I would love if we would check `null === $newSubject` for error, if we are already wrapping internal functions into OOP, let us use expections over ANSI-C style errors

src/PregWrapper.php
Slamdunk: What would be the added value of having the check done while I run `php-cs-fixer fix` in my apps?
The Exception would never be thrown because PHP-CS-Fixer would never be released with the *u-flag-thing* not well fixed for all the fixers.
If we check the *u-flag-thing* in all the travis build of PHP-CS-Fixer builds, we know it's everything ok before any release, so we can skip the check during an ordinary fixing run.

keradus: benefit is maintainability, if sth is protected only by tests it is not enough for me, as tests could not testing it for some weird reason after a year or class will be considered stable and not internal anymore, so any dev using it outside Fixer directly would not have the check performed.

Slamdunk: Ok then

keradus:
@FriendsOfPHP/php-cs-fixer 
I am :+1: for putting this into 2.2, not master. It will fix dozen of encoding issues, and as a side effect ease LTS branch merging. Do you agree guys ?

kubawerlos:
So, first implementation done.

Few things worth notice:
- `preg_quote` is not included as it is not using patterns
- `preg_split` is implemented differently as I was unable to find a case where it return `false` for both UTF-8 and non-UTF-8 calls
- `@` are added and each usage is covered by test case
- 2 classes that have `PregWrapper` already used are only ones requiring changes in code - all other changes are simple replacements of `preg_` functions with wrapper - not included yet, to make code review easier

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -255,6 +256,42 @@ public function testThatDataProvidersAreCorrectlyNamed($testClassName, $dataProv
...
+        $strings = array_unique($strings);
+
+        $this->markTestSkipped('Will do after functions are wrapped');
+        //$this->assertNotContains('preg_filter', $strings);
+        //$this->assertNotContains('preg_grep', $strings);
+        //$this->assertNotContains('preg_match', $strings);
+        //$this->assertNotContains('preg_match_all', $strings);
+        //$this->assertNotContains('preg_replace', $strings);
+        //$this->assertNotContains('preg_replace_callback', $strings);
+        //$this->assertNotContains('preg_split', $strings);
```
keradus: ?

tests/PregWrapperTest.php
```diff
@@ -0,0 +1,133 @@
...
+        $this->expectExceptionMessage(sprintf('Error occurred when calling %s.', $function));
+        $this->expectExceptionCode(PREG_BACKTRACK_LIMIT_ERROR);
+
+        call_user_func_array(PregWrapper::class.'::'.$method, array_merge(['/(?:\D+|<\d+>)*[!?]/'], $arguments));
+    }
+
+    public function provideFailingCases()
+    {
+        return [
+            ['match', 'preg_match', ['foobar foobar foobar']],
```
julienfalque: Usually, a test method only covers a single method of the tested class and I must admit I don't feel very comfortable with having a tested method per test case instead. But if others are ok with that, then 👍 for me.

src/Fixer/Phpdoc/PhpdocTrimFixer.php
```diff
@@ -114,15 +115,15 @@ private function fixStart($content)
...
             '~
                 (\R[ \t]*(?:\*[ \t]*)?\S.*?) # last line with useful content
                 (?:
                     (?<!/\*\*)               # not preceded by a DocComment start
                     \R[ \t]*(?:\*[ \t]*)?    # lines without useful content
                 )+
                 (\R[ \t]*\*/$)               # DocComment end
-            ~xu',
+            ~x',
```
julienfalque: I would keep the `u` flag as it matches the most common scenario.

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+
+    /**
+     * @param string|string[] $pattern
+     *
+     * @return string|string[]
+     */
+    private static function addUtf8Modifier($pattern)
+    {
+        if (is_array($pattern)) {
+            return array_map('self::'.__FUNCTION__, $pattern);
```
julienfalque: Could be `__METHOD__` instead of `'self::'.__FUNCTION__` I think. Same in `removeUtf8Modifier()`.

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -255,6 +256,42 @@ public function testThatDataProvidersAreCorrectlyNamed($testClassName, $dataProv
...
 
+    /**
+     * @dataProvider provideClassesWherePregFunctionsAreForbiddenCases
+     *
+     * @param string $className
+     */
+    public function testThereIsNoPregFunctionUsedDirectly($className)
+    {
+        $rc = new \ReflectionClass($className);
+        $tokens = Tokens::fromCode(file_get_contents($rc->getFileName()));
```
julienfalque: Did you try running this test with Xdebug enabled? I think it might be very slow, maybe worth skipping it in such case like e.g. in [`ProjectCodeTest::provideDataProviderMethodNameCases()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.10.0/tests/AutoReview/ProjectCodeTest.php#L338).

kubawerlos: Yeah, was slow - I'll disable them when Xdebug is enabled.

What about going further and disable whole AutoReview for Xdebug?

tests/PregWrapperTest.php
kubawerlos: When I started working on the wrapper I wrote tests first and this way was more readable, in the cost of "following the art". `PregWrapper` is also in fact "ifology" and structural code (written as class), not OO code.

tests/AutoReview/ProjectCodeTest.php
julienfalque: If others are really slow, why not, maybe in another PR. But for me it was only super slow for the already disabled test, which uses `Tokens::fromCode()`. This is why I commented on this specific line.

kubawerlos:
@keradus @juliendufresne any idea how to handle PHP 5.3? I cannot reproduce the problem reported by Travis: http://sandbox.onlinephpfunctions.com/code/255e4b8c9762c6cc8e51a77245777f2481b42c46

kubawerlos:
@julienfalque @keradus PR updated, unfortunately fabbot.io cannot be satisfied in this case - what do you think about it? I'm not sure if we want to wrap `preg_` functions in `tests` directory - what do you think?

kubawerlos:
@julienfalque @keradus PR updated, unfortunately fabbot.io cannot be satisfied in this case - what do you think about it? I'm not sure if we want to wrap preg_ functions in tests directory - do we?

julienfalque:
You can ignore fabbot.io's report, the applied rule does not exist on branch 2.2 but it exists and is applied on branch 2.10.

kubawerlos:
@julienfalque I know we have to ignore fabbot.io report in this case - I mean what do you think about this PR? Can you continue review so we have this merged?

@keradus I've removed wrapper from tests as we spoke, feel free to do code review as well.

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ */
+final class PregWrapper
```
keradus: I would replace it with `Preg` name, so changes would be:
```diff
-preg_match(
+Preg::match(
```

shall be super easy to apply with any IDE or just sed

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+
+        throw new \RuntimeException('Error occurred when calling preg_match_all.', preg_last_error());
+    }
+
+    /**
+     * @param string|string[] $pattern
+     * @param string|string[] $replacement
+     * @param string|string[] $subject
+     * @param int             $limit
+     * @param int             $count
```
keradus: nullable? &?

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ */
+final class PregWrapper
+{
+    /**
+     * @param string   $pattern
+     * @param string   $subject
+     * @param string[] &$matches
```
keradus: nullable

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+
+        throw new \RuntimeException('Error occurred when calling preg_replace.', preg_last_error());
+    }
+
+    /**
+     * @param string|string[] $pattern
+     * @param callable        $callback
+     * @param string|string[] $subject
+     * @param int             $limit
+     * @param int             $count
```
keradus: null + &

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+     * @param string   $subject
+     * @param string[] &$matches
+     * @param int      $flags
+     * @param int      $offset
+     *
+     * @throws \RuntimeException
+     *
+     * @return int
+     */
+    public static function match($pattern, $subject, &$matches = null, $flags = 0, $offset = 0)
```
keradus: input pattern:
- must have `/u`
- must not have `/u`
- doesn't matter

?

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+final class PregWrapper
+{
+    /**
+     * @param string   $pattern
+     * @param string   $subject
+     * @param string[] &$matches
+     * @param int      $flags
+     * @param int      $offset
+     *
+     * @throws \RuntimeException
```
keradus: IMO we shall have dedicated exception thrown (one exception for all methods is enough)

tests/PregWrapperTest.php
```diff
@@ -0,0 +1,162 @@
...
+    /**
+     * @param string $pattern
+     * @param string $subject
+     *
+     * @dataProvider provideCommonCases
+     */
+    public function testMatch($pattern, $subject)
+    {
+        $expectedResult = preg_match($pattern, $subject, $expectedMatches);
+        $actualResult = PregWrapper::match($pattern, $subject, $actualMatches);
```
keradus: Which test is proving that `Preg::*` works when `preg_*` fails?

src/PregWrapper.php
```diff
@@ -0,0 +1,175 @@
...
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
```
keradus: There is missing reasoning why one shall use that class instead of `preg_*` methods directly

src/PregWrapper.php
kubawerlos: doesn't matter

tests/PregWrapperTest.php
kubawerlos: I've added couple.

src/PregWrapper.php
kubawerlos: Added, open for suggestion for better one.

src/Preg.php
```diff
@@ -0,0 +1,177 @@
...
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+/**
+ * This class replaces preg_* functions to better handle UTF8 strings.
```
keradus: What is that "better" ? please describe playing with `/u`, especially:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3446#discussion_r172073746
(so one would know shall he pass `/u`, shall he not, or all works)

src/PregException.php
```diff
@@ -0,0 +1,24 @@
...
+namespace PhpCsFixer;
+
+/**
+ * Exception that is thrown when PCRE function encounters an error.
+ *
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ */
+class PregException extends \RuntimeException
```
keradus: final

tests/PregTest.php
```diff
@@ -159,4 +159,22 @@ public function testSplitNonUtf8Pattern()
...
         $this->assertSame($expectedResult, $actual);
     }
+
+    public function testCorrectnessForUtf8String()
+    {
+        Preg::match('/./', 'àbc', $methodMatches);
+        preg_match('/./', 'àbc', $functionMatches);
+
+        $this->assertSame(array('à'), $methodMatches);
+        $this->assertNotSame(array('à'), $functionMatches);
```
keradus: here and in next, please extract expected argument into named var, to highly expose it's the same value for both assertions

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -361,6 +394,25 @@ public function provideDataProviderMethodNameCases()
         return $data;
     }
 
+    public function provideClassesWherePregFunctionsAreForbiddenCases()
+    {
+        if (extension_loaded('xdebug') && false === getenv('CI')) {
+            $this->markTestSkipped('Data provider too slow when Xdebug is loaded.');
```
julienfalque: Here I think the slow part is the test itself, not its data provider.

julienfalque: What I meant is that skip message is not totally accurate :p

kubawerlos: m'kay, done

keradus:
Thank you @kubawerlos.

PR #3451, opened 2018-01-15
Update findPHPUnit functions

BackEndTea:
Following the discussion in #3434

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
@@ -132,7 +132,8 @@ final class MyTest extends \PHPUnit_Framework_TestCase
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        foreach (array_reverse($this->findPhpUnitClasses($tokens)) as $indexes) {
+        $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
+        foreach ($phpUnitTestCaseIndicator->findPhpUnitClasses($tokens) as $indexes) {
             $this->fixPhpUnitClass($tokens, $indexes[0], $indexes[1]);
         }
     }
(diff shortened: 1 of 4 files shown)


SpacePossum:
Looks good :)
@keradus as we are about to have this same method in 3 PHPUnit related fixers, how you feel about introducing an `AbstractPHPUnitTestFixer` which provides it childs with this method to find PHPUnit classes in the Tokens?

keradus:
all the refactoring and unification of code shall be send to 2.2, as LTS

BackEndTea:
2.2 Does not have these fixers and therefor not the functions

keradus:
yey :+1:

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -174,23 +174,18 @@ private function detectIndent(Tokens $tokens, $index)
     /**
      * @param Tokens $tokens
      *
-     * @return int[][] array of [start, end] indexes from sooner to later classes
+     * @return \Generator
```
keradus: it has to be documented what is this generator producing

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -113,23 +113,18 @@ protected function createConfigurationDefinition()
...
      *
-     * @return int[][] array of [start, end] indexes from sooner to later classes
+     * @return \Generator
      */
     private function findPhpUnitClasses(Tokens $tokens)
     {
         $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
-        $phpunitClasses = [];
-
-        for ($index = 0, $limit = $tokens->count() - 1; $index < $limit; ++$index) {
```
keradus: in original solution, we were iterating each element of collection (having 1k+ tokens usually!) only once - `O(n)`.
now, its `O(2n)`.
In fact, for non-phpunit it will iterate only once as well, yet for phpunit, it will literally iterate almost all of the element twice, as usually test file has just one big test class.

SpacePossum: this could be fixed by keeping tracking of the close brace of the class, removing much of the overhead;
will your analyze is correct it is also incomplete, the new code provides two benefits;
- not returning an array after _all_ tokens are checked, the gain here is up to the caller
- no need for array flipping the results for some fixers which caused the returned array to be copied

keradus: most cases:
- iterate over all classes
- flip one-item array

performance loss here is significant

keradus:
ping @BackEndTea

BackEndTea:
@keradus & @SpacePossum  sorry for the delay, i optimized the code a bit and moved it to the `PhpUnitTestCaseIndicator` class. I'd like to hear your thoughts

SpacePossum:
@BackEndTea FYI I've add a little unit test method for this. @keradus feel free to squash my commit out before/during merging.

src/Indicator/PhpUnitTestCaseIndicator.php
```diff
@@ -48,4 +48,27 @@ public function isPhpUnitClass(Tokens $tokens, $index)
 
         return false;
     }
+
+    /**
+     * @param Tokens $tokens
+     *
+     * @return \Generator| [int,int] array of [start, end] indexes from sooner to later classes
```
keradus: `\Generator| [int,int]` syntax doesn't look valid

SpacePossum: if this is the last thing blocking a merge I'm happy to update this PR :)   what should it be, returned is a `generator` that provides `[int, int]`

keradus:
Thank you @BackEndTea.

BackEndTea:
Thank you @keradus @SpacePossum

PR #3454, opened 2018-01-16
Add StringLineEndingFixer

### LINKED ISSUE #2224: LineEndingInContentFixer
HI!
as original fixer stop changing the content (very good!) after merge of #2209, we need a second, risky fixer that will fix the line ending in content

### LINKED ISSUE #3452: Line endings in multi-line strings are not corrected
We've moved from `\r\n` line endings to `\n` in all of our PHP files and have used PHP-CS-Fixer for this migration. Unfortunately, PHP-CS-Fixer does not convert the line endings in multi-line strings. This makes sense as that's a change in behaviour. However in our case we really just want unified line endings.

Do you think it would make sense to add an additional rule for this?

```php
<?php

$finder = PhpCsFixer\Finder::create()->in(__DIR__);

    ->setRiskyAllowed(true)
    ->setRules([
        '@PSR1' => true,
        '@PSR2' => true,
        '@Symfony' => true,
        '@PHP70Migration' => true,
        'phpdoc_align' => false,
        'phpdoc_summary' => false,
        'ternary_to_null_coalescing' => true,
        'ordered_imports' => true,
        'phpdoc_order' => true,
        'general_phpdoc_annotation_remove' => ['author'],
        'concat_space' => ['spacing' => 'one'],
        'yoda_style' => false,
    ])
    ->setFinder($finder);
```

Hint: The `\r\n` and `\n` are just for clarity and not actually part of the source code.

```php
return 'my\r\n
multi\r\n
line\r\n
string';
```

```php
return 'my\r\n
multi\r\n
line\r\n
string';
```

```php
return 'my\n
multi\n
line\n
string';
```

iluuu1994:
Closes #3452
Closes #2224


keradus:
Closes #3452

src/Fixer/StringNotation/StringLineEndingFixer.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        return new FixerDefinition(
+            'All multi-line strings must use same line ending.',
+            [
+                new CodeSample(
+                    "<?php \$a = 'my\r\nmulti\nline\r\nstring';\r\n"
+                ),
+            ],
+            null,
+            'Changing the line endings of multi-line strings might change behaviour.'
```
keradus: Basically, you said here:
`sprintf("Doing %s might change behaviour.", $description)`.
I don't see any value here, change behaviour of what? in what way?

iluuu1994: @keradus Well, for one if you store the string anywhere, different line endings will be stored. String comparison could also potentially fail. Do you have a suggestion?

src/Fixer/StringNotation/StringLineEndingFixer.php
```diff
@@ -0,0 +1,84 @@
...
+        $ending = $this->whitespacesConfig->getLineEnding();
+
+        for ($index = 0, $count = count($tokens); $index < $count; ++$index) {
+            $token = $tokens[$index];
+
+            if ($token->isGivenKind([T_CONSTANT_ENCAPSED_STRING, T_ENCAPSED_AND_WHITESPACE])) {
+                $tokens[$index] = new Token([
+                    $token->getId(),
+                    preg_replace(
+                        "#\r\n|\n#",
```
Slamdunk: Just use `\R`, see https://secure.php.net/manual/en/regexp.reference.escape.php

src/Fixer/StringNotation/StringLineEndingFixer.php
keradus: I do know that this fixer is risky.
Yet, this line is suppose to describe **why** it's risky, and you are just copy-pasting description only

src/Fixer/StringNotation/StringLineEndingFixer.php
SpacePossum: don't forget the `\u` flag than, so; `'#\R#u'` (or `#(.*\R)#`) IIRC

src/Fixer/StringNotation/StringLineEndingFixer.php
iluuu1994: Ok, what about:

> Changing the line endings of multi-line strings might change string comparisons and outputs.

Of something along those lines.

src/Fixer/StringNotation/StringLineEndingFixer.php
iluuu1994: @Slamdunk Note that this is mostly copied from the `LineEndingFixer`. Should we make the same adjustments in that fixer as well?

src/Fixer/StringNotation/StringLineEndingFixer.php
```diff
@@ -0,0 +1,84 @@
...
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All multi-line strings must use same line ending.',
```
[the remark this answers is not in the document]
iluuu1994: Indeed, seems like a better choice of words.

src/Fixer/StringNotation/StringLineEndingFixer.php
iluuu1994: @SpacePossum Why is the `\u` flag needed here?

src/Fixer/StringNotation/StringLineEndingFixer.php
```diff
@@ -0,0 +1,84 @@
...
+ * @author Ilija Tovilo <ilija.tovilo@me.com>
+ */
+final class StringLineEndingFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_CONSTANT_ENCAPSED_STRING, T_ENCAPSED_AND_WHITESPACE]);;
```
ntzm: Double semicolon at the end of this line

iluuu1994: Ah dang it 😂

src/Fixer/StringNotation/StringLineEndingFixer.php
SpacePossum: I've added a test case to show, which is more easy than putting it here and hoping the line breaks are not changed by the GH comment system ;)

.php_cs.dist
```diff
@@ -54,6 +54,7 @@ $config = PhpCsFixer\Config::create()
         'single_line_comment_style' => true,
         'strict_comparison' => true,
         'strict_param' => true,
+        'string_line_ending' => true,
```
julienfalque: Should be after `single_line_comment_style`.

keradus: why?

SpacePossum: can you give an example for that?

julienfalque: Because I was tired and read `strict_line_ending` :D ... Nothing to do here.

SpacePossum: :D  thanks for reviewing

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2224
I think

keradus:
indeed

tests/Fixer/StringNotation/StringLineEndingFixerTest.php
```diff
@@ -0,0 +1,83 @@
...
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        $template = "<?php\n\$a=\n<<<'EOT'\n%s\n\nEOT;\n";
```
julienfalque: Please add similar test cases with heredoc (`<<<EOT`) instead of nowdoc (`<<<'EOT'`).

keradus: @iluuu1994 or @julienfalque could one of you add that one missing test?

@julienfalque , if I get you right, this PR is RTM from your perspective after that test is added ? So we would merge it

julienfalque: I added the test cases, we can merge the PR now I think :)

Please squash my commit when merging.

SpacePossum:
thanks @iluuu1994 , nice addition :D

PR #3455, opened 2018-01-17
NoEmptyCommentFixer - comment block detection for line ending different than LF

kubawerlos:
When line ending was different than LF the end of comment block was calculated incorrectly.


SpacePossum:
@keradus please squash my commit out during merging

keradus:
Thank you @kubawerlos.

PR #3462, opened 2018-01-22
Move classmap config to autoload-dev

dunglas:
We use PHP CS Fixer as a library in [API Platform's Schema Generator](https://github.com/api-platform/schema-generator) and we use [PHP Scoper](https://github.com/humbug/php-scoper) to prevent conflicts in the generated PHAR.
PHP Scoper breaks because of the following config, moving it in the `autoload-dev` section fixes the issue. /cc @theofidry

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -54,7 +54,10 @@
         "sort-packages": true
     },
     "autoload": {
-        "psr-4": { "PhpCsFixer\\": "src/" },
+        "psr-4": { "PhpCsFixer\\": "src/" }
+    },
+    "autoload-dev": {
+        "psr-4": { "PhpCsFixer\\Tests\\": "tests/" },
         "classmap": [
             "tests/Test/AbstractFixerTestCase.php",
             "tests/Test/AbstractIntegrationTestCase.php",
@@ -63,9 +66,6 @@
             "tests/Test/IntegrationCaseFactory.php",
             "tests/TestCase.php"
         ]
-    },
-    "autoload-dev": {
-        "psr-4": { "PhpCsFixer\\Tests\\": "tests/" }
     },
     "bin": ["php-cs-fixer"]
 }


theofidry:
I would also add that PHP-Scoper or not, this change makes more sense to me as I don't really see what's the point of having those test classes in the no dev classmap

dunglas:
@theofidry this change breaks the tests: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/331860925#L580

SpacePossum:
>I don't really see what's the point of having those test classes in the no dev classmap

We don't ship and/or support the reuse of tests that are internal and only made for testing this repo.; those are in the `tests` folder and in `autoload-dev`.
However we provide tests as part of the API, i.e. it is designed for reuse by others. Putting those in the autoload-dev` causes those not being available and as such are not reusable.

Why is `PHP Scoper` break on this, why is it concerned about those classes there?

dunglas:
If they are part of the API, what about moving them to `src/` (just as Symfony does) @SpacePossum?

SpacePossum:
I'm :+1: got that, however I can't recall the discussion about the movement of these classes.
There was something about the additional composer requirements that these classes have, but not the rest of the code in the `src` folder.
Pinging @keradus and @julienfalque , they had some good points on it.

theofidry:
>Why is PHP Scoper break on this, why is it concerned about those classes there?

Not sure yet and I need to look into it. As it stands it's trying to load that file but it's no longer there as ignored by PHP-Scoper which has been configured to not scan the tests files

keradus:
Back in old days we wanted to give users making custom rules power to use our "base" test classes. Great, we put them in `src`. Awesome !

Now, users kind-of could use them. Why? They have the code, but they don't have deps. They need to install deps on their own. Some of deps. Like we install `gecko` only for testing purposes, but we don't install `phpunit`, which is also needed to use our "base" test classes.
This is fucked up. So, obviously, we don't want to depend on phpunit in our non-dev requirements. Also, we want to expose "base" test classes.

So what we can do.
We could move the classes from `src` to tests, but that would make them unable to be used by other devs, as @SpacePossum mentioned, as basically disappearing public class is BC breaker.
We could move deps from dev-deps to prod-dev-deps, but we also don't want that to happen.

So, what is the outcome?
We deprecated those classes in `src`, put them in `tests`, yet to have them usable by external code, we expose those few files manually.

Ultimate solution is under development and basically, files will not be in `src` anymore under `3.0` line (as, again, this movement is a BC breaker).

Especially because of that, we have tests proving that status quo of those problematic files are not changed.

For that, this PR, while totally understandable, cannot be merged. I admit I f*-up on those files.
Sorry for inconvenience and thanks for patience in awaiting for new, clean solution.

dunglas:
Thanks for the detailed explanation @keradus! Closing this one as it cannot be merged.
What do you think about:

* Document the soft deps in "suggest" (and maybe throw an exception like "You need to install foo/bar to use this feature"
* Let base test classes in `src/`
* Alias them with `class_alias` to preserve BC?

keradus:
As they are only test-related classes, there is no need to suggest (and suggesting to install PHPUnit is ridiculous). There is not that much ppl who actually want to use those classes, and if they are, they will see error of PHP that parent class cannot be found. And since those classes are... only PHPUnit, no deal here to have dedicated handling of it ;)

They shall not be in `src` to due unresolve-able dependencies, like PHPUnit or gecko.
And having `PHPUnit` ... 4? 5? 6? soon 7? some of them? all of them? one need to control it over composer.

`class_alias` is not a solution, as original issues remains, and deprecated classes are not 1-1 as classes from tests.

theofidry:
Make sense. Well in the meantime @dunglas it means you need to add them to both box & php-scoper to make it work

PR #3466, opened 2018-01-23
no_whitespace_in_blank_line and no_blank_lines_after_phpdoc fixers bug

kubawerlos:
Priorities were wrong and caused problem as showed in test case


keradus:
super nice, RTM if CIs are green

keradus:
Thank you @kubawerlos.

SpacePossum:
nice catch, thanks @kubawerlos !

src/Fixer/Phpdoc/NoBlankLinesAfterPhpdocFixer.php
```diff
@@ -61,7 +61,8 @@ class Bar {}
     public function getPriority()
     {
         // should be ran before the SingleBlankLineBeforeNamespaceFixer.
-        return 1;
+        // should be ran after the NoWhitespaceInBlankLineFixer.
```
GrahamCampbell: should be **run**, not **ran**

Issue #3467, opened 2018-01-23
Header comment gets reapplied


BackEndTea:
For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

#### The PHP version you are using (`$ php -v`):
=> PHP 7.1.13 (cli) (built: Jan  3 2018 08:27:24) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.10.0 Bowling Bear by Fabien Potencier and Dariusz Ruminski (429e3a2)

#### The command you use to run PHP CS Fixer:
=> ./php-cs-fixer-v2.phar fix --config=.php_cs.dist -v --allow-risky=yes"

#### The configuration file you are using, if any:
```php
<?php

$year = date('Y');
$header = <<<TXT
Copyright © 2017-$year Maks Rafalko

License: https://opensource.org/licenses/BSD-3-Clause New BSD License
TXT;
$finder = \PhpCsFixer\Finder::create()
   ->in('src')
   ->in('tests')
   ->exclude('Fixtures')
   ->in('app')
;

return \PhpCsFixer\Config::create()
    ->setRules([
        '@Symfony' => true,
        '@PHP70Migration:risky' => true,

        'concat_space' => ['spacing' => 'one'],
        'header_comment' => [
            'commentType' => 'PHPDoc',
            'header' => $header,
            'location' => 'after_open',
            'separate' => 'bottom',
        ],
        'is_null' => true,
        'phpdoc_align' => false,
        'phpdoc_summary' => false,
        'yoda_style' => false,
    ])
    ->setFinder($finder)
;

```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php
/**
 * Copyright © 2017-2018 Maks Rafalko
 *
 * License: https://opensource.org/licenses/BSD-3-Clause New BSD License
 */

namespace Infection\Tests\Filesystem\Exception;

```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php
/**
 * Copyright © 2017-2018 Maks Rafalko
 *
[...]

Issue #2590, opened 2017-03-10
Conflict between declare_strict_types and header_comment


sebastianbergmann:
```
$ php -v
PHP 7.1.2 (cli) (built: Feb 15 2017 08:01:35) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Xdebug v2.5.1, Copyright (c) 2002-2017, by Derick Rethans
```
```
$ php-cs-fixer -V
PHP CS Fixer version 2.1.1 by Fabien Potencier and Dariusz Ruminski (e0e33ce)
```
I have
```
'declare_strict_types' => true,
```
and
```
'header_comment' => ['header' => $header, 'separate' => 'none'],
```
in my `.php_cs` and
```php
<?php
/*
 * This file is part of foo
 *
 * (c) Sebastian Bergmann <sebastian@phpunit.de>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

declare(strict_types=1);

namespace SebastianBergmann\Foo;

class Bar
{
    // ...
```
is transformed into
```php
<?php
/*
 * This file is part of foo
 *
 * (c) Sebastian Bergmann <sebastian@phpunit.de>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

declare(strict_types=1);
/*
 * This file is part of foo
 *
 * (c) Sebastian Bergmann <sebastian@phpunit.de>
 *
 * For the full copyright and license information, please view the LICENSE
 * file that was distributed with this source code.
 */

namespace SebastianBergmann\Foo;

class Bar
{
    // ...
```

SpacePossum:
Hi,

Sorry you got hit by this issue. There is a bit of talk about here;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2058
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2062

Unfortunately nobody found the time to fix this yet.

keradus:
@sebastianbergmann, maybe you want to help with that ?

marcw:
I was hit by this issue. If you need to remove the header from a PHP file between the opening tag and the namespace declaration, you can use this command:

```find src -name '*.php' | xargs -L 1 sed -i '' '/^..php/,/^namespace .*;/{//!d;}'```

[...]

Issue #2363, opened 2016-12-06
Feature Request: Copyright header configuration in header_comment


bshaffer:
Because copyright headers should be pinned to the year a file was created, `header_comment` quickly becomes impractical, as the copyright year can be different depending on the file. Excluding the copyright year is not an option for everyone. 

It would be great to have options like this:

```php
return PhpCsFixer\Config::create()
    ->setRules([
        'header_comment' => [
            'ignore_copyright' => true,
            'default_copyright' => 'Copyright Brent Shaffer ' . date('Y')
        ]
    ]);

// or...

return PhpCsFixer\Config::create()
    ->setRules([
        'header_comment' => [
            'copyright' => [
                'ignore' => true,
                'default' => 'Copyright Brent Shaffer ' . date('Y')
            ]
        ]
    ])
```

Thoughts?

keradus:
if you want to keep a year in file header why not keep it up to date for all files?

bshaffer:
@keradus because Copyright headers should reflect the year the file was added, not the current year.

keradus:
why is that? created-current, as one didnt throw copyrights away
like 2015-2016

bshaffer:
I feel like newer files should not have a previous copyright year.

keradus:
To not copy-paste discussion let me hijack the thread and move forward discussion into https://github.com/GoogleCloudPlatform/php-docs-samples/pull/263#discussion_r90998582 ;)

SpacePossum:
Maybe we could try a diff. approach. How about only replacing a header with the configured header if the current header is different more than a configurable Levenshtein distance?

keradus:
:-1: for Levenshtein , if there is a typo in one file I want to have it fixed
:+1: for pattern matching

keradus:
@Wirone , I'm still tricked with the default button text to be simply "Close issue"

<img width="592" alt="Screenshot 2023-09-08 at 21 03 09" src="https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/assets/2716794/af6ffc54-821a-4812-bc61-061345fcdf60">

PR #3468, opened 2018-01-23
[WIP]Fix header comment & declare strict integration

BackEndTea:
This is a failing test case which 'proves' the bug in #3467

### DIFF
--- a/tests/AutoReview/FixerFactoryTest.php
+++ b/tests/AutoReview/FixerFactoryTest.php
@@ -69,6 +69,7 @@ final class FixerFactoryTest extends TestCase
             array($fixers['declare_strict_types'], $fixers['blank_line_after_opening_tag']),
             array($fixers['declare_strict_types'], $fixers['declare_equal_normalize']),
             array($fixers['declare_strict_types'], $fixers['single_blank_line_before_namespace']),
+            array($fixers['declare_strict_types'], $fixers['header_comment']),
             array($fixers['elseif'], $fixers['braces']),
             array($fixers['function_to_constant'], $fixers['native_function_casing']),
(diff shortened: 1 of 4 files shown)


kubawerlos:
The bug more of `DeclareStrictTypesFixer` itself than the combination of fixers, [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/tests/Fixer/Strict/DeclareStrictTypesFixerTest.php#L74) case exactly.

BackEndTea:
Since that is part of the test case i'm wondering what the best solution here is

* Check if the **exact** same header is applied below declare strict if it is present, and if so, remove it and apply it again before declare strict (if configuration says after opening tag)
* Make declare strict apply below header comment if it is present (Probably a BC break)

ping @keradus

SpacePossum:
for more background on this please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2590
(there are links to other issues/PR's as well related to this)

(possibly https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2363)

it would be awesome to this old issue resolved, it is a hard one though...

BackEndTea:
@SpacePossum 

What do you think of this solution:
>Check if the exact same header is applied below declare strict if it is present, and if so, remove it and apply it again before declare strict (if configuration says after opening tag)

julienfalque:
I think the solution proposed by @BackEndTea is the way to go. The fixer should check all possible positions of the header (`after_open` and `after_declare_strict` for now) and remove it from positions that are not the expected one.

This should prevent conflicts with `declare_strict_types` and fix an edge case: if one changes the configured position of the header, the fixer will duplicate the header even if the `declare` statement is already there, because currently the fixer only checks the configured position, not the other ones.

Issue #3474, opened 2018-01-26
Fixers should be forced to have a rationale


PowerKiKi:
I would like to start a discussion about the need for every single fixers to be documented with their rationale. This should then appear in the `describe` command, and in the documentation (`README.rst`).

For now all fixers must have a summary and may have a description. The description usually explains in details _what_ it does, but it rarely says _why_ we should enable that particular fixer. This makes it difficult and/or time consuming to decide whether or not new fixers should be enabled during an upgrade.

A recent example of this is the `StaticLambdaFixer`. From the `describe` command it is extremely clear _what_ it does, but there not even a hint of _why_ it may be a good idea to consider it:

```sh
$ ./vendor/bin/php-cs-fixer describe static_lambda
Description of static_lambda rule.
Lambdas not (indirect) referencing `$this` must be declared `static`.

Fixer applying this rule is risky.
Risky when using "->bindTo" on lambdas without referencing to `$this`.

Fixing examples:
 * Example #1.
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php
   -$a = function () use ($b)
   +$a = static function () use ($b)
    {   echo $b;
    };
   
   ----------- end diff -----------
```

This is even exacerbated by @localheinz who [kind of asked about it in the PR](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3187#issuecomment-340875577), but never gives an answer.

A counter-example of that is `ExplicitStringVariableFixer` whose description contains a rationale, and makes it much easier to make a decision about it:

```sh
$ ./vendor/bin/php-cs-fixer describe explicit_string_variable
Description of explicit_string_variable rule.
Converts implicit variables into explicit ones in double-quoted strings or heredoc syntax.
The reasoning behind this rule is the following:
- When there are two valid ways of doing the same thing, using both is confusing, there should be a coding standard to follow
- PHP manual marks `"$var"` syntax as implicit and `"${var}"` syntax as explicit: explicit code should always be preferred
- Explicit syntax allows word concatenation inside strings, e.g. `"${var}IsAVar"`, implicit doesn't
- Explicit syntax is easier to detect for IDE/editors and therefore has colors/hightlight with higher contrast, which is easier to read

Fixing examples:
 * Example #1.
   ---------- begin diff ----------
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php
   -$a = "My name is $name !";
[...]

Issue #3473, opened 2018-01-26
[3.0] Remove readme-command


SpacePossum:
I propose to remove the readme generate command and refactor to a dedicated script.
The command has little value to end users and does not need to be shipped with the release.
We do need it (of course) at dev time, however for that it is not needed (to list) as command.

kubawerlos:
@SpacePossum what do you think about moving it (or its replacement form #3475) to `dev-tools`? I did this in [custom fixers](https://github.com/kubawerlos/php-cs-fixer-custom-fixers/tree/v2.1.0/dev-tools/src) and it works well for some time.

SpacePossum:
indeed better to do with the doc refactor

julienfalque:
I added a TODO to #3475.

PR #3475, opened 2018-01-26
Rework documentation

julienfalque:
This PR depends on #4892 which should be merged first.

Currently, all the documentation appears in the root `README.rst` file. IMO this has several drawbacks:
* it's not easy to find a specific information;
* we have to summarize/strip out some details.

This PR proposes to move all details about the fixers into a root `doc` directory. The files will still be generated automatically, but we will also be able to manually add extra files to cover other parts of the documentation.


PowerKiKi:
This seems somewhat related to #3473, though it might not necessarily impact each other technically speaking.

Were you planning on publishing that documentation via https://readthedocs.org ?

I would suggest to use `docs` instead of `doc`, which seems slightly more popular (default for GitHub pages and Read the Docs).

I would also consider `rm -rf docs/rules` when starting the process to ensure that there is no left-over for fixers that got removed between versions.

Otherwise it's very close to what I had in mind for #3474. I think it would be a very nice improvement.

PowerKiKi:
What is the behavior of the `decribe` command for the fixer that would require a specific PHP version that you mentioned ?

julienfalque:
> Were you planning on publishing that documentation via https://readthedocs.org ?

Not for now, might be interesting.

> I would suggest to use `docs` instead of `doc`, which seems slightly more popular (default for GitHub pages and Read the Docs).

👍 for me.

> I would also consider `rm -rf docs/rules` when starting the process to ensure that there is no left-over for fixers that got removed between versions.

Yep, I planned to add it :)

> What is the behavior of the decribe command for the fixer that would require a specific PHP version that you mentioned ?

I think it just skips code samples that are not compatible with the PHP version it is running on. Not sure we can do the same here as the documentation has to be up-to-date whatever the PHP version you are using when contributing this project.

doc/rules/alias/pow_to_exponentiation.rst
```diff
@@ -0,0 +1,22 @@
...
+Example #1.
+
+.. code-block:: diff
+
+   --- Original
+   +++ New
+   @@ -1,2 +1,2 @@
+    <?php
+   - pow($a, 1);
+   + $a** 1;
```
carusogabriel: Minor: space

julienfalque:
Thinking about the PHP compatibility issue again. I'm willing to do the following:
- make the `documentation` command:
  - remove all rules documentation pages
  - generate all pages but those with code samples not compatible with current PHP version
- make the test suite check all pages have the expected content but skips those with code samples not compatible with current PHP version
- provide a Docker image with a PHP version that is compatible with all code samples (currently 7.2).

The Docker image would give contributors an alternative choice to update the documentation when needed. In case one still can't have the documentation up-to-date, maintainers will still be able to retrieve the branch and do it but that should not happen frequently.

WDYT?

PowerKiKi:
I am not sure I understand the benefit of the docker image. Would it not be simpler to have an automation to generate documentation in a controlled environnement (I'm thinking Travis here and its [GitHub Pages Deployment](https://docs.travis-ci.com/user/deployment/pages/)) ?

Also I don't think tests should assert the actual content of every single page of the docs, but rather the mechanic to produce it, so it should be possible to test that mechanic with all PHP versions.

ntzm:
If you only have the documentation building on CI, there would be no way of verifying it works and is correct until you make a PR, which would be annoying for many contributors. I'm +1 for the docker image.

PowerKiKi:
If you want to double-check the documentation for a fixer modified by your own PR, you most likely are running the correct PHP version that allow you to develop the PR in the first place, aren't you ?

And for people who want to work specifically on the documentation generation process, it does not seems too far-fetched to require PHP 7.2 to have a complete output. To me, the intersection of people who cannot install PHP 7.2, and people who can install docker, is probably very small.

The docker thing seems to be a waste of your time to set up in the first place. On the other hand, I'm not gonna stop you if think it is necessary.

ntzm:
Any reason for choosing RST instead of MD for the new files?

julienfalque:
I've rebased the PR on the 2.15 branch and worked a bit more on it. I consider everything is ready except the missing tests to check whether the doc is up-to-date (as was done with the README file). Please have a look at [the new README](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1b59948982a8b853544dbfc3d3830e992f1c69b1/README.md) and tell me what you think about it.

About the highest PHP version required to update the documentation, I updated the related instruction in the CONTRIBUTING file:

> - Update documentation: `php php-cs-fixer documentation`. This requires the highest version of PHP supported by PHP CS Fixer. If it is not installed on your system, you can run it in a Docker container instead: `docker run -it --rm --user="$(id -u):$(id -g)" -w="/app" --volume="$(pwd):/app" php:7.3-cli php php-cs-fixer documentation`.

I used the Docker command and it works fine. We still might receive PRs where a maintainer will have to update the documentation because the contributor could not do it, but this should be rare. Would do you think?

doc/rules/alias/pow_to_exponentiation.rst
julienfalque: This diff is generated using the actual `pow_to_exponentiation` rule, which means the corresponding fixer does not add the space. This is already covered by the `binary_operator_spaces` rule anyway so nothing to do here IMO.

dmvdbrugge:
I like the "first step" of the separation so to say: the readme itself is nice and clean. I did expect the "list of available rules" to be the full list though, not just an index. As it stands now, you either have to know what a rule does by its name, or you have to click every rule to see. This reduces searchability. Add to that the fact that the classification is a bit random, I never know where to look. Keeping it just the big alphabetical list with all details would imho be better.

I haven't checked, but does this also affect the output of `php-cs-fixer help fix`? Because I never liked all the rules being printed there too, I mean, that's not what you expect when looking for help for the command itself, right? If you ask me it should just be what "Usage" now is.

As for building the docs, I'm fine with docker, but I would like to see a "simple" command for it, even though right now it already is just a one-liner. (a `build-documentation.sh` would do imho)

I do think that CI needs to build them as well, and check if all is ok.

Lastly, I agree with the `doc` -> `docs` comment which you seemed to agree with but didn't actually do.

doc/cookbook_fixers.rst
```diff
@@ -0,0 +1,561 @@
+==============================================
```
dmvdbrugge: `CONTRIBUTING.md` still links to the old version

julienfalque: Fixed, thanks.

julienfalque:
@dmvdbrugge I added rules summary to the list of rules but not the configuration details, I'm afraid the page would contain too much information, and the initial purpose of this PR was to avoid this. What do you think?

dmvdbrugge:
That's already better indeed. Can you also add deprecated and/or risky where applicable? For example:

- [`silenced_deprecation_error`](#) *(deprecated, risky)*
    Ensures deprecation notices are silenced.

I think that would give the most information while still being as minimal as possible. Then you can read all of the what, and for the why/how you can click.

dmvdbrugge:
The list of rules headings I see are generated based on the namespace, which makes sense, but also causes inconsistency with the rule summaries:

- Php Unit
    Referred to as PHPUnit in the summaries
- Php Tag
    Referred to as PHP tags in the summaries
- Phpdoc
    Referred to as PHPDoc in the summaries

Do you think there is value in just mapping these cases, to have the desired consistency?

doc/rules/language_construct/silenced_deprecation_error.rst
```diff
@@ -0,0 +1,27 @@
+===================================
+Rule ``silenced_deprecation_error``
+===================================
```
dmvdbrugge: I don't know if GitHub is just rendering this wrong, but when looking at it [(link for convenience)](https://github.com/julienfalque/PHP-CS-Fixer/blob/doc/doc/rules/language_construct/silenced_deprecation_error.rst), it's not really styled and thus just a blob of text where it's unsure what the first warning is, and where the summary starts.

I'd say do the summary and description either before or after the warnings, but not in between. (And the styling should be fixed.)

doc/rules/language_construct/silenced_deprecation_error.rst
```diff
@@ -0,0 +1,27 @@
...
+
+   You should use ``error_suppression`` instead.
+
+Ensures deprecation notices are silenced.
+
+.. warning:: Using this rule is risky.
+
+   Silencing of deprecation errors might cause changes to code behaviour.
+
+Samples
```
dmvdbrugge: Maybe just call them `Examples`? It's a bit strange to have a heading called `Samples`, and then the enumeration calling them `Example #X.`

dmvdbrugge:
The single-rule description pages could also list the rulesets the rule is part of. That's currently in the main readme as well. I wouldn't do it in the list, but it deserves a spot on the single-rule page imho.

doc/rules/language_construct/silenced_deprecation_error.rst
julienfalque: I would keep the current ordering:
- deprecation notice (if deprecated)
- summary
- description (if available)
- risk details (if risky)

The main issue is that Github has no specific styling for admonitions such as `.. warning::`, not sure what we can do about it.

dmvdbrugge: I know you won't like this answer, but manual styling is always an option:

**Warning:** deprecated, see `other_rule`

This fixer does X

*Risky: because Y*

😜

doc/rules/language_construct/silenced_deprecation_error.rst
julienfalque: Done.

doc/rules/language_construct/silenced_deprecation_error.rst
julienfalque: Indeed I don't like this solution: it would work for Github but it means we would not use dedicated markup provided by reStructuredText and thus no possibility to leverage it for rendering the documentation somewhere else such as e.g. at https://cs.symfony.com.

doc/rules/alias/backtick_to_shell_exec.rst
```diff
@@ -0,0 +1,27 @@
...
+Description
+-----------
+
+Conversion is done only when it is non risky, so when special chars like
+single-quotes, double-quotes and backticks are not used inside the command.
+
+Examples
+--------
+
+Example #1.
```
PowerKiKi: I would make this a header, so it is linkable and have better layout, as well as remove the full stop

```suggestion
Example #1
~~~~~~~~
```

doc/rules/language_construct/silenced_deprecation_error.rst
dmvdbrugge: I get what you're saying, but then perhaps the links to [cs.symfony.com](https://cs.symfony.com) need to be more prevalent, if that is the preferred way of reading the docs.

doc/rules/alias/backtick_to_shell_exec.rst
julienfalque: Done.

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -43,7 +43,9 @@ final class ProjectCodeTest extends TestCase
      */
     private static $classesWithoutTests = [
         \PhpCsFixer\Console\SelfUpdate\GithubClient::class,
+        \PhpCsFixer\Console\Command\DocumentationCommand::class,
         \PhpCsFixer\Doctrine\Annotation\Tokens::class,
+        \PhpCsFixer\Documentation\DocumentationGenerator::class,
```
julienfalque: I added these classes here because I'm not sure testing them really has value. I mean, if they fail somehow, we should see it with docs being out of date or having unexpected changes. Shall I add tests anyway?

julienfalque:
The new documentation now has some tests. I consider this PR ready for review.

julienfalque:
The new documentation now has some tests. I consider this PR ready for review.

@dmvdbrugge I think reworking the `fix` command help should actually be done in a separate PR.

.travis.yml
```diff
@@ -91,8 +91,6 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.3
-            before_script:
-                - php php-cs-fixer fix --rules @PHP71Migration,@PHP71Migration:risky,blank_line_after_opening_tag -q || travis_terminate 1
```
julienfalque: I had to remove this because it applied changes that don't actually comply with our CS rules so the subsequent fixer run failed:
```diff
-dirname(dirname(__DIR__))
+dirname(__DIR__,2)
```
Note the missing space between "," and "2".

See https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/594286444#L1047.

I'm not sure why this line was here. Also, should we fix `CombineNestedDirnameFixer`?

SpacePossum: >I'm not sure why this line was here. Also, should we fix CombineNestedDirnameFixer?

I think so, `CombineNestedDirnameFixer` list `should run before MethodArgumentSpaceFixer` so there definitely is a priority issue we should resolve.

julienfalque: This is already covered by a priority test. The issue here is that `method_argument_space` is not part of the rules applied in the first run (the one I removed).

We could restore this line by either:
- making `CombineNestedDirnameFixer` always add the space without relying on a subsequent rule;
- merge default rules of the project with the one added explicitly here (so maybe the subsequent run would become redudant).

SpacePossum: I think the first option has most benefits for all. I also think this is a bug in the current fixer so could be part of the next patch release, WDYT?

julienfalque: See #4892.

julienfalque:
Documentation command is now in a dedicated script: `dev-tools/doc` (ref #3473).

README.md
```diff
@@ -0,0 +1,54 @@
...
+code, you know that fixing them by hand is tedious, especially on large
+projects. This tool does not only detect them, but also fixes them for you.
+
+## Installation
+
+See [installation instructions](./doc/installation.rst).
+
+## Documentation
+
+* [Usage](./doc/usage.rst)
```
SpacePossum, line 28: I think it would be good to have a very simpel command example, like `php-cs-fixer fix ./src` or something and than link to the doc file.

julienfalque: [I reworked the file a bit](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/9e95ec0524429d1ef363db2e5c5da8f696bffe13/README.md), please let me know what you think.

SpacePossum: looks very good  :)

julienfalque:
Note: won't merging this break something on https://cs.symfony.com? ping @keradus (and @fabpot maybe?)

fabpot:
Maybe it's time to archive cs.symfony.com?

src/Documentation/DocumentationGenerator.php
```diff
@@ -0,0 +1,416 @@
...
+            return strcmp(\get_class($a), \get_class($b));
+        });
+
+        $documentation = <<<'RST'
+=======================
+List of Available Rules
+=======================
+RST;
+
+        $currentGroup = null;
```
kubawerlos, line 95: What is the profit of having the groups? Would that help anyhow to someone looking for a rule?

src/Documentation/DocumentationGenerator.php
```diff
@@ -0,0 +1,416 @@
...
+use PhpCsFixer\FixerDefinition\CodeSampleInterface;
+use PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSampleInterface;
+use PhpCsFixer\Preg;
+use PhpCsFixer\RuleSet;
+use PhpCsFixer\StdinFileInfo;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Utils;
+
+class DocumentationGenerator
```
kubawerlos, line 37: `@internal` and `final`?

doc/rules/language_construct/silenced_deprecation_error.rst
dmvdbrugge: > Maybe it's time to archive cs.symfony.com?
> _-- fabpot a few comments below_

If this happens, then what? Github becomes the main source of documentation? In that case, I'd like to point out this styling point again 😉

CONTRIBUTING.md
```diff
@@ -4,6 +4,20 @@ If you need any help, don't hesitate to ask the community on [Gitter](https://gi
 
 ## Quick Guide
 
+### Fixers
```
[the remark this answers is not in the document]
julienfalque: Done.

CONTRIBUTING.md
```diff
@@ -4,6 +4,20 @@ If you need any help, don't hesitate to ask the community on [Gitter](https://gi
...
+
+A *fixer* is a class that tries to fix one CS issue (a ``Fixer`` class must
```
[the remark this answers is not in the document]
julienfalque: Done.

CONTRIBUTING.md
```diff
@@ -13,7 +27,7 @@ If you need any help, don't hesitate to ask the community on [Gitter](https://gi
 * Make changes.
 * If you are adding functionality or fixing a bug - add a test! Prefer adding new test cases over modifying existing ones.
 * Make sure there is no trailing spaces in code: `./check_trailing_spaces.sh`.
-* Regenerate README: `php php-cs-fixer readme > README.rst`. Do not modify `README.rst` manually!
+* Update documentation: `php php-cs-fixer documentation`. This requires the highest version of PHP supported by PHP CS Fixer. If it is not installed on your system, you can run it in a Docker container instead: `docker run -it --rm --user="$(id -u):$(id -g)" -w="/app" --volume="$(pwd):/app" php:7.3-cli php php-cs-fixer documentation`.
```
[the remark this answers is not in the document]
julienfalque: Yep.

README.md
```diff
@@ -0,0 +1,72 @@
...
+following command to fix the files PHP files in the `src` directory:
+
+```console
+$ tools/php-cs-fixer/vendor/bin/php-cs-fixer fix src
+```
+
+See [usage](./doc/usage.rst), list of [built-in rules](./doc/rules/index.rst)
+and [configuration file](./doc/config.rst) documentation for more details.
+
+## Helpers
```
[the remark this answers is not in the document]
julienfalque: I chose "Editor integration".

README.md
```diff
@@ -0,0 +1,72 @@
...
+* [Sublime Text](https://github.com/benmatselby/sublime-phpcs)
+* [Vim](https://github.com/stephpy/vim-php-cs-fixer)
+* [VS Code](https://github.com/junstyle/vscode-php-cs-fixer)
+
+## Community
+
+The PHP CS Fixer is maintained on GitHub at https://github.com/FriendsOfPHP/PHP-CS-Fixer
+bug reports and ideas about new features are welcome there.
+
+You can talk to us at https://gitter.im/PHP-CS-Fixer/Lobby about the project,
```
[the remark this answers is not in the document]
julienfalque: Done.

check_trailing_spaces.sh
```diff
@@ -7,6 +7,7 @@ files_with_trailing_spaces=$(
         -not -path "./.git/*" \
         -not -path "./dev-tools/bin/*" \
         -not -path "./dev-tools/vendor/*" \
+        -not -path "./doc/rules/*" \
```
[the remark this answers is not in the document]
julienfalque: I don't think we can avoid having this exception. Some code samples contain trailing whitespaces on purpose. We would have to find a way to display trailing whitespaces with something else than whitespaces.

doc/config.rst
```diff
@@ -0,0 +1,72 @@
+===========
+Config file
+===========
+
+Instead of using command line options to customize the rule, you can save the
```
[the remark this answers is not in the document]
julienfalque: Done.

doc/config.rst
```diff
@@ -0,0 +1,72 @@
...
+project configuration in a ``.php_cs.dist`` file in the root directory of your project.
```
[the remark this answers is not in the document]
julienfalque: IMO the current wording is correct: the general workflow of `.dist` files is to create them first, and copy them afterwards, if needed.

However I do believe having a `.dist` file for PHP CS Fixer does not make sense and I would drop support for it. But that discussion is out of scope here.

doc/config.rst
```diff
@@ -0,0 +1,72 @@
...
+        ->setFinder($finder)
+    ;
+
+**NOTE**: ``exclude`` will work only for directories, so if you need to exclude file, try ``notPath``.
+Both ``exclude`` and ``notPath`` methods accept only relative paths to the ones defined with the ``in`` method.
+
+See `Symfony\\Finder <https://symfony.com/doc/current/components/finder.html>`_
+online documentation for other `Finder` methods.
+
+You may also use a blacklist for the rules instead of the above shown whitelist approach.
```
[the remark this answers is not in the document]
julienfalque: Actually this was changed in #5008, I kept the same wording.

doc/config.rst
```diff
@@ -0,0 +1,72 @@
...
+The following example shows how to use all ``Symfony`` rules but the ``full_opening_tag`` rule.
+
+.. code-block:: php
+
+    <?php
+
+    $finder = PhpCsFixer\Finder::create()
+        ->exclude('somedir')
```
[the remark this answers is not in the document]
julienfalque: Done.

doc/cookbook_fixers.rst
```diff
@@ -0,0 +1,563 @@
...
+    * @author Your name <your@email.com>
+    */
+   final class RemoveCommentsFixer extends AbstractFixer
+   {
+       /**
+        * {@inheritdoc}
+        */
+       public function getDefinition()
+       {
+           // Return a definition of the fixer, it will be used in the README.rst.
```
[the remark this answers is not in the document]
julienfalque: Fixed.

doc/cookbook_fixers.rst
```diff
@@ -0,0 +1,563 @@
...
+* Forked FriendsOfPHP/PHP-CS-Fixer into your own GitHub Account.
+* Cloned your forked repository locally.
+* Installed the dependencies of PHP CS Fixer using Composer_.
+* You have read `CONTRIBUTING.md`_.
+
+Step by step
+------------
+
+For this step-by-step, we are going to create a simple fixer that
+removes all comments of the code that are preceded by ';' (semicolon).
```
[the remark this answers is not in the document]
julienfalque: Fixed.

src/Console/Command/DocumentationCommand.php
```diff
@@ -0,0 +1,90 @@
...
+        ];
+
+        $filesystem = new Filesystem();
+
+        foreach ($fixers as $fixer) {
+            $class = \get_class($fixer);
+            $paths[$class] = $path = $this->generator->getFixerDocumentationFilePath($fixer);
+
+            $directory = \dirname($path);
+            $filesystem->mkdir($directory, 0755);
```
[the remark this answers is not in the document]
julienfalque: Indeed.

src/Documentation/DocumentationGenerator.php
julienfalque: Done.

src/Documentation/DocumentationGenerator.php
julienfalque: I do find reading the list with groups easier: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ff12f49c185da6eac090595519e5902b2305283d/doc/rules/index.rst

doc/config.rst
```diff
@@ -0,0 +1,72 @@
+===========
+Config file
+===========
+
+Instead of using command line options to customize the rule, you can save the
+project configuration in a ``.php_cs.dist`` file in the root directory of your project.
+The file must return an instance of `PhpCsFixer\\ConfigInterface <https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.15.7/src/ConfigInterface.php>`_
```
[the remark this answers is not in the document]
julienfalque: Done.

doc/rules/language_construct/silenced_deprecation_error.rst
julienfalque: Agreed, I'll try to find a workaround.

tests/AutoReview/ProjectCodeTest.php
julienfalque: No answers, I assume this is considered OK.

julienfalque:
Changes from #4890 integrated here.

SpacePossum:
Thank you @julienfalque.

SpacePossum:
Great to see this in!
This is going to help a lot of our users : D

keradus:
awesome work, @julienfalque ;)

PR #3476, opened 2018-01-26
Add PhpUnitTestCaseStaticMethodCallsFixer

Slamdunk:
```diff
 class MyTest extends \PHPUnit_Framework_TestCase
 {
     public function testMe()
     {
-        $this->assertSame(1, 2);
+        static::assertSame(1, 2);
     }
 }
```

- [x] Convert `PHPUnit\Framework\Assert` static methods
- [x] Convert `PHPUnit\Framework\TestCase` static methods
- [x] Adapt `PhpUnitStrictFixer` to check for `self::` and `static::` calls too: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3477
- [x] Adapt `PhpUnitDedicateAssertFixer` to check for `self::` and `static::` calls too: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3478
- [x] Adapt `PhpUnitConstructFixer` to check for `self::` and `static::` calls too: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3479
- [x] Add to `.php_cs.dist` with `$this->` syntax: only after review

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1276,6 +1276,20 @@ Choose from the list of available rules:
   - ``style`` (``'annotation'``, ``'prefix'``): whether to use the @test annotation or
     not; defaults to ``'prefix'``
 
+* **php_unit_test_case_static_method_calls**
+
+  Calls to ``PHPUnit\Framework\TestCase`` static methods must all be of the
+  same type, either ``$this->``, ``self::`` or ``static::``.
+
+  *Risky rule: risky when PHPUnit methods are overridden or not accessible, or when project has PHPUnit incompatibilities.*
+
+  Configuration options:
+
+  - ``call_type`` (``'self'``, ``'static'``, ``'this'``): the call type to use for referring
+    to PHPUnit methods; defaults to ``'static'``
+  - ``methods`` (``array``): dictionary of ``method`` => ``call_type`` values that
+    differ from the default strategy; defaults to ``[]``
+
 * **php_unit_test_class_requires_covers**
 
   Adds a default ``@coversNothing`` annotation to PHPUnit test classes that
--- a/src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
@@ -1,0 +1,428 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+use Symfony\Component\OptionsResolver\Exception\InvalidOptionsException;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpUnitTestCaseStaticMethodCallsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @internal
+     */
+    const CALL_TYPE_THIS = 'this';
+
+    /**
+     * @internal
+     */
+    const CALL_TYPE_SELF = 'self';
+
+    /**
+     * @internal
+     */
+    const CALL_TYPE_STATIC = 'static';
+
+    private $allowedValues = [
+        self::CALL_TYPE_THIS => true,
+        self::CALL_TYPE_SELF => true,
+        self::CALL_TYPE_STATIC => true,
+    ];
+
+    private $staticMethods = [
+        // Assert methods
+        'anything' => true,
+        'arrayHasKey' => true,
+        'assertArrayHasKey' => true,
+        'assertArrayNotHasKey' => true,
+        'assertArraySubset' => true,
+        'assertAttributeContains' => true,
+        'assertAttributeContainsOnly' => true,
+        'assertAttributeCount' => true,
+        'assertAttributeEmpty' => true,
+        'assertAttributeEquals' => true,
+        'assertAttributeGreaterThan' => true,
+        'assertAttributeGreaterThanOrEqual' => true,
+        'assertAttributeInstanceOf' => true,
+        'assertAttributeInternalType' => true,
+        'assertAttributeLessThan' => true,
+        'assertAttributeLessThanOrEqual' => true,
+        'assertAttributeNotContains' => true,
+        'assertAttributeNotContainsOnly' => true,
+        'assertAttributeNotCount' => true,
+        'assertAttributeNotEmpty' => true,
+        'assertAttributeNotEquals' => true,
+        'assertAttributeNotInstanceOf' => true,
+        'assertAttributeNotInternalType' => true,
+        'assertAttributeNotSame' => true,
+        'assertAttributeSame' => true,
+        'assertClassHasAttribute' => true,
+        'assertClassHasStaticAttribute' => true,
+        'assertClassNotHasAttribute' => true,
+        'assertClassNotHasStaticAttribute' => true,
+        'assertContains' => true,
+        'assertContainsOnly' => true,
+        'assertContainsOnlyInstancesOf' => true,
+        'assertCount' => true,
+        'assertDirectoryExists' => true,
+        'assertDirectoryIsReadable' => true,
+        'assertDirectoryIsWritable' => true,
+        'assertDirectoryNotExists' => true,
+        'assertDirectoryNotIsReadable' => true,
+        'assertDirectoryNotIsWritable' => true,
+        'assertEmpty' => true,
+        'assertEqualXMLStructure' => true,
+        'assertEquals' => true,
+        'assertFalse' => true,
+        'assertFileEquals' => true,
+        'assertFileExists' => true,
+        'assertFileIsReadable' => true,
+        'assertFileIsWritable' => true,
+        'assertFileNotEquals' => true,
+        'assertFileNotExists' => true,
+        'assertFileNotIsReadable' => true,
+        'assertFileNotIsWritable' => true,
+        'assertFinite' => true,
+        'assertGreaterThan' => true,
+        'assertGreaterThanOrEqual' => true,
+        'assertInfinite' => true,
+        'assertInstanceOf' => true,
+        'assertInternalType' => true,
+        'assertIsReadable' => true,
+        'assertIsWritable' => true,
+        'assertJson' => true,
+        'assertJsonFileEqualsJsonFile' => true,
+        'assertJsonFileNotEqualsJsonFile' => true,
+        'assertJsonStringEqualsJsonFile' => true,
+        'assertJsonStringEqualsJsonString' => true,
+        'assertJsonStringNotEqualsJsonFile' => true,
+        'assertJsonStringNotEqualsJsonString' => true,
+        'assertLessThan' => true,
+        'assertLessThanOrEqual' => true,
+        'assertNan' => true,
+        'assertNotContains' => true,
+        'assertNotContainsOnly' => true,
+        'assertNotCount' => true,
+        'assertNotEmpty' => true,
+        'assertNotEquals' => true,
+        'assertNotFalse' => true,
+        'assertNotInstanceOf' => true,
+        'assertNotInternalType' => true,
+        'assertNotIsReadable' => true,
+        'assertNotIsWritable' => true,
+        'assertNotNull' => true,
+        'assertNotRegExp' => true,
+        'assertNotSame' => true,
+        'assertNotSameSize' => true,
+        'assertNotTrue' => true,
+        'assertNull' => true,
+        'assertObjectHasAttribute' => true,
+        'assertObjectNotHasAttribute' => true,
+        'assertRegExp' => true,
+        'assertSame' => true,
+        'assertSameSize' => true,
+        'assertStringEndsNotWith' => true,
+        'assertStringEndsWith' => true,
+        'assertStringEqualsFile' => true,
+        'assertStringMatchesFormat' => true,
+        'assertStringMatchesFormatFile' => true,
+        'assertStringNotEqualsFile' => true,
+        'assertStringNotMatchesFormat' => true,
+        'assertStringNotMatchesFormatFile' => true,
+        'assertStringStartsNotWith' => true,
+        'assertStringStartsWith' => true,
+        'assertThat' => true,
+        'assertTrue' => true,
+        'assertXmlFileEqualsXmlFile' => true,
+        'assertXmlFileNotEqualsXmlFile' => true,
+        'assertXmlStringEqualsXmlFile' => true,
+        'assertXmlStringEqualsXmlString' => true,
+        'assertXmlStringNotEqualsXmlFile' => true,
+        'assertXmlStringNotEqualsXmlString' => true,
+        'attribute' => true,
+        'attributeEqualTo' => true,
+        'callback' => true,
+        'classHasAttribute' => true,
+        'classHasStaticAttribute' => true,
+        'contains' => true,
+        'containsOnly' => true,
+        'containsOnlyInstancesOf' => true,
+        'countOf' => true,
+        'directoryExists' => true,
+        'equalTo' => true,
+        'fail' => true,
+        'fileExists' => true,
+        'getCount' => true,
+        'getObjectAttribute' => true,
+        'getStaticAttribute' => true,
+        'greaterThan' => true,
+        'greaterThanOrEqual' => true,
+        'identicalTo' => true,
+        'isEmpty' => true,
+        'isFalse' => true,
+        'isFinite' => true,
+        'isInfinite' => true,
+        'isInstanceOf' => true,
+        'isJson' => true,
+        'isNan' => true,
+        'isNull' => true,
+        'isReadable' => true,
+        'isTrue' => true,
+        'isType' => true,
+        'isWritable' => true,
+        'lessThan' => true,
+        'lessThanOrEqual' => true,
+        'logicalAnd' => true,
+        'logicalNot' => true,
+        'logicalOr' => true,
+        'logicalXor' => true,
+        'markTestIncomplete' => true,
+        'markTestSkipped' => true,
+        'matches' => true,
+        'matchesRegularExpression' => true,
+        'objectHasAttribute' => true,
+        'readAttribute' => true,
+        'resetCount' => true,
+        'stringContains' => true,
+        'stringEndsWith' => true,
+        'stringStartsWith' => true,
+
+        // TestCase methods
+        'any' => true,
+        'at' => true,
+        'atLeast' => true,
+        'atLeastOnce' => true,
+        'atMost' => true,
+        'exactly' => true,
+        'never' => true,
+        'onConsecutiveCalls' => true,
+        'once' => true,
+        'returnArgument' => true,
+        'returnCallback' => true,
+        'returnSelf' => true,
+        'returnValue' => true,
+        'returnValueMap' => true,
+        'setUpBeforeClass' => true,
+        'tearDownAfterClass' => true,
+        'throwException' => true,
+    ];
+
+    private $conversionMap = [
+        self::CALL_TYPE_THIS => [[T_OBJECT_OPERATOR, '->'], [T_VARIABLE, '$this']],
+        self::CALL_TYPE_SELF => [[T_DOUBLE_COLON, '::'], [T_STRING, 'self']],
+        self::CALL_TYPE_STATIC => [[T_DOUBLE_COLON, '::'], [T_STATIC, 'static']],
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Calls to `PHPUnit\Framework\TestCase` static methods must all be of the same type, either `$this->`, `self::` or `static::`.',
+            [
+                new CodeSample(
+                    '<?php
+final class MyTest extends \PHPUnit_Framework_TestCase
+{
+    public function testMe()
+    {
+        $this->assertSame(1, 2);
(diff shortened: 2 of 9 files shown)


keradus:
>  Add to `.php_cs.dist`: only after review

big :-1: for that
This topic is coming back multiple times around phpunit.
Assertions are declared as static methods, indeed. It is due to how they were copy-pasted from JUnit ages ago.
It even happen it's officially documented now: http://phpunit.readthedocs.io/en/latest/assertions.html#static-vs-non-static-usage-of-assertion-methods
yet, keep in mind that official docs uses `->` convention everywhere, and most devs are following it. We too.

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,300 @@
...
+        'throwException' => true,
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'PHPUnit assertion calls must be static.',
```
keradus: add option to do the reverse, ensure all calls are `->`

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,300 @@
...
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpUnitStaticAssertionCallsFixer extends AbstractFixer
+{
+    private static $staticMethods = [
```
keradus: big :-1: for having this list.
it's easy to use, yet it will be a nightmare to maintain, with every new phpunit release, whenever new method is exposed, one will need to add it to this list...

Slamdunk: I agree, but what are the solution? PhpUnit may not be in the autoloaded dependencies.

keradus: maybe there is no solution before adding context-aware fixers

kubawerlos: > will be a nightmare to maintain

> PhpUnit may not be in the autoloaded dependencies.

What about writing test if all of assertions from PHPUnit installed during Travis job are on the list. It will result with surprising test fail, but with good message will be easy to fix. Not the cleanest solution, but will do the job.

Slamdunk: It would fail on a version and pass on another, or viceversa

kubawerlos: No, it won't, it would check if all of assertions from version are on the list and raise and error if any is missing, not when there is any extra one.

Slamdunk: If a method is in this fixer but not in phpunit, I would like to not fix it. In other words, I want to stay updated not only for new methods, but also for deprecated and removed ones.
Still, we can start with the test you propose, better than nothing.

Slamdunk:
> > Add to .php_cs.dist: only after review
> 
> big 👎 for that
> This topic is coming back multiple times around phpunit.
> Assertions are declared as static methods, indeed. It is due to how they were copy-pasted from JUnit ages ago.
> It even happen it's officially documented now: http://phpunit.readthedocs.io/en/latest/assertions.html#static-vs-non-static-usage-of-assertion-methods
> yet, keep in mind that official docs uses -> convention everywhere, and most devs are following it. We too.

Is it ok for you to add it to `.php_cs.dist` with fixer configured to `$this->`?

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
Slamdunk: Configuration added. Do I need to change the fixer name as well?

tests/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixerTest.php
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\PhpUnit\PhpUnitStaticAssertionCallsFixer
+ */
+final class PhpUnitStaticAssertionCallsFixerTest extends AbstractFixerTestCase
+{
+    public function testFixerContainsAllPhpunitStaticMethodsInItsList()
+    {
+        $assertionRefClass = new \ReflectionClass(Assert::class);
```
Slamdunk: @kubawerlos I've added the test that ensures all methods of `PHPUnit\Framework\Assert` are in the list. However we can't do this for all `PHPUnit\Framework\TestCase` methods too because there are too many false positive.

kubawerlos: @Slamdunk `PHPUnit\Framework\TestCase` is too much indeed, also from the name of fixer I would expect only method started with "assert" to be considered.

tests/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixerTest.php
```diff
@@ -0,0 +1,236 @@
...
+        $updatedStaticMethodsList = $assertionRefClass->getMethods(\ReflectionMethod::IS_STATIC | \ReflectionMethod::IS_PUBLIC);
+
+        $fixerRefClass = new \ReflectionClass(PhpUnitStaticAssertionCallsFixer::class);
+        $staticMethodsProperty = $fixerRefClass->getProperty('staticMethods');
+        $staticMethodsProperty->setAccessible(true);
+        $staticMethods = $staticMethodsProperty->getValue();
+        $staticMethodsProperty->setAccessible(false);
+
+        foreach ($updatedStaticMethodsList as $method) {
```
kubawerlos: What about using data provider instead of `foreach`?

Slamdunk: What would be the added value of having 155 tests on 1 data, instead of 1 test with 155 data?
Unlike the tests that use the `doFix` helper, here there is no environoment to set up, nor diff to output in case of fail.

kubawerlos: It would separate preparing test data from actually testing it, plus if more than 1 check fail all problems would be displayed at once.

Slamdunk: Now all the eventual missing methods are all listed at once.

Slamdunk:
@keradus may we add this to 2.11?

Slamdunk:
@keradus is anything blocking this PR to be tagged for the next release?

keradus:
I thought that #3478 shall be merged first

keradus:
`$this->` please do, we do follow this syntax in whole project

README.rst
```diff
@@ -1164,6 +1164,17 @@ Choose from the list of available rules:
 
   *Risky rule: this fixer may change functions named setUp or tearDown outside of PHPUnit tests, when a class is wrongly seen as a PHPUnit test.*
 
+* **php_unit_static_assertion_calls**
+
+  PHPUnit assertion calls must all be of the same type.
```
keradus: what type ?

Slamdunk: It is specified in the Configuration options

README.rst
```diff
@@ -1164,6 +1164,17 @@ Choose from the list of available rules:
...
+
+  *Risky rule: risky when PHPUnit methods are overridden or not accessible, or when project has PHPUnit incompatibilities.*
+
+  Configuration options:
+
+  - ``call_type`` (``'self'``, ``'static'``, ``'this'``): the call type to use for referring
+    to PHPUnit methods; defaults to ``'static'``
```
keradus: please change default to `this`. i do know that official, syntax, method header is static, yet even creator is using `->` syntax, docs says so, it's most popular usage. doing anything else by default will be confusing for users

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,351 @@
...
+        'logicalNot' => true,
+        'logicalOr' => true,
+        'logicalXor' => true,
+        'markTestIncomplete' => true,
+        'markTestSkipped' => true,
+        'matches' => true,
+        'matchesRegularExpression' => true,
+        'objectHasAttribute' => true,
+        'readAttribute' => true,
+        'resetCount' => true,
```
keradus: so, i see it's not only about assertions (vide rule name: `php_unit_static_assertion_calls`)

README.rst
Slamdunk: I'll do so, but I hope it will change soon as static analysis tools get widespread and strict calls become mandatory.

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,351 @@
...
+        'stringContains' => true,
+        'stringEndsWith' => true,
+        'stringStartsWith' => true,
+
+        // TestCase methods
+        'any' => true,
```
keradus: from which version of phpunit you collected the methods ?
how did you generate it? maybe we could reuse it later, to add method of newer phpunit when it will be released ?

maintaining this list manually will be problematic...

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,351 @@
...
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpUnitStaticAssertionCallsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    const CALL_TYPE_THIS = 'this';
```
keradus: `@internal, @private`

tests/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixerTest.php
keradus: either drop methods from `TestCase` or remove `assert` part from the name.

what are those false positives when iterating over `TestCase` methods? could you give an example ?

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
Slamdunk: Ok for `@internal`.

I don't see any usage `@private` in the `src/` folder, can you confirm you want `@private` in the PHPDoc too?

README.rst
keradus: you read description to determine do you want to use this rule and maybe configure it, user won't read options definition if he can't understand description

README.rst
keradus: you remind me myself from 3 years ago when i was trying to persuade Sebastian ;)
in the meantime, it is like it is...

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
keradus: then internal is fine - goal is the same, to not expose this as BC promise

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('call_type', 'The call type to use for referring to PHPUnit methods.'))
```
keradus: I believe we shall clone policy of `binary_operator_spaces` rule.
we have `default` strategy we want to apply, and then we have `operators` (here: `methods`) when we can override strategy.

eg, one may want to call all assertions with `->`, but `setUpBeforeClass` with `::`

README.rst
keradus: *** it, let's be strict and spread the word. long live SCA.

keradus:
> 7.2 fails because of the "exceeded the timeout of 60 seconds"

yep, it's weird. please take a look at #3627

README.rst
Slamdunk: > (keep static as default)

I'm confused: do you want `$this` or `static` as default?

tests/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixerTest.php
Slamdunk: The false positive where generated by a bug in the fixer test.
I've fixed it and now I've moved all the algorithm to consider `PHPUnit\Framework\TestCase` as the unique reference.

README.rst
Slamdunk: Description updated, hope is enough now.

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
Slamdunk: Because of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3476#discussion_r176371831, would now be `php_unit_testcase_static_method_calls` better?

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
Slamdunk: `binary_operator_spaces` is a style rule only.

This one is a SCA helper: if you want strict calls, you can't have exceptions.

I'm against your request, but of course I'll adopt it if you don't consider the above a valid point.

README.rst
keradus: you convinced me. static

tests/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixerTest.php
keradus: nice :+1:

src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php
keradus: I'm lookinig at the usage. like here, or in biggest framework - Symfony, assertions are called on `$this`, but `setUpBeforeClass` on class, not instance context, with `::`.
for that, it's not possible to use this rule without change of current "style", which is never easy to do. (well, easy to apply, hell to resolve conflicts on all PRs)

Slamdunk: > assertions are called on `$this`, but `setUpBeforeClass` on class, not instance context, with `::`.

What you describe here is a custom exception more than a custom style.

What about a list of excluded method, that will not be considered for fixing?
Something like:
```php
$config = [
    'php_unit_testcase_static_method_calls' => [
        'exclude_methods' => ['setUpBeforeClass', 'tearDownAfterClass']
    ],
];

Comment (keradus) [reply src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php]: with that option, you will not be able to enforce `->assertTrue` but `::setUpBeforeClass`.
you will only enforce `->assertTrue` and leave `setUpBeforeClass` as is, even if it was called on `$this`.
with my proposal, you enforce style for every method

File: src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php (hunk 7 above)
Comment (Slamdunk) [reply src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php]: Fixer name updated.

File: src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php (hunk 10 above)
Comment (Slamdunk) [reply src/Fixer/PhpUnit/PhpUnitStaticAssertionCallsFixer.php]: Done

**Comment by Slamdunk** [comment]
Bug found when using `$this->` for calls inside static methods, let me the day to fix it :)

**Comment by Slamdunk** [comment]
> Bug found when using `$this->` for calls inside static methods, let me the day to fix it :)

Fixed

> `$this->` please do, we do follow this syntax in whole project

Done

File: src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php [hunk 11]
```diff
@@ -0,0 +1,447 @@
...
+        'throwException' => true,
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Calls to PHPUnit\Framework\TestCase static methods must all be of the same type, either `$this->`, `self::` or `static::`.',
```
julienfalque: Please surround `PHPUnit\Framework\TestCase` with `` ` ``.

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     * @param int    $endIndex
+     */
+    private function fixPhpUnitClass(Tokens $tokens, $startIndex, $endIndex)
+    {
+        static $conversionMap = [
```
julienfalque: Please don't use a static variable.

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
Slamdunk: Done

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
Slamdunk: Moved to a class private variable.
Still static btw: class variables shouldn't have memory issues, have they?

keradus: everything static is global and persistent, as PHP can't say would you use it laster or not.
non-static can be freed after instance is destroyed

Slamdunk: Static variables removed

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
```diff
@@ -0,0 +1,447 @@
...
+
+            $functionIndex = $tokens->getPrevTokenOfKind($functionIndex, [[T_FUNCTION]]);
+        }
+
+        $prev = $tokens->getPrevMeaningfulToken($functionIndex);
+        if (!$tokens[$prev]->isGivenKind(T_STATIC)) {
+            $prev = $tokens->getPrevMeaningfulToken($prev);
+        }
+
+        return $tokens[$prev]->isGivenKind(T_STATIC);
```
julienfalque: If the function signature is `static final private function ...`, this will return `false`.

Slamdunk: Thank you, fixed

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
```diff
@@ -0,0 +1,449 @@
...
+
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('call_type', 'The call type to use for referring to PHPUnit methods.'))
+                ->setAllowedTypes(['string'])
+                ->setAllowedValues(array_keys($this->allowedValues))
+                ->setDefault('static')
+                ->getOption(),
+            (new FixerOptionBuilder('methods', 'Dictionary of `method` => `call_type` values that differ from the default strategy.'))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([static function ($option) use ($thisFixer) {
```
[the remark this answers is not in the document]
Slamdunk: I agree that sounds verbose, but I find it more clear about the intentions.
Here I explicitly tell that this function will never be bound to anything but the fixer.
Otherwise someone in the future may change the binding (for example in `setAllowedValues`) and s**t happens.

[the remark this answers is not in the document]
julienfalque: Well, tests are here to prevent s**t from happening aren't they :p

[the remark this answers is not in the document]
Slamdunk: Language constructs are stronger than tests, but I agree that a standard should be preferred over a perfect, but isolated, solution.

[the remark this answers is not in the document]
keradus: someone can always rebind sth, and having that protection here is IMO not worth it, if we don't protect ourselves everywhere else. So either we forbid to rebind on language level, having all lambdas static, or we forbid to do so by "good code rule" of "don't rebind".
Here, aliasing `$this` to pass an alias instead of using value directly is confusing, imho.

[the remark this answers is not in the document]
Slamdunk: Ok, I'll revert it tomorrow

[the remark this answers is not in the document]
Slamdunk: Can't do that: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/FixerConfiguration/FixerOption.php#L77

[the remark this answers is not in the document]
Slamdunk: That's code introduced by you @julienfalque :smile:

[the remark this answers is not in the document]
julienfalque: Damn, forgot about that one. Yep, using closures bound to a fixer for configuration caused big memory leaks. This is probably still true so you actually need to keep the closure static :/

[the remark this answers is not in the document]
Slamdunk: If can be any help I can import within the `use` only the variables I use, but i'm not an expert in PHP memory leaks / gc, so don't know if would be an improvement or a degradation.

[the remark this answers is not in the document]
julienfalque: The memory leak is caused when the closure is bound to the fixer it is declared in:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/852339dc22539ea5d4ef5be0c1f83b4db3ec43b1/src/FixerConfiguration/FixerOption.php#L153-L179

so your current code is fine.

keradus:
Thank you @Slamdunk.

PR #3477, opened 2018-01-26
PhpUnitStrictFixer: handle static calls

Slamdunk:

PR #3478, opened 2018-01-26
PhpUnitDedicateAssertFixer: handle static calls

Slamdunk:


Slamdunk:
> 👍 but I would merge it as a new feature.

I would not :\\
Let's listen what others say.

keradus:
@SpacePossum , what would be your opinion? bugfix or feature ?

keradus:
@SpacePossum , @localheinz , what would be your opinion? bugfix or feature ?

SpacePossum:
I like the change very much even though I nearly never use the static calls, I know of some people who prefer those.
My vote however is `feature` or `improvement`. I would not go for `bugfix` as the cases here that are being fixed were never intended to be fixed. So yeah, really nice additional so lets release it in a new minor :)

Slamdunk:
Rebased to master.

src/Fixer/PhpUnit/PhpUnitDedicateAssertFixer.php
```diff
@@ -178,23 +178,37 @@ public function getPriority()
...
-            end($candidate);
-            $index = $this->getAssertCandidate($tokens, key($candidate));
-            if (is_array($index)) {
-                $index = $this->fixAssert($tokens, $index);
-            }
-
-            ++$index;
+        foreach ($searchSequences as $searchSequence) {
+            $index = 1;
             $candidate = $tokens->findSequence($searchSequence, $index);
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3477#discussion_r164154150

keradus: ping @Slamdunk , could you please take a look at that ? :)
we don't want to slower this rule by 3 times

Slamdunk: Sorry for the late reply: this isn't strictly as slow as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3477#discussion_r164154150 I don't think we can improve much the performance without making this method much more complex and unreadable.

keradus: well, this makes the fixer to run over whole tokens collection not once, but 3 times, isn't it ?

Slamdunk: Algorithm optimized

Slamdunk: > i would propose to drop findSequence at all here, iterate tokens manually, search for T_STRING matching the assertion, if it does then ensure prev and prev-prev tokens are matching $this->/self::/static:: and continue with current fixing logic

I've applied this algorithm before reading your modified comment, and I wrote it like you described. Please stop reading out mind, I'm scared you'll start fixing our thoughts too.

keradus: :D
well, I'm proud of fixing few guys thoughts already, you know, fixing the initial mindset "I don't need tests" :)

keradus:
btw, @Slamdunk , could you please take a look at failing 7.2? it';s somehow related to phpdbg: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3627

keradus:
Thank you @Slamdunk.

PR #3484, opened 2018-01-30
Create Tokens::findBlockStart

ntzm:
Having `findBlockEnd` with a boolean flag to invert the behaviour is very confusing, makes the code harder to read. I propose we add a new `findBlockStart` method that replicates this behaviour in a more readable way.

Am I right in thinking I should target 2.2 with this, ~as this is an internal class~ it should be OK to add new features on patch releases?

### DIFF
--- a/src/Fixer/ArrayNotation/NoWhitespaceBeforeCommaInArrayFixer.php
+++ b/src/Fixer/ArrayNotation/NoWhitespaceBeforeCommaInArrayFixer.php
@@ -91,11 +91,11 @@ final class NoWhitespaceBeforeCommaInArrayFixer extends AbstractFixer
     private function skipNonArrayElements($index, Tokens $tokens)
     {
         if ($tokens[$index]->equals('}')) {
-            return $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $index, false);
+            return $tokens->findBlockStart(Tokens::BLOCK_TYPE_CURLY_BRACE, $index);
         }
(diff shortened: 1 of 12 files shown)


SpacePossum:
I like this one :+1: 
If others agree, please deprecate the usage of the last argument (the flag) so we can remove it on 3.0.

keradus:
> as this is an internal class (...)

it is not internal class ;)
we could add new feature in "component", as we release minor version according to new rules only (as for most users, they don't cares about using this repo as a lib, yet we cant break BC for them!).
due to that, :+1: for adding new method on 2.2, but deprecation have to be added in next minor, due to SEMVER rules.

ntzm:
How would I go about deprecating a specific parameter?

SpacePossum:
Deprecating the param has to be done in a new PR because it can only be done in a new minor.
The implementation would be something like:
- let the public methods `findStart` and `findEnd` call a new private method (directly, so find `findStart` no longer calls `findEnd`
- in the `findEnd` method, check the number of arguments passed, if to much trigger the deprecation. An example can be found here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.10.2/src/FixerDefinition/FixerDefinition.php#L44

ntzm:
Sorry accidentally deleted my comment instead of editing it 🤦‍♂️

SpacePossum:
haha no worries man :)

keradus:
Thank you @ntzm.

PR #3492, opened 2018-02-01
PhpdocScalarFixer - Add callback pesudo-type to callable type

### LINKED ISSUE #3356: Conversion of `callback` pesudo-type to `callable` type
Note to self to add this to the fixer that already does `integer` -> `int` conversions.

carusogabriel:
This resolves #3356


keradus:
Thank you @carusogabriel.

PR #3494, opened 2018-02-01
Add missing PHPDoc param type

ntzm:


src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -474,7 +474,7 @@ private function getNewOrder(array $uses, Tokens $tokens)
     }
 
     /**
-     * @param $indexes
+     * @param array $indexes
```
keradus: if we are doing this, array of what ?

keradus: typehinting in method header is missing then

src/Fixer/Import/OrderedImportsFixer.php
```diff
@@ -474,11 +474,11 @@ private function getNewOrder(array $uses, Tokens $tokens)
     }
 
     /**
-     * @param $indexes
+     * @param array[] $indexes
```
keradus: wow, this is array of arrays? really? what's inside? could you give example?

ntzm: They are built here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/760f404720db334e4a52bde5daf83123ea152f2e/src/Fixer/Import/OrderedImportsFixer.php#L410-L416

SpacePossum: the alternative to avoid this was `continue 2;` IIRC ;)

SpacePossum:
is fabbot outdated? I can't recall how the `for (;;)` should look like in SF style

src/Fixer/Import/OrderedImportsFixer.php
keradus: no, solution is to use real DTO and never use assoc arrays

keradus:
behaviour of rule was modified on higher branch to handle this case better, and on 2.2 we kept code matching old behaviour, as we are running fixer as well, so better to have failing fabbot than fixer from our perspective

keradus:
Thank you @ntzm.

PR #3464, opened 2018-01-22
IncludeFixer - fix incorrect order of fixing

kubawerlos:
Lines 65, 129, 133 and 134 are the actual fix - other changes are small cleanup while I was trying to understand the fixer.

closes #3463

PR #3496, opened 2018-02-01
Bug in Tokens::removeLeadingWhitespace

kubawerlos:
It's bug, right? @SpacePossum I believe this is the problem causing your test case from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3464 failing.


ntzm:
I think I had a similar issue here: https://github.com/ntzm/PHP-CS-Fixer/blob/151b10d4b32103811e27f979944f0122e005d5e8/tests/Fixer/Operator/StandardizeIncrementFixerTest.php#L176-L179

which I mitigated here: https://github.com/ntzm/PHP-CS-Fixer/blob/151b10d4b32103811e27f979944f0122e005d5e8/src/Fixer/Operator/StandardizeIncrementFixer.php#L155-L167

tests/Tokenizer/TokensTest.php
```diff
@@ -1045,20 +1045,46 @@ public function provideEnsureWhitespaceAtIndexCases()
...
+    }
+
+    public function provideRemoveLeadingWhitespaceCases()
+    {
+        return array(
+            array(
+                7,
+                " \t",
+                "<?php echo 1;//\n//",
+                "<?php echo 1;//\n       //",
```
SpacePossum: This is the test that fails that I, as end user of the method, at first found confusing, but in the end makes sense.
In order words; I would update this test to pass and not touch the method.
I think it is up to the caller of the method (for example fixers) to check if it is save to remove the leading space. WDYT?

kubawerlos: Well, the original failing test was to show that piece of code - `$bar;` - when is directly after one-line comment and `removeLeadingWhitespace` is called for it, will be commented out.

In the meaning of PHP token parser current behaviour is expected, but as we improve the parser to be more reasonable I believe this is the case to do it - I cannot imagine a case when having the current behaviour would be expected.

SpacePossum: The new test shows the same case and issue as yours, I just put into a provider construct.

So; lets think a bit more about what to expect; lets say for:

(I used `|` to hint a line break and stop GH UI to trim the line)

```
echo 1;  |
          |
 |
     |
   echo 2;
```

If we ask to remove leading spacing before `echo 2;` should it also remove the line breaks because it can be done?
So:

```
echo 1;echo 2;
```

Or will the new version of the method only remove spacing till the first previous line break?
So:

```
echo 1;  |
          |
 |
     |
echo 2;
```

Or remove all spacing before it but always leave it on its own line:

```
echo 1;  |
echo 2;
```

or:
```
echo 1;|
echo 2;
```

When we figured out what we want, we can update :)

kubawerlos: I'd say `echo 1;echo 2;`.

But it's different when it's with the comment (one-lines of course).

Having:
```php
// hello
    echo 1;
```

The new line after word "hello" is actually part of the comment - let's say if we have a fixer to add `trigger_error` after comment containing `@todo` it would have to move over the new line character and place new tokens "in the middle" of the `T_WHITESPACE` token (if whitespaces were line in the example above). So similarly we should treat mentioned new line when removing leading whitespaces.

keradus: As creator of the remove(leading|trailing)whitespaces methods...
@SpacePossum , in your example, output shall be `echo 1;echo 2;`.

----

Issue of Kuba is that `//` and `#` aren't contain trailing newline char.
See `WhitespacyCommentTransformer`.

Actually, I would consider this as bug, and fix it by ensuring that this case would work as well:
```php
            array(
                7,
                null,
                "<?php echo 1;//\necho 2;",
                "<?php echo 1;//\n       echo 2;",
            ),
```
(comments are special, removing \n here would change meaning of code)

SpacePossum: alright lets go with that :+1:

kubawerlos:
@SpacePossum @keradus PR updated, any suggestion for improvement?

tests/Tokenizer/TokensTest.php
SpacePossum: please add tests for:
```
echo 1;  |
          |
 |
     |
echo 2;
```
and
```
echo 1;  // 1|
          |
 |
     |
echo 2;
```

src/Tokenizer/Tokens.php
```diff
@@ -908,8 +908,22 @@ public function overrideRange($indexStart, $indexEnd, $items)
...
-            $this->clearAt($index - 1);
+        if (isset($this[$index - 1]) && $this[$index - 1]->isWhitespace()) {
+            // if the token candidate to remove is preceded by single line comment we do not consider the new line after this comment
+            // so we split the T_WHITESPACE token into 2 - newline character and the rest and call the function again
+            if (isset($this[$index - 2]) && $this[$index - 2]->isComment() && '/*' !== substr($this[$index - 2]->getContent(), 0, 2)) {
+                list($emptyString, $newLineCharacter, $remainingWhitespaces) = preg_split('/^(\R)/', $this[$index - 1]->getContent(), -1, PREG_SPLIT_DELIM_CAPTURE);
+                if ('' !== $remainingWhitespaces) {
+                    $this->overrideAt($index - 1, new Token(array(T_WHITESPACE, preg_replace('/^\R(.*)$/', '$1', $remainingWhitespaces))));
+                    $this->insertAt($index - 1, new Token(array(T_WHITESPACE, preg_replace('/^(\R).*$/', '$1', $newLineCharacter))));
+                    $this->removeLeadingWhitespace($index + 1, $whitespaces);
```
keradus: don't make 2 `T_WHITESPACE` tokens one after another, recursion is not needed

keradus:
Thank you @kubawerlos.

PR #3506, opened 2018-02-05
Require GeckoPHPUnit ^3.1.1 for PHPUnit 7 conflict

Slamdunk:
Ref: https://github.com/GeckoPackages/GeckoPHPUnit/pull/56

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "ext-tokenizer": "*",
         "composer/semver": "^1.4",
         "doctrine/annotations": "^1.2",
-        "gecko-packages/gecko-php-unit": "^2.0 || ^3.0",
+        "gecko-packages/gecko-php-unit": "^2.0 || ^3.1.1",
         "php-cs-fixer/diff": "^1.2",
         "symfony/console": "^3.2 || ^4.0",
         "symfony/event-dispatcher": "^3.0 || ^4.0",


jakzal:
Why's this package in the `require` section instead of `require-dev`?

julienfalque:
@jakzal It is used by [base test classes](https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.10.0/src/Test) provided for third party libraries, see #2757. Those classes have been deprecated so we will be able to move the package back to `require-dev` section when they will be removed, so in next major version.

Slamdunk:
Better wait https://github.com/GeckoPackages/GeckoPHPUnit/pull/57

SpacePossum:
I think best would be to remove the dependency. Current use is very minimal so shouldn't be hard. That way the project doesn't get stuck on the old package.

jakzal:
Indeed. I didn't expect this kind of package to be a dependency of a tool like php-cs-fixer. 

The dependency can still be made optional. Installing the dependency could be suggested when someone tries to use one of the classes that need it.

julienfalque:
Making the dependency optional and adding a proper check that suggests to install it would be great indeed.

SpacePossum:
replaced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3526

SpacePossum:
Lets not do the optional-tests-requirements-discussion again.
Sane way is to ditch the gecko package, keep the hidden dependency on PHPUnit for unit tests (same as Twig does it and other so...) for BC reasons and on 3.0 and up remove the API-test completely.

Issue #1692, opened 2016-01-21
[FR] ExitToDieAliasFixer (and v.v. by config)


SpacePossum:
feature request, I would like to have a fixer to changes all `exit` calls to `die` (and prop. v.v. for other people ;) )

keradus:
isn't die alias for exit ?

SpacePossum:
It is (they have the same token id), it is a language construct to it is not picked up by the `AliasFunctionsFixer`.

keradus:
then I would simply convert alias to non-alias (not v.v.) in `AliasFunctionsFixer`.

SpacePossum:
:+1:

gharlan:
:-1: because it is not a function.

SpacePossum:
Other opions? I like a diff. name (i.e. dedicated fixer) because (as @gharlan stated) it is not a function.

keradus:
I think that @gharlan refuse to incorporate it into `alias_functions`. Create a new fixer then ;)

SpacePossum:
New fixer it is :)

btw. I would like to add configuration options to normalize the default usage; 
```php
exit("");
exit('');
exit(0);
exit();
exit;
```

SpacePossum:
I'm closing this as I use a diff. system for keeping tracking on stuff I want to do one day when I've time. Please don't reopen, I like to have a small list of issues on GH to manage. Thanks!

PR #3517, opened 2018-02-08
NoAliasFunctionsFixer: add die => exit

Slamdunk:
https://secure.php.net/manual/en/function.die.php

### DIFF
--- a/src/Fixer/Alias/NoAliasFunctionsFixer.php
+++ b/src/Fixer/Alias/NoAliasFunctionsFixer.php
@@ -29,6 +29,7 @@ final class NoAliasFunctionsFixer extends AbstractFixer
     private static $aliases = [
         'chop' => 'rtrim',
         'close' => 'closedir',
+        'die' => 'exit',
         'doubleval' => 'floatval',
         'fputs' => 'fwrite',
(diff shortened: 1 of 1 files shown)


Slamdunk:
Mmm, didn't expect tests to fail 😮

Slamdunk:
Ah yea, die end exit are not functions, but language constructs, so dynamic generated test cases can never happen. Will fix tests soon, with dedicated test case for die.

SpacePossum:
good idea :+1:  but it should not be part of this fixer IMHO

maybe something of value is still in this issue;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1692#issuecomment-173545467

julienfalque:
It's not a function but it's kinda used like one, adding it to this fixer is fine IMO, but I would be ok with a separate fixer as well.

carusogabriel:
Here's a list of all PHP aliases https://secure.php.net/manual/en/aliases.php#aliases. Maybe can help you :smile:

Slamdunk:
The fact that it is a language contruct and not a funtion break `NoAliasFunctionsFixer` algorithm. I'll open a new PR with dedicated fixer.

PR #3518, opened 2018-02-09
FixerFactoryTest - Test each priority test file is listed as test

SpacePossum:


tests/AutoReview/FixerFactoryTest.php
```diff
@@ -279,16 +280,78 @@ function (array $case) use ($self) {
...
+            if ($candidate->isDot()) {
+                continue;
+            }
+
+            $fileName = $candidate->getFilename();
+            $this->assertTrue($candidate->isFile(), sprintf('Expected only files in the priority integration test directory, got "%s".', $fileName));
+            $this->assertFalse($candidate->isLink(), sprintf('No (sym)links expected the priority integration test directory, got "%s".', $fileName));
+            $this->assertSame(
+                1,
+                preg_match('#^([^,\s]+),([^,\s\d]+)(?:_\d{1,3})?\.test$#', $fileName, $matches),
```
julienfalque: Why not using `[a-z]+` for fixer names?

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -279,16 +280,78 @@ function (array $case) use ($self) {
...
+                sprintf('File with unexpected name "%s" in the priority integration test directory.', $fileName)
+            );
+
+            $fixerName1 = $matches[1];
+            $fixerName2 = $matches[2];
+
+            $caseFound = false;
+            foreach ($priorityCases as $priorityCase) {
```
julienfalque: I would process the `$priorityCases` array before the main `foreach` to avoid this nested `foreach`.

tests/AutoReview/FixerFactoryTest.php
SpacePossum: good point, should go for;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/src/FixerNameValidator.php#L31

src/FixerNameValidator.php
```diff
@@ -27,10 +27,10 @@
      */
     public function isValid($name, $isCustom)
     {
-        if (!$isCustom) {
```
keradus: what is the change of this function ?

julienfalque: Just reversing the condition, so not related to the PR but still a nice change IMO: the "core" naming format is now the default rule, instead of being an exception to it.

SpacePossum:
rebased and squashed

keradus:
Thank you @SpacePossum.

PR #3519, opened 2018-02-09
Fix typo

SpacePossum:
cc @GrahamCampbell

### DIFF
--- a/tests/IntegrationTest.php
+++ b/tests/IntegrationTest.php
@@ -73,7 +73,7 @@ final class IntegrationTest extends AbstractIntegrationTestCase
                 ], true)) {
                     static::markTestIncomplete(sprintf(
                         'Integration test `%s` was defined as explicit priority test, but no priority conflict was detected.'
-                        ."\n".'Either integration test needs to be extended or be moved from `priority` to `misc`.'
+                        ."\n".'Either integration test needs to be extended or moved from `priority` to `misc`.'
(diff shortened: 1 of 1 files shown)


julienfalque:
Do you mind handling https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3466#discussion_r165825898 here?

SpacePossum:
that one should be done on 2.2, while this typo is on 2.10 and up only so we would still still need another PR for it I'm afraid

tests/IntegrationTest.php
```diff
@@ -73,8 +73,8 @@ protected static function assertRevertedOrderFixing(IntegrationCase $case, $fixe
                 ], true)) {
                     static::markTestIncomplete(sprintf(
                         'Integration test `%s` was defined as explicit priority test, but no priority conflict was detected.'
-                        ."\n".'Either integration test needs to be extended, or test moved from `priority` to `misc`.'
-                        ."\n".'Bud don\'t do it blindly - it deserves investigation!',
+                        ."\n".'Either integration test needs to be extended or be moved from `priority` to `misc`.'
```
julienfalque: Not sure about _"or be moved"_, maybe _"or moved"_ would be better?

keradus:
oh boy... we need a spellchecker somehow :D

keradus:
Thank you @SpacePossum.

PR #3521, opened 2018-02-09
Use HTTPS

carusogabriel:
Use HTTPS when possible, and also simplifying `https://secure.php.net` to just `https://php.net` :+1:

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocInlineTagFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocInlineTagFixerTest.php
@@ -153,7 +153,7 @@ final class PhpdocInlineTagFixerTest extends AbstractFixerTestCase
         $cases[] = array(
             '<?php
     /**
-     * {@link http://www.ietf.org/rfc/rfc1035.text)
+     * {@link https://www.ietf.org/rfc/rfc1035.txt)
      */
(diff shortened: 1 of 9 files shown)


SpacePossum:
good idea, thanks :)

doc/schema.json
```diff
@@ -1,5 +1,5 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
+    "$schema": "https://json-schema.org/draft-04/schema#",
```
julienfalque: Looks like the certificate for this one is not valid, I get a "This connection is not secure" page when trying to reach it.

tests/Fixer/Phpdoc/PhpdocInlineTagFixerTest.php
```diff
@@ -153,7 +153,7 @@ public function provideTestFixInlineDocCases()
         $cases[] = array(
             '<?php
     /**
-     * {@link http://www.ietf.org/rfc/rfc1035.text)
+     * {@link https://www.ietf.org/rfc/rfc1035.text)
```
julienfalque: This one is a 404, it should be `.txt`, not `.text`.

keradus:
Thank you @carusogabriel.

PR #3523, opened 2018-02-09
[WIP] Create TernaryToElvisOperatorFixer

### LINKED ISSUE #3485: Ternary to elvis fixer
When the first and the second operands are the same:

```diff
-$a = $b ? $b : $c;
+$a = $b ?: $c;
```

carusogabriel:
This resolves #3485 by creating a new `TernaryToElvisOperatorFixer`.

(Target WIP as this is my first Fixer, and I'm trying it :sweat_smile: )

### DIFF
--- a/src/Fixer/Operator/TernaryToElvisOperatorFixer.php
+++ b/src/Fixer/Operator/TernaryToElvisOperatorFixer.php
@@ -1,0 +1,72 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Operator;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ */
+final class TernaryToElvisOperatorFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Converts a ternary operation with first and second operands identicals to an elvis operation (?:).',
+            [
+                new CodeSample(
+                    "<?php \$foo = \$bar ? \$bar : baz;\n"
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound(['?', ':']);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->equals('?')) {
+                continue;
+            }
+
+            $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
+            $prevToken = $tokens[$prevTokenIndex];
+
+            $nextTokenIndex = $tokens->getNextMeaningfulToken($index);
+            $nextToken = $tokens[$nextTokenIndex];
+
+            if (!$prevToken->equals($nextToken)) {
+                continue;
+            }
+
+            // previous and before content of ternary operation are identical, continue...
+            $tokens->clearTokenAndMergeSurroundingWhitespace($nextTokenIndex);
+        }
+    }
+}
--- a/tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
+++ b/tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\Operator;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Operator\TernaryToElvisOperatorFixer
+ */
+final class TernaryToElvisOperatorFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixCases()
+    {
+        return [
+            ['<?php $foo = $bar ?  : $bax;'],
+            ["<?php \$foo = \$bar ? 'bar' : \$bax;"],
+            'Common fix case.' => [
+                '<?php $foo = $bar ?  : $bax;',
+                '<?php $foo = $bar ? $bar : $bax;',
+            ],
+            'Array index fix case.' => [
+                "<?php \$foo = \$bar['quux'] ?  : \$bax;",
+                "<?php \$foo = \$bar['quux'] ? \$bar['quux'] : \$bax;",
+            ],
+            'Minimal number of tokens case.' => [
+                '<?php
+$foo=$bar?:$bax;',
+                '<?php
+$foo=$bar?$bar:$bax;',
+            ],
+            'With comments case.' => [
+                '<?php $foo = $bar /* foo */ ? /* bar */  : $baz;',
+                '<?php $foo = $bar /* foo */ ? /* bar */ $bar : $baz;',
+            ],
+            'With arguments in different lines.' => [
+                '<?php $foo = $bar
+? '.'
+: $baz;',
+                '<?php $foo = $bar
+? $bar
+: $baz;',
+],
+            'With arguments in dirrefent lines with comments.' => [
+                '<?php $foo = $bar // check
+?  // value if true
+: $baz; // value if false',
+                '<?php $foo = $bar // check
+? $bar // value if true
+: $baz; // value if false',
+            ],
+        ];
+    }
+}
--- a/README.rst
+++ b/README.rst
@@ -1444,6 +1444,11 @@ Choose from the list of available rules:
 
   Standardize spaces around ternary operator.
 
+* **ternary_to_elvis_operator**
+
+  Converts a ternary operation with first and second operands identicals
+  to an elvis operation (?:).
+
 * **ternary_to_null_coalescing** [@PHP70Migration, @PHP71Migration]
 
   Use ``null`` coalescing operator ``??`` where possible. Requires PHP >= 7.0.


src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,60 @@
...
+
+            $nextTokenIndex = $tokens->getNextTokenOfKind($index, [$prevToken]);
+            $nextToken = $tokens[$nextTokenIndex];
+
+            if (!$prevToken->equals($nextToken)) {
+                continue;
+            }
+
+            // previous and before content of ternary operation are equal, continue...
+            $tokens->clearRange($nextTokenIndex, $nextTokenIndex + 1);
```
carusogabriel: I'm stuck here, I don't know how to remove this whitespace:
```diff
-'<?php $foo = $bar ?: $bax;'
+'<?php $foo = $bar ? : $bax;'
```

julienfalque:
Looks like a good start 👍 

Please add more test cases with comments and/or newlines, e.g.:
```php
$foo = $bar/* a */?/* b */$bar/* c */:/* d */$baz;

$foo = $bar
    ? $bar
    : $baz
;

$foo = $bar // a
    ? $bar // b
    : $baz // c
;
```

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
```diff
@@ -0,0 +1,48 @@
...
+namespace PhpCsFixer\Tests\Fixer\Operator;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Gabriel Caruso <carusogabriel34@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Operator\TernaryToElvisOperator
```
julienfalque: Missing `Fixer` at the end of covered class name.

Slamdunk:
What about adding
```php
$foo = !empty($bar) ? $bar : $baz;
```
?

carusogabriel:
@julienfalque Thanks, added them, and know I'll try to figure out how to make them pass.

@Slamdunk In tests or in the fixer? Is this always valid, in this case?

keradus:
@carusogabriel, idea of @Slamdunk is that
`$foo = !empty($bar) ? $bar : $baz;`
means
`$foo = $bar ? $bar : $baz;`
means
`$foo = $bar ?: $baz;`
so it could be fixed as well.

yet, condition is not only `!empty($bar)` but also a loooot of different variations, like `!!$bar`, `!!!!$bar`, `!!!empty($bar)`. I'm not sure is it worth to fix them @Slamdunk , let us focus on most obvious case.

julienfalque:
@keradus maybe that could be yet another fixer that simplifies boolean conditions? E.g.:
```diff
-if (!empty($foo)) {
+if ($foo) {
     // foo
-} elseif ((bool) $bar) {
+} elseif ($bar) {
     // bar
 }

File: src/Fixer/Operator/TernaryToElvisOperatorFixer.php [hunk 3]
```diff
@@ -0,0 +1,69 @@
...
+            'Converts a ternary operation with first and second operands identicals to a elvis operatorion (?:).',
+            [
+                new CodeSample(
+                    "<?php \$foo = \$bar ? \$bar : baz;\n"
+                ),
+            ]
+        );
+    }
+
+    public function isCandidate(Tokens $tokens)
```
keradus: not sure how we go in most of the cases, but I would add `inheritdoc`

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,69 @@
...
+
+            $nextTokenIndex = $tokens->getNextMeaningfulToken($index);
+            $nextToken = $tokens[$nextTokenIndex];
+
+            if (!$prevToken->equals($nextToken)) {
+                continue;
+            }
+
+            // previous and before content of ternary operation are identical, continue...
+            $tokens->clearRange($nextTokenIndex - 1, $nextTokenIndex + 1);
```
keradus: in continuation of
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3523#discussion_r167305641

Let us describe what is what:
`$foo ? $foo : $bar`
let us assume terms for it:
`condition ? val_if_yes : val_if_no`.

So, here, you drop everything around `val_if_yes`, which could be... whitespace? comment? `?` symbol, like in case `Minimal number of tokens case.`. That's why tests are currently failing.

IMHO, this new rule shall only drop `val_if_yes`, doing
```diff
-$foo ? $foo : $bar
+$foo ? : $bar
```

Yes, it could be ugly. But it also won't mess the comments, won't drop real code, and won't mess up with whitespaces.
But hey, you know what? for all that handling of dirty whitespaces, we do have `TernaryOperatorSpacesFixer` already !
And it does know how to deal with multiline whitespaces, so you don't need to take care about it.
And in the end, we write fixers to be idempotent (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/doc/COOKBOOK-FIXERS.md#background )

So what to do?
1. update src to clear only what is needed
```diff
-$tokens->clearRange($nextTokenIndex - 1, $nextTokenIndex + 1);
+$tokens->clearTokenAndMergeSurroundingWhitespace($nextTokenIndex);
```
2. add multiline test case (don't worry, it doesn't need to look perfectly, it is only to prove that given fixer is not messing up the code, like with current `Minimal number of tokens case.`. to look pretty in this "complex" case, see step 3)
3. add priority integration test that new fixers run before `TernaryOperatorSpacesFixer` and how would look output of code fixed by our 2 fixers one after another


I hope this will help you a bit !

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @covers \PhpCsFixer\Fixer\Operator\TernaryToElvisOperatorFixer
+ */
+final class TernaryToElvisOperatorFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideFixCases
+     *
+     * @param mixed      $expected
+     * @param null|mixed $input
```
keradus: expected is always `string`
input is always `?string`

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
```
keradus: please add return typehint as array, so we will have a way to autoupgrade from docblock to php7 typehint when we will bump required php version

keradus:
@carusogabriel looks pretty nice already, I dropped few comments ;)

@julienfalque, sounds reasonable

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
carusogabriel: @keradus Thanks very much for this explanation, it healed a lot of doubts that I had. Gonna work on that and your suggestions :construction:

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
carusogabriel: Sorry, missed that one

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
carusogabriel: Oh, this was filed by php-cs-fixes itself. Maybe something related to #3516?

keradus: you mean that `@param mixed ...` were added ? yes, if there is docblock that doesn't document params, they are added with generic `mixed`. as, like here, just by `looking at code` we don't know what the types are, so fixer put generix `mixed` pseudotype.
yet, by writing the code, we do know what are the types, so we could use `string / ?string`.
putting `mixed` is some kind of reminder "maybe you shall document the types?" ;)

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,72 @@
...
+        return $tokens->isAllTokenKindsFound(['?', ':']);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->equals('?')) {
```
carusogabriel: Is there some `findSequence` we can practice here? Like: `?`, `something`, `:`

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
```diff
@@ -0,0 +1,69 @@
...
+                '<?php $foo = $bar /* foo */ ? /* bar */  : $baz;',
+                '<?php $foo = $bar /* foo */ ? /* bar */ $bar : $baz;',
+            ],
+            'With arguments in dirrefent lines with comments.' => [
+                '<?php $foo = $bar // check
+?  // value if true
+: $baz; // value if false',
+                '<?php $foo = $bar // check
+? $bar // value if true
+: $baz; // value if false',
```
carusogabriel: @julienfalque I couldn't add a:
```php
'With arguments in different lines.' => [
                '<?php $foo = $bar
? 
: $baz;',
                '<?php $foo = $bar
? $bar
: $baz;',
],
```
because of `./check_trailing_spaces.sh` :disappointed:

carusogabriel:
@keradus Something from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3523#discussion_r167445266 item 3: should I overwrite `getPriority()`? I didn't get 
> add priority integration test that new fixers run before TernaryOperatorSpacesFixer and how would look output of code fixed by our 2 fixers one after another

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
carusogabriel: @keradus Should we add this to everyone? :smile:

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
julienfalque: This is because the string contains a trailing space. If you actually need this space, you can write it like this:
```php
'With arguments in different lines.' => [
                '<?php $foo = $bar
? '.'
: $baz;',
                '<?php $foo = $bar
? $bar
: $baz;',
],
```

tests/Fixer/Operator/TernaryToElvisOperatorFixerTest.php
```diff
@@ -0,0 +1,81 @@
...
+            ['<?php $foo = $bar ?  : $bax;'],
+            ["<?php \$foo = \$bar ? 'bar' : \$bax;"],
+            'Common fix case.' => [
+                '<?php $foo = $bar ?  : $bax;',
+                '<?php $foo = $bar ? $bar : $bax;',
+            ],
+            'Array index fix case.' => [
+                "<?php \$foo = \$bar['quux'] ?  : \$bax;",
+                "<?php \$foo = \$bar['quux'] ? \$bar['quux'] : \$bax;",
+            ],
```
carusogabriel: This case will need some work :construction_worker_man:

Slamdunk:
> yet, condition might be not only `!empty($bar)`, but also a loooot of different variations, like `!!$bar`, `!!!!$bar`, `!!!empty($bar)`. I'm not sure is it worth to fix them @Slamdunk , let us focus on most obvious case.

@keradus you are right, complexity doesn't have an end, but I would make an exception for `!empty($bar)`, it is a too common construct.
I would add a `!empty($bar)` test case to be fixed, and anything that previously has another binary operator would fail, so:
```php
'fix simple not empty case' => [
    '<?php $foo = $bar ?: $baz;',
    '<?php $foo = !empty($bar) ? $bar : $baz;',
],
'dont touch complex ones empty case (1)' => [
    '<?php $foo = !!empty($bar) ? $bar : $baz;',
],
'dont touch complex ones empty case (2)' => [
    '<?php $foo = !!!empty($bar) ? $bar : $baz;',
],
'dont touch complex ones empty case (3)' => [
    '<?php $foo = !$bar ? $bar : $baz;',
],
'dont touch complex ones empty case (4)' => [
    '<?php $foo = !!$bar ? $bar : $baz;',
],
```

ntzm:
`!empty($x)` isn't the same as `$x` though, because it does an `isset` check underneath:

```php
$a = ['foo' => 'bar'];
!empty($a['baz']); // false
$a['baz']; //  Undefined index: baz
```

keradus:
@carusogabriel 

in answer tohttps://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3523#issuecomment-364791455

How to ensure proper priority between 2 fixers: `FooFixer` (`foo` rule) and `BarFixer` (`bar` rule)?
Let say we want to ensure that `foo` runs before `bar`.
Configuring proper priority could be problematic, for that we will follow Test Driven Development here, first preparing the tests, then adjusting code to ensure priority relation is proper and tests are passing.
1. Create new explicit priority unit test case in `PhpCsFixer\Tests\AutoReview\FixerFactoryTest::provideFixersPriorityCases`
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/tests/AutoReview/FixerFactoryTest.php#L60
by adding `[$fixers['foo'], $fixers['bar']]` item into returned array
2. Creat new implicit priority integration test case by creating file:
`tests/Fixtures/Integration/priority/foo,bar.test`
```
--TEST--
Integration of fixers: foo,bar.
--RULESET--
{"foo": true, "bar": true}
--EXPECT--
<?php
// fill expected code

--INPUT--
<?php
// fill input code
// ensure that this code sample is violating both, `foo` and `bar` rule in a way that would be problematic if rules would be applied in wrong order
```
3. Override `getPriority` of `FooFixer` to put comment it has to run before `BarFixer` (or `BarFixer` to put a comment that it has to run after `FooFixer`).
4. Run the tests. If they are failing, adjust value returned by `getPriority` to have tests passing.

carusogabriel:
@keradus Thank you very much. I'll update this PR ASAP :)

carusogabriel:
Closing as I won't be able to finish it :disappointed:

keradus:
why @carusogabriel ? was there anything missing except priority changes?

TomasVotruba:
@carusogabriel I can help if you need.

PR #3526, opened 2018-02-12
Remove gecko dependency

### LINKED ISSUE #3551: gecko-packages/gecko-php-unit conflicts with PHPUnit 7.0
PHP-CS-Fixer depends on gecko-packages/gecko-php-unit, but [gecko-packages/gecko-php-unit is abandoned](https://github.com/GeckoPackages/GeckoPHPUnit), and it prevents me from installing PHPUnit 7.0.

```text
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit 7.0.1 conflicts with gecko-packages/gecko-php-unit[v3.1.1].
    - phpunit/phpunit 7.0.1 conflicts with gecko-packages/gecko-php-unit[v3.1.1].
    - phpunit/phpunit 7.0.1 conflicts with gecko-packages/gecko-php-unit[v3.1.1].
    - Installation request for phpunit/phpunit ^7.0.1 -> satisfiable by phpunit/phpunit[7.0.1].
    - Installation request for gecko-packages/gecko-php-unit (locked at v3.1.1) -> satisfiable by gecko-packages/gecko-php-unit[v3.1.1].
```

SpacePossum:
this is the only way forward with this
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3506
closes #3551

### DIFF
--- a/src/Test/Constraints/XmlMatchesXsdConstraint.php
+++ b/src/Test/Constraints/XmlMatchesXsdConstraint.php
@@ -39,7 +39,7 @@ final class XmlMatchesXsdConstraint extends \PHPUnit_Framework_Constraint
         // replace first only
         $needle = 'http://www.w3.org/2001/xml.xsd';
         if (false !== $pos = strpos($xsd, $needle)) {
-            $xsd = substr_replace($xsd, 'file://'.str_replace('\\', '/', __DIR__).'/xml.xsd', $pos, strlen($needle));
+            $xsd = substr_replace($xsd, 'file:///'.str_replace('\\', '/', __DIR__).'/xml.xsd', $pos, strlen($needle));
         }
 
         $this->xsd = $xsd;
--- a/src/Test/Constraints/XmlMatchesXsdConstraint.php
+++ b/src/Test/Constraints/XmlMatchesXsdConstraint.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer\Test\Constraints;
 
 /**
+ * @author SpacePossum
+ *
  * @internal
  */
 final class XmlMatchesXsdConstraint extends \PHPUnit_Framework_Constraint
--- a/src/Test/Constraints/XmlMatchesXsdConstraint.php
+++ b/src/Test/Constraints/XmlMatchesXsdConstraint.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Test\Constraints;
+
+/**
+ * @internal
+ */
+final class XmlMatchesXsdConstraint extends \PHPUnit_Framework_Constraint
+{
+    /**
+     * @var string[]
+     */
+    private $xmlConstraintErrors = array();
+
+    /**
+     * @var string
+     */
+    private $xsd;
+
+    /**
+     * @param string $xsd
+     */
+    public function __construct($xsd)
+    {
+        parent::__construct();
+
+        $this->xsd = str_replace('http://www.w3.org/2001/xml.xsd', 'file:////'.__DIR__.'/xml.xsd', $xsd);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function toString()
+    {
+        return 'matches XSD';
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function failureDescription($other)
+    {
+        if (is_string($other)) {
+            return sprintf("%s %s.\n%s", $other, $this->toString(), implode("\n", $this->xmlConstraintErrors));
+        }
+
+        if (is_object($other)) {
+            $type = sprintf('%s#%s', get_class($other), method_exists($other, '__toString') ? $other->__toString() : '');
+        } elseif (null === $other) {
+            $type = 'null';
+        } else {
+            $type = gettype($other).'#'.$other;
+        }
+
+        return $type.' '.$this->toString();
+    }
+
+    /**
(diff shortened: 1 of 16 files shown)


SpacePossum:
`composer-require-checker` cannot be told to read `autoload-dev` as well it seems :|

keradus:
> `composer-require-checker` cannot be told to read `autoload-dev` as well it seems :|

it shouldn't load dev, that's the idea of checker. would your project work with only prod-deps

tests/Constraints/XMLMatchesXSDConstraint.php
```diff
@@ -0,0 +1,143 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Constraints;
+
+/**
+ * @internal
+ */
+final class XMLMatchesXSDConstraint extends \PHPUnit_Framework_Constraint
```
keradus: is it copy-paste from gecko ?

SpacePossum: yeah it is

SpacePossum:
>it shouldn't load dev, that's the idea of checker. would your project work with only prod-deps

If it doesn't load dev, than it shouldn't check for unknown symbols in the dev folders either I think?

julienfalque:
The report is legit: the classes you ported from Gecko package are defined in the `PhpCsFixer\Tests` namespace, which is similar to moving the package back to `require-dev`: it won't be available when using Composer with `--no-dev`.

SpacePossum:
OK, thanks. This is such a energy-draining issue I take any idea to get it resolved.

tests/Constraints/XMLMatchesXSDConstraint.php
keradus: then I'm not confident, put you as author at least ;P
maybe, let it be raised for official phpunit package ?

SpacePossum:
```
/home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Test/Constraints/XMLMatchesXSDConstraint.php:33	The parameter $XSD is not named in camelCase.
/home/travis/build/FriendsOfPHP/PHP-CS-Fixer/src/Test/Constraints/XMLMatchesXSDConstraint.php:33	The variable $XSD is not named in camelCase.
```

¯\_(ツ)_/¯

keradus:
name it `$xsd`

SpacePossum:
and class name? `XmlMatchesXsdConstraint` ? can't we white list abbreviations?

keradus:
abbreviations shall be camelcased properly. like `is HTML` is camel cased into `isHtml`. same for XML and XSD

keradus:
@SpacePossum , if possible, avoid squashing ;)
or was that just a rebase ?

src/Test/Constraints/XmlMatchesXsdConstraint.php
```diff
@@ -0,0 +1,134 @@
...
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Test\Constraints;
+
+/**
+ * @internal
```
keradus: @SpacePossum , please put you as author here

keradus:
@SpacePossum , I'm getting lost, windows is failing, could you help ?

SpacePossum:
thanks for the debugging @keradus
I'm trying here as well
This is a good example why the gecko has to go, Windows and Mac CI was never setup :(

keradus:
funny thing is that when we use this class in our tests, it works nicely,
it just failing on it's own tests :|

src/Test/Constraints/XmlMatchesXsdConstraint.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param string $xsd
+     */
+    public function __construct($xsd)
+    {
+        parent::__construct();
+
+        // replace first only
+        $needle = 'http://www.w3.org/2001/xml.xsd';
+        if (false !== $pos = strpos($xsd, $needle)) {
+            $xsd = substr_replace($xsd, 'file://'.str_replace('\\', '/', __DIR__).'/xml.xsd', $pos, strlen($needle));
```
julienfalque: Why do you need to replace this URI?

SpacePossum: Technically you don't.... as long as you are online and the resource is available because if you're not than the xsd cannot be pulled and thing breaks.

julienfalque:
Test fixed :)

@keradus please squash my commit when merging.

src/Test/Constraints/XmlMatchesXsdConstraint.php
keradus: maybe it's worth to document it btw

src/Test/Constraints/XmlMatchesXsdConstraint.php
```diff
@@ -0,0 +1,142 @@
...
+     * @param string $xsd
+     */
+    public function __construct($xsd)
+    {
+        parent::__construct();
+
+        // replace first only
+        $needle = 'http://www.w3.org/2001/xml.xsd';
+        if (false !== $pos = strpos($xsd, $needle)) {
+            $xsd = substr_replace($xsd, 'file:///'.str_replace('\\', '/', __DIR__).'/xml.xsd', $pos, strlen($needle));
```
keradus: ok, I must say that I'm super confused about triple `///` and that magical `str_replace` :(

julienfalque: The third slash is part of the path (`/C:\foo.txt`), it is required so that the URI can be parsed while omitting the hostname.

Not sure replacing Windows directory separators `\` with regular slashes `/` is required, but it works :)

keradus:
Thank you @SpacePossum.

TomasVotruba:
@SpacePossum Thank you so much for this :+1:

PR #3527, opened 2018-02-12
Deprecate last param of Tokens::findBlockEnd

### LINKED ISSUE #3525: Deprecate last param of Tokens::findBlockEnd
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3484#issuecomment-363559723

ntzm:
Closes #3525

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -440,6 +440,10 @@ class Tokens extends \SplFixedArray
      */
     public function findBlockEnd($type, $searchIndex, $findEnd = true)
     {
+        if (3 === func_num_args()) {
+            @trigger_error('Argument #3 of Tokens::findBlockEnd is deprecated and will be removed in 3.0, use Tokens::findBlockStart or drop the argument.', E_USER_DEPRECATED);
+        }
+
         return $this->findOppositeBlockEdge($type, $searchIndex, $findEnd);
     }
 
@@ -1227,10 +1231,10 @@ class Tokens extends \SplFixedArray
 
     /**
      * @param int  $type        type of block, one of BLOCK_TYPE_*
-     * @param int  $searchIndex index of starting brace
-     * @param bool $findEnd     if method should find block's end or start
+     * @param int  $searchIndex index of opening or closing brace
+     * @param bool $findEnd     find end or start?
      *
-     * @return int index of opposite brace
+     * @return int index of the opposite brace
      */
     private function findOppositeBlockEdge($type, $searchIndex, $findEnd)
     {
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -1235,10 +1235,10 @@ class Tokens extends \SplFixedArray
 
     /**
      * @param int  $type        type of block, one of BLOCK_TYPE_*
-     * @param int  $searchIndex index of opening or closing brace
-     * @param bool $findEnd     find end or start?
+     * @param int  $searchIndex index of starting brace
+     * @param bool $findEnd     if method should find block's end or start
      *
-     * @return int index of the opposite brace
+     * @return int index of opposite brace
      */
     private function findOppositeBlockEdge($type, $searchIndex, $findEnd)
     {
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -441,7 +441,11 @@ class Tokens extends \SplFixedArray
     public function findBlockEnd($type, $searchIndex, $findEnd = true)
     {
         if (3 === func_num_args()) {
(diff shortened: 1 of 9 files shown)


ntzm:
Travis failing is unrelated to changes I made, do they need fixing on this PR?

tests/Tokenizer/TokensTest.php
```diff
@@ -772,6 +772,17 @@ public function testFindBlockEndInvalidStart()
...
 
+    /**
+     * @expectedDeprecation Argument #3 of Tokens::findBlockEnd is deprecated and will be removed in 3.0, use Tokens::findBlockStart or drop the argument.
+     * @group legacy
+     */
+    public function testFindBlockEndLastParameterDeprecated()
+    {
+        $tokens = Tokens::fromCode('<?php ${$bar};');
+
+        $tokens->findBlockEnd(Tokens::BLOCK_TYPE_DYNAMIC_VAR_BRACE, 4, false);
```
julienfalque: You should still check the result of this call as it must continue to work even if deprecated. You should also check that the deprecation is triggered if the third argument is passed `true`.

tests/Tokenizer/TokensTest.php
```diff
@@ -865,8 +876,8 @@ public function assertFindBlockEnd($expectedIndex, $source, $type, $searchIndex)
         Tokens::clearCache();
         $tokens = Tokens::fromCode($source);
 
-        $this->assertSame($expectedIndex, $tokens->findBlockEnd($type, $searchIndex, true));
-        $this->assertSame($searchIndex, $tokens->findBlockEnd($type, $expectedIndex, false));
+        $this->assertSame($expectedIndex, $tokens->findBlockEnd($type, $searchIndex));
+        $this->assertSame($searchIndex, $tokens->findBlockEnd($type, $expectedIndex));
```
julienfalque: Shouldn't this one become `findBlockStart(...)`?

src/Tokenizer/Tokens.php
```diff
@@ -1284,6 +1233,73 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
...
     }
 
+    /**
+     * @param int  $type        type of block, one of BLOCK_TYPE_*
+     * @param int  $searchIndex index of opening or closing brace
+     * @param bool $findEnd     find end or start?
+     *
+     * @return int index of the opposite brace
+     */
+    private function findOppositeBlockEdge($type, $searchIndex, $findEnd)
```
keradus: for sake of maintability of LTS, this shall be targetting 2.2 branch.
so `findBlockEnd` would be just a proxy method on 2.2, without raising a deprecation

src/Tokenizer/Tokens.php
```diff
@@ -440,62 +440,11 @@ public function ensureWhitespaceAtIndex($index, $indexOffset, $whitespace)
...
-        $startIndex = $searchIndex;
-        $endIndex = $this->count() - 1;
-        $indexOffset = 1;
-
-        if (!$findEnd) {
-            list($startEdge, $endEdge) = [$endEdge, $startEdge];
-            $indexOffset = -1;
-            $endIndex = 0;
+        if (3 === func_num_args()) {
+            @trigger_error('Argument #3 of Tokens::findBlockEnd is deprecated and will be removed in 3.0, use Tokens::findBlockStart or drop the argument.', E_USER_DEPRECATED);
```
keradus: I would be more specific.
if `$findEnd` is true then drop the argument, if false then use different methods.
we could trigger different error depends on the condition, so end user doesn't have to double-check what he shall do in his case.

src/Tokenizer/Tokens.php
keradus: btw, if some of the deprecation fixers can be ported to 2.2 line, please do so as well

keradus:
> Travis failing is unrelated to changes I made, do they need fixing on this PR?

hm... Travis is green for target branch:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/branches

ntzm:
The errors are to do with the docblocks that I didn't touch

keradus:
ok, on branch check this check is not performed, as file that is causing the error is not modified on last merge into that branch.
probably the issue was introduced directly into 2.2 branch, yet we don't run phpmd on 2.2 line, that's why issue was introduced to 2.2 and detected only on later branch.

For that, could you prepare a fix in this PR, and if Travis will be happy, backport it to 2.2 line ?
Thanks !
If you want to go even further, maybe we shall backport phpmd into 2.2 branch as well, but up to you !

src/Tokenizer/Tokens.php
ntzm: I'm going to open a separate PR to port this to 2.2, but keep this one open for the deprecation and changes to > 2.2 code

keradus:
could any of method replacements to use non-deprecated version could be applied for 2.2 branch ?

keradus:
Thank you @ntzm.

PR #3531, opened 2018-02-12
Backport PHPMD to LTS version to ease maintainability

keradus:
Let us try...
backport of #2304 

fabbot.io failure is not related

if all would be fine, merge into higher branch shall bring **no** change at all

### DIFF
--- a/src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
+++ b/src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
@@ -73,6 +73,10 @@ final class NoExtraConsecutiveBlankLinesFixer extends AbstractFixer implements C
 
     /**
      * {@inheritdoc}
+     *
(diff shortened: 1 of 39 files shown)


src/Fixer/Whitespace/NoExtraConsecutiveBlankLinesFixer.php
```diff
@@ -73,6 +73,10 @@
 
     /**
      * {@inheritdoc}
+     *
+     * @SuppressWarnings(PHPMD.SwitchStatement)
+     *
+     * @todo drop warning suppression
```
julienfalque: `@todo` later? Or did you forgot to handle it in this PR?

keradus: @todo for non 2.2 branch,
actually, it's already resolved in #2304, just didn't wanted to backport it as it's not just style that changed but also a logic (in yet another PR)

PR #3532, opened 2018-02-13
Implement Tokens::findOppositeBlockEdge

ntzm:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3527#discussion_r167798841

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -392,55 +392,7 @@ class Tokens extends \SplFixedArray
      */
     public function findBlockEnd($type, $searchIndex, $findEnd = true)
     {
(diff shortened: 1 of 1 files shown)


src/Tokenizer/Tokens.php
```diff
@@ -1199,6 +1151,66 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
         $this->clearAt($nextIndex);
     }
 
+    /**
+     * @param int  $type        type of block, one of BLOCK_TYPE_*
+     * @param int  $searchIndex index of starting brace
+     * @param bool $findEnd     if method should find block's end, default true, otherwise method find block's start
```
julienfalque: The parameter has no default value.

keradus:
Thank you @ntzm.

PR #3529, opened 2018-02-12
MultilineWhitespaceBeforeSemicolonsFixer: static call bug

Slamdunk:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3327

Ping @egircys 

```diff
--- Expected
+++ Actual
@@ @@
 '<?php
 Foo::bar()
-    ->baz()
-;
+    ->baz();
```

PR #3534, opened 2018-02-13
MultilineWhitespaceBeforeSemicolonsFixer - support static calls

ntzm:
Ref #3529 

Not really a bug as it was never supported, so targetting master

Just duplicated all the `$foo->` tests and changed them to `Foo::`, let me know if you want any other cases added

### DIFF
--- a/tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
+++ b/tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
@@ -124,6 +124,81 @@ $this
 
     ;',
             ],
+            [
+                '<?php
+                    Foo::bar() // test
+;',
+                '<?php
+                    Foo::bar() // test
+                    ;',
+            ],
+            [
+                '<?php
+                    Foo::bar() # test
+;',
+                '<?php
+                    Foo::bar() # test
+
+
+                ;',
+            ],
+            [
+                '<?php
+self
+    ::setName(\'readme1\')
+    ->setDescription(\'Generates the README\');
+',
+                '<?php
+self
+    ::setName(\'readme1\')
+    ->setDescription(\'Generates the README\')
+;
+',
+            ],
+            [
+                '<?php
+self
+    ::setName(\'readme2\')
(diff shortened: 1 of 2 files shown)


keradus:
Note:
target shall be the same as it will be decided in
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3478

egircys:
Thank you @ntzm !

tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
```diff
@@ -424,6 +499,265 @@ public function provideSemicolonForChainedCallsFixCases()
...
+                ?>',
+                '<?php
+
+                    Service
+                        ::method2();
+                ?>',
+            ], [
+                '<?php
+
+                    Service::method1()
```
egircys: Should it support also multiple static calls?
```php
Service::method1()
    ::method2()
    ->method3()
;
```

ntzm: What would `method1` return to allow that?

ntzm: Ah like something that returns a class string, I think it should be allowed but I want to know what others think

egircys: Yes, something like this.. not sure if there are any real-life cases for it.
```php
<?php

class A 
{
    public static function fa()
    {
        return 'B';
    }
}

class B
{
    public static function fb()
    {
        return new C();
    }
}

class C
{
    public function fc()
    {
        return 'c';
    }
}

echo A::fa()
    ::fb()
    ->fc();

```

ntzm: Added some tests for chained static calls, thanks @egircys

keradus:
Travis is failing on PHP 5.6.
Could you take a look @ntzm ?

keradus:
Thank you @ntzm.

PR #3538, opened 2018-02-15
Don't use third parameter of Tokens::findBlockStart

ntzm:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3527#issuecomment-365867150


keradus:
Thank you @ntzm.

PR #3539, opened 2018-02-15
Update UnifiedDiffOutputBuilder from gecko-packages/gecko-diff-output-builder usage after it was incorporated into sebastian/diff

keradus:


SpacePossum:
if all green, lets :ship: 
thanks for the backporting man :)

keradus:
thanks for making strict udiffer part of main package ;)

SpacePossum:
>thanks for making strict udiffer part of main package ;)

you're welcome, I'm happy it all turned out well : D
test are green, so lets ship it!

Issue #3541, opened 2018-02-16
SCA: use `composer-normalize`


keradus:
Target: 2.2 LTS
Travis stage: SCA

We shall ensure at Travis that our composer is normalized.

keradus:
@localheinz , could you help with this one ?

localheinz:
@keradus 

Absolutely, give me a moment!

PR #3542, opened 2018-02-16
Enhancement: Run composer-normalize on Travis CI

localheinz:
This PR

* [x] runs `composer normalize` on Travis CI

Related to #3541.


.travis.yml
```diff
@@ -53,6 +53,7 @@ jobs:
                 - ./check_trailing_spaces.sh || travis_terminate 1
                 - if [ -n "$COMMIT_SCA_FILES" ]; then ./dev-tools/vendor/bin/phpmd `echo "$COMMIT_SCA_FILES" | grep -Ev "^tests/Fixtures" | xargs | sed 's/ /,/g'` text phpmd.xml || travis_terminate 1; fi
                 - ./dev-tools/vendor/bin/composer-require-checker check composer.json --config-file=.composer-require-checker.json
+                - composer require --dev localheinz/composer-normalize:0.5.0 && composer normalize --dry-run
```
localheinz: @keradus 

Not sure if this is what you like, I can imagine you would prefer me to add `localheinz/composer-normalize` to `/dev-tools/composer.json`, is that so?

If so I need to make some adjustments to `localheinz/composer-normalize` to allow for pointing to `composer.json` in a different directory.

localheinz:
Looks like I need to make some adjustments to the output, got extra empty lines: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/342438646#L681-L876.

.travis.yml
keradus: I was thinking about adding it to `/dev-tools/composer.json` indeed
would that be a big change ?

dev-tools/composer.json
```diff
@@ -1,8 +1,9 @@
 {
     "require": {
-        "php": "^7.0"
+        "php": "^7.1"
```
localheinz: Needed to update because of requirement in `localheinz/composer-normalize`. See https://github.com/localheinz/composer-normalize/pull/30.

.travis.yml
```diff
@@ -53,6 +53,7 @@ jobs:
                 - ./check_trailing_spaces.sh || travis_terminate 1
                 - if [ -n "$COMMIT_SCA_FILES" ]; then ./dev-tools/vendor/bin/phpmd `echo "$COMMIT_SCA_FILES" | grep -Ev "^tests/Fixtures" | xargs | sed 's/ /,/g'` text phpmd.xml || travis_terminate 1; fi
                 - ./dev-tools/vendor/bin/composer-require-checker check composer.json --config-file=.composer-require-checker.json
+                - composer normalize -d ./dev-tools ./../composer.json --dry-run
```
localheinz: @keradus 

Added the `file` argument, and fixed the issue with the extra new lines in the diff, take a look at https://github.com/localheinz/composer-normalize/compare/0.5.0...0.6.0. Hope you like it!

keradus: looks great ;)
actually, let us run it against both composer files, because "why not" :)

btw, why `composer/normalize` instead of `/vendor/bin/composer-normalize`? ah, as it's a plugin, not a "binary" app. that's why path was not there previously.

keradus: `travis_terminate 1` to previous line, it was not there only as it was last command

dev-tools/composer.json
```diff
@@ -1,8 +1,9 @@
...
     },
     "require-dev": {
+        "localheinz/composer-normalize": "0.6.0",
```
keradus: do we need to lock to concrete version?
why not `^0.6` ?

.travis.yml
```diff
@@ -52,7 +52,8 @@ jobs:
                     && (echo "UNKNOWN FILES DETECTED" && travis_terminate 1) || echo "NO UNKNOWN FILES"
                 - ./check_trailing_spaces.sh || travis_terminate 1
                 - if [ -n "$COMMIT_SCA_FILES" ]; then ./dev-tools/vendor/bin/phpmd `echo "$COMMIT_SCA_FILES" | grep -Ev "^tests/Fixtures" | xargs | sed 's/ /,/g'` text phpmd.xml || travis_terminate 1; fi
-                - ./dev-tools/vendor/bin/composer-require-checker check composer.json --config-file=.composer-require-checker.json
+                - ./dev-tools/vendor/bin/composer-require-checker check composer.json --config-file=.composer-require-checker.json || travis_terminate 1; fi
+                - composer normalize -d ./dev-tools ./../composer.json --dry-run
```
keradus: btw, why `composer/normalize` instead of `/vendor/bin/composer-normalize`? ah, as it's a plugin, not a "binary" app. that's why path was not there previously.

keradus:
Thank you @localheinz.

localheinz:
Thank you, @keradus!

keradus:
no, thank you !

keradus:
merged up to 3.0

PR #3547, opened 2018-02-17
MultilineWhitespaceBeforeSemicolonsFixer - chained call for a return fix

egircys:
Fixes the case when the chained call is on a instantiated object within the return statement.

```php
return (new Foo($bar))
    ->baz();

return (new Foo($bar))
    ->baz()
;
````

### DIFF
--- a/tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
+++ b/tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
@@ -424,6 +424,62 @@ $this
                         ->method2();
                 ?>',
             ],
+            [
+                '<?php
+
+                    function foo($bar)
+                    {
+                        if ($bar === 1) {
+                            $baz
+                                ->bar()
+                            ;
+                        }
+
+                        return (new Foo($bar))
+                            ->baz()
+                        ;
+                    }
+                ?>',
+                '<?php
+
+                    function foo($bar)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ntzm** [comment]
What about

```php
$foo = (new Foo($bar))
    ->baz();
```
?

egircys:
It's already covered, but i've added a test case for it :)
> What about
> 
> $foo = (new Foo($bar))
>     ->baz();
> ?

ntzm:
Ah cool just making sure!

tests/Fixer/Semicolon/MultilineWhitespaceBeforeSemicolonsFixerTest.php
```diff
@@ -424,6 +424,66 @@ public function provideSemicolonForChainedCallsFixCases()
...
+
+                    $foo = (new Foo($bar))
+                        ->baz();
+
+                    function foo($bar)
+                    {
+                        $foo = (new Foo($bar))
+                            ->baz();
+                    }
+;
```
julienfalque: Are those `;` at the end of new test cases intended?

egircys: actually not! :) removed it :)

keradus:
I see no new rule nor new option.
I'm downgrading this from feature to bugfix.

keradus:
Thank you @egircys.

PR #3549, opened 2018-02-19
DescribeCommand - use our Differ wrapper class, not external one directly

keradus:
Like we do on 2.2 line.

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -12,9 +12,8 @@
 
 namespace PhpCsFixer\Console\Command;
 
-use PhpCsFixer\Diff\v3_0\Differ;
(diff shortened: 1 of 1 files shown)


keradus:
ah, accidentally already merged into 2.11, not 2.10. fixing the milestone.
good idea Julien, I was just too fast :(
in the end, no big deal, 2.10 is not LTS and gonna be dropped soon

keradus:
Actually, I couldn't use 2.10. as 2.10 is using diff 1.2 and only master is using 1.3, so master was only possible choice :D

julienfalque:
But the `UnifiedDiffer` class already exists on 2.10, so it should be possible anyway, as the diff dependency is wrapped in this class, no?

keradus:
ha, good point ;)

SpacePossum:
but it had different implementation, so 2.10 would be a BC break ;)

keradus:
we don't have BC promise on txt output of any command.

PR #3556, opened 2018-02-20
DX: Add AutoReview/ComposerTest

keradus:
Inspired by #3555 

So, during merging one line into another, I will not mess up with branch aliasing

### DIFF
--- a/tests/AutoReview/ComposerTest.php
+++ b/tests/AutoReview/ComposerTest.php
@@ -43,7 +43,7 @@ final class ComposerTest extends TestCase
         $this->assertTrue(isset($aliases['dev-master']), 'The only branch that could be aliased is "dev-master".');
 
         $this->assertSame(
-            $this->convertAppVersiontoAliasedVersion(Application::VERSION),
+            $this->convertAppVersionToAliasedVersion(Application::VERSION),
             $aliases['dev-master'],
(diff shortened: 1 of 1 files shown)


tests/AutoReview/ComposerTest.php
```diff
@@ -0,0 +1,60 @@
...
+ * @group auto-review
+ */
+final class ComposerTest extends TestCase
+{
+    public function testBranchAlias()
+    {
+        $composerJson = json_decode(file_get_contents(__DIR__.'/../../composer.json'), true);
+
+        if (!isset($composerJson['extra']['branch-alias'])) {
+            $this->assertTrue(true); // composer.json doesn't contain branch alias, all good!
```
SpacePossum: ? maybe better to just increase the assert count and not an always pass assert?

keradus: I was looking for that at night, could you please remind me the method ?

kubawerlos: `addToAssertionCount`

keradus: That's why I couldn't find it :D 
```
ker@dus:~/github/PHP-CS-Fixer λ git grep addToAssertionCount
ker@dus:~/github/PHP-CS-Fixer λ 
```

PR #3557, opened 2018-02-21
AbstractDoctrineAnnotationFixer: edge case bugfix

Slamdunk:
```
    [RuntimeException]
    Index invalid or out of range

  PhpCsFixer\AbstractDoctrineAnnotationFixer->nextElementAcceptsDoctrineAnnotations()
    in ./vendor/friendsofphp/php-cs-fixer/src/AbstractDoctrineAnnotationFixer.php at line 52
  PhpCsFixer\AbstractDoctrineAnnotationFixer->applyFix()
    in ./vendor/friendsofphp/php-cs-fixer/src/AbstractFixer.php at line 73
  PhpCsFixer\AbstractFixer->fix()
    in ./vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php at line 190
  PhpCsFixer\Runner\Runner->fixFile()
    in ./vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php at line 132
  PhpCsFixer\Runner\Runner->fix()
    in ./vendor/friendsofphp/php-cs-fixer/src/Console/Command/FixCommand.php at line 219
  PhpCsFixer\Console\Command\FixCommand->execute()
    in ./vendor/symfony/console/Command/Command.php at line 252
  [ ... ]
```

### DIFF
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -207,6 +207,10 @@ abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements 
     {
         do {
             $index = $tokens->getNextMeaningfulToken($index);
+
+            if (null === $index) {
+                return false;
+            }
         } while ($tokens[$index]->isGivenKind(array(T_ABSTRACT, T_FINAL)));
 
         if ($tokens[$index]->isClassy()) {
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -208,9 +208,9 @@ abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements 
         do {
             $index = $tokens->getNextMeaningfulToken($index);
 
-            if (null === $index) {
-                return false;
-            }
+//             if (null === $index) {
+//                 return false;
+//             }
         } while ($tokens[$index]->isGivenKind(array(T_ABSTRACT, T_FINAL)));
 
         if ($tokens[$index]->isClassy()) {
--- a/src/AbstractDoctrineAnnotationFixer.php
+++ b/src/AbstractDoctrineAnnotationFixer.php
@@ -208,9 +208,9 @@ abstract class AbstractDoctrineAnnotationFixer extends AbstractFixer implements 
         do {
             $index = $tokens->getNextMeaningfulToken($index);
 
-//             if (null === $index) {
-//                 return false;
-//             }
+            if (null === $index) {
+                return false;
+            }
         } while ($tokens[$index]->isGivenKind(array(T_ABSTRACT, T_FINAL)));
 
         if ($tokens[$index]->isClassy()) {
--- a/tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
+++ b/tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
@@ -39,7 +39,18 @@ final class DoctrineAnnotationBracesFixerTest extends AbstractDoctrineAnnotation
      */
     public function provideFixWithBracesCases()
     {
-        return $this->createTestCases(array(
+        $edgeCases = array(
+            array(
+                '<?php
+
+/**
+ * @see \User getId()
+ */
+',
+            ),
+        );
+
+        return $edgeCases + $this->createTestCases(array(
             array('
 /**
  * @Foo()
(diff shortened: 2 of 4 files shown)


julienfalque:
Occured with a DocComment at the end of a file?

Slamdunk:
In production? No, at start of a regular class file.

src/AbstractDoctrineAnnotationFixer.php
```diff
@@ -215,6 +219,10 @@ private function nextElementAcceptsDoctrineAnnotations(PhpTokens $tokens, $index
 
         while ($tokens[$index]->isGivenKind(array(T_PUBLIC, T_PROTECTED, T_PRIVATE, T_FINAL, T_ABSTRACT))) {
             $index = $tokens->getNextMeaningfulToken($index);
+
+            if (null === $index) {
```
julienfalque: Not sure this can ever happen: no meaningful token after `public`, `protected`, `private`, `final` or `abstract` would be a parse error.

tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
```diff
@@ -39,7 +39,18 @@ public function testFixWithBraces($expected, $input = null)
      */
     public function provideFixWithBracesCases()
     {
-        return $this->createTestCases(array(
+        $edgeCases = array(
```
julienfalque: Without the fix, does this edge case also fail with other Doctrine Annotation fixers? If so, please add it to related test classes as well.

Slamdunk:
> Then I'm not sure the new test case actually covers the error you got.

It covers the first `AbstractDoctrineAnnotationFixer` change I made, because I follow TDD.

src/AbstractDoctrineAnnotationFixer.php
Slamdunk: I added this second change only to be safe, but not covered by any test (this specific addition).

tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
Slamdunk: Yes, but to be fair there should be an `AbstractDoctrineAnnotationFixerTest`, it's not a good thing to have repeated test cases all over subclasses.

src/AbstractDoctrineAnnotationFixer.php
Slamdunk: Sorry for that, done.

tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
julienfalque: I would still add the case in existing test classes. @keradus @SpacePossum WDYT?

julienfalque:
@Slamdunk I think the changes here still don't cover the initial error you had. Did you try putting the whole PHP file that triggered it as a test case?

Slamdunk:
https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/347214072#L802

julienfalque:
What I meant is that the test case you added represents a PHP file that contains a DocComment only, which, indeed, fails and is fixed by your patch. But you said the file that actually triggered the error when you used the tool contains two DocComments followed by a class definition. Unless I'm missing something, `$tokens->getNextMeaningfulToken()` should not return `null` when processing such file.

Slamdunk:
Likely I was wrong on the original file: I only know that I tested this patch in that repository and worked.
Sorry if I get you to spend time on my erroneous words :cry:

tests/Fixer/DoctrineAnnotation/DoctrineAnnotationBracesFixerTest.php
SpacePossum: I agree with both of you :)
If there is an `AbstractDoctrineAnnotationFixer` than there should be a `AbstractDoctrineAnnotationFixerTest` (there should be, maybe introduce it here? not sure if it is a lot of work), if there is not than the test should be added to the utests of all of the effected children

Slamdunk: Done

julienfalque:
@keradus Higher branches have an additional Doctrine Annotation fixer. Do you mind adding the test case in the related test class while merging? Or do you prefer this to be done in a dedicated PR?

keradus:
@julienfalque , please do me a favor and open dedicated PR for it, then we will know I didnt fucked up when applying changes locally - as there is no CI for my local changes (well, there is, when it's already merged and it's too late;P)

keradus:
Thank you @Slamdunk.

PR #3559, opened 2018-02-22
Don't expose new files under Test namespace

keradus:
as on higher branches, this namespace is deprecated

Appendix to #3526


keradus:
FYI @SpacePossum @julienfalque

PR #3560, opened 2018-02-22
SelfAccessorFixer - mark as risky

Slamdunk:
https://3v4l.org/LrvSc

Behavoir present in [Yii2](https://github.com/yiisoft/yii2):
https://github.com/yiisoft/yii2/blob/4f80cda7130da5259ce93fe41eb681cbb70f30be/framework/base/BaseObject.php#L84-L87

S****y inheritance code, but still this fixer changes the user code behaviour.


keradus:
I see that our AutoReview is doing great job :D

great catch, but extra sad for me :(

yet, IMO, we could try to fix it by not changing static method calls. WDYT ?

Slamdunk:
> yet, IMO, we could try to fix it by not changing static method calls. WDYT ?

After years of usage of this Fixer, this is the first time I see it, and Yii2 is gonna remove that funcionality in the next release.

I'm not gonna spend time on this edge case: risky comment is enough for me.

julienfalque:
:+1: for keeping the behavior as-is and mark it as risky.

keradus:
Thank you @Slamdunk.

PR #3561, opened 2018-02-22
PHPUnit5 - add in place missing compat layer for PHPUnit6

keradus:
I'm so happy now...

as always, our usage of PHPUnit push it to the limits or even above, and it's not fully ready for that.


keradus:
merging fast as it's fixing broken 2.10 line (CIs are red),
if one would like to make it better - please go ahead and send a PR, it's much welcome!

Issue #3567, opened 2018-02-23
All config options should follow the same casing rules


ntzm:
Most options seems to follow `snake_case`, but some like `header_comment` have `commentType`. Would be good to normalise these

keradus:
Good point.

Things to do:
- create an AutoReview test that check if all options are snake_case,
- for those which not, make new option that suppresses old option (`comment_type` over `commentType`), deprecate old option and make it an exception to new AutoReview check
- make dedicated PR for 3.0 line to remove deprecated options

keradus:
@ntzm , would you like to dive into it?

ntzm:
I would love to!

ntzm:
So in terms of branches:

- The AutoReview targets 2.2
- The new options that suppress the old options targets 2.2
- The deprecation of non snake_case options targets master
- The removal of deprecated options targets 3.0

Is this correct?

mvorisek:
@SpacePossum is there a roadmap when the 3.0 version will be releases?

SpacePossum:
Hi,
I missed the last checking with the maintainers, but AFAIK there is no target date besides "this year" (I hope we make that one : ) ).

keradus:
all identified cases were marked as aliased and provided with both casing on 2.x line, and then cleaned up on 3.0-dev

PR #3568, opened 2018-02-23
New Autoreview: Correct option casing

ntzm:
Ref: #3567

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -272,12 +272,14 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``multiLineExtendsEachSingleLine`` (``bool``): whether definitions should be
-    multiline; defaults to ``false``
-  - ``singleItemSingleLine`` (``bool``): whether definitions should be single line
-    when including a single item; defaults to ``false``
-  - ``singleLine`` (``bool``): whether definitions should be single line; defaults
-    to ``false``
+  - ``multi_line_extends_each_single_line`` (``bool``): whether definitions should
+    be multiline; defaults to ``false``; DEPRECATED alias:
+    ``multiLineExtendsEachSingleLine``
+  - ``single_item_single_line`` (``bool``): whether definitions should be single
+    line when including a single item; defaults to ``false``; DEPRECATED alias:
+    ``singleItemSingleLine``
+  - ``single_line`` (``bool``): whether definitions should be single line; defaults
+    to ``false``; DEPRECATED alias: ``singleLine``
 
 * **class_keyword_remove**
 
@@ -474,8 +476,8 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``commentType`` (``'comment'``, ``'PHPDoc'``): comment syntax type; defaults to
-    ``'comment'``
+  - ``comment_type`` (``'comment'``, ``'PHPDoc'``): comment syntax type; defaults to
+    ``'comment'``; DEPRECATED alias: ``commentType``
   - ``header`` (``string``): proper header content; required
   - ``location`` (``'after_declare_strict'``, ``'after_open'``): the location of the
     inserted header; defaults to ``'after_declare_strict'``
@@ -790,10 +792,11 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``importsOrder`` (``array``, ``null``): defines the order of import types; defaults
-    to ``null``
-  - ``sortAlgorithm`` (``'alpha'``, ``'length'``): whether the statements should be
-    sorted alphabetically or by length; defaults to ``'alpha'``
+  - ``imports_order`` (``array``, ``null``): defines the order of import types; defaults
+    to ``null``; DEPRECATED alias: ``importsOrder``
+  - ``sort_algorithm`` (``'alpha'``, ``'length'``): whether the statements should be
+    sorted alphabetically or by length; defaults to ``'alpha'``; DEPRECATED
+    alias: ``sortAlgorithm``
 
 * **php_unit_construct** [@Symfony:risky]
 
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -17,6 +17,7 @@ use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\DefinedFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerConfiguration\AliasedFixerOption;
 use PhpCsFixer\FixerConfiguration\AllowedValueSubset;
 use PhpCsFixer\FixerConfiguration\FixerOptionInterface;
 use PhpCsFixer\FixerFactory;
@@ -525,6 +526,10 @@ EOF
                             $line .= 'defaults to <comment>'.self::toString($option->getDefault()).'</comment>';
                         } else {
                             $line .= 'required';
+                        }
+
+                        if ($option instanceof AliasedFixerOption) {
+                            $line .= '; DEPRECATED alias: <comment>'.$option->getAlias().'</comment>';
                         }
 
                         foreach (self::wordwrap($line, 72) as $index => $line) {
--- a/src/FixerConfiguration/AliasedFixerOption.php
+++ b/src/FixerConfiguration/AliasedFixerOption.php
@@ -1,0 +1,103 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\FixerConfiguration;
+
+/**
+ * @author ntzm
+ *
+ * @internal
+ *
+ * @todo 3.0 Drop this class
+ */
+final class AliasedFixerOption implements FixerOptionInterface
+{
+    /**
+     * @var FixerOptionInterface
+     */
+    private $fixerOption;
+
+    /**
+     * @var string
+     */
+    private $alias;
+
+    public function __construct(FixerOptionInterface $fixerOption, $alias)
+    {
+        $this->fixerOption = $fixerOption;
+        $this->alias = $alias;
+    }
+
+    /**
+     * @return string
+     */
+    public function getAlias()
+    {
+        return $this->alias;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getName()
+    {
+        return $this->fixerOption->getName();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDescription()
+    {
+        return $this->fixerOption->getDescription();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function hasDefault()
+    {
+        return $this->fixerOption->hasDefault();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefault()
+    {
+        return $this->fixerOption->getDefault();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAllowedTypes()
+    {
+        return $this->fixerOption->getAllowedTypes();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getAllowedValues()
+    {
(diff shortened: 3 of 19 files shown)


README.rst
```diff
@@ -276,11 +276,11 @@ Choose from the list of available rules:
 
   Configuration options:
 
-  - ``multiLineExtendsEachSingleLine`` (``bool``): whether definitions should be
-    multiline; defaults to ``false``
-  - ``singleItemSingleLine`` (``bool``): whether definitions should be single line
-    when including a single item; defaults to ``false``
-  - ``singleLine`` (``bool``): whether definitions should be single line; defaults
+  - ``multi_line_extends_each_single_line`` (``bool``): whether definitions should
```
ntzm: Should aliases be documented?

keradus: yes, look at `around_argument_assignments` and follow the approach

src/FixerConfiguration/FixerOptionBuilder.php
```diff
@@ -108,6 +113,18 @@ public function setNormalizer(\Closure $normalizer)
         return $this;
     }
 
+    /**
+     * @param string $preferredName
+     *
+     * @return $this
+     */
+    public function setAlias($preferredName)
```
keradus: this is not internal class, in perfect world, we don't want to have aliases. For that ,we don't want to expose this "new feature" to users and allow them to use it - as this means we need to support it in long term.
Instead, make new, internall class perhaps?
take a look at `FixerConfigurationResolverRootless` - when we wanted all options to have root, but not all of them did, we added temporary class to ease development of it.

ntzm: Yeah I thought as much! Thanks for the pointer :)

src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -62,6 +62,10 @@ public function resolve(array $options)
                 $alias = $option->getAlias();
 
                 if (array_key_exists($alias, $options)) {
+                    if (array_key_exists($name, $options)) {
+                        throw new \LogicException(sprintf('Aliased option %s/%s is passed multiple times.', $name, $alias));
```
keradus: too late for me to check it, but it shall be `InvalidFixerConfigurationException` or `InvalidOptionsException`

src/FixerConfiguration/FixerOptionBuilder.php
ntzm: I've implemented decorator classes for `FixerOption` and `FixerOptionBuilder`, I think it solves the issues you brought up!

src/FixerConfiguration/FixerConfigurationResolver.php
ntzm: I couldn't throw `InvalidFixerConfigurationException` from here because I need access to the fixer name which I don't believe is possible in here. Throwing symfony's `InvalidOptionsException` feels a little hacky, but it means it's upgraded to a `InvalidFixerConfigurationException` by `AbstractFixer::configure`

src/FixerConfiguration/AliasedFixerOptionBuilder.php
```diff
@@ -0,0 +1,96 @@
...
+ */
+
+namespace PhpCsFixer\FixerConfiguration;
+
+/**
+ * @author ntzm
+ *
+ * @internal
+ */
+final class AliasedFixerOptionBuilder
```
ntzm: If we decide to go for this decorator approach, would be worth making a `FixerOptionBuilderInterface` that this and `FixerOptionBuilder` implement

tests/Test/AbstractFixerTestCase.php
```diff
@@ -193,6 +213,39 @@ protected function lintSource($source)
         }
     }
 
+    protected function configureFixerWithAliasedOptions(array $configuration)
```
keradus: this is base class for our public-api test class, we shall not add this logic here

keradus: ping @ntzm 
let us make that one finished & merged ;)

julienfalque:
Shouldn't we deprecate options that have the wrong casing?

src/Console/Command/HelpCommand.php
```diff
@@ -508,6 +509,10 @@ function (FixerOptionInterface $optionA, FixerOptionInterface $optionB) {
                             $line .= 'required';
                         }
 
+                        if ($option instanceof AliasedFixerOption) {
+                            $line .= '; alias: <comment>'.$option->getAlias().'</comment>';
```
SpacePossum: don't we need an escape on `$option->getAlias()`? maybe wrap this in a `sprintf` as well

tests/Test/AbstractFixerTestCase.php
ntzm: Sorry! Got distracted with other things, will work on this tomorrow evening

ntzm:
@julienfalque I was under the impression that the deprecation should happen in the master branch not 2.2?

src/Console/Command/HelpCommand.php
ntzm: I was following line 486, no escaping used there

tests/Test/AbstractFixerTestCase.php
keradus: Great to hear !

keradus:
`trigger_error(E_USER_DEPRECATED, ...` shall not go to 2.2, only master
docs update that option is deprecated could go to 2.2, sure

eventually, both of those changes shall be applied ;)

julienfalque:
Ah indeed, deprecations must be merged on master... But if we are to actually deprecate wrong options, shouldn't we avoid giving them visibility via aliases? I mean, aliases give the impression that both names are fine and choice is up to final user.

keradus:
we shall put stress on that aliased options are deprecated, but one shall not un-document aliased values - as then, one is not able to search for docs of them

julienfalque:
I know. I meant to simply deprecate wrong options like we did for already deprecated one: mark the current ones as `deprecated` in the doc and create the new options that replace them.

keradus:
@ntzm , it looks like this one is very close to be ready, could you finish it please ?

ntzm:
Yes sorry, will finish tonight

keradus:
please remove `WIP` from PR title whenever you finish @ntzm , thanks !

ntzm:
This is no longer WIP :)

src/FixerConfiguration/AliasedFixerOptionBuilder.php
julienfalque: I'd rather add a `setAlias()` method to `FixerOptionBuilder`.

tests/FixerConfiguration/AliasedFixerOptionTest.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->setExpectedException('LogicException', 'No default value defined.');
+        $option->getDefault();
+    }
+
+    public function testGetAllowedTypes()
+    {
+        $option = new AliasedFixerOption(new FixerOption('foo', 'Bar.'), 'baz');
+        $this->assertNull($option->getAllowedTypes());
+
+        $option = new AliasedFixerOption(new FixerOption('foo', 'Bar.', true, null, array('bool')), 'baz');
```
julienfalque: Please use `@dataProvider`s to extract multiple cases from the test implementation itself (same for other tests).

tests/Test/AbstractFixerWithAliasedOptionsTestCase.php
```diff
@@ -0,0 +1,80 @@
...
+ * @author ntzm
+ *
+ * @internal
+ */
+abstract class AbstractFixerWithAliasedOptionsTestCase extends AbstractFixerTestCase
+{
+    /**
+     * @var null|\PhpCsFixer\Fixer\ConfigurableFixerInterface
+     */
+    private $fixerWithAliasedConfig;
```
julienfalque: Please override `tearDown()` or `tearDownAfterClass()` to unset this property after tests and avoid memory leaks.

src/FixerConfiguration/AliasedFixerOptionBuilder.php
julienfalque: I'd rather add a `setAlias()` method to `FixerOptionBuilder`.

ntzm:
Will sort those issues soon @julienfalque, on holiday atm

julienfalque:
See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3568/files#r178953652 as well.

tests/FixerConfiguration/AliasedFixerOptionTest.php
ntzm: As these tests are copied from `FixerOptionTest`, shall I change the tests there as well? Or shall I open a separate PR for that?

src/FixerConfiguration/AliasedFixerOptionBuilder.php
ntzm: @julienfalque see: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3568#discussion_r170391280

tests/FixerConfiguration/AliasedFixerOptionTest.php
julienfalque: I'd say separate PR :)

tests/Test/AbstractFixerWithAliasedOptionsTestCase.php
julienfalque: Yep, but this is assuming PHPUnit will stop referencing the instance of the test as soon as all of its test methods have been executed. In my experience this is rarely (never?) true. For that reason I avoid using properties in tests at all. And if I have to, I always use `tearDown` or `tearDownAfterClass` to unset them.

tests/FixerConfiguration/AliasedFixerOptionTest.php
```diff
@@ -25,31 +25,104 @@
...
-        $this->assertSame('foo', $option->getName());
+        $this->assertSame($expected, $input->getName());
     }
 
-    public function testGetDescription()
+    public function provideGetNameCases()
     {
-        $option = new AliasedFixerOption(new FixerOption('foo', 'Bar.'), 'baz');
-        $this->assertSame('Bar.', $option->getDescription());
+        return array(
```
julienfalque: You can simplify data set to the bare minimum: the expected name. Since you expect the result to be exactly what you pass to the `FixerOption` constructor, you can use `$expected` as both the assertion's expected value and as `FixerOption` parameter in the test itself. You can probably simplify other data sets the same way.

tests/Test/AbstractFixerWithAliasedOptionsTestCase.php
ntzm: Done 👍

ntzm:
Anything else I need to do to get this approved?

src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -58,6 +59,19 @@ public function resolve(array $options)
         foreach ($this->options as $option) {
             $name = $option->getName();
 
+            if ($option instanceof AliasedFixerOption) {
+                $alias = $option->getAlias();
+
+                if (array_key_exists($alias, $options)) {
+                    if (array_key_exists($name, $options)) {
```
julienfalque: Please add the following comment before this line: `// @TODO on 2.12: trigger a deprecation notice and add a test for it`.

src/FixerConfiguration/AliasedFixerOption.php
```diff
@@ -0,0 +1,101 @@
...
+ */
+
+namespace PhpCsFixer\FixerConfiguration;
+
+/**
+ * @author ntzm
+ *
+ * @internal
+ */
+final class AliasedFixerOption implements FixerOptionInterface
```
julienfalque: Please add the following annotation is the DocComment of this class, `AliasedFixerOptionBuilder` and `AbstractFixerWithAliasedOptionsTestCase`: `@TODO on 3.0: drop this class`.

tests/Test/AbstractFixerWithAliasedOptionsTestCase.php
```diff
@@ -0,0 +1,87 @@
...
+
+    protected function tearDown()
+    {
+        parent::tearDown();
+
+        unset($this->fixerWithAliasedConfig);
```
julienfalque: Set the property to `null` rather than using `unset()`, which drops the property itself.

keradus:
please re-run readme generation

keradus:
Thank you @ntzm.

keradus:
@ntzm , can you resolve TODOs for 2.12 on master branch ?

PR #3572, opened 2018-02-28
DX: allow for more phpunit-speedtrap versions to support more PHPUnit versions

keradus:
basically newly released v3 added support for PHPUnit v7


SpacePossum:
AppVeyor is sad, can we retrigger?

keradus:
retrigger, but It wasn't because of code, but issues with PHP installation itself. let us see now

PR #3573, opened 2018-02-28
minor: add test: multi-line comments before else indented correctly.

demiankatz:
See also discussion on #3359.

TODO
- [x] Create test case to demonstrate problem
- [x] Implement fix for problem

### DIFF
--- a/tests/Fixer/Basic/BracesFixerTest.php
+++ b/tests/Fixer/Basic/BracesFixerTest.php
@@ -1775,6 +1775,39 @@ if (1) {
             ],
             [
                 '<?php
+if (1) {
+    if (2) {
+        $a = "a";
+    } elseif (3) {
+        $b = "b";
+    // comment line 1
+    // comment line 2
+    // comment line 3
+    // comment line 4
+    } else {
+        $c = "c";
+    }
+    $d = "d";
+}',
+                '<?php
+if (1) {
+    if (2) {
+        $a = "a";
+    } elseif (3) {
+        $b = "b";
+        // comment line 1
+        // comment line 2
+// comment line 3
+            // comment line 4
+    } else {
+        $c = "c";
+    }
+    $d = "d";
+}',
+                self::CONFIGURATION_OOP_POSITION_SAME_LINE,
+            ],
+            [
+                '<?php
 foreach ($numbers as $num) {
     for ($i = 0; $i < $num; ++$i) {
         $a = "a";


demiankatz:
This is currently work in progress -- I have implemented a test but am finding the solution to the problem challenging. Input is welcomed!

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3359/files#diff-0b7fa2dc2594eae7a18836044e3287d8R476
here, we do know that "next token" is either last meaningful token before `}` or it's comment for next block already (so just before `} elseif...` for example)

for that, we could add here that if "next token" is comment, we do still add that whitespaces, but also check if "next next token" is comment as well, and if so we adjust whitespace before it as well.

that would be my first guess at least...

tests/Fixer/Basic/BracesFixerTest.php
```diff
@@ -1768,6 +1768,35 @@ public function __construct(
...
+    $d = "d";
+}',
+                '<?php
+if (1) {
+    if (2) {
+        $a = "a";
+    } elseif (3) {
+        $b = "b";
+        // comment line 1
+        // comment line 2
```
keradus: just to be fancy, let us extend it to:
```
+        // comment line 1
+        // comment line 2
+            // comment line 3
+    // comment line 4
```

demiankatz: Good idea!

demiankatz:
@keradus, I've put about an hour into this without making any headway. It seems like all four lines of comment are being processed as a unit, and I can't seem to figure out where or how to break them apart. I guess my fundamental lack of understanding is how the "cursor" is flowing through the code. Do I want to change something in ensureWhitespaceAtIndexAndIndentMultilineComment? Do I need to change something about the loop at a higher level? Apologies for not making greater progress -- alas, my time is very limited.

ntzm:
How it works currently is how I expect it to work, is this definitely a bug?

SpacePossum:
>How it works currently is how I expect it to work, is this definitely a bug?

I would say so; if the fixer moves comments it should move those as a "block" and not just the last line of a multi line comment or the last comment of a comment "block".

demiankatz:
@ntzm, the bug is that if you have multiple comments in a row, only the first one gets de-indented. I would expect either:

<pre>
    // line 1
    // line 2
    } else {
</pre>

or:

<pre>
        // line 1
        // line 2
    } else {
</pre>

but definitely not:

<pre>
    // line 1
        // line 2
    } else {
</pre>

This last case is what we're actually getting right now.

ntzm:
Ignore me, I was reading the test case wrong! Sorry

SpacePossum:
The testcase in the PR currently passes on latest of the main branch, as such I think this PR can be closed (the linked issue is also closed)

demiankatz:
I'm getting a static code analysis failure from CI, but I'm not sure why -- the error message doesn't seem to relate to my code. Is this an existing problem in master, or am I missing something?

julienfalque (approved):

julienfalque:
Thank you @demiankatz.

Issue #3570, opened 2018-02-27
GeneralPhpdocAnnotationRemoveFixer - does not remove given annotation


MarcusFulbright:
When trying to use the rule `general_phpdoc_annotation_remove` to remove some `@noinspection` annotations, I get get the following error: `[InvalidArgumentException]  Cannot set empty content for id-based Token.`

The particulars of my set up are below: 

#### The PHP version you are using (`$ php -v`):
7.1.11

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
2.10.3

#### The command you use to run PHP CS Fixer:
vendor/bin/php-cs-fixer fix --config=.php_cs.dist.php -vvv

#### The configuration file you are using, if any:
Only Rule: ` 'general_phpdoc_annotation_remove' => ['annotations' => ["noinspection"]] `

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
/** @noinspection PhpAssignmentInConditionInspection */
while ($something = myFunction($foo) {
  //code 
}

I'd expect the annotation to get removed.

while ($something = myFunction($foo)) {
  //code 
}

SpacePossum:
Hi @MarcusFulbright and thanks for reporting.

From the sample and config you provided I created a unit test to prove the bug and proposed a fix here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3574

PR #3574, opened 2018-03-01
GeneralPhpdocAnnotationRemoveFixer - remove PHPDoc if no content is left

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3570

### DIFF
--- a/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
+++ b/src/Fixer/Phpdoc/GeneralPhpdocAnnotationRemoveFixer.php
@@ -93,7 +93,11 @@ function foo() {}',
                 $annotation->remove();
             }
 
-            $tokens[$index] = new Token(array(T_DOC_COMMENT, $doc->getContent()));
+            if ('' === $doc->getContent()) {
+                $tokens->clearTokenAndMergeSurroundingWhitespace($index);
+            } else {
+                $tokens[$index] = new Token(array(T_DOC_COMMENT, $doc->getContent()));
+            }
         }
     }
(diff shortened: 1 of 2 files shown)


keradus:
@TomasVotruba , either that one or similar one you have also faced if I remember correctly.
just pinging you, maybe you will find it useful.

keradus:
Thank you @SpacePossum.

TomasVotruba:
@keradus Thanks for the ping. I'll wait for 3.0 release then verify.

keradus:
well, it will be released under 2.2 ;)

TomasVotruba:
Sure. I meant it in the chunk upgrade with other changes.   I put all my os time to Rector now

TomasVotruba:
@keradus I just check it and this only applied when some annotations are removed by this fixer. Not to other empty doc blocks.

The fix is needed in `NoEmptyPhpdocFixer`

keradus:
yep, this change affects only `GeneralPhpdocAnnotationRemoveFixer`
if you want to change `NoEmptyPhpdocFixer`, open a PR :+1:

TomasVotruba:
Have you changed your view on this? IIRC you've declined it before

keradus:
and then more ppl wanted it, you have proof of that here - this PR.
when it's one-vs-one, status quo is maintained, but if more ppl want some change, well, i don't need to agree with it to accept it.

TomasVotruba:
Allright: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3604

PR #3578, opened 2018-03-04
Add interface for deprecated options

julienfalque:


src/Console/Command/DescribeCommand.php
```diff
@@ -209,6 +210,14 @@ private function describeRule(OutputInterface $output, $name)
                     $line .= '<comment>required</comment>';
                 }
 
+                if ($option instanceof DeprecatedFixerOption) {
+                    $line .= '. <error>DEPRECATED</error>: '.preg_replace(
+                        '/(`.+?`)/',
+                        '<info>$1</info>',
+                        lcfirst(OutputFormatter::escape($option->getDeprecationMessage()))
```
SpacePossum: shouldn't the escape be the last operation here (after the `lcfirst`)?

SpacePossum:
awesome stuff :)   :+1:

src/Console/Command/DescribeCommand.php
julienfalque: Escaping doesn't have any impact on alphanumeric characters so the result would be the same.

SpacePossum: it used to be and it might be again in the future, I think it is best not to rely on the implementation details about which character(s) are used for the escaping by SF, but than again you are right here about the result being the same for now so :)

julienfalque: You're right, maybe worth creating a dedicated PR as we do the same in several other places.

keradus:
Thank you @julienfalque.

PR #3585, opened 2018-03-08
Add ReturnAssignmentFixer

SpacePossum:
```
$ php php-cs-fixer describe return_assignment
Description of return_assignment rule.
Non global, static or indirect referenced variables should not be assigned and directly returned by a function or method.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,5 +1,4 @@
    <?php
    function a() {
   -    $a = 1;
   -    return $a;
   +    return 1;
    }
   
```

after almost 2 years;
Closes #1853


src/Fixer/ReturnNotation/ReturnAssignmentFixer.php
```diff
@@ -0,0 +1,349 @@
...
+
+        if (!$tokens[$nextIndex]->isWhitespace()) {
+            $tokens->insertAt($nextIndex, new Token([T_WHITESPACE, ' ']));
+            ++$inserted;
+        }
+
+        return $inserted;
+    }
+
+    private function clearIfSave(Tokens $tokens, $index)
```
SpacePossum: TODO better name

src/Fixer/ReturnNotation/ReturnAssignmentFixer.php
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        if ($tokens[$index]->isWhitespace() && $tokens[$tokens->getPrevNonWhitespace($index)]->isComment()) {
+            return;
+        }
+
+        $tokens->clearAt($index);
+    }
+
+    private function isSuperGlobal(Token $token)
```
SpacePossum: TODO:
missing PHPDoc

src/Fixer/ReturnNotation/ReturnAssignmentFixer.php
```diff
@@ -0,0 +1,349 @@
...
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // FIXME
+        // should be run before white space lines, extra empty lines,
```
keradus: and before `blank_line_before_return`/`blank_line_before_statement`

SpacePossum: I don't think that is needed, the fixer removes the line where the statement was on if the line is empty

keradus: it's not about removing, it's about adding new line.

```php
function foo()
{
    $foo = [1, 2, 3];
    $bar = array_filter($foo, ...);

    return $bar;
}
```

what would be the outcome?

```php
function foo()
{
    $foo = [1, 2, 3];
    return array_filter($foo, ...);
}
```

so far, so good, right?

but, `blank_line_before_return` want to put blank line before `return`.
if it will be executed on original code, it will say that all is fine.
if it will be executed on fixed code, it will add new line.

for that, `return_assignment` shall be executed before `blank_line_before_return`

SpacePossum:
rebased and squashed

keradus:
please re-run fixer

README.rst
```diff
@@ -1364,6 +1364,11 @@ Choose from the list of available rules:
     ones; defaults to ``['getrandmax' => 'mt_getrandmax', 'rand' =>
     'mt_rand', 'srand' => 'mt_srand']``
 
+* **return_assignment**
+
+  Non global, static or indirect referenced variables should not be
```
greg0ire: "indirectly referenced"?

SpacePossum: :+1:
I update this when everything else in this PR is approved

keradus:
Thank you @SpacePossum.

keradus:
magic :D

Issue #3286, opened 2017-12-01
How to disable xdebug when running php-cs-fixer?


alexciarlillo:
I am trying to run php-cs-fixer with xdebug disabled. I found an old issue for Atom here: https://github.com/pfefferle/atom-php-cs-fixer/issues/27 but that suggestion no longer seems to work. If you run php with -n, you no longer have the phar extension. I then tried

`php -n -d extension=phar.so /usr/local/bin/php-cs-fixer fix ...` but then received errors about undefined constants. 

I want xdebug enabled *most* of the time because I use it heavily while testing. Is there an easy foolproof way to disable it only when running php-cs-fixer?

#### The PHP version you are using (`$ php -v`):
=> 7.1.11

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.8.3

julienfalque:
AFAIK there is no easy way to disable an extension on demand when running a PHP command. `php -n` will prevent loading *all* ini settings including extensions, but then you disable extensions that are required by PHP CS Fixer to run, which is probably the cause of the undefined constants errors you get.

I would suggest to disable Xdebug by default for CLI instead, and enable it on demand. You will speed up execution not only for PHP CS Fixer but *any* PHP command (Composer?), and slow down the ones that actually need it only. I published a little [tool](https://github.com/julienfalque/xdebug) that might help you with that.

---
Composer has a piece of code that allows it to restart itself with Xdebug disabled. We recently rejected doing the same here (#3084) but now they are extracting this piece of code into its own [package](https://github.com/composer/xdebug-handler), maybe we could consider using it?

keradus:
could be interesting, yet https://github.com/composer/xdebug-handler/issues/5 is must have to start using it.

sanmai:
It appears to be that composer/xdebug-handler#5 was patched on 20 Dec 2017. Can we move forward with this issue?

ad-si:
I'm having the same issue. Is there an easy solution by now?

keradus:
yes. disable your xdebug globally and run it on demand only.

ad-si:
Well, that was easy 😅. `phpdismod xdebug` did the job for me on ubuntu. Thanks!

JoseMMorales:
If using Ubuntu you can enter $ sudo phpdismod xdebug to disable that service :)

PR #3587, opened 2018-03-09
Use XdebugHandler to avoid perormance penalty

AJenbo:


SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3286

php-cs-fixer
```diff
@@ -52,8 +52,14 @@ try {
     }
 }
 
+use Composer\XdebugHandler\XdebugHandler;
 use PhpCsFixer\Console\Application;
 
+// Restart if xdebug is loading, unless the environment variable PHP_CS_FIXER_ALLOW_XDEBUG is set.
+$xdebug = new XdebugHandler('PHP_CS_FIXER', '--ansi');
```
keradus: why we need to specify colors here ?

php-cs-fixer
```diff
@@ -52,8 +52,14 @@ try {
     }
 }
 
+use Composer\XdebugHandler\XdebugHandler;
 use PhpCsFixer\Console\Application;
 
+// Restart if xdebug is loading, unless the environment variable PHP_CS_FIXER_ALLOW_XDEBUG is set.
```
keradus: `is loaded` ?

keradus:
Hi, thanks for PR @AJenbo , I'm happy we could finally use this package and you decided to help!
Please, could you change target branch for 2.2 LTS ?

php-cs-fixer
AJenbo: Because with out it the subprocess will think it is in a text only terminal and color and progress-bars won't output correctly.

keradus: hm... how it would behave if console is missing colors indeed and we pass `--ansi` ? ?

AJenbo: It only passes the parameter if the parent is in a terminal that supports color. This is not to to set the parameter but only to tell xdebughandler what the parameter is called in case it needs to use it. Other applications call it "--color" and others don't even have such a feature, so XdebugHandler can simply know the name of the parameter before hand, but again, it is only passed to the child process if relevant.

AJenbo:
Opened a new PR for 2.2 here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3590

PR #3590, opened 2018-03-09
Use XdebugHandler to avoid perormance penalty

AJenbo:


AJenbo:
Fix type "loading" -> "loaded"

SpacePossum:
the warning would still be valid if the command is used outside of the phar I think?

julienfalque:
I assumed the Xdebug handler works for any PHP CLI process, not only phar files. @AJenbo is that right? If so the warning would never be shown, unless one explicitly disables the handler at runtime somehow, but then showing a warning would not make sense as the user *does* want to use Xdebug.

AJenbo:
Correct, setting the constant PHP_CS_FIXER_ALLOW_XDEBUG will allow running with xdebug, I'll update the PR with the warning stripped

SpacePossum:
> unless one explicitly disables the handler at runtime somehow

if I use the command in my app I by-pass the handler, but not explicit (I use my own bootstrap and not the `/php-cs-fixer` file)

julienfalque:
I'd say that in this case your app should be responsible of showing the warning, not each single command.

SpacePossum:
fair enough :)

src/Console/WarningsDetector.php
```diff
@@ -57,13 +57,6 @@ public function detectOldVendor()
         }
     }
 
-    public function detectXdebug()
-    {
-        if (extension_loaded('xdebug')) {
-            $this->warnings[] = 'You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.';
```
keradus: I would still expect this if `PHP_CS_FIXER_ALLOW_XDEBUG` is set to be honest...

julienfalque: Because using `PHP_CS_FIXER_ALLOW_XDEBUG` means you enable Xdebug intentionally. I would at least add another environment variable that allows to disable the message.

julienfalque:
ping @AJenbo could you add some tests (using the tool as a PHAR and as a regular PHP file) please?

AJenbo:
@julienfalque not sure how to set that up but I will give it a stab.

php-cs-fixer
```diff
@@ -54,8 +54,14 @@ try {
     }
 }
 
+use Composer\XdebugHandler\XdebugHandler;
 use PhpCsFixer\Console\Application;
 
+// Restart if xdebug is loaded, unless the environment variable PHP_CS_FIXER_ALLOW_XDEBUG is set.
+$xdebug = new XdebugHandler('PHP_CS_FIXER', '--ansi');
```
keradus: just about this one,
if we provide `--ansi` on ansi-supported terminal, it pass it to our console application and it will use colours. great.

in general, `--ansi` is `<info> --ansi</info> Force ANSI output`.

while I agree that this is needed for terminals that support colours, i'm not sure how it would behave in terminal which doesn't support colours...

maybe we shall pass it only if terminal [hasColorSupport](https://github.com/symfony/console/blob/0d62adf9f4dc46a837560dee07b2b94675ddc492/Output/StreamOutput.php#L92) ?

AJenbo: XdebugHandler only passes it to the sub process if the parent supports colors. All this does is tell XdebugHandler what parameter is used by this application, not that it should apply it.

AJenbo:
Your most welcome :-)

PR #3592, opened 2018-03-10
Support PHPUnit 7

keradus:


SpacePossum:
rebased, conflict resolved and forced pushed
this should also trigger the CI again

keradus:
not sure why fabbot is asking to use long array syntax... cc @julienfalque

julienfalque:
I'm getting a blank page when trying to see fabbot reports for your latest commits.

keradus:
:|

PR #3594, opened 2018-03-10
ElseifFixer - Bug with alternative syntax

kubawerlos:
Previously the code:
`if ($x) { foo(); } else if ($y): bar(); endif;`
was changed to:
`if ($x) { foo(); } elseif ($y): bar(); endif;`
which is not correct PHP syntax.


keradus:
Thank you @kubawerlos.

PR #3596, opened 2018-03-12
Bug with duplicating header comment

kubawerlos:
Fix for https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2590 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3467.

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -125,7 +125,7 @@ echo 1;
             }
 
             $this->insertHeader($tokens, $headerNewIndex);
-        } elseif ($this->getHeaderAsComment() !== $tokens[$headerCurrentIndex]->getContent()) {
+        } elseif ($headerCurrentIndex < $headerNewIndex || $this->getHeaderAsComment() !== $tokens[$headerCurrentIndex]->getContent()) {
             $tokens->clearTokenAndMergeSurroundingWhitespace($headerCurrentIndex);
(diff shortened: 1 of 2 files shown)


julienfalque:
At first glance, this looks like a partial implementation of what @BackEndTea proposed in #3468, right? If so, do you want to go further here?

julienfalque:
ping @kubawerlos :)

kubawerlos:
It might help in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3468 yet this focuses only on HeaderCommentFixer - the header placing will need further discussion if we want also add third location.

keradus:
if it fixes 2590, shouldn't we add integration test for `declare_strict_types and header_comment` (it may be `misc` itest, not necessary `priority` one) ?

kubawerlos:
@keradus not sure about integration test as this only makes sure comment won't be duplicated. Best case I created is:
```
--TEST--
Integration of fixers: declare_strict_types,header_comment.
--RULESET--
{"declare_strict_types": true, "header_comment": {"header": "Foo header"}}
--REQUIREMENTS--
{"php": 70000}
--EXPECT--
<?php



declare(strict_types=1);

/*
 * Foo header
 */

namespace Bar;

class Baz
{
}

--INPUT--
<?php

/*
 * Foo header
 */

declare(strict_types=1);

namespace Bar;

class Baz
{
}

```

It looks like "declare_strict_types" is not used there.

PR #3597, opened 2018-03-12
DeclareStrictTypesFixer - fix bug of removing line

kubawerlos:
Fixes: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2058

Removed integration test is no longer valid as `DeclareStrictTypesFixer` won't mess with trailing new lines anymore.


src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -161,10 +161,6 @@ private function insertSequence(Tokens $tokens)
         }
 
         $content = $tokens[1 + $endIndex]->getContent();
-        if (false !== strpos($content, "\n")) {
-            return;
-        }
-
-        $tokens[1 + $endIndex] = new Token(array(T_WHITESPACE, $lineEnding.ltrim($content)));
+        $tokens[1 + $endIndex] = new Token(array(T_WHITESPACE, $lineEnding.ltrim($content, $this->whitespacesConfig->getIndent())));
```
julienfalque: Using `$this->whitespacesConfig->getIndent()` as second parameter of `ltrim()` looks weird, why not use `" \t"`?

kubawerlos: I don't know, `" \t"` seems to be better choice indeed.

keradus:
Thank you @kubawerlos.

PR #3602, opened 2018-03-14
Add new header comment location

IchHabRecht:
This pull requests adds a new location for the header_comment configuration: `'after_namespace'`. If configured, the header is inserted after the namespace declaration. In none namespace declaration is found, the default `'after_declare_strict'` configuration is used to find the correct header position.

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -160,7 +160,7 @@ echo 1;
                 ->setDefault(self::HEADER_COMMENT)
                 ->getOption(),
             (new FixerOptionBuilder('location', 'The location of the inserted header.'))
-                ->setAllowedValues(['after_open', 'after_declare_strict'])
+                ->setAllowedValues(['after_open', 'after_declare_strict', 'after_namespace'])
                 ->setDefault('after_declare_strict')
                 ->getOption(),
             (new FixerOptionBuilder('separate', 'Whether the header should be separated from the file content with a new line.'))
@@ -221,6 +221,23 @@ echo 1;
             return 1;
         }
 
+        if ('after_namespace' === $this->configuration['location']) {
+            return $this->findHeaderCommentInsertionIndexForNamespace($tokens);
+        }
+
+        return $this->findHeaderCommentInsertionIndexForDeclareStrict($tokens);
+    }
+
+    /**
+     * Find the index where the header comment must be inserted if location is after_declare_strict.
+     *
+     * @param Tokens $tokens
+     *
+     * @return int
+     */
+    private function findHeaderCommentInsertionIndexForDeclareStrict(Tokens $tokens)
+    {
+        $index = $tokens->getNextMeaningfulToken(0);
         if (!$tokens[$index]->isGivenKind(T_DECLARE)) {
             return 1;
         }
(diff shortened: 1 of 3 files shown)


SpacePossum:
Hi @IchHabRecht and thanks for the proposal :)
I like the idea and hope to review soon.

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea/
```
ntzm: This should be placed in your global gitignore :)

IchHabRecht: Done :-)

src/Fixer/Comment/HeaderCommentFixer.php
```diff
@@ -258,6 +275,36 @@ private function findHeaderCommentInsertionIndex(Tokens $tokens)
...
+        }
+
+        if (null === $index) {
+            return $this->findHeaderCommentInsertionIndexForDeclareStrict($tokens);
+        }
+
+        $next = $tokens->getNextMeaningfulToken($index);
+        while (null !== $next && !$tokens[$next]->equals(';')) {
+            $next = $tokens->getNextMeaningfulToken($next);
+        }
```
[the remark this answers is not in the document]
IchHabRecht: No, it would insert the header above the namespace. I don't think there is a good solution where to insert the header. As the declaration would have to insert it *after* the namespace (thus after closing curly bracket) but that IMHO would not make any sense. I guess the expectation would be to insert the header *inside* but that's not covered by the configuration.

kubawerlos:
This fixer brings third place for header comment - which makes tests much more complicated.

Lets consider having following template:
```php
<?php
/**
 * {pos1}
 */
declare(strict_types=1);
/**
 * {pos2}
 */
namespace Foo;
/**
 * {pos3}
 */
class Bar
{
}
```

What if `pos1` = `pos3` = `"Hello World"` and `pos2` = `"Hello Universe"` and config is `['header' => 'Hello Milky Way', 'location' => 'after_open']`? Do we remove `pos3` as it's the same as the one we will update? Maybe `pos2` as well? Or only replace the first one?

My first solution was to refactor the current fixer - so it locates all possible places for the comment, updates the configured one and removes others. But what if the comment at `pos2` is meaningful and someone want to update only `pos1`?

I'd say let's start with bunch of tests to cover all cases.

IchHabRecht:
To be honest, defining what to do with existing comments is not the scope of this patch. Of course that should be done, but IHMO is not defined yet nor working correctly (currently I get two time the same header if I switch between both existing options).

@kubawerlos in your example pos3 and the new header aren't the same and I would advice not to remove anything.

kubawerlos:
What if they are the same? Then remove it?

What I meant is having 2 locations we have issues like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2590 and https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3467, so adding third location without properly handling two is not the best way.

ntzm:
Would be good to get some tests in here

ntzm:
@IchHabRecht Any chance you have the time to add some tests here?

Wirone:
Thank you @IchHabRecht for your work, but I am going to close this because:
- it's massively outdated (2018!) and requires solid rebase
- it does not have tests
- there wasn't any interest in this feature for years 🤷‍♂️

Issue #2480, opened 2017-01-26
no_unused_imports no working correct


DjThossi:
Hey in our project we got some trouble with newest version (2.0.0) of php-cs-fixer and no_unused_imports rule. 

If the name of an imported class (e.g. Design) is anywhere else used (e.g. $this->design or //design) the import is not removed even if it should be removed.

PHP version we are using:
```
$ php -v
PHP 5.6.29-1~dotdeb+7.1 (cli) (built: Dec  9 2016 16:30:46) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.4.0, Copyright (c) 2002-2016, by Derick Rethans
```

PHP CS Fixer version we are using:
```
$ php-cs-fixer -V
PHP CS Fixer version 2.0.0 by Fabien Potencier and Dariusz Ruminski
```

The command we use to run the fixer 
```
$ bin/php-cs-fixer fix -v
```

The configuration file we are using
```
<?php

return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setRules(array(
        'no_unused_imports' => true,
    ))
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->in(__DIR__ . '/src')
    );
```

A minimal sample of valid PHP code that is not fixed correctly
```
<?php
namespace App\Example;

use App\Test1;
use App\Test2;
use App\Test3;

class Example
{
    protected function getExample($config)
    {
        //test1
        'test2';
        $this->test3;
    }
}
```

The fixed version of that code after you run the fixer and the version you expected.
```
<?php
namespace App\Example;

class Example
{
    protected function getExample($config)
    {
        //test1
        'test2';
        $this->test3;
    }
}
```

GrahamCampbell:
Are you saying that the test1 import is not removed? If so, that's the expected behaviour, to allow people to reference class names from comments.

SpacePossum:
>Are you saying that the test1 import is not removed?

It seems all imports are left intact.
I agree with @GrahamCampbell that `test1` should not be removed. The imports `test2` and `test3` could be removed. IMO it is not a bug, but a lacking feature (i.e. missed possibility). Love to see a PR for this!

keradus:
Closed as discussed.

[...]

PR #3445, opened 2018-01-14
Rewrite NoUnusedImportsFixer

julienfalque:
Includes #3289, which should be merged first.

Fixes #2480 and partially addresses #2814 as a bugfix. New options are still to be implemented in another PR so the issue should not be closed.

I tested it on Symfony repo (2.7 and master). It only removed unused `use` statements that were not detected before, no actually used `use` statement were removed.

PR #3603, opened 2018-03-15
Fix no unused imports fixer

IchHabRecht:
This bugfix prevents finding false position usages of use statements that are quoted (either single or double quotes).

Maybe related to #2480


julienfalque:
Hi, thanks for taking time to work on this issue but this is already covered by #3445, which drops regular expression usage to prevent such false positive. I'm closing this one as duplicate.

PR #3607, opened 2018-03-16
PhpdocVarWithoutNameFixer - update sample with @ type

SpacePossum:
PhpdocVarWithoutNameFixer fixes `@type` as well, however this was not part of the sample nor the unit tests


keradus:
Thank you @SpacePossum.

PR #3610, opened 2018-03-17
make phar extension optional

ktomk:
the utility itself does not need the phar extension if used in the source
version but it previously crashed when invoked w/o the phar extension b/c
there was an unchecked use while trying to map the phar.

fix is to check for the phar class and only if it exists try with phar
mapping. otherwise just take the source version as it is not a phar
anyway.

### DIFF
--- a/php-cs-fixer
+++ b/php-cs-fixer
@@ -41,11 +41,17 @@ set_error_handler(function ($severity, $message, $file, $line) {
     }
 });
 
-try {
+$require = true;
+if (class_exists('Phar')) {
     // Maybe this file is used as phar-stub? Let's try!
-    Phar::mapPhar('php-cs-fixer.phar');
-    require_once 'phar://php-cs-fixer.phar/vendor/autoload.php';
-} catch (PharException $e) {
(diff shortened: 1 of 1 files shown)


keradus:
hi @ktomk ! Your reasoning sounds damn valid if you ask me!
I see that some jobs started to fail at Travis, but that's probably due to travis update of meta-versions like `nightly`, not sth specific to this PR

keradus:
did you actually got any issue on your env about lack of Phar ext? I thought it's standard already

ktomk:
Your mileage may vary, but what I did to speed things up was invoking PHP with the `-n` switch which loads no extensions at all. So I then needed to add one extension after the other on the command line and with this small edit as phar is not needed to run it.

In the end I did this to not load xdebug for speed reasons, just in case you're wondering.

keradus:
oooh, `-n` was the reason, cool to know ;)
yeah, i removed xdebug totally, never using it after phpdbg was born ;)

keradus:
Thank you @ktomk.

PR #3615, opened 2018-03-20
Detect fabbot.io

julienfalque:
Ref: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2935#issuecomment-373520493.

Thank you very much @fabpot for adding the environment variable :)

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -43,15 +43,17 @@ $config = PhpCsFixer\Config::create()
 ;
 
 // special handling of fabbot.io service if it's using too old PHP CS Fixer version
-try {
-    PhpCsFixer\FixerFactory::create()
-        ->registerBuiltInFixers()
-        ->registerCustomFixers($config->getCustomFixers())
-        ->useRuleSet(new PhpCsFixer\RuleSet($config->getRules()));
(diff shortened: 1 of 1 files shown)


.php_cs.dist
```diff
@@ -43,15 +43,17 @@ $config = PhpCsFixer\Config::create()
...
-try {
-    PhpCsFixer\FixerFactory::create()
-        ->registerBuiltInFixers()
-        ->registerCustomFixers($config->getCustomFixers())
-        ->useRuleSet(new PhpCsFixer\RuleSet($config->getRules()));
-} catch (PhpCsFixer\ConfigurationException\InvalidConfigurationException $e) {
-    $config->setRules(array());
-} catch (UnexpectedValueException $e) {
-    $config->setRules(array());
+if ('1' === getenv('FABBOT_IO')) {
```
keradus: we don't control what is the value here, if it's 1, maybe it's like true/false, maybe it's version and it will be 2, 3, 4 in the future, we don't know, we don't control. what we care... is that env var is exposed or not.
imo, we shall drop `'1' ===` part

julienfalque: Indeed, thanks for making the change :)

keradus:
Thank you @julienfalque.

PR #3617, opened 2018-03-20
Tests stability patches

ktomk:
Diverse patches, mainly on test expectations, to make the suite green after checkout.

Additionally a fix removing a caching iterator as it behaved unstable in the fix linting iteration.

### DIFF
--- a/src/Runner/FileLintingIterator.php
+++ b/src/Runner/FileLintingIterator.php
@@ -20,7 +20,7 @@ use PhpCsFixer\Linter\LintingResultInterface;
  *
  * @internal
  */
-final class FileLintingIterator extends \CachingIterator
+final class FileLintingIterator extends \IteratorIterator
 {
     /**
      * @var LintingResultInterface
--- a/src/Runner/FileLintingIterator.php
+++ b/src/Runner/FileLintingIterator.php
@@ -20,7 +20,7 @@ use PhpCsFixer\Linter\LintingResultInterface;
  *
  * @internal
  */
-final class FileLintingIterator extends \IteratorIterator
+final class FileLintingIterator extends \CachingIterator
 {
     /**
      * @var LintingResultInterface
--- a/tests/Fixer/Comment/HeaderCommentFixerTest.php
+++ b/tests/Fixer/Comment/HeaderCommentFixerTest.php
@@ -516,8 +516,8 @@ declare(strict_types=1)?>',
             array("<?php\nphpinfo();\n?><hr/>"),
             array("  <?php\n"),
             array('<?= 1?>'),
-            array('<?= 1?><?php'),
-            array("<?= 1?>\n<?php"),
+            array("<?= 1?><?php\n"),
+            array("<?= 1?>\n<?php\n"),
         );
     }
 
--- a/tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
+++ b/tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
@@ -548,16 +548,6 @@ final class NoEmptyStatementFixerTest extends AbstractFixerTestCase
                 );
         }
 
-        // HHVM parses '<?=' as T_ECHO instead of T_OPEN_TAG_WITH_ECHO
-        // test the fixer doesn't break anything
-        if (ini_get('short_open_tag') && !defined('HHVM_VERSION')) {
-            $cases[] =
-                array(
-                    '<?= ',
-                    '<?= ;',
-                );
-        }
-
         if (count($cases) < 1) {
             $this->markTestSkipped('No short tag tests possible.');
         }
--- a/tests/Fixer/Whitespace/NoWhitespaceInBlankLineFixerTest.php
+++ b/tests/Fixer/Whitespace/NoWhitespaceInBlankLineFixerTest.php
@@ -39,7 +39,7 @@ final class NoWhitespaceInBlankLineFixerTest extends AbstractFixerTestCase
     {
         return array(
             array(
-                '<?php',
+                "<?php\n",
             ),
             array(
                 '<?php  ',


ktomk:
If fine with this **LinterInterface::isAsync()** has proven superfluous as well which actually kind of makes sense when chewing this a bit. Why as for type differences via the common interface (async or not async linter on the same linter interface). Your mileage may vary, just saying. Let me know if we can change that interface, I would recommend to remove that method from it.

src/Runner/Runner.php
```diff
@@ -124,9 +124,10 @@ public function fix()
             $this->cacheManager
         );
 
-        $collection = $this->linter->isAsync()
-            ? new FileCachingLintingIterator($fileFilteredFileIterator, $this->linter)
-            : new FileLintingIterator($fileFilteredFileIterator, $this->linter);
+        $collection = new FileLintingIterator(
+            new \CachingIterator($fileFilteredFileIterator),
```
keradus: could you please describe what is a benefit of using `CachingIterator` in code you proposed ?

tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
```diff
@@ -548,16 +548,6 @@ public function provideWithShortOpenTagCases()
                 );
         }
 
-        // HHVM parses '<?=' as T_ECHO instead of T_OPEN_TAG_WITH_ECHO
```
keradus: did hhvm fixed the bug ?

keradus:
> Additionally a fix removing a caching iterator as it behaved unstable in the fix linting iteration.

elaborate what is unstable, show logs, etc...

src/Runner/Runner.php
ktomk: The CachingIterator was used before (just via extended and even somewhat re-implemented which I considered superfluous so I removed that one and replaced it with this one).

tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
ktomk: Please see the linked comment for that change in the commit message. I removed it due to get the suite green. I can not say whether or not they fixed the bug.

src/Runner/Runner.php
keradus: so, you didn't answer my question.
what is the benefit of using `CachingIterator` in code you are proposing?

I will answer you - none.

That "superfluous" class you have removed was responsible for running async linter on 2nd core, allowing one core lint file A and other core to fix file B.
With your changes, you drop the parallelism. Sometimes, it's better to understand before dropping the code.

ktomk: @keradus: Core as processor cores? I was indeed not aware of that. Caching both do actually help to have the same behaviour. I'll update the PR.

tests/Fixer/Semicolon/NoEmptyStatementFixerTest.php
SpacePossum: @keradus 

the test case is invalid on HHVM and PHP, we just never saw the issue because the tests never run on Travis and the majority of the contributors has `short_open_tag` disabled

```
1) PhpCsFixer\Tests\Fixer\Semicolon\NoEmptyStatementFixerTest::testCasesWithShortOpenTag with data set #1 ('<?= ', '<?= ;')
Failed asserting that 'PHP Parse error: syntax error, unexpected ';' on line 1.\n
\n
Source:\n
<?= ;' is null.
```

```
[12:08 PM]-[possum@zoo1]-[/home/possum/work/PHP-CS-Fixer]-[git master] 
$ php -v
PHP 7.2.3-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Mar  6 2018 11:18:25) ( NTS )
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.3-1+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies

[12:08 PM]-[possum@zoo1]-[/home/possum/work/PHP-CS-Fixer]-[git master] 
$ cat /etc/php/7.2/cli/php.ini|grep short_open_tag
151:; short_open_tag
197:short_open_tag = On
```

tests/Fixer/Comment/HeaderCommentFixerTest.php
```diff
@@ -516,8 +516,8 @@ public function provideDoNotTouchCases()
             array("<?php\nphpinfo();\n?><hr/>"),
             array("  <?php\n"),
             array('<?= 1?>'),
-            array('<?= 1?><?php'),
-            array("<?= 1?>\n<?php"),
+            array("<?= 1?><?php\n"),
+            array("<?= 1?>\n<?php\n"),
```
SpacePossum: tested, both are good fixes :+1:

tests/Fixer/Whitespace/NoWhitespaceInBlankLineFixerTest.php
```diff
@@ -39,7 +39,7 @@ public function provideFixCases()
     {
         return array(
             array(
-                '<?php',
+                "<?php\n",
```
SpacePossum: :+1:

keradus:
@ktomk , revert last commit, then we could merge

src/Runner/Runner.php
keradus: yes, cores as cpu cores.
I think you don't understand what `Caching` in `CachingIterator` stands for.
Using CachingIterator for non-async linter will only increase memory usage, without giving any benefit

ktomk:
@keradus: Indeed I was throwing out the baby w/ the bathwater ... for the reasons you wrote. I must admit it was not entirely clear for me what that caching was for (as I've been written yesterday already). Anyway the update is far less invasive. Let's see if Travis likes it. Local tests for me w/ PHP 5.6 and 7.2 were fine again.

Do you have interest to batch up more than two files within iteration? E.g. four parallel processes in linting? currently my setups I/we normally use  eight parallel tasks like so:

~~~
find ./*.php bin application library/what-not tests -xdev -type f \
  \( -name '*.php' -o -name '*.phtml' \) -print0 \
  | 2>/dev/null xargs -0 -n1 -P8 php -d short_open_tag=0 -l \
  | grep -v '^No syntax errors detected'

test $? -eq 1
~~~

Just asking I hacked together one batching iterator recently (w/ tests) that fills up a batch of X iterations from the decorated iterator and then iterates over the batch. Would just need another decorator that does the linting (or a re-use of the non-caching one). Then the number of parallel lints is configurable.

keradus:
running the linter on second core and running whole app in parallel calls are to different things. for later, simply use `parallel` (GNU)

src/Runner/FileLintingIterator.php
```diff
@@ -20,7 +20,7 @@
  *
  * @internal
  */
-final class FileLintingIterator extends \IteratorIterator
+final class FileLintingIterator extends \CachingIterator
```
keradus: so, let me ask same question here as well.
could you please describe what is a benefit of using `CachingIterator` in code you proposed ?

ktomk: Both iterators then provide the same iteration. w/o that one, the FileLinitingIterator has a different order than the other one. I can not say 100% for sure why that is, but I debugged it the last three days and this change does make that integration test green w/ PHP 5.6 (process linter) and PHP 7.2 (the other linter). Sorry for not being able to share more on this, it's just I can't say. But feel free to ask more, I've looked in there extensively the last days.

keradus: order of items in iteration is not stable regardless which iterator you gonna use. basically when we read the files from drive, we could get different order each time.
I can see no benefit of this change, but I see CPU power wasted. Please revert it.

ktomk: Both of the decorating iterators are feed w/ the same order of files. So for a reason the caching iterator changes the order of current elements but I do not understand why that happens. Do you have any idea? It's not the order of files returned from disk, it is the same as far I've seen.

keradus: order of iterating is not defined as stable anywhere, and it doesn't need to be stable tbh.

again, to have this PR merged, revert this change.
if you want this change to happen, send it as separated PR with dedicated utest that fails without change and works with it.

ktomk: Yes, we can separate this problem, I'll refresh the PR. Just wondering, what is a utest? Alternatively the fixture could be changed to have the order undefined. Should work as well (will make the test a little bit more convoluted, but likely not much compared to its current state).

keradus: utest - unit test
fixture and undefined order - it shall be done indeed, order of reading files from drive is not defined, as we don't sort files after read from drive.

SpacePossum:
as this PR is now (with the revert done by @keradus ) , RTM from me :+1:

keradus:
Thank you @ktomk.

PR #3618, opened 2018-03-20
DateTimeImmutableFixer - grooming

keradus:
appendix to #3260

### DIFF
--- a/src/Fixer/ClassUsage/DateTimeImmutableFixer.php
+++ b/src/Fixer/ClassUsage/DateTimeImmutableFixer.php
@@ -128,8 +128,7 @@ final class DateTimeImmutableFixer extends AbstractFixer implements FixerInterfa
         }
 
         if ($isUsedWithLeadingBackslash || $isUsedAlone && ($isInNamespace && $isImported || !$isInNamespace)) {
-            $tokens->clearAt($index);
-            $tokens->insertAt($index, new Token([T_STRING, 'DateTimeImmutable']));
+            $tokens[$index] = new Token([T_STRING, 'DateTimeImmutable']);
             if ($isInNamespace && $isUsedAlone) {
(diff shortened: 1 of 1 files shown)


src/Fixer/ClassUsage/DateTimeImmutableFixer.php
```diff
@@ -128,8 +128,7 @@ private function fixClassUpdate(Tokens $tokens, $index, $isInNamespace, $isImpor
         }
 
         if ($isUsedWithLeadingBackslash || $isUsedAlone && ($isInNamespace && $isImported || !$isInNamespace)) {
-            $tokens->clearAt($index);
-            $tokens->insertAt($index, new Token([T_STRING, 'DateTimeImmutable']));
+            $tokens[$index] = new Token([T_STRING, 'DateTimeImmutable']);
             if ($isInNamespace && $isUsedAlone) {
                 $tokens->insertAt($index, new Token([T_NS_SEPARATOR, '\\']));
```
keradus: @kubawerlos 
with this line, we are adding new items to collection during iterating over it.
for that, we shall either:
- iterate backward
- adjust iteration limit
- don't inject this token, but replace import - is it possible, or in some cases class usage is not replaced?

kubawerlos: @keradus your changes are what it suppose to be, I cannot say why I did `clearAt`/`insertAt`.

keradus: I know my cleanup is good ;)
yet, issue i'm pointing to is doing `insertAt` while iterating forward in main loop

kubawerlos: Ok, got it now: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3632

You can merge this PR.

PR #3624, opened 2018-03-21
ArrayIndentationFixer: missing handling of case when value starts on new line

Slamdunk:
Ping @julienfalque (ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3135)

```diff
--- Expected
+++ Actual
@@ @@
 '<?php
 $foo = [
     'foo' =>
-        'Value'
-        . ' on'
-        . ' new-line',
+    'Value'
+    . ' on'
+    . ' new-line',
 ];'
```

### DIFF
--- a/tests/Fixer/Whitespace/ArrayIndentationFixerTest.php
+++ b/tests/Fixer/Whitespace/ArrayIndentationFixerTest.php
@@ -220,6 +220,42 @@ INPUT
                 <<<'EXPECTED'
 <?php
 $foo = [
+    'foo' =>
+        'My'
+        . ' very'
+        . ' long'
+        . ' string',
+    'bar' =>
+        'My'
+        . ' second'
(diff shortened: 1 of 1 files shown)


julienfalque:
Yep, this is already known (https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3135#issuecomment-370251632). We decided to postpone supporting this case to be able to merge the PR. Also this is a missing feature rather than a bug IMO.

I do want to work on this but I think it might be tricky so I probably won't take it soon, if anyone wants to give it a try in the meantine, that would be great :)

Slamdunk:
Ok.
For the records I will not take the try for now: anyone keen feel free to pull this PR as a patch and start a new PR with the solution.

julienfalque:
Can be replaced by #4654.

PR #3627, opened 2018-03-21
Fix tests execution under phpdbg

keradus:
collecting coverage is failing the build...
even when we already have annotation to not collect coverage for test that is failing...


keradus:
For some reason that I was unable to hunt down, execution of `proc_open` (execution of command, like we do for smoke tests, for example) under `phpdbg` is never-ending process, for which we hit timeout exception. I tried to investigate is it related to PHP ver/listeners, but apparently it is not. I'm unable to find why it started failing and was passing week ago.
I was thinking about solution when problematic tests are grouped and executed separately, which actually works. Then I extended it to have a group not only for manually pointed problematic tests, but to all tests that collects no coverage.

Could you give opinion, @julienfalque , @SpacePossum , @localheinz , please?

PR #3623, opened 2018-03-20
Fix: Keep rules sorted

localheinz:
This PR

* [x] keeps rules sorted in `.php_cs.dist`

PR #3629, opened 2018-03-21
ProjectFixerConfigurationTest - test rules are sorted

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3623

### DIFF
--- a/tests/AutoReview/ProjectFixerConfigurationTest.php
+++ b/tests/AutoReview/ProjectFixerConfigurationTest.php
@@ -34,14 +34,13 @@ final class ProjectFixerConfigurationTest extends TestCase
 
     protected function setUp()
     {
-        $file = __DIR__.'/../../.php_cs.dist';
-        $this->assertFileExists($file);
+        $file = __DIR__.'/../../.php_cs.dist';       
         $this->config = require $file;
-        $this->assertInstanceOf(Config::class, $this->config);
     }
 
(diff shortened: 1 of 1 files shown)


tests/AutoReview/ProjectFixerConfigurationTest.php
```diff
@@ -26,22 +27,40 @@
...
+     * @var \PhpCsFixer\Config
+     */
+    private $config;
+
+    protected function setUp()
     {
-        /** @var \PhpCsFixer\Config $config */
-        $config = require __DIR__.'/../../.php_cs.dist';
+        $file = __DIR__.'/../../.php_cs.dist';
+        $this->assertFileExists($file);
```
keradus: assertions in set up? it doesn't look good

SpacePossum: not sure which way to refactor, any preference?

keradus: replace whole body with
```php
$this->config = require __DIR__.'/../../.php_cs.dist';
```

if it's not the file, it will crash the `require` call, no need for assertion
if it won't return Config, usage of `$this->config` will crash, no need for assertion

keradus:
Thank you @SpacePossum.

PR #3632, opened 2018-03-21
DateTimeImmutableFixer bug with adding tokens while iterating over them

kubawerlos:


keradus:
Thank you @kubawerlos.

PR #3633, opened 2018-03-22
Moving SameStringsConstraint to autoload files

tomas7777:
Moving `SameStringsConstraint.php` file to autoload files, since it causes tests to fail.
https://getcomposer.org/doc/04-schema.md#classmap 
```
This map is built by scanning for **classes** in all .php and .inc files in the given directories/files.
```
Since there aren't any classes defined in the file, making custom fixers and trying to run tests causes them to fail.


keradus:
Sorry for late answer, but I wanted to double check this PR, and that was a good call.

First of all, I do agree that we have fucked up "dev" related classes exposed in `src`, those classes are needed only for development purposes like yours, and sadly those classes needs PHPUnit as dependency. Yet, for regular user, he doesn't need PHPUnit at all to run PHP CS Fixer. For that, it's not a requirement in our `composer.json` file (only dev-req). That means it is a hidden dependency for devs like you, who extend the ecosystem. Ultimate solution is to create dedicated package with all dev-related files.

Sadly, your proposal will break the installation for users who don't need PHPUnit:
(note: I checked out to your changes before executing the following test)

```
ker@dus:/tmp/xxx λ cat composer.json 
{
    "repositories": [
        {
            "type": "path",
            "url": "~/github/PHP-CS-Fixer"
        }
    ],
    "require": {
        "friendsofphp/php-cs-fixer": "*@dev"
    },
    "config": {
        "optimize-autoloader": true
    }
}
ker@dus:/tmp/xxx λ composer install -q
ker@dus:/tmp/xxx λ php vendor/autoload.php 
PHP Warning:  Class 'PHPUnit_Runner_Version' not found in /home/keradus/github/PHP-CS-Fixer/tests/Test/Constraint/SameStringsConstraint.php on line 16
PHP Fatal error:  Uncaught Error: Class 'PHPUnit\Runner\Version' not found in /home/keradus/github/PHP-CS-Fixer/tests/Test/Constraint/SameStringsConstraint.php:19
Stack trace:
#0 /tmp/xxx/vendor/composer/autoload_real.php(66): require()
#1 /tmp/xxx/vendor/composer/autoload_real.php(56): composerRequireb4c3e05ccb886ff01a06ba144e3efcb5('fb17371873febc4...', '/tmp/xxx/vendor...')
#2 /tmp/xxx/vendor/autoload.php(7): ComposerAutoloaderInitb4c3e05ccb886ff01a06ba144e3efcb5::getLoader()
#3 {main}
  thrown in /home/keradus/github/PHP-CS-Fixer/tests/Test/Constraint/SameStringsConstraint.php on line 19
```

keradus:
could you take a look? #3645

tomas7777:
#3645 seems to have fixed the issues.

keradus:
for that, I'm closing this PR. Thanks for bringing the topic and verify the fix

PR #2207, opened 2016-09-25
Added ReturnTypeDeclarationFixer

keradus:
Support for `Return Type Declarations` (part of #1119)

Issue #3558, opened 2018-02-21
MethodChainingIndentationFixer: fix call contents


Slamdunk:
For this snippet:

```php
$foo = $bar
            ->baz([
                1 => 2,
            ])
->xyz()
;
```

following https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3528 currently `MethodChainingIndentationFixer` does:

```diff
 $foo = $bar
-            ->baz([
+    ->baz([
                 1 => 2,
             ])
-->xyz()
+    ->xyz()
 ;
```
What I would like is:

```diff
 $foo = $bar
-            ->baz([
-                1 => 2,
-            ])
+    ->baz([
+        1 => 2,
+    ])
-->xyz()
+    ->xyz()
 ;
```

julienfalque:
:+1: 

It should also do this:

```diff
 $foo
-        ->bar([
-                1 => 2,
-            ])
-                ->baz()
+    ->bar([
+            1 => 2,
+        ])
+    ->baz()
;
```

i.e. the indentation of the array should be adjusted, not fixed. I did the same in #3135.

Slamdunk:
> i.e. the indentation of the array should be adjusted, not fixed.

I agree, but what would you do on this?

```php
$foo
    ->bar()            // ok
        ->baz([        // will go back by 4 spaces
            1 => 2     // will go back by 4 spaces
])                     // how to handle this?
;
```

julienfalque:
We could either:
* not change the indentation of the array because, as a whole, it already has the minimum indentation (so moving the `->baz([` row only)
* unindent all possibles lines (like most IDEs would do when selecting the array and pressing `Shift+Tab`), the array would be actually changed, not ajusted, but the indentation is already wrong anyway

I personally prefer the second choice.

Slamdunk:
> I personally prefer the second choice.

Me too

Slamdunk:
@julienfalque the `has PR` label should be removed at the time of writing

julienfalque:
Indeed, thanks :)

Wirone:
Closing, because as of now Fixer 3.16 does exactly what is expected

<img width="896" alt="image" src="https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/assets/600668/7da90a59-6463-4e11-bbc6-4c6f182d56db">

I believe indentation of an array is better like this, which is like in @Slamdunk's initial report.

PR #3640, opened 2018-03-28
Add PhpdocToReturnTypeFixer

Slamdunk:
```diff
 /**
  * @return \My\Bar
  */
-function my_foo()
+function my_foo(): \My\Bar
 {}
```

- [x] Mark as *risky*
- [x] Require ^7.0
- [x] Fix ^7.1 [*void return*, *iterable*]
- [x] Fix ^7.2 [*object type*]
- [x] Fix ^7.1 [*nullable types*]
- [x] Config to fix scalar types, default to **false** because of unexpected results if *stryct_types* are enabled or not
- [x] `@return static` => `: self`
- [x] ~Config to overwrite existing mismatching types: PHPDoc wins~ can't do this, because `FunctionNotation\ReturnTypeDeclarationFixer` changes the output of this fixer
- [x] Ignore mixed types
- [x] Ignore [generics](https://wiki.php.net/rfc/generics) and array of types
- [x] Run before `FunctionNotation\ReturnTypeDeclarationFixer` #2207
- [x] Run before `Import\FullyQualifiedStrictTypesFixer` #3276 
- [x] Run ~before~ integration test with `FunctionNotation\VoidReturnFixer` #2740
- [x] Run after `Phpdoc\PhpdocScalarFixer`

Suggestions for the name of the fixer are welcome.


keradus:
> don't need this because FunctionNotation\VoidReturnFixer skips if return annotation is present

could you please add `misc` integrity test between them, so we won't hit the priority issue if one day one would change `VoidReturnFixer` to not skip on annotation ?

Slamdunk:
> could you please add misc integrity test between them, so we won't hit the priority issue if one day one would change VoidReturnFixer to not skip on annotation ?

Didn't know about `misc`, wrote that comment because IntegrationTest failed.

Added.

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
```diff
@@ -0,0 +1,307 @@
...
+
+            if (1 !== Preg::match($classRegex, $returnType)) {
+                continue;
+            }
+
+            $startIndex = $tokens->getNextTokenOfKind($index, ['{', ';']);
+
+            if ($this->hasReturnTypeHint($tokens, $startIndex)) {
+                continue;
+            }
```
Slamdunk: For the records, segmented `if () { continue; }` are intended: force a clear mapping between feature/bug and code-coverage.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,200 @@
...
+namespace PhpCsFixer\Tests\Fixer\FunctionNotation;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ *
+ * @internal
+ *
+ * @requires PHP 7.0
```
julienfalque: Shouldn't the tests run on PHP 5.6 as well and assert that nothing is changed?

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,200 @@
...
+            'nullable native type' => [
+                '<?php /** @return null|array */ function my_foo(): ?array {}',
+                '<?php /** @return null|array */ function my_foo() {}',
+                70100,
+            ],
+            'skip mixed nullable types' => [
+                '<?php /** @return null|Foo|Bar */ function my_foo() {}',
+            ],
+            'skip generics' => [
+                '<?php /** @return array<int, bool> */ function my_foo() {}',
```
julienfalque: Why ignoring those? An `array` return type would be valid wouldn't it?

README.rst
```diff
@@ -1293,6 +1293,18 @@ Choose from the list of available rules:
...
+* **phpdoc_to_return_type**
+
+  Takes ``@return`` annotation of non-mixed types and adjusts accordingly
+  the function signature. Requires PHP >= 7.0.
+
+  *Risky rule: modifies the signature of functions.*
+
+  Configuration options:
+
+  - ``scalar_types`` (``bool``): fix also scalar types; may have unexpected
```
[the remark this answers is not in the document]
Slamdunk: My reasons not to do what you describe:

1. https://3v4l.org/dicJv && https://3v4l.org/0nRCb
1. https://twitter.com/Ocramius/status/978204573801250816

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
```diff
@@ -0,0 +1,307 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        // These cause syntax errors.
+        static $blacklistFuncNames = [
```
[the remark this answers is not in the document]
Slamdunk: May I ask you why? Local scope is easier to maintain.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: Indeed it would be.
I'm just not confortable enough supporting something that isn't widely accepted.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,200 @@
...
+            ],
+            'skip array of types' => [
+                '<?php /** @return Foo[] */ function my_foo() {}',
```
Slamdunk: For those asking why not supporting `Foo[]` after reading https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3640#discussion_r177971967 it's because it could also be an `Iterator`, so we can't be sure `[]` means `array`

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: I've done like [VoidReturnFixerTest](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/tests/Fixer/FunctionNotation/VoidReturnFixerTest.php#L22) running the tests against 5.6 would only hit the `isCandidate` method:

https://github.com/Slamdunk/PHP-CS-Fixer/blob/58f511b4a1a56aa27379d31bdc303029fca5ae50/src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php#L78-L81

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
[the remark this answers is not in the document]
julienfalque: I think they can easily create memory leaks as the variable is not removed when instances of the class are destroyed.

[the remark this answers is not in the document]
Slamdunk: Done.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
julienfalque: Generics syntax is not widely used indeed, but `Foo[]` syntax on the other hand is very common, I think it would be nice to at least support this one.

Slamdunk: see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3640#discussion_r177973321

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
julienfalque: I think `int[]` syntax only means array of integers. [phpDocumentor's doc](https://docs.phpdoc.org/guides/types.html#arrays) and [PSR-5](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#abnf) seem to say so.

IMO if a function can return an array or an iterator, it should have something like e.g. `@return iterable` or `@return Foo[]|\Traversable`.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
julienfalque: IMO `VoidReturnFixerTest` is not a good example. Current tests do not prove that the fixer does nothing on PHP 5.6. I would still remove the `@requires` annotation and either assert nothing is changed in any test case when running on PHP 5.6 or add specific tests for `isCandidate()`.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: `array of types` support added.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: Test onto PHP < 7.0 added.

README.rst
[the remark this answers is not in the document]
julienfalque: Yeah, already got similar issues. IMO the real issue is not using scalar type hints but using them *without* `declare(strict_types=1);`.

julienfalque:
It would be nice to do the same with parameter typehints. Do we want this fixer to handle those too? Or do we want a new fixer?

Slamdunk:
> It would be nice to do the same with parameter typehints.

I skipped parameter typehints because PHP had them long time before return types, even for arrays.

I suppose that if a programmer wrote a good PHPDoc with object type hint, he/she also wrote it explicitely.
If so, the only [**userful**] types involved in such fixer would be:

1. scalar ones
1. callback, iterable, Closure (introduced recently)

For scalar ones, hell no please https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3640#discussion_r177972761.
For new ones, ok but it would add huge complexity for such little cases.

> Do we want this fixer to handle those too? Or do we want a new fixer?

With above consideration, it would be a new fixer for me.

julienfalque:
I still don't see the problem with scalar typehints (with strict types enabled).

Slamdunk:
> I still don't see the problem with scalar typehints (with strict types enabled).

A lot of database abstraction layers don't strictly respect typing, or simply can't for example when returning an integer greater than 2^64-1.
I have experience on a lot of legacy apps where int/float typing is just a waste of time.

julienfalque:
That doesn't mean no project would benefit from using them. You could still disable the fixer on those legacy projects. Anyway, do you think the code needed to implement such fixer would be very different from this one?

Slamdunk:
Honestly I don't know.

It is sad to say this, but in the current year I'm not able anymore to develop FOSS for purposes that aren't useful for me. I still miss #3553, #3558, #3624, #3644, #3648 :(

But I'm open to merge any PR against my PR-branch.

julienfalque:
No worries, I'm not asking you to implement parameters typehints support here :) I just want to know whether implementing it in the same fixer would be interesting (maybe as a new option later), in which case it would be better to find a more generic name before merging your PR.

Slamdunk:
> it would be better to find a more generic name before merging your PR.

I would prefer a different fixer creating a shared helper for both.

julienfalque:
@SpacePossum @keradus WDYT? Should we anticipate a broader scope for this fixer or assume it will be a new one? And what about https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3640#discussion_r177965593?

Beside these questions, the PR is ready IMO.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,211 @@
...
+            ],
+            'skip generics' => [
+                '<?php /** @return array<int, bool> */ function my_foo() {}',
+            ],
+            'array of types' => [
+                '<?php /** @return Foo[] */ function my_foo(): array {}',
+                '<?php /** @return Foo[] */ function my_foo() {}',
+            ],
+            'skip array of array of types' => [
+                '<?php /** @return Foo[][] */ function my_foo() {}',
```
julienfalque: Why skipping these?

Slamdunk: To be aligned with https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.11.1/src/DocBlock/Annotation.php#L30

julienfalque: Mmh I think it's an oversight rather than an explicit _the `Foo[][]` syntax is not supported_, I'd say this should still be supported.

Slamdunk: Another reason for me is to be protective: `Foo[][]` is rare, and why would a sane OOP architecture return an array of arrays?

I would like to keep the regex as simple as possible.

SpacePossum: what is `Foo[][]` ? `array<array<Foo>>` ?

Slamdunk: It could be

README.rst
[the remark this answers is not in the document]
SpacePossum: I'm happy with both default `on` or `off`, I do like having the option though for doing more granular fixing

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
```diff
@@ -0,0 +1,325 @@
...
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpdocToReturnTypeFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @var array
+     */
+    private $blacklistFuncNames = [
```
SpacePossum: shouldn't this list all [magic methods](http://php.net/manual/en/language.oop5.magic.php) ?

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,211 @@
...
+            ],
+            'nullable array of types' => [
+                '<?php /** @return null|Foo[] */ function my_foo(): ?array {}',
+                '<?php /** @return null|Foo[] */ function my_foo() {}',
+                70100,
+            ],
```
SpacePossum: I think it would be nice to have some complex cases, like
```php
class A
{
    /** @return Foo*/  # some comment
   final/**/static/**/function/**/A/**/(/**/)/**/{#comment 2
   }
}
```

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
```diff
@@ -0,0 +1,325 @@
...
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 1; 0 < $index; --$index) {
+            if (!$tokens[$index]->isGivenKind(T_FUNCTION)) {
```
SpacePossum: if it is a `lambda` you can continue here as well (as long as PHP doesn't allow return type hinting on `lambda`'s)

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
Slamdunk: No, only those 3 cannot declare a return type.

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
Slamdunk: Lambdas are allowed to have a return type: https://3v4l.org/NmbNN

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: Thanks, added

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
julienfalque: Whatever it is exactly, the root type is always `array`.

src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
SpacePossum: well TIL, thanks :+1:

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: Still, this PR rely on https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.11.1/src/DocBlock/Annotation.php#L30 , I should change this to support `Foo[][]`. Do have I?

I know it sounds a bit irrational, and it is, but I'm still against it :(

Slamdunk:
@keradus may we include this in the [2.12](https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestone/104) milestone?

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
SpacePossum: `Foo[]` is an array (or supports array access), but I still think `Foo[][]` is not a type description and therefore is used wrongly.

Slamdunk:
> Yet, I dared to run this rule over repo. Effect? crash with fatal error. About what? Incompatible interfaces.
> When there is annotation it's moved to singature, which is great.
> But when there is no annotation or annotation is `@inheritdoc`, fixer can't do a thing, for which project is not running at all after changes.
> 
> While I strongly believe in value of this rule, having it in current form would bring IMO more harm than good, as it will automate only little piece of code asking user to adjust manually all the rest.

I followed the workflow you described in my projects, both new and legacy, and with non-trivial effort fixing the manual things I found a log of bugs/improvements.
You are right, this PR is not ready for **automation**, but I consider it ready to improve userland codebases and give more value than harm.

Stated this, I also agree on your POV and this PR shouldn't be merged as a production-ready feature.

> I see 2 possibilities:
>     * freeze the PR until PHP CS Fixer will support context-aware rules (no ETA on that)
>     * create special, internal interface like ExperimentalFixerInterface, mark current fixer with it

I don't trust stalled/freezed features: only maintaners and the main developer are aware of the possible new feature, a lot of effort is required to keep it up to date, and no one gets value in the meantime.

I like your second proposal: all users can explore the feature, contribute, give feedbacks and ideas; or can also downvote it.
Either the feature will prosper or die, its future is reached much faster and better.

> hide experimental fixers in readme

Little [OT] about this: ok to make experimental features "*tricky*" to enable, but why hiding them? Proper disclaimers are enough for me. If disclaimers are done right, everything can be put well visible.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
keradus: how can one define 2-d dimension of sth ?
eg, when we have `$cartesianCoordinate[$locationX][$locationY] = new City(...)`
it's `City[][]`

keradus:
> but why hiding them?

because if we would decide to merge it, I would like to merge #2918

both are problematic from same point of view - they need context.
And we can get poor's men context by enforcing rule is executed in env that has access to whole project (and not only modified files) and composer autoloader, and even deps.
Super easy to achieve silly way, like in mentioned 2918, but one needs to know how to properly run such rule (eg installation via phar will always fail). benefit? we can fix every single method, even if it has no docs, but it's parent has docs.
but that silly way, without doing it properly, will break things like standard CI integration, execution rule from phar, having non-standard autoloader, caching, and so, and so - super bigger thing => that's the reason why not make this public, as it will break a lot in tool itself if one run rule like any other rule.

julienfalque:
:+1: for an `ExperimentalFixerInterface`. I would still show the fixer in the main README, with an explicit note about why the fixer is experimental, just like with risky fixers.

SpacePossum:
Maybe we can add and start working on/with this in a `nightly` type of branch and releases? or maybe in the 3.0 branch?

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
SpacePossum: I think I'm just slow here, but if you return `City[][]` it will be an array of what? Not `City`, than you would return `City[]` , so of what type would the returned item be?

julienfalque:
If we don't show it in the README, I think almost nobody will know the fixer exists and use it, reducing chances users try it, report bugs, suggest improvements and maybe contribute some of them. Something that is not documented simply doesn't exist.

keradus:
Actually I... agree. 
And that's good alternative for me to "not merge until context is ready".
why?

we can merge, but not expose. We already maintain (eg all bc breakers with v3 would be applied, like for all other printers), but we will announce no sooner than we at least have damn good docs what it means it's experimental, how to use it, how to use it not, what's not working.

for that, we can merge, stop paying long-living PR debt, and start working on that docs in separated PR, that would be merged maybe in 2, 3 weeks, depends how fast one would do it.

README.rst
```diff
@@ -1304,6 +1304,18 @@ Choose from the list of available rules:
 
   Docblocks should only be used on structural elements.
 
+* **phpdoc_to_return_type**
```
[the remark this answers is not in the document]
Slamdunk: Sorry for late reply.
I agree, `param` should be handled as well.
Few notes:

1. Before facing this, "experimental" discussion should reach a solution
1. `return` part was easy to code, `param` maybe not so much: I would split the fixer into two fixers to reduce complexity and only at the end, if it'll be possible, merge them back together. I'm used to this practice to get to the result easier and faster
1. I'm too much busy this year, I think I cannot contribute to this for the following months :frowning_face:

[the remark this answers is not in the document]
keradus: Fully understandable. No worry.

Then, let us keep the scope here to touch only `return`.
In such case, to have it merged faster, can I still ask for few minutes of yours ?

Please extend `FixerDefinition` to contain info that "This rule is *experimental* and is not covered with BC promise." and stress that "`@return` annotation is crucial to fixer to make changes, signatures of methods without it (no docblock, inheritdocs) will not be fixed - this automate part of the work, but not everything, manual action is still required !" or sth like that?

Just as a text in definition. no new interface or special handling of "experimental" rules. That way, we would ship it still for 2.12 and postpone the problem of "experimental" rules to future us.

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
```diff
@@ -0,0 +1,231 @@
...
+            'null alone cannot be a return type' => [
+                '<?php /** @return null */ function my_foo() {}',
+            ],
+            'skip mixed types' => [
+                '<?php /** @return Foo|Bar */ function my_foo() {}',
+            ],
+            'nullable type' => [
+                '<?php /** @return null|Bar */ function my_foo(): ?Bar {}',
+                '<?php /** @return null|Bar */ function my_foo() {}',
+                70100,
```
keradus: super trivial, but would be nice to have test-case showing behaviour on PHP <7, currently, it's not defined by test

README.rst
```diff
@@ -1374,6 +1374,18 @@ Choose from the list of available rules:
...
+
+  EXPERIMENTAL: Takes ``@return`` annotation of non-mixed types and adjusts
+  accordingly the function signature. Requires PHP >= 7.0.
+
+  *Risky rule: [1] This rule is EXPERIMENTAL and is not covered with backward compatibility promise. [2] ``@return`` annotation is mandatory for the fixer to make changes, signatures of methods without it (no docblock, inheritdocs) will not be fixed. [3] Manual actions are required is inherited signatures are not properly documented. [4] ``@inheritdocs`` support is under construction.*
```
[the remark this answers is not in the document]
Slamdunk: Yes, you asked me to **stress** the user about this fixer version's brittleness

tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php
Slamdunk: Yes it is. Take a look at the `testFix` function: if you are running tests with a PHP_VERSION lower than the third param, the second param becomes both the expected and the input values.

README.rst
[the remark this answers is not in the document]
Slamdunk: Typo fixed, thanks

README.rst
```diff
@@ -1374,6 +1374,18 @@ Choose from the list of available rules:
...
+
+  Configuration options:
+
+  - ``scalar_types`` (``bool``): fix also scalar types; may have unexpected
+    behaviour due to PHP bad type coercion system; defaults to ``false``
```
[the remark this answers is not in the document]
Slamdunk: Changed.
I am still againts this though: I took too much pain in the ass for this.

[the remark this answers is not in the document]
keradus: I understand, that's why we have option.
At PHP CS Fixer, we try to promote us much strictness about code as possible, that's why I asked.
Thanks for the change, even if not fully agreeing with it

SpacePossum:
Thanks @Slamdunk :tada: :100:

keradus:
experimental feature which risky level is so high, that one needs for sure to manually adjust code after fixer is executed. it's like changing the game here.

PR #3644, opened 2018-03-28
Add PhpdocFullyQualifiedTypesFixer

Slamdunk:
Reference: #3276
What about fixing PHPDoc too?

```diff
 use Foo\Bar;

 /**
- * @param \Foo\Bar $foo
+ * @param Bar $foo
  */
 function foo($foo) {}
```

Ping @veewee


keradus:
3276 was about changing FQCN to short CN, why doing the reverse ?

Slamdunk:
> 3276 was about changing FQCN to short CN, why doing the reverse ?

The topic reports the Expected VS Actual diff of the test, nothing to reverse here.

veewee:
@Slamdunk : that could indeed be a new feature. Currently I did not touch any docblocks. Personally, I am removing these kind of hard to maintain `@param` docblocks since they do not make any sense when using strict types anymore.

Slamdunk:
> Personally, I am removing these kind of hard to maintain @param docblocks since they do not make any sense when using strict types anymore.

Heh, if only we could...

> that could indeed be a new feature. Currently I did not touch any docblocks.

I pinged you because #3276 seems a complex work and maybe you faces some issue I need to know before building this new feature.

veewee:
Didn't face any issues regarding to the doblocks.
The fixer itself isn't that complex, the hard part was abstracting the logic into generic analyzers.
It should be fairly simple to detect the function docblock and change the param / return tags.

You could add a `fixFunctionDocblocks()` method underneeth this line:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/src/Fixer/Import/FullyQualifiedStrictTypesFixer.php#L97

The way I think it should work:

- Search for previous docblock token and make sure it is the doblock of current function.
- Parse param and return tags. (There is a class for that)
- Split up the method `detectAndReplaceTypeWithShortType()` in `detectShortType()` and `replaceFqcnWithShortType()` or something like that
- Reuse the `detectShortType` to replace the param / return tags
- replace old docblocks with new docblocks
- party like it's 1999

keradus:
> The topic reports the Expected VS Actual diff of the test, nothing to reverse here.

I treated your opening post as feature request definition, not as output of failing test :D 

little :+1: for the feature, but indeed, it's better to move it to prototype if possible, but that would be harder, so probably it's worth to do easier feature first.

Slamdunk:
> I treated your opening post as feature request definition, not as output of failing test :D

You are right, it was misleading. I do so for bug reports, but it is wrong for new feature .

> it's better to move it to prototype if possible

I don't understand whatyou mean

ostrolucky:
I'll try this one. Although it will be different fixer, in Phpdoc namespace - to have better ability to deal with priority conflicts. FullyQualified**StrictTypes**Fixer is wrong for such a thing anyway.

julienfalque:
:+1: for this. I'm fine if the current fixer also deals with PHPDocs rather than creating a new fixer.

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * {@inheritdoc}
+     *
+     * Must run before PhpdocAlignFixer.
+     * Must run after CommentToPhpdocFixer, PhpdocIndentFixer, PhpdocScalarFixer, PhpdocToCommentFixer, PhpdocTypesFixer.
+     */
+    public function getPriority()
+    {
+        /*
+         * Should be run before all other docblock fixers apart from the
```
SpacePossum: is this a copy-paste-comment?

src/AbstractPhpdocTypesFixer.php
```diff
@@ -56,6 +56,7 @@ public function isCandidate(Tokens $tokens)
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
+        $tokensForAnalysis = clone $tokens;
```
SpacePossum: this is very expensive, if you iterate backwards;

```php
    /**
     * {@inheritdoc}
     */
    {
        for ($i = \count($tokens) - 1; $i >= 0; --$i) {
            if (!$tokens[$i]->isGivenKind(T_DOC_COMMENT)) {
                continue;
            }

            $doc = new DocBlock($tokens[$i]->getContent());
            $annotations = $doc->getAnnotationsOfType($this->tags);

            if (empty($annotations)) {
                continue;
            }

            foreach ($annotations as $annotation) {
                $this->fixTypes($tokens, $annotation);
            }

            $tokens[$i] = new Token([T_DOC_COMMENT, $doc->getContent()]);
        }
    }
```

all tests pass. However, I'm not sure why you are cloning here so maybe this doesn't work?

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    protected function normalize(Tokens $tokens, $type)
+    {
+        if (0 !== strpos($type, '\\')) {
+            return $type;
+        }
+
+        if ((new TypeAnalysis(substr($type, 1), 1, 2))->isReservedType()) {
```
SpacePossum: just for my uinderstanding, the `1, 2` are dummy value right?

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
```diff
@@ -0,0 +1,79 @@
...
+            return $type;
+        }
+
+        return (new TypeShortNameResolver())->resolve($tokens, $type);
```
SpacePossum: is there nice way to prevent creating a new instance of `new TypeShortNameResolver()` for each candidate? (maybe private property of the fixer)

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
Slamdunk: Yes , I've started from `PhpdocScalarFixer` and it seemed to me a good idea to retain part of it.

src/AbstractPhpdocTypesFixer.php
Slamdunk: Well, I've started with touching `\PhpCsFixer\AbstractPhpdocTypesFixer` as minimum as possible, and without your fix the loop gets infinite because `\PhpCsFixer\Tokenizer\Resolver\TypeShortNameResolver` calls (somewhere in its dependencies) `\PhpCsFixer\Tokenizer\TokensAnalyzer` which resets the cursor with `$this->tokens->rewind();`

I'll apply your fix soon :+1:

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
Slamdunk: You're right, comment is coming

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
Slamdunk: `TypeShortNameResolver` is a stateless class, and I've read few times @julienfalque to prefer not saving anything privately due to performance reason.

Moreover, `FullyQualifiedStrictTypesFixer` do the same:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/c8afb599858876e95e8ebfcd97812d383fa23f02/src/Fixer/Import/FullyQualifiedStrictTypesFixer.php#L155

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
SpacePossum: ok, than we should back up this comments claim with priority test fixtures so we no longer need this comment :+1:

Slamdunk: So, I've added all the priority tests that I've could imaging, and then had to delete most of them except two cause of a test allowing only meaningful priority tests :smiley:

Slamdunk:
Could someone please fix the PR label to `kind/feature`?

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
SpacePossum: That is my concern :smiley: so a lot PHPDoc fixer claim that they must run before a whole bunch of other fixers, yet creating a priority test to prove it seems not possible.

tests/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixerTest.php
```diff
@@ -0,0 +1,236 @@
...
+ * @return object
+ */
+function foo($a, $b) {}',
+            ],
+            'Handle union types' => [
+                '<?php
+use Foo\Bar;
+
+/**
+ * @return null|int|Bar
```
mvorisek, line 223: add some more complex phpdoc types like nested type, generic, array shape, conditional type

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
mvorisek: this is no longer true

src/Fixer/Phpdoc/PhpdocFullyQualifiedTypesFixer.php
```diff
@@ -0,0 +1,80 @@
...
+     * {@inheritdoc}
+     */
+    protected function normalize(Tokens $tokens, $type)
+    {
+        if (0 !== strpos($type, '\\')) {
+            return $type;
+        }
+
+        // Dummy start and end indexes, we only need isReservedType result
+        if ((new TypeAnalysis(substr($type, 1), 1, 2))->isReservedType()) {
```
mvorisek, line 74: since https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7188/files#diff-9eb8d1884736f293d0a3ea5127a6713fea7a744e293f41b48b263e910297a4bdR66 the `1, 2` dummy values are not needed

Wirone:
Actually it was implemented in #7622 so we can close this 🙂. Thanks @Slamdunk for your work!

PR #3652, opened 2018-03-29
CiIntegrationTest - run tests with POSIX sh, not Bash

### LINKED ISSUE #3650: tests/Smoke/CiIntegrationTest.php doesn't work without Bash
After fixing #3544 it is still not possible to run PHP-CS-Fixer's unit tests in an environment that doesn't have Bash, because tests/Smoke/CiIntegrationTest.php fails with

> env: can't execute 'bash': No such file or directory

`BashScriptExecutor` here should be replaced with `ShellScriptExecutor` that uses `sh` instead of `bash`:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/90efa83e60f00d848e56c6484f6a4eb8b03c4684/tests/Smoke/CiIntegrationTest.php#L15

keradus:
Closes #3650

### DIFF
--- a/tests/Smoke/CiIntegrationTest.php
+++ b/tests/Smoke/CiIntegrationTest.php
@@ -12,7 +12,7 @@
 
 namespace PhpCsFixer\Tests\Smoke;
 
-use Keradus\CliExecutor\BashScriptExecutor;
+use Keradus\CliExecutor\ScriptExecutor;
 use Keradus\CliExecutor\CommandExecutor;
 use PhpCsFixer\Tests\TestCase;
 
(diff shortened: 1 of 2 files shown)


tests/Smoke/CiIntegrationTest.php
```diff
@@ -251,6 +251,9 @@ private static function executeCommand($command)
 
     private static function executeScript(array $scriptParts)
     {
-        return BashScriptExecutor::create($scriptParts, self::$fixtureDir)->getResult();
+        // @TODO: drop $scriptInit, for now it's needed, as defaut `set -eu` is causing our scripts to crash
+        $scriptInit = array('#!/bin/sh', 'set -e', '');
```
keradus: @EvgenyOrekhov, I remember that you wanted to fix `s/e/eu` globally.
during that, could you handle this one as well ?

EvgenyOrekhov: I created #3654 for `set -eu` and referenced this PR there.

PR #3656, opened 2018-03-30
DX: Clean ups

SpacePossum:
-

### DIFF
--- a/tests/AutoReview/ComposerTest.php
+++ b/tests/AutoReview/ComposerTest.php
@@ -43,7 +43,7 @@ final class ComposerTest extends TestCase
     /**
      * @param string $version
      *
-     * @return strting
+     * @return string
      */
     private function convertAppVersionToAliasedVersion($version)
     {
--- a/tests/AutoReview/FixerFactoryTest.php
+++ b/tests/AutoReview/FixerFactoryTest.php
@@ -79,7 +79,6 @@ final class FixerFactoryTest extends TestCase
             array($fixers['general_phpdoc_annotation_remove'], $fixers['phpdoc_trim']),
             array($fixers['general_phpdoc_annotation_remove'], $fixers['no_empty_phpdoc']),
             array($fixers['indentation_type'], $fixers['phpdoc_indent']),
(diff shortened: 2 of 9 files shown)


tests/Console/Command/SelfUpdateCommandTest.php
```diff
@@ -132,7 +132,7 @@ public function testExecute(
 
         $commandTester = $this->execute($command, $input, $decorated);
 
-        self::assertSame($expectedFileContents, file_get_contents($this->getToolPath()));
+        self::assertStringEqualsFile($this->getToolPath(), $expectedFileContents);
```
keradus: :-1:
first value shall be expected value, not actual value.
you will get wrong error message in case of failure

SpacePossum:
```
assertStringEqualsFile(string $expectedFile, string $actualString[, string $message = ''])
```
here we can't swap those, as one is content and the other a path to file

SpacePossum:
is Travis failing on
```
Remaining deprecation notices (21)
  21x: `\Keradus\CliExecutor\BashScriptExecutor` is deprecated, use `\Keradus\CliExecutor\ScriptExecutor` instead.
    16x in CiIntegrationTest::testIntegration from PhpCsFixer\Tests\Smoke
    4x in CiIntegrationTest::tearDown from PhpCsFixer\Tests\Smoke
    1x in CiIntegrationTest::setUpBeforeClass from PhpCsFixer\Tests\Smoke
```

or I'm missing some real issue here?

tests/Console/Command/SelfUpdateCommandTest.php
keradus: yep, that;'s why this assertion is not used here, but simply `assertSame`.
we could use `assertFileEqualsString`, but such assertion doesn't exists.

Still, you proposed to use wrong assertion, as it assert different thing than we want. please revert that one

keradus:
ref #3652

keradus:
Thank you @SpacePossum.

PR #3658, opened 2018-03-30
DX: Clean ups

SpacePossum:


tests/Indicator/PhpUnitTestCaseIndicatorTest.php
```diff
@@ -0,0 +1,125 @@
...
+
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tests\TestCase;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ *
+ * @internal
+ * @coversNothing
```
SpacePossum: not sure at this one, cc @keradus

keradus: why not `$covers \PhpCsFixer\Indicator\PhpUnitTestCaseIndicator` ?

SpacePossum: sounds good! I'm just not to familiar with these annotations I think

SpacePossum:
haha sorry at that organisation (great name btw :P), is there a way to escape such an annotation in a commit message?

keradus:
for my best knowledge, nope

SpacePossum:
rebased and forced pushed after resolving conflict

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -62,7 +62,6 @@ public function provideFixersPriorityCases()
             [$fixers['array_indentation'], $fixers['binary_operator_spaces']],
             [$fixers['array_syntax'], $fixers['binary_operator_spaces']],
             [$fixers['array_syntax'], $fixers['ternary_operator_spaces']],
-            [$fixers['backtick_to_shell_exec'], $fixers['escape_implicit_backslashes']],
```
keradus: could you elaborate on this one? are we sure there is no prioritiy conflict between them ?

I assume you didn't saw warning from https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/tests/IntegrationTest.php#L71 
(as if you would and willingly drop the test here, you would also drop the expection there)

keradus: Also, `backtick_to_shell_exec` has
> // Should run before escape_implicit_backslashes

comment in `getPriority`

SpacePossum: on second thought I revered the change, making it more easy to pass this PR

keradus:
Thank you @SpacePossum.

PR #3659, opened 2018-03-30
use php-cs-fixer/diff ^2.0

SpacePossum:


keradus:
please retarget to 2.11

SpacePossum:
that would be a BC on 2.11 I think, as we've different differ-delegated classes/options there

keradus:
good point, we remove all but strict udiff at 3.0

SpacePossum:
thanks @keradus :)  nice to see all this finds its place!

PR #3660, opened 2018-03-30
Fix  do not rely on order of fixing in CiIntegrationTest

### LINKED ISSUE #3171: One CiIntegration test failing on PHP 7.1
I have recently installed fresh Ubuntu 16.04.3 LTS with both PHP 5.6 and 7.1 and GIT (version 2.7.4).

I have checkout `master` branch and run `composer install` with PHP 5.6. And run tests and all went fine. I have switched to PHP 7 (by switched I mean changed path of `/usr/bin/php`). And then I have one test failing (I have NOT updated dependencies):

```bash
vendor/bin/phpunit tests/CiIntegrationTest --filter=testIntegration#0
PHPUnit 5.5.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.1.10-1+ubuntu16.04.1+deb.sury.org+1
Configuration: PHP-CS-Fixer/phpunit.xml.dist

Testing PhpCsFixer\Tests\CiIntegrationTest
F                                                                                                                                                                                                                                      1 / 1 (100%)

 1. 112ms to run PhpCsFixer\Tests\CiIntegrationTest:testIntegration with data set #0


Time: 147 ms, Memory: 4.00MB

There was 1 failure:

1) PhpCsFixer\Tests\CiIntegrationTest::testIntegration with data set #0 ('random-changes', array('touch dir\ a/file.php', 'rm -r dir\ c', 'echo "" >> dir\ b/file\ b.php', 'echo "echo 1;" >> dir\ b/file\ b.php', 'git add .', 'git commit -m "Random changes" -q'), array('dir a/file.php', 'dir b/file b.php'), array('4', '--path-mode=intersection -- d... b.php', '--path-mode=intersection', '--', 'dir a/file.php', 'dir b/file b.php'), 'S.')
Failed asserting that 'Loaded config default from ".php_cs.dist".
.S
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error' matches PCRE pattern "/^(PHP needs to be a minimum version of PHP 5\.6\.0 and maximum version of PHP 7\.2\.\*\.
Ignoring environment requirements because `PHP_CS_FIXER_IGNORE_ENV` is set\. Execution may be unstable\.
)?(You are running PHP CS Fixer with xdebug enabled\. This has a major impact on runtime performance\.
If you need help while solving warnings, ask at https\:\/\/gitter\.im\/PHP\-CS\-Fixer, we will help you\!
)?Loaded config default from "\.php_cs\.dist"\.
S\.
Legend\: \?\-unknown, I\-invalid file syntax, file ignored, S\-Skipped, \.\-no changes, F\-fixed, E\-error$/".

PHP-CS-Fixer/tests/CiIntegrationTest.php:166
```

The difference is an order of checked files - `.S` instead of `S.`. I have checked at some other Ubuntu 16.04 instance with the same PHP 7.1 and GIT versions and the test is fine. This only happens on my PC.

PHP 5.6 modules:
```bash
php -m
[PHP Modules]
calendar
Core
ctype
curl
date
dom
ereg
exif
fileinfo
filter
ftp
gd
gettext
hash
iconv
igbinary
intl
json
libxml
mbstring
memcached
mhash
msgpack
openssl
pcntl
pcre
PDO
Phar
posix
readline
Reflection
session
shmop
SimpleXML
sockets
SPL
standard
sysvmsg
sysvsem
sysvshm
tokenizer
wddx
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```

PHP 7.1 modules:
```bash
php -m
[PHP Modules]
calendar
Core
ctype
curl
date
dom
exif
fileinfo
filter
ftp
gd
gettext
hash
iconv
igbinary
intl
json
libxml
mbstring
memcached
msgpack
mysqli
mysqlnd
openssl
pcntl
pcre
PDO
pdo_mysql
pdo_sqlite
Phar
posix
readline
Reflection
session
shmop
SimpleXML
sockets
SPL
sqlite3
standard
sysvmsg
sysvsem
sysvshm
tokenizer
wddx
xml
xmlreader
xmlwriter
xsl
Zend OPcache
zip
zlib

[Zend Modules]
Zend OPcache
```

Does anyone have an idea why this happens?

kubawerlos:
Closes #3171

### DIFF
--- a/tests/Smoke/CiIntegrationTest.php
+++ b/tests/Smoke/CiIntegrationTest.php
@@ -136,19 +136,22 @@ Ignoring environment requirements because `PHP_CS_FIXER_IGNORE_ENV` is set. Exec
 If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
 ';
 
-        $executionDetails = "Loaded config default from \".php_cs.dist\".
-${expectedResult3Files}
-Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error";
-
-        $this->assertRegExp(
-            sprintf(
-                '/^(%s)?(%s)?%s$/',
-                preg_quote($optionalIncompatibilityWarning, '/'),
-                preg_quote($optionalXdebugWarning, '/'),
-                preg_quote($executionDetails, '/')
-            ),
-            $result3->getError()
+        $pattern = sprintf(
+            '/^(?:%s)?(?:%s)?%s\n([\.S]{%d})\n%s$/',
+            preg_quote($optionalIncompatibilityWarning, '/'),
+            preg_quote($optionalXdebugWarning, '/'),
+            preg_quote('Loaded config default from ".php_cs.dist".', '/'),
+            strlen($expectedResult3Files),
+            preg_quote('Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error', '/')
         );
+
+        $this->assertRegExp($pattern, $result3->getError());
+
+        preg_match($pattern, $result3->getError(), $matches);
+
+        $this->assertArrayHasKey(1, $matches);
+        $this->assertSame(substr_count($expectedResult3Files, '.'), substr_count($matches[1], '.'));
+        $this->assertSame(substr_count($expectedResult3Files, 'S'), substr_count($matches[1], 'S'));
 
         $this->assertRegExp(
             '/^\s*Checked all files in \d+\.\d+ seconds, \d+\.\d+ MB memory used\s*$/',


tests/Smoke/CiIntegrationTest.php
```diff
@@ -136,20 +136,23 @@ public function testIntegration(
...
-        $this->assertRegExp(
-            sprintf(
-                '/^(%s)?(%s)?%s$/',
-                preg_quote($optionalIncompatibilityWarning, '/'),
-                preg_quote($optionalXdebugWarning, '/'),
-                preg_quote($executionDetails, '/')
-            ),
-            $result3->getError()
+        $pattern = sprintf(
+            '/^(?:%s)?(?:%s)?%s\n([\.S]{%d})\n%s$/',
```
keradus: :+1: for non-capturing groups ;)

tests/Smoke/CiIntegrationTest.php
```diff
@@ -136,20 +136,23 @@ public function testIntegration(
...
+            preg_quote('Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error', '/')
         );
 
+        $this->assertRegExp($pattern, $result3->getError());
+
+        preg_match($pattern, $result3->getError(), $matches);
+
+        $this->assertArrayHasKey(1, $matches);
+        $this->assertSame(substr_count($expectedResult3Files, '.'), substr_count($matches[1], '.'));
+        $this->assertSame(substr_count($expectedResult3Files, 'S'), substr_count($matches[1], 'S'));
```
keradus: What if one would add case with `'F'` ? he wouldn't know that `F` wasn't tested and only `'.'` and `'S'` were.
Would be nice to reuse `\PhpCsFixer\Console\Output\ProcessOutput::$eventStatusMap` (after little refactoring, don't use private property in tests ;) ).

alternative - ensure that `(count of '.' + count of 'S') === strlen($expectedResult3Files)`

kubawerlos: I have no concept for "little refactoring" (best I came with is extracting statuses to classes value object `Status` and collection `StatusCollection` - it's not a little change and seems like overkill - we won't be extending the list anyway) and doing it only for the use of this test does not feel right either.

I've ensured in RegEx that only `.` and `S` are in the list, so if someone add case with `F` then `assertRegExp` will fail - handling all 6 type seemed a little YAGNI, or should I go in that direction?

keradus: nah, if you ensured only `.S` are allowed, it's fine as well

ktomk:
Just FYI, I was finally able to apply the changes on my end as well, I went with a similar but slightly different fixture correction. You can find it in the cross-referenced PR #3667.

keradus:
Thank you @kubawerlos.

ktomk:
Can confirm green w/ this patch on my end as well. Thanks for fixing!

PR #3661, opened 2018-03-30
 Fix: covers annotation for NoAlternativeSyntaxFixerTest

kubawerlos:
Just spotted this, NoAlternativeSyntaxFixerTest should cover its class.


keradus:
Thank you @kubawerlos.

PR #3664, opened 2018-03-31
Bump PHPUnit versions

carusogabriel:
As we have some tests with `--prefer-lowest` :)


keradus:
Please provide reasoning for such change.

carusogabriel:
> we bump them only when needed

@keradus And I just read the `CONTRIBUTING`:

```
- Don't bump version.
```

Sorry, may bad :)

keradus:
no worry!
thanks anyway ;)

Issue #3608, opened 2018-03-17
Can't Follow Contribution Guide as Tests Fail On Checkout


ktomk:
When following https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/CONTRIBUTING.md and checking out the branch which is eligible to make changes to (that is 2.2 I think), invoking Phpunit:
~~~
$ composer install
# ...
$ vendor/bin/phpunit
~~~

fails with tests.

I'm running PHP 7.2. something. Which PHP version do I need to configure for Composer so that the test-suite runs?

ktomk:
Trying to pin-point the culprit, but no luck so far. Thought it was xdebug, but w/o xdebug get a different error then (first was a test failure not taking the you're using xdebug warning message into account):

~~~
php -n -dextension=ctype -dextension=json -dextension=tokenizer -dextension=dom -dextension=mbstring  vendor/bin/phpunit --stop-on-failure 
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.3-1+ubuntu16.04.1+deb.sury.org+1
Configuration: /home/tom/PhpstormProjects/PHP-CS-Fixer/phpunit.xml.dist

Testing 
.............................................................   61 / 9683 (  0%)
# ...
....................................................S........ 5246 / 9683 ( 54%)
...............................................F

Time: 5.88 seconds, Memory: 118.00MB

There was 1 failure:

1) PhpCsFixer\Tests\Fixer\Comment\HeaderCommentFixerTest::testDoNotTouch with data set #5 ('<?= 1?><?php')
Failed asserting that 'PHP Parse error: syntax error, unexpected end of file on line 1.

Source:
<?= 1?><?php' is null.

/home/tom/PhpstormProjects/PHP-CS-Fixer/tests/Test/AbstractFixerTestCase.php:164
/home/tom/PhpstormProjects/PHP-CS-Fixer/tests/Fixer/Comment/HeaderCommentFixerTest.php:508

--

There were 31 incomplete tests:
#...
~~~

SpacePossum:
Hi and thanks for reporting :)

Can you try applying this diff to see if this is the cause/fix ?

```diff
diff --git a/tests/Fixer/Comment/HeaderCommentFixerTest.php b/tests/Fixer/Comment/HeaderCommentFixerTest.php
index 1c63733..d2c62a2 100644
[...]

PR #3667, opened 2018-04-01
undefined order of files (tppka patch 4), fixes #3171

### LINKED ISSUE #3171: One CiIntegration test failing on PHP 7.1

ktomk:
non invasive fix to deal w/ an inconsistency which is to adopt test
fixtures for undefined order of file result display on standard output
created by the file fixing iteration within the integration test.

props @keradus comment [1] to not rely on the order of fixing and earlier
[2] to not align iterator's flow just for that.

note: there is a similar approach to this in #3660.

refs:

- [1] https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3171#issuecomment-374572917

- [2] https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3617#discussion_r175851429

- #3171

- #3608

- #3660

### DIFF
--- a/tests/Smoke/CiIntegrationTest.php
+++ b/tests/Smoke/CiIntegrationTest.php
@@ -136,17 +136,19 @@ Ignoring environment requirements because `PHP_CS_FIXER_IGNORE_ENV` is set. Exec
 If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
 ';
 
-        $executionDetails = "Loaded config default from \".php_cs.dist\".
-${expectedResult3Files}
-Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error";
+        $executionDetails = 'Loaded config default from ".php_cs.dist".
+@@@
+Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error';
 
+        $patternBuffer = sprintf(
+            '/^(?:%s)?(?:%s)?%s$/',
+            preg_quote($optionalIncompatibilityWarning, '/'),
+            preg_quote($optionalXdebugWarning, '/'),
+            preg_quote($executionDetails, '/')
+        );
+        $patternBuffer = strtr($patternBuffer, array('@@@' => $expectedResult3Files));
         $this->assertRegExp(
-            sprintf(
-                '/^(%s)?(%s)?%s$/',
-                preg_quote($optionalIncompatibilityWarning, '/'),
-                preg_quote($optionalXdebugWarning, '/'),
-                preg_quote($executionDetails, '/')
-            ),
+            $patternBuffer,
             $result3->getError()
         );
 
@@ -180,7 +182,7 @@ Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes,
                     'dir b/file b.php',
                     '',
                 ),
-                'S.',
+                '(?:S[.]|[.]S)',
             ),
             array(
                 'changes-including-dist-config-file',
@@ -201,7 +203,7 @@ Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes,
                     '',
                     '',
                 ),
-                '...',
+                '\Q...\E',
             ),
             array(
                 'changes-including-custom-config-file-creation',
@@ -220,7 +222,7 @@ Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes,
                     '',
                     '',
                 ),
-                '...',
+                '\Q...\E',
             ),
             array(
                 'changes-including-composer-lock',
@@ -239,7 +241,7 @@ Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes,
                     '',
                     '',
                 ),
-                '...',
+                '\Q...\E',
             ),
         );
     }


tests/Smoke/CiIntegrationTest.php
```diff
@@ -180,7 +182,7 @@ public function provideIntegrationCases()
                     'dir b/file b.php',
                     '',
                 ),
-                'S.',
+                '(?:S[.]|[.]S)',
```
keradus: this approach will be problematic when we will have case not with 2 files processed, but with more, like 10, 20, etc....

keradus:
what is the benefit of this change comparing to #3660, which got proposed earlier ?

tests/Smoke/CiIntegrationTest.php
ktomk: Indeed it may, but that would also imply to make up ones mind about what the future might bring and as that is full of many different possibilities and outcomes I personally would not deem it so fitting for a fix and development discussion. In the concrete scenario of the fixture this is with two files (in the same order at iteration but different order at output).

ktomk:
@keradus: I have not looked much into the #3660 before compiling my patch as I realized it later. From what I see now, one (subtle) benefit might be that it is more concise as it encodes the expectation in the pattern of the fixture instead of processing a result of such a pattern afterwards. As written it's pretty similar in terms that it touches the fixture, but a bit different as it doesn't change the assertion (it does extra work on compiling the pattern compared to earlier though). Just take it as an (additional) approach which from my end made it green and use it for inspiration/see it as part of the discussion we started earlier. I think next to exchanging arguments it's also good to see the code w/ it. You don't have to take this PR if you prefer the other, I'm fine if you take it for reading only. However I would love to get some feedback what you think about it as we discussed the topic earlier. But please don't feel obliged to anything.

tests/Smoke/CiIntegrationTest.php
keradus: even with having 5 files only (like 3x`.` and 2x`S`) this is gonna be more and more complex.
and even now, with just 2 files, data providers in this PR are more complex than original code.

keradus:
This is one of the possible fixes, and it works, thus thanks for that ;) Simply, different approach was proposed earlier.
And to be honest, personally, I prefer to have simpler data provider and more complex testing method, than the reverse. For that, thanks once more, but let us stay with already merged solution.

Issue #3153, opened 2017-10-12
YodaStyleFixer with function calls


pocketrocket:
The PHP version you are using:
```bash
$ php -v
PHP 7.1.8 (cli) (built: Aug 16 2017 17:19:54) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

PHP CS Fixer version you are using:
```bash
$ php-cs-fixer -v
PHP CS Fixer 2.6.0 Local Tavern by Fabien Potencier and Dariusz Ruminski
```

The command you use to run the fixer (and the configuration file you are using if any).
```bash
$ php-cs-fixer fix --verbose --rules='{"yoda_style":true}' [PATH_TO_FILE]
```

Before Fixer:
```php
<?php
$foo = 123;
$bar = $foo === count($foo);
```

After Fixer (actually):
```php
<?php
$foo = 123;
$bar = $foo === count($foo);
```

After Fixer (expected):
```php
<?php
$foo = 123;
$bar = count($foo) === $foo;
```

SpacePossum:
Hi and thanks for the report.

The behavior of the YodaStyle fixer you observed is intended. It was discussed in various PR's and issue, both in this repo and over at Symfony.
There are requests for adding more options to this fixer to change its behavior, for example a similar discussion is going on here https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3121

Please let us know if you've time to help out!

julienfalque:
@SpacePossum Are you sure this is expected? I think the fixer is supposed to move the variable from the left side to the right side of the operator.

#3674 Fixes this issue.

SpacePossum:
Pretty sure, the fixer is the result of a very long discussion over multiple PR's and reverts, 3121 is a nice start/end on the process. I'm not in any way saying it is the _only_ good way the fixer could work and I would welcome new options, however we should be careful not to change the default (and prevent doing the whole specification discussion again ;))
Part of the confusion and the long discussion also resulted in less good naming in the current implementation, such as the method name `isVariable` used to detect that, but ended up detected candidates (which are not always variables only).

SpacePossum:
[...]

PR #3674, opened 2018-04-02
YodaStyleFixer - Fix variable detection for multidimensional arrays

### LINKED ISSUE #3335: Weird changes for the YodaStyleFixer
PHP: 7.2.0
PHP-CS-Fixer: 2.9.0
Command: `php-cs-fixer fix`
Configuration file:
```php
<?php

    ->in(__DIR__.'/src')
    ->in(__DIR__.'/tests')
    ->in(__DIR__.'/features/bootstrap')
    ->in(__DIR__.'/app')
    ->in(__DIR__.'/api')
    ->exclude('cache')
    ->exclude('logs')
    ->exclude('sessions')
    ->notName('bootstrap.php.cache')
;

    ->setRules(array(
        '@Symfony' => true,
        'array_syntax' => array('syntax' => 'long'),
        'doctrine_annotation_braces' => true,
        'doctrine_annotation_indentation' => true,
        'doctrine_annotation_spaces' => true,
        'heredoc_to_nowdoc' => true,
        'no_useless_else' => true,
        'ordered_imports' => true,
    ))
    ->setFinder($finder)
    ->setCacheFile(__DIR__.'/build/.php_cs.cache')
;
```

Here is the diff produced by the tool:

```diff
             if (isset($referenceMessagesByDomainAndKey[$key])
-                && $messages['en'] === $referenceMessagesByDomainAndKey[$key]['en']
-                && $messages['fr'] === $referenceMessagesByDomainAndKey[$key]['fr']) {
+                && $referenceMessagesByDomainAndKey[$key]['en'] === $messages['en']
+                && $referenceMessagesByDomainAndKey[$key]['fr'] === $messages['fr']) {
                 continue;
             }
```

I expected no changes there, as I have variables on both sides.

julienfalque:
Fixes #3335 and replaces #3505.

### DIFF
--- a/src/Fixer/ControlStructure/YodaStyleFixer.php
+++ b/src/Fixer/ControlStructure/YodaStyleFixer.php
@@ -544,23 +544,29 @@ final class YodaStyleFixer extends AbstractFixer implements ConfigurationDefinit
             }
 
             // $a-> or a-> (as in $b->a->c)
-            if ($current->isGivenKind($expectString ? T_STRING : T_VARIABLE) && $next->isGivenKind(T_OBJECT_OPERATOR)) {
+            if ($current->isGivenKind([T_STRING, T_VARIABLE]) && $next->isGivenKind(T_OBJECT_OPERATOR)) {
                 $index = $tokens->getNextMeaningfulToken($nextIndex);
                 $expectString = true;
 
                 continue;
             }
 
-            // $a[...] or a[...] (as in $c->a[$b])
-            if ($current->isGivenKind($expectString ? T_STRING : T_VARIABLE) && $next->equals('[')) {
-                $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE, $index + 1);
+            // $a[...], a[...] (as in $c->a[$b]), $a{...} or a{...} (as in $c->a{$b})
+            if (
+                $current->isGivenKind($expectString ? T_STRING : T_VARIABLE)
+                && $next->equalsAny(['[', [CT::T_ARRAY_INDEX_CURLY_BRACE_OPEN, '{']])
+            ) {
+                $index = $tokens->findBlockEnd(
+                    $next->equals('[') ? Tokens::BLOCK_TYPE_INDEX_SQUARE_BRACE : Tokens::BLOCK_TYPE_ARRAY_INDEX_CURLY_BRACE,
+                    $index + 1
+                );
                 if ($index === $end) {
                     return true;
                 }
 
                 $index = $tokens->getNextMeaningfulToken($index);
 
-                if (!$tokens[$index]->isGivenKind(T_OBJECT_OPERATOR)) {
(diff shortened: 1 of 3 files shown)


julienfalque:
PR scope reduced as discussed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3153#issuecomment-379725393.

SpacePossum:
Thanks for picking this up :)
I was struggling to get the possible repeated access detected correctly as well,
for example
```
$c2->a{a}{a}']
$c1->a{a}->a']
$a[1][2][3]/**/[4]
```

for more sample, please see this test; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3505/files

I didn't test your solution yet, will do soon!

julienfalque:
In the tests you added, which one are supposed to be swapped? All that should be detected as a variable (`$expected = true`)?

SpacePossum:
I think;
```php
+            [true, '$a'],
+            [true, '$a+1'],
+            [true, '$b->a'],
+            [true, '$c->a[a]'],
+            [true, '$c1->a[a]->a'],
+            [true, '$c2->a[a][a]'],
+            [true, '$d->a[$a]'],
+            [true, '$c->a{a}'],
+            [true, '$c1->a{a}->a'],
+            [true, '$c2->a{a}{a}'],
+            [true, '$d->a{$a}'],
```
and these
```
$c2->a{a}{a}']
$c1->a{a}->a']
$a[1][2][3+5]/**/[4*2-1]
```
for sure

julienfalque:
So `$a+1` should be detected as a variable but it's not. Does it mean you want me to add the following test case?

```php
['<?php return $b === $a+1;'],
```

SpacePossum:
that looks like a mistake in my testset, sorry about that one. IIRC the bug is handling single level array access different than deeper levels. So $a[1] vs $a[1][2] and deeper. Also for $a->$A vs $a->$a->$a and combinations of that.

julienfalque:
@SpacePossum I added more test cases from `provideFixCases()` in your PR. Most already pass (some with the changes in the fixer, some even without) so I didn't add them all. I'm still not sure how to proceed with test cases in `provideIsVariableCases()` to be honest. Could you show how those would look like if used in `provideFixCases()`?

SpacePossum:
I've added more test in your branch to make it a bit more easy I hope. I'm fine with not having dedicated test for `provideIsVariableCases`-especially as the method name is misleading anyway ;)

So, after all the awesome work there is one case left I could find:

```php
<?php return $foo->$a === $foo->$b->$c;
```

IMHO this should not be fixed/touched, as both the left and right side of the `===` are the same type of access to something, WDYT?

julienfalque:
Thanks, I'll have a look shortly!

> IMHO this should not be fixed/touched, as both the left and right side of the === are the same type of access to something, WDYT?

I agree it should not be changed though I think it's because both sides are assignable, not because they are of the same access type.

julienfalque:
Green :)

SpacePossum:
nice, thanks! looking good @julienfalque :)

keradus:
Thank you @julienfalque.

PR #3675, opened 2018-04-02
[WIP] Fix: handle missing fixers in fabbot.io

kubawerlos:


.php_cs.dist
```diff
@@ -82,10 +82,15 @@ $config = PhpCsFixer\Config::create()
 // special handling of fabbot.io service if it's using too old PHP CS Fixer version
 if (false !== getenv('FABBOT_IO')) {
     try {
-        PhpCsFixer\FixerFactory::create()
-            ->registerBuiltInFixers()
-            ->registerCustomFixers($config->getCustomFixers())
-            ->useRuleSet(new PhpCsFixer\RuleSet($config->getRules()));
+        $config = PhpCsFixer\FixerFactory::create()
```
julienfalque: You are overriding the `Config` instance in `$config` with a `FixerFactory` instance. Is that expected? I think this is the reason fabbot failed to run on your PR.

.php_cs.dist
```diff
@@ -82,10 +82,15 @@ $config = PhpCsFixer\Config::create()
...
+            ->registerBuiltInFixers();
+        $config->useRuleSet(new PhpCsFixer\RuleSet(array_filter(
```
julienfalque: This will detect unknown rules but not unknown/invalid config for known rules. Shall we go further to cover all cases?

.php_cs.dist
kubawerlos: Yeah, must been tired that day. Fixed.

.php_cs.dist
kubawerlos: Fixing having unknown rules solves current problem, as we agree how to solve it let's go with invalid config then - or is there any clean solution how to do it in `array_filter`?

keradus:
request for debug purposes:
- change target branch to master
- touch any file that's under PHP CS Fixer control
- let confirm if fabbot will fail or not

then, after fabbot is not failing and we finish the review, I will ask to retarget the branch again

.php_cs.dist
keradus: you could create a fixer from rule (`braces` -> `BracesFixer`) and then call `->configure(...)` to know if it's failing with config or not

.php_cs.dist
```diff
@@ -81,23 +81,26 @@ $config = PhpCsFixer\Config::create()
...
-        $config->setRules([]);
-    }
+    $fixerFactory = PhpCsFixer\FixerFactory::create();
+    $fixerFactory->registerBuiltInFixers();
+    $ruleSet = new PhpCsFixer\RuleSet($config->getRules());
+    $config->setRules(array_filter(
+        $ruleSet->getRules(),
+        function ($config, $name) use ($fixerFactory) {
+            try {
+                $fixerFactory->useRuleSet(new PhpCsFixer\Ruleset([$name => $config]));
```
keradus: 1. will crash
```diff
-Ruleset
+RuleSet
```

2. won't work as you think:
```php
<?php

$fixerFactory = PhpCsFixer\FixerFactory::create();
$fixerFactory->registerBuiltInFixers();

$fixerFactory->useRuleSet(new PhpCsFixer\RuleSet(['strict_param' => true]));
$fixerFactory->useRuleSet(new PhpCsFixer\RuleSet(['braces' => true]));

> In FixerFactory.php line 167:
>  Rule "braces" does not exist.  
```


this proves one thing - you need to move this logic from config file to internal class, and then add tests for it.

kubawerlos:
Ok, I commented out all rules except the one for short array syntax, yet fabbot.io still wants to have long syntax. What am I missing?

kubawerlos:
As you can see after first commit fabbot.io is passing, then second commit - which only adds `getenv('FABBOT_IO');` after`return` - makes it failing again, wanting to change array syntax.

@fabpot can you point me what is wrong? Why https://fabbot.io/report/FriendsOfPHP/PHP-CS-Fixer/3675/a9c52ddbea176be8c34d554b6397f20d44ac6312 wants to change array syntax? Was there different config loaded for some reason?

kubawerlos:
@julienfalque first attempt with external class to filter failed due it was not whitelisted, similar to what is happening now - after I moved filtering to `.php_cs.dist`. Can you think a way we could make it with the whitelist limitation?

keradus:
Ok, so the whitelisted approach without any docs for it, sandbox approach without docs for it as well, while both being good approaches for sure, make it impossible for us to consume fabbot.io.
Having in mind fact we also want to switch from php-based config into yaml-based one, we won't be able to use fancy workarounds for fabbot.io in our codebase.

fabbot.io is mostly eating our food by running PHP CS Fixer, and adds few more nice to have checks.
But it doesn't work for us. In the end, for all but 2.2 LTS, our branch is ahead of fabbot.io, and we have rules fabbot.io is not aware about.

For that, sadly, remembering that most of the PRs are failing on fabbot, remembering false-positive hints provided by fabbot.io that our contributors follow, to realize they need to revert it because fabbot is outdated - which brings big confusion, after being asked few times to do so, I would, sadly, propose to disable fabbot.io for PHP CS Fixer.

@FriendsOfPHP/php-cs-fixer , please advise

fabpot:
Would it make sense to use the master branch on fabbot.io?

keradus:
If you ask me, for builds of regular repos (all but PHP CS Fixer), it shall not use `@dev`, thus non-stable version.

For builds of PHP CS Fixer repo, using `master` would decrease, but not eliminate the issue, consider following cases:
- new PR that would add new fixer and include it to `.php_cs.dist` file - then, fabbot is not aware about that rule
- new PR for 3.x branch, fabbot would crash hardly on it, as it supports v2, not v3`@dev`

julienfalque:
I see only one reliable way to fix this. On this repository, fabbot should retrieve the branch that triggers the build and run the root `php-cs-fixer` file (i.e. like done on Travis for example). For other repositories, it should check whether PHP CS Fixer is installed with Composer and use it. Otherwise it should fallback to latest release (currently 2.11).

julienfalque:
Fabbot seems to work fine again on other PRs.

PR #3676, opened 2018-04-02
RunnerTest - workaround for failing Symfony v2.8.37

kubawerlos:
[This](https://github.com/symfony/symfony/pull/26337/files) change was released with Symfony v2.8.37 today.

### DIFF
--- a/tests/Runner/RunnerTest.php
+++ b/tests/Runner/RunnerTest.php
@@ -106,7 +106,7 @@ final class RunnerTest extends TestCase
     {
         $errorsManager = new ErrorsManager();
 
-        $path = __DIR__.DIRECTORY_SEPARATOR.'..'.DIRECTORY_SEPARATOR.'Fixtures'.DIRECTORY_SEPARATOR.'FixerTest'.DIRECTORY_SEPARATOR.'invalid';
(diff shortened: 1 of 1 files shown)


kubawerlos:
Failing build example: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/361316355

keradus:
woooot :|

keradus:
it's reverted on Sf, so if we would merge it, next release of Sf would make our tests to fail again...

julienfalque:
Thought about it too but actually it should work anyway: what Symfony reverted is normalizing paths in its results. But if the given path is already normalized, normalizing it has no effect anyway, so if Symfony reverts the normalization on its side, it should have no effect on our side.

keradus:
```diff
-/foo/.../bar
+/foo
```

that's the normalization we are talking about.
2.8.37 does it, 2.8.36 and 2.8.28`@dev` doesn't.
if we will do it on our side, it will fail on 2.8.28, won't it ?

julienfalque:
No, because we normalize it *before* passing it to the finder, which will then use it as-is so the test will still pass.

keradus:
didn't spot that. awesome !

keradus:
Thank you @kubawerlos.

Issue #2723, opened 2017-04-25
Method chaining and commas in arrays


julienfalque:
I sometimes use method chaining inside arrays. In such case, I move the comma on the next line for readability like I do with semicolons outside of arrays. When applying `no_whitespace_before_comma_in_array` rule, the comma is moved next to the last closing parenthesis:

```diff
 [
     $foo
         ->foo()
-        ->foo()
-    ,
+        ->foo(),
     $bar
         ->bar()
-        ->bar()
-    ,
+        ->bar(),
 ]
```

I would like the fixer to not move the comma when it is on a new line like in this example. Do you think it makes sense? If yes, do you consider this a bugfix or a feature request (so probably with a new configuration option to enable this behavior)?

keradus:
new config option (this feature) is ok for that

keradus:
@julienfalque is this request still valid ?

julienfalque:
It is (at least for me), but haven't got the time to work on it yet.

PR #3679, opened 2018-04-04
Add MultilineWhitespaceBeforeStatementEndFixer

julienfalque:
Implements #2723.

### DIFF
--- a/src/Fixer/Whitespace/WhitespaceBeforeStatementEndFixer.php
+++ b/src/Fixer/Whitespace/WhitespaceBeforeStatementEndFixer.php
@@ -12,11 +12,10 @@ declare(strict_types=1);
  * with this source code in the file LICENSE.
  */
 
-namespace PhpCsFixer\Fixer\Semicolon;
+namespace PhpCsFixer\Fixer\Whitespace;
 
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
-use PhpCsFixer\Fixer\DeprecatedFixerInterface;
 use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerConfigurationResolverInterface;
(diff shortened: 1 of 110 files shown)


src/AbstractMultilineWhitespaceBeforeChainedCallsEndCharacterFixer.php
```diff
@@ -0,0 +1,274 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Graham Campbell <graham@alt-three.com>
+ * @author Egidijus Girčys <e.gircys@gmail.com>
+ *
+ * @internal
+ */
+abstract class AbstractMultilineWhitespaceBeforeChainedCallsEndCharacterFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
```
julienfalque: I also had to invalidate some Composer cache to fix a test failure.

src/AbstractMultilineWhitespaceBeforeChainedCallsEndCharacterFixer.php
```diff
@@ -0,0 +1,274 @@
...
+
+    /**
+     * @internal
+     */
+    const STRATEGY_NEW_LINE_FOR_CHAINED_CALLS = 'new_line_for_chained_calls';
+
+    /**
+     * @var string
+     */
+    private $character;
```
keradus: what is that character, semantically ?

julienfalque: It's the character that ends a method calls chain, i.e. `;` or `,`. I'll rename the property to `$chainedCallsEndCharacter`.

julienfalque: Property renamed.

src/AbstractMultilineWhitespaceBeforeChainedCallsEndCharacterFixer.php
keradus: if logic of 2 fixer is exactly the same but `$chainedCallsEndCharacter`, it proves that it's the very same fixer that shall have configuration - instead of having 2 artificial fixers

julienfalque: Shall I deprecate `multiline_whitespace_before_semicolons` in favor of a new fixer (e.g. `multiline_whitespace_before_chained_call_end_characters`) with a `characters` option?

keradus: is there a need for chained call indeed?
```php
$foo
    ->bar()
;
```

is not chained call ;)

julienfalque: True, I'll rename things, but that doesn't answer my question :p

keradus: I wanted to be sure first, as that was pretty damn big part of name ;)

yes, I would do the rename+proxy+deprecate trio

julienfalque: PR reworked.

README.rst
```diff
@@ -798,14 +798,27 @@ Choose from the list of available rules:
...
+
+  Forbid multi-line whitespace before a statement end (comma or semicolon)
+  or moves it to the next line for chained calls.
+
+  Configuration options:
+
+  - ``comma_strategy`` (``'new_line_for_chained_calls'``, ``'no_new_line'``, ``'none'``):
+    strategy to apply to comma; defaults to ``'no_new_line'``
+  - ``semicolon_strategy`` (``'new_line_for_chained_calls'``, ``'no_new_line'``,
+    ``'none'``): strategy to apply to semicolon; defaults to ``'no_new_line'``
```
keradus: for minimize diff size,
we like to have trailing comma and semicolons on next line.

Can we reflect that in default settings of this new rule, please?
```
            'comma_strategy' => 'no_new_line',
            'semicolon_strategy' => 'new_line_for_chained_calls',
```

Then, please add rule to local `.php_cs.dist` file

keradus:
OK, I got lost.
what's the difference between
`'no_whitespace_before_comma_in_array' => true`
and
`'multiline_whitespace_before_statement_end' => ['comma_strategy' => 'no_new_line', 'semicolon_strategy' => 'none']`

README.rst
julienfalque: Did you mean
```php
            'comma_strategy' => 'new_line_for_chained_calls',
            'semicolon_strategy' => 'new_line_for_chained_calls',
```
?

julienfalque:
```php
    'no_whitespace_before_comma_in_array' => true,
```
removes any whitespace whereas
```php
    'multiline_whitespace_before_statement_end' => ['comma_strategy' => 'no_new_line'],
```
removes any whitespace that contains a newline.

keradus:
OK, I found exactly that test.
My confusion go higher...

`'no_whitespace_before_comma_in_array' => true,`
and
`'multiline_whitespace_before_statement_end' => ['comma_strategy' => 'no_new_line'],`

will do the same for multiline whitespaces.... ?

README.rst
keradus: i got confused by multi vs single...

yes, I would opt for the `new_line_for_chained_calls` x2 as defaults, as it minimizes diff size

julienfalque: Done.

julienfalque:
Indeed. Note this is also true with the fixer being deprecated here:
```php
'no_whitespace_before_comma_in_array' => true,
```
and
```php
'multiline_whitespace_before_semicolons' => [
    'strategy' => 'no_multi_line',
],
```
have the same effect when whitespaces contain new lines.

I suggest to do the following in this PR:
- rename `multiline_whitespace_before_statement_end` to `whitespace_before_statement_end`;
- add a new `no_whitespace` strategy;
- deprecate `no_whitespace_before_comma_in_array` in favor of `whitespace_before_statement_end` with `no_whitespace` strategy.

WDYT?

keradus:
Would that one, unified rule allow for achieving this? (copy-pasted from itest):
```
--EXPECT--
<?php
return [
    'foo',
    'foo',
    $bar
        ->bar()
        ->baz()
    ,
];

--INPUT--
<?php
return [
    'foo'   ,
    'foo'
    ,
    $bar
        ->bar()
        ->baz(),
];
```

?

if so, I'm all for it :+1:

keradus:
ping @julienfalque

julienfalque:
Expect tests to fail for now.

So...

* `no_whitespace`: always removes spaces;
* `no_new_line`: pretty much the same but only if a newline is found;
* `new_line_for_chained_calls`: adds a new line if none is found *and* the expression is a method calls chain.

I'm thinking `no_new_line` strategy should be dropped because it's a bit weird: why would one want to entirely remove spaces if and only if they contain a new line? Maybe the initial intention of this strategy was the same as `no_whitespace`. But dropping it would be a BC break regarding the (now deprecated ) `multiline_whitespace_before_semicolons` rule.

Also, `new_line_for_chained_calls` does not *remove* spaces when the expression isn't a method call chain. IMO it should, but that would also be a BC break regarding `multiline_whitespace_before_semicolons`. I also think the name is not good: later we might want to cover other kinds of expressions like e.g.
```php
return
    $foo
    && $bar
;
```
or
```php
return
    $foo ?
    $bar :
    $baz
;
```
which are not chained method calls but should follow the same whitespace rule as with chained method calls.

I have no idea which direction to follow. Shall I find better strategy names? Allow using multiple strategies at once? Maybe use completely different options?

keradus:
what do you think about my proposal, @julienfalque  ?

keradus:
ping @julienfalque ?

SpacePossum:
ping @julienfalque :)

julienfalque:
No need to ping, I didn't forget about my WIP branches. I just don't have the time to finish those right now but I'll do it eventually :)

SpacePossum:
sorry to ping you man, take your time, great to hear it is still on your list :D

julienfalque:
PR reworked. I didn't apply the new fixer to the codebase yet to ease review so builds will fail but tests should pass.

I reverted proxying the new fixer in the deprecated ones because they don't have the exact same behavior anymore:
- `no_whitespace_before_comma_in_array`: the new fixer also removes whitespaces before commas in function calls (I don't think we already have a fixer for that which would have to be deprecated as well?)
- `multiline_whitespace_before_semicolons`: the new fixer has no equivalent option to `no_multi_line` (removing whitespaces only when they contain a newline).

This means there is some duplicated code now, espacially with `multiline_whitespace_before_semicolons` and the new `whitespace_before_statement_end`. Not sure how to refactor for now.

@keradus I applied your proposal but used `new_line_for_multiline_statement` instead of `diff_optimized` because it's not explicit enought about what it does IMO.

keradus:
@julienfalque , it's labeled as WIP, can you confirm is it still WIP or ready for review ?

julienfalque:
Honestly I'm not sure, it's quite old now :)

I think this part from my previous comment:

> This means there is some duplicated code now, espacially with `multiline_whitespace_before_semicolons` and the `new whitespace_before_statement_end`. Not sure how to refactor for now.

is still to be addressed, but maybe it's not a blocker. Looks like a rebase is needed too. Apart from that, the PR should be ready for review.

Once review process will be done, I'll run the fixer on the codebase so that builds pass.

kubawerlos:
Can `multiline_whitespace_before_semicolons` be a proxy to `whitespace_before_statement_end`?

julienfalque:
> Can `multiline_whitespace_before_semicolons` be a proxy to `whitespace_before_statement_end`?

@kubawerlos No, `multiline_whitespace_before_semicolons` only removes newlines inside whitespace tokens. The removal of singleline whitespaces is currently handled by `no_singleline_whitespace_before_semicolons` (which I forgot to deprecate the same way). The new `whitespace_before_statement_end` rule does both.

julienfalque:
Not sure how to make fabbot happy here, any idea?

keradus:
let's ignore fabbot here

src/Fixer/Whitespace/WhitespaceBeforeStatementEndFixer.php
```diff
@@ -0,0 +1,295 @@
...
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+final class WhitespaceBeforeStatementEndFixer extends AbstractFixer implements ConfigurableFixerInterface, WhitespacesAwareFixerInterface
+{
+    public function getDefinition(): FixerDefinitionInterface
+    {
+        return new FixerDefinition(
+            'Forbid multi-line whitespace before a statement end (comma or semicolon) or moves it to the next line for multiline statements.',
```
SpacePossum, line 35: Why isn't the multi line part not part of the rule name?
Why are only `,` and `;` considered statement ends and not `?>` and `]`, `}` etc.?

SpacePossum:
With 111 files changes, isn't this new rule the exception rather than the current CS? I mean, can we keep the current CS with the new rule or no?

src/Fixer/Whitespace/WhitespaceBeforeStatementEndFixer.php
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3679#issuecomment-830164657

> The new whitespace_before_statement_end rule does both.

hmmm now I'm confused, maybe the description is out dated?

Wirone:
@keradus @julienfalque do we still need/want this? It's massively outdated and introduces changes in the codebase that I don't like (as @SpacePossum I would rather want to keep current placing for commas in chained calls). Also, cognitive load in this PR is over 9000 😆. Maybe we should close it along with the related issue?

julienfalque:
This is still something I'd like to have control over when working on PHP code, but we can configure it to keep current style in this project specifically. I may rebase this PR but I already did that a lot of times so I'd like you to confirm it has a chance to be merged before I try to revice it, otherwise let's close it 😄

julienfalque:
Closing for now, I'll reopen if I can find some time and motivation to recover this.

PR #3680, opened 2018-04-05
DX: Tokens - removeLeadingWhitespace and removeTrailingWhitespace must act in same way

SpacePossum:
The difference in (not) taking multi line comments into account when clearing the space makes it very confusing.


SpacePossum:
(rebased and squashed only, no changes)

keradus:
Thank you @SpacePossum.

PR #3681, opened 2018-04-05
CiIntegrationTest - fix incompatibility from 2.x line

keradus:
on 2.x, by default line contains only "dots", eg 
```
..S.
```
on 3.x, by default line contains also % progress, eg 
```
..S.           (4/8) 50%
```


keradus:
FYI @kubawerlos , you updated file recently ;)

PR #3682, opened 2018-04-05
Add LogicalOperatorsShortVariationFixer

kubawerlos:
As never seen `and` or `or` used internationally.


SpacePossum:
solid idea, I think however it duplicates https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2577 , maybe something can be reused or a little team up can be done around this topic? cc @hkdobrev

kubawerlos:
Ok, commented there.

Issue #3683, opened 2018-04-06
php_unit_strict rule should not convert assertEquals method to assertSame when delta param is filled


tifabien:
#### The PHP version you are using (`$ php -v`):
=>PHP 7.1.12-3+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Dec 14 2017 15:37:13) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.12-3+ubuntu16.04.1+deb.sury.org+1, Copyright (c) 1999-2017, by Zend Technologies

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.11.1 Grey Devil by Fabien Potencier and Dariusz Ruminski

I'm having an issue with the php_unit_strict rule. Indeed when the delta parameter is filled in the assertEquals method, the rule converts the method into assertSame but both method signature don't match.
I think this rule should not touch asertEquals methods when delta is filled and should not raise an error.

What do you think?

keradus:
indeed.
could you raise a fix PR please ?

tifabien:
Of course no problem! I will not be available for the next days but I'll work on it after!

tifabien:
Ooops too slow to react... Sorry

SpacePossum:
no worries @tifabien :+1:

PR #3684, opened 2018-04-06
PhpUnitStrictFixer - Do not fix if not correct amount of argument are used.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3683


SpacePossum:
note to self for other PR for this fixer; we could/should also only fix (static)method calls within classes/traits we think are (part of)  PHPUnit classes

src/Fixer/PhpUnit/PhpUnitStrictFixer.php
```diff
@@ -116,6 +119,19 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 }
 
                 $sequenceIndexes = array_keys($sequence);
+
+                $argumentsCount = \count(
+                    $argumentsAnalyzer->getArguments(
```
julienfalque: I would use `countArguments` instead of `getArguments`.

keradus:
Thank you @SpacePossum.

Issue #3653, opened 2018-03-29
Require empty line after class definition and before it end


erickskrauch:
#### The PHP version you are using (`$ php -v`):
PHP 7.1.8 

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
PHP CS Fixer 2.11.1

#### The command you use to run PHP CS Fixer:
php-cs-fixer --diff --dry-run --stop-on-violation --using-cache=no -v fix

#### The configuration file you are using, if any:
```php
<?php
$finder = Symfony\Component\Finder\Finder::create()
    ->in(['api', 'common', 'console'])
    ->notPath('*/runtime')
    ->name('*.php')
    ->ignoreDotFiles(true)
    ->ignoreVCS(true);

return PhpCsFixer\Config::create()
    ->setRules([
        '@PSR2' => true,
        'braces' => [
            'allow_single_line_closure' => false,
            'position_after_functions_and_oop_constructs' => 'same',
        ],
        'blank_line_before_statement' => true,
        'cast_spaces' => [
            'space' => 'none',
        ],
        'linebreak_after_opening_tag' => true,
        'no_whitespace_in_blank_line' => true,
        'non_printable_character' => [
            'use_escape_sequences_in_strings' => true,
        ],
        'ordered_imports' => true,
        'single_quote' => true,
        'whitespace_after_comma_in_array' => true,
        'trailing_comma_in_multiline_array' => true,
    ])
    ->setRiskyAllowed(true)
    ->setFinder($finder);
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php
namespace common\rbac;

final class Roles {

    public const ACCOUNTS_WEB_USER = 'accounts_web_user';

}
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
   1) project-name/common/rbac/Roles.php (class_definition, braces)
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
 <?php
 namespace common\rbac;
 
 final class Roles {
-
     public const ACCOUNTS_WEB_USER = 'accounts_web_user';
-
 }
 

      ----------- end diff -----------
```

 * with the changes you expected instead:
```php
No changes
```

--------------------------

I want to have one blank line after defining the class and before closing it:

```php
class Test {

    // Any contents here

}
```

[...]

Issue #823, opened 2014-12-12
BracesFixer has too many responsibilities


ceeram:
For instance #797 is also caused by this fixer, which i would not expect. The braces fixer seems to also fix proper indentation for other things, which imo should be a separate fixer

keradus:
It may be easier to split if one fixer could make another one as dependency (can't run without it)

GrahamCampbell:
Yeh, that could make things easier.

keradus:
Issue is reported for 2 years. As it's valid, it's not an easy task to do. Closing to clean up the things.
Even after that, if one is feeling ready for the task - feel free to contribute !

julienfalque:
I'm reopening because this still causes a lot of bug reports/feature requests related to this.

julienfalque:
I've started working on this issue.

PR #3688, opened 2018-04-07
[WIP] Add BlankLineAroundClassBodyFixer

erickskrauch:
Implements https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3653.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -297,7 +297,7 @@ Choose from the list of available rules:
   Configuration options:
 
   - ``apply_to_anonymous_classes`` (``bool``): whether this fixer should be applied
-    to anonymous classes; defaults to ``false``
+    to anonymous classes; defaults to ``true``
   - ``blank_lines_count`` (``int``): adjusts an amount of the blank lines; defaults
     to ``1``
 
--- a/README.rst
+++ b/README.rst
@@ -289,6 +289,18 @@ Choose from the list of available rules:
   Ensure there is no code on the same line as the PHP open tag and it is
   followed by a blank line.
 
+* **blank_line_around_class_body**
+
+  Ensure that class body contains one blank line after class definition
+  and before its end.
+
+  Configuration options:
+
+  - ``apply_to_anonymous_classes`` (``bool``): whether this fixer should be applied
+    to anonymous classes; defaults to ``false``
+  - ``blank_lines_count`` (``int``): adjusts an amount of the blank lines; defaults
+    to ``1``
+
 * **blank_line_before_return**
 
   An empty line feed should precede a return statement. DEPRECATED: use
--- a/src/Fixer/ClassNotation/BlankLineAroundClassBodyFixer.php
+++ b/src/Fixer/ClassNotation/BlankLineAroundClassBodyFixer.php
@@ -134,7 +134,7 @@ new class extends Foo {
                 ->getOption(),
             (new FixerOptionBuilder('apply_to_anonymous_classes', 'whether this fixer should be applied to anonymous classes.'))
                 ->setAllowedTypes(['bool'])
-                ->setDefault(false)
+                ->setDefault(true)
                 ->getOption(),
         ]);
     }
--- a/src/Fixer/ClassNotation/BlankLineAroundClassBodyFixer.php
+++ b/src/Fixer/ClassNotation/BlankLineAroundClassBodyFixer.php
@@ -1,0 +1,162 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+use PhpCsFixer\Utils;
+
+/**
+ * @author ErickSkrauch <erickskrauch@ely.by>
+ */
+final class BlankLineAroundClassBodyFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface, WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Ensure that class body contains one blank line after class definition and before its end.',
+            [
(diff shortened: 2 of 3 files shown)


erickskrauch:
Should I really follow `fabbot.io` suggestions about replacing short array declaration via `array()`?

ntzm:
Fabbot looks like it's using the wrong rule set, think you can ignore it

ntzm:
Yes you'll have to use `@require PHP 5.6` on the test cases that require PHP 5.6

erickskrauch:
Meh, nope. I added `@requires` annotation, but it didn't help: the test is still executed. What am I doing wrong?

erickskrauch:
@ntzm thank you! I don't even know how I did not notice it 🙄

README.rst
```diff
@@ -289,6 +289,18 @@ Choose from the list of available rules:
...
 
+* **blank_line_around_class_body**
+
+  Ensure that class body contains one blank line after class definition
+  and before its end.
+
+  Configuration options:
+
+  - ``apply_to_anonymous_classes`` (``bool``): whether this fixer should be applied
+    to anonymous classes; defaults to ``false``
```
keradus:
```php
$classInstance = new class () {


   public function foo() { }



}
```

why allowing for not applying the change here ?

erickskrauch: Oh, I missed this. It should be `true` by default but must be configurable to support our code style (in anonymous classes we do not leave empty lines).

keradus: sounds messy. why different rule depends is it anonymous or not?
what's the case? could you show examples? I can imagine one-liner anon class that we don't want to break to multiple lines, but, especially with example I gave, why not fix it ?

erickskrauch: > why different rule depends is it anonymous or not?

Because it's common to us to have code like this:

```php
<?php
class TestAmqpComponent extends \common\components\RabbitMQ\Component {

    public function getConnection() {
        return new class extends \PhpAmqpLib\Connection\AbstractConnection {
            public function __construct(
                $user,
                $password,
                $vhost,
                $insist,
                $login_method,
                $login_response,
                $locale,
                \PhpAmqpLib\Wire\IO\AbstractIO $io,
                $heartbeat
            ) {
                // do nothing
            }
        };
    }

}
```

Or like this (sorry for a screenshot):



We don't make whitespace for anonymous classes 'cause it's mainly inlined in some another code and feels like a simple statement and not really class definition, so it should be inlined.

-------------

But as I said earlier this fixer should be applied to anonymous classes by default. Maybe better I should replace it with `anonymous_class_blank_lines_count`?

keradus: sounds better, to not have hack to opt the feature out for some cases, but simply has config for anon and config for non-anon classes.

SpacePossum:
This fixer is in conflict with some other fixers I think. On a more general topic, is this a wide accepted standard that we want to support in the project directly (i.e. maybe this would be better as a custom fixer for the project that wants this style)?

erickskrauch:
@SpacePossum we have already discussed it here https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3653#issuecomment-381408068. @keradus said, that such fixer with the more generic implementation can be useful to provide users possibility more variously configure their code style. I'm only awaiting his reply with [my latest suggestion](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3653#issuecomment-382183474).

keradus:
ha, I did gave the talk here, assuming the discussion moved from "issue" to "PR", now I see it's back on "issue" level...
well, in the end it doesnt need to be me - @SpacePossum , if you want, investigate discussion on "issue" level as well ;)

SpacePossum:
back on implementation level then ;)  
I think there is a conflict with the rules `no_extra_consecutive_blank_lines` and with the  `class_attributes_separation`  (I think)

kubawerlos:
What about changing default number of lines to `0` and deprecating `no_extra_consecutive_blank_lines`?

IllyaMoskvin:
Unfortunately fails with v2.15.1 due to #4232.

```
[Error]
        Call to undefined method PhpCsFixer\Utils::splitLines()
```

@SpacePossum I believe this also conflicts with `braces` (cf. #823). This might not be the best place to discuss this, but it feels like `braces` and `class_attributes_separation` are overreaching their scope by removing blank lines around class body. Shouldn't that behavior be confined to `no_extra_blank_lines`? Feels like a bug.

erickskrauch:
This PR has been in existence for over a year. I am ready to complete it, but only on the condition that it is really needed in this project and will be merged.

@IllyaMoskvin, If you are interested in this functionality as a separate library, [use our configuration](https://github.com/elyby/php-code-style#using-our-fixers).

SpacePossum:
```
This might not be the best place to discuss this, but it feels like braces and class_attributes_separation are overreaching their scope by removing blank lines around class body. Shouldn't that behavior be confined to no_extra_blank_lines? Feels like a bug.
```

`braces` I agree, this is coming back in multiple PR's and issue, so if someone could check if `no_extra_blank_lines` indeed does the same we can move forward removing it from the `braces` fixer. For now I think `class_attributes_separation` is correct.

On this PR, I didn't see this as part of a standard of a bigger community and I'm not sure this should be in the core. Maybe @julienfalque , @keradus and @kubawerlos can give a vote as well so we all know where we stand on this one, thanks!

kubawerlos:
I've already down-voted the issue it covers and didn't change my mind since then ;)

Seems very specific - like some silly rule to remove PHPStrom's auto generated header comments - and IMHO should be in the repo which needs that format for some reason.

kubawerlos:
@IllyaMoskvin to write a custom fixer all you need to do actually is implementing `FixerInterface`, adding it in config with `registerCustomFixers` and adding it to rules passed in `setRules`.

Using other custom fixers can be simple as [this](https://github.com/kubawerlos/php-cs-fixer-custom-fixers#usage).

The majority of fixers proposed as PR here are good and very good ideas, that a lot of people would want to use. Only some of them - like this one - are not followed by any community and - what's bother me more - conflicting with actually existing fixers - we have a fixer that removes the empty lines this one would add.

IllyaMoskvin:
@kubawerlos Thanks, I'm aware of that now. Let's add that to the readme.

If you search this repo for `registerCustomFixers`, you won't find any docs about the functionality. Even its PHPDoc block is lacking:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/37cc9c62f71a4cd49d9427aea729954f0ca35fe2/src/ConfigInterface.php#L105-L112

The only way users can find out about this functionality is by (1) digging through GitHub issues for threads like this, (2) reading through the code, where the docs are lacking on this topic, or (3) searching GitHub for repos with custom fixers. I've yet to find a tutorial about it online.

I'd bet $20 that 99.9% of PHP-CS-Fixer users will never know about this functionality because AFAICT, there is no mention of it in the readme. Which is a shame, because it's great to have this option.

> Only some of them - like this one - are not followed by any community and - what's bother me more - conflicting with actually existing fixers - we have a fixer that removes the empty lines this one would add.

Right. I see what you're saying, but I think there's no harm in making things more configurable. Why have a fixer that can only remove empty lines, when you could have a fixer that can specify the number of empty lines? I agree with [your comment above](#issuecomment-453542484).

shakaran:
@erickskrauch Could you rebase the pull request for avoid the conflicts? Thanks

erickskrauch:
@shakaran, the original issue [has been closed](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3653#issuecomment-609389991), so I have no reason to work on this PR anymore. If the project maintainers confirm that this PR can still be merged, then I'll return to work on it.

You can use [our package](https://github.com/elyby/php-code-style#elyblank_line_around_class_body) if you need this particular fixer.

julienfalque:
Closing as it was decided not to integrate this in the main codebase (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3653#issuecomment-609389991).

Issue #1686, opened 2016-01-19
PhpDoc: no extra empty lines


keradus:
Target: 1.x line

Code before:

``` php
<?php

/**
 * Description.
 *
 *
 * @var int
 */
$foo = 1;
```

Code after:

``` php
<?php

/**
 * Description.
 *
 * @var int
 */
$foo = 1;
```

GrahamCampbell:
We already have a fixer that nearly does that.

keradus:
Updated feature specification to not partially duplicate `PhpdocTrimFixer`

GrahamCampbell:
I didn't originally add a fixer like this on purpose because the integration testing across lots of php repos showed that sometimes multiple lines breaks where indeed wanted, say when using complex markdown or code examples inside the phpdoc.

keradus:
It should not touch content of part of docblock (like annotation/description/summary), but the space between them.
And even if we could not reach that - the fixer would be useful for some repos ;)

GrahamCampbell:
Sure. This should be quite easy to implement.

keradus:
```php
/**
 * Summary.
 *
 *
 * Description.
 *
 *
 * @var int
 *
 *
 * @return int
 */
$foo = 1;
```

Extra blank line shall be removed after Summary and after Description.
One between annotations is already removed by `phpdoc_separation`, probably it shall not do it anymore and it shall be done only by new fixer.

julienfalque:
Why reopening the issue?

SpacePossum:
because I think there are still cases where PhpDoc's have extra empty lines which can/should be removed. Example:

```php
final class ClassKeywordRemoveFixerTest extends AbstractFixerTestCase
{
    /**
     * @param string      $expected
     * @param null|string $input
     *
     *
     *
     *
     * @dataProvider provideFixCases
     */
    public function testFix($expected, $input = null)
    {
        $this->doTest($expected, $input);
    }
}
```

keradus:
would you be fine guys to trim any consecutive blank lines, not only ones after summary/description ?
(as one, single fixer, no config)

keradus:
ref #3802

PR #3691, opened 2018-04-07
Add PhpdocTrimAfterDescriptionFixer

nobuf:
Implements https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1686; a new fixer to remove extra blank lines after Summary and after Description in phpdocs.

```diff
 <?php

 /**
  * Description.
  *
- *
  * @var int
  */
 $foo = 1;
```

It expects to be used with phpdoc_trim, phpdoc_separation.

First pull request here :)

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
@@ -1,0 +1,193 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Phpdoc;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\DocBlock\ShortDescription;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Nobu Funaki <nobu.funaki@gmail.com>
+ */
+final class PhpdocTrimAfterDescriptionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Removes extra blank lines after summary and after description in PHPDoc.',
+            [
+                new CodeSample(
+                    '<?php
+/**
+ * Summary.
+ *
+ *
+ * Description.
+ *
+ *
+ * @param string $foo
+ */
+function fnc($foo) {}
+'
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            $doc = new DocBlock($token->getContent());
+            $summaryEnd = (new ShortDescription($doc))->getEnd();
+
+            if (null === $summaryEnd) {
+                continue;
+            }
+
+            $this->fixSummary($doc, $summaryEnd);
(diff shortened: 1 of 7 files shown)


src/DocBlock/DocBlock.php
```diff
@@ -162,6 +162,36 @@ public function getContent()
         return implode($this->lines);
     }
 
+    /**
+     * Find the line number of the line containing the end of the short
+     * description, if present.
+     *
+     * @return null|int
+     */
+    public function findShortDescriptionEnd()
```
keradus: For now, this class was exposing informations from docblock text "as is", this is the first method that actually provide some analysis logic. 
For that, I'm not feeling 100% sure that this class is proper place for it, especially as it means that we are extending our BC promise to end users who are using our repo as library.
I understand the need of extraction and for sure it's way better than copy-pasting between fixers, but please extract it to dedicated, standalone class.
We already had that discussion previously for tokenizer. We have `Tokens\Analyzer` namespace with "random", standalone analysis that are internal and which are not covered by any BC promise, while main `Tokens` class is. Similar approach could be done here. Dont be afraid of small classes ;)

tests/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixerTest.php
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * @author Nobu Funaki <nobu.funaki@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Phpdoc\PhpdocTrimAfterDescriptionFixer
+ */
+final class PhpdocTrimAfterDescriptionFixerTest extends AbstractFixerTestCase
+{
+    public function testFix()
```
keradus: could you please rearrange tests? 

do one method that do testing (so `testFix`), but with data provider instead of multiple testing methods.

your data provider would look like
```
return [
    'case foo' => [$expected1, $input1],
    'case bar' => [$expected2, $input2],
    ...,
];
```

src/DocBlock/DocBlock.php
nobuf: I get the backward compatibility concern. I've created `ShortDescription` class with `@internal` and `final` keyword.

src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
```diff
@@ -0,0 +1,159 @@
...
+            $summaryEnd = (new ShortDescription($doc))->getEnd();
+
+            if (null === $summaryEnd) {
+                continue;
+            }
+
+            $this->fixSummary($doc, $summaryEnd);
+            $this->fixDescription($doc, $summaryEnd);
+
+            $tokens[$index] = new Token([T_DOC_COMMENT, $doc->getContent()]);
```
SpacePossum: not sure, but shouldn't we check if the content remained the same so no new token has to be set? (cc @keradus )

src/DocBlock/DocBlock.php
SpacePossum: thanks, composition looks a lot better now :+1:  :)

src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/src/Tokenizer/Tokens.php#L314

src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+final class PhpdocTrimAfterDescriptionFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Removes extra blank lines after Summary and after Description in phpdocs.',
```
ntzm: Can `Summary` and `Description` be lowercase instead? And write `PHPDoc` instead of `phpdocs` thanks!

src/Fixer/Phpdoc/PhpdocTrimAfterDescriptionFixer.php
```diff
@@ -0,0 +1,159 @@
...
+        foreach ($doc->getLines() as $index => $line) {
+            if ($line->containsATag()) {
+                return $index;
+            }
+        }
+
+        return $index; // no Annotation, return the last line
+    }
+
+    private function reverseFindLastUsefulContent(DocBlock $doc, $from)
```
ntzm: to help understandability of the code, could you add docblocks describing the type of parameters on methods that aren't fully-typed like this one?

nobuf:
The fixer definition is updated, and `@param` are present in all methods.

Travis CI got failed, though maybe just the CI  wasn't stable? I can't trigger a rebuild.
Let me know if I should rebase against `master` and make this PR a single commit.

keradus:
Thank you @nobuf.

nobuf:
My pleasure. Cheers! @keradus

TomasVotruba:
How is this fixer called in 2.12? I cannot find it under this or similar names

TomasVotruba:
Here is it: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3802

PR #3692, opened 2018-04-08
DX: Optimize tests

julienfalque:
This PR reduces usage of properties in test classes or clean them after tests are run.

A few local tests gave the following results:

-- | Without changes | With changes | Delta
-- | ---- | ---- | ----
**Execution time** | ~23.5s | ~16.5s | -29.8%
**Memory usage** | 132.00MB | 124.00MB | -6.5%

### DIFF
--- a/tests/Test/AbstractFixerTestCase.php
+++ b/tests/Test/AbstractFixerTestCase.php
@@ -14,15 +14,12 @@ namespace PhpCsFixer\Tests\Test;
 
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
-use PhpCsFixer\FixerFactory;
 use PhpCsFixer\Linter\CachingLinter;
 use PhpCsFixer\Linter\Linter;
 use PhpCsFixer\Linter\LinterInterface;
-use PhpCsFixer\RuleSet;
 use PhpCsFixer\Tests\TestCase;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
-use PhpCsFixer\Utils;
 use Prophecy\Argument;
 
 /**
@@ -42,11 +39,6 @@ abstract class AbstractFixerTestCase extends TestCase
      */
     protected $fixer;
 
-    /**
-     * @var null|string
-     */
-    private $fixerClassName;
-
     protected function setUp()
     {
         parent::setUp();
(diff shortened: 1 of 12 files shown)


tests/Fixer/ControlStructure/NoUnneededControlParenthesesFixerTest.php
```diff
@@ -29,8 +29,6 @@
 
     public static function setUpBeforeClass()
     {
-        parent::setUpBeforeClass();
```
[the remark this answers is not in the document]
julienfalque: I don't expect PHPUnit itself to add anything in those methods. If anything is added, that would be in our own internal classes or libraries we manage, which means we would be aware when calling the parent methods becomes required so we could make the changes at that moment.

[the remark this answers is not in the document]
keradus: argument is valid against any parent class, doesn't matter is in phpunit or not, in general ;)
and then you would try to look up for any classes that has the method overwritten? looks like extra job to do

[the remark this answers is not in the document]
julienfalque: Reverted parent calls removal and added them where missing.

tests/Test/AbstractFixerTestCase.php
```diff
@@ -54,36 +46,22 @@ protected function setUp()
...
 
-    /**
-     * Create fixer factory with all needed fixers registered.
-     *
-     * @return FixerFactory
-     */
-    protected function createFixerFactory()
-    {
-        return FixerFactory::create()->registerBuiltInFixers();
+        $this->linter = null;
```
keradus: sometimes we follow `unset($this->foo)`, sometimes `$this->foo = null`.
any reason why to do it 2 different ways ?

julienfalque: I'm just used to clear properties by assigning `null` to them.

I was about to say that I'm going to use `unset()` everywhere but I checked something: https://3v4l.org/aNCUO. As you can see, `unset()` completely destroys properties just like it completely destroys regular variables, so I suggest using `$this->foo = null` everywhere instead.

keradus: Sounds like new Fixer....

SpacePossum: ` I suggest using $this->foo = null everywhere instead.` :+1:

keradus: No Julien, your changes are great. Just saying that new rule would be nice, so we would never do unset on property again ;)

SpacePossum: I think an exception is OK here, we've enough PR's to process as it :)

julienfalque: `unset()` calls on properties removed (actually only one found on 2.2, didn't check higher branches).

keradus: don't do it, we will do higher branches with new fixer

keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3695

tests/Test/AbstractFixerTestCase.php
```diff
@@ -54,36 +46,22 @@ protected function setUp()
...
-    protected function getFixerName()
+    protected function createFixer()
     {
-        $reflection = new \ReflectionClass($this);
+        $fixerClassName = preg_replace('/^(PhpCsFixer)\\\\Tests(\\\\.+)Test$/', '$1$2', get_called_class());
 
-        $name = preg_replace('/FixerTest$/', '', $reflection->getShortName());
-
-        return Utils::camelCaseToUnderscore($name);
+        return new $fixerClassName();
```
keradus: well, it works only because we follow convention super strictly, if one would create `FooFixer` that `getName` would return `bar_baz`, this implementation wouldn't work, while original one would remain to work

just pointing this out (if we ever hit that problem).

i'm fine with the change

julienfalque: Maybe better to add a (final?) `testGetName()` method in this abstract class?

keradus:
@julienfalque , please resolve conflicts

keradus:
Thank you @julienfalque.

PR #3693, opened 2018-04-09
Update readme for homebrew-php deprecation

fiskie:
The homebrew-php tap has been removed, but the php-cs-fixer formula was moved to homebrew-core. This PR addresses the change.


keradus:
Hi !
thanks for contributing, looks like valid change indeed ;)
Please, change target branch for 2.2 and check out contributing guidelines to figure out why tests are failing (hint: readme file is autogenerated, we don't change it manually)

keradus:
is it? I still see `homebrew-php` used there

PR #3698, opened 2018-04-14
YodaStyleFixer - Add always_move_variable option

### LINKED ISSUE #3153: YodaStyleFixer with function calls

julienfalque:
Closes #3153. Should be merged after #3674.

### DIFF
--- a/src/Fixer/ControlStructure/YodaStyleFixer.php
+++ b/src/Fixer/ControlStructure/YodaStyleFixer.php
@@ -21,6 +21,7 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
 
 /**
  * @author Bram Gotink <bram@gotink.me>
@@ -33,6 +34,11 @@ final class YodaStyleFixer extends AbstractFixer implements ConfigurationDefinit
      * @var array<int|string, Token>
      */
     private $candidatesMap;
+
+    /**
+     * @var array<string, null|bool>
+     */
+    private $candidateTypesConfiguration;
 
     /**
      * @var array<int|string>
@@ -76,6 +82,14 @@ final class YodaStyleFixer extends AbstractFixer implements ConfigurationDefinit
                         'less_and_greater' => null,
                     ]
                 ),
+                new CodeSample(
+                    '<?php
+return $foo === count($bar);
+',
+                    [
+                        'always_move_variable' => true,
+                    ]
+                ),
             ]
         );
     }
@@ -113,6 +127,10 @@ final class YodaStyleFixer extends AbstractFixer implements ConfigurationDefinit
             (new FixerOptionBuilder('less_and_greater', 'Style for less and greater than (`<`, `<=`, `>`, `>=`) statements.'))
                 ->setAllowedTypes(['bool', 'null'])
                 ->setDefault(null)
+                ->getOption(),
+            (new FixerOptionBuilder('always_move_variable', 'Whether variables should always be on non assignable side when applying Yoda style.'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
                 ->getOption(),
         ]);
     }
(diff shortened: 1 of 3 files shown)


src/Fixer/Phpdoc/PhpdocAlignFixer.php
```diff
@@ -228,7 +228,7 @@ private function fixDocBlock($content)
             // update
             foreach ($items as $j => $item) {
                 if (null === $item['tag']) {
-                    if ($item['desc'][0] === '@') {
+                    if ('@' === $item['desc'][0]) {
```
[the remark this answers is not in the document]
SpacePossum: I think this change is good and indeed shows the bug you are fixing in the other PR

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -114,6 +119,14 @@ protected function createConfigurationDefinition()
                 ->setAllowedTypes(['bool', 'null'])
                 ->setDefault(null)
                 ->getOption(),
+            (new FixerOptionBuilder('always_move_variable', 'Whether variables should always be on non assignable side when applying Yoda style.'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
```
julienfalque: By the way I would enable these options in our `.php_cs.dist` file and set their default value to `true` on 3.0 (or even drop the options). WDYT?

SpacePossum: I would like these new defaults, however I would not drop the options as it would stop SF from updating to 3.0 I think, but that is just me guessing.

julienfalque:
Rebased.

SpacePossum:
Please add a code sample showing how the new options work.

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -481,17 +506,35 @@ private function isOfLowerPrecedence(Token $token)
...
+        }
+
+        if ($strict) {
+            if ($tokens[$start]->equals('(')) {
+                return false;
+            }
+
+            for ($index = $start; $index <= $end; ++$index) {
+                if (
+                    $tokens[$index]->isGivenKind([T_POW, T_IS_SMALLER_OR_EQUAL, T_IS_GREATER_OR_EQUAL, T_SL, T_SR, T_INSTANCEOF, T_BOOL_CAST, T_OBJECT_CAST, T_ARRAY_CAST, T_STRING_CAST, T_DOUBLE_CAST, T_INT_CAST])
```
SpacePossum: can we use `TokensAnalyzer::isBinaryOperator` here?

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -598,7 +646,26 @@ private function isVariable(Tokens $tokens, $index, $end)
...
+        return !$this->isConstant($tokens, $start, $end);
+    }
+
+    private function isConstant(Tokens $tokens, $index, $end)
+    {
+        $token = $tokens[$index];
+
+        // -1 or -1.2
+        if ($index === $end - 1) {
+            return $token->equals('-') && $tokens[$end]->isGivenKind([T_LNUMBER, T_DNUMBER]);
```
SpacePossum: I think `$tokens[$end]` might a comment/PHPDoc here; `$a/**/=/**/-/**/1/**/;`

julienfalque:
> Please add a code sample showing how the new options work.

Done.

I also removed the `always_move_constant` option I introduced, I noticed it actually has no effect: constants are moved even without the option. Not sure why I introduced it. Maybe something fixed in upstream branches.

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -481,17 +506,35 @@ private function isOfLowerPrecedence(Token $token)
...
+    private function isVariable(Tokens $tokens, $start, $end, $strict)
     {
-        if ($end === $index) {
-            return $tokens[$index]->isGivenKind(T_VARIABLE);
+        if ($start === $end) {
+            return $tokens[$start]->isGivenKind(T_VARIABLE);
+        }
+
+        if ($strict) {
+            if ($tokens[$start]->equals('(')) {
```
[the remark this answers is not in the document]
SpacePossum: I was thinking about cases like `((( ($foo) === (1))))` or any construct that is supported within `$strict` with the meaning less extra `()` added. Maybe I miss understand, but the new options will not support fixing these cases by design?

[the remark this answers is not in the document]
SpacePossum: I think for `if (  (($foo))) === ((1)) )` it is called twice for `(($foo))` and `((1))` and not for `($foo)` and `(1)`, but I've to re-check TBH :)

keradus:
Thank you @julienfalque.

PR #3704, opened 2018-04-15
DX: Tokens - fix naming

keradus:
appendix for #3680
`save` vs `safe`


SpacePossum:
I'll never get the `v` vs. `f` rule, not with `safe` nor `life` or whatever, it a defect in my brain,
anyway, thanks for fixing!

PR #3709, opened 2018-04-17
Add SetTypeToCastFixer

SpacePossum:
```
$ php php-cs-fixer describe set_type_to_cast
Description of set_type_to_cast rule.
Cast shall be used, not `settype`.

Risky when the `settype` function is overridden.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,3 +1,3 @@
    <?php
   -settype($foo, "integer");
   -settype($bar, "string");
   -settype($bar, "null");
   +$foo = (int) $foo;
   +$bar = (string) $bar;
   +$bar = null;
   
```

might require some priority tweaking with other rules

### DIFF
--- a/src/Fixer/Alias/SetTypeToCastFixer.php
+++ b/src/Fixer/Alias/SetTypeToCastFixer.php
@@ -1,0 +1,251 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Alias;
+
+use PhpCsFixer\AbstractFunctionReferenceFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Analyzer\ArgumentsAnalyzer;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class SetTypeToCastFixer extends AbstractFunctionReferenceFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Cast shall be used, not `settype`.',
+            [
+                new CodeSample(
+                    '<?php
+settype($foo, "integer");
+settype($bar, "string");
+settype($bar, "null");
+'
+                ),
+            ],
+            null,
+            'Risky when the `settype` function is overridden or when used as the 2nd or 3rd expression in a `for` loop .'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_CONSTANT_ENCAPSED_STRING, T_STRING, T_VARIABLE]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $map = [
+            'array' => [T_ARRAY_CAST, '(array)'],
+            'bool' => [T_BOOL_CAST, '(bool)'],
+            'boolean' => [T_BOOL_CAST, '(bool)'],
+            'double' => [T_DOUBLE_CAST, '(float)'],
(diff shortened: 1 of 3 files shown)


src/Fixer/Alias/SetTypeToCastFixer.php
```diff
@@ -0,0 +1,223 @@
...
+        $secondArgumentStart,
+        $closeParenthesisIndex
+    ) {
+        $tokens->clearTokenAndMergeSurroundingWhitespace($closeParenthesisIndex);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($secondArgumentStart);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($commaIndex);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($openParenthesisIndex);
+        $tokens->clearTokenAndMergeSurroundingWhitespace($functionNameIndex);
+
+        $tokens->removeTrailingWhitespace(
```
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3680  ;)

tests/Fixer/Alias/SetTypeToCastFixerTest.php
```diff
@@ -0,0 +1,161 @@
...
+                '<?php A::settype($foo, "integer");',
+            ],
+            'member call' => [
+                '<?php $a->settype($foo, "integer");',
+            ],
+            'unknown type' => [
+                '<?php $a->settype($foo, "foo");',
+            ],
+            'no null cast' => [
+                '<?php $a->settype($foo, "null");',
```
ntzm: the above two are testing on member call when they shouldn't be

tests/Fixer/Alias/SetTypeToCastFixerTest.php
```diff
@@ -0,0 +1,161 @@
...
+                '<?php (float) $foo;',
+                '<?php settype($foo, "double");',
+            ],
+            [
+                '<?php (float) $foo;',
+                '<?php settype($foo, "float");',
+            ],
+            'int full caps' => [
+                '<?php (int) $foo;',
+                '<?php settype($foo, "INT");',
```
ntzm: should it be lower casing them? or is that a job for `lowercase_cast`?

SpacePossum: this is the input, new cast is always in lowercase

tests/Fixer/Alias/SetTypeToCastFixerTest.php
SpacePossum: copy-paste error :(  good catch!

keradus:
`settype` ? kerad is crying :(
where did you faced it ? :(

ntzm:
Seeing as

> Return Values ¶
> Returns TRUE on success or FALSE on failure.

I think we shouldn't fix if assigning somewhere

ntzm:
Also it modifies by reference, so surely should fix to `$foo = (array) $foo` instead of just `(array) $foo`?

SpacePossum:
>I think we shouldn't fix if assigning somewhere

I think the fixer doesn't `fix` if assigning

SpacePossum:
>Also it modifies by reference, so surely should fix to $foo = (array) $foo instead of just (array) $foo?

indeed, seems I lost a commit somewhere...

SpacePossum:
> where did you faced it :(

on one of those older code bases :(

Slamdunk:
Nice work and nice try, but the correct fix for [settype](https://secure.php.net/manual/en/function.settype.php) usage in PHP should be:
```diff
-settype($foo, "integer");
+settype($foo, "integer"); // Really?
+// Look at me, pay attention to my lips
+// and repeat after me: "I shall not mess with PHP typecasting"
+// Now, I'm good this time and I'll only add the next line for now
+// @TODO Refactor this typesystem mess
```
That should be enforced on a secon run by:

```diff
-settype($foo, "integer");
+throw new \Exception('settype($foo, "integer");'); // I TOLD YA!
```

keradus:
@SpacePossum , you removed `WIP` flag, but Travis is still sad. Could you make him happy?

SpacePossum:
I fixed the test, but than added `settype` `null` support, breaking the space control around the statements.
Hope to fix this later.

ntzm:
> or when used as the 2nd or 3rd expression in a `for` loop

I don't want to know how you found this out 😆

SpacePossum:
>I don't want to know how you found this out

consider this check

```php
            if (!$tokens[$prev]->isGivenKind(T_OPEN_TAG) && !$tokens[$prev]->equalsAny([';', '{'])) {
                continue; // return value of the function is used
            }
```

Therefore `for(; settype($a, 'int'); settype($a, 'int')) {}` both calls are fixed, changing the meaning of the code. It is just such an edge case I think it is not worth checking ;)

keradus:
is there anything left here ?

keradus:
Thank you @SpacePossum.

PR #3713, opened 2018-04-20
Use HTTPS whenever possible

fabpot:


README.rst
```diff
@@ -40,7 +40,7 @@ your system:
 
 .. code-block:: bash
 
-    $ wget http://cs.sensiolabs.org/download/php-cs-fixer-v2.phar -O php-cs-fixer
+    $ wget https://cs.sensiolabs.org/download/php-cs-fixer-v2.phar -O php-cs-fixer
```
fabpot: We might not want to use HTTPS. WDYT?

julienfalque: Why would we want to _not_ use HTTPS?

SpacePossum:
Thanks, awesome :) 
Great to see the releases on `cs.sensiolabs.org` can be fetch over HTTPS.

Practical question at the maintainers, do we want the changes on `2.2` as well (`1.13` still exists as branch but is EOL so doesn't matter to much I think)?

julienfalque:
I'd say 2.2 for the README changes and 2.11 for `TypeAnalysis` as it does not exist in branch 2.2.

README.rst
keradus: well, we wanted HTTPS since beginning, simply https wasn't working properly on this subdomain originally

keradus:
thanks for pointing this out @fabpot !

as a side effect, you can see your PR is failing on fabbot, which is asking to use long-array syntax, even if we have config to use short syntax. Can you take a look?

README.rst
julienfalque: I know, but now it does and I see no reason not to use it, hence my question :)

src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
```diff
@@ -22,9 +22,9 @@
      *
      * More info:
      *
-     * @see http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.types
-     * @see http://php.net/manual/en/reserved.other-reserved-words.php
-     * @see http://php.net/manual/en/language.pseudo-types.php
+     * @see https://secure.php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration.types
```
carusogabriel: We can remove the `secure` subdomain

keradus:
Thank you @fabpot.

PR #3717, opened 2018-04-22
Test Cache::toJson throws exception on invalid data

ntzm:


src/Cache/Cache.php
```diff
@@ -79,7 +79,7 @@ public function toJson()
             'hashes' => $this->hashes,
         ));
 
-        if (false === $json) {
+        if (null === $json || false === $json) {
```
julienfalque: Is `json_encode()` really returning `null` in case of failure?

tests/Cache/CacheTest.php
```diff
@@ -194,6 +194,25 @@ public function provideCanConvertToAndFromJsonCases()
...
+    {
+        $this->setExpectedException(
+            'UnexpectedValueException',
+            'Can not encode cache signature to JSON, error: "Malformed UTF-8 characters, possibly incorrectly encoded". If you have non-UTF8 chars in your signature, like in license for `header_comment`, consider enabling `ext-mbstring` or install `symfony/polyfill-mbstring`.'
+        );
+
+        $invalidUtf8Sequence = "\xB1\x31";
+
+        $signature = $this->prophesize('PhpCsFixer\Cache\SignatureInterface');
+        $signature->getPhpVersion()->willReturn($invalidUtf8Sequence);
```
julienfalque: I would rather use the invalid UTF-8 sequence in `getRules()` as it is more likely to return such value.

src/Cache/Cache.php
ntzm: Didn't read the docs correctly, updated for a better solution

tests/Cache/CacheTest.php
```diff
@@ -194,6 +194,27 @@ public function provideCanConvertToAndFromJsonCases()
         );
     }
 
+    public function testToJsonThrowsExceptionOnInvalid()
+    {
+        $this->setExpectedException(
```
julienfalque: Please set this expectation right before calling `toJson()`.

keradus:
Thanks @ntzm !
Merged manually into #3723

PR #3718, opened 2018-04-22
Add missing FixerFactory tests

ntzm:


tests/FixerFactoryTest.php
```diff
@@ -250,6 +251,82 @@ public function testNoDoubleConflictReporting()
         );
     }
 
+    public function testSetWhitespacesConfig()
+    {
+        $factory = new FixerFactory();
+        $config = new WhitespacesFixerConfig();
+
+        $fixer = $this->prophesize('PhpCsFixer\Fixer\WhitespacesAwareFixerInterface');
+        $fixer->getName()->willReturn('foo');
```
julienfalque: This method isn't called so you don't need to define it.

tests/FixerFactoryTest.php
```diff
@@ -250,6 +251,82 @@ public function testNoDoubleConflictReporting()
...
+        $fixer->setWhitespacesConfig($config)->shouldBeCalled();
+
+        $factory->registerFixer($fixer->reveal(), false);
+
+        $factory->setWhitespacesConfig($config);
+    }
+
+    public function testRegisterFixerInvalidName()
+    {
+        $this->setExpectedException('UnexpectedValueException', 'Fixer named "0" has invalid name.');
```
julienfalque: The exception expectation should be set as late as possible before executing the code the code that is supposed to throw it. This will make the test fail if the exception is thrown before the expectation is set. I know it's very unlikely here but I think it's a good practice that we should always follow :)

tests/FixerFactoryTest.php
```diff
@@ -250,6 +251,82 @@ public function testNoDoubleConflictReporting()
...
+
+        $factory = new FixerFactory();
+
+        $fixer = $this->prophesize('PhpCsFixer\Fixer\ConfigurableFixerInterface');
+        $fixer->getName()->willReturn('foo');
+
+        $factory->registerFixer($fixer->reveal(), false);
+
+        $factory->useRuleSet(new RuleSet(array(
+            'foo' => 'bar',
```
julienfalque: You could use a data provider to test more values (e.g. `null`, `array()`, ...).

tests/FixerFactoryTest.php
```diff
@@ -250,6 +251,82 @@ public function testNoDoubleConflictReporting()
...
+        )));
+    }
+
+    public function testConfigurableFixerIsConfigured()
+    {
+        $fixer = $fixer = $this->prophesize('PhpCsFixer\Fixer\ConfigurableFixerInterface');
```
julienfalque: `$fixer` is assigned twice.

tests/FixerFactoryTest.php
ntzm: If I don't define it I get

```
1) PhpCsFixer\Tests\FixerFactoryTest::testSetWhitespacesConfig
Prophecy\Exception\Call\UnexpectedCallException: Method call:
  - getName()
on Double\WhitespacesAwareFixerInterface\P19 was not expected, expected calls were:
  - setWhitespacesConfig(exact(PhpCsFixer\WhitespacesFixerConfig:00000000010848210000000006cade73 Object (
    'indent' => '    '
    'lineEnding' => '
'
)))
```

julienfalque: Ah yes, I read the code too fast.

tests/FixerFactoryTest.php
ntzm: I can't test anything falsey here as it gets filtered out here https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/RuleSet.php#L294, not sure if that's the intended behaviour or not

julienfalque: I can't say whether this is intended either, @SpacePossum @keradus do you have an idea?

SpacePossum: if you comment out that line maybe a test will tell, but I think it is by design to remove rules set to `false`, for example to remove a rule in the result set that got it in through inheritance but set to false by child. That besides `false` others like `null` are filtered as well might be an implementation detail though.

keradus: that is exactly what the comment above that line states for:
>  filter out all resolvedRules that are off

only rules configured to `true | array` are supposed to survive

ntzm: An empty array also gets filtered out here

keradus: empty array is invalid configuration

keradus:
Thanks @ntzm !
Merged manually into #3723

tests/FixerFactoryTest.php
julienfalque: Those values are invalid but currently most are silently ignored instead of throwing an exception. I wonder if this behavior is intended.

PR #3719, opened 2018-04-22
Add more custom token tests

ntzm:


keradus:
Thanks @ntzm !
Merged manually into #3723

PR #3722, opened 2018-04-22
Add tests for CachingLinter

ntzm:


tests/Linter/CachingLinterTest.php
```diff
@@ -0,0 +1,88 @@
...
+            'bar.php' => '<?php echo "baz";',
+            'baz.php' => '<?php echo "foobarbaz";',
+        ));
+
+        $result = $this->prophesize('PhpCsFixer\Linter\LintingResultInterface');
+
+        $sublinter = $this->prophesize('PhpCsFixer\Linter\LinterInterface');
+        $sublinter->lintFile($fs->url().'/foo.php')->shouldBeCalledTimes(1)->willReturn($result->reveal());
+        $sublinter->lintFile($fs->url().'/bar.php')->shouldNotBeCalled();
+        $sublinter->lintFile($fs->url().'/baz.php')->shouldBeCalled()->willReturn($result->reveal());
```
keradus: why we count number of calls to foo and bar, but not baz ?

tests/Linter/CachingLinterTest.php
```diff
@@ -0,0 +1,88 @@
...
+
+        $linter = new CachingLinter($sublinter->reveal());
+
+        $linter->lintFile($fs->url().'/foo.php');
```
julienfalque: Please add assertions on the result of `lintFile()` calls here and `lintSource()` in the following test.

tests/Linter/CachingLinterTest.php
ntzm: Will update

tests/Linter/CachingLinterTest.php
ntzm: Will do, thanks 👍

tests/Linter/CachingLinterTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $linter = new CachingLinter($sublinter->reveal());
+
+        $results = array(
+            $linter->lintFile($fs->url().'/foo.php'),
+            $linter->lintFile($fs->url().'/foo.php'),
+            $linter->lintFile($fs->url().'/bar.php'),
+            $linter->lintFile($fs->url().'/baz.php'),
+        );
+
+        $this->assertContainsOnlyInstancesOf('PhpCsFixer\Linter\LintingResultInterface', $results);
```
julienfalque: This assertion is nice but not sufficient: the `CachingLinter` may return any instance of `LintingResultInterface` without making the assertion fail. You should check that each call returns `$result`, which would prove the `CachingLinter` just returns the result provided by the decorated `$sublinter`.

Maybe that would require to mock several results though, as you call `lintFile()` on different files.

ntzm: Thanks, will update

keradus:
Thanks @ntzm !
Merged manually into #3723

PR #3723, opened 2018-04-22
Extend tests coverage

ntzm:


ntzm:
Thanks @keradus, in the future should I group new tests and stuff into one PR?

keradus:
for cleaning/refactoring, yeah. (for features/bugfixes - nope)
even if this PR will collect other 10 (or so), it will still be smaller than most of the PRs

tests/FileRemovalTest.php
```diff
@@ -0,0 +1,93 @@
...
+use PhpCsFixer\FileRemoval;
+
+/**
+ * @author ntzm
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\FileRemoval
+ */
+final class FileRemovalTest extends TestCase
```
keradus: Whole idea of `FileRemoval` is to still remove the file even if php process crashed and destructor was not called - we do it with shutdown function.
Not sure how to test this...

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * @author ntzm
+ *
+ * @internal
+ *
+ * @see \PhpCsFixer\Tests\FileReaderTest::testReadStdin
+ *
+ * A stream wrapper class that pretends to be php://stdin, and returns:
+ * - <?php echo "foo";
+ * - <?php echo "bar";
```
keradus: this is not how `php://stdin` will behave for us:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.11/src/StdinFileInfo.php#L34

tests/FileRemovalTest.php
julienfalque: What about [`@runInSeparateProcess`](https://phpunit.readthedocs.io/en/7.1/annotations.html#runinseparateprocess)? Maybe we could have one test in a separate process and others depending on it so they are run after and can check if the file was actually removed after the first test? I have no idea whether this is even possible though :p

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
ntzm: How else to test that it caches?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.2/src/FileReader.php#L56-L600

tests/FileReaderTest.php
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @covers \PhpCsFixer\FileReader
+ */
+final class FileReaderTest extends TestCase
+{
+    public function testCreateSingleton()
+    {
+        $instance = FileReader::createSingleton();
+
+        $this->assertSame($instance, FileReader::createSingleton());
```
julienfalque: You could add an `assertInstanceOf()` assertion on `$instance`.

tests/FileReaderTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $this->assertSame('<?php echo "hi";', $reader->read($fs->url().'/foo.php'));
+    }
+
+    public function testReadStdinCaches()
+    {
+        $reader = new FileReader();
+
+        stream_wrapper_unregister('php');
+        stream_wrapper_register('php', 'PhpCsFixer\Tests\Fixtures\Test\FileReaderTest\StdinFakeStream');
```
julienfalque: Nice trick!

tests/FileReaderTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $this->assertSame('<?php echo "foo";', $reader->read('php://stdin'));
+        $this->assertSame('<?php echo "foo";', $reader->read('php://stdin'));
+
+        stream_wrapper_restore('php');
```
julienfalque: This should be done in `tearDownAfterClass()` in case an assertion fails (i.e. throws an exception).

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
julienfalque: I knew that about POST requests in HTTP. So this is true about CLI as well?

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
```diff
@@ -0,0 +1,67 @@
...
+    private $contents = array(
+        '<?php echo "foo";',
+        '<?php echo "bar";',
+    );
+
+    private static $contentsIndex = 0;
+
+    private $hasRead = false;
+
+    public function stream_open($path)
```
keradus: it's following some magical interface, isn't it?

julienfalque: Yep, magical: you must follow [streamWrapper class prototype](https://secure.php.net/manual/en/class.streamwrapper.php).

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
ntzm: Ah OK I get you now, will update, thanks

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
keradus: maybe nice to link it in the class

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
julienfalque: Woops, I misread the doc: it tells `php://input` is not (always?) seekable prior to PHP 5.6 but I couldn't find documentation regarding `php://stdin` not being seekable.

keradus:
```
ker@dus:~/github/FriendsOfPHP/PHP-CS-Fixer λ cat x.php 
<?php

var_dump([
    file_get_contents('php://stdin'),
    file_get_contents('php://stdin'),
]);
ker@dus:~/github/FriendsOfPHP/PHP-CS-Fixer λ echo "foo" | php -f x.php
array(2) {
  [0]=>
  string(4) "foo
"
  [1]=>
  string(0) ""
}
```

ntzm: Seems to be seekable in all PHP 5+ versions https://3v4l.org/PoJlP

keradus: well, even iin your link,
2nd run on stdin is seekable under 5.6 but not 5.5.
also, `file_get_contents` is not rewind, eg on 7.0

julienfalque: That means this custom stream wrapper class could be dropped on 2.11 branch, right?

julienfalque: Nope, `php://stdin` does not seem to be seekable at all on PHP 7.2:

```console
 $ cat test.php
<?php

$stdin1 = fread(STDIN, 3);
fseek(STDIN, 0);
$stdin2 = fread(STDIN, 3);

var_dump($stdin1, $stdin2);

 $ echo "foo" | php test.php
PHP Warning:  fseek(): stream does not support seeking in test.php on line 4
string(3) "foo"
string(1) "
"
```

ntzm: Weird.. So the behaviour wanted is after returning the content just to return an empty string forever?

tests/FileRemovalTest.php
ntzm: How's that solution?

julienfalque: I was rather thinking about creating a real file in e.g. `/tmp` and observe it without explicitly ending the PHP separate process to let PHPUnit run normally (no `exit`); and then have another test in the main process that checks that the real file does not exist anymore. Maybe this could be done using the [`@depends` annotation](https://phpunit.readthedocs.io/en/7.1/annotations.html#depends).

ntzm: Nearly got it working like that, just one thing:
Because the initial test (create file, observe file) runs in a separate process, it invokes tearDownAfterClass after it finishes. I'm using tearDownAfterClass to clean up the created file in case an assertion fails, so the files are deleted by tearDownAfterClass instead of FileRemoval. Any ideas?

julienfalque: Maybe a static property like `$removeFileOnTearDown` set to `true` by default, and set to `false` in the test that creates and observes the file: this way you would know in the `tearDownAfterClass` method if you can delete the file.

ntzm: Perfect, thanks a lot!

tests/Fixtures/Test/FileReaderTest/StdinFakeStream.php
keradus: the behaviour wanted in php? no, but it works like that
the behaviour wanted in mock? yes, to mimic php one

tests/FileRemovalTest.php
```diff
@@ -0,0 +1,131 @@
...
+{
+    private static $removeFilesOnTearDown = true;
```
keradus: just by reading the code, one may thing that as soon as `testShutdownRemovesObservedFilesSetup` is executed then this flag is set to false, thus we would never cleanup those files.
IMO it's worth to document the background between this variable and correlation with `testShutdownRemovesObservedFilesSetup`.

Solution works like a charm, but is super counter intuitive, at least for me.

keradus:
damn. fabbot is wrongly suggesting to fix paths that are excluded in config.
cc @fabpot

tests/FileRemovalTest.php
ntzm: Yeah I'm not 100% happy with it, but I can't figure out a better way of doing it, it feels like we're abusing the `@runInSeparateProcess` a little. Will document 👍

tests/FileReaderTest.php
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * @author ntzm
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\FileReader
+ */
+final class FileReaderTest extends TestCase
+{
+    public static function tearDownAfterClass()
```
ntzm: think this should be on tearDown in case someone adds more tests after `testReadStdinCaches`

tests/FileRemovalTest.php
keradus: exactly. and especially for that, docu is nice to have ;)

tests/FileReaderTest.php
keradus: well, `testA` creates file, then we want to drop file after `testB`.
`testC`, if it would be ever crated, has nothing to do with that file, and moving cleaning up to `tearDown` would mean it's executed after each tests, so either wasting CPU or we need more fancy logic to determine when it shall do the cleanup and when it shall not

keradus: i'm fine with `tearDownAfterClass` solution, let us just document it (as we just aligned) ;)

keradus:
Thank you @ntzm.

PR #3724, opened 2018-04-23
BlankLineBeforeStatementFixer - Add case and default as options

dmvdbrugge:
Includes testcases. Also lower priority, because now it needs to run after NoExtraBlankLinesFixer as that removes possible whitespace this fixer tries to add. Which is the reason these additions are wanted.

### DIFF
--- a/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
+++ b/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
@@ -166,6 +166,47 @@ while (true) {
     }
 
     /**
+     * @dataProvider provideFixWithCaseCases
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFixWithCase($expected, $input = null)
+    {
+        $this->fixer->configure([
+            'statements' => ['case'],
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixWithCaseCases()
+    {
+        return [
+            [
+                '<?php
+switch ($a) {
+    case 1:
+        return 1;
+
+    case 2:
+        return 2;
+}',
+                '<?php
+switch ($a) {
+    case 1:
+        return 1;
+    case 2:
+        return 2;
+}',
+            ],
+        ];
+    }
+
+    /**
      * @dataProvider provideFixWithContinueCases
      *
      * @param string      $expected
@@ -302,6 +343,47 @@ do {
 } while (true);
 $foo = "bar";
 declare(ticks=1);',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideFixWithDefaultCases
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFixWithDefault($expected, $input = null)
+    {
+        $this->fixer->configure([
+            'statements' => ['default'],
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixWithDefaultCases()
+    {
+        return [
+            [
+                '<?php
+switch ($a) {
+    case 1:
+        return 1;
+
+    default:
+        return 2;
+}',
+                '<?php
+switch ($a) {
+    case 1:
+        return 1;
+    default:
+        return 2;
+}',
             ],
         ];
     }
@@ -1228,7 +1310,7 @@ function foo() {
 
     public function provideFixWithMultipleConfigStatementsCases()
     {
-        $allStatements = [
+        $statementsWithoutCaseOrDefault = [
             'break',
             'continue',
             'declare',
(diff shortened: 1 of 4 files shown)


SpacePossum:
Hi @dmvdbrugge and thanks for your PR :)
I like the idea! Could you update the FixerFactory test (in the autoreview namespace) together with adding a priority test fixture to provide a way to test the new priority change?

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
```diff
@@ -306,6 +357,57 @@ public function provideFixWithDeclareCases()
...
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixWithDefaultCases()
+    {
+        return [
+            [
+                '<?php
```
SpacePossum: This case is doubled and therefore overcomplete. This is because in your next test the `expected` result is the same as this one. When you test with both `input` and `expected` not only will it test that after the `input` is fixed it is changed into the `expected` result, but it also tests that the `expected` result is not changed when provided to the fixer. Two birds one stone ;)

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
```diff
@@ -165,6 +165,57 @@ public function provideFixWithBreakCases()
...
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixWithCaseCases()
+    {
+        return [
+            [
+                '<?php
```
SpacePossum: same double thingy, see comment below

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
dmvdbrugge: That's some awesome magic 😉 removed.

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
dmvdbrugge: Removed as well

dmvdbrugge:
> Could you update the FixerFactory test (...) together with adding a priority test fixture

Done.

> Please add a test with both a default and multiple case statements

Done.

> test that the default behavior of the fixer is not changed (so no BC break)

This was already guaranteed by one of the existing tests, but I extended it so that future changes still guarantee no BC break.

SpacePossum:
thanks @dmvdbrugge :)

dmvdbrugge:
So I guess the state of this PR is "waiting to be mergeable because it's not up-to-date with base"?

Do you want me to merge base in, or rebase on top of it?

SpacePossum:
You need one more vote of a member/maintainer, but I think you'll get it ;)

Besides that keradus will need some time to merge PR's and make releases. This might take some time based on his agenda as he typically merges a whole bunch of things and releases it all in one go.

Till than, feel free to rebase or merge in the changes from the base (as merge-commits will be removed during the merging of the PR anyway)  :)

dmvdbrugge:
> You need one more vote of a member/maintainer

Ah, I had not seen that requirement anywhere. Might be an idea to add that to CONTRIBUTING.md? (Or am I just blind?)

> This might take some time based on his agenda

No hurry, just wondering :)

> Till than, feel free to rebase or merge in the changes from the base (as merge-commits will be removed during the merging of the PR anyway) :)

Ok, so I rebased (and again: nice magic)

Thanks for the answers! This is my first time contributing to a repo that has such strict rules 😆

keradus:
> Ah, I had not seen that requirement anywhere. Might be an idea to add that to CONTRIBUTING.md? (Or am I just blind?)

it's not hard rule. Like, if thing is trivial, one approval is enough.
If not, we like to have 2 of them.
As ppl are often running PHP CS Fixer blindly, trusting it won't break their code, we want to double check PRs to prevent that happening.


> This is my first time contributing to a repo that has such strict rules 

could you elaborate please? what strictness you are referring to? tests in code, 2nd approval, or... ?

keradus:
Thank you @dmvdbrugge.

keradus:
one more magic post mortem @dmvdbrugge , I automatically squashed (and pushed back to your fork) all commits of your PR ;)

SpacePossum:
thanks @dmvdbrugge :+1: 

and thanks @keradus for the black magic : D

dmvdbrugge:
> > This is my first time contributing to a repo that has such strict rules
>
> could you elaborate please? what strictness you are referring to? tests in code, 2nd approval, or... ?

I ment it in a positive way, but ye the 2nd approval, the vast amount of tests and how they work, a contributing.md that is not just "PR's welcome" but has actual requirements, 4 different build processes (even though 1 is misbehaving currently 😆).

Maybe strict wasn't the best choice of words, but again I ment it positive.
Thanks for the feedback and all the magic. Whooo, contributor now 🎉 😉

keradus:
Great to hear ;)
Now, we are waiting for more :D

Issue #3727, opened 2018-04-28
InvalidArgumentException when using phpdoc_indent on huge file


ntzm:
For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

#### The PHP version you are using (`$ php -v`):
=> PHP 7.2.4

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.11.1

#### The command you use to run PHP CS Fixer:
=> vendor/bin/php-cs-fixer fix yaf/yaf_namespace.php --rules phpdoc_indent -vvv

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):

https://github.com/JetBrains/phpstorm-stubs/blob/master/yaf/yaf_namespace.php

I get:

```
Loaded config default.
Using cache file ".php_cs.cache".
E
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error

Fixed all files in 0.071 seconds, 12.000 MB memory used

Files that were not fixed due to errors reported during fixing:
   1) ~/Code/phpstorm-stubs/yaf/yaf_namespace.php


        [InvalidArgumentException]
        Id must be an int, got "NULL".


      PhpCsFixer\Tokenizer\Token->__construct()
        in ~/Code/phpstorm-stubs/vendor/friendsofphp/php-cs-fixer/src/Fixer/Phpdoc/PhpdocIndentFixer.php at line 108
      PhpCsFixer\Fixer\Phpdoc\PhpdocIndentFixer->applyFix()
        in ~/Code/phpstorm-stubs/vendor/friendsofphp/php-cs-fixer/src/AbstractFixer.php at line 73
      PhpCsFixer\AbstractFixer->fix()
        in ~/Code/phpstorm-stubs/vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php at line 190
      PhpCsFixer\Runner\Runner->fixFile()
        in ~/Code/phpstorm-stubs/vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php at line 132
      PhpCsFixer\Runner\Runner->fix()
        in ~/Code/phpstorm-stubs/vendor/friendsofphp/php-cs-fixer/src/Console/Command/FixCommand.php at line 220
[...]

PR #3731, opened 2018-05-03
PhpdocIndentFixer -  crash fix

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3727

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocIndentFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocIndentFixer.php
@@ -105,7 +105,11 @@ class DocBlocks
 
             $newPrevContent = $this->fixWhitespaceBeforeDocblock($prevToken->getContent(), $indent);
             if ($newPrevContent) {
-                $tokens[$prevIndex] = new Token(array($prevToken->getId(), $newPrevContent));
+                if ($prevToken->isArray()) {
+                    $tokens[$prevIndex] = new Token(array($prevToken->getId(), $newPrevContent));
+                } else {
+                    $tokens[$prevIndex] = new Token($newPrevContent);
+                }
             } else {
                 $tokens->clearAt($prevIndex);
             }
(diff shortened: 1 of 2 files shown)


src/Fixer/Phpdoc/PhpdocIndentFixer.php
```diff
@@ -105,7 +105,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
 
             $newPrevContent = $this->fixWhitespaceBeforeDocblock($prevToken->getContent(), $indent);
             if ($newPrevContent) {
-                $tokens[$prevIndex] = new Token(array($prevToken->getId(), $newPrevContent));
+                $tokens[$prevIndex]->setContent($newPrevContent);
```
SpacePossum: not 100% sure about this fix because I don't know why a new Token was created here

tests/Fixer/Phpdoc/PhpdocIndentFixerTest.php
```diff
@@ -38,7 +38,27 @@ public function provideFixIndentCases()
...
+class Dispatcher
+{
+}
+',
+        );
+
+        $expected = $input = "<?php\n";
+        $template = "%s/** */\nfunction f%d() {}\n";
+
+        for ($i = 0; $i < 10; ++$i) {
```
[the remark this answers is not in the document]
julienfalque: I don't find this test very relevant and I would remove it.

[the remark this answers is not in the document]
SpacePossum: done :+1:

src/Fixer/Phpdoc/PhpdocIndentFixer.php
SpacePossum: hmmm... fair enough :)
do you have a suggestion how to proceed here? I'm not sure what would be best

julienfalque: I think you have to check whether `$prevToken->getId()` is null and only pass the content as parameter of `Token` constructor in such case instead of an array (id + content).

keradus:
Thank you @SpacePossum.

PR #3733, opened 2018-05-06
Disable Composer optimized autoloader by default

julienfalque:
Using the optimized autoloader of Composer in a dev environment has a drawback: if you rename or drop a class, you have to dump the autoloader again, which is a bit cumbersome. If you don't know that or forget about it, it becomes painful to debug.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -34,7 +34,7 @@ install:
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --no-interaction --no-progress --prefer-stable --no-ansi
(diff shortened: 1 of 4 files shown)


keradus:
cc @localheinz ,
I know you are using this flag as well. Did you experienced problems with it ?

SpacePossum:
There is no need, however there should be performance gain. With the changes in this PR, when someone uses requires the project with composer it wouldn't get the optimized class map generated for it by default would it?

julienfalque:
This will have no impact on projects that require `friendsofphp/php-cs-fixer` because [`config` is a root-only property of the composer.json schema](https://getcomposer.org/doc/04-schema.md#config), which means it is only taken into account inside the project where it is defined like e.g. `require-dev`.

keradus:
Thank you @julienfalque.

Issue #2979, opened 2017-08-17
[RFC] Remove @param and @return when they don't provide new informations


michaelcontento:
**Description:**

I'd like to propose the idea of a "strip down useless PHPDoc"-fixer that:
- Removes `@param` records when they don't provide additional informations
- Removes `@return` records when they don't provide additional informations

**Before:**

```php
<?php

class Example
{
    /**
     * @param \stdClass $obj
     * @return \stdClass
     */
    public function canBeRemoved(\stdClass $obj): \stdClass
    {
    }

    /**
     * @param \stdClass|array $mixed
     * @return \stdClass
     */
    public function canNotBeRemoved($mixed): \stdClass
    {
    }

    /**
     * @param \stdClass $obj
     * @param int|bool $mixed
     * @return \stdClass
     */
    public function firstParamCanBeRemoved(\stdClass $obj, $mixed): \stdClass
    {
    }
}
```

**After:**

```php
<?php

class Example
{
    public function canBeRemoved(\stdClass $obj): \stdClass
    {
    }

    /**
     * @param \stdClass|array $mixed
     */
    public function canNotBeRemoved($mixed): \stdClass
    {
    }

    /**
     * @param int|bool $mixed
     */
    public function firstParamCanBeRemoved(\stdClass $obj, $mixed): \stdClass
    {
    }
}
```

keradus:
Hi.
This feature was accepted and even discussed how to do it as part of bigger, more generic rule.
Need to find that discussion.
@localheinz , I remember I was discussing it with you and you thumbed this issue as well, do you remember our discussion?

localheinz:
@keradus 

Cannot recall, but I remember countless times re-iterating personal rules for when a method needs a docblock and when it doesn't to former team members, hehe.

In the light of that, a fixer that solves this problem would be a nice addition to the tool belt!

localheinz:
However, I would suggest to

* remove entire docblocks only when they do not contain any additional information
* not remove parts of a docblock


<?php

class Example
{
    /**
     * @param \stdClass $obj
     * @return \stdClass
     */
    public function canBeRemoved(\stdClass $obj): \stdClass
    {
    }

    /**
     * @param \stdClass|array $mixed
     * @return \stdClass
     */
    public function canNotBeRemoved($mixed): \stdClass
    {
    }

    /**
     * @param \stdClass $obj
     * @param int|bool $mixed
     * @return \stdClass
     */
[...]

PR #3734, opened 2018-05-06
Add NoSuperfluousPhpdocTagsFixer

julienfalque:
Implements #2979.


gharlan:
What about this case?

```php
/**
 * @param Foo $foo
 */
function bar(?Foo $foo) {}
```

(forgotten `|null` in phpdoc..)

Will the phpdoc be removed?

julienfalque:
No, the current version only removes PHPDocs types when they match the actual typehints exactly. In some cases, the difference can be intended.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -0,0 +1,241 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should run before NoEmptyPhpdocFixer
```
ntzm: How should this run in relation to `phpdoc_return_self_reference`?

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -0,0 +1,280 @@
...
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
```
ntzm: Can you add cases for:

- Namespaced + imported classed:
```php
use Foo\Bar;

/**
 * @param Bar
 */
function foo(Bar $bar) {}
```
- Namespaced + not imported classes:
```php
/**
 * @param \Foo\Bar
 */
function foo(\Foo\Bar $bar) {}
```
```php
/**
 * @param Foo\Bar
 */
function foo(Foo\Bar $bar) {}
```
- Mix of imported and not imported:
```php
use Foo\Bar;

/**
 * @param \Foo\Bar
 */
function foo(Bar $bar) {}
```
- Imported aliased classes:
```php
use Foo\Bar as Baz;

/**
 * @param Foo\Bar
 */
function foo(Baz $bar) {}
```

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
julienfalque: Nice catch! Test added and priority updated.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
julienfalque: Thank you, I added the tests and updated the code as most were failing.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -0,0 +1,482 @@
...
+}',
+                '<?php
+class Foo {
+    /**
+     * @param Bar $bar
+     *
+     * @return Baz
+     */
+    public function doFoo(Bar $bar): Baz {}
+}',
```
keradus: please add case with input:
```
class Foo {
    /**
     * @param Bar $bar
     *
     * @return Baz
     */
    public function doFoo(Bar $bar) {}
}',
```

input is typehinted on language level, but output is not.
will annotation for input will be removed, while annotation for output will remain?


-------------------

```
class Foo {
    /**
     * @param Bar $bar
     * @param Bax|Baz $baxz
     */
    public function doFoo(Bar $bar, $baxz) {}
}',
```

will `bar` annotation be removed, and `baxz` not?

src/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixer.php
```diff
@@ -101,6 +101,14 @@ public function isCandidate(Tokens $tokens)
         return count($tokens) > 10 && $tokens->isTokenKindFound(T_DOC_COMMENT) && $tokens->isAnyTokenKindsFound([T_CLASS, T_INTERFACE]);
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return 10;
```
keradus: comment with reasoning is missing

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $actualTypes = [$info['type']];
+        if ($info['allows_null']) {
+            $actualTypes[] = 'null';
+        }
+
+        return $annotationTypes === $this->toComparableNames($actualTypes, $symbolShortNames);
+    }
+
+    private function toComparableNames(array $types, array $symbolShortNames)
```
keradus: what are the input arrays?
what is the output ?

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -0,0 +1,264 @@
...
+            return $this->parseTypeHint($tokens, $tokens->getNextMeaningfulToken($colonIndex));
+        }
+
+        return [
+            'type' => null,
+            'allows_null' => true,
+        ];
+    }
+
+    private function parseTypeHint(Tokens $tokens, $index)
```
keradus: typehinting for index is missing.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -0,0 +1,264 @@
...
+
+            if ($tokens[$index]->isGivenKind(T_FUNCTION)) {
+                return $index;
+            }
+        } while ($tokens[$index]->isGivenKind([T_ABSTRACT, T_FINAL, T_STATIC, T_PRIVATE, T_PROTECTED, T_PUBLIC]));
+
+        return null;
+    }
+
+    private function getArgumentsInfo(Tokens $tokens, $start, $end)
```
keradus: `ArgumentsAnalyzer` won't help here ?

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
julienfalque: Test cases added.

src/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixer.php
julienfalque: Ah yep, forgot that one, thanks.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
julienfalque: PHPDoc added.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
julienfalque: PHPDoc added.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
julienfalque: Didn't thought about that class. At first glance I'm not sure it can provide all information the fixer needs, but maybe it's worth adding a new method to it :)

keradus: IIRC, for now it finds indexes of args. so not doing everything you need here, but half done

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -212,6 +212,42 @@ function foo(Baz $bar) {}',
...
+     * @return Baz
+     */
+    public function doFoo(Bar $bar) {}
+}',
+            ],
+            'with_not_all_params_superfluous' => [
+                '<?php
+class Foo {
+    /**
+     * @param Bax|Baz $baxz
```
keradus: so...
usually, in repos I saw, ppl like to have all `@param`+`@return` if at least one of param/return can't be done on language level, and remove whole docblock when all of them are typehinted.

would it be big thing to add such option ?

julienfalque: I don't think so, but I would postpone that to a separate PR because that would still require some code and review time and I do want this PR to be merged before next minor release :)

keradus: that's the thing. I would like to have such option already in 2.12

keradus:
IMO, this shall be default behaviour:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2979#issuecomment-323092084

adding option to define behaviour later means that default behaviour won't be good

julienfalque:
While I'm :+1: for an option like `only_when_all_superfluous`, I'm :-1: to enable it by default: with PHP 7, PHPDocs should only be their to fill the remaining gaps, nothing more. This would also be in contradiction with the default behavior of `phpdoc_add_missing_param_annotation` (option `only_untyped` defaults to `true`).

keradus:
convinced with `only_untyped`

keradus:
Thank you @julienfalque.

TomasVotruba:
Sometimes the code tells that some types is KNOWN to be unknown, almost any value like:

```php
/**
 * @param mixed $value
 */
public function compute($value): int
{
    // ...
}
```

Seeing this:

```php
public function compute($value): int
{
    // ...
}
```

is more often than not forgotten typehint in method. And other fixer takes care of that to make the programmer add that typehint or at least annotation.

To bad this fixer removes `mixed` and there is no way to turn it off. I'd like to use it along with [`RemoveUselessDocBlockFixer`](https://github.com/symplify/codingstandard#block-comment-should-only-contain-useful-information-about-types) but I can't for this reason.

Otherwise great job! :+1:

julienfalque:
I don't understand your comment. `@param mixed $value` has the exact same meaning as not having a typehint for `$value`. The annotation doesn't add any information and any automated tool should be able to detect the missing typehint with or without it, right?

TomasVotruba:
That's why I wrote:

> more often than not forgotten typehint in method. Other fixers/sniff takes care of that to make the programmer add that typehint or at least annotation.

It adds a value of knowing and is consistent with `*TypeHintSniffs`.

> any automated tool should be able to detect the missing typehint

How can that tool know it it's mixed or if its forgotten?

julienfalque:
> How can that tool know if it's mixed or if it's forgotten?

It can't, and I don't think adding `@param mixed $value` helps.

TomasVotruba:
It's matter or probability. Is this mixed or not?


```php
public function get($id): int
{
    // ...
}
```

julienfalque:
It definitely is: there is not type hint. You may expect `$id` to be an `int` (e.g. auto increment primary key) or a `string` (UUID). Again, this remains true if you add `/** @param mixed $id */`.

I don't understand why having the annotation would help guessing that an actual type hint is expected and forgotten. This should be done by analyzing what the method does and how it is used, which is out of the scope of PHP CS Fixer.

TomasVotruba:
This often happens in code-reviews for new programmers, so: "There should be `int` typehint" is a common comment.

But having a CS tool that handle this can create blindness and skip this and then application crashes on invalid type somewhere on production.

That's why I think this is dangerous to enforce.

TomasVotruba:
Also regarding to possible future compatibility: https://wiki.php.net/rfc/mixed-typehint, this typehint is needed in the doc.

Same as `object` was before.

PR #3735, opened 2018-05-06
Add LowercaseStaticReferenceFixer

kubawerlos:
```diff
<?php
class Foo extends Bar
{
    public function baz()
    {
        return 42;
    }

    public function qux()
    {
-        return SELF::baz();
+        return self::baz();
    }

    public function quux()
    {
-        return STATIC::baz();
+        return static::baz();
    }

    public function quuz()
    {
-        return PARENT::baz();
+        return parent::baz();
    }
}
```

I've added it to `@Symfony` rule set as it seems to fit other lowercasing fixers and is not violated anywhere there, so won't make any problems.

### DIFF
--- a/src/Fixer/Casing/LowercaseStaticReferenceFixer.php
+++ b/src/Fixer/Casing/LowercaseStaticReferenceFixer.php
@@ -1,0 +1,105 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Casing;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ */
+final class LowercaseStaticReferenceFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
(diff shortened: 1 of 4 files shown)


src/Fixer/Casing/LowercaseStaticReferenceFixer.php
```diff
@@ -0,0 +1,90 @@
...
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind([T_STATIC, T_STRING])) {
+                continue;
+            }
+
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
+            if (!$tokens[$nextIndex]->isGivenKind([T_PAAMAYIM_NEKUDOTAYIM])) {
+                continue;
```
julienfalque: What about supporting `instanceof` as well, e.g. `$foo instanceof self`? And typehints?

kubawerlos: @julienfalque done, ready for review.

SpacePossum:
Thanks for working on this, great idea for a new fixer :)

Please add the following test cases for `static`:
```php
    <?php
    $a = STATIC function() {};
```
```php
    <?php
    class A
    {
        public function B()
        {
            STATIC $a; echo $a;
        }
    }
```
```php
    <?php
    class A
    {
        public function B()
        {
            $collection = $static ? new STATIC($b) : new self();
        }
    }
```
```php
    <?php
    class A
    {
        STATIC public function A()
        {
        }
    }
```
```php
    <?php
        $a = function () {
            STATIC $B = false;
            if ($B) {
                echo 1;
            }
            $B = true;
        };
```

src/Fixer/Casing/LowercaseStaticReferenceFixer.php
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_STATIC, T_STRING]);
+    }
+
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
```
SpacePossum: I think this could be a `foreach ($tokens as $index => $token)`, yet I don't consider it blocking ;)

kubawerlos: Yes, it could and it is now :)

SpacePossum: thanks :D

keradus:
Thank you @kubawerlos.

PR #3738, opened 2018-05-10
ReturnAssignmentFixer description update

kubawerlos:
The original description gave me a little headache. I guess the intention of @SpacePossum was to read it as:
`Non (global, static or indirectly) referenced variables ...`
I understood it as:
`(Non global), static or indirectly referenced variables ...`
Wouldn't the new version be more readable?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1377,8 +1377,8 @@ Choose from the list of available rules:
 
 * **return_assignment**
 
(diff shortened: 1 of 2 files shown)


SpacePossum:
:+1: 
I'm not very good with these descriptions, having a technical-copywriter here would be awesome ;)

keradus:
Thank you @kubawerlos.

PR #3740, opened 2018-05-11
NoUnusedImportsFixer - remove SF exception

SpacePossum:
This should be handled in the configuration of the SF project itself, not at this location anymore.


keradus:
Thank you @SpacePossum.

PR #3748, opened 2018-05-11
PhpUnitStrictFixer - extend risky note

### LINKED ISSUE #3746: php_unit_strict should not use assertSame on values that might be objects
`php_unit_strict` is marked risky because "risky when any of the functions are overridden"

In reality, it exchanges all instances of `assertEquals` to `assertSame`, which breaks tests with objects that are equal but not identical. Likewise `assertNotEquals` and `assertNotSame` will fail too.

I doubt it's reasonable to statically analyse the tests to determine whether its parameters are objects, but if either side is a literal it should be doable. If this is a wontfix, please update the readme to warn about it.

jnvsor:
Closes #3746

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitStrictFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitStrictFixer.php
@@ -73,7 +73,7 @@ final class MyTest extends \PHPUnit_Framework_TestCase
                 ),
             ),
             null,
-            'Risky when any of the functions are overridden.'
+            'Risky when any of the functions are overridden or when testing object equality.'
         );
     }
 
--- a/README.rst
+++ b/README.rst
@@ -818,7 +818,7 @@ Choose from the list of available rules:
   PHPUnit methods like ``assertSame`` should be used instead of
   ``assertEquals``.
 
-  *Risky rule: risky when any of the functions are overridden.*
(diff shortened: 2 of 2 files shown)


src/Fixer/PhpUnit/PhpUnitStrictFixer.php
```diff
@@ -74,7 +74,7 @@ public function testSomeTest()
                 ),
             ],
             null,
-            'Risky when any of the functions are overridden.'
+            'Risky when any of the functions are overridden, or you are testing object equality.'
```
SpacePossum: Thanks for the fast and nice proposal!
I would drop the Oxford Comma and rephrase to;
 `'Risky when any of the functions are overridden or when testing object equality.'`
just to stay in the same style of the rest of most of the README, we kind of like to have the same style with everything here, not stating one is better than the other ^_^

jnvsor: Fixed

> we kind of like to have the same style with everything here

That's kind of the whole point of `php-cs-fixer` right? :smile: 

How can I tell which is the lowest maintained branch?

julienfalque: The lowest maintained branch is 2.2 currently. I think we don't have any documentation about that.

keradus:
Thank you @jnvsor.

PR #3745, opened 2018-05-11
Add PhpUnitInternalClassFixer

BackEndTea:
Fixes #3741 

Adds `@internal` annotation to phpunit test classes. This fixes #3741 if used in combination with the `final_internal_class` fixer.

PR #3749, opened 2018-05-11
Make sure PHPUnit is cased correctly in fixers descriptions

kubawerlos:
With a special [dedication](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3745/files#diff-88b99bb28683bd5b7e3a204826ead112R1126) for @BackEndTea.

### DIFF
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -58,6 +58,9 @@ final class FixerTest extends TestCase
         $fixerIsConfigurable = $fixer instanceof ConfigurationDefinitionFixerInterface;
 
         $this->assertRegExp('/^[A-Z`].*\.$/', $definition->getSummary(), sprintf('[%s] Description must start with capital letter or a ` and end with dot.', $fixerName));
+        $this->assertNotContains('phpdocs', $definition->getSummary(), sprintf('[%s] `PHPDoc` must not be in the plural in description.', $fixerName), true);
+        $this->assertCorrectCasing($definition->getSummary(), 'PHPDoc', sprintf('[%s] `PHPDoc` must be in correct casing in description.', $fixerName));
+        $this->assertCorrectCasing($definition->getSummary(), 'PHPUnit', sprintf('[%s] `PHPUnit` must be in correct casing in description.', $fixerName));
 
         $samples = $definition->getCodeSamples();
         $this->assertNotEmpty($samples, sprintf('[%s] Code samples are required.', $fixerName));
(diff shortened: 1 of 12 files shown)


tests/AutoReview/FixerTest.php
```diff
@@ -231,4 +233,13 @@ private static function assertStringIsNotEmpty($actual, $message = '')
         self::assertInternalType('string', $actual, $message);
         self::assertNotEmpty($actual, $message);
     }
+
+    /**
+     * @param mixed  $actual
```
julienfalque: I would either write `@param string $actual` or add a type check in the method.

kubawerlos: Fixed, obviously should be `string` (I've copied it from the method above and missed that).

julienfalque:
:+1:

Would you keep PHPDoc or PHPDocs?

keradus:
`PHPDoc`
https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md

tests/AutoReview/FixerTest.php
```diff
@@ -58,7 +58,9 @@ public function testFixerDefinitions(FixerInterface $fixer)
         $fixerIsConfigurable = $fixer instanceof ConfigurationDefinitionFixerInterface;
 
         $this->assertRegExp('/^[A-Z`].*\.$/', $definition->getSummary(), sprintf('[%s] Description must start with capital letter or a ` and end with dot.', $fixerName));
-        $this->assertPhpUnitCasing($definition->getSummary(), sprintf('[%s] `PHPUnit` must be in correct casing in description.', $fixerName));
+        $this->assertNotContains('phpdocs', $definition->getSummary(), sprintf('[%s] `PHPDoc` must not be in the plural in description.', $fixerName), true);
```
keradus: is it case sensitive or not?
if it is, then it's not enough

tests/AutoReview/FixerTest.php
```diff
@@ -235,11 +239,12 @@ private static function assertStringIsNotEmpty($actual, $message = '')
     }
 
     /**
-     * @param string $actual
+     * @param string $haystack
+     * @param string $needle
      * @param string $message
      */
-    private static function assertPhpUnitCasing($actual, $message)
+    private static function assertCorrectCasing($haystack, $needle, $message)
```
keradus: expected value, here expected casing, shall be first param, to mimic native PHPUnit's assertions params order

tests/AutoReview/FixerTest.php
kubawerlos: [4th](https://github.com/sebastianbergmann/phpunit/blob/7.1.5/src/Framework/Assert.php#L250) parameter causes it's not.

keradus:
Thank you @kubawerlos.

Issue #3730, opened 2018-05-03
DX: allow failure of composer install on local setups


SpacePossum:
I would like to propose to mark this test skipped if _not_ running in the CI and only if the error is triggered by the timeout. For DX reasons I think it is better to be able to run the test suits locally even when no there is no internet connection available.

```
There was 1 error:

1) PhpCsFixer\Tests\Smoke\InstallViaComposerTest::testInstallationIsPossible
Symfony\Component\Process\Exception\ProcessTimedOutException: The process "composer install -q" exceeded the timeout of 60 seconds.

/home/possum/work/PHP-CS-Fixer/vendor/symfony/process/Process.php:1160
/home/possum/work/PHP-CS-Fixer/vendor/symfony/process/Process.php:382
/home/possum/work/PHP-CS-Fixer/vendor/symfony/process/Process.php:201
/home/possum/work/PHP-CS-Fixer/vendor/keradus/cli-executor/src/CommandExecutor.php:65
/home/possum/work/PHP-CS-Fixer/tests/Smoke/InstallViaComposerTest.php:71
```

SpacePossum:
closing, as not worth the trouble keeping this as a ticket in the pool

PR #3753, opened 2018-05-16
[DX] Allow composer install fail on non-CI setup.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3730

The issue is that each command that is executed through the `CommandExecutor` has a max. run time of 60 seconds. This is not always enough when running the tests with a slow internet connection or no connection at all.

### DIFF
--- a/tests/Smoke/InstallViaComposerTest.php
+++ b/tests/Smoke/InstallViaComposerTest.php
@@ -74,6 +74,8 @@ final class InstallViaComposerTest extends TestCase
         } catch (ProcessTimedOutException $e) {
             if (false === getenv('CI')) {
                 self::markTestSkipped('Install by `composer` failed. Details:'."\n".$e->getMessage());
+            } else {
+                throw $e;
             }
         }
 
--- a/tests/Smoke/InstallViaComposerTest.php
+++ b/tests/Smoke/InstallViaComposerTest.php
@@ -16,6 +16,7 @@ use Keradus\CliExecutor\CommandExecutor;
 use PhpCsFixer\Utils;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Process\Exception\ProcessTimedOutException;
 
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
@@ -33,13 +34,13 @@ final class InstallViaComposerTest extends TestCase
         try {
             CommandExecutor::create('php --version', __DIR__)->getResult();
         } catch (\RuntimeException $e) {
-            self::markTestSkipped('Missing `php` env script. Details:'."\n".$e->getMessage());
+            self::markTestSkipped("Missing `php` env script. Details:\n".$e->getMessage());
         }
 
         try {
             CommandExecutor::create('composer --version', __DIR__)->getResult();
         } catch (\RuntimeException $e) {
-            self::markTestSkipped('Missing `composer` env script. Details:'."\n".$e->getMessage());
+            self::markTestSkipped("Missing `composer` env script. Details:\n".$e->getMessage());
         }
     }
 
@@ -68,7 +69,14 @@ final class InstallViaComposerTest extends TestCase
             json_encode($initialComposerFileState, Utils::calculateBitmask(array('JSON_PRETTY_PRINT')))
         );
 
-        $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        try {
+            $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        } catch (ProcessTimedOutException $e) {
+            if (false === getenv('CI')) {
+                self::markTestSkipped('Install by `composer` failed. Details:'."\n".$e->getMessage());
+            }
+        }
+
         $this->assertSame(0, CommandExecutor::create('composer dump-autoload --optimize', $tmpPath)->getResult()->getCode());
         $this->assertSame(0, CommandExecutor::create('php vendor/autoload.php', $tmpPath)->getResult()->getCode());
         $this->assertSame(0, CommandExecutor::create('vendor/bin/php-cs-fixer --version', $tmpPath)->getResult()->getCode());


tests/Smoke/InstallViaComposerTest.php
```diff
@@ -68,7 +69,14 @@ public function testInstallationIsPossible()
             json_encode($initialComposerFileState, Utils::calculateBitmask(array('JSON_PRETTY_PRINT')))
         );
 
-        $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        try {
+            $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        } catch (ProcessTimedOutException $e) {
+            if (false === getenv('CI')) {
```
keradus: not CI shall be the indicator.

Imagine travis is dead, yet we want to release.
Then, one will run tests locally and make a release. and if this test would fail, and smoke tests are super important to pass for making a release, it must not be skipped.

We already had that kind of issue with phars, when they are there sometimes, but not always, normally we want to skip if they are not present, but not when preparing the release. For that, we introduced `PHP_CS_FIXER_TEST_ALLOW_SKIPPING_PHAR_TESTS` flag.

I believe we shall use same approach here. Either by introducing new flag, or replacing existing flag with sth like `PHP_CS_FIXER_TEST_ALLOW_SKIPPING_SMOKE_TESTS` and use it in both places.
True by default, false when we run build.

tests/Smoke/InstallViaComposerTest.php
```diff
@@ -68,7 +69,16 @@ public function testInstallationIsPossible()
             json_encode($initialComposerFileState, Utils::calculateBitmask(array('JSON_PRETTY_PRINT')))
         );
 
-        $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        try {
+            $this->assertSame(0, CommandExecutor::create('composer install -q', $tmpPath)->getResult()->getCode());
+        } catch (ProcessTimedOutException $e) {
```
keradus: it doesn't cover situation when there is no `composer` available (eg under docker)

SpacePossum: that failure would not result in a `ProcessTimedOutException` I think?

keradus: nope, that's why this `catch` doesn't cover it

tests/Smoke/InstallViaComposerTest.php
SpacePossum: While I think I understand this; it is not the current case (before this PR).

Looking at the `setUp` in the test it is already skipped when for `composer` is not available at all, leaving the same concern that the smoke test is skipped.

So if we going to update to use a new flag I propose to use this flag to (not) skip the current tests in this file and update the `ProjectCodeTest` as well to (not) skip the tests either when running the `build`.

keradus: It's not the state for this smoke test of `install via composer`, but it is the state for `phar tests`.
Difference is that currently, `install via composer` is skipped on missing system deps, but not on crashing commands, you want to skip if crashed, and that's not sth I want to have, as it will make manual releases risky.

I don't get it, how are `ProjectCodeTest` related to this topic?

SpacePossum: >It's not the state for this smoke test of install via composer, but it is the state for phar tests.

OK, than why is the test skipped if this requirement for the test is not matched during a build?

>you want to skip if crashed

when not doing a build and running the test locally without an internet connection and not on CI I think this test can be skipped.

>I don't get it, how are ProjectCodeTest related to this topic?

Because it is the exact same thing; we allow skipping tests when not on CI when running a build. I don't get why you would allow skipping tests because it slow (`ProjectCodeTest`) or when `composer` is not installed (current state) when doing a build, but not allow tests to be skipped when there is no internet connection when not running a build.

The last one is really bad DX I think, I started skipping running the test locally and just open PR's to see what happens because this test cannot be trusted; it fails not because my code changes, it fails because it has dependencies that are not matched. If I remove composer it is skipped, if I have composer but no internet connection it fails.
There is no logic in there; it should both fail or both pass.

keradus: > OK, than why is the test skipped if this requirement for the test is not matched during a build?

because nobody introduced such safe-flag for this test class so far.

> when not doing a build and running the test locally without an internet connection and not on CI I think this test can be skipped.

no, it shall fail and I shall be forced to run it again, after fixing internet connection

> I don't get why you would allow skipping tests because it slow (ProjectCodeTest)

because `ProjectCodeTest` is part of `AutoReview` mechanism that abuses PHPUnit as tests runner, `ProjectCodeTest` is not real test like `unit/integration/smoke` we have

> or when `composer` is not installed (current state)

because nobody introduced such safe-flag for this test class so far.

> If I remove composer it is skipped, if I have composer but no internet connection it fails.
There is no logic in there; it should both fail or both pass.

That's the thing - I totally agree that by default they shall be skipped if system-/internet-requirements are not met.
but just by default, and for release build I would like to be more strict. We can have higher requirements for release build than for development build, it's fine.

tests/Smoke/InstallViaComposerTest.php
keradus: ack

tests/Smoke/InstallViaComposerTest.php
keradus: > OK, than why is the test skipped if this requirement for the test is not matched during a build?

because nobody introduced such safe-flag for this test class so far.

> when not doing a build and running the test locally without an internet connection and not on CI I think this test can be skipped.

no, it shall fail and I shall be forced to run it again, after fixing internet connection

> I don't get why you would allow skipping tests because it slow (ProjectCodeTest)

because `ProjectCodeTest` is part of `AutoReview` mechanism that abuses PHPUnit as tests runner, `ProjectCodeTest` is not real test like `unit/integration/smoke` we have

> or when `composer` is not installed (current state)

because nobody introduced such safe-flag for this test class so far.

> If I remove composer it is skipped, if I have composer but no internet connection it fails.
There is no logic in there; it should both fail or both pass.

That's the thing - I totally agree that by default they shall be skipped if system-/internet-requirements are not met.
but just by default, and for release build I would like to be more strict. We can have higher requirements for release build than for development build, it's fine.

SpacePossum:
Closing as doing this in an acceptable way is not something I can commit to. Would still be nice to see resolved, yet not very important.

Issue #3754, opened 2018-05-16
YodaStyleFixer does not like space between array name and leading [


phil-davis:
#### The PHP version you are using (`$ php -v`):
=> PHP 7.1.17-1+ubuntu16.04.1+deb.sury.org+1 (cli) (built: May  5 2018 04:55:21) ( NTS )

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.11.1 Grey Devil by Fabien Potencier and Dariusz Ruminski

#### The command you use to run PHP CS Fixer:
=> lib/composer/bin/php-cs-fixer fix -vvv --diff --allow-risky yes

#### The configuration file you are using, if any:
See https://github.com/owncloud/coding-standard/releases/tag/1.0.1

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php
trait Logging {
	public function example($arrayOfStuff) {
		if ("" !== $arrayOfStuff ['key']) {
			echo "Yes";
		}
	}
}
```

 * with unexpected changes applied when running PHP CS Fixer:
There is an error:
```
lib/composer/bin/php-cs-fixer fix -vvv --diff --allow-risky yes
You are running PHP CS Fixer with xdebug enabled. This has a major impact on runtime performance.
If you need help while solving warnings, ask at https://gitter.im/PHP-CS-Fixer, we will help you!
Loaded config ownCloud coding standard from "/home/phil/ownCloud/core/.php_cs.dist".
Using cache file ".php_cs.cache".
[...]

PR #3755, opened 2018-05-16
YodaStyleFixer - handle space between var name and index

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3754


keradus:
Thank you @SpacePossum.

PR #3756, opened 2018-05-16
Fix Yoda when space between  and [ - issue 3754

phil-davis:
Fixes issue #3754 

After a var name, like ``$this->myArray`` then there might be a white-space token before the ``[`` that introduces the array key. Make sure to skip over that white-space token before looking for the ``[`` and its ``]`` mate.


phil-davis:
PR #3755 already has the fix.

keradus:
wow, thanks for opening the initial issue and super kudos for fixing it yourself :+1:
I'm glad to see that community fixed it in parallel ;)

Issue #3760, opened 2018-05-17
binary_operator_spaces not working in closure type argument


xenmayer:
PHP => 7.2.4
PHP-CS-FIXER=> 2.11.1

Command
```bash
php-cs-fixer fix /path/to/file.php
```

Config
```php
<?php

return PhpCsFixer\Config::create()
                        ->setRiskyAllowed(true)
                        ->setRules(
                            [
                                '@Symfony' => true,
                                '@Symfony:risky' => false,
                                '@PHP71Migration' => true,
                                'array_syntax' => ['syntax' => 'short'],
                                'array_indentation' => true,
                                'dir_constant' => true,
                                'heredoc_to_nowdoc' => true,
                                'linebreak_after_opening_tag' => true,
                                'modernize_types_casting' => true,
                                'no_multiline_whitespace_before_semicolons' => true,
                                'no_unreachable_default_argument_value' => true,
                                'no_useless_else' => true,
                                'no_useless_return' => true,
                                'ordered_class_elements' => true,
                                'ordered_imports' => true,
                                'phpdoc_add_missing_param_annotation' => ['only_untyped' => false],
                                'phpdoc_order' => true,
                                'doctrine_annotation_braces' => true,
                                'doctrine_annotation_indentation' => true,
                                'doctrine_annotation_spaces' => true,
                                'psr4' => true,
                                'no_php4_constructor' => true,
                                'no_short_echo_tag' => true,
                                'declare_equal_normalize' => [
                                    'space' => 'single'
                                ],
                                'binary_operator_spaces' => [
                                    'default' => 'align_single_space_minimal'
                                ],
                                'concat_space' => [
                                    'spacing' => 'one'
                                ],
                                'no_trailing_whitespace' => true,
                                'no_whitespace_in_blank_line' => true,
                                'semicolon_after_instruction' => true,
                                'align_multiline_comment' => true,
                                'doctrine_annotation_array_assignment' => true,
                                'general_phpdoc_annotation_remove' => ['annotations' => ["author", "package"]],
                                'list_syntax' => ["syntax" => "short"],
                                'phpdoc_types_order' => ['null_adjustment'=> 'always_last'],
                                'single_line_comment_style' => true,
                            ]
                        )
                        ->setCacheFile(__DIR__.'/.php_cs.cache')
                        ->setFinder(
                            PhpCsFixer\Finder::create()
                                             ->in(__DIR__)
                        );

```

Needed rule
```php
[...]

PR #3762, opened 2018-05-18
BinaryOperatorSpacesFixerTest - add failing tests

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3248
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3760


SpacePossum:
This root cause of this issue is when the fixer is inserting placeholder for aligning all `( )` blocks are skipped. This is done to prevent alignment where not wanted, for example:

```php
<?php

for($i = 1; $i < 3; ++$i) {
   $a = $i; // do not align this `=` with the one in the `for` statement above
}
```

other cases are `function calls` and `function definitions`.

While effective, the skipping is also so "wide" that valid cases (those that could be fixed) are missed.
Based on the input of both related issues I've created unit tests to show where the missed opportunities are.

PR #3765, opened 2018-05-18
Fix binary-prefixed double-quoted strings to single quotes

ntzm:


src/Fixer/StringNotation/SingleQuoteFixer.php
```diff
@@ -54,17 +54,36 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                '"' === $firstCharacter &&
+                $this->compatibleWithSingleQuotes($content)
             ) {
                 $content = substr($content, 1, -1);
                 $content = str_replace(array('\\"', '\\$'), array('"', '$'), $content);
                 $tokens[$index] = new Token(array(T_CONSTANT_ENCAPSED_STRING, '\''.$content.'\''));
             }
+
+            if (
+                0 === stripos($content, 'b"') &&
```
ntzm: would `'b"' === substr($content, 0, 2)` be preferred here?

SpacePossum:
nice catch,
we should really make a fixer to ditch the `b` at some point, it will never have any value since PHP6 didn't work out as planned ;)

src/Fixer/StringNotation/SingleQuoteFixer.php
SpacePossum: I would you the one from your comment, as there is no need to search the whole content for `b"` which is what `stripos` will do, but just my vote ;)

src/Fixer/StringNotation/SingleQuoteFixer.php
```diff
@@ -54,17 +54,46 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                '"' === $firstCharacter &&
+                $this->isCompatibleWithSingleQuotes($content)
             ) {
                 $content = substr($content, 1, -1);
-                $content = str_replace(array('\\"', '\\$'), array('"', '$'), $content);
+                $content = $this->addSlashesToCharacters($content);
                 $tokens[$index] = new Token(array(T_CONSTANT_ENCAPSED_STRING, '\''.$content.'\''));
             }
+
+            if (
```
julienfalque: The body of this `if` and the one before it are pretty much the same except handling of first character. Removing the `b` flag from `$content` if present, then keeping existing process and adding the flag back would allow you to do less changes.

ntzm: Good point, fixed!

keradus:
Thank you @ntzm.

PR #3766, opened 2018-05-18
Add NoBinaryStringFixer

ntzm:


julienfalque:
What about converting `(binary)` cast to `(string)`?

ntzm:
Is that worth doing in the same fixer?

julienfalque:
You're right, this should be a different fixer.

src/Fixer/StringNotation/NoBinaryStringFixer.php
```diff
@@ -0,0 +1,64 @@
...
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind([T_CONSTANT_ENCAPSED_STRING, T_START_HEREDOC])) {
```
SpacePossum: I think NOWDOC could also be fixed?

```php
<?php

echo b<<<'EOT'
test
EOT;
?>
```

please add a HEREDOC code sample as well :)

keradus: it's not like i was in the middle of merging here, and then boom, comment....
merge stopped just in time

keradus:
@ntzm , FYI - I have squashed your PR. feel free to push your original branch (with `--force`) or use squashed one

src/Fixer/StringNotation/NoBinaryStringFixer.php
SpacePossum: sorry about that one, you guys are just way to fast for me today ;)

ntzm: NOWDOCs are dealt with here, they have a token of T_START_HEREDOC though 😆

ntzm:
Pushed using force, added a HEREDOC code sample

src/Fixer/StringNotation/NoBinaryStringFixer.php
SpacePossum: what, damn PHP tokenizer haha, thanks man!

keradus:
Thank you @ntzm.

PR #3770, opened 2018-05-20
Handle binary flags in heredoc_to_nowdoc

ntzm:


src/Fixer/StringNotation/HeredocToNowdocFixer.php
```diff
@@ -101,7 +101,7 @@ private function convertToNowdoc(Token $token)
     {
         return new Token(array(
             $token->getId(),
-            Preg::replace('/(?<=^<<<)([ \t]*)"?([^\s"]+)"?/', '$1\'$2\'', $token->getContent()),
```
SpacePossum: the `<=^` part is no longer needed?

keradus: I'm so proud of you, Julien ;)

keradus:
Thank you @ntzm.

PR #3774, opened 2018-05-21
PhpUnitTestClassRequiresCoversFixer - Remove unneeded loop and use phpunit indicator class

BackEndTea:
Refactor of the `PhpUnitTestClassRequiresCoversFixer` to reduce unneeded looping of token sets.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -66,90 +66,87 @@ final class MyTest extends \PHPUnit_Framework_TestCase
     {
         $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
 
-        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
-            if (!$tokens[$index]->isGivenKind(T_CLASS)) {
-                continue;
-            }
-
-            $prevIndex = $tokens->getPrevMeaningfulToken($index);
-
-            // don't add `@covers` annotation for abstract base classes
-            if ($tokens[$prevIndex]->isGivenKind(T_ABSTRACT)) {
-                continue;
-            }
-
-            if (!$phpUnitTestCaseIndicator->isPhpUnitClass($tokens, $index)) {
-                continue;
-            }
-
-            $prevIndex = $tokens->getPrevMeaningfulToken($index);
-            $index = $tokens[$prevIndex]->isGivenKind(T_FINAL) ? $prevIndex : $index;
-
-            $indent = $tokens[$index - 1]->isGivenKind(T_WHITESPACE)
-                ? Preg::replace('/^.*\R*/', '', $tokens[$index - 1]->getContent())
-                : '';
-
-            $prevIndex = $tokens->getPrevNonWhitespace($index);
-            $doc = null;
-            $docIndex = null;
-
-            if ($tokens[$prevIndex]->isGivenKind(T_DOC_COMMENT)) {
-                $docIndex = $prevIndex;
-                $docContent = $tokens[$docIndex]->getContent();
-
-                // ignore one-line phpdocs like `/** foo */`, as there is no place to put new annotations
-                if (false === strpos($docContent, "\n")) {
-                    continue;
-                }
-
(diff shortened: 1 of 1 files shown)


src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -66,90 +66,87 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
         $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
 
-        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
-            if (!$tokens[$index]->isGivenKind(T_CLASS)) {
-                continue;
-            }
-
-            $prevIndex = $tokens->getPrevMeaningfulToken($index);
+        foreach ($phpUnitTestCaseIndicator->findPhpUnitClasses($tokens) as $indexes) {
```
SpacePossum: this used to loop from the last token to the first, no it is the other way around, are you sure this is without issues when the fixer add a lot of tokens?

BackEndTea: This fixer should not add much tokens, but it has the same issue as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3745#discussion_r190040263, so it might be best to wait on the results of that.

SpacePossum: please update after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3745/files#diff-60a412e1b3e8bf5c206429800960dd55R58 :)

SpacePossum: ping @BackEndTea this one is close to merging with a little tweak I think :)

BackEndTea: The default of `findPhpUnitClasses` is to start at the bottom right now. Should i add true as a second argument to make this more explicit here?

SpacePossum: ah nice, all good than :+1:  :)

keradus:
this one is old, and has some conflicts now.
i was just about to merge it, can you resolve conflicts first, @BackEndTea ? thanks !

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -151,5 +148,21 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
 
             $tokens[$docIndex] = new Token([T_DOC_COMMENT, implode('', $lines)]);
         }
+
+        $lines = $doc->getLines();
```
keradus: this, till end of the method, is duplicated code. why is that? maybe a merge conflict solved wrongly?

BackEndTea: You're correct, seems like a rebase that went wrong somewhere, i fixed it now.

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -66,75 +66,73 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
-                    ]);
-                    ++$docIndex;
-                }
-
-                $doc = new DocBlock($tokens[$docIndex]->getContent());
+                ++$docIndex;
             }
 
+            $doc = new DocBlock($tokens[$docIndex]->getContent());
+        }
```
dmvdbrugge: By having moved this close-brace up, the remainder of the method is indented wrongly.

This means that either `braces` isn't doing its job correctly, or you didn't run `php-cs-fixer fix`, and none of the PR-style-checks are picking it up 😕

Can you run `php-cs-fixer fix` and see if it indents correctly?

SpacePossum:
Thank you @BackEndTea.

BackEndTea:
Thank you @SpacePossum

PR #3777, opened 2018-05-21
EscapeImplicitBackslashesFixer - handle binary strings

ntzm:


src/Fixer/StringNotation/EscapeImplicitBackslashesFixer.php
```diff
@@ -102,7 +102,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
         $doubleQuoteOpened = false;
         foreach ($tokens as $index => $token) {
             $content = $token->getContent();
-            if ($token->equals('"')) {
+            if ('"' === $content || 'b"' === strtolower($content)) {
```
keradus: :-1:
don't grab content to compare it directly.
here, we don't know what was the token type.

ntzm: New approach better?

keradus:
Thank you @ntzm.

PR #3783, opened 2018-05-27
CI: Split COLLECT_COVERAGE job

keradus:
To allow job with coverage to fail,
as I have no clue why it's failing... :(


keradus:
While issue is not solved for sure, I'm gonna merge it as-is for now, so we can proceed with merging other PRs...
if one have idea how to fix failing code coverage, help is more than welcome !

PR #3793, opened 2018-05-31
DX: enable no_superfluous_phpdoc_tags rule

keradus:
What do you prefer guys ? shall we use `no_superfluous_phpdoc_tags` ?
Please vote ;)


keradus:
cc @SpacePossum @julienfalque @localheinz

SpacePossum:
If it makes a PHPDoc completely empty I'm :+1: , if not I like complete listings better, so in its current form I'm not in favor of the changes.

julienfalque:
:+1: even with partial removing :)

Slamdunk:
I don't know if I can vote.
To me partial docs look so misunderstanding / confusing

localheinz:
I agree with @Slamdunk and @SpacePossum.

ntzm:
I'm 👍for this PR, I think some devs are used to the PHPDoc with all the information, but IMO it's completely useless if it doesn't add anything.

keradus:
for now, there is no config option that would allow to drop position only if all params/return would be dropped.
if one would raise such PR, it would be highly appreciated ;)

for now, closing without mergin

PR #3794, opened 2018-05-31
DX - AutoReview - options sanity

SpacePossum:
While we allow `null`/`empty` values in the API we shouldn't use that combination ourselves because:
- more info is better ;)
- when leaving out the HelpCommand/Readme will fail to be generated

### DIFF
--- a/tests/AutoReview/FixerTest.php
+++ b/tests/AutoReview/FixerTest.php
@@ -15,6 +15,8 @@ namespace PhpCsFixer\Tests\AutoReview;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\DefinedFixerInterface;
 use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerConfiguration\AllowedValueSubset;
+use PhpCsFixer\FixerConfiguration\FixerOptionInterface;
 use PhpCsFixer\FixerDefinition\FileSpecificCodeSampleInterface;
 use PhpCsFixer\FixerDefinition\VersionSpecificCodeSampleInterface;
 use PhpCsFixer\FixerFactory;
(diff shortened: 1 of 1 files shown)


tests/AutoReview/FixerTest.php
```diff
@@ -128,8 +130,25 @@ public function testFixerDefinitions(FixerInterface $fixer)
...
+
+                $this->assertFalse(
+                    (null === $types || 0 === count($types)) && (null === $values || 0 === count($values)),
+                    sprintf('[%s] Option %s has no allowed types and values set.', $fixerName, $option->getName())
+                );
+
+                if (null !== $values) {
+                    foreach ($values as $value) {
+                        if ($value instanceof AllowedValueSubset) {
+                            $this->assertContains('array', $types, sprintf('[%s] Option %s has allowed value of type "AllowedValueSubset" but not allowed type "array".', $fixerName, $option->getName()));
```
keradus: can you show an example for what definition this assertion would fail?
(i'm not fully getting it)

SpacePossum:
```php
            (new FixerOptionBuilder('elements', sprintf('List of classy elements; \'%s\'.', implode("', '", $types))))
                // ->setAllowedTypes(['array']) this would fail, because `array` is not set as allow type
                ->setAllowedValues([new AllowedValueSubset($types)])
                ->setDefault(['const', 'method', 'property'])
                ->getOption(),
        ]);
```

SpacePossum: thinking of it, we could also do it the other way around; if the allowed values are defined, there is no need to define the types... hmmm

keradus: yeah, let us not mix types with values. only one shall be set.

:+1:

keradus: ultimately (3.0 to not introduce bc breaks?), would be nice to validate in runtime, not on tests run

SpacePossum:
this needs more looking into to figure out if there is any value at all

PR #3796, opened 2018-05-31
DX: StdinTest - do not assume name of folder, into which project was cloned

keradus:


tests/Smoke/StdinTest.php
```diff
@@ -45,7 +45,7 @@ public function testFixingStdin()
                     $fileResult->getError()
                 ),
                 'output' => str_ireplace(
-                    str_replace('/', DIRECTORY_SEPARATOR, 'PHP-CS-Fixer/tests/Fixtures/Integration/set/@PSR2.test-in.php'),
+                    str_replace('/', DIRECTORY_SEPARATOR, basename(realpath($cwd)) . '/tests/Fixtures/Integration/set/@PSR2.test-in.php'),
```
kubawerlos: Can we use `$inputFile` as it's exactly the path suffix?

keradus: yes we do ;)

PR #3798, opened 2018-05-31
DX: enable native_function_invocation

keradus:
please vote


keradus:
ping @FriendsOfPHP/php-cs-fixer

ntzm:
What are the before and after benchmarks?

kubawerlos:
Why not all native functions?

keradus:
@ntzm , minimal compared to fun with tokenizer
@kubawerlos , see discussion at #3048

PR #3799, opened 2018-05-31
DX: php_unit_test_case_static_method_calls - use default config

keradus:
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3476/files#r176360233

please vote

### DIFF
--- a/src/Console/Output/ErrorOutput.php
+++ b/src/Console/Output/ErrorOutput.php
@@ -65,7 +65,7 @@ final class ErrorOutput
             $message = $e->getMessage();
             $code = $e->getCode();
             if (0 !== $code) {
-                $message .= " (${code})";
+                $message .= " ({$code})";
             }
 
             $length = max(\strlen($class), \strlen($message));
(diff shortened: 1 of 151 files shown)


keradus:
ping @FriendsOfPHP/php-cs-fixer

SpacePossum:
:-1:  as most of the ((F)OSS) project I work on use `$this->`, I also like the current style better. All in all a matter of taste I think ;)

keradus:
so, let me remind what I referenced into:
1. we are already using self:: and static::, eg in our custom assertions, in which we can't use `$this->`, as our methods are static
2. SCA tools got confused by calling static method on `$this`

SpacePossum:
@keradus you've the majority of the voted, congrats!  :D 
can you prepare this PR again so it can be merged? thanks!

keradus:
would you want it to target `master` or LTS branch ?

SpacePossum:
I would vote LTS as to have least number of differences between the branches

keradus:
done

kubawerlos:
Thank you @keradus

PR #3800, opened 2018-05-31
Symfony:risky ruleset: include set_type_to_cast rule

keradus:


SpacePossum:
:+1:  we might even add it to the Symfony set as there are no hits there.

PR #3801, opened 2018-05-31
NativeFunctionInvocationFixe - fix buggy config validation

keradus:


SpacePossum:
nice catch :)
I've update the test to be even more strict, feel free to drop that commit if you thing it is to much.

keradus:
I will keep it (but squash during merge)

keradus:
Thank you @SpacePossum.

SpacePossum:
thanks @keradus !

PR #3804, opened 2018-06-01
Cleanup: remove useless constructor comment

kubawerlos:
It's the only one such comment in the repo.

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -146,8 +146,6 @@ final class ConfigurationResolver
     private $fixerFactory;
 
     /**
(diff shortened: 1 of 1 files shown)


kubawerlos:
Do you mean something like [this](https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/master/src/Fixer/NoUselessConstructorCommentFixer.php)?

SpacePossum:
exactly, looks like a great start :)

keradus:
agreeing here, fixer would be nice :+1:

keradus:
Thank you @kubawerlos.

PR #3806, opened 2018-06-02
ReturnAssignmentFixer with PHPDoc @var for the variable

kubawerlos:
We have to do something with such annotation as `phpdoc_to_comment` would want to change it if it wasn't run before - which means of course that in practice it will happen after flushing cache.

### DIFF
--- a/tests/Fixer/ReturnNotation/ReturnAssignmentFixerTest.php
+++ b/tests/Fixer/ReturnNotation/ReturnAssignmentFixerTest.php
@@ -407,6 +407,22 @@ var names are case insensitive */ }
 var names are case insensitive */ return $a   ;}
                 ',
             ],
+            [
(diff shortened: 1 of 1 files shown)


SpacePossum:
Not sure if the fixer should remove the doc.. have to think about it. For now it is an overseen priority issue of sure.

ntzm:
IMO that code should be kept as-is, psalm will complain with the `totallyTyped` option otherwise

SpacePossum:
when `$bar` is removed I expected leaving ` /** @var ValueFromSomeService $bar */` will also trigger a warning as it documents the type of a variable that does not exists?

dmvdbrugge:
Commenting for backreference: this is case 14 of #3844 (and I would like to see this fixed 😉)

Issue #2646, opened 2017-03-28
Multiline comments to one-line


zored:
Is that possible to convert multiline comments and PHPDocs with into one-line?
Only for comments with one line of useful information.

## Example
Before:
```php
/**
 * @Example
 */
```

After:
```php
/** @Example */
```

SpacePossum:
Hi @zored 

At the moment we do not have a rule/fixer to do that.
Could you elaborate or provide a use case for when you would want this to happen?

zored:
@SpacePossum, hi.

This rule would save 2 extra lines of code per each one-line comment block. Code will be more compact and easier to read.

For example, I have model objects with many properties that only have '@var' type definitions. And each property has getter with only '@return' annotation. They occupy 'N x 2 x 2' lines of code (where 'N' is number of properties).

Or '@inhertDoc' that occupies two extra lines of code.

Thanks for reply :)

julienfalque:
I won't use such fixer personally, but the use case seems valid to me.

zored:
Currently using [my own simple script](https://gist.github.com/zored/8ffde76c8ec1617756de6156b2c63218) for conversion.

keradus:
I wouldn't use that fixer as well. but it's not blocking the fixer to become accepted.
Actually, I would use reverse fixer for methods/classes/properties, but one-line docblock for regular vars.
Would be great to see it that generic and configurable.

zored:
This code:
```php
<?php
class Hi
{
    /** @return string */
    public function hello($string)
    {
        return $string;
    }
}
```

Fixes comment into:
```php
    * @param mixed $string
/** @return string */
```

So existing code should be aware about this type of comments.

@keradus, yes, options for this rule will be great. I think that some users will prefer to use multiline one-liners because they are easier to modify.

keradus:
> Fixes comment into:

what is fixes comment into that invalid piece of code? PHP CS Fixer or your custom script? if first, we have very sad bug here.

zored:
@keradus PHP CS Fixer 2.1.2.
The good news is that this one-line format rarely used in method PHPDocs  😃

keradus:
ok, let us move discussion about invalid phpdoc modification here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2654
here, let us focus only on `multiline vs oneline docblock` ;)

BackEndTea:
[...]

PR #3810, opened 2018-06-03
 Introduction of SingleToMultiLinePhpdocFixer 

BackEndTea:
Fixes: #2646 

If preferred, i can target the changes to the Doc block  class to the current LTS (2.2?), as its features are useful for a number of other fixers. 

```bash
$ ./php-cs-fixer describe single_to_multi_line_phpdoc
Description of single_to_multi_line_phpdoc rule.
Changes doc blocks from single to multi line, or reversed.

* const ('multi', 'single'): whether const blocks should be single or multi line; defaults to 'multi'
* property ('multi', 'single'): whether property doc blocks should be single or multi line; defaults to 'multi'
* method ('multi', 'single'): whether method doc blocks should be single or multi line; defaults to 'multi'
```
Fixing examples:
```diff
<?php
class Foo{
-    /** @var bool */
+    /**
+     * @var bool
+     */
    public $var;
}
```

 * Example #2. Fixing with configuration: ['property' => 'single'].
```diff
<?php
class Foo{
-    /**
-    * @var bool
-    */
+    /** @var bool */
    public $var;
 }
```


SpacePossum:
Nice work and :+1:  for concept.
About priorities there is some stuff to check out I think, like when rules remove lines (`phpdoc_no_useless_inheritdoc` and others) or add lines (`phpdoc_add_missing_param_annotation` and possibly others), together with locking down the priority with `phpdoc_to_comment`

tests/Fixtures/Integration/priority/general_phpdoc_annotation_remove,single_to_multi_line_phpdoc.test
```diff
@@ -0,0 +1,28 @@
+--TEST--
+Integration of fixers: general_phpdoc_annotation_remove,single_to_multi_line_phpdoc.
+--RULESET--
+{"general_phpdoc_annotation_remove": {"annotations": ["test"] }, "single_to_multi_line_phpdoc": {"method": "single"}}
```
ntzm: extra space after `["test"]`

src/Fixer/Phpdoc/SingleToMultiLinePhpdocFixer.php
```diff
@@ -0,0 +1,144 @@
...
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return int
+     */
+    private function getDocBlockIndex(Tokens $tokens, $index)
+    {
+        do {
+            $index = $tokens->getPrevNonWhitespace($index);
+        } while ($tokens[$index]->isGivenKind([T_PUBLIC, T_PROTECTED, T_PRIVATE, T_FINAL, T_ABSTRACT, T_COMMENT]));
```
ntzm: what about `T_VAR`?

hkdobrev:
@BackEndTea Do you think that could be finalised for 2.14?

BackEndTea:
I won't have time until somewhere next week to work on this.

TomasVotruba:
@BackEndTea Thanks for your work, this is awesome feature :+1:

src/Fixer/Phpdoc/SingleToMultiLinePhpdocFixer.php
dmvdbrugge: And what about `T_STATIC`?

tests/Fixtures/Integration/priority/general_phpdoc_annotation_remove,single_to_multi_line_phpdoc.test
keradus: done

keradus:
> I won't have time until somewhere next week to work on this.

@BackEndTea , feel free to ping us when you consider this ready.
in the meantime, i'm changing the milestone for this PR

BackEndTea:
@keradus Sorry for the long delay on this, I kinda forgot about it.

Currently this only works for class methods, properties and constants, and i'm not 100% sure how to expand it to also work with functions, classes and other doc blocks.

jaikdean:
I agree with SpacePossum, it would be great to see this released as it stands.

BackEndTea:
@SpacePossum & @keradus  Sorry for the long wait on this, but I have updated it, and it is ready for review.

I have rebased this and fixed a few remarks people had.
Current concerns are:
* The fact that its 'incomplete' e.g. only class properties, constants and methods. But this should work as is.
* Priorities with other fixes
* The fact that this includes updates to the `DocBlock` class. This isn't a final class and could be extended by others.

keradus:
@BackEndTea ,
the CI is failing, this is a blocker to proceed with the PR :(
can you fix it, please ?

staabm:
@BackEndTea do you have time to work on this?

gharlan:
I think we should find a better name for this one, because it can be used for both directions.

dmvdbrugge:
> I think we should find a better name for this one, because it can be used for both directions.

I think replacing `to` with `or` makes clear that it's purpose is to make sure it's always either single or multi, so `SingleOrMultiLinePhpdocFixer`. In fact, I see now that the branch is already called that 😆

localheinz:
Perhaps `PhpdocLineSpanFixer`?

SpacePossum:
rebased, conflicts resolved, fixed unit tests, fixed code style to new standard, rename the fixer, forced pushed, in preparation of merging

SpacePossum:
Thank you @BackEndTea.

TomasVotruba:
Thank you @BackEndTea, gonna deprecate my own rule now :)

PR #3811, opened 2018-06-04
Priority conflict: NoUnneededControlParenthesesFixer & ReturnAssignmentFixer

Slamdunk:
I don't know how to fixthis: priority numbers of the involved fixers are already stuck with other fixers.

Maybe we should can fix `NoUnneededControlParenthesesFixer` for the original variable assignment form, or maybe not, I don't know :disappointed:

### DIFF
--- a/tests/AutoReview/FixerFactoryTest.php
+++ b/tests/AutoReview/FixerFactoryTest.php
@@ -192,6 +192,7 @@ final class FixerFactoryTest extends TestCase
             [$fixers['pow_to_exponentiation'], $fixers['no_spaces_inside_parenthesis']],
             [$fixers['protected_to_private'], $fixers['ordered_class_elements']],
             [$fixers['return_assignment'], $fixers['blank_line_before_statement']],
+            [$fixers['return_assignment'], $fixers['no_unneeded_control_parentheses']],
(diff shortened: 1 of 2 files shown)


SpacePossum:
if `NoUnneededControlParenthesesFixer` is patched to fix these cases as well;

```
    return (true
        ? 1
        : 2
    );
```

than we should be good?

Slamdunk:
Yup

SpacePossum:
than we know how to fix ;)

julienfalque:
So... feature rather than bugfix?

Slamdunk:
Well, wait: the fact that I had to run PHP-CS-Fixer **twice** is a *bug*, and showing the bug is the intent of this PR.

The fact that the bug may not be an integration one, is another story that do not affect the type of flaw, which is a bug.

SpacePossum:
the bug is the priority issue, but if that issue is gone if we add the feature to the `NoUnneededControlParenthesesFixer`

SpacePossum:
After rechecking; if you use `no_unneeded_control_parentheses` with `return` it is fixed as you expect, if you don't configure `return` there is no prio issue; therefore I close this one.

PR #3812, opened 2018-06-05
Add FopenFlagOrderFixer & FopenFlagsFixer

### LINKED ISSUE #3417: FOpenFlagOrderFixer - vs 'b'/'t'
ref #3405

```diff
-fopen($foo, 'b+r')
+fopen($foo, 'r+b')
```
```diff
-fopen($foo, 't+w', $a, $b)
+fopen($foo, 'w+t', $a, $b)
```
```diff
fopen($foo, $bar, $a, $b)
```

SpacePossum:
closes #3417

ping @ntzm , hope you can help me out with:
- define the order of all flags
- ~do we want to have an option to remove flags when found, like `t`~
- ~do we want to have an option to add flags when not found, like `b`~

### DIFF
--- a/src/AbstractFopenFlagFixer.php
+++ b/src/AbstractFopenFlagFixer.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer;
+
+use PhpCsFixer\Tokenizer\Analyzer\ArgumentsAnalyzer;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @internal
+ *
+ * @author SpacePossum
+ */
+abstract class AbstractFopenFlagFixer extends AbstractFunctionReferenceFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_STRING, T_CONSTANT_ENCAPSED_STRING]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $argumentsAnalyzer = new ArgumentsAnalyzer();
+
+        $index = 0;
+        $end = $tokens->count() - 1;
+        while (true) {
+            $candidate = $this->find('fopen', $tokens, $index, $end);
+
+            if (null === $candidate) {
+                break;
+            }
+
+            $index = $candidate[1]; // proceed to '(' of `fopen`
+
+            // fetch arguments
+            $arguments = $argumentsAnalyzer->getArguments(
+                $tokens,
+                $index,
+                $candidate[2]
+            );
+
+            $argumentsCount = \count($arguments); // argument count sanity check
+
+            if ($argumentsCount < 2 || $argumentsCount > 4) {
+                continue;
+            }
+
+            $argumentStartIndex = array_keys($arguments)[1]; // get second argument index
+
+            $this->fixFopenFlagToken(
+                $tokens,
+                $argumentStartIndex,
+                $arguments[$argumentStartIndex]
+            );
+        }
+    }
+
+    abstract protected function fixFopenFlagToken(Tokens $tokens, $argumentStartIndex, $argumentEndIndex);
+}
--- a/src/Fixer/FunctionNotation/FopenFlagOrderFixer.php
+++ b/src/Fixer/FunctionNotation/FopenFlagOrderFixer.php
@@ -1,0 +1,127 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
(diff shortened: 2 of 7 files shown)


keradus:
> do we want to have an option to remove flags when found, like `t`
> do we want to have an option to add flags when not found, like `b`

That could be different fixer that would ensure there is always `b` and never `t`.
Splitting it into 2 options doesn't make sense to me - from manual:
> For portability, it is strongly recommended that you always use the 'b' flag when opening files

SpacePossum:
ping @ntzm could you provide a string that shows all possible flags in the correct order? Than this fixer can be finished :)

ntzm:
`b` should always be last (apart from `+`, e.g. `wb`, `wb+`), but the order of the others doesn't matter. Might be worth putting them in alphabetical order for uniformity.

SpacePossum:
thanks @ntzm :)
I think this one should be about finished than, let me know if anything is still open

SpacePossum:
@ntzm anything you would like me to change here?

src/Fixer/FunctionNotation/FopenFlagOrderFixer.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param Tokens $tokens
+     * @param int    $argumentStartIndex
+     * @param int    $argumentEndIndex
+     */
+    private function fixFopenFlagToken(Tokens $tokens, $argumentStartIndex, $argumentEndIndex)
+    {
+        $argumentFlagIndex = null;
+
+        for ($i = $argumentStartIndex; $i <= $argumentEndIndex; ++$i) {
+            if ($tokens[$i]->equalsAny([[T_WHITESPACE], [T_COMMENT], [T_DOC_COMMENT]])) {
```
ntzm: I think `isGivenKind([T_WHITESPACE, T_COMMENT, T_DOC_COMMENT])` may be more optimal here

ntzm:
Think something is wrong with my GitHub notifications, seem to be missing some, sorry about the delay

src/AbstractFopenFlagOrderFixer.php
```diff
@@ -0,0 +1,75 @@
...
+
+use PhpCsFixer\Tokenizer\Analyzer\ArgumentsAnalyzer;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @internal
+ *
+ * @author SpacePossum
+ */
+abstract class AbstractFopenFlagOrderFixer extends AbstractFunctionReferenceFixer
```
dmvdbrugge: I would suggest removing `Order` from this name, as only one of the extending classes actually has to do with ordering.

src/RuleSet.php
```diff
@@ -166,6 +166,7 @@
             'dir_constant' => true,
             'ereg_to_preg' => true,
             'error_suppression' => true,
+            'fopen_flag_order' => true,
```
dmvdbrugge: Do we want the other rule as well?

src/Fixer/FunctionNotation/FopenFlagsFixer.php
```diff
@@ -0,0 +1,93 @@
...
+            $contentQuote = $content[1]; // `'` or `"`
+            $mode = substr($content, 2, -1);
+        } else {
+            $binPrefix = '';
+            $mode = substr($content, 1, -1);
+        }
+
+        if (strlen($mode) > 13) { // 13 === length 'r+w+a+x+c+etb'
+            return; // sanity check to be less risky
+        }
```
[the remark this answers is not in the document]
SpacePossum: line 77 shouldn't be, beside that I think the shared logic shouldn't be made abstract as it will limit other possible fixers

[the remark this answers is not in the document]
dmvdbrugge: As the abstract fixer is called `AbstractFopenFlagFixer` all extensions should operate only on the flags of fopen. I would thus like to make the case that actually *all* logic in this function *except* lines 77-86 should be in the parent class. The method now only needs to take in `string $mode` and return either a fixed or non-altered `$mode`. For the other fixer, the lines are 78-83 (plus the implode of 84).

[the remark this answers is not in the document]
SpacePossum: I disagree; other fixers might want to operate on more tokens, like complex arguments, and not (only) when the mode is set as literal.

src/AbstractFopenFlagFixer.php
```diff
@@ -0,0 +1,75 @@
...
+                break;
+            }
+
+            $index = $candidate[1]; // proceed to '(' of `fopen`
+
+            // fetch arguments
+            $arguments = $argumentsAnalyzer->getArguments(
+                $tokens,
+                $index,
+                $tokens->findBlockEnd(Tokens::BLOCK_TYPE_PARENTHESIS_BRACE, $index)
```
dmvdbrugge: This is already done in `$this->find()` and available as `$candidate[2]`.

SpacePossum: nice catch, fixed

keradus:
Thank you @SpacePossum.

tests/Fixer/FunctionNotation/FopenFlagOrderFixerTest.php
```diff
@@ -0,0 +1,145 @@
...
+                '<?php
+                    $a = \fopen($foo, b"bw+r+c");
+                    $b = \fopen($foo, B"bw+rc");
+                    $c = \fopen($foo, b\'bw+r+c\');
+                    $d = \fopen($foo, B\'bw+rc\');
+                ',
+            ],
+            'common typos' => [
+                '<?php
+                     $a = fopen($a, "b+r");
```
prisis: If i run this new fixer all my `r+b` are now `b+r`.

shouldn't this be `r+b` and not `b+r`?

if you look on http://php.net/manual/en/function.fopen.php you find this

To use these flags, specify either 'b' or 't' as the last character of the mode parameter.

PR #3785, opened 2018-05-28
PhpUnitDedicateAssertFixer - fix to assertCount too

SpacePossum:
closes #3750

```
            'assert same' => [
                '<?php $this->assertCount(1, $a);',
                '<?php $this->assertSame(1, count($a));',
            ],
```

for more example please see the tests

PR #3813, opened 2018-06-05
PhpUnitDedicateAssertFixer - fix "sizeOf" same as "count".

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3785#discussion_r192593870

### DIFF
--- a/tests/Fixer/PhpUnit/PhpUnitDedicateAssertFixerTest.php
+++ b/tests/Fixer/PhpUnit/PhpUnitDedicateAssertFixerTest.php
@@ -353,6 +353,29 @@ $a#
      */
     public function testAssertCount($expected, $input = null)
     {
+        if (null === $input) {
+            $expected = sprintf($expected, 'count');
+        } else {
+            $input = sprintf($input, 'count');
+        }
+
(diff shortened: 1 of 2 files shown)


tests/Fixer/PhpUnit/PhpUnitDedicateAssertFixerTest.php
```diff
@@ -356,6 +356,23 @@ public function testAssertCount($expected, $input = null)
...
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideTestAssertCountCases
+     */
+    public function testAssertCountFromSame($expected, $input = null)
+    {
+        if (null === $input) {
+            $expected = str_replace('count', 'sizeOf', $expected);
```
julienfalque: This will never run any test case with entirely lowercased `sizeof`, which is probably the most common usage. I'd rather rewrite the data provider so that different casings are tested just like with `count`.

SpacePossum: :+1: fixed by using some templating

PR #3815, opened 2018-06-05
NoSuperfluousPhpdocTagsFixer - support array/callable type hints

gharlan:


julienfalque:
Thanks @gharlan! This is an oversight in my initial PR, I would consider this a bugfix. Could you add return types in the PHP 7+ test cases you added ?

keradus:
Thank you @gharlan.

PR #3823, opened 2018-06-09
NativeConstantInvocationFixer - better constant detection

gharlan:
Some previous failing test cases:

```php
M_PI::foo();

function foo(M_PI $a) {}

function foo(): M_PI {}

$foo instanceof M_PI;

class x implements FOO, BAR, BAZ {}

use X\Y\{FOO, BAR as BAR2, BAZ};

class Foo { use Bar, M_PI { Bar::baz insteadof M_PI; } }

M_PI: goto M_PI; // ;)
```


SpacePossum:
how about these as well than ;)
```php
yield M_PI;
yield M_PI => M_PI;
```

gharlan:
@SpacePossum The check is done like this: "All T_STRING tokens are constant invocations, except those matching these conditions".
So the challenge is to find all cases where a T_STRING is _not_ a constant invocation.

But of course it is good to have also different test cases for the actual constant invocations, so I will add your test case! 👍

Slamdunk:
:+1:

Slamdunk:
Shouldn't the new test cases also be added to the fixer's test, so refactoring TokenAnalyzer in the future will not mess up the specific fixer?

julienfalque:
Indeed, they should be added to the fixer's test too.

PR #3824, opened 2018-06-09
DX: Support PHPUnit 7.2

keradus:


phpunit.xml.dist
```diff
@@ -1,13 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <phpunit
-    xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
keradus: without this change, we get:
```
  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 22:
  - Element 'phpunit', attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation': The attribute '{https://www.w3.org/2001/XMLSchema-instance}noNamespaceSchemaLocation' is not allowed.

  Test results may not be as expected.
```

phpunit.xml.dist
```diff
@@ -1,13 +1,12 @@
...
     xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd"
     backupGlobals="false"
     backupStaticAttributes="false"
     beStrictAboutChangesToGlobalState="true"
     beStrictAboutOutputDuringTests="true"
-    beStrictAboutTestSize="true"
```
keradus: option was renamed ages ago, but till 7.2 there were no validation of it

composer.json
```diff
@@ -41,10 +41,10 @@
         "mikey179/vfsStream": "^1.6",
         "php-coveralls/php-coveralls": "^2.1",
         "php-cs-fixer/accessible-object": "^1.0",
-        "php-cs-fixer/phpunit-constraint-isidenticalstring": "^1.0",
-        "php-cs-fixer/phpunit-constraint-xmlmatchesxsd": "^1.0",
+        "php-cs-fixer/phpunit-constraint-isidenticalstring": "^1.0.1",
+        "php-cs-fixer/phpunit-constraint-xmlmatchesxsd": "^1.0.1",
         "phpunit/phpunit": "^5.7.27 || ^6.5.8 || ^7.1",
-        "phpunitgoodpractices/traits": "^1.4",
+        "phpunitgoodpractices/traits": "^1.5",
```
keradus: before 1.5, we were not able to use PHPUnit 7.2

PR #3825, opened 2018-06-09
UX: Provide full diff for code samples

keradus:
Our code samples are super condensed and each line of them has a purpose.

For example, diff for first code sample of `native_function_invocation` is:
```
   --- Original
   +++ New
   @@ -2,9 +2,9 @@
    
    function baz($options)
    {
   -    if (!array_key_exists("foo", $options)) {
   +    if (!\array_key_exists("foo", $options)) {
            throw new \InvalidArgumentException();
        }
    
   -    return json_encode($options);
   +    return \json_encode($options);
    }
   
```

The second sample is this:
```
 * Example #2. Fixing with configuration: ['exclude' => ['json_encode']].
   --- Original
   +++ New
   @@ -2,7 +2,7 @@
    
    function baz($options)
    {
   -    if (!array_key_exists("foo", $options)) {
   +    if (!\array_key_exists("foo", $options)) {
            throw new \InvalidArgumentException();
        }
    
   
```

in fact, they are using very same input code, just different configuration. But in later, we don't know that, as we don't see `json_encode` in diff at all.

------------------------

Proposed PR would change 2nd example into:
```
 * Example #2. Fixing with configuration: ['exclude' => ['json_encode']].
   --- Original
   +++ New
   @@ -1,10 +1,10 @@
    <?php
    
    function baz($options)
    {
   -    if (!array_key_exists("foo", $options)) {
   +    if (!\array_key_exists("foo", $options)) {
            throw new \InvalidArgumentException();
        }
    
        return json_encode($options);
    }
   
```

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -13,7 +13,7 @@
 namespace PhpCsFixer\Console\Command;
 
 use PhpCsFixer\Differ\DiffConsoleFormatter;
-use PhpCsFixer\Differ\UnifiedDiffer;
+use PhpCsFixer\Differ\FullDiffer;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\DefinedFixerInterface;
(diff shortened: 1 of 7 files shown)


SpacePossum:
:+1: for the idea, we could even create a real `full differ` on `sebastianbergmann/diff` one day.

keradus:
What would be the difference in generated output between differ in this PR and that `full differ` you imagine about? TBH, not sure if there is a need - for udiffer, it was important for us, we wanted to use it heavily and in the and, a lot of ppl were looking for udiffer implementation in PHP.
But `full differ` ? Well, there is not that much usage of it in the wild, and here, at our `code samples` case, it's super edge case.
Long story short - not sure if it's worth to spend time on it

PR #3834, opened 2018-06-19
Add --default-indentation option to use other style of indentation su…

johnnytemp:
…ch as 2 spaces. Useful for editor plugin to make it follow indentation of current document.

My use case is that I want to preserve the indentation style of an opened file (which is mostly using tabs originally), during some auto format tidying, such as when typing '}' will auto format the block by default.  I'm using vscode-php-cs-fixer, and I preferred --default-indentation to be passed by that plugin and give the opened file's current indentation.

Creating .php_cs config file, which could force the indentation to a single style, will not work either. It is because some newer files are changed to use 4 spaces indentation.  Letting an older file partially indentation with spaces & some with tabs seems awful.  Follow current file's major indentation is preferred in this project.

### DIFF
--- a/src/Config.php
+++ b/src/Config.php
@@ -21,12 +21,13 @@ use PhpCsFixer\Fixer\FixerInterface;
  */
 class Config implements ConfigInterface
 {
+    private static $defaultIndent = '    ';
     private $cacheFile = '.php_cs.cache';
     private $customFixers = [];
     private $finder;
     private $format = 'txt';
     private $hideProgress = false;
-    private $indent = '    ';
+    private $indent; // default to $defaultIndent
     private $isRiskyAllowed = false;
     private $lineEnding = "\n";
     private $name;
@@ -37,6 +38,14 @@ class Config implements ConfigInterface
     public function __construct($name = 'default')
     {
         $this->name = $name;
+        $this->indent = Config::$defaultIndent;
+    }
+
+    /**
+     * Used by --default-indentation option.  Not recommend to be used in .php_cs config file.
+     */
+    public static function setDefaultIndent($indent) {
+        Config::$defaultIndent = $indent;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Config.php
```diff
@@ -37,6 +38,14 @@ class Config implements ConfigInterface
     public function __construct($name = 'default')
     {
         $this->name = $name;
+        $this->indent = Config::$defaultIndent;
+    }
+
+    /**
+     * Used by --default-indentation option.  Not recommend to be used in .php_cs config file.
+     */
+    public static function setDefaultIndent($indent) {
```
keradus: nope.
statics are bad.
and this looks like damn hack.

src/Console/Command/FixCommand.php
```diff
@@ -127,6 +128,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $passedConfig = $input->getOption('config');
         $passedRules = $input->getOption('rules');
 
+        $indent = $input->getOption('default-indentation');
+        Config::setDefaultIndent($indent);
```
keradus: as you see with other options, options are passed to `ConfigurationResolver` to resolve their final value, after optional option and config consideration.

using this static hook is no go.

-----------------------

For now, we were accepting any `ConfigInterface` instance for this command.
Now, you bounded new input option with concrete implementation - `Config`.
**This new option won't work at all for all ppl using their custom config class.**

johnnytemp: > the whole tool is about standardization and unification, option to bypass it and have different rules for different files doesn't make me happy, personally

I understand your care about ideal standardization.  However, not all cases and user preferences perfectly match the strict model.  That's why it allow set custom rules to disable some PSR2 standardization, and have Config::setIndent().  Having necessary customization is good, so that the tool can be more useful for different people, and more widely used as a good code formatter.

johnnytemp:
Update: changed without static method, as new pull request:
  https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3847

PR #3835, opened 2018-06-21
Fix priority between PHPDoc return type fixers

julienfalque:


tests/Fixtures/Integration/priority/phpdoc_to_return_type,no_superfluous_phpdoc_tags.test
```diff
@@ -0,0 +1,26 @@
...
+--INPUT--
+<?php
+
+class Test {
+    /**
+     * @return int
+     */
+    public function foo() {
+    }
+}
```
SpacePossum: maybe best to update to
```
/**
 * @return int
 */
public function foo() {
    return 1;
}
```

so when the fixer get updates to check if some is actually returned this test still works

julienfalque: Updated.

SpacePossum:
Thanks @julienfalque.

SpacePossum:
thanks @keradus as well :)

PR #3839, opened 2018-06-22
MethodArgumentSpaceFixer - add empty line incorrectly

### LINKED ISSUE #3457: Trailing Space in a comment before the code to be fixed adds unnecessary empty line
Having some code with a potential fix and a comment with a trailing space right before the code, results to an empty line being added in the fix.
If the comment has no trailing space, fixer works as expected.

=> PHP 7.1.8

=> 2.10.0

=> `php-cs-fixer fix a.php`

#### Sample PHP code (a.php):
```php
<?php

function foo() : void
{
    // there is a trailing space in this comment! 
    $var1 = (
        $var2 ?? 0);
}

```

```php
<?php

function foo() : void
{
    // there is a trailing space in this comment!
    $var1 = (
        $var2 ?? 0
 
    );
}

```
> In the resulted code above, there is an empty line between `$var2 ?? 0` and the closing parenthesis

```php
<?php

function foo() : void
{
    // there is a trailing space in this comment!
    $var1 = (
        $var2 ?? 0
    );
}
```

SpacePossum:
closes #3457

### DIFF
--- a/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
+++ b/src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
@@ -285,6 +285,10 @@ final class MethodArgumentSpaceFixer extends AbstractFixer implements Configurat
         return true;
     }
 
+    /**
+     * @param Tokens $tokens
+     * @param int    $startFunctionIndex
+     */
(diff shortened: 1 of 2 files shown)


keradus:
Thank you @SpacePossum.

PR #3847, opened 2018-06-29
Add --default-indentation option to use other style of indentation su…

johnnytemp:
…ch as 2 spaces. Useful for editor plugin to make it follow indentation of current document.  (.php_cs config file's setIndent() still higher precedence)

Pull request update from this: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3834

BTW:  it allow editor plugin to config custom indent char (2 spaces or tab) without overhead of .php_cs

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -188,6 +188,8 @@ Complete configuration for rules can be supplied using a ``json`` formatted stri
 
     $ php php-cs-fixer.phar fix /path/to/project --rules='{"concat_space": {"spacing": "none"}}'
 
+The ``--default-indentation`` option (e.g. pass ``"  "`` or ``"\t"``) will use given indentation instead of 4 spaces as default.
+
 The ``--dry-run`` flag will run the fixer without making changes to your files.
 
 The ``--diff`` flag can be used to let the fixer output all the changes it makes.
--- a/src/Config.php
+++ b/src/Config.php
@@ -26,7 +26,7 @@ class Config implements ConfigInterface
     private $finder;
     private $format = 'txt';
     private $hideProgress = false;
-    private $indent = '    ';
+    private $indent; // initializes to null for a state meaning not explicitly set -> getIndent() will return default '    '.  @see isIndentSet()
     private $isRiskyAllowed = false;
     private $lineEnding = "\n";
     private $name;
@@ -93,10 +93,22 @@ class Config implements ConfigInterface
 
     /**
      * {@inheritdoc}
+     *
+     * If setIndent() is not called with non-null indent, the indent given in
+     * fix command's --default-indentation option is used (for this class only).
+     * Otherwise, it use '    '.
      */
     public function getIndent()
     {
-        return $this->indent;
+        return null !== $this->indent ? $this->indent : '    ';
+    }
+
+    /**
+     * Check whether a (non-null) indent is set via setIndent().
+     */
+    public function isIndentSet()
+    {
+        return null !== $this->indent;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


keradus:
I'm :-1: for this, it's super edge case and it shall be control over control file.

@SpacePossum , @julienfalque , @kubawerlos @ntzm , @localheinz 
do you see value of this change?
(I am not talking about implementation, just about idea)

julienfalque:
:-1: for me as well for the same reasons you gave in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3834#pullrequestreview-129821154.

SpacePossum:
We already allow passing the `rules config` through the `cli` and some other items that are also part of the `config` like `cache` control. So in concept I think it could make sense to allow passing `line ending`, `indenting` and some of the others as well.

However, I think the current issue presented is small and the proposed direction is not inline with where we want to go; a better, more unified configuration setup that can be set through config-file or passed by command line, and where the resolving is very simple (to make, maintain and to understand by end-users). This would make integration with other tooling also better.

So for now; I don't think we should add more features to the command line I/O or configuration in general, as the chances are big that these will be redone in the next major. The more we add, the more we are going to break in the future for our end-users.

johnnytemp:
@SpacePossum Yes, I am of great desire to see a rule config from command line like `indenting` for indent style, as you mentioned.

Only allowing .php_cs to setIndent() in current situation is a limitation to usage.  It makes the tool only suitable for code base which is already using ideally consistent single indentation style.   BTW, I don't think private projects with multiple indentation styles is necessarily minority people, and forcing them to normalize from all over is no good.

Hope it could be solved in future.  Thanks.

johnnytemp:
BTW, I tell you a sad fact.  One of my project agreed to use tabs indentation.  We are not maturely set up config to force to commit in one style, as you may.  One factor which "pollute" our code may be from non-flexible plugin or tools which format code not according to the editor's current indent style, but accordingly to their default indent style.  Other factor may be new editor mis-config.  Thus, after many factors, the code base is mixed with tabs & spaces indentation...

kubawerlos:
Anything against PSR2 would be 👎 from me - this should be our community standard - I remember there was a request here a year back or so to allow `true` and `false` to be uppercase - and we managed to convince an author to make it ~the right~ PSR2 way.

@johnnytemp in one of my old projects we have tabs and all the doubtful stuff - I loved to run `phpcs` against the codebase to see the number of issues - it was over a million.

ntzm:
I don't think this project should cater for people who have mixed indentation. The whole idea of php-cs-fixer is to make code conform to a set of standards, not allow random deviations.

SpacePossum:
@johnnytemp sorry to hear about the `tabs`, most of us got caught in such a project at times so we know the struggle ;)
What might help in the project is to create a `.editorconfig` file and a `.php_cs.dist` file in the root of the project. Juniors not very familiar with setting up IDE's to match the project needs might still be using the right style because of the files. Medior and senior developers are provided with easy tooling to fix the CS to the rules they've made, leaving not much excuse to not do it.
Secondly you could add the fixer and its config in the CI chain to capture more violations.
Still no quarantines it will all work out, but is merely an example of how to maybe improve the situation without this PR having being merged.
I'm closing the PR for now as I think this issue should be part of a bigger configuration overhaul for the next major release. If you see us discussing it or working on it in the (near) future, please join!

Best of luck in the project (and don't get eaten up about CS, in the end work is work ;))

PR #3848, opened 2018-07-02
Add install/update instructions for PHIVE to the README

### LINKED ISSUE #3838: Document PhiVE release channel

SpacePossum:
closes #3838

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,6 +85,15 @@ Globally (homebrew)
 
     $ brew install php-cs-fixer
 
+Locally (PHIVE)
+~~~~~~~~~~~~~~~
+
+Install `PHIVE <https://phar.io>`_ and issue the following command:
+
+.. code-block:: bash
+
+    $ phive install php-cs-fixer # use `--global` for global install
+
 Update
 ------
 
@@ -123,6 +132,13 @@ You can update ``php-cs-fixer`` through this command:
 .. code-block:: bash
 
     $ brew upgrade php-cs-fixer
+
+Locally (PHIVE)
+~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: bash
+
+    $ phive update php-cs-fixer
 
 Usage
 -----
--- a/src/Console/Command/ReadmeCommand.php
+++ b/src/Console/Command/ReadmeCommand.php
@@ -131,6 +131,15 @@ Globally (homebrew)
 
     $ brew install php-cs-fixer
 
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -85,6 +85,15 @@ Globally (homebrew)
...
     $ brew install php-cs-fixer
 
+Locally (PHIVE)
+~~~~~~~~~~~~~~~~~~~
+
+Install `PHIVE <https://phar.io>`_ and issue the following command:
+
+.. code-block:: bash
+
+    $ phive install php-cs-fixer # use `--global` for global install, not recommended
```
keradus: I would drop `, not recommended`. It's opinionated, and even while I agree with it, we don't put that for other ways of installation, don't we ?
Here, i brings the confusion that installing sth with PHIVE globally is not recommended, it's not true.
We simply don't recommend to install dev-tools globally (as their versions depends on project), regardless which way of installation was chosen

keradus: i would keep it

keradus: i would keep it

src/Console/Command/ReadmeCommand.php
```diff
@@ -131,6 +131,15 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
     $ brew install php-cs-fixer
 
+Locally (PHIVE)
+~~~~~~~~~~~~~~~~~~~
```
kubawerlos: Why so long? Other underline-ish lines are with the length of the text above.

SpacePossum: I'll admit, the lengthy `~` is a result of copying another (sub)header, I'll update soon, thanks for the sharp review :+1:   :)

README.rst
```diff
@@ -124,6 +133,13 @@ You can update ``php-cs-fixer`` through this command:
 
     $ brew upgrade php-cs-fixer
 
+Locally (PHIVE)
+~~~~~~~~~~~~~~~~~~~
```
kubawerlos: No so fast, this one is also too long ;)

SpacePossum: oh no, I really really _really_ thought I had this covered, f**k me, this is a lesson learned on my "I wished wasn't needed "-list :(
aaarrggh I'm sorry man, you called me out on it before and still I messed it up, f**k me :(

keradus: no worry. instead of fixing that, let us run rst / md linters on CI :D

Issue #3864, opened 2018-07-03
ordered_imports - sort_function


josephzidell:
When using the `ordered_imports` option and setting to `alpha`, it sorts case-insensitive by using `strcasecmp`. Our application requires sorting by case.

I can see 3 options:

1. Extend the class and override `sortAlphabetically()`, but the class is final. Arghhh
2. Replace the entire class and change that one line. Also Arghhh
3. Adding a configurable option `sort_function` that defaults to `strcasecmp`. I'd be happy to pop a PR.

Thoughts?

josephzidell:
Related to #3591 but different and simpler.

kubawerlos:
> Our application requires sorting by case.

I understood it as first by case, then alphabetically, e.g.
```
use apple;
use basket;
use Aaron;
use Bob;
```
or

```
use Aaron;
use Bob;
use apple;
use basket;
```
so the option would be like `SORT_CASE_APLHA`.

josephzidell:
@keradus @SpacePossum Seems like there is a need for more flexibility, the question being the ideal way to implement. I agree that leaving it open-ended can be dangerous, and want to suggest an alternative to @SpacePossum's idea: we whitelist (and test) certain sort functions, and the user can choose, defaulting to `strcasecmp`. ATM, the only other option would be `strcmp`, and we can allow more if/when others need.

@kubawerlos, the second way

SpacePossum:
The reasons why the maintainers are not really in favor of opening the logic of the rule are of different nature;
- the initial PR opened a way that leaves all control to the user. This could be dangerous, but also when done wrongly might cause other issues that the user doesn't know about on first sight. Like creating priority issues with other fixers, or corrupt internal state of the application (for example by creating an invalid Tokens collection). Even when done right in the current version, we might break it in another version as we change the internals working. We like to keep our BC promise as much as possible ;)
- allowing `strcmp` as input isn't the best UX IMHO, having an option called `alphabetically_case` might be more descriptive. 
- Using an option over a function name will also allow for changing the logic over time when needed, while using the `strcmp` function leaves not a lot of options.
- maybe counter intuitive but the goal of the fixer is not to provide a framework of sorts to use for fixing CS as one likes, we like to support standards or wide use styles (for various reasons), so opening it up for any implementation is not something in the scope

[...]

PR #3865, opened 2018-07-03
Allow custom sorting for ordered imports

josephzidell:
Solves #3864


ntzm:
It's a definite no from me, this would create an injection point in my project [PHP-CS-Fixer](https://github.com/ntzm/PHP-CS-Fixer-Playground) and similar applications

keradus:
Closing, as too many deciders disagree with idea, first on initial issue #3864 , then here in PR.
As SpacePossum said, it's possible to extend, but not to allow everything.

If you will like to implement the feature request, you are more than welcome, yet please first align the direction how to do it in ticket, and when it comes to implement it in new PR - remember that each config of every single fixer has to be documented in readme (here, you created new option, but it's not documented)

Issue #3857, opened 2018-07-03
(Priority?) issue with phpdoc_annotation_without_dot (and whitespace)


dmvdbrugge:
This is either a priority issue, where `no_trailing_whitespace_in_comment` needs to run before `phpdoc_annotation_without_dot`, or the latter should take whitespace into account.

#### The PHP version you are using (`$ php -v`):
```
PHP 7.1.18 (cli) (built: May 25 2018 19:18:59) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.18, Copyright (c) 1999-2018, by Zend Technologies
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
`PHP CS Fixer 2.12.1 Long Journey by Fabien Potencier and Dariusz Ruminski`

#### The command you use to run PHP CS Fixer:
`vendor/bin/php-cs-fixer fix -v --using-cache=no`

#### The configuration file you are using, if any:
Basically the same as [dmvdbrugge/dynamic-components](https://github.com/dmvdbrugge/dynamic-components/blob/5dfba1d7ea9573b2f2d75468556c988019be541c/.php_cs.dist), except `explicit_string_variable`, `header_comment`, `no_alternative_syntax`, and `no_unset_on_property`

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

/**
 * @throws \Exception If something goes wrong.     
 */
function throwMe()
{
}
```

* after first run:
```diff
--- Original
+++ New
@@ @@
 /**
- * @throws \Exception If something goes wrong.     
+ * @throws \Exception If something goes wrong.
  */
 function throwMe()
```

* after second run:
```diff
--- Original
+++ New
@@ @@
 /**
- * @throws \Exception If something goes wrong.
+ * @throws \Exception if something goes wrong
  */
 function throwMe()
```

* expected:
Immediately arrive at the last step

Meta: this is case 13 of #3844

dmvdbrugge:
> following a dot it want to fix

[...]

PR #3867, opened 2018-07-04
PhpdocAnnotationWithoutDotFixer - Handle trailing whitespaces

kubawerlos:
Solves #3857 - I guess @SpacePossum will have some more fancy test cases.

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
@@ -139,6 +139,26 @@ final class PhpdocAnnotationWithoutDotFixerTest extends AbstractFixerTestCase
      *                   and Returns `false` otherwise.
      */',
             ],
+            [
+                '<?php
+    /**
+     * @throws \Exception having whitespaces after dot, yet I am fixed
+     */',
+                '<?php
+    /**
+     * @throws \Exception having whitespaces after dot, yet I am fixed.   '.'
+     */',
+            ],
+            [
+                '<?php
+    /**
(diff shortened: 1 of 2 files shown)


SpacePossum:
this looks awesome to me, thanks so much :)
hope that we get a second review on this

SpacePossum:
(on a side note, if this fixer is about changing text to a sentence without a stop maybe we should unify what we consider a stop (so not only dots), like done [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.2.20/src/Fixer/Phpdoc/PhpdocSummaryFixer.php#L123))

tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
```diff
@@ -137,6 +137,16 @@ public function provideFixCases()
...
+     */',
+            ],
+            [
+                '<?php
+    /**
+     * @throws \Exception having whitespaces after dot, yet I am fixed
+     */',
+                '<?php
+    /**
+     * @throws \Exception having whitespaces after dot, yet I am fixed.   '.'
```
ntzm: Please add test case for trailing tab character too, as we are using `\h`

keradus: ping @kubawerlos

kubawerlos: Sorry, I've missed that. Done.

keradus:
Thank you @kubawerlos.

PR #3873, opened 2018-07-05
Add the native_function_invocation fixer in the Symfony:risky ruleset

stof:
Symfony wants this fixer applied for compiler-optimized functions (we were the ones requesting the feature).

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -43,6 +43,7 @@ $config = PhpCsFixer\Config::create()
         'method_argument_space' => ['on_multiline' => 'ensure_fully_multiline'],
         'method_chaining_indentation' => true,
         'multiline_comment_opening_closing' => true,
+        'native_function_invocation' => false,
         'no_alternative_syntax' => true,
         'no_binary_string' => true,
         'no_extra_blank_lines' => ['tokens' => ['break', 'continue', 'extra', 'return', 'throw', 'use', 'parenthesis_brace_block', 'square_brace_block', 'curly_brace_block']],
--- a/README.rst
+++ b/README.rst
@@ -869,7 +869,7 @@ Choose from the list of available rules:
 
   Function defined by PHP should be called using the correct casing.
 
-* **native_function_invocation**
+* **native_function_invocation** [@Symfony:risky]
 
   Add leading ``\`` before function invocation to speed up resolving.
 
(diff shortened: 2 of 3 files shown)


keradus:
hi @stof, great to see you here again ;)
yeah, discussion about this rule, it's expectation and configuration was longs indeed, finally we manage to cover (most of) users requirements, not only one from Sf ;)
`native_constant_invocation` was already added ;)

be aware that tests are failing, do you want to handle it yourself ?

SpacePossum:
>I'm currently preparing the PR running it.

looking forward to that diff ^_^

stof:
I was fixing the tests, but I see that @SpacePossum already did it.

stof:
@SpacePossum the Symfony PR is referenced just before my comment, in case you want to look at it.

.php_cs.dist
```diff
@@ -43,6 +43,7 @@ $config = PhpCsFixer\Config::create()
         'method_argument_space' => ['on_multiline' => 'ensure_fully_multiline'],
         'method_chaining_indentation' => true,
         'multiline_comment_opening_closing' => true,
+        'native_function_invocation' => false,
```
keradus: to be dropped when number of open PR shrinks down a bit

we do follow Sf ruleset and add some more strict rules on top of it, but we do not exclude things from Sf ruleset.

SpacePossum: could be any good argument to let https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3798 pass

stof: #3798 is indeed the one doing this removal (when it was adding it manually)

keradus: indeed, silly me, forgot about it...
I added milestone there to not lose track of it

keradus:
looks good, yet we have always pushed change of rulesets to MINOR release, changing target to `2.13@dev`

keradus:
it;s really fair config, it does not enforce everything, just what is optimized by compiler.
all good imo

stof:
@SpacePossum the Symfony rule is also configured to favor readability: we only apply it to the compiler-optimized functions, and only in namespaced code (for non-namespaced code, calls are always fully-qualified, and so always optimized).

Btw, that's also the reason for my change in #3876

SpacePossum:
Thank you @stof.

PR #3876, opened 2018-07-05
NativeConstantInvocationFixer - add the scope option

### LINKED ISSUE #3872: NativeConstantInvocationFixer - should not prepend a backslash when the file is not namespaced
If the file is not namespaced, all native calls are already fully-qualified. So prepending a backslash is not necessary there.

stof:
Closes #3872 

The way to implement the option is heavily inspired by the existing feature in NativeFunctionInvocationFixer, except that I reused the NamespacesAnalyzer implementation instead of copy-pasting `\PhpCsFixer\Fixer\FunctionNotation\NativeFunctionInvocationFixer::getUserDefinedNamespaces`

I also updated the Symfony ruleset to restrict the fixer to the namespaced scope as adding the `\` when the code is already in the global scope has no benefit for performance (as both cases are fully qualified) while increasing visual burden (which was the reason to limit it to only a few constants in the Symfony ruleset).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -864,6 +864,8 @@ Choose from the list of available rules:
     ``get_defined_constants``. User constants are not accounted in this list
     and must be specified in the include one; defaults to ``true``
   - ``include`` (``array``): list of additional constants to fix; defaults to ``[]``
+  - ``scope`` (``'all'``, ``'namespaced'``): only fix constant invocations that are made
+    within a namespace or fix all; defaults to ``'all'``
 
 * **native_function_casing** [@Symfony]
 
--- a/src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
+++ b/src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
@@ -18,6 +18,8 @@ use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\NamespaceAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\NamespacesAnalyzer;
 use PhpCsFixer\Tokenizer\Analyzer\NamespaceUsesAnalyzer;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
@@ -48,6 +50,17 @@ final class NativeConstantInvocationFixer extends AbstractFixer implements Confi
             'Add leading `\` before constant invocation of internal constant to speed up resolving. Constant name match is case-sensitive, except for `null`, `false` and `true`.',
             [
                 new CodeSample('<?php var_dump(PHP_VERSION, M_PI, MY_CUSTOM_PI);'.PHP_EOL),
+                new CodeSample(
+                    '<?php
+namespace space1 {
+    echo PHP_VERSION;
+}
+namespace {
+    echo M_PI;
+}
+',
+                    ['scope' => 'namespaced']
+                ),
                 new CodeSample(
                     '<?php var_dump(PHP_VERSION, M_PI, MY_CUSTOM_PI);'.PHP_EOL,
                     [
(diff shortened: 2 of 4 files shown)


stof:
@keradus anything missing here ?

keradus:
Hi @stof , 
what is needed is a review. Sadly, another 85 awaits for my attention, but dont worry - PRs can be reviewed by community as well, it doesn't need to be me ;) Thanks for understanding and patience.

tests/Fixer/ConstantNotation/NativeConstantInvocationFixerTest.php
```diff
@@ -450,6 +450,35 @@ public function testDoNotFixImportedConstants()
     M_PI,
     M_EULER
 );
+EOT;
+
+        $this->doTest($expected, $input);
+    }
+
+    public function testFixScopedOnly()
```
dmvdbrugge: I don't know if it matters anything, but I'm missing a (scoped only) test with no `namespace` keyword at all.

```php
<?php

echo PHP_VERSION . PHP_EOL;
```

Should be untouched

README.rst
```diff
@@ -864,6 +864,8 @@ Choose from the list of available rules:
     ``get_defined_constants``. User constants are not accounted in this list
     and must be specified in the include one; defaults to ``true``
   - ``include`` (``array``): list of additional constants to fix; defaults to ``[]``
+  - ``scope`` (``'all'``, ``'namespaced'``): only fix constant invocations that are made
```
dmvdbrugge: The order of options is `all, namespaced`, however the description order is `namespaced, all`. Could be confusing.

stof: that's exactly the same wording than for the `scope` option of the `native_function_invocation` fixer.

dmvdbrugge: Fix that one too? 😇 I dunno, I'll leave that call to (one of) the maintainers.

keradus:
ping @FriendsOfPHP/php-cs-fixer @localheinz @kubawerlos @ntzm 
can you help with a review here? thanks !

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -225,6 +212,59 @@ protected function createConfigurationDefinition()
                 ->setAllowedValues([$constantChecker])
                 ->setDefault(['null', 'false', 'true'])
                 ->getOption(),
+            (new FixerOptionBuilder('scope', 'Only fix constant invocations that are made within a namespace or fix all.'))
```
ntzm: As we've added this option, can we also add some a test or two to check that `'scope' => 'all'` works correctly

stof: well, all the non-modified tests are running with the `all` scope

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -225,6 +212,59 @@ protected function createConfigurationDefinition()
...
+                ->setAllowedValues(['all', 'namespaced'])
+                ->setDefault('all')
+                ->getOption(),
         ]);
     }
+
+    private function fixConstantInvocations(Tokens $tokens, $start, $end)
```
SpacePossum: missing PHPDoc to typehint `$start` and `$end`

SpacePossum: If a file with multiple namespaces is fixed this method might be called multiple times (based on the new config option). 
The `$useDeclarations` found and the token analyzer (created later on) are based on the complete tokens collection, these can be reused as long as no tokens are inserted.

README.rst
dmvdbrugge: I wasn't comparing anything to anything, I just ment in this specific line in itself. It reads `a, b: description of b, or a.`

SpacePossum: Im not saying that you were, still no change needed IMHO

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
stof: but this method inserts tokens.

Actually, a better way (for both this fixer and the function invocation one) would be to analyze use statements only for the current namespace anyway. but the use statement analyzer does not support that currently.

stof: the composer.json says that PHP-Cs-Fixer supports PHP 5.6+

dmvdbrugge: > PHPDoc

stof: hmm, sorry

keradus:
Thank you @stof.

PR #3877, opened 2018-07-05
NamespacesAnalyzer - Optimize performance

stof:
Instead of looping on all tokens, even inside the namespace it identified, the analyzer now continues the analysis after the end of the identified namespace, thanks to the fact that namespaces cannot be nested.

When implementing #3876, I compared the NamespacesAnalyzer implementation with `\PhpCsFixer\Fixer\FunctionNotation\NativeFunctionInvocationFixer::getUserDefinedNamespaces`. The NamespacesAnalyzer also identifies the namespace name instead of skipping over it (and returns the global namespace too). But I found out that NativeFunctionInvocationFixer was more optimized, as it was not re-analyzing all tokens inside a namespace. This makes NamespacesAnalyzer implement the same optimization.

### DIFF
--- a/src/Tokenizer/Analyzer/NamespacesAnalyzer.php
+++ b/src/Tokenizer/Analyzer/NamespacesAnalyzer.php
@@ -29,7 +29,9 @@ final class NamespacesAnalyzer
     {
         $namespaces = [];
 
-        foreach ($tokens as $index => $token) {
+        for ($index = 1, $count = \count($tokens); $index < $count; ++$index) {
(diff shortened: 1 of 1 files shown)


keradus:
Thank you @stof.

stof:
> We do push code refactoring to oldest maintained branch,

I was not aware of that. Good to know.

keradus:
for next MINOR branch we do push only what MINOR is reserved for - features ;)
all the rest, like bugs, docs, typos, refactoring - to lowest possible branch.

PR #3878, opened 2018-07-05
NativeFunctionInvocationFixer - use the NamespacesAnalyzer to remove duplicated code

stof:
Instead of maintaining 2 different analyzers to find the beginning and the end of a namespace block, this reuses the existing NamespacesAnalyzer.

See #3877 which optimizes it based on existing optimization in the implementation I'm deleting here.


keradus:
Thank you @stof.

Issue #2159, opened 2016-08-26
[RFC] Inline docblock fixer


phansys:
This RFC is for inline docblocks, since there are different ways to write them:

```
/** @var $someVar \SomeClass */
```

```
/* @var \SomeClass $someVar */
```

keradus:
:+1: for idea

phansys:
Somebody knows another notation for inline docblocks? Moreover, anyone knows how to name or identify each of them?

GrahamCampbell:
The bottom option is **not** actual doc. It's just a comment.

keradus:
wait wait wait, I misread it.

@phansys , you want to fix this:
```diff
-/** @var $someVar \SomeClass */
+/** @var \SomeClass $someVar */
```

or this as well:
```diff
-/* @var $someVar \SomeClass */
+/** @var \SomeClass $someVar */
```

?

phansys:
It depends in the result of the "C" part of this RFC @keradus. I know some IDEs (like NetBeans) that treat the 2nd example as a docblock, regardless it isn't really a docblock as @GrahamCampbell pointed. I'm completely open to your concerns about this.

GrahamCampbell:
It's incorrect to fix a comment to a docblock. A comment must never be interpreted as phpdoc, and nobody does that?

phansys:
This is an example of what I mean, but the use case exposed is just for type hinting AFAIK: https://blogs.oracle.com/netbeansphp/entry/defining_a_variable_type_in. Should we ignore these cases then?

GrahamCampbell:
Looks like those people there have made a mistake, or don't know how php works. :trollface:

I suspect they just missed out a star by mistake.

rob006:
> I know some IDEs (like NetBeans) that treat the 2nd example as a docblock

AFAIK **all** IDEs treats the 2nd example as docblock and it is only safe way for define typehints for local variables - NetBeans (and Eclipse?) ignores `/** @var $someVar \SomeClass */`

GrahamCampbell:
Gah. I'm going to declare my position on this as "meh". I give zero f***s about optimizing anything for an IDE. PHPDoc is for humans IMO.

phansys:
You mean Coding Style is not for humans?

GrahamCampbell:
Transforming doc for IDEs is outside the realm of cs. Moving about spacing in phpdoc is cs, but changing the fundamental format is not.

[...]

PR #3881, opened 2018-07-05
Add PhpdocVarAnnotationCorrectOrderFixer

kubawerlos:
Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2159

### DIFF
--- a/tests/Fixer/Alias/NoAliasFunctionsFixerTest.php
+++ b/tests/Fixer/Alias/NoAliasFunctionsFixerTest.php
@@ -40,7 +40,7 @@ final class NoAliasFunctionsFixerTest extends AbstractFixerTestCase
         $cases = [];
 
         foreach (['internalSet', 'imapSet'] as $setStaticAttributeName) {
-            /** @var $aliases string[] */
+            /** @var string[] $aliases */
             $aliases = $this->getStaticAttribute(\PhpCsFixer\Fixer\Alias\NoAliasFunctionsFixer::class, $setStaticAttributeName);
 
             foreach ($aliases as $alias => $master) {
--- a/tests/Fixer/Phpdoc/PhpdocVarAnnotationCorrectOrderFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocVarAnnotationCorrectOrderFixerTest.php
@@ -1,0 +1,165 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\Phpdoc;
+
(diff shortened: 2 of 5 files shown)


tests/Fixer/Phpdoc/PhpdocVarAnnotationCorrectOrderFixerTest.php
```diff
@@ -0,0 +1,130 @@
...
+        ];
+
+        yield [
+            '<?php
+/** @var Foo|Bar|mixed|int $someWeirdLongNAME__123 */
+',
+            '<?php
+/** @var $someWeirdLongNAME__123 Foo|Bar|mixed|int */
+',
+        ];
```
ntzm: Please add cases for changes that are not expected:

```
/** @var */
/** @var $foo */
/** @var Bar */
```

And multi-line descriptions:

```
/**
 * @var $bar Foo long description
 *               goes here
 */
```

dmvdbrugge:
What is the reasoning behind not fixing `@param`?

kubawerlos:
@dmvdbrugge the alternative for 2 is not a acceptable as we cannot assume some will use `PhpdocNoAliasTagFixer`.

`@param` is not fixed as the idea is to fix `@var` cases - I've never seen `@param` with variable name before type and `@param` more often has description than `@var`.

I'm not sure now, maybe we can aim to have `PhpdocCorrectOrderOfTypeAndNameFixer` to handle all Phpdoc types that have type and name the same, yet I prefer to add [`mixed`](https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/v1.5.1/tests/Fixer/PhpdocParamTypeFixerTest.php#L108) when type is missing instead of getting first word of string behind the name as it can do more harm like:

```diff
 /**
- * @param $foo This is foo
+ * @param This $foo us foo
  */

**Comment by dmvdbrugge** [comment]
> the alternative for 2 is not a acceptable as we cannot assume some will use `PhpdocNoAliasTagFixer`.

Was thinking that as well, removed the suggestion.

As for the `@param`, makes sense. So only those 2 points for me.

**Comment by kubawerlos** [comment]
@dmvdbrugge `@type` added, yet I cannot find valid case for priority integration test with `PhpdocVarWithoutNameFixer` - it handles well name with both orders:
```diff
 <?php
 /**
- * @var FooInterface $foo
+ * @var FooInterface
  */
 $foo;
 
 /**
- * @var $bar BarInterface
+ * @var BarInterface
  */
 $bar;
```

tests/Fixer/Alias/NoAliasFunctionsFixerTest.php
```diff
@@ -40,7 +40,7 @@ public function provideFixCases()
         $cases = [];
 
         foreach (['internalSet', 'imapSet'] as $setStaticAttributeName) {
-            /** @var $aliases string[] */
+            /** @var string[] $aliases */
```
dmvdbrugge: So nice that it immediately adds value to the project itself 🎉

keradus:
Thank you @kubawerlos.

phansys:
Thank you @kubawerlos! :tada:

PR #3882, opened 2018-07-05
PhpdocAnnotationWithoutDotFixer - Handle empty line in comment

kubawerlos:
Solves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3857#issuecomment-402697765 - yet, feels too easy - I'm pessimistic about this one - can we easily trick it? @SpacePossum @dmvdbrugge what do you think?

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
@@ -159,6 +159,36 @@ final class PhpdocAnnotationWithoutDotFixerTest extends AbstractFixerTestCase
      * @throws \Exception having tabs after dot, yet I am fixed.		'.'
      */',
             ],
+            [
+                '<?php
+    /**
+     * This is a broken phpdoc
+     * @param string $str surprisingly, it is a string
+
+     */
+    function fixMe($str) {}',
+                '<?php
+    /**
+     * This is a broken phpdoc
+     * @param string $str Surprisingly, it is a string.
+
+     */
+    function fixMe($str) {}',
+            ],
+            [
+                '<?php
+    /**
+     * @return bool|null returns `true` if the class has a single-column ID
+                         Returns `false` otherwise.
+     */
+    function fixMe() {}',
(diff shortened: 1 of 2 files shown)


tests/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixerTest.php
```diff
@@ -139,6 +139,22 @@ public function provideFixCases()
...
+    /**
+     * This is a broken phpdoc
+     * @param string string Surprisingly, it is a string
+
+     */
+    function fixMe($str) {}',
+                '<?php
+    /**
+     * This is a broken phpdoc
+     * @param string string Surprisingly, it is a string.
```
dmvdbrugge: We can see that the fixer works by the removal of the dot, but this should be `string $str` instead of `string string`, and then `Surprisingly` will be lowercased as well.

SpacePossum:
The solutions, this one and resolving the priority, don't have to be exclusive, sorry if I caused confusion about that.

For this PR, I like the direction :+1: 
How would something like

```php
    /**
     * @return bool|null Returns `true` if the class has a single-column ID
                        and Returns `false` otherwise.
     */',
```

be fixed in the new situation?

kubawerlos:
@SpacePossum not nicely - `PhpCsFixer\DocBlock\DocBlock::getAnnotations` will lose 2nd line at all.

SpacePossum:
can you add this to the tests?

```
    /**
     * @return bool|null Returns `true` if the class has a single-column ID.
                         Returns `false` otherwise.
     */
```

If the fixer removes the dot after `ID` but not after `otherwise` I'm happy, if it doesn't removes any dot than it is fine as well for me. It is nice to have a test to see how the fixer handles invalid PHPDocs, that is all :)

SpacePossum:
Thanks for the test, this PR looks good to me.
Maybe another member can have look as well?

keradus:
@kubawerlos , can you rebase please ? we have some conflicts due to recent merging session

keradus:
Thank you @kubawerlos.

PR #3884, opened 2018-07-06
NoSuperfluousPhpdocTagsFixer - handle null in every position

### LINKED ISSUE #3849: Priority issue: phpdoc_types_order should run before no_superfluous_phpdoc_tags
```
```

```php
/**
 * @param null|string $myStr
 */
public function testFunc(?string $myStr);
```

* after first run:
```php
/**
 * @param string|null $myStr
 */
public function testFunc(?string $myStr);
```

* after second run:
```php
public function testFunc(?string $myStr);
```

* expected:

This _could_ also be seen as a bug in `no_superfluous_phpdoc_tags` that it doesn't think `null|string` is superfluous, whereas it (rightfully) does think so about `string|null`.

Meta: this is case 4 of #3844

dmvdbrugge:
… instead of last only.

This fixes #3849

### DIFF
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -255,8 +255,8 @@ class Foo {
     /**
      * Normalizes types to make them comparable.
      *
-     * Converts given types to lowercase and replaces imports aliases with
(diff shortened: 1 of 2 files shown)


julienfalque:
Note: my commit can be squashed when merging :)

SpacePossum:
thanks @dmvdbrugge and @julienfalque :+1:

PR #3885, opened 2018-07-06
Add priority to AlignMultilineCommentFixer

### LINKED ISSUE #3860: Priority: array_indentation should run before align_multiline_comment
```
```

```php
$config->add([
              /*
               * Aligned with the values
               */
    'this' => 'is a bug',
    'but'  => 'this is too',
]);
```

* after first run:
```php
$config->add([
    /*
               * Aligned with the values
               */
    'this' => 'is a bug',
    'but'  => 'this is too',
]);
```

* after second run:
```php
$config->add([
    /*
     * Aligned with the values
     */
    'this' => 'is a bug',
    'but'  => 'this is too',
]);
```

* expected:

Meta: this is case 26 of #3844

dmvdbrugge:
It should run after ArrayIndentationFixer.

This fixes #3860


julienfalque:
Note: my commit can be squashed when merging :)

dmvdbrugge:
I'm not sure I agree with the added commit, it serves no use other than taking away the focus of why it was added in the first place.

julienfalque:
I added invalid indent on other lines to make `array_indentation` usage more explicit. This doesn't change the result in any way.

> the comment is now no longer valid, since the aligning is even more off haha

What do you mean?

dmvdbrugge:
The comment reads `Aligned with the values` (to illustrate the point). That no longer holds true because you invalidated the original alignments.

julienfalque:
The comment is still aligned with values in the expected result, and never was in the input. Adding invalid indent didn't change that.

dmvdbrugge:
Sorry but no, look again: it's aligned with the keys in the expected result. It was aligned with the values in my original input.

julienfalque:
Oh I see what you meant now. But I still prefer the current input: it explicitly shows that indentation of the array changes, while aligning the comment with values (instead of keys) has few value for this test IMO. If we keep current input, the comment should be updated indeed.

dmvdbrugge:
The indentation of the array already changed, the comment gets moved. Or at least the first line of the comment. That is being done by `array_indentation` and is the exact reason this change is needed, look at the example code in the issue. I prefer to not "overcomplicate" the testcase, and only show the reason for the priority.

julienfalque:
Ok, I removed my commit.

dmvdbrugge:
-Just tested `method_chaining_indentation`: it doesn't touch the comment at all (and thus doesn't break alignment).
- For `method_argument_space` I fail to see the case sorry.

Issue #3302, opened 2017-12-06
Snake or camel casing for phpunit tests


BackEndTea:
My proposal is a fixer that enforces either

```php
public function test_it_does_a_test() {}
```
or
```php
public function testItDoesATest() {}
```
for **PHPUnit tests**.

Ive seen the discussion in #1044, but i feel that reducing the scope of a fixer like this to only PHPUnit tests makes it safe to use. 
Any helper functions or dataproviders and the like would not be affected by the fixer, only the tests (either prefixed with test or having a doc block containing ```@test```.

The fixer would only have to adjust the function names it finds, and function names in any ```@depends``` tags in doc blocks.

If there is any desire for a fixer like this i'm willing to shoot a PR this way.

keradus:
is there any big community using underscore convension for tests ?

Slamdunk:
> is there any big community using underscore convension for tests ?

I don't know, but I used camelCase since ever (maybe 10 years) and snake_case since this week, and can be much more readable if the method names are longer than 100 chars, which is a good practice if you have a complex domain with thousand edge cases for every single feature/spec.

keradus:
you are using it globally, or just for test method names?
if later, use proper reporter that do this for you, eg `testdox`

Slamdunk:
Just test method names.
Testdox is  only useful when running tests: I need it to be useful also when reading them.

keradus:
why do you consider long name inside `src` as readable, but in `tests` not?

Slamdunk:
There is a fundamental difference between the two environments: in `src` methods name are composed solely of a verb and a noun because a method shall have a single responsability and a simple way of acting; in `tests` for most of the edge cases we wrote entire phrases explaining what the test code can't explain well or quickly.

PR #3886, opened 2018-07-06
Add PhpUnitMethodCasingFixer

Slamdunk:
Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3302
```diff
 class MyTest extends \PhpUnit\FrameWork\TestCase
 {
-    public function test_my_code() {}
+    public function testMyCode() {}
 }
```

- [x] Implement base fixer
- [x] Take care of `@depends`
- [x] Run after `PhpUnitTestAnnotationFixer`
- [x] Deprecate `PhpUnitTestAnnotationFixer` *case* config

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
@@ -1,0 +1,274 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+use PhpCsFixer\Utils;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpUnitMethodCasingFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @internal
+     */
+    const CAMEL_CASE = 'camel_case';
+
+    /**
+     * @internal
+     */
+    const SNAKE_CASE = 'snake_case';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Enforce camel (or snake) case for PHPUnit test methods, following configuration.',
+            [
+                new CodeSample(
+                    '<?php
+class MyTest extends \\PhpUnit\\FrameWork\\TestCase
+{
+    public function test_my_code() {}
+}
+'
+                ),
+                new CodeSample(
+                    '<?php
+class MyTest extends \\PhpUnit\\FrameWork\\TestCase
+{
+    public function testMyCode() {}
+}
+',
+                    ['case' => self::SNAKE_CASE]
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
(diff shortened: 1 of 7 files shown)


src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -108,6 +108,7 @@ protected function createConfigurationDefinition()
             (new FixerOptionBuilder('case', 'Whether to camel or snake case when adding the test prefix'))
                 ->setAllowedValues(['camel', 'snake'])
                 ->setDefault('camel')
+                ->setDeprecationMessage('Use `php_unit_method_casing` fixer instead.')
```
keradus: can `php_unit_method_casing` do same thing as this option ?

Slamdunk: Well, theoretically no because someone could want to transform `/** @test */ function fooBar` in `function test_fooBar` and only with this option you can do this.
But does this example make sense?

keradus: then, if deprecated feature is not fully represented by new feature, the difference shall be mentioned as addition to "use X instead"

Slamdunk: Sounds right, but I have no idea what could I wrote more to explain the intentions.
If someone used to use *case* option for Annotation fixer, the only possibility that comes into my mind is that the developer is taking care of method casing in PHPUnit classes.
The name of the new fixer is self explanatory, as this is the exact purpose.

keradus: well, actually, when I saw the new rule name, i though it's about proper casing of `assertsame` or `ExPeCtEdExCePtIoN`...

src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
```diff
@@ -0,0 +1,277 @@
...
+     * @param string $functionName
+     *
+     * @return string
+     */
+    private function updateMethodCasing($functionName)
+    {
+        if (self::CAMEL_CASE === $this->configuration['case']) {
+            $newFunctionName = $functionName;
+            $newFunctionName = str_replace('_', ' ', $newFunctionName);
+            $newFunctionName = ucwords($newFunctionName);
```
kubawerlos: `ucwords` takes delimiter as 2nd param, so you can put there `_` and avoid remove `str_replace` above.

src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
```diff
@@ -0,0 +1,277 @@
...
+        } else {
+            $newFunctionName = Utils::camelCaseToUnderscore($functionName);
+        }
+
+        return $newFunctionName;
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int$index
```
kubawerlos: Seems like you have found a bug in `phpdoc_align` - shouldn't this line fail style checking?

src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
```diff
@@ -0,0 +1,277 @@
...
+     * @return bool
+     */
+    private function isTestMethod(Tokens $tokens, $index)
+    {
+        // Check if we are dealing with a (non abstract, non lambda) function
+        if (!$this->isMethod($tokens, $index)) {
+            return false;
+        }
+
+        // if the function name starts with test its a test
```
kubawerlos: `its` -> `it's` or `it is`

src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
```diff
@@ -0,0 +1,277 @@
...
+        $functionNameIndex = $tokens->getNextMeaningfulToken($index);
+        $functionName = $tokens[$functionNameIndex]->getContent();
+
+        if ($this->startsWith('test', $functionName)) {
+            return true;
+        }
+        // If the function doesn't have test in its name, and no doc block, its not a test
```
kubawerlos: 1st `its` is correct, 2nd isn't -> https://data.grammarbook.com/blog/pronouns/1-grammar-error/

src/Fixer/PhpUnit/PhpUnitMethodCasingFixer.php
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * @param string $needle
+     * @param string $haystack
+     *
+     * @return bool
+     */
+    private function startsWith($needle, $haystack)
+    {
+        $len = strlen($needle);
```
kubawerlos: (Personal) I'd drop `$len` here and put `strlen` directly in `substr` to have one-liner.

Slamdunk:
@kubawerlos thank you very much for the review.
Just note that I stole almost 90% of the code from [PhpUnitTestAnnotationFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.12.2/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php) so all your feedback should be addressed there too, and I don't know if I should fix there too or open another PR.

kubawerlos:
@Slamdunk I think no need to fix it here - maybe in separate PR it's worth to give try to initialize this fixer in `PhpUnitTestAnnotationFixer` and proxy the fixing.

keradus:
Thank you @Slamdunk.

SpacePossum:
this tripped https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/419018045#L1863
should be an easy fix

keradus:
i will handle it

keradus:
ref #3979

kaiwa:
Thanks for your work @Slamdunk, exactly what I needed. Snake case in tests ftw ;-)

PR #3890, opened 2018-07-08
StrictParamFixer - make it case-insensitive

kubawerlos:


kubawerlos:
Due to function names being case-insensitive I would rather call this a "bug", not "enhancement".

SpacePossum:
Thanks @kubawerlos.

PR #3891, opened 2018-07-09
PhpdocNoEmptyReturnFixer - account for null[]

### LINKED ISSUE #3869: PhpdocNoEmptyReturnFixer is too greedy
```
```

```php
/**
 * @return string[]|NULL[].
 */
```

* after first run:
```php
/**
 * @return NULL[]|string[]
 */
```

* after second run:
```php
```

* expected:
```php
/**
 * @return string[]|null[]
 */
```

Meta: this is case 48 of #3844

dmvdbrugge:
This fixes #3869

### DIFF
--- a/src/DocBlock/Annotation.php
+++ b/src/DocBlock/Annotation.php
@@ -227,6 +227,22 @@ class Annotation
     }
 
     /**
+     * Get the normalized types associated with this annotation, so they can easily be compared.
+     *
+     * @return string[]
+     */
+    public function getNormalizedTypes()
+    {
+        $normalized = array_map(static function ($type) {
+            return strtolower($type);
+        }, $this->getTypes());
+
+        sort($normalized);
+
+        return $normalized;
+    }
+
+    /**
      * Remove this annotation by removing all its lines.
      */
     public function remove()
--- a/src/Fixer/Phpdoc/PhpdocNoEmptyReturnFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocNoEmptyReturnFixer.php
@@ -17,7 +17,6 @@ use PhpCsFixer\DocBlock\Annotation;
 use PhpCsFixer\DocBlock\DocBlock;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Preg;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
@@ -105,7 +104,9 @@ function foo() {}
      */
     private function fixAnnotation(DocBlock $doc, Annotation $annotation)
     {
-        if (1 === Preg::match('/@return\s+(void|null)(?!\|)/', $doc->getLine($annotation->getStart())->getContent())) {
+        $types = $annotation->getNormalizedTypes();
+
+        if (1 === count($types) && ('null' === $types[0] || 'void' === $types[0])) {
             $annotation->remove();
         }
     }
--- a/tests/DocBlock/AnnotationTest.php
+++ b/tests/DocBlock/AnnotationTest.php
@@ -342,6 +342,29 @@ final class AnnotationTest extends TestCase
         ];
     }
 
+    /**
+     * @param string[] $expected
+     * @param string   $input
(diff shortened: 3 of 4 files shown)


src/Fixer/Phpdoc/PhpdocNoEmptyReturnFixer.php
```diff
@@ -105,7 +105,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
      */
     private function fixAnnotation(DocBlock $doc, Annotation $annotation)
     {
-        if (1 === Preg::match('/@return\s+(void|null)(?!\|)/', $doc->getLine($annotation->getStart())->getContent())) {
+        if (1 === Preg::match('/@return\s+(void|null)(?!\||\[)/', $doc->getLine($annotation->getStart())->getContent())) {
```
keradus: maybe instead of negative lookahead, let us ensure there is whitespace character after `null` ? or use DocBlock component here and rely on logic that is already there...

dmvdbrugge: There doesn't have to be a whitespace character, as discussed before on gitter. But you wanted me to fix that case as separate PR 😉

keradus: still, I believe we shall switch to DocBlock, and if issue exist fix it there, for all fixers, and not here, for single fixer

tests/Fixer/Phpdoc/PhpdocNoEmptyReturnFixerTest.php
```diff
@@ -123,6 +123,19 @@ public function testOtherDoNothing()
...
+
+        $this->doTest($expected);
+    }
+
+    public function testYetAnotherDoNothing()
+    {
+        $expected = <<<'EOF'
+<?php
+    /**
+     * @return null[]|string[]
```
SpacePossum: could you add a test to make sure the change is case-insensitive, i.e. `NULL[]` and `VOID[]` get fixed as well?

dmvdbrugge: Nice catch, they didn't because `Annotation::getTypes` isn't normalized. Fix incoming.

tests/DocBlock/AnnotationTest.php
```diff
@@ -342,6 +342,29 @@ public function provideTypesCases()
         ];
     }
 
+    /**
+     * @param string   $input
+     * @param string[] $expected
+     *
+     * @dataProvider provideNormalizedTypesCases
+     */
+    public function testNormalizedTypes($input, $expected)
```
SpacePossum: please flip the arguments

dmvdbrugge: Done

src/DocBlock/Annotation.php
```diff
@@ -226,6 +226,15 @@ public function setTypes(array $types)
         $this->clearCache();
     }
 
+    public function getNormalizedTypes()
```
SpacePossum: please add a PHPDoc with a return type hint, should be `@return string[]` I think

src/DocBlock/Annotation.php
```diff
@@ -226,6 +226,15 @@ public function setTypes(array $types)
...
+    {
+        $normalized = array_map('strtolower', $this->getTypes());
```
SpacePossum: maybe we want this refactored to not rely on the function name, but I'll leave that to a second reviewer :)

keradus: yes, we don't want to call functions by string with their name, as it's impossible to analyse with static code analysis
please use anonymous function as a wrapper here

ntzm: Would be good if there were a fixer for that ;)

dmvdbrugge: I did it this way because I had seen that in `NativeConstantInvocationFixer`. Want me to fix it there as well, or just here?

dmvdbrugge: Fixed here, let me know if you want me to fix the other one too.

keradus:
Thank you @dmvdbrugge.

PR #3893, opened 2018-07-09
Fix handling /** and */ on the same line as the first and/or last annotation

dmvdbrugge:
In both `Annotation::remove()` and `PhpdocNoEmptyReturnFixer`

This PR is sort-of combined with #3891

### DIFF
--- a/src/DocBlock/Annotation.php
+++ b/src/DocBlock/Annotation.php
@@ -248,7 +248,23 @@ class Annotation
     public function remove()
     {
         foreach ($this->lines as $line) {
-            $line->remove();
+            if ($line->isTheStart() && $line->isTheEnd()) {
+                // Single line doc block, remove entirely
+                $line->remove();
+            } elseif ($line->isTheStart()) {
+                // Multi line doc block, but start is on the same line as the first annotation, keep only the start
+                $content = Preg::replace('#(\s*/\*\*).*#', '$1', $line->getContent());
+
+                $line->setContent($content);
+            } elseif ($line->isTheEnd()) {
+                // Multi line doc block, but end is on the same line as the last annotation, keep only the end
+                $content = Preg::replace('#(\s*)\S.*(\*/.*)#', '$1$2', $line->getContent());
+
+                $line->setContent($content);
+            } else {
+                // Multi line doc block, neither start nor end on this line, can be removed safely
+                $line->remove();
+            }
         }
 
         $this->clearCache();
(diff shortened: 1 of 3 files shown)


src/DocBlock/Annotation.php
```diff
@@ -232,7 +232,23 @@ public function setTypes(array $types)
     public function remove()
     {
         foreach ($this->lines as $line) {
-            $line->remove();
+            if ($line->isTheStart() && $line->isTheEnd()) {
```
dmvdbrugge: I can also make this line a `if (!( ... xor ...)) {` and then remove the else case (because the body is the same), however I chose to do it like this for readability and case separation.

If you want to go for the `! xor`, tell me and I'll change it.

keradus: no xor please, operators like `or` or `and` have different prios than `||` or `&&`

tests/DocBlock/AnnotationTest.php
```diff
@@ -213,6 +213,46 @@ public function provideRemoveCases()
         return $cases;
     }
 
+    /**
+     * @param string $content
+     * @param string $expected
+     *
+     * @dataProvider provideRemoveEdgeCasesCases
+     */
+    public function testRemoveEdgeCases($content, $expected)
```
SpacePossum: please flip the arguments to `$expected, $input` to be consistent with the rest of the code base

dmvdbrugge: Done

keradus:
@dmvdbrugge , @kubawerlos ,

are this and #4035 PRs handling the very same issue? how shall we proceed on those two ?

dmvdbrugge:
Seems to me like they could live together, they don't even touch the same files. They also solve a problem that looks the same but is not entirely equal, even though it could result in the same outcome.

My suggestion would be just merge both.

kubawerlos:
I'd like them both to be merged as well.

Reading them again it got me thinking why `Line::isTheStart` checks if `false !== strpos($this->content, '/**');`? Wouldn't make more sense to check if it is `0`?

dmvdbrugge:
> Wouldn't make more sense to check if it is `0`?

Can't because of whitespace, I think. Could make it a regex though.

I hadn't even looked into the `Line` class itself, but now you caused me to 😉and I see that `containsUsefulContent` excludes start and end lines, which it should not (because as we both discovered, start and end lines actually could contain useful content). Shall I fix that in this PR too?

julienfalque:
@dmvdbrugge I allowed myself to rebase your work to fix a conflict. I also rewrote test cases in `AnnotationTest` to make them easier to read. For me this can be merged now.

SpacePossum:
Thank you @dmvdbrugge.

PR #3894, opened 2018-07-09
Add the lowest maintained branch to CONTRIBUTING.md

dmvdbrugge:


SpacePossum:
I'm +0 here, because this information will be outdated every minor and the current older branches (which cause the confusion in the first place) won't show this information.

dmvdbrugge:
> this information will be outdated every minor

Is that so? 2.2 has been maintained until only recently.

> and the current older branches (which cause the confusion in the first place) won't show this information.

I disagree: when you go to just the "homepage" of the repo, **_that_** contributing should always be leading. No source of confusion.

There probably is enough internal notice or memo's or dunno what, but to outsiders it's not clear what the current lowest is. Only other thing you can do is to always remove unmaintained branches, but as we speak, both 2.2 and 2.11 are still there. How else should I know what branch to target?

ntzm:
This information needs to be somewhere, as it isn't obvious. It's not too much effort so long as someone remembers to update it, so 👍 here

SpacePossum:
>There probably is enough internal notice or memo's or dunno what, but to outsiders it's not clear what the current lowest is.

maybe and on the second part; I never disputed that

> Only other thing you can do is to always remove unmaintained branches, but as we speak, both 2.2 and 2.11 are still there. How else should I know what branch to target?

well no; it not the only thing, another way could be to add to the no longer supported branched `contributing` file that .. well. it is no longer support

SpacePossum:
>as someone remembers to update 

I hope you'll ping us on that @ntzm on the releases ;)

dmvdbrugge:
> another way could be to add to the no longer supported branched `contributing` file that .. well. it is no longer support

Which means that contributors should always look through all existing branches to see if it's still supposed to be maintained. Also: updating no-longer-maintained-branches' contributing to reflect they're no longer maintained is almost harder to forget to do and as much work as just updating the "main" `contributing` file.

dmvdbrugge:
> Specify that lowest maintained branch === lowest existing branch and that's it.

In that case I'm just closing, as that essentially is what it already says, and holds true again now that you deleted the leftover branches 😉

keradus:
Well, removal of letfover branches was not trivial, as if I would just delete them - all PR against them would be automatically closed. For that, we kept them so far.

But yes, I agree that it brought the confusion, I did manually change target of them to 2.12 and drop 2.2 / 2.11.

Thanks for bringing the topic to the table :+1:

PR #3895, opened 2018-07-09
FunctionsAnalyzer - false positive for constant and function definition

kubawerlos:
Class is `@internal` so rename is fine.


SpacePossum:
Thanks @kubawerlos.

PR #3906, opened 2018-07-12
Handle unset index in FunctionsAnalyzer::isGlobalFunctionCall

kubawerlos:
Third time lucky with the function?


src/Tokenizer/Analyzer/FunctionsAnalyzer.php
```diff
@@ -29,6 +29,10 @@
      */
     public function isGlobalFunctionCall(Tokens $tokens, $index)
     {
+        if (!isset($tokens[$index])) {
+            return false;
```
keradus: throw invalid argument, if the index does not exist in the collection

keradus:
So, I do see some nice changes in Analyzer, but what actually fails due to that issue? any fixer?

kubawerlos:
Ha, I thought I can achieve it in one "nasty" fixer I'm working on, but it is still called inside of token's collection loop.

It seems it is only to make the function bullet proof then.

Issue #3904, opened 2018-07-11
 Add ImplodeCallFixer


kubawerlos:
Function `implode` can be called in 3 ways, one - not [documented](http://php.net/manual/en/function.implode.php) - which should be fixed:
```diff
 <?php
-implode($pieces , '');
+implode("", $pieces );
```

and two that are correct:
```
<?php
implode("", $pieces);
implode($pieces);
```

How should we handle these 2 signatures? I could not find any conclusive reason one is better than the other - should the fixer have an option `empty_string` with values `ensure` and `remove`? If so what would be the default (and why?) and would that default be our default here?

Or maybe one signature is somehow better and should be recommended?

SpacePossum:
:+1:  for fixing to `implode("", $pieces);` , as it is more explicit and used in same way as the counter part `explode`

gharlan:
Maybe `implode($pieces , '')` will be deprecated in 7.4:

https://wiki.php.net/rfc/deprecations_php_7_4

> The following are still Work-In-Progress and ideas:
> * Unify parameter order for implode() to match explode()

kubawerlos:
The `implode($pieces , '')` case is obvious how to handle - I'm asking how to approach 2 correct, documented usages.

gharlan:
I agree to @SpacePossum, we should fix it to the explicit one `implode('', $pieces)`.
I would not add a configuration option for it, until it is explicitly demanded by someone who wants to use `implode($pieces)`.

gharlan:
[...]

PR #3907, opened 2018-07-12
Add ImplodeCallFixer

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3904

Adding to `@Symfony` rule set was suggested by @keradus.


src/Fixer/FunctionNotation/ImplodeCallFixer.php
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $functionNameIndex
+     *
+     * @return int[]
+     */
+    private function getArgumentIndices(Tokens $tokens, $functionNameIndex)
```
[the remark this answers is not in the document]
kubawerlos: Indeed, with a small change, but saves few lines.

README.rst
```diff
@@ -714,6 +714,11 @@ Choose from the list of available rules:
 
   Convert ``heredoc`` to ``nowdoc`` where possible.
 
+* **implode_call** [@Symfony]
+
+  Function ``implode`` must be called with 2 arguments in the documented
+  order.
+
```
keradus: `Risky rule: risky when the function ``implode`` is overridden.`

src/Fixer/FunctionNotation/ImplodeCallFixer.php
```diff
@@ -0,0 +1,139 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $functionsAnalyzer = new FunctionsAnalyzer();
+
+        foreach ($tokens as $index => $token) {
+            if (!$functionsAnalyzer->isGlobalFunctionCall($tokens, $index)) {
```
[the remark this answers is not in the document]
kubawerlos: I want all the global function checking to be delegated to `FunctionsAnalyzer` class (so it can be reused e.g in `CombineNestedDirnameFixer` and here). Maybe other function `isGlobalFunctionCallWithName` or 3rd parameter to `isGlobalFunctionCall`?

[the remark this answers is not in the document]
keradus: For reasoning of this PR - I see no value of extending the Analyzer.
it can be raised as idea in separated PR, which would proof it's value (eg, simplify more than one place)

dmvdbrugge:
Is it too much to ask for config to only fix order, not add 2nd param when it's not there? (I like my 1-param implodes...)

keradus:
IMO, let us finish it first config-less, to have working MVP, then we can extend it, if needed

tests/Fixer/FunctionNotation/ImplodeCallFixerTest.php
```diff
@@ -0,0 +1,106 @@
...
+        yield [
+            '<?php
+                implode(
+                    \'\', $foo
+                );',
+            '<?php
+                implode(
+                    $foo
+                );',
+        ];
```
SpacePossum: please add;
```php
<?php
implode(# 1
$foo/* 2.1 */,# 2.2
""# 3
);
```
and
```php
implode(# 1
# 2
$foo# 3
# 4
)# 5
;
```

SpacePossum: Thanks for adding the tests :)
Sorry for dropping another request, I just thought of it, can you add a test with a lot of candidates?
Something like;
```php
<?php
implode($a);implode($a);implode($a);implode($a);implode($a);implode($a);
// comment for testing
implode($a);implode($a);implode($a);implode($a);implode($a);implode($a);
```

kubawerlos: No problem, added.

keradus:
Thank you @kubawerlos.

SpacePossum:
Thanks @kubawerlos , great fixer and love to see your work going from 3rd party fixer to the core so even more people will enjoy it!
On projects at work this already kicked over some CI's and after fixing we now enjoy more clean code :D
:+1:

PR #3908, opened 2018-07-12
StrictParamFixer - fix edge case

kubawerlos:
No matter how stupid, this is valid PHP syntax, so should be fixed, right?


SpacePossum:
I would say so, the current [`return` ](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/src/Fixer/Strict/StrictParamFixer.php#L80) statement seems off.
Having a test case with a file containing multiple blocked-namespaces which include function imports will show more issues around this.

keradus:
Thank you @kubawerlos.

Issue #3161, opened 2017-10-18
Order @param annotations in the order they are defined in the method


ntzm:
```php
/**
 * @param string $foo
 * @param string $bar
 */
function baz(string $bar, string $foo) { }
```

should be turned into

```php
/**
 * @param string $bar
 * @param string $foo
 */
function baz(string $bar, string $foo) { }
```

grubersjoe:
I am also highly interested in this feature.

A basic question: Would this be implemented in a new `PhpdocParamOrderFixer` class or could this be integrated into the existing `PhpdocOrderFixer`? I'm thinking about creating a MR.

julienfalque:
Great to hear @grubersjoe! :)

IMO this should be a dedicated fixer, `PhpdocParamOrderFixer` would be fine.

Wirone:
Implemented in #3909 and will be available in the upcoming release 🙂.

PR #3909, opened 2018-07-12
feature: Order of PHPDoc @param annotations

grubersjoe:
This PR addresses #3161. Another quite old PR is still open to fix the same issue, but seems inactive: #2385.

The fixer reorders `@param` annotations in the DocBlock according to the formal parameters of the function / method.

Example:

```php
<?php
class C {
    /**
     * @param bool   $a
     * @param string $c
     * @param string $b
     */
    public function foo($a, $b, $c) {}
}
```

to

```php
class C {
    /**
     * @param bool   $a
     * @param string $b
     * @param string $c
     */
    public function foo($a, $b, $c) {}
}
```
Superfluous `@param` annotations are ignored and listed in previous order after the valid ones.

Thoughts?


tests/Fixer/Phpdoc/PhpdocParamOrderFixerTest.php
```diff
@@ -0,0 +1,473 @@
...
+    /**
+     * @param $b
+     * @param bool $c
+     * @param int $a
+     */
+    public function m(int $a, $b, bool $c) {}
+}
+EOT;
+        $this->doTest($expected, $input);
+    }
```
ntzm: Please add tests for comments between tokens:
```php
public/*1*/function/*2*/m/*3*/(/*4*/){}
```

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,174 @@
...
+        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Orders all @param annotations in DocBlocks according to method signature.',
```
ntzm: Please surround `@param` with backticks so it is formatted correctly

grubersjoe: Done

tests/Fixer/Phpdoc/PhpdocParamOrderFixerTest.php
grubersjoe: I've added one.

grubersjoe:
Should I squash all commits into a single one?

dmvdbrugge:
> Superfluous @param annotations are ignored and listed in previous order after the valid ones.

Why? I'd say reorder everything correctly, and let the superfluous fixer remove the superfluous ones. If that fixer is not enabled (for reasons) I think they would want them to be sorted correctly...

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,174 @@
...
+
+        // Concatenate the sorted annotations
+        $orderedAnnotations = $validParams + $invalidParams;
+
+        // Rewrite the param annotations
+        foreach ($paramAnotations as $i => $docAnnotation) {
+            $docAnnotation->remove();
+            $doc
+                ->getLine($docAnnotation->getStart())
+                ->setContent($orderedAnnotations[$i]);
```
dmvdbrugge: The fixer crashes on this line if I test it with a nested doc block.

I added this test case
```php
    public function testNestedPhpdoc()
    {
        $expected = <<<'EOT'
<?php

/**
 * @param string[] $array
 * @param callable $callback {
 *     @param string $value
 *     @param int    $key
 *     @param mixed  $userdata
 * }
 * @param mixed    $userdata
 *
 * @return bool
 */
function string_array_walk(array &$array, callable $callback, $userdata = null) {}
EOT;

        $input = <<<'EOT'
<?php

/**
 * @param callable $callback {
 *     @param string $value
 *     @param int    $key
 *     @param mixed  $userdata
 * }
 * @param string[] $array
 * @param mixed    $userdata
 *
 * @return bool
 */
function string_array_walk(array &$array, callable $callback, $userdata = null) {}
EOT;

    }
```
And the result is this:
```
Undefined offset: 2
 php-cs-fixer/vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:108
 php-cs-fixer/src/Fixer/Phpdoc/PhpdocParamOrderFixer.php:150
 php-cs-fixer/src/Fixer/Phpdoc/PhpdocParamOrderFixer.php:79
 php-cs-fixer/src/AbstractFixer.php:75
 php-cs-fixer/tests/Test/AbstractFixerTestCase.php:127
 php-cs-fixer/tests/Fixer/Phpdoc/PhpdocParamOrderFixerTest.php:572
 php-cs-fixer/vendor/phpunitgoodpractices/traits/src/ExpectationViaCodeOverAnnotationTrait7.php:39
```

dmvdbrugge: I suggest you fix only "first level" annotations, the same way as can be seen [in PR 3911](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911/files#diff-4aae1b68fe4f27312caade85cd01c158R115).

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * @param Annotation[] $docParams
+     * @param string       $identifier
+     *
+     * @return null|int
+     */
+    private function findParamAnnotationByIdentifier(array $docParams, $identifier)
+    {
+        $regex = sprintf('/\*\s*@param\s*\w*\s*\%s/', $identifier);
```
kubawerlos: It would be cleaner to use `preg_quote` instead of escaping `%s`.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
kubawerlos: Can you add a test case with `@param Foo|Bar $x`, `@param ?Foo $x` and `@param Foo[]|\Bar\Baz $x`?

grubersjoe: @dmvdbrugge I'll look into it.

@kubawerlos Do you mean a testcase like this (I'm a bit confused about the three `$x`)?

```php
<?php
class C {
    /**
     * @param Foo|Bar $y
     * @param ?Foo $z
     * @param Foo[]|\Bar\Baz $x
     */
    public function m($x, $y, $z) {}
}
```

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
grubersjoe: Good call. The regex has also changed to support more complex types like `Foo[]|\Bar\Baz`. It doesn't check the order of operators, though. I'm not sure if it's necessary to take this into account. Invalid `@param` annotations would be possible right now, but I don't know if it should be the fixer's task to check validity thoroughly.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
kubawerlos: Yeah, some mix of really "creative" types.

kubawerlos:
Another case (obfuscated from an existing project, I wouldn't came with such case by myself, I'm not that creative) that will crash the fixer:
```php
<?php
/**
 * @param Foo $fooBar
 * @param $fooSomethingNotMatchingTheName
 * @param OtherClassLorem $x
 */
function f(Foo $fooBar, Payment $foo, OtherClassLoremIpsum $y)
{
}
```

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,208 @@
...
+            preg_quote('[]|\?'),
+            substr($identifier, 1) // remove starting `$`
+        );
+
+        foreach ($docParams as $i => $param) {
+            $blockStart = Preg::match('/\s*{\s*$/', $param->getContent());
+            $blockEnd = Preg::match('/\s*}\s*$/', $param->getContent());
+
+            if ($blockStart) {
+                $inBlock = true;
```
dmvdbrugge: The fact that you only do true/false for `$inBlock` suggests that this still breaks for a double-nested (or more) doc block. Nesting could happen infinitely deep, and params in a nested block could have the same name as one block higher in the chain. I don't know if this is a case we want to support, but the spec allows it.

grubersjoe: You are right. I've adapted the code to support deeply nested Docblocks and added tests.

grubersjoe:
That's right @kubawerlos. I forgot to also check for types like `Foo<Bar, Baz>`. I run the fixer for the whole project and now only correct fixes are performed (see diff below). Some annotations are superfluous and listed behind (as intended). For example `$isPhpdocCandidate` in `CommentsAnalyzerTest::testPhpdocCandidate`.

From my perspective the fixer seems to work correctly now and might be merged.

```diff
diff --git a/src/Fixer/Alias/SetTypeToCastFixer.php b/src/Fixer/Alias/SetTypeToCastFixer.php
index a02f78a7..40327314 100644
--- a/src/Fixer/Alias/SetTypeToCastFixer.php
+++ b/src/Fixer/Alias/SetTypeToCastFixer.php
@@ -196,15 +196,15 @@ private function removeSettypeCall(
         $tokens->clearAt($functionNameIndex); // we'll be inserting here so no need to merge the space tokens
         $tokens->clearEmptyTokens();
     }
 
     /**
      * @param Tokens $tokens
-     * @param Token  $castToken
      * @param int    $functionNameIndex
      * @param Token  $argumentToken
+     * @param Token  $castToken
      */
     private function fixSettypeCall(
         Tokens $tokens,
         $functionNameIndex,
         Token $argumentToken,
         Token $castToken
diff --git a/tests/Doctrine/Annotation/TokenTest.php b/tests/Doctrine/Annotation/TokenTest.php
index 0b94857c..0f544cb2 100644
--- a/tests/Doctrine/Annotation/TokenTest.php
+++ b/tests/Doctrine/Annotation/TokenTest.php
@@ -93,14 +93,14 @@ public function provideIsTypeCases()
         ];
     }
 
     /**
      * @dataProvider provideIsNotTypeCases
      *
-     * @param int       $type
      * @param int|int[] $types
+     * @param int       $type
      */
     public function testIsTypeReturnsFalse($types, $type)
     {
         $token = new Token();
 
         $token->setType($type);
diff --git a/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php b/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
index a99d49ce..3652e649 100644
--- a/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
+++ b/tests/Fixer/ClassNotation/OrderedClassElementsFixerTest.php
@@ -327,15 +327,15 @@ protected function abc() {
 EOT
             ],
         ];
     }
 
     /**
+     * @param array       $configuration
      * @param string      $expected
      * @param null|string $input
-     * @param array       $configuration
      *
      * @dataProvider provideFix71Cases
      * @requires PHP 7.1
      */
     public function testFix71(array $configuration, $expected, $input = null)
     {
diff --git a/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php b/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
index 0c817133..b9ef9a22 100644
--- a/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
+++ b/tests/Fixer/ControlStructure/NoUselessElseFixerTest.php
@@ -773,14 +773,14 @@ public function provideConditionsWithoutBracesCases()
         ];
 
         return $cases;
     }
 
     /**
-     * @param string            $input
      * @param string<int, bool> $indexes
+     * @param string            $input
      *
      * @dataProvider provideIsInConditionWithoutBracesCases
      */
     public function testIsInConditionWithoutBraces($indexes, $input)
     {
         $reflection = new \ReflectionObject($this->fixer);
diff --git a/tests/RuleSetTest.php b/tests/RuleSetTest.php
index eb357dbd..42e02269 100644
--- a/tests/RuleSetTest.php
+++ b/tests/RuleSetTest.php
@@ -33,14 +33,14 @@ public function testCreate()
         $ruleSet = RuleSet::create();
 
         $this->assertInstanceOf(\PhpCsFixer\RuleSet::class, $ruleSet);
     }
 
     /**
-     * @param string     $ruleName
      * @param string     $setName
+     * @param string     $ruleName
      * @param array|bool $ruleConfig
      *
      * @dataProvider provideAllRulesFromSetsCases
      */
     public function testIfAllRulesInSetsExists($setName, $ruleName, $ruleConfig)
     {
@@ -67,14 +67,14 @@ public function testIfAllRulesInSetsExists($setName, $ruleName, $ruleConfig)
         } catch (InvalidForEnvFixerConfigurationException $exception) {
             // ignore
         }
     }
 
     /**
-     * @param string $ruleName
      * @param string $setName
+     * @param string $ruleName
      *
      * @dataProvider provideAllRulesFromSetsCases
      */
     public function testThatThereIsNoDeprecatedFixerInRuleSet($setName, $ruleName)
     {
         $factory = new FixerFactory();
diff --git a/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php b/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
index c19f5ead..aee24976 100644
--- a/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
+++ b/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
@@ -181,14 +181,14 @@ public function testPhpdocCandidateAcceptsOnlyComments()
         $this->expectException(\InvalidArgumentException::class);
 
         $analyzer->isBeforeStructuralElement($tokens, 2);
     }
 
     /**
-     * @param bool   $isPhpdocCandidate
      * @param string $code
+     * @param bool   $isPhpdocCandidate
      *
      * @dataProvider providePhpdocCandidateCases
      */
     public function testPhpdocCandidate($code)
     {
         $tokens = Tokens::fromCode($code);
@@ -233,14 +233,14 @@ public function providePhpdocCandidateCases()
             ['<?php /* @var string $s */ print($s);'],
             ['<?php /* @var string $s */ echo($s);'],
         ];
     }
 
     /**
-     * @param bool   $isPhpdocCandidate
      * @param string $code
+     * @param bool   $isPhpdocCandidate
      *
      * @dataProvider provideNotPhpdocCandidateCases
      */
     public function testNotPhpdocCandidate($code)
     {
         $tokens = Tokens::fromCode($code);
diff --git a/tests/Tokenizer/TokensTest.php b/tests/Tokenizer/TokensTest.php
index 3456c918..85ca4256 100644
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -511,14 +511,14 @@ public function bar()
         $this->assertCount(1, $found[T_FUNCTION]);
         $this->assertArrayHasKey(9, $found[T_FUNCTION]);
     }
 
     /**
      * @param string  $source
-     * @param Token[] $expected tokens
      * @param int[]   $indexes  to clear
+     * @param Token[] $expected tokens
      *
      * @dataProvider provideGetClearTokenAndMergeSurroundingWhitespaceCases
      */
     public function testClearTokenAndMergeSurroundingWhitespace($source, array $indexes, array $expected)
     {
         $this->doTestClearTokens($source, $indexes, $expected);
```

grubersjoe:
Okay I added the fixer it to `.php_cs.dist` and squashed all commits. Or should this be done in another PR?

SpacePossum:
In the opening of this PR a reference is made to another PR that has staled.
There are a few points there that are relevant for the PR as well:
- dealing with `@param` 's that are overcomplete, i.e. these reference a parameter that is not part of the signature
- dealing with missing `@param`'s (since there is a fixer that (might) add those, see below)

I'm not stating this fixer should add or remove `@params` (IMHO it shouldn't!), but we should check how this fixer deals with these scenarios and lock it down in unit test. Secondly, there might be a priority issue here with `no_superfluous_phpdoc_tags` and `phpdoc_add_missing_param_annotation` I think. Maybe someone could look into this? 

I really like to have this fixer in :)

grubersjoe:
@SpacePossum I already added test cases for these two scenarios:

```php
public function testOnlyParamsUndocumented()
    {
        $expected = <<<'EOT'
<?php
class C {
    /**
     * @param $a
     * @param $b
     * @param $c
     * @param $d
     */
    public function m($a, $b, $c, $d, $e, $f) {}
}
EOT;
        $input = <<<'EOT'
<?php
class C {
    /**
     * @param $a
     * @param $c
     * @param $d
     * @param $b
     */
    public function m($a, $b, $c, $d, $e, $f) {}
}
EOT;
    }

    public function testOnlyParamsSuperfluousAnnotations()
    {
        $expected = <<<'EOT'
<?php
class C {
    /**
     * @param $a
     * @param $b
     * @param $c
     * @param $superfluous2
     * @param $superfluous1
     * @param $superfluous3
     */
    public function m($a, $b, $c) {}
}
EOT;
        $input = <<<'EOT'
<?php
class C {
    /**
     * @param $a
     * @param $superfluous2
     * @param $b
     * @param $superfluous1
     * @param $c
     * @param $superfluous3
     */
    public function m($a, $b, $c) {}
}
EOT;
    }
```

Superfluous parameters are listed in previous order after the valid @param annotations. As you said I also think that it's not the concern of this fixer to add or remove annotations. Other fixers should handle that if need be.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,209 @@
...
+        foreach ($tokens as $index => $token) {
+            if (!$token->isGivenKind(T_DOC_COMMENT)) {
+                continue;
+            }
+
+            // Check for function signature
+            $functionSequence = $tokens->findSequence([[T_FUNCTION], [T_STRING], '('], $index);
+
+            if (null === $functionSequence) {
+                continue;
```
SpacePossum: Since no `end` has been given for the sequence find the result `null` means no more functions are found in the token collection. Therefore I think a `break` or even a `return` can be done here.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Fetches the tokens of a function signature.
+     *
+     * @param Tokens            $tokens
+     * @param array<int, Token> $functionSequence
+     *
+     * @return array
+     */
+    private function getFuncParameterTokens(Tokens $tokens, array $functionSequence)
```
SpacePossum: maybe `PhpCsFixer\Tokenizer\Analyzer\ArgumentsAnalyzer::getArguments` can be used for this?

grubersjoe: I checked this out. Unfortunately this function works differently. It "returns start and end token indexes of arguments". So ` $c = null` (note the leading space) leads to a single space as start and `null` as end. So in the following case you get `null` instead of `$c`:

```php
function m(array &$a, callable $b, $c = null, $d) {}

// $paramNames =
Array
(
    [8] => $a
    [13] => $b
    [18] => null
    [25] => $d
)
```

Therefore I would keep my function, since it seems to work fairly well.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
grubersjoe: Alright, I used a `return`.

src/Fixer/Phpdoc/PhpdocParamOrderFixer.php
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @param Token   $docBlockToken
+     * @param Token[] $funcParamTokens
+     *
+     * @return DocBlock
+     */
+    private function sortDocBlockParamAnnotations(Token $docBlockToken, array $funcParamTokens)
+    {
+        $doc = new DocBlock($docBlockToken->getContent());
+        $paramAnotations = $doc->getAnnotationsOfType('param');
```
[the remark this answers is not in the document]
grubersjoe: Yes, that's better. Done.

grubersjoe:
I added all reqested changes @SpacePossum. What's next :)?

kubawerlos:
Can you do `rebase`? Right now diff shows adding redundant docs in `tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php`

grubersjoe:
Done

grubersjoe:
Done

SpacePossum:
@grubersjoe thanks for all the work :D

>What's next :)?

What is next is for some of the maintainers to find some time to test and review a bit more,
currently we're are all more busy than we can keep up with all the contributions (luxury problem :) ) so please bear with us a bit longer!

ntzm:
Here's some test cases I'd like to see:

## Other annotations mixed in with params

```
/**
 * @param mixed $a
 * @throws FooException
 * @param mixed $b
 */
public function foo($b, $a) {}
```

## Non-public methods

- Private methods
- Protected methods
- Static methods
- Final methods
- Abstract methods

## Interfaces

- Methods defined on interfaces

## Closures

```
/**
 * @param mixed $b
 * @param mixed $a
 */
$foo = function ($a, $b) {
};
```

grubersjoe:
I've begun to add tests for these cases. The first category "Other annotations mixed in with params" is rather tricky and I could use some advice. This is the current test case:

```php
$expected = <<<'EOT'
<?php
/**
 * Method description
 * over multiple lines
 *
 * @param int   $a Multiline param
 *                 description
 * @param mixed $b
 * @param mixed $superflous
 * @return array
 * @throws Exception
 */
function foo($a, $b) {}
EOT;
        $input = <<<'EOT'
<?php
/**
 * Method description
 * over multiple lines
 *
 * @param mixed $b
 * @param mixed $superflous
 * @return array
 * @throws Exception
 * @param int   $a Multiline param
 *                 description
 */
function foo($a, $b) {}
EOT;
```
I think it is reasonable to list the ordered @param annotations first (superflous ones afterwards)  and then all other annotations and tags in their previous order (@return and @throws in above test case). Their order might be incorrect as well, but this should be another fixers concern I guess.

It's no problem to fetch all other annotations and simply append them to a list of `$orderedAnnotations`:

```php
Array
(
    [0] => ' * @param int   $a Multiline param *                 description'
    [1] =>  '* @param mixed $b'
    [2] =>  '* @param mixed $superflous'
    [3] =>  '* @return array'
    [4] =>  '* @throws Exception'
)
```
So as long as only annotations are used within the docblock, this approach would work. However, where I currently struggle is the question, how to handle other text comments in the doc block (`Method description \n over multiple lines`). I came up with the following to find plain text lines, but this will also fail with multiline parameter descriptions (because only the first line contains a tag):

```php
private function getUsefulDocBlockLines(DocBlock $doc)
{
    $usefulLines = [];

    foreach ($doc->getLines() as $line) {
        if ($line->containsUsefulContent() && !$line->containsATag()) {
            $usefulLines[] = $line->getContent();
        };
    }

    return $usefulLines;
}
```
You could of course parse the docblock line by line and check every annotation for their beginning and end to identify all "other" lines. But I'm not sure if this really is the easiest way to do this.

Does someone have a better idea?

grubersjoe:
I've added tests for all above cases. I had to adapt my previous approach of sorting the params to support the first category (other annotations mixed in between):

Before, the parameter lines (and only those) were overwritten with the parameters in order. Therefore annotations in between would simply stay, where they had been. Now, all annotations between the first and last param annotation line are sorted. Consequently, everything in between is listed in previous order *immediately* after the param block.

The corresponding test case:

```php
$expected = <<<'EOT'
<?php
/**
 * [c1] Method description
 * [c2] over multiple lines
 *
 * @see Baz
 *
 * @param int   $a Long param
 *                 description
 * @param mixed $b
 * @param mixed $superflous1 With text
 * @param int $superflous2
 * @return array Long return
 *               description
 * @throws Exception
 * @throws FooException
 */
function foo($a, $b) {}
EOT;
        $input = <<<'EOT'
<?php
/**
 * [c1] Method description
 * [c2] over multiple lines
 *
 * @see Baz
 *
 * @param mixed $b
 * @param mixed $superflous1 With text
 * @return array Long return
 *               description
 * @param int $superflous2
 * @throws Exception
 * @param int   $a Long param
 *                 description
 * @throws FooException
 */
function foo($a, $b) {}
EOT;
```

ntzm:
I am happy with the current behaviour of this fixer, but I think some integration tests are needed to make sure it doesn't conflict with other fixers that add, remove or otherwise modify `@param` tags.

luckyraul:
@grubersjoe @ntzm What should we do to merge this PR?

grubersjoe:
Hey. Sorry for the super long delay. I'm going to rebase this onto the current master and add the desired integration tests. I think afterwards this might be mergable again...

SpacePossum:
thanks @grubersjoe : D

grubersjoe:
So, sorry for the delay - again. Real life came in between :see_no_evil:. I've rebased my branch and added this fixer to the integration test in `tests/Fixtures/Integration/misc/phpdocs.test`.

Please let me know if more integration tests are needed or something else needs to be handled.

SpacePossum:
Thanks for the updates :)

Currently the CI has some troubles with the tooling itself and not some much with the code.

I tested you new rule on the symfony master code base and it made one change (pretty clean code base it seems ;) )
However I'm not sure this is a good change:

```diff
 interface CacheInterface
 {
     /**
-     * @param callable(CacheItem):mixed $callback Should return the computed value for the given key/item
      * @param float|null                $beta     A float that controls the likeliness of triggering early expiration.
      *                                            0 disables it, INF forces immediate expiration.
      *                                            The default (or providing null) is implementation dependent but should
      *                                            typically be 1.0, which should provide optimal stampede protection.
+     * @param callable(CacheItem):mixed $callback Should return the computed value for the given key/item
      *
      * @return mixed The value corresponding to the provided key
      */
      public function get(string $key, callable $callback, float $beta = null);
```

grubersjoe:
I've tried it myself with Symfony's master branch and I also got one change only, but it's correct... However, I enabled this fixer only. What ruleset did you use?

```
php php-cs-fixer fix ~/symfony/src --verbose --rules=phpdoc_param_order
```

```diff
 /**
  * Test the returned data when AccessDecisionManager is a TraceableAccessDecisionManager.
  *
  * @param string $strategy             strategy returned by the AccessDecisionManager
- * @param array  $voters               voters returned by AccessDecisionManager
  * @param array  $decisionLog          log of the votes and final decisions from AccessDecisionManager
+ * @param array  $voters               voters returned by AccessDecisionManager
  * @param array  $expectedVoterClasses expected voter classes returned by the collector
  * @param array  $expectedDecisionLog  expected decision log returned by the collector
  *
  * @dataProvider providerCollectDecisionLog
  */
  public function testCollectDecisionLog(string $strategy, array $decisionLog, array $voters, array $expectedVoterClasses, array $expectedDecisionLog): void
```

SpacePossum:
```
$ php php-cs-fixer fix /home/possum/work/symfony/src --rules=phpdoc_param_order --using-cache=no -vvv 
Loaded config default.

1) /home/possum/work/symfony/src/Symfony/Component/Cache/CacheInterface.php (phpdoc_param_order)
```

on:
>commit 9ae92fc98394551b6cdb78aec7d3407633ea04ef (HEAD -> feature-phpdoc-param-order, grubersjoe/feature-phpdoc-param-order)

and symfony:

```
$ git log
commit 5c338cc7ab1bbb614811f7beb560547f5de56723 (HEAD -> master, origin/master)
Merge: c997026c91 413af69e1d
Author: Nicolas Grekas <nicolas.grekas@gmail.com>
Date:   Mon Jun 11 16:42:15 2018 +0200
```

kubawerlos (approved):

Wirone:
Thank you very much @grubersjoe 🍻

grubersjoe:
Oh wow. I didn't think this would end up being merged after all this time but thank you for picking it up 😄👍.

Wirone:
I try to clean the backlog and ensure that contributors' work is not wasted, I believe people can take advantage of many things that are waiting here 🙂. 60 left 😅.

PR #3911, opened 2018-07-13
PhpdocVarWithoutNameFixer - fix for properties only

### LINKED ISSUE #3112: Change rules of PhpdocVarWithoutNameFixer
`phpdoc_var_without_name` fixer has some internal rules to decide whether the `$varname` part of a `@var` PHPDoc tag should be removed:
- ignore DocComments when they are on a single line, e.g. `/** @var Foo $foo */`;
- ignore `@var` tags that are not the first tag in the DocComment.

IMO these rules are wrong, I would replace them with a new rule: the fixer should simply ignore DocComments that don't apply to a class property.

dmvdbrugge:
This fixes #3112

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
@@ -23,6 +23,7 @@ use PhpCsFixer\Tokenizer\Tokens;
 
 /**
  * @author Graham Campbell <graham@alt-three.com>
+ * @author Dave van der Brugge <dmvdbrugge@gmail.com>
  */
 final class PhpdocVarWithoutNameFixer extends AbstractFixer
 {
@@ -32,7 +33,7 @@ final class PhpdocVarWithoutNameFixer extends AbstractFixer
     public function getDefinition()
     {
         return new FixerDefinition(
-            '`@var` and `@type` annotations should not contain the variable name.',
+            '`@var` and `@type` annotations of classy properties should not contain the name.',
             [new CodeSample('<?php
 final class Foo
 {
@@ -66,7 +67,7 @@ final class Foo
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isTokenKindFound(T_DOC_COMMENT);
+        return $tokens->isTokenKindFound(T_DOC_COMMENT) && $tokens->isAnyTokenKindsFound(Token::getClassyTokenKinds());
     }
 
     /**
@@ -79,21 +80,32 @@ final class Foo
                 continue;
             }
 
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
+
+            if (null === $nextIndex) {
+                continue;
+            }
+
+            // For people writing static public $foo instead of public static $foo
+            if ($tokens[$nextIndex]->isGivenKind(T_STATIC)) {
+                $nextIndex = $tokens->getNextMeaningfulToken($nextIndex);
+            }
+
+            // We want only doc blocks that are for properties and thus have specified access modifiers next
+            if (!$tokens[$nextIndex]->isGivenKind([T_PRIVATE, T_PROTECTED, T_PUBLIC, T_VAR])) {
+                continue;
+            }
+
             $doc = new DocBlock($token->getContent());
 
-            // don't process single line docblocks
-            if (1 === \count($doc->getLines())) {
-                continue;
+            $firstLevelLines = $this->getFirstLevelLines($doc);
+            $annotations = $doc->getAnnotationsOfType(['type', 'var']);
+
+            foreach ($annotations as $annotation) {
+                if (isset($firstLevelLines[$annotation->getStart()])) {
+                    $this->fixLine($firstLevelLines[$annotation->getStart()]);
+                }
             }
-
-            $annotations = $doc->getAnnotationsOfType(['param', 'return', 'type', 'var']);
-
-            // only process docblocks where the first meaningful annotation is @type or @var
-            if (!isset($annotations[0]) || !\in_array($annotations[0]->getTag()->getName(), ['type', 'var'], true)) {
-                continue;
-            }
-
-            $this->fixLine($doc->getLine($annotations[0]->getStart()));
 
             $tokens[$index] = new Token([T_DOC_COMMENT, $doc->getContent()]);
         }
@@ -108,5 +120,32 @@ final class Foo
         if (isset($matches[0][0])) {
             $line->setContent(str_replace($matches[0][0], '', $content));
         }
+    }
+
+    /**
+     * @return Line[]
+     */
+    private function getFirstLevelLines(DocBlock $docBlock)
+    {
+        $nested = 0;
+        $lines = $docBlock->getLines();
+
+        foreach ($lines as $index => $line) {
+            $content = $line->getContent();
+
+            if (Preg::match('/\s*\*\s*}$/', $content)) {
+                --$nested;
+            }
+
+            if ($nested > 0) {
+                unset($lines[$index]);
+            }
+
+            if (Preg::match('/\s\{$/', $content)) {
+                ++$nested;
+            }
+        }
+
+        return $lines;
     }
 }
--- a/tests/Fixer/Phpdoc/PhpdocVarWithoutNameFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocVarWithoutNameFixerTest.php
@@ -56,114 +56,237 @@ final class PhpdocVarWithoutNameFixerTest extends AbstractFixerTestCase
             'testFixVar' => [
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var string Hello!
      */
-
+    public $foo;
+}
 EOF
                 ,
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var string $foo Hello!
      */
-
+    public $foo;
+}
 EOF
                 ,
             ],
             'testFixType' => [
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var int|null
      */
-
+    public $bar;
+}
 EOF
                 ,
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var int|null $bar
      */
-
+    public $bar;
+}
 EOF
                 ,
             ],
             'testDoNothing' => [
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var Foo\Bar This is a variable.
      */
-
+    public $bar;
+}
 EOF
-            ],
-            'testFixVarWithOtherAnnotation' => [
-                <<<'EOF'
-<?php
-    /**
-     * @var string Hello!
-     *
-     * @deprecated
-     */
-
-EOF
-                ,
-                <<<'EOF'
-<?php
-    /**
-     * @var string $foo Hello!
-     *
-     * @deprecated
-     */
-
-EOF
-                ,
             ],
             'testFixVarWithNestedKeys' => [
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var array {
      *     @var bool   $required Whether this element is required
      *     @var string $label    The display name for this element
      * }
      */
-
+     public $options;
+}
 EOF
                 ,
                 <<<'EOF'
 <?php
+
+class Foo
+{
     /**
      * @var array $options {
      *     @var bool   $required Whether this element is required
      *     @var string $label    The display name for this element
      * }
      */
-
+     public $options;
+}
 EOF
             ],
             'testSingleLine' => [
                 <<<'EOF'
-                <?php
+<?php
+
+class Foo
+{
+    /** @var Foo\Bar */
+    public $bar;
+}
+EOF
+                ,
+                <<<'EOF'
+<?php
+
+class Foo
+{
     /** @var Foo\Bar $bar */
-    $bar;
+    public $bar;
+}
 EOF
                 ,
             ],
-            'testEmpty' => [
+            'testSingleLineProtected' => [
                 <<<'EOF'
-                <?php
-    /**
-     *
-     */
+<?php
 
+class Foo
+{
+    /** @var Foo\Bar */
+    protected $bar;
+}
+EOF
+                ,
+                <<<'EOF'
+<?php
+
+class Foo
+{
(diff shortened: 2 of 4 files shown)


src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
```diff
@@ -68,21 +68,30 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
 
-            $doc = new DocBlock($token->getContent());
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
+            $nextToken = isset($tokens[$nextIndex]) ? $tokens[$nextIndex] : null;
 
-            // don't process single line docblocks
-            if (1 === count($doc->getLines())) {
-                continue;
+            // For people writing static public $foo instead of public static $foo
+            if ($nextToken && $nextToken->isGivenKind([T_STATIC])) {
```
ntzm: Better to be strict and check `$nextToken !== null` or `$nextToken instanceof Token`

src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
```diff
@@ -68,21 +68,30 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                $nextIndex = $tokens->getNextMeaningfulToken($nextIndex);
+                $nextToken = isset($tokens[$nextIndex]) ? $tokens[$nextIndex] : null;
             }
 
-            $annotations = $doc->getAnnotationsOfType(['param', 'return', 'type', 'var']);
-
-            // only process docblocks where the first meaningful annotation is @type or @var
-            if (!isset($annotations[0]) || !in_array($annotations[0]->getTag()->getName(), ['type', 'var'], true)) {
+            // We want only doc blocks that are for properties and thus have specified access modifiers next
+            if (!$nextToken || !$nextToken->isGivenKind([T_PRIVATE, T_PROTECTED, T_PUBLIC])) {
```
ntzm: Strict check here too

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
```diff
@@ -1,26 +0,0 @@
---TEST--
-Integration of fixers: phpdoc_to_comment,phpdoc_var_without_name.
```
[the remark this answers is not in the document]
dmvdbrugge: It's a useless test that tests nothing, both fixers have no effect on the given code, as is already proven by their unit tests. There is no interaction going on between these fixers, so this test serves no use.

"Do not remove tests" sounds like a rule without reason. Can you tell me why this test should exist?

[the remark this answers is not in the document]
keradus: for dropping test you assume it's pointless
for keeping it, we ensure that when test was introduced to ensure something works, we don't break that by accident, because we happily drop testing that.

I prefer to have 10 tests that test nothing than drop one test that one only thought it doesn't test a thing.

Test here was added as proof that false-positive is not happening anymore, so prove that bug we had was fixed. That means originally, this code was wrongly changed.
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2191

src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
```diff
@@ -68,21 +68,30 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
             }
 
-            $doc = new DocBlock($token->getContent());
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
+            $nextToken = isset($tokens[$nextIndex]) ? $tokens[$nextIndex] : null;
 
-            // don't process single line docblocks
-            if (1 === count($doc->getLines())) {
-                continue;
+            // For people writing static public $foo instead of public static $foo
```
ntzm: Any tests to demonstrate this behaviour? Only seems to be tests for `public` properties atm, could do with ones for `static`, `private`, `protected` and `var`

src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
```diff
@@ -20,6 +20,7 @@
 use PhpCsFixer\Preg;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
+use const T_VAR;
```
kubawerlos: What is the reason for this import?

dmvdbrugge: PhpStorm settings I use in personal projects

`use` :allthethings:

Should've seen and removed that. Will do so now 😄

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
```diff
@@ -8,19 +8,35 @@ Integration of fixers: phpdoc_to_comment,phpdoc_var_without_name.
 --EXPECT--
 <?php
 
-$first = true;// needed because by default first docblock is never fixed.
+/**
```
kubawerlos: What is the point of this test? `phpdoc_to_comment` is not doing anything here (neither it was before changes).

Maybe... maybe we can have a docblock with a var name to remove, and `phpdoc_to_comment` would change it to comment before `phpdoc_var_without_name` would remove the name - but personally I'd remove this test.

kubawerlos:
@keradus @ntzm can you check it and update you review status?

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
dmvdbrugge: This is exactly why I initially deleted this test, however @keradus was totally against it, as can be seen [in the discussion above](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911#discussion_r202309642).

src/Fixer/Phpdoc/PhpdocVarWithoutNameFixer.php
```diff
@@ -68,21 +69,30 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                $nextIndex = $tokens->getNextMeaningfulToken($nextIndex);
+                $nextToken = isset($tokens[$nextIndex]) ? $tokens[$nextIndex] : null;
             }
 
-            $annotations = $doc->getAnnotationsOfType(['param', 'return', 'type', 'var']);
-
-            // only process docblocks where the first meaningful annotation is @type or @var
-            if (!isset($annotations[0]) || !\in_array($annotations[0]->getTag()->getName(), ['type', 'var'], true)) {
+            // We want only doc blocks that are for properties and thus have specified access modifiers next
+            if (null === $nextToken || !$nextToken->isGivenKind([T_PRIVATE, T_PROTECTED, T_PUBLIC, T_VAR])) {
```
SpacePossum: Maybe you can utilize `TokensAnalyzer::getClassyElements` to prevent have this kind of logic duplicated in the code base. Please add a unit test for nested anonymous classes as well.

dmvdbrugge: - [X] Nested anonymous classes testcase
- [ ] Try `TokensAnalyzer::getClassyElements`

dmvdbrugge: I've looked into `TokensAnalyzer::getClassyElements` however I don't see it becoming cleaner code:

- If working forward from the doc block (like now) we still need a way to match it to the classy element it belongs to.
- If working backwards from the classy elements, we still need to traverse back to find the actual matching doc block if any (and not any intermediary comment that happens to be like `/** this */`) which can be any number of tokens away from the index of the classy element.

enumag:
Can this be merged?

SpacePossum:
@kubawerlos or members/maintainer please review/approve as well

enumag:
Another month passed... Can someone please take a look? @keradus?

dmvdbrugge:
I rebased onto latest 2.12

Travis fails on something completely unrelated to this PR 😕

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
```diff
@@ -8,19 +8,35 @@ Integration of fixers: phpdoc_to_comment,phpdoc_var_without_name.
 --EXPECT--
```
keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911#discussion_r202366775
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911/files#r208303314

basically, this test got rewritten.

I just tested locally original test against code changes proposed by this PR - the test is failing. This means that the rule behind the fixer got modified, and test that has to be changed (regardless if it's utest or itest) proves that.
I find this to be a BC breaker - if a fixer apply single rule in multiple places (like here - for properties and for variables), and you want to adjust the scope and allow to fix only some of places - it deserves configurable option. If you want to fix only some of places without being able to configure it, it's a BC breaker for the rule.

Now, if I would merge it, nobody would be able to apply the rule for variables.

keradus: ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911#discussion_r202366775
ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3911/files#r208303314

basically, this test got rewritten.

I just tested locally original test against code changes proposed by this PR - the test is failing. This means that the rule behind the fixer got modified, and test that has to be changed (regardless if it's utest or itest) proves that.
I find this to be a BC breaker - if a fixer apply single rule in multiple places (like here - for properties and for variables), and you want to adjust the scope and allow to fix only some of places - it deserves configurable option. If you want to fix only some of places without being able to configure it, it's a BC breaker for the rule.

Now, if I would merge it, nobody would be able to apply the rule for variables.

dmvdbrugge: @julienfalque can you weigh in here? Wasn't this the entire point of your issue?

keradus: if someone wants to apply the rule only for properties, i'm fine with that.
but allowing to do so must not disallow to apply the rule for variables, as this is the behaviour we currently have - thus changing the rule in a way that variables will not be modified anymore is BC breaker.

I agree - i am not fully aware about the concerns or issues and i'm far from judging should we fix only properties or everyone - I'm only watching after keeping BC.

julienfalque: From [phpDocumentator doc](https://docs.phpdoc.org/references/phpdoc/tags/var.html) and [PSR-19 PHPDoc tags](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#description-16):
> The `@var` tag MUST contain the name of the element it documents. An exception to this is when property declarations only refer to a single property. In this case the name of the property MAY be omitted.

IMO the current behavior of the fixer is wrong, I consider changing it a bugfix and I'm fine with breaking BC here.

tests/Fixer/Phpdoc/PhpdocVarWithoutNameFixerTest.php
```diff
@@ -172,21 +295,156 @@ public function provideFixVarCases()
...
+                {
+                    /**
+                     * @var $stringVar string
+                     */
+                    public $stringVar;
+                };
+            }
+        };
+    }
+}
```
julienfalque: @dmvdbrugge Could you add a test case with multiple properties such as e.g.:
```php
class Foo {
    /**
     * @var int $foo
     */
    private $foo, $bar;
}
```

In such case, the variable name should not be removed from the `@var` tag as it would become ambiguous.

Also [PSR-19](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc-tags.md#examples-15) shows a sample with variation of this scenario but I'm not sure we should explicitly support it for now as PSR-19 is still a draft.

```php
class Foo {
    private
        /**
         * @var int
         */
        $foo,
        /**
         * @var int
         */
        $bar;
}
```

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
keradus: If the fixer would be:
> `@var` tag must follow 5.17 of PSR-19 proposal

I would agree.

Yet, it's different:
```
ker@dus:~/github/PHP-CS-Fixer λ php php-cs-fixer describe phpdoc_var_without_name
Description of phpdoc_var_without_name rule.
`@var` and `@type` annotations should not contain the variable name.
```

then, i would suggest to:
- make one fixer like `phpdoc_property_var_without_name` (this PR, just new fixer instead of modifying existing)
- deprecate existing fixer
- add fixer like `phpdoc_variable_var_with_name` (our current, existing fixer does the oposite) - of course, a separated rule

enumag:
ping @dmvdbrugge

SpacePossum:
Thank you @dmvdbrugge.

tests/Fixtures/Integration/misc/phpdoc_to_comment,phpdoc_var_without_name.test
SpacePossum: I considered the old working of the fixer a bug as well, together with other members and as such merged this.
I've updated the description to be more precise on the working of the fixer. If anyone needs the `phpdoc_variable_var_with_name` fixer than a PR is welcome.

tests/Fixer/Phpdoc/PhpdocVarWithoutNameFixerTest.php
SpacePossum: The scope of this issue was to only fix classy properties. I think that the request made here is out of scope and can be resolved with a new issue and PR

SpacePossum:
I've update the PR and added some tests and made other minor adjustments.

Thanks @dmvdbrugge for all the work, thanks to the reviewers as well!

dmvdbrugge:
Thank you @SpacePossum for the merge and additional work. I apologise for being a bit unresponsive on this 😅

PR #3913, opened 2018-07-13
TokensAnalyzer - fix isConstantInvocation

gharlan:
```php
echo "... $foo[bar] .."
```

`bar` is not a constant in this case.

### DIFF
--- a/src/Tokenizer/TokensAnalyzer.php
+++ b/src/Tokenizer/TokensAnalyzer.php
@@ -355,6 +355,15 @@ final class TokensAnalyzer
             }
         }
 
+        // check for array in double quoted string: `"..$foo[bar].."`
+        if ($this->tokens[$prevIndex]->equals('[') && $this->tokens[$nextIndex]->equals(']')) {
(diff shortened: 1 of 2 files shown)


Slamdunk:
I disagree: https://3v4l.org/PCZG4

Tricky and silly edge case, but still

gharlan:
@Slamdunk in line 3 `bar` is a constant. but because the constant does not exist, the string "bar" is used (warning).

In line 5 `bar` is not a constant (no warning).
The fix here is only about line 5.

gharlan:
https://3v4l.org/2AroY

SpacePossum:
we don't have a similar issue with `echo "... $foo{bar} .."`?

gharlan:
@SpacePossum it seems not to be valid syntax: https://3v4l.org/RNv7L

dmvdbrugge:
What about
```php
echo "{$foo[BAR]}";
```
Is that already covered? I don't see that case tested. Does it make sense to add?

gharlan:
@dmvdbrugge added your test case.

keradus:
Thank you @gharlan.

Issue #3883, opened 2018-07-06
Add a rule to force-add or force-remove `?` on nullable types


nicolas-grekas:
Sorry if this already exists, I didn't find it with a quick lookup,

Since PHP 7.1, nullable arguments can be written with or without the `?` prefix:
Form 1: `function (string $s = null) {}`
vs
Form 2: `function (?string $s = null) {}`

It would be great to have a rule to normalize to either form 1 or 2.
(On Symfony, we would stick to Form 1 for consistency with branches that still support PHP 5.5)

SpacePossum:
Hi,
Sounds like a good idea, we currently don't have a rule for this nor an open request for it, so all good :+1:

gharlan:
@keradus I think for this case the NoUnreachableDefaultArgumentValueFixer should be updated.

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/src/Fixer/FunctionNotation/NoUnreachableDefaultArgumentValueFixer.php#L102

`isTypehintedNullableVariable` should be replaced by
`isNonNullableTypehintedNullableVariable`

keradus:
Closing due to merge of #4401

PR #3914, opened 2018-07-13
NoUnreachableDefaultArgumentValueFixer - remove `null` for nullable typehints

gharlan:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3883#issuecomment-402994377


keradus:
I like the idea :+1:

Travis needs to be fixed (provided case is running on too low PHP engine, it has to run on high enough)

keradus:
Thank you @gharlan.

Issue #3697, opened 2018-04-14
[Meta] PHP 7.3 support


julienfalque:
This issue lists PHP 7.3 changes that are relevant to this project and tracks related issues and PR.

#### Syntax changes that require new tests to make sure the tool is compatible with PHP 7.3:

- [x] [Flexible Heredoc and Nowdoc Syntaxes](https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes) (#3915)
- [x] [list() Reference Assignment](https://wiki.php.net/rfc/list_reference_assignment)
- [x] `instanceof` now allows literals as the first operand, in which case the result is always `false`.
- [x] [Allow a trailing comma in function calls](https://wiki.php.net/rfc/trailing-comma-function-calls)

Progress tracker: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4184

#### Relevant changes that could lead to new features (not required to make the tool compatible with PHP 7.3):

- [ ] `hrtime` function has been added
- [ ] `is_countable` function has been added => #4236
- [ ] add trailing comma to multiline function calls => #4237

#### Sources:
- https://github.com/php/php-src/blob/PHP-7.3/NEWS
- https://github.com/php/php-src/blob/PHP-7.3/UPGRADING

ntzm:
Would be nice to get a fixer to pass the new `JSON_THROW_ON_ERROR` flag to `json_` methods when 7.3 is released

julienfalque:
Could be nice indeed, but that should be a separate feature request. This issue mostly tracks changes making _existing_ code compatible with PHP 7.3 :)

keradus:
and then, clean up manual handling of error, huh? gonna be tricky ;(

SpacePossum:
- `is_countable` function
- [Allow a trailing comma in function calls](https://wiki.php.net/rfc/trailing-comma-function-calls)

The last one will be a bit of work here I think.
It also seems to be missing from the releases note, maybe that gets corrected before the release though.

julienfalque:
Thanks @SpacePossum, I edited the list. Note that I think we don't need to care about `is_countable()` to be compatible with PHP 7.3. Having rules for it would be a separate feature request, and one was just submitted by @ntzm: #3711 ;)

keradus:
whenever we care about native functions, we need to take care of `is_countable` now as well.

SpacePossum:
[...]

PR #3915, opened 2018-07-14
Add HeredocIndentationFixer

gharlan:
#3697 / https://wiki.php.net/rfc/flexible_heredoc_nowdoc_syntaxes

Requires PHP 7.3

Input:

```php
<?php
    $a = [
        <<<'EOD'
abcdef
    ghi
jkl
EOD
    ];
```

is fixed to:

```php
<?php
    $a = [
        <<<'EOD'
            abcdef
                ghi
            jkl
            EOD
    ];
```

### DIFF
--- a/src/Fixer/Whitespace/HeredocIndentationFixer.php
+++ b/src/Fixer/Whitespace/HeredocIndentationFixer.php
@@ -1,0 +1,150 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Whitespace;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Gregor Harlan
+ */
+final class HeredocIndentationFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Heredoc/nowdoc content must be properly indented. Requires PHP >= 7.3.',
+            [
+                new VersionSpecificCodeSample(
+                    <<<'SAMPLE'
+<?php
+    $a = <<<EOD
+abc
+    def
+EOD;
+
+SAMPLE
+                    ,
+                    new VersionSpecification(70300)
+                ),
+                new VersionSpecificCodeSample(
+                    <<<'SAMPLE'
+<?php
+    $a = <<<'EOD'
+abc
+    def
+EOD;
+
+SAMPLE
+                    ,
+                    new VersionSpecification(70300)
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return \PHP_VERSION_ID >= 70300 && $tokens->isTokenKindFound(T_START_HEREDOC);
+    }
+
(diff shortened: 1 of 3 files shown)


dmvdbrugge:
Is it strange that I'd prefer it to be like this? It would be the same as other opening/closing pairs.
```
<?php
    $a = [
        <<<'EOD'
            abcdef
                ghi
            jkl
        EOD
    ];
```

gharlan:
It would change the string content, so it would be risky.

This would be non-risky:

```php
<?php
    $a = [
        <<<'EOD'
        abcdef
            ghi
        jkl
        EOD
    ];
```

If it is desired, there could be a configuration option for choosing this style, or the style from my first comment.

tests/Fixer/Whitespace/HeredocIndentationFixerTest.php
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * @dataProvider provideFixCases
+     * @requires PHP 7.3
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
```
dmvdbrugge: I see no test with nested here/now doc, might be nice to add.

gharlan:
But I think the style from my first comment is better, especially in other cases, like function calls:

Input:

```php
<?php
    foo(<<<'EOD'
abcdef
    ghi
jkl
EOD
    );
```

Style 1 (current indentation + 1):

```php
<?php
    foo(<<<'EOD'
        abcdef
            ghi
        jkl
        EOD
    );
```

Style 2 (current indentation + 0):

```php
<?php
    foo(<<<'EOD'
    abcdef
        ghi
    jkl
    EOD
    );
```

dmvdbrugge:
Oh yeah I agree, definitely current +1.

What I commented goes against the standard unfortunately (but understandably too, how else would they determine what whitespace to strip).

tests/Fixer/Whitespace/HeredocIndentationFixerTest.php
```diff
@@ -0,0 +1,158 @@
...
+<?php
+    foo(<<<EOD
+$abc
+    $def
+{$ghi}
+EOD
+    );
+INPUT
+                ,
+            ],
```
ntzm: Can we add some tests for tab indent too?

src/Fixer/Whitespace/HeredocIndentationFixer.php
```diff
@@ -0,0 +1,152 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return \PHP_VERSION_ID >= 70300 && $tokens->isAnyTokenKindsFound([T_START_HEREDOC]);
```
ntzm: can be `$tokens->isTokenKindFound(T_START_HEREDOC)` instead

src/Fixer/Whitespace/HeredocIndentationFixer.php
```diff
@@ -0,0 +1,152 @@
...
+            $tokens[$index] = new Token([T_ENCAPSED_AND_WHITESPACE, $content]);
+        } else {
+            $tokens->insertAt($index, new Token([T_ENCAPSED_AND_WHITESPACE, $indent]));
+        }
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return null|string
```
SpacePossum: is `null` a real possible value here? the calling code seems to assume a string is always returned (it concats on the returned value)

gharlan: fixed

gharlan:
@SpacePossum it is fixed now.
(the errors seem not to be not related to this pr)

src/Fixer/Whitespace/HeredocIndentationFixer.php
```diff
@@ -0,0 +1,150 @@
...
+    {
+        $indent = $this->getIndentAt($tokens, $start).$this->whitespacesConfig->getIndent();
+
+        Preg::match('/^[ \t]*/', $tokens[$end]->getContent(), $matches);
+        $currentIndent = $matches[0];
+
+        $content = $indent.substr($tokens[$end]->getContent(), \strlen($currentIndent));
+        $tokens[$end] = new Token([T_END_HEREDOC, $content]);
+
+        if ($end === $start + 1) {
```
SpacePossum: I think this early return can be done before the indent checks (i.e. first thing in this method)

gharlan: In the line before the indentation of `T_END_HEREDOC` is fixed. So this `return` can not be moved..

SpacePossum: i see, sorry for the noise :)

keradus:
Thank you @gharlan.

SpacePossum:
thanks @gharlan :tada:  :)

PR #3916, opened 2018-07-14
New ruleset "@PhpCsFixer"

gharlan:
I suggest to add new rulesets "@PhpCsFixer" and "@PhpCsFixer:risky".

Reason: I would like to use the standards recommended by this org.

I would only add fixers to these rulesets, that are really recommended by this org, not fixers you are using because of your min php version (like list_syntax=long).


localheinz:
@gharlan 

I think it would be great if rulesets could be concrete classes, perhaps for 3.0?

keradus:
sounds interesting and I see ppl like it.
(i didn't added it myself for a reason of what @localheinz is asking for)

yet, we have already some conflicts here,
@gharlan , can you target 2.12 branch, please ?
it would easier the maintainance

gharlan:
> @gharlan , can you target 2.12 branch, please ?

done.

keradus:
Thank you @gharlan.

keradus:
Thank you @gharlan.

keradus:
I propagated the idea up to 3.0, considering changes we done in config file on those branches

PR #3918, opened 2018-07-14
[WIP] Format code in risky rule reasons

ntzm:
In RST, you can't have code in italics blocks, so we have a lot of `` in the readme. I propose we only italic the `Risky rule:` part of it, so the rest can be formatted correctly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -394,7 +394,7 @@ Choose from the list of available rules:
   Comments with annotation should be docblock when used on structural
   elements.
 
-  *Risky rule: risky as new docblocks might mean more, e.g. a Doctrine entity might have a new column in database.*
+  *Risky rule:* risky as new docblocks might mean more, e.g. a Doctrine entity might have a new column in database.
 
 * **compact_nullable_typehint**
 
@@ -413,7 +413,7 @@ Choose from the list of available rules:
 
   Class ``DateTimeImmutable`` should be used instead of ``DateTime``.
 
-  *Risky rule: risky when the code relies on modifying ``DateTime`` objects or if any of the ``date_create*`` functions are overridden.*
+  *Risky rule:* risky when the code relies on modifying ``DateTime`` objects or if any of the ``date_create*`` functions are overridden.
 
 * **declare_equal_normalize** [@Symfony]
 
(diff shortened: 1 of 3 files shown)


ntzm:
Alternatively, we could replace 
\*Risky rule: foo \`\`bar\`\` baz\*
with
\*Risky rule: foo\* \`\`bar\`\` \*baz\*
which would keep the italics throughout, just terminating before each code block and starting after

ntzm:
My regex skills are failing me 😢 I've got `$help = Preg::replace('#(\`\`[^`]+\`\`) #m', '* $1 *', $help);` to terminate the `*` before and start it up again, but I need to use a lookbehind to assert the line starts with a `*`. Any help?

README.rst
```diff
@@ -1643,25 +1643,25 @@ Choose from the list of available rules:
 
   Lambdas not (indirect) referencing ``$this`` must be declared ``static``.
 
-  *Risky rule: risky when using "->bindTo" on lambdas without referencing to ``$this``.*
+  *Risky rule:* risky when using "->bindTo" on lambdas without referencing to ``$this``.
```
dmvdbrugge: Unrelated, but shouldn't this be ` ``->bindTo`` `?

ntzm: Yes it should 😁

dmvdbrugge:
Lookbehind is relative to the thing you're trying to match, which would make looking back to the start of the line "non-fixed width", which is an error, according to https://regex101.com/r/Ji8pRi/1/

Also that space in your regex is wrong, as can be seen in the README.rst part commented on above, it would fail the ` ``this``.`

I don't know if this can be solved with regexes, but then again I'm not that good with 'em (while this codebase is full of them 😱).

SpacePossum:
a more stable solution could be to create a new style tag `<risky>`,
add the style to the `HelpCommand` ; `->setStyle('risky', new OutputFormatterStyle('white', 'red'));`
update `"   | *Risky rule: %s.*\n",` to `"   | <risky>Risky rule</risky>: %s.\n",`
now in the `ReadmeCommand` you can "just" match on the tags, making your life more easy ;)

kubawerlos:
Does this PR still make any sense since we have https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3475? I think we can close it.

SpacePossum:
@ntzm closing for now as there hasn't been worked on this for a bit and we have another PR open for changes on this topic, let me know if you plan on working it though than we can reopen the PR, thanks for all the work!

PR #3924, opened 2018-07-16
FullyQualifiedStrictTypesFixer - space bug

### LINKED ISSUE #3861: fully_qualified_strict_types should not remove the space between type and param
Which it does in case of `&` being present. If that's a case you don't want to account for (because objects are already passed by ref anyway, I don't know who wrote that code either), at least add a priority case, that `fully_qualified_strict_types` should run *before* `function_typehint_space` (which fixes it).

```
```

```php
function withReference(\Exception &$e)
{
}
```

* after first run:
```php
function withReference(Exception&$e)
{
}
```

* after second run:
```php
function withReference(Exception &$e)
{
}
```

* expected:

Meta: this is case 35 of #3844

dmvdbrugge:
This fixes #3861


keradus:
Thank you @dmvdbrugge.

Issue #3902, opened 2018-07-11
FR: Add 'final' keyword to classes


JarJak:
Feature Request

According to https://ocramius.github.io/blog/when-to-declare-classes-final/

Make classes final, if they are not extended somewhere, they implement an interface, and no other public methods are defined.

SpacePossum:
Hi and thanks for the suggestion,

Is this something you feel not covered by the rule `final_internal_class` ?

JarJak:
I mean something like https://github.com/Symplify/CodingStandard/blob/master/src/Fixer/Solid/FinalInterfaceFixer.php

SpacePossum:
I think this request is similar to https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3923 and the related PR of it, therefore I'm closing this one in favor of discussing it in one place.

There is a lot of work and talk about this issue so please @JarJak hop over to the other issue and PR to give your input if you can :)

PR #3927, opened 2018-07-17
Add FinalClassFixer

### LINKED ISSUE #3923: New fixer: Make classes final
Apart from:

- Doctrine entities
- Abstract classes

Any others?

Slamdunk:
Closes #3923 #3902 

Two notes:

#### 1. No exception and no configuration are intentional

Beside [Doctrine entities](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/architecture.html#entities) (and there are rumors that Doctrine 3 will allow final entities), and of course abstract classes, there is no single reason not to declare all classes final.

If you want to subclass a class, mark the parent class as abstract and create two child classes, one empty if necessary: you'll gain much more fine grained type-hinting.

If you need to mock a standalone class, create an interface, or maybe it's a value-object that shouldn't be mocked at all.

If you need to extend a standalone class, create an interface and use the composite pattern.

If you aren't ready yet for serious OOP, go with [FinalInternalClassFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3183), it's fine.

Further reading: [When to declare classes final](https://ocramius.github.io/blog/when-to-declare-classes-final/)

#### 2. Doctrine entities detection is brittle as of yet

- [x] Please point me out how to improve import and use detection for Doctrine Entities: https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/reference/basic-mapping.html
I don't have the full picture on how they can be marked as such, and how this projects handles import detections

### DIFF
--- a/src/Fixer/ClassNotation/FinalClassFixer.php
+++ b/src/Fixer/ClassNotation/FinalClassFixer.php
@@ -1,0 +1,60 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractProxyFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class FinalClassFixer extends AbstractProxyFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All classes must be final, except abstract ones and Doctrine entities.',
+            [
+                new CodeSample(
+                    '<?php
+class MyApp {}
+'
+                ),
+            ],
+            'No exception and no configuration are intentional. Beside Doctrine entities and of course abstract classes, there is no single reason not to declare all classes final. '
+            .'If you want to subclass a class, mark the parent class as abstract and create two child classes, one empty if necessary: you\'ll gain much more fine grained type-hinting. '
+            .'If you need to mock a standalone class, create an interface, or maybe it\'s a value-object that shouldn\'t be mocked at all. '
+            .'If you need to extend a standalone class, create an interface and use the Composite pattern. '
+            .'If you aren\'t ready yet for serious OOP, go with FinalInternalClassFixer, it\'s fine.',
+            'Risky when subclassing non-abstract classes.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createProxyFixers()
+    {
+        $fixer = new FinalInternalClassFixer();
+        $fixer->configure([
+            'annotation-white-list' => [],
+            'consider-absent-docblock-as-internal-class' => true,
+        ]);
+
+        return [$fixer];
+    }
+}
--- a/src/Fixer/ClassNotation/FinalInternalClassFixer.php
+++ b/src/Fixer/ClassNotation/FinalInternalClassFixer.php
@@ -20,6 +20,7 @@ use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
 use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Preg;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 use Symfony\Component\OptionsResolver\Options;
@@ -73,7 +74,7 @@ final class FinalInternalClassFixer extends AbstractFixer implements Configurati
      */
     public function isCandidate(Tokens $tokens)
     {
-        return $tokens->isAllTokenKindsFound([T_CLASS, T_DOC_COMMENT]);
+        return $tokens->isTokenKindFound(T_CLASS);
     }
 
     /**
@@ -143,8 +144,12 @@ final class FinalInternalClassFixer extends AbstractFixer implements Configurati
             (new FixerOptionBuilder('annotation-black-list', 'Class level annotations tags that must be omitted to fix the class, even if all of the white list ones are used as well. (case insensitive)'))
                 ->setAllowedTypes(['array'])
                 ->setAllowedValues($annotationsAsserts)
-                ->setDefault(['@final', '@Entity', '@ORM'])
+                ->setDefault(['@final', '@Entity', '@ORM\Entity'])
                 ->setNormalizer($annotationsNormalizer)
+                ->getOption(),
+            (new FixerOptionBuilder('consider-absent-docblock-as-internal-class', 'Should classes without any DocBlock be fixed to final?'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
                 ->getOption(),
         ]);
     }
@@ -157,23 +162,26 @@ final class FinalInternalClassFixer extends AbstractFixer implements Configurati
      */
     private function isClassCandidate(Tokens $tokens, $index)
     {
-        if ($tokens[$tokens->getPrevMeaningfulToken($index)]->isGivenKind([T_ABSTRACT, T_FINAL])) {
+        if ($tokens[$tokens->getPrevMeaningfulToken($index)]->isGivenKind([T_ABSTRACT, T_FINAL, T_NEW])) {
             return false; // ignore class; it is abstract or already final
         }
 
         $docToken = $tokens[$tokens->getPrevNonWhitespace($index)];
 
         if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
-            return false; // ignore class; it has no class-level PHPDoc
+            return $this->configuration['consider-absent-docblock-as-internal-class'];
         }
 
         $doc = new DocBlock($docToken->getContent());
         $tags = [];
 
         foreach ($doc->getAnnotations() as $annotation) {
-            $tag = strtolower($annotation->getTag()->getName());
-            if (isset($this->configuration['annotation-black-list'][$tag])) {
-                return false; // ignore class: class-level PHPDoc contains tag that has been black listed through configuration
+            Preg::match('/@\S+(?=\s|$)/', $annotation->getContent(), $matches);
+            $tag = strtolower(substr(array_shift($matches), 1));
+            foreach ($this->configuration['annotation-black-list'] as $tagStart => $true) {
+                if (0 === strpos($tag, $tagStart)) {
+                    return false; // ignore class: class-level PHPDoc contains tag that has been black listed through configuration
+                }
             }
 
             $tags[$tag] = true;
--- a/tests/Fixer/ClassNotation/FinalClassFixerTest.php
+++ b/tests/Fixer/ClassNotation/FinalClassFixerTest.php
@@ -1,0 +1,93 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\ClassNotation;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\ClassNotation\FinalClassFixer
+ */
+final class FinalClassFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixCases
+     */
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
+            ['<?php /** @Entity */ class MyEntity {}'],
+            ['<?php use Doctrine\ORM\Mapping as ORM; /** @ORM\Entity */ class MyEntity {}'],
+            ['<?php /** @entity */ class MyEntity {}'],
+            ['<?php use Doctrine\ORM\Mapping as ORM; /** @orm\entity */ class MyEntity {}'],
+            ['<?php abstract class MyAbstract {}'],
+            ['<?php trait MyTrait {}'],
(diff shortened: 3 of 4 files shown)


src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,110 @@
...
+    private function isDoctrineEntity(Token $docToken)
+    {
+        if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
+            return false;
+        }
+
+        $doc = new DocBlock($docToken->getContent());
+        foreach ($doc->getAnnotations() as $annotation) {
+            $tag = strtolower($annotation->getTag()->getName());
+            $content = strtolower($annotation->getContent());
```
ntzm: can remove this line and change the line below:

```diff
-if ('entity' === $tag || false !== strpos($content, '@orm\entity')) {
+if ('entity' === $tag || false !== stripos($content, '@orm\entity')) {
```

src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,110 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All classes must be final, except abstract ones and Doctrine entities.',
+            [
+                new CodeSample('<?php class MyApp {}'.PHP_EOL),
```
ntzm: This will produce different results on windows and unix machines, should be `"<?php class MyApp {}\n"`

src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,113 @@
...
+    private function isDoctrineEntity(Token $docToken)
+    {
+        if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
+            return false;
+        }
+
+        $doc = new DocBlock($docToken->getContent());
+        foreach ($doc->getAnnotations() as $annotation) {
+            $tag = strtolower($annotation->getTag()->getName());
+            if ('entity' === $tag || false !== stripos($annotation->getContent(), '@orm\entity')) {
```
ntzm: What about `@Entity`? I'm not overly familiar with Doctrine but when I used it for a project I was following [this](https://www.doctrine-project.org/projects/doctrine-orm/en/2.6/tutorials/getting-started.html) which uses `@Entity`

tests/Fixer/ClassNotation/FinalClassFixerTest.php
```diff
@@ -0,0 +1,81 @@
...
+                '<?php class MyClass extends MyAbstract {}',
+            ],
+            [
+                '<?php final class MyClass implements MyInterface {}',
+                '<?php class MyClass implements MyInterface {}',
+            ],
+            [
+                '<?php /** @codeCoverageIgnore */ final class MyEntity {}',
+                '<?php /** @codeCoverageIgnore */ class MyEntity {}',
+            ],
```
ntzm: Can we add some tests for:

- Multiple classes in one file
- Interfaces (don't get changed obviously :D)
- PHPDoc with mentions of "Entity" but not an actual annotation

ntzm: Thanks @Slamdunk, anyone know of a way of marking comments as resolved without deleting them? 😆

Slamdunk: Approve the PR?

Slamdunk: Nope, but I get your approval :stuck_out_tongue:

keradus: @ntzm - yes, with hand made comment like `changes applied` ;)

src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,113 @@
...
+ */
+final class FinalClassFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All classes must be final, except abstract ones and Doctrine entities.',
```
keradus: one more case before we will be able to merge it
classes marked with `@final` annotation shall not be marked with `final` keyword.

eg, if PHP CS Fixer v2 has some class that is not final, fixer would make it final, but that's BC breaker, so we would move `final` keyword into `@final` annotation and plan to make it real final in v3.

src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,113 @@
...
+     */
+    private function isDoctrineEntity(Token $docToken)
+    {
+        if (!$docToken->isGivenKind(T_DOC_COMMENT)) {
+            return false;
+        }
+
+        $doc = new DocBlock($docToken->getContent());
+        foreach ($doc->getAnnotations() as $annotation) {
+            $tag = strtolower($annotation->getTag()->getName());
```
keradus: question: it's really case insensitive ?

tests/Fixer/ClassNotation/FinalClassFixerTest.php
```diff
@@ -0,0 +1,90 @@
...
+    public function testFix($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
+            ['<?php /** @Entity */ class MyEntity {}'],
+            ['<?php use Doctrine\ORM\Mapping as ORM; /** @ORM\Entity */ class MyEntity {}'],
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3927/files#r203016406
if yes, please add next case with different casing as well ;)

src/Fixer/ClassNotation/FinalClassFixer.php
Slamdunk: This would be against the purpose of the fixer.

`@final` is not a standard annotation (AFAIK), and if you have a non trivial DocBlock for the class (more than 10 lines) it would be very easy to miss out the tag (in a contributor PR f.e.) and therefore the class could end up without the final php keyword.

Being a BC breaker is intentional. I want maintainers to expect that adopting this fixer automatically prevent each and every inheritance (future) break.

As a maintainer myself, I rely on (risky) fixers also to spot such dangerous cases.

tests/Fixer/ClassNotation/FinalClassFixerTest.php
```diff
@@ -0,0 +1,90 @@
...
+            ],
+            [
+                '<?php final class A {} abstract class B {} final class C {}',
+                '<?php class A {} abstract class B {} class C {}',
+            ],
+            [
+                '<?php /** @internal Map my app to an @Entity */ final class MyMapper {}',
+                '<?php /** @internal Map my app to an @Entity */ class MyMapper {}',
+            ],
+        ];
```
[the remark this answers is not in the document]
Slamdunk: I don't understand: you want to skip Exceptions, don't you?

I want to make Exceptions final as well instead.

src/Fixer/ClassNotation/FinalClassFixer.php
keradus: there is no docblock PSR approved you know? :)
yet, this annotation is highly used.

I totally agree of relying on rule as a watcher to not introduce new classes without `final`!
Yet, like at PHP CS Fixer, we already have the situation when we have non-final classes, and we can not mark them as final. So, we can either do nothing, which is bad, or mark them with `@final` and mark as real final in next MAJOR release. But doing the later will prevent us from using the rule.

src/Fixer/ClassNotation/FinalClassFixer.php
Slamdunk: I've tested it, case-insensitiveness is required

tests/Fixer/ClassNotation/FinalClassFixerTest.php
Slamdunk: I've tested it, case-insensitiveness is required, test cases added

tests/Fixer/ClassNotation/FinalClassFixerTest.php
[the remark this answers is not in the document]
keradus: I want to have case that will prove that
```php
<?php echo Exception::class;
```

will not get wrongly "fixed" into:
```php
<?php echo final Exception::class;
```


as even if it works now, in next year, when one will try to change sth, he can break it, if there would be no test for such case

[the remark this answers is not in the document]
Slamdunk: Done

src/Fixer/ClassNotation/FinalClassFixer.php
```diff
@@ -0,0 +1,117 @@
...
+                    '<?php
+class MyApp {}
+'
+                ),
+            ],
+            'No exception and no configuration are intentional. Beside Doctrine entities and of course abstract classes, there is no single reason not to declare all classes final. '
+            .'If you want to subclass a class, mark the parent class as abstract and create two child classes, one empty if necessary: you\'ll gain much more fine grained type-hinting. '
+            .'If you need to mock a standalone class, create an interface, or maybe it\'s a value-object that shouldn\'t be mocked at all. '
+            .'If you need to extend a standalone class, create an interface and use the Composite pattern. '
+            .'If you aren\'t ready yet for serious OOP, go with FinalInternalClassFixer, it\'s fine.',
```
Slamdunk: Extended description added

src/Fixer/ClassNotation/FinalClassFixer.php
keradus: On the other side, without it the rule will be used either in green field project or project which already follow the rule.
The rule won't help community to follow the rule if they are not following it yet; or they will start it and do the BC breaks

Slamdunk: The only two cases that come into my mind are:

1. A project didn't add any DocBlock to state (in the current version) that a class shouldn't be used in the userland, so adding `final` would be a BC break anyway
1. A project, like this, already had a dedicated DocBlock, so `FinalInternalClassFixer` would fit much better for the transition period

With the configuration you ask, it would be much easier to mark `annotation-white-list` option as optional in `FinalInternalClassFixer`, and drop this PR.

keradus: do you mean that this PR can be basically a proxy fixer to
`final_internal_class => ['annotation-black-list'=> ['@Entity', '@ORM'], 'annotation-white-list ' => []]`
?

Slamdunk: Yes

SpacePossum: maybe I'm slow on this one, but if we already have a fixer to this what is the added value of another?

Slamdunk: That there is no way to escape this fixer. Once you enable it, your library will never have inheritance BC breaks anymore, since subclassing would be totally forbidden for non abstract classes.

On the other hand, `final_internal_class` provide much weaker guarantees, because it relies on DocBlock to take decisions, and DocBlocks are much harder to track/review.

What we could do is proxy this fixer to `final_internal_class => ['annotation-black-list'=> ['@Entity', '@ORM'], 'annotation-white-list ' => []]` to reduce code duplication.

keradus: that can make sense.
ultimately, new code from this PR is nice and test cases brings a lot of value.

i would suggest to merge as much logic & tests as possible into `final_internal_class`, and then to the proxy-ing

then, I would actually rename `final_internal_class` into `final_class` and change default config on it on 3.0, as final by default is better, and exception of `@internal`/`@final` are ways of dealing with legacy code and looks more like a configuration per project from perspective.

cc @Slamdunk @SpacePossum

SpacePossum: I really like the suggestion of @keradus here, TBH I think it is a requirement for this PR. This is because without it this PR would bring a mixture of a proxy fixer with new logic which might cause confusion.
So lets aim for `final` as much as possible with the new fixer proposed here and allow for the detailed exceptions of the current fixer, win-win :)

src/Fixer/ClassNotation/FinalClassFixer.php
ntzm: Perhaps we should refer to it by its documented name, `final_internal_class`

src/Fixer/ClassNotation/FinalClassFixer.php
Slamdunk: Sorry for the late edit. I've proxified the code to `FinalInternalClassFixer` as agreed.

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -157,23 +162,26 @@ protected function createConfigurationDefinition()
...
         }
 
         $doc = new DocBlock($docToken->getContent());
         $tags = [];
 
         foreach ($doc->getAnnotations() as $annotation) {
-            $tag = strtolower($annotation->getTag()->getName());
-            if (isset($this->configuration['annotation-black-list'][$tag])) {
-                return false; // ignore class: class-level PHPDoc contains tag that has been black listed through configuration
+            Preg::match('/@\S+(?=\s|$)/', $annotation->getContent(), $matches);
```
Slamdunk: A lot of online doc about Doctrine refers to entities with the tag `@ORM\Entity`, which is useful also because you have to write only one `use` for everything Doctrine-related an entity needs.
But the backslash breaks what is considered a tag in this regex:

https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/110bc117adbbda23ac7770a670aa95947123476e/src/DocBlock/Tag.php#L70

Without this change `@ORM\Entity` can be handled properly: `other` would be too generic IMHO.

src/Fixer/ClassNotation/FinalInternalClassFixer.php
```diff
@@ -143,9 +144,13 @@ protected function createConfigurationDefinition()
             (new FixerOptionBuilder('annotation-black-list', 'Class level annotations tags that must be omitted to fix the class, even if all of the white list ones are used as well. (case insensitive)'))
                 ->setAllowedTypes(['array'])
                 ->setAllowedValues($annotationsAsserts)
-                ->setDefault(['@final', '@Entity', '@ORM'])
+                ->setDefault(['@final', '@Entity', '@ORM\Entity'])
```
Slamdunk: This may or may not be considered a BC break, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3927/files#r268091523
Note: it is not strictly necessary for this PR, I can also explicit `annotation-black-list` in `FinalClassFixer` with the annotations I need, and revert this line, but to me it's worth discussing it

keradus: tbh i would rather consider this as a bugfix (cc @julienfalque ), and each bugfix is kind-of bc break

keradus:
Thank you @Slamdunk.

rvanlaak:
@Slamdunk started experimenting with this rule. What about Doctrine repositories that are extending `EntityRepository`? 

Anotating as `@ORM\EntityRepository` to let the csfixer rule ignore that class does work, but as `EntityRepository()` does not exist as anotation the application thereafter will not run.

Any suggestions aside of anotating as `@final`?

Slamdunk:
I'm sorry as far as I remember entity repositories are not affected by the proxy issue needed for entities, why would you need them non-final?

PR #3928, opened 2018-07-17
Add FinalPublicMethodForAbstractClassFixer

Slamdunk:
```diff
 abstract class AbstractMachine
 {
-    public function start()
+    final public function start()
     {}
 }
```

#### No exception and no configuration are intentional

If you want to override a method, use the [Template method pattern](https://en.wikipedia.org/wiki/Template_method_pattern).

This fixer enforces API encapsulation in an inheritance environment.

I left out magic methods:

1. Because some of them are not API
1. And the rest have issues with SPL releated array classes etc

Further reading: [When to declare methods final](https://slamdunk.github.io/blog/when-to-declare-methods-final/)

### DIFF
--- a/src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
+++ b/src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
@@ -1,0 +1,168 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class FinalPublicMethodForAbstractClassFixer extends AbstractFixer
+{
+    /**
+     * @var array
+     */
+    private $magicMethods = [
+        '__construct' => true,
+        '__destruct' => true,
+        '__call' => true,
+        '__callstatic' => true,
+        '__get' => true,
+        '__set' => true,
+        '__isset' => true,
+        '__unset' => true,
+        '__sleep' => true,
+        '__wakeup' => true,
+        '__tostring' => true,
+        '__invoke' => true,
+        '__set_state' => true,
+        '__clone' => true,
+        '__debuginfo' => true,
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All public methods of abstract classes should be final.',
+            [
+                new CodeSample(
+                    '<?php
+
+abstract class AbstractMachine
+{
+    public function start()
+    {}
+}
+'
+                ),
+            ],
+            'Enforce API encapsulation in an inheritance architecture. '
+            .'If you want to override a method, use the Template method pattern.',
+            'Risky when overriding public methods of abstract classes'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_ABSTRACT, T_PUBLIC, T_FUNCTION]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $classes = array_keys($tokens->findGivenKind(T_CLASS));
+
+        while ($classIndex = array_pop($classes)) {
+            $prevToken = $tokens[$tokens->getPrevMeaningfulToken($classIndex)];
+            if (!$prevToken->isGivenKind([T_ABSTRACT])) {
+                continue;
+            }
+
+            $classOpen = $tokens->getNextTokenOfKind($classIndex, ['{']);
+            $classClose = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $classOpen);
+
+            $this->fixClass($tokens, $classOpen, $classClose);
+        }
+    }
+
+    /**
+     * @param int $classOpenIndex
+     * @param int $classCloseIndex
+     */
+    private function fixClass(Tokens $tokens, $classOpenIndex, $classCloseIndex)
+    {
+        for ($index = $classCloseIndex - 1; $index > $classOpenIndex; --$index) {
+            // skip method contents
+            if ($tokens[$index]->equals('}')) {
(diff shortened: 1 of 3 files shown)


src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
```diff
@@ -0,0 +1,166 @@
...
+
+abstract class AbstractMachine
+{
+    public function start()
+    {}
+}
+'
+                ),
+            ],
+            null,
```
keradus: actually, for this and previous PR of yours.
you gave some really nice reasoning in opening comments.

why not put them here, to extended description ?

tests/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixerTest.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    public function provideFixCases()
+    {
+        $original = $fixed = $this->getClassElementStubs();
+        $fixed = str_replace('public function f1', 'final public function f1', $fixed);
+        $fixed = str_replace('public static function f4', 'final public static function f4', $fixed);
+        $fixed = str_replace('static public function f7', 'final static public function f7', $fixed);
+
+        return [
```
keradus: I believe we don't have a case with non-final, non-abstract class

Slamdunk: I'm sorry I don't get what you are asking

keradus: we have test cases for:
`final class`, `trait`, `abstract class`, but I don't see one for regular `class`, without `final`/`abstract` modifiers

Slamdunk: You are right, added

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class FinalPublicMethodForAbstractClassFixer extends AbstractFixer
+{
+    /**
+     * @var array
+     */
+    private $magicMethods = [
```
SpacePossum: this is a near copy of the list here `https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.13.1/src/Fixer/Casing/MagicMethodCasingFixer.php#L26` , maybe we can move this list to the `PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer` ?

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
```diff
@@ -0,0 +1,167 @@
...
+            'Risky when overriding public methods of abstract classes'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_CLASS);
```
SpacePossum: `isAllTokenKindsFound([T_CLASS, T_ABSTRACT, T_PUBLIC, T_FUNCTION)`

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
```diff
@@ -0,0 +1,167 @@
...
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $classes = array_keys($tokens->findGivenKind(T_CLASS));
```
SpacePossum: `TokensAnalyzer::getClassyElements` might be helpful here, also for dealing with nested anonymous classes within the body of an abstract class

SpacePossum: Reducing code that does the same comes with the typical OOP/maintaining benefits. For example like for PHP7.3 finding out all the code that deals with functions (/method) calls would've been easier if it was all done through one analyzer. So who knows, maybe PHP will change something related to class elements in the future. Having all code that deals with that using the `TokensAnalyzer::getClassyElements` would make it easier to find all the use cases that need to checked/fixed :)

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
SpacePossum: I would be :+1: for that  , but lets hear from others as well

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
```diff
@@ -0,0 +1,167 @@
...
+            $prevIndex = $tokens->getPrevMeaningfulToken($index);
+            $prevToken = $tokens[$prevIndex];
+            if ($prevToken->isGivenKind(T_STATIC)) {
+                $index = $prevIndex;
+                $prevIndex = $tokens->getPrevMeaningfulToken($index);
+                $prevToken = $tokens[$prevIndex];
+            }
+            // skip already final methods
+            if ($prevToken->isGivenKind([T_FINAL])) {
+                continue;
```
SpacePossum: I think `$index = $prevIndex;` can be applied here as well?

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
SpacePossum: could you do this change @Slamdunk ? after that this PR seems RTM to me (I can do the move later if you can't, just let me know :) )

Slamdunk: For this PR `FunctionsAnalyzer` could be useful, but for [MagicMethodCasingFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.13.1/src/Fixer/Casing/MagicMethodCasingFixer.php#L26)? That fixer must have the full map anyway because it needs a casing map that for `FunctionsAnalyzer` would be useless

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
Slamdunk: Done, thank you

src/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixer.php
SpacePossum: I think both fixers can build the map they need based on a list of magic method names, so we only need to maintain that list at one place?

Slamdunk: @keradus is this discussion blocking the merge of the PR?

tests/Fixer/ClassNotation/FinalPublicMethodForAbstractClassFixerTest.php
```diff
@@ -0,0 +1,163 @@
...
+            'final-class' => ["<?php final class MyClass { ${original} }"],
+            'trait' => ["<?php trait MyClass { ${original} }"],
+            'interface' => [
+                '<?php interface MyClass {
+                    public function f1();
+                    public static function f4();
+                    static public function f7();
+                }',
+            ],
+            'magic-methods' => [
```
SpacePossum: please add a test with a magic method in different casing, like `__CLONE(){}`

Slamdunk: Good catch, done :+1:

keradus:
(rebased + fix small issues causing CI to fail)

keradus:
Thank you @Slamdunk.

PR #3930, opened 2018-07-18
Run NoSuperfluousPhpdocTagsFixer before PhpdocAddMissingParamAnnotationFixer

ossinkine:
PhpdocAddMissingParamAnnotationFixer adds missing parameter with `mixed` type but NoSuperfluousPhpdocTagsFixer considers it is superfluous and remove it.

### DIFF
--- a/src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
+++ b/src/Fixer/FunctionNotation/PhpdocToReturnTypeFixer.php
@@ -129,7 +129,7 @@ function my_foo()
     {
         // should be run after PhpdocScalarFixer.
         // should be run before ReturnTypeDeclarationFixer, FullyQualifiedStrictTypesFixer, NoSuperfluousPhpdocTagsFixer.
-        return 8;
+        return 12;
     }
 
     /**
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -62,8 +62,8 @@ class Foo {
      */
     public function getPriority()
     {
-        // should run before NoEmptyPhpdocFixer
-        return 6;
+        // should run before NoEmptyPhpdocFixer and PhpdocAddMissingParamAnnotationFixer
+        return 11;
     }
 
     /**
--- a/src/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocReturnSelfReferenceFixer.php
@@ -107,7 +107,7 @@ class Sample
     public function getPriority()
     {
         // must run before NoSuperfluousPhpdocTagsFixer
-        return 10;
+        return 12;
     }
 
     /**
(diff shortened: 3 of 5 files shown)


ntzm:
Can you add an integration test?

julienfalque:
Well, `phpdoc_add_missing_param_annotation` will always add a `@param` tag that exactly matches the argument definition, which makes it superfluous and thus a legitimate candidate for removal by `no_superfluous_phpdoc_tags`. IMO there is no priority issue here: using both rules at the same time doesn't make sense.

ntzm:
It should conflict then right @julienfalque?

julienfalque:
@ossinkine `@param mixed $foo` means you can pass any type of value to `$foo`, which is also what having no typehint in the definition of `$foo` means, hence the rule considers it superfluous. I see no value in adding such tag. On the other hand, if you add a description to the tag, then it adds information that are not available in the argument definition itself so it won't be removed.

@ntzm If you mean we should trigger an error when using both fixers, we don't have a way to do so and I don't think we want to. We already have some fixers for which we can pass options that don't make much sense but we don't trigger errors in such case.

ntzm:
@julienfalque we have `FixerFactory::getFixersConflicts` https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/src/FixerFactory.php#L221

julienfalque:
Wow, I didn't even know that :D

Then yes, I would rather use that to prevent using both rules at the same time.

ossinkine:
`phpdoc_add_missing_param_annotation` means you should explicit set up a type in docblock even it's `mixed` and this is not superfluous.
`no_superfluous_phpdoc_tags` removes tags with types like `string`, `bool` and so on if they are specified in method signature.

julienfalque:
> `no_superfluous_phpdoc_tags` removes tags with types like string, bool and so on if they are specified in method signature.

Not exactly: the rule removes tags that are superfluous, that is tags that don't add any information that isn't already available in the argument or return type definitions. This is why it considers `@param mixed` superfluous: no typehint could be the same as a `mixed` typehint, if it was possible.

ossinkine:
I described above my opinion. If method argument has no typehint and docblock has no `@param` tag you can not say for sure that the argument type is really a `mixed`. Perhaps the developer simply forgot to add a typehint and the argument is actually integer for example. If you see a docblock tag, you can safely say that the argument type is `mixed`, because this is clearly written. When a `mixed` typehint appears in the PHP, then such docblock tag will also become superfluous.

ossinkine:
I think we should add an option which says PHPDoc tag is superfluous only when method signature has the same typehint.

artem328:
Hi,

I faced with similar issue. I would consider phpdoc is superfluous if it doesn't have any information except of method signature. So what I want to achieve is full phpdoc if I have either method comment, or comment for parameter, or annotation. Can we add some option to this rule? Or maybe it's new rule?

ossinkine:
@artem328 It's not clear for me, what exactly do you mean? Can you give an example?

artem328:
@ossinkine 

E.g we have a method:

```php
public function foo(int $bar, string $baz): string {
    // code here
}
```

```php
/**
 * @param int $bar
 * @param string $baz
 *
 * @return string
 */
```
The PHPdoc above just copying the method signature, and is not usefull. So I would not keep it.

But if we have phpdoc like this:
```php
/**
 * This is description for method foo
 *
 * @param int $bar first parameter
 * @param string $baz
 *
 * @return string
 */
```
I would keep it, because PHPdoc now contains description and comment for parameter, and I don't want it becomes like this:
```php
/**
 * This is description for method foo
 *
 * @param int $bar first parameter
 */
```

So the main idea is:
* Remove phpdoc if it describes method signature
* Keep full phpdoc, if it contains any comments

PR #3929, opened 2018-07-17
Split calculation of code coverage to 2 runs

kubawerlos:

PR #3931, opened 2018-07-18
Use PHPUnitGoodPractices\Traits only for PHPUnit 7+

kubawerlos:


kubawerlos:
What would be the loss? Traits will be used a bit later (instead of "Fast test" stage in "Test" one - which we can be easily bring back by bumping PHP version for "Fast test").

The benefit is code coverages works - what I observed is when trait uses other trait (like [here](https://github.com/PHPUnitGoodPractices/Traits/blob/v1.5.1/src/ExpectationViaCodeOverAnnotationTrait.php#L48)) `phpdbg` - in the form of `phpdbg_start_oplog` and `phpdbg_end_oplog` (used [here](https://github.com/sebastianbergmann/php-code-coverage/blob/6.0.7/src/Driver/PHPDBG.php#L45-L55)) - cannot close some resources (which leads to a problem described [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3929#issuecomment-405922620)).

It is repeatable - if I leave only `ExpectOverSetExceptionTrait` it works well. If I leave any other trait (all others have the `*7` version "imported") it fails.

PR #3936, opened 2018-07-19
Add allow_untyped option to NoSuperfluousPhpdocTagsFixer

ossinkine:
This PR continues #3930.
I described my point of view in that PR conversation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1047,6 +1047,11 @@ Choose from the list of available rules:
   Removes ``@param`` and ``@return`` tags that don't provide any useful
   information.
 
+  Configuration options:
+
+  - ``allow_untyped`` (``bool``): whether to remove superfluous ``@param`` annotations
+    for parameters with specified type only; defaults to ``false``
+
 * **no_trailing_comma_in_list_call** [@Symfony]
 
   Remove trailing commas in list function calls.
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -15,6 +15,9 @@ namespace PhpCsFixer\Fixer\Phpdoc;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\DocBlock\Annotation;
 use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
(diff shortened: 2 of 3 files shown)


ossinkine:
In particular, such a requirement also has PHPStan

SpacePossum:
```
`@param mixed $foo` means you can pass any type of value to $foo, which is also what having no typehint in the definition of $foo means, hence the rule considers it superfluous. I see no value in adding such tag. 
```
(https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3930#issuecomment-406184572)

The value is adding itself, it states that the param is documented, i.e. the author did not forgot to document the param. When there is more than one param and one is typed and one mixed I like those both to be documented and only the typed ones.

This is why I typically don't use the `NoSuperfluousPhpdocTagsFixer`. I think the fixer does what it should do so I don't agree on adding the option by this PR. If you want to keep the params; remove the fixer from your config :)

That does leave the initial valid issue report on the conflict between the fixers that we should try to resolve in some other way.

ossinkine:
I respect your point, but what should I do in my case?
The coding standards in our company require, among other things, the following rules:
1. Always explicitly specify the type of each argument and the return type (typehint or docblock). PHPStan helps with this (why this rule was added you can read here https://github.com/phpstan/phpstan-strict-rules/issues/14). The very existence of a rule in PHPStan means that someone other than me has encountered a similar issue.
2. Docblock should not contain superfluous tags `@param`. Superfluous one means that the method signature already contains the corresponding typehint. NoSuperflousPhotsTagsFixer can help with this except for the `mixed` type because it contradicts the first rule.

Adding an option to the fixer will make it more flexible, solve my issue and does not break BC.

dmvdbrugge:
Add a description to the `@param mixed $var`s and it will no longer consider them superfluous.

Also, mixed is not always really mixed, it can't always handle _everything_. While a lot of work, you could specify what it _actually_ accepts. (And use the first option in other cases.)

SpacePossum:
In the sample the `mixed` is most likely not correct; you cannot serialize most `resource`-es so the doc could've been made better ;)
Adding the description might be the best way to go to explain why something is mixed.

I can see both sides now, so I'm +0 on this change now.

ossinkine:
Actually in Symfony implementation the serializer can get any type (even resource) but throws exception if serialization of this type does not supported.
We can argue this is good or bad, but not in this MR. This works that way and this needs to be documented.

julienfalque:
#4002 was just merged with similar feature so I'm closing here. Thank you for your time.

PR #3937, opened 2018-07-22
LowercaseStaticReferenceFixer - Fix "Parent" word in namespace

kubawerlos:


src/Fixer/Casing/LowercaseStaticReferenceFixer.php
```diff
@@ -86,7 +86,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
             }
 
             $prevIndex = $tokens->getPrevMeaningfulToken($index);
-            if ($tokens[$prevIndex]->isGivenKind([T_CONST, T_DOUBLE_COLON, T_FUNCTION, T_OBJECT_OPERATOR, T_PRIVATE, T_PROTECTED, T_PUBLIC])) {
+            if ($tokens[$prevIndex]->isGivenKind([T_CONST, T_DOUBLE_COLON, T_FUNCTION, T_NAMESPACE, T_NS_SEPARATOR, T_OBJECT_OPERATOR, T_PRIVATE, T_PROTECTED, T_PUBLIC])) {
```
keradus: why `T_NAMESPACE`? I don't see test for it

kubawerlos: It will handle `<?php namespace Foo\Parent;` case, but it wold be handed by `T_NS_SEPARATOR` in next token search, so you're right - I've added one more test.

SpacePossum:
Thank you @kubawerlos.

PR #3939, opened 2018-07-23
Add PhpUnitSizeClassFixer

malukenho:
_All PHPUnit test cases should have `@small`, `@medium` or `@large` annotation to enable run time limits._

Configuration options:

   - ``group`` (``'large'``, ``'medium'``, ``'small'``): define a specific group to be used
    in case no group is already in use; defaults to ``'small'``

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1357,6 +1357,16 @@ Choose from the list of available rules:
 
   *Risky rule: this fixer may change functions named ``setUp()`` or ``tearDown()`` outside of PHPUnit tests, when a class is wrongly seen as a PHPUnit test.*
 
+* **php_unit_size_class**
+
+  All PHPUnit test cases should have ``@small``, ``@medium`` or ``@large``
+  annotation to enable run time limits.
+
+  Configuration options:
+
+  - ``group`` (``'large'``, ``'medium'``, ``'small'``): define a specific group to be used
+    in case no group is already in use; defaults to ``'small'``
+
 * **php_unit_strict** [@PhpCsFixer:risky]
 
   PHPUnit methods like ``assertSame`` should be used instead of
--- a/src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
@@ -1,0 +1,279 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\Annotation;
+use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\DocBlock\Line;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use SplFileInfo;
+
+/**
+ * @author Jefersson Nathan <malukenho.dev@gmail.com>
+ */
+final class PhpUnitSizeClassFixer extends AbstractFixer implements WhitespacesAwareFixerInterface, ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All PHPUnit test cases should have `@small`, `@medium` or `@large` annotation to enable run time limits.',
+            [new CodeSample("<?php\nclass MyTest extends TestCase {}\n")],
+            'The special groups [small, medium, large] provides a way to identify tests that are taking long to be executed.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_STRING]);
+    }
+
+    protected function applyFix(SplFileInfo $file, Tokens $tokens)
+    {
+        $phpUnitTestCaseIndicator = new PhpUnitTestCaseIndicator();
+
+        foreach ($phpUnitTestCaseIndicator->findPhpUnitClasses($tokens, true) as $indexes) {
+            $this->markClassSize($tokens, $indexes[0]);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('group', 'Define a specific group to be used in case no group is already in use'))
+                ->setAllowedValues(['small', 'medium', 'large'])
+                ->setDefault('small')
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     */
+    private function markClassSize(Tokens $tokens, $startIndex)
+    {
+        $classIndex = $tokens->getPrevTokenOfKind($startIndex, [[T_CLASS]]);
+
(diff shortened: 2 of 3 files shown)


README.rst
```diff
@@ -1237,6 +1237,10 @@ Choose from the list of available rules:
   - ``types`` (a subset of ``['normal', 'final', 'abstract']``): what types of
     classes to mark as internal; defaults to ``['normal', 'final']``
 
+* **php_unit_size_class**
+
+  All PHPUnit test classes should have a @small, @medium or @large annotation.
```
ntzm: Please wrap with backticks to format like code: e.g. `@small`

tests/Fixer/PhpUnit/PhpUnitSizeClassFixerTest.php
```diff
@@ -0,0 +1,344 @@
...
+        return [
+            'It does not change normal classes' => [
+                '<?php
+
+class Hello
+{
+}
+',
+            ],
+            'It marks a test class as internal' => [
```
ntzm: Lots of references to `internal` where it should be `small` or similar

tests/Fixer/PhpUnit/PhpUnitSizeClassFixerTest.php
```diff
@@ -0,0 +1,344 @@
...
+
+if (class_exists("Foo\Bar")) {
+    /**
+     * @author malukenho again
+     *
+     *
+     * @covers \Other\Class
+     *
+     * @small
+     */
```
ntzm: Can a test be added like this but for tab indent?

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,279 @@
...
+            [new CodeSample("<?php\nclass MyTest extends TestCase {}\n")]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before FinalInternalClassFixer
```
SpacePossum: I'm not sure this is true, the priority test added and the entry in the FixerFactoryTest do not backup this claim either.
Maybe this is a copy-paste left over form another fixer?

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,279 @@
...
+        $this->createDocBlock($tokens, $docBlockIndex);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $i
+     *
+     * @return bool
+     */
+    private function isAllowedByConfiguration(Tokens $tokens, $i)
```
SpacePossum: what is the rationale behind filtering candidate classes on `final`, `abstract`, none of the previous?

tests/Fixer/PhpUnit/PhpUnitSizeClassFixerTest.php
```diff
@@ -0,0 +1,344 @@
...
+namespace PhpCsFixer\Tests\Fixer\PhpUnit;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @internal
+ *
+ * @author Jefersson Nathan <malukenho@phpse.net>
+ *
+ * @covers \PhpCsFixer\Fixer\PhpUnit\PhpUnitInternalClassFixer
```
SpacePossum: I think this is a copy-paste leftover, the test does not cover this fixer ;)

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,279 @@
...
+ */
+final class PhpUnitSizeClassFixer extends AbstractFixer implements WhitespacesAwareFixerInterface, ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All PHPUnit test cases should have @small, @medium or @large annotation.',
```
keradus: please, use long description field to provide reasoning why they should have it.

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,261 @@
...
+ * @author Jefersson Nathan <malukenho@phpse.net>
+ */
+final class PhpUnitSizeClassFixer extends AbstractFixer implements WhitespacesAwareFixerInterface, ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3939#discussion_r205985892

dmvdbrugge:
I feel like it doesn't make much sense to make `@small` the default. Isn't it more logical to make `@medium` default, which can than be either increased or decreased when determined so? Could also be an option which one you'd want. But then the same comment applies to the default of that option 😉

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,261 @@
...
+            'All PHPUnit test cases should have @small, @medium or @large annotation to enable run time limits.',
```
ntzm: Tests are failing due to these `@` not being wrapped in backticks

tests/Fixer/PhpUnit/PhpUnitSizeClassFixerTest.php
malukenho: Done

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
malukenho: Done

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
malukenho: Done

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,284 @@
...
+            'The special groups [small, medium, large] provides a way to identify tests that are taking long to be executed.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isTokenKindFound(T_CLASS);
```
SpacePossum: `$tokens->isAllTokenKindsFound([T_CLASS, T_STRING])` , as a single anonymous class is not a candidate

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    public function getPriority()
```
SpacePossum: if there are no priority constraints for the fixer you can remove this method and leave it to the parent to pick the default value.

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,284 @@
...
+        ]);
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $startIndex
+     */
+    private function markClassSize(Tokens $tokens, $startIndex)
+    {
+        $classIndex = $tokens->getPrevTokenOfKind($startIndex, [[T_CLASS]]);
```
SpacePossum: maybe we can update `findPhpUnitClasses` to return the index of `T_CLASS` as third value in the returned array (like to hear some opinions on that first though)

malukenho: I may be wrong, but I think that it is out of scope for a fixer.

malukenho:
@keradus @SpacePossum is there something else missing here?

src/Fixer/PhpUnit/PhpUnitSizeClassFixer.php
```diff
@@ -0,0 +1,279 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'All PHPUnit test cases should have `@small`, `@medium` or `@large` annotation to enable run time limits.',
+            [new CodeSample("<?php\nclass MyTest extends TestCase {}\n")],
+            'The special groups [small, medium, large] provides a way to identify tests that are taking long to be executed.'
```
keradus: and why one shall add those annotation ? phpunit doesn't require it.

malukenho: Yes, you are right, PHPUnit doesn't require it because by default tests are treated as `@small`.
One may set a time-out for each of this group, and then that makes sense to require it on tests.

malukenho:
Any news on this one?

SpacePossum:
@malukenho 
Thanks for all the work, it looks good to me.
Please squash my commit and rebase on top of 2.14.

Kindly note that is just my opinion, we need some more support for merging I think :)

malukenho:
@SpacePossum done!

keradus:
Thank you @malukenho.

PR #3944, opened 2018-07-24
ExplicitStringVariableFixer - fix array handling

### LINKED ISSUE #3943: ExplicitStringVariableFixer - wrong behavior for arrays
```php
<?php $a = "My name is $array[MY_CONSTANT] !";
```

```php
<?php $a = "My name is {$array[MY_CONSTANT]} !";
```

```php
<?php $a = "My name is {$array['MY_CONSTANT']} !";
```

It is a test case from ExplicitStringVariableFixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/1c9553ac16b1d10ea891c6f739aa9215d16de024/tests/Fixer/StringNotation/ExplicitStringVariableFixerTest.php#L117-L118

But it is wrong, `MY_CONSTANT` is not a constant in that case (see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3913), so it needs quotes in the explicit style.

gharlan:
fixes #3943


keradus:
Thank you @gharlan.

PR #3947, opened 2018-07-26
Add .gitignore and .gitattributes support to the config by the Finder

jdufresne:
Many PHP projects use NPM to build frontend assets. NPM installs third party packages to the `node_modules` directory. These packages sometimes contain PHP files, for example jquery-ui examples. Avoid running php-cs-fixer on these files in the default configuration.

### DIFF
--- a/src/Finder.php
+++ b/src/Finder.php
@@ -31,6 +31,7 @@ class Finder extends BaseFinder
             ->ignoreDotFiles(true)
             ->ignoreVCS(true)
             ->exclude('vendor')
+            ->exclude('node_modules')
         ;
     }
 }


dmvdbrugge:
While an edge case, I think it can be fixed in a broader, generalised way, by parsing (and excluding the results of) ignore files like `.gitignore`. That, however seems more like something to suggest to Symfony for the Finder component. I'd 👍 that too.

`node_modules` _should_ be in there anyway 😉

ntzm:
I think ignoring things in `.gitignore` would be a good idea, there's a ticket for adding support for it in symfony/finder https://github.com/symfony/symfony/issues/26714

SpacePossum:
I'm not in favor of listing more excludes on default, but I do like the idea in general.
Therefore I've tagged this is as WIP as I think having this through the Finder package is the way to go.

jdufresne:
Thanks for all the feedback. I agree, using `.gitignore` is a better and more general solution.

keradus:
@jdufresne , maybe you can help with this topic by solving it directly in Symfony Finder component ?

SpacePossum:
unblocked through; https://github.com/symfony/symfony/pull/30448

keradus:
let's call `ignoreVCSIgnored` if it's available

julienfalque:
Replaced by #4720.

PR #3948, opened 2018-07-28
DX: cleanup - remove redundant @param annotations

kubawerlos:


keradus:
actually, those are not just redundant, they are simply wrong :D

kubawerlos:
those are wrong by being redundant ;)

SpacePossum:
Thank you @kubawerlos.

Issue #3917, opened 2018-07-14
setFinder and arrays


uuf6429:
My original issue was that I ended up creating a finder instance and then excluding all files not in a list of files. Thinking there must be a better way I looked at the [documentation of `setFinder()`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/0790602be97425d79331dfa5ca920983e7b6faeb/src/ConfigInterface.php#L124) and at the actual implementation.

The PHPDoc there mentions `string[]` and in the code there were a few places were an array is allowed. Eventually, `ConfigurationResolver->iterableToTraversable` is called, where the array is converted to an `ArrayIterator`, regardless of content. Anyway, using an array of strings causes the following error:
```
In FileFilterIterator.php line 58:

  Expected instance of "\SplFileInfo", got "string".
```

Instead, simply passing an array of `\SplFileInfo` makes it work.

So maybe, that PHPDoc line should be:
```php
/**
 * @param iterable|\SplFileInfo[]|\Traversable $finder
```

Also, maybe it would be helpful to mention this in the readme as some kind of tip.

SpacePossum:
Hi @uuf6429 

Thanks for raising this issue. The incorrect typehint has been [fixed](https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/f06a457f7a9774eaa7908578a4c845dd61b1d5a2#diff-c88507a1ae2b9206a3ee7c42b0436ab7R40) and the fix was been released.
We no longer maintain the older versions so I'm not sure if there is anything left we can fix here.
Did you come across this erroneous doc somewhere else?

uuf6429:
@SpacePossum, apologies for the wrong link, I meant to link to the master branch:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/ConfigInterface.php#L124
What does the master branch represent in this case?

Also, regarding the fix itself, shouldn't it say somewhere that the final result should be a bunch of `\SplFileInfo`? `iterable|\Traversable` does tell as much. Also, in the fix, arrays are still technically allowed (but this fact ist not documented).

I haven't seen this documented elsewhere, by the way.

SpacePossum:
Ah I see now, the doc for `getFinder` was fixed, but `setFinder` was indeed not fixed.
I think we want to keep supporting `\SplFileInfo[]` so both docs should be fixed.
[...]

PR #3949, opened 2018-07-28
DX: cleanup - fixing PHPDoc types in ConfigInterface

kubawerlos:
Solves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3917

Simply following the https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3917#issuecomment-408616281

### DIFF
--- a/src/ConfigInterface.php
+++ b/src/ConfigInterface.php
@@ -37,7 +37,7 @@ interface ConfigInterface
     /**
      * Returns files to scan.
      *
-     * @return iterable|\Traversable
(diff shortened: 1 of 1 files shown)


kubawerlos:
Sure, I just saw the comment in an issue and had the repo open so went with this PR.

kubawerlos:
It's over 1 year since last activity, I'm not even sure if this is valid or if PHPStan would handle this case at some point. Let's close it for now, we always can have it re-open if needed.

Issue #3874, opened 2018-07-05
Circle CI needs to be migrated to v2


stof:
Circle CI builds are displaying a warning about running on v1 which will be shutdown after August 31, 2018.

So migrating the CI to Circle CI v2 is needed before the end of the summer.

keradus:
when v2 was launched, PHP was not offered for it (at least officially).
I hope it's available now...

stof:
I don't think it is offered officially on v1 either. the current config is installing PHP on its own: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/ceec561308e27f2778c869d82e0c08109ff27d17/circle.yml#L15

stof:
I have absolutely no experience with Circle CI. I opened this issue due to seeing a warning when looking at the build for one of my PRs.

PR #3950, opened 2018-07-29
Circle CI v2 yml

siad007:
Added a new circle ci version 2 config file.
Note that i could not tested this but maybe it is still helpful for you.

Related to #3874 .


keradus:
Hi @siad007 ! Very first - big thanks for support on this topic! Looks promising, indeed!

Sadly, for PHP CS Fixer, the whole point of using CircleCI is to run tests on MacOS (same as we use Travis for Linux and AppVeyor for Windows)

While using docker image instead of build php using brew on host OS directly, we are not testing on MacOS anymore:
```
ker@dus:/tmp λ docker run -it --rm circleci/php:7.2 php -r "print_r(posix_uname()['sysname']);"
Linux
```

Can you update your PR please? thanks !

kubawerlos:
Wasn't the purpose of CircleCI to check if the fixer will work on macOS? Ping @keradus

siad007:
@keradus @kubawerlos added your changes right now and it semms ok. But do you have an idea why the former linux based config produced the test failures?

kubawerlos:
@siad007 in `.circleci/config.yml` in line 22 there are trailing whitespaces.

.gitattributes
```diff
@@ -20,6 +20,7 @@ box.json export-ignore
 check_trailing_spaces.sh export-ignore
 phpmd.xml export-ignore
 phpunit.xml.dist export-ignore
+.circleci export-ignore
```
kubawerlos: Can we move it a little bit up to keep alphabetic order? And also add `/` after the directory name to point it is directory?

siad007:
Ah and btw 8.2 is not supported any longer so i updated to xcode 9.0.

keradus:
Thank you @siad007.

PR #3952, opened 2018-07-30
AbstractFixerTestCase - drop testing method already provided by trait

keradus:


SpacePossum:
Thanks mister @keradus

Issue #3956, opened 2018-08-01
Withespace after opening tag no trimed


waghanza:
Hi,

I have a file starting with a withespace after the opening tag.
```php
<?php 
```

Using
~~~
composer exec php-cs-fixer -- fix --rules=@PSR1,@PSR2 src/
~~~

does not trim this withespace, but
~~~
composer exec php-cs-fixer -- fix --rules=@PSR1,@PSR2,@Symfony src/
~~~
does

+ **PHP** : `7.1.17`
+ **PHP CS Fixer** : `2.12.2`

Regards,

SpacePossum:
`no_trailing_whitespace` rule should be updated to check for this special case

SpacePossum:
`<?php ` cannot be fixed to `<?php` because that is not valid PHP

SpacePossum:
```
<?php

```
with space after `<?php`  can be trimmed (but the newline must be there than)

PR #3957, opened 2018-08-01
NoTrailingWhitespaceFixer - trim space after opening tag

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3956

That's what I call edge case :)

The iteration over `$tokens` has to be reversed, so the whitespace after the `T_OPEN_TAG` is already trimmed (otherwise we could replace space with another space) when handing the open tag.

### DIFF
--- a/src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
+++ b/src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
@@ -62,17 +62,37 @@ final class NoTrailingWhitespaceFixer extends AbstractFixer
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
-        foreach ($tokens as $index => $token) {
+        for ($index = count($tokens) - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+            if (
+                $token->isGivenKind(T_OPEN_TAG)
+                && $tokens->offsetExists($index + 1)
+                && $tokens[$index + 1]->isWhitespace()
+                && 1 === Preg::match('/(.*)\h$/', $token->getContent(), $openTagMatches)
+                && 1 === Preg::match('/^(\R)(.*)$/s', $tokens[$index + 1]->getContent(), $whitespaceMatches)
+            ) {
+                $tokens[$index] = new Token([T_OPEN_TAG, $openTagMatches[1].$whitespaceMatches[1]]);
+                if ('' === $whitespaceMatches[2]) {
+                    $tokens->clearAt($index + 1);
+                } else {
+                    $tokens[$index + 1] = new Token([T_WHITESPACE, $whitespaceMatches[2]]);
+                }
+
+                continue;
+            }
+
             if (!$token->isWhitespace()) {
                 continue;
             }
 
-            $lines = Preg::split("/([\r\n]+)/", $token->getContent(), -1, PREG_SPLIT_DELIM_CAPTURE);
+            $lines = Preg::split('/(\\R+)/', $token->getContent(), -1, PREG_SPLIT_DELIM_CAPTURE);
             $linesSize = count($lines);
 
             // fix only multiline whitespaces or singleline whitespaces at the end of file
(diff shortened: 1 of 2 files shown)


tests/Fixer/Whitespace/NoTrailingWhitespaceFixerTest.php
```diff
@@ -96,6 +96,25 @@ public function provideFixCases()
...
+                "<?php\necho 'Hello World';",
+                "<?php  \necho 'Hello World';",
+            ],
+            [
+                "<?php\necho 'Hello World';",
+                "<?php	\necho 'Hello World';",
+            ],
+            [
+                '<?php ', // do not trim this as "<?php" is not valid PHP
+            ],
```
SpacePossum: can you add `don't fix` test for;
`"<?php\t\t"` two tabs
`"<?php  "` two spaces

or should those be trimmed down to `<?php ` single space .. .hmmm ;)  (I don't think these need fixing, but a test would be nice)

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
```diff
@@ -62,7 +62,25 @@ public function isCandidate(Tokens $tokens)
...
+            $token = $tokens[$index];
+            if (
+                $token->isGivenKind(T_OPEN_TAG)
+                && $tokens->offsetExists($index + 1)
+                && $tokens[$index + 1]->isWhitespace()
+                && 1 === Preg::match('/(.*)\h$/', $token->getContent(), $openTagMatches)
+                && 1 === Preg::match('/^(\R)(.*)$/', $tokens[$index + 1]->getContent(), $whitespaceMatches)
+            ) {
+                $tokens[$index] = new Token([T_OPEN_TAG, $openTagMatches[1].$whitespaceMatches[1]]);
+                if (empty($restOfWhitespace)) {
```
SpacePossum: can we avoid `empty` here? , `$restOfWhitespace` is not defined here I think?

tests/Fixer/Whitespace/NoTrailingWhitespaceFixerTest.php
kubawerlos: I'd say trim them without the character that is part of `T_OPEN_TAG`.

src/Fixer/Whitespace/NoTrailingWhitespaceFixer.php
kubawerlos: Fixed

tests/Fixer/Whitespace/NoTrailingWhitespaceFixerTest.php
```diff
@@ -96,6 +96,37 @@ public function provideFixCases()
...
+                '<?php ',
+                '<?php  ',
+            ],
+            [
+                "<?php\t",
+                "<?php\t\t",
+            ],
+            [
+                '<?php ', // do not trim this as "<?php" is not valid PHP
+            ],
```
SpacePossum: please add this case;
```php
[
"<?php\n      \n   \n    "',
]
```

kubawerlos: @SpacePossum added, I hope you mean 2nd one with fix (as the first set of spaces is in non-blank line).

tests/Fixer/Whitespace/NoTrailingWhitespaceFixerTest.php
```diff
@@ -96,6 +96,44 @@ public function provideFixCases()
...
+            [
+                "<?php\n      \n   \n    ",
+            ],
+            [
+                "<?php\n   \n    ",
```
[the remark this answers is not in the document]
kubawerlos: The name of the fixer is a little misleading, here is description:

> Remove trailing whitespace at the end of non-blank lines.

So, the only non-blank line is the one with open tag.

keradus:
Thank you @kubawerlos.

PR #3604, opened 2018-03-15
NoEmptyPhpdocFixer - do not leave empty line after docs removal

TomasVotruba:
As suggested here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3574#issuecomment-373484579

Tests should fail now.

I can fix/complete tests, after initial review

PR #3961, opened 2018-08-07
remove complete line if becomes empty after removal

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3604

This add a method ~~to the `Tokens` class~~ to remove a token and if that causes that the line on which the token was to become completely empty remove that line as well.
Adding this method is in the first commit, the naming is terrible IMHO but lets discuss if the naming is the issue or maybe we want some traits to bring this kind of functionality and not add it to the tokens class.

~~The other commits show some fixers that benefit from this new method.~~

~~There are a bunch of other candidate fixers that could use this new method (from the top of my head);~~
~~- useless elseif~~
~~- empty comment (blocked by transformer issue)~~
~~- header comment (when removing the header comment/PHPDoc)~~
~~- ReturnAssignmentFixer~~

~~I left those out because of other issues or just because I'm running out of time.~~

### DIFF
--- a/src/Tokenizer/Manipulator/TokenRemover.php
+++ b/src/Tokenizer/Manipulator/TokenRemover.php
@@ -1,0 +1,175 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Manipulator;
+
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @internal
+ */
+final class TokenRemover
+{
+    /**
+     * @var Tokens
+     */
+    private $tokens;
+
+    public function __construct(Tokens $tokens)
+    {
+        $this->tokens = $tokens;
+    }
+
+    public function clearToken($index)
+    {
+        $previous = $this->tokens->getNonEmptySibling($index, -1);
+        $next = $this->tokens->getNonEmptySibling($index, 1);
+
+        // --- if nothing before the token check space after it ----------
+
+        if (null === $previous) {
+            // remove trailing space if any
+            if (false === strpos($this->tokens[$index]->getContent(), "\n") && null !== $next && $this->tokens[$next]->isWhitespace()) {
+                $linebreakAfterTokenStrIndex = strpos($this->tokens[$next]->getContent(), "\n");
+
+                if (false === $linebreakAfterTokenStrIndex) {
+                    if (null === $this->tokens->getNonEmptySibling($next, 1)) {
+                        $this->tokens->clearTokenAndMergeSurroundingWhitespace($next);
+                    }
+                } else {
+                    $newContent = substr($this->tokens[$next]->getContent(), $linebreakAfterTokenStrIndex + 1);
+
+                    if ('' === $newContent || false === $newContent) { // false check for PHP5.6
+                        $this->tokens->clearTokenAndMergeSurroundingWhitespace($next);
+                    } else {
+                        $this->tokens[$next] = new Token([T_WHITESPACE, $newContent]);
+                    }
+                }
+            }
+
+            $this->tokens->clearTokenAndMergeSurroundingWhitespace($index);
+
+            return;
+        }
+
+        // --- check space before the token ------------------------------
+
+        if (!$this->tokens[$previous]->isWhitespace() && !$this->tokens[$previous]->isGivenKind(T_OPEN_TAG)) {
+            $this->tokens->clearTokenAndMergeSurroundingWhitespace($index);
+
+            return; // there is more content on the same line before the token
+        }
+
+        $prePrevious = null;
+        $linebreakBeforeTokenStrIndex = strrpos($this->tokens[$previous]->getContent(), "\n");
+
+        if (false === $linebreakBeforeTokenStrIndex && $this->tokens[$previous]->isWhitespace()) {
+            // test for open tag with line break followed by space without linebreak followed by token to remove
+            $prePrevious = $this->tokens->getNonEmptySibling($previous, -1);
+
+            if (null !== $prePrevious) {
+                $linebreakBeforeTokenStrIndex = strpos($this->tokens[$prePrevious]->getContent(), "\n");
+            }
+        }
+
+        if (false === $linebreakBeforeTokenStrIndex) {
+            $this->tokens->clearTokenAndMergeSurroundingWhitespace($index);
+
+            return; // there is more content on the same line before the token
+        }
+
+        // --- check space after the token -------------------------------
+
+        $linebreakAfterTokenStrIndex = false;
+
+        if (null !== $next) {
+            if (!$this->tokens[$next]->isWhitespace()) {
+                $this->tokens->clearTokenAndMergeSurroundingWhitespace($index);
+
+                return; // more content on the same line after the token
+            }
+
+            $linebreakAfterTokenStrIndex = strpos($this->tokens[$next]->getContent(), "\n");
+
+            if (false === $linebreakAfterTokenStrIndex && null !== $this->tokens->getNonEmptySibling($next, 1)) {
+                $this->tokens->clearTokenAndMergeSurroundingWhitespace($index);
+
+                return; // more content on the same line after the token (here trailing space that is not the last token)
+            }
+        }
+
+        // --- clear line break before the token -------------------------
+
+        if (null !== $prePrevious) {
+            $this->tokens->clearTokenAndMergeSurroundingWhitespace($previous);
+
+            // open tag -> space without line break -> token -> [nothing/trailing space with or without line break]
+            if (null === $next || false === $linebreakAfterTokenStrIndex) {
+                $this->tokens[$prePrevious] = new Token([T_OPEN_TAG, rtrim($this->tokens[$prePrevious]->getContent()).' ']);
+            } else {
+                // remove the line break from the next space token and not from the open tag
+                $newContent = substr($this->tokens[$next]->getContent(), $linebreakAfterTokenStrIndex + 1);
+                if ('' === $newContent || false === $newContent) { // false check for PHP5.6
+                    $this->tokens->clearTokenAndMergeSurroundingWhitespace($next);
+                } else {
+                    $this->tokens[$next] = new Token([T_WHITESPACE, $newContent]);
+                }
+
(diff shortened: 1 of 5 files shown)


tests/Tokenizer/TokensTest.php
```diff
@@ -29,6 +29,587 @@
...
+    }
+
+    /**
+     * @param int    $index    token index to clear
+     * @param string $expected PHP source expected after clearing
+     * @param string $input    PHP source input
+     *
+     * @dataProvider provideClearTokenMergeSurroundingWhitespaceRemoveLineIfLeftEmpty2Cases
+     */
+    public function testClearTokenMergeSurroundingWhitespaceRemoveLineIfLeftEmpty2($index, $expected, $input)
```
kubawerlos: Can we have one test test and compare tokens collection like here and not code like in the `testClearTokenMergeSurroundingWhitespaceRemoveLineIfLeftEmpty`?

tests/Tokenizer/TokensTest.php
```diff
@@ -29,6 +29,587 @@
...
+/* 17d*/;# 18d
+',
+            ],
+            // --- special cases
+            '18/' => [
+                0,
+                '',
+                '<?php ',
+            ],
+            '20/' => [
```
kubawerlos: Are these indices somehow useful?

tests/Tokenizer/TokensTest.php
```diff
@@ -29,6 +29,587 @@
...
+    '.'
+     '.'
+    # delete this line only 22
+     '.'
+    '.'
+   '.'
+  '.'
+ ',
+            ],
+        ];
```
kubawerlos: Can you add this case:
```php
    [
        1,
        '<?php
',
        '<?php
TRY_TO_REMOVE_LAST_ELEMENT',
    ],
```
?

tests/Tokenizer/TokensTest.php
SpacePossum: Here we cannot because the tests here show some edge cases causing hard to fix issues.
For example:
```
            'D3' => [
                0,
                " \necho 14;",
                "<?php\n \necho 14;",
            ],
```

By removing `<?php` all other tokens stay the same, however when re-tokenizing the new content " \necho 14;" these are different. This is not unique for the new method, it as a limitation of our Tokens model.

tests/Tokenizer/TokensTest.php
SpacePossum: When a case fails it more easy to find back which one it is.

tests/Tokenizer/TokensTest.php
SpacePossum: this: '<?php
TRY_TO_REMOVE_LAST_ELEMENT' does not lint.

However I added:
```
            [
                1,
                '<?php ',
                '<?php
;',
            ],
```

tests/Tokenizer/TokensTest.php
SpacePossum: I removed the confusing indices

tests/Tokenizer/TokensTest.php
SpacePossum: when removing the token `;` (@ index 1)  the result is; `<?php\n` , so now there is an empty line left after the removal (before there wasn't) so the function will remove the line,
however `<?php` has a different meaning than `<?php\n`, therefore it is changed to `<?php ` which has the same meaning (https://3v4l.org/VJKHA)

kubawerlos: Oh, I thought `<?php` would result with syntax error or at least `T_INLINE_HTML` - not with pair of tokens.

Anyway - this would make conflict with `no_trailing_whitespace`.

SpacePossum: I think there is no conflict (on this level) as `no_trailing_whitespace`  will not trim ['<?php ']( https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/tests/Fixer/Whitespace/NoTrailingWhitespaceFixerTest.php#L120) (as it would change the meaning of the token)

keradus:
> This add a method to the Tokens class to remove a token and if that causes that the line on which the token was to become completely empty remove that line as well.
Adding this method is in the first commit, the naming is terrible IMHO but lets discuss if the naming is the issue or maybe we want some traits to bring this kind of functionality and not add it to the tokens class.

the introduced method is almost 150 lines and it's hard to read/understand/change

We have BC promise under `Tokens`. We shall not add this method there.
IMHO, similar as we have `Analyzers`, we can have sth like `Manipulators`

Let set up the plan for this PR. What about having:
- extract new logic from Tokens into separated class
- apply to other rules, as well
?

also, it may happen that `clearTokenAndMergeSurroundingWhitespace` would no longer be needed...

keradus:
(IMHO, not a prio PR, i was just checking what was it, to know shall I merge it now - let's focus on other PRs first and come back to this one, this one looks more complicated, let's close simpler ones first)

src/Tokenizer/Tokens.php
```diff
@@ -1207,6 +1207,139 @@ public function isPartialCodeMultiline($start, $end)
         return false;
     }
 
+    public function clearTokenMergeSurroundingWhitespaceRemoveLineIfLeftEmpty($index)
+    {
+        $previous = $this->getNonEmptySibling($index, -1);
+        $next = $this->getNonEmptySibling($index, 1);
+
+        // --- check space around the token ------------------------------
```
kubawerlos: This comment is to remove, right?

src/Tokenizer/Tokens.php
```diff
@@ -1207,6 +1207,139 @@ public function isPartialCodeMultiline($start, $end)
         return false;
     }
 
+    public function clearTokenMergeSurroundingWhitespaceRemoveLineIfLeftEmpty($index)
```
kubawerlos: Wouldn't name `clearTokenAndRemoveLineIfLeftEmpty` be enough?

SpacePossum:
anything left todo here?

kubawerlos:
@SpacePossum can you take a look at failing test from here: https://github.com/kubawerlos/php-cs-fixer-custom-fixers/pull/218 ?

tests/Tokenizer/TokensTest.php
```diff
@@ -29,6 +29,625 @@ final class TokensTest extends TestCase
...
+                    echo 1;
+                ',
+                '<?php
+                    echo 27;
+                    /* */echo 1;
+                ',
+            ],
+            [
+                5,
+                '<?php echo 18;'.'
```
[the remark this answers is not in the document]
SpacePossum: thanks, removed!

SpacePossum:
```
Could you please add an integration test with single_blank_line_at_eof and one or more other rules proving that the empty line at end of file is not removed?
```

not sure if I understand, that fixer is not touched in this PR

julienfalque:
@SpacePossum Indeed, the new method has to be called si nothing to do, please forget my comment :)

src/Tokenizer/Tokens.php
SpacePossum: renamed inside the new class

SpacePossum:
I've put the thing in a class in the `Manipulators` NS as suggested by @keradus 
I don't see much overhead to make the method more easy, it is just dealing with PHP code so....
I also still see a use for `clearTokenAndMergeSurroundingWhitespace` so not sure why we would want to remove it.
Also removed all the usages and other changes that were in this PR in the hope this can be reviewed and done with first. After that we can start using this to clean up a lot of duplicate code in the various fixers and let even more fixers remove tokens and lines in a better fashion.

src/Tokenizer/Manipulators/TokenRemover.php
```diff
@@ -0,0 +1,162 @@
...
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Manipulators;
```
kubawerlos: In `PhpCsFixer\Tokenizer` we have `Analyzer`, `Generator`, `Resolver` and `Transformer`, so let's make this `Manipulator`.

src/Tokenizer/Tokens.php
```diff
@@ -1223,13 +1223,21 @@ public function clearTokenAndMergeSurroundingWhitespace($index)
 
         $prevIndex = $this->getNonEmptySibling($index, -1);
 
-        if ($this[$prevIndex]->isWhitespace()) {
+        if (null === $prevIndex) {
```
kubawerlos: Can't the changes in `Tokens` class be reverted? Or is this addressing some issue (maybe separate PR then)?

SpacePossum: the changes are needed for this PR, not sure what extracting these give us?

kubawerlos:
@SpacePossum what's the status of this PR? Will we be able to include this in [2.15.6](https://github.com/FriendsOfPHP/PHP-CS-Fixer/milestone/138)?

SpacePossum:
Not sure, I'm rewriting this PR for over a year towards the liking of reviewers, but no agreement has been reached this should be added to the code base.

kubawerlos:
I'd say after changing `Manipulators` to `Manipulator` and fix for that failing test I'm for merging this.

SpacePossum:
:+1: I moved it to the other namespace

src/Tokenizer/Manipulator/TokenRemover.php
```diff
@@ -0,0 +1,166 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Manipulator;
+
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+final class TokenRemover
```
julienfalque: `@internal`?

kubawerlos:
@SpacePossum please add this test case:
```php
        [
            '<?php
/* other comment
 */
',
            '<?php
/* other comment
 *//* to remove */
',
        ],
```

SpacePossum:
thanks @kubawerlos  ! test and fix added, please review again when you've the time :)

kubawerlos:
There are still at least 2 problems.

Code like:
```php
<?php
/* remove me */
```

without any whitespace after the comment is changed to `<?php ` when should be to `<?php\n`.

Other is that empty token is added somehow: https://github.com/kubawerlos/php-cs-fixer-custom-fixers/pull/218/checks?check_run_id=811898353 - I'm not sure it's worth investigating and creating test or simple condition before creating token is enough.

SpacePossum:
>without any whitespace after the comment is changed to <?php when should be to <?php\n.

I'm not sure if I agree, shouldn't the complete line be removed by this method?

Second thing is worth looking into, thanks! :)

Issue #2571, opened 2017-02-28
Fixer for native classes casing


jtojnar:
Similarly to `NativeFunctionCasingFixer`, there should be a fixer for [native classes](https://secure.php.net/manual/en/reserved.classes.php) like `Exception`, `stdClass` or `Closure`.

SpacePossum:
Thanks for the suggestion @jtojnar all :+1:  for it 
I just saw I labelled your request and dropped you a +1 but never a reaction ;)
Please let us know if we can help out getting this implemented :)

Wirone:
Closing, as this was introduced in #6262 and works as expected:

<img width="897" alt="image" src="https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/assets/600668/247501a6-0cfd-48af-9071-2f6dde0d713d">

PR #3969, opened 2018-08-12
Add DeclaredClassCasingFilter

siad007:
Related to #2571


src/Fixer/Casing/NativeClassCasingFixer.php
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Classes should be called using the correct casing.',
```
ntzm:
```diff
-should be called using
+should be referred to using
```

src/Fixer/Casing/NativeClassCasingFixer.php
```diff
@@ -0,0 +1,110 @@
...
+            if ($name = $this->nativeClass($index, $token, $tokens)) {
+                $tokens[$index] = new Token([T_STRING, self::$nativeClassNames[$name]]);
+            }
+        }
+    }
+
+    /**
+     * Get the lower case name of the native class or null.
+     *
+     * @param $index
```
ntzm: Needs a type

src/Fixer/Casing/NativeClassCasingFixer.php
```diff
@@ -0,0 +1,110 @@
...
+     * @param Token  $token
+     * @param Tokens $tokens
+     *
+     * @return bool|string
```
ntzm: `null|string`

src/Fixer/Casing/NativeClassCasingFixer.php
```diff
@@ -0,0 +1,110 @@
...
+                ]
+            )
+            &&
+            !(
+                $tokens[$beforeClassName]->isGivenKind(T_NS_SEPARATOR)
+                &&
+                $tokens[$tokens->getPrevMeaningfulToken($beforeClassName)]->isGivenKind([T_STRING])
+            )
+            ? $lower
+            : null;
```
ntzm: This logic is very hard to grok imo, if it was split into multiple `if`s it may be more readable

src/Fixer/Casing/NativeClassCasingFixer.php
```diff
@@ -0,0 +1,110 @@
...
+final class NativeClassCasingFixer extends AbstractFixer
+{
+    private static $nativeClassNames;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        if (null === self::$nativeClassNames) {
+            foreach (get_declared_classes() as $class) {
```
ntzm: Personally, I use this project as a library sometimes, so it's not always guaranteed that this will only target native classes.

ntzm:
So the scope here has changed to use declared classes instead of native classes. I don't like this change because the output of the fixer can be changed depending on what classes are defined before construction, which in my case of using it as a library, it could be anything. Also, different environments will produce different results, such as:

- I am using `ext-pdo` on my local environment, which defines the `PDO` class
- I have some code that uses `PDo`
- On my local, when I run the fixer, it fixes it correctly to `PDO`
- However, on my CI setup, I might not have `ext-pdo` enabled, so it wouldn't fix that case

SpacePossum:
I don't share that concern, if you use different PHP builds/SAPI you get different results from a range of fixers already. It is also a one way fix, so on one system it fixes more than on another, but it never reverts a good fix.

src/Fixer/Casing/DeclaredClassCasingFixer.php
```diff
@@ -0,0 +1,116 @@
...
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author siad007
+ */
+final class DeclaredClassCasingFixer extends AbstractFixer
+{
+    private static $declaredClassNames;
```
ntzm: Can you add a PHPDoc with `@var` to explain what type this property has?

src/Fixer/Casing/DeclaredClassCasingFixer.php
```diff
@@ -0,0 +1,116 @@
...
+     * @param Tokens $tokens
+     *
+     * @return null|string
+     */
+    private function declaredClass($index, Token $token, Tokens $tokens)
+    {
+        $beforeClassName = $tokens->getPrevMeaningfulToken($index);
+        $lower = strtolower($token->getContent());
+
+        $isDeclaredClass = null;
```
ntzm: When I see a method or variable starting with `is` I assume it's a boolean, I think this variable name could be clearer.

ntzm:
@siad007 would you like someone to finish off this PR if you haven't got time to? It's a very useful fixer

src/Fixer/Casing/DeclaredClassCasingFixer.php
```diff
@@ -0,0 +1,116 @@
...
+final class DeclaredClassCasingFixer extends AbstractFixer
+{
+    private static $declaredClassNames;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        if (null === self::$declaredClassNames) {
+            foreach (get_declared_classes() as $class) {
```
kubawerlos: I'm sceptical about this line - `get_declared_classes` will not return classes that are declared, but those declared in the current script - depending how autoloader was called - which mean one day it will return some classes and another day more classes - adding some fixer to config might accidentally "discover" more classes.

When I was working on similar [fixer](https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/v1.13.0/src/Fixer/InternalClassCasingFixer.php#L90) I have ended with only internal classes to handle.

siad007:
@ntzm i have fixed the last pending issues right now

SpacePossum:
@siad007 thanks for the update.
I added another commit, please re-review @ntzm and @kubawerlos  . Note that this fixer does not fix classes outside of the global namespace, so even if more classes are already loaded it, for example it will never fix `Foo\bar`, only `\Foo` and `Foo` (if needed)

tests/Fixer/Casing/DeclaredClassCasingFixerTest.php
```diff
@@ -0,0 +1,139 @@
...
+
+                class test
+                {
+                    use exception;
+                }
+                ',
+            ],
+            [
+                '<?php
+                use stdclass as exception;
```
ntzm: Should this not be fixed to

```
use stdClass as exception;
```

?

SpacePossum: @siad007 can you pick this up? please, also rebase and squash this, then change the target branch to master please :)

SpacePossum:
Thanks @siad007 for all the work here!

The feature has been added to the code base here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/6262 , with big difference that it only fixes internal class names and addresses some other feedback items from here, thanks all :+1:

PR #3987, opened 2018-08-31
Run HeaderCommentFixer after NoBlankLinesAfterPhpdocFixer

StanAngeloff:
This ensures the output is stable when `HeaderCommentFixer` is configured with `["separate" => "bottom", "commentType" => "PHPDoc"]` and the target file has no `namespace` or `declare()` construct.

Test case added, test suite passes.


SpacePossum:
Thanks for this PR @StanAngeloff! Looking good so far, not sure what is/was up with circleci, lets see what it does now

keradus:
Thank you @StanAngeloff.

PR #3995, opened 2018-09-18
Change default config of native_function_invocation

dunglas:
According to the following comment of @nicolas-grekas, the default configuration of `native_function_invocation` should be changed to prepend a backslash only to functions that can be optimized by the PHP engine:

> Adding more functions as opcodes would make the engine slower, because routing more opcodes requires more CPU. So this will not happen :)
> The policy is to never add \, except when it provides a significant advantage, i.e. for those special functions that php-cs-fixer/fabbot handle.

https://github.com/symfony/panther/pull/61#discussion_r212794498

This PR changes the default config accordingly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -889,7 +889,7 @@ Choose from the list of available rules:
   - ``include`` (``array``): list of function names or sets to fix. Defined sets are
     ``@internal`` (all native functions), ``@all`` (all global functions) and
     ``@compiler_optimized`` (functions that are specially optimized by Zend);
-    defaults to ``['@internal']``
+    defaults to ``['@compiler_optimized']``
   - ``scope`` (``'all'``, ``'namespaced'``): only fix function calls that are made
     within a namespace or fix all; defaults to ``'all'``
   - ``strict`` (``bool``): whether leading ``\`` of function call not meant to have it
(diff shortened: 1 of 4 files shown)


dmvdbrugge:
Not sure if I personally agree to this, but if this is what's wanted, it's a bc-break so it has to target `3.0` not `master`.

keradus:
Changing default configuration of rule is BC breaker indeed, @dmvdbrugge , good spot :+1:

Sadly, I see sth wrong happen with this branch - I see lot of not related changes in diff.

dmvdbrugge:
Probably because he just targeted the other branch, instead of rebasing first

dmvdbrugge:
It also shows @keradus that you have not yet merged the 2.13 release into 3.0

dunglas:
> Sadly, I see sth wrong happen with this branch - I see lot of not related changes in diff.

I tried the "change base" button provided by GitHub, but it doesn't rebase automatically. This is fixed now.

dunglas:
@keradus done

SpacePossum:
sadly the tests are failure stopping this PR from being merged

SpacePossum:
Thank you @dunglas.

PR #4000, opened 2018-09-23
FinalStaticAccessFixer - Introduction

### LINKED ISSUE #3999: Contradiction between PHPUnit rules when they are used with defaults
There is a contradiction between `php_unit_internal_class`, `final_internal_class`, and `php_unit_test_case_static_method_calls` rules when they are used together with default options.

1. `php_unit_internal_class` marks PHPUnit test classes as internal
2. `final_internal_class` makes them final
3. `php_unit_test_case_static_method_calls` makes all calls to PHPUnit methods use the `static` keyword, **but there is no point in using the `static` keyword in a final class**

I propose to change the default value of the `call_type` option of the `php_unit_test_case_static_method_calls` rule from `'static'` to `'self'`.

ntzm:
Partially handles #3999.

Any other integration tests needed?


SpacePossum:
Please add the following tests.

to `public function provideFixCases()` (think will pass)

```php
            'in multiple methods and classes' => [
                '<?php
                    final class X {
                        public function b() { echo self::class; }
                        public function c() { echo self::class; }
                    }
                    final class Y {
                        public static $var = 2;
                        public function b() { echo self::class; }
                        public function c() { static $a = 0; echo self::class.$a; }
                    }
                    class Foo {
                        public function b() { echo static::class; }
                        public function c() { echo static::class; }
                    }
                    final class Z {
                        public function b() { echo self::class; }
                        public function c() { return static function(){}; }
                    }
                ',
                '<?php
                    final class X {
                        public function b() { echo static::class; }
                        public function c() { echo static::class; }
                    }
                    final class Y {
                        public static $var = 2;
                        public function b() { echo static::class; }
                        public function c() { static $a = 0; echo static::class.$a; }
                    }
                    class Foo {
                        public function b() { echo static::class; }
                        public function c() { echo static::class; }
                    }
                    final class Z {
                        public function b() { echo static::class; }
                        public function c() { return static function(){}; }
                    }
                ',
            ],
```

to `public function provideFix70Cases()` (think will fail)

```php
'property and nested anonymous class' => [
                '<?php final class A { public $b = self::class; public function foo(){ return new class { public $b = static::class; }; }}',
                '<?php final class A { public $b = static::class; public function foo(){ return new class { public $b = static::class; }; }}',
            ],
```

ntzm:
Thanks for your review @SpacePossum, good spot! Nested anonymous classes should be ignored now.

src/Fixer/ClassNotation/FinalStaticAccessFixer.php
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return int
+     */
+    private function getEndOfAnonymousClass(Tokens $tokens, $index)
+    {
+        $instantiationBraceStart = $tokens->getNextMeaningfulToken($index);
```
SpacePossum: can't you jump directly to `$bodyBraceStart = $tokens->getNextTokenOfKind($index, ['{']);` ? I think an anonymous always have a body  (`{` `}`) and you are never interested in the tokens between `new class` and the end `}`

ntzm: I tried that! See the other test cases I added, but imagine this:

```php
new class (function () {}) {};
```

SpacePossum: Very nice catch!
i wonder if we've more places where this case was not considered

ntzm: I was wondering that, it's definitely worth having a look

src/Fixer/ClassNotation/FinalStaticAccessFixer.php
```diff
@@ -0,0 +1,153 @@
...
+ */
+final class FinalStaticAccessFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Converts `static` access to `self` access in final classes.',
```
keradus: I do believe we shall add an reasoning why it's good to do this.

Like, we do have `summary`, but we are lacking `description`.

I can provide description why we shall use `static` in final classes.
If we have final class - both, static and self will work the same way. Then, if for any reason one is opening that class for modification - by removing the `final` modifier from class declaration - he will forget to change `self` to `static` in all places...
For that, I'm also not a big fan of adding it to local PHP CS Fixer's config

-----------------

For this reasoning, I'm giving the review "request changes".
This does not mean "we won't merge the PR, ever", but "we have open topic that deserves to be addressed, don't merge before that".

SpacePossum: Changing from `self` to the looser `static` and forgetting it at some spots is not a concern I share, in other words I would use this rule. However, I agree, improving documentation is a good idea, so please do so :)

@ntzm please update the docs and resolve the conflicts

kubawerlos: What about making this configurable to choose what one want to achieve? `self` or `static`.

ntzm: Why would one want to use `static` in a final class?

SpacePossum: >For that, I'm also not a big fan of adding it to local PHP CS Fixer's config

I removed it, reverted the changes, rebased and resolved the conflicts on top of master and squashed to one commit

SpacePossum:
Thank you @ntzm.

PR #4002, opened 2018-09-25
NoSuperfluousPhpdocTagsFixer - Allow `mixed` in superfluous PHPDoc by configuration

MortalFlesh:
Hello,

I'm facing a problem with `NoSuperfluousPhpdocTagsFixer`.

## Problem
When it is used with phpstan strict rules,  which needs an explicit annotation type even if it is mixed, I must choose either of them - or ignore the other.

## Solution
So I've decided to make a `NoSuperfluousPhpdocTagsFixer` configurable, to allow a `mixed` annotation for these cases.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1087,6 +1087,11 @@ Choose from the list of available rules:
   Removes ``@param`` and ``@return`` tags that don't provide any useful
   information.
 
+  Configuration options:
+
+  - ``allow_mixed`` (``bool``): whether type ``mixed`` without description is allowed
+    (``true``) or considered superfluous (``false``); defaults to ``false``
+
 * **no_trailing_comma_in_list_call** [@Symfony]
 
   Remove trailing commas in list function calls.
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -15,6 +15,9 @@ namespace PhpCsFixer\Fixer\Phpdoc;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\DocBlock\Annotation;
 use PhpCsFixer\DocBlock\DocBlock;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\FixerDefinition\VersionSpecification;
@@ -25,7 +28,7 @@ use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
-final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer
+final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
      * {@inheritdoc}
@@ -132,6 +135,19 @@ class Foo {
 
             $tokens[$index] = new Token([T_DOC_COMMENT, $docBlock->getContent()]);
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('allow_mixed', 'Whether type `mixed` without description is allowed (`true`) or considered superfluous (`false`)'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
+                ->getOption(),
+        ]);
     }
 
     private function findDocumentedFunction(Tokens $tokens, $index)
@@ -255,7 +271,7 @@ class Foo {
         $annotationTypes = $this->toComparableNames($annotation->getTypes(), $symbolShortNames);
 
         if (['mixed'] === $annotationTypes && null === $info['type']) {
-            return true;
+            return !$this->configuration['allow_mixed'];
         }
 
         $actualTypes = null === $info['type'] ? [] : [$info['type']];
--- a/tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
+++ b/tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
@@ -602,4 +602,60 @@ class Foo {
             ],
         ];
     }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixWithAllowedMixedCases
+     */
+    public function testFixWithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed' => true]);
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixWithAllowedMixedCases()
+    {
+        $cases = $this->provideFixCases();
+        $cases['no typehint mixed'] = [
+            '<?php
+class Foo {
+    /**
+     * @param mixed $bar
+     *
+     * @return mixed
+     */
+    public function doFoo($bar) {}
+}',
+        ];
+
+        return $cases;
+    }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixPhp70Cases
+     * @requires PHP 7.0
+     */
+    public function testFixPhp70WithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed' => true]);
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixPhp71Cases
+     * @requires PHP 7.1
+     */
+    public function testFixPhp71WithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed' => true]);
+        $this->doTest($expected, $input);
+    }
 }


src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -296,4 +325,11 @@ function ($type) use ($symbolShortNames) {
 
         return $normalized;
     }
+
+    private function resolveConfiguration()
+    {
+        if (null !== $this->configuration['allow_mixed_superfluous_phpdoc'] && true === $this->configuration['allow_mixed_superfluous_phpdoc']) {
```
ntzm: The first condition here isn't necessary

julienfalque:
What does PHPStan complains about exactly? I don't see the point in enforcing a `mixed` type annotation. Maybe this is a bug or a missing feature in PHPStan rather than PHP CS Fixer.

ntzm:
I think PHPStan sees a difference between a parameter that has no type (implicitly mixed) and an explicitly mixed parameter

dmvdbrugge:
I was wondering why this felt so familiar. There is another open PR that wants sort of the same: #3936

I must say the current PR is better, but the feature itself is still questionable.

README.rst
```diff
@@ -1087,6 +1087,11 @@ Choose from the list of available rules:
   Removes ``@param`` and ``@return`` tags that don't provide any useful
   information.
 
+  Configuration options:
+
+  - ``allow_mixed_superfluous_phpdoc`` (``bool``): whether to consider a mixed type
```
dmvdbrugge: The description is the wrong way around: the option says "allow" with default `false` (thus removing) but the description suggests that on default it is NOT considered superfluous (thus keeping).

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -25,8 +28,21 @@
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
-final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer
+final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    /** @var bool */
+    private $shouldFixMixed = true;
```
dmvdbrugge: If you call this `$allowMixed` with default `false`, then you could just do `$this->allowMixed = $this->configuration['allow_mixed_superfluous_phpdoc'];` while resolving the configuration. Only thing then to do is to negate the return in `annotationIsSuperfluous` and then you no longer have a sort of double negation, plus consistency, because currently everything is talking about "allow" in the config and test, but not in the actual code.

OndraM:
The general idea of PHPStan strict rules is to by type strict and type safe, so it forces you to *always* declare a type of a parameter or return type:

> Require parameter and return typehints for functions and methods (either native or phpDocs)

https://github.com/phpstan/phpstan-strict-rules/blob/master/README.md

Because of this, PHP's implicit "mixed" (= undeclared) type is considered as unsafe (meaning "you forgot to declare type!") and causes a warning. But if you in some scenario actually want to use "mixed" as a typehint, you must declare it explicitly (using PHPDoc), to make it obvious you didn't forgot to declare the type and the mixed type is intentional. But in this case this fixer will attempt to remove the phpdoc, because mixed actually is "superfluous" from the PHP point of view (but not from "I want to be sure I didn't forget to declare types" person point of view).

Also note there is [RFC](https://wiki.php.net/rfc/mixed-typehint) targeted for PHP 7.4/8.0, which may add mixed typehint. But until then, there is no other way than use PHPDoc to tell the mixed is intentional.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
MortalFlesh: yea.. great point :+1:

README.rst
MortalFlesh: I've tried to make it better - if you come up with even better description, I'd like to update it :)

dmvdbrugge: I would rename the option to `allow_mixed` (just like the property in the fixer), with description:
```
whether type `mixed` without description is allowed (`true`) or considered superfluous (`false`)
```

dmvdbrugge:
Ha, you're not the only one, join the waiting list 😆

In the meantime, I had already left an additional comment which you neither responded nor reacted to.

MortalFlesh:
@dmvdbrugge I missed that, the comment was hidden by github as `outdated` 🙃
I've just send a fix - your wording is better :+1: 

I can wait longer ofc, I was just curious if it will continue somehow..

ntzm:
Be patient, it will be merged :)

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -593,4 +593,96 @@ public function doFoo(iterable $bar, ?int $baz): ?array {}
...
+
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixWithAllowedMixedCases
+     */
+    public function testFixWithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed_superfluous_phpdoc' => true]);
```
dmvdbrugge: You renamed the config option, but forgot to fix your tests.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -593,4 +593,96 @@ public function doFoo(iterable $bar, ?int $baz): ?array {}
...
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixPhp70WithAllowedMixedCases
+     * @requires PHP 7.0
+     */
+    public function testFixPhp70WithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed_superfluous_phpdoc' => true]);
```
dmvdbrugge: You renamed the config option, but forgot to fix your tests.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -593,4 +593,96 @@ public function doFoo(iterable $bar, ?int $baz): ?array {}
...
+    /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixPhp71Cases
+     * @requires PHP 7.1
+     */
+    public function testFixPhp71WithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed_superfluous_phpdoc' => true]);
```
dmvdbrugge: You renamed the config option, but forgot to fix your tests.

MortalFlesh:
@dmvdbrugge it is fixed. I was sloppy and wanted to just push it right away. 

Btw should I rebase it against master now? Or wait till it will be accepted and you will rebase it before merge to master?

dmvdbrugge:
You can squash / rebase / do whatever you like until told otherwise 😁

In the end, the merger will squash all anyway, but the best way to guarantee a smooth process is if the branch is up-to-date with base.

What I'd do is rebase now. (But that's just me.)

julienfalque:
You can rebase your branch to make it up-to-date your target branch but please don't squash commits, mergers will take care of that. In the meantime, keeping separate commits help reviewers understand successive changes you make without reading the entire diff each time.

keradus:
@ntzm , I see your review as requesting changes. Can you re-review ?

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -296,4 +325,9 @@ function ($type) use ($symbolShortNames) {
 
         return $normalized;
     }
+
+    private function resolveConfiguration()
+    {
+        $this->allowMixed = $this->configuration['allow_mixed'];
```
SpacePossum: Hi and thanks for the PR :)

Could you move this one-liner in-line in the `configure`-method? than rebase one more time, after that I think it ready (but not 100% my call) to merge

MortalFlesh: Hi, it is fixed and rebased against current master.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -25,8 +28,21 @@
...
+    private $allowMixed = false;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->allowMixed = $this->configuration['allow_mixed'];
```
julienfalque: IMO there is no need for an extra `$allowMixed` property as it simply duplicates `$this->configuration['allow_mixed']`. I would drop it.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -602,4 +602,96 @@ public function doFoo(iterable $bar, ?int $baz): ?array {}
...
+    public function testFixPhp70WithAllowedMixed($expected, $input = null)
+    {
+        $this->fixer->configure(['allow_mixed' => true]);
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixPhp70WithAllowedMixedCases()
+    {
+        $cases = $this->provideFixPhp70Cases();
+        $cases['no typehint mixed'] = [
```
julienfalque: This test is the same as the one added above which has no PHP 7.0 specific syntax. You can drop this one. Same for the PHP 7.1 test below.

MortalFlesh: Wel.. it uses the original provider and just upsert a `no typehint mixed` case, so all previous cases must pass as well - meaning, I didn't break anything by this option..

Should I still drop these tests?

julienfalque: I think that for PHP 7.0 and 7.1 tests using `allow_mixed => true`, the extra `no typehint mixed` case isn't necessary (already tested in `testFixWithAllowedMixed()`) so you can drop the specific data providers and use the existing ones directly.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -25,8 +28,16 @@
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
-final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer
+final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
```
julienfalque: Overriding this method isn't necessary anymore.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
MortalFlesh: Oh you meant _drop providers_ not drop tests.. I makes sense now :+1:

PR #4007, opened 2018-09-27
Fixes cookbook for fixers

greeflas:
- Updated file header and array syntax
- Fixed code style in code samples

### DIFF
--- a/doc/COOKBOOK-FIXERS.md
+++ b/doc/COOKBOOK-FIXERS.md
@@ -49,9 +49,10 @@ Put this content inside:
 <?php
 
 /*
- * This file is part of the PHP CS utility.
+ * This file is part of PHP CS Fixer.
  *
  * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
  *
  * This source file is subject to the MIT license that is bundled
  * with this source code in the file LICENSE.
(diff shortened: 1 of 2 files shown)


ntzm:
See #3118

greeflas:
@ntzm but what about other fixes? :)

ntzm:
What other fixes? They are all code style related

greeflas:
@ntzm PR #3118 has only indentation fixes. This PR is updates array syntax, file header and adds additional PHPDoc blocks.

greeflas:
@dmvdbrugge ok. 
1. For me as for developer of fixers would be good to see updated doc with updated syntax.
2. Fixer will not add additional PHPDoc that was added in this PR.

SpacePossum:
I know the code should be fixed after running the tool, but as a reader being told to do on thing and than have it changed directly after by a tool feels odd.

If was only on sample-code that is to be fix later by the rule of the cookbook than I get it, but having the skeleton code in shape is still up to date is still a good thing IMHO.

doc/COOKBOOK-FIXERS.md
```diff
@@ -113,10 +115,15 @@ use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
  * @author Your name <your@email.com>
  *
  * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\Comment\RemoveCommentsFixer
```
keradus: :+1:

keradus:
Thank you @greeflas.

Issue #3989, opened 2018-09-03
Don't force multiline anonymous classes


nicolas-grekas:
I should be able to create an anonymous class on a single line (only when no body is defined):
```
$foo = new class() extends bar {};
```

Right now, php-cs-fixer enforces a line between the braces:
```
$foo = new class() extends bar {
};
```

keradus:
Can you raise a PR, @nicolas-grekas ?

nicolas-grekas:
Nope sorry I need help also.

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4012 by @kubawerlos :+1:

Issue #2782, opened 2017-05-18
class definitions in one line


bwoebi:
Would it be possible to set some option which retains one liner class defs like:

```
class FooException extends Foo {} // or maybe { }
```

… currently the braces fixer (I don't really want to disable it…) transforms that into:

```
class FooException extends Foo {
}
```

which is a totally unnecessary newline. [And is driving me nuts mhm…]

(If requested I may also provide a PR.)

SpacePossum:
The current configuration options of the `braces` fixer do not provide what you want.
If you want you can add a new option through a PR.
You can follow something like the `allow_single_line_closure` option as an example how to add the option, and later how to provide sample code and documentation for it. 
As a new option I would suggest calling it something like `allow_single_line_anonymous_class`

bwoebi:
I've tried to implement it, but when I've fixed it on the braces fixer, the ClassDefinitionFixer reverted it.
Now I tried to fix it on the ClassDefinition fixer, and the braces fixer reverted it.

I somehow needed to add an option for both of them. Not sure what I can do to just need a single option here?

Concretely, the code I used for both:

```diff
diff --git a/src/Fixer/Basic/BracesFixer.php b/src/Fixer/Basic/BracesFixer.php
index f00ea6b8..3eef382f 100644
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -160,6 +160,10 @@ protected function createConfigurationDefinition()
                 ->setAllowedTypes(['bool'])
                 ->setDefault(false)
                 ->getOption(),
+            (new FixerOptionBuilder('allow_single_line_class', 'Whether single line class notation should be allowed.'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
+                ->getOption(),
             (new FixerOptionBuilder('position_after_functions_and_oop_constructs', 'whether the opening brace should be placed on "next" or "same" line after classy constructs (non-anonymous classes, interfaces, traits, methods and non-lambda functions).'))
                 ->setAllowedValues([self::LINE_NEXT, self::LINE_SAME])
                 ->setDefault(self::LINE_NEXT)
@@ -297,9 +301,14 @@ function ($item) {
             }
 
             if (
-                $this->configuration['allow_single_line_closure']
-                && $token->isGivenKind(T_FUNCTION)
-                && $tokensAnalyzer->isLambda($index)
+                (
[...]

PR #4012, opened 2018-10-02
BracesFixer - add "allow_single_line_anonymous_class_with_empty_body" option

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3989.

Regarding https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2782 - I cannot see how `ClassDefinitionFixer` is affecting braces.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -329,6 +329,9 @@ Choose from the list of available rules:
 
   Configuration options:
 
+  - ``allow_single_line_anonymous_class_with_empty_body`` (``bool``): whether single
+    line anonymous class with empty body notation should be allowed;
+    defaults to ``false``
   - ``allow_single_line_closure`` (``bool``): whether single line lambda notation
     should be allowed; defaults to ``false``
   - ``position_after_anonymous_constructs`` (``'next'``, ``'same'``): whether the
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -159,6 +159,10 @@ class Foo
     protected function createConfigurationDefinition()
     {
         return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('allow_single_line_anonymous_class_with_empty_body', 'Whether single line anonymous class with empty body notation should be allowed.'))
+                ->setAllowedTypes(['bool'])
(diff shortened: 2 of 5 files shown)


SpacePossum:
Thanks for picking this up!
Wondering, would a more generic `allow_single_line_empty_body` make sense at some point?
Like `while(++$a <1){}` , `$a = function(){};` and `$a = new class () implements C {};` ?

dmvdbrugge:
Don't forget the `class CustomException extends \Exception {}` 😜

In other words, I agree with
> Wondering, would a more generic allow_single_line_empty_body make sense at some point?

kubawerlos:
@SpacePossum @dmvdbrugge I'm not sure about the change - take a look at failing tests. Making this option more generic (e.g. allowing `class Foo {}`) might not be what @nicolas-grekas meant when requesting https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3989.

kubawerlos:
ping @SpacePossum @dmvdbrugge

kubawerlos:
@dmvdbrugge also could be comma with `getNextNonWhitespace`, possibly something else as well.

The question here is should we have generic option to allow all empty bodies, one to allow only anonymous classes or maybe both? @nicolas-grekas?

SpacePossum:
@kubawerlos can you update the PR to allow only single line anonymous classes? But still allow for something like;
`new class() {} // test`  and `new class() { /* comment */ } // test` 

I think that is the initial request and we can always allow other constructs later as well

mvorisek:
@kubawerlos Please see https://groups.google.com/forum/?fromgroups#!topic/php-fig/VWcw3NnNHnM

The `BracesFixer` class is already huge and as few as possible options should exist.

In the discussion above I proposed: Add a blank line before the closing bracket if and only if a new line is before the opening bracket, there is no content inside the block (code nor comment) and the closing bracket is the last part of that expression (except semicolon).

SpacePossum:
@kubawerlos please rebase to get the conflict out and this is good to go, thanks for the work! :+1:

SpacePossum:
Thank you @kubawerlos.

PR #4017, opened 2018-10-07
Suppress warnings emitted by token_get_all

petk:
Hello, when `token_get_all()` encounters the octal sequence greater than `\377` it emits a warning.

See: https://wiki.php.net/rfc/octal.overload-checking

For example:

```php
<?php

var_dump("\542");
```
Emits this warning when running php-cs-fixer on such file:
```
Warning: Octal escape sequence overflow \542 is greater than \377
```

Thanks...

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -1030,8 +1030,8 @@ class Tokens extends \SplFixedArray
         $this->setSize(0);
 
         $tokens = \defined('TOKEN_PARSE')
-            ? token_get_all($code, TOKEN_PARSE)
-            : token_get_all($code);
+            ? @token_get_all($code, TOKEN_PARSE)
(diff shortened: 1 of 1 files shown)


petk:
I currently have no idea of anything else. So for now, I've opened an issue #4018 to not forget about it... It's an edge case, probably even token_get_all bug itself ... I'll check. If someone comes with a better patch sooner, please do.

SpacePossum:
This is real odd one because of the way it is implemented with PHP itself, nice catch :)
Not sure how to deal with for now, thanks for opening the issue as well!

ntzm:
We could `set_error_handler` with `E_WARNING`, with a closure that explicitly ignores that error, then restore after?

petk:
Indeed. Let me check if I can add a better pull request here. Yes, testing this... I know something better should be done :/

PR #4019, opened 2018-10-07
PhpdocTypesFixer - allow for configuration

keradus:
eg, `callback` alias type was in charge before 5.4, later it's not used.
eg `boolean`, usually may mean `bool` primitive, but sometimes it is a real custom class...

while for most cases it shall be fixed, there are few cases when not. let's make this configurable

ref https://github.com/sebastianbergmann/phpunit/pull/3327


keradus:
cc @localheinz, cc @sebastianbergmann

SpacePossum:
Thanks @keradus.

Issue #3843, opened 2018-06-28
NoUnsetOnPropertyFixer breaks on function call as key in array class-property


dmvdbrugge:
#### The PHP version you are using (`$ php -v`):
```
PHP 7.1.18 (cli) (built: May 25 2018 19:18:59) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.18, Copyright (c) 1999-2018, by Zend Technologies
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
`PHP CS Fixer 2.12.1 Long Journey by Fabien Potencier and Dariusz Ruminski`

#### The command you use to run PHP CS Fixer:
`vendor/bin/php-cs-fixer fix --rules=no_unset_on_property filename.php`

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
unset($this->profile->getSettings()[Settings::RANDOM_SETTING]);
```

 * with unexpected changes applied when running PHP CS Fixer:
Linting error after fixing

 * with the changes you expected instead:
None, it shouldn't have changed it.

The `getSettings()` call returns a magic subclass of `ArrayObject`, however the fixer thinks it's a property. I know it's ugly syntax but it's used all over the place because `$this->profile->settings` was changed to protected, and all occurrences were replaced with the getter (because lazy loading).

keradus:
@SpacePossum , can you take a look ?

SpacePossum:
I might but let me ping the author of the fixer as well @BackEndTea ;)

BackEndTea:
I'll take a look either later today or tomorrow

BackEndTea:
I created a test case for 
```php
unset($this->profile->getSettings()[Settings::RANDOM_SETTING]);
```
But it does not get changed by the fixer. Is there any other place in the file that uses an unset which might cause thsi bug?

dmvdbrugge:
Before reporting the issue I tried disabling this fixer, and there is no problem. Re-enabling the fixer and commenting out this line, and again no problem.

It might conflict with other fixers or its surrounding code, which I will be able to test again tomorrow.

dmvdbrugge:
So yeah... different line now, don't know if inconsistent or just my bad 😨
[...]

PR #4020, opened 2018-10-08
NoUnsetOnPropertyFixer - fix for function call with class constant inside

kubawerlos:


src/Fixer/LanguageConstruct/NoUnsetOnPropertyFixer.php
```diff
@@ -197,7 +197,22 @@ private function createTokensSubSet(Tokens $tokens, $startIndex, $endIndex)
      */
     private function doTokensInvolveProperty(Tokens $tokens)
     {
-        return $tokens->isAnyTokenKindsFound([T_PAAMAYIM_NEKUDOTAYIM, T_OBJECT_OPERATOR]);
+        for ($index = 0; $index < $tokens->count(); ++$index) {
+            if ('(' === $tokens[$index]->getContent()) {
+                $index = $tokens->findBlockEnd(
+                Tokens::BLOCK_TYPE_PARENTHESIS_BRACE,
```
SpacePossum: indenting is bit off here I think

SpacePossum:
looking good, please fix the CS minor, Travis failure is not related (existing test is to slow)

dmvdbrugge:
Does this fix #3843 too? (Would be awesome 😉)

kubawerlos:
Unfortunately no, I see a good test case there so will try to solve it as well.

dmvdbrugge:
I guess we should mention here that this is being superseeded by #4022? 😉

Issue #3511, opened 2018-02-06
Fixer for line breaks around operators


EvgenyOrekhov:
Similar to <https://eslint.org/docs/rules/operator-linebreak>.

Slamdunk:
Totally 👍 for this.

Developers who put operator at the end of line should be punished: either they love short meaningless variable names or scroll horizontally to understand the code :rage4:

Slamdunk:
@SpacePossum whatsapp? You agree, don't you?

kubawerlos:
> Developers who put operator at the end of line should be punished

With a year ban for programming, at least.

SpacePossum:
I don't agree on developers being punished for not following a CSS-style. I wouldn't down vote a fixer for this, however there are a lot of FR's opened lately without the intent to develop those it seems.

peter-gribanov:
> Developers who put operator at the end of line should be punished: either they love short meaningless variable names or scroll horizontally to understand the code :rage4:

In my opinion, exactly the opposite. Placing statements at the beginning of a line shifts the expression code relative to the first line and complicates reading the entire condition code. Using operators at the end of a line is a popular, well-established approach used in many languages.

See example [link](https://eslint.org/docs/rules/operator-linebreak) on ESLint from topic starter. The operator at the beginning is an exception, not a rule.

> Some developers find that placing operators at the beginning of the line makes the code more readable.

This is normal.


![image](https://user-images.githubusercontent.com/1954436/104935937-ea074200-59bc-11eb-89f4-c8dbe6d3a396.png)

This is controversial, but acceptable.


![image](https://user-images.githubusercontent.com/1954436/104935852-d0fe9100-59bc-11eb-8384-85addfca20f4.png)

This style makes reading difficult because of the offset of the condition code.


![image](https://user-images.githubusercontent.com/1954436/104936137-2aff5680-59bd-11eb-98ed-d0f7f50b826c.png)

kubawerlos:
> This is normal.

> This is controversial, but acceptable.

> This style makes reading difficult because of the offset of the condition code.

First of all, those are your personal opinions. Someone can just switch the positions in your examples keeping the description.

As in the proverb: one rabbi will say “beginning”; another rabbi will say “end”.

There is no good or wrong style, just a matter of preference.

----

For me, the`beginning` version makes having code like this:
```
var foo = arg1
        + arg2
        - arg3
        + arg4;
```
more readable than:

```
[...]

PR #4021, opened 2018-10-08
OperatorLinebreakFixer - Introduction

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3511

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1367,6 +1367,18 @@ Choose from the list of available rules:
   There should not be space before or after object ``T_OBJECT_OPERATOR``
   ``->``.
 
+* **operator_linebreak**
+
+  Operators - when multiline - must always be at the beginning or at the
+  end of the line.
+
+  Configuration options:
+
+  - ``only_booleans`` (``bool``): whether to limit operators to only boolean ones;
+    defaults to ``false``
+  - ``position`` (``'beginning'``, ``'end'``): whether to place operators at the
+    beginning or at the end of the line; defaults to ``'beginning'``
+
 * **ordered_class_elements** [@PhpCsFixer]
 
   Orders the elements of classes/interfaces/traits.
--- a/README.rst
+++ b/README.rst
@@ -1367,7 +1367,7 @@ Choose from the list of available rules:
   There should not be space before or after object ``T_OBJECT_OPERATOR``
   ``->``.
 
-* **operator_linebreak**
+* **operator_linebreak** [@PhpCsFixer]
 
   Operators - when multiline - must always be at the beginning or at the
   end of the line.
--- a/src/Fixer/Operator/OperatorLinebreakFixer.php
+++ b/src/Fixer/Operator/OperatorLinebreakFixer.php
@@ -137,6 +137,10 @@ function foo() {
                 continue;
             }
 
+            if ($this->belongsToGoToLabel($tokens, $index)) {
+                continue;
+            }
+
             if ($referenceAnalyzer->isReference($tokens, $index)) {
                 continue;
             }
@@ -310,5 +314,27 @@ function foo() {
         }
 
         return false;
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return bool
+     */
+    private function belongsToGoToLabel(Tokens $tokens, $index)
+    {
+        if (!$tokens[$index]->equals(':')) {
+            return false;
+        }
+
+        $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($index);
+
+        if (!$tokens[$prevMeaningfulTokenIndex]->isGivenKind(T_STRING)) {
+            return false;
+        }
+
+        $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($prevMeaningfulTokenIndex);
+
+        return $tokens[$prevMeaningfulTokenIndex]->equalsAny([';', '{', '}', [T_OPEN_TAG]]);
     }
 }
--- a/src/Fixer/Operator/OperatorLinebreakFixer.php
+++ b/src/Fixer/Operator/OperatorLinebreakFixer.php
@@ -1,0 +1,314 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Operator;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\CaseAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\ReferenceAnalyzer;
+use PhpCsFixer\Tokenizer\Analyzer\SwitchAnalyzer;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ */
+final class OperatorLinebreakFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @internal
+     */
+    const BOOLEAN_OPERATORS = [[T_BOOLEAN_AND], [T_BOOLEAN_OR], [T_LOGICAL_AND], [T_LOGICAL_OR], [T_LOGICAL_XOR]];
+
+    /**
+     * @internal
+     */
+    const NON_BOOLEAN_OPERATORS = ['%', '&', '*', '+', '-', '.', '/', ':', '<', '=', '>', '?', '^', '|', [T_AND_EQUAL], [T_CONCAT_EQUAL], [T_DIV_EQUAL], [T_DOUBLE_ARROW], [T_IS_EQUAL], [T_IS_GREATER_OR_EQUAL], [T_IS_IDENTICAL], [T_IS_NOT_EQUAL], [T_IS_NOT_IDENTICAL], [T_IS_SMALLER_OR_EQUAL], [T_MINUS_EQUAL], [T_MOD_EQUAL], [T_MUL_EQUAL], [T_OBJECT_OPERATOR], [T_OR_EQUAL], [T_PAAMAYIM_NEKUDOTAYIM], [T_PLUS_EQUAL], [T_POW], [T_POW_EQUAL], [T_SL], [T_SL_EQUAL], [T_SR], [T_SR_EQUAL], [T_XOR_EQUAL]];
+
+    /**
+     * @var string
+     */
+    private $position = 'beginning';
+
+    /**
+     * @var array<array<int|string>|string>
+     */
+    private $operators = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Operators - when multiline - must always be at the beginning or at the end of the line.',
+            [
+                new CodeSample('<?php
+function foo() {
+    return $bar ||
+        $baz;
+}
+'),
+                new CodeSample(
+                    '<?php
+function foo() {
+    return $bar
+        || $baz;
+}
+',
+                    ['position' => 'end']
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('only_booleans', 'whether to limit operators to only boolean ones'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
+                ->getOption(),
+            (new FixerOptionBuilder('position', 'whether to place operators at the beginning or at the end of the line'))
+                ->setAllowedValues(['beginning', 'end'])
+                ->setDefault($this->position)
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->operators = self::BOOLEAN_OPERATORS;
+        if (!$this->configuration['only_booleans']) {
+            $this->operators = array_merge($this->operators, self::NON_BOOLEAN_OPERATORS);
+            if (\PHP_VERSION_ID >= 70000) {
+                $this->operators[] = [T_COALESCE];
+                $this->operators[] = [T_SPACESHIP];
+            }
+        }
+        $this->position = $this->configuration['position'];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $referenceAnalyzer = new ReferenceAnalyzer();
+
+        $excludedIndices = $this->getExcludedIndices($tokens);
+
+        $index = $tokens->count();
+        while ($index > 1) {
+            --$index;
+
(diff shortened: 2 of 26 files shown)


src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,229 @@
...
+        '+=',
+        '<<',
+        '<<=',
+        '>>',
+        '>>=',
+        '^=',
+        '**',
+        '**=',
+        '<=>',
+        '??',
```
dmvdbrugge: Shouldn't `?`, `:`, and `?:` be here as well, if you include `??`?

README.rst
```diff
@@ -1186,6 +1186,18 @@ Choose from the list of available rules:
   There should not be space before or after object ``T_OBJECT_OPERATOR``
   ``->``.
 
+* **operator_linebreak**
+
+  Binary operators must always be at the beginning or at the end of the
+  line.
```
dmvdbrugge: I think this misses "when multiline". Might be obvious to most, but I had a "what?" moment (the "ah hah" came soon after 😉)

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,229 @@
...
+
+    /** @var string[] */
+    private $operators;
+
+    /** @var string */
+    private $position = 'beginning';
+
+    public function __construct()
+    {
+        $this->operators = \array_merge(self::BOOLEAN_OPERATORS, self::NON_BOOLEAN_OPERATORS);
```
dmvdbrugge: Isn't it kinda weird to always do an `array_merge`, and then optionally override it with the unmerged if config says so? I'd expect it to be the other way around for performance reasons (aka only do the merging when configured to do so).

I know it's the default config, but that shouldn't matter.

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,229 @@
...
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        if (isset($configuration['only_booleans']) && $configuration['only_booleans']) {
+            $this->operators = self::BOOLEAN_OPERATORS;
+        }
+        $this->position = isset($configuration['position']) ? $configuration['position'] : $this->position;
```
dmvdbrugge: What's the use of setting the property's default if you're overriding it here anyway, optionally with its own value?

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,229 @@
...
+            $this->fixMoveToTheEnd($tokens);
+        }
+    }
+
+    private function fixMoveToTheBeginning(Tokens $tokens)
+    {
+        for ($index = 0; $index < $tokens->count(); ++$index) {
+            $tokenContent = \strtolower($tokens[$index]->getContent());
+
+            if (!\in_array($tokenContent, $this->operators, true)) {
```
dmvdbrugge: So uhm, isn't the runtime of this one fixer going to be enormous, doing an `in_array` for each and every token? I propose either making the operators the keys, or doing an `array_flip` to achive the same, and then here do either `array_key_exists` or `isset`.

This obviously holds for `fixMoveToTheEnd` too.

tests/Fixer/Operator/OperatorLinebreakFixerTest.php
```diff
@@ -0,0 +1,207 @@
...
+            '<?php
+function foo() {
+    return isThisTheRealLife() || // First comment
+        // Second comment
+        // Third comment
+        isThisJustFantasy();
+}
+',
+        ];
+    }
```
dmvdbrugge: I'd like to see some nested testcases, you know, something like

```
return $a
    && (
        $b
        || $c
    )
    && $d;
```

(And then some variations, with comments too.)

src/Fixer/Operator/OperatorLinebreakFixer.php
dmvdbrugge: Also, it has a default value, doesn't that mean that it's always set anyway?

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,278 @@
...
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isRisky()
```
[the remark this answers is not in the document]
kubawerlos: None :(

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return null|int[]
+     */
+    private function getOperatorIndices(Tokens $tokens, $index)
+    {
+        if (null === $this->operators) {
```
SpacePossum: can `$this->operators` be resolved in the `configure` method?

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,262 @@
...
+     * @param Tokens $tokens
+     * @param int[]  $indices
+     * @param int    $direction
+     *
+     * @return Token[]
+     */
+    private function getReplacementsAndClear(Tokens $tokens, array $indices, $direction)
+    {
+        return array_map(
+            function ($index) use ($tokens, $direction) {
```
SpacePossum: this can be `static function` I think? no need to leak the scope in here

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,262 @@
...
+            $this->position = $configuration['position'];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return true;
```
SpacePossum: if `only_booleans` is configured I think a token-of-type-of-interest-is-found might be justified

src/Fixer/Operator/OperatorLinebreakFixer.php
kubawerlos: Done

src/Fixer/Operator/OperatorLinebreakFixer.php
kubawerlos: I thought we have `static_lambda` set up for the repo.

src/Fixer/Operator/OperatorLinebreakFixer.php
kubawerlos: Done

GrahamCampbell:
Any news on this fixer?

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,247 @@
...
+    private $operators;
+
+    /**
+     * @return FixerDefinition
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Operators - when multiline - must always be at the beginning or at the end of the line.',
+            [new CodeSample('<?php
```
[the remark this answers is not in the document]
kubawerlos: Done

kubawerlos:
@dmvdbrugge @GrahamCampbell @julienfalque @SpacePossum I've reworked the fixers A LOT.

The previous version didn't handle `&` (which can be "bitwise and" or reference) or `:` (can be part of ternary operator or part of `switch`).

kubawerlos:
No activity for over half a year, so I guess no one is interested in having it here, and the fixer is already implemented here: https://github.com/kubawerlos/php-cs-fixer-custom-fixers#OperatorLinebreakFixer

julienfalque:
> I guess no one is interested in having it here

I am :)

kubawerlos:
> I am :)

@julienfalque then go go go with review :stuck_out_tongue_winking_eye:

tests/Tokenizer/Analyzer/Analysis/CaseAnalysisTest.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
```
GrahamCampbell, line 3: Needs to be PHP 5.6 compat. ;)

tests/Tokenizer/Analyzer/Analysis/SwitchAnalysisTest.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
```
GrahamCampbell, line 3: Same

tests/Tokenizer/Analyzer/ReferenceAnalyzerTest.php
```diff
@@ -0,0 +1,137 @@
+<?php
+
+declare(strict_types=1);
```
GrahamCampbell, line 3: Same

tests/Fixer/Operator/OperatorLinebreakFixerTest.php
```diff
@@ -0,0 +1,518 @@
...
+            '&=', '|=', '^=', '<<=', '>>=', // Bitwise assignment
+            '==', '===', '!=', '<>', '!==', '<', '>', '<=', '>=',  // Comparison
+            'and', 'or', 'xor', '&&', '||', // Logical
+            '.', '.=', // String
+            '->', // Object
+            '::', // Scope Resolution
+        ];
+
+        if (\PHP_VERSION_ID >= 70000) {
+            var_dump('aaaa');
```
julienfalque, line 478: Whoops ;)

tests/Tokenizer/Analyzer/Analysis/SwitchAnalysisTest.php
```diff
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
```
GrahamCampbell, line 3: Please rebase this PR. :)

kubawerlos: Done

mvorisek:
@kubawerlos can this be merged?

kubawerlos:
@mvorisek I guess now after getting `RTM` label it can :)

src/Fixer/Operator/OperatorLinebreakFixer.php
```diff
@@ -0,0 +1,314 @@
...
+{
+    /**
+     * @internal
+     */
+    const BOOLEAN_OPERATORS = [[T_BOOLEAN_AND], [T_BOOLEAN_OR], [T_LOGICAL_AND], [T_LOGICAL_OR], [T_LOGICAL_XOR]];
+
+    /**
+     * @internal
+     */
+    const NON_BOOLEAN_OPERATORS = ['%', '&', '*', '+', '-', '.', '/', ':', '<', '=', '>', '?', '^', '|', [T_AND_EQUAL], [T_CONCAT_EQUAL], [T_DIV_EQUAL], [T_DOUBLE_ARROW], [T_IS_EQUAL], [T_IS_GREATER_OR_EQUAL], [T_IS_IDENTICAL], [T_IS_NOT_EQUAL], [T_IS_NOT_IDENTICAL], [T_IS_SMALLER_OR_EQUAL], [T_MINUS_EQUAL], [T_MOD_EQUAL], [T_MUL_EQUAL], [T_OBJECT_OPERATOR], [T_OR_EQUAL], [T_PAAMAYIM_NEKUDOTAYIM], [T_PLUS_EQUAL], [T_POW], [T_POW_EQUAL], [T_SL], [T_SL_EQUAL], [T_SR], [T_SR_EQUAL], [T_XOR_EQUAL]];
```
mvorisek, line 41: is this complete list, what about `**`, `instanceof`, `<=>`?

kubawerlos: what about them? do you want `**` and `<=>` to be removed and `instanceof` to be added? ;)

That does not make a sense - `instanceof` is not operator according to [docs](https://www.php.net/manual/en/language.operators.php) - it would make sense to add it as it works like operator or wouldn't? Ping @SpacePossum and @julienfalque for an opinion.

mvorisek: I see, spaceship is on line 107 and `**` is probably covered by `T_POW`. As this list contains also operators like `::` (`T_PAAMAYIM_NEKUDOTAYIM`), then I think `instanceof` should be added.

at least for non-boolean operators, I would very like to see a support for removing the line break completely.

kubawerlos: Once again `::` is on the list of operators and `instanceof` - I think we can extend the list, but let's merge this with the list we have (all the operators from https://www.php.net/manual/en/language.operators.php) and then discuss new items separately.

> Also, at least for non-boolean operators, I would very like to see a support for removing the line break completely.

Good idea, with the `position` value `none` and similarly like above, let's do it in separate PR after this one is merged.

GrahamCampbell:
Should this rule be added to the php-cs-fixer ruleset?x

src/Fixer/Operator/OperatorLinebreakFixer.php
mvorisek: > Good idea, with the `position` value `none` and similarly like above, let's do it in separate PR after this one is merged.

I support merging this asap, but I think we would need a different config to support different `position` for different types. Not sure how much breaking changes are an issue in `master`, but I think the config should change to something like:
- `position_boolean_operator` =>`beginning`
- `position_other_operator` =>`none` (or maybe `remove`, also, this must support false)

with these values beiing default and drop `only_booleans` completely.

mvorisek:
> 
> 
> Should this rule be added to the php-cs-fixer ruleset?

I would like to see this in `@Symfony` ruleset with config described in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4021#discussion_r499048329

src/Fixer/Operator/OperatorLinebreakFixer.php
kubawerlos: Breaking changes are only for released stuff only - if this is merged we can change it as much as we can until it is released.

SpacePossum:
@kubawerlos please review my changes. The goto-label edge case I found when running on the symfony source folder. I've copy-pasted the fix from https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5113/files so maybe someone can add a todo for making a colon-analyzer or something like that.

@GrahamCampbell done

@mvorisek I'm hesitant as there are a lot of fixes on the symfony source, so probably even far more in the symfony ecosystem. I would be more confident to add it to their set if a core member would request it. Than we don't have to deal with disappointments after 2.17 release like we had in the past releases. Thanks for the suggestion though!

src/Fixer/Operator/OperatorLinebreakFixer.php
SpacePossum: Nice suggestion!
I think leaving out `instance of` is a good call, but happy to see real life examples where this could add value.
For the new options, we can adjust until released and even after that we can add new options and deprecate the current ones (less ideal surely). As such, maybe best to start a new issue/PR for expected changes and discuss there, thanks!

kubawerlos:
@SpacePossum looks good, are you sure we want to add it to `@PhpCsFixer` in this PR instead of applying it to `2.15` and then merge the changes up?

SpacePossum:
Yeah good point, I might run it on 2.15 later on, but as the fixer won't be part of it new cases might get in anyways.
I was wondering though, I've added the `boolean` one only, because we (and on SF even more) some hits on constructs like;

```php
$foo =
<<<TXT

TXT;
```

getting fixed to;

```php
$foo
    =
<<<TXT

TXT;
```

WDYT, is this good, or something we want to make configurable on an operator bases?

SpacePossum:
Thank you @kubawerlos.

peter-gribanov:
Why is the default value `beginning`?

kubawerlos:
@peter-gribanov what are possible answers? ;)

peter-gribanov:
> what are possible answers? ;)

In this project, it was customary to indicate operators at the end. I see no explanation as to why you changed the coding style for this project.

kubawerlos:
> why you changed the coding style for this project.

I did not.

peter-gribanov:
> I did not.

Sorry. It was @SpacePossum who [change CS](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4021/commits/00ddcf6f7dcc9b2a087e141fd200b84090f04aa4).

mvorisek:
Normalizing to beginning is definitely better, at least in my opinion. I would also this is defacto standard and I would expect at least 80% of the operators in the top 1000 packages to be at the beginning. Feel free to present hard data to discuss more :)

GrahamCampbell:
> Sorry. It was @SpacePossum who change CS.

It was not a change. Most of the files in this project were already using that style.

PR #4022, opened 2018-10-09
NoUnsetOnPropertyFixer - refactor and bugfixes

kubawerlos:
Replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4020
Fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3843

### DIFF
--- a/src/Fixer/LanguageConstruct/NoUnsetOnPropertyFixer.php
+++ b/src/Fixer/LanguageConstruct/NoUnsetOnPropertyFixer.php
@@ -15,7 +15,7 @@ namespace PhpCsFixer\Fixer\LanguageConstruct;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Tokenizer\CT;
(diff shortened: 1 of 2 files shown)


src/Fixer/LanguageConstruct/NoUnsetOnPropertyFixer.php
```diff
@@ -71,142 +71,161 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
      *
-     * @return Token[]
+     * @return bool
      */
-    private function createReplacementUnset(Tokens $filteredTokens)
+    private function isProperty(Tokens $tokens, $index, $endIndex)
     {
-        if ($filteredTokens[0]->isWhitespace()) {
-            $filteredTokens->clearAt(0);
+        while ($tokens[$index]->isWhitespace() || $tokens[$index]->isComment()) {
```
ntzm: Can we use `getNextMeaningfulToken` here instead?

kubawerlos: We can even remove it with updating parent call.

keradus:
Thank you @kubawerlos.

PR #4023, opened 2018-10-10
Enhancement: Update localheinz/composer-normalize

localheinz:
This PR

* [x] updates `localheinz/composer-normalize`

💁‍♂️ For reference, see https://github.com/localheinz/composer-normalize/compare/0.6.0...0.9.0.

### DIFF
--- a/dev-tools/composer.json
+++ b/dev-tools/composer.json
@@ -3,7 +3,7 @@
         "php": "^7.1"
     },
     "require-dev": {
-        "localheinz/composer-normalize": "^0.6.0",
+        "localheinz/composer-normalize": "^0.9.0",
         "maglnet/composer-require-checker": "^1.0",
         "mi-schi/phpmd-extension": "^4.2",
         "phpmd/phpmd": "^2.4.3"


keradus:
60 commits without changelog is hard to catch up

please, target 2.12 LTS

dev-tools/composer.json
```diff
@@ -3,7 +3,7 @@
         "php": "^7.1"
     },
     "require-dev": {
-        "localheinz/composer-normalize": "^0.6.0",
+        "localheinz/composer-normalize": "~0.9.0",
```
keradus: why not caret ?

localheinz:
@keradus 

Good point, should probably start keeping a changelog!

localheinz:
@keradus @SpacePossum 

Since this is admittedly the first time that I have written a changelog, perhaps you can take a look at https://github.com/localheinz/composer-normalize/pull/94 and let me know if this is good? 

I would appreciate it!

keradus:
Thank you @localheinz.

PR #4025, opened 2018-10-11
Add phpdoc_types_order rule to Symfony's ruleset

carusogabriel:
As suggested by @julienfalque in https://github.com/symfony/symfony/pull/28780, besides add this new rule due to https://github.com/symfony/symfony/pull/28675, we should also enforces it in Symfony's ruleset here too :blush:


SpacePossum:
thanks @carusogabriel  :+1:

PR #4027, opened 2018-10-11
PhpdocAnnotationWithoutDotFixer - add failing cases

keradus:
introduced in #3882

ref https://github.com/symfony/symfony/pull/28814
ref https://github.com/symfony/symfony/pull/28817

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocAnnotationWithoutDotFixer.php
@@ -78,8 +78,8 @@ function foo ($bar) {}
                 $content = $annotation->getContent();
 
                 if (
-                    1 !== Preg::match('/[.。]\h*(\R|$)/u', $content)
-                    || 0 !== Preg::match('/[.。](?!\h*(\R|$))/u', $content, $matches)
+                    1 !== Preg::match('/[.。]\h*$/u', $content)
+                    || 0 !== Preg::match('/[.。](?!\h*$)/u', $content, $matches)
                 ) {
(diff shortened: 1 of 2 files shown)


keradus:
@kubawerlos , can you take a look on this one, please ?

kubawerlos:
I see no simple solution for this - the actually broken PHPDoc (with missing `*`) handled in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3882 IMHO should not be handled in the fixer, but in `DocBlock`/`Tag` class - as all broken PHPDocs should, like the `@return` added in this PR (the content of tag - returned by `Tag` class - is the first line only, so here fixer behaves correctly).

What do you think about reverting #3882 for now, then updating `DocBlock`/`Tag` to allow missing `*` and then getting back to the cases with broken PHPDoc?

keradus:
Changing DocBLock\* has unpredictable consequences :(

kubawerlos:
We could give it a try... or assume that PHPDocs with gaps in asterisk are not handles as its have asterisk where missing => remove both test cases with missing asterisk and revert #3882 "fix".

keradus:
indeed, when asterisks are missing, phpdoc is malformed and let not modify it

keradus:
can you take a look, @kubawerlos ?

SpacePossum:
Thanks @keradus.

SpacePossum:
(not sure why GH doens't state this PR as `merged` but as `closed`)

keradus:
(as I see squashed commit went to target branch, but PR was not updated with squashed commit)

PR #4029, opened 2018-10-12
ExplicitStringVariableFixer - add failing test case with execution operators

kubawerlos:
What should be the fix? Converting variable or leaving it untouched? Because definitely not `RuntimeException: Index invalid or out of range`.

Ping @Slamdunk

### DIFF
--- a/tests/Fixer/StringNotation/ExplicitStringVariableFixerTest.php
+++ b/tests/Fixer/StringNotation/ExplicitStringVariableFixerTest.php
@@ -177,6 +177,9 @@ EOF;
                 '<?php $a = B"end {$a[1]}";',
                 '<?php $a = B"end $a[1]";',
             ],
+            [
+                '<?php $a = `echo $foo`;',
+            ],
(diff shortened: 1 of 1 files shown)


kubawerlos:
How come? For:
```php
<?php
$x = '"foo"';
var_dump(`echo $x`);
var_dump(`echo ${x}`);
var_dump(`echo {$x}`);
```

I'm getting:
```
string(4) "foo
"
string(4) "foo
"
string(4) "foo
"
```

Slamdunk:
Ah, the value of the variable must be shell escaped to work, didn't remember that!

So, backtick to me is a beast, I propose:

1. Leave backtick operator untouched
1. Run backtick to shell exec fixer with higher priority, and thus creating the appropriate integration tests
1. Point this out in this fixer description

kubawerlos:
Suppressed by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4036

PR #4030, opened 2018-10-13
Add get_required_files and user_error aliases

ntzm:


kubawerlos:
Although, shouldn't branch `2.12` be the target?

ntzm:
I think I see this as a new feature rather than a bug fix

dmvdbrugge:
> I think I see this as a new feature rather than a bug fix

In which case it should target master, and get milestone 2.14
It seems however, @SpacePossum disagrees with you

dmvdbrugge:
(And I submitted too fast...)

I think I agree with the feature not bug fix approach, also taking into account how some of my previous things were "classified".

SpacePossum:
it is not exact science,  I'm happy with 2.12, 2.13 and master here TBH

ntzm:
I think I will target `master` instead, forgot 2.13 was the stable

kubawerlos:
> SpacePossum added `feature` `RTM` labels
> SpacePossum added this to the `2.13.1` milestone

@SpacePossum both agrees and disagrees this to be feature :)

SpacePossum:
haha, true, what to say, I like this to be in that is all, I'll leave the details to the specialists :)

SpacePossum:
Thank you @ntzm.

PR #4031, opened 2018-10-13
Correct FixerOptionBuilder::getOption return type

ntzm:


SpacePossum:
shouldn't this be `FixerOptionInterface` only?

SpacePossum:
ok :)   I'm :+1:  for that one interface only

keradus:
Thank you @ntzm.

PR #4033, opened 2018-10-13
Raise deprecation warnings on usage of deprecated aliases

ntzm:
Forgot to do this one (sorry 😬), reference: #3568

However, I need a way to still have the aliases tested. I could add a `@expectedDeprecation` for every test that uses a deprecated alias, but I was wondering if there was an easier way of doing this, maybe in the `AbstractFixerWithAliasedOptionsTestCase::doTest`.

### DIFF
--- a/src/FixerConfiguration/FixerConfigurationResolver.php
+++ b/src/FixerConfiguration/FixerConfigurationResolver.php
@@ -63,10 +63,11 @@ final class FixerConfigurationResolver implements FixerConfigurationResolverInte
                 $alias = $option->getAlias();
 
                 if (\array_key_exists($alias, $options)) {
-                    // @TODO 2.12 Trigger a deprecation notice and add a test for it
                     if (\array_key_exists($name, $options)) {
                         throw new InvalidOptionsException(sprintf('Aliased option %s/%s is passed multiple times.', $name, $alias));
                     }
+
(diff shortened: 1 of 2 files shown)


src/FixerConfiguration/FixerConfigurationResolver.php
```diff
@@ -63,7 +63,8 @@ public function resolve(array $options)
                 $alias = $option->getAlias();
 
                 if (array_key_exists($alias, $options)) {
-                    // @TODO 2.12 Trigger a deprecation notice and add a test for it
+                    @trigger_error(sprintf('Option %s is deprecated, use %s instead.', $alias, $name), E_USER_DEPRECATED);
```
SpacePossum: please use double quotes, like;  `'Option "%s" is deprecated, use "%s" instead.'`

SpacePossum:
no idea how to nicely test those cases...

keradus:
@julienfalque can correct me if I', wrong, but deprecations shall be released under MINOR, not PATCH release

julienfalque:
> no idea how to nicely test those cases :(

@SpacePossum is there still something left to do here in your opinion?

SpacePossum:
note; rebased and squashed @ntzm

SpacePossum:
Thank you @ntzm.

PR #4035, opened 2018-10-13
DocBlock - handle removing of first or last line

kubawerlos:
For a PHPDoc:
```php
/** @foo
 * @bar
 */
```
removing annotation `foo` with `Annotation::remove` will also remove `/**` which will result with broken PHPDoc syntax, leading to PHP syntax error after fixing.

What should happen in such case (and similar with closing `*/`)? Should we aim ro changing `DocBlock` constructor to split into pieces differently?


dmvdbrugge:
I think the `remove` methods should take begin/end of block into account when performing their action, I feel like I've worked around this before, maybe even in an open PR (not sure).

It will not hinder removing an entire (empty) docblock anyway because that should just remove the `Token`.

SpacePossum:
I would be expect;

```php
/**
 * @bar
 */
```

as result.
And;

```php
/** @bar*/
```

for

```php
/** @foo @bar */
```

maybe we can add this test as well;

```php
/** @bar
 * @foo */
```

dmvdbrugge:
I don't think `/** @foo @bar */` valid syntax to begin with though. `@bar` is and should be considered part of the text of `@foo`.

SpacePossum:
alright, please lets have that case in the tests as well (even if untouched by the rule)

kubawerlos:
> I think the `remove` methods should take begin/end of block into account when performing their action

@dmvdbrugge neither `Assertion` not `Line` is aware if it contains the beginning of the PHPDoc (multiple `/**` inside of PHPDoc are valid syntax) so I went one level higher to `DocBlock` class itself.

After removing some lines (no matter if via `Assertion` or directly) the `getContent` function must return valid content.

dmvdbrugge:
> neither Assertion not Line is aware if it contains the beginning of the PHPDoc 

Sorry but you're wrong: line is aware. In fact, I remembered I already had a PR for this: #3893
It's been sitting there, quietly waiting review...

kubawerlos:
The `Line::isTheStart` may return `true` while not being start line e.g.
```php
<?php
/**
/**
 * Foo
 */
```

dmvdbrugge:
That however (if true) is a bug in `Line::isTheStart` that should be fixed. Also, I don't think that's a valid docblock. The GH-syntax-highlighting even breaks on it.

kubawerlos:
It is a valid DocBlock (from the PHP tokenizer perspective) and to fix it we would have to pass the information if it's the start in constructor - not sure if we would want to go this way.

kubawerlos:
After merging https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3893 tests for GeneralPhpdocAnnotationRemoveFixer are passing.

PR #4038, opened 2018-10-14
CommentToPhpdocFixer - handling nested PHPDoc

kubawerlos:


tests/Fixer/Comment/CommentToPhpdocFixerTest.php
```diff
@@ -233,6 +233,26 @@ public function provideTestCases()
...
+ * @var string $foo
+ */
+$foo = 1;
+EOT
+                ,
+                <<<'EOT'
+<?php /* header comment */ $foo = true;
+
+// This is my var
+// /** @var string $foo */
```
dmvdbrugge: This isn't "nested", this is "commented out". This input should not change imho.

kubawerlos: It cannot be declined - current behaviour is not ok, [test case](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4038/commits/9e6c6b9b249260cf0fce966c022f6aaf673f7f4c) is failing due to [breaking](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/441227326) PHP syntax.

I'm open how to handle this - we may leave such a comment without change, we can change `*/` to `*-/` or remove both `/**` and `*/`

dmvdbrugge: > current behaviour is not ok, test case is failing due to breaking PHP syntax.

ok yeah I may have sounded harsher than intended, this is indeed not ok.

> we may leave such a comment without change

I think this should be the way to go here, as per my already given reason. Correct me if I'm wrong but the aim of this fixer is to fix comments that were intended (or used) as doc block in the first place, whereas this case is not, right?

keradus: Good point that original comment was crashing the fixer, yet I believe we shall handle it differently

`// /** @var string $foo */`
is not intended to be documenting next structural element, as it's commented out.
I agree with @dmvdbrugge that handling shall be changed to not try to convert original comments into phpdoc comment, but instead we shall leave it unchanged.

src/Fixer/Comment/CommentToPhpdocFixer.php
```diff
@@ -155,18 +155,26 @@ private function fixCommentMultiLine(Tokens $tokens, array $indices)
...
         $indent = Utils::calculateTrailingWhitespaceIndent($tokens[$startIndex - 1]);
 
-        $content = '/**'.$this->whitespacesConfig->getLineEnding();
+        $newContent = '/**'.$this->whitespacesConfig->getLineEnding();
         $count = max($indices);
 
         for ($index = $startIndex; $index <= $count; ++$index) {
             if ($tokens[$index]->isComment()) {
-                $content .= $indent.' *'.$this->getMessage($tokens[$index]->getContent()).$this->whitespacesConfig->getLineEnding();
+                $newContent .= $indent.' *'.$this->getMessage($tokens[$index]->getContent()).$this->whitespacesConfig->getLineEnding();
```
SpacePossum: before concat'ing the new content the `if (false !== strpos($newContent, '*/')) {` check can be done I think? allowing for an early return

keradus:
Thank you @kubawerlos.

Issue #4026, opened 2018-10-11
Performance investigation


staabm:
while crunching together initial numbers in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4024#issuecomment-428994403 I did some blackfire profiles

my findings:

linting a single file:
`blackfire run php ./PHP-CS-Fixer/php-cs-fixer fix ./sabre-dav/lib/CalDAV/Calendar.php`

https://blackfire.io/profiles/739e8c50-2345-4f4b-a157-e94b9b220029/graph
![grafik](https://user-images.githubusercontent.com/120441/46814772-b7038c00-cd79-11e8-99f9-4bdb3f2dbad2.png)

=> PhpCsFixer\Tokenizer\Token::equals() and PhpCsFixer\Tokenizer\Token::isGivenKind() seem to be slow

linting a folder of files (pretty similar picture):
`blackfire run php ./PHP-CS-Fixer/php-cs-fixer fix ./sabre-dav/lib/CalDAV/`

https://blackfire.io/profiles/7db4aa88-ece8-4fd4-9850-a1a7467bf1a4/graph
![grafik](https://user-images.githubusercontent.com/120441/46815438-36de2600-cd7b-11e8-9caf-f7768e8165c4.png)

=> PhpCsFixer\Tokenizer\Token::equals(), PhpCsFixer\Tokenizer\Tokens::extractTokenKind(), PhpCsFixer\Tokenizer\Tokens::offsetSet() seem to be slow

staabm:
> can you elaborate on that ?
> what I can see is that they are called plenty of times, and well, they will, due to their responsibility

those methods are called so frequently that very minor changes can lead to walltime speed win. alternatively one could check the calling locations whether some of those calls can be saved, e.g. by using temp-variables or similar

keradus:
Great work @staabm !
Closing due to merge of #4046, #4047, #4048 .

-----------------

After saying that, if anyone has idea how to improve performance even further, PRs are welcome !

staabm:
Thx for merging. I am very proud to contribute to this great tool.

keradus:
Thanks for your patience as well ;)

ntzm:
Awesome work on this @staabm

PR #4039, opened 2018-10-14
Inlined extract-token-kind on the hot path

staabm:
as identified in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4026 we inline calls to `extractTokenKind()` which leads to a 5-7% speed increase.

most calls of this method happen thru the changed code as can be seen in the blackfire profile

image

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -1396,7 +1396,11 @@ class Tokens extends \SplFixedArray
      */
     private function registerFoundToken($token)
     {
-        $tokenKind = $this->extractTokenKind($token);
+        // inlined extractTokenKind() call on the hot path
+        $tokenKind = $token instanceof Token
+            ? ($token->isArray() ? $token->getId() : $token->getContent())
+            : (\is_array($token) ? $token[0] : $token)
+        ;
 
         if (!isset($this->foundTokenKinds[$tokenKind])) {
             $this->foundTokenKinds[$tokenKind] = 0;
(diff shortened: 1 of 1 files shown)


src/Tokenizer/Tokens.php
```diff
@@ -1396,7 +1396,11 @@ private function changeCodeHash($codeHash)
      */
     private function registerFoundToken($token)
     {
-        $tokenKind = $this->extractTokenKind($token);
+        // inlined extractTokenKind() call on the hot path
+        $tokenKind = $token instanceof Token
+            ? ($token->isArray() ? $token->getId() : $token->getContent())
+            : (\is_array($token) ? $token[0] : $token)
+        ;
```
keradus: on big fan of this.
DRY is violated. Tokenizer classes are heart of PHP CS Fixer, having code copy-pasted is not helping with a maintenance :(

SpacePossum:
Is the overhead created by the making the function call or can we get better performance by updating the method to only except `Token`?

This would make the `instanceof` check not needed, I think the calls in `unregisterFoundToken` and `registerFoundToken` can be optimized like this.

staabm:
The function is called millions of times, so yeah - mostly function call overhead

staabm:
> Is the overhead created by the making the function call or can we get better performance by updating the method to only except Token?
This would make the instanceof check not needed, I think the calls in unregisterFoundToken and registerFoundToken can be optimized like this.

Could be a further improvement for a separate PR

staabm:
re-submitted with new merge base

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4048

PR #4040, opened 2018-10-14
inline $other->getPrototype() to speedup equals()

staabm:
as measured in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4026 we try to speedup equals() by inlinening the $this->getPrototype() call.

also we use FQCN for native functions, because equals() is called a lot.

this speedsup the process by 4-5%.

see https://blackfire.io/profiles/compare/f4eefae8-6d09-4ab4-a833-f828d74a7b52/graph
image

### DIFF
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -156,7 +156,19 @@ class Token
      */
     public function equals($other, $caseSensitive = true)
     {
-        $otherPrototype = $other instanceof self ? $other->getPrototype() : $other;
+        if ($other instanceof self) {
+            // inlined getPrototype() on this very hot path
+            if (!$other->isArray) {
+                $otherPrototype = $other->content;
+            } else {
+                $otherPrototype = [
+                    $other->id,
+                    $other->content,
+                ];
+            }
+        } else {
+            $otherPrototype = $other;
+        }
 
         if ($this->isArray !== \is_array($otherPrototype)) {
             return false;


src/Tokenizer/Token.php
```diff
@@ -156,34 +156,44 @@ public function clearChanged()
...
             return false;
         }
 
         if (!$this->isArray) {
-            return $this->content === $otherPrototype;
+            return $this->content === $other;
         }
 
-        if (array_key_exists(0, $otherPrototype) && $this->id !== $otherPrototype[0]) {
+        if (\array_key_exists(0, $other) && $this->id !== $other[0]) {
```
keradus: please revert backslashing.
we rely on `native_function_invocation` rule, that's it

src/Tokenizer/Token.php
```diff
@@ -156,34 +156,44 @@ public function clearChanged()
      */
     public function equals($other, $caseSensitive = true)
     {
-        $otherPrototype = $other instanceof self ? $other->getPrototype() : $other;
+        if ($other instanceof self) {
```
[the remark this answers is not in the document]
keradus: well, we see a failing test...

src/Tokenizer/Token.php
staabm: this path is very very hot, and paying this additional namespace lookup here makes a difference.

ntzm: How much difference?

src/Tokenizer/Token.php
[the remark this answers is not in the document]
staabm: tests fixed

src/Tokenizer/Token.php
staabm: after https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4041 its no longer measurable.

PR #4042, opened 2018-10-14
Use dedicated assertRegExp assertions

carusogabriel:
Looks like this assertion isn't present in the `2.12` branch.


keradus:
2.12 is using same PHPUnit that master branch

PR #4043, opened 2018-10-14
NativeFunctionInvocationFixer - add option to remove redundant backslashes

kubawerlos:
If we add it to `@Symfony` rule set with `true` then in Symfony v4.1.6 26 files would be affected, and here - at master - 6 files.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -926,6 +926,8 @@ Choose from the list of available rules:
     defaults to ``['@internal']``
   - ``scope`` (``'all'``, ``'namespaced'``): only fix function calls that are made
     within a namespace or fix all; defaults to ``'all'``
+  - ``strict`` (``bool``): whether leading ``\`` of function call not meant to have it
+    should be removed; defaults to ``false``
 
 * **new_with_braces** [@Symfony, @PhpCsFixer]
 
--- a/src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
+++ b/src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
@@ -115,19 +115,19 @@ namespace {
     {
         parent::configure($configuration);
 
-        $uniqueConfiguredExclude = \array_unique($this->configuration['exclude']);
+        $uniqueConfiguredExclude = array_unique($this->configuration['exclude']);
 
         // Case sensitive constants handling
-        $constantsToEscape = \array_values($this->configuration['include']);
+        $constantsToEscape = array_values($this->configuration['include']);
         if (true === $this->configuration['fix_built_in']) {
-            $getDefinedConstants = \get_defined_constants(true);
+            $getDefinedConstants = get_defined_constants(true);
             unset($getDefinedConstants['user']);
             foreach ($getDefinedConstants as $constants) {
-                $constantsToEscape = \array_merge($constantsToEscape, \array_keys($constants));
+                $constantsToEscape = array_merge($constantsToEscape, array_keys($constants));
             }
         }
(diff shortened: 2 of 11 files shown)


README.rst
```diff
@@ -916,6 +916,8 @@ Choose from the list of available rules:
     defaults to ``['@internal']``
   - ``scope`` (``'all'``, ``'namespaced'``): only fix function calls that are made
     within a namespace or fix all; defaults to ``'all'``
+  - ``strict`` (``bool``): whether leading ``\`` from function class not meant to have
```
SpacePossum: >from function class

I think this should be `from function call` ?

keradus:
> If we add it to @Symfony rule set with true then in Symfony v4.1.6 26 files would be affected, and here - at master - 6 files.

@nicolas-grekas , WDYT ?

nicolas-grekas:
We could enable it, so that things have only one valid way to be written.

keradus:
:+1:

@kubawerlos , can you update ruleset definition, in that case ?

kubawerlos:
@keradus updated

@nicolas-grekas should I make PR to Symfony to remove extra `\` to avoid surprise when this is released? If so what branch to target?

nicolas-grekas:
no need, let's wait for the fabbot update and we'll do it in a batch starting from 2.8

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -115,44 +115,44 @@ public function configure(array $configuration = null)
...
                 unset($constantsToEscape[$constantIndex]);
             }
         }
 
-        $caseInsensitiveConstantsToEscape = \array_diff(
-            \array_unique($caseInsensitiveConstantsToEscape),
-            \array_map('strtolower', $uniqueConfiguredExclude)
+        $caseInsensitiveConstantsToEscape = array_diff(
+            array_unique($caseInsensitiveConstantsToEscape),
+            array_map('strtolower', $uniqueConfiguredExclude)
```
keradus: :(

SpacePossum:
@kubawerlos friendly heads up about the conflict here

kubawerlos:
@SpacePossum update, I had to remove it from Travis's "strict" job as it was conflicting with the `@Symfony` configuration. Why did we had it there anyway?

SpacePossum:
The test was added so that we will know that if we would make all function calls within the project native calls (were applicable) we would still be good, as we currently do not enforce this.

I think this is still a good idea though, but I see the trouble arise here, therefore I need some time to think about removing it as proposed in the PR.

kubawerlos:
@SpacePossum we can create separate step to apply these rules and run only tests or (more hacky) update the config with `native_function_invocation` to override the one from `@Symfony`

SpacePossum:
@kubawerlos  please add tests @ `NativeConstantInvocationFixerTest` to cover the new configuration option, than I think this PR is good to go :)

About the test, if I run the original fix command and than the unit tests the test passes. So I'm not sure where this would fail now?

SpacePossum:
Sorry for being slow; please ignore my first remark and about the test; I agree with your previous suggestion; 

>we can create separate step to apply these rules and run only tests

keradus:
I had drop `native_function_invocation` from migration again.
We had it to see if we run on optimized build.
Now, definition of `what` optimized means changed due to better understanding of `\` resolution, and default setting of it is not what we now consider as best practice. `native_function_invocation` we have in our config we use internally is enough for CI.

keradus:
Thank you @kubawerlos.

PR #4046, opened 2018-10-16
Token - Added fast isset() path to token->equals()

staabm:
looking up the existance of a property via isset() is waaaaay faster then array_key_exists().
we still need array_key_exists() as a fallback, when properties exist with a null-value to work properly. this case is very rare, therefore this can be considered a speedup.

this change speeds up php-cs-fixer by 5.6%.

image

### DIFF
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -166,11 +166,11 @@ class Token
             return $this->content === $otherPrototype;
         }
 
-        if (array_key_exists(0, $otherPrototype) && $this->id !== $otherPrototype[0]) {
+        if ($this->id !== $otherPrototype[0]) {
             return false;
         }
 
-        if (array_key_exists(1, $otherPrototype)) {
+        if (isset($otherPrototype[1])) {
             if ($caseSensitive) {
                 if ($this->content !== $otherPrototype[1]) {
                     return false;


src/Tokenizer/Token.php
```diff
@@ -166,11 +166,11 @@ public function equals($other, $caseSensitive = true)
             return $this->content === $otherPrototype;
         }
 
-        if (array_key_exists(0, $otherPrototype) && $this->id !== $otherPrototype[0]) {
+        if ((isset($otherPrototype[0]) || array_key_exists(0, $otherPrototype)) && $this->id !== $otherPrototype[0]) {
```
keradus: I'm wondering now... is there a case where index `0` does not exist ?

staabm: I dont know too, therefore I played it safe for BC reasons

keradus: If I would be sure then I would not say that "I'm wondering" :)

we do need check for index 1, as it's optional, one can call this method with `[type, value]` or `[type]`.

try to drop condition for index 0 (always compare the values under 0-index) and see if any test is failing ;)

staabm: Looks pretty good so far :-)

src/Tokenizer/Token.php
```diff
@@ -166,11 +166,11 @@ public function equals($other, $caseSensitive = true)
             return $this->content === $otherPrototype;
         }
 
-        if (array_key_exists(0, $otherPrototype) && $this->id !== $otherPrototype[0]) {
+        if ($this->id !== $otherPrototype[0]) {
             return false;
         }
 
-        if (array_key_exists(1, $otherPrototype)) {
```
keradus: I believe we can drop array_key_exists, one condition (regardless which one) is enough

staabm:
Not sure why the coverage build is failling.. do I need todo something about that?

kubawerlos:
Coverage build is failing for a long time and is not mandatory, the problem is that deployment stage failed - if you rebase your branch all should be fine.

keradus:
kudos @staabm !!!

PR #4047, opened 2018-10-16
Token - inline $other->getPrototype() to speedup equals()

staabm:
as measured in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4026 we try to speedup equals() by inlinening the $this->getPrototype() call.

~~also we use FQCN for native functions, because equals() is called a lot.~~

this speedsup the process by 4-5%.

see https://blackfire.io/profiles/compare/f4eefae8-6d09-4ab4-a833-f828d74a7b52/graph
image

### DIFF
--- a/src/Tokenizer/Token.php
+++ b/src/Tokenizer/Token.php
@@ -156,7 +156,21 @@ class Token
      */
     public function equals($other, $caseSensitive = true)
     {
-        $otherPrototype = $other instanceof self ? $other->getPrototype() : $other;
+        if ($other instanceof self) {
+            // Inlined getPrototype() on this very hot path.
+            // We access the private properties of $other directly to save function call overhead.
+            // This is only possible because $other is of the same class as `self`.
+            if (!$other->isArray) {
+                $otherPrototype = $other->content;
+            } else {
+                $otherPrototype = [
+                    $other->id,
+                    $other->content,
+                ];
+            }
+        } else {
+            $otherPrototype = $other;
+        }
 
         if ($this->isArray !== \is_array($otherPrototype)) {
             return false;


staabm:
not sure what the failling coverage built should tell me.. do I need to add addtitional unit tests?

src/Tokenizer/Token.php
```diff
@@ -156,7 +156,19 @@ public function clearChanged()
      */
     public function equals($other, $caseSensitive = true)
     {
-        $otherPrototype = $other instanceof self ? $other->getPrototype() : $other;
+        if ($other instanceof self) {
+            // inlined getPrototype() on this very hot path
+            if (!$other->isArray) {
```
keradus: oh boy, i hate this "feature" of PHP tbh...

we basically access the internal, private property of some class, which works only because we run it from method of same class, yet different instance...
for me, it's violation of encapsulation...

maybe at least a comment documenting that we doing it and what for...

staabm: I agree that this is not a feature which should be used all over the codebase.

but its a nice "shortcut" to squeeze out performance on very hot code-paths

src/Tokenizer/Token.php
```diff
@@ -158,6 +158,8 @@ public function equals($other, $caseSensitive = true)
     {
         if ($other instanceof self) {
             // inlined getPrototype() on this very hot path
+            // we access the private properties of $other directly to save function call overhead.
+            // this is only possible because $other is of the same class as `self`
```
keradus: Please fix punctuation. Hard to read when first sentence has no full stop and last sentence is not starting with upper cased letter.

staabm: should be fixed.

keradus:
Kudos @staabm !!!

staabm:
Thx

SpacePossum:
thanks @staabm :+1:  :)

PR #4048, opened 2018-10-16
Tokens - inlined extractTokenKind() call on the hot path

staabm:
as identified in https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4026 we inline calls to `extractTokenKind()` which leads to a 5-7% speed increase.

most calls of this method happen thru the changed code as can be seen in the blackfire profile

image


keradus:
kudos @staabm !!!

PR #4049, opened 2018-10-16
Tokens - use parent::offset*() methods when moving items arround in insertAt()

staabm:
since we only move already existing items arround, we directly call into SplFixedArray::offset* methods.

that way we get arround additional overhead Tokens class adds with overridden offset* methods.

blackfire shows a 35-40% runtime improvement in overall walltime.
in my testbenchmark this renders a ~20 second improvement.

https://blackfire.io/profiles/compare/4a0c1058-5910-446d-9641-8bcf2823cc63/graph

image

### DIFF
--- a/src/Tokenizer/Tokens.php
+++ b/src/Tokenizer/Tokens.php
@@ -877,10 +877,14 @@ class Tokens extends \SplFixedArray
 
         $oldSize = \count($this);
         $this->changed = true;
+        $this->blockEndCache = [];
         $this->setSize($oldSize + $itemsCnt);
 
+        // since we only move already existing items around, we directly call into SplFixedArray::offset* methods.
+        // that way we get around additional overhead this class adds with overridden offset* methods.
         for ($i = $oldSize + $itemsCnt - 1; $i >= $index; --$i) {
-            $this[$i] = $this->offsetExists($i - $itemsCnt) ? $this[$i - $itemsCnt] : new Token('');
+            $oldItem = parent::offsetExists($i - $itemsCnt) ? parent::offsetGet($i - $itemsCnt) : new Token('');
+            parent::offsetSet($i, $oldItem);
         }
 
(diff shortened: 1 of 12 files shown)


keradus:
> in my testbenchmark this renders a ~20 second improvement.

20 seconds improvement for 1minute / 1hour / etc run ?

staabm:
From initially 1min10 secs

src/Tokenizer/Tokens.php
```diff
@@ -877,18 +877,23 @@ public function insertAt($index, $items)
...
         $oldSize = \count($this);
         $this->changed = true;
+        $this->blockEndCache = [];
         $this->setSize($oldSize + $itemsCnt);
 
+        // since we only move already existing items around, we directly call into SplFixedArray::offset* methods.
+        // that way we get around additional overhead this class adds with overridden offset* methods.
         for ($i = $oldSize + $itemsCnt - 1; $i >= $index; --$i) {
-            $this[$i] = $this->offsetExists($i - $itemsCnt) ? $this[$i - $itemsCnt] : new Token('');
+            $oldItem = parent::offsetExists($i - $itemsCnt) ? parent::offsetGet($i - $itemsCnt) : new Token('');
```
staabm: I tried to use parent::offsetGet() before offsetExists and determine property existance based on offsetGet returning NULL. 
Goal: reduce the number of offsetExists() calls

But this didn‘t worked out - not sure why.
Since this changes already is a major perf improvement over the current code, we should check that again in a separate PR

keradus:
Thank you @staabm.

staabm:
@keradus just a question: did you intentionally mix all these unrelated changes into my PR?

keradus:
Maybe recent GitHub issues are somehow related...
https://status.github.com/messages

staabm:
I have no problem with the current state. just curious whether this was intenionally or not.

thx

PR #4050, opened 2018-10-18
Remove unecessary callbacks for array_* functions

carusogabriel:
Not sure if this should became a Fixer :thinking:


kubawerlos:
Downvote [reason](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3065#discussion_r139208206).

carusogabriel:
@kubawerlos Thanks for the explanation. I've automated a tool for that (see https://github.com/rectorphp/rector/pull/699) and with this discussion, I'll change it to do the opposite :)

PR #4051, opened 2018-10-18
composer: allow PHP 7.3 to allow Travis testing

TomasVotruba:
At the moment this blocks testing any project/package that uses PHP CS Fixer.

https://travis-ci.org/TomasVotruba/tomasvotruba.cz/jobs/443022040

PHP 7.3 is availble: https://github.com/travis-ci/travis-ci/issues/9717#issuecomment-429564626 
and stable out in ~2 months. This would help to test code before

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || >=7.0 <7.3",
+        "php": "^5.6 || >=7.0 <7.4",
         "ext-json": "*",
         "ext-tokenizer": "*",
         "composer/semver": "^1.4",


keradus:
It doesn't matter what PHP is available by default on travis and what is not.

PHP CS Fixer is not yet, sadly, supporting PHP 7.3. For that, we won't claim in our composer file that we do.

If you want to help, please visit https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697

TomasVotruba:
> It doesn't matter what PHP is available by default on travis and what is not.

It does, since this package is the only blocker left.

> PHP CS Fixer is not yet, sadly, supporting PHP 7.3. For that, we won't claim in our composer file that we do.

Test are passing. The `composer.json` is for code, not for features.

SpacePossum:
we already test on 7.3 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/.travis.yml#L123)

>The composer.json is for code dependencies, not for features.

indeed, we do not state the tool works on PHP7.3 in composer because we don't know if that is the case, 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697 has more background information on that

what is important for us is that the fixer not only works on PHP7.3, but that it does not break PHP7.3 code when "fixing" it.

TomasVotruba:
> what is important for us is that the fixer not only works on PHP7.3, but that it does not break PHP7.3 code when "fixing" it.

I see. That's more helpful.

How would you test that without allowing to install?

SpacePossum:
>How would you test that without allowing to install?

people can test locally with a checkout of the tool, than create unit tests and fixes and PR those here :)

TomasVotruba:
I would love to, but I can't install it :(

I have many fixers that might break on PHP 7.3 code, I wanted to add 7.3 test suite to Travis to see.

kubawerlos:
Use `--ignore-platform-reqs` as temporary fix.

Issue #4054, opened 2018-10-19
PhpdocToParamTypeFixer


jg-development:
Hi,
I hava adapted the PhpdocToReturnTypeFixer fixer and created a PhpdocToParamTypeFixer.
The goal is to convert all phpdoc params to the corresponding param type hint.
Could anybody please review this class and tests for feedback?

Greetings Jan

julienfalque:
Hi, this feature was already requested in #3258 so I'm closing this one as duplicate, thanks.

Issue #3258, opened 2017-11-15
RFC: PHPDoc types to typehints


umpirsky:
It would be nice to have fixer that takes param types form PHPDoc and adds them as type hints where applicable.

julienfalque:
See #3640 (for return types only).

SpacePossum:
done through https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4056

PR #4056, opened 2018-10-20
PhpdocToParamTypeFixer - Introduction

jg-development:
like #4054 add PhpdocToParamTypeFixer
but from master to master

Finishes #3258


jg-development:
Are there any objections or can it be merged?

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,317 @@
...
+    {
+        return \PHP_VERSION_ID >= 70000 && $tokens->isTokenKindFound(T_FUNCTION);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // NoSuperfluousPhpdocTagsFixer.
```
dmvdbrugge: Please comment not only what priority-relation it has, but also if it has to be run before or after said fixer.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,317 @@
...
+        return 8;
```
dmvdbrugge: While you did add a test fixture, I miss the relation in
```
\PhpCsFixer\Tests\AutoReview\FixerFactoryTest::provideFixersPriorityCases()
```

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,290 @@
...
+            'invalid class 2' => [
+                '<?php /** @param \\Foo\\\\Bar */ function my_foo() {}',
+            ],
+            'blacklisted class methods' => [
+                '<?php
+
+                    class Foo
+                    {
+                        /** @param Bar */
+                        function __construct() {}
```
dmvdbrugge: Shouldn't you have the params, to be sure it's untouched? This case isn't doing anything anyway now because of doc/signature mismatch.

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,290 @@
...
+            ],
+            'static is skipped' => [
+                '<?php
+                    class Foo {
+                        /** @param static $foo */ function my_foo($foo) {}
+                    }
+                ',
+            ],
+            'skip resource special type' => [
+                '<?php /** @param resource */ function my_foo() {}',
```
dmvdbrugge: Shouldn't you have the params, to be sure it's skipped?
This case isn't doing anything anyway now because of doc/signature mismatch.

Same for the following 3.

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,290 @@
...
+            ],
+            'root class' => [
+                '<?php /** @param \My\Bar $foo */ function my_foo(\My\Bar $foo) {}',
+                '<?php /** @param \My\Bar $foo */ function my_foo($foo) {}',
+            ],
+            'interface' => [
+                '<?php interface Foo { /** @param Bar $bar */ function my_foo(Bar $bar); }',
+                '<?php interface Foo { /** @param Bar $bar */ function my_foo($bar); }',
+            ],
+            'invalid void param on ^7.1' => [
```
dmvdbrugge: You say `^7.1` however you pass no version (contrary to the next case). One of them has got to be wrong.

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,290 @@
...
+                '<?php /** @param null|void $foo */ function my_foo($foo) {}',
+            ],
+            'iterable return on ^7.1' => [
+                '<?php /** @param iterable $counter */ function my_foo(iterable $counter) {}',
+                '<?php /** @param iterable $counter */ function my_foo($counter) {}',
+                70100,
```
dmvdbrugge: You test `iterable` and `void`, thereby proving that the version-specific types work. I do wonder, do we want `object` tested as well, just because it has a different version?

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,317 @@
...
+    private $versionSpecificTypes = [
+        'void' => 70100,
+        'iterable' => 70100,
+        'object' => 70200,
+    ];
+
+    /**
+     * @var string
+     */
+    private $classRegex = '/^\\\\?[a-zA-Z_\\x7f-\\xff](?:\\\\?[a-zA-Z0-9_\\x7f-\\xff]+)*(?<array>\[\])*$/';
```
dmvdbrugge: Is this something we have/want in a more general place? I can't imagine it's only used here, especially seeing there already is a `PhpdocToReturnTypeFixer`.

After checking, I see the exact regex is listed there, and who knows where more. This is just duplication.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,317 @@
...
+        'resource' => true,
+        'null' => true,
+        'static' => true,
+    ];
+
+    /**
+     * @var array
+     */
+    private $versionSpecificTypes = [
+        'void' => 70100,
```
dmvdbrugge: I don't know about others, but I like lists like these to be sorted, preferably by key.
If the sorting by version is wanted though, please sort the same versions by key.

_**edit:** same goes for the skipped types and blacklisted functions, please sort them._

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
jg-development: ok I clearify this

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
jg-development: I was/am not familiar with this class and the priorityCases ... I will implement it

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
jg-development: yeah, you are right I fix this

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
jg-development: added an object test for 7.2

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
jg-development: Correct, this is a duplication from PhpdocToReturnTypeFixer, but not anywhere else.
As long as there aren`t more fixers I would not create an abstract, parent or whatever class for one line.

dmvdbrugge: I'm not talking about creating a new class, I'm saying maybe there already is a class where this is more in place. Like for example, would it make sense to add it to `Annotation`? Just thinking out loud here.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
jg-development: The sorting is like in the PhpdocToReturnTypeFixer Class. If we want to change the sorting, it should be in all classes or not?

jg-development: In that case ... OK from my side... sorted it by key + value

tests/AutoReview/FixerFactoryTest.php
```diff
@@ -194,6 +194,7 @@ public function provideFixersPriorityCases()
             [$fixers['phpdoc_to_return_type'], $fixers['fully_qualified_strict_types']],
             [$fixers['phpdoc_to_return_type'], $fixers['no_superfluous_phpdoc_tags']],
             [$fixers['phpdoc_to_return_type'], $fixers['return_type_declaration']],
+            [$fixers['phpdoc_to_param_type'], $fixers['no_superfluous_phpdoc_tags']],
```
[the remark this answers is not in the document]
jg-development: argh ... fixed

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,288 @@
...
+            ],
+            'invalid class 2' => [
+                '<?php /** @param \\Foo\\\\Bar */ function my_foo() {}',
+            ],
+            'blacklisted class methods' => [
+                '<?php
+
+                    class Foo
+                    {
+                        /** @param $bar Bar */
```
dmvdbrugge: This annotation is the wrong way around, is that on purpose?

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixCases()
+    {
+        return [
+            'no phpdoc param' => [
+                '<?php function my_foo() {}',
```
dmvdbrugge: For this and the next 3 cases: yes they are all invalid, however, there are no parameters to typehint anyway, so the fixer can't do anything anyway. Shouldn't they have a parameter?

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
jg-development: a little bit, but I think I can split this test in 2

jg-development:
any open tasks or comments for me?

jg-development:
I updated a fix for stop searching the param variable with the last token.
I tested the fixer with several projects and everything was working fine.
Can somebody merge the request please?

dmvdbrugge:
Patience my friend, so far, only 1 contributor (me) has looked at your code, no actual maintainers seem to have taken interest yet.

tests/Fixer/FunctionNotation/PhpdocToParamTypeFixerTest.php
```diff
@@ -0,0 +1,293 @@
...
+            ],
+            'non-root class with single int param' => [
+                '<?php /** @param int $bar */ function my_foo(int $bar) {}',
+                '<?php /** @param int $bar */ function my_foo($bar) {}',
+            ],
+            'non-root class with single float param' => [
+                '<?php /** @param float $bar */ function my_foo(float $bar) {}',
+                '<?php /** @param float $bar */ function my_foo($bar) {}',
+            ],
+            'non-root class with multiple string param' => [
```
julienfalque: This test case is the same as the previous one.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,326 @@
...
+                if (null === $variableIndex) {
+                    continue;
+                }
+
+                if ('(' === $tokens[$variableIndex - 1]->getContent()) {
+                    if (!$tokens[$variableIndex]->isGivenKind([T_VARIABLE])) {
+                        continue;
+                    }
+                } else {
+                    if ($this->hasParamTypeHint($tokens, $variableIndex - 2)) {
```
julienfalque: I think there is no test case covering this scenario.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,326 @@
...
+                        $paramType = $types[0];
+                    }
+
+                    if ('null' !== $null) {
+                        continue;
+                    }
+
+                    $isNullable = true;
+
+                    if (\PHP_VERSION_ID < 70100) {
```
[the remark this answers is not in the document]
julienfalque: I meant the version check can be done before checking the types, so at the very beginning of the `if`:
```php
                 if (2 === $typesCount) {
                     if (\PHP_VERSION_ID < 70100) {
                         continue;
                     }
                     
                     $null = $types[0];
                     // ...
```

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
julienfalque: You refactored (and changed) these `if` conditions but you didn't add any test so I think the scenario is still not covered. Please add a test case that proves the typehint isn't changed if already defined, e.g.:
```php
'typehint already defined' => [
    '<?php /** @param string $foo */ function foo(int $foo) {}'
],
```

jg-development:
@julienfalque :
you are right, i added the changes regarding your comments

TomasVotruba:
tl;dr;

I use [`TypeHintDeclarationSniff`](https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/TypeHintDeclarationSniff.php) that does the same job for many years.

But it works only with local scope and thus is buggy in cases with parent class or interfaces:

**3rd party code**

```php
<?php

interface Sniff
{
    /**
     * @param int $position
     */
    public function proces(File $file, $position);
}
```

**My code** "fixed":

```diff
 <?php

 final class MySniff implements Sniff
 {
     /**
      * @param int $position
      */
-    public function proces(File $file, $position)
+    public function proces(File $file, int $position)
     {
     }
 }
```

Does this fixer deals with that?

jg-development:
Hi @TomasVotruba :
The fixer would fix the interface and the class seperatly.

The fixer itself does not look for a parent class or interface method and compare the new output.
In general fixers are called by file, there is no option to "lookup" for the parent class.
I think this cannot be done by the fixer alone, this is a designchange of the php-cs-fixer.
The only quick possible solution I can think of, is an option to add namespaces like "\3rdParty\"
and if an file uses this namespace, it will be ignored.
But automatic lookup by the fixer itself is not possible in this structure.
Greetings Jan

TomasVotruba:
Hi @jg-development, thanks for the answer. So it will break the code in such case?

jg-development:
yes it does

julienfalque:
This PR is looking pretty good but the fixer seems to have a lot of code from `PhpdocToReturnTypeFixer`. Wouldn't it be better to refactor those fixers to avoid duplicating the code?

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
TomasVotruba: > Like for example, would it make sense to add it to Annotation?

:+1: Or even better `AnnotationRegex` or similar, to state purporse of such class. 

Also, this should be a constant, since code should not change this value.

```php
AnnotationRegex::CLASS_NAME
```

jg-development:
@TomasVotruba :
Some kinds of param can be done, but not something like this:
Foo[][]
The problem is the DocBlock\Annotation Class. The regex for content and types is not ready for such cases.

TomasVotruba:
Why not extend it? `Foo[][]` is pretty clear

jg-development:
possible, but in my feeling not part of this story/task. another issue with extending the annotation class to return the correct types and this can be done.
And I think there could be some other fixers, that could use the change, too.

Personally, than I would change the fixer to identify all types and generate the annotation from the result. Like $hasNull + $hasIterable = ?array and so on.

TomasVotruba:
Well, then `null|Foo[]` → `?array` is already supported, 
I'd expect `null|Foo[][]` → `?array` to work as well.

But I understand you want to merge this huge PR and this might be changed on `@return` as well, thus scope for another PR.

jg-development:
Ok, I did a refactor of the fixer and modified the annotation class for multiple array annotations.
This had an impact on the phpdoctoreturntypefixer as well (imho positive).

New covered use cases are:
```
            'nullable and mixed types of arrays' => [
                '<?php /** @param null|Foo[]|Bar[] $foo */ function my_foo(?array $foo) {}',
                '<?php /** @param null|Foo[]|Bar[] $foo */ function my_foo($foo) {}',
                70100,
            ],
            'nullable and array and array of types' => [
                '<?php /** @param null|Foo[]|array $foo */ function my_foo(?array $foo) {}',
                '<?php /** @param null|Foo[]|array $foo */ function my_foo($foo) {}',
                70100,
            ],
            'nullable array of array of types' => [
                '<?php /** @param null|Foo[][] $foo */ function my_foo(?array $foo) {}',
                '<?php /** @param null|Foo[][] $foo */ function my_foo($foo) {}',
                70100,
            ],
            'array and iterable param' => [
                '<?php /** @param Foo[]|iterable $foo */ function my_foo(array $foo) {}',
                '<?php /** @param Foo[]|iterable $foo */ function my_foo($foo) {}',
                70100,
            ],
```

TomasVotruba:
@jg-development Amazing work :clap:

jg-development:
ok ... was not aware that switch case is not allowed .... using kind of polymorphism now

jg-development:
@SpacePossum 
Is there anything still todo from my side or does the merge just takes some time/timing?
Greetings Jan

kubawerlos:
@jg-development what about [this comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4056#issuecomment-443421382)?

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
jg-development: I converted it to a local constant .... PhpdocToParamTypeFixer::CLASS_REGEX

jg-development:
I resolved all issues and updated the branch. Maybe a quick look over it and it can be merged?

jg-development:
Can somebody give me an update for this.... waiting for approval about 1 1/2 month now

SpacePossum:
Sorry for the long wait, we are trying to catch up
(@julienfalque  maybe you can review again?)

jg-development:
hi,
any updates on this pull?

tests/Fixtures/Integration/misc/phpdoc_to_param_type.test
```diff
@@ -0,0 +1,17 @@
+--TEST--
+Integration of fixers: phpdoc_to_param_type.
+--RULESET--
+{"phpdoc_to_param_type": true}
```
julienfalque: The purpose of these integration tests is to check that several fixers work fine when used together. This test only uses `phpdoc_to_param_type` so it can be deleted (the scenario is already covered by unit tests of the fixer I think).

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,458 @@
...
+            if (0 < $nsIndex) {
+                $newTokens[] = new Token([T_NS_SEPARATOR, '\\']);
+            }
+            $newTokens[] = new Token([T_STRING, $value]);
+        }
+        $newTokens[] = new Token([T_WHITESPACE, ' ']);
+        $tokens->insertAt($index, $newTokens);
+    }
+
+    private function typeIsOfIterable($type)
```
julienfalque: I think those `typeIs*()` functions are superfluous and I would drop them. The comparaisons can be done directly where needed.

Also, the result of a `===` expression is already a boolean, no need to use a ternary.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,458 @@
...
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Jan Gantzert <jan@familie-gantzert.de>
+ */
+final class PhpdocToParamTypeFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    const CLASS_REGEX = '/^\\\\?[a-zA-Z_\\x7f-\\xff](?:\\\\?[a-zA-Z0-9_\\x7f-\\xff]+)*(?<array>\[\])*$/';
```
julienfalque: Please add `@internal` annotation to the constants as they are not meant to be used outside this class.

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,458 @@
...
+        $nextFunction = $tokens->getNextTokenOfKind($index, [[T_FUNCTION]]);
+        $variableIndex = $tokens->getNextTokenOfKind($index, [[T_VARIABLE]]);
+        if (\is_int($nextFunction) && $variableIndex > $nextFunction) {
+            return null;
+        }
+        if (!isset($tokens[$variableIndex])) {
+            return null;
+        }
+        $variableToken = $tokens[$variableIndex]->getContent();
+        Preg::match('/@param\s*[^\s]+\s*([^\s]+)/', $paramTypeAnnotation->getContent(), $paramVariable);
```
julienfalque: This regular expression might have unexpected results for tags that use generics, e.g. in `@param array<int, string> $foo` the variable would be `string>`. I suggest adding several cases with generics in the tests.

jg-development:
Hi @julienfalque ,
I did the changes as requested/reviewed.

jg-development:
any updates? all changes are made

jg-development:
seriously guys ... i am doing this for over 6 months now.
approved nearly two month, but everytime I has to recheck, rebase or migrate to new standards.
I am getting tired of this ... will this anytime merged or should we just reject this?

jg-development:
ok ... removed all the merge commits ... rebased and updated the branch.
can this be merged? it was approved by @TomasVotruba and all is green.

TomasVotruba:
I'm only contributor here, but it would be nice not to keep PRs rotten

jg-development:
+1 for merging plz

tigitz:
Hello @julienfalque 

Symfony community would greatly benefit from this fixer as it's moving towards type hinted source code like you can see here:
- https://github.com/symfony/symfony/pull/33154
- https://github.com/symfony/symfony/issues/32179

Is there anything we could do to make the final push here ?

Thanks everyone for your hard work in order to make it happen.

/cc @derrabus @nicolas-grekas

jg-development:
@tigitz 
thx for the support ... I am trying it for month now, but no reaction.
I am currently using the fixer in our company without any problems

tigitz:
@jg-development thanks for your work, hopefully a new dynamic with Symfony community support behind will help move things forward here. I'm sure maintainers would love to take care of this feature when they'll find the right moment for them :slightly_smiling_face: 

Meanwhile, I've tried your fixer on symfony codebase and here's a test case that fails:
```php

            'union type containing a scalar' => [
                '<?php /** @param int|stdClass $bar */ function my_foo($bar) {}',
            ],
```

It generates `function my_foo(intstdClass$bar)`.

Seems like union types containing at least 1 scalar fails to be properly fix.

jg-development:
@tigitz 
wow ... an int OR a stdClass as a param .... never thought of THAT ;-)

tigitz:
@jg-development haha I understand your surprise but it's actually a dummy example to show that every union type having at least a single scalar fails, like the more "real life" example of `string|array`

jg-development:
@tigitz 
ok I added a change for multiple types exept nullable and iterable plus array.
plz recheck maybe I forgot something.

nicolas-grekas:
Some ideas, I didn't check how they work now:
`array|\Traversable` => `iterable`
`Foo[]` => `array`
`array|Foo[]` => `array`
`iterable|Foo[]` => `iterable`
`iterable<Foo>` => `iterable`
`array<Foo>` => `iterable`

all these rules except the first should keep the annotation in place to not remove useful info

jg-development:
array|\Traversable => iterable ==> now supported
Foo[] => array ==> already works
array|Foo[] => array ==>  now supported
iterable|Foo[] => iterable ==> now supported
iterable<Foo> => iterable ==> not supported
array<Foo> => array ==> not supported

Currently the generics are not supported. You can ad a pull request to it, if you want to.

tigitz:
Thanks @jg-development !

Another small failing test case:

jg-development:
@tigitz 
that is true. I have no coice to count the number of different param types here.
I added the commit and the test given by you.

tigitz:
Great !

Last one I can infer from my tests of your fixer on symfony code base:
```
'reference symbol' => [
                '<?php /** @param array $bar */ function my_foo(array &$bar) {}',
                '<?php /** @param array $bar */ function my_foo(&$bar) {}',
            ],
            'reference symbol in php doc' => [
                '<?php /** @param array &$bar */ function my_foo(array &$bar) {}',
                '<?php /** @param array &$bar */ function my_foo(&$bar) {}',
            ],
```
Gives respectively:

```
--- Expected
+++ Actual
@@ @@
-'<?php /** @param array &$bar */ function my_foo(array &$bar) {}'
+'<?php /** @param array &$bar */ function my_foo(&$bar) {}'
```

```
--- Expected
+++ Actual
@@ @@
-'<?php /** @param array $bar */ function my_foo(array &$bar) {}'
+'<?php /** @param array $bar */ function my_foo(&array $bar) {}'
```

It should be :ok_hand: after that

jg-development:
@tigitz 
Not a nice solution in the commit, but it works

tigitz:
You're going to hate me :stuck_out_tongue: , here's 3 new failing cases:

```
'scalar type and array of same scalar type' => [
                '<?php /** @param string|string[] $bar */ function my_foo($bar) {}',
            ],
            'scalar type and array' => [
                '<?php /** @param string|array $bar */ function my_foo($bar) {}',
            ],
            'union type containing null and non scalar type' => [
                '<?php /** @param string|Transition|null $bar */ function my_foo($bar) {}',
            ],
```

Again, thanks for your work, we're getting closer :wink:

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,463 @@
...
+        }
+
+        $doc = new DocBlock($tokens[$index]->getContent());
+
+        return $doc->getAnnotationsOfType('param');
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param $index
```
derrabus:
```suggestion
     * @param int $index
```

src/Fixer/FunctionNotation/PhpdocToParamTypeFixer.php
```diff
@@ -0,0 +1,463 @@
...
+     * @param $paramTypeAnnotation
```
derrabus:
```suggestion
     * @param Annotation $paramTypeAnnotation
```

jg-development:
@tigitz 
ok last one was a bit tricky but I commited it

tigitz:
Found another failing one:
```
'non scalar type or array' => [
                '<?php /** @param array|InputDefinition $bar */ function my_foo($bar) {}',
            ],
```

```
--- Expected
+++ Actual
@@ @@
-'<?php /** @param array|InputDefinition $bar */ function my_foo($bar) {}'
+'<?php /** @param array|InputDefinition $bar */ function my_foo(arrayInputDefinition $bar) {}'
```

jg-development:
@tigitz 
fixed this too

tigitz:
Great, now onto variadics :slightly_smiling_face: 

There's different way to express them through phpdoc, most standards (PSR5 and phpDocumentator) don't take them into accounts. However PHPStorm has a faire range of support as documented here: https://stackoverflow.com/a/53804818 so I think it would be really beneficial to add a preliminary support at least.

Here are two minimal cases:
```
'variadic param type dot notation' => [
                '<?php /** @param string ...$bar */ function my_foo(string ...$bar) {}',
                '<?php /** @param string ...$bar */ function my_foo(...$bar) {}',
            ],
            'variadic param type array notation' => [
                '<?php /** @param string[] $bar */ function my_foo(string ...$bar) {}',
                '<?php /** @param string[] $bar */ function my_foo(...$bar) {}',
            ],
```

I guess there could be some edge cases in combination with `&` and `array|\Traversable ...$bar` but I would suggest to restrain from going to deep for this first version

jg-development:
@tigitz 
is the second example correct?
@param string[] $bar => string ...$bar
shouldnt that be array ...$bar

jg-development:
I think we are coming to the point, that this fixer should finally be merged.
The superspecial features shouldn´t be part of the initial fixer.
Can this be merged now?

tigitz:
> @tigitz
> is the second example correct?
> @param string[] $bar => string ...$bar
> shouldnt that be array ...$bar

Yes you're totally right, my mistake comes from my previous reasoning:
```php
<?php

/** @var string[] $strings **/
$strings = ['foo','bar'];

function stringTyped(string ...$strings) {};
function arrayTyped(array ...$strings){};

stringTyped($strings); // KO
stringTyped(...$strings); // OK
arrayTyped($strings); // OK
```

tigitz:
On the expected input->output of this fixer it LGTM :+1: 

One thing to note or maybe warn users from is, people tend to document a narrower type than what the real implementation allows:

```php
class Foo extends \ArrayAccess {
    /**
     * @param string $name
     */
    public function offsetGet($name){}
}
```

So the fixer will output :

```php
class Foo extends \ArrayAccess {
    /**
     * @param string $name
     */
    public function offsetGet(string $name){ }
}
```

Which breaks declaration compatibility with `ArrayAccess::offsetGet($offset)`

Symfony code base "suffers" from this in some places but it's a matter of "should the doc be the strict reflection of what PHP runtime would allow ? or should it be what would the programmer expects as a type when he designed its signature".

In the case of the latter, obviously doing a straight forward translation from doc to type the same way this fixer is doing will lead to problems like highlighted above.

jg-development:
ok guys .... this is waiting about a year now.
can this PLEASEEEEEE be merged into master?

julienfalque:
@jg-development Please be patient, merging this PR requires more review from PHP CS Fixer team. This is done on free time and we do not always have enough. I do want this to be merged as much as you do, and I'm pretty sure it will be done at some point :)

SpacePossum:
```
    /**
     * @param string $locale
     * @param null   $style
     * @param null   $pattern
     *
     * @return \NumberFormatter
     */
    abstract protected function getNumberFormatter($locale = 'en', $style = null, $pattern = null);
```

gets fixed to

```
    /**
     * @param string $locale
     * @param null   $style
     * @param null   $pattern
     *
     * @return \NumberFormatter
     */
    abstract protected function getNumberFormatter(string $locale = 'en', ? $style = null, ? $pattern = null);
```

```
    /**
     * Sorts files and directories by name.
     *
     * This can be slow as all the matching files and directories must be retrieved for comparison.
     *
     * @param bool $useNaturalSort Whether to use natural sort or not, disabled by default
     *
     * @return $this
     *
     * @see SortableIterator
     */
    public function sortByName(/* bool $useNaturalSort = false */)
    {
        $useNaturalSort = 0 < func_num_args() && func_get_arg(0);

        $this->sort = $useNaturalSort ? Iterator\SortableIterator::SORT_BY_NAME_NATURAL : Iterator\SortableIterator::SORT_BY_NAME;

        return $this;
    }
```

to

```
    /**
     * Sorts files and directories by name.
     *
     * This can be slow as all the matching files and directories must be retrieved for comparison.
     *
     * @param bool $useNaturalSort Whether to use natural sort or not, disabled by default
     *
     * @return $this
     *
     * @see SortableIterator
     */
    public function sortByName(/* bool $useNaturalSort = false */)
    {
        bool $useNaturalSort = 0 < func_num_args() && func_get_arg(0);

        $this->sort = $useNaturalSort ? Iterator\SortableIterator::SORT_BY_NAME_NATURAL : Iterator\SortableIterator::SORT_BY_NAME;

        return $this;
    }
```

@ see https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/Finder/Finder.php#L441

SpacePossum:
I think the two cases that cause lint errors can be prevented and patched in this PR, can you work on this @jg-development ? Looking very promising so would like to wrap up these two cases and merge this as experimental feature :)

jg-development:
@SpacePossum 
Hi,
I added to tests for your requirements and fixed it.
see https://github.com/jg-development/PHP-CS-Fixer/commit/7aa35177015891afe8a4a3e8f54fbefca4be716d
Greetings Jan

SpacePossum:
Thanks for the fast fix! 

May I bother you to ask for a rebase and squash to one commit, I know this isn't what we typically do during review (or even ask not do :blush: ), but since I think this is close to merging it would help me a lot to see one clean commit on the lasted of master here.

jg-development:
@SpacePossum 
what do you exactly mean? all my commits squash into to one commit?

jg-development:
@SpacePossum 
finally all in one commit

jg-development:
@SpacePossum 
ok .... the diff now looks good. ready to merge

SpacePossum:
it seems I can push the squashed commit back to your master, so I'll merging using;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4583

you are still the author of the commit

SpacePossum:
Thanks @jg-development for all the work and patience, great to have this :D

PR #4061, opened 2018-10-22
Use dedicated PHPUnit assertions

carusogabriel:


ntzm:
Would be better to add this logic to PhpUnitDedicateAssertFixer

SpacePossum:
it weakens the tests, for example;

```php
    public function testABC()
    {
        $a = 'abc';
        $this->assertContains('a', $a);
        $this->assertTrue(\in_array('a', $a, true));
    }
```

first passes, second fails

carusogabriel:
@SpacePossum Thanks for that. I'll follow what @ntzm said and improve our Fixers to automatically fix them

Issue #4011, opened 2018-10-02
Allow removing the "b" mode from fopen calls


nicolas-grekas:
Since #3812, php-cs-fixer keeps adding the 'b' mode to fopen() calls.
But since PHP ~4~5.something, this is already the default.
It should be disabled by default IMHO.

SpacePossum:
Hi @nicolas-grekas 
Can you point to the code where PHP makes sure this is the default? 
AFAIK the default is left to the SAPI and version of PHP in use, leaving room for different behavior of the same code on different environments when the flag is omitted.

Secondly, do you want a rule to remove the flag or would removing the current rule that adds it from the Symfony set be sufficient?

nicolas-grekas:
This commit I think: https://github.com/php/php-src/commit/fa342d2404a9c809d0a182d80798da2abd3a98f7

SpacePossum:
:+1:  that settles it than :)
do you still want a fixer to remove the `t` flag as well, or just the `b` flag?

nicolas-grekas:
The "t" flag should never be used for sure (it's not portable).
About the "b", I think it's fine to not care. At least we could remove it from the Symfony ruleset I think.

ntzm:
Welp, that's my bad, part of the documentation is out of date :(

keradus:
let's simply update the fixer and ruleset, nice finding @nicolas-grekas

SpacePossum:
reopen to discuss changing the default config (and on what release (branch))

dbu:
the documentation of php itself "strongly recommends" to not rely on the default "b" and set it explicitly: https://www.php.net/manual/en/function.fopen.php

SpacePossum:
I think this issue has been resolved here;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/69b973a4ee21bc2c8d1a10b37b37c31c1379ee69#diff-1e418beac3b4a9cf60ecd4ca4e2975a8R171

so by default the SF will remove the `b` flag.

@dbu I think the documentation of PHP is outdated on this topic, however I config my projects to keep adding the `b` for the same `strong recommends` :)

Issue #3983, opened 2018-08-29
fopen_flag_order fixer changing all r+b to b+r and w+b to b+w


prisis:
For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

#### The PHP version you are using (`$ php -v`):
PHP 7.2.0 (cli) (built: Jan  4 2018 01:20:58) ( NTS )
with Zend OPcache v7.2.0, Copyright (c) 1999-2017, by Zend Technologies
with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.13.0 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski

#### The command you use to run PHP CS Fixer:
=> php vendor/bin/php-cs-fixer fix src/ --rules=fopen_flag_order,fopen_flags --allow-risky=yes

#### The configuration file you are using, if any:
....

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

\fopen('php://temp', 'r+b');
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
\fopen('php://temp', 'b+r');
```

 * with the changes you expected instead:
```php
=> \fopen('php://temp', 'r+b')
```

shouldn't this still be r+b and not be changed to b+r?

To use these flags, specify either 'b' or 't' as the last character of the mode parameter, from http://php.net/manual/en/function.fopen.php.

keradus:
ping @SpacePossum , can you have a look ?

SpacePossum:
I cannot reproduce this problem.

```bash
[08:27 AM]-[possum@zoo1]-[/home/possum/work/PHP-CS-Fixer]-[git (v2.13.0)] 
$ php php-cs-fixer fix tmp/test.php -vvv --dry-run --diff --rules=fopen_flag_order,fopen_flags
Loaded config default from "/home/possum/work/PHP-CS-Fixer/.php_cs.dist".
Paths from configuration file have been overridden by paths provided as command arguments.
.
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error

Checked all files in 0.004 seconds, 10.000 MB memory used
[...]

PR #4064, opened 2018-10-23
Ignore invalid mode strings, add option to remove the "b" flag.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3983
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4011

- ignore invalid mode-strings
- add config option to add or remove the `b` flag

### DIFF
--- a/src/Fixer/FunctionNotation/FopenFlagsFixer.php
+++ b/src/Fixer/FunctionNotation/FopenFlagsFixer.php
@@ -13,6 +13,9 @@
 namespace PhpCsFixer\Fixer\FunctionNotation;
 
 use PhpCsFixer\AbstractFopenFlagFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Token;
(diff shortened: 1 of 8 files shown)


src/Fixer/FunctionNotation/FopenFlagsFixer.php
```diff
@@ -21,21 +24,42 @@
...
+    }
+
     /**
      * {@inheritdoc}
      */
     public function getDefinition()
     {
         return new FixerDefinition(
-            'The flags in `fopen` calls must contain `b` and must omit `t`.',
+            'The flags in `fopen` calls must omit `t` and `b` must be used consistent.',
```
ntzm: I think this would read better:

```suggestion
            'The flags in `fopen` calls must omit `t`, and `b` must be omitted or included consistently.',
```

SpacePossum: Im following the [SF doc standard](https://symfony.com/doc/current/contributing/documentation/standards.html) ; 
`Punctuation: avoid the use of Serial (Oxford) Commas;`

ntzm: Reading it left-to-right reads like this to me:

> The flags in `fopen` calls must omit `t` and `b`

then you get to

> must be...

and you have to backtrack to understand the sentence fully. Also, I believe an Oxford Comma is only considered that when listing a series of three or more things.

SpacePossum: :+1:  merged, can you fix the readme one as well?

src/Fixer/FunctionNotation/FopenFlagsFixer.php
```diff
@@ -21,21 +24,42 @@
...
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('b_mode', 'The `b` flag must be used (`true`) or omitted (`false`).'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(true)
```
keradus: what is expected behaviour under 3.0 ?

SpacePossum: same i guess, or do you want me to change it on 3.0?

ntzm: I think `false`, as mentioned in #4011 it's the default for all supported PHP versions

PR #4071, opened 2018-10-29
DX: do not insert Token when calling removeLeadingWhitespace/removeTrailingWhitespace from Tokens

kubawerlos:
Previously when calling `Tokens::removeLeadingWhitespace`/`Tokens::removeTrailingWhitespace` it could happen that new `Token` was inserted to collection - which would result with changes of indices.

### DIFF
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -1175,6 +1175,48 @@ echo $a;',
     }
 
     /**
+     * Action that begins with the word "remove" should not change the size of collection.
+     */
+    public function testRemovingLeadingWhitespaceWillNotIncreaseTokensCount()
+    {
+        $tokens = Tokens::fromCode('<?php
+                                    // Foo
+                                    $bar;');
+        $originalCount = $tokens->count();
+
+        $tokens->removeLeadingWhitespace(4);
+
+        $this->assertSame($originalCount, $tokens->count());
+        $this->assertSame(
+            '<?php
+                                    // Foo
+$bar;',
+            $tokens->generateCode()
+        );
+    }
+
+    /**
+     * Action that begins with the word "remove" should not change the size of collection.
+     */
+    public function testRemovingTrailingWhitespaceWillNotIncreaseTokensCount()
+    {
+        $tokens = Tokens::fromCode('<?php
+                                    // Foo
+                                    $bar;');
+        $originalCount = $tokens->count();
+
+        $tokens->removeTrailingWhitespace(2);
+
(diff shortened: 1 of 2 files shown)


keradus:
I would love to see an unit test for that

tests/Tokenizer/TokensTest.php
```diff
@@ -1174,6 +1174,21 @@ public function provideRemoveTrailingWhitespaceCases()
...
+    public function testRemovingLeadingWhitespaceWillNotIncreaseTokensCount()
+    {
+        $tokens = Tokens::fromCode('<?php
+                                    // Foo
+                                    $bar;');
+        $originalCount = $tokens->count();
+
+        $tokens->removeLeadingWhitespace(4);
+
+        $this->assertLessThanOrEqual($originalCount, $tokens->count());
```
keradus: `<=` ? doesn't sound strict. what exactly we are asserting here ?

kubawerlos: That we won't increase the size of token's collection

tests/Tokenizer/TokensTest.php
```diff
@@ -1174,6 +1174,36 @@ public function provideRemoveTrailingWhitespaceCases()
...
+    public function testRemovingTrailingWhitespaceWillNotIncreaseTokensCount()
+    {
+        $tokens = Tokens::fromCode('<?php
+                                    // Foo
+                                    $bar;');
+        $originalCount = $tokens->count();
+
+        $tokens->removeTrailingWhitespace(2);
+
+        $this->assertSame($originalCount, $tokens->count());
```
SpacePossum: please add:
```php
        $this->assertSame(
            '<?php
                                    // Foo
                                    $bar;',
            $tokens->generateCode()
        );
```

tests/Tokenizer/TokensTest.php
```diff
@@ -1174,6 +1174,36 @@ public function provideRemoveTrailingWhitespaceCases()
...
+    public function testRemovingLeadingWhitespaceWillNotIncreaseTokensCount()
+    {
+        $tokens = Tokens::fromCode('<?php
+                                    // Foo
+                                    $bar;');
+        $originalCount = $tokens->count();
+
+        $tokens->removeLeadingWhitespace(4);
+
+        $this->assertSame($originalCount, $tokens->count());
```
SpacePossum: please add
```php
        $this->assertSame(
            '<?php
                                    // Foo
$bar;',
            $tokens->generateCode()
        );
```

SpacePossum:
Thank you @kubawerlos.

PR #4073, opened 2018-10-30
IsNullFixer - fix function detection

kubawerlos:


SpacePossum:
Thank you @kubawerlos.

Issue #4013, opened 2018-10-03
FileFilterIterator can skip a file if it has a symlink (depends on the order the files are read)


rybakit:
#### The PHP version you are using (`$ php -v`):
```bash
PHP 7.2.10 (cli) (built: Sep 11 2018 07:06:00) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.10, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.1, Copyright (c) 2002-2018, by Derick Rethans
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
```bash
PHP CS Fixer 2.13.0 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski (7136aa4)
```

#### The command you use to run PHP CS Fixer:
```bash
PHP_CS_FIXER_ALLOW_XDEBUG=1 vendor/bin/php-cs-fixer fix --config=.php_cs --dry-run --using-cache=no --path-mode=intersection var/cs/qux.php
```

#### The configuration file you are using, if any:
```php
$finder = PhpCsFixer\Finder::create()->in(__DIR__.'/var')
    ->name('*.php')
    ->sortByName() // <- make sure we read the symbolic link first
;

return PhpCsFixer\Config::create()
    ->setFinder($finder)
    ->setRiskyAllowed(true)
    ->setRules([
        '@Symfony' => true,
        '@Symfony:risky' => true,
    ])
;
```

#### Directory structure:

```
var
    cs
        foo
            bar.php -> ../qux.php
        qux.php
```

When `bar.php` is read first, it will be filtered (because it's a symlink to `qux.php`) and the real path to `qux.php` will be added to `$visitedElemets`: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2f48a42f8a0edf9a75a7f34e1e494eb355c5dbf2/src/Runner/FileFilterIterator.php#L66-L70
On a next iteration, when `qux.php` is read, it will be ignored, as `$visitedElemets` already holds the path.

Possible fix:

```diff
-        $path = $file->getRealPath();
+        $path = $file->isLink() ? (string) $file : $file->getRealPath();
```

SpacePossum:
thanks for reporting @rybakit , I've greated a PR that I hope resolves this issue

PR #4074, opened 2018-11-01
FileFilterIterator - do not filter out files that need fixing

SpacePossum:
..., but ignore other symlinks to those files.

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4013


tests/Runner/FileFilterIteratorTest.php
```diff
@@ -188,4 +188,31 @@ public function testInvalidIterator()
 
         iterator_to_array($filter);
     }
+
+    public function testFileIsAcceptedAfterFilteredAsSymlink()
```
ntzm: I think this is a unix-only feature, although I don't know much about symlinks in Windows

SpacePossum: :+1: 
hmm maybe I could refactor making the symlink on the fly so the tests passes on all OS'es, not sure  though...  maybe someone with more Windows knowledge can assist here

keradus: it would be `mklink` on Windows, but I'm not convinced it's worth to pursuit

in the end, we don't want to test test `is_link` logic of internal PHP function on our side

PR #4082, opened 2018-11-08
Failing test case for non-utf8 diff

ostrolucky:
Hello, I work on some legacy projects and PHP-CS-Fixer differ does not work properly there, because it has some issues with non utf8 files


SpacePossum:
thanks for reporting, should be fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4093 (I hope)

ostrolucky:
It does, thank you

PR #4085, opened 2018-11-08
Priority: AlignMultilineComment should run before every PhpdocFixer

### LINKED ISSUE #3879: Priority: align_multiline_comment should be run before every phpdoc_ fixer
See also #3857

```
```

```php
/**
 * This is a broken phpdoc
 *
 * @param string $str The next line has no *.

 * @return int The number of runs needed to fix this is 2 :( .
 */
function fixMe($str);
```

* after first run:
```php
/**
 * This is a broken phpdoc
 *
 * @param string $str The next line has no *.
 *
 * @return int The number of runs needed to fix this is 2 :( .
 */
function fixMe($str);
```

* after second run:
```php
/**
 * This is a broken phpdoc.
 *
 * @param string $str the next line has no *
 *
 * @return int the number of runs needed to fix this is 2 :(
 */
function fixMe($str);
```

* expected:

Meta: this is case 61 of #3844

dmvdbrugge:
I made use of the graph I generated with https://github.com/dmvdbrugge/fixer-prio-graph

I also found a hidden priority between LineEnding and Braces, which I had to add in this PR as well, because otherwise the `@Symfony_whitespace.test` would fail.

Fixes #3879


src/Fixer/Basic/BracesFixer.php
```diff
@@ -127,7 +127,7 @@ public function bar($baz)
     public function getPriority()
     {
         // should be run after the ElseIfFixer, NoEmptyStatementFixer and NoUselessElseFixer
-        return -25;
+        return 33;
```
keradus: worth to mention `LineEnding` relation

src/AbstractNoUselessElseFixer.php
```diff
@@ -26,7 +26,7 @@
     public function getPriority()
     {
         // should be run before NoWhitespaceInBlankLineFixer, NoExtraBlankLinesFixer, BracesFixer and after NoEmptyStatementFixer.
-        return 25;
+        return 34;
```
keradus: why that change?

src/Fixer/ControlStructure/ElseifFixer.php
```diff
@@ -36,6 +36,14 @@ public function getDefinition()
         );
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        return 34;
```
keradus: why that change ?

src/Fixer/ControlStructure/NoAlternativeSyntaxFixer.php
```diff
@@ -67,7 +67,7 @@ public function isCandidate(Tokens $tokens)
      */
     public function getPriority()
     {
-        return 1;
+        return 35;
```
keradus: why that change ?

src/Fixer/ControlStructure/NoUnneededCurlyBracesFixer.php
```diff
@@ -52,7 +52,7 @@ public function getDefinition()
     public function getPriority()
     {
         // must be run before NoUselessElseFixer and NoUselessReturnFixer.
-        return 26;
+        return 35;
```
keradus: why that change ?

src/Fixer/Semicolon/NoEmptyStatementFixer.php
```diff
@@ -42,7 +42,7 @@ public function getPriority()
     {
         // should be run before the BracesFixer, CombineConsecutiveUnsetsFixer, NoExtraBlankLinesFixer, MultilineWhitespaceBeforeSemicolonsFixer, NoSinglelineWhitespaceBeforeSemicolonsFixer,
         // NoTrailingCommaInListCallFixer, NoUselessReturnFixer, NoWhitespaceInBlankLineFixer, SpaceAfterSemicolonFixer, SwitchCaseSemicolonToColonFixer.
-        return 26;
+        return 35;
```
keradus: ?

tests/Fixtures/Integration/priority/line_ending,braces.test
```diff
@@ -0,0 +1,6 @@
+--TEST--
+Integration of @Symfony [whitespace version].
```
keradus: i don't think so

keradus:
I see plenty changes of `getPriority` methods for Fixers that are not Phpdoc*/Braces/LineEnding, without any reasoning added for those changes (keep in mind that when we do non-0 prio, we put a comment why - and those comments were either not added in your PR, or not updated while update of prio)

src/Fixer/ControlStructure/NoAlternativeSyntaxFixer.php
keradus: and for other changes...

dmvdbrugge:
@keradus the answer to all your "why" is because of all the existing priorities. See the graph I mentioned, see all defined priorities, and see that all tests still pass.

keradus:
>  See the graph I mentioned, see all defined priorities, and see that all tests still pass.

half of priorities are not covered with i-tests, eg even the one you are just saying you are fixing in this PR:
> AlignMultilineComment should run before every PhpdocFixer

You added a priority relation between `AlignMultilineComment` and `Phpdoc*Fixer`s, but you did not added integration tests for that.

For that, I'm not happy to change a priority without adding comment with reasoning and integration test for it.

keradus:
Priority is one of the most tricky part of this tool. Maybe let's try to split this PR into atomic pieces, eg for what I understand, `LineEnding and Braces` can be extracted to separated PR.
Smaller PRs, easier to review & manual QA, easier to spot code changes that shall not be done or are missing tests.

keradus:
> > Smaller PRs

> This really is not that big of a PR...

You say this PR is small in terms of diff size. Indeed, it is.
Yet, I'm saying this PR is huge in terms of complexity, impact and potential risk.

Don't get me wrong - your work here is super nice, simply part of the code you are touching is super fragile, and any mistake here has big consequences.
And it's not trivial to validate if the changes are good or not. "small diff" you say? few lines to read, hard to understand the changes, impact impossible to predict, especially with missing tests.

And well, outcome of it is clear, nobody wants to review it (don't worry, I don't want to sound harsh here, this PR is on my watchlist and won't get forgotten for sure (even if I can't spend as much time with it as I would like to)), as nobody can say "it's safe to release this PR" - I'm trying to ask you for changes, that would make us more safe to release it.

-------------------------

> > I see plenty changes of getPriority methods for Fixers that are not Phpdoc*/Braces/LineEnding, without any reasoning added for those changes

> As I mentioned before, it's a cascading change (again, see the graph, which visualises all current defined relations).

and for that, as it's all-in-one, hard to see out of the box which change in code is related to what change we want to accomplish.


> > half of priorities are not covered with i-tests

> So we should write them. Please don't let that stop existing PRs that try to fix priority-problems.
> (...)
> Yes, sorry, none of the "special phpdoc cases" have integration tests as far as I could see.

So, if some legacy code has no tests, it's not an excuse to not create tests for new code/scenarios.
I'm not asking you to add in this PR tests for existing-not-covered cases.
I'm asking you to add them for new priority relation cases you are adding.
You added changes in "special cases" - `provideFixersPrioritySpecialPhpdocCases` https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/56557496da75f1feec1e12377c500f7a9b4688c2/tests/AutoReview/FixerFactoryTest.php#L219 . Those are that legacy cases without proper testing, and those are ignored in our test guard: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/56557496da75f1feec1e12377c500f7a9b4688c2/tests/AutoReview/FixerFactoryTest.php#L310 .
We shall not add more tests that are not covered by the guard.


> > For that, I'm not happy to change a priority without adding comment with reasoning and integration test for it.

> The comment is there though. I could link to the issue as well. The reasoning I hope is clear to you: It fixes the `*`s of docblocks, they could be missing. And while that still is a valid docblock, there are fixers that won't run (correctly) in that case.

The comment part, especially after `I will boyscout these` (kudos :) ), is good part of the changes.
Lack of integration tests for newly configured priority relationship, is not.


> > Maybe let's try to split this PR into atomic pieces, eg for what I understand, LineEnding and Braces can be extracted to separated PR.

> In theory yes, however that would block the current PR which needs to change those priorities, that's why it was found in the first place, and needed to be fixed as well.

I prefer to merge multiple small, easy and safe PRs, one after another, steps by steps, instead of having one problematic, long-living PR.

src/Fixer/ControlStructure/NoAlternativeSyntaxFixer.php
keradus: > AlignMultilineComment should run before every PhpdocFixer

> `AlignMultilineCommentFixer` needed to change from -40 to +30

would adjusting priority of `PhpdocFixer` instead of `AlignMultilineCommentFixer` produce less priority changes ?

dmvdbrugge: Unfortunately, no. If it would've, I'd have done that instead 😉

You can see that in the graph as well.

dmvdbrugge:
I pulled out the `LineEnding and Braces` part into #4245 together with the added comments.

julienfalque:
@dmvdbrugge #4245 was merged, can you please rebase this PR?

dmvdbrugge:
@julienfalque rebase done, had to work in a "new" (since a year ago) fixer, but all seems good again.

keradus:
( i rebase to re-trigger the CIs with newest changes )

SpacePossum:
Hi @dmvdbrugge ,

>AlignMultilineComment should run before every PhpdocFixer

where does this come from? I would like to right the priority integration test so we can move forward with this PR :)

src/Fixer/ClassNotation/SingleTraitInsertPerStatementFixer.php
```diff
@@ -43,8 +43,8 @@ final class Example
 
     public function getPriority()
     {
-        // must be run before Braces and SpaceAfterSemicolonFixer
-        return 1;
+        // should be run before Braces, SpaceAfterSemicolonFixer
```
kubawerlos: Why changing `should` to `must`? It must run in that order for single run fixing, `should` suggest it's optional.

dmvdbrugge: Consistency: most other priority-comments use `should`

dmvdbrugge:
> > AlignMultilineComment should run before every PhpdocFixer
> 
> where does this come from? I would like to write the priority integration test so we can move forward with this PR :)

See the issue linked in the PR description: #3879 and the issue that issue also refers to: #3857 both issues that you've been involved in (more than a year ago, so I get it if you don't remember 😉)

SpacePossum:
Both linked issues are about PHPDoc's AFAIK. This PR also states that the LineEndingFixer must run after the BracesFixer, I cannot see the priority issue, I cannot find it in the linked issues, this is a big part of why some PR's take a long time to get merged, because I don't merge PR's without tests especially if I don't understand the need for the change.

dmvdbrugge:
"LineEnding and Braces" has been pulled out and is merged (by you) already. This PR is (since that) purely about phpdoc.

SpacePossum:
so this is outdated? https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4085/files#diff-b4335f803c5a32ac9837972e11e4ddafR58

dmvdbrugge:
No it's not. The pulled-out code places the priority, but because Braces already ran at -25 there was no need to add a number for LineEnding. However, because this PR puts Braces at +34, LineEnding needs to change with it. (To still run before it, keeping the priority.)

SpacePossum:
ok, so why is Braces changed to 34 as it is not a PHPDoc fixer?

dmvdbrugge:
It might be a lot to ask, but please read the entire discussion in this PR again, because all is explained already. TLDR: the required change is the title (AlignMultilineComment before the Phpdoc*), all other changes is to keep priority correct. See the graphs.

keradus:
a lot of integration tests were already added to cover some of those missing priority cases.
would you mind rebase and see what is still missing to cover with tests, @dmvdbrugge ?
i know it's a lot to ask, no need to rush on it ;)

src/Fixer/ClassNotation/SingleTraitInsertPerStatementFixer.php
kubawerlos: @dmvdbrugge really? Can you point a fixer in `2.15` that is using should? ;)

dmvdbrugge: See the other files in this diff, most of them use "should" and comma-separation. There is another inconsistency using "be" or not but whatever.

I do find it interesting that *this* is the point you comment on after all this time 😄

kubawerlos: > See the other files in this diff

I did not ask about the diff, but the `2.15` branch ;)

dmvdbrugge: I meant of course the unchanged lines that can also be seen in the diff, which is based upon 2.15

If you meant to say that in the last 2 years something has changed (which for some reason is not visible in the diff), please just do so.

kubawerlos: I was hoping you find it yourself e.g. during doing rebase, but ok, this happened: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4758

dmvdbrugge: I'm currently not able to rebase (work...) but I can always read comments. I suspected something like that might've happened, but a simple "this has changed now" would've been a more effective way of re-opening the discussion, instead of challenging me on something that was correct when I wrote it (and according to the unchanged parts of the diff still is, blame github for that).

Thanks for the link, I will keep that in mind when I get around to rebasing this.

keradus:
I let myself integrate the integration test into the PR, with having the actual test proving the changes are solving the issue, I'm feeling much more comfortable merging this.

I synced the branch (ignore the merge commit and fabbot complain, I gonna squash it during merging anyway) and put milestone back.

keradus:
@kubawerlos , i believe we should aim to drop numbers and have relations, like @SpacePossum described here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5078#issuecomment-683424449

keradus:
The failure we see in the test for PHP8 build is caused by not-known-before fixers relation between `method_argument_space` and `function_declaration`, and as we were not aware about the relation, we did not have explicit test for it.
Now, we accidentally discovered it with PHP8 integration test.

Having all relations discovered and documented with dedicated test is almost impossible, and changing the priorities like in this PR is putting us at risk to introduce such issues.
This time we were lucky that one of the test, even when not designed for it, discovered the problem.

FYI @dmvdbrugge @julienfalque @SpacePossum @kubawerlos

keradus:
Trying to solve above issue in #5552 ,
then we should unblock this PR

keradus:
Thank you @dmvdbrugge.

dmvdbrugge:
Thanks a lot for finishing this @keradus and my apologies for being awol.

PR #4088, opened 2018-11-10
Add Travis test with strict types

kubawerlos:


src/DocBlock/Line.php
```diff
@@ -42,6 +42,8 @@ public function __construct($content)
      * Get the string representation of object.
      *
      * @return string
+     *
+     * @deprecated will be removed in 3.0, use getContent
```
keradus: minor: why
major: how it's related to topic of this PR ?

.travis.yml
```diff
@@ -92,6 +92,14 @@ jobs:
             stage: Test
             php: 7.2
 
+        -
+            <<: *STANDARD_TEST_JOB
+            stage: Test
+            php: 7.2
+            env: STRICT_TYPES=1
+            before_script:
+                - PHP_CS_FIXER_FUTURE_MODE=1 php php-cs-fixer fix --quiet --rules=blank_line_after_opening_tag,declare_strict_types
```
keradus: we already do this, and more:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.12/.travis.yml#L126

SpacePossum:
I like the change to remove the `_toString()` method/call. The test is a good idea as well but as keradus pointed out we already do it :)

So maybe update the PR to only keep the `_toString()` deprecation?

src/DocBlock/Line.php
kubawerlos: > minor: why ?

why what? We add deprecations in major releases, aren't we?

> major: how it's related to topic of this PR ?

To make the strict pass I had to update the call, anyway - moved to separate PR: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4098

.travis.yml
kubawerlos: Yeah, but this job is allowed to fail, which is happening since I remember :(

kubawerlos:
@keradus @SpacePossum what about making the strict check mandatory?

(Travis will fail until [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4098) is merged)

SpacePossum:
maybe we can add the `strict_types` to the standard PHP 7 job without all the other risky ones from the migration sets?

src/DocBlock/Line.php
keradus: > why what? We add deprecations in major releases, aren't we?

why to deprecate it at all ?

kubawerlos:
@SpacePossum like this?

.travis.yml
keradus: then let it not allow to fail on newest stable PHP release, and let allow nightly to fail only because it's unstable PHP, not because we run strict checks there

kubawerlos: So like this or in separate job or how?

src/DocBlock/Line.php
SpacePossum: `__toString ` is not the best of method names here as you get the content of the Line, not a description of the object. We cannot throw exceptions from the method or later on change the signature because it is a magic method. `echo ltrim($line->__toString());` doesn't like nice to me and  `echo ltrim($line);` triggers a warning under strict type, so than you stuck with `echo ltrim((string) $line);`, or the suggestion here `echo ltrim($lime->getContent())` which looks more descriptive to me. Than again, I'm from Java world where `toString` was only used for debugging.

.travis.yml
keradus: almost like this - here is the proper place - 7.2 job, but run all extra rules, not just those randomly picked

keradus:
btw, please target 2.12, nothing that we shall benefit only since 2.14

.travis.yml
kubawerlos: Done (those weren't random - `declare_strict_types` plus `blank_line_after_opening_tag` to cleanup after it).

.travis.yml
```diff
@@ -91,6 +91,8 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.2
+            before_script:
+                - php php-cs-fixer fix --rules @PHP71Migration,@PHP71Migration:risky,blank_line_after_opening_tag,native_function_invocation -q || travis_terminate 1
```
keradus: i believe we can drop it from `php: nightly` now, what do you think? @kubawerlos @SpacePossum

SpacePossum:
Thank you @kubawerlos.

PR #4091, opened 2018-11-12
Adjust misleading sentence in CONTRIBUTING.md

ostrolucky:
In previous message it wasn't clear how to find which branch is still maintained. However, when branch is no longer deleted, it's just deleted. This means users don't need to concern about which branch is maintained.


dmvdbrugge:
This should target 2.12 though 😉

ostrolucky:
Oh the irony. Addressed.

keradus:
Thank you @ostrolucky.

Issue #4086, opened 2018-11-09
NoUnusedImportsFixer removes use trait


dmvdbrugge:
So I have a test-class in a namespace, and a trait in the same namespace (different file, both adhering PSR-4). The test uses the trait somewhere in the middle, just above the case that actually uses it (yes I have colleagues with annoying habits).

`no_unused_imports` wrongly removes the use of the trait.

The sample code sketches the situation, however does not reproduce it 😞
I have not found a way to actually reproduce it yet, however on the actual code (cannot share...) it's consistently being removed.

_Weird thing 1_: when I manually move the use to be the first thing in the class, it stays.
_Weird thing 2_: when I instead run with `ordered_class_elements` too (via commandline:)
```
--rules='{"ordered_class_elements":{"order":["use_trait","constant_public","constant_protected","constant_private","property_public","property_protected","property_private"]},"no_unused_imports":true}'
```
even though it has a higher priority than `no_unused_imports`, the use gets removed and the `ordered_class_elements` has not triggered.

#### The PHP version you are using (`$ php -v`):
```
PHP 7.1.23 (cli) (built: Oct 11 2018 16:21:11) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.1.23, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
```
PHP CS Fixer 2.13.1 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski
```

#### The command you use to run PHP CS Fixer:
```
vendor/bin/php-cs-fixer fix -v --using-cache=no --rules=no_unused_imports <the_file.php>
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

namespace Inside\Some\TestNamespace;

use PHPUnit\Framework\TestCase;

class NoUnusedImportsBug extends TestCase
{
    public function testOneThing()
    {
        self::assertTrue(true);
    }

    use MyTrait;
    
    public function testSomethingUsingTrait()
    {
        // This function comes from MyTrait
        $this->traitFunc();

        self::assertTrue(true);
    }
}
```

[...]

PR #4092, opened 2018-11-12
UseTransformer - simplify/optimize

SpacePossum:
found when looking into;
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4086

(does not fix it though)

### DIFF
--- a/src/Tokenizer/Transformer/UseTransformer.php
+++ b/src/Tokenizer/Transformer/UseTransformer.php
@@ -51,21 +51,20 @@ final class UseTransformer extends AbstractTransformer
     {
         if ($token->isGivenKind(T_USE) && $this->isUseForLambda($tokens, $index)) {
             $tokens[$index] = new Token([CT::T_USE_LAMBDA, $token->getContent()]);
-        }
 
-        if (!$token->isClassy()) {
             return;
         }
 
-        $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
-        $prevToken = null === $prevTokenIndex ? null : $tokens[$prevTokenIndex];
+        // Only search inside class/trait body for `T_USE` for traits.
+        // Cannot import traits inside interfaces or anywhere else
 
-        if ($prevToken->isGivenKind(T_DOUBLE_COLON)) {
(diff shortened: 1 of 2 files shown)


src/Tokenizer/Transformer/UseTransformer.php
```diff
@@ -51,22 +51,17 @@ public function process(Tokens $tokens, Token $token, $index)
     {
         if ($token->isGivenKind(T_USE) && $this->isUseForLambda($tokens, $index)) {
             $tokens[$index] = new Token([CT::T_USE_LAMBDA, $token->getContent()]);
-        }
 
-        if (!$token->isClassy()) {
```
SpacePossum: narrowed down to `class` or `trait` as `interfaces` cannot import `traits`

src/Tokenizer/Transformer/UseTransformer.php
```diff
@@ -51,22 +51,17 @@ public function process(Tokens $tokens, Token $token, $index)
...
             return;
         }
 
-        $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
-        $prevToken = null === $prevTokenIndex ? null : $tokens[$prevTokenIndex];
+        // Only search inside class/trait body for `T_USE` for traits.
+        // Cannot import traits inside interfaces or anywhere else
 
-        if ($prevToken->isGivenKind(T_DOUBLE_COLON)) {
```
SpacePossum: this was done to check for `Foo::class`, however in such statement `class` is not a `T_CLASS` and so not `classy`

tests/Tokenizer/Transformer/UseTransformerTest.php
```diff
@@ -75,6 +75,62 @@ public function provideProcessCases()
...
+                CT::T_USE_LAMBDA,
+                CT::T_USE_TRAIT,
+            ]
+        );
+    }
+
+    public function provideFix70Cases()
+    {
+        return [
+            'nested anonymous classes' => [
```
SpacePossum: this use case of nested anonymous classes didn't exists

keradus:
Thank you @SpacePossum.

PR #4096, opened 2018-11-12
DX: fixing typo

kubawerlos:
Some noob could not correctly write word "argument" :)

After discovering this I'd created [this](https://gist.github.com/kubawerlos/d187e42d209d00dda2cea6284e42a220) test - do you think it would be useful? To unify "indexes" vs "indixes", "regex" vs "regexp" and more, in future.


kubawerlos:
> detecting typos by whitelisting accepted words ? then we need to whitelist whole dict...

"Only" 840 words :)

> I would choose "regex" here ;)

I meant "regex" vs "regexp".

SpacePossum:
I like the idea of having a basic spelling check rule, aren't there like open standards and open dictionaries? 
Maybe someone knows how fabbot does it?

PR #4097, opened 2018-11-12
DX: namespace casing

kubawerlos:
Just spotted this - not sure if it's worth to write an AutoReview test to prevent it in future.


keradus:
Thank you @kubawerlos.

Issue #4075, opened 2018-11-01
HeaderComment - tool generates invalid PHP


SpacePossum:
#### The PHP version you are using (`$ php -v`):
PHP 7.2.11-2+ubuntu16.04.1+deb.sury.org+1 (cli) (built: Oct 15 2018 11:40:04) ( NTS )

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
2.12 / 2.13

#### The command you use to run PHP CS Fixer:
$ php-cs-fixer fix tmp/test2.php --dry-run -vvv --diff

#### The configuration file you are using, if any:
```php
<?php

$header = <<<'EOF'
/**
 * Foo
 */
EOF;

$finder = PhpCsFixer\Finder::create()
    ->in(__DIR__)
;

return PhpCsFixer\Config::create()
    ->setRules([
        'header_comment' => [
            'header' => $header,
            'location' => 'after_open',
            'separate' => 'bottom'
        ],
    ])
    ->setFinder($finder)
;
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
```php
<?php

```

 * with unexpected changes applied when running PHP CS Fixer:
[none]
```
$ php-cs-fixer fix tmp/test2.php --dry-run -vvv --diff
Loaded config default from "/home/possum/work/PHP-CS-Fixer/.php_cs.dist".
Using cache file ".php_cs.cache".
Paths from configuration file have been overridden by paths provided as command arguments.
E
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error

Checked all files in 0.006 seconds, 10.000 MB memory used

Files that were not fixed due to errors reported during linting after fixing:
   1) /home/possum/work/PHP-CS-Fixer/tmp/test2.php

                                                                                         
        [PhpCsFixer\Linter\LintingException]                                             
        PHP Parse error: syntax error, unexpected '*', expecting end of file on line 6.  
                                                                                         

      Applied fixers: header_comment
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
 <?php
+/*
+ * /**
+ *  * Foo
+ *  */
+ */
+

```

 * with the changes you expected instead:
```php
<?php
/**
 * Foo
 */

```

--- Update ---

Clearly this is an user error for providing the header as PHPDoc already. 
However we could detect this case during configuration of the rule and throw a nice exception message and not crash during fixing.

[...]

PR #4099, opened 2018-11-13
HeaderCommentFixer - throw exception on invalid header configuration

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4075

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -13,6 +13,7 @@
 namespace PhpCsFixer\Fixer\Comment;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
 use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
 use PhpCsFixer\Fixer\WhitespacesAwareFixerInterface;
 use PhpCsFixer\FixerConfiguration\AliasedFixerOptionBuilder;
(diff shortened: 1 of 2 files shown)


kubawerlos:
What about `comment` type? It also cannot have `*/`.

SpacePossum:
thanks @kubawerlos , updated :+1:

src/Fixer/Comment/HeaderCommentFixer.php
```diff
@@ -145,14 +146,20 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
             (new FixerOptionBuilder('header', 'Proper header content.'))
                 ->setAllowedTypes(['string'])
-                ->setNormalizer(static function (Options $options, $value) {
+                ->setNormalizer(static function (Options $options, $value) use ($fixerName) {
                     if ('' === trim($value)) {
                         return '';
                     }
 
+                    if (false !== strpos($value, '*/')) {
+                        throw new InvalidFixerConfigurationException($fixerName, 'Cannot use \'*/\' in header.');
```
kubawerlos: Can't `$this->getName()` be instead of `$fixerName` here?

dmvdbrugge: I first thought the same, however, this is a static closure, thus not bound to `$this`. See http://php.net/manual/en/functions.anonymous.php#functions.anonymous-functions.static

kubawerlos: Do we need this closure to be static?

Issue #3767, opened 2018-05-19
`(unset)` cast to `= null` fixer


ntzm:
```diff
-$b = (binary) $a;
+$b = (string) $a;
```

```diff
-$b = (unset) $a;
+$b = null;
```

`(unset)` is deprecated as of PHP 7.2 though

keradus:
sounds like 2 different logics for 2 different rules

ntzm:
Case 1 is handled in #3780

PR #4102, opened 2018-11-13
Add NoUnsetCastFixer

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3767

```
$ php php-cs-fixer describe no_unset_cast
Description of no_unset_cast rule.
Using `$foo = (unset) $bar` must not be used, but `$foo = null;`.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,2 +1,2 @@
    <?php
   -$a = (unset) $b;
   +$a =  null;
   
```

### DIFF
--- a/src/Fixer/CastNotation/NoUnsetCastFixer.php
+++ b/src/Fixer/CastNotation/NoUnsetCastFixer.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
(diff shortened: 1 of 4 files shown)


src/Fixer/CastNotation/NoUnsetCastFixer.php
```diff
@@ -0,0 +1,86 @@
...
+ */
+final class NoUnsetCastFixer extends AbstractFixer
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Using `$foo = (unset) $bar` must not be used, but `$foo = null;`.',
```
kubawerlos: > **Using** ... must not be **used**

To not repeat "using"/"used" maybe drop "Using"?

SpacePossum:
Suggestions for adding this fixer to a set?

keradus:
Thank you @SpacePossum.

PR #4103, opened 2018-11-13
NoExtraBlankLinesFixer - fix candidate detection

SpacePossum:
technically a bug, however the logic effected is for fast checking for candidate, so in the end the effect of the bug is "only" slower code


SpacePossum:
I was looking to prove the issue, but I couldn't really. Looking at 2.12 we use 
>$tokens->getNextTokenOfKind($index, [';', [T_CLOSE_TAG]]);

and

>$siblingIndex = $tokens->getNextTokenOfKind($siblingIndex, [[T_COMMENT]]);

so maybe there are more oddities (like unnecessary wrapping into an array). However the fix here is that `T_CLOSE_TAG` is never a match IIUC.

PR #4104, opened 2018-11-13
Change transformers order, fixing untransformed T_USE

### LINKED ISSUE #4086: NoUnusedImportsFixer removes use trait

dmvdbrugge:
Fixes #4086


tests/AutoReview/TransformerTest.php
```diff
@@ -63,4 +64,40 @@ public function provideTransformerCases()
...
+        $this->assertLessThan(
+            $first->getPriority(),
+            $second->getPriority(),
+            sprintf('"%s" should have less priority than "%s"', \get_class($second), \get_class($first))
+        );
+    }
+
+    public function provideTransformerPriorityCases()
+    {
+        if (\PHP_VERSION < 70100) {
```
keradus: please drop it, as you already used `@requires PHP 7.1` annotation

keradus:
please, add code sample provided in #4086 to `NoUnusedImportsFixerTest.php` as well, that way:
- we will have a proof that fix to Transformer fixes the Fixer
- when one would change Transformer logic, we will be sure that Fixer still works properly

tests/AutoReview/TransformerTest.php
```diff
@@ -63,4 +64,40 @@ public function provideTransformerCases()
 
         return $transformersArray;
     }
+
+    /**
+     * @requires PHP 7.1
+     *
+     * @dataProvider provideTransformerPriorityCases()
```
keradus: I don't believe `dataProvider` annotation was intended to be used with parenthesis
```php
public const REGEX_DATA_PROVIDER = '/@dataProvider\s+([a-zA-Z0-9._:-\\\\x7f-\xff]+)/';
```

and https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.data-providers

tests/AutoReview/TransformerTest.php
dmvdbrugge: Sadly this is required, because there is something (might be phpunit itself) that calls dataproviders, even if they are unused. And because not all transformers are available, the "big list" would error, but I wanted to maintain the same format as for the fixer priority cases.

src/Tokenizer/Transformers.php
```diff
@@ -64,8 +64,8 @@ public static function create()
      */
     public function transform(Tokens $tokens)
     {
-        foreach ($tokens as $index => $token) {
-            foreach ($this->items as $transformer) {
+        foreach ($this->items as $transformer) {
+            foreach ($tokens as $index => $token) {
```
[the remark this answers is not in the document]
dmvdbrugge: You are partially correct, however the priority already suggests one transformer should run before another, which is now even better enforced.

dmvdbrugge:
> please, add code sample (...) to `NoUnusedImportsFixerTest.php` as well

Done, and I also added it to `OrderedClassElementsFixerTest` as that suffers from the same issue (which explains "_Weird thing 2_").

tests/AutoReview/TransformerTest.php
keradus: then, why we don't need condition like that with all `@requires PHP x.y` we have in other places ?

keradus:
for next time - please don't squash commits of yours,
now, one needs to re-review everything, instead of just few fixes you made according to comments :(

tests/AutoReview/TransformerTest.php
```diff
@@ -63,4 +64,40 @@ public function provideTransformerCases()
 
         return $transformersArray;
     }
+
+    /**
+     * @requires PHP 7.1
```
keradus: why we don't want to test if priority is set properly below 7.1 ?
like, if I want to run tool at PHP 7.0, I still want transformers to be applied in proper order, right ?

tests/Tokenizer/TransformersTest.php
```diff
@@ -0,0 +1,73 @@
...
+    public function testUsingTrait()
+    {
+        $this->testTraitFunction();
+    }
+}
+
+SOURCE
+                ,
+                46,
+                CT::T_USE_TRAIT,
```
keradus: let's do `[46 => CT::T_USE_TRAIT]` instead, so we will be able to check more indexes in single test

tests/AutoReview/TransformerTest.php
dmvdbrugge: Because of the way the test currently gets the transformers, which I just reused. I can probably rewrite that.

tests/AutoReview/TransformerTest.php
dmvdbrugge: It's not about testing if certain functionality works at a certain version, it's just a requirement for the test itself because of the way it works, priority shouldn't change with versions anyway.

However like I said, I can probably rework this.

tests/Tokenizer/TransformersTest.php
dmvdbrugge: Ha, I had thought of that, but didn't think you would approve. I'll gladly change this 😄

dmvdbrugge:
> for next time - please don't squash commits of yours,
> now, one needs to re-review everything, instead of just few fixes you made according to comments :(

My apologies, I had misinterpreted some earlier comments in this repo. I usually don't even do it this way. Won't happen again.

keradus:
don't worry, nothing big (squashing or not). sometimes, when we can't split ourselves, we ask original creator to squash, but it is very rarely happening

src/Tokenizer/Transformers.php
```diff
@@ -64,9 +64,11 @@ public static function create()
      */
     public function transform(Tokens $tokens)
     {
-        foreach ($tokens as $index => $token) {
-            foreach ($this->items as $transformer) {
-                $transformer->process($tokens, $token, $index);
+        foreach ($this->items as $transformer) {
+            if (\PHP_VERSION_ID >= $transformer->getRequiredPhpVersionId()) {
```
keradus: please, move it back where it belongs.

if this check would be in `registerTransformer` - it's executed `once per transformer`
if it would be here - it's executed `once per transforer x per file` :(

dmvdbrugge: This is consistent with the fixers though, which do the same. But I'll move it back if you wish. That however will convolute the tests again.

keradus:
I let myself do some dirty things with this PR, now CI shall be green, let's wait and see

PR #4105, opened 2018-11-13
change classy to class in a few places

codysnider:


README.rst
```diff
@@ -339,7 +339,7 @@ Choose from the list of available rules:
     structures; defaults to ``'same'``
   - ``position_after_functions_and_oop_constructs`` (``'next'``, ``'same'``): whether
     the opening brace should be placed on "next" or "same" line after
-    classy constructs (non-anonymous classes, interfaces, traits, methods
+    class constructs (non-anonymous classes, interfaces, traits, methods
```
keradus: `class constructs (non-anonymous classes, interfaces, traits, ...)`

interface is not `class`, it's class-alike, thus `classy`

SpacePossum:
Hi @codysnider,

In the places where `classy` is used here we mean more than just `class` (like `interfaces` and `traits` as well), therefore I'm closing this PR. Thanks for the suggestion through PR though ! :)

PR #4093, opened 2018-11-12
[Preg:: issue] udiff output fix

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4082

PR #4107, opened 2018-11-14
Preg::split - fix for non-UTF8 subject

kubawerlos:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4093

### DIFF
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -36,12 +36,12 @@ final class Preg
     public static function match($pattern, $subject, &$matches = null, $flags = 0, $offset = 0)
     {
         $result = @preg_match(self::addUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
-        if (false !== $result) {
+        if (false !== $result && PREG_NO_ERROR === preg_last_error()) {
             return $result;
         }
 
         $result = @preg_match(self::removeUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
-        if (false !== $result) {
+        if (false !== $result && PREG_NO_ERROR === preg_last_error()) {
             return $result;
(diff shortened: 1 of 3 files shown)


ntzm:
Shall we use `PREG_NO_ERROR` instead of `0`? More verbose and catches if it's redefined it a later release of PHP (highly unlikely ;) )

kubawerlos:
Ok, now I need some help, I cannot reproduce the errors, all tests are passing for me.

kubawerlos:
Okay, found the reason: https://bugs.php.net/bug.php?id=74183

See the difference: https://3v4l.org/eF5DO#v7032

src/Preg.php
```diff
@@ -36,12 +36,12 @@
     public static function match($pattern, $subject, &$matches = null, $flags = 0, $offset = 0)
     {
         $result = @preg_match(self::addUtf8Modifier($pattern), $subject, $matches, $flags, $offset);
-        if (false !== $result) {
+        if (false !== $result && PREG_NO_ERROR === preg_last_error()) {
```
ntzm: Perhaps this logic can be split into a private method like `isErroneousResult($result)`?

kubawerlos: I was thinking about it, but `$result` would have to be compared sometimes to `false` and sometimes to `null` - having both in the private method would be confusing for the reader and having another argument in method call was looking less readable either.

ntzm: Makes sense

SpacePossum:
Thank you @kubawerlos.

SpacePossum:
Thanks @ostrolucky  as well :)

PR #4109, opened 2018-11-15
NoBlankLines*: fix removing lines consisting only of spaces

kubawerlos:
We have 3 `NoBlankLines*` fixers and they behave differently for lines consisting only of spaces:
- `NoBlankLinesAfterClassOpeningFixer` - removes line of spaces only when next line is indented
- `NoBlankLinesAfterPhpdocFixer` - same as above
- `NoBlankLinesBeforeNamespaceFixer` - removes indentation of the next line

What would be the proper behaviour (apart of removing indentation of next line what is obviously a bug) - to keep lines consisting only of spaces or to remove them?

_Tests are to demonstrate the problem, obviously have to be fixed as well._


kubawerlos:
By making a proxy for `2.12` and deprecation for `master`?

dmvdbrugge:
> how about merging those into `no_extra_blank_lines` ?

There's a difference between "no extra" (meaning at most 1) or "no" (meaning 0).

I'd be for combining the ones pointed out into a single fixer (just like the `blank_line_before_statement`), however think that it should be a new one. (With the deprectations ofc.)

SpacePossum:
if I expect 0 than 1 is `extra` to me

dmvdbrugge:
But how are you going to configure where you expect zero and where one? Sounds like configuration hell. I for one, wouldn't want `NoBlankLinesBeforeNamespaceFixer` while I do want the others.

kubawerlos:
Okay, as there obviously going to be discussion about new fixer to remove blank lines let's focus here about fixing the bugs only.

tests/Fixtures/Integration/priority/no_whitespace_in_blank_line,no_blank_lines_after_phpdoc.test
```diff
@@ -1,22 +0,0 @@
---TEST--
```
keradus: please do not remove tests.
ever

keradus:
Thank you @kubawerlos.

tests/Fixtures/Integration/priority/no_whitespace_in_blank_line,no_blank_lines_after_phpdoc.test
kubawerlos: Oh, yeah, I haven't thought about moving it from `priority` to `misc`. Next time I will remember.

PR #4112, opened 2018-11-19
NoSuperfluousPhpdocTagsFixer - Add missing code sample, groom tests

keradus:
Appendix for #4002

### DIFF
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -42,18 +42,29 @@ final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements Config
 class Foo {
     /**
      * @param Bar $bar
+     * @param mixed $baz
      */
-    public function doFoo(Bar $bar) {}
+    public function doFoo(Bar $bar, $baz) {}
 }
 '),
+                new CodeSample('<?php
+class Foo {
+    /**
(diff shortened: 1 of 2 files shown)


tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -102,6 +104,18 @@ class Foo {
      */
     public function doFoo($bar) {}
 }',
+                ['allow_mixed' => false],
+            ],
+            'allow_mixed=>true' => [
+                '<?php
+class Foo {
+    /**
+     *
```
julienfalque: The `@param mixed $bar` annotation is missing here (to prove the fixer doesn't remove it). Maybe a `@return mixed` would be nice too.

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -42,18 +42,29 @@ public function getDefinition()
...
-    public function doFoo(Bar $bar) {}
+    public function doFoo(Bar $bar, $baz) {}
 }
 '),
+                new CodeSample('<?php
+class Foo {
+    /**
+     * @param Bar $bar
+     * @param mixed $baz
+     */
```
julienfalque: What about adding a `@return mixed` annotation?

SpacePossum:
Thank you @keradus.

PR #4113, opened 2018-11-20
OrderedInterfacesFixer - Introduction

### LINKED ISSUE #4108: [Feature request] Order of class interfaces
Having the following input:

```php
<?php

class Example implements CheckInterface, AuthInterface {}
```

a `ordered_implements` rule should order the import alphanumerically to achieve the following output:

```php
<?php

class Example implements AuthInterface, CheckInterface {}
```

Other orders could be `length` and `none` (identical to `ordered_imports`).

This rule should not be risky, as interfaces cannot overwrite each other when being implemented.

dmvdbrugge:
Fixes #4108

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1241,6 +1241,19 @@ Choose from the list of available rules:
     should be sorted alphabetically or by length, or not sorted; defaults
     to ``'alpha'``; DEPRECATED alias: ``sortAlgorithm``
 
+* **ordered_interfaces**
+
+  Orders the interfaces in an ``implements`` or ``interface extends`` clause.
+
+  *Risky rule: risky for ``implements`` when specifying both an interface and its parent interface, because PHP doesn't break on ``parent, child`` but does on ``child, parent``.*
+
+  Configuration options:
+
+  - ``direction`` (``'ascend'``, ``'descend'``): which direction the interfaces should
+    be ordered; defaults to ``'ascend'``
+  - ``order`` (``'alpha'``, ``'length'``): how the interfaces should be ordered;
+    defaults to ``'alpha'``
+
 * **php_unit_construct** [@Symfony:risky, @PhpCsFixer:risky]
 
   PHPUnit assertion method calls like ``->assertSame(true, $foo)`` should be
--- a/src/Fixer/ClassNotation/OrderedInterfacesFixer.php
+++ b/src/Fixer/ClassNotation/OrderedInterfacesFixer.php
@@ -1,0 +1,231 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ClassNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Dave van der Brugge <dmvdbrugge@gmail.com>
+ */
+final class OrderedInterfacesFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /** @internal */
+    const OPTION_DIRECTION = 'direction';
+
+    /** @internal */
+    const OPTION_ORDER = 'order';
+
+    /** @internal */
+    const DIRECTION_ASCEND = 'ascend';
+
+    /** @internal */
+    const DIRECTION_DESCEND = 'descend';
+
+    /** @internal */
+    const ORDER_ALPHA = 'alpha';
+
+    /** @internal */
+    const ORDER_LENGTH = 'length';
+
+    /**
+     * Array of supported directions in configuration.
+     *
+     * @var string[]
+     */
+    private $supportedDirectionOptions = [
+        self::DIRECTION_ASCEND,
+        self::DIRECTION_DESCEND,
+    ];
+
+    /**
+     * Array of supported orders in configuration.
+     *
+     * @var string[]
+     */
+    private $supportedOrderOptions = [
+        self::ORDER_ALPHA,
+        self::ORDER_LENGTH,
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Orders the interfaces in an `implements` or `interface extends` clause.',
+            [
+                new CodeSample(
+                    "<?php\n\nfinal class ExampleA implements Gamma, Alpha, Beta {}\n\ninterface ExampleB extends Gamma, Alpha, Beta {}\n"
+                ),
+                new CodeSample(
+                    "<?php\n\nfinal class ExampleA implements Gamma, Alpha, Beta {}\n\ninterface ExampleB extends Gamma, Alpha, Beta {}\n",
+                    [self::OPTION_DIRECTION => self::DIRECTION_DESCEND]
+                ),
+                new CodeSample(
+                    "<?php\n\nfinal class ExampleA implements MuchLonger, Short, Longer {}\n\ninterface ExampleB extends MuchLonger, Short, Longer {}\n",
+                    [self::OPTION_ORDER => self::ORDER_LENGTH]
+                ),
+                new CodeSample(
+                    "<?php\n\nfinal class ExampleA implements MuchLonger, Short, Longer {}\n\ninterface ExampleB extends MuchLonger, Short, Longer {}\n",
+                    [
+                        self::OPTION_ORDER => self::ORDER_LENGTH,
+                        self::OPTION_DIRECTION => self::DIRECTION_DESCEND,
+                    ]
+                ),
+            ],
+            null,
+            "Risky for `implements` when specifying both an interface and its parent interface, because PHP doesn't break on `parent, child` but does on `child, parent`."
+        );
+    }
+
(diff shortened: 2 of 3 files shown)


src/Fixer/ClassNotation/OrderedImplementsFixer.php
```diff
@@ -0,0 +1,208 @@
...
+
+    /** @internal */
+    const ORDER_LENGTH = 'length';
+
+    /**
+     * Array of supported directions in configuration.
+     *
+     * @var string[]
+     */
+    private $supportedDirectionOptions = [
```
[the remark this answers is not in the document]
dmvdbrugge: Would've been a `private const` if the language level accepted that. Does that change your opinion?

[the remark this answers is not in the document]
julienfalque: Not really. Having the arrays defined in `setAllowedValues()` directly would be enough to indicate their meaning IMO. Anyway if you really prefer keeping those properties, I'm fine :)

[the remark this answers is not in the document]
dmvdbrugge: In that case I'm keeping, to remain consistent with both `OrderedClassElementsFixer` and `OrderedImportsFixer` (which is why it was this way to begin with).

README.rst
```diff
@@ -1213,6 +1213,17 @@ Choose from the list of available rules:
   - ``sortAlgorithm`` (``'alpha'``, ``'none'``): how multiple occurrences of same type
     statements should be sorted; defaults to ``'none'``
 
+* **ordered_implements**
+
+  Orders the interfaces in an ``implements`` clause.
```
SpacePossum: would sorting multiple `extends` of an interface very similar to the logic in this fixer?

tests/Fixer/ClassNotation/OrderedImplementsFixerTest.php
```diff
@@ -0,0 +1,159 @@
...
+            'multiple' => [
+                '<?php class A implements MuchLonger, Longer, Short {}',
+                '<?php class A implements MuchLonger, Short, Longer {}',
+            ],
+            'mixed' => [
+                '<?php class T implements Partially\Q\C\N, /* Who mixes these? */ UnNamespaced, \F\Q\C\N {}',
+                '<?php class T implements /* Who mixes these? */ UnNamespaced, \F\Q\C\N, Partially\Q\C\N {}',
+            ],
+        ];
+    }
```
SpacePossum: please add some tests with nested anonymous classes

README.rst
dmvdbrugge: Would need additional checks to see if it applies and where to fix. (Aka `isCandidate` and the current first lines of `applyFix`.)

The rest of the logic should be the same indeed.

And now you gonna ask me to add that as well? 😜

SpacePossum: it can wait, after this one maybe some abstraction can be done to create an additional fixer so we would have both ;)

tests/Fixer/ClassNotation/OrderedImplementsFixerTest.php
dmvdbrugge: 👍 Done

README.rst
dmvdbrugge: Code-wise my preference would be separate. However as a user, I think single fixer would be better. Would you do one without doing the other? But giving the option is nice as well.

I can see it as both a separate fixer or as part of this one, so I'll leave that up to you guys to decide.

julienfalque: I would go with a single fixer, but that means implementing it in this PR as it would probably require to rename the fixer.

SpacePossum: single fixer makes sense to me :)
that would mean:
- rename the fixer
- candidate detection would become 'implements' or 'extends && interface' is found
- update the check if candidate token in the fix method
- I wouldn't update the config, I don't see why someone would sort `implements` and `extends` differently, but sure up for discussion ;)

tests/Fixer/ClassNotation/OrderedImplementsFixerTest.php
SpacePossum: thanks :)

src/Fixer/ClassNotation/OrderedImplementsFixer.php
```diff
@@ -0,0 +1,208 @@
...
+                    $normalized .= str_replace('\\', ' ', $token->getContent());
+                    ++$actualInterfaceIndex;
+                }
+
+                $interfaces[$interfaceIndex]['normalized'] = $normalized;
+                $interfaces[$interfaceIndex]['originalIndex'] = $interfaceIndex;
+            }
+
+            usort($interfaces, function (array $first, array $second) {
+                $score = self::ORDER_LENGTH === $this->configuration[self::OPTION_ORDER]
```
SpacePossum: if sorting on length on normalized data, would '\A\B\C' not considered less of length than 'ABCD' ?

tests/Fixer/ClassNotation/OrderedImplementsFixerTest.php
```diff
@@ -0,0 +1,197 @@
...
+            ],
+            'FQCN' => [
+                '<?php class T implements \F\Q\C\N, \F\Q\I\N {}',
+                '<?php class T implements \F\Q\I\N, \F\Q\C\N {}',
+            ],
+            'mixed' => [
+                '<?php class T implements \F\Q\C\N, Partially\Q\C\N, /* Who mixes these? */ UnNamespaced {}',
+                '<?php class T implements /* Who mixes these? */ UnNamespaced, \F\Q\C\N, Partially\Q\C\N {}',
+            ],
+            'nested anonymous classes' => [
```
SpacePossum: since we still support >PHP7 on master I'm afraid you have to move this to separate test and provide for PHP7 and up :(

tests/Fixer/ClassNotation/OrderedImplementsFixerTest.php
```diff
@@ -0,0 +1,197 @@
...
+                "<?php class T implements\n// Here's B\nB,\n// Here's A\nA\n{}",
+            ],
+            'no whitespace' => [
+                '<?php class T implements/*An*/AnInterface/*end*/,/*Second*/SecondInterface{}',
+                '<?php class T implements/*Second*/SecondInterface,/*An*/AnInterface/*end*/{}',
+            ],
+            'FQCN' => [
+                '<?php class T implements \F\Q\C\N, \F\Q\I\N {}',
+                '<?php class T implements \F\Q\I\N, \F\Q\C\N {}',
+            ],
```
SpacePossum: Since you are not iterating the tokens backwards and yet changing the set, I would like a test with multiple fix cases in one file.
For example sorting alpha:
```
<?php
class A1 extends Z\X\Y, A\B\C {
}

class B2 extends Z\X, A\B {
}

class C3 extends Z\X, A {
}

class D4 extends Z\X\V, B\V, A\B {
}

class E5 extends Z\X\V, Y\V, X\B, U\B {
}
```

dmvdbrugge: There already are multiple cases in the nested anonymous classes test.

src/Fixer/ClassNotation/OrderedImplementsFixer.php
dmvdbrugge: If you're looking at actual length, `A\B\C\D` **is longer** than `Abcde` so it should be marked as longer, normalised or not.

Also, I looked at `OrderedImportsFixer` for inspiration, which handles it the same.

README.rst
dmvdbrugge: > single fixer makes sense to me :)
> that would mean:
 
- [X] rename the fixer
- [X] candidate detection would become 'implements' or 'extends && interface' is found
- [X] update the check if candidate token in the fix method
- [X] I wouldn't update the config
- [X] update code samples
- [X] update tests
- [X] update readme
- [ ] push commit

src/Fixer/ClassNotation/OrderedInterfacesFixer.php
```diff
@@ -0,0 +1,221 @@
...
+            if (!$token->isGivenKind(T_IMPLEMENTS)) {
+                if (!$token->isGivenKind(T_EXTENDS)) {
+                    continue;
+                }
+
+                $nameTokenIndex = $tokens->getPrevMeaningfulToken($index);
+                $interfaceTokenIndex = $tokens->getPrevMeaningfulToken($nameTokenIndex);
+                $interfaceToken = $tokens[$interfaceTokenIndex];
+
+                if (null === $interfaceToken || !$interfaceToken->isGivenKind(T_INTERFACE)) {
```
SpacePossum: I think `null === $interfaceToken` is never true here?

keradus:
I let myself resolve the conflicts, readme was not regenerated after changes

dmvdbrugge:
I will mark as risky. However, in the case of our own code base, isn't it bad practise anyway to have this interface-duplication? I mean, `DefinedFixerInterface extends FixerInterface`, why `implements` them both? (Also: why does php break one way but not the other? Strange...)

SpacePossum:
>isn't it bad practise anyway to have this interface-duplication? 

it doesn't harm to be explicit, it could also be because it makes merging over the various branches more easy, but maybe it is (indeed) overcomplete and can be cleaned up some day in another PR

>why does php break one way but not the other? Strange...

I have no idea, it really puzzles me!, sadly it does so please update to be `risky`

keradus:
I wouldn't consider it a bad idea. If you want to implement interfaces A and B, you may not know, not control, or even not care internal detail that B also implement A under the hood. Yet, this rule can break the code in current PHP engine, so we need risky flag with good explanation to the user why is it so.

src/Fixer/ClassNotation/OrderedInterfacesFixer.php
keradus: @dmvdbrugge , can you answer this as well, please ?

dmvdbrugge: I see we don't start fixing files that are invalid to begin with, so no this will never be null. I'll remove the check.

dmvdbrugge:
@keradus all changes have been processed (like, 3 months ago 😜) could you update your review?

SpacePossum:
Thank you @dmvdbrugge.

keradus:
> all changes have been processed (like, 3 months ago stuck_out_tongue_winking_eye) could you update your review?

post mortem, @dmvdbrugge - pushing fixes to PR doesn't mark the PR as changed for repo maintainer (or at least to me). without new comment (!= commit), it's hard to notice that sth changed

Issue #4057, opened 2018-10-20
VisibilityRequiredFixer removes comments between access modifier and variable name


pcf0:
#### The PHP version you are using (`$ php -v`):
=> PHP 7.2.11

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.13.0

#### The command you use to run PHP CS Fixer:
=> php-cs-fixer fix --rules visibility_required -- Test.php

#### The configuration file you are using, if any:
no config

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php
declare(strict_types=1);

class Test
{
  private /* comment */ $private = 'private';
};
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php
declare(strict_types=1);

class Test
{
  private $private = 'private';
};
```

 * with the changes you expected instead:
no change, comment not removed

keradus:
btw, what is the purpose of such comment ?

PR #4114, opened 2018-11-21
VisibilityRequiredFixer - don't remove comments

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4057

### DIFF
--- a/tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
+++ b/tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
@@ -532,9 +532,11 @@ const F=1; }',
     {
         $expected = '<?php
 class A
-{#
-public static function#
-AB#
+{# We will have a function below
+# It will be static
+# and awesome
+public static function# <- this is the function
+AB# <- this is the name
 (#
 )#
 {#
@@ -544,11 +546,11 @@ AB#
 
         $input = '<?php
 class A
-{#
-static#
-#
-function#
-AB#
+{# We will have a function below
+static# It will be static
+# and awesome
+function# <- this is the function
+AB# <- this is the name
 (#
 )#
 {#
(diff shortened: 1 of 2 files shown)


tests/Fixer/ClassNotation/VisibilityRequiredFixerTest.php
```diff
@@ -595,4 +597,91 @@ function A()
...
+     */
+    public function testKeepingComment()
+    {
+        $this->fixer->configure(['elements' => ['property', 'method', 'const']]);
+
+        $this->doTest(
+            '<?php
+                class Foo
+                {
+                    /* constant */ private const BAR = 3;
```
julienfalque: Why swapping the comment with `private`?

kubawerlos: Because of the checking in function `isKeywordPlacedProperly`.

Do you think I should change it to:
```php
    private function isKeywordPlacedProperly(Tokens $tokens, $keywordIndex, $comparedIndex)
    {
        for ($index = $comparedIndex - 1; $index > $keywordIndex; --$index) {
            if (!$tokens[$index]->isGivenKind([T_COMMENT, T_DOC_COMMENT, T_WHITESPACE])) {
                return false;
            }
            if (1 === Preg::match('/\R/', $tokens[$index]->getContent())) {
                return false;
            }
        }

        return true;
    }
```
?

julienfalque: I don't know the implementation so I cannot answer this question properly for now. All I'm saying is that having such change when running the tool on a codebase would be quite unexpected.

kubawerlos: You can see `isKeywordPlacedProperly` function in this PR - I've extracted the logic to it for easier understanding.

> having such change when running the tool on a codebase would be quite unexpected.

Would it be? Current behaviour is removing such comment anyway.

keradus: it's better to move the comment rather than dropping it, for sure :+1: 
it's not perfect indeed, I agree, yet input code is not perfect as well, and we are making it slightly better

keradus:
Thank you @kubawerlos.

PR #4118, opened 2018-11-23
Fixed brew install command

sarelvdwalt:
`brew upgrade php-cs-fixer` won't work on fresh install, it will yield: `Error: php-cs-fixer not installed`
Fixed to be `brew install php-cs-fixer`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -131,7 +131,7 @@ You can update ``php-cs-fixer`` through this command:
 
 .. code-block:: bash
 
-    $ brew upgrade php-cs-fixer
+    $ brew install php-cs-fixer
 
 Locally (PHIVE)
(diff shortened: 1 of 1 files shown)


SpacePossum:
Hi @sarelvdwalt

Thanks for the suggestion :)
You are right that "brew upgrade php-cs-fixer won't work on fresh install," . However in the `readme` under "Installation" it states `$ brew install php-cs-fixer`, your PR tries to change the instructions under "Update". I think the current doc is ok, but not 100% sure as I haven't use brew for a bit.

sarelvdwalt:
Ah golly, sorry I was being daft. Sorry I'll close the PR :)

PR #4119, opened 2018-11-25
PhpUnitNoExpectationAnnotationFixer - fix extracting content from annotation

kubawerlos:
PHPUnit [regex](https://github.com/sebastianbergmann/phpunit/blob/7.4.4/src/Util/Test.php#L66) does not care about content before `@expectedException`.

_And yes, I have found such case, not invented it by myself ;)_


tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
```diff
@@ -546,6 +546,33 @@ public function testMessageOnMultilinesWithAnotherSpaceAndTag()
...
+
+            aaa();
+        }
+    }',
+                '<?php
+    final class MyTest extends \PHPUnit_Framework_TestCase
+    {
+        /**
+         * Double "@" is/was below
+         * @@expectedException FooException
```
keradus: What if we would have sth like
```php
/**
 * We are providing invalid input, for that we @expectedException FooException
*/
```

?

I would like to not drop the meaningful part of comment

kubawerlos: Noting, fixer won't touch this line as first we use `DocBlock::getAnnotationsOfType` and it requires `@` to be start of the comment/line.

I'd say let's take care of this case in another PR - I can do it right after this one is merged, would that be ok?

keradus: then, let's add the case in this PR to prove it's not modified, as you said.

i just want to be sure that we won't change behaviour of fixer for that case after some refactoring that will happen one day...

kubawerlos: Done

Issue #2103, opened 2016-07-30
multiple_use analogue for traits


Strate:
As for now, therer is seems no rule, which forces each use of trait be on separate line:

``` php
class Some
{
  use SomeTrait, OtherTrait;  
}
```

should be reformatted by the rule to:

``` php
class Some
{
  use SomeTrait;  
  use OtherTrait;
}
```

SpacePossum:
indeed, check 1.12 and master and both don't do that, nice FR!

phansys:
I think it could be added as an option at "single_class_element_per_statement" fixer.

SpacePossum:
or maybe `single_import_per_statement`? not use....

keradus:
`single_class_element_per_statement`
trait import is not element of class. it's an instruction to import some elements

SpacePossum:
check and agreed :)
`single_class_element_per_statement` it is, anyone got time to pick this up?

keradus:
I think you not understand my comment in a way I hope to express, sorry.
I'm against making it in `single_class_element_per_statement`, as import trait statement is not element of class itself.

SpacePossum:
I did misread indeed,  new fixer than?

keradus:
I would say that having some other opinions would be great ;)

keradus:
@Strate are you interested in creating PR for this feature request ?

Strate:
@keradus I hope, but I am not sure that I'll have enought spare time for this. So, let keep it vacant :)

PR #4121, opened 2018-11-27
SingleTraitInsertPerStatementFixer - Introduction

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2103

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1704,6 +1704,10 @@ Choose from the list of available rules:
   - ``strings_containing_single_quote_chars`` (``bool``): whether to fix
     double-quoted strings that contains single-quotes; defaults to ``false``
 
+* **single_trait_insert_per_statement** [@Symfony, @PhpCsFixer]
+
+  Each trait ``use`` must be done as single statement.
+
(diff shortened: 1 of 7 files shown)


ntzm:
Not sure about the name, `single_trait_import_per_statement` or `single_trait_use_per_statement` makes more sense imo

SpacePossum:
I looked at the [docs](http://php.net/manual/en/language.oop5.traits.php) for the correct terminology but didn't found anything concluding, so all names are fine with me :)

gharlan:
Please add some tests with `use` statements with `{ }`, like this (example from http://php.net/manual/de/language.oop5.traits.php#example-214):

```php
class Talker {
    use A, B {
        B::smallTalk insteadof A;
        A::bigTalk insteadof B;
    }
}
```

Do we want to change this one? 
It seems to work: https://3v4l.org/oGJJr, but I would prefer to not change it.

SpacePossum:
Suggestions for adding this fixer to a set?

PR #4123, opened 2018-11-27
Added support of PHP 7.3

sergeyklay:


keradus:
Hi ! First of all, thanks for collaborating !

Indeed, PHP CS Fixer is not yet compatible with PHP 7.3,
as each minor release of PHP introduces changes in syntax, for regular users it's a new feature, yet for us, who operate on low-level of tokenization, this is BC breaker, and without adjusting the project it will crash on trying to fix 7.3 code.

Just allowing to run on PHP 7.3 is not enough. We can run tool on PHP 7.3 to fix code written in php 7.2-, but we don't understand code written in PHP 7.3.
For that, I'm closing this PR.

If you want to help, please visit https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3697

----------------

You can still enforce composer to install PHP CS Fixer on PHP 7.3 anyway with `--ignore-platform-reqs`.
Then, you can set up `PHP_CS_FIXER_IGNORE_ENV` env var to enforce PHP CS Fixer to run under PHP 7.3, but you may experience unexpected issues.

Issue #3553, opened 2018-02-20
Add lowercase hints fixer


Slamdunk:
```diff
-function foo(sTrIng $str): sTrIng {}
+function foo(string $str): string {}

-function foo(ITERABLE $arr): ITERABLE {}
+function foo(iterable $arr): iterable {}

-function foo(Int $num): Int {}
+function foo(int $num): int {}

-function foo(fLOAT $num): fLOAT {}
+function foo(float $num): float {}
```

julienfalque:
Maybe in type castings as well:
```diff
-$foo = (sTrInG) $foo;
+$foo = (string) $foo;
```

Slamdunk:
There is already https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/src/Fixer/CastNotation/LowercaseCastFixer.php for that

julienfalque:
Ah right, I searched in the `Casing` namespace only :)

Then I would deprecate this one in favor of the new one, which would cover both type hints and type castings, WDYT?

Slamdunk:
I'm ok: before spending time on code I need @keradus feedback if he's ok with this approach.

SpacePossum:
better not call it `lowercase` something, as you could fix the casing of `stdClass` (and other predefined classes as well)

keradus:
If we are able to prepare general solution that will fix casing everywhere for built-in types, I'm :+1:
Like we do for true/false/null constants, we don't care where they are used, we always fix casing.

Slamdunk:
@SpacePossum this should be tagged with `has PR`

PR #4126, opened 2018-11-29
NativeFunctionTypeDeclarationCasingFixer - Introduction

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3553

```
$ php php-cs-fixer describe native_function_type_declaration_casing
Description of native_function_type_hint_casing rule.
Type hints (of functions) defined by PHP should be called using the correct casing.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,4 +1,4 @@
    <?php
   -function Foo(CALLABLE $bar) {
   +function Foo(callable $bar) {
        return 1;
    }
   
 * Example #2.
   --- Original
   +++ New
   @@ -1,5 +1,5 @@
    <?php
   -function Foo(INT $a): Bool
   +function Foo(int $a): bool
    {
        return true;
    }
   
 * Example #3.
   --- Original
   +++ New
   @@ -1,5 +1,5 @@
    <?php
   -function Foo(Iterable $a): VOID
   +function Foo(iterable $a): void
    {
        echo 'Hello world';
    }
   
 * Example #4.
   --- Original
   +++ New
   @@ -1,5 +1,5 @@
    <?php
   -function Foo(Object $a)
   +function Foo(object $a)
    {
        return 'hi!';
    }
   
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -929,6 +929,10 @@ Choose from the list of available rules:
   - ``strict`` (``bool``): whether leading ``\`` of function call not meant to have it
     should be removed; defaults to ``false``
 
+* **native_function_type_declaration_casing** [@Symfony, @PhpCsFixer]
+
+  Native type hints for functions should use the correct case.
+
 * **new_with_braces** [@Symfony, @PhpCsFixer]
 
   All instances created with new keyword must be followed by braces.
--- a/src/Fixer/Casing/NativeFunctionTypeDeclarationCasingFixer.php
+++ b/src/Fixer/Casing/NativeFunctionTypeDeclarationCasingFixer.php
@@ -1,0 +1,179 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Casing;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\TypeAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class NativeFunctionTypeDeclarationCasingFixer extends AbstractFixer
+{
+    /**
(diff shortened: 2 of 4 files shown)


ntzm:
Is this to work with class methods as well? Tests and description should be updated to reflect that if so.

src/Fixer/Casing/NativeFunctionTypeHintCasingFixer.php
```diff
@@ -0,0 +1,175 @@
...
+     * @param int    $index
+     */
+    private function fixFunctionArgumentTypes(Tokens $tokens, $index)
+    {
+        foreach ($this->functionsAnalyzer->getFunctionArguments($tokens, $index) as $argument) {
+            $this->fixArgumentType($tokens, $argument->getTypeAnalysis());
+        }
+    }
+
+    private function fixFunctionReturnType(Tokens $tokens, $index)
```
ntzm: Need phpdoc for `$index`

SpacePossum: :+1:   done

SpacePossum:
>Is this to work with class methods as well? Tests and description should be updated to reflect that if so.

I've updated some tests, I don't think the doc is off as methods are function (not always true the otherway around)

src/Fixer/Casing/NativeFunctionTypeHintCasingFixer.php
```diff
@@ -0,0 +1,179 @@
...
+        $this->functionsAnalyzer = new FunctionsAnalyzer();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Type hints (of functions) defined by PHP should be called using the correct casing.',
```
ntzm: I think this would read better:

```suggestion
            'Native type hints for functions should use the correct case.',
```

ntzm:
Oh one last thing from me, could you add a test for `interface` as well?

ntzm:
LGTM 👍

src/Fixer/Casing/NativeFunctionTypeHintCasingFixer.php
```diff
@@ -0,0 +1,179 @@
...
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\TypeAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class NativeFunctionTypeHintCasingFixer extends AbstractFixer
```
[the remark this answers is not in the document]
localheinz: Also see http://php.net/manual/en/functions.arguments.php#functions.arguments.type-declaration:

> **Note**:
> Type declarations were also known as type hints in PHP 5.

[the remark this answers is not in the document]
julienfalque: Oh I didn't know the PHP documentation officially renamed _type hints_ to _type declarations_. :+1: for reflecting that in PHP CS Fixer.

SpacePossum:
Suggestions for adding this fixer to a set? (maybe a PHP migration one, SF, projects one?)

SpacePossum:
maybe the sets `[@Symfony, @PhpCsFixer]` are OK? :)

PR #4127, opened 2018-11-29
LowercaseConstantsFixer - Fix case with properties using constants as their name

srathbone:


SpacePossum:
Thank you @srathbone.

PR #4128, opened 2018-11-30
DX: cleanup - AbstractFunctionReferenceFixer - use FunctionsAnalyzer

kubawerlos:
This is the last place to change for `FunctionsAnalyzer::isGlobalFunctionCall` unification.

### DIFF
--- a/src/AbstractFunctionReferenceFixer.php
+++ b/src/AbstractFunctionReferenceFixer.php
@@ -12,7 +12,7 @@
 
 namespace PhpCsFixer;
 
-use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer;
 use PhpCsFixer\Tokenizer\Tokens;
 
 /**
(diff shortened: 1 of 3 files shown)


SpacePossum:
Clean up looks good!
However I would like some tests for the `AbstractFunctionReferenceFixer::find` method (currently there are no dedicated tests for it). as to show what supports (global/no-global, namespaced/global, calling function/method, function signatures, stuff like that ;) )

kubawerlos:
Are you sure we want to have tests dedicated to this function? Now it is covered by 132 tests from different fixers. Also, we don't test other abstract classes.

To test it I'd have to create a dummy fixer in `PhpCsFixer\Tests\Fixtures` extending `AbstractFunctionReferenceFixer` and proxy `find` function, right?

SpacePossum:
yeah I was thinking of something like; https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4128/commits/a38dca7683e96bb0d16aaf66ba1c442c86e715d6

(feel free to push out again if you don't like it)

keradus:
Thank you @kubawerlos.

keradus:
Thank you @SpacePossum

Issue #4130, opened 2018-12-03
blank_line_after_namespace enforces zero intentation on line after blank line


InvisibleSmiley:
#### The PHP version you are using (`$ php -v`):
7.1.23

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
2.13.1

#### The command you use to run PHP CS Fixer:
vendor/bin/php-cs-fixer fix --dry-run /path/to/file.php -v --diff

#### The configuration file you are using, if any:
Custom, too long/irrelevant to post here. You only need to know two things:
1. Our code style requires one level of indentation for **everything** except for the opening tag
2. The test file becomes valid if I disable blank_line_after_namespace.

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php
    declare(strict_types=1);

    namespace My\NS;

    use Exception;

    /**
     * Example class
     *
     * @author Me
     */
    class X extends Y {

        /**
         * Example method
         *
         * @throws Exception
         * @return void
         */
        public function test(): void {
            throw new Exception();
        }

    }

```

 * with unexpected changes applied when running PHP CS Fixer:
```
-    use Exception;
+use Exception;
```
(When fixing this, please also mind/test the case without any "use" statements!)

 * with the changes you expected instead:
None

SpacePossum:
Hi!

Looking about your configuration you want the fixer to changes your code style to be compliant with PSR2.
As such, following the standard it will remove the indentation. The first sample of standard shows this is valid; https://www.php-fig.org/psr/psr-2/#11-example

InvisibleSmiley:
FYI I certainly do **not** want to apply PSR-2 or make our code compliant with it. We are using a completely custom coding standard, part of which is the requirement for one level of indentation of everything but the opening tag. We have tons (read: hunderds of thousands of lines) of existing code that is already compliant with said custom code style. We even created a custom php_codesniffer ruleset for it. I'm just trying to make PHP-CS-Fixer work with it, not change our standard.

The above is just an excerpt to show the particular issue I have.

SpacePossum:
misread indeed, sorry about that, I think your case might be related to this fix:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4109

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4132

PR #4132, opened 2018-12-03
BlankLineAfterNamespaceFixer - do not remove indent, handle comments

SpacePossum:
prove of https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4130

### DIFF
--- a/src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
+++ b/src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
@@ -85,8 +85,12 @@ final class BlankLineAfterOpeningTagFixer extends AbstractFixer implements White
             $tokens[0] = new Token([$token->getId(), rtrim($token->getContent()).$lineEnding]);
         }
 
-        if (!$tokens[1]->isWhitespace() && false === strpos($tokens[1]->getContent(), "\n")) {
-            $tokens->insertAt(1, new Token([T_WHITESPACE, $lineEnding]));
+        if (false === strpos($tokens[1]->getContent(), "\n")) {
+            if ($tokens[1]->isWhitespace()) {
+                $tokens[1] = new Token([T_WHITESPACE, $lineEnding.$tokens[1]->getContent()]);
+            } else {
+                $tokens->insertAt(1, new Token([T_WHITESPACE, $lineEnding]));
+            }
         }
     }
 }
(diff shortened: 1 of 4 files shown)


kubawerlos:
What about:
```php
<?php
namespace Foo; /** comment 1 */              /**
 * comment 2
 */
    use Bar;
```
Should we add new line before `use` or before `comment 2` (as it has break line itself)?

SpacePossum:
I think it would be fine to not fix in that case, sometimes people write stuff we cannot make look better :(

src/Fixer/PhpTag/BlankLineAfterOpeningTagFixer.php
```diff
@@ -85,8 +85,12 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
             $tokens[0] = new Token([$token->getId(), rtrim($token->getContent()).$lineEnding]);
         }
 
-        if (!$tokens[1]->isWhitespace() && false === strpos($tokens[1]->getContent(), "\n")) {
-            $tokens->insertAt(1, new Token([T_WHITESPACE, $lineEnding]));
+        if (false === strpos($tokens[1]->getContent(), "\n")) {
```
kubawerlos: Is it OK to check only fofr `\n` instead of RegEx with `\R`?

SpacePossum: we support "\n" and "\r\n" but not "\r" only (old Mac thing I think?), so I think this is fine (we do this on other places as well)

keradus:
(dropping milestone as this is WIP)

tests/Fixer/NamespaceNotation/BlankLineAfterNamespaceFixerTest.php
```diff
@@ -138,6 +145,33 @@ class C {}\r",
...
+            ],
+            [
+                '<?php
+    namespace My\NS;
+
+    class X extends Y {}',
+            ],
+            [
+                '<?php
+namespace My\NS; // comment
```
julienfalque: What about adding a test case with a `/* comment */`?

SpacePossum: :+1:  added

SpacePossum:
Thank you @kubawerlos.

PR #4134, opened 2018-12-04
[7.3] SquareBraceTransformer - nested array destructuring not handled correctly

SpacePossum:
found when testing/preparing for PHP 7.3


src/Tokenizer/Transformer/SquareBraceTransformer.php
```diff
@@ -93,6 +93,19 @@ private function transformIntoDestructuringSquareBrace(Tokens $tokens, $index)
 
         $tokens[$index] = new Token([CT::T_DESTRUCTURING_SQUARE_BRACE_OPEN, '[']);
         $tokens[$endIndex] = new Token([CT::T_DESTRUCTURING_SQUARE_BRACE_CLOSE, ']']);
+
+        $previousMeaningfulIndex = $index;
+        $index = $tokens->getNextMeaningfulToken($index);
+
+        while (null !== $index && $index < $endIndex) {
```
kubawerlos: Is the first condition (`null !== $index`) possible to achieve? The loop iterates with `getNextMeaningfulToken` and will finally get to position `$endIndex`.

PR #4129, opened 2018-11-30
Fix: Symfony 4.2 deprecations

kubawerlos:
- [x] [Fix Travis CI](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4138)
- [x] [Fix CircleCI](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4139)
- [x] Manage deprecations in [`keradus/cli-executor`](https://github.com/keradus/cli-executor/pull/2)
- [x] Update `composer.json` with new `keradus/cli-executor` version

PR #4138, opened 2018-12-09
Fix Travis CI

kubawerlos:
[This](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3639) is PR that introduced `symfony/lts` usage - when you go to Travis log you can [see](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/359259428) (click on the line with "symfony/lts" to expand logs) that it was failing from the begging. It started to break builds after Composer 1.8.0 was released. Run this series of commands to see:
```bash
mkdir test173 test180
composer self-update 1.7.3
composer require --no-update symfony/lts:v4.0 --working-dir=test173
composer self-update 1.8.0
composer require --no-update symfony/lts:v4.0 --working-dir=test180
cat test173/composer.json
cat test180/composer.json

```

P.S. I have disabled deprecations related to Symfony 4.2 to verify all CI builds - I'd like to enable it in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4129.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,10 +63,6 @@ jobs:
             stage: Fast Test
             php: 7.0
             install:
-                # Composer: enforce given Symfony components version
-                - if [ "$SYMFONY_VERSION" != "" ]; then composer require --no-update symfony/force-lowest:$SYMFONY_VERSION; fi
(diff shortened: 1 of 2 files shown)


keradus:
Thanks for the work on this and pointing out which part of Travis is not working, @kubawerlos ;)

PR #4143, opened 2018-12-10
PhpUnitTestCaseStaticMethodCallsFixer - Add PHPUnit 7.5 new assertions

Slamdunk:
Reference: https://github.com/sebastianbergmann/phpunit/blob/7.5.0/ChangeLog-7.5.md

NOTE: tests don't fail in CI because of `phpunitgoodpractices/traits` locking newer versions of PHPUnit:

> $ composer.phar why phpunit/phpunit
> [...]
> phpunitgoodpractices/traits                        v1.6.0          requires                    phpunit/phpunit (~4.0.0 || ~4.1.0 || ~4.2.0 || ~4.3.0 || ~4.4.0 || ~4.5.0 || ~4.6.0 || ~4.8.0 || ~5.3.0 || ~5.4.0 || ~5.5.0 || ~5.6.0 || ~5.7.0 || ~6.0.0 || ~6.1.0 || ~6.2.0 || ~6.3.0 || ~6.4.0 || ~6.5.0 || ~7.0.0 || ~7.1.0 || ~7.2.6 || ~7.3.0)

On my local setup I run the tests without `phpunitgoodpractices/traits`


SpacePossum:
nice update to newer PHPUnit :)  :+1: 

maybe these should be added using something like `PhpUnitTargetVersion::fulfills` ?

Slamdunk:
You are right. We should have used `PhpUnitTargetVersion` since the beginning, but for some unknowable reason we didn't.

I have to say currently I'm unable to code this improvement: if it's ok for you all, I'll leave this PR as it is and create a dedicated Issue for the future.

keradus:
for most PHPUnit fixers we do have PHPUnit target version, as manipulation of assertion (like changing one assertion (`::assertSame(true === $x)`) into another (`::assertTrue($x)`) may fail, if later one does not exist (yet, as someone is using old version).

adding support for `::assertX` vs `->assertX` is not that problematic that we need to know target version, as long as one doesn't implement assertion on his own - but that's exactly why Fixer is already marked as risky, and we don't increase the riskiness by supporting more assertions

keradus:
Tested with 7.5: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/473161761#L654

keradus:
Thank you @Slamdunk.

PR #4147, opened 2018-12-12
HeaderCommentFixer - throw exception on invalid header configuration (II)

SpacePossum:
Follow up on https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4099 where I missed this case.

cc @kubawerlos

### DIFF
--- a/src/Fixer/Comment/HeaderCommentFixer.php
+++ b/src/Fixer/Comment/HeaderCommentFixer.php
@@ -169,8 +169,10 @@ echo 1;
                         return '';
                     }
 
-                    if (false !== strpos($value, '*/')) {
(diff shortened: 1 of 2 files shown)


dmvdbrugge:
The case is more complex than you think:

```php
<?php
/**
 * This is valid ?>
 */

/*
 * This is valid too ?>
 */

/* And so is this ?> */
```

```php
<?php
// This is the case that breaks ?>
```

```php
<?php
# This breaks too ?>
```

(Tested locally on php 7.1 at least, and the github syntax highlighting seems to agree)

dmvdbrugge:
And I see now that the only cases that break aren't even supported by `HeaderCommentFixer` so this PR is unnecessary, right?

kubawerlos:
I agree with @dmvdbrugge - the fixer only support comments where `?>` is fine.

PR #4150, opened 2018-12-14
Update WhitespacesFixerConfig.php

koromerzhin:
With https://mlocati.github.io/php-cs-fixer-configurator/#configurator you can choose the indent (1 space to 8 spaces or tab)

### DIFF
--- a/src/WhitespacesFixerConfig.php
+++ b/src/WhitespacesFixerConfig.php
@@ -27,7 +27,7 @@ final class WhitespacesFixerConfig
     public function __construct($indent = '    ', $lineEnding = "\n")
     {
         $testIndent = ["\t"];
-        for ($i = 1; $i <= 8 ; ++$i) {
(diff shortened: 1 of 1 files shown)


ntzm:
The only supported indents are 2 spaces, 4 spaces and tabs. This is reflected throughout other fixers as well.

kubawerlos:
> The only supported indents are 2 spaces, 4 spaces and tabs. This is reflected throughout other fixers as well.

and [tests](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.13.1/tests/WhitespacesFixerConfigTest.php#L55).

SpacePossum:
Hi @koromerzhin ,

As stated by others we support a few indentation options but not all, this is by design. Do you know of a standard or any major (F)OSS project that uses a different indentation style? For us this is important because we want to promote standards and/or support styles that are used a lot (so more support for this is more likely)

PR #4152, opened 2018-12-14
Fix indentation_type

koromerzhin:
this bug comes from '   ' which was misinterpreted


ntzm:
1. Current tests are failing
2. Could you add some tests to prove this bug?

PR #4153, opened 2018-12-16
PhpUnitFqcnAnnotationFixer - handle only PhpUnit classes

kubawerlos:
- handle only PhpUnit classes
 - do not add backslash to `self` and `static`


kubawerlos:
> * Native PHP functions and classes are qualified.

This happens currently.

> * The classes specified by `use` are resolved.

Definitely out of scope for this fixer, yet good idea for a new fixer.

> * Any class method reference that isn't resolved remains unqualified.

That would be against what this fixer should do and I can't see why anyone would like skipping unresolved classes.

> * And of course, `self` is not prefixed.

Fixed in this PR.

> * Finally, the entire list is sorted alphabetically. (Bonus points: group PHP functions, resolved class methods, unresolved class methods, `self` methods and `static` methods.)

Sorting annotations would be quite interesting fixer - don't forget we could have multi-line annotations.

keradus:
Thank you @kubawerlos.

PR #4163, opened 2018-12-24
Enhancement: Update localheinz/composer-normalize

localheinz:
This PR

* [x] updates `localheinz/composer-normalize`

💁‍♂️ For reference, see https://github.com/localheinz/composer-normalize/compare/1.0.0...1.1.0.

### DIFF
--- a/dev-tools/composer.json
+++ b/dev-tools/composer.json
@@ -3,7 +3,7 @@
         "php": "^7.1"
     },
     "require-dev": {
-        "localheinz/composer-normalize": "^1.0.0",
+        "localheinz/composer-normalize": "^1.1.0",
         "maglnet/composer-require-checker": "^1.0",
         "mi-schi/phpmd-extension": "^4.2",
         "phpmd/phpmd": "^2.4.3"


localheinz:
@keradus 

I don't mind - personally, I never run 

```
$ composer update
```

to update a dependency (unless in CI), but only

```
$ composer require
```

From my point of view, this documents nicely the minimum required version, that is all.

keradus:
> From my point of view, this documents nicely the minimum required version, that is all.

keep in mind:
> > We don't update anything (code/config) to benefit from new version. What's the value of bumping it ?

would PHP CS Fixer run with `localheinz/composer-normalize` 1.0? yes. 1.1? yes. 1.x? yes...

dev-tools/composer.json
```diff
@@ -3,7 +3,7 @@
         "php": "^7.1"
     },
     "require-dev": {
-        "localheinz/composer-normalize": "^1.0.0",
+        "localheinz/composer-normalize": "^1.1.0",
```
keradus: let's write it as `^1.1` (no need for PATCH number) and merge it once, without future updates for each release, as long as there is nothing to update on our side (like change of config or sth)

kubawerlos: What if we would remove these 4 deps from `composer.json` and install them via `composer require`? That wouldn't slow things, would it?

keradus: if we would do that, what would be the benefit ?

kubawerlos: We'd have latest stable versions of those packages (I assume there won't be any conflicts between them) and will avoid bumping version constraint - `composer-require-checker` has already `1.1` tag, `mi-schi/phpmd-extension` has `4.3` and `phpmd/phpmd` has `2.6`.

localheinz:
Ha, no problem!

dev-tools/composer.json
kubawerlos: I didn't want to `composer require localheinz/composer-normalize:^1`, but simply `composer require localheinz/composer-normalize` to cover future major version bumps - I believe we have very little chance to stuck on BC break of these packages.

keradus: then I would say no to that.
don't install tool without providing major version.

you may hope that if you currently have v1, that v2 won't provide BC break that we need to consider, but can you say the same between v1 and v5 ?

PR #4166, opened 2018-12-27
DX: do not use AbstractFunctionReferenceFixer when no need to

kubawerlos:
`PhpUnitExpectationFixer` and `PhpdocAddMissingParamAnnotationFixer` are not using `AbstractFunctionReferenceFixer`.


keradus:
Thank you @kubawerlos.

PR #4171, opened 2018-12-28
Fix CircleCI cache

kubawerlos:
Currently on `2.12` we only save cache, never restore it :(

PR split into 2 commits:
 - first one to save cache when having content to cache - what's important with a different key
 - second one to restore it - the reason for split is to see after second commit the build time change

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,15 +3,13 @@ version: 2
 jobs:
     build:
         macos:
-            xcode: '9.0'
+            xcode: '10.1.0'
         steps:
             - checkout
 
-            - save_cache:
-                key: dependency-cache
-                paths:
-                    - ~/.composer
-                    - ~/Library/Caches/Homebrew
+            - restore_cache:
+                keys:
+                    - cache-{{ checksum "composer.json" }}
 
             - run: brew update
             - run: brew install php@7.2
@@ -22,5 +20,12 @@ jobs:
             - run: php composer.phar global show hirak/prestissimo -q || php composer.phar global require --no-interaction --no-progress --optimize-autoloader hirak/prestissimo
             - run: php composer.phar install --optimize-autoloader --no-interaction --no-progress --no-suggest
             - run: php composer.phar info -D | sort
+
+            - save_cache:
+                key: cache-{{ checksum "composer.json" }}
+                paths:
+                    - ~/.composer
+                    - ~/Library/Caches/Homebrew
+
             - run: vendor/bin/phpunit
             - run: PHP_CS_FIXER_FUTURE_MODE=1 php php-cs-fixer --diff --dry-run -v fix


.circleci/config.yml
```diff
@@ -22,5 +16,12 @@ jobs:
             - run: php composer.phar global show hirak/prestissimo -q || php composer.phar global require --no-interaction --no-progress --optimize-autoloader hirak/prestissimo
             - run: php composer.phar install --optimize-autoloader --no-interaction --no-progress --no-suggest
             - run: php composer.phar info -D | sort
+
+            - save_cache:
+                key: source-v1-{{ .Branch }}-{{ .Revision }}
```
keradus: don't cache by revision, please, too detailed.

homebrew we can cache by php version, but we have only one currently, so we can cache always
composer's vendor we can cache by php version (also not relevant) and PR target branch (2.12/2.13/master/3.0),
but even if we would not, most of the deps would be cached properly, as 2.12 vs 3.0 has very little deps change.

So, even just adding `restore_cache` would be enough, no need for interpolated cache key

kubawerlos: Okay, do you know how to speed up `brew update`? It takes usually 5 minutes out of 8 for the build.

.circleci/config.yml
```diff
@@ -7,11 +7,9 @@ jobs:
...
             - checkout
 
-            - save_cache:
-                key: dependency-cache
-                paths:
-                    - ~/.composer
-                    - ~/Library/Caches/Homebrew
+            - restore_cache:
+                keys:
+                    - source-v1-{{ checksum "composer.json" }}
```
kubawerlos: @keradus what about this key?

.circleci/config.yml
keradus: no, i'm not using mac

.circleci/config.yml
keradus: no hard opinion on this one

```
ker@dus:~/github/FriendsOfPHP/PHP-CS-Fixer λ composer info | wc -l
64
ker@dus:~/github/FriendsOfPHP/PHP-CS-Fixer λ git diff 2.12..3.0 composer.json
diff --git a/composer.json b/composer.json
index 5df4718f..b42e19aa 100644
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
         "composer/semver": "^1.4",
         "composer/xdebug-handler": "^1.2",
         "doctrine/annotations": "^1.2",
-        "php-cs-fixer/diff": "^1.3",
+        "php-cs-fixer/diff": "^2.0",
         "symfony/console": "^3.4.17 || ^4.1.6",
         "symfony/event-dispatcher": "^3.0 || ^4.0",
         "symfony/filesystem": "^3.0 || ^4.0",
```

if we have 64 package installed, do we really want to not use cache of them, if **one** package is different ? installing 63 packages from cache + one fresh vs 64 fresh...

kubawerlos: Yes, and then it will cache them and reuse again - bear in mind that CircleCI will not save the cache when there already is anything saved under the key - take a look [here](https://circleci.com/gh/FriendsOfPHP/PHP-CS-Fixer/5470)  in "Save Cache" action - 1 month old cache, impossible to track what is there.

kubawerlos:
Okay, summarising:
 - updating xcode from 9.0 to 10.0.1 reduced build time (precisely `brew update`) by 2.5~3 minutes
 - fixing cache gave a very little boost (composer installing `prestissimo` and rest of deps took up to 20 seconds, after fix almost 0)

.circleci/config.yml
keradus: oh damn it !
ok, then your proposal of using checksum makes perfect sense !

why `source-v1` prefix ? what does it stand for ?

kubawerlos: I have no idea - it's from the [docs](https://circleci.com/docs/2.0/caching/#source-caching) - never thought about it. Example seems to cache source and we would cache dependencies and Homebrew - what name would you propose? maybe simply `cache`?

keradus:
Thank you @kubawerlos.

PR #4176, opened 2018-12-30
Add support for braces / position_before_control_structures

jclaveau:
This option allows to force a line break before "else", "elseif", "finally" and "catch" to preserve verticality between all the control structures of conditions and exceptions handling. 

I found this kind of prettyprint in a project and realized it was really helpful when you need to duplicate or comment a whole case (several lines including braces) without making the code broken: 

```php
if (/* first case */) {
    // whatever 
}
// elseif (/* case you want to patch (commented so easy to compare) */) {
//     // what was done before your patch
// }
elseif (/* copy of the case that you're patching */) {
     // your patch experiments
}
else (/* ... */) {
    /// ...
}
``` 

This participated to make my workflow really smoother and promotes a rule trait which is rare in mainstream standards even if it's a plus for readability, I think: verticality.
It also makes it shorter to convert the "if {}" and "else {}" statements to "elseif {}" in both ways.

Hoping you'll consider this behavior and thank you very much for your work!


jclaveau:
I reopened it based on a dedicated branch : https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4178

Issue #3837, opened 2018-06-22
indentation_type unexpected indentation


amitbisht511:
#### The PHP version you are using (`$ php -v`):
=>7.1.11

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.12.1

#### The command you use to run PHP CS Fixer:
=> Using Sublime text (3.1.1) plugin [php cs fixer](https://github.com/adael/SublimePhpCsFixer)
=> Also tried in visual studio code (1.24.1)

#### The configuration file you are using, if any:
```php
return PhpCsFixer\Config::create()
    ->setRules(array(
        '@PSR2' => true,
        'method_argument_space' => array('on_multiline' =>'ignore','keep_multiple_spaces_after_comma' => false),
    ))
    ->setIndent("  ")
    ->setLineEnding("\n")
    ->setUsingCache(false)
    ->setCacheFile(__DIR__.'\php_cs.cache')
;
```
same if i set on_multiline='ensure_fully_multiline' 

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php

if (true) {
  if (true) {
    (new stdClass())->asdf(
      'text',
      'text2'
    );
  }
}

```

 * with unexpected changes applied when running PHP CS Fixer:
```php
if (true) {
  if (true) {
    (new stdClass())->asdf(
    'text',
    'text2'
  );
  }
}
```
if i run PHP CS Fixer again 

```php
if (true) {
  if (true) {
    (new stdClass())->asdf(
  'text',
  'text2'
  );
  }
}
```

 * with the changes you expected instead:
No changes

DanielRuf:
Same here. Ths produces problems with arguments on multiple lines.

This seems to work as workaround:
```
return PhpCsFixer\Config::create()
    ->setUsingCache(false)
    ->setRules([
        '@PSR1' => true,
        '@PSR2' => true,
        'indentation_type' => false
    ])
    ->setIndent("  ")
    ->setFinder($finder)
;
```

Running this multiple times changes it again.

whereisaaron:
Getting the same problems using 2.15.1 with 'php cs fixer' extension for VS Code.

It is not even idempotent, each time you format the same file it changes as multi-line arrays and function call arguments march left by 2 spaces each time like @amitbisht511's example.

```
.\php-cs-fixer.phar -V
[...]

PR #4183, opened 2018-12-31
IndentationTypeFixer - fix handling 2 spaces indent

kubawerlos:
For https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3837

### DIFF
--- a/src/Fixer/Whitespace/IndentationTypeFixer.php
+++ b/src/Fixer/Whitespace/IndentationTypeFixer.php
@@ -101,8 +101,8 @@ final class IndentationTypeFixer extends AbstractFixer implements WhitespacesAwa
         $indent = $this->indent;
 
         // change indent to expected one
-        $content = Preg::replaceCallback('/^(?:    )+/m', static function ($matches) use ($indent) {
-            return str_replace('    ', $indent, $matches[0]);
+        $content = Preg::replaceCallback('/^(?:    )+/m', function ($matches) use ($indent) {
+            return $this->getExpectedIndent($matches[0], $indent);
         }, $content);
 
         return new Token([$tokens[$index]->getId(), $content]);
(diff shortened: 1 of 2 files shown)


src/Fixer/Whitespace/IndentationTypeFixer.php
```diff
@@ -144,4 +144,19 @@ static function (array $matches) use ($indent) {
...
         return new Token([T_WHITESPACE, $newContent]);
     }
+
+    /**
+     * @param string $content
+     * @param string $indent
+     *
+     * @return string mixed
+     */
+    private function getExpectedIndent($content, $indent)
```
AdryanSignor: There isn't **$this** context within the function. Can you use static declaration?
Call getExpectedIndent with **self** token within anonymous function for Preg::replaceCallback, will
prevent anonymous function relocation.

keradus: static === global === evil
try to mock static method

no static here

tests/Fixer/Whitespace/IndentationTypeFixerTest.php
```diff
@@ -324,4 +324,69 @@ static function ($key) {
...
+if (true) {
+  if (true) {
+\t(new stdClass())->foo(
+\t  'text',
+\t  'text2'
+\t);
+  }
+}",
+            ],
+            ['<?php
```
julienfalque: This case is redundant with the subsequent one (as `doTest()` asserts that the expected part doesn't get changed by the fixer when used as input).

SpacePossum: :+1:  force pushed it out

SpacePossum:
Thank you @kubawerlos.

PR #4185, opened 2018-12-31
[7.3] ImplodeCallFixer - add tests for PHP7.3

kubawerlos:


tests/Fixer/FunctionNotation/ImplodeCallFixerTest.php
```diff
@@ -146,4 +146,44 @@ public function provideFixCases()
...
+     * @dataProvider provideFix73Cases
+     */
+    public function testFix73($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFix73Cases()
+    {
+        yield ['<?php implode("", $foo, );'];
```
keradus: i don't think we need to yield static array that is not generated

keradus: this one is provided by next case as well

tests/Fixer/FunctionNotation/ImplodeCallFixerTest.php
```diff
@@ -146,4 +146,42 @@ public function provideFixCases()
...
+     * @dataProvider provideFix73Cases
+     */
+    public function testFix73($expected, $input)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFix73Cases()
+    {
+        yield [
```
keradus: i don't think we need to yield static array that is not generated

kubawerlos: Original tests were with `yield` (made by me :smile:) so I continued with it - it makes one level of array less in data providers.

keradus:
Thank you @kubawerlos.

tests/Fixer/FunctionNotation/ImplodeCallFixerTest.php
TomasVotruba: Better make it consistent thought, one way, or another.

There is Rector for yield approach - https://github.com/rectorphp/rector/blob/master/config/level/phpunit/phpunit-yield-data-provider.yml

kubawerlos: OMG, I've wasted a lot of time to change some nasty data providers manually...

Is there maybe a Rector to change data providers to [static](https://github.com/sebastianbergmann/phpunit/blob/7.5.1/src/Util/Test.php#L862) methods?

keradus: no statics, please...

TomasVotruba: @kubawerlos No more manual labor for programmers :D I'm not sure what you mean. Create issues at Rector with example and I'll see what we can do.

PR #4186, opened 2018-12-31
Tokens - fix removal of leading/trailing whitespace with empty token in collection

kubawerlos:
It might happen that `Tokens` collection contain a sequence:
 - `T_WHITESPACE`
 - `new Token('')` (from `Tokens::clearAt(N - 1)`)
 - a token at index `N`

When we call `Tokens::removeLeadingWhitespace(N)` currently `Tokens` will try with offset `-1`, checks that it is not whitespace and does nothing.

(Of course similar situation applies to `Tokens::removeTrailingWhitespace`)

### DIFF
--- a/tests/Tokenizer/TokensTest.php
+++ b/tests/Tokenizer/TokensTest.php
@@ -1174,6 +1174,30 @@ echo $a;',
         return $cases;
     }
 
+    public function testRemovingLeadingWhitespaceWithEmptyTokenInCollection()
+    {
+        $code = "<?php\n    /* I will be removed */MY_INDEX_IS_THREE;foo();";
+        $tokens = Tokens::fromCode($code);
+        $tokens->clearAt(2);
+
+        $tokens->removeLeadingWhitespace(3);
+
+        $tokens->clearEmptyTokens();
+        $this->assertTokens(Tokens::fromCode("<?php\nMY_INDEX_IS_THREE;foo();"), $tokens);
+    }
+
+    public function testRemovingTrailingWhitespaceWithEmptyTokenInCollection()
+    {
+        $code = "<?php\nMY_INDEX_IS_ONE/* I will be removed */    ;foo();";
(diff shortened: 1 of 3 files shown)


tests/Tokenizer/TokensTest.php
```diff
@@ -1174,6 +1174,26 @@ public function provideRemoveTrailingWhitespaceCases()
...
+    }
+
+    public function testRemovingTrailingWhitespaceWithEmptyTokenInCollection()
+    {
+        $tokens = Tokens::fromCode("<?php\nMY_INDEX_IS_ONE/* I will be removed */    ;foo();");
+        $tokens->clearAt(2);
+
+        $tokens->removeTrailingWhitespace(1);
+
+        $this->assertSame("<?php\nMY_INDEX_IS_ONE;foo();", $tokens->generateCode());
```
keradus: can we use `assertTokens` instead ? I would like to ensure that output collection is proper

keradus: like this, but use `clearEmptyTokens` on `$tokens`

tests/Tokenizer/TokensTest.php
```diff
@@ -1174,6 +1174,26 @@ public function provideRemoveTrailingWhitespaceCases()
...
+
+        $this->assertSame("<?php\nMY_INDEX_IS_THREE;foo();", $tokens->generateCode());
+    }
+
+    public function testRemovingTrailingWhitespaceWithEmptyTokenInCollection()
+    {
+        $tokens = Tokens::fromCode("<?php\nMY_INDEX_IS_ONE/* I will be removed */    ;foo();");
+        $tokens->clearAt(2);
+
+        $tokens->removeTrailingWhitespace(1);
```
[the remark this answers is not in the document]
kubawerlos: Exactly how what the fix will do ;)

keradus:
Thank you @kubawerlos.

PR #4190, opened 2019-01-01
Travis CI - fix config

kubawerlos:
We allow build job with collecting coverage to fail, but since it was changed to use PHP 7.3 it has more flags and make builds [failing](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/474074754) thus we need to update the config.


keradus:
Thank you @kubawerlos.

PR #4173, opened 2018-12-29
[7.3] PowToExponentiationFixer - add support for PHP7.3

SpacePossum:

PR #4198, opened 2019-01-01
[7.3] PowToExponentiationFixer - adding to PHP7.3 integration test

kubawerlos:
The support was added in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4173.


SpacePossum:
I think the SCA job fails here because we give PHPMD the modified files from the PR, which in this case do not contain any `.php`-source files.

SpacePossum:
Thank you @kubawerlos.

PR #4200, opened 2019-01-01
[7.3] ModernizeTypesCastingFixer - support PHP 7.3

kubawerlos:
Ping @kalessil (as original author) for review.

### DIFF
--- a/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
+++ b/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
@@ -91,9 +91,17 @@ final class ModernizeTypesCastingFixer extends AbstractFunctionReferenceFixer
                     continue;
                 }
 
+                $paramContentEnd = $closeParenthesis;
+                $commaCandidate = $tokens->getPrevMeaningfulToken($paramContentEnd);
+                if ($tokens[$commaCandidate]->equals(',')) {
+                    $tokens->removeTrailingWhitespace($commaCandidate);
+                    $tokens->clearAt($commaCandidate);
+                    $paramContentEnd = $commaCandidate;
+                }
+
                 // check if something complex passed as an argument and preserve parenthesises then
                 $countParamTokens = 0;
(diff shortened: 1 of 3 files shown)


src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -91,9 +91,17 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                     continue;
                 }
 
+                $endIndex = $closeParenthesis;
```
kalessil: $paramContentEnd instead of $endIndex would be more explaining here.

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -91,9 +91,17 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                $prevIndex = $tokens->getPrevMeaningfulToken($endIndex);
```
kalessil: $commaCandidate instead of $prevIndex would be also self-explaining.

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -91,9 +91,17 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
+                if ($tokens[$prevIndex]->equals(',')) {
+                    $tokens->removeTrailingWhitespace($prevIndex);
```
kalessil: As edge case, people can place space before the comma as well =)

kubawerlos: Do you mean like in the 2nd test case I've added?

kalessil: Yep, seems you already covered it =)

SpacePossum:
Thank you @kubawerlos.

PR #4203, opened 2019-01-01
DX: fix indents

kubawerlos:


keradus:
change done manually or automated?

kubawerlos:
> change done manually or automated?

Both ;) I mean I have run `phpcs` with `PSR2.Methods.FunctionCallSignature.Indent`, but it indented a lot of `new CodeSample(\n`, so I reverted those to left only the obvious cases.

The problem is we have mixed indents like [this](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.13.2/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php#L89) - compare line `90` and `108`.

Of course when `CodeSample` constructor takes more than 1 argument we have to indent each of them, but with 1 argument we can also do `new CodeSample('<?php` and I wasn't sure which way we want to go.

keradus:
> The problem is we have mixed indents like this - compare line 90 and 108.

108 is good, 90 is not

> I wasn't sure which way we want to go.

```php
foo('a
b
c, 'd
e
f');
```

or
```php
foo(
   'a
b
c,
    'd
e
f');
```

I would use later

kubawerlos:
Yes, with 2 arguments it's easy, what about with 1?
```php
foo('a
b
c');
```
or
```php
foo(
    'a
b
c'
);
```

Keep in mind there migth be couple of `foo` calls in a row - some with 1 argument and some with more - like these from link - lines 90 i 108 calls are with 2 argument, while line 76 is with 1.

PR #4204, opened 2019-01-01
DX: Tokens - do not unregister/register found tokens when collection is not changing

kubawerlos:
This makes sence, right?

Only run `unregisterFoundToken`/`registerFoundToken` when changing collection.


keradus:
Thank you @kubawerlos.

PR #4205, opened 2019-01-01
DX: PhpdocAlignFixer - refactor to use DocBlock

kubawerlos:
As the change itself - using `DocBlock` instead of reinventing the wheel way - feels right there is also a surprise when this will be merged ;)

This is last one for today - Happy New Year to everybody! Long live the open source!

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocAlignFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocAlignFixer.php
@@ -13,6 +13,7 @@
 namespace PhpCsFixer\Fixer\Phpdoc;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\DocBlock\DocBlock;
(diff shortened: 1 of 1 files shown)


src/Fixer/Phpdoc/PhpdocAlignFixer.php
```diff
@@ -378,7 +375,7 @@ private function getIndent($verticalAlignIndent, $leftAlignIndent = 1)
     {
         $indent = self::ALIGN_VERTICAL === $this->align ? $verticalAlignIndent : $leftAlignIndent;
 
-        return str_repeat(' ', $indent);
+        return \str_repeat(' ', $indent);
```
keradus: ?

kubawerlos: ¿

keradus:
Thank you @kubawerlos.

PR #4213, opened 2019-01-02
[7.3] PHP7.3 integration tests

SpacePossum:


tests/Fixtures/Integration/misc/PHP7_3.test
```diff
@@ -103,3 +112,10 @@ rand($a, $b, ); // `random_api_migration` rule
 settype($foo, "integer", ); // `set_type_to_cast` rule
 in_array($b, $c, ); // `strict_param` rule
 foo($a === null, ); // `yoda_style` rule
+$a =(unset)$z; // `no_unset_cast` rule
```
keradus: didn't you pointed out it's pointless to add this rule here ?

SpacePossum: yeah, but could do no harm to prove that here by test?

SpacePossum:
failing tests are:
- PHPMD is run without any files (since no php files are touched in the PR) 
- timeout when using collecting-coverage (this test is allowed to fail)

keradus:
Thank you @SpacePossum.

PR #3974, opened 2018-08-16
Psr4Fixer - class with anonymous class

kubawerlos:

PR #4217, opened 2019-01-02
Psr0Fixer - class with anonymous class

kubawerlos:
Similar to https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3974


keradus:
Thank you @kubawerlos.

Issue #4218, opened 2019-01-02
Method name gets lower-cased


umulmrum:
#### The PHP version you are using (`$ php -v`):
=> 7.1.25

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.13.1 (download of current version 2.13.2 failed with message "Unable to download new version v2.13.2 from the server.", didn't investigate yet)

#### The command you use to run PHP CS Fixer:
=> php-cs-fixer fix

#### The configuration file you are using, if any:
```php
return PhpCsFixer\Config::create()
    ->setRules(array(
        '@Symfony' => true,
        'phpdoc_to_comment' => false,
        'phpdoc_var_without_name' => false,
    ))
;
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
public function &Next()
{
    ...
}
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
public function &next()
{
    ...
}
```
(method name is lowercased)

 * with the changes you expected instead:
No changes expected.

Interesting bits:
- When I remove the `&`, no changes are made (as expected).
- In the same class there are multiple methods that return references, but only `Next()` and `Current()` were lowercased, not e.g. `Previous()`.
- When changing the method name, no changes are made.
- When adding a method `Key()`, the same problem occurs.
- The class doesn't implement any interfaces and does not derive from another class.

The class is named `TIterator`, so it's close to PHP's `Iterator` interface. Maybe the fixer thinks the class implements `Iterator`?

Thanks for help!

kubawerlos:
Hi @umulmrum,

the fixer to blame is `native_function_casing` (part of `@Symfony`).

In future for better investigation try to run `php php-cs-fixer fix --diff --dry-run -vvv` - you will see something like this:
```
Legend: ?-unknown, I-invalid file syntax, file ignored, S-Skipped, .-no changes, F-fixed, E-error
   1) FileName.php (native_function_casing)
      ---------- begin diff ----------
--- Original
+++ New
@@ @@
 
-    public function &Next()
+    public function &next()
```

PR #4219, opened 2019-01-02
 NativeFunctionCasingFixer - handle T_RETURN_REF 

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4218

Ping @umulmrum (as reporter) and @SpacePossum (as original author) for review.

### DIFF
--- a/src/Fixer/Casing/NativeFunctionCasingFixer.php
+++ b/src/Fixer/Casing/NativeFunctionCasingFixer.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Fixer\Casing;
 use PhpCsFixer\AbstractFixer;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
 use PhpCsFixer\Tokenizer\Token;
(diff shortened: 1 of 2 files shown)


src/Fixer/Casing/NativeFunctionCasingFixer.php
```diff
@@ -67,6 +68,9 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
             }
 
             $functionNamePrefix = $tokens->getPrevMeaningfulToken($index);
+            if ($tokens[$functionNamePrefix]->isGivenKind(CT::T_RETURN_REF)) {
+                $functionNamePrefix = $tokens->getPrevMeaningfulToken($functionNamePrefix);
+            }
             if ($tokens[$functionNamePrefix]->isGivenKind([T_DOUBLE_COLON, T_NEW, T_OBJECT_OPERATOR, T_FUNCTION])) {
```
gharlan: I would add `CT::T_RETURN_REF` to this array instead.

kubawerlos: Thanks, I thought `CT::T_RETURN_REF` would also be in `$next1 = & Next($array1);` case, but it is not.

keradus:
Thank you @kubawerlos.

PR #4220, opened 2019-01-02
NativeFunctionInvocationFixer - namespaced strict to remove backslash

kubawerlos:
When we want only namespaced calls to have `\` in strict mode we should remove `\` for global namespace call.

### DIFF
--- a/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
+++ b/src/Fixer/FunctionNotation/NativeFunctionInvocationFixer.php
@@ -182,7 +182,7 @@ $c = get_class($d);
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
         if ('all' === $this->configuration['scope']) {
-            $this->fixFunctionCalls($tokens, $this->functionFilter, 0, \count($tokens) - 1);
+            $this->fixFunctionCalls($tokens, $this->functionFilter, 0, \count($tokens) - 1, false);
 
             return;
         }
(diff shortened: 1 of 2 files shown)


keradus:
Thank you @kubawerlos.

keradus:
ah, damn it. it shall be targetting 2.12...

keradus:
fixing milestone... `strict` was added in 2.14

voskobovich:
Can I have more information about the bug? I don't understand why we remove the slash in the function call.

kubawerlos:
@voskobovich take a look at the test - we remove backslash only for function in global namespace when fixer configured scope is `namespaced` and we are in strict mode.

voskobovich:
@kubawerlos Ok. I understand what you are doing. But I do not understand why this is necessary? Hard specification of the global namespace for a function should speed up its search when executing code.

kubawerlos:
@voskobovich this is not necessary, you can use scope `all`, yet it was bugged when scope `namespaced` was used.

> Hard specification of the global namespace for a function should speed up its search when executing code.

That is not true - opcode is the same, not matter if backaslash is used: https://3v4l.org/aK4ZR/vld#output and https://3v4l.org/NCSFH/vld#output.

voskobovich:
@kubawerlos Thank! This was new information for me))

keradus:
PHP CS Fixer sharing the knowledge and good practices ;)

mfn:
Can I use this rule for the opposite, ensuring that all `\` are in fact removed (please don't discuss pro/cons => different topic).

I read the current readme for "native_function_invocation" but I'm unsure.
> - exclude (array): list of functions to ignore; defaults to []
> - include (array): list of function names or sets to fix. Defined sets are @internal (all native functions), @all (all global functions) and @compiler_optimized (functions that are specially optimized by Zend); defaults to ['@internal']
> - scope ('all', 'namespaced'): only fix function calls that are made within a namespace or fix all; defaults to 'all'
> - strict (bool): whether leading \ of function call not meant to have it should be removed; defaults to false

I figured maybe with this I can achieve this, but it doesn't work:
```php
    'native_function_invocation' => [
      'exclude' => ['@all'],
      'strict' => true,
    ],
```
Still adds `\` everywhere and haven't noticed removing it where it's present.

Thanks :)

mfn:
> ensuring that all `\` are in fact removed

Apologies, I re-read my comment and I think it's not really clear:
the goal would be **not** to have function imports but just remove the `\` on global functions (and not exchanging the `\` for an import)

kubawerlos:
@mfn try:
```php
  'native_function_invocation' => [
      'include' => [],
      'strict' => true,
    ],
```

mfn:
How on eart…

Re-reading the readme _I guess_ it makes sense but I still fail to distill this from the text 🤷‍♀️

Anyway, _the one combination_ I did not try 😅, it works, so thank you 👍

PR #4224, opened 2019-01-03
FunctionToConstantFixer - handle T_RETURN_REF

SpacePossum:
same bug as #4219

ping @kubawerlos


keradus:
Thank you @SpacePossum.

PR #4232, opened 2019-01-04
DX: remove Utils::splitLines

kubawerlos:
This is the surprise mentioned [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4205#issue-241654313).

Class `Utils` is internal so no [backward compatibility](https://symfony.com/doc/current/contributing/code/bc.html) break.


keradus:
Thank you @kubawerlos.

PR #4233, opened 2019-01-04
Add official support for PHP 7.3

keradus:


composer.json
```diff
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || >=7.0 <7.3",
+        "php": "^5.6 || ^7.0",
```
julienfalque: We don't block unsupported future versions anymore?

PR #4235, opened 2019-01-04
DX: more specific @param types

kubawerlos:
A source was this part of the PHPStan report:
```
 ------ ------------------------------------------------------------------ 
  Line   Fixer/Import/SingleLineAfterImportsFixer.php                      
 ------ ------------------------------------------------------------------ 
  86     Binary operation "+=" between array and int results in an error.  
 ------ ------------------------------------------------------------------ 

 ------ ---------------------------------------------------------------- 
  Line   Fixer/LanguageConstruct/NoUnsetOnPropertyFixer.php              
 ------ ---------------------------------------------------------------- 
  204    Binary operation "+" between string and 1 results in an error.  
  222    Binary operation "+" between string and 1 results in an error.  
 ------ ---------------------------------------------------------------- 
```

### DIFF
--- a/src/Tokenizer/TokensAnalyzer.php
+++ b/src/Tokenizer/TokensAnalyzer.php
@@ -64,7 +64,7 @@ final class TokensAnalyzer
      *
      * @param bool $perNamespace Return namespace uses per namespace
      *
-     * @return array|array[]
+     * @return int[]|int[][]
      */
     public function getImportUseIndexes($perNamespace = false)
     {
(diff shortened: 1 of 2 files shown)


src/Tokenizer/TokensAnalyzer.php
```diff
@@ -64,7 +64,7 @@ public function getClassyElements()
      *
      * @param bool $perNamespace Return namespace uses per namespace
      *
-     * @return array|array[]
+     * @return int[]|int[][]
```
SpacePossum: the `int[]` is really an empty `array` here I think, than again `int[]` might be a better fit

Issue #4239, opened 2019-01-05
ModernizeTypesCastingFixer - breaks operator precedence


guilliamxavier:
`[]` and `**` have higher precedence than `(TYPE)` casts (http://php.net/manual/en/language.operators.precedence.php).

#### The PHP version you are using (`$ php -v`):
=> 7.2.10

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.12.6 (same issue with 2.13.3 and 2.14.0)

#### The command you use to run PHP CS Fixer:
=> `php-cs-fixer fix -vvv --using-cache=no --rules=modernize_types_casting --allow-risky=yes sample.php`

#### The configuration file you are using, if any:
(none)

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

$x = 42;
$y = strval($x)[0]; // ***
var_dump($y);

$x = 3.25;
$y = intval($x)**2; // ***
var_dump($y);
```
=> output (`$ php sample.php`):
```
string(1) "4"
int(9)
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php

$x = 42;
$y = (string) $x[0]; // ***
var_dump($y);

$x = 3.25;
$y = (int) $x**2; // ***
var_dump($y);
```
=> new output:
```
string(0) ""
int(10)
```

 * with the changes you expected instead:
```php
<?php

$x = 42;
$y = ((string) $x)[0]; // ***
var_dump($y);

$x = 3.25;
$y = ((int) $x)**2; // ***
var_dump($y);
```
=> (output: same as original)

SpacePossum:
thanks for reporting @guilliamxavier

PR #4240, opened 2019-01-05
ModernizeTypesCastingFixer - fix for operators with higher precedence

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4239

### DIFF
--- a/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
+++ b/src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
@@ -110,6 +110,10 @@ final class ModernizeTypesCastingFixer extends AbstractFunctionReferenceFixer
 
                 $preserveParenthesises = $countParamTokens > 1;
 
+                $afterCloseParenthesisIndex = $tokens->getNextMeaningfulToken($closeParenthesis);
+                $afterCloseParenthesisToken = $tokens[$afterCloseParenthesisIndex];
+                $wrapInParenthesises = $afterCloseParenthesisToken->equals('[') || $afterCloseParenthesisToken->isGivenKind(T_POW);
+
                 // analyse namespace specification (root one or none) and decide what to do
                 $prevTokenIndex = $tokens->getPrevMeaningfulToken($functionName);
                 if ($tokens[$prevTokenIndex]->isGivenKind(T_NS_SEPARATOR)) {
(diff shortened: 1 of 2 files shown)


src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -138,6 +145,10 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                     $tokens->removeTrailingWhitespace($functionName);
                 }
 
+                if ($wrapInParenthesises) {
+                    $tokens->insertAt($closeParenthesis, new Token(')'));
```
guilliamxavier: Thanks for the reactivity, but what will this result for e.g. `<?php $foo = strval($x + $y)[0];`? What about combining with the existing `$preserveParenthesises`?

guilliamxavier: OK I just tried your fix and the result is `<?php $foo = ((string) ($x + $y))[0];`, which is good! :+1: might be worth to add this case to the tests ;)

SpacePossum:
ping @kalessil as original author, if you time for a little review please do so :)

src/Fixer/CastNotation/ModernizeTypesCastingFixer.php
```diff
@@ -110,6 +110,10 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
 
                 $preserveParenthesises = $countParamTokens > 1;
 
+                $afterCloseParenthesisIndex = $tokens->getNextMeaningfulToken($closeParenthesis);
+                $afterCloseParenthesisToken = $tokens[$afterCloseParenthesisIndex];
+                $wrapInParenthesises = $afterCloseParenthesisToken->equals('[') || $afterCloseParenthesisToken->isGivenKind(T_POW);
```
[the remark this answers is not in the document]
kalessil: Ah, ok - I see now from the bug-report. If we need to only take care of lines (from the link) before typecasting, then looks good.

And now I wondering if it's time for me to get some sleep or PhpStorm AST tree is incorrect. Let's hope it's the first option ;)

keradus:
Thank you @kubawerlos.

PR #4242, opened 2019-01-05
DX: let's fix UPGRADE.md file

kubawerlos:
To be honest I have found out just today that we have the list of renames in `UPGRADE.md` and it is quite out-of-date.

Let's fix that - of course with the only acceptable way - with adding test it won't get out-of-date again.


keradus:
the upgrade file is valid. you mixed up upgrade for 2.0 with content we have for 3.0:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/UPGRADE-v3.md

kubawerlos:
Oh, in deed it's to 2.0, I assumed it's to latest version, so we would want to not have any deprecated fixers as we remove them from `RuleSet` class.

Issue #4159, opened 2018-12-20
`no_unused_imports` - allow keeping imports from the same namespace


arnaud-lb:
The `no_unused_imports` fix is removing imports of the same namespace.

Removing imports of the same namespace can be harmful: It makes it more difficult to move the code in an other namespace later (because we have to make sure that all classes have been imported).

**And more importantly**, if a class is _implicitly_ imported, it's easy to accidentally shadow it by importing an other class with the same name from an other namespace.

Also, imports are free (in terms of performance), so imports of the same namespace are not really an issue.

I believe that stopping this behaviour, and maybe adding a new fix that imports all used classes, would result in code that's safer to edit.

arnaud-lb:
Since a _feature request_ tag has been added, I've submitted a PR to disable this behaviour by default: #4243. I will also submit a fixer that imports all used classes, later in a separate PR.

kubawerlos:
> It makes it more difficult to move the code in an other namespace later (because we have to make sure that all classes have been imported).

I wouldn't consider it as an issue, it does not seems to be problem for any modern IDE.

> **And more importantly**, if a class is _implicitly_ imported, it's easy to accidentally shadow it by importing an other class with the same name from an other namespace.

What do you mean? Something like [this](https://3v4l.org/e8DXZ):
```php
<?php
namespace Foo;
class Test { const FROM = 'foo'; }

namespace Bar;
class Test { const FROM = 'bar'; }

namespace Foo;
use Bar\Test;
echo Test::FROM;
```
?

arnaud-lb:
> What do you mean? Something like this:

Probably. Just to be sure that we are speaking of the same thing, let's say that we have this code:

```
<?php
// Foo/Hello.php

namespace Foo;

class Hello
{
    public function __construct(Request $request)
    {
        ...
    }
}
```

Notice that `Foo/Hello.php` is not importing `Request`. This is not absolutely required, because PHP will default to using the current namespace. So, `Request` implicitly refers to `Foo\Request` (which was the intent, so it's ok).

Later, in the same file, we want to use `Symfony\Component\HttpFoundation\Request`. So we type `$httpFoundationRequest = new Request()`, and make the steps to import `Request`.

This is where troubles can happen: If you import `Symfony\Component\HttpFoundation\Request` as `Request`, the line `public function __construct(Request $request)` that refered to `Foo\Request` now refers to `Symfony\Component\HttpFoundation\Request`.

[...]

PR #4243, opened 2019-01-07
NoUnusedImportsFixer - Do not remove imports of the same namespace (if configured)

arnaud-lb:
Removing imports of the same namespace is harmful (see #4159).

This PR disables this behaviour by default in the `no_unused_imports` fix. It also adds a `treat_same_namespace_as_unused` option (false by default) that allows to restore the old behaviour.

I will submit a fixed to add imports for classes of the same namespace in a separate PR.

### DIFF
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -13,6 +13,9 @@
 namespace PhpCsFixer\Fixer\Import;
 
 use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Preg;
@@ -26,7 +29,7 @@ use PhpCsFixer\Tokenizer\Tokens;
 /**
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
-final class NoUnusedImportsFixer extends AbstractFixer
+final class NoUnusedImportsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
      * {@inheritdoc}
@@ -110,8 +113,23 @@ final class NoUnusedImportsFixer extends AbstractFixer
                 }
             }
 
-            $this->removeUsesInSameNamespace($tokens, $currentNamespaceUseDeclarations, $namespace);
+            if ($this->configuration['treat_same_namespace_as_unused']) {
+                $this->removeUsesInSameNamespace($tokens, $currentNamespaceUseDeclarations, $namespace);
+            }
         }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('treat_same_namespace_as_unused', 'whether to treat imports in the same namespace as unused'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(true)
+                ->getOption(),
+        ]);
     }
 
     private function importIsUsed(Tokens $tokens, NamespaceAnalysis $namespace, array $ignoredIndexes, $shortName)
--- a/tests/Fixer/Import/NoUnusedImportsFixerTest.php
+++ b/tests/Fixer/Import/NoUnusedImportsFixerTest.php
@@ -1044,6 +1044,925 @@ EOF
     }
 
     /**
+     * @param string      $expected
+     * @param null|string $input
+     *
+     * @dataProvider provideFixTreatSameNamespaceAsUsedCases
+     */
+    public function testFixTreatSameNamespaceAsUsed($expected, $input = null)
+    {
+        $this->fixer->configure([
+            'treat_same_namespace_as_unused' => false,
+        ]);
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixTreatSameNamespaceAsUsedCases()
+    {
+        return [
+            'simple' => [
+                <<<'EOF'
+<?php
+
+use Foo\Bar;
+use Foo\Bar\FooBar as FooBaz;
+use SomeClass;
+
+$a = new Bar();
+$a = new FooBaz();
+$a = new SomeClass();
+
+use Symfony\Annotation\Template;
+use Symfony\Doctrine\Entities\Entity;
+use Symfony\Array123\ArrayInterface;
+
+class AnnotatedClass
+{
+    /**
+     * @Template(foobar=21)
+     * @param Entity $foo
+     */
+    public function doSomething($foo)
+    {
+        $bar = $foo->toArray();
+        /** @var ArrayInterface $bar */
+    }
+}
+EOF
+                ,
+                <<<'EOF'
+<?php
+
+use Foo\Bar;
+use Foo\Bar\Baz;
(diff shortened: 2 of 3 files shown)


SpacePossum:
Hi @arnaud-lb and thanks for all your work here!

Am I correct to see that you change the default behavior of the rule? If so it would be a BC break (at least in this project ;) ) so maybe it would be better to add the option here and later on propose to change the default behavior (on the next major) as it would allow for discussion on that topic :)
Secondly, it would be nice to have test for both configuration options. You added a lot of test already (big :+1: ), but please at test for both values for the new config option starting with the most simple ones.

kubawerlos:
> I will submit a fixed to add imports for classes of the same namespace in a separate PR.

I would downvote such a PR as I believe unused imports are not worth to keep in prospect of [future](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) changing namespace of the class.

arnaud-lb:
@SpacePossum alright, I will change the PR to keep the current behaviour by default :)

The tests I've added are testing both configurations. I will make this more explicit.

arnaud-lb:
@kubawerlos: YAGNI is one of my favorite acronym principles. When following it, we lean towards thinking that if a hypothetical feature is easy to add later, we should do so. This prevents introducing complexity that may not actually be needed in the future, or worse, that may not match the use-cases when the feature is eventually needed.

However, I don't think that it applies here. We are not keeping namespace declarations for a hypothetical need: We are keeping them for an actual, current threat: When some code uses a class of the same namespace, and this class is not explicitly imported, it's easy to accidentally shadow it by importing an other class with the same name from an other namespace.

On the other hand, I don't really understand the need for removing imports of the same namespace. They make the code clearer / more explicit IMHO. I also don't see how they could hurt in any way.

kubawerlos:
@arnaud-lb 
> it's easy to accidentally shadow it

What do you mean? Have you seen my example from the [issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4159#issuecomment-452023050)?

> Our goal as developpers is to make the code easy to read and to change.

Easy to read? Definitely. But easy to change? I would'n say it so important to keep unused imports to be prepared for it.

> I also don't see how they could hurt in any way.

I look at it from the opposite side - they don't do any good - let's get rid of them.

arnaud-lb:
> > it's easy to accidentally shadow it

> What do you mean? Have you seen my example from the issue?

I've just answered to your comment in the issue

> I look at it from the opposite side - they don't do any good - let's get rid of them.

I've shown that in some cases, they do at least some good :) They also have the merit of making things more explicit, which have some value for readability IMHO.

src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -110,10 +113,25 @@ function (NamespaceUseAnalysis $useDeclaration) use ($namespace) {
...
 
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('treat_same_namespace_as_unused', 'whether to treat imports in the same namespace as unused'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(false)
```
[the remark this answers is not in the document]
arnaud-lb: done 👍

tests/Fixer/Import/NoUnusedImportsFixerTest.php
```diff
@@ -29,6 +29,9 @@ final class NoUnusedImportsFixerTest extends AbstractFixerTestCase
      */
     public function testFix($expected, $input = null)
     {
+        $this->fixer->configure([
+            'treat_same_namespace_as_unused' => false,
```
kubawerlos: Why all tests should use non-default configuration?

Please leave all original tests as they were and add new ones with this option set to `false`.

arnaud-lb: done @kubawerlos

arnaud-lb: ping @kubawerlos :)

dmvdbrugge:
Shouldn't this target master instead, as it's a new feature?

arnaud-lb:
@keradus could you explain why the statu quo is good practice, and what kind of benefits it provides ? I'm under the impression that the only reason why imports of the same namespace are bad is because PHPStorm (wrongfully) complains about them.

arnaud-lb:
Also, the [Symfony Coding standards document](https://github.com/symfony/symfony-docs/blame/23b409b325e675a8d7e1894cc8300f8775c3b235/contributing/code/standards.rst#L197) clearly mentions that a `use` statement must be added for every class that is not part of the global namespace:

```
* Add a ``use`` statement for every class that is not part of the global namespace;
```

I think that this is the best thing to do regarding `use` statements:

 - Having an import statement for every class, including classes of the same namespace, makes the code more explicit, especially without an IDE (for example: during code review on Github). With an IDE, this make no difference / no harm (functionally, and also visually because use statements are collapsed).
 - Code that is lacking explicit use statements for classes of the same namespace is trickier to move, because we can lose track of exactly which classes are referenced, and mistakes can be made while adding these statements after the fact, resulting in broken code.
 - And, more importantly, code that uses implicit imports is prone to [accidental shadowing](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4159#issuecomment-452699451), resulting in broken code.

In short, enforcing the addition of `use` statements of the same namespace makes the code more explicit, and less prone to breakages.

On the other hand, removing `use` statements of the same namespace has no benefits at all, because:

 - Use statements have zero overhead. They don't result in any file include or anything else. The PHP compiler uses them to resolve class names at compile time. At run time, all class names are already in full-qualified-class-name form; use statements do not exist anymore at this stage. (In case the phrase "compile time" feels weird here, remember that the PHP interpreter does not interpret the code directly: If first needs to compile it to byte code, that is then executed by the PHP virtual machine.)
 - Use statements are invisible in advanced environments, like IDEs, so the extra lines don't hurt the reader.
 - In more basic environments, like text editors or Github, use statements are important information, because the users have no tools to resolve the FQCN referenced by a class name in the code. In these environments, the only way to know exactly what class is used by some code is to look at use statements; and the lack of them is really harmful.

keradus:
sorry, but sadly i have not enough time to dive super deep into this discussion.

Idea behind adding ref statement into docs of Sf was to define to not import global classes, just that.
Golden rule of Sf CS is to mimic what you see around - and whole codebase of Sf is following, how you called it, `treat_same_namespace_as_unused`.
The rule was originally introduced into PHP CS Fixer by Fabien, creator of Symfony, and it was one of very first rules in the repo (even before v1 was born).
For that, I don't consider your argument about Sf CS to be valid.

I do understand your point of view and your argument. I'm simply not agreeing with it. I would - if you would enforce the need of explicit import inside same namespace on php language itself, but it's not there. I agree - I would like to have it working like that, but it's not.

After saying that, keep in mind that we want PHP CS Fixer to serve multiple ppl, because the tool is already quite complex and not easy to maintain. For that, any change shall serve multiple ppl, not just me or just you. If this functionality (in terms "i want to use this feature", not "code of PR looks OKish") will gather more supporters, I will reconsider to merge it.
Thank you for understanding.

stollr:
Although this discussion happened long ago. I wanted to show my support for this option. Keeping imports of classes of the same namespace is useful for refactoring in growing projects. I explicitly use them and it is a bit frustrating if a tool forces you to change your code style and it is not possible to adjust it :-/

Pixelao:
I wanted to express my support for this approach. Having imports from the same namespace can be beneficial in some cases. My team deliberately employs this method.

In my opinion, this behavior should be configurable; maybe not the default option, but at least there should be an option to choose.

Wirone:
@stollr @Pixelao I've reopened the issue, but PR is massively outdated and requires work. Maybe you would like to contribute and refresh this change (behind the config option)?

PR #4245, opened 2019-01-07
LineEndingFixer - BracesFixer - Priority

dmvdbrugge:
Separated the `line_ending`/`braces`-priority relation and some comments from #4085

### DIFF
--- a/src/Fixer/Basic/BracesFixer.php
+++ b/src/Fixer/Basic/BracesFixer.php
@@ -126,7 +126,7 @@ class Foo
      */
     public function getPriority()
     {
-        // should be run after the ElseIfFixer, NoEmptyStatementFixer and NoUselessElseFixer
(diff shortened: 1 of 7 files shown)


SpacePossum:
Thanks for all the proviling!
I didn't catch up on all the bigger PR's and comments, but I wander; can't avoid this conflict if the `braces` uses the line ending that `line_ending` enforces? Just curious :)

dmvdbrugge:
I just checked, braces never _inserts_ hardcoded newlines, they come from the whitespace config. It does do some string-magic with `\n`s though, however at the moment I wouldn't know how to actually fix that (without spending hours for deeply understanding `braces`).

SpacePossum:
note to self, 7.3 failures are not related to this PR

SpacePossum:
Thank you @dmvdbrugge.

Issue #3996, opened 2018-09-20
php-cs-fixer chokes on this file


flack:
As discussed in #3224, here's my standalone report:

#### The PHP version you are using (`$ php -v`):
```
PHP 7.2.10-0ubuntu0.18.04.1 (cli) (built: Sep 13 2018 13:45:02) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.10-0ubuntu0.18.04.1, Copyright (c) 1999-2018, by Zend Technologies
    with Xdebug v2.6.0, Copyright (c) 2002-2018, by Derick Rethans
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
```
PHP CS Fixer 2.13.0 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski
```

#### The command you use to run PHP CS Fixer:
```
php-cs-fixer fix -vvv --rules=no_spaces_after_function_name,elseif,braces,encoding,full_opening_tag,single_blank_line_at_eof,function_declaration,line_ending,method_argument_space,no_trailing_whitespace_in_comment,no_closing_tag,switch_case_semicolon_to_colon,switch_case_space,no_trailing_whitespace,whitespace_after_comma_in_array,whitespace_after_comma_in_array,no_whitespace_before_comma_in_array,no_empty_statement t_user.php 
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):

I had to rename the file to `.txt`, otherwise github wouldn't let me upload it. Should be changed back beofre trying to reproduce

[t_user.txt](https://github.com/FriendsOfPHP/PHP-CS-Fixer/files/2401190/t_user.txt)

keradus:
worst case scenario: skip files that are bigger than 1MB or sth...

flack:
@keradus I don't think the file size is the problem here. I just deleted half the lines and ran again (630 KB), and still nothing. 

My first try to anonymize the file was to do `file_get_contents`, replace everything with `x` and write it back to a new file with `file_put_contents`. This new file (also 1.4MB) is processed by `php-cs-fixer` in 0.172s, whereas the version attached here never seems to finish (longest I tried was 10 minutes). My (completely uneducated) guess would be that it has something to do with the encoding

flack:
[...]

PR #4250, opened 2019-01-09
POC Tokens::insertSlices

keradus:
ref #4221
cc @kubawerlos , this is what we discussed in priv


tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
```diff
@@ -257,6 +257,20 @@ public function bar($foo)
...
+',
+            ],
+            'fix multiple calls in single code' => [
+                '<?php
+
+\json_encode($foo);
+\strlen($foo);
+\strlen($foo);
+',
+                '<?php
```
keradus: This, unchanged `Tokens` collection is size of 20.
The fixer has to add 3 new tokens to this collection.

original code has to run `offsetSet` 45 times, as it has to move all tokens from strlen, then middle strlen, then from json_encode - which means that last strlen was moved 3 times!
my proposal has to run `offsetSet` only 20 times, as it has to move one token no more than once

keradus:
why I believe this is a good idea?
I pushed a fix for #3996 to this PR
see the time difference without and with this PR:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3996#issuecomment-455617637

tests/AutoReview/FixerTest.php
```diff
@@ -215,6 +216,79 @@ public function testDeprecatedFixersHaveCorrectSummary(FixerInterface $fixer)
...
 
+    /**
+     * Blur filter that find candidate fixer for performance optimization to use only `insertSlices` instead of multiple `insertAt` if there is no other collection manipulation.
+     *
+     * @dataProvider provideFixerDefinitionsCases
+     */
+    public function testFixerUseInsertSlicesWhenOnlyInsertionsArePerformed(FixerInterface $fixer)
+    {
+        $reflection = new \ReflectionClass($fixer);
+        $tokens = Tokens::froMCode(file_get_contents($reflection->getFileName()));
```
dmvdbrugge: Typo: uppercase `M` here 😉

keradus: nice catch :+1:

src/Fixer/ArrayNotation/WhitespaceAfterCommaInArrayFixer.php
```diff
@@ -48,35 +48,32 @@ public function isCandidate(Tokens $tokens)
...
-                $tokens->insertAt($i + 1, new Token([T_WHITESPACE, ' ']));
+            for ($i = $endIndex - 1; $i > $startIndex; --$i) {
+                $i = $this->skipNonArrayElements($i, $tokens);
+                if ($tokens[$i]->equals(',') && !$tokens[$i + 1]->isWhitespace()) {
+                    $tokensToInsert[$i + 1] = new Token([T_WHITESPACE, ' ']);
+                }
             }
         }
+
+        if ($tokensToInsert) {
```
kubawerlos: `[] === $tokensToInsert`?

src/Tokenizer/Tokens.php
```diff
@@ -869,7 +869,27 @@ public function findSequence(array $sequence, $start = 0, $end = null, $caseSens
     public function insertAt($index, $items)
     {
         $items = \is_array($items) || $items instanceof self ? $items : [$items];
-        $itemsCnt = \count($items);
+
+        $this->insertSlices([$index => $items]);
+    }
+
+    /**
+     * @TODO: docs
```
kubawerlos: do?

src/Tokenizer/Tokens.php
```diff
@@ -869,7 +869,27 @@ public function findSequence(array $sequence, $start = 0, $end = null, $caseSens
...
+     *
+     * @internal
+     *
+     * 5 => tokenA
+     * 7 => [tokenB, tokenC]
+     *
+     * @param mixed $slices
```
kubawerlos: `array<int, Token>`?

src/Tokenizer/Tokens.php
```diff
@@ -869,7 +869,27 @@ public function findSequence(array $sequence, $start = 0, $end = null, $caseSens
...
+     */
+    public function insertSlices($slices)
+    {
+        $itemsCnt = 0;
```
kubawerlos: `$itemsCount` (or `$itemCount`) - the point is not to abbreviate.

src/Tokenizer/Tokens.php
dmvdbrugge: If I look at the description it should be `array<int, Token|Token[]>`

keradus: indeed, i simply didn't documented this methods (as I put todo), as i wanted the idea to be validated first ;)

src/Fixer/ArrayNotation/WhitespaceAfterCommaInArrayFixer.php
dmvdbrugge: You mean `!==`

keradus: done

src/Tokenizer/Tokens.php
keradus: done

src/Tokenizer/Tokens.php
keradus: done

src/Tokenizer/Tokens.php
```diff
@@ -869,31 +869,61 @@ public function findSequence(array $sequence, $start = 0, $end = null, $caseSens
...
 
         // since we only move already existing items around, we directly call into SplFixedArray::offset* methods.
         // that way we get around additional overhead this class adds with overridden offset* methods.
-        for ($i = $oldSize + $itemsCnt - 1; $i >= $index; --$i) {
-            $oldItem = parent::offsetExists($i - $itemsCnt) ? parent::offsetGet($i - $itemsCnt) : new Token('');
-            parent::offsetSet($i, $oldItem);
-        }
+        foreach ($slices as $index => $slice) {
+            $slice = \is_array($slice) || $slice instanceof self ? $slice : [$slice];
+            $sliceCnt = \count($slice);
```
dmvdbrugge: Nonblocking: I guess this should become `sliceCount` as well, given @kubawerlos's reasoning

src/Tokenizer/Tokens.php
```diff
@@ -869,31 +869,61 @@ public function findSequence(array $sequence, $start = 0, $end = null, $caseSens
...
-        if (0 === $itemsCnt) {
+        $this->insertSlices([$index => $items]);
+    }
+
+    /**
+     * Insert a slices or individual Tokens into multiple places in a single run.
+     *
+     * @internal
+     *
+     * @param array<int, Token|Token[]> $slices
```
dmvdbrugge: What I'm missing in this doc is the fact that this `int` key is used as insert-index, while it is mentioned in for example `insertAt`. Leaving it out here may cause one to think that it's not the same idea.

Also, isn't `Tokens` also accepted, thus shouldn't the typehint be `array<int, Token|Token[]|Tokens>`?

keradus:
@julienfalque , @SpacePossum , what do you think about the proposal in this PR? especially do we want to promote new method as new "best practice" or not?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4250/files#diff-0ae0077906c99aaf35085e05c46cefd3R284

> I think the issue is not related to the overall length, so all keys/values can be probably changed to single char like `0` (zero) or `''` (empty string).

it can, i was just too lazy to modify example provided by original reporter.

src/Tokenizer/Tokens.php
keradus: :+1:

keradus:
> Is the name `insertSlices` fitting situation? Quite often (if not most times) we will use this to insert array of tokens. Wouldn't `insertMultiple` (or `insertMany`) better describe what the function is doing?

for me, `insertMultiple`/`insertMany` is more fitting actual behaviour of `insertAt`, where we are pointing a single index, and adding multiple tokens to it:
```php
$tokens->insertAt(15, $tokensCollection)
$tokens->insertAt(10, [$tokenA, $tokenB]);
$tokens->insertAt(5, $tokenC);
```

I'm trying to express that we are inserting different slices to different indexes at single call:
```php
$tokens->insertSlices([
    5 => $tokenC,
    10 => [$tokenA, $tokenB],
    15 => $tokensCollection,
]);
```

mvorisek:
I was hoping this PR with solve https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5279, but even with this PR running CS Fixer against https://github.com/mvorisek/TCPDF/tree/psr12_cs is slow and CS Fixer does not finish within an hour.

keradus:
@mvorisek :
- please avoid thread hijacking
- I already shared that this PR is NOT helping with topic you raised - https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5279#issuecomment-740656516
- if you want your edge-case topic covered, consider the suggestion raised here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4702#issuecomment-740749024

PR #4251, opened 2019-01-10
NativeFunctionInvocationFixer - Don't consider imported functions global

### LINKED ISSUE #3051: NativeFunctionInvocationFixer is incompatible with imported functions
The following code would add a `\` while the function being called is known *not* to be the internal one at compile time, due to the use statement:

```php
namespace Foo;

use function GuzzleHttp\json_encode;

echo json_encode('foo');
```

Triggering this bug requires having a function with the same name than an internal function, but this happens.

jyggen:
This PR fixes #3051 by making sure that `FunctionsAnalyzer::isGlobalFunctionCall()` only returns `true` if the function name isn't imported using `use function`.

Running `NamespaceUsesAnalyzer::getDeclarationsFromTokens()` on the same set of tokens over and over again (in for example `NativeFunctionInvocationFixer`, where `FunctionsAnalyzer::isGlobalFunctionCall()` is called within a loop) feels suboptimal, but not sure what a better solution would be.


tests/Fixer/FunctionNotation/NativeFunctionInvocationFixerTest.php
```diff
@@ -564,6 +564,38 @@ public function & strlen($name) {
...
+    {
+        $this->doTest($expected, $input);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideFixWithImportedFunctionCases()
+    {
+        return [
```
julienfalque: Those test cases don't prove the bug you are fixing: by default, the fixer only adds the leading `\` to native PHP functions, which `bar()` and `baz()` aren't. You should configure the fixer before calling `doTest()`:
```php
$this->fixer->configure([
    'include' => ['@all'],
]);
```
I would also add a test case with a native PHP function such as `json_decode()`.

bendavies:
just came across this bug. thanks for the fix!

SpacePossum:
Thanks @jyggen for all the work!\
I made a different fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4846 as such I'm closing this one, thanks again!

PR #4259, opened 2019-01-13
PsrAutoloadingFixer - introduction

kubawerlos:
As `psr_0` and `psr_4` fixers share a lot of code after discussion with @keradus we want to merge them and introduce this fixer to replace them.


src/Fixer/Basic/PsrAutoloadingFixer.php
```diff
@@ -0,0 +1,213 @@
...
+                    '<?php
+namespace PhpCsFixer\FIXER\Basic;
+class InvalidName {}
+',
+                    new \SplFileInfo(__FILE__),
+                    ['dir' => realpath(__DIR__.'/../..')]
+                ),
+            ],
+            null,
+            'This fixer may change your class name, which will break the code that is depended on old name.'
```
localheinz: I realize that this is copied from the deprecated fixer, but what do you think about adjusting the wording?

I think it should be either `is dependent on` or `depends on` - what do you think?

src/Fixer/Basic/Psr4Fixer.php
```diff
@@ -12,89 +12,57 @@
...
-class InvalidName {}
-',
-                    new \SplFileInfo(__FILE__)
-                ),
-            ],
-            null,
-            'This fixer may change your class name, which will break the code that is depended on old name.'
+            \array_slice($this->fixer->getDefinition()->getCodeSamples(), 0, 1),
+            $this->fixer->getDefinition()->getDescription(),
+            $this->fixer->getDefinition()->getRiskyDescription()
```
localheinz: Not that it is going to save a ton of time, but what do you think about introducing a temporary variable for the proxied fixer's definition?

src/Fixer/Basic/PsrAutoloadingFixer.php
kubawerlos: Let's change to the latter (as it is shorter).

SpacePossum:
@kubawerlos can rebase and squash this PR for me?

I was wondering about two things:
- PSR0 en PSR4 have the same code samples, yet one has only, why?
- with PSR0 and PSR4 deprecated, how to configure the new one to do either PSR0 or PSR4?

kubawerlos:
@SpacePossum rebased and squashed

> PSR0 en PSR4 have the same code samples, yet one has only one, why is that?

PsrAutoloadingFixer has 2 code samples - 1st without configuration and 2nd with configuration and Psr4Fixer is not configurable fixer, so it cannot have the 2nd sample

> with PSR0 and PSR4 deprecated, how to configure the new one to do either PSR0 or PSR4?

If you use PsrAutoloadingFixer without `dir` then it works as `psr0` or `psr4` without config - that was the reason for the merge - both work the same with default config. Adding `dir` makes the fixer works as `psr0` with a `dir`.

SpacePossum:
thanks for explaining, I leave the review to someone with more knowledge on this topic

src/Fixer/Basic/Psr0Fixer.php
```diff
@@ -12,146 +12,65 @@
...
+final class Psr0Fixer extends AbstractProxyFixer implements ConfigurationDefinitionFixerInterface, DeprecatedFixerInterface
 {
+    /**
+     * @var PsrAutoloadingFixer
+     */
+    private $fixer;
+
+    public function __construct()
+    {
+        $this->fixer = new PsrAutoloadingFixer();
```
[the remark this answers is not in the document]
kubawerlos: @julienfalque then for method `configure` I'd have to change code to something like:
```php
reset($this->proxyFixers)->configure($configuration);
```
and something similar for `getDefinition` method - or having the original content of this method which would be copy/paste with the new fixer.

I'd figured that adding `fixer` property would be more readable.

kubawerlos:
@keradus is this still something you would like to have merged?

SpacePossum:
@kubawerlos please rebase, after that RTM :+1:

kubawerlos:
@SpacePossum rebased

SpacePossum:
Thank you @kubawerlos.

PR #4261, opened 2019-01-14
CI: Use Ubuntu 16.04

carusogabriel:
Ubuntu 16.04 is already available for usage: https://docs.travis-ci.com/user/reference/xenial/#using-xenial.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 language: php
 
 git:


SpacePossum:
Thanks for bringing this to our attention :)

keradus:
if merged, then for 2.12

yet, the fact it exist doesn't mean we have to use it.
why would we? no reasoning was provided, I don't see a need to use non-default os as well.
I'm tend to agree with https://github.com/sebastianbergmann/phpunit/pull/3476 .
I don't want to remember that i need to manually drop it, when 18th would be default, but we have configured 16.

keradus:
Closing due to discussion.

Thank you @carusogabriel

PR #4263, opened 2019-01-14
DX: AppVeyor - bump PHP version

keradus:


dmvdbrugge:
Do we really want to run only a single php7 case? It's a matrix right, why not run all off 7.x? (Yes I know those are on travis as well, but do we really want to support latest-only for windows?)

keradus:
to not waste time
we care about unix users most.
then, we have basic checks for Windows and Mac.

dmvdbrugge:
Then isn't there a specifier for "latest stable" or something? I mean, manually upping all the time shouldn't be required either right?

keradus:
i don't know it. i don't also need to always have last version there - i just wanted to have 5.x latest, 7.x update is by the way

Issue #3026, opened 2017-09-06
lowercase_constants: Enable different casing for true/false and null constants.


RomanHodulak:
In the company where I work, we use lowercase true/false constants and uppercase NULL constant. Since lowercase_constants cannot differentiate between those, it is impossible for us to set proper rules.

Could you add the option to set different casing for null and true/false?

keradus:
big :-1:
it's violating PSR-2 in general and consistency of casing those trivials

SpacePossum:
lets see if there is a bigger community which follows this conversion, till than I'm +0

SpacePossum:
I'm closing this as there does not seem a bigger use of this standard. This doesn't mean I'm judging the standard wrong or whatever.

However the scope of this project is to provide a way to fix PHP code to a wide used standard. Besides advocating usage of wide used standards it is also we'll have more support for maintaining code related to such standards as there is a bigger community interest in it.

You can always create a custom fixer to suit your project needs, copy-pasting the fixer, renaming it and adjust it to fix `null` to uppercase shouldn't be hard. If you decide to give it a try please let me know if you want me to help out with if you get stuck with it.

RomanHodulak:
Thank you for your comments. We've discussed this and ended up changing the standard, so our code uses all these constants in lowercase 🙂

SpacePossum:
great to hear :)  thanks for reporting back!

PR #4275, opened 2019-01-21
Issue #4274: Let lowercase_constants directive to be configurable.

### LINKED ISSUE #4274: lowercase_constants: Enable configuration.
This issue is to have the `lowercase_constants` directive configurable. (Issue #3026)
The configurable option would let you choose in between 'upper' or 'lower' case syntax.

Here's an example on how it would be:

```yaml
lowercase_constants:
   syntax: upper
```
or
```yaml
lowercase_constants:
   syntax: lower
```

The default would be `lower` to be backward compatible.

Why this request ?
CMS's like Drupal 8 are using uppercase constants (TRUE, FALSE, NULL) and I would like to have a proper tool that follows the Drupal 8 "standard" in our custom code at European Commission.

=> PHP 7.2.14

=> PHP CS Fixer 2.14.1-DEV Sunrise by Fabien Potencier and Dariusz Ruminski

=> Via grumphp

```yaml
        array_syntax:
            syntax: short
        blank_line_after_opening_tag: true
        blank_line_before_statement:
            statements:
                - break
                - continue
                - declare
                - return
                - throw
                - try
        include: true
        # Drupalism:
        class_attributes_separation: false
        declare_equal_normalize:
            space: single
        declare_strict_types: true
        is_null: true
        lowercase_static_reference: true
        # Drupalism: lowercase_constants (unable to configure it in php-cs-fixer. Provide a patch?)
        native_function_invocation: false
        # Drupalism:
        no_blank_lines_after_class_opening: false
        no_closing_tag: true
        no_extra_blank_lines:
            tokens:
                - extra
                # Drupalism: - curly_brace_block
                - parenthesis_brace_block
                - square_brace_block
                - return
                - throw
                - use
                - use_trait
        no_trailing_whitespace_in_comment: true
        no_unneeded_final_method: true
        single_blank_line_at_eof: true
        single_line_after_imports: true
        strict_comparison: true
        strict_param: true
        trailing_comma_in_multiline_array: true
        yoda_style:
            equal: false
            identical: false
            less_and_greater: false
            always_move_variable: false
```

```php
if ($var === null) {
  // Do something
}
```

```php
if ($var === null) {
  // Do something
}
```

```php
if ($var === NULL) {
  // Do something
}
```

drupol:
Fix #4274 

Depends on #4362 

This issue is to have the `lowercase_constants` directive configurable. (Issue #3026)
The configurable option would let you choose in between 'upper' or 'lower' case syntax.

Here's an example on how it would be:

```yaml
lowercase_constants:
   case: upper
```
or
```yaml
lowercase_constants:
   case: lower
```

The default would be `lower` to be backward compatible.

Why this request ?
CMS's like Drupal 8 are using uppercase constants (TRUE, FALSE, NULL) and I would like to have a proper tool that follows the Drupal 8 "standard" in our custom code at European Commission.

=> PHP 7.2.14

=> PHP CS Fixer 2.14.1-DEV Sunrise by Fabien Potencier and Dariusz Ruminski

=> Via grumphp

```yaml
        array_syntax:
            syntax: short
        blank_line_after_opening_tag: true
        blank_line_before_statement:
            statements:
                - break
                - continue
                - declare
                - return
                - throw
                - try
        include: true
        # Drupalism:
        class_attributes_separation: false
        declare_equal_normalize:
            space: single
        declare_strict_types: true
        is_null: true
        lowercase_static_reference: true
        # Drupalism: lowercase_constants (unable to configure it in php-cs-fixer. Provide a patch?)
        native_function_invocation: false
        # Drupalism:
        no_blank_lines_after_class_opening: false
        no_closing_tag: true
        no_extra_blank_lines:
            tokens:
                - extra
                # Drupalism: - curly_brace_block
                - parenthesis_brace_block
                - square_brace_block
                - return
                - throw
                - use
                - use_trait
        no_trailing_whitespace_in_comment: true
        no_unneeded_final_method: true
        single_blank_line_at_eof: true
        single_line_after_imports: true
        strict_comparison: true
        strict_param: true
        trailing_comma_in_multiline_array: true
        yoda_style:
            equal: false
            identical: false
            less_and_greater: false
            always_move_variable: false
```

```php
if ($var === null) {
  // Do something
}
```

```php
if ($var === null) {
  // Do something
}
```

```php
if ($var === NULL) {
  // Do something
}
```


src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,20 @@
  *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
-final class LowercaseConstantsFixer extends AbstractFixer
+final class LowercaseConstantsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
+    private $fixCallback;
```
dmvdbrugge: This misses (a doc block with) a typehint.

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,20 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->resolveFixCallback();
```
dmvdbrugge: Why the extra method, and not just place the oneliner here?

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -50,6 +78,8 @@ public function isCandidate(Tokens $tokens)
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
+        $callback = $this->fixCallback;
```
dmvdbrugge: I'm having a little issue with the naming of all this. It's a callable yes, but it is in no way a _callback_. It's (just) the fix function.

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -99,4 +129,9 @@ private function isNeighbourAccepted(Tokens $tokens, $index)
 
         return !$token->isGivenKind($forbiddenTokens);
     }
+
+    private function resolveFixCallback()
+    {
+        $this->fixCallback = sprintf('strto%s', strtolower($this->configuration['syntax']));
```
dmvdbrugge: I'm not sure I like this "clever" way.

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,20 @@
  *
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
  */
-final class LowercaseConstantsFixer extends AbstractFixer
+final class LowercaseConstantsFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
```
dmvdbrugge: I find it a bit strange to have a `Lowercase...Fixer` that _uppercases_ things, but I guess that's how it has to be seeing that renaming is not an option.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: I will fix today after work.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Good point. Will do.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Ok, will make something meaningful.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Ok, got it.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Indeed, had the same thing in mind. I will update accordingly today evening.

drupol:
Branch has been rebased against master, as requested by @dmvdbrugge in first review comment.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Fixed.

src/Fixer/Casing/LowercaseConstantsFixer.php
dmvdbrugge: It's still not a `$callback` 😉

drupol: I renamed it to `$fixFunction`, do you suggest something else ?

src/Fixer/Casing/LowercaseConstantsFixer.php
dmvdbrugge: The reason `lowercase_constants: false` doesn't work is because that disables the rule (meaning no fixing either way). But that was not my point, as even if that were a way to configure it, it would still be a `Lowercase...Fixer` that uppercases things.

And like I said, renaming is not an option (because of BC), so there is no solution.

src/Fixer/Casing/LowercaseConstantsFixer.php
dmvdbrugge: You renamed the property but not the variable, was the point I unsuccessfully was trying to make.

drupol: oops, fixed now.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Mmh... I understand but what I had in mind was the following:

* Not having `lowercase_constants`: Do not do anything
* Having `lowercase_constants: false` means to uppercase them
* Having `lowercase_constants: true` means to lowercase them

drupol: Ok, thanks for the explanation, make sense.

SpacePossum:
Hi and thanks for opening this PR.

Before I review the PR in detail, I'm wondering. The suggestion to uppercase the keywords is not in line with [PSR2](https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull) , however this is part of the Drupal standard? Do you know why Drupal prefers this style or if this is legacy they might want to move away from?

drupol:
Hello,

No I don't know why using uppercase for keywords was chosen in Drupal.
I'm pretty sure that it won't change in Drupal 7 and 9 and I don't know either if they are going to move away from it in Drupal 9.
However, I really wish they do.

That said, having this configurable in PHP-CS-Fixer shouldn't be a bad thing, I think.

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,25 @@
...
+    private $fixFunction;
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->fixFunction = 'strto'.strtolower($this->configuration['syntax']);
```
kubawerlos: Looks like magic to me - concatenating 2 strings to to have function name.

I'd use anonymous function instead.

drupol: Going to fix this now.

drupol: @kubawerlos Could you check again, I just pushed an update. Thx.

drupol: Just for my personnal information, could you also explain why, in this specific case, using an anonymous function would be better ?

SpacePossum: we prefer lambda functions because;
- these are more explicit
- static code analyzers understand the code

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,29 @@
...
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $functionName = 'strto'.strtolower($this->configuration['syntax']);
+
+        $this->fixFunction = function ($string) use ($functionName) {
+            return $functionName($string);
+        };
```
SpacePossum: I think @kubawerlos meant something like:

```
if ('upper' === $this->configuration['syntax']) {
     $this->fixFunction = static function ($string) {
          return strtoupper($string);
     };
} else {
     $this->fixFunction = static function ($string) {
          return strtolower($string);
     };
}
```

so not calling a function by using string operations to get to the name :)

SpacePossum:
I'm +1 as this is a standard of a bigger community (Drupal) even if against PSR, however I know of and respect the opinions of people who feel different about this change.

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Perfect, thanks @SpacePossum !

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Ok I will fix it asap.

dmvdbrugge:
Final nitpicking remark from me: I find "syntax" a strange choice for configuration option, "case" or "casing" would be more clear and intuitive.

drupol:
Ok guys, I will fix this, it will be ready for tomorrow.

SpacePossum:
(maybe `case` would be best, as we already used that as config option else where, but `casing` we did not)

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: Fixed, thanks for the insight !

drupol:
@dmvdbrugge @SpacePossum Ready for review again.

SpacePossum:
please regenerate the `readme` with `php php-cs-fixer readme > README.rst` :)

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -24,8 +27,33 @@
...
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        $this->fixFunction = static function ($string) {
+            return strtolower($string);
+        };
+
+        if ('upper' === $this->configuration['case']) {
```
SpacePossum: I think `if else` would be a better fit, so the first lambda is not created if we run with the `upper` case

drupol: Ok, I thought it was a good idea as 'lower' will be the default value.
Going to do the change now.

drupol: Fixed, waiting for the tests.

drupol:
All good @SpacePossum

src/Fixer/Casing/LowercaseConstantsFixer.php
```diff
@@ -59,7 +102,7 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 $this->isNeighbourAccepted($tokens, $tokens->getPrevMeaningfulToken($index)) &&
                 $this->isNeighbourAccepted($tokens, $tokens->getNextMeaningfulToken($index))
             ) {
-                $tokens[$index] = new Token([$token->getId(), strtolower($token->getContent())]);
+                $tokens[$index] = new Token([$token->getId(), $fixFunction($token->getContent())]);
```
keradus: -1 for calling lambdas like that, no way to properly run SCA on `$lambda(...)` call.

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+namespace PhpCsFixer\Tokenizer\Analyzer;
+
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\TypeAnalysis;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @internal
+ */
+final class ClassAnalyzer
```
keradus: `class` or `classy` ? originally code here was taken from `getClassyDefinitionInfo`.

can i analyse only an class, or interface and trait as well ?

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+{
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @param $classIndex
+     *
+     * @return array
```
keradus: array of.... ?

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+     */
+    public function getClassDefinition(Tokens $tokens, $classIndex)
+    {
+        return $this->getClassDefinitionType($tokens, $classIndex);
```
keradus: no proxy methods, please.
this brings no value

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+
+        return new TypeAnalysis($type, $typeStartIndex, $typeEndIndex);
+    }
+
+    /**
+     * @param \PhpCsFixer\Tokenizer\Tokens $tokens
+     * @param $startIndex
+     * @param $label
+     *
+     * @return array
```
keradus: array of ?

keradus: especially when it's assoc

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+                $implementsInfo['multiLine'] = true;
+            }
+        }
+
+        return $implementsInfo;
+    }
+
+    /**
+     * @param $tokens
+     * @param $classIndex
```
keradus: no types?

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
...
+     *
+     * @return array
```
keradus: of what ?

tests/Fixer/Import/FullyQualifiedStrictTypesFixerTest.php
```diff
@@ -47,6 +47,39 @@ public function testCodeWithoutReturnTypes($expected, $input = null)
...
+     *
+     * @param mixed      $expected
+     * @param null|mixed $input
+     */
+    public function testCodeWithExtendsTypes($expected, $input = null)
+    {
+        $this->doTest($expected, $input);
+    }
+
+    public function provideCodeWithExtendsTypesCases()
```
keradus: can you please raise the PR to add it to 2.12 line?
i don't see how this test is related to feature you are adding, so let's push it to as lowest branch as possible

README.rst
```diff
@@ -812,6 +812,11 @@ Choose from the list of available rules:
 
   The PHP constants ``true``, ``false``, and ``null`` MUST be in lower case.
 
+  Configuration options:
+
+  - ``case`` (``'lower'``, ``'upper'``): whether to use the ``upper`` or ``lower`` case
```
keradus: so, `lowercase_constants` will make sth uppercase ?
we shall rename in BC way

src/Tokenizer/Analyzer/ClassAnalyzer.php
```diff
@@ -0,0 +1,132 @@
+<?php
```
keradus: I will leave it as a blocker here
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4310/files#r271581322

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: What do you have in mind then? What do you prefer?

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: It's only for classes.

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: Fixed in #4362

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: Fixed in #4362

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: Fixed in #4362

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: Fixed in #4362

src/Tokenizer/Analyzer/ClassAnalyzer.php
drupol: Fixed in #4362

README.rst
drupol: What do you propose then?

src/Fixer/Casing/LowercaseConstantsFixer.php
drupol: I do not use anonymous function anymore.

tests/Fixer/Import/FullyQualifiedStrictTypesFixerTest.php
drupol: This was probably a mistake that I did, I rebased the branch and removed unrelated code, please discard it then.

drupol:
@keradus @dmvdbrugge @SpacePossum Any of you guys to do a new review ?

README.rst
localheinz: @drupol 

For an example, see https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.15/src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php.

drupol: @localheinz Thanks for the example, but I don't understand what you're trying to show me :-)

SpacePossum: I think, what is meant is that the name of the rule `lowercase_constants`  feels odd for something that might make things upper case. The rename in BC form would;
- think of a new rule name, like `constants_case` for example
- create the file for the fixer by that name and copy paste everything from the current fixer you got here to the new file
- change the class name etc. in that file so it matches and passes
- (do the same for the test files of the fixer)
- than turn the current fixer into a proxy fixer, which directs calls on it to the new one (sample of a proxy fixer is [MethodSeparationFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.12.7/src/Fixer/ClassNotation/MethodSeparationFixer.php) which proxies to ClassAttributesSeparationFixer)

I can help do all of that. However I think first we need to get some more :+1:  's to prevent doing all that work while not making the merge.

drupol: Thanks @SpacePossum , I've implemented everything in my last commit and did a rebase as well.

drupol: All the tests are passing except on Travis, it seems that it's unrelated to this PR.

src/Fixer/Casing/ConstantCaseFixer.php
```diff
@@ -0,0 +1,143 @@
...
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * Fixer for rules defined in PSR2 ¶2.5.
```
kubawerlos: Now, this is misleading - it suggests the fixer is to make code follow PSR2, but only default configuration does that, with "upper" option it is to break PSR2.

src/Fixer/Casing/ConstantCaseFixer.php
```diff
@@ -0,0 +1,143 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        if ('lower' === $this->configuration['case']) {
+            $this->fixFunction = 'strtolower';
```
kubawerlos: Can we have closure here? AFAIK we avoid having code as a string in the project.

src/Fixer/Casing/ConstantCaseFixer.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        if ('upper' === $this->configuration['case']) {
```
kubawerlos: `else` instead of this `if`?

src/Fixer/Casing/ConstantCaseFixer.php
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'The PHP constants `true`, `false`, and `null` MUST be in lower case.',
```
kubawerlos: "lower" -> "correct" / "proper"

src/Fixer/Casing/ConstantCaseFixer.php
drupol: Fixed. I also updated the author. It was just a copy paste of the previous Fixer.

src/Fixer/Casing/ConstantCaseFixer.php
drupol: Fixed.

src/Fixer/Casing/ConstantCaseFixer.php
drupol: I'm always trying to avoid `else|elseif` conditions to have a easier code to read and to usually reduce the cyclomatic complexity. Is this really necessary ?

src/Fixer/Casing/ConstantCaseFixer.php
drupol: Reworded.

SpacePossum:
Thank you @drupol.

drupol:
THANKS !!!

PR #4277, opened 2019-01-22
Change method arg alignment

Billz95:
Fixes #3712 
As mentioned in the above issue, the code will not be aligned properly when forced into multiline (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3712#issue-316138499)
```php
<?php
function a($a, $b,
           $c, $d) {
}
```
would be fixed to
```diff
--- Original
+++ New
@@ @@
 <?php

-function a($a, $b,
-           $c, $d) {
+function a(
+    $a,
+    $b,
+           $c,
+    $d
+) {
 }
```
However, the expected result is 
```diff
--- Original
+++ New
@@ @@
 <?php

-function a($a, $b,
-           $c, $d) {
+function a(
+    $a,
+    $b,
+    $c,
+    $d
+) {
 }
```
This is caused by function `MethodArgumentSpaceFixer::fixNewline`, where it simply returns without applying any fix when the token after `,` is newline.

This is not ideal, from the above code snippet, we can see the token after `$b,` is `newline + many spaces`. Instead, what we expect should be `newline + indentation`.

Therefore, by simply removing `$this->isNewline($token[$index + 1])` from early return check, it fixes this indentation issue.

Any related unit tests has also been modified to allow proper alignment.

However, I'm not sure whether this will have any other side effect, if so, please point it out. Thanks!

PR #4280, opened 2019-01-27
MethodArgumentSpaceFixer - Fix method argument alignment

### LINKED ISSUE #3712: method_argument_space with multiline arguments
=> 7.2.4

=> 2.11.1

```php
<?php
function a($a, $b,
           $c, $d) {
}
```
```shell
php-cs-fixer fix sandbox.php --rules=@PSR2 --dry-run --diff
```
```diff
--- Original
+++ New
@@ @@
 <?php

-function a($a, $b,
-           $c, $d) {
+function a(
+    $a,
+    $b,
+           $c,
+    $d
+) {
 }
```

Extra spaces before `$c` could be removed. Perhaps this isn't `MethodArgumentSpaceFixer`'s responsibility, or it could have a new option `ensure_aligned` if `ensure_fully_multiline` is `true`.

Here's a test case with expected output:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/compare/master...nobuf:MethodArgumentSpaceFixer-with-multiline-args

Billz95:
Fixes #3712 
Original PR #4277 

New PR is made in order to target long term support version (2.12),
**For more discussion please see #4277** 

Below is a duplicated description of #4277 for ease of reading:

As mentioned in the above issue, the code will not be aligned properly when forced into multiline (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3712#issue-316138499)
```php
<?php
function a($a, $b,
           $c, $d) {
}
```
would be fixed to
```diff
--- Original
+++ New
@@ @@
 <?php

-function a($a, $b,
-           $c, $d) {
+function a(
+    $a,
+    $b,
+           $c,
+    $d
+) {
 }
```
However, the expected result is 
```diff
--- Original
+++ New
@@ @@
 <?php

-function a($a, $b,
-           $c, $d) {
+function a(
+    $a,
+    $b,
+    $c,
+    $d
+) {
 }
```
This is caused by function `MethodArgumentSpaceFixer::fixNewline`, where it simply returns without applying any fix when the token after `,` is newline.

This is not ideal, from the above code snippet, we can see the token after `$b,` is `newline + many spaces`. Instead, what we expect should be `newline + indentation`.

Therefore, by simply removing `$this->isNewline($token[$index + 1])` from early return check, it fixes this indentation issue.

Any related unit tests has also been modified to allow proper alignment.

However, I'm not sure whether this will have any other side effect, if so, please point it out. Thanks!

### DIFF
--- a/src/Console/Command/ReadmeCommand.php
+++ b/src/Console/Command/ReadmeCommand.php
@@ -271,7 +271,7 @@ EOF;
         //      `description <http://...>`_
 
         $help = Preg::replaceCallback(
-           '#`(.+)`\s?\(<url>(.+)<\/url>\)#',
+            '#`(.+)`\s?\(<url>(.+)<\/url>\)#',
             static function (array $matches) {
                 return sprintf('`%s <%s>`_', str_replace('\\', '\\\\', $matches[1]), $matches[2]);
             },
--- a/src/Fixer/Alias/BacktickToShellExecFixer.php
+++ b/src/Fixer/Alias/BacktickToShellExecFixer.php
@@ -41,7 +41,7 @@ final class BacktickToShellExecFixer extends AbstractFixer
             'Converts backtick operators to `shell_exec` calls.',
             [
                 new CodeSample(
-<<<'EOT'
+                    <<<'EOT'
 <?php
 $plain = `ls -lah`;
 $withVar = `ls -lah $var1 ${var2} {$var3} {$var4[0]} {$var5->call()}`;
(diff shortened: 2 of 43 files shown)


Billz95:
For commits that only contains the relevant fix, please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4280/commits/525ce169b60111bae01efa724476c42c6f5f6355

keradus:
Thank you @Billz95.

Issue #3601, opened 2018-03-14
Feature Request: Line Number with Checkstyle Format


ChrGriffin:
I'd love if this tool could be used to flag errors for developers to fix, rather than fixing them automatically (in the interest of teaching junior devs to format code properly the first time, rather than rely on an automated tool to catch it for them). We can _sort of_ do that now by executing `php-cs-fixer fix ./ --dry-run --diff`, but the given output is very large and doesn't actually identify what the error _was_, just the change that was made. In a perfect world I'd be able to write `php-cs-fixer fix ./app --dry-run --diff > output.txt` and see a much more truncated list with the filename, line number, and rule that is being violated.

I'm aware of one other such tool, CodeSniffer, that would offer me (sort of) this functionality, but in all honesty I far prefer this package's configuration method.

EDIT: on further review, the `--format=checkstyle` option _almost_ gives me what I want -- my only issue is that I can't get the line number with each item.

keradus:
when the formats were introduced one after another (last one - checkstyle indeed), there were no good differ available. Thanks to amazing job of SpacePossum, we can now generate diff in udiff format.

We are already prepare that checkstyle format could contain linenumber (vide https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/doc/checkstyle.xsd#L30), we were thinking about it even without having good differ.

Would you like to update https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.10/src/Report/CheckstyleReporter.php to expose line ? :)

ChrGriffin:
Sure thing, I'll take a crack at it.

SpacePossum:
Currently the check style report contains entries for each fixer, this makes the line number not useful in some situations. For example one fixer may change a series of tokens on line 2 and than the next fixer moves the tokens to line 20 or removes all tokens leaving no "line 2" in the file. Not sure how to proceed here in a meaningful matter.

nesl247:
I think at least when --dry-run is used, the line number the error is on is all that matters since fixers are not run. When one issue is fixed, the line numbers the errors were on will obviously change.

SpacePossum:
If you run with just one rule, than yeah the line numbers will always map back to that one rule and you should be good :)

If you run multiple rules (fixers) you still run into the same issue, in dry-run or not, as the fixed result from the first rule is the input of the next, of which the fixed/changed output is the input of the 3rd etc.
[...]

PR #4288, opened 2019-01-31
Add Gitlab Reporter

hco:
Gitlab does support a subset of the CodeClimate JSON Format.
I initially called the Reporter "CodeClimate", but that would "require" more fields that we currently cannot fill appropriately from a report generator, I think.
So I decided to rename it to "Gitlab Reporter".

It will result in a pull request annotation like this:
image

I will be happy to maintain this reporter, if problems arise.

In a next step I would be happy to add meaningful line numbers, if someone can assist me on that, but after reading https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3601 it does not seem to be so easy, and I think the current state is enough for an "MVP" ;-)

### DIFF
--- a/src/Report/GitlabReporter.php
+++ b/src/Report/GitlabReporter.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Report;
+
+use Symfony\Component\Console\Formatter\OutputFormatter;
+
+/**
+ * Generates a report according to gitlabs subset of codeclimate json files.
+ *
+ * @author Hans-Christian Otto <c.otto@suora.com>
+ *
+ * @internal
+ */
+final class GitlabReporter implements ReporterInterface
+{
+    public function getFormat()
+    {
+        return 'gitlab';
+    }
+
+    /**
+     * Process changed files array. Returns generated report.
+     *
+     * @param ReportSummary $reportSummary
+     *
+     * @return string
+     */
+    public function generate(ReportSummary $reportSummary)
+    {
+        $report = [];
+        foreach ($reportSummary->getChanged() as $fileName => $change) {
+            foreach ($change['appliedFixers'] as $fixerName) {
+                $report[] = [
+                    'description' => $fixerName,
+                    'location.path' => $fileName,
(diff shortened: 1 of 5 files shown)


keradus:
Hi @hco , thanks for your PR, idea looks interesting indeed.
Before going into details of the code, we need to ensure that format is valid.
please upload schema of this format to `/doc`

hco:
@keradus I think Gitlab does not provide a schema :-(

Or would a self-written schema help?

hco:
Oh - and thanks for the fast feedback!

keradus:
as a first step i would suggest you asking gitlab for schema, maybe it's there, but they forgot to publish it, or simply it's published but we can't find it? it happened before ;)

I'm totally up for having a gitlab support, just let us ensure it works the best and it's stable, meaning we have the schema first.

keradus:
@hco , any update on this ? had you manage to find/ask for gitlab official schema ?

hco:
I asked, but got no response yet. :(

hco:
So the current state is that they are pointing me to code climate, which would require us to generate more fields, which are not necessary for supporting gitlab.
In addition, code climate does not currently provide an official schema. Asked them as well.

Anything I can do while we wait?

hco:
No response from Code Climate yet. 

I currently think that the only think I could do is write my own schema for GitLab-CI.
Do you think that this would give us more benefit than the existing tests? 🤔

Do you have other suggestions on what I should do? 

The Travis build failure is nothing I need to work on, right?

src/Report/GitlabReporter.php
```diff
@@ -0,0 +1,52 @@
...
+                $report[] = [
+                    'description' => $fixerName,
+                    'location.path' => $fileName,
+                    'fingerprint' => md5($fileName.$fixerName),
+                    'location.lines.begin' => 0, // line numbers are required in the format, but not available to reports
+                ];
+            }
+        }
+
+        return json_encode($report);
```
SpacePossum: I this should be something like:
```php
$encoded = json_encode($report);

return $reportSummary->isDecoratedOutput() ? OutputFormatter::escape(encoded) : encoded;
```

hco:
Done. Thanks for the feedback!

The change did not involve any test, though. I think this is outside of the scope of this PR, but maybe there should be a testcase added for this to the `AbstractReporterTestCase`, right?

If you agree, I can tackle that later aswell.

SpacePossum:
Thanks for the changes, this PR looks good to me :)

If you've some time for a test for the `AbstractReporterTestCase` please give it a shot. I remember hunting down the not-escaped bug on the previous reporters, so if we can catch it in the future by a test it would great. For me it is not blocking for this PR though.

SpacePossum:
Thank you @hco.

PR #4292, opened 2019-02-02
NoSuperfluousPhpdocTagsFixer - Make null only type not considered superfluous

### LINKED ISSUE #4269: NoSuperfluousPhpdocTagsFixer removes "@return null" tags
The `no_superfluous_phpdoc_tags` removes `@param`/`@return` tags with only `null` as type. Those tags shouldn't be removed: they can't be superfluous because `null` can't be used as a type declaration.

```diff
 /**
- * @param null $foo
- * @return null
  */
 function foo($foo)
 {
     return null;
 }
```

julienfalque:
Fixes #4269.


keradus:
Thank you @julienfalque.

PR #4293, opened 2019-02-03
Add official support for PHP 7.3

keradus:
Backports #4233
7.3 support proved to be stable, we can use it for LTS


winkbrace:
Thanks! <3

PR #4303, opened 2019-02-11
FullyQualifiedStrictTypesFixer - Fix the short type detection when a question mark (nullable) is prefixing it.

### LINKED ISSUE #4302: Fix FullyQualifiedStrictTypesFixer when return types starts with ?
For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

=>7.1.26

=> 2.14

When using the directive `fully_qualified_strict_types`

```php
<?php

declare(strict_types = 1);

namespace Bar\Boo;

class Foo {
  public function random(): ?/Bar/Foo/Int {
    // Returns an instance of /Bar/Foo/Int or Null.
  }
}
```

```php
// Nothing changes.
```

```php
<?php

declare(strict_types = 1);

namespace Bar\Boo;

use Bar\Foo\Int;

class Foo {
  public function random(): ?Int {
    // Returns an instance of /Bar/Foo/Int or Null.
  }
}
```

drupol:
Fix #4302

### DIFF
--- a/src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
+++ b/src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
@@ -60,6 +60,11 @@ final class TypeAnalysis implements StartEndTokenAwareAnalysis
     private $endIndex;
 
     /**
+     * @var bool
+     */
+    private $nullable;
+
+    /**
      * @param string $name
      * @param int    $startIndex
      * @param int    $endIndex
@@ -67,6 +72,13 @@ final class TypeAnalysis implements StartEndTokenAwareAnalysis
     public function __construct($name, $startIndex, $endIndex)
     {
         $this->name = $name;
+        $this->nullable = false;
+
+        if (0 === strpos($name, '?')) {
+            $this->name = substr($name, 1);
+            $this->nullable = true;
+        }
+
         $this->startIndex = $startIndex;
         $this->endIndex = $endIndex;
     }
@@ -101,5 +113,13 @@ final class TypeAnalysis implements StartEndTokenAwareAnalysis
     public function isReservedType()
     {
         return \in_array($this->name, self::$reservedTypes, true);
+    }
+
+    /**
+     * @return bool
+     */
+    public function isNullable()
+    {
+        return $this->nullable;
     }
 }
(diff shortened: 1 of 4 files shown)


src/Tokenizer/Analyzer/Analysis/TypeAnalysis.php
```diff
@@ -76,7 +76,9 @@ public function __construct($name, $startIndex, $endIndex)
      */
     public function getName()
     {
-        return $this->name;
+        return $this->isNullable() ?
```
keradus: do it once, in constructor

drupol: Fixed.

drupol:
Here's the patch for branch 2.14: https://gist.github.com/drupol/c43a4ba8712782e104194fdc015c8ede
We need it for our project.

keradus:
thanks for the patch @drupol .
yet, to not waste your time on it, if you send a PR against older branch, after PR would be merged, that branch would also be merged into higher branch - and usually git does that without conflicts, and for small conflicts I resolve them myself.
it's very rare that for same change we ask original author to prepare independent patches for different branches - it's only in case of super complex change - in other words, don't prepare the "other branch patches" except if you was explicitly asked for ;)

drupol:
Allrigthty!

Thanks.

tests/Fixer/Import/FullyQualifiedStrictTypesFixerTest.php
```diff
@@ -36,6 +36,19 @@ public function testCodeWithReturnTypes($expected, $input = null)
         $this->doTest($expected, $input);
     }
 
+    /**
+     * @requires PHP 7.1
+     *
+     * @dataProvider provideCodeWithReturnTypesCasesWithNullableCases
+     *
+     * @param mixed      $expected
```
kubawerlos: ~~Can we have `string` here, and in the line below, instead of `mixed`?~~
Actually, can we have `string` instead of `mixed` in the whole file?

drupol: I think yes.

SpacePossum: done

SpacePossum:
Thank you @drupol.

drupol:
Thanks ! Very needed here :)

PR #4311, opened 2019-02-13
WIP: TernaryOperatorMultilineFixer : Standardize the ternary operators written on multiple lines

Gwalchmei:
I did not found any consensus on how to write ternary operators on multiple lines but I feel it should be consistent across a given project. That's why I decided to write this Fixer with 3 strategies :
- Force the operators to a single line
- Put the operators at the end of lines
- Put the operators at the beginning of lines

This PR is not finished because it does not handle the nested operators well (see the failing case in the test). However I open it because I did not find a clever way to fix them and hoped someone could help find a less complicated solution than listing all the nested cases.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1753,6 +1753,19 @@ Choose from the list of available rules:
 
   Removes extra spaces between colon and case value.
 
+* **ternary_operator_multiline**
+
+  Standardize multi-lines ternary operators.
+
+  Configuration options:
+
+  - ``ignore-single-line`` (``bool``): should ignore the single line operators;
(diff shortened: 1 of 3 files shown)


julienfalque:
Hi @Gwalchmei, sorry for the late reply. I think most of this is already handled by `operator_linebreak` rule. Shall we close this PR?

Wirone:
Closing this because:
- It's WIP for more than 4 years
- ternaries (also short form) are handled by `operator_linebreak`, as pointed out by @julienfalque 
- no response from the PR's author since PR was created

Issue #4312, opened 2019-02-14
PhpUnitTestCaseStaticMethodCallsFixer - produces syntax error


ro0NL:
#### The PHP version you are using (`$ php -v`):
7.2.15 

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
2.14.1

#### The command you use to run PHP CS Fixer:
php-cs-fixer fix --dry-run src/ -vvv

#### The configuration file you are using, if any:
https://github.com/msgphp/msgphp/blob/master/.php_cs.dist

```
Files that were not fixed due to errors reported during linting after fixing:
   1) /app/src/Eav/Tests/Entity/Fields/AttributeFieldTest.php

                                                                   
        [PhpCsFixer\Linter\LintingException]                       
        PHP Parse error: syntax error, unexpected '=' on line 38.  
                                                                   

      Applied fixers: php_unit_test_case_static_method_calls
```

See https://travis-ci.org/msgphp/msgphp/jobs/492145928#L198 for error during build.

I thought it was related to #4285, and as such fixed by #4286. But it seems this is not the case :(

SpacePossum:
Hi and thanks for reporting.

After checking I think the issues are `PhpUnitTestCaseStaticMethodCallsFixer`.
I see the fixer:
- touching not-method class (`$this->attribute = 1;`)
- touching method calls within a anonymous class inside a PHHUnit class

Seen on both 2.12 and 2.14

kubawerlos:
@SpacePossum take a look at PR mentioned above ;)

SpacePossum:
: D
I had this issue open for so long, I never saw the PR reference since I didn't reload the page :man_facepalming:

PR #4314, opened 2019-02-15
PhpUnitTestCaseStaticMethodCallsFixer - fix for having property with name as method to update

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4312

Pinging @Slamdunk (as original author) and @ro0NL (as reporter) for review.

### DIFF
--- a/tests/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixerTest.php
+++ b/tests/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixerTest.php
@@ -433,6 +433,19 @@ EOF
                     'call_type' => PhpUnitTestCaseStaticMethodCallsFixer::CALL_TYPE_THIS,
                 ],
             ],
+            [
+                <<<'EOF'
+<?php
+class HavingPropertyWithNameAsMethodToUpdateTest extends PHPUnit
+{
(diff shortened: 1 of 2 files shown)


ro0NL:
seems to work for me :+1:

tests/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixerTest.php
```diff
@@ -433,6 +433,19 @@ public function baz()
...
                 ],
             ],
+            [
+                <<<'EOF'
+<?php
+class HavingPropertyWithNameAsMethodToUpdateTest extends PHPUnit
+{
+    public function foo()
+    {
+        $this->attribute = 42;
```
Slamdunk: Test and fix is ok, but I think it isn't 100% clear enough what you are testing, for someone who look at this the first time. The class name of course helps, but I would use a more wide spread `$this->assertSame = 42` for example.

SpacePossum: :+1:  we can have both I think :)

SpacePossum:
Thank you @kubawerlos.

Issue #4241, opened 2019-01-05
NoUnsetCastFixer - breaks higher-precedence operators and some side effects


guilliamxavier:
#### The PHP version you are using (`$ php -v`):
=> 7.2.10

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.14.0

#### The command you use to run PHP CS Fixer:
=> `php-cs-fixer fix -vvv --using-cache=no --rules=no_unset_cast sample.php`

#### The configuration file you are using, if any:
(none)

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):

Given (don't mind the compact style):

```php
<?php

function f($x) {
    var_dump($x); // note: side effect
    return $x;
}
class C {
    public static function sf($x) { return f($x); }
    public function mf($x) { return f($x); }
}
class A extends ArrayObject {
    public function offsetGet($x) { return parent::offsetGet(f($x)); }
}

$f = 'f';
$c = new C();
$a = new A([42]);
$n = 3;
$i = 0;
```

then the following are correctly preserved:

| before | after |
| --- | --- |
| `$r = (unset) f(1);` | (no change) |
| `$r = (unset) C::sf(2);` | (no change) |
| `$r = (unset) (new C())->mf(3);` | (no change) |

but the following are not:

| before | after | problem |
| --- | --- | --- |
| `$r = (unset) $f(1);` | `$r =  null(1);` | Fatal error |
| `$r = (unset) $c::sf(2);` | `$r =  null::sf(2);` | Fatal error |
| `$r = (unset) $c->mf(3);` | `$r =  null->mf(3);` | [not applied because of] Parse error |
| `$r = (unset) $a[0];` | `$r =  null[0];` | lost side effect (`$a->offsetGet(0)`) |
| `$r = (unset) $n**f($n);` | `$r =  null**f($n);` | `$r` = `0` instead of `null` |
| `$r = (unset) $i++;` | `$r =  null++;` | [not applied because of] Parse error |

SpacePossum:
@guilliamxavier thanks for reporting, good find!
Do you plan on proposing a fix through a PR? I might give it a shot but I'm also kind of busy so please let me know :)

SpacePossum:
thanks for the details @guilliamxavier :)

I give it a try soon, checking if after the variable is either `;` or `?>` seems the easiest to do. It will still cover the most common cases.

```
[...]

PR #4316, opened 2019-02-17
NoUnsetCastFixer - Test for higher-precedence operators

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4241

### DIFF
--- a/src/Fixer/CastNotation/NoUnsetCastFixer.php
+++ b/src/Fixer/CastNotation/NoUnsetCastFixer.php
@@ -70,6 +70,11 @@ final class NoUnsetCastFixer extends AbstractFixer
             return;
         }
 
+        $afterVar = $tokens->getNextMeaningfulToken($varIndex);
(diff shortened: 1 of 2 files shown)


src/Fixer/CastNotation/NoUnsetCastFixer.php
```diff
@@ -70,6 +70,11 @@ private function fixUnsetCast(Tokens $tokens, $index)
             return;
         }
 
+        $afterVar = $tokens->getNextMeaningfulToken($varIndex);
+        if (null === $varIndex || !$tokens[$afterVar]->equalsAny([';', [T_CLOSE_TAG]])) {
```
guilliamxavier: Should be `null === $afterVar` probably?

SpacePossum: oh yes, copy-paste eror :man_facepalming:

PR #4320, opened 2019-02-22
Add support for outputting to stdout

### LINKED ISSUE #2011: Output fixed php to stdout
Google foo and a bit of trial and error has gotten me so far:

`cat test.php | php-cs-fixer fix -`
Allows stdin! But this doesn't output the fixed php.

`cat test.php | php-cs-fixer fix --diff -`
Outputs to stdout, but only a diff.

It would be useful to be able to pipe in php, and have the actual fixed up php returned to stdout.

My particular use case is vim, it's awkward to be forced to save a file, before having vim run php-cs-fixer and then have vim auto-reload the modified file.

Much better would be to pipe the contents of the modified buffer into php-cs-fixer and replace it with the returned stdout. Perhaps a --output option?
This would add a lot of flexibility to php-cs-fixer.

`:%!php-cs-fixer fix -o - -`

garygreen:
Fixes #2011

This PR adds basic support for outputting the result to stdout with the `--stdout` command line parameter.

Only one file is supported, though if you did pass multiple files it would just stream them all to stdout. In reality, if your wanting stdout then your probably not caring much for multiple files. But in any case, better support for that could be added in the future.

## Use case for this PR

To improve formatter extensions/plugins for e.g. Visual Studio Code, Atom, Sublime, etc.

If for example you had a `formatOnSave` option configured, the workflow in editors at the moment is this:

1. Document Saved (to disk)
2. Exec php-cs-fixer (takes 2+ secs)
3. Editor polls/detects change to file on disk
4. Reload document

If at any point during this the user has changed the file, the editor will essentially corrupt it and show an error message, it'll then force you to re-open the file, discarding any changes you made. This interrupts your workflow and isn't a good experience. There is currently NO way around this because php-cs-fixer doesn't even allow outputting to a different file, as a workaround.

With this PR, the workflow becomes much improved:

1. Intercept document about to be saved (e.g. in vscode, `onWillSaveTextDocument`)
2. Std-in the code to php-cs-fixer (response 2+ secs)
3. Std-out from cs fixer
4. Update contents of document
5. Save to disk

This avoids multiple disk saves, it also allows you to cancel/discard the output from php-cs-fixer if the document has subsequently been changed.

### DIFF
--- a/tests/Report/NullReporterTest.php
+++ b/tests/Report/NullReporterTest.php
@@ -1,0 +1,65 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Report;
+
+use PhpCsFixer\Report\NullReporter;
+
+/**
+ * @author Boris Gorbylev <ekho@ekho.name>
+ * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Report\NullReporter
+ */
+final class NullReporterTest extends AbstractReporterTestCase
+{
+    public function createSimpleReport()
+    {
+    }
+
+    public function createWithDiffReport()
+    {
+    }
+
+    public function createWithAppliedFixersReport()
+    {
+    }
+
+    public function createWithTimeAndMemoryReport()
+    {
+    }
+
+    public function createComplexReport()
+    {
+    }
+
+    protected function createReporter()
+    {
+        return new NullReporter();
+    }
+
+    protected function getFormat()
+    {
+        return 'null';
+    }
+
+    protected function createNoErrorReport()
+    {
+    }
+
+    protected function assertFormat($expected, $input)
+    {
+        $this->markTestSkipped('NullReporterTest skipped');
+    }
+}
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -111,6 +111,7 @@ final class FixCommand extends Command
                     new InputOption('format', '', InputOption::VALUE_REQUIRED, 'To output results in other formats.'),
                     new InputOption('stop-on-violation', '', InputOption::VALUE_NONE, 'Stop execution on first violation.'),
                     new InputOption('show-progress', '', InputOption::VALUE_REQUIRED, 'Type of progress indicator (none, run-in, estimating, estimating-max or dots).'),
+                    new InputOption('stdout', '', InputOption::VALUE_NONE, 'Send output to stdout'),
                 ]
             )
             ->setDescription('Fixes a directory or a file.')
@@ -144,17 +145,21 @@ final class FixCommand extends Command
                 'stop-on-violation' => $input->getOption('stop-on-violation'),
                 'verbosity' => $verbosity,
                 'show-progress' => $input->getOption('show-progress'),
+                'stdout' => $input->getOption('stdout'),
             ],
             getcwd(),
             $this->toolInfo
         );
 
+        // Supress any text-based errors when target is stdout.
         $reporter = $resolver->getReporter();
-
-        $stdErr = $output instanceof ConsoleOutputInterface
-            ? $output->getErrorOutput()
-            : ('txt' === $reporter->getFormat() ? $output : null)
-        ;
+        if ($resolver->isStdOut()) {
+            $stdErr = null;
+        } else {
+            $stdErr = $output instanceof ConsoleOutputInterface
+                ? $output->getErrorOutput()
+                : ('txt' === $reporter->getFormat() ? $output : null);
+        }
 
         if (null !== $stdErr) {
             if (null !== $passedConfig && null !== $passedRules) {
@@ -213,7 +218,8 @@ final class FixCommand extends Command
             $resolver->isDryRun(),
             $resolver->getCacheManager(),
             $resolver->getDirectory(),
-            $resolver->shouldStopOnViolation()
+            $resolver->shouldStopOnViolation(),
+            $resolver->isStdOut()
         );
 
         $this->stopwatch->start('fixFiles');
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -130,6 +130,7 @@ final class ConfigurationResolver
         'stop-on-violation' => null,
         'using-cache' => null,
         'verbosity' => null,
+        'stdout' => null,
     ];
 
     private $cacheFile;
@@ -555,6 +556,14 @@ final class ConfigurationResolver
     }
 
     /**
+     * @return bool
+     */
+    public function isStdOut()
+    {
+        return true === $this->options['stdout'];
+    }
+
+    /**
      * Compute file candidates for config file.
      *
      * @return string[]
@@ -617,6 +626,10 @@ final class ConfigurationResolver
      */
     private function getFormat()
     {
+        if ($this->isStdOut()) {
+            return 'null';
+        }
+
         if (null === $this->format) {
             $this->format = null === $this->options['format']
                 ? $this->getConfig()->getFormat()
--- a/src/Report/NullReporter.php
+++ b/src/Report/NullReporter.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Report;
+
+/**
+ * @author Boris Gorbylev <ekho@ekho.name>
+ *
+ * @internal
+ */
+final class NullReporter implements ReporterInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getFormat()
+    {
+        return 'null';
+    }
+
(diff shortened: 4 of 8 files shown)


tests/Report/NullReporterTest.php
```diff
@@ -0,0 +1,65 @@
...
+        return 'null';
+    }
+
+    protected function createNoErrorReport()
+    {
+    }
+
+    protected function assertFormat($expected, $input)
+    {
+        $this->markTestSkipped('NullReporterTest skipped');
```
keradus: no

keradus:
I don't see allowing to configure any run to produce output on stdout as good option, especially as you said, for multuple-files run it would not be good.

when input comes from drive, we store it back on it.
i would rather allow to produce result on stdout when input was from stdin.

and yes, we already support stdin
```
ker@dus:~/github/PHP-CS-Fixer λ cat aaa.php | php php-cs-fixer fix - --dry-run --diff
Loaded config default from "~/github/PHP-CS-Fixer/.php_cs.dist".
   1) php://stdin
--- Original
+++ New
@@ @@
-<?php echo 123;
+<?php
+
+
+echo 123;
 

Checked all files in 0.003 seconds, 12.000 MB memory used
```

garygreen:
> i would rather allow to produce result on stdout when input was from stdin.

How would that work? Do you mean not have a public option `--stdout` and instead just internally FORCE outputting to stdout whenever the input is stdin? That doesn't seem like a good alternative to this PR, imo.

Also - currently stdin only supports ONE file, so it would be fair to assume that stdout will only support one file as well. This should be documented.

garygreen:
## Use case for this PR

To improve formatter extensions/plugins for e.g. Visual Studio Code, Atom, Sublime, etc.

If for example you had a `formatOnSave` option configured, the workflow in editors at the moment is this:

1. Document Saved (to disk)
2. Exec php-cs-fixer (takes 2+ secs)
3. Editor polls/detects change to file on disk
4. Reload document

If at any point during this the user has changed the file, the editor will essentially corrupt it and show an error message, it'll then force you to re-open the file, discarding any changes you made. This interrupts your workflow and isn't a good experience. There is currently NO way around this because php-cs-fixer doesn't even allow outputting to a different file, as a workaround.

With this PR, the workflow becomes much improved:

1. Intercept document about to be saved (e.g. in vscode, `onWillSaveTextDocument`)
2. Std-in the code to php-cs-fixer (response 2+ secs)
3. Std-out from cs fixer
4. Update contents of document
5. Allow editor to save

This avoids multiple disk saves, it also allows you to cancel/discard the output from php-cs-fixer if the document has subsequently been changed.

garygreen:
In reference to multiple files - this isn't something I am particularly interested in adding support for, as I'm primarily focused on improving the use case as described above.

To support multiple files in stdout - an implementation detail for something like JSON-lines would need to be considered. This isn't something easily done without numerous design considerations and complexities. But in any case, I would imagine the design would look similar to how ripgrep has recently introduced JSON formatter, see this issue for inspiration/implementation details: https://github.com/BurntSushi/ripgrep/issues/244#issuecomment-394677435

tests/Report/NullReporterTest.php
garygreen: I did this because I wasn't entirely sure what to test, to be honest. Any other constructive advice would be appreciated.

FlickerBean:
@keradus why do you not see an output to stdout switch as a good option? This is pretty standard Unix behaviour, and valid use cases have been pointed out.

Even use cases such as piping the output of php-cs-fixer to another command are made impossible without this feature.

For a command line tool who's primary function is essentially to manipulate text, not being able to output to stdout seems bizzare.

garygreen:
It's worth noting that nearly all unix utilities/tools have a limitation of being only able to work on one file as stdin -> stdout (with some exceptions). However, piping is still an extremely powerful and useful tool in all utilities to avoid unnecessary intermediary disk saves and to have tighter integration with other utilities.

tests/Report/NullReporterTest.php
keradus: then I would expect `TODO` here to be honest or sth. now, you gave us impression that "test is done"

keradus: let see together how to do it after redefining whole idea of how to do this PR - maybe the reporter will be gone, maybe not...

keradus:
@FlickerBean , @garygreen , I have nothing against producing result on stdout instead of file and i do see the value of piping etc. my comments were not about "we not gonna merge it ever" or "we don't like this feature", yet "proposed solution doesn't look optimal". Sorry for confusion.

having an option switch to produce to stdout (let me ignore all hacky `if`s for a while for it) doesn't soung good for the reason that @garygreen already mentioned - it's possible to use it for a fixing of whole dir, having multiple files outputted to stdout. this gonna be mess.

instead of allowing to use it always, let's rely on the fact that we can already fix piped content from stdin.
IMO, let's do sth like that. what do you think about it?
- don't touch stderr. any content there is good content we want to still have. it's not stdout
- create new differ, that produce not real diff but whole new file only
- on stdout, we print only the proper output that can be grabbed by other tool, let's have sth like `RawFileOutputReporter` that gives only raw output of (new) differ
- allow to use `RawFileOutputReporter` only when tool was called with stdin, it can even assume to use new differ if it was not configured differently

remember to not break BC contract that is already in place

garygreen:
> * don't touch stderr. any content there is good content we want to still have. it's not stdout

Would the contents of that error message form part of the output? I would say that only if it's obvious that php-cs-fixer had errors and output as proper exit status code, otherwise you would be fooled into thinking the fixed file contents IS the error message, unless you had a proper way of distinguishing it in the piped stream.

> * allow to use `RawFileOutputReporter` only when tool was called with stdin, it can even assume to use new differ if it was not configured differently

I still think there's value in allowing stdout even when your not using stdin. It's a useful testing tool for one.

The proposal for using differ/reporter correctly is probably better than the "hacky" way (as you put it) that I've implemented it in this PR. It seems to me you have a clearer idea for implementation details, so I'm inclined to leave it up to you to finish/cleanup this PR, if your happy to do that? 😃

keradus:
extra point - what will happen if file marked for fixing:
- cache was used and it says to not do this file
- has invalid syntax, it would be skipped by tool due to that
- has valid syntax, but PHP CS Fixer failed to change it (code after fixing doesn't pass linting, PHP CS Fixer does not write those fixes to drive or show on diff)

garygreen:
> * cache was used and it says to not do this file

This PR doesn't change anything to do with the cache. So I would say it works exactly the way it does now, it just changes the output from file to stdout, that is all. Cache will still/won't be used depending on options configured, exactly as before.

>* has invalid syntax, it would be skipped by tool due to that
> * has valid syntax, but PHP CS Fixer failed to change it (code after fixing doesn't pass linting, PHP CS Fixer does not write those fixes to drive or show on diff)

These two points really relate to setting a proper "failed" exit status code. I'm not entirely sure if it's possible to configure an exit code with a custom message, do you happen to know more on that?

keradus:
> Would the contents of that error message form part of the output? I would say that only if it's obvious that php-cs-fixer had errors and output as proper exit status code, otherwise you would be fooled into thinking the fixed file contents IS the error message, unless you had a proper way of distinguishing it in the piped stream.

sorry, i totally not get this paragraph.
- stdout is used by tools to produce outcome that has BC policy on it and that can be used as input to other tool
- stderr is used by tools to inform about errors, warnings, progress, any kind of meta info that shall not be processed by other tool
- yes, on OS level you say which stream you are grabbing, by default you grab only stdout, not stderr

> I still think there's value in allowing stdout even when your not using stdin. It's a useful testing tool for one.

i'm not saying there is no value of that for you. but sth that has value only for you may not end up as official part of sourcecode

> so I'm inclined to leave it up to you to finish/cleanup this PR, if your happy to do that?

I won't do it for sure, while I see value of better supporting of stdout, I don't need it personally, and it's not a top-prio for this tool as whole. If you don't wish to continue on this PR, simply close it.

> This PR doesn't change anything to do with the cache. So I would say it works exactly the way it does now, it just changes the output from file to stdout, that is all. Cache will still/won't be used depending on options configured, exactly as before.

so, one expected to have "fixed" output, yet due to caching he got empty outcome. how can he say if input file was empty or outcome was not produce due to cache. it can be problematic.


> These two points really relate to setting a proper "failed" exit status code. I'm not entirely sure if it's possible to configure an exit code with a custom message, do you happen to know more on that?

we already have exit statuses for those cases: https://github.com/FriendsOfPHP/PHP-CS-Fixer#exit-codes
my question is same as for cache - what would be the outcome for end-user/tooling ?
like if someone plug PHP CS Fixer for his tooling/IDE/whatever, and then due to caching or any error that tooling pipeline overwritten his whole files to empty ones, it could be an issue

garygreen:
> * yes, on OS level you say which stream you are grabbing, by default you grab only stdout, not stderr

In which case, shall I change it to still write to `stdErr`? As any messages/errors sent to that stream won't form part of the default one on stdout?

> so, one expected to have "fixed" output, yet due to caching he got empty outcome. how can he say if input file was empty or outcome was not produce due to cache. it can be problematic.

Interesting points about caching, hadn't really considered it. I personally would expect it to always return the formatted code, even if it was unchanged. So maybe when it detects that it's not changed/in cache, it will just return the original source code, that way you can carry on piping it if needed.

keradus:
> In which case, shall I change it to still write to `stdErr`? As any messages/errors sent to that stream won't form part of the default one on stdout?

i said that only contracted output of reported is written to stdOut, all the rest to stdErr. don't send it to stdErr. for that, I dont get the "still write to stdErr" part.

garygreen:
Ohhh look, Google's Brotli CLI tool does `--stdout` even though it supports multiple files.

image

😸

garygreen:
@keradus From various comments I'm a bit confused about what needs to be done exactly. Could you explicitly list what you would like changed in terms of implementation before your ok to merge this in?

Epskampie:
To me, @keradus  suggestion of only outputting to `stdout` when `stdin` is used sounds very reasonable and logical.

In the meantime I just use [this small script](https://gist.github.com/Epskampie/427c512722b8e6c4cbf5f1340b87986b) for stdin/stdout but it would be better if it would be in the core.

Issue #4144, opened 2018-12-10
[PHPUnit] Add migration for 7.5 assertInternalType


Slamdunk:
References: https://github.com/sebastianbergmann/phpunit/blob/7.5.0/ChangeLog-7.5.md

> Implemented https://github.com/sebastianbergmann/phpunit/issues/3368: Added assertIsArray(), assertIsBool(), assertIsFloat(), assertIsInt(), assertIsNumeric(), assertIsObject(), assertIsResource(), assertIsString(), assertIsScalar(), assertIsCallable(), assertIsIterable(), assertIsNotArray(), assertIsNotBool(), assertIsNotFloat(), assertIsNotInt(), assertIsNotNumeric(), assertIsNotObject(), assertIsNotResource(), assertIsNotString(), assertIsNotScalar(), assertIsNotCallable(), assertIsNotIterable() as alternatives to assertInternalType() and assertNotInternalType()
> 
> The methods assertInternalType() and assertNotInternalType() are now deprecated. There is no behavioral change in this version of PHPUnit. Using these methods will trigger a deprecation warning in PHPUnit 8 and in PHPUnit 9 these methods will be removed.

TomasVotruba:
AST integration in Rector for inspiration: https://github.com/rectorphp/rector/pull/825/commits/a59e35678ce16d9c3ea9c1f9116eb60553832d94#diff-d98925ff1d61912b3a56a882ffafeae3

One drawbacks is that `boolean`/`integer` are internaly converted to `bool`/`int`.

```diff
-$this->assertNotInternalType('bool', $value, 'message');
+$this->assertIsNotBool($value, 'message');

-$this->assertNotInternalType('boolean', $value, 'message');
+$this->assertIsNotBool($value, 'message');
```

TomasVotruba:
@SpacePossum Could you elaborate on :confused: ?

SpacePossum:
The linking to other tooling/project that use AST confuses (:confused:) me because this tool does not use AST, it works with tokens.

TomasVotruba:
It's not about AST, it's about test case that you can copy while writing the fixer - https://github.com/rectorphp/rector/commit/a59e35678ce16d9c3ea9c1f9116eb60553832d94#diff-fb3d7cfa2c5b2dca79374f0551cbaeec

SpacePossum:
I see now, thanks for the help!

TomasVotruba:
[...]

PR #4328, opened 2019-03-04
Add PhpUnitDedicateAssertInternalTypeFixer

Slamdunk:
Closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4144

- [x] Update `RuleSet`
- [x] Add integration test with `PhpUnitDedicateAssertInternalTypeFixer`

I intentionally created a new fixer instead of updating [PhpUnitDedicateAssertInternalTypeFixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.14.2/src/Fixer/PhpUnit/PhpUnitDedicateAssertFixer.php) because the codes need too different logics.

I don't really like the names, they may confuse the users, any suggestion is welcome, but I would like to keep the two fixers in two different classes.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitDedicateAssertInternalTypeFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitDedicateAssertInternalTypeFixer.php
@@ -1,0 +1,200 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\PhpUnit;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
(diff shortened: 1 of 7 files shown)


src/Fixer/PhpUnit/PhpUnitDedicateAssertInternalTypeFixer.php
```diff
@@ -0,0 +1,190 @@
...
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Indicator\PhpUnitTestCaseIndicator;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+use PhpCsFixer\Tokenizer\TokensAnalyzer;
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class PhpUnitDedicateAssertInternalTypeFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
```
Slamdunk: I see `ConfigurationDefinitionFixerInterface` is deprecated, but I don't find the correct alternative to implement the `createConfigurationDefinition` method: help is welcome (It's a lot of time I can't follow PHP-CS-Fixer internal evolutions :cry:)

SpacePossum: I think you are looking you are looking for `PhpCsFixer\Fixer\ConfigurableFixerInterface`, if you extend the `AbstractFixer` and implement this interface you can keep the `createConfigurationDefinition` the same

SpacePossum:
Thank you @Slamdunk.

PR #4335, opened 2019-03-15
Cache - add indent and line ending to cache signature

### LINKED ISSUE #4330: Changing indent doesn't invalidate cache
=> `7.3.2`

=> `2.14.0 Sunrise`

Changing the indent style by changing (or initially setting) the value passed to `setIndent()` doesn't invalidate the cache, thus requiring either bypassing the cache or changing one of the rules to be able to get the indentation change to apply.

dmvdbrugge:
Add indent and line ending to cache signature

Fixes #4330

### DIFF
--- a/tests/Cache/FileHandlerTest.php
+++ b/tests/Cache/FileHandlerTest.php
@@ -15,6 +15,7 @@ namespace PhpCsFixer\Tests\Cache;
 use PhpCsFixer\Cache\Cache;
 use PhpCsFixer\Cache\FileHandler;
 use PhpCsFixer\Cache\Signature;
+use PhpCsFixer\Cache\SignatureInterface;
 use PhpCsFixer\Tests\TestCase;
 
 /**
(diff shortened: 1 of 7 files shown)


SpacePossum:
Thanks for the PR. Looks good to me, all changes are on internal classes/interfaces so no BC breaks :)

@localheinz could you take a look as well as author of the original cache files?

tests/Cache/FileHandlerTest.php
```diff
@@ -229,4 +188,21 @@ private function getFile()
     {
         return __DIR__.'/.php_cs.cache';
     }
+
+    /**
+     * @return SignatureInterface
+     */
+    private function getSignature()
```
localheinz: What do you think about using `create` instead of `get` as prefix, to indicate that we aren't retrieving the same instance, but creating a new instance every time the method is invoked?

dmvdbrugge: Makes sense, so I did it 😄

SpacePossum:
Thank you @dmvdbrugge.

PR #4336, opened 2019-03-15
Collect coverage with PCOV

kubawerlos:
TODO:
- [x] merge and release https://github.com/PHP-CS-Fixer/phpunit-constraint-isidenticalstring/pull/3
- [x] merge and release https://github.com/PHP-CS-Fixer/phpunit-constraint-xmlmatchesxsd/pull/5
- [x] [prepare](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4337) `PhpUnitTestCaseStaticMethodCallsFixerTest` for PHPUnit 8

Optional:
- cache cloning `krakjoe/pcov` (will be faster that PHP 5.6 without it, that's why optional)
- [allow PHPUnit 8](https://github.com/PHPUnitGoodPractices/Traits/pull/41) for `phpunitgoodpractices/traits`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,8 +118,24 @@ jobs:
 
                 # Composer: boost installation
                 - composer global show hirak/prestissimo -q || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
+
+                # Require PHPUnit 8
+                - composer require --dev --no-update phpunit/phpunit:^8
+
+                # Install PCOV
+                - |
+                    git clone --single-branch --branch=v1.0.2 --depth=1 https://github.com/krakjoe/pcov
+                    cd pcov
+                    phpize
+                    ./configure
+                    make clean install
+                    echo "extension=pcov.so" > $HOME/.phpenv/versions/$TRAVIS_PHP_VERSION/etc/conf.d/pcov.ini
+                    cd $TRAVIS_BUILD_DIR
+            before_script:
+                # Make code compatible with PHPUnit 8
+                - PHP_CS_FIXER_FUTURE_MODE=1 php php-cs-fixer fix --rules=void_return -q tests || return 0
             script:
-                - phpdbg -qrr vendor/bin/phpunit --testsuite coverage --exclude-group covers-nothing --coverage-clover build/logs/clover.xml || travis_terminate 1
+                - vendor/bin/phpunit --testsuite coverage --exclude-group covers-nothing --coverage-clover build/logs/clover.xml || travis_terminate 1
                 - php vendor/bin/php-coveralls -v
 
         -
@@ -152,4 +168,3 @@ jobs:
 
     allow_failures:
         - php: 7.4snapshot
-        - env: COLLECT_COVERAGE=1
--- a/composer.json
+++ b/composer.json
@@ -38,10 +38,10 @@
         "mikey179/vfsstream": "^1.6",
         "php-coveralls/php-coveralls": "^2.1",
         "php-cs-fixer/accessible-object": "^1.0",
-        "php-cs-fixer/phpunit-constraint-isidenticalstring": "^1.0.1",
-        "php-cs-fixer/phpunit-constraint-xmlmatchesxsd": "^1.0.1",
+        "php-cs-fixer/phpunit-constraint-isidenticalstring": "^1.1",
+        "php-cs-fixer/phpunit-constraint-xmlmatchesxsd": "^1.1",
         "phpunit/phpunit": "^5.7.27 || ^6.5.8 || ^7.1",
-        "phpunitgoodpractices/traits": "^1.5.1",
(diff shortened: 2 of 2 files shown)


kubawerlos:
Problem: PHPUnit 8 requires function from `TestCase` (such as `setUp`) to have `void` return type, thus installing PHPUnit with current codebase is not possible. But, to have PCOV we need PHPUnit 8.

Either way we need some compromise:
 - we allow PHPUnit 8 in `composer.json` and every job installing it must add `void` return types before running tests
- we don't allow PHPUnit 8 in `composer.json`, but require it only when collecting code coverage

I went with 2nd one as requiring less changes.

@SpacePossum can toy remove the `WIP` label? Let the review begin.

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
```diff
@@ -82,6 +82,7 @@ final class PhpUnitTestCaseStaticMethodCallsFixer extends AbstractFixer implemen
         'assertClassNotHasAttribute' => true,
         'assertClassNotHasStaticAttribute' => true,
         'assertContains' => true,
+        'assertContainsEquals' => true,
```
keradus: other PR, perhaps ?
like, it's nice to see this change in changelog, without dedidacted PR, it would not be visible

keradus:
can you elaborate about switch to pcov, perhaps? some reasoning, comparison, sth ?
probably a link to sth already existing would be enough

src/Fixer/PhpUnit/PhpUnitTestCaseStaticMethodCallsFixer.php
kubawerlos: Ok, I've reopened https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4337

kubawerlos:
Switch to PCOV is for a simple reason - phpdbg was not capable of calculating coverage for our test suite - it was hitting the limit of opened stream resources, explained [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3929#issuecomment-405922620), inspired by [this](https://github.com/sebastianbergmann/php-code-coverage/blob/7.0.3/.travis.yml#L35) with some benchmarks [here](https://kubawerlos.github.io/slides/2019-03-26_Warszawskie_Dni_Informatyki_Wprowadzenie_do_PHPUnit.html#/34).

.travis.yml
```diff
@@ -118,8 +118,24 @@ jobs:
 
                 # Composer: boost installation
                 - composer global show hirak/prestissimo -q || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
+
+                # Require PHPUnit 8
+                - composer require --dev --no-update phpunit/phpunit:^8
+
+                # Install PCOV
+                - |
+                    git clone --branch=release --depth=1 https://github.com/krakjoe/pcov
```
keradus: `--single-branch` ?
maybe pinpoint to tag instead of branch, that will change one day ?

.travis.yml
```diff
@@ -118,8 +118,24 @@ jobs:
...
+                    cd pcov
+                    phpize
```
keradus: just wondering can we not build it each time, but maybe have prebuilt package we can use locally as well?

.travis.yml
keradus: `git clone --single-branch --branch=v1.0.1 --depth=1 https://github.com/krakjoe/pcov` ?
ref https://github.com/sebastianbergmann/php-code-coverage/pull/673

kubawerlos: [Docs](https://github.com/krakjoe/pcov/blob/develop/INSTALL.md#installation) says:

> use `release` branch for last stable released version

.travis.yml
kubawerlos: I was afraid it would take long and be problematic, but it turned out to be fast - no more than 5 seconds.

What do you mean by prebuilt package? I imagine pcov will be official PHP extension, one day.

.travis.yml
kubawerlos: I cannot see any improvement with `--single-branch` - `--depth=1` already does the job.

keradus: so, release would be last stable 1.x or last stable 2.x ? you don't know. let us rely on BC versioning, and not random stable

.travis.yml
keradus: like a php docker image with phpize already built.
so, one would be able to execute code coverage as well on his local machine, using same command to generate it, without need to manually modify his own php installation on his host

kubawerlos: Can we do it separately? I'm not sure my docker's knowledge is good enough.

SpacePossum:
Thank you @kubawerlos.

PR #4340, opened 2019-03-18
Travis: build against 7.4snapshot instead of nightly

Slamdunk:
At the time of creating this PR, in Travis we have:

|TAG|PHP -v|
|-|-|
|`nightly`|8.0.0-dev|
|`7.4snapshot`|7.4.0-dev|

Ref: https://twitter.com/nikita_ppv/status/1094897743594770433

I think we can safely ignore PHP 8 for now, and instead check PHP 7.4 build results.

This PR base branch in `2.12` as this should also be merged in `2.14`, `3.0` and `master`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,7 +125,7 @@ jobs:
         -
             <<: *STANDARD_TEST_JOB
             stage: Test
-            php: nightly
+            php: 7.4snapshot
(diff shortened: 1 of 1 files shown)


kubawerlos:
Would it be bad if we have both? Still all would finish before `5.6`.

Slamdunk:
> Would it be bad if we have both?

8 is far far away, it would just be a waste of resources, both electric and maintainers time, to me.

> With `allow_failures` we ignore `7.4` as well.

I would remote it from `allow_failures`, I just didn't want to push anything except a fix in this PR.

SpacePossum:
Thank you @Slamdunk.

PR #4191, opened 2019-01-01
[7.3] NoWhitespaceBeforeCommaInArrayFixer - fix comma after heredoc-end

gharlan:
Input:

```php
$a = [<<<'EOD'
foo
EOD
    , <<<'EOD'
bar
EOD
];
```

With PHP < 7.3 this is not changed (like before). With PHP >= 7.3 it would be changed now to:

```php
$a = [<<<'EOD'
foo
EOD, <<<'EOD'
bar
EOD
];
```

PR #4342, opened 2019-03-18
[7.3] MethodArgumentSpaceFixer - fix comma after heredoc-end

gharlan:
Input:

```php
foo(
    <<<'EOD'
        bar
        EOD
    ,
    'baz'
);
```

With PHP 7.3 and new config option `after_heredoc` it is fixed to:

```php
foo(
    <<<'EOD'
        bar
        EOD,
    'baz'
);
```

(similar to #4191)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -845,6 +845,8 @@ Choose from the list of available rules:
 
   Configuration options:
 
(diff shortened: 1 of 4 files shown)


SpacePossum:
Thank you @gharlan.

PR #4343, opened 2019-03-19
Fix return void on "yield from" calls

jderusse:
the `yield from` token is not identical to `yield`

That PR fix does not change the signature of such method:
```
function foo() {
  yield from bar();
}
```

I don't figured out how to add a test that assert that the Fixer did not change the signature of the method...

### DIFF
--- a/src/Fixer/FunctionNotation/VoidReturnFixer.php
+++ b/src/Fixer/FunctionNotation/VoidReturnFixer.php
@@ -200,7 +200,7 @@ final class VoidReturnFixer extends AbstractFixer
                 continue;
             }
 
-            if ($tokens[$i]->isGivenKind(T_YIELD)) {
+            if ($tokens[$i]->isGivenKind(T_YIELD) || $tokens[$i]->isGivenKind(T_YIELD_FROM)) {
                 return false; // Generators cannot return void.
             }
 


SpacePossum:
Hi @jderusse 

Thanks for the PR :)

Here is a test case to prove the issue:

```
['<?php function foo($param) { yield from test(); }'],
```

to be added to `tests/Fixer/FunctionNotation/VoidReturnFixerTest.php` somewhere in `provideFixCases`.

Can you rebase/retarget the PR to 2.12 as well, that way we've the bug fix on all the supported branched :)

src/Fixer/FunctionNotation/VoidReturnFixer.php
```diff
@@ -200,7 +200,7 @@ private function hasVoidReturn(Tokens $tokens, $startIndex, $endIndex)
                 continue;
             }
 
-            if ($tokens[$i]->isGivenKind(T_YIELD)) {
+            if ($tokens[$i]->isGivenKind(T_YIELD) || $tokens[$i]->isGivenKind(T_YIELD_FROM)) {
```
SpacePossum: please update to `if ($tokens[$i]->isGivenKind([T_YIELD, T_YIELD_FROM])) {` :)

SpacePossum:
@see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4344

PR #4094, opened 2018-11-12
BracesFixer - NoTrailingWhitespaceInCommentFixer - Priority

dmvdbrugge:
Fixes #3880 ~when #4085 is merged first 😜~

~(Test will thus fail for now, and succeed after rebase when said PR is merged.)~

Separated from #4085 by just specifying a lower priority.

PR #4346, opened 2019-03-21
BracesFixer - Do not pull close tag onto same line as a comment

SpacePossum:
replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4094

this stops the fixer changing:

```php
<?php if ($a) { //
?>
```

into

```php
<?php if ($a) { // ?>
```


dmvdbrugge:
I disagree with the premise, I think it should pull it up.

dmvdbrugge:
What, no discussion?

SpacePossum:
There was between 3 people of 8 days?

dmvdbrugge:
Kubawerlos only commented on the code, I disagreed with the idea behind the PR. My comment got no reply from you. I don't call that a discussion.

SpacePossum:
I'm not sure what more there is to discuss, some people like this form `<?php if($a) { // ?>` and others don't. I didn't find that form in a major project FOSS project either.

dmvdbrugge:
Ok, so there's the discussion, thanks. Not much I can counter with, if it boils down to stylistic choice only.

Would've been nice to have had that comment earlier though, instead of ignoring me for a week and then just merging, especially since I was the bug reporter and even submitted a PR for it.

SpacePossum:
sorry about that, I just picked of all the RTM PR's without looking into it much more, I'll try to pay more attention next time

PR #4350, opened 2019-03-28
StrictParamFixer - Don't detect functions in use statements

bolmstedt:
StrictParamFixer detects things like `use Safe\base64_decode;` and tries to fix the very next function, usually something like `__construct($foo)` by adding a `true` to it, making it something like `__construct($foo, true)`.

Failing test example:
```php
<?php
use function \base64_decode;
foo($bar);
```
(The leading backslash is the minimum requirement to trigger the bug, without it, the already existing tests will pass, because they check if the previous token is a `use function`.)

This change tries to make sure it's actually a function call that is detected.

### DIFF
--- a/tests/Fixer/Strict/StrictParamFixerTest.php
+++ b/tests/Fixer/Strict/StrictParamFixerTest.php
@@ -155,6 +155,16 @@ final class StrictParamFixerTest extends AbstractFixerTestCase
         array_keys($foo, $bar);
     }',
             ],
+            [
+                '<?php
+    use function \base64_decode;
+    foo($bar);',
+            ],
+            [
+                '<?php
+    use function Baz\base64_decode;
+    foo($bar);',
+            ],
         ];
     }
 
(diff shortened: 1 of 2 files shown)


tests/Fixer/Strict/StrictParamFixerTest.php
```diff
@@ -155,6 +155,11 @@ public function __construct($foo, $bar) {}
         array_keys($foo, $bar);
     }',
             ],
+            [
+                '<?php
+    use function Foo\base64_decode;
+    foo($foo);',
```
kubawerlos: Can we have here `baz` instead of `foo` to point that function's name don't have to be the same as namespace?

bolmstedt: I reduced the test case to it's simplest form:
```php
<?php
use function \base64_decode;
foo($bar);',
```

SpacePossum:
Thank you @bolmstedt.

PR #4353, opened 2019-04-01
Add more priority tests

SpacePossum:


tests/Fixtures/Integration/misc/method_separation,indentation_type.test
```diff
@@ -0,0 +1,23 @@
+--TEST--
+Integration of fixers: method_separation, indentation_type
+--RULESET--
+{"class_attributes_separation": true, "indentation_type": true}
```
kubawerlos: Not matching description.

SpacePossum: good catch, I also created it in the wrong directory and didn't update the FixerFactoryTest :man_facepalming:

PR #4354, opened 2019-04-02
Show not so usefull verbose output only with -v

Gamesh:
Currently these 3 lines are outputed for every single file php-cs-fixer scans
```
Loaded config default from ".../.php_cs.dist".
```

which adds a lot of noise on pre-commit hooks and there is no way to turn it off

### DIFF
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -173,7 +173,7 @@ final class FixCommand extends Command
                 sprintf(
                     'Loaded config <comment>%s</comment>%s.',
                     $resolver->getConfig()->getName(),
-                    null === $configFile ? '' : ' from "' . $configFile . '"'
+                    null === $configFile ? '' : ' from "'.$configFile.'"'
                 ),
                 OutputInterface::VERBOSITY_VERBOSE
             );
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -183,7 +183,7 @@ final class FixCommand extends Command
                 if (is_file($cacheFile)) {
                     $stdErr->writeln(
                         sprintf('Using cache file "%s".', $cacheFile),
-                        OutputInterface::VERBOSITY_VERY_VERBOSE
+                        OutputInterface::VERBOSITY_VERBOSE
                     );
                 }
             }
--- a/src/Console/Command/FixCommand.php
+++ b/src/Console/Command/FixCommand.php
@@ -169,12 +169,22 @@ final class FixCommand extends Command
             }
 
             $configFile = $resolver->getConfigFile();
-            $stdErr->writeln(sprintf('Loaded config <comment>%s</comment>%s.', $resolver->getConfig()->getName(), null === $configFile ? '' : ' from "'.$configFile.'"'));
+            $stdErr->writeln(
+                sprintf(
+                    'Loaded config <comment>%s</comment>%s.',
+                    $resolver->getConfig()->getName(),
+                    null === $configFile ? '' : ' from "' . $configFile . '"'
+                ),
+                OutputInterface::VERBOSITY_VERBOSE
+            );
 
             if ($resolver->getUsingCache()) {
                 $cacheFile = $resolver->getCacheFile();
                 if (is_file($cacheFile)) {
-                    $stdErr->writeln(sprintf('Using cache file "%s".', $cacheFile));
+                    $stdErr->writeln(
+                        sprintf('Using cache file "%s".', $cacheFile),
+                        OutputInterface::VERBOSITY_VERY_VERBOSE
+                    );
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Console/Command/FixCommand.php
```diff
@@ -169,12 +169,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                OutputInterface::VERBOSITY_VERBOSE
+            );
 
             if ($resolver->getUsingCache()) {
                 $cacheFile = $resolver->getCacheFile();
                 if (is_file($cacheFile)) {
-                    $stdErr->writeln(sprintf('Using cache file "%s".', $cacheFile));
+                    $stdErr->writeln(
+                        sprintf('Using cache file "%s".', $cacheFile),
+                        OutputInterface::VERBOSITY_VERY_VERBOSE
```
SpacePossum: please make this the same level; `VERBOSITY_VERBOSE`

SpacePossum:
Hi @Gamesh and thanks for your PR!

Could you target 2.12 as it is the lowest supported branch? It will bring this enhancement to all supported versions :)

src/Console/Command/FixCommand.php
Gamesh: changed, felt like it was something even more verbose

Gamesh:
@SpacePossum should i just change the base or cherry pick the commits to 2.12 branch?

kubawerlos:
Would it be hard to have test to ensure no extra ouptut is displayed without verbose mode?

Gamesh:
> Would it be hard to have test to ensure no extra ouptut is displayed without verbose mode?
@kubawerlos we could probably create an integration test for it, do you have a good example i can look at already in your tests folder?

kubawerlos:
`ReadmeCommand` has a [test](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.14.2/tests/Console/Command/ReadmeCommandTest.php) to verify it's up to date - I'd imagine something similar - with checking output against some RegEx, or simply if it has anything more that version message.

SpacePossum:
@Gamesh a rebase on top of 2.12 and than updating the target here in the PR would be best :)

There is a test; `tests/Console/Command/FixCommandTest.php` this one already fetches the output, however it asserts the exit code only.

src/Console/Command/FixCommand.php
```diff
@@ -169,12 +169,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
             $configFile = $resolver->getConfigFile();
-            $stdErr->writeln(sprintf('Loaded config <comment>%s</comment>%s.', $resolver->getConfig()->getName(), null === $configFile ? '' : ' from "'.$configFile.'"'));
+            $stdErr->writeln(
+                sprintf(
+                    'Loaded config <comment>%s</comment>%s.',
+                    $resolver->getConfig()->getName(),
+                    null === $configFile ? '' : ' from "'.$configFile.'"'
+                ),
+                OutputInterface::VERBOSITY_VERBOSE
```
keradus: please no. we had wasted so many time because user was not aware that he is using a config file

src/Console/Command/FixCommand.php
```diff
@@ -169,12 +169,22 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            );
 
             if ($resolver->getUsingCache()) {
                 $cacheFile = $resolver->getCacheFile();
                 if (is_file($cacheFile)) {
-                    $stdErr->writeln(sprintf('Using cache file "%s".', $cacheFile));
+                    $stdErr->writeln(
+                        sprintf('Using cache file "%s".', $cacheFile),
+                        OutputInterface::VERBOSITY_VERBOSE
```
keradus: again, we wasted time multiple times investigating why app is not checking all the files, while it was simply ignoring them due to cache

src/Console/Command/FixCommand.php
```diff
@@ -184,7 +194,11 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         if (null !== $stdErr && $resolver->configFinderIsOverridden()) {
             $stdErr->writeln(
-                sprintf($stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s', 'Paths from configuration file have been overridden by paths provided as command arguments.')
+                sprintf(
+                    $stdErr->isDecorated() ? '<bg=yellow;fg=black;>%s</>' : '%s',
+                    'Paths from configuration file have been overridden by paths provided as command arguments.'
```
keradus: this is most critical notice from all 3 of them, don't hide it.

Gamesh:
the problem is that there is no way to hide this output in any way and it produces noise a lot of it, if we use it as a pre-commit hook, developer modifies 20 files and gets 60 lines of this noise and then in between something meaningful might get outputted and you miss it. We are not removing it we're just hiding it under verbose flag so we can control it. 
Current implementation outputs it to STDERR so we can't even filter it with grep as it works only on STDOUT
Also we explored using -q, that hides the messages but it also hides the final output of fixed files which is something we need, so it's no go also

i'm open to other suggestions, maybe a config option that by default leaves the current behavior on.
We seem to have a lot of output but it's on the same level of verbosity

Gamesh:
since we don't have an agreement and found another way and i don't want to make your life as a maintainer harder, i'm closing this review :) 
But i still think much of the output in daily usage is too verbose and could be hidden. Sure it's useful when debugging, but not so much in daily actual usage. It would be beneficial to make some of the output verbose or very verbose only

keradus:
> developer modifies 20 files and gets 60 lines

hint: modified 20 files? run tool once, not 20 times.

> We are not removing it we're just hiding it under verbose flag so we can control it.

sadly, regular user run tool without debug, see unexpected (for him) outcome, he doesn't understand why, he doesn't run verbose flag

Gamesh:
> hint: modified 20 files? run tool once, not 20 times.

on that you're right but there is no way to list the files that were changed, so we opted for per file run.

> sadly, regular user run tool without debug, see unexpected (for him) outcome, he doesn't understand why, he doesn't run verbose flag

could also be true, or when we don't understand we usually run tools in verbose mode. For most tools that's what it's for, to debug, to see what wget/curl or other tools are doing behind the scenes

Issue #2166, opened 2016-09-01
[RFC] Import types with `use` statements only for non native types


phansys:
This proposed fixer should perform the following transformation:
**ACTUAL:**

``` php
namespace Example;

use SomeCustomType;
use Some\Qcn;
use RuntimeException;

class SomeClass
{
    public function myMethod()
    {
        throw new RuntimeException();
    }
}
```

**EXPECTED:**

``` php
namespace Example;

use SomeCustomType;
use Some\Qcn;

class SomeClass
{
    public function myMethod()
    {
        throw new \RuntimeException();
    }
}
```

keradus:
i remember there were request to do the other way around

keradus:
:-1: for it from me, hard to implement when you want to remove import only for PHP-defined global classes and not user-defined global classes and have it working for all different PHPs, also, it makes execution slower... if one want to have it he could keep it without fixer I would say, if one want to automatically do the importing (Eg only before creating release) the fixer would be nice.

but yeah, let us wait for more comments.

GrahamCampbell:
Looks doable with little pain, but we might have to be careful when it comes to updating phpdoc.

GrahamCampbell:
The phpdoc side of things maybe doesn;t make this worth the effort.

localheinz:
I was just thinking about it, and wanted to open a ticket, then I ended up here.

I like it!

gharlan:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4355 is merged, should we close this?
(the fixer does not differientiate between native/non-native elements, only between global/non-global)

SpacePossum:
Yeah, looks good to me. Feel free to comment if I missed a detail in the discussion. Thanks again @gharlan for writing the fixer!

PR #4355, opened 2019-04-02
GlobalNamespaceImportFixer - Introduction

### LINKED ISSUE #1309: Add ImportGlobalClassesFixer
I'd like to see a fixer that adds import statements for fully-qualified global classes.

Input:

``` php
class MyClass
{
    public function doSomething()
    {
        $date = new \DateTime();
        // ...
    }
}
```

Output:

``` php
use DateTime;

class MyClass
{
    public function doSomething()
    {
        $date = new DateTime();
        // ...
    }
}
```

### LINKED ISSUE #2739: request for additional configuration option for native_function_invocation
I've finally been getting around to using php-cs-fixer 2.x (not waiting around for #2553 to be resolved), and I ended up experimenting with it on vimeo/psalm#152, but some of the tests produce a rather verbose diff file ([nearly 7000 lines)(https://gist.github.com/SignpostMarv/840cac5070e7b5e840588b8e51903c18#file-native_function_invocation-only-diff)).

The argument behind `native_function_invocation` is to speed up resolving, and most things that improve performance are good things, but having such large diffs in pull requests is probably a bad idea (for several reasons that I won't go into at the moment).

So a thought occurred to me; `use function strtolower;` produces the same opcodes as `\strtolower()`; see [3v4l.org: backslash](https://3v4l.org/gvU56/vld#output) vs. [3v4l.org: use function](https://3v4l.org/IUsaC/vld#output).

Since `use function` [has been available since 5.6](https://secure.php.net/manual/en/language.namespaces.importing.php), would it be possible to have a configuration option added to `native_function_invocation` to enable smaller diffs to be generated with the same performance improvement? i.e.:
```php
$rules = [
    'native_function_invocation' => [
        'import' => true, // defaults to false to preserve existing behaviour
    ],
];
```

for convenience/those not wanting to click on 3val.org links:
```php
<?php

namespace Foo;

use function strtolower;

strtolower('Foo');
```

```sh
Finding entry points
Branch analysis from position: 0
Jump found. (Code = 62) Position 1 = -2
filename:       /in/IUsaC
function name:  (null)
number of ops:  4
compiled vars:  none
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   7     0  E >   INIT_FCALL                                               'strtolower'
         1        SEND_VAL                                                 'Foo'
         2        DO_ICALL                                                 
         3      > RETURN                                                   1
```

```php
<?php

namespace Foo;

\strtolower('bar');
```

```sh
Finding entry points
Branch analysis from position: 0
Jump found. (Code = 62) Position 1 = -2
filename:       /in/gvU56
function name:  (null)
number of ops:  4
compiled vars:  none
line     #* E I O op                           fetch          ext  return  operands
-------------------------------------------------------------------------------------
   5     0  E >   INIT_FCALL                                               'strtolower'
         1        SEND_VAL                                                 'bar'
         2        DO_ICALL                                                 
         3      > RETURN                                                   1
```

### LINKED ISSUE #4347: Add no_root_imports fixer
With this fixer, an import such as `use PDO;` would be removed and all references to `PDO` in that file would be replaced with `\PDO`. Root imports are those which do not contain a backslash (`\`) character, excluding the first. The following are root imports:

* `PDO`
* `\PDO`

The following are not root imports:

* `Foo\Bar`.

The rationale behind this fixer is to make it clear, when reading the code, whether the built-in implementation is used or a custom implementation. That is, when code follows the *no root imports* style, we can be confident that code occurences of `\PDO` refer to the standard implementation, and `PDO` refers to a custom implementation, because PHP built-in classes are not namespaced.

gharlan:
* closes #1309 
* refs #2166 (this pr does not differientiate between native/non-native elements, only between global/non-global)
* closes #2739
* closes #4347

---

The fixer can import global classes/functions/constants:

Input:

```php
<?php

namespace Foo;

if (\count($x)) {
    /** @var \DateTimeImmutable $d */
    $d = new \DateTimeImmutable();
    $p = \M_PI;
}
```

Output:

```php
<?php

namespace Foo;
use DateTimeImmutable;
use function count;
use const M_PI;

if (count($x)) {
    /** @var DateTimeImmutable $d */
    $d = new DateTimeImmutable();
    $p = M_PI;
}
```

Global functions/constants without leading `\` are not imported. The slash can be added by `Native(Constant|Function)InvocationFixer` before.

The fixer can also do the reverse fix, adding the backslash to imported global classes/functions/constants. But the fixer does not remove the imports, this can be done by `NoUnusedImportsFixer`.


README.rst
```diff
@@ -716,6 +716,19 @@ Choose from the list of available rules:
...
+  Imports or fully qualifies global classes/functions/constants.
+
+  Configuration options:
+
+  - ``import_classes`` (``false``, ``null``, ``true``): whether to import, not import or
+    ignore global classes; defaults to ``null``
+  - ``import_constants`` (``false``, ``null``, ``true``): whether to import, not import or
+    ignore global constants; defaults to ``null``
+  - ``import_functions`` (``false``, ``null``, ``true``): whether to import, not import or
+    ignore global functions; defaults to ``null``
```
gharlan: At the moment the default value for all options is `null`. But this means the fixer wouldn't do anything without configuration.
This seems not to be allowed.
So which default value should we use here?

julienfalque: There already are some fixers that can be configured in a way that does not really make sense. Usually there is no validation error in such case so maybe the current setup is fine.

Though maybe we could split the fixer into three new ones, which would allow to remove `null` from allowed configuration values.

gharlan: I get this error with phpunit:

> [global_namespace_import] Has no sample for default configuration.

But I can not add a sample for the default configuration, because samples without any change after fixing, are rejected by the tests.

So we have to change the defaults to `true` (or `false`), but we can still accept `null`.

Bilge:
Why has this stalled?

README.rst
julienfalque: I would default to:
- `import_classes`: `true`
- `import_functions`: `null`
- `import_constants`: `null`

as this seems to be the most widespread way of using imports.

svnldwg:
Very interested in this. Let me know if I can support.

gharlan:
ready for review

gharlan:
just for testing purposes (!), I applied the fixer (with `'import_*' => true`) to this repo: https://github.com/gharlan/PHP-CS-Fixer/commit/6c1f2730d466eb7ddd1f1e5ef87049e447b8a4d3
seems to work :)

BobMali:
Not familiar with the workflow. When is this going to be released?

kubawerlos:
@BobMali one day ;)

Currently, there is a lot of PRs waiting for review, any help would be great.

Even - or maybe especially - testing the fixer in practice will help - if you have some legacy project - apply the fixer and see if all is good - if not, comment with test cases that are not yet handled.

SpacePossum:
On the branch I tried:

```
    ->setRules([
        'global_namespace_import' => [
            'import_classes' => true,
        ],
    ])
```

resulting in:

```
$ php php-cs-fixer fix -vvv src/Report/JunitReporter.php --dry-run --diff
F
   1) src/Report/JunitReporter.php (global_namespace_import)
$ php php-cs-fixer fix -vvv src/Report/JunitReporter.php --dry-run --diff
F
   1) src/Report/JunitReporter.php (global_namespace_import)
```

which generated the import twice;

```diff
--- Original
+++ New
@@ @@
 use Symfony\Component\Console\Formatter\OutputFormatter;
+use RuntimeException;
+use DOMElement;
+use DOMDocument;
+use DomElement;
@@ @@
         if (!\extension_loaded('dom')) {
-            throw new \RuntimeException('Cannot generate report! `ext-dom` is not available!');
+            throw new RuntimeException('Cannot generate report! `ext-dom` is not available!');
         }
 
-        $dom = new \DOMDocument('1.0', 'UTF-8');
+        $dom = new DOMDocument('1.0', 'UTF-8');
         $testsuites = $dom->appendChild($dom->createElement('testsuites'));
-        /** @var \DomElement $testsuite */
+        /** @var DomElement $testsuite */
@@ @@
 
-    private function createSuccessTestCase(\DOMDocument $dom, \DOMElement $testsuite)
+    private function createSuccessTestCase(DOMDocument $dom, DOMElement $testsuite)
@@ @@
     /**
-     * @param \DOMDocument  $dom
-     * @param \DOMElement   $testsuite
+     * @param DOMDocument  $dom
+     * @param DOMElement   $testsuite
      * @param ReportSummary $reportSummary
      */
-    private function createFailedTestCases(\DOMDocument $dom, \DOMElement $testsuite, ReportSummary $reportSummary)
+    private function createFailedTestCases(DOMDocument $dom, DOMElement $testsuite, ReportSummary $reportSummary)
@@ @@
     /**
-     * @param \DOMDocument $dom
+     * @param DOMDocument $dom
      * @param string       $file
      * @param array        $fixResult
      * @param bool         $shouldAddAppliedFixers
      *
-     * @return \DOMElement
+     * @return DOMElement
      */
-    private function createFailedTestCase(\DOMDocument $dom, $file, array $fixResult, $shouldAddAppliedFixers)
+    private function createFailedTestCase(DOMDocument $dom, $file, array $fixResult, $shouldAddAppliedFixers)
     {
         $appliedFixersCount = \count($fixResult['appliedFixers']);
 
         $testName = str_replace('.', '_DOT_', Preg::replace('@\.'.pathinfo($file, PATHINFO_EXTENSION).'$@', '', $file));
 
         $testcase = $dom->createElement('testcase');
         $testcase->setAttribute('name', $testName);
         $testcase->setAttribute('file', $file);
         $testcase->setAttribute('assertions', (string) $appliedFixersCount);
 
         $failure = $dom->createElement('failure');
         $failure->setAttribute('type', 'code_style');
         $testcase->appendChild($failure);
 
         if ($shouldAddAppliedFixers) {
             $failureContent = "applied fixers:\n---------------\n";
 
             foreach ($fixResult['appliedFixers'] as $appliedFixer) {
                 $failureContent .= "* {$appliedFixer}\n";
             }
         } else {
             $failureContent = "Wrong code style\n";
         }
 
         if (!empty($fixResult['diff'])) {
             $failureContent .= "\nDiff:\n---------------\n\n".$fixResult['diff'];
         }
 
         $failure->appendChild($dom->createCDATASection(trim($failureContent)));
 
         return $testcase;
     }
 }
```

gharlan:
@SpacePossum good catch! Please check again.

SpacePossum:
Thank you @gharlan.

SpacePossum:
@gharlan can you check if this makes sense:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/00b8cc9cdc4a6c86d168ac55fc403150cd418832

emodric:
Hi! Is it possible to ignore PHPDocs with this fixer, i.e. only replace global namespaces with imports in actual code?

emodric:
The usecase is really consistency. Across organization, we have a rule to always use FQCN in doc blocks and for consistency sake, that includes classes from global namespace too.

So, e.g. we always use:

```
/** @var \DateTimeInterface $date */
$date = ...;
```

instead of:

```
use DateTimeInterface;

/** @var DateTimeInterface $date */
$date = ...;
```

julienfalque:
Isn't removing this exception from code style rules a better alternative?

emodric:
With what?

julienfalque:
I mean: you want to apply `global_namespace_import`, which means the code style you want to follow is:

- import all symbols
- except when they are in a PHPDoc

There doesn't seem to be an actual need for the second part other than arbitrary code style choice. Can't the code style be updated to remove the second part then? `global_namespace_import` would fix everything automatically.

If consistency is what matters, this is the way to go IMO: symbols would always be imported instead of the current mixed situation where some are imported and other are fully qualified.

emodric:
We don't like having code which purpose is only to enable comments or vice versa. That is, we strongly feel that imports should be related only to actual code and PHPDoc should have FQCNs to make them more clear. Same thing with annotations. We don't use them for the same reason.

Anyway, we always relied on IDE to make sure classes are imported in code, we'll continue to do so. We're using this fixer for global functions and constants really, so no harm, no foul.

Issue #4356, opened 2019-04-03
Fails to properly parse syntax when function used inside of {}


ikari7789:
Not sure what configuration setting exactly is causing the issue, but I had to go in and disable the linting in order to figure out what was happening (which isn't very direct now in v2) and the error message is basically useless since it reverts the change without telling you anything about what the issue was.

We have a particular function which has a deep replace on array keys, and does some string conversion to camel case, and thus we have a line line this:

```php
$this->{camel_case($key)}[$arrayKey] = $arrayValue;
```

When php-cs-fixer parses this file, it attempts to change it to:

```php
$this->{camel_case($key)}list($arrayKey) = $arrayValue;
```

I've provided all the relevant information below as well as a code snippet that should reproduce the issue (at least, I was able to reproduce it on my local machine).

#### The PHP version you are using (`$ php -v`):
=> PHP 7.3.3 (cli) (built: Mar  8 2019 16:40:07) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.3, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.3.3, Copyright (c) 1999-2018, by Zend Technologies

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.14.2 Sunrise by Fabien Potencier and Dariusz Ruminski

#### The command you use to run PHP CS Fixer:
=> php-cs-fixer fix

#### The configuration file you are using, if any:
```php
<?php

$finder = PhpCsFixer\Finder::create()
    ->exclude(['.git', 'build', 'storage', 'vendor'])
    ->in(__DIR__);

/**
 * To know the effect of any of the following rules, please execute:
 *
 *     php-cs-fixer describe <rule_name>
 *
 */
$config = PhpCsFixer\Config::create()
    ->setRiskyAllowed(false)
    ->setRules([
        'array_syntax' => ['syntax' => 'short'],
        'binary_operator_spaces' => [
            'operators' => [
                '=' => 'align_single_space_minimal',
                '=>' => 'align_single_space_minimal',
            ],
        ],
        'blank_line_after_namespace' => true,
        'blank_line_after_opening_tag' => true,
        'blank_line_before_statement' => true,
        'braces' => true,
        'cast_spaces' => true,
        'class_definition' => true,
        'class_keyword_remove' => false,
        'combine_consecutive_issets' => true,
[...]

PR #4357, opened 2019-04-03
Fix short list syntax detection.

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4356

### DIFF
--- a/src/Tokenizer/Transformer/SquareBraceTransformer.php
+++ b/src/Tokenizer/Transformer/SquareBraceTransformer.php
@@ -47,6 +47,15 @@ final class SquareBraceTransformer extends AbstractTransformer
     /**
      * {@inheritdoc}
      */
+    public function getPriority()
+    {
+        // must run after CurlyBraceTransformer
+        return -1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     public function getRequiredPhpVersionId()
     {
         // Short array syntax was introduced in PHP 5.4, but the fixer is smart
--- a/tests/Fixer/ListNotation/ListSyntaxFixerTest.php
+++ b/tests/Fixer/ListNotation/ListSyntaxFixerTest.php
@@ -88,6 +88,17 @@ $a#
 ;#',
         ];
 
+        $cases[] = ['<?php
+
+class Test
+{
+    public function updateAttributeKey($key, $value)
+    {
+        $this->{camel_case($attributes)}[$key] = $value;
+    }
+}',
+        ];
+
         return $cases;
     }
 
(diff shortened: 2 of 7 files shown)


tests/Fixer/ListNotation/ListSyntaxFixerTest.php
```diff
@@ -88,6 +88,17 @@ public function provideToLongCases()
 ;#',
         ];
 
+        $cases[] = ['<?php
+
+class Test
+{
+    public function updateAttributeKey($key, $value)
+    {
+        $this->{camel_case($attributes)}[$key] = $value;
```
kubawerlos: Is `camel_case` intentional? Got me confused seeing words "camel case" in snake case ;)

SpacePossum: it is a copy paste of the reported sample that failed, I hadn't notice this until you pointed it out ;)

ikari7789: Sorry, was a copy + paste from our Laravel app. `camel_case()` is a helper function. ^^;

src/Tokenizer/Transformer/SquareBraceTransformer.php
```diff
@@ -44,6 +44,15 @@ public function getCustomTokens()
         ];
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // must run after CurlyBraceTransformer
```
SpacePossum: both `BraceClassInstantiationTransformer` and `CurlyBraceTransformer` check for `CT::T_ARRAY_SQUARE_BRACE_CLOSE` which comes from this transformer, so there is a priority issue...

SpacePossum:
I looked into the priority issue some more. The ideal situation would be:

```
CurlyBraceTransformer
	after SquareBraceTransformer because of:
		T_ARRAY_SQUARE_BRACE_CLOSE

SquareBraceTransformer
	after CurlyBraceTransformer because of:
		CT::T_DYNAMIC_PROP_BRACE_CLOSE
		CT::T_DYNAMIC_VAR_BRACE_CLOSE
		CT::T_ARRAY_INDEX_CURLY_BRACE_CLOSE
```

however, without splitting up the transformers this cannot be resolved.
The reason why CurlyBraceTransformer needs T_ARRAY_SQUARE_BRACE_CLOSE is for cases like:

```php
<?php $a = $b[0]{1};
```

Currently this is not an issue, because the transformer already/also accepts `]`.
The only other meaning `]` that could become an issue would be destructuring;

```php
<?php [$a]{1} = 1;
```

however these cases are always invalid in syntax and during execution.

Issue #3900, opened 2018-07-10
FunctionToConstant - update `get_class`


SpacePossum:
Currently `get_class($this)` is not fixed.
Update the logic to change `get_class($this)` to:
> `self::class` if the class is final
> `static::class` if not final

as such the rule must run after `final_internal_class`

note by kerad:
`@Symfony` ruleset

ntzm:
However, https://3v4l.org/8Ts0d

SpacePossum:
alrighty,
maybe if `get_class` is used within a closure, fixture or trait, don't fix, alternative would be throw more details in the risky-text

ntzm:
I'd vote for making it risky personally

keradus:
if the file is final, I would like to see this change as well, even if there were no `get_class` originally:
```diff
-static::class
+self::class
```
but other would not.

IMO, 2 different rules, good priority and integration test, voila

fabpot:
I wanted to create an issue for the same. In Symfony, we have a lot of `get_class($this)` that could be converted to `static::class` (100% of them actually). Is it still something you want to consider adding to PHP-CS-Fixer?

keradus:
@fabpot , for sure sth we want to have.

@ntzm , concern raised by you - `self::class vs get_class($this)` is that `get_class` works as `static::class, not `self::class`.

SpacePossum:
added in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4742

PR #4358, opened 2019-04-04
SelfStaticAccessorFixer - Introduction

SpacePossum:
```
$ php php-cs-fixer describe self_static_accessor
Description of self_static_accessor rule.
Inside a final class or anonymous class `self::X` should be preferred to `static::X`.

Fixing examples:
 * Example #1.
   --- Original
   +++ New
   @@ -1,15 +1,15 @@
    <?php
    final class Sample
    {
        private static $A = 1;
    
        public function getBar()
        {
   -        return static::class.static::test().static::$A;
   +        return self::class.self::test().self::$A;
        }
    
        private static function test()
        {
            return 'test';
        }
    }
   
 * Example #2.
   --- Original
   +++ New
   @@ -1,7 +1,7 @@
    <?php
    $a = new class() {
        public function getBar()
        {
   -        return static::class;
   +        return self::class;
        }
    };
   
```
following:
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3900#issuecomment-404857159
- https://github.com/symfony/symfony/pull/27860


src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_CLASS, T_STATIC, T_DOUBLE_COLON]);
+    }
+
+    public function getPriority()
+    {
+        // must be run after FinalInternalClassFixer
+        return -10;
```
gharlan: and after FunctionToConstantFixer, because of `get_called_class()` -> `static::class` -> `self::class`

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $tokensAnalyzer = new TokensAnalyzer($tokens);
+
+        $classIndex = $tokens->getNextTokenOfKind(0, [[T_CLASS]]);
+
+        while (null !== $classIndex) {
+            if (
+                $tokens[$tokens->getPrevMeaningfulToken($classIndex)]->isGivenKind([T_FINAL])
```
gharlan:
```suggestion
                $tokens[$tokens->getPrevMeaningfulToken($classIndex)]->isGivenKind(T_FINAL)
```

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $classIndex of T_CLASS of `final` or `anonymous` class
+     */
+    private function fixClass(Tokens $tokens, $classIndex)
+    {
+        // figure out where the classy starts
```
gharlan: not trait or interface here, only real classes, so no need for class**y** ;)

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,129 @@
...
+        $tokensAnalyzer = new TokensAnalyzer($tokens);
+
+        $classIndex = $tokens->getNextTokenOfKind(0, [[T_CLASS]]);
+
+        while (null !== $classIndex) {
+            if (
+                $tokens[$tokens->getPrevMeaningfulToken($classIndex)]->isGivenKind([T_FINAL])
+                || $tokensAnalyzer->isAnonymousClass($classIndex)
+            ) {
+                $this->fixClass($tokens, $classIndex);
```
gharlan:
```php
final class Foo 
{
    function bar() 
    {
        $c = new class() {
            function baz() 
            {
                static::class;
            }
        };
    }
}
```

In this case, `fixClass` is called two times. But `static::` is already replaced in first call, for outer class.
So I think this should be optimized, to not call `fixClass` again for inner classes.

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,154 @@
...
+                    // TODO fix anonymous classes inside lambda
+
+                    // figure out where the lambda ends
+                    $index = $tokens->findBlockEnd(Tokens::BLOCK_TYPE_CURLY_BRACE, $lambdaOpenIndex);
+                }
+
+                continue;
+            }
+
+            if (!$tokens[$index]->isGivenKind([T_STATIC])) {
```
gharlan: no array needed

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,154 @@
...
+                continue;
+            }
+
+            $staticIndex = $index;
+
+            $index = $tokens->getNextMeaningfulToken($index);
+            if (!$tokens[$index]->isGivenKind([T_DOUBLE_COLON])) {
```
gharlan: no array needed

src/Fixer/ClassNotation/SelfStaticAccessorFixer.php
```diff
@@ -0,0 +1,154 @@
...
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound([T_CLASS, T_STATIC, T_DOUBLE_COLON]);
```
gharlan: `isAllTokenKindsFound`

SpacePossum:
>new static();

done

>$foo instanceof static

TODO, thanks for the review!  :)

maxhelias:
@SpacePossum what do you think about add this rule the `@PhpCsFixer:risky` rule set ?

SpacePossum:
I would like that :) Lets see what others think!

(ping @ @julienfalque and @keradus)

keradus:
we would have a lot of conflicts to be honest...
#4537

PR #4366, opened 2019-04-09
Unset the auxillary variable $a

GrahamCampbell:
Otherwise it remains in scope, even if falsy: https://3v4l.org/RRGda.


SpacePossum:
Thank you @GrahamCampbell.

PR #4375, opened 2019-04-14
extend ruleset "@PHP73Migration"

gharlan:


README.rst
```diff
@@ -835,7 +835,7 @@ Choose from the list of available rules:
 
   *Risky rule: risky when any of the functions are overridden.*
 
-* **method_argument_space** [@PSR2, @Symfony, @PhpCsFixer]
+* **method_argument_space** [@PSR2, @Symfony, @PhpCsFixer, @PHP73Migration]
```
keradus: how is it related to 7.3 ?

README.rst
```diff
@@ -1175,7 +1175,7 @@ Choose from the list of available rules:
   There should not be an empty ``return`` statement at the end of a
   function.
 
-* **no_whitespace_before_comma_in_array** [@Symfony, @PhpCsFixer]
+* **no_whitespace_before_comma_in_array** [@Symfony, @PhpCsFixer, @PHP73Migration]
```
keradus: how is it related to 7.3 ?

README.rst
```diff
@@ -1809,11 +1809,11 @@ Choose from the list of available rules:
 
   Standardize spaces around ternary operator.
 
-* **ternary_to_null_coalescing** [@PHP70Migration, @PHP71Migration]
+* **ternary_to_null_coalescing** [@PHP70Migration, @PHP71Migration, @PHP73Migration]
 
   Use ``null`` coalescing operator ``??`` where possible. Requires PHP >= 7.0.
 
-* **trailing_comma_in_multiline_array** [@Symfony, @PhpCsFixer]
+* **trailing_comma_in_multiline_array** [@Symfony, @PhpCsFixer, @PHP73Migration]
```
keradus: how is it related to 7.3 ?

README.rst
```diff
@@ -1831,7 +1831,7 @@ Choose from the list of available rules:
 
   Unary operators should be placed adjacent to their operands.
 
-* **visibility_required** [@PSR2, @Symfony, @PhpCsFixer, @PHP71Migration]
+* **visibility_required** [@PSR2, @Symfony, @PhpCsFixer, @PHP71Migration, @PHP73Migration]
```
keradus: how is it related to 7.3 ?

gharlan: `@PHP73Migration` extends `@PHP71Migration`

README.rst
gharlan: the `after_heredoc` option is desired while migrating to php 7.3. That is the reason, why I've added the fixer with that option to the ruleset.

README.rst
gharlan: the `after_heredoc` option is desired while migrating to php 7.3. That is the reason, why I've added the fixer with that option to the ruleset.

README.rst
gharlan: the `after_heredoc` option is desired while migrating to php 7.3. That is the reason, why I've added the fixer with that option to the ruleset.

gharlan:
@keradus It would be nice when a ruleset could define `if fixer X is enabled, then set option Y to Z`.

gharlan:
But because we do not have this possibility at the moment, I think it is ok to add the three fixers with `after_heredoc` option to the PHP73Migration ruleset.
The migration rulesets are opinionated anyway (not everybody who is migrating to 7.1 wants the declare-strict-fixer). So I think it is ok, that we add the three fixers to the ruleset as our opinionated suggestion when migrating a code base to 7.3.

README.rst
keradus: ah, fair

keradus:
Idea behind 7.3 ruleset is to have `heredoc_indentation` in it, as this syntax came with 7.3.
I'm not sure that adding few more rules to it, not related to 7.3, is the best idea.

Maybe we shall follow similar idea we have for phpunit rules?
`target ('5.2', '5.6', 'newest'): target version of PHPUnit; defaults to 'newest'`

here, we can make sth like
deprecate `after_heredoc` and add
```
heredoc_handling ('remove', 'keep', `best_for_php_version`): ..., defaults to ``best_for_php_version``
```

what do you think ?

gharlan:
> Maybe we shall follow similar idea we have for phpunit rules?
> `target ('5.2', '5.6', 'newest'): target version of PHPUnit; defaults to 'newest'`

This would be a global configuration, like newline and indentation config?

gharlan:
I think a global configuration for the php target version would be nice.
In my opinion we could deprecate the `after_heredoc` option then without a new option. I can not imagine that someone (with target version 7.3) wants to use `method_argument_space` but wants to keep the newline between heredoc end and the comma.

gharlan:
We could remove the option without deprecation, because the option is not released yet.

keradus:
> This would be a global configuration, like newline and indentation config?

No, this would not be global config, it would be rule config. Look at example of phpunit_ rules I gave.


> I think a global configuration for the php target version would be nice.

We dont have a good mechanism for global rules config

gharlan:
Hmm then I don't understand your proposal, why is it better than the `after_heredoc` option?
We would still need to add the three fixers to the migration rule, but we would only change the `after_heredoc` option with the target php option.
I'm ok with this, but I don't see the benefit.

I thought we search a solution where we don't need to add the three fixers to the migration rule, but users with target php 7.3 would get the fixes after heredoc end anyway (when using these fixers).

keradus:
I let myself extract non-problematic part of this PR to #4375 keeping your authorship

Unfortunately, i'm out of home for loooong and had no time to properly think about this proposal. maybe some other maintainers would ?

kubawerlos:
@gharlan can you update this PR? It looks fine for me to have these 3 `after_heredoc` options in PHP 7.3 migration set.

@keradus did you had an opportunity to think about this PR?

@SpacePossum @julienfalque WDYT about this?

julienfalque:
IMO it is simpler if we either:
1. add the `after_heredoc` option like proposed here, making the ruleset a bit more opinionated, but users can easily opt out
2. remove the option and keep the ruleset more related to PHP *upgrade*, users can easily enable the option anyway

The alternative solution proposed by @keradus tries to keep the best of bost options but requires much more work, really not worth it IMO.

gharlan:
> @gharlan can you update this PR? It looks fine for me to have these 3 `after_heredoc` options in PHP 7.3 migration set.

I can do it, if there is a consensus to merge it. Otherwise let's close the pr.

SpacePossum:
>can you update this PR? It looks fine for me to have these 3 after_heredoc options in PHP 7.3 migration set.

:+1:

GrahamCampbell:
What's the status here?

SpacePossum:
rebased, conflicts resolved, force pushed to @gharlan his branch (!) , if the tests are green on all version, than RTM :+1:

GrahamCampbell (approved):

SpacePossum:
Thank you @gharlan.

PR #4376, opened 2019-04-14
PhpdocTypesFixer > Remove spaces around pipes

ruudk:
Sometimes, when adding multiple types in `@param` or `@return`, people add spaces before or after the pipe. That should be cleaned up as it's not standard.

### DIFF
--- a/tests/Fixer/Phpdoc/PhpdocTypesFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocTypesFixerTest.php
@@ -233,4 +233,54 @@ EOF;
 
         $this->fixer->configure(['groups' => ['__TEST__']]);
     }
+
+    public function testRemoveSpacesBeforeAndAfterPipe()
+    {
+        $expected = <<<'EOF'
+<?php
+    /**
+     * @param self|array|Foo $bar
+     *
+     * @return int|float|callback
+     */
+
+EOF;
+
+        $input = <<<'EOF'
+<?php
+    /**
+     * @param self | array | Foo $bar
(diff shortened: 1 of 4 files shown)


kubawerlos:
This seems valid, although there are new lines in `Annotation`, but no unit tests for this change (I mean new cases in `AnnotationTest`).

ruudk:
Good one... Added more tests.

ruudk:
@SpacePossum Anything I can do to bring this PR forward? Just rebased on master. Happy to make more changes.

tests/Fixer/Phpdoc/PhpdocTypesFixerTest.php
```diff
@@ -233,4 +233,54 @@ public function testWrongConfig()
...
+     *
+     * @return int|float|callback
+     */
+
+EOF;
+
+        $input = <<<'EOF'
+<?php
+    /**
+     * @param self | array | Foo $bar
```
keradus:
```
Description of phpdoc_types rule.
The correct case must be used for standard PHP types in PHPDoc.
```

Removing extra spaces has nothing to do with this fixer.
Your feature looks nice, but you can't just pick random fixer and add new logic there.
Make new rule (no spaces between types) as new fixer.

kubawerlos:
@ruudk FYI https://github.com/kubawerlos/php-cs-fixer-custom-fixers/pull/194

ruudk:
@kubawerlos Isn't it an idea to propose this to be included in PHP-CS-Fixer?

PR #4377, opened 2019-04-14
CommentsAnalyzer - fix for declare before header comment

kubawerlos:


SpacePossum:
can you add a test for:

```php
<?php

/**
 * 1
 */

declare(strict_types=1);

/**
 * 2
 */

namespace App\Controller\CMS\Capsule;

/**
 * 3
 */

final class Foo{}
```

so it show _only_ the first PHPDoc is considered a header?

php-cs-fixer
```diff
@@ -11,7 +11,7 @@
  * with this source code in the file LICENSE.
  */
 
-/**
+/*
  * @author Fabien Potencier <fabien@symfony.com>
  * @author Dariusz Rumiński <dariusz.ruminski@gmail.com>
```
keradus: drop this comment in general, it duplicates file header comment

SpacePossum:
Thank you @kubawerlos.

PR #4378, opened 2019-04-14
PhpUnitNoExpectationAnnotationFixer - annotation in single line doc comment

kubawerlos:


src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -265,7 +269,7 @@ private function extractContentFromAnnotation(Annotation $annotation)
             return '';
         }
 
-        $content = $matches[1];
+        $content = Preg::replace('/\*+\/$/', '', $matches[1]);
```
GrahamCampbell, line 272: You need to match exactly 2 stars followed by a space. One star is not phpdoc, it's a comment.

kubawerlos: That applies to the beginning of the comment, this RegEx is to match at least one `*`, followed by `$` at the end of the comment.

GrahamCampbell: Ah, yes, I misread those slashes at the start.

kubawerlos: Anything else regarding reviewing this then?

SpacePossum:
Thank you @kubawerlos.

PR #4379, opened 2019-04-15
ExplicitStringVariableFixer - add test case for variable as an array key

kubawerlos:
Is the test case correct? Ping @Slamdunk for an answer and help with fixing if so.


Slamdunk:
Sounds legit: https://3v4l.org/dHF7s

Slamdunk:
Fix published in https://github.com/kubawerlos/PHP-CS-Fixer/pull/3

Also added test cases for `"My name is $array[$foo][$bar] !"`: https://3v4l.org/ZhNsX

Slamdunk:
@SpacePossum not a `WIP` anymore

PR #4380, opened 2019-04-16
PHP7.4 - Add "str_split" => "mb_str_split" mapping.

SpacePossum:
@ see https://wiki.php.net/rfc/mb_str_split

### DIFF
--- a/src/Fixer/Alias/MbStrFunctionsFixer.php
+++ b/src/Fixer/Alias/MbStrFunctionsFixer.php
@@ -27,18 +27,19 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
     /**
      * @var array the list of the string-related function names and their mb_ equivalent
      */
-    private static $functions = [
+    private static $functionsMap = [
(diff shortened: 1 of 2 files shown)


JarJak:
This one should check first PHP version installed, no? Just got str_split replaced with mb_str_split on PHP 7.2...

src/Fixer/Alias/MbStrFunctionsFixer.php
```diff
@@ -85,8 +86,15 @@ public function isCandidate(Tokens $tokens)
      */
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
+        $functions = array_filter(
+            self::$functionsMap,
+            static function ($mapping) {
+                return \function_exists($mapping['alternativeName']);
```
keradus: this was causing the issue. when we execute this, it's always truth, as we have the mb polyfill on execution time...

keradus: ref #4529

keradus: cc @SpacePossum

PR #4384, opened 2019-04-20
MethodArgumentSpaceFixer - fix for on_multiline:ensure_fully_multiline with trailing comma in function call

kubawerlos:


src/Fixer/FunctionNotation/MethodArgumentSpaceFixer.php
```diff
@@ -421,6 +421,11 @@ private function fixNewline(Tokens $tokens, $index, $indentation, $override = tr
             return;
         }
 
+        $nextMeaningfulTokenIndex = $tokens->getNextMeaningfulToken($index);
+        if ($tokens[$nextMeaningfulTokenIndex]->equals(')')) {
```
keradus: Should this change be really applied only for master branch ?

julienfalque: The bug exists on branch 2.12 therefore the fix should be merged on that branch.

SpacePossum:
Thank you @kubawerlos.

Issue #3628, opened 2018-03-21
CommentToPhpdocFixer - Allow to ignore some tags


julienfalque:
I think it would be nice if `comment_to_phpdoc` could be configured to ignore a set of tags. My use case is the following:
```php
// @todo do something later
$foo = 'foo';
```
It is common to use regular comments for `@todo` tags and I would like those to be left untouched.

kubawerlos:
Right now it's under heavy refactoring as it was released too early: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3614

SpacePossum:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3614 has been merged so I think this can be picked up by someone or do you plan on more related-changes for this fixer @kubawerlos ?

kubawerlos:
After the merge the example is not valid anymore - only comments on structural elements will be candidates for PHPDoc. @julienfalque @SpacePossum do you still think allowing to ignore some tags would be useful?

SpacePossum:
if `//@todo xyx` and`// @FIXME abc` are no longer changed to PHPDoc's I think this can be closed indeed, I leave the call to @julienfalque (as reporter of this issue)

julienfalque:
Indeed the example isn't valid anymore but `// @TODO` would still be changed to a DocComment if placed before e.g. a class property. I would leave this feature request open.

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4385 thanks @kubawerlos

PR #4385, opened 2019-04-22
CommentToPhpdocFixer - allow to ignore tags

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3628

Ping @julienfalque for review, I've made the list case insensitive - I figured it's more natural, would you agree?

### DIFF
--- a/tests/Fixer/Comment/CommentToPhpdocFixerTest.php
+++ b/tests/Fixer/Comment/CommentToPhpdocFixerTest.php
@@ -26,11 +26,13 @@ final class CommentToPhpdocFixerTest extends AbstractFixerTestCase
     /**
      * @param string      $expected
      * @param null|string $input
+     * @param array       $config
      *
      * @dataProvider provideTestCases
      */
-    public function testFix($expected, $input = null)
+    public function testFix($expected, $input = null, array $config = [])
     {
+        $this->fixer->configure($config);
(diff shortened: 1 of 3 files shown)


keradus:
ping @julienfalque , can you take a look please?

tests/Fixer/Comment/CommentToPhpdocFixerTest.php
```diff
@@ -246,6 +248,70 @@ public function provideTestCases()
 EOT
                 ,
             ],
+            [
+                <<<'EOT'
+<?php /* header comment */ $foo = true;
+
+// @todo do something later
+$foo = 1;
```
julienfalque: Please add a test case with a comment before a structural element, I think the added tests here would still pass without the `ignored_tags`, thus they don't prove the option works as expected.

kubawerlos: What do you mean? Isn't this case with a comment before a structural element? Can you prepare the one you are thinking about?

julienfalque: Actually I misunderstood the code, sorry.

SpacePossum:
Thank you @kubawerlos.

PR #4390, opened 2019-04-26
DX: unify Cache types

kubawerlos:


keradus:
thanks @kubawerlos 
incorporated into #4393 , we don't need one-liner DX as separated PR.

PR #4392, opened 2019-04-26
DX: fix properties types

kubawerlos:


src/Fixer/LanguageConstruct/FunctionToConstantFixer.php
```diff
@@ -29,12 +29,12 @@
 final class FunctionToConstantFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
 {
     /**
-     * @var string[]
+     * @var array<string, array<int, Token>>
```
keradus:
```diff
-* @var array<string, array<int, Token>>
+* @var array<string, Token[]>
```

to indicate that numeric index is 0,1,2,3... and not 5,8,124,6 ?
same for later

src/Fixer/ControlStructure/YodaStyleFixer.php
```diff
@@ -36,7 +36,7 @@ final class YodaStyleFixer extends AbstractFixer implements ConfigurationDefinit
     private $candidatesMap;
 
     /**
-     * @var array<string, null|bool>
+     * @var array<int|string, null|bool>
```
keradus: :+1:

src/Linter/CachingLinter.php
```diff
@@ -25,7 +25,7 @@ final class CachingLinter implements LinterInterface
     private $sublinter;
 
     /**
-     * @var array<string, LintingResultInterface>
+     * @var array<int, LintingResultInterface>
```
keradus: :+1:

keradus:
thanks @kubawerlos
incorporated into #4393 , we don't need one-liner DX as separated PR.

PR #4393, opened 2019-04-26
DX: add missing explicit return types

kubawerlos:
Error was:
```
------ ----------------------------------------------------------------------------------------------------------------------- 
  Line   Console/Command/DescribeCommand.php                                                                                    
 ------ ----------------------------------------------------------------------------------------------------------------------- 
  110    Method PhpCsFixer\Console\Command\DescribeCommand::execute() should return int|null but empty return statement found.  
 ------ -----------------------------------------------------------------------------------------------------------------------
```


keradus:
Thank you @kubawerlos.

PR #4394, opened 2019-04-26
DX: remove reference from PHPDoc

kubawerlos:
`&` in PHPDoc is for [intersection types](https://medium.com/@ondrejmirtes/union-types-vs-intersection-types-fd44a8eacbb).

### DIFF
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -25,7 +25,7 @@ final class Preg
     /**
      * @param string        $pattern
      * @param string        $subject
-     * @param null|string[] &$matches
+     * @param null|string[] $matches
      * @param int           $flags
      * @param int           $offset
      *
@@ -51,7 +51,7 @@ final class Preg
     /**
      * @param string        $pattern
      * @param string        $subject
-     * @param null|string[] &$matches
+     * @param null|string[] $matches
      * @param int           $flags
      * @param int           $offset
      *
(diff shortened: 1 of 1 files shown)


kubawerlos:
Well, it usually is e.g. PHPStorm's PHP stubs does not have it.

> keradus requested changes

What changes would that be?

keradus:
> What changes would that be?

github has no specific option for "PR disapproved"


I did not got what you meant with `stubs`.
but once more - we don't use `&` for types intersection, as we don't use it between two types. we use it ahead of variable name, to indicate it's passed via reference

kubawerlos:
> we use it ahead of variable name, to indicate it's passed via reference

Why? [Symfony](https://github.com/symfony/symfony/blob/v4.2.8/src/Symfony/Component/PropertyAccess/PropertyAccessorInterface.php#L42) doesn't do this, neither [PHP Parser](https://github.com/nikic/PHP-Parser/blob/v4.2.1/lib/PhpParser/PrettyPrinterAbstract.php#L952), [PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/8.1.3/src/Framework/MockObject/Builder/InvocationMocker.php#L82) or [PHPStan](https://github.com/phpstan/phpstan/blob/0.11.5/src/Dependency/DependencyResolver.php#L192).

I'd say the information of how it's passed is useless.

keradus:
that's totally different argument than "& is for intersection" :)

keradus:
thank you, incorporated into #4393

PR #4401, opened 2019-05-04
Add NullableTypeDeclarationForDefaultNullValueFixer

### LINKED ISSUE #3883: Add a rule to force-add or force-remove `?` on nullable types

HypeMC:
Resolves #3883

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1226,6 +1226,17 @@ Choose from the list of available rules:
 
   Logical NOT operators (``!``) should have one trailing whitespace.
 
+* **nullable_type_declaration_for_default_null_value**
+
+  Adds or removes ``?`` before type declarations for parameters with a
+  default ``null`` value.
+
+  Configuration options:
+
+  - ``use_nullable_type_declaration`` (``bool``): whether to add or remove ``?``
+    before type declarations for parameters with a default ``null`` value;
+    defaults to ``true``
+
 * **object_operator_without_whitespace** [@Symfony, @PhpCsFixer]
 
   There should not be space before or after object ``T_OBJECT_OPERATOR``
--- a/src/Fixer/FunctionNotation/NullableTypeDeclarationForDefaultNullValueFixer.php
+++ b/src/Fixer/FunctionNotation/NullableTypeDeclarationForDefaultNullValueFixer.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\FunctionNotation;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\ArgumentAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\FunctionsAnalyzer;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author HypeMC
+ */
+final class NullableTypeDeclarationForDefaultNullValueFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Adds or removes `?` before type declarations for parameters with a default `null` value.',
+            [
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction sample(string \$str = null)\n{}\n",
+                    new VersionSpecification(70100)
+                ),
+                new VersionSpecificCodeSample(
+                    "<?php\nfunction sample(?string \$str = null)\n{}\n",
+                    new VersionSpecification(70100),
+                    ['use_nullable_type_declaration' => false]
+                ),
+            ],
+            'Rule is applied only in a PHP 7.1+ environment.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return \PHP_VERSION_ID >= 70100 && $tokens->isAllTokenKindsFound([T_FUNCTION, T_VARIABLE]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // Should run before NoUnreachableDefaultArgumentValueFixer
+        return 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('use_nullable_type_declaration', 'Whether to add or remove `?` before type declarations for parameters with a default `null` value.'))
+                ->setAllowedTypes(['bool'])
+                ->setDefault(true)
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $functionsAnalyzer = new FunctionsAnalyzer();
+
+        for ($index = $tokens->count() - 1; $index >= 0; --$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind(T_FUNCTION)) {
+                continue;
+            }
+
+            $arguments = $functionsAnalyzer->getFunctionArguments($tokens, $index);
+
+            $this->fixFunctionParameters($tokens, $arguments);
+        }
+    }
+
+    /**
+     * @param Tokens             $tokens
+     * @param ArgumentAnalysis[] $arguments
+     */
+    private function fixFunctionParameters(Tokens $tokens, array $arguments)
+    {
+        foreach (array_reverse($arguments) as $argumentInfo) {
+            // If the parameter doesn't have a type declaration or a default value null we can continue
+            if (
+                !$argumentInfo->hasTypeAnalysis()
+                ||
+                'null' !== $argumentInfo->getDefault()
+            ) {
+                continue;
+            }
+
+            $argumentTypeInfo = $argumentInfo->getTypeAnalysis();
+            if (true === $this->configuration['use_nullable_type_declaration']) {
+                if (!$argumentTypeInfo->isNullable()) {
+                    $tokens->insertAt($argumentTypeInfo->getStartIndex(), new Token([CT::T_NULLABLE_TYPE, '?']));
+                }
+            } else {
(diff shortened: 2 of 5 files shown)


src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($iter);
+
+            // Check whether the parameter has a type hint
+            if (!$tokens[$prevMeaningfulTokenIndex]->isGivenKind([CT::T_ARRAY_TYPEHINT, T_CALLABLE, T_STRING])) {
+                continue;
+            }
+
+            // Class namespaces are tokenized into multiple tokes, we need to reach the first one
```
phansys:
```suggestion
            // Class namespaces are tokenized into multiple tokens, we need to reach the first one
```

HypeMC: @phansys Done

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
```diff
@@ -0,0 +1,240 @@
...
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\FunctionNotation;
+
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @requires PHP 7.1
```
julienfalque: The tests should run on PHP lower than 7.1 to show that none of the test cases are changed on PHP 5.6 or 7.0, including cases that are changed on PHP 7.1+.

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
```diff
@@ -0,0 +1,171 @@
...
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author HypeMC
+ */
+final class NullableTypeHintForDefaultNullValueFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
```
julienfalque: I would rename the class to `NullableTypeDeclarationForDefaultNullValueFixer`.

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
```diff
@@ -0,0 +1,240 @@
...
+        yield ['<?php function foo(?\Bar\Baz & $param = null) {}'];
+        yield ['<?php function foo(?array &$param = null) {}'];
+        yield ['<?php function foo(?array & $param = null) {}'];
+        yield ['<?php function foo(?callable &$param = null) {}'];
+        yield ['<?php function foo(?callable & $param = null) {}'];
+        yield ['<?php $foo = function (?Baz $param = null) {};'];
+        yield ['<?php $foo = function (?Baz &$param = null) {};'];
+        yield ['<?php $foo = function (?Baz & $param = null) {};'];
+        yield ['<?php class Test { public function foo(?Bar\Baz $param = null) {} }'];
+        yield ['<?php class Test { public function foo(?self $param = null) {} }'];
```
julienfalque: Please add test cases with comments and/or whitespaces between `?` and the type.

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
```diff
@@ -0,0 +1,240 @@
...
+            '<?php function foo(?iterable $param = null) {}',
+        ];
+        yield [
+            '<?php $foo = function(array $a = null,
+                    array $b = null, array     $c = null, array
+                    $d = null) {};',
+            '<?php $foo = function(?array $a = null,
+                    ?array $b = null, ?array     $c = null, ?array
+                    $d = null) {};',
+        ];
```
julienfalque: Please add test cases with comments and/or whitespaces between `?` and the type.

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
```diff
@@ -0,0 +1,171 @@
...
+        return -1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('use_nullable_type_hint', 'Whether to add or remove `?` before type hints for parameters with a default `null` value.'))
```
julienfalque: I would rename the option to `use_nullable_type_declaration`.

In the description, _type hints_ should be _type declarations_.

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
```diff
@@ -0,0 +1,171 @@
...
+            $prevMeaningfulTokenIndex = $tokens->getPrevMeaningfulToken($iter);
+
+            // We've reached the opening braces so we're done
+            if ($prevMeaningfulTokenIndex <= $startParenthesisIndex) {
+                break;
+            }
+
+            // If the parameter doesn't have a default value null we can continue
+            if (
+                !$tokens[$tokens->getNextMeaningfulToken($variableTokenIndex + 1)]->equals('=')
```
julienfalque: I think adding 1 to the index will fail if there is no whitespace between the variable token and `=`. Same with adding 2 to find `null` in the next condition (e.g. `string $foo=null`). Please add test cases for those scenarios.

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
HypeMC: @julienfalque Nullable type hints are not supported on lower php versions, a parse error is thrown https://3v4l.org/jPGvV .

julienfalque: Yes, test cases that remove the `?` should not run on PHP 7.1. However test cases that add it on PHP 7.1+ should run on PHP 5.6 and 7.0 but should expect no change.

HypeMC:
@julienfalque All the requested changes have been made (except the one with the tests being run on older php versions) & can be viewed in this commit https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4401/commits/7143e6d009e2c316e0d376a017269eef15401dbe.

Since the fixer's `isCandidate` method has the following condition, `\PHP_VERSION_ID >= 70100`, running the tests on older php versions results in the test failing with the messegae `Fixer must be a candidate for input code.`, eg https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/563275126. I don't see how the test can be run on older php versions, any help would be appreciated.

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
```diff
@@ -0,0 +1,171 @@
...
+ */
+final class NullableTypeHintForDefaultNullValueFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Adds or removes `?` before type hints for parameters with a default `null` value.',
```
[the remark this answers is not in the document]
HypeMC: Done

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
HypeMC: Done

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
HypeMC: Done

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
HypeMC: Done

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
HypeMC: Done

src/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixer.php
HypeMC: Yep, you were right,these cases did fail, fixed now.

tests/Fixer/FunctionNotation/NullableTypeHintForDefaultNullValueFixerTest.php
HypeMC: As mentioned below, not sure how to do this one.

julienfalque:
For test cases that expect no change, the arguments passed to `doTest()` method change: parameter 1 is the input _and_ output code, and parameter 2 is null.

The error you get means parameter 2 is not null and thus treated as input code to be fixed, which requires the `isCandidate()` method to return `true`.

I would either:
- remove the `@requires` annotation on the test method and conditionally pass `null` as parameter 2 to `doTest()` (when PHP version is lower than 7.1)
- add a specific test that runs on PHP lower than 7.1 and always pass `null` as parameter 2 to `doTest()`.

julienfalque:
@HypeMC looking good :+1: 

As this is a new feature, could you please edit the PR to target branch `master` and rebase your work on it?

HypeMC:
@julienfalque Done.

SpacePossum:
Nice MR, great idea :)

The analyzing of the arguments in the fixer is very close to what the [ArgumentsAnalyzer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.15.1/src/Tokenizer/Analyzer/ArgumentsAnalyzer.php) does, could we utilize that one (maybe together with the FunctionAnalyzer if that helps)?

HypeMC:
@SpacePossum Done, I've used the `FunctionAnalyzer`.

drupol:
@HypeMC Rebase against latest master ?

HypeMC:
@drupol Done.

drupol:
Good, nice feature :-)

HypeMC:
@kubawerlos Tnx for the test, however, there seem to be some problems with the priority. The `NoUnreachableDefaultArgumentValueFixer` has a higher priority the `NullableTypeDeclarationForDefaultNullValueFixer`, but that priority is needed cause of the `CompactNullableTypehintFixer`. Not sure how to go about this one.

HypeMC:
@kubawerlos The only way I was able to make this work was by removing the dependency to the `CompactNullableTypehintFixer` and adding the white space removal to this fixer instead (as you suggested) https://github.com/HypeMC/PHP-CS-Fixer/pull/2

HypeMC:
@kubawerlos Any news on this one, are you ok with the changes suggested in https://github.com/HypeMC/PHP-CS-Fixer/pull/2 ? :smile:

kubawerlos:
@HypeMC please do not `push --force` as it hides what is new and what changes since one gave approve. The person who will do the merge will squash commits anyway.

SpacePossum:
Thank you @HypeMC.

PR #4404, opened 2019-05-06
FileLintingIterator - fix current value on end/invalid

SpacePossum:
Current implementation keeps the last value and does not reset the value to `null` at the end of an iteration or when an invalid value has been encountered.

### DIFF
--- a/src/Runner/FileLintingIterator.php
+++ b/src/Runner/FileLintingIterator.php
@@ -28,7 +28,7 @@ final class FileLintingIterator extends \IteratorIterator
     private $currentResult;
 
     /**
-     * @var LinterInterface
+     * @var null|LinterInterface
      */
     private $linter;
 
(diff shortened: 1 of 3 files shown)


src/Runner/FileLintingIterator.php
```diff
@@ -48,18 +48,14 @@ public function next()
     {
         parent::next();
 
-        if ($this->valid()) {
-            $this->currentResult = $this->handleItem($this->current());
-        }
+        $this->currentResult = $this->valid() ? $this->handleItem($this->current()) : null;
```
keradus: what issue you experienced without setting this value to `null` if position is not valid?
also, looks like typehint of `$currentResult` property is wrong

SpacePossum: >what issue you experienced without setting this value to null if position is not valid?

when not valid getting the current value returns the previous valid value, which is unexpected

>also, looks like typehint of $currentResult property is wrong

I forgot to update to `|null` indeed

PR #4407, opened 2019-05-08
Move test classmaps to autoload-dev

szepeviktor:


szepeviktor:
It was there for a reason.

Issue #4279, opened 2019-01-24
Allow passing multiple filenames without a configuration


jdufresne:
When reporting an issue (bug) please provide the following information:

#### The PHP version you are using (`$ php -v`):
```
$ php --version
PHP 7.2.14 (cli) (built: Jan  8 2019 09:59:17) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.6.1, Copyright (c) 2002-2018, by Derick Rethans
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
```
$ ./vendor/bin/php-cs-fixer -V
PHP CS Fixer 2.13.1 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski
```

#### The command you use to run PHP CS Fixer:
```
$ ./vendor/bin/php-cs-fixer fix path/to/file1.php path/to/file2.php
In ConfigurationResolver.php line 578:
                                                    
  For multiple paths config parameter is required.  
```

#### The configuration file you are using, if any:
None!

---

I've read through the rationale as to why this occurs here:

- https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/1903#discussion_r114061746
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/2390#issuecomment-266567608

I'm asking if the project can reconsider this decision and allow multiple paths without a configuration file. Typically, POSIX CLI tools are built to allow an arbitrary number of paths. If you take a look at other static analysis and code formatting tools, you'll see this is very common. Standard POSIX tools like `xargs` are built with this in mind.

My project chooses not to use a PHP-CS-Fixer configuration file. The goal is to have minimal fuss w/r/t code formatting and instead accept community defaults. As a result, I get the above error when using php-cs-fixer with multiple files.

This became a particular pain point when integrating with [pre-commit](https://github.com/pre-commit/pre-commit). pre-commit is built to pass all changed files to a tool in a single command. Requests to allow a "one at a time feature" have been [rejected](https://github.com/pre-commit/pre-commit/issues/771) with the response "the best way to improve the community is to improve the specific linter to take filenames in a more standard way." So here I am.

[...]

PR #4408, opened 2019-05-09
POC Add support for passing multiple files to fix when rules has been giv…

SpacePossum:
…en as well.

@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4279#issuecomment-490362053

### DIFF
--- a/tests/Console/Command/FixCommandTest.php
+++ b/tests/Console/Command/FixCommandTest.php
@@ -31,6 +31,7 @@ final class FixCommandTest extends TestCase
         $this->expectException(
             \PhpCsFixer\ConfigurationException\InvalidConfigurationException::class
         );
+
         $this->expectExceptionMessageRegExp(
             '#^Empty rules value is not allowed\.$#'
         );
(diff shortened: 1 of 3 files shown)


tests/Console/Command/FixCommandTest.php
```diff
@@ -56,6 +57,25 @@ public function testEmptyFormatValue()
...
+                    __FILE__,
+                    __DIR__.'/DescribeCommandTest.php',
+                ],
+                '--rules' => 'switch_case_semicolon_to_colon',
+            ]
+        );
+
+        $output = $cmdTester->getDisplay(true);
+
+        static::assertNotFalse(strpos($output, 'Loaded config default.'));
```
keradus: it's not obvious why loading 2 unrelated things (fixing multiple files + passing rules by cli param) is making tool to ignore the config file.

`php-cs-fixer fix src/path1 src/path2 --rules=foo` and boom, whole `.php_cs.dist` got ignored

jasonmccreary: Are you claiming `php-cs-fixer fix src/path1 src/path2 --rules=foo` runs currently? Cause it does not.

keradus: i'm not claiming. i'm saying.
after change proposed with this PR, running given command would ignore configuration file totally, which is wrong.

SpacePossum: > running given command would ignore configuration file totally, which is wrong.

given your specs, I think it is totally fine

SpacePossum:
ok

tests/Console/Command/FixCommandTest.php
keradus: if we ignore the config, we shall at least give user the note about it.
i don't want any implicit actions simply.

maybe let's have the `--config=default` or `--no-config` to allow for such cases?

SpacePossum: `Passing both `config` and `rules` options is not possible. `

It is already implicit which config file will be used when you use `rules` because you are not allowed to pass one explicit. So making it explicit the default config with the commandline arguments override seems best to me.

keradus: yeah, i guess it was a preparation for killing --rules in general

PR #4411, opened 2019-05-10
Allow multiple paths with rules

jasonmccreary:
This will open the restriction when passing multiple paths from requiring the `config` parameter to allow the `rules` parameter to be passed instead.

This was opened based on multiple previous discussions, notable issue [#4279](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4279#issuecomment-490224873).

### DIFF
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -577,8 +577,8 @@ final class ConfigurationResolver
 
         if ($this->isStdIn() || 0 === \count($path)) {
             $configDir = $this->cwd;
-        } elseif (1 < \count($path)) {
-            throw new InvalidConfigurationException('For multiple paths config parameter is required.');
+        } elseif (1 < \count($path) && null === $this->options['rules']) {
+            throw new InvalidConfigurationException('For multiple paths the config or rules parameter is required.');
         } elseif (is_file($path[0]) && $dirName = pathinfo($path[0], PATHINFO_DIRNAME)) {
             $configDir = $dirName;
         } else {
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -296,16 +296,30 @@ final class ConfigurationResolverTest extends TestCase
         $resolver->getReporter();
     }
 
-    public function testResolveConfigFileChooseFileWithPathArrayWithoutConfig()
+    public function testResolveConfigFileChooseFileWithPathArrayWithoutConfigOrRules()
     {
         $this->expectException(InvalidConfigurationException::class);
-        $this->expectExceptionMessageRegExp('/^For multiple paths config parameter is required\.$/');
+        $this->expectExceptionMessageRegExp('/^For multiple paths the config or rules parameter is required\.$/');
 
         $dirBase = $this->getFixtureDir();
 
         $resolver = $this->createConfigurationResolver(['path' => [$dirBase.'case_1/.php_cs.dist', $dirBase.'case_1/foo.php']]);
 
         $resolver->getConfig();
+    }
+
+    public function testResolveConfigFileChooseFileWithPathArrayAndRules()
+    {
+        $dirBase = $this->getFixtureDir();
+
+        $resolver = $this->createConfigurationResolver([
+            'rules' => '@PSR2',
+            'path' => [$dirBase.'case_1/.php_cs.dist', $dirBase.'case_1/foo.php']
+        ]);
+
+        $resolver->getConfig();
+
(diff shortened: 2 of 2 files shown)


jasonmccreary:
I am not familiar with the codebase, so although it works, this check may be too low level. Glad to move it and adjust accordingly if a core member wants to 👉 me in the right direction.

SpacePossum:
Thanks for the PR!

It is a duplicate of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4408 , I've no preference which way the we go at this point :)

jasonmccreary:
Build failures appear to be unrelated to my changes. Glad to help resolve, but seems it's at the project level.

jasonmccreary:
@SpacePossum just noticed that. May want to combine them though as I have an updated exception message + a previous test correct, yet keep your implementation of the new feature.

jasonmccreary:
@SpacePossum don't mean to be a pain, anyway we could get this merged? Glad to help combine the two PRs. Just tell me what you want and I'll make the change.

SpacePossum:
As soon as we get an OK from another member on the direction of this I'm happy to merge the two PR's and continue on this :)

tests/Console/ConfigurationResolverTest.php
```diff
@@ -317,6 +331,8 @@ public function testResolveConfigFileChooseFileWithPathArrayAndConfig()
             'path' => [$dirBase.'case_1/.php_cs.dist', $dirBase.'case_1/foo.php'],
         ]);
 
+        $resolver->getConfig();
```
keradus: why?

src/Console/ConfigurationResolver.php
```diff
@@ -577,8 +577,8 @@ private function computeConfigFiles()
 
         if ($this->isStdIn() || 0 === \count($path)) {
             $configDir = $this->cwd;
-        } elseif (1 < \count($path)) {
-            throw new InvalidConfigurationException('For multiple paths config parameter is required.');
+        } elseif (1 < \count($path) && null === $this->options['rules']) {
+            throw new InvalidConfigurationException('For multiple paths the config or rules parameter is required.');
```
keradus: so now, when `--rules` were passed, config will be taken from `$path[0]` implicitly? this is not good, if provided multiple paths for fixing will be different projects.

`/my/user/profile: $ php-cs-fixer fix projectA projectB`, now, `projectB` will be fixed using configuration of `projectA`
requiring an config file explicitly is removing that wrong usage of the tool, as person calling the tool knows that he is applying one config for multiple paths.

( @SpacePossum ref #4408 - using `defaultConfig` instead of one from `$path[]` is not better )

src/Console/ConfigurationResolver.php
```diff
@@ -577,8 +577,8 @@ private function computeConfigFiles()
 
         if ($this->isStdIn() || 0 === \count($path)) {
             $configDir = $this->cwd;
-        } elseif (1 < \count($path)) {
-            throw new InvalidConfigurationException('For multiple paths config parameter is required.');
+        } elseif (1 < \count($path) && null === $this->options['rules']) {
```
keradus: using `--path-mode=intersection` will not work properly after this change, as it will intersect with implicit `$path[0]` or default config

jasonmccreary: Should probably add a test for that as a contributor such as myself would have no way to tell that implicit overlap.

keradus: not having explicit config to be used creates a lot of issues, instead of covering each of non-working variants with test confirming it doesn't work, we forbidden to not provide the explicit config when problematic params combination was provided.
saying that, i'm always happy to merge more good tests.

jasonmccreary:
No problem. As I said originally, wasn't familiar with the codebase so I needed some direction. Nonetheless, I think there's a use case here to be covered. But no worries, I have been using my fork for a few weeks now anyway.

keradus:
if the use case is #4279, then i believe the solution is there already.

it's easy to build sth for concrete usage scenario only, it's not so easy when you build sth that should work for more users with different usage scenarios

jasonmccreary:
What's the solution?

keradus:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4279#issuecomment-496496696

jasonmccreary:
You may be over-simplifying the use case. I don't want to be forced to create a configuration file always. It would be nice to apply a very specific rule (or set of rules, like `@PSR2` to a project) to a set of directories as noted in the [original comment](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4279#issuecomment-490224873).

Again, I may be missing some combination of options. However, I have yet to see a one-line solution for the use case above.

PR #4412, opened 2019-05-14
PHP 7.4 - Tests for support

SpacePossum:


keradus:
should we do a meta ticket on what we should support exactly ?

PR #4413, opened 2019-05-14
Add AbstractConfigTestCase

Slamdunk:
This test enforces a project to have all the fixers specified, either active or not, so it's always up to date. This can be viewed as a more aggressive form of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3047.

Usage:
```php
final class ConfigTest extends \PhpCsFixer\Test\AbstractConfigTestCase
{
    public function testPhpCsFixerConfigIsUpToDate()
    {
        $this->doTestAllDefaultRulesAreSpecified(require dirname(__DIR__) . '/.php_cs');
    }
}
```

- [x] Check all rules are specified, either by name or RuleSet
- [x] Check all specified rules exist and are not deprecated
- [x] Check all rules are ordered alphabetically

It's 2 years I use this test in my project (for example in [Slamdunk/php-cs-fixer-extensions](https://github.com/Slamdunk/php-cs-fixer-extensions/blob/master/tests/ConfigTest.php)) and is sooo useful.

If you like it, I'll update the README too.

### DIFF
--- a/src/Test/AbstractConfigTestCase.php
+++ b/src/Test/AbstractConfigTestCase.php
@@ -57,14 +57,14 @@ abstract class AbstractConfigTestCase extends TestCase
         $diff = array_diff($currentRules, $availableRules);
         static::assertEmpty($diff, sprintf("The following fixers are specified but non existing or deprecated:\n- %s", implode(\PHP_EOL.'- ', $diff)));
 
-        $currentSets = array_values(array_filter(array_keys($configRules), static function (string $fixerName): bool {
+        $currentSets = array_values(array_filter(array_keys($configRules), static function ($fixerName) {
             return isset($fixerName[0]) && '@' === $fixerName[0];
         }));
         $defaultSets = $ruleSet->getSetDefinitionNames();
         $intersectSets = array_values(array_intersect($defaultSets, $currentSets));
         static::assertSame($intersectSets, $currentSets, sprintf('Rule sets must be ordered as the appear in %s', RuleSet::class));
 
-        $currentRules = array_values(array_filter(array_keys($configRules), static function (string $fixerName): bool {
+        $currentRules = array_values(array_filter(array_keys($configRules), static function ($fixerName) {
             return isset($fixerName[0]) && '@' !== $fixerName[0];
         }));
         $orderedCurrentRules = $currentRules;
--- a/src/Test/AbstractConfigTestCase.php
+++ b/src/Test/AbstractConfigTestCase.php
@@ -43,10 +43,10 @@ abstract class AbstractConfigTestCase extends TestCase
         $fixerFactory->registerCustomFixers($config->getCustomFixers());
         $fixers = $fixerFactory->getFixers();
 
-        $availableRules = array_filter($fixers, function (FixerInterface $fixer) {
+        $availableRules = array_filter($fixers, static function (FixerInterface $fixer) {
             return !$fixer instanceof DeprecatedFixerInterface;
         });
-        $availableRules = array_map(function (FixerInterface $fixer) {
+        $availableRules = array_map(static function (FixerInterface $fixer) {
             return $fixer->getName();
         }, $availableRules);
         sort($availableRules);
@@ -57,7 +57,16 @@ abstract class AbstractConfigTestCase extends TestCase
         $diff = array_diff($currentRules, $availableRules);
         static::assertEmpty($diff, sprintf("The following fixers are specified but non existing or deprecated:\n- %s", implode(\PHP_EOL.'- ', $diff)));
 
-        $currentRules = array_keys($configRules);
+        $currentSets = array_values(array_filter(array_keys($configRules), static function (string $fixerName): bool {
+            return isset($fixerName[0]) && '@' === $fixerName[0];
+        }));
+        $defaultSets = $ruleSet->getSetDefinitionNames();
+        $intersectSets = array_values(array_intersect($defaultSets, $currentSets));
+        static::assertSame($intersectSets, $currentSets, sprintf('Rule sets must be ordered as the appear in %s', RuleSet::class));
+
+        $currentRules = array_values(array_filter(array_keys($configRules), static function (string $fixerName): bool {
+            return isset($fixerName[0]) && '@' !== $fixerName[0];
+        }));
         $orderedCurrentRules = $currentRules;
         sort($orderedCurrentRules);
         static::assertSame($orderedCurrentRules, $currentRules, 'Fixers must be alphabetically ordered');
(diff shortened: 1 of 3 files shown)


Slamdunk:
```
1) PhpCsFixer\Tests\AutoReview\ComposerTest::testBranchAlias
Failed asserting that Array &0 (
    'dev-master' => '2.15-dev'
) is identical to Array &0 (
    'dev-master' => '2.16-dev'
).
```
immagine

src/Test/AbstractConfigTestCase.php
```diff
@@ -0,0 +1,74 @@
...
+use PhpCsFixer\ConfigInterface;
+use PhpCsFixer\Fixer\DeprecatedFixerInterface;
+use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\FixerFactory;
+use PhpCsFixer\RuleSet;
+use PHPUnit\Framework\TestCase;
+
+abstract class AbstractConfigTestCase extends TestCase
+{
+    final protected function doTestAllDefaultRulesAreSpecified(ConfigInterface $config)
```
keradus: what do you mean by `specified` ?

src/Test/AbstractConfigTestCase.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $currentRules = array_keys($rules);
+        sort($currentRules);
+
+        $fixerFactory = new FixerFactory();
+        $fixerFactory->registerBuiltInFixers();
+        $fixerFactory->registerCustomFixers($config->getCustomFixers());
+        $fixers = $fixerFactory->getFixers();
+
+        $availableRules = array_filter($fixers, static function (FixerInterface $fixer) {
```
keradus: those are availableFixers, only after ->getName() we get availableRules

src/Test/AbstractConfigTestCase.php
keradus: to be honest i got lost what this method does

tests/Test/AbstractConfigTestCaseTest.php
```diff
@@ -0,0 +1,152 @@
...
+use PhpCsFixer\RuleSet;
+use PhpCsFixer\Test\AbstractConfigTestCase;
+use PHPUnit\Framework\ExpectationFailedException;
+
+/**
+ * @internal
+ *
+ * @covers \PhpCsFixer\Test\AbstractConfigTestCase
+ */
+final class AbstractConfigTestCaseTest extends AbstractConfigTestCase
```
keradus: so, this class is testing `AbstractConfigTestCase` helper, and same time it inherits from it ?

tests/Test/AbstractConfigTestCaseTest.php
```diff
@@ -70,6 +71,27 @@ public function testRuleSetsAreHandled()
         }
     }
 
+    public function testRuleSetsMustBeOrderedAsTheyAppearInRuleSetClass()
```
keradus: ruleset in config must be ordered same as in ruleset class?
why??

Slamdunk: Set definitions are processed in the order they appear. For example, regarding the rule `visibility_required`, if you use:

```php
$config->setRules([
    '@PHP71Migration' => true,
    '@PSR2' => true,
]);
```
The `fix` run will **not** fix visibility for `const` because `@PSR2` resets the rule to the default value which doesn't fix constants, and this is likely not what the user expects.

I'll add the _why_ in the test name.

tests/Test/AbstractConfigTestCaseTest.php
Slamdunk: > so, this class is testing `AbstractConfigTestCase` helper, and same time it inherits from it ?

Exactly, pretty much like
https://github.com/sebastianbergmann/phpunit/blob/e4390fae89bf940f430604f6fd4b58804a530816/tests/unit/Framework/AssertTest.php#L27-L36

> shouldn't `AbstractConfigTestCase` be used by `ConfigTest`,

`ConfigTest` tests what a Config is and how a Config behaves, `AbstractConfigTestCase` tests how much a used and configured Config adhere to the standards where are establishing here.

>  same as `FooFixerTest extends AbstractFixerTestCase` ?

I don't get this suggestion: in this case the tested class would be `FooFixer`, but here I need to test (in this comparison) `AbstractFixerTestCase`.
Why? Because, unlike `AbstractFixerTestCase`, `AbstractConfigTestCase` will not be `@internal` as it is intended to be used by the final user, and for the same reason it's placed under the `src/` folder.

src/Test/AbstractConfigTestCase.php
Slamdunk: I mean that if you upgrade PHP-CS-Fixer from [2.14](https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.14.0) to [2.15](https://github.com/FriendsOfPHP/PHP-CS-Fixer/releases/tag/v2.15.0) without updating your `.php_cs`, the test will fail because the new [`php_unit_size_class`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3939) rule isn't configured: to pass the test you need to set it either to false or to whatever the rule allows you.

Maybe I should change the name of the function to `doTestAllDefaultRulesAreConfigured`, shouldn't I?

src/Test/AbstractConfigTestCase.php
Slamdunk: Right, thanks

src/Test/AbstractConfigTestCase.php
```diff
@@ -0,0 +1,74 @@
...
+
+        // RuleSet strips all disabled rules
+        foreach ($configRules as $name => $value) {
+            if ('@' === $name[0]) {
+                continue;
+            }
+            $rules[$name] = $value;
+        }
+
+        $currentRules = array_keys($rules);
```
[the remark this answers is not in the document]
Slamdunk: Sorry, I've redacted the variable name as a more meaningful one

src/Test/AbstractConfigTestCase.php
Slamdunk: I'm sorry: I've entirely refactored the class extracting all the methods I could, now it should be straightforward to follow its logic.

kubawerlos: `doTestAllBuiltInRulesAreConfigured`?

Slamdunk: I like it, done :+1:

SpacePossum:
near all checks have been added during the refactor of the RuleSet and RuleSets, so others I added here https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5930 , I think this covers all, let me know if not than we can reopen, thanks for all the work!

Slamdunk:
@SpacePossum does this mean that `3.0` dist version has a PHPUnit test that I can use withing my project that accomplish the goal of this PR?

SpacePossum:
We no longer provide any tests that can be extended on, or at least we don't hold a BC promise on any test logic AFAIK. This was removed in 3.0. There was/is to much work to maintain by the team so we had to reduce the scope of the project and removed the support on extendable tests.

PR #4416, opened 2019-05-21
NoUnusedImports - Fix imports detected as used in namespaces

### LINKED ISSUE #4410: no_unused_imports not working correctly
It seems like `php-cs-fixer` do not remove all unused imports.
I have provided a class (Laravel Form Request) as an example.
I'm importing 3 classes (Laravel Models), `User`, `Role`, `App`.
When running `php-cs-fixer` command listed below, it only removes `Role`, but leaves `User` and `App` even though they're not being used either.

---

For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

=> 7.2.17

=> 2.14.0

=> php-cs-fixer fix app/Admin/Http/Requests/Api/SubscribeRequest.php

```php
<?php

    ->exclude('tests/Fixtures')
    ->in(__DIR__);

    ->setRiskyAllowed(true)
    ->setRules([
        '@PSR2' => true,
        'array_indentation' => true,
        'trailing_comma_in_multiline_array' => true,
        'method_chaining_indentation' => true,
        'binary_operator_spaces' => true,
        'strict_comparison' => true,
        'standardize_not_equals' => true,
        'strict_param' => true,
        'concat_space' => ['spacing' => 'one'],
        'no_unused_imports' => true,
        'no_useless_else' => true,
        'blank_line_before_statement' => true,
        'whitespace_after_comma_in_array' => true,
    ])
    ->setFinder($finder);

return $config;
```

```php
<?php

namespace App\Admin\Http\Requests\Api;

use App\Models\User;
use App\Models\Role;
use App\Models\App;
use App\Http\Requests\Request;

class SubscribeRequest extends Request
{
    /**
     * Determine if the user is authorized to make this request.
     *
     * @return bool
     */
    public function authorize()
    {
        return true;
    }

    /**
     * Get the validation rules that apply to the request.
     *
     * @return array
     */
    public function rules()
    {
        //
    }
}
```

```php
<?php

namespace App\Admin\Http\Requests\Api;

use App\Models\User;
use App\Models\App;
use App\Http\Requests\Request;

class SubscribeRequest extends Request
{
    /**
     * Determine if the user is authorized to make this request.
     *
     * @return bool
     */
    public function authorize()
    {
        return true;
    }

    /**
     * Get the validation rules that apply to the request.
     *
     * @return array
     */
    public function rules()
    {
        //
    }
}
```

```php
<?php

namespace App\Admin\Http\Requests\Api;

use App\Http\Requests\Request;

class SubscribeRequest extends Request
{
    /**
     * Determine if the user is authorized to make this request.
     *
     * @return bool
     */
    public function authorize()
    {
        return true;
    }

    /**
     * Get the validation rules that apply to the request.
     *
     * @return array
     */
    public function rules()
    {
        //
    }
}
```

julienfalque:
Partially fixes #4410.

### DIFF
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -126,6 +126,15 @@ final class NoUnusedImportsFixer extends AbstractFixer
             $token = $tokens[$index];
 
             if (
+                $token->isGivenKind(T_NAMESPACE)
+                && $tokens[$tokens->getNextMeaningfulToken($index)]->isGivenKind(T_STRING)
+            ) {
+                $index = $tokens->getNextTokenOfKind($index, [';']);
+
+                continue;
+            }
+
+            if (
                 $token->isGivenKind(T_STRING)
                 && 0 === strcasecmp($shortName, $token->getContent())
                 && !$tokens[$tokens->getPrevMeaningfulToken($index)]->isGivenKind([T_NS_SEPARATOR, T_CONST, T_OBJECT_OPERATOR])
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -114,9 +114,18 @@ final class NoUnusedImportsFixer extends AbstractFixer
         }
     }
 
+    /**
+     * @param Tokens            $tokens
+     * @param NamespaceAnalysis $namespace
+     * @param array<int, int>   $ignoredIndexes
+     * @param string            $shortName
+     *
+     * @return bool
+     */
     private function importIsUsed(Tokens $tokens, NamespaceAnalysis $namespace, array $ignoredIndexes, $shortName)
     {
-        for ($index = $namespace->getScopeStartIndex(); $index <= $namespace->getScopeEndIndex(); ++$index) {
+        $namespaceEndIndex = $namespace->getScopeEndIndex();
+        for ($index = $namespace->getScopeStartIndex(); $index <= $namespaceEndIndex; ++$index) {
             if (isset($ignoredIndexes[$index])) {
                 $index = $ignoredIndexes[$index];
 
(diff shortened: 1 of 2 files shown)


SpacePossum:
looking good, please check https://github.com/julienfalque/PHP-CS-Fixer/pull/8 for one edge case

julienfalque:
Thank you @SpacePossum :)

erickskrauch:
Got the same bug. Is it possible to merge this fix?

src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -114,9 +114,18 @@ function (NamespaceUseAnalysis $useDeclaration) use ($namespace) {
...
 
+    /**
+     * @param Tokens            $tokens
+     * @param NamespaceAnalysis $namespace
+     * @param array<int, int>   $ignoredIndexes
+     * @param string            $shortName
+     *
+     * @return bool
+     */
     private function importIsUsed(Tokens $tokens, NamespaceAnalysis $namespace, array $ignoredIndexes, $shortName)
```
kubawerlos: Can we rename this method to `isImportUsed`?

src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -114,9 +114,18 @@ function (NamespaceUseAnalysis $useDeclaration) use ($namespace) {
...
     {
-        for ($index = $namespace->getScopeStartIndex(); $index <= $namespace->getScopeEndIndex(); ++$index) {
+        $namespaceEndIndex = $namespace->getScopeEndIndex();
```
kubawerlos: Is this assignment making the code more clear? Personally I'd put `$namespace->getScopeEndIndex()` directly in `for`, so it "makes pair" with `getScopeStartIndex` in single line.

src/Fixer/Import/NoUnusedImportsFixer.php
julienfalque: I prefer the current name, it feels more natural.

src/Fixer/Import/NoUnusedImportsFixer.php
julienfalque: Actually I don't remember why I made this change. Maybe I wanted to reuse the variable but then I didn't. Or I wanted to avoid a method call on each `for` iteration. I can revert it if you think it's best.

src/Fixer/Import/NoUnusedImportsFixer.php
kubawerlos: How come it is more natural? The question "is import used?" is the right way to ask, not "import is used?". Also `is` at the beginning is the "classic" prefix for returning boolean.

julienfalque: I always try to think of methods as orders to give rather than questions to ask: _"Tell me whether this **import is used**"_ vs. _"**Is** this **import used**?"_.

src/Fixer/Import/NoUnusedImportsFixer.php
julienfalque: I would keep the variable to avoid calling `$namespace->getScopeStartIndex()` for each iteration so I mark this conversation as resolved. Let me know if you prefer reverting the change instead.

src/Fixer/Import/NoUnusedImportsFixer.php
kubawerlos: > I always try to think of methods as orders

Me too, when possible, yet this case is yes/no question after all.

julienfalque: Method renamed.

SpacePossum:
Thank you @julienfalque.

PR #4418, opened 2019-05-21
PhpUnitNamespacedFixer - properly translate classes which do not follow translation pattern

### LINKED ISSUE #4417: PHP Parse error @PHPUnit60Migration
When using the _@PHPUnit60Migration:riky_ ruleset (effectively the **PhpUnitNamespacedFixer**) PHP-CS-Fixer was not able to fix files due to resulting the fixes into a PHP Parse error:

> PHP Parse error: syntax error, unexpected 'Return' (T_RETURN), expecting identifier (T_STRING)

This is when changing the class **PHPUnit_Framework_MockObject_Stub_Return** into the new namespaced  class **PHPUnit\Framework\MockObject\Stub\Return**.

Changing the class-name string `PHPUnit_Framework_MockObject_Stub_Return` to `PHPUnit_Framework_MockObject_Stub_ReturnStub` results in the correct new type `PHPUnit\Framework\MockObject\Stub\ReturnStub`.

---

~~~
php7.1 -n -dextension=tokenizer.so -dextension=json.so -dextension=ctype.so -dextension=iconv.so \
  -f../PHP-CS-Fixer/php-cs-fixer -- fix -vvv \
  --rules=@PHPUnit60Migration:risky,-php_unit_mock \
  tests
~~~

ktomk:
Fixes #4417

When cs-fix the name-space of the class

    PHPUnit_Framework_MockObject_Stub_Return

it results in a new class name which causes a PHP parse error b/c of the
reserved word "return" as the new class-name.

Phpunit has renamed that class to ...\ReturnStub (5 -> 6).

Fix is to translate it to the "new" name so that it transposes correctly.

Fixer: PhpUnitNamespacedFixer

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
@@ -33,6 +33,18 @@ final class PhpUnitNamespacedFixer extends AbstractFixer implements Configuratio
     private $originalClassRegEx;
 
     /**
+     * Class Mappings.
+     *
+     *  * [original classname => new classname] Some classes which match the
+     *    original class regular expression do not have a same-compound name-
+     *    space class and need a dedicated translation table. This trans-
+     *    lation table is defined in @see configure.
+     *
+     * @var array|string[] Class Mappings
+     */
+    private $classMap;
+
+    /**
      * {@inheritdoc}
      */
     public function getDefinition()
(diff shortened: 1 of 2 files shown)


ktomk:
Another candidate that could benefit from the translation is:

* `PHPUnit_Framework_Error` -> `PHPUnit\Framework\Error\Error`

ktomk:
Added a full translation table, could only **not** find the following Phpunit 5 counterparts:

* PHPUnit_Extensions_TicketListener (-> PHPUnit\Extensions\TicketListener)
* PHPUnit_Extensions_PhptTestSuite (-> PHPUnit\Extensions\PhptTestSuite)
* PHPUnit_Extensions_TestDecorator (-> PHPUnit\Extensions\TestDecorator)
* PHPUnit_Extensions_RepeatedTest (-> PHPUnit\Extensions\RepeatedTest)
* PHPUnit_Extensions_GroupTestSuite (-> PHPUnit\Extensions\GroupTestSuite)
* PHPUnit_Util_Log_JSON (-> PHPUnit\Util\Log\JSON)
* PHPUnit_Util_Log_TAP (-> PHPUnit\Util\Log\TAP)
* PHPUnit_Util_String (-> PHPUnit\Util\String)
* PHPUnit_Util_PHPTest (-> PHPUnit\Util\PHPTest)

keradus:
Nice PR @ktomk !

src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
```diff
@@ -81,10 +86,41 @@ public function configure(array $configuration = null)
...
         if (PhpUnitTargetVersion::fulfills($this->configuration['target'], PhpUnitTargetVersion::VERSION_6_0)) {
             $this->originalClassRegEx = '/^PHPUnit_\w+$/i';
+            $this->originalClassTranslations = [
+                'PHPUnit_Extensions_PhptTestCase' => 'PHPUnit_Runner_PhptTestCase',
+                'PHPUnit_Framework_Constraint' => 'PHPUnit_Framework_Constraint_Constraint',
+                'PHPUnit_Framework_Constraint_StringMatches' => 'PHPUnit_Framework_Constraint_StringMatchesFormatDescription',
+                'PHPUnit_Framework_Constraint_JsonMatches_ErrorMessageProvider' => 'PHPUnit_Framework_Constraint_JsonMatchesErrorMessageProvider',
+                'PHPUnit_Framework_Constraint_PCREMatch' => 'PHPUnit_Framework_Constraint_RegularExpression',
+                'PHPUnit_Framework_Constraint_ExceptionMessageRegExp' => 'PHPUnit_Framework_Constraint_ExceptionMessageRegularExpression',
+                'PHPUnit_Framework_Constraint_And' => 'PHPUnit_Framework_Constraint_LogicalAnd',
```
keradus: if I get this right, `PHPUnit_Framework_Constraint_LogicalAnd`  was never existing. It was introduced alread with `\` (`PHPUnit\Framework\Constraint\LogicalAnd`).

for that, currently, your proposal:
- takes original name
- translates it to fake name that would follow transition pattern
- applies transition pattern

we can simplify it that this table can contain already ready to use name with `\`, and then:
- if original name exists in dictionary, use name from dictionary
- if not, apply transition pattern

keradus:
please, change target branch to 2.12 LTS line

src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
```diff
@@ -32,6 +32,11 @@ final class PhpUnitNamespacedFixer extends AbstractFixer implements Configuratio
      */
     private $originalClassRegEx;
 
+    /**
+     * @var array|string[]
+     */
+    private $originalClassTranslations;
```
keradus: please add reasoning for this array in description
like why we need to have it in first place. for now, it's hidden knowledge why some classes are put here and some are not

src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
ktomk: This is equally possible, yes. It would spare any fake names. This is straight to update, but needs more branching in the old translation function. It might be more clear however when looking at the bare data that the map is old => new. I'll change it to that, this can be part of the next review iteration.

ktomk:
About the tests, is there some boilerplate I can reuse? I have it functionally tested here as I run the conversion currently which is how I've detected the original error. So if you can point to some file with a link would be helpful to get me started with fixer tests.

ktomk:
@keradus I fully addressed your comments.

src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
ktomk: thanks for pointing to that, I removed the indirection and made it a more direct old => new class-map now and explained that it assists the regex which provides the foremost class detection and helps out for those cases where it matches but *"do[es] not have a same-compound namespace class and need a dedicated translation table."*. Also referencing *`configure()`* method which now has the array configuration below the regex nicely paired.

src/Fixer/PhpUnit/PhpUnitNamespacedFixer.php
ktomk: There are no more fake names, it's just old => new now, and new is already with namespace separators.

keradus:
Thank you @ktomk.

ktomk:
@keradus: You're welcome, and thank you for having me here. Let's see for the next Phpunit conversion :)

PR #4426, opened 2019-05-31
DX: handle deprecations of symfony/event-dispatcher:4.3

keradus:


src/Runner/FileFilterIterator.php
```diff
@@ -105,6 +105,15 @@ private function dispatchEvent($name, Event $event)
             return;
         }
 
-        $this->eventDispatcher->dispatch($name, $event);
+        // BC compatibility < Sf 4.3
+        if (
```
kubawerlos: Could we have the condition like [this](https://github.com/symfony/contracts/blob/v1.1.1/EventDispatcher/EventDispatcherInterface.php#L16) one?

keradus: not really, you can have `PsrEventDispatcherInterface` installed anyway while having Sf component v3

keradus: also, you may have sf component v3.3 with deprecation warning and not have `psr/event-dispatcher` installed

PR #4431, opened 2019-06-02
DX: test that default config is not passed in RuleSet

kubawerlos:


src/RuleSet.php
```diff
@@ -322,7 +320,7 @@ final class RuleSet implements RuleSetInterface
         ],
         '@PHPUnit50Migration:risky' => [
             '@PHPUnit48Migration:risky' => true,
-            'php_unit_dedicate_assert' => ['target' => PhpUnitTargetVersion::VERSION_5_0],
```
keradus: please revert
this is 50migration ruleset, and default target will not be 5_0 on 3.0

kubawerlos: How to modify test? Exclude all `@PHPUnitXXMigration` sets?

keradus: not really. need to think about best approach here

kubawerlos: So let's [ensure](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4433) the rule set are targeting proper PHPUnit's version.

SpacePossum:
Thank you @kubawerlos.

PR #4432, opened 2019-06-02
DX: remove empty sets from RuleSet

kubawerlos:
As requested by @keradus

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -19,7 +19,7 @@ $finder = PhpCsFixer\Finder::create()
 $config = PhpCsFixer\Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
-        '@PHP56Migration' => true,
+        '@PHP56Migration:risky' => true,
         '@PHPUnit60Migration:risky' => true,
         '@PhpCsFixer' => true,
         '@PhpCsFixer:risky' => true,
--- a/tests/RuleSetTest.php
+++ b/tests/RuleSetTest.php
@@ -30,7 +30,7 @@ final class RuleSetTest extends TestCase
 {
     public function testCreate()
     {
-        $ruleSet = RuleSet::create();
+        $ruleSet = new RuleSet();
 
         static::assertInstanceOf(\PhpCsFixer\RuleSet::class, $ruleSet);
     }
@@ -252,19 +252,9 @@ final class RuleSetTest extends TestCase
      */
     public function testSetDefinitionsAreSorted($setDefinitionName)
     {
-        $ruleSet = RuleSet::create();
+        $reflection = new AccessibleObject(new RuleSet());
 
-        $method = new \ReflectionMethod(
-            \PhpCsFixer\RuleSet::class,
-            'getSetDefinition'
-        );
-
-        $method->setAccessible(true);
-
-        $setDefinition = $method->invoke(
-            $ruleSet,
-            $setDefinitionName
-        );
+        $setDefinition = $reflection->getSetDefinition($setDefinitionName);
 
         $sortedSetDefinition = $setDefinition;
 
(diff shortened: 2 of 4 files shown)


keradus:
i had to request it sooo long ago ;) can't remember about this

tests/RuleSetTest.php
```diff
@@ -276,6 +276,33 @@ public function testSetDefinitionsAreSorted($setDefinitionName)
...
+    public function testSetDefinitionIsNotEmpty($setDefinitionName)
+    {
+        $ruleSet = RuleSet::create();
+
+        $method = new \ReflectionMethod(
+            \PhpCsFixer\RuleSet::class,
+            'getSetDefinition'
+        );
+
+        $method->setAccessible(true);
```
keradus: `AccessibleObject` perhaps? it's already used in this file

tests/RuleSetTest.php
```diff
@@ -276,6 +266,21 @@ public function testSetDefinitionsAreSorted($setDefinitionName)
...
     }
 
+    /**
+     * @dataProvider provideSetDefinitionNameCases
+     *
+     * @param string $setDefinitionName
+     */
+    public function testSetDefinitionIsNotEmpty($setDefinitionName)
+    {
+        $reflection = new AccessibleObject(RuleSet::create());
```
keradus: `new RuleSet()` ?

kubawerlos: Sure, done, any particular reason to change it here? I simply wonder why we even have this static constructor as it makes no difference comparing to normal creation.

keradus: leftover (kept for bc reasons) of old php versions, where you could do `Foo::create()->bar();`, but you couldn't do it when using `new`

.php_cs.dist
```diff
@@ -19,7 +19,7 @@ $finder = PhpCsFixer\Finder::create()
 $config = PhpCsFixer\Config::create()
     ->setRiskyAllowed(true)
     ->setRules([
-        '@PHP56Migration' => true,
```
keradus: please mention that ruleset was removed in UPGRADE-v3.md

.php_cs.dist
```diff
@@ -19,7 +19,7 @@ $finder = PhpCsFixer\Finder::create()
...
+        '@PHP56Migration:risky' => true,
```
keradus: we don't need it here, it's part of `@PHPUnit60Migration:risky` that we are using already

kubawerlos: Is it? I cannot see [it](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3.0/src/RuleSet.php#L358).

keradus: It is.

kubawerlos: How?
I see `PHP56Migration:risky` is part of `@PHP70Migration:risky`.
Then `@PHP70Migration:risky` is part of `@PHP71Migration:risky`.
And that's it, how it becomes a part of `@PHPUnit60Migration:risky`?

kubawerlos: @keradus take a look at: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4432/commits/7b65acacc2d589fcafe48e86e6310a8491b3fd63 - I have removed it and now it's OK to have `pow` function (which changing to `**` is part of `@PHP56Migration:risky` set).

SpacePossum: `PHPUnit60Migration:risky` has `@PHPUnit56Migration:risky` it does indeed not have `PHP56Migration:risky` IIUC

keradus: Aaah. Indeed, I read them both as phpunit rulesets. My bad, sorry

SpacePossum: no worries :)

SpacePossum:
Thank you @kubawerlos.

PR #4433, opened 2019-06-02
DX: test to ensure @PHPUnitMigration rule sets are correctly defined

kubawerlos:
This test will ensure that all @PHPUnitMigration rule sets have the rules defined with target that is largest possible to the rule set PHPUnit's version it applies.

### DIFF
--- a/tests/RuleSetTest.php
+++ b/tests/RuleSetTest.php
@@ -16,6 +16,8 @@ use PhpCsFixer\AccessibleObject\AccessibleObject;
 use PhpCsFixer\ConfigurationException\InvalidForEnvFixerConfigurationException;
 use PhpCsFixer\Fixer\ConfigurableFixerInterface;
 use PhpCsFixer\Fixer\DeprecatedFixerInterface;
+use PhpCsFixer\Fixer\FixerInterface;
+use PhpCsFixer\Fixer\PhpUnit\PhpUnitTargetVersion;
 use PhpCsFixer\FixerFactory;
 use PhpCsFixer\RuleSet;
 
@@ -516,6 +518,38 @@ final class RuleSetTest extends TestCase
         $ruleSet->getRuleConfiguration('_not_exists');
     }
 
+    /**
+     * @dataProvider providePHPUnitMigrationSetDefinitionNameCases
+     *
+     * @param string $setName
+     */
+    public function testPHPUnitMigrationTargetVersions($setName)
+    {
+        $ruleSet = new RuleSet([$setName => true]);
+
+        foreach ($ruleSet->getRules() as $ruleName => $ruleConfig) {
+            $targetVersion = true === $ruleConfig ? $this->getDefaultPHPUnitTargetOfRule($ruleName) : $ruleConfig['target'];
+
+            static::assertPHPUnitVersionIsLargestAllowed($setName, $ruleName, $targetVersion);
+        }
+    }
+
+    /**
+     * @return string[][]
+     */
+    public function providePHPUnitMigrationSetDefinitionNameCases()
+    {
+        $setDefinitionNames = RuleSet::create()->getSetDefinitionNames();
+
+        $setDefinitionPHPUnitMigrationNames = array_filter($setDefinitionNames, function ($setDefinitionName) {
+            return 1 === preg_match('/^@PHPUnit\d{2}Migration:risky$/', $setDefinitionName);
+        });
+
+        return array_map(static function ($setDefinitionName) {
+            return [$setDefinitionName];
+        }, $setDefinitionPHPUnitMigrationNames);
+    }
+
     private static function assertSameRules(array $expected, array $actual, $message = '')
     {
         ksort($expected);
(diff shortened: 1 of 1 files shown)


tests/RuleSetTest.php
```diff
@@ -516,6 +516,53 @@ public function testGetMissingRuleConfiguration()
...
+     * @dataProvider providePHPUnitMigrationSetDefinitionNameCases
+     *
+     * @param string $setName
+     */
+    public function testPHPUnitMigrationTargetVersions($setName)
+    {
+        $ruleSet = new RuleSet([$setName => true]);
+
+        foreach ($ruleSet->getRules() as $ruleName => $ruleConfig) {
+            if (true === $ruleConfig) {
```
keradus: it's not readable what we are doing it.
make a private helper?
`$targetVersion = true === $ruleConfig ? $this->getDefaultPHPUnitTargetOfRule($ruleName) : $ruleConfig;`

tests/RuleSetTest.php
```diff
@@ -607,4 +654,54 @@ private static function getRuleSetDefinitionsToTestWith()
...
+     * @param string $targetVersion
+     */
+    private static function assertVersionIsLargestAvailable($setName, $ruleName, $targetVersion)
+    {
+        $maximumPHPUnitVersion = preg_replace('/^@PHPUnit(\d)(\d)Migration:risky$/', '$1.$2', $setName);
+
+        $factory = new FixerFactory();
+        $factory->registerBuiltInFixers();
+        $factory->useRuleSet(new RuleSet([$ruleName => true]));
+        $fixer = current($factory->getFixers());
```
keradus: dry is violated. `getFixerByName(string $rule): FixerInterface` ?

tests/RuleSetTest.php
```diff
@@ -607,4 +654,54 @@ private static function getRuleSetDefinitionsToTestWith()
...
+        foreach ($fixer->getConfigurationDefinition()->getOptions() as $option) {
+            if ('target' === $option->getName()) {
+                $availableVersions = $option->getAllowedValues();
+                $availableVersions = array_diff($availableVersions, ['newest']);
+
+                break;
+            }
+        }
+
+        $availableVersions = array_filter(
```
keradus: `$availableVersions` will be empty if no `target` option was defined for fixer

tests/RuleSetTest.php
```diff
@@ -607,4 +654,54 @@ private static function getRuleSetDefinitionsToTestWith()
...
+
+        $factory = new FixerFactory();
+        $factory->registerBuiltInFixers();
+        $factory->useRuleSet(new RuleSet([$ruleName => true]));
+        $fixer = current($factory->getFixers());
+
+        foreach ($fixer->getConfigurationDefinition()->getOptions() as $option) {
+            if ('target' === $option->getName()) {
+                $availableVersions = $option->getAllowedValues();
+                $availableVersions = array_diff($availableVersions, ['newest']);
```
keradus:
```
src/Fixer/PhpUnit/PhpUnitTargetVersion.php:36:    const VERSION_NEWEST = 'newest';
```

tests/RuleSetTest.php
```diff
@@ -607,4 +654,54 @@ private static function getRuleSetDefinitionsToTestWith()
...
+            $fixer->getName(),
+            $setName,
+            $targetVersion,
+            implode('", "', $availableVersions)
+        ));
+
+        rsort($availableVersions);
+        $largestAvailableVersion = reset($availableVersions);
+
+        static::assertSame($largestAvailableVersion, $targetVersion, sprintf(
```
keradus: the method name is `assertVersionIsLargestAvailable`

are we really testing that version is largest from available ones?

`@PHPUnit50Migration:risky` has `php_unit_dedicate_assert: ['target' => '5.0']`
but `php_unit_dedicate_assert` can be configured for target 5.6.

yet, we don't want to assert that we use largest available target for this rule, as we don't want to target 5.6 in 5.0 ruleset

tests/RuleSetTest.php
kubawerlos: Luckily all the rules for PHPUnit migrations have `target` option, handling a fixer without it looks like YAGNI to me.

tests/RuleSetTest.php
kubawerlos: For `@PHPUnit50Migration:risky` version available are only up to `5.0` (according to migration's name), so 5.6 would not be available for this migration.

Maybe different naming, "acceptable" or something?

tests/RuleSetTest.php
keradus: hoping to find an item in collection and not handling scenario when it's not there looks like asking for troubles in the future.
1. nobody will remember that when one will add new phpunit rule that will not have target option, he has to go to ruleset test and adjust sth not related
2. when option will not be found, we will not return string but null in current implementation, leading to type violation of method prototype

tests/RuleSetTest.php
keradus: i got what your logic does after reading it 5 times, because the current method name is misleading - claiming it does sth else.

if you have no good name for assertion, at least describe what this assertion does.

tests/RuleSetTest.php
kubawerlos: Allright, I've added the checking.

tests/RuleSetTest.php
kubawerlos: I've renamed function and variable, hopefully to be more descriptive, what do you think about them now?

SpacePossum:
Thank you @kubawerlos.

PR #4435, opened 2019-06-03
SingleSpaceAfterConstructFixer - Introduction

### LINKED ISSUE #3981: php-cs-fixer does not fix double whitespace between return and the return value (Language constructs must be followed by a single space; expected 1 space but found "  ")
=> 7.0.10

=> 2.13.0

=> php-cs-fixer fix ./src/Helpers/DateTime.php --verbose --config=./.php_cs

```php
$finder = (whatever...)
    ->setUsingCache(false)
    ->setRules([
        '@Symfony' => true,
        '@PSR2' => true,
        'align_multiline_comment' => true,
        'array_indentation' => true,
        'combine_consecutive_issets' => true,
        'combine_consecutive_unsets' => true,
        'no_binary_string' => true,
        'no_superfluous_elseif' => true,
        'not_operator_with_successor_space' => true,
        'phpdoc_order' => true,
        'phpdoc_trim_consecutive_blank_line_separation' => true,
        'phpdoc_types_order' => true,
        'yoda_style'=> null,
    ])
    ->setFinder($finder)
;
```

```php
if (! function_exists('getFormattedDateTime')) {
    function getFormattedDateTime($dateTimeMysql, $newFormat, $locale = null)
    {
        if (empty($locale)) {
            $locale = currentLocale();
        }
        $dateTime = new \DateTime($dateTimeMysql);
        $format = new \IntlDateFormatter(trans('Core::datetime.locale', [], $locale), \IntlDateFormatter::FULL, \IntlDateFormatter::FULL);
        $format->setPattern($newFormat);

        return  $format->format($dateTime);
    }
}
```

```php
// same, no change
if (! function_exists('getFormattedDateTime')) {
    function getFormattedDateTime($dateTimeMysql, $newFormat, $locale = null)
    {
        if (empty($locale)) {
            $locale = currentLocale();
        }
        $dateTime = new \DateTime($dateTimeMysql);
        $format = new \IntlDateFormatter(trans('Core::datetime.locale', [], $locale), \IntlDateFormatter::FULL, \IntlDateFormatter::FULL);
        $format->setPattern($newFormat);

        return  $format->format($dateTime);
    }
}
```

```php
// one space between return and $format
if (! function_exists('getFormattedDateTime')) {
    function getFormattedDateTime($dateTimeMysql, $newFormat, $locale = null)
    {
        if (empty($locale)) {
            $locale = currentLocale();
        }
        $dateTime = new \DateTime($dateTimeMysql);
        $format = new \IntlDateFormatter(trans('Core::datetime.locale', [], $locale), \IntlDateFormatter::FULL, \IntlDateFormatter::FULL);
        $format->setPattern($newFormat);

        return $format->format($dateTime);
    }
}
```

image

localheinz:
This PR

* [x] introduces a `SingleSpaceAfterConstructFixer`
* [x] adds the `single_space_after_construct` fixer to the `@PhpCsFixer` rule set

Fixes #3981.

💁‍♂ Not sure about this one, there are a quite a bunch of fixers which are dealing with spaces themselves, what do you think?

### Handles

* [x] `abstract`
* [x] `as`
* [x] `break`
* [x] `case`
* [x] `catch`
* [x] `class` (when declaring classes)
* [x] `clone`
* [x] `const`
* [x] `const_import`
* [x] `continue`
* [x] `do`
* [x] `echo` 
* [x] `else` 
* [x] `elseif` 
* [x] `extends`
* [x] `final`
* [x] `finally`
* [x] `for`
* [x] `foreach`
* [x] `function`
* [x] `function_import`
* [x] `global`
* [x] `goto` 
* [x] `if`
* [x] `implements`
* [x] `include`
* [x] `include_once`
* [x] `instanceof`
* [x] `insteadof`
* [x] `interface`
* [x] `new`
* [x] `php_open`
* [x] `open_tag_with_echo`  
* [x] `print` 
* [x] `private`
* [x] `protected`
* [x] `public`
* [x] `require`
* [x] `require_once`
* [x] `return` 
* [x] `static` (when referring to variables, fields, functions, methods)
* [x] `switch` 
* [x] `throw`
* [x] `trait`
* [x] `try`
* [x] `use`
* [x] `use_lambda`
* [x] `use_trait`
* [x] `var`
* [x] `while`
* [x] `yield`
* [x] `yield from`

## What about these?

* [ ] parameter type declaration
* [ ] property type declaration

## Wait until PHP 8 has been released and we started supporting it

* [ ] `match`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2053,6 +2053,31 @@ Choose from the list of available rules:
   - ``strings_containing_single_quote_chars`` (``bool``): whether to fix
     double-quoted strings that contains single-quotes; defaults to ``false``
 
+* **single_space_after_construct** [@PhpCsFixer]
+
+  Ensures a single space after language constructs.
+
+  Configuration options:
+
+  - ``constructs`` (a subset of ``['abstract', 'as', 'break', 'case', 'catch',
+    'class', 'clone', 'const', 'const_import', 'continue', 'do', 'echo',
+    'else', 'elseif', 'extends', 'final', 'finally', 'for', 'foreach',
+    'function', 'function_import', 'global', 'goto', 'if', 'implements',
+    'include', 'include_once', 'instanceof', 'insteadof', 'interface',
+    'new', 'open_tag_with_echo', 'php_open', 'print', 'private',
+    'protected', 'public', 'require', 'require_once', 'return', 'static',
+    'throw', 'trait', 'try', 'use', 'use_lambda', 'use_trait', 'var',
+    'while', 'yield', 'yield_from']``): list of constructs which must be
+    followed by a single space; defaults to ``['abstract', 'as', 'break',
+    'case', 'catch', 'class', 'clone', 'const', 'const_import', 'continue',
+    'do', 'echo', 'else', 'elseif', 'extends', 'final', 'finally', 'for',
+    'foreach', 'function', 'function_import', 'global', 'goto', 'if',
+    'implements', 'include', 'include_once', 'instanceof', 'insteadof',
+    'interface', 'new', 'open_tag_with_echo', 'php_open', 'print',
+    'private', 'protected', 'public', 'require', 'require_once', 'return',
+    'static', 'throw', 'trait', 'try', 'use', 'use_lambda', 'use_trait',
+    'var', 'while', 'yield', 'yield_from']``
+
 * **single_trait_insert_per_statement** [@Symfony, @PhpCsFixer]
 
   Each trait ``use`` must be done as single statement.
--- a/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
+++ b/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
@@ -1,0 +1,272 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\LanguageConstruct;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\Fixer\ConfigurationDefinitionFixerInterface;
+use PhpCsFixer\FixerConfiguration\AllowedValueSubset;
+use PhpCsFixer\FixerConfiguration\FixerConfigurationResolver;
+use PhpCsFixer\FixerConfiguration\FixerOptionBuilder;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\FixerDefinition\VersionSpecification;
+use PhpCsFixer\FixerDefinition\VersionSpecificCodeSample;
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class SingleSpaceAfterConstructFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @var array<string, null|int>
+     */
+    private static $tokenMap = [
+        'abstract' => T_ABSTRACT,
+        'as' => T_AS,
+        'break' => T_BREAK,
+        'case' => T_CASE,
+        'catch' => T_CATCH,
+        'class' => T_CLASS,
+        'clone' => T_CLONE,
+        'const' => T_CONST,
+        'const_import' => CT::T_CONST_IMPORT,
+        'continue' => T_CONTINUE,
+        'do' => T_DO,
+        'echo' => T_ECHO,
+        'else' => T_ELSE,
+        'elseif' => T_ELSEIF,
+        'extends' => T_EXTENDS,
+        'final' => T_FINAL,
+        'finally' => T_FINALLY,
+        'for' => T_FOR,
+        'foreach' => T_FOREACH,
+        'function' => T_FUNCTION,
+        'function_import' => CT::T_FUNCTION_IMPORT,
+        'global' => T_GLOBAL,
+        'goto' => T_GOTO,
+        'if' => T_IF,
+        'implements' => T_IMPLEMENTS,
+        'include' => T_INCLUDE,
+        'include_once' => T_INCLUDE_ONCE,
+        'instanceof' => T_INSTANCEOF,
+        'insteadof' => T_INSTEADOF,
+        'interface' => T_INTERFACE,
+        'new' => T_NEW,
+        'open_tag_with_echo' => T_OPEN_TAG_WITH_ECHO,
+        'php_open' => T_OPEN_TAG,
+        'print' => T_PRINT,
+        'private' => T_PRIVATE,
+        'protected' => T_PROTECTED,
+        'public' => T_PUBLIC,
+        'require' => T_REQUIRE,
+        'require_once' => T_REQUIRE_ONCE,
+        'return' => T_RETURN,
+        'static' => T_STATIC,
+        'throw' => T_THROW,
+        'trait' => T_TRAIT,
+        'try' => T_TRY,
+        'use' => T_USE,
+        'use_lambda' => CT::T_USE_LAMBDA,
+        'use_trait' => CT::T_USE_TRAIT,
+        'var' => T_VAR,
+        'while' => T_WHILE,
+        'yield' => T_YIELD,
+        'yield_from' => null,
+    ];
+
+    /**
+     * @var array<string, int>
+     */
+    private $fixTokenMap = [];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        if (70000 <= \PHP_VERSION_ID) {
+            self::$tokenMap['yield_from'] = T_YIELD_FROM;
+        }
+
+        $this->fixTokenMap = [];
+
+        foreach ($this->configuration['constructs'] as $key) {
+            $this->fixTokenMap[$key] = self::$tokenMap[$key];
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Ensures a single space after language constructs.',
+            [
+                new CodeSample(
+                    '<?php
+
+throw  new  \Exception();
+'
+                ),
+                new CodeSample(
+                    '<?php
+
+echo  "Hello!";
+',
+                    [
+                        'constructs' => [
+                            'echo',
+                        ],
+                    ]
+                ),
+                new VersionSpecificCodeSample(
+                    '<?php
+
+yield  from  baz();
+',
+                    new VersionSpecification(70000),
+                    [
+                        'constructs' => [
+                            'yield_from',
+                        ],
+                    ]
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAnyTokenKindsFound(array_values($this->fixTokenMap)) && !$tokens->hasAlternativeSyntax();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        $tokenKinds = array_values($this->fixTokenMap);
+
+        for ($index = $tokens->count() - 2; $index >= 0; --$index) {
+            $token = $tokens[$index];
+
+            if (!$token->isGivenKind($tokenKinds)) {
+                continue;
+            }
+
+            $whitespaceTokenIndex = $index + 1;
+
+            if (';' === $tokens[$whitespaceTokenIndex]->getContent()) {
+                continue;
+            }
+
+            if (
+                $token->isGivenKind(T_STATIC)
+                && !$tokens[$tokens->getNextMeaningfulToken($index)]->isGivenKind([T_FUNCTION, T_VARIABLE])
+            ) {
+                continue;
+            }
+
+            if ($token->isGivenKind(T_OPEN_TAG)) {
+                if ($tokens[$whitespaceTokenIndex]->equals([T_WHITESPACE]) && false === strpos($token->getContent(), "\n")) {
+                    $tokens->clearAt($whitespaceTokenIndex);
+                }
+
+                continue;
+            }
+
+            if ($token->isGivenKind(T_CLASS) && $tokens[$tokens->getNextMeaningfulToken($index)]->equals('(')) {
+                continue;
+            }
+
+            if ($token->isGivenKind(T_RETURN) && $this->isMultiLineReturn($tokens, $index)) {
+                continue;
+            }
+
+            if (!$tokens[$whitespaceTokenIndex]->equals([T_WHITESPACE])) {
+                $tokens->insertAt($whitespaceTokenIndex, new Token([T_WHITESPACE, ' ']));
+            } elseif (' ' !== $tokens[$whitespaceTokenIndex]->getContent()) {
+                $tokens[$whitespaceTokenIndex] = new Token([T_WHITESPACE, ' ']);
+            }
+
+            if (
+                70000 <= \PHP_VERSION_ID
+                && $token->isGivenKind(T_YIELD_FROM)
+                && 'yield from' !== strtolower($token->getContent())
+            ) {
+                $tokens[$index] = new Token([T_YIELD_FROM, Preg::replace(
+                    '/\s+/',
+                    ' ',
+                    $token->getContent()
+                )]);
+            }
+        }
+    }
+
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder('constructs', 'List of constructs which must be followed by a single space.'))
+                ->setAllowedTypes(['array'])
+                ->setAllowedValues([new AllowedValueSubset(array_keys(self::$tokenMap))])
+                ->setDefault(array_keys(self::$tokenMap))
+                ->getOption(),
+        ]);
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return bool
+     */
+    private function isMultiLineReturn(Tokens $tokens, $index)
+    {
+        ++$index;
+        $tokenFollowingReturn = $tokens[$index];
+
+        if (
+            !$tokenFollowingReturn->isGivenKind(T_WHITESPACE)
+            || false === strpos($tokenFollowingReturn->getContent(), "\n")
+        ) {
+            return false;
+        }
+
+        $nestedCount = 0;
+
+        for ($indexEnd = \count($tokens) - 1, ++$index; $index < $indexEnd; ++$index) {
+            if (false !== strpos($tokens[$index]->getContent(), "\n")) {
+                return true;
+            }
+
+            if ($tokens[$index]->equals('{')) {
+                ++$nestedCount;
+            } elseif ($tokens[$index]->equals('}')) {
+                --$nestedCount;
+            } elseif (0 === $nestedCount && $tokens[$index]->equalsAny([';', [T_CLOSE_TAG]])) {
+                break;
+            }
+        }
+
+        return false;
+    }
+}
--- a/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
@@ -305,7 +305,7 @@ public function testItDoesSomething() {}}'.$this->whitespacesConfig->getLineEndi
             return 'test_'.$functionName;
         }
 
-        return'test'.ucfirst($functionName);
+        return 'test'.ucfirst($functionName);
     }
 
     /**
--- a/tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
+++ b/tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
@@ -1,0 +1,3086 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tests\Fixer\LanguageConstruct;
+
+use PhpCsFixer\ConfigurationException\InvalidFixerConfigurationException;
+use PhpCsFixer\Tests\Test\AbstractFixerTestCase;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ *
+ * @internal
+ *
+ * @covers \PhpCsFixer\Fixer\LanguageConstruct\SingleSpaceAfterConstructFixer
+ */
+final class SingleSpaceAfterConstructFixerTest extends AbstractFixerTestCase
+{
+    /**
+     * @dataProvider provideInvalidConstructCases
+     *
+     * @param mixed $construct
+     */
+    public function testConfigureRejectsInvalidControlStatement($construct)
+    {
+        $this->expectException(InvalidFixerConfigurationException::class);
+
+        $this->fixer->configure([
+            'constructs' => [
+                $construct,
+            ],
+        ]);
+    }
+
+    /**
+     * @return array
+     */
+    public function provideInvalidConstructCases()
+    {
+        return [
+            'null' => [null],
+            'false' => [false],
+            'true' => [true],
+            'int' => [0],
+            'float' => [3.14],
+            'array' => [[]],
+            'object' => [new \stdClass()],
+            'unknown' => ['foo'],
+        ];
+    }
+
+    /**
+     * @dataProvider provideFixWithAbstractCases
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFixWithAbstract($expected, $input = null)
+    {
+        $this->fixer->configure([
+            'constructs' => [
+                'abstract',
+            ],
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixWithAbstractCases()
+    {
+        return [
+            [
+                '<?php abstract class Foo {};',
+                '<?php abstract  class Foo {};',
+            ],
+            [
+                '<?php abstract class Foo {};',
+                '<?php abstract
+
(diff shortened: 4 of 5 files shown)


src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,96 @@
...
+     * @var int[]
+     */
+    private static $tokenKinds = [
+        T_BREAK,
+        T_CONTINUE,
+        T_ECHO,
+        T_NEW,
+        T_PRINT,
+        T_RETURN,
+        T_THROW,
```
localheinz: Could list a bunch of additional statements here, however, some of these statements are already handled by other fixers, for example:

```diff
diff --git a/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php b/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
index d8864775..2a70edb7 100644
--- a/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
+++ b/src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
@@ -30,8 +30,15 @@ final class SingleSpaceAfterConstructFixer extends AbstractFixer
         T_BREAK,
         T_CONTINUE,
         T_ECHO,
+        T_ELSE,
+        T_ELSEIF,
+        T_IF,
+        T_INCLUDE,
+        T_INCLUDE_ONCE,
         T_NEW,
         T_PRINT,
+        T_REQUIRE,
+        T_REQUIRE_ONCE,
         T_RETURN,
         T_THROW,
     ];
```

localheinz: I think that for consistency, all of these should be added to the fixer. This should probably even belong to the `@PSR2` ruleset, and then all of the other fixers would not need to worry about whitespace adjustment.

What do you think?

julienfalque: Indeed, PSR-2 requires to have a single space after most of those keywords:
> Control structure keywords MUST have one space after them

Though I don't think this applies to e.g. `echo` or `print`. Maybe the list should be configurable so that the `@PSR2` ruleset enables the fixer with the tokens related to this rule only?

localheinz: Sounds good, @julienfalque!

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,153 @@
...
+        'break' => T_BREAK,
+        'continue' => T_CONTINUE,
+        'echo' => T_ECHO,
+        'goto' => T_GOTO,
+        'new' => T_NEW,
+        'print' => T_PRINT,
+        'return' => T_RETURN,
+        'throw' => T_THROW,
+        'yield' => T_YIELD,
+        'yield_from' => T_YIELD_FROM,
```
localheinz: Let me know if you have any ideas for constructs (and/or statements) I could add here!

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
```diff
@@ -0,0 +1,421 @@
...
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixWithContinueCases()
+    {
+        return [
+            [
+                '<?php continue 9000; ?>',
```
keradus: this is duplicated by next case

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
```diff
@@ -0,0 +1,421 @@
...
+        ];
+    }
+
+    /**
+     * @dataProvider provideFixWithGotoCases
+     *
+     * @param string      $expected
+     * @param null|string $input
+     */
+    public function testFixWithGoto($expected, $input = null)
```
keradus: looks like we can just have a single `testFix` method. we don't need separated method for fixing `echo`, separated method for fixing `continue` and so...

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
```diff
@@ -0,0 +1,421 @@
...
+
+    public function provideFixWithYieldFromCases()
+    {
+        return [
+            [
+                '<?php function foo() { yield from [2, 12, 13]; } ?>',
+            ],
+            [
+                '<?php function foo() { yield from [2, 12, 13]; } ?>',
+                '<?php function foo() { yield from  [2, 12, 13]; } ?>',
```
keradus: do we have an multiline example like
```php
echo
        123;
```
?

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: How so?

I want to ensure that we have a single space, right?

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: I took a look at `BlankLineBeforeStatementFixerTest`, so I figured it would make sense to test individual configuration options.

Should I merge into one?

keradus: you can test individual configs even when having only one `testFix`.
`testFix($expected, $input = null, array $config = [])` (we even sometimes have `expected, input, config, file`)

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: Added!

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
keradus:
```php
            [
                '<?php return 9000;',
            ],
            [
                '<?php return 9000;',
                '<?php return  9000;',
            ],
```

first scenario will test
`<?php return 9000;` got not changed

2nd scenario will test
`<?php return 9000;` got not changed
`<?php return  9000;` got changed into `<?php return 9000;`

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: Adjusted!

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: Hmm, but doesn't the second scenario test that

```php
<?php return  9000;
```

is fixed to 

```php
<?php return 9000;
```

?

keradus: if you have `expected + input` it test 2 things, as I said.
- `input` was fixed into `expected`
- `expected` is not modified when fixing it

keradus:
btw, CI is constantly failing on broken AutoReview

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
localheinz: Ha, I see - wasn't aware of it.

localheinz:
Not sure how to deal with the failure related to the constant `T_YIELD_FROM` not defined on PHP 5.6, what do you think?

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
derrabus: `require`, `include` and their `*_once` counterparts.

localheinz: Thank you, @derrabus!

localheinz:
A bunch of tests are currently missing.

IllyaMoskvin:
Just want to say that this seems like a great feature. I'm new to using this tool, so I might have missed something, but the only other way I've found to fix things like `if(` to `if (` is by using the `braces` fixer. I'm hesitant to do so due to how many responsibilities it has (cf. #472, #823). I think this is a nice addition for users who'd like to make those choices in a more granular manner.

IllyaMoskvin:
Found what looks like a bug. Running the fixer on this file:

```php
<?php

if (1 === 1)
{
    echo 'foo';
}
else
{
    echo 'bar';
}
```

...will produce the following output:

```php
<?php

if (1 === 1)
{
    echo 'foo';
}
else {
    echo 'bar';
}
```

Uhh please ignore the non-standard brace style, we don't usually do this. Just spotted it in the diff. It feels like removing the newline after the `else` is beyond the intended scope of this fixer.

IllyaMoskvin:
Another bug:

```php
<?php

return;
```

...results in this:

```php
<?php

return ;
```

This seems to impact `break` and `continue` the most. I can see adding spaces being useful for these constructs in some cases (e.g. `continue(2)`). Just thinking of scope. Can we make it so that there's no space added if the construct is immediately followed by a semicolon?

localheinz:
@IllyaMoskvin 

Thank you for reporting!

This definitely needs work. I’m on vacation right now, will take a look when back behind my computer!

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
dmvdbrugge: `clone`?

README.rst
```diff
@@ -1780,6 +1780,17 @@ Choose from the list of available rules:
...
+* **single_space_after_construct**
+
+  Ensures a single space after language constructs.
+
+  Configuration options:
+
+  - ``constructs`` (a subset of ``['break', 'continue', 'echo', 'goto', 'new',
+    'print', 'return', 'throw', 'yield', 'yield_from']``): list of constructs
+    which must be followed by a single space; defaults to ``['break',
+    'continue', 'goto', 'new', 'return', 'throw', 'yield']``
```
dmvdbrugge: Why is `yield` default, but `yield from` not?

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
localheinz: @derrabus 

Implemented for

- `include`
- `include_once`
- `require`
- `require_once`

localheinz: @dmvdbrugge 

Implemented for

* `clone`

README.rst
localheinz: @dmvdbrugge 

Currently failing because of support for PHP 5.6 - not sure yet how to work around it.

dmvdbrugge: By only adding them if the version allows? There are more fixers with version-specific configuration.

localheinz:
@derrabus @dmvdbrugge @julienfalque @IllyaMoskvin @keradus 

Let's see if the tests pass (`yield from` requires PHP 7).

Can you think of something else, perhaps, which of these constructs should be defaults?

dmvdbrugge:
Opinion: I would say all of them should be default (even `yield from` if the running version allows it).

dmvdbrugge:
Also the remaining points in the initial comment, you ask
> (should it?)

In my opinion: yes (surprise surprise 😜)

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configure(array $configuration = null)
+    {
+        parent::configure($configuration);
+
+        if (70000 <= \PHP_VERSION_ID && \defined('T_YIELD_FROM')) {
+            self::$tokenMap = array_merge(self::$tokenMap, [
```
dmvdbrugge: These are both "hardcoded" arrays. Doesn't it make more sense to directly add the values, instead of `array_merge`?

localheinz:
Does anyone have an idea how to work around the tests failing on PHP 5.6?

See https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/609203996#L791-L800:

```
1) PhpCsFixer\Tests\Console\Command\ReadmeCommandTest::testIfReadmeFileIsCorrect
README.rst file is not up to date! Do not modify it manually! Regenerate readme with command: `php php-cs-fixer readme > README.rst`.
--- Expected
+++ Actual
@@ @@
     'goto', 'include', 'include_once', 'new', 'print', 'require',
-    'require_once', 'return', 'throw', 'yield']``): list of constructs which
-    must be followed by a single space; defaults to ``['break', 'case',
-    'clone', 'continue', 'echo', 'goto', 'include', 'include_once', 'new',
-    'print', 'require', 'require_once', 'return', 'throw', 'yield']``
+    'require_once', 'return', 'throw', 'yield', 'yield_from']``): list of
+    constructs which must be followed by a single space; defaults to
+    ``['break', 'case', 'clone', 'continue', 'echo', 'goto', 'include',
+    'include_once', 'new', 'print', 'require', 'require_once', 'return',
+    'throw', 'yield', 'yield_from']``
```

julienfalque:
👋 @localheinz. I'm currently working on breaking down the `braces` rule into several smaller rules with less responsibilities. As you noticed, one of `braces`' responsibilities is to ensure control structure keywords such as `if`, `else` but also `declare` are followed by a single space. Do you mind adding support for these keywords in your PR?

kubawerlos:
> Does anyone have an idea how to work around the tests failing on PHP 5.6?

@localheinz I'd say test `README.rst` is up to date only on PHP 7.4.

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class SingleSpaceAfterConstructFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @var array
+     */
+    private static $tokenMap = [
```
kubawerlos: [More](https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/v1.17.0/src/Fixer/SingleSpaceAfterStatementFixer.php#L21)?

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,180 @@
...
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Andreas Möller <am@localheinz.com>
+ */
+final class SingleSpaceAfterConstructFixer extends AbstractFixer implements ConfigurationDefinitionFixerInterface
+{
+    /**
+     * @var array
```
kubawerlos:
```suggestion
     * @var array<string, int>
```

localheinz:
@kubawerlos 

Thank you for your feedback - I'm a bit busy at the moment, but might get around to it next week!

guilliamxavier:
Hello, thanks for the work! What do you think about adding a test for the following fix?

```diff
-echo$a;
+echo $a;
```

(yes, `echo$a;` without space actually [works](https://3v4l.org/R1J4h)...)

Same for `return` (https://3v4l.org/2btMT) and maybe others?

SpacePossum:
please add;
```
            T_ARRAY,
            T_EMPTY,
            T_EVAL,
            T_EXIT,
            T_ISSET,
            T_LIST,
            T_UNSET,
            T_VARIABLE,
            T_DECLARE,
``

**Comment by guilliamxavier** [comment]
@SpacePossum: that would change `declare(strict_types=1);` to `declare (strict_types=1);`, `$f($x)` to `$f ($x)`, `unset($a, $b);` to `unset ($a, $b);` etc. 😕

File: tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php [hunk 10]
```diff
@@ -0,0 +1,746 @@
...
+                'return',
+            ],
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixWithReturnCases()
+    {
+        return [
```
julienfalque: Please add a test case with a multiline return value, e.g.:
```php
return
    $a
    && $b
;
```
IMO in this case the newline after `return` is legit and should not be removed. This probably applies to more keywords (e.g. `yield`).

ossinkine:
Hey guys, looks like this PR is ready for a long time. What is keeping us from merging?

@localheinz @derrabus @julienfalque @keradus @kubawerlos @dmvdbrugge

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,180 @@
...
+        'include_once' => T_INCLUDE_ONCE,
+        'new' => T_NEW,
+        'print' => T_PRINT,
+        'require' => T_REQUIRE,
+        'require_once' => T_REQUIRE_ONCE,
+        'return' => T_RETURN,
+        'throw' => T_THROW,
+        'yield' => T_YIELD,
+        'yield_from' => null,
+    ];
```
rquadling, line 53: Can we have T_OPEN_TAG_WITH_ECHO and T_CLOSE_TAG in this list too? (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/5138)

SpacePossum: and maybe `T_AS` as well? as in `foreach($foo as $b){}`

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
localheinz: Ha, interesting, @kubawerlos - wasn't aware you already had a fixer for this case!

Thank you!

localheinz:
For the record, I have update the list of tokens this fixer should handle, thanks to the work of @kubawerlos on https://github.com/kubawerlos/php-cs-fixer-custom-fixers/blob/v1.17.0/src/Fixer/SingleSpaceAfterStatementFixer.php#L21.

tests/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixerTest.php
```diff
@@ -0,0 +1,2617 @@
...
+            ],
+            [
+                '<?php use /* foo */Foo\Bar;',
+                '<?php use/* foo */Foo\Bar;',
+            ],
+            [
+                '<?php use /* foo */Foo\Bar;',
+                '<?php use  /* foo */Foo\Bar;',
+            ],
+            // not detected as T_USE, why?
```
localheinz, line 2488: @julienfalque @keradus SpacePossum 

Does anyone have an idea why these cases are not detected?

Unless I'm mistaken, these usages of `use` should also be `T_USE`, shouldn't they?

julienfalque: This is because `PhpCsFixer\Tokenizer\Transformer\UseTransformer` changes those to `CT::T_USE_TRAIT` and `CT::T_USE_LAMBDA`.

localheinz: @julienfalque 

Ha, good to know!

I will add these as separate tokens, then!

localheinz:
@derrabus @dmvdbrugge @julienfalque @keradus @kubawerlos @rquadling 

Can I ask you to take a look?

localheinz:
There are a bunch of case I don't know how to handle - not sure which tokens I should use to identify and fix these cases:

### `use const`

```diff
-use const  Foo\BAR;
+use const Foo\BAR;
```
### `use function`

```diff
-use function  Foo\bar;
+use function Foo\bar;
```

guilliamxavier:
You can quickly debug tokens by printing `Tokens::fromCode('<?php ...')->toJson()`

So I see that `use const` and `use function` respectively give [`T_USE`, `T_WHITESPACE`, `CT::T_CONST_IMPORT`] and [`T_USE`, `T_WHITESPACE`, `CT::T_FUNCTION_IMPORT`]

localheinz:
Thank you, @guilliamxavier!

SpacePossum:
can you maybe add T_OPEN_TAG_WITH_ECHO as well?

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,220 @@
...
+            } elseif (' ' !== $tokens[$whitespaceTokenIndex]->getContent()) {
+                $tokens->clearAt($whitespaceTokenIndex);
+                $tokens->insertAt($whitespaceTokenIndex, new Token([T_WHITESPACE, ' ']));
+            }
+
+            if (
+                70000 <= \PHP_VERSION_ID
+                && $token->isGivenKind(T_YIELD_FROM)
+                && 'yield from' !== strtolower($token->getContent())
+                && 1 === Preg::match('/^yield\s+from$/i', $token->getContent())
```
SpacePossum, line 199: this will always be true, otherwise it is not Tokenized as T_YIELD_FROM so for performance the check can be removed

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,220 @@
...
+            ) {
```
SpacePossum, line 200: you can override in one go;
```
{
                $tokens[$index] = new Token([T_YIELD_FROM, Preg::replace(
                    '/\s+/',
                    ' ',
                    $token->getContent()
                )]);
}
```

src/Fixer/LanguageConstruct/SingleSpaceAfterConstructFixer.php
```diff
@@ -0,0 +1,220 @@
...
+                $token->isGivenKind(T_STATIC)
+                && !$tokens[$tokens->getNextMeaningfulToken($index)]->isGivenKind([T_FUNCTION, T_VARIABLE])
+            ) {
+                continue;
+            }
+
+            if (!$tokens[$whitespaceTokenIndex]->equals([T_WHITESPACE])) {
+                $tokens->insertAt($whitespaceTokenIndex, new Token([T_WHITESPACE, ' ']));
+            } elseif (' ' !== $tokens[$whitespaceTokenIndex]->getContent()) {
+                $tokens->clearAt($whitespaceTokenIndex);
```
SpacePossum, line 191: same as https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4435/files#r508393721
you can do `$tokens[$whitespaceTokenIndex] = new Token([T_WHITESPACE, ' ']);`

SpacePossum:
I've added support for T_OPEN, please pull @localheinz  :)

localheinz:
Thank you, @SpacePossum!

SpacePossum:
```diff
-            $this->translator = new class() implements TranslatorInterface {
+            $this->translator = new class () implements TranslatorInterface {
```

I think we should not change this one (PHP7 anonymous class case)

SpacePossum:
I think the only case open is what @julienfalque also commented, multi line `return` statements;

```diff
-        return
-            Preg::match('/(^|\R)\h*[^@\s]\N*/', $this->cleanComment($parts[$index - 1]))
+        return Preg::match('/(^|\R)\h*[^@\s]\N*/', $this->cleanComment($parts[$index - 1]))
             || Preg::match('/^.*?\R\s*[^@\s]/', $this->cleanComment($parts[$index + 1]))
         ;
```

we want to fix these, maybe skip, maybe option?

localheinz:
@SpacePossum 

> I think the only case open is what @julienfalque also commented, multi line `return` statements;

Ha, good point!

localheinz:
@julienfalque @SpacePossum 

I think I got the detection of multi-line return statements right - do you want to take a look?

Is there a need to handle something like this?

```php
return
    $foo ?
    $bar :
    $baz;
```

localheinz:
Unless there are any objections, I think this is good!

🤓 

Sorry for taking so long to push this through!

src/Fixer/PhpUnit/PhpUnitTestAnnotationFixer.php
```diff
@@ -305,7 +305,7 @@ private function addTestPrefix($functionName)
             return 'test_'.$functionName;
         }
 
-        return'test'.ucfirst($functionName);
+        return 'test'.ucfirst($functionName);
```
localheinz, line 308: Ha, nice - this is the only place where the fixer fixed something so far in the code base!

OskarStark:
Great work, thank you 😃

localheinz:
Thank you, @OskarStark!

SpacePossum:
looks great @localheinz :+1: 
can you review my changes in this commit: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4435/commits/be1aeb1722a812107b8c5883bea8287c4c01412f ?

if any others can review that would be great, but I think this RTM :)

SpacePossum:
please squash to one commit :)

localheinz:
@SpacePossum 

Hang on a moment!

localheinz:
@SpacePossum 

Squashed and rebased!

SpacePossum:
Thanks so much @

drupol:
Nice!

localheinz:
Thank you, @derrabus, @dmvdbrugge, @drupol, @guilliamxavier, @IllyaMoskvin, @julienfalque, @keradus, @kubawerlos, @OskarStark, @ossinkine, @rquadling, and @SpacePossum!

guilliamxavier:
Many thanks 🎉 

Now what about adding it to the `@Symfony` ruleset (cf. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4998#issuecomment-653890485)?
(and maybe also to the `@PSR2` ruleset with a subset of constructs (cf. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4435#discussion_r289762802)?)

Also, how to ensure that the priorities are right, e.g. that this will apply *after* NoSpacesAfterFunctionNameFixer (so that we can finally close #4817)?

localheinz:
@guilliamxavier 

I have made a note about that and might look into that tomorrow!

drupol:
Hello,

Despite that I really like this new feature, I have an issue with it.

It tries to fix:

```php
interface A extends
    B,
    C,
    E
{
    // Something
}
```

into 

```php
interface A extends B,
    C,
    E
{
    // Something
}
```

WDYT about this example? Do you think it should still fix something?

SpacePossum:
Hi @drupol :wave:  great to see you again : )

Your example looks indeed like a valid case to fix differently than is currently done by the fixer. Can I bother you with asking to open a new fresh issue for this?

localheinz:
@drupol 

I will be happy to take a look!

drupol:
Hello mates,

Glad to see you around as well!

I'll open a new issue!

PR #4440, opened 2019-06-04
SimpleToComplexStringVariableFixer - Fix $ bug

dmvdbrugge:
Account for $ being the last character of the encapsulating string, it needs to be escaped in order to not form a new simple string variable.


SpacePossum:
Thank you @dmvdbrugge.

PR #4445, opened 2019-06-06
DX: static call of markTestSkippedOrFail

kubawerlos:


SpacePossum:
Thank you @kubawerlos.

Issue #3988, opened 2018-09-03
Single line "throw"


nicolas-grekas:
On symfony/symfony, we keep asking people to not split lines that start with "throw" (that's our CS, it helps grepping the code for errors.)
It'd be awesome if php-cs-fixer could do it for us.

Related to #3168 (method's signature also should be on a single line.)

keradus:
Can you raise a PR, @nicolas-grekas ?

nicolas-grekas:
Nope sorry I need help also.

PR #4452, opened 2019-06-23
Add SingleLineThrowFixer

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3988

@nicolas-grekas should we have this as part of `@Symfony` ruleset?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1841,6 +1841,10 @@ Choose from the list of available rules:
   - ``comment_types`` (a subset of ``['asterisk', 'hash']``): list of comment types
     to fix; defaults to ``['asterisk', 'hash']``
 
+* **single_line_throw**
+
+  Throwing exception must be done in single line.
+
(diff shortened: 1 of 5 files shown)


nicolas-grekas:
It don't the rules of the project regarding updating existing rulesets, but we would definitely enable this one the Symfony repo. Thanks for working on it!

julienfalque:
:+1: for adding the rule to the `@Symfony` ruleset.

SpacePossum:
can you add these test case:

```php
<?php throw new Exception( // 0
                "Foo", // 1
                0 // 2
            ); //3
```

```php
<?php throw new Exception(  0 /*
1
2
3
*/
  /*4*/              "Foo", /*5*/
/*6*/0 /*7*/);
```

keradus:
(rebased + fix small issues causing CI to fail)

keradus:
Thank you @kubawerlos.

PR #4458, opened 2019-06-30
Add PHPStan

julienfalque:
Let's start with level 0 and increase it progressively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ jobs:
         -
             stage: Static Code Analysis
             php: 7.2
-            env: COMPOSER_FLAGS="--no-dev --prefer-stable"
+            env: COMPOSER_FLAGS="--prefer-stable"
             install:
                 - ./dev-tools/install.sh
 
@@ -51,6 +51,7 @@ jobs:
                     | grep -v tests/TestCase.php \
                     && (echo "UNKNOWN FILES DETECTED" && travis_terminate 1) || echo "NO UNKNOWN FILES"
                 - ./check_trailing_spaces.sh || travis_terminate 1
+                - ./dev-tools/tools/phpstan analyse
                 - if [ -n "$COMMIT_SCA_FILES" ]; then ./dev-tools/vendor/bin/phpmd `echo "$COMMIT_SCA_FILES" | grep -Ev "^tests/Fixtures" | xargs | sed 's/ /,/g'` text phpmd.xml || travis_terminate 1; fi
                 - ./dev-tools/tools/composer-require-checker check composer.json --config-file $(realpath .composer-require-checker.json) || travis_terminate 1
                 - composer normalize -d ./dev-tools ./../composer.json --dry-run
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,0 +1,10 @@
+parameters:
+    level: 0
+    paths:
+        - src
+        - tests
+    excludes_analyse:
+        - tests/Fixtures
+    ignoreErrors:
+        - '/^.+::__construct\(\) does not call parent constructor from SplFileInfo\.$/'
+        - '/^Return typehint of method PhpCsFixer\\Tests\\Test\\.+::createIsIdenticalStringConstraint\(\) has invalid type PHPUnit_Framework_Constraint_IsIdentical\.$/'
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -2,6 +2,7 @@
     "symbol-whitelist" : [
         "Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface",
         "PhpCsFixer\\Tests\\Test\\Constraint\\SameStringsConstraint",
+        "PhpCsFixer\\Tests\\Test\\IsIdenticalConstraint",
         "PHPUnit\\Framework\\Constraint\\IsIdentical",
         "PHPUnit\\Framework\\TestCase",
         "PHPUnit\\Runner\\Version",
--- a/src/Console/Command/DescribeNameNotFoundException.php
+++ b/src/Console/Command/DescribeNameNotFoundException.php
@@ -37,6 +37,8 @@ final class DescribeNameNotFoundException extends \InvalidArgumentException
     {
         $this->name = $name;
         $this->type = $type;
+
+        parent::__construct();
     }
 
     /**
--- a/tests/Smoke/PharTest.php
+++ b/tests/Smoke/PharTest.php
@@ -12,6 +12,7 @@
 
 namespace PhpCsFixer\Tests\Smoke;
 
+use Keradus\CliExecutor\CliResult;
 use Keradus\CliExecutor\CommandExecutor;
 use PhpCsFixer\Console\Application;
 use PhpCsFixer\Console\Command\DescribeCommand;
--- a/tests/Test/AbstractFixerTestCase.php
+++ b/tests/Test/AbstractFixerTestCase.php
@@ -31,6 +31,7 @@ use Prophecy\Argument;
 abstract class AbstractFixerTestCase extends TestCase
 {
     use AssertTokensTrait;
+    use IsIdenticalConstraint;
 
     /**
      * @var LinterInterface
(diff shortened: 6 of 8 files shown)


.travis.yml
```diff
@@ -71,6 +71,7 @@ jobs:
                 - composer info -D | sort
             script:
                 - vendor/bin/phpunit || travis_terminate 1
+                - dev-tools/vendor/bin/phpstan analyse --level=0 --configuration=phpstan.neon src tests
```
kubawerlos: Can we have here only `dev-tools/vendor/bin/phpstan analyse`? Configuration path is not needed (we would benefit from `phpstan.neon`/`phpstan.neon.dist` pair) and level and paths can be move to config file.

phpstan.neon
```diff
@@ -0,0 +1,3 @@
+parameters:
```
kubawerlos: Better add here:
```
parameters:
    level: 0
    paths:
        - src
        - tests
```
and rename the file to `phpstan.neon.dist`

.travis.yml
julienfalque: Done.

phpstan.neon
julienfalque: What is the benefit of having a `.dist` file?

kubawerlos: Exactly the same as for `.php_cs.dist` or `phpunit.xml.dist` - to test new config locally, e.g. you can increase the level or try some exclude - we would have some - like in [`Tokens::insertAt`](https://github.com/phpstan/phpstan/issues/2106).

julienfalque: What prevents one from editing the `phpstan.neon` file and eventually commit changes? I never saw a legit use case for `.php_cs.dist`. `phpunit.xml.dist` makes sense when one wants custom local options for e.g. output, but PHPStan does not have such option I think.

dmvdbrugge: It follows the pattern of "config files like these *with* `.dist` are to be committed while *without* are to be ignored." I find that a nice separation, and enough reason to do it. (So hereby also the request to add the without version to `.gitignore`)

julienfalque: I'm feeling like we would follow this pattern for the sake of following this pattern. Why would one want to maintain a specific local file rather than editing the tracked one? What use case would need to have a local file that permanently differs from the tracked one?

localheinz: I agree with @julienfalque - I never had a need for `.dist` files.

dev-tools/composer.json
```diff
@@ -7,7 +7,8 @@
         "localheinz/composer-normalize": "^1.1",
         "maglnet/composer-require-checker": "^2.0",
         "mi-schi/phpmd-extension": "^4.3",
-        "phpmd/phpmd": "^2.6"
+        "phpmd/phpmd": "^2.6",
+        "phpstan/phpstan": "^0.11"
```
SpacePossum: we might want;
```
phpstan/phpstan-deprecation-rules
phpstan/phpstan-strict-rules
phpstan/phpstan-phpunit
phpstan/phpstan-php-parser
```

as well?

julienfalque: Indeed, but let's add them later, just as we'll rise the analyse level later.

julienfalque:
Question: we currencly have most PHP tools installed in `dev-tools` directory via a dedicated `composer.json` to avoid conflicts with PHP CS Fixer own dependencies. But we are starting to have a lot of tools and some might conflict with others. What do you think about splitting this `composer.json` by tool, i.e.:
* dev-tools/phpmd
* dev-tools/phpstan
* ...
?

julienfalque:
Finally green (the remaining Travis failure does not seem to be related to the changes in this PR).

I moved PHPStan installation to its own directory like I suggested in [my previous comment](#issuecomment-508706376).

drupol:
Hi,

Maybe we should evaluate [Grumphp](https://www.hansgrohe.be/nl/articledetail-raindance-douchepaneel-lift-180-2jet-opbouw-27008400) ?
One single configuration file that would do all the checks.

What do you think?

keradus:
we evaluated it quite a lot time ago (over a year, i guess).
it was not mature enough and we faced few issues (same time, we helped to solve some of them)

maybe it's mature now for us, maybe not. we would need to reevaluate. lets not mix multiple topics in single pr.

phpstan.neon
Slamdunk: I agree with @julienfalque too

Slamdunk:
> I moved PHPStan installation to its own directory like I suggested in [my previous comment](#issuecomment-508706376).

There is also the shim version of PHPStan for this exact reason: https://github.com/phpstan/phpstan-shim

julienfalque:
> There is also the shim version of PHPStan for this exact reason: https://github.com/phpstan/phpstan-shim

I didn't know about it. I have no preference for one over the other so I would keep `phpstan/phpstan`.

.travis.yml
```diff
@@ -35,6 +35,7 @@ jobs:
                 - composer info -D | sort
 
                 - ./dev-tools/bin-download.sh
+                - travis_retry composer install --working-dir=dev-tools/phpstan $DEFAULT_COMPOSER_FLAGS
```
keradus: why not part of https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.15/dev-tools/composer.json ?

julienfalque: See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4458#issuecomment-508706376

This:
- prevents conflicts with dependencies of other tools;
- allows locking PHPStan version to prevent random build failures when a new PHPStan version brings new errors.

I would do the same with other tools but this is out of the scope of this PR.

keradus: do we really have dependencies conflict?
for locking the version, we can do it without extra composer.json file

julienfalque: > do we really have dependencies conflict?

We don't, for now.

> for locking the version, we can do it without extra composer.json file

Using a lock file removes the need to resolve dependencies so installation is faster, I would keep it.

keradus: my point - it's enough to have one, single dev-tools/composer.json, not one per tool.

keradus: maybe https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4536 ? i wanted to switch long time ago, anyway

julienfalque: 👍 for using PHIVE, I'll update my PR.

keradus: then, maybe just review #4526 and I will merge it, the phpstan is already in my pr

.travis.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         -
             stage: Static Code Analysis
             php: 7.2
-            env: COMPOSER_FLAGS="--no-dev --prefer-stable"
```
keradus: it was `--no-dev` to not run `composer-require-checker` with dev dependencies of main composer file installed

tests/Test/IsIdentical.php
```diff
@@ -0,0 +1,56 @@
...
+
+use PhpCsFixer\PhpunitConstraintIsIdenticalString\Constraint\IsIdenticalString;
+use PHPUnit\Framework\Constraint\IsIdentical as PhpUnitIsIdentical;
+
+/**
+ * @internal
+ *
+ * @todo Remove me when usages will end up in dedicated package.
+ */
+trait IsIdentical
```
keradus: let either name it or put to subnamespace of `Constraint`

tests/Test/IsIdentical.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @return string
+     */
+    private static function getIsIdenticalStringConstraintClassName()
+    {
+        foreach ([
+            IsIdenticalString::class,
+            PhpUnitIsIdentical::class,
+            'PHPUnit_Framework_Constraint_IsIdentical',
```
keradus: can we already use `PHPUnit_Framework_Constraint_IsIdentical::class` , perhaps ?

.travis.yml
julienfalque: I need to install dev dependencies so that PHPStan does not complain about missing classes. I think Composer Require Checker ignores dev dependencies anyway.

.travis.yml
julienfalque: Rebased.

tests/Test/IsIdentical.php
julienfalque: The class does not actually exist, maybe the line can be dropped instead.

tests/Test/IsIdentical.php
julienfalque: Trait renamed to `IsIdenticalConstraint`.

SpacePossum:
Thank you @julienfalque.

PR #4467, opened 2019-07-17
ToolInfo - fix access to reference without checking existence

### LINKED ISSUE #4465: Fixer fails with undefined reference
PHP 7.3.7
PHP CS Fixer 2.15.1
Composer version 1.8.6

Problem: running the fixer throws "Undefined index: reference" in ToolInfo.php line 68. https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.15/src/ToolInfo.php#L68

My composer.lock generates without "reference" in "dist" sections. When running an older version of composer the reference is always null, I guess the newer version just doesn't add the whole key when it is null.

Fix: The isset in line 67 should check for the whole dist/reference, not just dist.

black-silence:
fixes #4465

### DIFF
--- a/src/ToolInfo.php
+++ b/src/ToolInfo.php
@@ -64,7 +64,7 @@ final class ToolInfo implements ToolInfoInterface
 
         $versionSuffix = '';
 
-        if (isset($package['dist'])) {
+        if (isset($package['dist']['reference'])) {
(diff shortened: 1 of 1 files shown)


SpacePossum:
Thank you @black-silence.

PR #4472, opened 2019-07-23
Fix non-static closure unbinding this on PHP 7.4

kelunik:
Fixes 'Unbinding $this of closure is deprecated in .../friendsofphp/php-cs-fixer/src/FixerConfiguration/FixerOption.php:174'


SpacePossum:
Thank you @kelunik.

PR #4476, opened 2019-07-26
FunctionsAnalyzer - add "isTheSameClassCall" for correct verifying of function calls

kubawerlos:


src/Tokenizer/Analyzer/FunctionsAnalyzer.php
```diff
@@ -96,4 +96,31 @@ public function getFunctionReturnType(Tokens $tokens, $methodIndex)
...
         return new TypeAnalysis($type, $typeStartIndex, $typeEndIndex);
     }
+
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     *
+     * @return bool
+     */
+    public function isTheSameClassCall(Tokens $tokens, $index)
```
SpacePossum: I like the PR as a good clean up, I'm not sure about the name of this method though. 
Also, not sure about adding a new method to this class following the reasoning [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3961#issuecomment-450569021), maybe a new analyzer would be a better fit?

kubawerlos: I'm open for suggestion for new names (analyzer and method) ;)

`MethodssAnalyzer`? the function name is the best I can think of.

julienfalque: This class is `@internal` so I think adding this method here is fine.

keradus:
Thank you @kubawerlos.

PR #4477, opened 2019-07-26
DX: control names of public methods in test's classes

kubawerlos:


tests/AutoReview/ProjectCodeTest.php
```diff
@@ -463,6 +492,10 @@ static function (SplFileInfo $file) {
             iterator_to_array($finder, false)
         );
 
+        $classes = array_filter($classes, function ($class) {
```
SpacePossum: can be `static`

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -208,34 +208,63 @@ public function testThatSrcClassesNotExposeProperties($className)
...
+     *
+     * @param string $testClassName
+     */
+    public function testThatPublicMethodsAreCorrectlyNamed($testClassName)
+    {
+        $reflectionClass = new \ReflectionClass($testClassName);
+
+        $publicMethods = array_filter(
+            $reflectionClass->getMethods(\ReflectionMethod::IS_PUBLIC),
+            function (\ReflectionMethod $reflectionMethod) use ($reflectionClass) {
```
SpacePossum: can be `static`

kubawerlos: @SpacePossum Why don't we have `static_lambda` enabled in the `PhpCsFixer` rule set?

tests/AutoReview/ProjectCodeTest.php
```diff
@@ -208,34 +208,63 @@ public function testThatSrcClassesNotExposeProperties($className)
...
+
+        if ($publicMethods === []) {
+            $this->addToAssertionCount(1); // no methods to test, all good!
+        }
+
+        foreach ($publicMethods as $method) {
+            static::assertRegExp(
+                '/^(test|provide|setUpBeforeClass$|tearDownAfterClass$)/',
+                $method->getName(),
+                sprintf('Public method "%s::%s" is not properly named.', $reflectionClass->getName(), $method->getName())
```
keradus: when i would run newly written tests and see this msg as an outcome in console, i would have no clue why given method is not "properly named"

tests/AutoReview/ProjectCodeTest.php
SpacePossum: it is a risky fixer and when applied causes some of the tests to fail because of the mock framework we use (IIRC), but yeah, I would like it on ;)

tests/AutoReview/ProjectCodeTest.php
kubawerlos: wouldn't you go then to the test and see the regex? can you [propose](https://thenextweb.com/dd/2018/10/16/github-launches-suggested-changes-to-make-it-easier-to-collaborate-within-pull-requests/) the message?

SpacePossum:
Thank you @kubawerlos.

Issue #1812, opened 2016-03-10
Extra brackets removal fixer proposal


GrahamCampbell:
For example:

``` php
$stuff = ($this->baz) ? $this->foo() : $this->bar();
```

We don't need the brackets around `$this->baz`.

SpacePossum:
:+1:

SpacePossum:
@keradus good pointers, but that kind of code.. oh no ^-^

gharlan:
Extra cautions is also necessary for all assignment operators and for `and`, `or` and `xor`.

```php
($a = foo()) ? 1 : 2;

($a and $b) ? 1 : 2;
```

interduo:
Example2 of doing useless bracketing:
`$stuff = ($this ? $something1 : $something2);`

keradus:
Hey! I'm closing this as no commited effort to this by anyone. The idea is valid, yet it didn't get any movement around it code-wise for 8 years. I see no value in keeping it open due to that.
If any of you see a value in it and are ready to implement it, please raise a PR.

PR #4479, opened 2019-07-28
IncludeFixer - remove braces when the statement is wrapped in block

kubawerlos:
Ping @umpirsky, @sstok and @keradus (as authors) for review.

### DIFF
--- a/src/Tokenizer/Analyzer/BlocksAnalyzer.php
+++ b/src/Tokenizer/Analyzer/BlocksAnalyzer.php
@@ -1,0 +1,70 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Tokenizer\Analyzer;
+
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ */
+final class BlocksAnalyzer
+{
+    /**
+     * @param Tokens   $tokens
+     * @param null|int $openIndex
+     * @param null|int $closeIndex
+     *
+     * @return bool
+     */
+    public function isBlock(Tokens $tokens, $openIndex, $closeIndex)
+    {
+        if (null === $openIndex || null === $closeIndex) {
+            return false;
+        }
+
+        if (!$tokens->offsetExists($openIndex)) {
+            return false;
+        }
+
+        if (!$tokens->offsetExists($closeIndex)) {
+            return false;
+        }
+
+        $blockType = $this->getBlockType($tokens[$openIndex]);
+
+        if (null === $blockType) {
+            return false;
+        }
+
+        return $closeIndex === $tokens->findBlockEnd($blockType, $openIndex);
+    }
+
+    /**
+     * @param Token $token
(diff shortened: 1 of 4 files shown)


tests/Tokenizer/Analyzer/BlocksAnalyzerTest.php
```diff
@@ -0,0 +1,87 @@
...
+        return [
+            ['<?php foo(1);', 2, 4],
+            ['<?php foo((1));', 3, 5],
+            ['<?php foo((1));', 2, 6],
+            ['<?php foo(1, 2, 3);', 2, 10],
+            ['<?php foo(1, bar(2, 3), 4);', 2, 16],
+            ['<?php $foo["bar"];', 2, 4],
+            ['<?php [1, 2, 3];', 1, 9],
+            ['<?php $foo = function ($x) { return $x + 10; };', 7, 9],
+            ['<?php $foo = function ($x) { return $x + 10; };', 11, 22],
```
SpacePossum: can you add some array destructuring samples as well (in both syntax styles)?

SpacePossum:
@kubawerlos why target master and not 2.12 here?

kubawerlos:
@SpacePossum after all it's a new feature, right? Also I've separated the changes into few commits, so in case of bugfix for the fixer on `2.12` we can cherry-pick the first one.

SpacePossum:
The changes to the rule/fixer are a nice improvement on the existing logic which would be great to have on 2.12. For that I'm happy to see this in 2.12, but if others think it cannot than fine as well :+1:

dmvdbrugge:
I don't see new options, it's not an entire new fixer, I could see this as a "bugfix" (of a "forgotten" edge case) and thus targeting 2.12

SpacePossum:
@kubawerlos would you mind to retarget to 2.12?

kubawerlos:
@SpacePossum done

src/Tokenizer/Analyzer/BlocksAnalyzer.php
```diff
@@ -0,0 +1,66 @@
...
+
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author Kuba Werłos <werlos@gmail.com>
+ *
+ * @internal
+ */
+final class BlocksAnalyzer
```
keradus: please add some docs for it, out of the blue, i don't see it as generic analyzer that deserve an separated, reusable class

src/Tokenizer/Analyzer/BlocksAnalyzer.php
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * @param Tokens   $tokens
+     * @param null|int $openIndex
+     * @param null|int $closeIndex
+     *
+     * @return bool
+     */
+    public function isBlock(Tokens $tokens, $openIndex, $closeIndex)
+    {
+        if (!$tokens->offsetExists($openIndex)) {
```
keradus: https://www.php.net/manual/en/splfixedarray.offsetexists.php
it expects `int` as parameter, you are giving `null|int`.

src/Tokenizer/Analyzer/BlocksAnalyzer.php
kubawerlos: It's more about SRP than re-usability.

Through, I was thinking about using it when working on https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1812

src/Tokenizer/Analyzer/BlocksAnalyzer.php
kubawerlos: Yet, it there are tests for `null` and are passing.

Anyway, I've added the check for `null` to make it more clear.

SpacePossum:
Thank you @kubawerlos.

PR #4488, opened 2019-08-02
MethodArgumentSpaceFixer - Regression

SpacePossum:
bug is in the current 2.12 branch, but is not released yet

I think this is regression after https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/75b55b529e09b32f3d3215b4429939154e0c5fab

causes 3.0 currently to fail as well

ping @keradus


julienfalque:
@SpacePossum I tried to rebase your branch on latest 2.15 and I noticed it contains the same test but with the change being expected. I would say the current behavior of 2.15 is correct and this PR can be closed.

SpacePossum:
thank you @julienfalque !

PR #4493, opened 2019-08-06
Add echo_tag_syntax rule

### LINKED ISSUE #2990: Normalizing simple echo tags (short and long forms)
We currently have the `no_short_echo_tag` to expand code from `<?=` to `<?php echo`.

#2520 was an attempt to do the reverse, but I think it has been closed mainly because of lack of interest.

I can take care of this, by implementing to a configurable fixer that can merge the two fixers (`no_short_echo_tag` and `opening_tag_plus_echo_to_short_echo_tag`), into a new `short_echo_tag` fixer.

This fixer could have
- a `format` option (required), with a value of:
  - `'short'` to convert `<?php echo` and `<?php print` to `<?=`
  - `'long'` to convert `<?=` to `<?php echo` or `<?php print`
- a `printOrEcho` option (optional, defaulting to `'echo'`), with a value of:
  - `'echo'` to convert `<?=` to `<?php echo`
  - `'print'` to convert `<?=` to `<?php print`

We could also add more functionalities to this fixer, but those would maybe be better implemented in another fixer. I'm thinking about normalizing spaces and trailing semi-colons (eg formatting `<?php echo 2;?>` as `<?= 2 ?>` or `<?=2?>` or whatever).

Thoughts?

mlocati:
We have a fixer (`no_short_echo_tags`) that expands short echo tags (`<?= $foo ?>`) to long echo tags (`<?php echo $foo ?>`).

What about adding a new fixer (`short_echo_tags`) that can also do the reverse?

Fix #2990


src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver([
+            (new FixerOptionBuilder(self::OPTION_FORMAT, 'How the fixer should process short/long echo tags'))
+                ->setAllowedValues($this->supportedFormatOptions)
+                ->setDefault(self::FORMAT_SHORT)
```
kubawerlos: Why not `self::FORMAT_SHORT`?

mlocati: > Why not `self::FORMAT_SHORT`?

@kubawerlos I'm not sure I understand... it *is* `self::FORMAT_SHORT`...

kubawerlos: @mlocati yes, sorry, I meant why not `self::FORMAT_LONG` - it would be consistent with `no_short_echo_tag`

mlocati: @kubawerlos I agree: see 451129f30ac68adfa68723b156872b18c5293b14

Big-Shark:
@kubawerlos Any update about this?

README.rst
```diff
@@ -1711,6 +1712,20 @@ Choose from the list of available rules:
...
   *Risky rule: risky when the ``settype`` function is overridden or when used as the 2nd or 3rd expression in a ``for`` loop .*
 
+* **short_echo_tag** [@PhpCsFixer]
+
+  Replace short-echo ``<?=`` with long format ``<?php echo``/``<?php print``
+  syntax, or vice-versa.
+
+  Configuration options:
+
+  - ``format`` (``'long'``, ``'short'``): how the fixer should process short/long echo
```
kubawerlos: The description could be more clear, like from `no_mixed_echo_print` - `the desired language construct`.

dev-tools/trigger-website.sh
```diff
@@ -4,7 +4,7 @@ set -eu
 TOKEN=$1
 MSG_SUFFIX=${2:-''}
 
-if [ ! -z "$MSG_SUFFIX" ]
+if [ -n "$MSG_SUFFIX" ]
```
kubawerlos: Can we revert this change?

src/Fixer/PhpTag/NoShortEchoTagFixer.php
```diff
@@ -12,16 +12,17 @@
...
+use PhpCsFixer\Fixer\DeprecatedFixerInterface;
 use PhpCsFixer\FixerDefinition\CodeSample;
 use PhpCsFixer\FixerDefinition\FixerDefinition;
-use PhpCsFixer\Tokenizer\Token;
-use PhpCsFixer\Tokenizer\Tokens;
 
 /**
  * @author Vincent Klaiber <hello@vinkla.com>
+ *
+ * @deprecated in 2.16, proxy to ShortEchoTagFixer
```
kubawerlos: Version is now out of date, I don't think we need it at all.

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        if (self::FORMAT_SHORT === $this->configuration[self::OPTION_FORMAT]) {
+            if (null !== $tokens->findSequence([[\T_OPEN_TAG], [\T_ECHO]])) {
```
kubawerlos: In [Cookbook](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.0/doc/COOKBOOK-FIXERS.md) the method `isCandidate` has mentioning that it "has to be ultra cheap to execute" - I'm not sure if `findSequence` is fine to have here - @keradus, @SpacePossum, @julienfalque WDYT?

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(SplFileInfo $file, Tokens $tokens)
+    {
+        if (self::FORMAT_SHORT === $this->configuration[self::OPTION_FORMAT]) {
+            $this->longToShort($file, $tokens);
+        } else {
```
kubawerlos: There is no point to pass `$file` in above and below lines.

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+            $this->shortToLong($file, $tokens);
+        }
+    }
+
+    private function longToShort(SplFileInfo $file, Tokens $tokens)
+    {
+        $offset = 0;
+        for (;;) {
+            $foundEcho = $tokens->findSequence([[\T_OPEN_TAG], [\T_ECHO]], $offset);
+            $foundPrint = $tokens->findSequence([[\T_OPEN_TAG], [\T_PRINT]], $offset);
```
kubawerlos: Searching bor both and lower using only one - seems like waste of CPU.
Why not with classic `for` loop:
```
for ($i = 0 ; $i < count; $i++)
    if $i is not T_OPEN_TAG -> continue;
    $nextMeanigful = ...;
    if $nextMeanigful is not echo or print -> continue;
    if for complex option
    override
```

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+
+    private function shortToLong(SplFileInfo $file, Tokens $tokens)
+    {
+        if (self::LONG_FUNCTION_PRINT === $this->configuration[self::OPTION_LONG_FUNCTION]) {
+            $echoToken = [\T_PRINT, 'print'];
+        } else {
+            $echoToken = [\T_ECHO, 'echo'];
+        }
+        $offset = $tokens->count() - 1;
+        for (;;) {
```
kubawerlos: A classic loop would be more explicit or at least merge `$offset` and `$found` - they could be simply one variable, preferably named `$index`.

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+            $offset = $found - 1;
+            $replace = [new Token([\T_OPEN_TAG, '<?php ']), new Token($echoToken)];
+            if (!$tokens[$found + 1]->isWhitespace()) {
+                $replace[] = new Token([\T_WHITESPACE, ' ']);
+            }
+            $tokens->overrideRange($found, $found, $replace);
+        }
+    }
+
+    private function isComplexCode(Tokens $tokens, $index)
```
kubawerlos: I'm not sure id I understand it - can you add few more tests for this option?

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+                $semicolonFound = true;
+            } elseif ($semicolonFound && !$token->isWhitespace()) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
```
kubawerlos: No need for FQCN in this PHPDoc (BTW seems like an idea for new fixer).

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+     * @param int                          $openTagIndex
+     * @param int                          $echoTagIndex
+     *
+     * @return \PhpCsFixer\Tokenizer\Token[]
+     */
+    private function buildLongToShortTokens(Tokens $tokens, $openTagIndex, $echoTagIndex)
+    {
+        $result = [new Token([\T_OPEN_TAG_WITH_ECHO, '<?='])];
+
+        // Skip to the first non-whitespace token between $openTagIndex and $echoTagIndex
```
kubawerlos: From description it looks like `Tokens::getNextNonWhitespace`.

src/Fixer/PhpTag/ShortEchoTagFixer.php
```diff
@@ -0,0 +1,242 @@
...
+                $found = $foundPrint;
+            }
+            $startRange = key($found);
+            $offset = $startRange + 1;
+            if (!$this->configuration[self::OPTION_SHORT_ALWAYS] && $this->isComplexCode($tokens, $offset)) {
+                continue;
+            }
+            next($found);
+            $endRange = key($found);
+            $newTokens = $this->buildLongToShortTokens($tokens, $startRange, $endRange);
```
kubawerlos: Can't we simply remove the `echo`/`print` token with the following whitespace - if any - and override the tag open?

tests/Fixer/PhpTag/NoShortEchoTagFixerTest.php
```diff
@@ -31,7 +31,7 @@ final class NoShortEchoTagFixerTest extends AbstractFixerTestCase
      */
     public function testFix($expected, $input = null)
     {
-        if (!ini_get('short_open_tag')) {
+        if (\PHP_VERSION_ID < 50400 && !ini_get('short_open_tag')) {
```
kubawerlos: We don't support PHP version lower than `5.4`, so this change is pointless.

README.rst
mlocati: :+1: 94dc2c3dd377c5c08cd7c04c91a5ec1864bb20c5

dev-tools/trigger-website.sh
mlocati: Why? I agree that it's not strictly related to this PR, but IMHO the code is more clear...

For example, which of the following PHP code examples do you prefer?
```php
if (!($a === false)) {
```
vs
```php
if ($a !== false)) {
```

Furthermore, it fixes [this warning](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/568483121#L633)

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: If there's a faster way to check if `$tokens` contains `<?php echo` or `<?php print` I'd be glad to update the code...

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: Not exactly the same: see the `between $openTagIndex and $echoTagIndex` part of the description

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: It's not that simple. For example:
```
<? /* comment */ echo 1 ?>
```
should become
```
<?= /* comment */ 1 ?>
```

(it's included in the fixer tests).

src/Fixer/PhpTag/NoShortEchoTagFixer.php
mlocati: Done: 8ec8c60451e2e31844b1c0f81c1627b759c716ed

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: Done: f7a7d1a345ad755458c40708a79ec650141f59ef

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: There are already tests for it. BTW I added a PHPDoc comment to this method: dadc11e152f13bb15854f57ce4dc5eccf8c99f13

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: I do prefer FQCN in PHPDocs, and that's for these reasons:

1. If we need to reference a class only in PHPDocs, do we need to add a `use` statement at the beginning of the file?
If yes, it's quite useless to me.
If no: what if we add at a later time a reference to it in the code: should we fix all the previous PHPDocs?

2. consider this example:
    ```
    <?php
    use Namespace1\ClassName;
    use Namespace2\ClassName as ClassNameAlias;
    
    /**
     * @return ClassNameAlias
     */
    function foo() {
    }
    ```
    In order to know what `ClassNameAlias` actually is, humans have to read the `use` statements at the top of the file.

BTW, since this project doesn't use FQCN in comments, I updated this class: 9031768b641b9fb3ae82ea0b48053d0fd28cd5fe

tests/Fixer/PhpTag/NoShortEchoTagFixerTest.php
mlocati: Done: c3535e6f59568504b9a81418418a2194af780bcc

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: Done: 5b3d5eb6ce6373aa1cfdf73752814061f7913ed7

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: I simplified the code a bit: see 7feb5bee336c2c6ee6551f9ec47df45e5032b2cb

dev-tools/trigger-website.sh
kubawerlos: Sorry, I was thinking it's some leftover after old PRs.

src/Fixer/PhpTag/ShortEchoTagFixer.php
kubawerlos: We don't have to search for the sequence - see the `isCandidate` method in other fixers - it could use `findSequence` in few of them, but the goal is the method is fast, not super accurate - I'm not saying this is right (maybe more complex `isCandidate` could save some files from being checked) - but it is the approach all other fixers follow.

Simple `$tokens->isAnyTokenKindsFound([T_ECHO, T_PRINT])` would be fine, I guess.

src/Fixer/PhpTag/ShortEchoTagFixer.php
kubawerlos: I don't understand - that's exactly what I'm suggesting - `echo`  and ` ` after it are removed and `<?' becomes `<?=`.

src/Fixer/PhpTag/ShortEchoTagFixer.php
kubawerlos: Can we have a test for this "not exactly"? All tests are passing for https://github.com/mlocati/PHP-CS-Fixer/pull/1

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: You are absolutely right: 372f92e0f07fcb61ba78b83c4b5c918e4ae1ca69

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: Sorry, I didn't understand what you meant.

About removing the whitespace, we have two options: remove the whitespaces after the PHP opening tag, or remove the whitespaces after the `echo`/`print`.

Consider these two cases:

```
<?php /*comment*/ echo
  1
?>

<?php
  /*comment*/ echo 1
?>
```

what should be the short-echo version of them?

src/Fixer/PhpTag/ShortEchoTagFixer.php
mlocati: Yuu are right :+1:. That's because for sure the next non white-space token can only be an integer whose value is at max `$echoTagIndex`, it's better to use `getNextNonWhitespace()`

tests/Fixer/PhpTag/NoShortEchoTagFixerTest.php
julienfalque: The `if` can be completely removed here too.

julienfalque: The `if` can be completely removed here too.

README.rst
```diff
@@ -1799,6 +1800,20 @@ Choose from the list of available rules:
 
   *Risky rule: risky when the ``settype`` function is overridden or when used as the 2nd or 3rd expression in a ``for`` loop .*
 
+* **short_echo_tag** [@PhpCsFixer]
```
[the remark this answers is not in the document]
mlocati: Done: ac53c2c0ab452cd9f2ecf8fce6f2fa6b07e66231

tests/Fixer/PhpTag/NoShortEchoTagFixerTest.php
mlocati: Done: 2721a030d64f2f00ef3952a5077857b545609a25

README.rst
```diff
@@ -1799,6 +1800,20 @@ Choose from the list of available rules:
...
+
+  Replace short-echo ``<?=`` with long format ``<?php echo``/``<?php print``
```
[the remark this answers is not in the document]
mlocati: Done: f986d5f05950fc1d832bdbe0775615dc45864241

src/RuleSet.php
```diff
@@ -255,6 +254,7 @@ final class RuleSet implements RuleSetInterface
             'phpdoc_var_annotation_correct_order' => true,
             'protected_to_private' => true,
             'return_assignment' => true,
+            'echo_tag_syntax' => true,
```
julienfalque: Should be moved after `compact_nullable_typehint` for alphabetical order.

mlocati: Done: dbefb78ccb2a48ea59f31abc94cc0d2358b4718e

mlocati:
There's still [this issue](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493#discussion_r350040138) to be fixed.

How should these cases be translated from long-echo to short-echo tags?

1.
    ```php
    <?php /*comment*/ echo
      1
    ?>
    ```
2.
    ```php
    <?php
      /*comment*/ echo 1
    ?>
    ```
3.
    ```php
    <?php
      /*comment*/
      echo
      1
    ?>
    ```

julienfalque:
IMO keeping current behavior is fine.

mlocati:
> IMO keeping current behavior is fine.

Ok, so I'll update the code as per @kubawerlos [suggestion](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493#discussion_r348714652).

mlocati:
> Ok, so I'll update the code as per @kubawerlos [suggestion](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493#discussion_r348714652).

Nope, I can't find a way to implement https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493#discussion_r348714652 in a faster/simpler way.

So, unless @kubawerlos sees a way to implement it, IMHO this PR is ready to be merged...

julienfalque:
Could you please rebase onto latest master to get your branch up-to-date?

mlocati:
> Could you please rebase onto latest master to get your branch up-to-date?

Done.

src/Fixer/PhpTag/EchoTagSyntaxFixer.php
```diff
@@ -0,0 +1,246 @@
...
+    /** @internal */
+    const FORMAT_LONG = 'long';
+
+    /** @internal */
+    const LONG_FUNCTION_ECHO = 'echo';
+
+    /** @internal */
+    const LONG_FUNCTION_PRINT = 'print';
+
+    private $supportedFormatOptions = [
```
kubawerlos: Can be constant as well (same for `$supportedLongFunctionOptions`) - it is not changing during the run.

README.rst
```diff
@@ -598,6 +598,20 @@ Choose from the list of available rules:
...
+  Replaces short-echo ``<?=`` with long format ``<?php echo``/``<?php print``
+  syntax, or vice-versa.
+
+  Configuration options:
+
+  - ``format`` (``'long'``, ``'short'``): the desired language construct; defaults to
+    ``'long'``
+  - ``long_function`` (``'echo'``, ``'print'``): the function to be used to expand the
+    short echo tags; defaults to ``'echo'``
+  - ``short_always`` (``bool``): always render short-echo tags even in case of
```
kubawerlos: `handle_complex_code`? `short_always` looks meaningless.

mlocati: `handle_complex_code` with a boolean value? I don't think it's very clear...

julienfalque: What about `shorten_simple_statements_only`, defaulting to `true`? A bit verbose but very explicit :)

src/Fixer/PhpTag/EchoTagSyntaxFixer.php
mlocati: Done in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493/commits/034546b04580b786f795945c9f191dfc9cb8e218

README.rst
mlocati: Done in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4493/commits/034546b04580b786f795945c9f191dfc9cb8e218

julienfalque:
@mlocati Before merging this PR we should squash commits. As there is a commit from @kubawerlos, could you please squash all consecutive commits from same author?

mlocati:
> Before merging this PR we should squash commits. As there is a commit from @kubawerlos, could you please squash all consecutive commits from same author?

Done.

julienfalque:
Thank you @mlocati.

mlocati:
So, I submitted this PR 16 months ago, it was reviewed 3 months later, and still today we don't have a release including it?
You should really redesign your release process guys, I think I made a huge mistake investing my time contributing to PHP-CS-Fixer.

SpacePossum:
2.17 has been released with your rule. We are working on a new release model so we can release far more often.
Thank for all your work and for your patience!

Issue #4396, opened 2019-04-28
binary_operator_spaces should have a single_space_minimal option


VincentLanglet:
In rule: binary_operator_spaces,
There is an `align_single_space` and an `align_single_space_minimal` option.

I had some difficulties to understand the way they worked but if I'm not wrong 
- `align_single_space`: Align operator with at least one space before and after
- `align_single_space_minimal`: Align operator with exactly one space before and after

So there should be
- `single_space`: At least one space before and after
- `single_space_minimal`: Exactly one space before and after (The actual `single_space` behaviour)

SpacePossum:
Hi!

I think your observations are correct for the two align options; one will align to the operators and assures a single space, the other will do the same but assures minimal space on the right side of the operators.

For the not-aligning options, `single_space` is indeed your `single_space_minimal` description.

So if understand correctly you want a non-aligning option that assure a single space, but allows for multiple spaces?
So:
```php
$a = $b+$c;
```
would be fixed to
```php
$a = $b + $c;
```
however;
```php
$a = $b              +      $c;
```
would not be fixed?

VincentLanglet:
Yes @SpacePossum 

I didn't look at the code, but I think it wouldn't be difficult to implement this option.
I think the more coherent should be to rename the `single_space` option as `single_space_minimal` and writing the real `single_space` option, but it implies breaking changes...

SpacePossum:
I think the options were added later and therefore the naming might not be as coherent as could be, however we took a practical way of adding these without BC breaking changes.

For the requested missing option; I'm not sure I see a use case for this one. Currently having multiple spacing left and/or right of the operator is only possible for alignment reasons. I think that is valid, but having such spacing for a single operator is not a code style I've seen around much. Do you maybe have an example project that uses this code style?

VincentLanglet:
In our code base, we aligned `=` when we declared const,
```
    const GENDER_MISTER  = 'gender.mister';
    const GENDER_MADAM   = 'gender.mrs';
    const GENDER_MISS    = 'gender.miss';
    const GENDER_COMPANY = 'gender.company';
    const GENDER_OTHER   = 'gender.other';
    const GENDER_UNKNOWN = 'gender.unknown';
```
But we use one space for anywhere else.

I know it's not a classic usecase.

stevelacey:
[...]

PR #4497, opened 2019-08-09
BinaryOperatorSpacesFixer - Single space minimum option

stevelacey:
Assuming this is of interest I'll tidy it up as per the contribution guide.

This allows a config like this:

```
        'binary_operator_spaces' => [
            'default' => 'single_space',
            'operators' => [
                '=' => 'single_space_minimum',
                '=>' => 'single_space_minimum',
            ]
        ],
```

Which allows a less drastic change than either forcing single spaces or forcing alignment for all `=` and `=>` but still enforcing at least one space – which are currently the only available options afaics.

The naming of this is tricky, I think the other names are a bit confusing, I also thought about calling this `single_space_or_more` to distinguish it from `align_single_space_minimal`. Is that better?

An alternative (or extra) implementation I'd probably prefer would be `single_space_or_align`... but that'd be a lot harder to throw together than this was.

For comparison, on my current project, using `single_space` or `align_single_space_minimal` resulted in 300+ file changes, given there's a whole mix of legitimately aligned and single-spaced things, whilst `single_space_minimum` caused 5 file changes, sniping out the specific problem I wanted to address: zero spaces.

Thoughts? Shall I tidy this up and add the docs and tests?

Related: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4396

### DIFF
--- a/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
+++ b/src/Fixer/Operator/BinaryOperatorSpacesFixer.php
@@ -41,6 +41,11 @@ final class BinaryOperatorSpacesFixer extends AbstractFixer implements Configura
     /**
      * @internal
      */
+    const SINGLE_SPACE_MINIMUM = 'single_space_minimum';
+
+    /**
+     * @internal
+     */
     const NO_SPACE = 'no_space';
 
     /**
@@ -87,6 +92,7 @@ final class BinaryOperatorSpacesFixer extends AbstractFixer implements Configura
         self::ALIGN_SINGLE_SPACE,
         self::ALIGN_SINGLE_SPACE_MINIMAL,
         self::SINGLE_SPACE,
+        self::SINGLE_SPACE_MINIMUM,
         self::NO_SPACE,
         null,
     ];
@@ -341,6 +347,12 @@ $foo = \json_encode($bar, JSON_PRESERVE_ZERO_FRACTION | JSON_PRETTY_PRINT);
             return;
         }
 
+        if (self::SINGLE_SPACE_MINIMUM === $this->operators[$tokenContent]) {
+            $this->fixWhiteSpaceAroundOperatorToAtLeastOneSingleSpace($tokens, $index);
+
+            return;
+        }
+
         if (self::NO_SPACE === $this->operators[$tokenContent]) {
             $this->fixWhiteSpaceAroundOperatorToNoSpace($tokens, $index);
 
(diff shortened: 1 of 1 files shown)


stevelacey:
@SpacePossum @keradus ?

stevelacey:
@julienfalque @nicolas-grekas?

nicolas-grekas:
No opinion on my side as I don't think I'd use this rule. Single space is my all time favorite :)

SpacePossum:
Hi and apologies for getting back so late.

Am I correct to understand you want something like this:

```php
$a =$b&&$c; // fix
$d =$a   &&    $e; // don't fix
```

being fixed to this:

```php
$a =$b && $c; // fix
$d =$a   &&    $e; //don't fix
```

? 
It might help having unit tests as an example :)

Im +0 on adding the option, but leave the remark that Im very much in favor to rename the options in 3.0.0 because there is a lot of confusion we could get rid of :+1:

stevelacey:
@SpacePossum yes, whilst those examples aren't really what I'm after, that's what the effect would be, in short, I want to enforce a minimum of 1 space, without enforcing or disallowing alignment.

If it were trivial to write "single space or alignment" as a rule I'd do that, but that's pretty complex.

I am happy to spend more time on this and submit update this PR with tests... if it'd ever merge.

SpacePossum:
>if it'd ever merge.

I cant promise that as we would have to see what others think as well, but I would vote for it :)
There are a few details to sort out though, not sure if you want to work on this or if you want to wait for more people expressing support for your PR in concept?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -393,6 +405,33 @@ private function fixWhiteSpaceAroundOperatorToSingleSpace(Tokens $tokens, $index
...
+    /**
+     * @param Tokens $tokens
+     * @param int    $index
+     */
+    private function fixWhiteSpaceAroundOperatorToAtLeastOneSingleSpace(Tokens $tokens, $index)
+    {
+        // fix white space after operator
+        if ($tokens[$index + 1]->isWhitespace()) {
+            $content = $tokens[$index + 1]->getContent();
+            if (!preg_match('/^\s+$/', $content) && false === strpos($content, "\n") && !$tokens[$tokens->getNextNonWhitespace($index + 1)]->isComment()) {
```
SpacePossum: >if (!preg_match('/^\s+$/', $content) && false === strpos($content, "\n")

would it be possible for content to contain `"\n"` after the preg match check?

src/Fixer/Operator/BinaryOperatorSpacesFixer.php
```diff
@@ -393,6 +405,33 @@ private function fixWhiteSpaceAroundOperatorToSingleSpace(Tokens $tokens, $index
...
+                $tokens[$index - 1] = new Token([T_WHITESPACE, ' ']);
+            }
+        } else {
+            $tokens->insertAt($index, new Token([T_WHITESPACE, ' ']));
+        }
+    }
+
     /**
      * @param Tokens $tokens
      * @param int    $index
```
SpacePossum: needs:
- README regenerate
- unit tests
- code sample in the fixer

AJenbo:
I would like to voice my support for this, having one space with optional alignment is ideal imo.

VincentLanglet:
@stevelacey This option would be great !

I'm just scared about the confusion
SINGLE_SPACE -> ALIGN_SINGLE_SPACE_MINIMAL
SINGLE_SPACE_MINIMUM -> ALIGN_SINGLE_SPACE

Wirone:
@stevelacey if you still want this to be merged, please provide tests for this feature.

PR #4500, opened 2019-08-12
NoSuperfluousPhpdocTags - Add remove_inheritdoc option

julienfalque:
`@inheritDoc` is superfluous when not surrounded by text because it is already the default behavior.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1122,6 +1122,7 @@ Choose from the list of available rules:
 
   - ``allow_mixed`` (``bool``): whether type ``mixed`` without description is allowed
     (``true``) or considered superfluous (``false``); defaults to ``false``
+  - ``remove_inheritdoc`` (``bool``): remove ``@inheritDoc`` tags; defaults to ``false``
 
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -1122,6 +1122,7 @@ Choose from the list of available rules:
 
   - ``allow_mixed`` (``bool``): whether type ``mixed`` without description is allowed
     (``true``) or considered superfluous (``false``); defaults to ``false``
+  - ``remove_inheritdoc`` (``bool``): remove ``@inheritDoc`` tags; defaults to ``false``
```
kubawerlos: `remove_superfluous_inheritdoc`? Seeing `remove_inheritdoc` I'd expect all occurrences would be removed, not  only

> when not surrounded by text because it is already the default behavior.

julienfalque: I thought about it, but the name of the rule already says it's about superfluous tags. Wouldn't that be an unnecessary repetition?

julienfalque:
What about enabling this rule for the project?

SpacePossum:
Thank you @julienfalque.

PR #4516, opened 2019-08-24
DX: Lock binary SCA tools versions

keradus:
Travis is red for all branches - I can confirm that it's due to Travis unattended update of some libraries, let's lock them.

#4494 was fixing the symptom (already merged), but let us not have same issue again


keradus:
Screenshot from 2019-08-24 17-38-56

keradus:
ref #4494

PR #4406, opened 2019-05-06
NoSuperfluousElseifFixer - fix invalid escape sequence in character class

remicollet:
Try to fix #4405

PR #4517, opened 2019-08-24
Verify PCRE pattern before use

ktomk:
To give a detailed PregException message making any problems w/ the
pattern visible.

Add to all the Preg::...($pattern, ....) methods as getting details of any
pattern related errors later on is not easily possible b/c of multi-
matching w/ and w/o UTF-8 mode (see b29848e).

NOTE: Replacement parameters are not checked.

Related: #4406
Reference: b29848ed609c652150e0a2813cefe5b9b70a9e1d

### DIFF
--- a/src/Preg.php
+++ b/src/Preg.php
@@ -45,7 +45,7 @@ final class Preg
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_match.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
     }
 
     /**
@@ -71,7 +71,7 @@ final class Preg
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_match_all.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
     }
 
     /**
@@ -97,7 +97,7 @@ final class Preg
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_replace.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
     }
 
     /**
@@ -123,7 +123,7 @@ final class Preg
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_replace_callback.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
     }
 
     /**
@@ -148,7 +148,7 @@ final class Preg
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_split.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
     }
 
     /**
@@ -180,10 +180,54 @@ final class Preg
             return '';
         }
 
-        $delimiter = substr($pattern, 0, 1);
+        $delimiter = $pattern[0];
 
         $endDelimiterPosition = strrpos($pattern, $delimiter);
 
         return substr($pattern, 0, $endDelimiterPosition).str_replace('u', '', substr($pattern, $endDelimiterPosition));
+    }
+
+    /**
+     * Create PregException.
+     *
+     * Create the generic PregException message and if possible due to finding
+     * an invalid pattern, tell more about such kind of error in the message.
+     *
+     * @param int      $error
+     * @param string   $method
+     * @param string[] $patterns
+     *
+     * @return PregException
+     */
+    private static function newPregException($error, $method, array $patterns)
+    {
+        foreach ($patterns as $pattern) {
+            $last = error_get_last();
+            $result = @preg_match($pattern, '');
+
+            if (false !== $result) {
+                continue;
+            }
+
+            $code = preg_last_error();
+            $next = error_get_last();
+
+            if ($last !== $next) {
+                $message = sprintf(
+                    '(code: %d) %s',
+                    $code,
+                    preg_replace('~preg_[a-z_]+[()]{2}: ~', '', $next['message'])
+                );
+            } else {
+                $message = sprintf('(code: %d)', $code);
+            }
+
+            return new PregException(
+                sprintf('%s(): Invalid PCRE pattern "%s": %s (version: %s)', $method, $pattern, $message, PCRE_VERSION),
+                $code
+            );
+        }
+
+        return new PregException(sprintf('Error occurred when calling %s.', $method), $error);
     }
 }
--- a/tests/PregTest.php
+++ b/tests/PregTest.php
@@ -27,7 +27,7 @@ final class PregTest extends TestCase
     public function testMatchFailing()
     {
         $this->expectException(PregException::class);
-        $this->expectExceptionMessage('Error occurred when calling preg_match.');
+        $this->expectExceptionMessage('Preg::match(): Invalid PCRE pattern ""');
 
         Preg::match('', 'foo', $matches);
     }
@@ -47,10 +47,113 @@ final class PregTest extends TestCase
         static::assertSame($expectedMatches, $actualMatches);
     }
 
+    public function providePatternValidationCases()
+    {
+        return [
+            'invalid_blank' => ['', null, PregException::class],
+            'invalid_open' => ["\1", null, PregException::class, "'\1' found"],
+            'valid_control_character_delimiter' => ["\1\1", 1],
+            'invalid_control_character_modifier' => ["\1\1\1", null, PregException::class, ' Unknown modifier '],
+            'valid_slate' => ['//', 1],
+            'valid_paired' => ['()', 1],
+            'null_byte_injection' => ['()'."\0", null, PregException::class, ' Null byte in regex '],
+            'paired_non_utf8_only' => ["((*UTF8)\xFF)", null, PregException::class, 'UTF-8'],
+            'valid_paired_non_utf8_only' => ["(\xFF)", 1],
+            'php_version_dependent' => ['([\\R])', 0, PregException::class, 'Compilation failed: escape sequence is invalid '],
+        ];
+    }
+
+    /**
+     * @dataProvider providePatternValidationCases
+     *
+     * @param $pattern
+     * @param null|int    $expected
+     * @param null|string $expectedException
+     * @param null|string $expectedMessage
+     */
+    public function testPatternValidation($pattern, $expected = null, $expectedException = null, $expectedMessage = null)
+    {
+        $setup = function () use ($expectedException, $expectedMessage) {
+            $i = 0;
+
+            if (null !== $expectedException) {
+                ++$i;
+                $this->expectException($expectedException);
+            }
+
+            if (null !== $expectedMessage) {
+                ++$i;
+                $this->expectExceptionMessage($expectedMessage);
+            }
+
+            return (bool) $i;
+        };
+
+        try {
+            $actual = Preg::match($pattern, "The quick brown \xFF\x00\\xXX jumps over the lazy dog\n");
+        } catch (\Exception $ex) {
+            $setup();
+
+            throw $ex;
+        }
+
+        if (null !== $expected) {
+            static::assertSame($expected, $actual);
+
+            return;
+        }
+
+        $setup() || $this->addToAssertionCount(1);
+    }
+
+    /**
+     * @dataProvider providePatternValidationCases
+     *
+     * @param string      $pattern
+     * @param null|int    $expected
+     * @param null|string $expectedException
+     * @param null|string $expectedMessage
+     */
+    public function testPatternsValidation($pattern, $expected = null, $expectedException = null, $expectedMessage = null)
+    {
+        $setup = function () use ($expectedException, $expectedMessage) {
+            $i = 0;
+
+            if (null !== $expectedException) {
+                ++$i;
+                $this->expectException($expectedException);
+            }
+
+            if (null !== $expectedMessage) {
+                ++$i;
+                $this->expectExceptionMessage($expectedMessage);
+            }
+
+            return (bool) $i;
+        };
+
+        try {
+            $buffer = "The quick brown \xFF\x00\\xXX jumps over the lazy dog\n";
+            $actual = $buffer !== Preg::replace((array) $pattern, 'abc', $buffer);
+        } catch (\Exception $ex) {
+            $setup();
+
+            throw $ex;
+        }
+
+        if (null !== $expected) {
+            static::assertSame((bool) $expected, $actual);
+
+            return;
+        }
+
+        $setup() || $this->addToAssertionCount(1);
+    }
+
     public function testMatchAllFailing()
     {
         $this->expectException(PregException::class);
-        $this->expectExceptionMessage('Error occurred when calling preg_match_all.');
+        $this->expectExceptionMessage('Preg::matchAll(): Invalid PCRE pattern ""');
 
         Preg::matchAll('', 'foo', $matches);
     }
@@ -73,7 +176,7 @@ final class PregTest extends TestCase
     public function testReplaceFailing()
     {
         $this->expectException(PregException::class);
-        $this->expectExceptionMessage('Error occurred when calling preg_replace.');
+        $this->expectExceptionMessageRegExp('~\Q\Preg::replace()\E: Invalid PCRE pattern "": \(code: \d+\) [^(]+ \(version: \d+~');
 
         Preg::replace('', 'foo', 'bar');
     }
@@ -96,7 +199,7 @@ final class PregTest extends TestCase
     public function testReplaceCallbackFailing()
     {
         $this->expectException(PregException::class);
-        $this->expectExceptionMessage('Error occurred when calling preg_replace_callback.');
+        $this->expectExceptionMessage('Preg::replaceCallback(): Invalid PCRE pattern ""');
 
         Preg::replaceCallback('', 'sort', 'foo');
     }
@@ -140,7 +243,7 @@ final class PregTest extends TestCase
     public function testSplitFailing()
     {
         $this->expectException(PregException::class);
-        $this->expectExceptionMessage('Error occurred when calling preg_split.');
+        $this->expectExceptionMessage('Preg::split(): Invalid PCRE pattern ""');
 
         Preg::split('', 'foo');
     }


ktomk:
There is an issue w/ the check by fabbot.io, I've seen changes in the file that were explicitly made to \array_map (with leading slash) so I did it the same and the cs-fixer in the project itself did not suggest any changes either. So I'm a little bit puzzled what is wanted here.

I must admit, I was even before filing the PR b/c other global function in the file like preg_match or preg_last_error etc. aren't prefixed w/ a slash.

julienfalque:
Isn't this going to add too much overhead (verifying each pattern _every_ time it is used)? IMO, covering code with tests should be enough to prevent wrong pattern usages.

src/Preg.php
```diff
@@ -35,6 +64,8 @@ final class Preg
      */
     public static function match($pattern, $subject, &$matches = null, $flags = 0, $offset = 0)
     {
+        $pattern = (string) self::verify($pattern);
```
kubawerlos:
```suggestion
        self::verify($pattern);
```

kubawerlos: Wouldn't that ⬆️ be enough?

src/Preg.php
```diff
@@ -22,6 +22,35 @@
...
+     * @param string|string[] $pattern
+     *
+     * @return string|string[]
+     */
+    public static function verify($pattern)
+    {
+        if (\is_array($pattern)) {
+            return \array_map(__METHOD__, $pattern);
+        }
+
```
kubawerlos: Can we have an early return if we already verified the pattern to avoid the too much overhead issue mentioned by @julienfalque ? Something like:
```
if (isset(self::$verifiedPatterns[$pattern])) {
    return;
}
```

src/Preg.php
```diff
@@ -22,6 +22,35 @@
...
+            return $pattern;
+        }
+        $next = error_get_last();
+        if ($last !== $next) {
+            $message = sprintf('(code: %d) %s', preg_last_error(), $next['message']);
+        } else {
+            $message = sprintf('(code: %d)', preg_last_error());
+        }
+
+        throw new PregException(
```
keradus: i kind of agree with @julienfalque here.
what is a benefit of manual validation rather than letting the `preg_*` validate during normal flow?

in both cases, we throw the `PregException`.

src/Preg.php
ktomk: @kubawerlos If you have static code analysis turned on in the IDE, you will get the notice that it's ambiguous (array is not a string). The cast to string makes the intended use clear (and runtime will give any nice array to string conversion warning if used in error).

src/Preg.php
ktomk: @kubawerlos We certainly are able to write that code. But I would like to mention that we should not.

I could now quote from Rob Pike 1989, however even more concretely from top of my head, the PCRE library already caches compiled patterns internally. At least this is what I remember from years ago when I wrote that CSS parser related lexer / parser thingy years ago you can still find on my Github page. It's PCRE regex based.,

/cc @julienfalque

src/Preg.php
ktomk: Yes, on that level it looks nearly identical. The main reason I tried to highlight in the commit message by writing:

> as getting details of any pattern related errors later on is not easily possible b/c of multi-
matching w/ and w/o UTF-8 mode (see b29848e).

Let me write the little story here:

So when I was executing the CS Fixer, I just was getting PregExceptions which resulted  that the fixer had failures. I could not see which kind of error there was in specific.

And as usual, you get these errors the time you don't need them.

As I'm using the source version I was able to find out w/ step debugging (which was hard as I had the issue w/ PHP 7.4 and my local Xdebug  version with Exception breakpoints in Phpstorm IDE did crash fully, so much for debugging this quick then).

It turned out that it was an invalid pattern. So it would have been good to me for knowing this, but the exact error information was hidden.

Maybe I was too inexperienced and I should have activated some scream first as the the calls suppress warnings (for good reason as two patterns are tested (by the match) for one pattern each).

I first tried to gain the additional information at the place where the exception is thrown (at the end of the function body) as I finally found the spot. And I thought how cruel, throwing an exception with no detailed error message. What would be better?

And then reading the code above I started to realize what made it hard so far to provide a detailed message. It would have certainly required to collect the error information on multiple places. If the pattern is valid first and then invalid this would be good to know, but certainly would introduce many changes, and in the end the issue was with the pattern provided via the parameter.

Now with the suggested changes applied on my box locally I was directly getting the error message and could address the problem.

That detailed error message did allow me in my case to easily spot the bug report in the redhat bugtracker which reminded me I need to update my src version that I had in use as there was a patch already applied and my version was just stale.

In this case I thought, wouldn't it be useful if for project collaborators such errors would be more visible upfront? As the work was already done I just filed this PR.

So this is just the story.

Replacement inputs are not checked (that is similarly possible, too).

So to finally answer your question. This was the benefit to me. A benefit in general is to fail early perhaps, as the pattern is passed stringly typed this is perhaps worth, but I don't need this change upstream. I thought it would be better to have these error messages upfront, but maybe having some contribution / development guidelines that show how to have the application creaming all over the place might have already sufficed as well w/o any further code changes and it's just me who typed too much.

ktomk:
> Isn't this going to add too much overhead (verifying each pattern _every_ time it is used)? IMO, covering code with tests should be enough to prevent wrong pattern usages.

For the overhead: I've not measured. In the past this never has been an issue for me, but I'm eager to learn.

About tests: The pattern was wrong in a specific PHP version. It was not wrong earlier. Unit tests aren't effective at run-time (when testing you normally favor a better message as well).

src/Preg.php
kubawerlos: > But I would like to mention that we should not.

@ktomk and I still stand on my opinion that we should, please do quote from Rob Pike 1989 because I am eager to learn what is wrong with an early return, which we are using it this project a lot.

src/Preg.php
kubawerlos: @ktomk so static code analysis turned on in the IDE would have problem that `$pattern` can be a string, but not that "cannot cast array<string>|string to string" (this is what PHPStan reports on that casting)? Not the best code analysis IMHO.

But that's not what I meant here - I've suggested to remove the assignment as the return value would always be equal to argument.

src/Preg.php
ktomk: @kubawerlos Don't get me wrong, I'm for early reaturns very often, if not always, for the structure of a method. The verification calls are at the top of each of those methods and for the error case they throw. This means the exit is pretty early.

Your suggestion as I understand it is to introduce an in-memory caching layer in form of an array for performance reasons which have not been demonstrated to the best of my current knowledge.

I told in a comment below that I have not measured the change. I know this is sort of weak from my end. This is purely because I don't know where the perf suite of the project is to do these measurements and I'm eager to learn more here.

As you have asked for the quote, it's mainly the rules 1+2 from the complexity section that sprung to my mind, even if not a C programmer I think it's worth to read the whole document which is not very large, but here is a limited quote:

> ...
> Complexity
>      Most programs are too complicated - that is, more complex than they need to be to solve their problems efficiently.  Why? Mostly it's because of bad design, but I will skip that issue here because it's a big one.  But programs are often complicated at the microscopic level, and that is something I can address here.
>      Rule 1.  You can't tell where a program is going to spend its time.  Bottlenecks occur in surprising places, so don't try to second guess and put in a speed hack until you've proven that's where the bottleneck is.
>     Rule 2.  Measure.  Don't tune for speed until you've measured, and even then don't unless one part of the code overwhelms the rest.
> ...

From: [Notes on Programming in C by Rob Pike; February 21, 1989](https://www.lysator.liu.se/c/pikestyle.html)

Even Pike write _rules_, please mind that this is in context of his writing. I do not want to say that these are rules in this project that I think should be followed, it is just that in my opinion this is generally good advice. And it's more generic.

In this case especially  - as I already wrote elsewhere - I think that PCRE caches the compiled patterns already.

So what I'm trying to say is, that if there is a performance issue, we need to find out first to address it. So if you have more data at hand, I really would love to here about it and if this is an issue which needs addressing I'm the last who says to not address it. But let's see if this is an issue first of all.

src/Preg.php
ktomk: @kubawerlos The static analysis I was referring to is Phpstorm EAP w/ EA Inspections, I introduced it in there b/c I got it marked. I personally have no problem to not have it, I thought it might be useful.

And please excuse my ignorance, I have not seen that Phpstan is available in the project, it should be no problem for me to have it run before pushing commits. If you can help me getting a quick start like which script to call in the project, would be great.

kubawerlos: It's not in project yet, currently on PR, I remembered about this as I have PHPStan complaining about the casting a lot in some legacy project I was adding it to - for me personally casting is a last resort - if I cannot make it better.

src/Preg.php
kubawerlos: This in-memory caching with single `if` would save a lot of calls for `preg_` functions - and the less calls to `preg_` function is worth to have it.

src/Preg.php
ktomk: @kubawerlos You raise a valid point here, just tried it out w/ Phpstan in the project configuration. It needs level 6 to trigger the issue and the overall code-base is full of it. Let the *Preg* class alone. This requires so much rewriting I'm pretty sure this is out of scope for this PR. However as I have it already running, let me know if you like to have such kind of changes in this PR, this is pretty straight forward to do w/ Phpstan, so just let me know.

src/Preg.php
ktomk: @kubawerlos Even if so, which different would it make? Is this worth to consider at all?

src/Preg.php
ktomk: @kubawerlos I just update the PR to have phpstan on level 7 (max AFAIK) pass for the new lines of code. This resolves the string cast issue. For the benefit of it only requires an array cast at the other places. Good hint here.

Still interested if you want to see fixes based on Phpstan error reporting for the overall PR, as written this is easy to file, but I want to know of this is really needed for acceptance here.

src/Preg.php
keradus: I see the value of exposing more details on exception.
I don't see a need to validate regex twice - one on explicit validation, one on internal validation. For me, it's like executing same action twice.
What's more, if John Doe would drop the pre-validation, outcome of the execution would be the same.

src/Preg.php
```diff
@@ -87,12 +91,14 @@ public static function matchAll($pattern, $subject, &$matches = null, $flags = P
      */
     public static function replace($pattern, $replacement, $subject, $limit = -1, &$count = null)
     {
-        $result = @preg_replace(self::addUtf8Modifier($pattern), $replacement, $subject, $limit, $count);
+        $patterns = self::verifyPatterns((array) $pattern);
```
keradus: if we don't provide an array, please not convert it to array. verifyPatterns can accept string|string[], like all other methods

src/Preg.php
```diff
@@ -180,10 +190,45 @@ private static function removeUtf8Modifier($pattern)
...
     }
+
+    /**
+     * @param string[] $patterns
+     *
+     * @return string[]
+     */
+    private static function verifyPatterns(array $patterns)
+    {
+        return \array_map('PhpCsFixer\Preg::verify', $patterns);
```
keradus: do not call method by string names.
one will rename method name and this reference will not be picked up by static code analysis/refactor tool

tests/PregTest.php
```diff
@@ -201,4 +202,26 @@ public function testCorrectnessForNonUtf8String()
         static::assertSame([\chr(224)], $methodMatches);
         static::assertNotSame([\chr(224)], $functionMatches);
     }
+
+    /**
+     * Preg class is final, to test these internals reflection is necessary.
```
keradus: don't test internals.
it's bad practice.

tests/PregTest.php
```diff
@@ -158,7 +159,7 @@ public function testSplitFailing()
             'PhpCsFixer\\PregException'
         );
         $this->expectExceptionMessage(
-            'Error occurred when calling preg_split.'
+            'Invalid PCRE pattern ""'
```
keradus: can we have an example with sth else than `""` as well ?

src/Preg.php
ktomk: I think this was an oversight I had in the first push, I changed the visibility to private. Let me double check. Yes, they are private already. Also moved to the end of the class. Refreshing on top of 2.12 now.

src/Preg.php
```diff
@@ -35,6 +35,8 @@ final class Preg
      */
     public static function match($pattern, $subject, &$matches = null, $flags = 0, $offset = 0)
     {
+        $pattern = self::verify($pattern);
```
kubawerlos: I still don't understand why is the assignment here - simply `self::verify($pattern);` would work the same.

ktomk: @kubawerlos What would help you to gain more understanding? A method does not represent a process result as much as a function that returns (spoiler: a method has no result). What is there to add? Should we go more abstract into math/logic here? What would help you?

tests/PregTest.php
ktomk: @keradus This is the original fixture IIRC. I'll take a look to put this more specifically under test w/ some more fixtures maybe on Preg::match specifically, I assume you're looking for more examples here which I think is a valid point.

src/Preg.php
ktomk: @keradus You only say what not to do, but you don't give an example what should be used instead. This is not helpful for me. How is it possible to not  use a string for the method name?

tests/PregTest.php
ktomk: Sure. Now the internals are publicly tested. Equally bad practice, but not that visible. Also a benefit of more invalid pattern examples. See as well your other comment about the fixtures.

src/Preg.php
keradus: i'm not talking about visibility of the method - i'm talking about fact that we in fact validate twice now

src/Preg.php
keradus: sorry, i remembered this topic was raised few times already and wrongly assumed that it was only oversight here.

```
grep -r array_ src -A5
src/Console/ConfigurationResolver.php:        $paths = array_filter(array_map(
src/Console/ConfigurationResolver.php-            static function ($path) {
src/Console/ConfigurationResolver.php-                return realpath($path);
src/Console/ConfigurationResolver.php-            },
src/Console/ConfigurationResolver.php-            $this->getPath()
src/Console/ConfigurationResolver.php-        ));
```

instead of `$paths = array_filter("realpath", $this->getPath());`

tests/PregTest.php
keradus: I am looking for example that would present the value of changes you are doing - now, we know it's there, but without explicit utest of it, it can be broken accidentally in future.
I understand you had to adjust existing tests to pass changed code.
It would be also great to add new cases to cover newly added functionality (even if the functionality is "only" better error messages)

keradus:
@ktomk , please, for a while, during the review, avoid doing the squashing.
it's really hard to see the changes you applied between my previous review and now, as everything is a single commit now :(

src/Preg.php
keradus: it was nothing about complexity, math or psalm.

I believe Kuba suggested this change:
```diff
-private static function verify($pattern): string
+private static function verify($pattern): void
```
as the only outcome of validation method is telling if input is valid or not, which `verify` method does by throwing the exception or not. no need to return anything

src/Preg.php
kubawerlos: I don't know, it would be great to see some benchmarks with no `verify`, with cached version and with non-cached version.

src/Preg.php
ktomk: @keradus This did already validate twice on failure before the change-set. So actually this validates only once in case of error (with all due respect: which if I understand you correctly would be in your benefit contrary to what you argue against).

However this could be a resolution. For all of those methods (IIRC four to six), just the error message function could be put to enrich the error message (or take care of the error signalling in full). This would also spare to have the `new` in all those places. It's sort of just a new idea, I can't say if this works out in the end, but it looks not too idiotic.

It's perhaps worth to make it a static method of the exception itself. Please let me know if you have some preferences here or just feedback in general before I change this. Also I would like to know if you're fine to have any of this static function public, not private.

If I touch the PregException class I would turn it final as well if it is not yet (and it does not break too many test-cases).

src/Preg.php
ktomk: There was a split, verifyPatterns for arrays, verify for strings. I perhaps should more properly type-hint that.

src/Preg.php
ktomk: Alright, I see. This is easy to change now for me. Very helpful, thanks!

src/Preg.php
ktomk: This is (only) true if there is a try-catch block which re-throws. I normally spare that by making use of the return value for the happy-path. I don't say it's perfect, but much more expressive if try-catch is left out as it was suggested.

And don't spare psalm, it's in PR already.

ktomk:
@keradus: If I only had read that earlier. I always rebase / amend locally after the build is fine and then push for the remote build resetting anything previous. I can leave fixup! etc. commits if that helps but was not aware earlier this was inconvenient. I'll do for future pushes.

src/Preg.php
ktomk: Was already type hinted. It's not a conversion but a cast. PHP is very good at this, I see it as less overhead and would really not suggest to change this.

src/Preg.php
ktomk: I update the PR which addresses this. Unfortunately I've ready your comment too late about squashing. Next changes will be separated.

src/Preg.php
kubawerlos: > much more expressive

You meant less expressive. If there is a return value person who sees it expect that this is different from the argument and would go into the function to see - otherwise can assume it's only to verify and does not change the variable.

ktomk: I see this more as a gate the data passes through. If the sub-process starts to misbehave this is more stable and stringent that is visible upfront which makes it more expressive for me, however I do see that there might be different views on that and also about it's expressiveness.

I care more about the data processing here than counting the number of code expressions.

src/Preg.php
ktomk: The verify method is gone for the current changes. I assume this addresses your comments.

src/Preg.php
ktomk: @kubawerlos: That method is gone. Me assumes this addresses your comment.

src/Preg.php
ktomk: This method is now gone, however would you still suggest to not cast before calling the method but preferable doe that kind of processing inside the method?

Even though I assume your comments are addressed due to removal of the method, I'm interested in your opinion as there is a new method now which uses the same pattern.

src/Preg.php
ktomk: @kubawerlos Was afk for some days and then needed to be avail for others elsewhere first, so please excuse the feedback-less time. On topic, just let me know if you've found out. I don't know how things are measured in this project still.

src/Preg.php
ktomk: @keradus Validating any of the pattern(s) is now only done in case the main method already error'ed. This should address your comment. I still don't get the build green though, so I might need some more builds until done.

ktomk:
I have added some changes to this PR. The verification of the patterns is now only done if any of the preg_* operations failed regardless in which unicode/non-unicode mode.

In case verification fails, the **PregException** message has details of the error. Otherwise the **PregException** message now contains the method name of the **Preg** class method which caused the error (it had the name of the preg_* method previously) and the behavior is otherwise unchanged.

Then I added another change-set on top as the Travis build was giving me a hard time due to installation failures of the dev-tools (for example see [build #18669
](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/581634045). This change is marked with three dots at the beginning as I needed a couple of retries to compile it. Let me know what to do with it, e.g. if it should be kept in this PR or moved into a PR of it's own or what else.

keradus:
@ktomk,

As discussed on https://gitter.im/PHP-CS-Fixer/Lobby:
> https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4517/commits/012026115ea4e5ab13f8f16fe537e7d432bc12d1 looks like great improvement !
please, extract it to separated PR (and drop it from current PR, perhaps), so we can have proper review etc. I must admit, i'm not so experienced with shell scripts, so probably i will have few questions about the changes.
also, consider adding the great description from commit message into the shell script file itself, so one can always come back to it easily

also, don't worry about earlier squashes, but now you see, while having atomic commits, it's also easier to extract sth to dedicated PR by simply cherry-picking

----

your description of changes for Preg error handling itself sounds good. I will take a look at code diff hopefully tomorrow

src/Preg.php
```diff
@@ -180,10 +180,52 @@ private static function removeUtf8Modifier($pattern)
...
         return substr($pattern, 0, $endDelimiterPosition).str_replace('u', '', substr($pattern, $endDelimiterPosition));
     }
+
+    /**
+     * Create PregException.
+     *
+     * Create the generic PregException message and if possible due to finding
+     * an invalid pattern, tell more about such kind of error in the message.
+     *
+     * @param $error
```
keradus: types are missing

src/Preg.php
```diff
@@ -180,10 +180,52 @@ private static function removeUtf8Modifier($pattern)
...
+     * Create the generic PregException message and if possible due to finding
+     * an invalid pattern, tell more about such kind of error in the message.
+     *
+     * @param int            $error
+     * @param string         $method
+     * @param array|string[] $patterns
+     *
+     * @return PregException
+     */
+    private static function newPregException($error, $method, array $patterns)
```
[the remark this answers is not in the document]
ktomk: I would not do that. The reason is that on first guide I use the rule to use as little visibility as necessary to get the job done.

As the error handling is also very specific in that routine (won't hurt it matures a bit first would be my guess) and PregException is a utility to Preg, I think for the moment it rests well there.

Albeit what you suggest I also suggested earlier in a comment and I think is possible. But the handling is that specific to **Preg** that I have stepped a bit back from it.

src/Preg.php
```diff
@@ -45,7 +45,7 @@ public static function match($pattern, $subject, &$matches = null, $flags = 0, $
             return $result;
         }
 
-        throw new PregException('Error occurred when calling preg_match.', preg_last_error());
+        throw self::newPregException(preg_last_error(), __METHOD__, (array) $pattern);
```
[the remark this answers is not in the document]
ktomk: We can and I think we also do. However, this call is necessary to obtain the original last error of the preg operation as testing a patterin inside that routine will overwrite it and this parameter is the default exception code which otherwise would be implicit only and at risk to get overwritten due to more preg operations in the subroutine.

Please let me know if this answers your question and if not, please help me to better understand the intend of your question.

src/Preg.php
```diff
@@ -180,10 +180,52 @@ private static function removeUtf8Modifier($pattern)
...
+
+    /**
+     * Create PregException.
+     *
+     * Create the generic PregException message and if possible due to finding
+     * an invalid pattern, tell more about such kind of error in the message.
+     *
+     * @param int            $error
+     * @param string         $method
+     * @param array|string[] $patterns
```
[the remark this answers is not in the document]
ktomk: Please find your comment addressed.

PR #4518, opened 2019-08-24
PhpUnitNoExpectationAnnotationFixer - fix handling expect empty exception message

ktomk:
An edge case I had with an empty @expectedExceptionMessage annotation
uncovered an unchecked return resulting in the undefined offset.

Fix is to fall-back to empty string which also solves the fixer for
me.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
@@ -264,8 +264,12 @@ final class MyTest extends \PHPUnit_Framework_TestCase
     {
         $tag = $annotation->getTag()->getName();
 
-        Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches);
+        if (1 !== Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches)) {
+            return '';
+        }
+
         $content = $matches[1];
+
         if (Preg::match('/\R/u', $content)) {
             $content = Preg::replace('/\s*\R+\s*\*\s*/u', ' ', $content);
         }
--- a/tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
+++ b/tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
@@ -41,6 +41,32 @@ final class PhpUnitNoExpectationAnnotationFixerTest extends AbstractFixerTestCas
     public function provideTestFixCases()
     {
         return [
+            'empty exception message' => [
+                '<?php
+    final class MyTest extends \PHPUnit_Framework_TestCase
+    {
+        /**
+         */
+        public function testFnc()
+        {
+            $this->setExpectedException(\FooException::class, \'\');
+
+            aaa();
+        }
+    }',
+                '<?php
+    final class MyTest extends \PHPUnit_Framework_TestCase
+    {
+        /**
+         * @expectedException FooException
+         * @expectedExceptionMessage
+         */
+        public function testFnc()
+        {
+            aaa();
+        }
+    }',
+            ],
             'expecting exception' => [
                 '<?php
     final class MyTest extends \PHPUnit_Framework_TestCase


src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -263,8 +263,9 @@ private function extractContentFromAnnotation(Annotation $annotation)
     {
         $tag = $annotation->getTag()->getName();
 
-        Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches);
-        $content = $matches[1];
+        $content = Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches)
+            ? $matches[1]
```
kubawerlos: Can we check the `$matches` size and return an empty string for no match? There is no point to assing empty string to `$content` and execute `preg_match` on it.

GrahamCampbell:
Actually, `@expectedExceptionMessage` will do nothing, since it only checks if the message is a prefix of the actual message, so the correct implementation should just not check the message at all in the conversion...

tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
```diff
@@ -41,6 +41,32 @@ public function testFix($expected, $input = null, array $config = [])
...
         return [
+            'empty exception message' => [
+                '<?php
+    final class MyTest extends \PHPUnit_Framework_TestCase
+    {
+        /**
+         */
+        public function testFnc()
+        {
+            $this->setExpectedException(\FooException::class, \'\');
```
GrahamCampbell: Should be just `$this->setExpectedException(\FooException::class);`.

ktomk: @GrahamCampbell I don't know. What I can say is that is must not: it works this way for the code-base I had the error with.

Not dropping that annotation value also allowed me (as a user) to spot later on the problem that originally no message was given. The annotation that is getting transposed here is about looking for that string somewhere in the exception message - this makes hardly any sense with an empty string - so the user might want to be able to learn this its own w/o having the fixer hiding an issue.

So it already originally made no sense and in my case I didn't want to have the error, but it's fine if I can spot later on a useless exception expectation with a different inspection).

Just so it's better understandable why I kept the use-less looking behaviour in.

So please let me know if this lowers your change suggestion.

GrahamCampbell: The previous code with the empty annotation message was pointless, and should have been removed. It's obviously always going to be true that the empty string is a prefix of any string.

Therefore, we should remove the tautology check.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
ktomk: @kubawerlos You mean something like these lines?:

```php
    /**
     * @param Annotation $annotation
     *
     * @return string
     */
    private function extractContentFromAnnotation(Annotation $annotation)
    {
        $tag = $annotation->getTag()->getName();

        if (
            Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches)
            && Preg::match('/\R/u', $matches[1])
        ) {
            return rtrim(Preg::replace('/\s*\R+\s*\*\s*/u', ' ', $matches[1]));
        }

        return '';
    }
```

It does not set the variable at all.

tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
ktomk: Well then it is perhaps better to let the error in so that the user can actually notice the problem as CS-Fixer fails. This is the way it helped me actually. I don't like to remove a program flaw that hides away other issues. When it was the last line of defence, then it should be kept albeit the reporting is not very precise.

GrahamCampbell: I'm just saying that when this happens, the generated code should not include the quotes. Just setup the code with only one argument, not two.

GrahamCampbell: Like I said, I think the generated code should be:

```
$this->setExpectedException(\FooException::class);
```

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
kubawerlos: @ktomk I meant like this:
```php
    private function extractContentFromAnnotation(Annotation $annotation)
    {
        $tag = $annotation->getTag()->getName();

        if (1 !== Preg::match('/@'.$tag.'\s+(.+)$/s', $annotation->getContent(), $matches)) {
            return '';
        }

        $content = $matches[1];

        if (Preg::match('/\R/u', $matches[1])) {
            return Preg::replace('/\s*\R+\s*\*\s*/u', ' ', $matches[1]);
        }

        return rtrim($content);
    }
```

Also, I wouldn't call it refactoring, I think of it as Boy Scout Rule.

tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
GrahamCampbell: https://github.com/ktomk/PHP-CS-Fixer/pull/1

ktomk: Like I wrote, I see the benefit in keeping the error message then.I don't want to be the person who introduces the burden then for others to see their problem places getting hidden by a fixer that should just port.

But maybe there is a solution: What I asked myself is, is there any possibility that a fixer can signal problematic spots for the user? I also have other fixer-fails situations which were all good to know about, like with these not given (empty) exception messages which signal a problem (the message has been forgotten for example) a warning with a problem description (text), file and line, perhaps offset, could be given?

I would be very fine w/ not having the error in PHP-CS-Fixer and perhaps also trying best to fix the code-issue (like you want it) but to signal the user that a fishy spot has been detected.

This would also allow me to suggest more problem places I was running over, which are of similar matter. Having PHP-CS-Fixer failing with errors helped me to spot the places, but a better location which code triggered an issue would have helped me a lot.

But this might be bike-shedding and very much driven from my concrete needs due to the issues I was running over, so it's more the question if something like such a reporting facility already exists.

ktomk:
Just some more info: If the exception message is set empty to asset for / an empty message is set as expectation Phpunit (in practise at least in one version) is looking for an empty message:

~~~
2) Ktomk\Pipelines\Runner\DirectoriesTest::testGetBaseDirectoryThrows
Failed asserting that exception message is empty but is 'Not a base directory: 'XDG_FOO42_HOME'. Known base directories are: "XDG_DATA_HOME", "XDG_CACHE_HOME"'.
~~~

Details about the Phpunit version:

~~~
PHPUnit 6.5.7 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.22-1+ubuntu18.04.1+deb.sury.org+1 with Xdebug 2.7.2
~~~

Just for the record.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
SpacePossum: looks better IMHO, I pushed the change in

SpacePossum:
The discussion between;

>$this->setExpectedException(\FooException::class, '');

and 

>$this->setExpectedException(\FooException::class);

seems semantic to me. This is because `''` is the default value for the message (and not `null` as the PHPDoc incorrectly states).

Changing to:

>$this->setExpectedException(\FooException::class, null);

seems more correct as when the annotation `@expectedExceptionMessage` has no value than the message itself is not checked, like with passing `null`. However, passing `null` can be valid, I can't see why someone would omit the value for the annotation.

So fixing to `''` seems fine to me, people can update their code to use `null` if needed or revert the change and remove useless annotation and fix the project again.

Besides all this, it seems `setExpectedException` is removed from [PHPUnit](https://github.com/sebastianbergmann/phpunit/commit/9b3f36158ce4bb00425b0d30e8b99cd3c18cd2db)  ....

SpacePossum:
Thank you @ktomk.

ktomk:
@SpacePossum: You're welcome. I'm glad I could be of help.

Issue #4520, opened 2019-08-26
Wrong strings for string fix in php_unit_no_expectation_annotation fixer


ktomk:
While making use of the php_unit_no_expectation_annotation fixer recently, I stumbled over a little flaw when strings are used for class-names: All strings are wrongly generated with a leading backslash character (`\`). Correct is to not have it.

This is also PHP's behaviour, when outputting a class in global or another namespace for the PHP versions that support the ::class constant:

Example:
```php
<?php namespace Foo\Bar;
class Baz {};
var_dump(Baz::class);
```
Output:
```
string(11) "Foo\Bar\Baz"
```
(find [this example on 3v4l](https://3v4l.org/DJiV4))

Which is also a correct FQCN. However it might have been that for annotations (which are getting converted here) it is of use in IDEs which do fully resolve any class-name from annotations to the namespace of the file, so a leading back-slash is sometimes added in front of 
FQCNs in annotations for code editing convenience reasons.

As the fixer changes into a string parameter, this is not necessary at all as class-names expressed as strings always need to point to the class in the global namespace, hence never require a leading backslash.

For the conversions I do I current have a regular expression based fixer after PHP-CS-Fixer to workaround this issue temporarily however as I'm even fiddled with this fixer I'm open to provide a pull request which addresses this issue. It will just change from backslash-ed to the correct string.

Example:

Input:

```php
<?php
    final class MyTest extends \PHPUnit_Framework_TestCase
    {
        /**
         * @expectedException FooException
         */
        public function testFnc()
        {
            aaa();
        }
    }
```

Output:

```php
<?php
    final class MyTest extends \PHPUnit_Framework_TestCase
    {
        /**
         */
        public function testFnc()
        {
            $this->setExpectedException('\FooException');

            aaa();
        }
    }
```

This issue basically suggests to remove the leading `\` from the string in:

```php
$this->setExpectedException('\FooException');
                             ^
```
to get into a correct FQCN:

```php
$this->setExpectedException('FooException');
                            ^^
```

Let me know if you see any issues with this as otherwise I might just file a PR.

GrahamCampbell:
Hmm, yes. I'd say that FQCNs as string literals should not have a leading slash.

GrahamCampbell:
I'd expect `\Foo::class` -> `'Foo'`.

Issue #4068, opened 2018-10-25
Import classes instead using FQCN with `php_unit_no_expectation_annotation` fixer


emodric:
#### The PHP version you are using (`$ php -v`):

```
PHP 7.2.11-2+ubuntu18.04.1+deb.sury.org+1 (cli) (built: Oct 15 2018 11:40:35) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Zend OPcache v7.2.11-2+ubuntu18.04.1+deb.sury.org+1, Copyright (c) 1999-2018, by Zend Technologies
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):

```
PHP CS Fixer 2.13.0 Yogi's BBQ by Fabien Potencier and Dariusz Ruminski (7136aa4)
```

#### The command you use to run PHP CS Fixer:

```
php-cs-fixer fix
```

#### The configuration file you are using, if any:

```php
<?php

return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setRules([
        'php_unit_no_expectation_annotation' => true,
    ])
    ->setFinder(
        PhpCsFixer\Finder::create()
            ->in(__DIR__)
    )
;
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

declare(strict_types=1);

namespace MyApp\Tests;

use PHPUnit\Framework\TestCase;
use MyApp\MyException;

final class MyTest extends TestCase
{
    /**
     * @expectedException \MyApp\MyException
     * @expectedExceptionMessage Test message.
     */
    public function testThrowsException(): void
    {
        throw new MyException;
    }
}
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php

declare(strict_types=1);

namespace MyApp\Tests;

use PHPUnit\Framework\TestCase;
use MyApp\MyException;

final class MyTest extends TestCase
{
    /**
     */
    public function testThrowsException(): void
    {
        $this->setExpectedException(\MyApp\MyException::class, 'Test message.');

        throw new MyException;
    }
}
```

 * with the changes you expected instead:
```php
<?php

declare(strict_types=1);

namespace MyApp\Tests;

use PHPUnit\Framework\TestCase;
use MyApp\MyException;

[...]

PR #4521, opened 2019-08-27
Remove superfluous leading backslash, closes 4520

ktomk:
In string context, there is no need to prefix the exception class-name
of the exception with any backslashes as these are always in the global
name-space.

Looks like the prefix is only useful for verbatim class names which is not
addressed in this issue (see #4068 for more details on verbatim class-
names and also a generic resolution suggestion for many fixers).

Additionally minor changes regarding standard (Phpstorm) inspections.

Closes: #4520
Related: #4068

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
@@ -245,9 +245,10 @@ final class MyTest extends \PHPUnit_Framework_TestCase
             $tokens[$docBlockIndex] = new Token([T_DOC_COMMENT, $doc->getContent()]);
             $tokens->insertAt($braceIndex + 1, $newMethods);
 
-            $tokens[$braceIndex + $newMethods->getSize() + 1] = new Token([
+            $whitespaceIndex = $braceIndex + $newMethods->getSize() + 1;
+            $tokens[$whitespaceIndex] = new Token([
                 T_WHITESPACE,
-                $this->whitespacesConfig->getLineEnding().$tokens[$braceIndex + $newMethods->getSize() + 1]->getContent(),
+                $this->whitespacesConfig->getLineEnding().$tokens[$whitespaceIndex]->getContent(),
             ]);
 
             $i = $docBlockIndex;
@@ -275,10 +276,10 @@ final class MyTest extends \PHPUnit_Framework_TestCase
     private function annotationsToParamList(array $annotations)
     {
         $params = [];
-        $exceptionClass = '\\'.ltrim($annotations['expectedException'], '\\');
+        $exceptionClass = ltrim($annotations['expectedException'], '\\');
 
         if ($this->configuration['use_class_const']) {
-            $params[] = $exceptionClass.'::class';
+            $params[] = "\\{$exceptionClass}::class";
         } else {
             $params[] = "'{$exceptionClass}'";
         }
--- a/tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
+++ b/tests/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixerTest.php
@@ -221,7 +221,7 @@ final class PhpUnitNoExpectationAnnotationFixerTest extends AbstractFixerTestCas
          */
         public function testFnc()
         {
-            $this->setExpectedException(\'\FooException\');
+            $this->setExpectedException(\'FooException\');
 
             aaa();
         }


ktomk:
Next to the original request, I again need assistance w/ the fabbot. It paints the build red however, the code changes suggested are not suggested by project QA tools.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -183,6 +184,7 @@ private function fixPhpUnitClass(Tokens $tokens, $startIndex, $endIndex)
         $tokensAnalyzer = new TokensAnalyzer($tokens);
 
         for ($i = $endIndex - 1; $i > $startIndex; --$i) {
+            // @noinspection NotOptimalIfConditionsInspection
```
julienfalque: :-1: for adding tools instructions in the code.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -130,7 +131,7 @@ public function isRisky()
     /**
      * {@inheritdoc}
      */
-    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
```
julienfalque: Please revert unrelated changes. They should be applied globally in dedicated PRs.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -245,9 +247,10 @@ private function fixPhpUnitClass(Tokens $tokens, $startIndex, $endIndex)
             $tokens[$docBlockIndex] = new Token([T_DOC_COMMENT, $doc->getContent()]);
             $tokens->insertAt($braceIndex + 1, $newMethods);
 
-            $tokens[$braceIndex + $newMethods->getSize() + 1] = new Token([
+            $i1 = $braceIndex + $newMethods->getSize() + 1;
```
julienfalque: Please rename the variable (e.g. `$whitespaceIndex`), `$i1` does not carry any useful information about the data is holds.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
ktomk: Right, this does not look good for the code, removed.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
ktomk: Right, even used in a loop, all the other indexes are more verbosely named, it was only the $i below which might have irritated me a little.

Please consider your comment addressed and thanks for your thoughtful direction.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
ktomk: This must have been my muscle memory on the spot, you say it totally right this change has nothing at all to do with the issue at hand and has been removed from the change-set.

Please consider this comment addressed.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -275,12 +276,12 @@ private function extractContentFromAnnotation(Annotation $annotation)
...
         $params = [];
-        $exceptionClass = '\\'.ltrim($annotations['expectedException'], '\\');
+        $exceptionClass = ltrim($annotations['expectedException'], '\\');
 
         if ($this->configuration['use_class_const']) {
-            $params[] = $exceptionClass.'::class';
+            $params[] = '\\'.$exceptionClass.'::class';
         } else {
-            $params[] = "'{$exceptionClass}'";
+            $params[] = "'${exceptionClass}'";
```
julienfalque: This change should be reverted as well.

ktomk: @julienfalque This is actually what PHP-CS-Fixer applied after I have changed that line. I also was puzzled by this, could it be that the fixer suggests a different bracket format if there is no bracket format in use?

I removed it from the change-set in any case, so please consider your comment addressed.

src/Fixer/PhpUnit/PhpUnitNoExpectationAnnotationFixer.php
```diff
@@ -275,10 +276,10 @@ private function extractContentFromAnnotation(Annotation $annotation)
     private function annotationsToParamList(array $annotations)
     {
         $params = [];
-        $exceptionClass = '\\'.ltrim($annotations['expectedException'], '\\');
+        $exceptionClass = ltrim($annotations['expectedException'], '\\');
 
         if ($this->configuration['use_class_const']) {
-            $params[] = $exceptionClass.'::class';
+            $params[] = '\\'.$exceptionClass.'::class';
```
kubawerlos: Can we have `sprintf` here?

julienfalque: I would even write `"\\{$exceptionClass}::class"`.

ktomk: If you can find a resolution would be great. I can't write both in there at the same time.

IMHO neither of the three is of any difference. The projects PHP-CS-Fixer configuration suggests `"\\${exceptionClass}::class"`for string interpolation, but only unless the variant @julienfalque suggested is used.

So if you come to a conclusion, I'm happy to file it:

* A; `'a'.$b.'c'` style (as originally filed) 
* B: sprintf() call
* C: `"{$a}efgh"` style
* D: `"${a}efgh"` style

From the existing style in the file I prefer C, but please let me know which is the general preference in the project as I think this is worth to have at hand directly.

ktomk: Myself concludes as even after a day, no result of the poll was given by others, that the resolution C is acceptable for this PR.

Thanks a lot for putting your efforts into the review, I really like to have multiple reviewers on a pull request and I know that this is nothing to expect.

SpacePossum:
Thank you @ktomk.

ktomk:
@SpacePossum: You're welcome. I'm glad I could be of help.

PR #4526, opened 2019-08-27
no need to merge

ktomk:
this commit is empty to trigger the build on 2.12 head for any upcomming
pull-request.


keradus:
out of the curiosity, what was a need for this action?
(maybe I can optimize sth on our side)

ktomk:
@keradus  Thanks for asking, but it was just me not seeing the woods for the tree. For the bigger picture: If a fixer fails *after* fixing due to PHP syntax errors (in the liniting part), it would be great for better error reporting. I can imagine this is not easy, so this is really a bigger picture. What I did here - just for explanation - was to fire the build against the base branch b/c I couldn't understand why the build with my changes against the build branch failed. After I've seen it green it was just a  blunder I had in my PR. So sorry for the interruption.

PR #4531, opened 2019-08-30
PhpdocAlignFixer - add "property-read" and "property-write" to allowed tags

### LINKED ISSUE #4530: Suggestion: phpdoc_align should support `@property-read`
Currently, `phpdoc_align` doesn't have support for [@property-read](https://docs.phpdoc.org/references/phpdoc/tags/property-read.html)

Trying to configure it, e.g. `'phpdoc_align' => [ 'tags' => [ 'property', 'property-read']]` causes an `InvalidOptionsException` to be thrown.

I suggest adding `property-read` here https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/37cc9c62f71a4cd49d9427aea729954f0ca35fe2/src/Fixer/Phpdoc/PhpdocAlignFixer.php#L62-L70 which should be backward compatible as long as the default config value is not altered.

kubawerlos:
Resolves #4530

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1509,9 +1509,10 @@ Choose from the list of available rules:
   Configuration options:
 
   - ``align`` (``'left'``, ``'vertical'``): align comments; defaults to ``'vertical'``
-  - ``tags`` (a subset of ``['param', 'property', 'return', 'throws', 'type',
-    'var', 'method']``): the tags that should be aligned; defaults to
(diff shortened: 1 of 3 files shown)


julienfalque:
Shouldn't this target `master`?

SpacePossum:
Thank you @kubawerlos.

Issue #4529, opened 2019-08-30
PHP 7.4 features invalidating PHP 7.2 code


LarsGrevelink:
By merging #4380 PHP 7.4 functionality seems to have entered the realm of existing rules which are applied to existing projects. If I understand the code changes correctly there should be a check whether the function exists but for some reason that is not respected. I created a small test case to verify this.

We're running on Docker PHP 7.2 environments so nothing too fancy there.

Are we misconfiguring our tools somewhere or should this be fixed on the CS Fixer side?

Cheers,
L

**PHP version:** 7.2.19
**PHP CS Fixer version:** PHP CS Fixer 2.15.2 Europe Round by Fabien Potencier and Dariusz Ruminski (c9d30fd)

**Command used;**
```
./php-cs-fixer fix --config=.php_cs test.php
```

**Configuration used;**
```php
<?php

return PhpCsFixer\Config::create()->setRules([
    'mb_str_functions' => true,
])->setLineEnding("\n")->setRiskyAllowed(true);
```

**Before running PHP CS Fixer (no changes);**
```php
<?php

var_dump(function_exists('mb_str_split'));
// bool(false)

str_split('splitting the string');
```

**With unexpected changes applied when running PHP CS Fixer;**
```php
<?php

var_dump(function_exists('mb_str_split'));
// bool(false)

mb_str_split('splitting the string');
// Fatal error: Uncaught Error: Call to undefined function mb_str_split() in /src/test.php on line 5
```

**With the changes you expected instead;**
```php
<?php

var_dump(function_exists('mb_str_split'));
// bool(false)

str_split('splitting the string');
```

LarsGrevelink:
Might be good to know that we're retrieving it via;

```
$ curl -L https://cs.symfony.com/download/php-cs-fixer-v2.phar -o php-cs-fixer
```

weshooper:
Came here to report the same issue. We're running PHP 7.3.6.

keradus:
looks like we need to revert #4380 , thanks for the report

LarsGrevelink:
@keradus It was still scratching my head since the code was very solid which was added. Unfortunately, a dependency used by the project adds a polyfill. This is why for the validator is always seems to exist even though it isn't.

Dependency: `symfony/console` which requires `symfony/polyfill-mbstring`

Code which causes the eventual issue;
[...]

PR #4534, opened 2019-08-31
MbStrFunctionsFixer - Add "str_split" => "mb_str_split" mapping [PHP7.4]

keradus:
This reverts commit d35c431600a927db0620a40ae2c70e8681c9e8f4.

Reapply #4380 by reverting #4533

### DIFF
--- a/src/Fixer/Alias/MbStrFunctionsFixer.php
+++ b/src/Fixer/Alias/MbStrFunctionsFixer.php
@@ -55,7 +55,7 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
         $this->functions = array_filter(
             self::$functionsMap,
             static function (array $mapping) {
-                return \function_exists($mapping['alternativeName']);
+                return \function_exists($mapping['alternativeName']) && (new \ReflectionFunction($mapping['alternativeName']))->isInternal();
             }
         );
     }
--- a/src/Fixer/Alias/MbStrFunctionsFixer.php
+++ b/src/Fixer/Alias/MbStrFunctionsFixer.php
@@ -28,6 +28,7 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
      * @var array the list of the string-related function names and their mb_ equivalent
      */
     private static $functionsMap = [
+        'str_split' => ['alternativeName' => 'mb_str_split', 'argumentCount' => [1, 2, 3]],
         'stripos' => ['alternativeName' => 'mb_stripos', 'argumentCount' => [2, 3]],
         'stristr' => ['alternativeName' => 'mb_stristr', 'argumentCount' => [2, 3]],
         'strlen' => ['alternativeName' => 'mb_strlen', 'argumentCount' => [1]],
@@ -41,6 +42,23 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
         'substr' => ['alternativeName' => 'mb_substr', 'argumentCount' => [2, 3]],
         'substr_count' => ['alternativeName' => 'mb_substr_count', 'argumentCount' => [2, 3, 4]],
     ];
+
+    /**
+     * @var array<string, array>
+     */
+    private $functions;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->functions = array_filter(
+            self::$functionsMap,
+            static function (array $mapping) {
+                return \function_exists($mapping['alternativeName']);
+            }
+        );
+    }
 
     /**
      * {@inheritdoc}
@@ -86,7 +104,7 @@ $a = substr_count($a, $b);
     protected function applyFix(\SplFileInfo $file, Tokens $tokens)
     {
         $argumentsAnalyzer = new ArgumentsAnalyzer();
-        foreach (self::$functionsMap as $functionIdentity => $functionReplacement) {
+        foreach ($this->functions as $functionIdentity => $functionReplacement) {
             $currIndex = 0;
             while (null !== $currIndex) {
                 // try getting function reference and translate boundaries for humans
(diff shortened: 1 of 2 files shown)


src/Fixer/Alias/MbStrFunctionsFixer.php
```diff
@@ -42,6 +43,23 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
...
+    private $functions;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->functions = array_filter(
+            self::$functionsMap,
+            static function (array $mapping) {
+                return \function_exists($mapping['alternativeName']);
```
keradus: this is work way to do the check. during runtime, we have ext-mb polyfilled, thus we alternativeName always exists

vide https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4380#discussion_r319718704
vide #4529

Slamdunk: [ReflectionFunctionAbstract::isInternal](https://www.php.net/manual/en/reflectionfunctionabstract.isinternal.php) can help: https://3v4l.org/Wdp27

keradus: Looking nice!
Yet, to be honest, i'm not sure is it the way we want to follow - implicit handling of decision "should we transform this function or not" by detecting `isInternal` or current runtime (that still can be different than end environment), or explicit configuration (like we do eg for PHPUnit - we aim version X or Y, regardless what is installed in repo we are fixing).
What do you think, @Slamdunk ?

tests/Fixer/Alias/MbStrFunctionsFixerTest.php
```diff
@@ -62,5 +62,18 @@ public function strtolower($a);
                 }',
             ],
         ];
+
+        if (\function_exists('mb_str_split')) {
```
Slamdunk: Tests shall be as much static as possible: regardless of the ongoing topic about how to fix this in the Fixer, the test should check the PHP version instead of existing functionality. I repeat: the test should do this, not what the test tests. /cc @SpacePossum

src/Fixer/Alias/MbStrFunctionsFixer.php
Slamdunk: I'm gonna talk about this particular Fixer only: this Fixer aims to not forget anywhere any non-multi-byte string comparison. Why should anyone could want to fix one function but not the others?

> by detecting `isInternal` or current runtime (that still can be different than end environment)

That would be a user's fault: this tool it is intended to be run with the same binary used by the production, so no worry to me about this point

> explicit configuration (like we do eg for PHPUnit - we aim version X or Y, regardless what is installed in repo we are fixing)

As a user myself of PHP-CS-Fixer the fact that PHPUnit related fixers are by explicit configuration represents just a lack of the tool, not a cool feature. What would be the purpose of having PHPUnit 999 installed and tells PHP-CS-Fixer to to transition the code from PHPUnit 1 to PHPUnit 2? The installed version and targeted transition one should always be the same.
Here I don't question what was done: it is a safety net for PHP-CS-Fixer maintainers and it's good for now; just talking about theory here :wink:

keradus: > Why should anyone could want to fix one function but not the others?

vide issue reported - because he does not had the other function (`mb_str_split`) and code execution after fixing was crashing for him (We had installed polyfill, he didn't).

for phpunit - well, one can has it installed as global dependency or phar file, and we cannot autodetect version.

agree - if we expect to run on same runtime (which we do), then it's not an issue if we don't modify the runtime on the fly (by adding pollyfills). then, maybe `isInternal` is good enough for us here.

Slamdunk:
Fix proposed to @SpacePossum : https://github.com/SpacePossum/PHP-CS-Fixer/pull/24

tests/Fixer/Alias/MbStrFunctionsFixerTest.php
SpacePossum: the PHP version check sounds better to me, but I'm pragmatic with unit tests so not a blocker for me either

SpacePossum:
TBH I didn't see the need for the revert as now we still have the same issue with all other function calls, just not the new one added.

The additional guard for the internal check sounds good to me, not sure if it out performs using a look up in the [`internal` functions array](https://www.php.net/manual/en/function.get-defined-functions.php) (if we start using this check for all fixers, it might be worth casing this kind of meta data about the functions somewhere?)

keradus:
> TBH I didn't see the need for the revert

well, not that much users of PHP CS Fixer already using PHP 7.4, and for all of them that are not using it nor the polyfill, we were breaking the result file.

src/Fixer/Alias/MbStrFunctionsFixer.php
```diff
@@ -42,6 +43,23 @@ final class MbStrFunctionsFixer extends AbstractFunctionReferenceFixer
...
+    private $functions;
+
+    public function __construct()
+    {
+        parent::__construct();
+
+        $this->functions = array_filter(
+            self::$functionsMap,
+            static function (array $mapping) {
+                return \function_exists($mapping['alternativeName']) && (new \ReflectionFunction($mapping['alternativeName']))->isInternal();
```
keradus: does it mean that we won't fix the function for end-user if he has php 7.0, but installed a polyfill ?

Slamdunk: Exactly, this was the purpose, wasn't it?

SpacePossum:
@keradus anything else you want here?

SpacePossum:
Thank you @keradus.

SpacePossum:
thanks @Slamdunk as well :D

PR #4536, opened 2019-08-31
DX: use PHIVE for dev tools

keradus:


keradus:
no, eg `localheinz/composer-normalize` has no phive support and `phpmd`, while having the support, support installing extensions only via composer

ondrejmirtes:
Hi, i think it's a step backwards to not have PHPStan in `require-dev`. With Phive you cannot take advantage of any PHPStan extensions because they're installed only by Composer. Phive does not have any way to enforce version constraints of dependencies.

On 0.11.x you can use `phpstan/phpstan-shim` package which has only a single dependency on `nikic/php-parser` v4. In a few weeks when 0.12 is released, main distribution package `phpstan/phpstan` will also work this way. This will help everyone to stay on the latest version without any risk of dependency conflicts.

SpacePossum:
I think it would be best to have one composer.json for the project and another for all development tooling like SCA, spelling checks, CS etc.

Thanks for all the work on PHPStan @ondrejmirtes ! great tool :)  :+1:  :100:

ondrejmirtes:
PHPStan 0.12 can be put into `require-dev` without any downside.

ondrejmirtes:
And thank you for the kind words :)

kubawerlos:
@SpacePossum something like this https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4639?

SpacePossum:
@kubawerlos that looks like a great start, sorry to see it stranded because of to much in it.

```
TL;DR the idea was to move out dev dependencies from main composer.json to free normal dependencies from any potential conflict, nothing more.
```

I agree here, although; sure we might get conflicts among the list of dev tools within the new composer.json, however that is better/easier to deal with than conflicts with the project itself.
We would still have the benefits of having the extensions for PHPStan :+1: 

```
require-dev must contain package we are using in tests folder.
we are using phpunit in tests folder heavily.
```

I agree with Keradus here, I also think this is the only one in `require-dev` we should keep. But maybe we can start by doing one of the tooling and take it from there? Baby steps and we'll get there ;)

staabm:
> no, eg localheinz/composer-normalize has no phive support 

Was just added to phive, see https://github.com/ergebnis/composer-normalize/issues/216

localheinz:
@staabm

Spotted a bug there, will update later and open a pull request here!

PR #4541, opened 2019-09-01
Enhancement: Use default name property to configure command names

### LINKED ISSUE #4539: Refactor how command name is exposed
As we already updated symfony/console component, we can do this update:

```diff
-const COMMAND_NAME = 'describe';
+protected static $defaultName = 'describe';
```

While change, remember that we would need to change AutoReview to not check that `COMMAND_NAME` exists but `$defaultName` is set

localheinz:
This PR

* [x] uses the `$defaultName` property to configure command names

Fixes #4539.

### DIFF
--- a/src/Console/Command/DescribeCommand.php
+++ b/src/Console/Command/DescribeCommand.php
@@ -48,7 +48,7 @@ use Symfony\Component\Console\Output\OutputInterface;
  */
 final class DescribeCommand extends Command
 {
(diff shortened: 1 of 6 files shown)


keradus:
@localheinz , can you point to 2.15 ?

localheinz:
@keradus 

Absolutely, just a moment!

keradus:
Thank you @localheinz.

localheinz:

PR #4546, opened 2019-09-05
DX: removing unnecessary variable initialization

kubawerlos:
Spotted when working on https://github.com/kubawerlos/php-cs-fixer-custom-fixers/pull/132

Both variables will be initialized in `if` and in `else`.

### DIFF
--- a/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
+++ b/src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
@@ -88,8 +88,6 @@ final class MyTest extends \PHPUnit_Framework_TestCase
             : '';
 
         $prevIndex = $tokens->getPrevNonWhitespace($index);
-        $doc = null;
-        $docIndex = null;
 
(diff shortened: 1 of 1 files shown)


SpacePossum:
Thank you @kubawerlos.

src/Fixer/PhpUnit/PhpUnitTestClassRequiresCoversFixer.php
```diff
@@ -88,8 +88,6 @@ private function addRequiresCover(Tokens $tokens, $startIndex)
             : '';
 
         $prevIndex = $tokens->getPrevNonWhitespace($index);
-        $doc = null;
-        $docIndex = null;
```
keradus: FYI:
i added it explicitly, i hate assigning a value in a `if` block to a variable that was not yet defined. `if` should create a variable-scope, but it doesn't in php :(

kubawerlos: When I see initialization like this I expect that there might be `if` condition which will override a value, but if not entering that condition will not and we will have to use the defaults. And in this case we always override it.

PR #4422, opened 2019-05-30
[HeredocIndentationFixer] Should be run before NoTrailingWhitespaceFixer

phansys:
With the proposed configuration, the expected result for this snippet
```php
<?php
····$a·=·<<<'EOD'
first line

ending line
EOD;
```
should be
```php
<?php
····$a·=·<<<'EOD'
········first line

········ending line
EOD;
```
instead of
```php
<?php
····$a·=·<<<'EOD'
········first line
········
········ending line
EOD;
```

PR #4548, opened 2019-09-07
HeredocIndentationFixer - remove whitespace in empty lines

gharlan:
replaces #4422

### DIFF
--- a/src/Fixer/Whitespace/HeredocIndentationFixer.php
+++ b/src/Fixer/Whitespace/HeredocIndentationFixer.php
@@ -94,8 +94,9 @@ SAMPLE
 
         Preg::match('/^[ \t]*/', $tokens[$end]->getContent(), $matches);
         $currentIndent = $matches[0];
+        $currentIndentLength = \strlen($currentIndent);
 
-        $content = $indent.substr($tokens[$end]->getContent(), \strlen($currentIndent));
+        $content = $indent.substr($tokens[$end]->getContent(), $currentIndentLength);
         $tokens[$end] = new Token([T_END_HEREDOC, $content]);
 
         if ($end === $start + 1) {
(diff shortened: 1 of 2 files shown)


tests/Fixer/Whitespace/HeredocIndentationFixerTest.php
```diff
@@ -176,6 +200,50 @@ public function provideFixCases()
...
+      def
+        '.'
+      EOD
+    );',
+            ],
+            [
+                /* EXPECTED */ '
+<?php
+    foo(<<<EOD
+
```
julienfalque: I'm not sure whitespaces that exist before fixing should be removed. IMO the fixer should only avoid adding new ones.

gharlan: I'm also not sure. But what would be the expected result for the given test case for you?
Keep the two spaces in the empty lines? Or change them to 4 spaces, because all other lines also get 2 more spaces indentation?

dmvdbrugge: As the fixer is called `HeredocIndentationFixer` it imho just fixes all things wrong with the indentation. Having 2 where it should be 4 is incorrect and should thus be fixed. It then seems weird to pad here and not in fully empty lines, because that would still give an inconsistency: 2 kinds of empty lines, just one with the "prefix" and one without.

I can therefor only come to the conclusion that it should either pad all or pad none. And as the complaint was that it added, I guess the only (already implemented) solution is to remove all spacing.

julienfalque: Agreed, let's keep current behavior (removing spaces in all blank lines).

gharlan: Side note: except in those lines, where the indentation of the empty line is more than the heredoc indentation (marked by the heredoc end). For this case see lines 203-224.
These spaces can not be removed because this fixer is not marked as risky.

src/Fixer/Whitespace/HeredocIndentationFixer.php
```diff
@@ -107,19 +108,35 @@ private function fixIndentation(Tokens $tokens, $start, $end)
                 continue;
             }
 
-            $regexEnd = $last && !$currentIndent ? '(?!$)' : '';
-            $content = Preg::replace('/(?<=\v)'.$currentIndent.$regexEnd.'/', $indent, $tokens[$index]->getContent());
+            $content = $tokens[$index]->getContent();
+
+            if ($currentIndent) {
```
dmvdbrugge: Aren't we usually explicit? (Also in the ternary on line 117 and the `elseif` on line 135)

julienfalque: Indeed, we prefer explicit comparisons.

SpacePossum:
Thank you @gharlan.

PR #4549, opened 2019-09-08
DX: use ::class whenever possible

keradus:


keradus:
shell may have a lot of false positives tbh
please extract the shell script part and let's review it separately

kubawerlos:
@keradus let have [this](https://github.com/keradus/PHP-CS-Fixer/pull/32/files) and go with the changes - I don't thinks we have to make automated check for this right now.

SpacePossum:
Thank you @keradus.

PR #4550, opened 2019-09-09
DX: travis_retry for dev-tools install

ktomk:
in Travis seeing errors like:

	gpg: keyserver timed out
	gpg: keyserver receive failed: keyserver error
	The command "./dev-tools/install.sh" failed and exited with 2 during .

which is doing GPG verification of the downloaded Phive utility and then
failing due to network failures w/ GPG key-servers.

a work-around is to retry done by travis_retry (retries 3 times by default
on failure).

this is also true for other network operations during dev-tools install.

as the install.sh script only installs by testing for files if already
installed the installation of each utility forms a transaction of its own.

to establish transaction behavior for the Phive installation, files are
now re-downloaded (unless Phive was already installed) and only after
successful GPG verification the Phive utility is installed as bin/phive.

for better verification of each utility installation per potential retry,
the installation check is done always (most often by invoking each utility
and displaying the version information). previously this was not necessary
as the script was not expected to be called over and over again.

change for all wget operations having the --no-clobber argument while
using the --output-document with a pathname to a regular file are not safe
any longer as wget exits w/ non-zero status when the file already exists.
therefore wget writes the file to stdout that is redirected to the target
file to make the operation fail-safe in context of an unattended build
that re-tries on failure.

utilities are now tested for being executable as only testing for being a
file as done previously is not enough as installation requires modifying
the execution bit.

additionally a minor improvement in extracting the single shellcheck file
from the tar archive by directly extracting the file removing the
management of the temporary directory that previously was an intermediate
temporary result of tar-file extraction.

these changes should also benefit calling the script again in interactive
use (e.g. $ ./install.sh).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ jobs:
             php: 7.2
             env: COMPOSER_FLAGS="--prefer-stable"
             install:
-                - ./dev-tools/install.sh
+                - travis_retry ./dev-tools/install.sh
 
                 - travis_retry composer update $DEFAULT_COMPOSER_FLAGS $COMPOSER_FLAGS
                 - composer info -D | sort
--- a/dev-tools/install.sh
+++ b/dev-tools/install.sh
@@ -1,42 +1,70 @@
 #!/bin/sh
-
+#
+# dev-tools install
+#
+# installation script for dev-tool utilities (phive, checkbashisms,
+# etc.) but not composer.
+#
+# script must be idempotent as to continue retrying in case of failure
+# (e.g. network timed out) when invoking again so that temporary i/o or
+# network problems can be dealt with by invoking the script again (and
+# again) until success. required for unattended build.
+#
+# usage: ./dev-tools/install.sh
+#    or: travis_retry ./dev-tools/install.sh
+#    or: ./install.sh
+#    or: ./install.sh || ./install.sh || ./install.sh
+#
 set -eu
 
 cd "$(dirname "$0")"
 
 mkdir -p bin
 
+###
+# temporary work-around: phive expects gpg2 as gpg, build box has gpg1 as gpg
+#   phive/phive:208 <https://github.com/phar-io/phive/issues/208>
+if [ "${TRAVIS-}" = true ]; then
+  sudo apt-get install dirmngr --install-recommends
+  [ ! -f "/usr/bin/gpg1"  ] && sudo mv -- "/usr/bin/gpg" "/usr/bin/gpg1"
+  sudo ln -sfT "/usr/bin/gpg2" "/usr/bin/gpg"
+fi
+
 VERSION_CB="2.0.0.2"
 VERSION_SC="stable"
 
 echo λλλ phive
-if [ ! -f bin/phive ]; then
-    wget --no-clobber --output-document=bin/phive https://phar.io/releases/phive.phar
-    wget --no-clobber --output-document=bin/phive.asc https://phar.io/releases/phive.phar.asc
+if [ ! -x bin/phive ]; then
+    wget -Obin/phive https://phar.io/releases/phive.phar
+    wget -Obin/phive.asc https://phar.io/releases/phive.phar.asc
     gpg --keyserver pool.sks-keyservers.net --recv-keys 0x9D8A98B29B2D5D79
     gpg --verify bin/phive.asc bin/phive
     chmod u+x bin/phive
-    bin/phive --version
 fi
+bin/phive --version
 
 echo λλλ checkbashisms
-if [ ! -f bin/checkbashisms ]; then
-    wget --no-clobber --output-document=bin/checkbashisms https://sourceforge.net/projects/checkbaskisms/files/${VERSION_CB}/checkbashisms/download
+if [ ! -x bin/checkbashisms ]; then
+    wget -Obin/checkbashisms https://sourceforge.net/projects/checkbaskisms/files/${VERSION_CB}/checkbashisms/download
     chmod u+x bin/checkbashisms
-    bin/checkbashisms --version
 fi
+bin/checkbashisms --version
 
 echo λλλ shellcheck
-if [ ! -f bin/shellcheck ]; then
-    wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${VERSION_SC}.linux.x86_64.tar.xz" | tar -xJv --directory bin shellcheck-${VERSION_SC}/shellcheck
-    mv "bin/shellcheck-${VERSION_SC}/shellcheck" bin/
-    rmdir "bin/shellcheck-${VERSION_SC}/"
-    bin/shellcheck --version
+if [ ! -x bin/shellcheck ]; then
+    wget -qO- "https://storage.googleapis.com/shellcheck/shellcheck-${VERSION_SC}.linux.x86_64.tar.xz" \
+        | tar -xJv -O shellcheck-${VERSION_SC}/shellcheck \
+        > bin/shellcheck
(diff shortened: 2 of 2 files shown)


ktomk:
Just a bit more context here. When filing #4517 I experienced Travis build errors on the new 2.15 base branch (was previously filing against 2.12 w/o this problem).

This is what fixed the build problems for me. After a [short exchange on gitter](https://gitter.im/PHP-CS-Fixer/Lobby?at=5d74dcb6b9abfa4bd36f50b9) w/ @keradus I extracted the fix and made it a PR of it's own.

dev-tools/install.sh
```diff
@@ -10,29 +25,31 @@ VERSION_CB="2.0.0.2"
 VERSION_SC="stable"
 
 echo λλλ phive
-if [ ! -f bin/phive ]; then
-    wget --no-clobber --output-document=bin/phive https://phar.io/releases/phive.phar
-    wget --no-clobber --output-document=bin/phive.asc https://phar.io/releases/phive.phar.asc
+if [ ! -x bin/phive ]; then
+    wget --output-document=- https://phar.io/releases/phive.phar > bin/phive
```
julienfalque: What's the difference between `--output-document=bin/phive` and `--output-document=- > bin/phive`? Isn't removing `--no-clobber` enough?

ktomk: @julienfalque It won't overwrite the existing file. Just removing --no-clobber would start to number the file(s).

dev-tools/install.sh
```diff
@@ -1,5 +1,20 @@
 #!/bin/sh
-
+#
+# dev-tools install
+#
+# installation script for dev-tool utilitites (phive, checkbashisms,
```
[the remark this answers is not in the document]
ktomk: @julienfalque I've addressed this typo in a fixup! commit on top.

dev-tools/install.sh
julienfalque: Are you sure about that? I just tried this locally:
```console
$ wget --quiet --output-document=composer.json https://raw.githubusercontent.com/FriendsOfPHP/PHP-CS-Fixer/2.15/composer.json
$ head -n 2 composer.json
{
    "name": "friendsofphp/php-cs-fixer",
$ echo test > composer.json
$ head -n 2 composer.json
test
$ wget --quiet --output-document=composer.json https://raw.githubusercontent.com/FriendsOfPHP/PHP-CS-Fixer/2.15/composer.json
$ head -n 2 composer.json
{
    "name": "friendsofphp/php-cs-fixer",
```

ktomk: No, not sure. I normally refer to the manpage when I'm in doubt. `wget -O - ... > file` is the equivalent of `wget -O file ...` as I just read there, so what you suggested should behave the same as in the script (when not using `--no-clobber`).

.travis.yml
```diff
@@ -162,3 +162,6 @@ jobs:
                     tags: true
             after_deploy:
                 - ./dev-tools/trigger-website.sh ${TRAVIS_TOKEN} ${TRAVIS_TAG}
+
+    allow_failures:
+        -   php: 7.4snapshot
```
julienfalque: This should be reverted, see #4559.

dev-tools/install.sh
julienfalque: Actually, even if the previous syntax doesn't overwrite the existing file, I think it's fine because it means the file has been downloaded in a previous attempt, so it is already the expected one.

So I propose to:
- keep the previous syntax (no output redirect) as it looks more readable to me (removing the `--no-clobber` option should be fine)
- keep the `wget` execution in a `if [ ! -f file ]` condition as was done before (no need to try to download the file if it is already here) but keeping the `chmod u+x` execution in its own `if [ ! -x file ]` condition
- execute `gpg --recv-keys` command *before* downloading the file: if fetching the key fails but the file has been downloaded, it won't be tried again

.travis.yml
ktomk: No problem to remove the commit, but it will make the build fail / unstable I fear.

dev-tools/install.sh
ktomk: I can modify this so it pleases your preference regarding wget syntax.

But I would restrain from adding more complexity than necessary as I find that rather cumbersome for the other parts.

also I don't understand where you see the re-try transaction broken by the change-set for phive/gpg (as long as bin/phive is not executable, there is retry).

julienfalque: The third point of my suggestion is necessary only if other items are applied. The current setup works fine. We might download the file even if we already did it successfully but it won't happen often so indeed, not worth the complexity.

ktomk: @julienfalque Just pushed an update. Rebased on current 2.15 as some travis build fixes are in there and added a new fixup! on top which removes relying on output redirection w/ wget and --output-file. I used -O as equivalent option name (short option) as it was in use in the script already to align w/ existing style. This should address your comments, so please take a look if this fits well in your eyes.

ktomk: Okay, retrying works however sometimes errors are of more permanent nature:

~~~
ERROR]    Failed to download load https://api.github.com/repos/maglnet/ComposerRequireChecker/releases: HTTP Code 403 
~~~

* Build: https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/595557561

As I can't re-trigger the build AFAIK, I might push some faux commit later today to trigger it again.

kubawerlos: @ktomk `git commit --allow-empty -m 'Trigger CI'`

ktomk:
Even this build goes towards green again, this is far from ready.

We stumbled over an issue with `phive` and the build box: https://github.com/phar-io/phive/issues/208

For now on the PHP-CS-Fixer travis build we have a work-around, but I would like to use this PR to continue helping finding a solution for the underlying issue so that `phive` can improve and at best the workaround is not needed at all.

julienfalque:
:+1: for waiting a proper fix rather than a workaround. Thank you for the time you spend on this.

ktomk:
Yesterday I cleaned the debug / workaround commits up a bit and removed the intermediate binaries from the history as the workaround is only needed until the upstream fix and it's only some lines at the beginning of the file.

Now as reading your comment I also inlined the fixup! commits so for it to become merge ready, and resolved all conversations, hopefully in your interest. If not let me know.

dev-tools/install.sh
```diff
@@ -1,42 +1,70 @@
...
 fi
+bin/shellcheck --version
 
 echo λλλ composer packages
 composer update
 composer info -D | sort
 
 echo λλλ phive packages
+
+echo "debug: #phive/phive:208"
```
keradus: do we need this `debug` print ?

julienfalque: I would keep it. It will reminds us there is some workaround code that should be removed as soon as the upstream bug is fixed.

ktomk: Ah you spotted it. I would also keep it. It marks the place where it failed w/o the workaround.

keradus: We can keep it as a comment if it's only to mark the place where it failed.

If John gonna call the script, he will have no clue what to do with this debug msg. IMO, It's confusing for users that are not aware about this phive issue (and who should not wonder about it, to be honest...)

I will let myself to comment it out and merge this PR.

keradus:
I blame myself to not merge it sooner. Sorry for that.
Big thanks for all the investigation and hard work to create this PR, @ktomk !

Please, don't hesitate to hop in and cleanup after the main issue would be fix on phive. Thanks ;)

Issue #4382, opened 2019-04-16
[Meta] PHP 7.4 support 


SpacePossum:
#### (Syntax) changes that require new tests to make sure the tool is compatible with PHP 7.4:

- [x] [Null Coalescing Assignment Operator](https://wiki.php.net/rfc/null_coalesce_equal_operator) (#4603)
- [x] [Typed Properties 2.0](https://wiki.php.net/rfc/typed_properties_v2) (#4593)
- [x] [Spread Operator in Array Expression](https://wiki.php.net/rfc/spread_operator_for_array) (#4602)
- [x] [Arrow Functions 2.0](https://wiki.php.net/rfc/arrow_functions_v2) (#4600)
- [x] [Change the precedence of the concatenation operator](https://wiki.php.net/rfc/concatenation_precedence) (couldn't find any issue with this on 2.15)
- [x] [Numeric Literal Separator](https://wiki.php.net/rfc/numeric_literal_separator) (#4596)

#### Other:

- [x] [Reflection for references](https://wiki.php.net/rfc/reference_reflection) (class casing) (couldn't find any issue with this on 2.15)
- [x] [Password Hashing Registry](https://wiki.php.net/rfc/password_registry) (function casing) (couldn't find any issue with this on 2.15)
- [x] [mb_str_split](https://wiki.php.net/rfc/mb_str_split) (#4380 -> #4534)
- [x] [New custom object serialization mechanism](https://wiki.php.net/rfc/custom_object_serialization) (#4381)

### TODO:

- [x] PHP 7.4 integration test (#4637)
- [x] do the same checks for rules added in v2.16 (#4638)
- [x] do the same checks for rules added on master (couldn't find any issue with this on master)
- [x] VisibilityRequiredFixer (#4641)

#### Sources:
[Approved RFC's list for PHP7.4](https://wiki.php.net/rfc#php_74)

zmitic:
Will it be possible to automatically add property typehints if constructor has assignment, but property itself doesn't have ``@var`` annotation?

Example:

```php
class Product
{
	private $category;

	public function __construct(Category $category)
	{
		$this->category = $category;
	}
}
```

The reason is that I never put ``@var`` annotations, PHPStorm is smart enough to figure the type from constructor and give me autocomplete.

keradus:
can be problematic in cases like
```php
        private $foo;
	public function __construct(Category $category)
	{
[...]

PR #4557, opened 2019-09-19
feat: Introduce `StaticPrivateMethodFixer`

Slamdunk:
References:
1. https://twitter.com/BackEndTea/status/1174225152034385920 (/cc @BackEndTea)
1. https://github.com/localheinz/phpstan-rules/issues/133

```diff
 class Foo
 {
     public function bar()
     {
-        return $this->baz();
+        return self::baz();
     }
     
-    private function baz()
+    private static function baz()
     {
         return 1;
     }
 }
```

- [x] Must not contain `$this` references
- [x] ~Must not contain non-static closures~ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4557#issuecomment-532988296 updated to: Must not contain closures
- [x] Must not contain `debug_backtrace()` calls
- [x] Must be marked as risky, no dynamic reference handled such as `$this->{$method}()` or `$$var` inside the method
- [x] Must run after `ProtectedToPrivateFixer`
- [x] Must run before `StaticLambdaFixer`


Ocramius:
Restrictions:

 * Must not contain non-static closures
 * Must not contain `debug_backtrace()` usage
 * Must not contain `$this`
 * Must not contain `$$var`

Slamdunk:
> Must not contain `debug_backtrace()` usage

Why? How could a code calling `debug_backtrace()` be affected by this change?
I mean affected runtime; the output would obviously be different, but I would not consider this as an issue.

> Must not contain `$$var`

This could only affects `$$var` referencing to `$this` or global variables AFAICS, we'll point this out in the Risky signature but not handle them

Ocramius:
> Why? How could a code calling `debug_backtrace()` be affected by this change?

`debbug_backtrace()[1]['object']` stops working

dkarlovi:
> Must not contain non-static closures
> Must run before StaticLambdaFixer

Wouldn't running after static lambda relax the first requirement?

BackEndTea:
> Wouldn't running after static lambda relax the first requirement?

These two kinda have circular dependency, idk if php-cs-fixer can handle that atm.

Slamdunk:
> > Must not contain non-static closures
> > Must run before StaticLambdaFixer
> 
> Wouldn't running after static lambda relax the first requirement?

Heh, that's tricky, it should run before to relax methods containing non-static closures, but also after to fix non-static closures using private methods now declared static.
As pointed out by @BackEndTea here's a circular dependency.

Ocramius:
You can solve the dependency by running these recursively, which would require changing the tool architecture.

Your exit condition being "nothing got fixed", or "nesting level limit reached".

dkarlovi:
I mean, it's the same effect as running it twice, I guess. It's a bit annoying if you have an unstable rule combo like this: you run CS and commit, your CI runs it again and fails because it changed something too.

Slamdunk:
@keradus always pushed for avoiding such architectures for clear reasons. I think I'll propose the first production-ready version of the Fixer to completely ignore private method containing any closure

Slamdunk:
> must not be declared by interface

@Ocramius you killed me :cry: this practically makes the Fixer impossible to be implemented without context awareness

dkarlovi:
download

BackEndTea:
> must not be declared by interface

Interfaces can't declare private methods right?

Slamdunk:
> Interfaces can't declare private methods right?

No, but the resulting new method can be present in the interface, and such creating a conflict

BackEndTea:
I don't understand, could you provide an example

Slamdunk:
I thought this https://3v4l.org/G6vS6 could be possible but turns out it can't, so either @Ocramius know something we don't or he was wrong; and it would be really weird for him to be wrong :stuck_out_tongue:

Ocramius:
Maybe annoying, but still  a good amount of improvement 👍

Ocramius:
@Slamdunk sorry, I missed the limitation to private methods in my thougt process 😛

dkarlovi:
TBH I really assumed you actually **could** put private methods in interfaces, because... you know, PHP. 

TIL you can't.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,288 @@
...
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class StaticPrivateMethodFixer extends AbstractFixer
+{
+    /**
+     * @var array
+     */
+    private $magicNames = [
+        '__call' => true,
+        '__callstatic' => true,
```
BackEndTea: `__callStatic` can't be non-static, so it could be removed here.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,288 @@
...
+                continue;
+            }
+
+            if ($this->skipMethod($tokens, $functionKeywordIndex, $methodOpen, $methodClose)) {
+                continue;
+            }
+
+            $fixedMethods[$methodName] = true;
+
+            $tokens->insertAt($functionKeywordIndex, [new Token([T_STATIC, 'static']), new Token([T_WHITESPACE, ' '])]);
```
BackEndTea: How does this work with final methods?

Even though it doesn't do much, private methods *can* be final.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,288 @@
...
+        // Must run after ProtectedToPrivateFixer, before StaticLambdaFixer
+        return 1;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function isCandidate(Tokens $tokens)
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_PRIVATE]);
```
BackEndTea: I think the fixer can also be applied inside of Traits

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,308 @@
...
+    private function baz()
+    {
+        return 1;
+    }
+}
+'
+                ),
+            ],
+            null,
+            'Risky when method contains dynamic generated calls to the instance, or the method is dynamically referenced.'
```
BackEndTea: This is also risky if the private method is referenced from a trait, or a private method from a trait is referenced, as i don't think the fixer can pick up on that.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: I'll refine this list even further, as a lot of this methods cannot be private, thank you

Slamdunk: I'll refine this list even further, as a lot of this methods cannot be private, thank you

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,308 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param Tokens         $tokens
+     * @param TokensAnalyzer $tokensAnalyzer
+     * @param int            $methodOpen
+     * @param int            $methodClose
+     * @param array          $fixedMethods
```
BackEndTea: Since this array has to be of a specific key/value type, how about making that explicit in the types here:
```suggestion
     * @param array<string, bool> $fixedMethods
```

Slamdunk: Didn't know this library uses generic annotations, thank you for the feedback

BackEndTea:
One more possible edge case that (may) not be relevant yet. How about $this etc in php 7.4 arrow functions

Slamdunk:
> One more possible edge case that (may) not be relevant yet. How about $this etc in php 7.4 arrow functions

I was planning to pinpoint it in an issue like https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4382 as I haven't a clear idea on the topic as of yet, nor on the way this library intend to support the new code structure

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: Mmm, I think not
```php
// file Foo.php
trait Foo
{
    private function bar() {}
}
// file Baz.php
class Baz
{
    use Foo;
    
    public function wut()
    {
        return $this->bar();
    }
}
```
I cannot fix Foo::bar because I would break Baz::wut behavior; well, not entirely break as php still run without issue, but a static analyzer would complain

Slamdunk: I'm sorry I just read https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4557#discussion_r326203871: yes we can fix Traits too pointing the possible side effects into the risky description.

But honestly the two scenario are much differently spread imho:

1. A class that use a private method of a trait: ok, bad but common, we can't use this fixer for Trait as we would (logically) break the code
1. A trait that use a private method of a class: WUT? Are you kidding me? Yeah it's feasible, but if you write crappy code it's not our fault

I feel safe to fix classes and not fix traits, what do you think?

BackEndTea: Seems like a good idea.

I don't think it will be a common problem, but it is one of those nasty edge cases.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: Thank you, added

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: Fixed, thanks

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,298 @@
...
+
+/**
+ * @author Filippo Tessarotto <zoeslam@gmail.com>
+ */
+final class StaticPrivateMethodFixer extends AbstractFixer
+{
+    /**
+     * @var array
+     */
+    private $magicNames = [
```
kubawerlos: Can it be constant?

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
```diff
@@ -0,0 +1,298 @@
...
+        '__clone' => true,
+        '__construct' => true,
+        '__destruct' => true,
+        '__serialize' => true,
+        '__set_state' => true,
+        '__sleep' => true,
+        '__unserialize' => true,
+        '__wakeup' => true,
+    ];
```
kubawerlos: What about other magic methods like `__call`? They can't be `static` either, right?

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: Yes, must it?

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
Slamdunk: I started with all magic methods in the test cases, but some like `__call` cannot be private (https://3v4l.org/gNl6u) and therefore cannot be added into them.

If needed I can still add all of them into the array, but cannot be tested.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
kubawerlos: I'd say yes, as for [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

Always when I see variable I'd expect it would be modified and when I see constant I know it won't be modified - and here we have variable that won't be modified.

Slamdunk: Of course. What stops me from doing this is that here we must use public constants as this project must support PHP versions that don't know about private constants.

Between public constants and private variables, I'd prefer private variables, WDYT?

kubawerlos: > Between public constants and private variables

we can mark constant `@internal`, it make a bit ~better~ less bad, but I won't insist.

Slamdunk:
@SpacePossum may we tag this for 2.17?

Wirone:
@Slamdunk I've updated this PR because it looks interesting 🙂. I've interactively rebased your branch, did `dev-tools/doc.php` on each commit to update docs at any point of your work process and fixed PHPStan/PHPUnit issues in last commit (there is only one fail in `fix-self` case - fixer skips method because it's private, and I am not sure what to do with it, especially that this case does not make much sense:

image

Let's resolve this last problem and process it further 😉.

src/Fixer/ClassNotation/StaticPrivateMethodFixer.php
kubawerlos: This discussion is outdated nowadyas, we can use `private const` now.

doc/rules/class_notation/static_private_method.rst
```diff
@@ -0,0 +1,44 @@
+==============================
+Rule ``static_private_method``
+==============================
+
+Converts private methods to ``static`` where possible.
```
keradus, line 5: I'm not convinced that we should promote this rule, I rather avoid using static where possible.

if you want to have it in, go ahead with merge @Wirone

kubawerlos (approved):

doc/rules/class_notation/static_private_method.rst
keradus: > do you have some reasoning

Overall, I'm not big sponsor of `static` as I see it close to `singleton` or `global`.
With that, I wouldn't promote using it (which the propose rule doing).
It has it's usage, just not to apply to whole codebase blindly.

Same time, proposed rule is limiting this to internals of class, only static methods. That's why I'm not fully against and "if you want to have it, go ahead with merge".


> Deprecating `static_lambda` ?

so that's exactly matching the part of "limited to internals" ;)

PR #4559, opened 2019-09-22
Allow 7.4snapshot to fail due to a bug on it

kubawerlos:
Bug: https://bugs.php.net/bug.php?id=78575

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -162,3 +162,6 @@ jobs:
                     tags: true
             after_deploy:
                 - ./dev-tools/trigger-website.sh ${TRAVIS_TOKEN} ${TRAVIS_TAG}
+
+    allow_failures:
+        - php: 7.4snapshot


src/Console/Command/SelfUpdateCommand.php
```diff
@@ -151,7 +151,8 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $tempFilename = \dirname($localFilename).'/'.basename($localFilename, '.phar').'-tmp.phar';
         $remoteFilename = $this->toolInfo->getPharDownloadUri($remoteTag);
 
-        if (false === @copy($remoteFilename, $tempFilename)) {
+        copy($remoteFilename, $tempFilename);
```
keradus: unfortunately, we should still silence the `copy` function to not get low-level warning, we are handling outcome differently.

keradus:
to be honest, looks like hacky workaround nobody will understand in half year.
please, report the issue to php-src.

kubawerlos:
@keradus bug is already reported at https://bugs.php.net/bug.php?id=78575 - let's gite it some votes.

We can allow PHP 7.4 to tail on Travis until it is fixed.

julienfalque:
I think I would still keep the build failing. It's easier to forget to check whether PHP 7.4 failed for a good reason when everything is green that to ignore failures due to this bug when only PHP 7.4 is red.

ktomk:
This is not PHP 7.4,  it's the snapshot. Would be same as with master. They normally fail in the build (speaking of my own projects), so I allow snapshots / nightlies / masters to fail. I only need to check them when I want to learn how it is going on and for upstream reporting (e.g. PHP bugs etc.).

Not allowing failure for unstable run-times in the build must ask for something very specific to  argue for the burden this creates (e.g. my PRs fail here b/c of that, see #4550). I would love to learn about the specifics that weight the burden.

kubawerlos:
@julienfalque I like to rely on CI checks and here we have clear situation that unstable PHP version fails CI due to its bug, We are far from supporting PHP 7.4, so why do we require it to pass tests? It's not like we would have dozens of failing tests when we allow it to fail.

I bet there would be people (including myself) checking the reason of PHP 7.4 failure even if failure would be allowed.

SpacePossum:
Im fine for allowing the snapshot to fail

SpacePossum:
Thank you @kubawerlos.

Issue #4496, opened 2019-08-08
Install composer locally question


danielspk:
Why is the installation not documented by composing the project locally? Are there any problems detected?

kubawerlos:
I have no idea, it's the way I install this library every time :)

ktomk:
@danielspk: If I may ask, what do you exactly mean by composing the project locally? Are you cloning sources from Github and then do composer install or do you mean to require via composer cs-fixer as a dev dependency?

danielspk:
I require via composer cs-fixer as a dev dependency.

It works correctly. It simply catches my attention that it is not documented in the README.md as a valid means of local installation.

keradus:
it's fully possible.
looks like a person who created this part of docs focused on more "fancy" cases leaving the obvious one not documented.
@danielspk , can you extend the documentation to cover that case? thanks !

danielspk:
Ready!

https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4560

ktomk:
@danielspk Thanks for the quick feedback and for taking care directly filing a PR.

You asked about if *"there [are] any problems detected?"* when adding PHP-CS-Fixer as a composer dev dependency.

If you make use of PHP-CS-Fixer in a project that also has non-dev dependencies which collide or intersect with PHP-CS-Fixer (e.g. Symfony components), composer needs to find install candidates across all of them. This might lead to upgrade/downgrade of non-dev requirements only due to dev requirements which most often is not wanted (and can also lead to unresolved deps under circumstances).

Apart from that, there aren't any problems that I know of. I use it as dev dependency in some projects, but I think I do not have any same dependencies in these.

To avoid such problems altogether, take a look at the phive installer from phar.io. PHP-CS-Fixer is available via it as well and you can have the tooling independent to your composer requirements. I can highly recommend it: https://phar.io/

keradus:
just to fulfill the picture - issue described above is relevant for any package installed via Composer, not just PHP CS Fixer

ktomk:
Absolutely what @keradus wrote, this is nothing specific to the PHP-CS-Fixer project, this is just how composer works in any composer based project.

IMHO for a self-containing build in a project this is sort of fine, but it comes w/ work if this goes more into tooling. It is also possible to care about that in the build, e.g. use a different vendor directory in the
build tool-chain, a different composer.json and what not. Or just phive. Or what not. ;)

[...]

PR #4560, opened 2019-09-23
Install and update locally by composer

danielspk:
Edit README.

Add info for install and update locally by composer.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -94,15 +94,6 @@ Install `PHIVE <https://phar.io>`_ and issue the following command:
 
     $ phive install php-cs-fixer # use `--global` for global install
 
-Locally (Composer)
-~~~~~~~~~~~~~~~
-
-Install `Composer <https://getcomposer.org/download/>`_ and issue the following command:
-
-.. code-block:: bash
-
-    $ ./composer.phar require --dev friendsofphp/php-cs-fixer
-
 Update
 ------
 
@@ -148,13 +139,6 @@ Locally (PHIVE)
 .. code-block:: bash
 
     $ phive update php-cs-fixer
-
-Locally (Composer)
-~~~~~~~~~~~~~~~
-
-.. code-block:: bash
-
-    $ ./composer.phar update friendsofphp/php-cs-fixer
 
 Usage
 -----
--- a/README.rst
+++ b/README.rst
@@ -94,6 +94,15 @@ Install `PHIVE <https://phar.io>`_ and issue the following command:
 
     $ phive install php-cs-fixer # use `--global` for global install
 
+Locally (Composer)
+~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


keradus:
@julienfalque , we are giving the options here, it's up to end user. you consider it wrong, one considers it's great, as he doesn't need to have global tools/multiple composer files/multiple package managers . both can use whatever we prefer. simply, let's document the option and that it. it's up to end user which one he will take.

keradus:
@danielspk , please take a look at README paragraph
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/2.15/CONTRIBUTING.md#quick-guide

SpacePossum:
>it's not npm, that's how composer works.

This true by itself so not much an argument to go on :)  In order words, just because composer acts this way and npm does not, doesn't mean it is a good idea to use composer for that use case (some might think so but I don't).

I agree with @julienfalque here and as such I think we should not document this way, we cannot document all possible ways to install so I think it would be best to only document ways we think is the best or "good" way.

julienfalque:
IMO #4608 is the exact reason why we should suggest *not* to install the tool as a dev dependency in the main `composer.json`: every two years, upgrading PHP CS Fixer to next Symfony major version suddenly becomes the top priority because it's installed in a lot of projects this way, so it prevents those projects from using the new Symfony version until PHP CS Fixer does.

This is wrong, the version of Symfony components used by PHP CS Fixer should remain an internal detail, we should be able to upgrade whenever we want/need it, not immediatly when a new Symfony major version is released.

keradus:
I like your reasoning, @julienfalque . but how to bind which PHP CS Fixer version should be used by given project? doing a global installation can do more harm.
what if i have 3 projects?
- project A using 2.12
- project B using 2.16
- project C using 1.12

let say that i had installed 2.12 globally.
the result of executing the tool will not be stable
- processing project A will pass
- processing project B will fail due to configured rules from 2.16
- processing project C will crash due to heavy bc breaks.

we need to provide a way to bind which PHP CS Fixer should be used for given project by having any kind of lock.
for that and completeness of having both, global and local installation steps , i like having the option described in readme.

julienfalque:
I'm not talking about installing PHP CS Fixer globally. It brings other issues indeed. I'm talking about installing it via a dedicated `composer.json` (e.g. `tools/php-cs-fixer/composer.json`). Nothing prevents this `composer.json` from having its own `composer.lock`.

The `require-dev` section should be for packages that are actually needed by the project to run in a non-prod environment. For example `symfony/debug` for a Symfony application. PHP CS Fixer is *not* required by the application to run, it is an external tool.

SpacePossum:
closing following https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4496#issuecomment-571927585

Issue #4430, opened 2019-06-01
Add missing schema for gitlab-reporter


keradus:
gitlab-reporter got introduced by #4288 ,
even if there is no official schema, we should define what schema we follow, so we would have BC promise on it and won't break it

PR #4563, opened 2019-09-27
GitlabReporter - fix report output

mjanser:
According to the [documentation](https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html#implementing-a-custom-tool) the location in the report should be a nested object.

### DIFF
--- a/src/Report/GitlabReporter.php
+++ b/src/Report/GitlabReporter.php
@@ -17,6 +17,8 @@ use Symfony\Component\Console\Formatter\OutputFormatter;
 /**
  * Generates a report according to gitlabs subset of codeclimate json files.
  *
+ * @see https://github.com/codeclimate/spec/blob/master/SPEC.md#data-types
+ *
  * @author Hans-Christian Otto <c.otto@suora.com>
  *
  * @internal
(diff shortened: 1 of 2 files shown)


SpacePossum:
Hi and thanks for the PR!

Could you add `@see https://github.com/codeclimate/spec/blob/master/SPEC.md#data-types` to the header so it becomes:

```
/**
 * Generates a report according to gitlabs subset of codeclimate json files.
 *
 * @see https://github.com/codeclimate/spec/blob/master/SPEC.md#data-types
 *
 * @author Hans-Christian Otto <c.otto@suora.com>
 *
 * @internal
 */
```

after that I'm :+1:  for merging in this

mjanser:
Sure, I added the reference to the documentation.

keradus:
please, can you solve https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4430 inside this PR?

mjanser:
What should I do to solve #4430 inside this PR?

SpacePossum:
Thank you @mjanser.

keradus:
I really don't like that we are doing some unspecified changes to reporter, that is supposed to be under BC promise, but we don't specify what the promise is - we have reporter without spec :/

cc @SpacePossum @mjanser

SpacePossum:
We've added a link to the spec: 
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4563#issuecomment-539068372

This PR fixes a bug so the tool generates the report following that spec, I think that is inline with BC?

keradus:
Link to a spec does not ensure that our implementation is following this spec.
validating outcome of our implementation with spec file (eg xsd) would.

SpacePossum:
Good idea, let me know when you've time to help out with that :)   

Till than I will merge bug fixes because I think these are more important than test case completeness.

PR #4564, opened 2019-09-29
Move readme-update command to Section 3

iwasherefirst2:
If one follows strictly the previous instructions, then the call in Section 1
from `php php-cs-fixer readme > README.rst` would lead to an error,
because no `FixerDefinition` has been provided yet. Therefore,
this commits suggest to move the readme-call after the definition
has been set.

### DIFF
--- a/doc/COOKBOOK-FIXERS.md
+++ b/doc/COOKBOOK-FIXERS.md
@@ -146,13 +146,6 @@ final class RemoveCommentsFixerTest extends AbstractFixerTestCase
     }
 }
 ```
-
-The files are created, one thing is still missing though: we need to
-update the README.md. Fortunately, PHP CS Fixer can help you here.
-Execute the following command in your command shell:
-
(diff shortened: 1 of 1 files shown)


doc/COOKBOOK-FIXERS.md
```diff
@@ -268,6 +261,11 @@ final class RemoveCommentsFixer extends AbstractFixer
     }
 }
 ```
+Next, we need to update the README.md.
```
julienfalque: Please remove the trailing space to make Travis happy.

doc/COOKBOOK-FIXERS.md
```diff
@@ -268,6 +261,11 @@ final class RemoveCommentsFixer extends AbstractFixer
...
```
kubawerlos:
```suggestion
Next, we need to update the `README.rst`.
```

SpacePossum:
Thank you @iwasherefirst2.

PR #4505, opened 2019-08-15
NoSuperfluousPhpdocTagsFixer - allow params that aren't on the signature

azjezz:
fixes #4504
fixes #3933

Issue #3398, opened 2018-01-06
Getting yoda fixes even though I disabled that option


mnapoli:
#### The PHP version you are using (`$ php -v`):
=> 7.2.0

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> `PHP CS Fixer 2.9.0 Speechless by Fabien Potencier and Dariusz Ruminski`

#### The command you use to run PHP CS Fixer:
=> `vendor/bin/php-cs-fixer fix --dry-run --diff --allow-risky yes`

#### The configuration file you are using, if any:

=> https://github.com/PHP-DI/PHP-DI/blob/323f76ec2aa70cb367d54f7dc9c1a5fe403a16a4/.php_cs#L21-L23

Extract:

```php
        'is_null' => [
            'use_yoda_style' => false,
        ],
```

#### Diff generated

```php
@@ @@
         $errorMessage = $this->isCompilable($value);
-        if ($errorMessage !== true) {
+        if (true !== $errorMessage) {
@@ @@
```

Expected: no changes.

The `if` should not be flipped since I disabled yoda conditions. I don't understand why php-cs-fixer changed its behavior, I did not change that code and I did not change the configuration.

keradus:
you have configured `is_null` fixer to use or not the yoda style, as `is_null($foo)` may change it to `null === $foo` or `$foo === null`, you have configuration for it. And that's it.
Presented code was not touched by `is_null` fixer, as there is no `is_null` invocation !
You are using `yoda_style` rule, configure it.

mnapoli:
I don't understand, this was not applying before. I maintain something has changed or there is a bug. FYI I am *not* using the `yoda_style`, it is not in my config file as you can see in the link, I've never used it and php-cs-fixer never offered to change my conditions before.

Anyway thanks.

mnapoli:
Well I was not crazy, I'm now getting this error on my CI:

>   [PhpCsFixer\ConfigurationException\InvalidConfigurationException]  
>   The rules contain unknown fixers (yoda_style).

This fix was not here before.

kubawerlos:
Travis has a job with`--prefer-lowest` and you have `"friendsofphp/php-cs-fixer": "^2.4"` as dependency in `composer.json` - `yoda_style` fixer was introduced in PHP CS Fixer 2.6, so you have to update `composer.json`.

julienfalque:
[...]

Issue #1112, opened 2015-03-28
Proposal: .php_cs.lock with list of fixers being (and not) applied


kalessil:
It would be helpful to have this file dumped after each CS Fixer run (similar to composer.lock file). 
E.g. for Php Inspections (EA Extended) to not apply "last comma in array" inspection to source code - just read the file and don't bother developers with extra setting management.

Some rough ideas for structure:

header_comment=yes
long_array_syntax=yes
strict=no

GrahamCampbell:
Very interesting idea. I'd vote for having it json formatted, like composer does.

Ping @bobthecow - you were concerned over bc breaks due to us introducing new fixers. This might be a good solution for you.

GrahamCampbell:
@kalessil A better way to do this as far as I'm concerned is for the lock file to just have the fixers to apply, rather than the whole yes/no deal, so we can just skip straight to that list instead of going through the config resolver like we do at the moment.

kalessil:
@GrahamCampbell : for the format it would be better ini/xml format as PHP/Java has functions from the box to work with these formats. To keep only applied fixers in lock file will work for me.

For file structure it can be more useful options: excluded resources (folders and files), preset id (psrN, Symfony). Maybe something else to state version/BC meta-data.

kalessil:
Update: JSON is fine, there is <a href="https://code.google.com/p/json-simple/">lightweight library</a> for java - just 20KB.

bobthecow:
:+1: Lock file sounds good to me, that'd fix all my concerns :)

keradus:
Regenerating this file after each PHP CS Fixer run is pointless - it must be generated once and then used. yaml is good with me too.

@GrahamCampbell , maybe create one such lock file and then no need to have it duplicated into StyleCI configuration?

One more thing: sb configured fixers inside `.php_cs`, then store it in lock file and change configuration. When running a fixer we will taki list from lock file even when config change ? And what if one run a fixer with `--fixers` option?
Such a lock file must be generated on demand, not always. Remember that there will be users that want to use full level like PSR2 or Symfony - even when new fixer appear.

bobthecow:
Something like `npm shrinkwrap` then?

kalessil:
Maybe one step further: declarative cs_fixer.json as composer again, and then lock-file just to state BC meta-info?

Then I can just use manifest and adjust my inspections accordingly, without any workarounds/

keradus:
[...]

PR #4566, opened 2019-10-01
Update symfony ruleset

gharlan:
https://github.com/symfony/symfony/blob/d2b66ff59295f47ad3448ea5bc9e826864da441f/.php_cs.dist#L13-L19


README.rst
```diff
@@ -1113,7 +1113,7 @@ Choose from the list of available rules:
 
   Replaces superfluous ``elseif`` with ``if``.
 
-* **no_superfluous_phpdoc_tags**
+* **no_superfluous_phpdoc_tags** [@Symfony, @PhpCsFixer]
```
gharlan: Do we want to adopt this here? In general the `@PhpCsFixer` ruleset is a superset of `@symfony`. But there would be many changes in this repo.
So what do you think about this rule? (I like it!)

julienfalque: :+1:

julienfalque:
Should this be merged in 2.15 rather than master?

SpacePossum:
I'm :+1:  for 2.15 without `no_superfluous_phpdoc_tags`,
+0 with it on `master`, I like all the other changes :)

gharlan:
Ok, for a first step I changed target to 2.15 and removed `no_superfluous_phpdoc_tags`.

src/RuleSet.php
```diff
@@ -174,7 +175,6 @@ final class RuleSet implements RuleSetInterface
             'ereg_to_preg' => true,
             'error_suppression' => true,
             'fopen_flag_order' => true,
-            'fopen_flags' => ['b_mode' => false],
```
SpacePossum: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4011 I think this should stay, it was disabled before in the symfony repo because it would add the `b`, now it will remove it (when we ship the next version of the tool that is)

SpacePossum:
Thanks looking good!

Maybe we could add `no_superfluous_phpdoc_tags` to the SF set with `['allow_mixed' => true, 'allow_unused_params' => true]` (through https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4505)
 as well now we added the config?

src/RuleSet.php
SpacePossum: done

SpacePossum:
Thank you @gharlan.

glensc:
neh, commit a0df28285e61c9fb52cca313ad64f2329a71a6ae (HEAD, origin/pr/4566, refs/bisect/bad) broke my setup ability to set `$rules['array_syntax'] = ['syntax' => 'long'];`
- https://github.com/glensc/php-cs-fixer-config/issues/2

glynnforrest:
@SpacePossum @julienfalque can you explain why this went into the 2.15 branch? An innocuous update from 2.15.3 to 2.15.4 broke a project I'm part of that uses `php-cs-fixer fix --dry-run` in CI.

I really appreciate all the work you do keeping up with the Symfony code style, but this going into a patch version feels unexpected! These feel like somewhat breaking changes that should at least go into the next minor version.

julienfalque:
@glynnforrest Please see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3398#issuecomment-355815211.

The purpose of a ruleset it to follow a specific code style without having to know what it applies exactly. If this code style evolves somehow, the ruleset becomes outdated and thus wrong so updating it feels more like a bugfix than a new feature to me.

glynnforrest:
Thanks @julienfalque, that explains it perfectly. We have multiple large codebases so each new rule has a wide-reaching impact. Perhaps it's better in our case to remove the Symfony ruleset and use our own to avoid any surprises.

> Please see #3398 (comment).

I appreciate seeing the backstory. It looks like this has been discussed already (https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1112), so I've got nothing more to add.

PR #4569, opened 2019-10-05
PhpdocOrderByValueFixer - Introduction

localheinz:
This PR

* [x] extracts a `PhpdocOrderByValueFixer`
* [x] deprecates the `PhpUnitOrderedCoversFixer`


README.rst
```diff
@@ -1398,9 +1398,19 @@ Choose from the list of available rules:
...
-* **php_unit_ordered_covers** [@PhpCsFixer]
+* **php_unit_ordered_annotations** [@PhpCsFixer]
 
-  Order ``@covers`` annotation of PHPUnit tests.
+  Order annotations of PHPUnit tests.
+
+  Configuration options:
+
+  - ``annotations`` (a subset of ``['covers', 'uses']``): list of annotations to
+    order, e.g. ``["covers"]``; defaults to ``[]``
```
localheinz: Should we include `covers` by default, or should the default be an empty `array`?

src/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixer.php
```diff
@@ -0,0 +1,167 @@
...
+                            'uses',
+                        ],
+                    ]
+                ),
+            ]
+        );
+    }
+
+    /**
+     * {@inheritdoc}
```
localheinz: Do we really need these DocBlocks?

README.rst
```diff
@@ -1402,9 +1402,19 @@ Choose from the list of available rules:
...
 
-* **php_unit_ordered_covers** [@PhpCsFixer]
+* **php_unit_ordered_annotations** [@PhpCsFixer]
 
-  Order ``@covers`` annotation of PHPUnit tests.
+  Order annotations of PHPUnit tests.
+
+  Configuration options:
+
+  - ``annotations`` (a subset of ``['covers', 'uses']``): list of annotations to
```
localheinz: Perhaps other annotations also make sense (for example, `@dataProvider`)?

src/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixer.php
julienfalque: We don't but I would leave it for consistency with the project. Maybe we'll revise this with #4500.

README.rst
SpacePossum: maybe add these: `@group`, `@author`, `@dataProvider` ?

README.rst
SpacePossum: I think `covers` as default is nice for the people migrating to the new fixer

SpacePossum:
I would like to merge this one, so if anyone has more this please let me know otherwise I resolve the minors and continue :)

tests/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixerTest.php
```diff
@@ -0,0 +1,306 @@
...
+                         * @dataProvider provide
+                         * @uses Bar
+                         */
+                        public function testMe() {}
+                    }
+                ',
+            ],
+        ];
+    }
+}
```
SpacePossum: maybe we can have a test with multiple annotations configured as well?

README.rst
localheinz: Crazy thing is that I heard from @epdenouden that some people write tests where the order in which data providers are configured matter.

😆

localheinz: @SpacePossum 

Not sure about `@author` - would this make sense outside of the scope of a `phpunit/phpunit` test case?

SpacePossum: well that is .. ehmm .... interesting XD    anyway... they can omit the option than : P

SpacePossum: Im definitely fine with leaving `author` out, I was more typing out loud than pushing specs :)

localheinz: @SpacePossum 

Had not been aware that `@author` is a thing as an annotation for `phpunit/phpunit`!

Screen Shot 2019-10-11 at 21 29 56

localheinz: Not a fan of using `@depends`, but does anyone know whether one can specify multiple `@depends`?

SpacePossum: I never new it was an alias, TIL :D

SpacePossum: https://phpunit.de/manual/6.5/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.examples.MultipleDependencies.php

some a sample of mulitple depends (not a big fan either btw)

tests/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixerTest.php
```diff
@@ -0,0 +1,954 @@
...
+                    /**
+                     * @uses Foo
+                     * @uses Bar
+                     * @covers Qux
+                     * @covers Baz
+                     */
+                    class FooTest extends \PHPUnit_Framework_TestCase {}
+                ',
+            ],
+            'preserve positions if other docblock parts are present' => [
```
localheinz: This test case currently fails.

🤔

README.rst
localheinz: @SpacePossum 

Added~

localheinz:
@julienfalque @SpacePossum 

Perhaps it makes sense to adjust the `PhpdocOrderFixer` to also take into account a range of other annotations?

I like

- having annotations in a specific order
- having annotations grouped (separated by an empty line)

What do you think?

SpacePossum:
I think that would be great :)
Do you want to incorporate the work done here in the new fixer into that one or keep these separated? I might be easier to have to prevent conflicts and make the configuration for the end user more easy... hmm... I might need to give it some more though, but it does sound good

tests/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixerTest.php
SpacePossum: I think we (you ;) ) should not need to trouble ourselves to much with `@author` annotations sticking to a piece of PHPDoc. 
For such details I think people can better refer to their VCS and only a list of authors per file can be handy. However, if they really want to keep the authors sticky to the rest of the text than they can remove the `@author` annotation from the configuration of the rule.

localheinz:
Maybe there’s room for more fixers?

- group annotations
- sort annotations within a group
- sort annotation groups

Personally, I would use all of them!

localheinz:
Build failures are currently related to failing to download static analysis tools.

README.rst
```diff
@@ -1445,9 +1445,20 @@ Choose from the list of available rules:
...
 
-* **php_unit_ordered_covers** [@PhpCsFixer]
+* **php_unit_ordered_annotations** [@PhpCsFixer]
 
-  Order ``@covers`` annotation of PHPUnit tests.
+  Order annotations of PHPUnit tests.
+
+  Configuration options:
+
+  - ``annotations`` (a subset of ``['author', 'covers', 'dataProvider', 'depends',
```
kubawerlos: What about `coversNothing` and `internal`? The more the merrier in this case, right?

src/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixer.php
```diff
@@ -0,0 +1,184 @@
...
+    {
+        return $tokens->isAllTokenKindsFound([T_CLASS, T_DOC_COMMENT]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function applyFix(\SplFileInfo $file, Tokens $tokens)
+    {
+        if (!\count($this->configuration['annotations'])) {
```
kubawerlos: Boolish, why not `[] === $this->configuration['annotations']`?

README.rst
keradus: ping @localheinz

keradus:
@localheinz , can you please rebase and handle newest code review ?

localheinz:
@keradus 

Apologies for the delay - will do!

README.rst
localheinz: @kubawerlos 

Not sure, why would anyone have multiple

- `@coversNothing`
- `@internal`

annotations, perhaps even followed by content?

Should I add them nonetheless?

Perhaps a case for an additional fixer, remove duplicate annotations?

tests/Fixer/PhpUnit/PhpUnitOrderedAnnotationsFixerTest.php
localheinz: @SpacePossum 

Fixed it, had something messed up in my expectations!

👍

localheinz:
@julienfalque @keradus @kubawerlos @SpacePossum 

I have

- rebased on top of `master`
- added handling of `@coversNothing`
- added handling of `@internal`

and I apologize for the delay.

If you find something else that needs to be addressed, I promise to sort it out quickly!

README.rst
kubawerlos: @localheinz oh, I was thinking the fixer will also order the groups of annotations :man_facepalming:

README.rst
```diff
@@ -1452,9 +1452,20 @@ Choose from the list of available rules:
...
 
-* **php_unit_ordered_covers** [@PhpCsFixer]
+* **php_unit_ordered_annotations** [@PhpCsFixer]
 
-  Order ``@covers`` annotation of PHPUnit tests.
+  Order annotations of PHPUnit tests.
+
+  Configuration options:
+
+  - ``annotations`` (a subset of ``['author', 'covers', 'coversNothing',
```
localheinz: Just thinking, since this fixer handles annotations which are not bound to `phpunit` test cases, such as

- `@author`
- `@internal`

and since we already have `PhpdocOrderFixer`, perhaps this fixer here should be renamed to `PhpdocOrderByValueFixer`?

Here's what I think:

### Current Fixers

- `PhpDocOrderFixer` groups annotations by name (`@param`, `@throws`, `@return`) and inserts an empty line between them

### New Fixers

- `PhpDocOrderByValueFixer` sorts configured annotations by value
- `PhpDocOrderByNameFixer` sorts configured annotations by name, and optionally inserts a blank line between them

### Deprecated Fixers

* `PhpDocOrderFixer` can be deprecated and proxy to the `PhpDocOrderByNameFixer`

localheinz: @julienfalque @keradus @kubawerlos @SpacePossum 

What do you think?

julienfalque: That's exactly what I was thinking when reading the changes, 👍 for me. Though I think we can merge this PR (after renaming the new fixer) and work on the other fixers in another PR?

localheinz: @julienfalque 

Sounds good!

Renamed the fixer!

julienfalque:
Thank you @localheinz.

localheinz:
Thank you, @julienfalque, @keradus, @kubawerlos, and @SpacePossum!

keradus:
kudos @localheinz

PR #4571, opened 2019-10-07
BlankLineBeforeStatementFixer - can now add blank lines before doc-comments

addiks:
This PR adds the "doccomment" token to the blank-line-before-statement-fixer.

Now the php-cs-fixer can automatically turn this:

```php
/** @var int $foo */
$foo = 123;
/** @var int $bar */
$bar = 456;
/** @var int $baz */
$baz = 789;
```

... into this:

```php
/** @var int $foo */
$foo = 123;

/** @var int $bar */
$bar = 456;

/** @var int $baz */
$baz = 789;
```

### DIFF
--- a/src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
+++ b/src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
@@ -43,6 +43,7 @@ final class BlankLineBeforeStatementFixer extends AbstractFixer implements Confi
         'continue' => T_CONTINUE,
         'declare' => T_DECLARE,
         'default' => T_DEFAULT,
+        'phpdoc' => T_DOC_COMMENT,
         'do' => T_DO,
         'exit' => T_EXIT,
         'for' => T_FOR,
--- a/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
+++ b/tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
@@ -1494,4 +1494,54 @@ enum UserStatus: string {
 ',
         ];
     }
+
+    /**
+     * @dataProvider provideFixWithDocCommentCases
+     */
+    public function testFixWithDocCommentCases(string $expected, string $input = null): void
+    {
+        $this->fixer->configure([
+            'statements' => ['phpdoc'],
+        ]);
+
+        $this->doTest($expected, $input);
+    }
+
+    public function provideFixWithDocCommentCases(): iterable
+    {
+        yield [
+            '<?php
+/** @var int $foo */
+$foo = 123;
+
+/** @var float $bar */
+$bar = 45.6;
+
+/** @var string */
+$baz = "789";
+',
+            '<?php
+/** @var int $foo */
+$foo = 123;
+/** @var float $bar */
+$bar = 45.6;
+/** @var string */
+$baz = "789";
+',
+        ];
+
+        yield [
+            '<?php
+/* header */
+
+/**
+ * Class description
+ */
+class Foo {
+    /** test */
+    public function bar() {}
+}
+',
+        ];
+    }
 }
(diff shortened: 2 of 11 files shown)


addiks:
Is there currently some issue with the CI?

```
λλλ phive packages

Phive 0.13.0 - Copyright (C) 2015-2019 by Arne Blankerts, Sebastian Heuer and Contributors

Fetching repository list

Downloading https://phar.io/data/repositories.xml

[ERROR]    Failed to download load https://api.github.com/repos/maglnet/ComposerRequireChecker/releases: HTTP Code 403 

The command "./dev-tools/install.sh" failed and exited with 4 during .

Your build has been stopped.
```

SpacePossum:
Looking into this a bit more, this fixer will only insert the empty line if the previous line ends with `}` or `;`, so it just might work after all. This makes sense for statements, not sure for PHPDocs. 
I still think it makes sense to make a dedicated fixer for this, like to hear input on this though. Even if we got to copy-paste relevant parts to a new fixer; it still looks like a nice feature and very promising solution.

addiks:
Honestly i don't know if this is the best solution or not. I do not have enough experience with this project to judge that. Currently i like it because it creates a feature i want just by adding one very simple line. That seems elegant to me.

Maybe i am more motivated to write a new fixer for this when someone more experienced crushes my hopes by showing how this simple solution fails in some situation i do not yet see. ;-)

In the meantime i write one or two actual new fixers for other features that i would like and explore the project that way...

SpacePossum:
I try to keep an eye out on the overall responsibilities for each fixer because we have/had a lot of issues with fixer doing to much, making these less flexible for later extension and adding configuration, causing priority issues we can't resolve and the only way out are BC changes on a next major which no one like because these ship rarely, causing issues and PR's to stale for months.

So...no, I didn't spot an issue with your code and I don't want to crush your hopes, sorry if it feels that 
way. What I did say about your code ` it still looks like a nice feature and very promising solution.` .

src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php
```diff
@@ -40,6 +40,7 @@ final class BlankLineBeforeStatementFixer extends AbstractFixer implements Confi
         'continue' => T_CONTINUE,
         'declare' => T_DECLARE,
         'default' => T_DEFAULT,
+        'doccomment' => T_DOC_COMMENT,
```
kubawerlos, line 43: Let's have the key named `doc_comment` (as there is already `require_once`).

localheinz:
@addiks 

Can you take another look, and ideally resolve conflicts?

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
```diff
@@ -397,6 +397,50 @@ public function provideFixWithDefaultCases()
...
+
+    /**
+     * @return array
+     */
+    public function provideFixWithDocCommentCases()
+    {
+        return [
+            [
+                '<?php
+/** @var int $foo */
```
kubawerlos, line 423: Why there is no line added before this doc comment?

kubawerlos (changes requested):
Please revert your changes on `logo.png`.

tests/Fixer/Whitespace/BlankLineBeforeStatementFixerTest.php
addiks: Because that is the file's doc-comment, which normally contains annotations like `@author`, it belongs at the top.
The purpose of this fixer is to separate doc-comments from other declarations above it, like here:
```php
$foo = 123;
/** @var int $baz */
$baz = 456;
```
The top-doccomment does not have another declaration above it, just the file's beginning (which it is normally describing). What should the file-doccomment be separated from?

I will revert the changes to `logo.png` as soon as I have a bit of time to set everything up again, most likely in the next few days. Maybe today. Thank you for the review.

kubawerlos: > What should the file-doccomment be separated from?

Could be from openning tab, but AFAIR there is other fixer to do that, also the same applies to code, not only PHPDoc.

addiks:
I have now reverted the changes to `logo.png`. This was a tricky one because git thought it was a text file and changed a CRLF to LF in that file, corrupting it in the process. The line `* text=auto eol=lf`in `.gitattributes` defines that all files are text files and that the mentioned automatic replacement must happen for all files, including PNG files. I have added an exception for PNG files to the `.gitattributes` files so that it stops corrupting the PNG file(s).

SpacePossum:
Thank you @addiks.

Issue #4574, opened 2019-10-08
After v2.13.0  cann't run in PHP 7.0.0


calchen:
#### The PHP version you are using (`$ php -v`):

```
PHP 7.0.0 (cli) (built: Mar 25 2018 21:06:27) ( ZTS )
Copyright (c) 1997-2015 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2015 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2015, by Zend Technologies
    with Xdebug v2.5.5, Copyright (c) 2002-2017, by Derick Rethans
```

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.13.1

#### The command you use to run PHP CS Fixer:
=> none

#### The configuration file you are using, if any:
=> none

After v2.13.0 (since v2.13.1). Dependence on  symfony/console require `^3.4.17 || ^4.1.6`, but symfony/console require php `^5.5.9|>=7.0.8`([here](https://github.com/symfony/console/blob/v3.4.17/composer.json)), so version after v2.13.0 could only run in PHP 7.0.8 or after.

calchen:
here is the pull request #4575

SpacePossum:
thanks for reporting, I wonder though.. hmmm shouldn't composer allow 7.0.8 for  ^7.0 https://getcomposer.org/doc/articles/versions.md#caret-version-range- ?

SpacePossum:
closing following https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4575 , let me know if can help you out with installing!

calchen:
I will follow Julien's advice, update my php version. And I think ^7.0 means start with 7 and greater than or equal to 7.0.0 ( >= 7.0.0 and < 8.0.0 ). The result of my test should confirm my guess.

```
root@c2e3c3254dc4:/app# php -v
PHP 7.0.0 (cli) (built: Dec 11 2015 00:57:25) ( NTS )
Copyright (c) 1997-2015 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2015 Zend Technologies

root@c2e3c3254dc4:/app# composer --version
Do not run Composer as root/super user! See https://getcomposer.org/root for details
Composer version 1.9.0 2019-08-02 20:55:32

root@c2e3c3254dc4:/app# composer init -q

[...]

PR #4575, opened 2019-10-08
Update composer.json

calchen:
here is my issue #4574

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^5.6 || >=7.0.8",
         "ext-json": "*",
         "ext-tokenizer": "*",
         "composer/semver": "^1.4",


composer.json
```diff
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": "^5.6 || ^7.0",
+        "php": "^5.6 || >=7.0.8",
```
kubawerlos:
```suggestion
        "php": "^5.6 || ^7.0.8",
```

SpacePossum:
Closing following the explaining of Julien. This project does not require PHP 7.0.8; the console component does. Composer will figure this out for you if you install on a PHP supported by both.

calchen:
> The current constraint already allows _any_ PHP 7 version, including yours (7.0.0). Not only will this change not fix your issue, it will add a new constraint to prevent your specific version (`>=7.0.8` and `^7.0.8` both reject version `7.0.0`).
> 
> Allowing PHP 7.0.0 would require to change the constraint in `symfony/console`, but if they claim they need PHP to be at least 7.0.8, they probably have a good reason.
> 
> I don't think there is much to do here. I'd suggest you update your PHP version.

thanks for your explaining, i will update my php version

calchen:
@julienfalque thanks for your explaining, i will update my php version

Issue #4133, opened 2018-12-04
Return error code on invalid file


julienfalque:
Currently, when PHP CS Fixer processes a file and somehow introduces a syntax error in it (because of a bug), the command exits with `0`. IMO the command should exit with an error code in such case.

Maybe we could also add a new option to make the command fail even for files that are invalid _before_ being processed. WDYT?

SpacePossum:
The first case sounds like a bug to me :)

Second case, sounds logical to me, so +1 as well

keradus:
> Currently, when PHP CS Fixer processes a file and somehow introduces a syntax error in it (because of a bug), the command exits with `0`.

we have the fail-safe mechanism, that if "fixing" process would corrupt the file, we don't apply it. For that, the whole CLI took worked. I agree it may make sense to raise warning in such case, maybe even we already have it, not sure

> Maybe we could also add a new option to make the command fail even for files that are invalid before being processed. WDYT?

we already have:
> 4 Some files have invalid syntax (only in dry-run mode).

SpacePossum:
> we have the fail-safe mechanism, that if "fixing" process would corrupt the file, we don't apply it. For that, the whole CLI took worked.

If a fixer find something that it should fix and than fails and causes an internal exception, we indeed do not apply it, however the file is than not fixed so the whole CI didn't work as expected (the result is that a file still needs fixing which wasn't done).

dmvdbrugge:
I agree with @SpacePossum here. In a pre-commit hook, I want to abort when a file could not be fixed (aka it failed fixing). If however the cli-tool says it succeeds (exit 0), how can I know that it failed?

dmvdbrugge:
To add even more to this: on lint exception, `--dry-run --diff` exits with 0 as well even though it has a diff and thus tried to fix the file. (So the 8 flag should've been set at least imho.)

keradus:
> If a fixer find something that it should fix and than fails and causes an internal exception, we indeed do not apply it, however the file is than not fixed so the whole CI didn't work as expected (the result is that a file still needs fixing which wasn't done).

We dont know that. Internal exception can happen on anything. But good reasoning for linting-after-fixing error

SpacePossum:
First part (exit code on lint failure after fixing) has been covered, second FR is still up for discussion.

keradus:
What is a 2nd FR part here?

julienfalque:
[...]

PR #4577, opened 2019-10-09
Fix command exit code on lint error after fixing fix.

SpacePossum:
@ see https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4133

From the docs `Exit code 64 - Exception raised within the application.` so this change is not a BC but a bug fix.

### DIFF
--- a/src/Console/Command/FixCommandExitStatusCalculator.php
+++ b/src/Console/Command/FixCommandExitStatusCalculator.php
@@ -31,10 +31,11 @@ final class FixCommandExitStatusCalculator
      * @param bool $hasChangedFiles
      * @param bool $hasInvalidErrors
      * @param bool $hasExceptionErrors
+     * @param bool $hasLintErrorsAfterFixing
      *
      * @return int
      */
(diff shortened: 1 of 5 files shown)


src/Console/Command/FixCommandExitStatusCalculator.php
```diff
@@ -31,10 +31,11 @@ final class FixCommandExitStatusCalculator
      * @param bool $hasChangedFiles
      * @param bool $hasInvalidErrors
      * @param bool $hasExceptionErrors
+     * @param bool $hasLintErrorsAfterFixing
      *
      * @return int
      */
-    public function calculate($isDryRun, $hasChangedFiles, $hasInvalidErrors, $hasExceptionErrors)
+    public function calculate($isDryRun, $hasChangedFiles, $hasInvalidErrors, $hasExceptionErrors, $hasLintErrorsAfterFixing)
```
SpacePossum: internal class, so no BC promise

PR #4579, opened 2019-10-10
Two new fixers: Blank line before else/elseif and before catch/finally

addiks:
This PR add's two similar fixers:
 - Blank line before else and elseif code-blocks
 - Blank line before catch and finally code-blocks

The first fixer (blank_line_before_else_block) turns this:
```php
if ($a) {
    foo();
} elseif ($b) {
    bar();
} else {
    baz();
}
```
... into this:
```php
if ($a) {
    foo();

} elseif ($b) {
    bar();

} else {
    baz();
}
```
And the second fixer (blank_line_before_catch_block) turns this:
```php
try {
    foo();
} catch (\Exception $e) {
    bar();
} finally {
    baz();
}
```
... into this:
```php
try {
    foo();

} catch (\Exception $e) {
    bar();

} finally {
    baz();
}
```

The reasoning on why someone (including me) might want this is relatively simple: Statements in a conditional code-block is more closely related to the condition of said code-block then to the condition of the next code-block or the previous code-block, so they (IMO) should be separated by a blank line.

I thought about merging these two very similar fixers into one fixer because the reasoning is the same for both cases: To group code-blocks by explicitly separating them from other blocks with a blank line. But these two fixers concern two different types of code-blocks (conditional and exception-handling), so a user may want to use one but not the other, even if i do not see the reasoning for that yet.
Instead of using two separate fixers i could also have used one and made it configurable, but that seems more complicated then just having two fixers and they *do* concern two different use-cases, so they could become different in the future.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -304,6 +304,10 @@ Choose from the list of available rules:
   Ensure there is no code on the same line as the PHP open tag and it is
   followed by a blank line.
 
+* **blank_line_before_else_block**
+
+  An empty line feed must precede any else or elseif codeblock.
+
 * **blank_line_before_return**
 
   An empty line feed should precede a return statement. DEPRECATED: use
--- a/README.rst
+++ b/README.rst
@@ -304,6 +304,10 @@ Choose from the list of available rules:
   Ensure there is no code on the same line as the PHP open tag and it is
   followed by a blank line.
 
+* **blank_line_before_catch_block**
+
+  An empty line feed must precede any catch or finally codeblock.
+
 * **blank_line_before_else_block**
 
   An empty line feed must precede any else or elseif codeblock.
--- a/src/Fixer/Whitespace/BlankLineBeforeCatchBlockFixer.php
+++ b/src/Fixer/Whitespace/BlankLineBeforeCatchBlockFixer.php
@@ -19,7 +19,7 @@ use PhpCsFixer\FixerDefinition\FixerDefinition;
 use PhpCsFixer\Tokenizer\Token;
 use PhpCsFixer\Tokenizer\Tokens;
 
-final class BlankLineBeforeElseBlockFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
+final class BlankLineBeforeCatchBlockFixer extends AbstractFixer implements WhitespacesAwareFixerInterface
 {
     /**
      * {@inheritdoc}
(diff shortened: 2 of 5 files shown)


SpacePossum:
Hi and thanks for the PR and all the work you out into it :)

I follow your reasoning, although I wouldn't use this style. For this project the aim is to help people out formatting their code towards a standard used by a bigger community. Do you feel the style your fixers change code is in use by such community?

addiks:
No, i don't think that this style is currently in use by any big community at large. I do however think that this is more by chance than by deliberate choice. I think most developers do not think about this and just copy the style from others (which actually is the right thing to do in most cases). I have not yet seen this being talked about in any style guide, although i may have missed something. Some people say that they prefer the variant without the blanks but provide no reason, so i think that it is just from habit.

As for reasoning, the only reasoning against this that i have seen is that blank lines are bad per'se because you should see as much code as possible at once, which i think is a bad reason. Code is not the most understandable when it is compressed the most. 

If something is too complex to be understood at once, it must be understood in parts. To do that, each part should be perceived separate from the other parts. Blank lines are used to separate parts of the program. Different parts should be separated when the cohesion between the different perceived parts is lower then the cohesion between the lines inside the block. The cohesion between a block and it's condition is higher then between it and the code before the condition, 
so it should get separated from said surrounding code, using a blank line.

Maybe i am wrong with this, then this PR should not be merged. But if so, i would like to know the reason behind it. Do you think it would help to ask others what they think about this? I could just ask anyone interested on twitter and see what people think using a poll. On the other hand, i dont want to make this any bigger then it needs to be. Do you maybe have any resources or discussions about this?

localheinz:
@addiks 

Why not adjust the [`PhpCsFixer\Fixer\Whitespace\BlankLineBeforeStatementFixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.15.3/src/Fixer/Whitespace/BlankLineBeforeStatementFixer.php) instead?

addiks:
Because the blank-line-before-statement-fixer inserts the blank line **directly** before the statement. In these cases, the blank line should be inserted before the `}` character that is before the actual statement (T_ELSE, T_ELSEIF, T_CATCH and T_FINALLY).

Blank-line-before-statement-fixer:
```php
    foo();
} 

else {
    ...
```
Blank-line-before-else-fixer:
```php
    foo();

} else {
    ...
```

PR #4584, opened 2019-10-11
Increase PHPStan level to 1

julienfalque:


.travis.yml
```diff
@@ -51,7 +51,7 @@ jobs:
                     | grep -v tests/TestCase.php \
                     && (echo "UNKNOWN FILES DETECTED" && travis_terminate 1) || echo "NO UNKNOWN FILES"
                 - ./check_trailing_spaces.sh || travis_terminate 1
-                - ./dev-tools/tools/phpstan analyse
+                - php -d auto_prepend_file=dev-tools/vendor/autoload.php ./dev-tools/tools/phpstan analyse
```
julienfalque: Prepeding the `dev-tools/vendor/autoload.php` is required so that the PHPStan Phar file is able to load `phpstan/phpstan-phpunit` extension but this feels very hacky and I can't say there is no downside. Any better idea?

phpstan.neon
```diff
@@ -1,5 +1,8 @@
+includes:
+    - dev-tools/vendor/phpstan/phpstan-phpunit/extension.neon
```
julienfalque: This extension helps PHPStan understand how PHPUnit works. Specifically it helps it know that `static::markTestSkipped()` ends the test execution.

.travis.yml
SpacePossum: maybe use a dedicated `composer.json` for PHPStan in a nested dir in `dev-tools` ?

localheinz: Perhaps specify using https://github.com/phpstan/phpstan#specify-paths-to-scan?

julienfalque: @SpacePossum Do you mean install PHPStan itself using Composer instead of Phive? Otherwise it won't fix the issue I think.

@localheinz I just tried and I doesn't work. As far as I can tell, these options are for parts of the scanned codebase that isn't autoloaded by Composer, not for extensions to PHPStan runtime.

SpacePossum: indeed, given the issues we have with Phive lately I wouldn't mind switch (back) to composer installs

kubawerlos: If we switch back to composer installs I'd vote to move all dev dependencies to `dev-tools/composer.json` and merge it to the main `composer.json` from there.

keradus:
Thank you @julienfalque.

Issue #4451, opened 2019-06-19
BracesFixer adds newline to dynamic static method call


ShiraNai7:
#### The PHP version you are using (`$ php -v`):
=> 7.3.6

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.15.1

#### The command you use to run PHP CS Fixer:
=> php php-cs-fixer.phar fix --rules=@PSR2 src

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php

function example()
{
    SomeClass::{$method}(new \stdClass());
}

```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php

function example()
{
    SomeClass::{$method}
    (new \stdClass());
}

```

 * with the changes you expected instead:
no changes expected

SpacePossum:
Thanks for reporting this bug @ShiraNai7 !
I've proposed a bug fix for it.

PR #4586, opened 2019-10-11
BracesFixer - handle dynamic static method call

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4451


keradus:
Thank you @SpacePossum.

Issue #4383, opened 2019-04-19
"Skipped" status in progress output is confusing


teohhanhui:
When the file is "skipped" because of the cache, the correct output should be `.` for "no changes", not `S` for "Skipped".

teohhanhui:
Rationale: The user only cares about whether the file is okay, or not. Imagine if phpunit or behat has a cache, and decides to output "Skipped" for cached tests. That would confuse the heck out of everyone.

SpacePossum:
Hi and thanks for the suggestion,

It's always helpful to hear what the end users find confusing and what can be tweaked, especially if you work on a project for a while as you might get blind for the details!

If PHPUnit had a cache and cached the test results I would prefer knowing it through some feedback. As such I think have different indicators for different states isn't that bad, but I agree that we don't explain the _why_ something is skipped.

I made a proposal to clarify the state, let me know what you think :)

teohhanhui:
As a matter of fact, PHPUnit does have a cache now, and it didn't change the progress output, as far as I could tell.

kubawerlos:
@teohhanhui PHPUnit uses the cache only to change the order of tests - it caches the results to be able to run the failed test first in the next run. No tests is skipped because it is cached.

Thinking from a perspective of PHPUnit user who discovers PHP CS Fixer when seeing `S` in progress then might think the file was skipped because it was marked to be skipped (like we can do in PHPUnit) and not because the file was already checked and have no changed since.

PR #4587, opened 2019-10-11
Output details - Explain why a file was skipped

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4383

### DIFF
--- a/src/Console/Output/ProcessOutput.php
+++ b/src/Console/Output/ProcessOutput.php
@@ -30,8 +30,8 @@ final class ProcessOutput implements ProcessOutputInterface
      */
     private static $eventStatusMap = [
         FixerFileProcessedEvent::STATUS_UNKNOWN => ['symbol' => '?', 'format' => '%s', 'description' => 'unknown'],
-        FixerFileProcessedEvent::STATUS_INVALID => ['symbol' => 'I', 'format' => '<bg=red>%s</bg=red>', 'description' => 'invalid file syntax, file ignored'],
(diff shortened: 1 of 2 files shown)


julienfalque:
I think a file can also be skipped when it's empty, regardless of the cache.

SpacePossum:
that is definitely true, but I decided to leave it out of the explanation because it is very rare case (who has empty `.php` files in their projects? ;) )

julienfalque:
Well... it happens :p

Seing an "S" for an empty file can be confusing, especially when cache is enabled. It will be worse if the output says that "S" is only for files skipped because of the cache.

SpacePossum:
@juliendufresne I will update :+1: 

@teohhanhui dont know why you throw that smiley at me, at least I tried and actually put time in your complaining, if you cannot be bothered to state what the issue is here now than maybe open a PR yourself? thanks

teohhanhui:
Sorry. I just find it confusing. Your work is definitely appreciated.

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4383#issuecomment-541212621

> As a matter of fact, PHPUnit does have a result cache now, and it didn't change the progress output (still just a dot), as far as I could tell.

teohhanhui:
My suggestion:

 We should only use `S` for empty files and such. Use `.` as usual for successful files that are "skipped" because of the cache.

GrahamCampbell:
This should probably target 2.16 or 3.x?

keradus:
docs improvement can go to lowest supported branch.

keradus:
Thank you @SpacePossum.

PR #4590, opened 2019-10-12
SwitchContinueToBreakFixer - Introduction

SpacePossum:
Description of switch_continue_to_break rule.
Switch case must not be ended with `continue` but with `break`.

```diff
--- Original
+++ New
@@ -1,5 +1,5 @@
<?php
    switch ($foo) {
        case 1:
-           continue;
+           break;
    }
```

```diff
--- Original
+++ New
@@ -1,15 +1,15 @@
<?php
switch ($foo) {
    case 1:
        while($bar) {
            do {
-                continue 3;
+                break 3;
            } while(false);
            
            if ($foo + 1 > 3) {
                continue;
            }
            
-            continue 2;
+            break 2;
        }
}
```

@ see https://www.php.net/manual/en/migration73.incompatible.php#migration73.incompatible.core.continue-targeting-switch

https://wiki.php.net/rfc/continue_on_switch_deprecation

https://github.com/php/php-src/pull/3364

### DIFF
--- a/src/Fixer/ControlStructure/SwitchContinueToBreakFixer.php
+++ b/src/Fixer/ControlStructure/SwitchContinueToBreakFixer.php
@@ -1,0 +1,259 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\ControlStructure;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Token;
+use PhpCsFixer\Tokenizer\Tokens;
+
+final class SwitchContinueToBreakFixer extends AbstractFixer
+{
+    private $switchLevels = [];
+
+    /**
+     * {@inheritdoc}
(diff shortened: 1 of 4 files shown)


tests/Fixer/ControlStructure/SwitchContinueToBreakFixerTest.php
```diff
@@ -0,0 +1,177 @@
...
+            ],
+            'do not fix cases' => [
+                '<?php
+switch($a) {
+    case 1:
+        while (false) {
+            continue;
+        }
+
+        while (false) continue 2;
```
julienfalque: Shouldn't this case be fixed as well? `continue 2` is targeting the parent `switch`. I would even make it a dedicated test case.

SpacePossum: thanks, fixed!

src/Fixer/ControlStructure/SwitchContinueToBreakFixer.php
```diff
@@ -0,0 +1,244 @@
...
+
+        $afterFollowingContinueIndex = $tokens->getNextMeaningfulToken($followingContinueIndex);
+
+        if (!$tokens[$afterFollowingContinueIndex]->equals(';')) {
+            return $afterFollowingContinueIndex; // if next not is `;` return without fixing, for example `continue 1 ? ><?php + $a;`
+        }
+
+        // check if continue {jump} targets a switch statement and if so fix it
+
+        $jump = (int) $followingContinueToken->getContent();
```
SpacePossum: TODO: test with `1_000` etc. on PHP7.4 https://wiki.php.net/rfc/numeric_literal_separator

SpacePossum: added

SpacePossum:
5.6 ..... ... ..  *sigh*

Issue #4511, opened 2019-08-19
Issues with phpdoc_to_return_type


nicolas-grekas:
I'm playing with this rule on the Symfony codebase, here are some issues I found while doing so:

- [x] `@return false` should become `: bool` (#4580)
- [x]  `@return object[][]` should become `: array` (#4589)
- [x]  `@return string[]|int[]` should become `: array` (#4591)
- [x]  `@return \Traversable|CacheItem[]` should become `: iterable` (#4591)
- [x]  `@return iterable|VoterInterface[]` should become `: iterable` (#4591)
- [ ]  `@return $this` should become `: object` (related to #3932), unless the method is final, in which case it should be `: self` to allow using it on PHP 7.1
- [x]  `@return static` should become `: object`, unless the method is final, in which case it should be `: self` to allow using it on PHP 7.1

For the last two, we've decided it should be the rule for the Symfony codebase, so could be a separate rule maybe (forcing `: object` when a `return $this`/`return static` annotation is found and a return type is found too)

Note that I find a lot of false-positive editions related to return-type covariance: eg a child method declares `@return Foo[]`, which leads to `: array`, while the parent says `: iterable`. But I guess we're hitting the limit of the tool and that's fine, there must be some.

SpacePossum:
checked off `static` as I think we handle is nicely now (including on [PHP8](https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/3057facfa4c5a70d2bcaec7c50069381b09de2f6) )

Wirone:
[I looked at tests](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/tests/Fixer/FunctionNotation/PhpdocToReturnTypeFixerTest.php) and I believe the only case missing is `@return $this`. I've marked todo item about `static` as resolved.

Wirone:
@nicolas-grekas what's the current status from the Symfony side? I believe you don't need the support for PHP 7.1 (`: object` return type). Maybe we can just close this? 🙂

nicolas-grekas:
Correct, thanks for the reminder.

PR #4591, opened 2019-10-13
Refactor conversion of PHPDoc to type declarations

julienfalque:
This PR:
* refactors parts of `PhpdocToParamTypeFixer` and `PhpdocToReturnTypeFixer` to ease sharing common logic.
* implements some points of #4511

### DIFF
--- a/src/DocBlock/Annotation.php
+++ b/src/DocBlock/Annotation.php
@@ -13,6 +13,8 @@
 namespace PhpCsFixer\DocBlock;
 
 use PhpCsFixer\Preg;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\NamespaceAnalysis;
+use PhpCsFixer\Tokenizer\Analyzer\Analysis\NamespaceUseAnalysis;
 
 /**
  * This represents an entire annotation from a docblock.
@@ -24,41 +26,6 @@ use PhpCsFixer\Preg;
  */
 class Annotation
 {
-    /**
-     * Regex to match any types, shall be used with `x` modifier.
-     *
-     * @internal
-     */
-    const REGEX_TYPES = '
-    # <simple> is any non-array, non-generic, non-alternated type, eg `int` or `\Foo`
-    # <array> is array of <simple>, eg `int[]` or `\Foo[]`
-    # <generic> is generic collection type, like `array<string, int>`, `Collection<Item>` and more complex like `Collection<int, \null|SubCollection<string>>`
-    # <type> is <simple>, <array> or <generic> type, like `int`, `bool[]` or `Collection<ItemKey, ItemVal>`
-    # <types> is one or more types alternated via `|`, like `int|bool[]|Collection<ItemKey, ItemVal>`
-    (?<types>
-        (?<type>
-            (?<array>
-                (?&simple)(\[\])*
-            )
-            |
-            (?<simple>
-                [@$?]?[\\\\\w]+
-            )
-            |
-            (?<generic>
-                (?&simple)
-                <
-                    (?:(?&types),\s*)?(?:(?&types)|(?&generic))
-                >
-            )
-        )
-        (?:
-            \|
-            (?:(?&simple)|(?&array)|(?&generic))
-        )*
-    )
-    ';
-
     /**
      * All the annotation tag names with types.
      *
@@ -119,13 +86,27 @@ class Annotation
     private $types;
 
     /**
+     * @var null|NamespaceAnalysis
+     */
+    private $namespace;
+
+    /**
+     * @var NamespaceUseAnalysis[]
+     */
+    private $namespaceUses;
+
+    /**
      * Create a new line instance.
      *
-     * @param Line[] $lines
+     * @param Line[]                 $lines
+     * @param null|NamespaceAnalysis $namespace
+     * @param NamespaceUseAnalysis[] $namespaceUses
      */
-    public function __construct(array $lines)
+    public function __construct(array $lines, $namespace = null, array $namespaceUses = [])
     {
         $this->lines = array_values($lines);
+        $this->namespace = $namespace;
+        $this->namespaceUses = $namespaceUses;
 
         $keys = array_keys($lines);
 
@@ -188,6 +169,33 @@ class Annotation
     }
 
     /**
+     * @return TypeExpression
+     *
+     * @internal
+     */
+    public function getTypeExpression()
+    {
+        return new TypeExpression($this->getTypesContent(), $this->namespace, $this->namespaceUses);
+    }
+
+    /**
+     * @return null|string
+     *
+     * @internal
+     */
+    public function getVariableName()
+    {
+        $type = preg_quote($this->getTypesContent(), '/');
+        $regex = "/@{$this->tag->getName()}\\s+{$type}\\s+(?<variable>\\$.+?)(?:[\\s*]|$)/";
+
+        if (Preg::match($regex, $this->lines[0]->getContent(), $matches)) {
+            return $matches['variable'];
+        }
+
+        return null;
+    }
+
+    /**
      * Get the types associated with this annotation.
      *
      * @return string[]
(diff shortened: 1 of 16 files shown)


kubawerlos:
Could we instead of abstract fixers introduce - as @SpacePossum suggested one day - a namespace of manipulators?

julienfalque:
Not sure what you mean exacly?

kubawerlos:
Similarly like we have `Tokenizer` we can create `Manipulator` and store there classes to be called by `PhpdocToParamTypeFixer` and `PhpdocToReturnTypeFixer`.

`AbstractPhpdocToTypeDeclarationFixer` ~does not have any property~ has properties that all are in practice constants, thus the class is set of functions that do stuff for the fixers, similar as all [analyzers](https://github.com/FriendsOfPHP/PHP-CS-Fixer/tree/v2.15.3/src/Tokenizer/Analyzer).

julienfalque:
Indeed, some parts of the new abstract class could probably be extracted to some "analyzer" class, but some are quite specific to these fixers. I'll have a closer look shortly.

keradus:
as this PR is also fixing a bugfix from 2.16 line, shouldn't it actually target 2.16.x as well ?

julienfalque:
This PR also introduces new public API in non-internal code (`DocBlock`) so it should be released in a new minor version according to SemVer. Though these new API are more of an detail of this refactoring than an actual new feature, so I'm fine merging this into 2.16. What do you prefer?

keradus:
can we backport bugfix only to 2.15?

julienfalque:
Well, at the time I created the PR, 2.16 was not released yet so targeting branch master was fine.

The initial purpose of this PR was to refactor things to ease maintenance. While working on it, I noticed it made implementing parts of #4511 easy. When I noticed it also fixes #4605, the work on this PR was already finished. I only added the test cases, which passed without further work. I'm not sure fixing those bugs without leveraging this refactoring is easy. I'll give it a try.

julienfalque:
Tests fixed in #4605.

GrahamCampbell:
This PR looks good, even if the 2nd follow-up PR to add a new fixer is not merged.

GrahamCampbell:
There are merge conflicts. This needs to be rebased, again. 😿

GrahamCampbell:
In fact, it looks like https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5092 kinda broke some of the sharing..

GrahamCampbell:
Though I guess we can regard adding the config stuff back in as a new feature.

GrahamCampbell:
// cc @SpacePossum

src/DocBlock/Annotation.php
```diff
@@ -185,6 +166,33 @@ public function getTag()
         return $this->tag;
     }
 
+    /**
+     * @return TypeExpression
+     *
+     * @internal until 2.17
```
GrahamCampbell, line 172: I don't think we need to do this. Just write internal, and we can delete this tag in 2.17 if we want, or indeed not!

GrahamCampbell:
NB, we will also need to add an example showing the scalar types config in use (for the tests to pass when merged up to 2.16). Something like:

```
                new VersionSpecificCodeSample(
                    '<?php
/** @param string|null $bar */
function my_foo($bar)
',
                    new VersionSpecification(70100),
                    ['scalar_types' => false]
                ),
```

src/DocBlock/Annotation.php
julienfalque: I prefer keeping the version so that if we forget to remove the tag, one still knows this is not internal anymore on 2.17 and higher.

julienfalque:
Rebased.

This PR now reverts removal of option `scalar_types` from #5092. I don't think the option is a much useful feature but the actual issue was that it was not used. Since this is a documented option, if we want to change/remove it, we should deprecate it first.

To avoid having a release that removes the option then a subsequent one that introduces it again, I suggest we include this in the upcoming 2.16.x release.

> we will also need to add an example showing the scalar types config in use

Added.

GrahamCampbell:
Ping @SpacePossum. Can this be merged now?

GrahamCampbell:
Looks like there are merge conflicts now. // cc @julienfalque

GrahamCampbell:
:+1:

GrahamCampbell:
Gah, conflicts again. NB This PR looks great to me @SpacePossum. Would be good to get this merged to avoid repeated merge conflicts.

GrahamCampbell:
Oh, this PR also needs updating to support the mixed type. // cc @SpacePossum

GrahamCampbell:
Would something like this do:

image

julienfalque:
I rebased this PR to fix the conflict but did not apply any extra change. Let's merge this first and work on PHP 8 compatibility separately please.

GrahamCampbell:
The reason I suggest that change is because support for mixed was added a few days ago, so this PR should not now break it.

GrahamCampbell:
I think the php 8 tests now fail after this pr, but were passing before.

GrahamCampbell (changes requested):
Broken unit tests on php 8

julienfalque:
@GrahamCampbell My bad, I didn't notice this was already supported on upstream branch, fixed as you suggested.

GrahamCampbell:
Needs rebasing again, against 2.17.

julienfalque:
Actually, if I remember correctly, all "new features" are marked `@internal until 2.18` (should it be 3.0?) so it can still be merged into 2.16.

GrahamCampbell:
I think 2.17 is the right place for this PR.

julienfalque:
Why?

GrahamCampbell:
Because it adds new features, and 2.16 should only be getting bug fixes?

julienfalque:
The PR:
* adds `scalar_types` option to `phpdoc_to_param_type` rule, but actually this is a revert of #5092 which can be considered a BC break
* adds new APIs that are marked `@internal` for now
* improves an existing fixer, which is something we use to merge in lower branches.

So IMO everything can be merged in 2.16. What do you think @SpacePossum @keradus @kubawerlos?

src/DocBlock/Annotation.php
```diff
@@ -187,6 +168,33 @@ public function getTag()
         return $this->tag;
     }
 
+    /**
+     * @return TypeExpression
+     *
+     * @internal until 2.18
```
GrahamCampbell, line 174: 2.19

src/DocBlock/Annotation.php
```diff
@@ -187,6 +168,33 @@ public function getTag()
...
+     */
+    public function getTypeExpression()
+    {
+        return new TypeExpression($this->getTypesContent(), $this->namespace, $this->namespaceUses);
+    }
+
+    /**
+     * @return null|string
+     *
+     * @internal until 2.18
```
GrahamCampbell, line 184: 2.19

GrahamCampbell (changes requested):
Have been testing this PR for a number of weeks, we no issues. Left just a couple of tiny comments.

src/DocBlock/Annotation.php
julienfalque: @keradus Do you plan a 2.19 release or shall we write 3.0 here?

julienfalque: I assumed 2.19 for now.

src/DocBlock/Annotation.php
```diff
@@ -187,6 +168,33 @@ public function getTag()
         return $this->tag;
     }
 
+    /**
+     * @return TypeExpression
+     *
+     * @internal until 2.19
```
keradus, line 174:
```suggestion
     * @internal
```

to speed up review. let's separately discuss if we would like to make it public

src/DocBlock/Annotation.php
```diff
@@ -187,6 +168,33 @@ public function getTag()
...
+     */
+    public function getTypeExpression()
+    {
+        return new TypeExpression($this->getTypesContent(), $this->namespace, $this->namespaceUses);
+    }
+
+    /**
+     * @return null|string
+     *
+     * @internal until 2.19
```
keradus, line 184:
```suggestion
     * @internal
```

to speed up review. let's separately discuss if we would like to make it applied for BC promise

keradus (approved):
crazy amount of work. kudos!

keradus:
Thank you @julienfalque.

PR #4592, opened 2019-10-13
PhpdocToPropertyTypeFixer - introduction

julienfalque:
This PR includes #4591 and depends on #4593 which should both be merged first.


SpacePossum:
:+1:  looking very nice :)

some minors I found;
the fixer crashes on abstract function like;
>abstract public function getFoo(): string;

which seems to be because;
```
            if ($tokens[$index]->isGivenKind(T_FUNCTION)) {
                $index = $tokens->getNextTokenOfKind($index, ['{']);
```

secondly, when there are a of candidate the fixer doesn't catch them all in the first run. Test sample;

```php
<?php

class ActivityLog
{
    /**
     * @var int
     */
    private $id;

    /**
     * @var string
     */
    private $id1;

    /**
     * @var string
     */
    private $id2;

    /**
     * @var int
     */
    private $id3;

    /**
     * @var string
     */
    private $id4;

    /**
     * @var string
     */
    private $id5;

    /**
     * @var string
     */
    private $id6;

    /**
     * @var string
     */
    private $id7;

    /**
     * @var int
     */
    private $id8;

    /**
     * @var string
     */
    private $id9;

    /**
     * @var int|null
     */
    private $id410;
}
```

julienfalque:
Thanks for the test cases, I fixed them :)

src/AbstractFixer.php
```diff
@@ -97,6 +97,9 @@ public function getName()
 
     /**
      * {@inheritdoc}
+     *
+     * Must run before PhpdocAlignFixer.
+     * Must run after CommentToPhpdocFixer, PhpdocIndentFixer, PhpdocScalarFixer, PhpdocToCommentFixer, PhpdocTypesFixer.
```
SpacePossum, line 102: I don't think this is true for every subclass of the AbstractFixer, at least not in design, I think it is better to document on the child classes only.

julienfalque: Fixed. Not sure why I added the comment here...

duxthefux:
+1 on that. Any estimation on when this can be merged?

kubawerlos:
@julienfalque can you split this PR to one adding new fixers and 2nd (targeted to 2.15) with refactoring of `DocBlock` classes?

julienfalque:
@kubawerlos Is #4591 what you want (this PR depends on it)? It currently targets 2.16 though.

kubawerlos:
@julienfalque yes, it is. Shame on me.

README.rst
```diff
@@ -1723,6 +1723,18 @@ Choose from the list of available rules:
   - ``scalar_types`` (``bool``): fix also scalar types; may have unexpected
     behaviour due to PHP bad type coercion system; defaults to ``true``
 
+* **phpdoc_to_property_type**
+
+  EXPERIMENTAL: Takes ``@var`` annotation of non-mixed types and adjusts
+  accordingly the property signature. Requires PHP >= 7.4.
+
+  *Risky rule: [1] This rule is EXPERIMENTAL and is not covered with backward compatibility promise. [2] ``@var`` annotation is mandatory for the fixer to make changes, signatures of properties without it (no docblock, inheritdocs) will not be fixed. [3] Manual actions are required if inherited signatures are not properly documented. [4] ``@inheritdocs`` support is under construction.*
```
Slamdunk, line 1731: We should also mention that nullable properties may be manually initialized if they are accessed before initialization.
The following code would indeed break with this fixer:
```php
class Foo
{
    /** @var null|Bar */
    private $bar;
    public function isBarGood(): bool
    {
        return null !== $this->bar;
    }
}
```
If you'd like to initialize nullable properties directly within this very fixer, I'd suggest to make it an option defaulted to `false`, because stricter code should always be favored.

julienfalque: I like your initialization option idea but I'd leave it to a subsequent PR so we can move forward with this one. I'll update the risky description :+1:

SpacePossum: I don't think we've to document the way PHP works but limit us to what the fixer does and the docs go in enough detail IMHO.
The suggestion for adding setting the null-value within this fixer is a +0 for me and could be added later if wanted, so for me out of scope of the PR. Note that when the property is initialized already in the constructor it should not be added to the property declaration for example.

Slamdunk: To both: regarding the option/new-fixer implementation, of course this is not the place, it's much better to land this as it is right now, and think about it later.

> I don't think we've to document the way PHP works but limit us to what the fixer does and the docs go in enough detail IMHO.

I disagree: prior 7.4 no one ever initialized nullable properties to null, and now all of a sudden this fixer breaks everything even though apparently should be harmless.

And what this fixer does is changing the behavior of the code.

```php
    /** @var null|Bar */
    private $bar;
```

Means the property may be null or Bar **and** is initialized to null.

```php
    private ?Bar $bar;
```

This instead solely means may be null or Bar, but uninitialized.

> Note that when the property is initialized already in the constructor it should not be added to the property declaration for example.

This is technically valid also outside constructor.

SpacePossum: >though apparently should be harmless

I think this where we disagree on, besides being marked as risky the whole description already yells twice that it is an experimental fixer, I don't understand how people could read it as being harmless.

Slamdunk: The risky description already list all the downside known, you either delete them all and leave just a risky label, or put them all, I don't understand why pinpointing only half of them :confused:

julienfalque: Description updated.

kubawerlos:
Sorry for the conflicts @julienfalque - can you solve them? I believe thee review would be quite straightforward here.

julienfalque:
Rebased #4591 and here, please review :)

GrahamCampbell:
I'd actually question if this belongs in php-cs-fixer. It's more a job for a refactoring tool like Rector, rather than a code style fixing tool.

Slamdunk:
> I'd actually question if this belongs in php-cs-fixer. It's more a job for a refactoring tool like Rector, rather than a code style fixing tool.

We already to a lot of refactoring in PHP-CS-Fixer. Moreover, PHP_CodeSniffer already does it: https://github.com/slevomat/coding-standard/blob/master/SlevomatCodingStandard/Sniffs/TypeHints/PropertyTypeHintSniff.php#L260-L296

GrahamCampbell:
I don't think we need to add more. It only adds more upgrading work every time a new version of PHP comes out.

Slamdunk:
I understand your point: it would be good for maintenance to have a fixer tool decoupled from PHP version, and maybe other one (like Rector) for the upgrade path.

But we are a bit late for such approach, aren't we?

SpacePossum:
@julienfalque please rebase and after that RTM :+1:

julienfalque:
@SpacePossum #4591 should be merged first.

composer.json
```diff
@@ -76,7 +76,6 @@
             "tests/Test/IntegrationCaseFactoryInterface.php",
             "tests/Test/InternalIntegrationCaseFactory.php",
             "tests/Test/IsIdenticalConstraint.php",
-            "tests/Test/TokensWithObservedTransformers.php",
             "tests/TestCase.php"
```
keradus, line 79:
```suggestion
            "tests/Test/TokensWithObservedTransformers.php"
            "tests/TestCase.php"
```

keradus:
Thank you @julienfalque.

julienfalque:
Thank you @keradus!

AJenbo:
thanks this will be useful

PR #4594, opened 2019-10-15
Braces - fix both single line comment styles

SpacePossum:


keradus:
Thank you @SpacePossum.

Issue #4582, opened 2019-10-11
Not removing unused function import


n3tw0rk-spy:
Not removing unused function import

#### The PHP version we are using (`$ php -v`):
PHP 7.3.6 (cli) (built: Jul  3 2019 20:59:35) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.3.6, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.7.2, Copyright (c) 2002-2019, by Derick Rethans

#### PHP CS Fixer version we are using (`$ php-cs-fixer -V`):
 PHP CS Fixer 2.15.3 Europe Round by Fabien Potencier and Dariusz Ruminski

#### The command we use to run PHP CS Fixer:
php-cs-fixer --no-interaction --diff -v fix 

#### The sample of configuration we are using:
```php
return PhpCsFixer\Config::create()
    ->setRiskyAllowed(true)
    ->setRules([
     // ...
     'no_unused_imports' => true,
     // ...
])
```

#### Minimum samples of PHP code (as plain text, not screenshots):
 * Before running PHP CS Fixer (no changes):
```php
<?php

declare(strict_types=1);

namespace App;

use App\UsedClass;
use App\UnusedClass;
use function League\Uri\create;

class Example 
{
    public function doesNotMatter()
    {
        return new UsedClass();
    }
}
```

 * Unexpected changes applied when running PHP CS Fixer:
```php
<?php

declare(strict_types=1);

namespace App;

use App\UsedClass;
use function League\Uri\create;

class Example 
{
    public function doesNotMatter()
    {
        return new UsedClass();
    }
}
```

 * The changes we expected instead:
```php
<?php

declare(strict_types=1);

namespace App;

use App\UsedClass;

class Example 
{
    public function doesNotMatter()
    {
        return new UsedClass();
    }
}
```

SpacePossum:
Hi and thanks for reporting!

There are two expectation that are not met for you, lets see what is up.

First, the removal of `use App\UsedClass;`, this is done because it is safe to do so because the import is not needed. The import is done within `namespace App;`  and the imported class itself is in the same namespace; `App` -> `App\UsedClass`.

Than the not removal of the used `use function League\Uri\create;` import. I did some testing on 2.15.3 and cannot recreate the problem. I tried to recreate the scenario as follows:

```
$ php php-cs-fixer fix test.php -vvv --dry-run --diff
[...]

PR #4595, opened 2019-10-15
Import cannot be used after `::` so can be removed

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4582

### DIFF
--- a/src/Fixer/Import/NoUnusedImportsFixer.php
+++ b/src/Fixer/Import/NoUnusedImportsFixer.php
@@ -90,7 +90,7 @@ final class NoUnusedImportsFixer extends AbstractFixer
         foreach ((new NamespacesAnalyzer())->getDeclarations($tokens) as $namespace) {
             $currentNamespaceUseDeclarations = array_filter(
                 $useDeclarations,
-                function (NamespaceUseAnalysis $useDeclaration) use ($namespace) {
+                static function (NamespaceUseAnalysis $useDeclaration) use ($namespace) {
                     return
                         $useDeclaration->getStartIndex() >= $namespace->getScopeStartIndex()
                         && $useDeclaration->getEndIndex() <= $namespace->getScopeEndIndex()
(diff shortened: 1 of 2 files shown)


src/Fixer/Import/NoUnusedImportsFixer.php
```diff
@@ -145,18 +145,20 @@ private function isImportUsed(Tokens $tokens, NamespaceAnalysis $namespace, arra
...
                 continue;
             }
 
-            if ($token->isComment() && Preg::match(
-                '/(?<![[:alnum:]])(?<!\\\\)'.$shortName.'(?![[:alnum:]])/i',
-                $token->getContent()
-            )) {
+            if ($token->isComment()
+                && Preg::match(
+                    '/(?<![[:alnum:]])(?<!\\\\)'.$shortName.'(?![[:alnum:]])/i',
```
roukmoute: Remove `/i`.

roukmoute:
I think you could add a new test:

```php
'in_same_namespace_3' => [
            <<<'EOF'
<?php

namespace Foo\Baz;

// My bar task.

EOF
            ,
            <<<'EOF'
<?php

namespace Foo\Baz;

use Foo\Bar;

// My bar task.

EOF
            ,
        ]
```

src/Fixer/Import/NoUnusedImportsFixer.php
keradus: the regex remain the same as it was. php imports are case insensitive.

GrahamCampbell: > php imports are case insensitive.

Not necessarily, actually. When it comes to autoloading, case matters if one has a case sensitive filesystem. So, while class fully qualified names are case insensitive, if the autoloader is later fired, case matters.

keradus:
Thank you @SpacePossum.

PR #4596, opened 2019-10-16
Ensure compatibility with PHP 7.4 numeric literal separator

julienfalque:
Related to #4382.


keradus:
Thank you @julienfalque.

PR #4597, opened 2019-10-16
Fix PHP 7.4 deprecation notices

julienfalque:
Related to #4382.


keradus:
Thank you @julienfalque.

PR #4583, opened 2019-10-11
PhpdocToParamTypeFixer - Introduction

SpacePossum:

PR #4605, opened 2019-10-21
PhpdocToParamTypeFixer - cover more cases

keradus:
ref #4583

@jg-development, @SpacePossum ,
can you pick it up, please?


julienfalque:
If you don't mind, I'd like to add those to #4591.

jg-development:
What about the string[]|int[] case?

julienfalque:
It should already be covered by #4591.

jg-development:
I am trying to look into that today

keradus:
Thank you @julienfalque.

PR #4606, opened 2019-10-21
Configure no_superfluous_phpdoc_tags for Symfony

keradus:
this PR is a placeholder.
codebase needs to be adjusted

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1109,7 +1109,7 @@ Choose from the list of available rules:
 
   Replaces superfluous ``elseif`` with ``if``.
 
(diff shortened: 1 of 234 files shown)


julienfalque:
I would override the configuration in the `@PhpCsFixer` rule set: `'allow_mixed' => false`.

keradus:
I decided to allow mixed so i can still grep for them `git grep "@param mixed"`
then, one day, we should stop using `mixed` at all.

SpacePossum:
:-1:   see also https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4566

PR #4608, opened 2019-10-21
Allow Symfony 5 components

l-vo:
Allow Symfony 5 components

cc @nicolas-grekas

### DIFF
--- a/.composer-require-checker.json
+++ b/.composer-require-checker.json
@@ -1,6 +1,8 @@
 {
     "symbol-whitelist" : [
+        "Symfony\\Contracts\\EventDispatcher\\Event",
         "Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface",
+        "Symfony\\Component\\EventDispatcher\\Event",
         "PhpCsFixer\\Tests\\Test\\Constraint\\SameStringsConstraint",
         "PhpCsFixer\\Tests\\Test\\IsIdenticalConstraint",
         "PHPUnit\\Framework\\Constraint\\IsIdentical",
--- a/.travis.yml
+++ b/.travis.yml
@@ -88,6 +88,7 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.2
+            env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="^5.0"
 
         -
             <<: *STANDARD_TEST_JOB
--- a/src/Console/ConfigurationResolver.php
+++ b/src/Console/ConfigurationResolver.php
@@ -302,11 +302,15 @@ final class ConfigurationResolver
     {
         if (null === $this->directory) {
             $path = $this->getCacheFile();
-            $filesystem = new Filesystem();
+            if (null === $path) {
+                $absolutePath = $this->cwd;
+            } else {
+                $filesystem = new Filesystem();
 
-            $absolutePath = $filesystem->isAbsolutePath($path)
-                ? $path
-                : $this->cwd.\DIRECTORY_SEPARATOR.$path;
+                $absolutePath = $filesystem->isAbsolutePath($path)
+                    ? $path
+                    : $this->cwd.\DIRECTORY_SEPARATOR.$path;
+            }
 
             $this->directory = new Directory(\dirname($absolutePath));
         }
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -1,0 +1,26 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Event;
+
+// Since PHP-CS-FIXER is PHP 5.6 compliant we can't always use Symfony Contracts (currently needs PHP ^7.1.3)
+// This conditionnal inheritance will be useless when PHP-CS-FIXER no longer supports PHP versions
+// inferior to Symfony/Contracts PHP minimal version
+if (class_exists(\Symfony\Contracts\EventDispatcher\Event::class)) {
+    class Event extends \Symfony\Contracts\EventDispatcher\Event
+    {
+    }
+} else {
+    class Event extends \Symfony\Component\EventDispatcher\Event
+    {
+    }
+}
--- a/tests/Console/ConfigurationResolverTest.php
+++ b/tests/Console/ConfigurationResolverTest.php
@@ -1142,6 +1142,47 @@ final class ConfigurationResolverTest extends TestCase
         ];
     }
 
+    public function provideGetDirectoryCases()
+    {
+        return [
+            [null, '/my/path/my/file', 'path/my/file'],
+            ['/my/path2/dir/.php_cs.cache', '/my/path2/dir/dir2/file', 'dir2/file'],
+            ['dir/.php_cs.cache', '/my/path/dir/dir3/file', 'dir3/file'],
+        ];
+    }
+
+    /**
+     * @dataProvider provideGetDirectoryCases
+     *
+     * @param null|string $cacheFile
+     * @param string      $file
+     * @param string      $expectedPathRelativeToFile
+     */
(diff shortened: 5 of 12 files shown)


src/Console/ConfigurationResolver.php
```diff
@@ -305,6 +305,10 @@ public function getDirectory()
     {
         if (null === $this->directory) {
             $path = $this->getCacheFile();
+            if (null === $path) {
+                return null;
```
keradus: this explicitly violate the method return type

src/Event/Event.php
```diff
@@ -0,0 +1,30 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Event;
+
+if (class_exists(\Symfony\Contracts\EventDispatcher\Event::class)) {
```
keradus: please add some comment why it's needed and when it would not be needed anymore (how to cleanup, when to cleanup)

src/Console/ConfigurationResolver.php
l-vo: Indeed... Modified, thanks :)

src/Event/Event.php
l-vo: Done :)

tests/Event/EventTest.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace PhpCsFixer\Tests\Event;
+
+use PhpCsFixer\Event\Event;
+use PhpCsFixer\Tests\TestCase;
+
+/**
+ * @internal
+ * @coversNothing
```
julienfalque: Should be:
```suggestion
 * @covers \PhpCsFixer\Event\Event
```

l-vo: Fixed, thank you :)

src/Event/Event.php
```diff
@@ -0,0 +1,33 @@
...
+} else {
+    class BaseEvent extends \Symfony\Component\EventDispatcher\Event
+    {
+    }
+}
+
+/**
+ * @author Laurent VOULLEMIER <laurent.voullemier@gmail.com>
+ */
+class Event extends BaseEvent
```
derrabus: The `BaseEvent` class looks unnecessary to me. You could directly define the `Event` class within the `if`/`else` block above.

```php
if (class_exists(\Symfony\Contracts\EventDispatcher\Event::class)) {
    class Event extends \Symfony\Contracts\EventDispatcher\Event
    {
    }
} else {
    class Event extends \Symfony\Component\EventDispatcher\Event
    {
    }
}
```

l-vo: Indeed... Thank you

tests/Event/EventTest.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+namespace PhpCsFixer\Tests\Event;
+
+use PhpCsFixer\Event\Event;
+use PhpCsFixer\Tests\TestCase;
+
+/**
+ * @internal
+ * @@covers \PhpCsFixer\Event\Event
```
keradus: `@@`

src/Console/ConfigurationResolver.php
```diff
@@ -305,11 +305,15 @@ public function getDirectory()
     {
         if (null === $this->directory) {
             $path = $this->getCacheFile();
-            $filesystem = new Filesystem();
+            if (null === $path) {
+                $absolutePath = $this->cwd;
```
keradus: can you elaborate on this change?
it does not look related to `Event` at first glance

keradus:
@l-vo , please ping us when 2nd point of https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4608#pullrequestreview-307590178 is ready

src/Console/ConfigurationResolver.php
derrabus: @keradus `Filesystem::isAbsolutePath()` does not accept `null` anymore, so that case is handled separately.

keradus: ah. thanks for clarification !

tests/Console/ConfigurationResolverTest.php
```diff
@@ -1143,6 +1143,47 @@ public function provideDeprecatedFixerConfiguredCases()
         ];
     }
 
+    public function provideGetDirectoryCases()
+    {
+        return [
+            [null, '/my/path/my/file', 'path/my/file'],
```
julienfalque: Shouldn't the expected relative path be `my/file` instead of `path/my/file`?

l-vo: Maybe in the case where there is no cache file,`getDirectory` doesn't behave as expected. But it's the current behavior. Fixing this potential issue is not the goal of this PR and I'm not sur this wouldn't be considered as a BC break.

l-vo:
@keradus CI steps to test against Symfony 4 and Symfony 5 components added

.travis.yml
```diff
@@ -78,6 +79,11 @@ jobs:
             php: 5.6
             env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
 
+        -   <<: *STANDARD_TEST_JOB
+            stage: Test
+            php: 7.1
+            env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="~3.0"
```
keradus: why `SYMFONY_DEPRECATIONS_HELPER=disabled` ?
why `PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1` ?

keradus:
@l-vo , can we reuse some of existing jobs?
i would really like to avoid extra 2 jobs

.travis.yml
l-vo: Job not used anymore, using `COMPOSER_FLAGS="--prefer-stable --prefer-lowest"` instead which already installs Symfony 3 components

l-vo:
@keradus PR modified to reuse existing jobs

alexander-schranz:
@julienfalque another way would be use the `.phar` for php-coveralls as it is just a dev dependency. Example: https://github.com/sulu/SuluContentBundle/blob/756944dbb17a27ed840fc12d0d13203ccb4e7aed/.circleci/config.yml#L135-L138

andreybolonin:
Screen Shot 2019-11-21 at 1 09 53 PM

l-vo:
@julienfalque rebased on 2.15

keradus:
@julienfalque , php-coveralls issue solved. can you review once more?
@l-vo , why changing the baseline to 2.15? if sth would not be compatible, we would need to fix it then on both, 2.15 and 2.16, instead of only 2.16

julienfalque:
@keradus [I did ask to rebase onto 2.15](4608#pullrequestreview-318660553). Travis build was green on 2.16 so merging 2.15 into 2.16 should be fine anyway.

.travis.yml
```diff
@@ -88,6 +89,7 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.2
+            env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="5.0.x-dev" MINIMUM_STABILITY=dev
```
julienfalque: Symfony 5.0 stable is available now, we can use it rather than the `-dev` version.

l-vo: Indeed... Fixed :)

julienfalque: Maybe `^5.0` instead of `~5.0.0`? Also the line with `$MINIMUM_STABILITY` isn't necessary anymore.

fabpot:
:+1: Looks great! That's the only missing piece to allow me to upgrade fabbot.io to Symfony 5.0 :)

TomasVotruba:
Bloker for https://github.com/getsentry/sentry-php/pull/925

julienfalque:
Thank you @l-vo.

.composer-require-checker.json
```diff
@@ -1,6 +1,8 @@
 {
     "symbol-whitelist" : [
+        "Symfony\\Contracts\\EventDispatcher\\Event",
         "Symfony\\Contracts\\EventDispatcher\\EventDispatcherInterface",
+        "Symfony\\Component\\EventDispatcher\\Event",
```
keradus: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4665

PR #4615, opened 2019-10-28
Remove workaround for dev-tools install reg. Phive

ktomk:
Removal of the temporary work-around to make Phive work on Travis. 2dadad5 (https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/2dadad5eb75eec98c8ccca4bb6f2668afae1038b#diff-3948854d59731218a3ce36138a86bcc3R24-R32).

Phive has been updated (0.13.2) to work in the Travis build of PHP-CS-Fixer.

NOTE: This was previously WIP, after updating / running w/ the fixed version problems to install Phive in the build can be seen for key download as keyservers are often down. This makes the build to fail early.

Phive-Issue: phar-io/phive#208 <https://github.com/phar-io/phive/issues/208>
Related-To: #4550 
Ref: 2dadad5eb75eec98c8ccca4bb6f2668afae1038b

### DIFF
--- a/dev-tools/install.sh
+++ b/dev-tools/install.sh
@@ -21,15 +21,6 @@ cd "$(dirname "$0")"
 
 mkdir -p bin
 
-###
-# temporary work-around: phive expects gpg2 as gpg, build box has gpg1 as gpg
-#   phive/phive:208 <https://github.com/phar-io/phive/issues/208>
-if [ "${TRAVIS-}" = true ]; then
-  sudo apt-get install dirmngr --install-recommends
-  [ ! -f "/usr/bin/gpg1"  ] && sudo mv -- "/usr/bin/gpg" "/usr/bin/gpg1"
-  sudo ln -sfT "/usr/bin/gpg2" "/usr/bin/gpg"
-fi
-
 VERSION_CB="2.0.0.2"
 VERSION_SC="stable"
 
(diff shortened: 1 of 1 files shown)


ktomk:
Just in case anyone wonders: This is not to be merged, just gathering data as upstream progresses. Build failures are expected.

ktomk:
@julienfalque Phive has been patched upstream, I cleaned up the PR here after I can confirm the fix working. CR pls.

For reference, PR which introduced the workaround was #4550 .

ktomk:
Just for the record as I experienced the same already yesterday when refreshing: When the keyserver times out to import the key to verify the phar package, with the current travis_retry  of three times it looks likely to fail three times in  a row. When I tried later on, all was good.

Maybe trying with more keyservers at the point of failure heps ([ref](https://github.com/tianon/gosu/issues/39)). Phive also internally does rotate keyservers ([ref](https://github.com/phar-io/phive/commit/fda19494898977600a1d0045af7c9bc683d21606#diff-5e5b2bcab99b42b28d693831aa29d429R3)).

ktomk:
@julienfalque and others: I think we should address point break of the whole Travis build as when the task w/ the dev-tools fails, the whole Travis build is marked a failure.

When it turns out that this breaks often in practice, it's a bummer as with current configuration. Travis will skip many other build steps and next to having the build red, it's also not clear for the state of the "rest" of the build:

* The step that easily breaks runs first, which is also the reason to push for the workaround in the first place
* On that step failure, other steps are stopped / not even executed (grey) and the build gathers no more data

keradus:
Thank you @ktomk.

PR #4626, opened 2019-11-04
FinalPublicMethodForAbstractClassFixer - Do not attempt to mark abstract public methods as final

localheinz:
This PR

* [x] asserts that the `FinalPublicMethodForAbstractClassFixer` does not mark an `abstract public`  method as `final`
* [x] adjusts the `FinalPublicMethodForAbstractClassFixer` to ignore `abstract public` methods

Follows #3928.


Slamdunk:
Ups, I'm sorry, nice catch :+1:

julienfalque:
Thank you @localheinz.

localheinz:
Thank you, @GrahamCampbell, @julienfalque, @kubawerlos, and @Slamdunk!

PR #4630, opened 2019-11-08
FullyQualifiedStrictTypesFixer - Ignore partial class names which look like FQCNs

localheinz:
This PR

* [x] ignores types when they are not references relative to the root namespace

### DIFF
--- a/src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
+++ b/src/Fixer/Import/FullyQualifiedStrictTypesFixer.php
@@ -152,6 +152,11 @@ class SomeClass
         }
 
         $typeName = $type->getName();
+
+        if (0 !== strpos($typeName, '\\')) {
+            return;
+        }
+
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Is this ready to be merged?

SpacePossum:
This looks like a solid fix to me :+1: 

Just wondering, maybe we could apply it to the `TypeShortNameResolver::resolve` itself?
So `->resolve($tokens, "\Foo")` just returns `"\Foo"` ?

localheinz:
@SpacePossum 

Can you explain what you mean?

SpacePossum:
checking the code again Im not sure what I was thinking :sweat_smile:

SpacePossum:
Thank you @localheinz.

localheinz:
Thank you, @GrahamCampbell, @julienfalque, and @SpacePossum!

PR #4632, opened 2019-11-08
NullableTypeDeclarationForDefaultNullValueFixer - fix for not lowercase "null"

kubawerlos:
Pinging @HypeMC for review as the original author.


julienfalque:
Thank you @kubawerlos.

PR #4635, opened 2019-11-09
Run PHPUnit for PHP 5.6 with ParaTest

kubawerlos:


kubawerlos:
No go for now as the bug failing Travis has fix only for PHP 7.

PR #4636, opened 2019-11-10
DX: do not check for PHPDBG when collecting coverage

kubawerlos:


.travis.yml
```diff
@@ -100,16 +99,10 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.3
-            env: COLLECT_COVERAGE=1
+            name: Collect coverage
```
keradus: can we name it like `7.3 | Collect coverage` ?
and then, previous one like `7.3 | With migration rules`

kubawerlos: @keradus done

keradus:
Thank you @kubawerlos.

PR #4639, opened 2019-11-11
DX: move all dev dependencies to dev-tools

kubawerlos:
This PR extracts `dev-require` and `dev-autoload` from `composer.json`, but also:

- cleans up AppVeyor, CircleCI and TravisCI configurations
- update TravisCI fast test to run PHP 7.2 instead of PHP 7.0
- updates `InstallViaComposerTest`, so it makes the copy for testing, not symlink

Should those 3 be extracted to separate PRs?

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,6 +35,7 @@ install:
 
 before_test:
     - cd C:\projects\php-cs-fixer
+    - php C:\tools\composer.phar remove --dev --no-update --working-dir=./dev-tools humbug/box localheinz/composer-normalize phpstan/phpstan-phpunit
     - php C:\tools\composer.phar update --no-ansi --no-interaction --no-progress --optimize-autoloader --prefer-stable --working-dir=dev-tools
     - php C:\tools\composer.phar info --direct --working-dir=dev-tools | sort
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -35,11 +35,11 @@ install:
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --optimize-autoloader --no-interaction --no-progress --prefer-stable --no-ansi
-    - php C:\tools\composer.phar info -D | sort
+    - php C:\tools\composer.phar update --no-ansi --no-interaction --no-progress --optimize-autoloader --prefer-stable --working-dir=dev-tools
+    - php C:\tools\composer.phar info --direct --working-dir=dev-tools | sort
 
 test_script:
     - cd C:\projects\php-cs-fixer
-    - vendor\bin\phpunit
+    - dev-tools\vendor\bin\phpunit
     - set PHP_CS_FIXER_FUTURE_MODE=1
     - php php-cs-fixer --diff --dry-run -v fix
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ env:
         - COMPOSER_FLAGS=""
 
 before_install:
-    # turn off XDebug
+    # turn off Xdebug
     - phpenv config-rm xdebug.ini || return 0
 
     # Composer: boost installation
@@ -78,7 +78,7 @@ jobs:
             php: '5.6'
             env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
             before_install:
-                # turn off XDebug
+                # turn off Xdebug
                 - phpenv config-rm xdebug.ini || return 0
 
                 # Composer: boost installation
@@ -93,7 +93,7 @@ jobs:
             php: '7.0'
             env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="~4.1.0"
             before_install:
-                # turn off XDebug
+                # turn off Xdebug
                 - phpenv config-rm xdebug.ini || return 0
 
                 # Composer: boost installation
@@ -107,7 +107,7 @@ jobs:
             stage: Test
             php: '7.1'
             before_install:
-                # turn off XDebug
+                # turn off Xdebug
                 - phpenv config-rm xdebug.ini || return 0
 
                 # Composer: boost installation
@@ -139,7 +139,7 @@ jobs:
                 - if [ $COLLECT_COVERAGE == 0 ]; then travis_terminate 0; fi
 
                 ## regular `before_install`
-                # turn off XDebug
+                # turn off Xdebug
                 - phpenv config-rm xdebug.ini || return 0
 
                 # Composer: boost installation
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,15 @@ jobs:
             stage: Test
             php: 5.6
             env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+            before_install:
+                # turn off XDebug
+                - phpenv config-rm xdebug.ini || return 0
+
+                # Composer: boost installation
+                - composer global show hirak/prestissimo --quiet || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
+
+                # Require PHPUnit 8
+                - composer remove --dev --no-update --working-dir=./dev-tools humbug/box phpstan/phpstan-phpunit
 
         -
             <<: *STANDARD_TEST_JOB
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ jobs:
 
         - &STANDARD_TEST_JOB
             stage: Fast Test
-            php: 7.3
+            php: 7.2
             install:
                 # Composer: enforce given Symfony components version
                 - if [ "$SYMFONY_VERSION" != "" ]; then composer global show symfony/flex -q || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS symfony/flex; fi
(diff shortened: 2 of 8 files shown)


.travis.yml
```diff
@@ -60,46 +60,73 @@ jobs:
...
                 - PHP_CS_FIXER_FUTURE_MODE=1 php php-cs-fixer --diff --dry-run -v fix
 
         -
             <<: *STANDARD_TEST_JOB
             stage: Test
-            php: 5.6
+            php: '5.6'
             env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
+            before_install:
+                # turn off XDebug
```
localheinz: See https://twitter.com/derickr/status/1184465999971700736.

😉

kubawerlos: @localheinz fixed, now you can safely review the rest of PR ;)

.appveyor.yml
```diff
@@ -35,11 +35,12 @@ install:
 
 before_test:
     - cd C:\projects\php-cs-fixer
-    - php C:\tools\composer.phar update --optimize-autoloader --no-interaction --no-progress --prefer-stable --no-ansi
-    - php C:\tools\composer.phar info -D | sort
+    - php C:\tools\composer.phar remove --dev --no-update --working-dir=./dev-tools humbug/box localheinz/composer-normalize phpstan/phpstan-phpunit
```
keradus: why `remove` ?

.travis.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     include:
         -
             stage: Static Code Analysis
-            php: 7.3
+            php: '7.3'
```
keradus: please revert

php-cs-fixer
```diff
@@ -56,7 +56,7 @@ if (class_exists('Phar')) {
 
 if ($require) {
     // OK, it's not, let give Composer autoloader a try!
-    $possibleFiles = [__DIR__.'/../../autoload.php', __DIR__.'/../autoload.php', __DIR__.'/vendor/autoload.php'];
+    $possibleFiles = [__DIR__.'/../../autoload.php', __DIR__.'/../autoload.php', __DIR__.'/vendor/autoload.php', __DIR__.'/dev-tools/vendor/autoload.php'];
```
keradus: it is a prod file. it must not be aware about `dev-tools`

.appveyor.yml
kubawerlos: One job runs on PHP 5.6 and these packages need 7+.

.travis.yml
kubawerlos: why? PHP version is not a float value, but a string. PHP 7.0 is not `7.0`, but `7` on the [list](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/builds/613530218).

composer.json
```diff
@@ -31,20 +31,6 @@
...
-    "require-dev": {
-        "johnkary/phpunit-speedtrap": "^1.1 || ^2.0 || ^3.0",
-        "justinrainbow/json-schema": "^5.0",
-        "keradus/cli-executor": "^1.2",
-        "mikey179/vfsstream": "^1.6",
-        "php-coveralls/php-coveralls": "^2.1",
-        "php-cs-fixer/accessible-object": "^1.0",
-        "php-cs-fixer/phpunit-constraint-isidenticalstring": "^1.1",
-        "php-cs-fixer/phpunit-constraint-xmlmatchesxsd": "^1.1",
-        "phpunit/phpunit": "^5.7.27 || ^6.5.14 || ^7.1",
```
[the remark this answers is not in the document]
kubawerlos: why "must"? what is the problem if they are not?

php-cs-fixer
kubawerlos: I agree that is the perfect solution.

Yet I can name [one file](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.0/src/Console/Command/HelpCommand.php#L287) that is aware of `dev-tools`.

Also, what harm can it make? Only if previous 3 paths are not matching that last one is even considered - and I don't believe would match for anyone.

In fact, only one test class was failing without this change - I can give it another try if you insist.

.travis.yml
keradus: i don't see anyhow this argument.
even if string, then `php: 7.4snapshot` and not `php: '7.4snapshot'`.
same, the `stage` is a string, but we don't use quotes. they are not needed

composer.json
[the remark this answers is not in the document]
keradus: that is a purpose of `require-dev`.
we are extending classes from phpunit package in our codebase - thus it must be there in composer.json file. if we use the class in src - then package should be in `dependencies`. if not, then in `devdependencies`.

removing dependencies directly used in code (dev or src) from composer file i consider a very bad practice. it is not the external tool that we are using out of code, like phpmd.

.travis.yml
```diff
@@ -60,46 +60,73 @@ jobs:
...
             <<: *STANDARD_TEST_JOB
             stage: Test
-            php: 7.2
+            php: '7.1'
+            before_install:
+                # turn off Xdebug
+                - phpenv config-rm xdebug.ini || return 0
+
+                # Composer: boost installation
+                - composer global show hirak/prestissimo --quiet || travis_retry composer global require $DEFAULT_COMPOSER_FLAGS hirak/prestissimo
```
keradus: DRY is violated

composer.json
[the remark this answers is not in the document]
keradus: maybe you should just suggest to Composer to drop the `require-dev` section at all?

php-cs-fixer
keradus: Point existing code smell is not the best excuse to introduce another...

i consider referencing to dev file available in dev env only to be a code smell

tests/Smoke/InstallViaComposerTest.php
```diff
@@ -75,6 +75,9 @@ public function testInstallationViaPathIsPossible()
                 [
                     'type' => 'path',
                     'url' => __DIR__.'/../..',
+                    'options' => [
+                        'symlink' => false,
```
keradus: any reasoning ?

.appveyor.yml
keradus: yet, on stable branch, there is no issue like that in AppVeyor
https://ci.appveyor.com/project/keradus/php-cs-fixer/builds/29070411

kubawerlos: Because those packages are already in dev-tools.

.travis.yml
kubawerlos: `stage` is a string because it is a string.

I was always thinking we want be explicit over implicit - I guess not in this case.

composer.json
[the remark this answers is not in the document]
kubawerlos: Purpose of `require-dev` is to lists packages required for developing this package - so following that reasoning all dependencies from `dev-tools` should be moved to main `composer.json`.

> if we use the class in src - then package should be in `dependencies`.

Is that rule about the src directory? ;) I believe (no-dev) dependencies are the ones needed in production - which in case of library is release - such as PHPUnit - because [some files](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.0/composer.json#L57) relying on it are in the release. Yet, no one ever consider moving it from `require-dev` to `require`.

> removing dependencies directly used in code (dev or src) from composer file i consider a very bad practice

I don't understand why? I mean the part for `dev`. For `src` it's obvious, but for `dev`? Even if PHP CS Fixer would be install via composer locally in project it's dev dependencies would be ignored.

TL;DR the idea was to move out dev dependencies from main `composer.json` to free normal dependencies from any potential conflict, nothing more.

php-cs-fixer
kubawerlos: > i consider referencing in prod file to dev file available in dev env only to be a code smell

me too, but for me in this case it's a valid exception to do that

.travis.yml
kubawerlos: I didn't know there is a rule to not repeat the code more then twice ;)

Do you know how can we fix this and currently existing duplication as well?

tests/Smoke/InstallViaComposerTest.php
kubawerlos: This test would fail - without it packages was installed via symlink and then calling `php-cs-fixer` would require `autoload.php` from project's root `vendor`, not the one created for the test.

php-cs-fixer
kubawerlos: on the second thought - and after cleaning the other smell - this needs fixing

.travis.yml
keradus: i see no reasoning why `stage` should not have quotes and `php` should
we have, on master line:
```
stage: Static Code Analysis
php: 7.4snapshot
```

both are strings. none is using quotes

composer.json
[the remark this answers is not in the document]
keradus: we don't care about conflicting dependencies of standalone tools, like phpmd. so we extracted it. cool.
we do care about conflicting dependencies of dev libraries we are using in our own codebase, eg phpunit, so let's keep them under main composer.json.

.travis.yml
keradus: probably each duplication to discuss separately. i'm talking about multiple places in travis file when you added same step over and over ago. not needed, as proven on master branch already.

tests/Smoke/InstallViaComposerTest.php
keradus: why it's not failing on master branch when symlink is not there?

kubawerlos: it uses original code, so all is good - with these changes there is no `vendor/autoload.php` as all was moved to `dev-tools`

PR #4643, opened 2019-11-13
DX: Update .gitattributes and move ci-integration.sh to root of the project

kubawerlos:
This PR:
 - adds leading `/` for paths in `.gitattributes` to indicate better they starts from project's root
 - excludes all hidden paths (row `/.*`) from distribution by assumption hidden path is the one starting with dot
 - fixes `box.json` to `box.json.dist` in `.gitattributes`
 - excludes `phpstan.neon` from distribution
 - moves `ci-integration.sh` from `dev-tool` to root
 - excludes `dev-tools` from distribution


.gitattributes
```diff
@@ -8,24 +8,18 @@ tests/Report/AbstractReporterTestCase.php export-ignore
 tests/Test/AbstractFixerWithAliasedOptionsTestCase.php export-ignore
 tests/Test/AbstractTransformerTestCase.php export-ignore
```
julienfalque: Could you add `tests/Test/IsIdenticalConstraint.php` and revert d481be5?

dev-tools/.gitattributes
```diff
@@ -0,0 +1,10 @@
+.* export-ignore
```
julienfalque: Why not ignoring the whole directory in the root `.gitattributes` file?

.gitattributes
kubawerlos: Done

dev-tools/.gitattributes
kubawerlos: Because we need `dev-tools/ci-integration.sh` in distribution and we have test for it, the one [failing](https://travis-ci.org/FriendsOfPHP/PHP-CS-Fixer/jobs/611568199) on first commit.

Also we cannot make this:
```
/dev-tools/
!/dev-tools/ci-integration.sh
```
because it's not allowed:
```
warning: Negative patterns are ignored in git attributes
Use '\!' for literal leading exclamation.
```

.gitattributes
julienfalque: Actually the trait is used in classes that are *not* excluded, so my comment was wrong, sorry :/

kubawerlos: Reverted :)

dev-tools/.gitattributes
julienfalque: This looks a bit overkill to me. What about moving the file somewhere else?

kubawerlos: To project's root?

julienfalque: Yes I guess. It seems required by the `--help` CLI option, it's not really a dev tool.

keradus:
Thank you @kubawerlos.

glensc:
Is ci-integration.sh something to be used by projects or it's strictly internal to php-cs-fixer itself? i could not find anything in the docs.

it's included in distribution, i.e not present in .gitattibutes:
- https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v3.0.0/.gitattributes

and the paths it contains, make it suitable to be used from projects from vendor dir.

PR #4644, opened 2019-11-14
Docs: add info about "-vv..."

voku:
This change is [Reviewable](https://reviewable.io/reviews/friendsofphp/php-cs-fixer/4644)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -166,8 +166,16 @@ NOTE: the output for the following formats are generated in accordance with XML 
 * ``junit`` follows the `JUnit xml schema from Jenkins </doc/junit-10.xsd>`_
 * ``checkstyle`` follows the common `"checkstyle" xml schema </doc/checkstyle.xsd>`_
 
+The ``--quiet`` Do not output any message.
 
 The ``--verbose`` option will show the applied rules. When using the ``txt`` format it will also display progress notifications.
+
+NOTE: if there is an error like "errors reported during linting after fixing", you can use this to be even more verbose for debugging purpose
+
+* ``--verbose=0`` or no option: normal
+* ``--verbose``, ``--verbose=1``, ``-v``: verbose
+* ``--verbose=2``, ``-vv``: very verbose
+* ``--verbose=3``, ``-vvv``: debug
 
 The ``--rules`` option limits the rules to apply to the
 project:
--- a/src/Console/Command/HelpCommand.php
+++ b/src/Console/Command/HelpCommand.php
@@ -71,8 +71,16 @@ NOTE: the output for the following formats are generated in accordance with XML 
 * ``junit`` follows the `JUnit xml schema from Jenkins </doc/junit-10.xsd>`_
 * ``checkstyle`` follows the common `"checkstyle" xml schema </doc/checkstyle.xsd>`_
 
+The <comment>--quiet</comment> Do not output any message.
 
 The <comment>--verbose</comment> option will show the applied rules. When using the ``txt`` format it will also display progress notifications.
+
+NOTE: if there is an error like "errors reported during linting after fixing", you can use this to be even more verbose for debugging purpose
+
+* ``--verbose=0`` or no option: normal
+* ``--verbose``, ``--verbose=1``, ``-v``: verbose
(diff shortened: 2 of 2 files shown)


README.rst
```diff
@@ -169,6 +169,8 @@ NOTE: the output for the following formats are generated in accordance with XML
 
 The ``--verbose`` option will show the applied rules. When using the ``txt`` format it will also display progress notifications.
 
+NOTE: if there is a error like "errors reported during linting after fixing", you can use ```-vvvv``` to be even more verbose for debugging purpose
```
julienfalque: I don't think `-vvvv` exists. Available options regarding verbosity are:
* `--quiet`, `-q`: quiet
* `--verbose=0` or no option: normal
* `--verbose`, `--verbose=1`, `-v`: verbose
* `--verbose=2`, `-vv`: very verbose
* `--verbose=3`, `-vvv`: debug

julienfalque: In reStructuredText ` ```...``` ` should be ` ``...`` ` (2×2 backquotes, not 2×3).

voku: added

voku: :+1:

README.rst
```diff
@@ -166,9 +166,17 @@ NOTE: the output for the following formats are generated in accordance with XML
 * ``junit`` follows the `JUnit xml schema from Jenkins </doc/junit-10.xsd>`_
 * ``checkstyle`` follows the common `"checkstyle" xml schema </doc/checkstyle.xsd>`_
 
+The ``--quiet`` Do not output any message. 
 
 The ``--verbose`` option will show the applied rules. When using the ``txt`` format it will also display progress notifications.
 
+NOTE: if there is a error like "errors reported during linting after fixing", you can use this to be even more verbose for debugging purpose
```
kubawerlos:
```suggestion
NOTE: if there is an error like "errors reported during linting after fixing", you can use this to be even more verbose for debugging purpose
```

README.rst
```diff
@@ -166,9 +166,17 @@ NOTE: the output for the following formats are generated in accordance with XML
...
 * ``checkstyle`` follows the common `"checkstyle" xml schema </doc/checkstyle.xsd>`_
 
+The ``--quiet`` Do not output any message. 
 
 The ``--verbose`` option will show the applied rules. When using the ``txt`` format it will also display progress notifications.
 
+NOTE: if there is an error like "errors reported during linting after fixing", you can use this to be even more verbose for debugging purpose
+
+* ``--verbose=0`` or no option: normal
+* ``--verbose``, --verbose=1, -v: verbose
```
kubawerlos: Don't we want `--verbose=1` and `-v` in backticks too?

voku: yep, updated.

kubawerlos:
Content is looking good, only checks to fix.

julienfalque:
I'm wondering if the current changes are really helpful. I mean those options are standard Symfony options and quite common amongst CLI tools in general. Also one can already find this information with `php-cs-fixer fix --help`.

Maybe something more helpful would be to actually add details about what each verbosity level shows/hides?

voku:
Just because these are default options, does not mean that the docs should not contain it. 

And yes, some extra information are maybe helpful. I think this is the relevant part?: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/src/Console/Output/ErrorOutput.php#L55

keradus:
please, check the contributing guidelines.
as you can see, all CIs are failing.

voku:
@keradus CIs fixed. I did not change any source code, so I thought that the failing tests are not related to these changes.

SpacePossum:
:+1: 
maybe best to target lowest stable so we've less differences over the branches?

julienfalque:
I still think it would be nice to not only list the different options and their verbosity level but also say what each level of verbosity will output, i.e. *if verbosity is at least [level], it will show [info]*.

keradus:
I agree with you, @julienfalque 

unfortunately, nobody is working on it and I believe it's better to merge small improvement than not merge any improvement (while awaiting for big improvement to be proposed).

keradus:
Thank you @voku.

keradus:
merged into LTS

Issue #4329, opened 2019-03-04
Check PHP extensions on runtime


keradus:
Same as we are checking PHP version on runtime, we shall check the extensions.
If the app is running from phar distribution, nothing had checked the extensions

keradus:
(idea raised by @waghanza on gitter)

PR #4645, opened 2019-11-14
Check PHP extensions on runtime

kubawerlos:
Resolves https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4329

### DIFF
--- a/php-cs-fixer
+++ b/php-cs-fixer
@@ -37,6 +37,19 @@ if (defined('HHVM_VERSION_ID')) {
     }
 }
 
+foreach (['json', 'tokenizer'] as $extension) {
+    if (!extension_loaded($extension)) {
+        fwrite(STDERR, sprintf("PHP extension ext-%s is missing from your system. Install or enable it.\n", $extension));
+
+        if (getenv('PHP_CS_FIXER_IGNORE_ENV')) {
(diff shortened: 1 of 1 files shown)


php-cs-fixer
```diff
@@ -37,6 +37,19 @@ if (defined('HHVM_VERSION_ID')) {
...
+        fwrite(STDERR, sprintf("PHP extension ext-%s is missing from your system. Install or enable it.\n", $extension));
+
+        if (getenv('PHP_CS_FIXER_IGNORE_ENV')) {
+            fwrite(STDERR, "Ignoring environment requirements because `PHP_CS_FIXER_IGNORE_ENV` is set. Execution may be unstable.\n");
+        } else {
+            exit(1);
+        }
+    }
+}
+unset($extension);
```
Slamdunk: You can wrap everything inside a volatile lambda to be sure context gets cleaned at the end:
```php
(static function(){
    $var = 1;
    echo $var;
})();
// $var is undefined now, don't have to remember the unset() call
```

julienfalque: I like the idea, though I'd rather do it in a separate PR to have atomic commits as there are other `unset()`s in other places.

kubawerlos: PHP 5.6 https://3v4l.org/rvgHm :(

julienfalque: Definitely later then :)

Though it might be time to drop support for old PHP versions.

keradus:
Thank you @kubawerlos.

Issue #4446, opened 2019-06-12
only psalm annotations not recognised as docblock and a * is striped


basz:
Hi,

Whenever I have a docblock annotation with only a `@psalm-*` annotation in it, cs fixer does not recognize it as a docblock and therefore strips the last `*` of the opening `/**`.

Might be just a configuration option, or a bug...

When reporting an issue (bug) please provide the following information:

#### The PHP version you are using (`$ php -v`):
PHP 7.2.19 (cli) (built: Jun  1 2019 00:03:09) ( NTS )
Copyright (c) 1997-2018 The PHP Group
Zend Engine v3.2.0, Copyright (c) 1998-2018 Zend Technologies
    with Xdebug v2.7.2, Copyright (c) 2002-2019, by Derick Rethans
    with Zend OPcache v7.2.19, Copyright (c) 1999-2018, by Zend Technologies

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
PHP CS Fixer 2.15.1 Europe Round by Fabien Potencier and Dariusz Ruminski

#### The command you use to run PHP CS Fixer:
php-cs-fixer fix -v --diff

#### The configuration file you are using, if any:
```php
private $defaults = [
        '@PHP71Migration' => true,
        '@PSR2' => true,
        '@Symfony' => true,
        'array_syntax' => ['syntax' => 'short'],
        'binary_operator_spaces' => [
            'align_double_arrow' => false,
            'align_equals' => false,
        ],
        'blank_line_after_namespace' => true,
        'blank_line_after_opening_tag' => true,
        'blank_line_before_return' => true,
        'braces' => true,
        'cast_spaces' => true,
        'class_definition' => true,
        'combine_consecutive_unsets' => true,
        'concat_space' => false,
        'declare_strict_types' => true,
        'elseif' => true,
        'encoding' => true,
        'full_opening_tag' => true,
        'function_declaration' => true,
        'function_typehint_space' => true,
        'hash_to_slash_comment' => true,
        'header_comment' => [
            'commentType' => 'PHPDoc',
            'header' => 'PLHW was here at `%package%` in `%year%`! Please create a .docheader in the project root and run `composer cs-fix`',
            'location' => 'after_open',
            'separate' => 'both',
        ],
        'include' => true,
        'indentation_type' => true,
        'line_ending' => true,
        'linebreak_after_opening_tag' => true,
        'lowercase_constants' => true,
[...]

PR #4649, opened 2019-11-18
PhpdocToCommentFixer - Ignore annotations

localheinz:
This PR

* [x] adjusts `PhpDocToComment` fixer to ignore annotations

💁‍♂ For reference, see https://github.com/vimeo/psalm/issues/2344.

/cc @theseer

closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4446

### DIFF
--- a/src/Fixer/Phpdoc/PhpdocToCommentFixer.php
+++ b/src/Fixer/Phpdoc/PhpdocToCommentFixer.php
@@ -88,6 +88,10 @@ foreach($connections as $key => $sqlite) {
                 continue;
             }
 
+            if ($commentsAnalyzer->isAnnotation($token)) {
+                continue;
+            }
+
             $tokens[$index] = new Token([T_COMMENT, '/*'.ltrim($token->getContent(), '/*')]);
         }
     }
--- a/src/Tokenizer/Analyzer/CommentsAnalyzer.php
+++ b/src/Tokenizer/Analyzer/CommentsAnalyzer.php
@@ -92,6 +92,11 @@ final class CommentsAnalyzer
         return false;
     }
 
+    public function isAnnotation(Token $token)
+    {
+        return 1 === Preg::match('#\/\*{2}\s+.*@\w+#m', $token->getContent());
+    }
+
     /**
      * Return array of indices that are part of a comment started at given index.
      *
--- a/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
+++ b/tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
@@ -302,4 +302,121 @@ $bar;',
             ['<?php /* Before anonymous function */ $fn = fn($x) => $x + 1;'],
         ];
     }
+
+    /**
+     * @dataProvider provideAnnotationCases
+     *
+     * @param string $code
+     */
+    public function testIsAnnotation($code)
+    {
+        $tokens = Tokens::fromCode($code);
+
+        $index = $tokens->getNextTokenOfKind(0, [[T_COMMENT], [T_DOC_COMMENT]]);
+
+        $token = $tokens[$index];
+
+        $analyzer = new CommentsAnalyzer();
+
+        static::assertTrue($analyzer->isAnnotation($token));
+    }
+
+    public function provideAnnotationCases()
+    {
+        return [
+            [
+                '<?php
+
+/**
+ * @var int $foo
+ */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/** @var int $foo */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/**
+ * @noinspection PhpParamsInspection
+ */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/** @noinspection PhpParamsInspection */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/** @psalm-suppress hmmm */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/**
+ * @psalm-suppress hmmm
+ */
+$foo = 900;
+',
+            ],
+        ];
+    }
+
+    /**
+     * @dataProvider provideNotAnnotationCases
+     *
+     * @param string $code
+     */
+    public function testIsNotAnnotation($code)
+    {
+        $tokens = Tokens::fromCode($code);
+
+        $index = $tokens->getNextTokenOfKind(0, [[T_COMMENT], [T_DOC_COMMENT]]);
+
+        $token = $tokens[$index];
+
+        $analyzer = new CommentsAnalyzer();
+
+        static::assertFalse($analyzer->isAnnotation($token));
+    }
+
+    public function provideNotAnnotationCases()
+    {
+        return [
+            [
+                '<?php
+
+/* @var int $foo */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/* @noinspection PhpParamsInspection */
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/* @psalm-suppress hmmm */
+#foo = 900;
+',
+            ],
+        ];
+    }
 }
--- a/tests/Fixer/Phpdoc/PhpdocToCommentFixerTest.php
+++ b/tests/Fixer/Phpdoc/PhpdocToCommentFixerTest.php
@@ -552,6 +552,17 @@ $loader = require __DIR__.\'/../vendor/autoload.php\';
 ',
         ];
 
+        $cases[] = [
+            '<?php
+
+foreach ($this->subscribers[$eventClassName] as $subscriber) {
+    /** @noinspection PhpParamsInspection */
+    /** @psalm-suppress UndefinedInterfaceMethod */
+    $subscriber->notify($event);
+}
+',
+        ];
+
         return $cases;
     }
 


localheinz:
@kubawerlos 

> Can we go with option `ignored_tags` like for [`CommentToPhpdocFixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4385/files) instead?

Makes a lot of sense, will adjust!

Taluu:
I don't think adding just an option to "ignore some tags" is really dx-friendly. There are A LOT of annotations tags (doctrine, psalm, php ea inspections, ... and so on).

I think that if a line in a comment starts with `* @`, it should be safe to consider that it's an annot...

kubawerlos:
> I think that if a line in a comment starts with `* @`, it should be safe to consider that it's an annot...

@Taluu it seems you are totally missing the point of [PHPDoc](https://github.com/php-fig/fig-standards/blob/master/proposed/phpdoc.md#3-definitions).

The first criteria if a comment is/should be PHPDoc is: if it's placed before the structural element or not, so relying on the content is not a solution.

> There are A LOT of annotations tags (doctrine, psalm, php ea inspections, ... and so on).

Doctrine's tags must be in PHPDocs, so this is not really good example.

Taluu:
> The DocBlock is the combination of a DocComment, which is a type of comment, and a PHPDoc entity.

DocBlocks (which are also annotations, as PHPDocs tags actually) shouldn't be transformed then.

kubawerlos:
Why so? If they are not placed before the structural element they should.

Taluu:
It's not because they're not placed above a structural element that they shouldn't be an annotation.

Think of `@var`, `@codeCoverageIgnore`, ... and so on. `@var` is a valid phpdoc tag, yet not always used on a structural element...

kubawerlos:
>  `@var` is a valid phpdoc tag, yet not always used on a structural element...

Like when? I'd say it always is.

theseer:
> > `@var` is a valid phpdoc tag, yet not always used on a structural element...
> 
> Like when? I'd say it always is.

Like at any inline location where a type declaration might be needed to get IDE autocomplete or have psalm annotations:

```php
/** @var $foo SomeType */
$foo = $obj->someMethod();
```

Not saying a "proper" fix would probably be to have correct return type on the method `someMethod` - but that may not be an option.

Or, scroll back to the top and read the referenced issue at psalm (https://github.com/vimeo/psalm/issues/2344) and the very reason why @localheinz created the PR.

theseer:
Actually, I think there are various - potentially arbitrary - reasons to have inline docblocks with annotations: To disable inspections, to add assertions for or otherwise configure static analyzers, code coverage generation, ...

Having `PhpDocToCommentFixer` unconditionally replace inline docblocks breaks all of them. It makes perfect sense to replace mere comments that just happend to use the "wrong" syntax of course, but a docblock regardless of its location with any type of annotation in it is more than just a comment.

kubawerlos:
Adding `/** @var $foo SomeType */` before variable is perfectly fine usage of PHPDoc.

> To disable inspections, to add assertions for or otherwise configure static analyzers, code coverage generation, ...

All these reasons seem to have defined a set of annotations that would be uses -> thus the idea of having a list of tags to ignore.

> Having `PhpDocToCommentFixer` unconditionally replace inline docblocks breaks all of them.

Not unconditionally, simply according to the PSR5, which clarifies what PHPDoc is and what isn't.

If your proposition is to ignore PSR5 and trying to guess if PHPDoc should be converted comment basing of its content I think you need to create an issue to discuss it there - not here in a pull request.

src/Tokenizer/Analyzer/CommentsAnalyzer.php
```diff
@@ -92,6 +92,11 @@ public function isBeforeStructuralElement(Tokens $tokens, $index)
         return false;
     }
 
+    public function isAnnotation(Token $token)
+    {
+        return 1 === Preg::match('#\/\*{2}\s+.*@\w+#m', $token->getContent());
```
SpacePossum: shouldn't this also do a type check? For example this string literal `$a = "/** @var int $foo */"` and HEREDOC/NOWDOC s etc., should all not be considered annotations I think?

julienfalque:
I've got this issue a few times and I admit it is quite annoying. Adding a whitelist option would keep current stricly PSR-5 compliant behavior but I think it's not very DX-friendly and I prefer ignoring any `/**` comment that starts with a tag.

SpacePossum:
I like the the proposal of @kubawerlos , but I can see why it might be a bit of trouble to configure the exceptions. To be clear `@var` already works fine AFAIK.

The cases I found are those from popular SCA's like `@psalm-suppress`  (but i prefer baselines file now).

I propose to add the ignore option in the next minor release. On the next major we release with a new default value for it containing `@psalm-suppress` and some other that people can provide here that are biting them.

GrahamCampbell:
For some background, the reason this fixer changes some phpdoc containing annotations to comments is because of the proposed PSR phpdoc spec, which says that doc comments must document structural elements, so the sort of doc comments you have in your tests are illegal according to that spec.

HOWEVER, things have changed since that spec was first drafted, and the role of phpdoc has changed. No longer is phpdoc just for documentation generators to show the public api of classes, but these days we have genuinely usable static analyzers, and indeed these tools are now dictating the syntax (and semantics) of phpdoc (and annotations), and the sensible locations for phpdoc within a piece of code has changed, to allow for, for example, the specification of the type of an internal variable: something that was not relevant, even in 2014 (since the tooling was just not there, back then).

With all this in mind, I think this is a good change, in principle. The outstanding review comments need to be addressed, as do the failing tests, then this PR should be ready for 2nd reviews (and merge, if the php-cs-fixer team are happy!). ;)

GrahamCampbell:
Unrelated, but phpstorm processes comments as phpdoc, in error. Kinda annoying, but I guess they do it because:

1. not everyone understands the difference between a comment and a doc comment
2. people have run php-cs-fixer on their code and it's changed their doc comments to comments :trollface:

src/Tokenizer/Analyzer/CommentsAnalyzer.php
```diff
@@ -92,6 +92,11 @@ public function isBeforeStructuralElement(Tokens $tokens, $index)
         return false;
     }
 
+    public function isAnnotation(Token $token)
```
julienfalque, line 95: Shouldn't it accepts the `Tokens` list and an index like other methods do?

tests/Tokenizer/Analyzer/CommentsAnalyzerTest.php
```diff
@@ -302,4 +302,121 @@ public function providePhpdocCandidatePhp74Cases()
...
+$foo = 900;
+',
+            ],
+            [
+                '<?php
+
+/* @psalm-suppress hmmm */
+#foo = 900;
+',
+            ],
```
julienfalque, line 419: Maybe we can add cases like e.g. `/** no "at" sign */` and `/** @ followed by a space */`?

src/Fixer/Phpdoc/PhpdocToCommentFixer.php
```diff
@@ -88,6 +88,10 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
                 continue;
             }
 
+            if ($commentsAnalyzer->isAnnotation($token)) {
```
kubawerlos, line 91: Can we have this configurable to maintain BC? An option `ignore_annotations` (with default to `false`) seems the way.

kubawerlos (changes requested):

VincentLanglet:
What's the status of this ?
I started https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5572 before discovering this PR.

The fact that `/** @psalm- */` annotation are modified is really annoying.

masi:
Please, merge this code. Having @psalm-suppress (and possible many other linter/checker annotations) as collateral damage is no fun.

kubawerlos:
@masi as you can see there is not agreement in the discussion - you can always disable the rule if does any harm to your code.

masi:
But the rule is fine for all comments unless they contain (certain) annotations.
Of course if Psalm would allow it's annotations within standard comments I would also be happy :)

kubawerlos:
How is it fine? In the current state it does not convert to comment if there is any annotation in PHPDoc which makes the description "Docblocks should only be used on structural elements" not being true.

I'd prefer this one to go through: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5572

masi:
I don't care about what the PHPDoc . I'm using annotations for my IDE and other tools. Eg I'm using Psalm and this tool insists on docblocs for its annotations.

For sake of consistency I prefer to use a single * for normal comments.

Perhaps I failed to see the subtleties. If #5572 is in, I'm happy.

Sorry, for any misunderstandings and the added confusion.

julienfalque:
I'm still in favor of *not* adding any option and ignore all `/** ... */` comments as soon as they contain an annotation by default. Maybe we could rename the fixer to something like `DocCommentToRegularCommentFixer` to avoid the PHPDoc rules?

Anyway we need to make a decision here /ping @keradus @SpacePossum

masi:
> I'm still in favor of _not_ adding any option and ignore all `/** ... */` comments as soon as they contain an annotation by default.
Make this behaviour an opt-in and everybody should be happy.

> Maybe we could rename the fixer to something like `DocCommentToRegularCommentFixer` to avoid the PHPDoc rules?
Or have two rules.

VincentLanglet:
I think this can be closed since the feature was introduced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/5572

PR #4655, opened 2019-11-21
Improve docs - README

mvorisek:
Based on https://github.com/symfony/symfony/blob/1b4ab8108542daddd6ca96caf38b464ff3cc459a/src/Symfony/Component/Finder/Finder.php#L317

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1934,6 +1934,7 @@ The example below will add two rules to the default list of PSR2 set rules:
     ;
 
 **NOTE**: ``exclude`` will work only for directories, so if you need to exclude file, try ``notPath``.
+Both ``exclude`` and ``notPath`` methods accept only relative paths to the ones defined with the ``in`` method.
 
 See `Symfony\\Finder <https://symfony.com/doc/current/components/finder.html>`_
 online documentation for other `Finder` methods.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -2022,7 +2022,7 @@ The example below will add two rules to the default list of PSR2 set rules:
         ->setFinder($finder)
     ;
 
-**NOTE**: ``exclude`` will work only for directories, so if you need to exclude file, try ``notPath``.
+**NOTE**: ``exclude`` will work only for directories, so if you need to exclude file, try ``notPath``. Both ``exclude`` and ``notPath`` methods accepts only relative paths to the ones defined with the ``in`` method.
```
kubawerlos:
```suggestion
**NOTE**: ``exclude`` will work only for directories, so if you need to exclude file, try ``notPath``. Both ``exclude`` and ``notPath`` methods accept only relative paths to the ones defined with the ``in`` method.
```

mvorisek: Plural - so it should be accept, should not?

mvorisek:
> 
> 
> Also tests are failing - see the README information at https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.0/CONTRIBUTING.md

What should I do? This is only improval of README, not code, I do not except any CI tests on this.

kubawerlos:
@mvorisek check the message from the failing test ;)

>  I do not except any CI tests on this

How do you think README is up to date with all the fixers in code?

mvorisek:
> 
> 
> @mvorisek check the message from the failing test ;)
> 
> > I do not except any CI tests on this
> 
> How do you think README is up to date with all the fixers in code?

Fixed.

keradus:
Thank you @mvorisek.

keradus:
(merged to 2.15 instead of master)

PR #4659, opened 2019-11-22
NoSuperfluousPhpdocTagsFixer - Make regex compatible with large input/subject

mvorisek:
Fix regex with greedy modifier to make it compatible with large input.

Needed for example for: https://github.com/PrestaShop/PrestaShop/blob/1.6.1.x/tools/tcpdf/tcpdf.php with rule:
```
    ['no_superfluous_phpdoc_tags' => array (
          'allow_mixed' => true,
          'allow_unused_params' => true,
        )
    ]
```
without the greedy operator it ends with PHP error `preg_last_error() === 6`.


julienfalque:
Looking good. Though, as a fix, this should target the lowest maintained branch instead of master (2.15 currently). Could you please rebase your work onto it and update this PR to target it?

mvorisek:
> 
> 
> Looking good. Though, as a fix, this should target the lowest maintained branch instead of master (2.15 currently). Could you please rebase your work onto it and update this PR to target it?

Done

kubawerlos:
> without the greedy operator it ends with PHP error preg_last_error() === 6.

Can we have a test that would cause that error without this fix?

mvorisek:
> 
> > Fix regex with greedy modifier to make it compatible with large input.
> 
> I see different change here.

Yeah, "greedy quantifier" to be more exact.

mvorisek:
> 
> 
> I've run `master` against mentioned files and all went well.
> 
With that file it was failing at least on WIndows 10 x64 with PHP 7.3.11 installed from Chocolatey.

kubawerlos:
AppVeyor is testing using Windows set up, so the test would fail there.

mvorisek:
> I've run `master` against mentioned files and all went well.

I posted here a link for never version for which the fixer was passing. Please retry this issue with: https://github.com/PrestaShop/PrestaShop/blob/1.6.1.x/tools/tcpdf/tcpdf.php

I am out of resources to provide tests for this. But probably some long string/test file to mimic the failing subject of the regex can be tested if needed.

mvorisek:
> 
> 
> I will put this to fridge until we have the change covered by test.
> IMHO it's not the best option to do "random" changes to regex without any test covering it

I just optimized a regex that should be already fully covered with tests. No extra functionality or changed behaviour.

julienfalque:
Note that using a possessive quantifier instead of a greedy one isn't just an optimization, it actually changes the behavior of the regular expression.

Though I think it still does what we want it to do here. Having a test case that only proves a performance issue on a very large input might be tricky so I'm ok to merge this without new test case.

keradus:
i disagree - it's not just refactoring where we aim to remain same functionality (=same tests) while simplifying code. we claim to fix the issue, but no issue presented in tests.

mvorisek:
> 
> 
> i disagree - it's not just refactoring where we aim to remain same functionality (=same tests) while simplifying code. we claim to fix the issue, but no issue presented in tests.

What about running all rules against top1000 PHP repos + a list of problematic (to fix) repos? I tried this locally and found a lot of other issues. A lot of CPU time, but it can discover unknown errors and monitor their decrease.

mvorisek:
Helpt wanted - please add tests.

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4783, thanks for the work in this PR @mvorisek :+1:

PR #4661, opened 2019-11-23
ExplicitStringVariableFixer - variables pair if one is already explicit

kubawerlos:
1st added test case works fine, 2nd one (when space between is removed) is failing.

Ping @Slamdunk for review.


julienfalque:
Thank you @kubawerlos.

Issue #4663, opened 2019-11-23
"non_printable_character" is throwing linting exception


mvorisek:
#### The PHP version you are using (`$ php -v`):
=> 7.3.11

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> 2.16.0

This file: https://github.com/PrestaShop/PrestaShop/blob/1.6.1.x/install-dev/langs/zh/install.php

can not be fixed with this very core/simple rule:
```
        'non_printable_character' => [
            'use_escape_sequences_in_strings' => true,
        ],
```

CLI error output:
```
        [PhpCsFixer\Linter\LintingException]                                                                            
        PHP Parse error: syntax error, unexpected '">点击这里</a>重新启动' (T_CONSTANT_ENCAPSED_STRING), expecting ')' on line 170.
```

mvorisek:
The issue is here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/3e5dd53a0ab1fbe87057b06a027d8d072457ca3e/src/Fixer/Basic/NonPrintableCharacterFixer.php#L153

currently the code is buggy, it does not escape `"` in newly create double quoted string.

very simple test file:
```
echo '"%s">​​过程'; // there is non-printable char after the ">"
// if it can not be copied from Github, you can create it locally by:
// file_put_contents('test_file__issue_csfxer_4663.php', 'echo \'' . hex2bin('2573223ee2808be2808be8bf87e7a88b') . '\';');
```

In general string like `'lorem<bad_char>ipsum'` can be simple escaped like: `'lorem' . "<bad_char_escaped>" . 'ipsum'`. Otherwise the original string may become harder to read and to fix (you will have to undo all escapes).

PR #4664, opened 2019-11-23
Fix double quote escaping in originally single quoted string

mvorisek:
Fix https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4663


GrahamCampbell:
@mvorisek Please add tests, before this PR can be considered.

mvorisek:
@GrahamCampbell Analyzed and fixed internally, out of internal resources for tests, help needed

mvorisek:
Fixed in PR https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4675

PR #4665, opened 2019-11-23
DX: .composer-require-checker.json - drop duplicated symbol

keradus:


keradus:
silly me, contract vs component

PR #4672, opened 2019-11-25
Feature add sort by depth algorithm for OrderedClassElementsFixer

sidux:
This adds the possibility the sort class methods by depth meaning called methods will be placed after the caller like in the option "Keep dependent methods together depth-first-order" of PhpStorm.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1301,9 +1301,9 @@ Choose from the list of available rules:
     'property_public', 'property_protected', 'property_private',
     'construct', 'destruct', 'magic', 'phpunit', 'method_public',
     'method_protected', 'method_private']``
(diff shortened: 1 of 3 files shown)


sidux:
The `sort_algorithm` was only string before i made it array and kept string (backward compatibility) so that we can use something like `['alpha', 'depth']` or 'alpha' as a `sort_algorithm` and that's the result on the `README.rst`

SpacePossum:
Thanks @sidux for the PR, looks promising (based on the unit test samples)! Need some more time to review it properly though

(and I've to check what is up with that README generation indeed)

sidux:
I'm closing this, since it doesn't seem that requested.

sidux:
I'm closing this, since it doesn't seem that requested.

PR #4673, opened 2019-11-27
FinalStaticAccessFixer - deprecate

julienfalque:
In v2.16, we added rules `final_static_access` and `self_static_accessor` but they almost do the same thing so we can actually drop one. I chose to deprecate `final_static_access` in favor of `self_static_accessor` because the former does not support anonymous classes while the latter does.

/cc @ntzm @SpacePossum as authors of the rules.


SpacePossum:
Thank you @julienfalque.

PR #4676, opened 2019-11-30
NoBlankLinesAfterPhpdocFixer — Do not strip newline between docblock and use statements

mollierobbert:
When a docblock is followed by a use statement, it is most likely a file-level docblock. File-level docblocks should be followed by a newline as per the phpDoc standard: https://docs.phpdoc.org/getting-started/your-first-set-of-documentation.html

Example:

```
<?php
/**
 * Some license info.
 */

use Foo\Bar;

/**
 * My class.
 */
class MyClass
...
```


SpacePossum:
Thanks for creating the PR @mollierobbert  :+1: 

Could you target your PR to 2.15? that is the lowest supported branch ATM and since we merge upstream from it your fix will end up by as many releases/users as possible

mollierobbert:
Thanks @SpacePossum. I wouldn't know how to do that in GitHub. I created a new PR targeted to 2.15: #4712

PR #4679, opened 2019-11-30
NativeConstantInvocationFixer - add "strict" flag

kubawerlos:
@nicolas-grekas is it OK to add it to `@Symfony` rule set (I did it as it is very similar "strict" as for `NativeFunctionInvocationFixer`)?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -970,6 +970,8 @@ Choose from the list of available rules:
   - ``include`` (``array``): list of additional constants to fix; defaults to ``[]``
   - ``scope`` (``'all'``, ``'namespaced'``): only fix constant invocations that are made
     within a namespace or fix all; defaults to ``'all'``
+  - ``strict`` (``bool``): whether leading ``\`` of constant invocation not meant to
+    have it should be removed; defaults to ``false``
 
 * **native_function_casing** [@Symfony, @PhpCsFixer]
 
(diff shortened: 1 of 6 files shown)


README.rst
```diff
@@ -956,6 +956,8 @@ Choose from the list of available rules:
   - ``include`` (``array``): list of additional constants to fix; defaults to ``[]``
   - ``scope`` (``'all'``, ``'namespaced'``): only fix constant invocations that are made
     within a namespace or fix all; defaults to ``'all'``
+  - ``strict`` (``bool``): whether leading ``\`` of constant call not meant to have it
```
SpacePossum: if would be prefer `whether leading ``\`` of constant usage not meant to have it should be removed` as constants are not really called, WDYT?

SpacePossum:
Im very much :+1:  for adding to the SF set

nicolas-grekas:
Me too: `\` should be removed on constants that don't need it.

src/Fixer/ConstantNotation/NativeConstantInvocationFixer.php
```diff
@@ -225,14 +225,18 @@ protected function createConfigurationDefinition()
                 ->setAllowedValues(['all', 'namespaced'])
                 ->setDefault('all')
                 ->getOption(),
+            (new FixerOptionBuilder('strict', 'Whether leading ``\`` of constant invocation not meant to have it should be removed.'))
```
julienfalque: `` ` `` should not be doubled here (they get tripled in the README file).

kubawerlos: Of course 🤦‍♂ - fixed.

SpacePossum:
Thank you @kubawerlos.

PR #4683, opened 2019-12-01
Moved brew install option higher up

agrublev:
It's the simplest choice for all tech-savvy mac users, and by being so low it means most of us will just run the first complicated set of steps instead <3

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,13 @@ Locally
 
 Download the `php-cs-fixer.phar`_ file and store it somewhere on your computer.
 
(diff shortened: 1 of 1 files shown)


SpacePossum:
Hi and thanks for the PR!

For now I'm going to close it as I think it might help some developers, however the argument for the change can be provided by `composer` - users as well. They might (after your change is applied) go through the trouble installing `brew` first and than read they can use `composer` :)

Thanks for using the tool and thinking about improving the documenting based on your experience, much appreciated!

PR #4685, opened 2019-12-02
Improve Symfony 5 compatibility

### LINKED ISSUE #4684: Error after upgrading to v2.16.1
I get the following error after upgrading to 2.16.1 (from 2.15.0):

```
PHP Fatal error:  Uncaught TypeError: Argument 2 passed to Symfony\Component\EventDispatcher\EventDispatcher::dispatch() must be an instance of Symfony\Component\EventDispatcher\Event or null, instance of PhpCsFixer\FixerFileProcessedEvent given, called in /vendor/friendsofphp/php-cs-fixer/src/Runner/Runner.php on line 300 and defined in /vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/EventDispatcher.php:37
```

=> 7.2.25

=> 2.16.1

=> vendor/bin/php-cs-fixer fix -v

```php
    ->exclude('var')
    ->exclude('node_modules')
    ->exclude('etc')
    ->exclude('build')
    ->exclude('vendor')
    ->exclude('web')
    ->in(__DIR__)
;

    ->setRules([
        '@Symfony' => true,
        '@Symfony:risky' => true,
        '@DoctrineAnnotation' => true,
        '@PHP71Migration' => true,
        '@PHP71Migration:risky' => true,
        '@PHPUnit75Migration:risky' => true,
        'multiline_whitespace_before_semicolons' => ['strategy' => 'new_line_for_chained_calls'],
        'multiline_comment_opening_closing' => true,
        'ordered_imports' => true,
        'native_function_invocation' => true,
        'explicit_indirect_variable' => true,
        'explicit_string_variable' => true,
        'escape_implicit_backslashes' => true,
        'mb_str_functions' => true,
        'logical_operators' => true,
        'php_unit_method_casing' => ['case' => 'snake_case'],
        'php_unit_test_annotation' => ['style' => 'annotation'],
        'no_unreachable_default_argument_value' => true,
        'declare_strict_types' => false,
        'void_return' => false,
        'single_trait_insert_per_statement' => false,
        'simple_to_complex_string_variable' => true,
        'no_superfluous_phpdoc_tags' => ['allow_mixed' => true],
    ])
    ->setRiskyAllowed(true)
    ->setUsingCache(true)
    ->setFinder($finder)
;
```

keradus:
Closes #4684

one can install contract on his own, even if he is using Sf version that does not rely on contracts,
for that, we need to check if actually Sf component relies on contract, not only the fact that contract exists.

TODO:
- [x] look for other places when we check that contract exists and nothing more (edit by @julienfalque: only found one in the related test, not sure it's worth the hassle)

### DIFF
--- a/src/Event/Event.php
+++ b/src/Event/Event.php
@@ -12,10 +12,13 @@
 
 namespace PhpCsFixer\Event;
 
+use Symfony\Component\EventDispatcher\EventDispatcher;
+use Symfony\Contracts\EventDispatcher\EventDispatcherInterface;
+
 // Since PHP-CS-FIXER is PHP 5.6 compliant we can't always use Symfony Contracts (currently needs PHP ^7.1.3)
-// This conditionnal inheritance will be useless when PHP-CS-FIXER no longer supports PHP versions
+// This conditional inheritance will be useless when PHP-CS-FIXER no longer supports PHP versions
 // inferior to Symfony/Contracts PHP minimal version
-if (class_exists(\Symfony\Contracts\EventDispatcher\Event::class)) {
+if (is_subclass_of(EventDispatcher::class, EventDispatcherInterface::class)) {
     class Event extends \Symfony\Contracts\EventDispatcher\Event
     {
     }


kbond:
I can confirm this fixes the issue. Here is my reproducer: https://github.com/kbond/php-cs-fixer-2.16.1-reproducer

X-Coder264:
Any chance of merging and tagging a new version with this soon? The PR is opened for almost a month already and it is blocking us from properly testing our package in CI (with different Symfony and PHP versions) -> https://github.com/trikoder/oauth2-bundle/pull/141

jvocampings:
Hello

do you know when this fix will be released ?

julienfalque:
Thank you @keradus.

Issue #4687, opened 2019-12-05
declare_strict_types: adds strict_types=1 even though strict_types=0 is set


JeroenVanOort:
#### The PHP version you are using (`$ php -v`):
=> PHP 7.3.11-1+ubuntu18.04.1+deb.sury.org+1

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
=> PHP CS Fixer 2.16.1 Yellow Bird

#### The command you use to run PHP CS Fixer:
=> php-cs-fixer fix test.php --config=.php_cs_test

#### The configuration file you are using, if any:
```php
<?php

declare(strict_types=1);

return PhpCsFixer\Config::create()
    ->setRules([
        'declare_strict_types' => true,
    ])
    ->setRiskyAllowed(true);
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
<?php declare(strict_types=0);
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
<?php declare(strict_types=1);
declare(strict_types=0);
```

 * with the changes you expected instead:
```php
<?php declare(strict_types=0);
```

julienfalque:
I don't think the rule supports ignoring files where strict mode is explicitly disabled. IMO this would be a new feature.

Though the current behavior (enabling strict mode everywhere) is buggy too: it should update the existing statement, i.e.:
```diff
-<?php declare(strict_types=0);
+<?php declare(strict_types=1);
```

JeroenVanOort:
While being very content with your swift reply, I'd like to share my thoughts on what part of this issue I consider a bug and what I consider a requested feature.

In my opinion, this rule's name and summary allow for two interpretations:
1. Require `strict_types` to be declared, regardless of it's value.
2. Require `strict_types=1`.

In my opinion, number 1 is the correct interpretation, as the name and summary do not explicitly specify a value. However, I do see a valid use case in number 2.

Therefore, I think that the bug is that the declaration is made twice while the file should have been left unchanged and that the requested feature is having a configurable option to allow forcing `strict_types=1`.

kubawerlos:
@JeroenVanOort 

> this rule's name and summary allow for two interpretations:

Agree, though the reason why it is risky ("forcing strict types will stop non strict code from working") make it clear that it should be the 2nd interpretations.

julienfalque:
[...]

PR #4689, opened 2019-12-07
DeclareStrictTypesFixer - fix for "strict_types" set to "0"

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4687

Ping @Seldaek, @SpacePossum (as authors) and @JeroenVanOort (as bug reporter) for review.

### DIFF
--- a/src/Fixer/Strict/DeclareStrictTypesFixer.php
+++ b/src/Fixer/Strict/DeclareStrictTypesFixer.php
@@ -82,52 +82,31 @@ final class DeclareStrictTypesFixer extends AbstractFixer implements Whitespaces
             return;
         }
 
-        $sequence = $this->getDeclareStrictTypeSequence();
-        $sequenceLocation = $tokens->findSequence($sequence, $searchIndex, null, false);
+        $sequenceLocation = $tokens->findSequence([[T_DECLARE, 'declare'], '(', [T_STRING, 'strict_types'], '=', [T_LNUMBER], ')'], $searchIndex, null, false);
         if (null === $sequenceLocation) {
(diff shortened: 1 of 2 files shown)


src/Fixer/Strict/DeclareStrictTypesFixer.php
```diff
@@ -82,62 +82,48 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
             return;
         }
 
-        $this->fixStrictTypesCasing($tokens, $sequenceLocation);
-    }
-
-    /**
-     * @return Token[]
-     */
-    private function getDeclareStrictTypeSequence()
```
GrahamCampbell: Why is this removed. The new code is slower and harder to read now, since the line is so long?

kubawerlos: @GrahamCampbell this is removed because the new code is faster (yes, it is) and easier to read now - I'd say one 152 characters long line is more readable than 27 lines long function.

julienfalque: I agree with @kubawerlos. Also static variables inside functions are easy memory leaks.

kubawerlos: To clarify:
- it is faster in runs that are not changing the code - for `findSequence` 6 new objects were created every time, and now none are created
- `getDeclareStrictTypeSequence` had to be updated anyway, as the check for `declare` could not use objects, because we have to search for any number, not only `1` - which is the core of this bugfix

julienfalque:
Thank you @kubawerlos.

PR #4691, opened 2019-12-08
Run Travis CI on stable PHP 7.4

kubawerlos:
- "Static Code Analysis" and "Fast Test" to be run on PHP 7.4
- each standard job to be run on different PHP (previously "Collect coverage" and "With migration rules" were run on the same PHP version)
- `nightly` instead of `7.4snapshot` as `assertUpcomingPhpVersionIsCoveredByCiJob`
- allow nightly to fail
- update `humbug/box` due to it's dependency `nikic/iter` having the [problem](https://github.com/nikic/iter/blob/v1.6.0/src/iter.fn.php#L3) with PHP 7.4

Needs https://travis-ci.community/t/some-extensions-are-missing-in-php-7-4-0-zip-gmp-sodium/6320 to be fixed

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ jobs:
     include:
         -
             stage: Static Code Analysis
-            php: 7.3
+            php: 7.4
             env: COMPOSER_FLAGS="--prefer-stable"
             install:
                 - travis_retry ./dev-tools/install.sh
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -81,12 +81,14 @@ jobs:
             <<: *STANDARD_TEST_JOB
             stage: Test
             php: 7.1
-            env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="~4.1.0"
+            name: 7.1 | Symfony 4.1
+            env: SYMFONY_DEPRECATIONS_HELPER=disabled PHP_CS_FIXER_TEST_USE_LEGACY_TOKENIZER=1 SYMFONY_VERSION="^4.1"
```
julienfalque: There's a subtle change here: now any 4.4.x version of Symfony can also be installed. Not sure this is desired.

kubawerlos: Reverted

julienfalque:
Thank you @kubawerlos.

Issue #4502, opened 2019-08-12
PSR-12 Support


julienfalque:
[PSR-12](https://www.php-fig.org/psr/psr-12/) just got approved.

I think PHP CS Fixer already supports almost all of its rules. The missing cases I could notice are the following:

---

- [5.1 if, elseif, else](https://www.php-fig.org/psr/psr-12/#51-if-elseif-else) ;
- [5.2 switch, case](https://www.php-fig.org/psr/psr-12/#52-switch-case) ;
- [5.3 while, do while](https://www.php-fig.org/psr/psr-12/#53-while-do-while) ;
- [5.4 for](https://www.php-fig.org/psr/psr-12/#54-for) ;
- [5.5 foreach](https://www.php-fig.org/psr/psr-12/#55-foreach) ;
- [5.6 try, catch, finally](https://www.php-fig.org/psr/psr-12/#56-try-catch-finally) ;

> [In control structures, ]expressions in parentheses MAY be split across multiple lines, where each subsequent line is indented at least once. When doing so, the first condition MUST be on the next line. Boolean operators between conditions MUST always be at the beginning or at the end of the line, not a mix of both.

I would create a fixer that could either:
- force boolean operators to always be at the beginning or the end of lines when a mix is found ;
- force boolean operators to always be at the beginning or the end regardless of their current position.

---

- [8. Anonymous Classes](https://www.php-fig.org/psr/psr-12/#8-anonymous-classes)

> The opening brace [of an anonymous class] MAY be on the same line as the class keyword so long as the list of implements interfaces does not wrap. If the list of interfaces wraps, the brace MUST be placed on the line immediately following the last interface.

I think this is currently covered by the rule `braces`. Not sure if we should add a new option to it or create a new fixer.

---

TODO

- [ ] support for boolean operator position in control structure expression
- [ ] support for anonymous class opening brace in case `implements` list wraps
- [ ] `@PSR12` ruleset

kbond:
What about [this](https://www.php-fig.org/psr/psr-12/#3-declare-statements-namespace-and-import-statements) - specifically the part about spaces between use class/function/const. It this possible currently?

dmvdbrugge:
No, the newlines between different use types is not yet possible, however there is a longstanding issue (without PR) addressing it: #3582

dmvdbrugge:
> > When doing so, the first condition MUST be on the next line.

Is this part already covered somewhere?

[...]

PR #4695, opened 2019-12-11
Initial PSR12 ruleset creation

evan-burrell:
Reference: https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4502#issuecomment-563958596

Create a new ruleset for PSR12, with single PSR1 rule

[PSR12-FIG](https://www.php-fig.org/psr/psr-12/)


SpacePossum:
Thanks for the starts on this :+1:

src/RuleSet.php
```diff
@@ -56,6 +56,9 @@ final class RuleSet implements RuleSetInterface
             'switch_case_space' => true,
             'visibility_required' => true,
         ],
+        '@PSR12' => [
+            '@PSR1' => true,
```
julienfalque: Why not `@PSR2`?

PR #4699, opened 2019-12-16
LineEndingFixer - handle "\r\r\n"

kubawerlos:
The added test is converting the input to expected, but it fails because:

> Code build on input code must match expected code.

We have two ways to fix it:
1. Update the RegEx `#\r\n|\n#` to `#\r+\n|\n#`
2. Update the RegEx `#\r\n|\n#` to `#\r\n|\r|\n#` and test case to:
```php
        $cases[] = [
            "<?php echo 'foo',\n\n'bar';",
            "<?php echo 'foo',\r\r\n'bar';",
        ];
```

Option 2 seems more right as the original `\r\r\n` is treated (at least in PHPStorm) as 2 line endings.

Ping @fabpot, @SpacePossum and @keradus as authors for opinion.

### DIFF
--- a/src/Fixer/Whitespace/LineEndingFixer.php
+++ b/src/Fixer/Whitespace/LineEndingFixer.php
@@ -67,7 +67,7 @@ final class LineEndingFixer extends AbstractFixer implements WhitespacesAwareFix
                     $tokens[$index] = new Token([
                         $token->getId(),
                         Preg::replace(
-                            "#\r\n|\n#",
+                            '#\R#',
                             $ending,
                             $token->getContent()
                         ),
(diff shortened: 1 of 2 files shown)


julienfalque:
What about `#\R#`?

kubawerlos:
Even better :)

julienfalque:
I'm not against changes proposed here but, if I remember correcly, the current behavior (not supporting `\r`) is on purpose because the syntax is only used on old systems.

kubawerlos:
The problem is that the current behavior is wrong - fixer cannot leave the code that requires fixing by itself.

SpacePossum:
Not sure if we left `\r` support in/out by design, or didn't bother trying to fix it because it is only used on older setups as we are talking about before macOS X which was released in 2002.

With the proposed changes in this PR, which users might be effected in a way that we might not want?

kubawerlos:
@SpacePossum not sure, some that have really old piece of code I guess, I doubt anyone is using `\r` in new code.

SpacePossum:
Can you propose the fix in the PR in a new commit? I think it will be fine TBH

SpacePossum:
Thank you @kubawerlos.

PR #3820, opened 2018-06-08
Braces - (re)indenting comment issues

SpacePossum:
fixes #3819
fixes #4814

replaces https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4700

PR #4700, opened 2019-12-16
BracesFixer must be run before NoTrailingWhitespaceInCommentFixer

kubawerlos:


tests/Fixtures/Integration/priority/braces,no_trailing_whitespace_in_comment.test
```diff
@@ -0,0 +1,27 @@
...
+    return foo($i);
+     */
+}
+
+--INPUT--
+<?php
+if (true) {
+    $i += 2;
+    return foo($i);
+/*
```
julienfalque: This comment does not contain any trailing space, so I think the test would still pass if you revert priority changes.

kubawerlos: If it would contain trailing spaces they would be handled by `no_trailing_whitespace_in_comment` during the run and all would be good.

The `braces` fixer will indent each line of the comment with 4 spaces making the empty line having "trailing_whitespace_in_comment", so it must run first from the pair.

> I think the test would still pass if you revert priority changes.

`IntegrationTest` verifies that swapping the declared order of tests is making the transformation fail.

julienfalque: > The braces fixer will indent each line of the comment with 4 spaces making the empty line having "trailing_whitespace_in_comment", so it must run first from the pair.

Ok, I understand the problem now. I think this is the wrong solution, though. IMO we should fix this the same way as was done for `heredoc_indentation` in #4548: changing `braces` to avoid adding the extra spaces in the first place, so that it doesn't rely on a subsequent rule to clean its mess.

SpacePossum:
I think the conflict goes away with https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3820 , maybe someone can review it?

SpacePossum:
replaced by https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3820

Issue #1334, opened 2015-07-29
[Proposal] Ordering Traits


vinkla:
Lets say I've included two traits within my class and I want to order them by name.

**Before:**

``` php
use SerializerTrait, DispatchesJobs;
```

**After:**

``` php
use DispatchesJobs, SerializerTrait;
```

Don't know if this will be a problem PHP-wise? Otherwise it would make a good feature.

stof:
I think this might change the behavior of the code if 2 traits are defining a method with the same name

vinkla:
If two traits insert a method with the same name, isn't a fatal error is produced? If the conflict is not explicitly resolved?

http://stackoverflow.com/a/25064632/4801875

gharlan:
I would like to see trait equivalents of `MultipleUseFixer` and `OrderedUseFixer` (namespace usages).

```php
use SerializerTrait, DispatchesJobs;

// After MultipleUseTraitFixer
use SerializerTrait;
use DispatchesJobs;

// After OrderedUseTraitFixer
use DispatchesJobs;
use SerializerTrait;
```

theofidry:
Yeah a fatal error is thrown so the order of the traits shouldn't have an impact.

keradus:
:+1: for feature, but it also must support aliasing and resolving conflicts.

SpacePossum:
bit similar but we could also order the interfaces a class implements by name...

JarJak:
Bump

ntzm:
@JarJak the best way of getting this implemented is to have a crack at it yourself! If you need any pointers feel free to as in the Gitter chat

SpacePossum:
added in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4701 by @julienfalque :)

PR #4701, opened 2019-12-17
OrderedTraitsFixer - introduction

julienfalque:
Implements #1334.


GrahamCampbell:
Surely this should be a risky fixer?

SpacePossum:
can you provide an example of such risky fix, conflict resolution for traits has to be done explicit already and that order does not matter

GrahamCampbell:
Ah, right. I see.

SpacePossum:
Thank you @julienfalque.

PR #4707, opened 2019-12-20
Fix typos

TysonAndre:


SpacePossum:
Thanks @TysonAndre !

Are these typos also on the 2.15 branch? If so please retarget so we get your fixes in as much as possible supported versions :)

julienfalque:
Thank you @TysonAndre.

PR #4708, opened 2019-12-21
Rename phpstan.neon to phpstan.neon.dist

GrahamCampbell:


svenluijten:
Needs adding in `.gitattributes` as well.

GrahamCampbell:
> Needs adding in .gitattributes as well.

No it doesn't? It already is?

GrahamCampbell:
Thanks. Updated.

SpacePossum:
Hi and thanks for the PR,

Can you provide any reasoning for this change? Like, would anyone outside of this project be effected by the file name, either before or after?

GrahamCampbell:
The same reasoning as the phpunit config has `.dist`. It allows local config to exist outside of the version control, without the extra extension.

svenluijten:
@SpacePossum `phpstan.neon` v `phpstan.neon.dist` works the same way as `phpunit.xml.dist`, which is explained best [here](https://stackoverflow.com/a/44764078/4795870) by Sebastian himself.

SpacePossum:
I know the concept, so ` This allows a developer to use a custom configuration without running the risk of accidentally checking it into version control.` applies to developers of this project. I see no reason why a developer in this project would want to create a custom version of this file without the intentions to add the changes. What would be a case for this (so not the generic concept if `dist` files, but why this change here in this PR)?

julienfalque:
I agree with @SpacePossum: I don't see a valid use case for this. This is useful for PHPUnit because the config file allows to define some output customization options, which can differ locally from the versioned file. This does not apply to PHPStan config file as far as I know.

Same reasoning regarding `.php_cs.dist` for which I think we should drop support.

SpacePossum:
closing after discussion, feel free to keep discussing and who knows maybe we change our minds!

GrahamCampbell:
If php-cs-fixer uses the phpstan baseline thing, then this conversion is quite useful.

GrahamCampbell:
Example from a `Makefile`:

```make
phpstan-analyze:
	@docker run -it -w /data -v ${PWD}:/data:delegated --entrypoint vendor/bin/phpstan --rm registry.gitlab.com/grahamcampbell/php:7.4-base analyze

phpstan-baseline:
	@sed -e '2,1d' < phpstan.neon.dist > phpstan.neon && docker run -it -w /data -v ${PWD}:/data:delegated --entrypoint vendor/bin/phpstan --rm registry.gitlab.com/grahamcampbell/php:7.4-base analyze -c phpstan.neon --no-ansi --no-progress --error-format baselineNeon > phpstan-baseline.neon && rm -f phpstan.neon || rm -f phpstan.neon

phpstan-no-baseline:
	@sed -e '2,1d' < phpstan.neon.dist > phpstan.neon && docker run -it -w /data -v ${PWD}:/data:delegated --entrypoint vendor/bin/phpstan --rm registry.gitlab.com/grahamcampbell/php:7.4-base analyze -c phpstan.neon && rm -f phpstan.neon || rm -f phpstan.neon
```

`phpstan.neon.dist`:

```neon
includes:
  - phpstan-baseline.neon
  - vendor/nunomaduro/larastan/extension.neon

parameters:
  paths:
    - app
    - database

  level: 3

  ignoreErrors:
    - '#Unsafe usage of new static#'
    - '#Call to an undefined method#'
    - '#Access to an undefined property#'
```

The makefile generates a `phpstan.neon` with the baseline removed, as an intermediate step.

Issue #4709, opened 2019-12-22
SingleTraitInsertPerStatement - wrong formatting


javaDeveloperKid:
#### The PHP version you are using (`$ php -v`):
7.2

#### PHP CS Fixer version you are using (`$ php-cs-fixer -V`):
2.15.1

#### The command you use to run PHP CS Fixer:
php-cs-fixer fix

#### The configuration file you are using, if any:
```php
=> ....................................
```

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
class Test
{
    use FirstTrait,
        SecondTrait,
        ThirdTrait;
}
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
class Test
{
    use FirstTrait;
    use
        SecondTrait;
    use
        ThirdTrait;
}
```

 * with the changes you expected instead:
```php
class Test
{
    use FirstTrait;
    use SecondTrait;
    use ThirdTrait;
}
```

SpacePossum:
Thanks for reporting @javaDeveloperKid and thanks @kubawerlos for picking this up :)

SpacePossum:
fixed in https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4710

PR #4710, opened 2019-12-23
SingleTraitInsertPerStatement - fix formatting for multiline "use"

kubawerlos:
Fixes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4709

### DIFF
--- a/tests/Fixer/ClassNotation/SingleTraitInsertPerStatementFixerTest.php
+++ b/tests/Fixer/ClassNotation/SingleTraitInsertPerStatementFixerTest.php
@@ -83,6 +83,24 @@ final class Example
 }
 ',
             ],
+            'simple III' => [
+                '<?php
+class Example
+{
+    use Foo;use Bar;
+
+    public function baz() {}
+}
+',
+                '<?php
+class Example
+{
+    use Foo, Bar;
+
+    public function baz() {}
+}
+',
+            ],
             'multiple' => [
                 '<?php
 final class Example
@@ -100,6 +118,44 @@ final class Example
     use Foo00, Bar01;
     use Foo10, Bar11, Bar110;
     use Foo20, Bar20, Bar200, Bar201;
+}
+',
+            ],
+            'multiple_multiline' => [
+                '<?php
+final class Example
+{
+    use Foo;
+    use Bar;
+    use Baz;
+}
+',
+                '<?php
+final class Example
+{
+    use Foo,
+        Bar,
+        Baz;
+}
+',
+            ],
+            'multiple_multiline_with_comment' => [
+                '<?php
+final class Example
+{
+    use Foo;
+    use Bar;
+//        Bazz,
+    use Baz;
+}
+',
+                '<?php
+final class Example
+{
+    use Foo,
+        Bar,
+//        Bazz,
+        Baz;
 }
 ',
             ],
@@ -126,9 +182,9 @@ class ZZ
 use#2
 Z/* 2 */ #3
 #4
-;use #5
+;#5
 #6
-T#7
+use T#7
 #8
 ;#9
 #10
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
What does it do with the following example:

```php
<?php
final class Example
{
    use Foo,
        Bar,
//         Bazz,
        Baz;
}
```

kubawerlos:
@GrahamCampbell now it does well

GrahamCampbell:
There seems to still be an issue:

image

The newline in between the `use` and `public function` is getting deleted.

kubawerlos:
@GrahamCampbell where is this happening? Because definitely not in `SingleTraitInsertPerStatementFixer` in this PR.

GrahamCampbell:
Could you add the following test case:

```php
<?php

class Example
{
    use Foo, Bar;

    public function baz() {}
}
```

it should be fixed to:

```php
<?php

class Example
{
    use Foo;
    use Bar;

    public function baz() {}
}
```

I think your code will actually fix it to:

```php
<?php

class Example
{
    use Foo;
    use Bar;
    public function baz() {}
}

File: tests/Fixer/ClassNotation/SingleTraitInsertPerStatementFixerTest.php [hunk 1]
```diff
@@ -81,6 +81,24 @@ final class Example
 {
     use Foo, Bar ;
 }
+',
+            ],
+            'simple III' => [
+                '<?php
+class Example
+{
+    use Foo;use Bar;
```
GrahamCampbell: Should be 
```
use Foo;
use Bar;
```

not:

```
use Foo;use Bar;
```

kubawerlos:
Added, through the expected result is different from each of your examples - we are not adding any new line if none between imports (as in very first test case), so this what would be: 
```php
<?php
class Example
{
    use Foo;use Bar;

    public function baz() {}
}
```

GrahamCampbell:
Right, so your fixer is fixing to the wrong thing, and then another fixer is breaking up the statements?

tests/Fixer/ClassNotation/SingleTraitInsertPerStatementFixerTest.php
kubawerlos: No, it shall not -> see the test case `simple I`.

GrahamCampbell:
// cc @SpacePossum. This fixer is meant to fix to the Symfony and PSR-12 style, but seems to not do that?

kubawerlos:
> Right, so your fixer is fixing to the wrong thing

WHAAAT? Are you kidding? See the issue it's fixing - there is a test case that covers this issue...

See the original tests - they are not meant to break the lines, only to do what the fixer describes to do, which is 'Each trait `use` must be done as single statement.' - STATEMENT, not a line.

> This fixer is meant to fix to the Symfony and PSR-12 style

Since when? It seems like overthinking. We have a fixer to split the imports into separate lines since a long time ago, this fixer is meant to be run before it and @SpacePossum thought about it as he wrote this integration test: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/v2.16.1/tests/Fixtures/Integration/priority/single_trait_insert_per_statement,braces.test

GrahamCampbell:
> See the original tests - they are not meant to break the lines,

Yes, I know. I am saying I think the common sense thing to do is to have the fixer split onto new lines, even though the original implementation didn't do this. This is why I have cced in the original fixer author.

GrahamCampbell:
Because of exactly the problem I am describing, with the example I provided. Trying to have separation of responsibilities just leads to it being harder to use, and producing output that is not desirable.

GrahamCampbell:
Though, maybe this is something we should discuss for **after** this PR. ;)

GrahamCampbell:
This PR itself looks fine.

SpacePossum:
> This fixer is meant to fix to the Symfony and PSR-12 style,

No, it is meant/designed to make multiple use statements

> Trying to have separation of responsibilities just leads to it being harder to use, and producing output that is not desirable.

I disagree, please provide more than an opinion and I might change my mind. 

For now, lets leave this out of the PR, the only issue I've seen so far is that there is no priority test between this fixer and `class_attributes_separation`.

SpacePossum:
Thank you @kubawerlos.

PR #4711, opened 2019-12-24
Ensure that files from "tests" directory in release are autoloaded

kubawerlos:
Ping @julienfalque as the breaker :smile: (author of `tests/Test/IsIdenticalConstraint.php`) for review.


julienfalque:
Thank you @kubawerlos.

PR #4712, opened 2019-12-26
NoBlankLinesAfterPhpdocFixer — Do not strip newline between docblock and use statements

mollierobbert:
When a docblock is followed by a use statement, it is most likely a file-level docblock. File-level docblocks should be followed by a newline as per the phpDoc standard: https://docs.phpdoc.org/getting-started/your-first-set-of-documentation.html

Example:

```
<?php
/**
 * Some license info.
 */

use Foo\Bar;

/**
 * My class.
 */
class MyClass
...
```


mollierobbert:
Replaces PR #4676.

SpacePossum:
Thank you @mollierobbert.

PR #4713, opened 2019-12-27
NoSuperfluousPhpdocTagsFixer - Remove superfluous phpdocs for typed properties

ruudk:
Changes superfluous PHPdocs like 
```php
class Foo {
    /**
     * @var null|Bar
     */
    private ?Bar $bar;
}
```

to this:
```php
class Foo {
    /**
     */
    private ?Bar $bar;
}
```

### DIFF
--- a/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
+++ b/src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
@@ -36,7 +36,7 @@ final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements Config
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Removes `@param` and `@return` tags that don\'t provide any useful information.',
+            'Removes `@param`, `@return` and `@var` tags that don\'t provide any useful information.',
             [
                 new CodeSample('<?php
 class Foo {
@@ -136,6 +136,8 @@ class Foo {
 
             if ($token->isGivenKind(T_FUNCTION)) {
                 $content = $this->fixFunctionDocComment($content, $tokens, $index, $shortNames);
+            } elseif ($token->isGivenKind(T_VARIABLE)) {
+                $content = $this->fixPropertyDocComment($content, $tokens, $index, $shortNames);
             }
 
             if ($this->configuration['remove_inheritdoc']) {
@@ -188,7 +190,7 @@ class Foo {
 
         $index = $tokens->getNextMeaningfulToken($docCommentIndex);
 
-        $kindsBeforeProperty = [T_STATIC, T_PRIVATE, T_PROTECTED, T_PUBLIC];
+        $kindsBeforeProperty = [T_STATIC, T_PRIVATE, T_PROTECTED, T_PUBLIC, CT::T_NULLABLE_TYPE, CT::T_ARRAY_TYPEHINT, T_STRING, T_NS_SEPARATOR];
 
         if (!$tokens[$index]->isGivenKind($kindsBeforeProperty)) {
             return null;
@@ -247,6 +249,38 @@ class Foo {
                 $annotation->remove();
             }
         }
+
+        return $docBlock->getContent();
+    }
+
+    /**
+     * @param string $content
+     * @param int    $docCommentIndex
+     *
+     * @return string
+     */
+    private function fixPropertyDocComment($content, Tokens $tokens, $docCommentIndex, array $shortNames)
+    {
+        $docBlock = new DocBlock($content);
+
+        $index = $tokens->getNextMeaningfulToken($docCommentIndex);
+
+        $kindsBeforeProperty = [T_STATIC, T_PRIVATE, T_PROTECTED, T_PUBLIC];
+
+        if (!$tokens[$index]->isGivenKind($kindsBeforeProperty)) {
+            return $content;
+        }
+
+        do {
+            $index = $tokens->getNextMeaningfulToken($index);
+
+            $propertyTypeInfo = $this->parseTypeHint($tokens, $index);
+            foreach ($docBlock->getAnnotationsOfType('var') as $annotation) {
+                if ($this->annotationIsSuperfluous($annotation, $propertyTypeInfo, $shortNames)) {
+                    $annotation->remove();
+                }
+            }
+        } while ($tokens[$index]->isGivenKind($kindsBeforeProperty));
 
         return $docBlock->getContent();
     }
@@ -330,7 +364,7 @@ class Foo {
         }
 
         return [
-            'type' => $type,
+            'type' => '' === $type ? null : $type,
             'allows_null' => $allowsNull,
         ];
     }
@@ -344,6 +378,8 @@ class Foo {
     {
         if ('param' === $annotation->getTag()->getName()) {
             $regex = '/@param\s+(?:\S|\s(?!\$))+\s\$\S+\s+\S/';
+        } elseif ('var' === $annotation->getTag()->getName()) {
+            $regex = '/@var\s+\S+(\s+\$\S+)?(\s+)([^$\s]+)/';
         } else {
             $regex = '/@return\s+\S+\s+\S/';
         }
--- a/tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
+++ b/tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
@@ -116,6 +116,49 @@ class Foo {
                 null,
                 ['allow_mixed' => true],
             ],
+            'allow_mixed=>false on property' => [
+                '<?php
+class Foo {
+    /**
+     */
+    private $bar;
+}',
+                '<?php
+class Foo {
+    /**
+     * @var mixed
+     */
+    private $bar;
+}',
+                ['allow_mixed' => false],
+            ],
+            'allow_mixed=>false on property with var' => [
+                '<?php
+class Foo {
+    /**
+     */
+    private $bar;
+}',
+                '<?php
+class Foo {
+    /**
+     * @var mixed $bar
+     */
+    private $bar;
+}',
+                ['allow_mixed' => false],
+            ],
+            'allow_mixed=>false on property but with comment' => [
+                '<?php
+class Foo {
+    /**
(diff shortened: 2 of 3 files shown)


tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -1327,6 +1327,174 @@ class Foo {
...
+    /**
+     */
+    private object $bar;
+}',
+                '<?php
+class Foo {
+    /**
+     * @var object
+     */
+    private object $bar;
```
julienfalque: Please add test cases:
* with the `allow_mixed` option;
* that remove `@var mixed` when there is no type declaration (can run on PHP lower than 7.4);
* with fully qualified class names, e.g: `@var \Foo\Bar\Baz`;
* where the annotation is *not* removed (because it does not match the actual type declaration), e.g.:
```php
/**
 * @var FooImplementation1|FooImplementation2
 */
private FooInterface $foo;
```

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -250,6 +252,57 @@ private function fixFunctionDocComment($content, Tokens $tokens, $functionIndex,
...
+    private function fixPropertyDocComment($content, Tokens $tokens, $docCommentIndex, array $shortNames)
+    {
+        $docBlock = new DocBlock($content);
+
+        $index = $tokens->getNextMeaningfulToken($docCommentIndex);
+
+        $kindsBeforeProperty = [T_STATIC, T_PRIVATE, T_PROTECTED, T_PUBLIC];
+
+        if (!$tokens[$index]->isGivenKind($kindsBeforeProperty)) {
+            return null;
```
julienfalque: This violates the return type (`string`). Maybe this `if` should be done in the caller function so we never get to this point?

ruudk: Fixed.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
ruudk: Added the tests. Are my assuptions with `\Foo\Bar` correct? Should it also support this:
```diff
 namespace App;
 use \Foo\Bar;

 /**
-  * @var \Foo\Bar
  */
 private Bar $foo;
```
?

julienfalque: Yes, this should be supported as well.

julienfalque:
Please do not amend/squash commits as it forces reviewers reading the entire diff on every push instead of only reading added changes.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -1327,6 +1327,245 @@ class Foo {
...
+            'phpdoc does not match property typehint' => [
+                '<?php
+class Foo {
+    /**
+     * @var FooImplementation1|FooImplementation2
+     */
+    private FooInterface $bar;
+}',
+            ],
+            'allow_mixed=>false' => [
```
julienfalque: The tests without actual type declaration should run on PHP lower than 7.4.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
```diff
@@ -1327,6 +1327,245 @@ class Foo {
...
+                '<?php
+class Foo {
+    /**
+     */
+    private $bar;
+}',
+                '<?php
+class Foo {
+    /**
+     * @var mixed
```
julienfalque: Please add the same test case but with a description after `mixed`. In this scenario the `@var` annotation must not be removed.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
ruudk: Already added in my squash commit last time.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
ruudk: Added.

tests/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixerTest.php
ruudk: Done.

ruudk:
@julienfalque Thanks for your review.

> Please do not amend/squash commits as it forces reviewers reading the entire diff on every push instead of only reading added changes.

Done, please have a look :)

julienfalque:
@keradus @SpacePossum @kubawerlos Do you consider this a whole new feature or an enhancement of the existing scope?

ruudk:
Description updated (new commit) + rebased with master

src/Fixer/Phpdoc/NoSuperfluousPhpdocTagsFixer.php
```diff
@@ -36,7 +36,7 @@ final class NoSuperfluousPhpdocTagsFixer extends AbstractFixer implements Config
     public function getDefinition()
     {
         return new FixerDefinition(
-            'Removes `@param` and `@return` tags that don\'t provide any useful information.',
+            'Removes `@param`, `@var` and `@return` tags that don\'t provide any useful information.',
```
kubawerlos: Let's have it in alphabetical order ;)

Also, you have to regenerate README.

ruudk: Done.

ruudk:
@julienfalque @keradus @SpacePossum @kubawerlos Anything I can do to get this merged?

julienfalque:
@ruudk You need to update the README file using the dedicated command: `php php-cs-fixer readme > README.rst`. This should fix CI builds.

Also, this PR will be a new feature once merged. When we add new functionality to existing rules, we usually do so using a new option disabled by default. This way, users that already rely on the rule won't have new changes applied to their code that are likely unexpected. Though in this case I would be :+1: to merge this without option (so always enabled). @SpacePossum @kubawerlos @keradus WDYT?

ruudk:
@julienfalque Thanks for giving me the command, I was not aware. Made the change.

ruudk:
I had to do a rebase to get rid of the merge commit, so that was a good opportunity to add everything as one commit.

ruudk:
Can we merge this?

julienfalque:
The PR looks good, yet our process requires at least two approvals from maintainers before merging important additions/changes.

Also we still need to agree whether:
* we merge it on 2.15 (enhancement) or master (new feature);
* if new feature, enabled by default or with a new option.

I think I'd actually merge this as an enhancement rather than a new feature. My reasoning is that the intent of the rule is to remove *all* superfluous PHPDoc tags. This includes `@var` tags on typed properties but those didn't exist yet at the time the rule was introduced.

ruudk:
@julienfalque I agree that this is an enhancement. 

@SpacePossum @kubawerlos @keradus What do you think? Can we merge this?

ruudk:
@julienfalque I see that @kubawerlos approved, so we can merge it 🥳

julienfalque:
Not so fast :)

> Also we still need to agree whether:
> * we merge it on 2.15 (enhancement) or master (new feature);
> * if new feature, enabled by default or with a new option.

This still needs to be decided.

SpacePossum:
I'd vote 2.15, no new option, as the fixer should remove superfluous Phpdoc's and this PR makes sure it does on 7.4 for properties

ruudk:
@SpacePossum Sounds like a plan. Can you give it a Github Approval? :)

SpacePossum:
I do you better, if you retarget to 2.15 I'll today or tomorrow :)

ruudk:
@SpacePossum Ah, sorry, I forgot that I targeted master. Tried to apply this on 2.15 but a lot has changed in that file. I don't want to rewrite this all, since I already spend a lot of time on this, so let's merge it to 2.17 then?

SpacePossum:
I see, can you try 2.16 than? I think for 2.16 and master the fixer is the same

ruudk:
@SpacePossum Done! :)

julienfalque:
I'm :-1: for merging this into 2.16. We should either merge it into 2.15 (current LTS) or master as new feature. 2.16 is for fixing bugs that don't exist on 2.15.

SpacePossum:
the issue here with the difference between 2.15 and master makes it not feasible to target 2.15, so 2.16 is the lowest supported candidate. There is no gain doing the change on master and not on 2.16 as far as I know.

kubawerlos:
> Tried to apply this on 2.15 but a lot has changed in that file.

That suggest some maybe important changes, so I'd insist to solve this and do merge to `2.15` then

> I already spend a lot of time on this

I bet still 100x less that @SpacePossum on maintaining this project ;) and since there are changes between `master` and `2.15` I imagine merging it to master would bring some conflicts in future.

SpacePossum:
I checked and it is not that hard to get this on 2.15.

@ruudk your changes to the fixer test cases can go in without adjustments.
For the fixer itself, copy paste your version of the 2.16, now remove the `remove_inheritdoc` option from the definition, the private method `removeSuperfluousInheritDoc` can go as well, leave the rest. Regenerate the README and it should be fine :)
Let me know if this doesn't work for you, than I can do it for you so you keep authorship of the change.

Please make a new branch of the current one, or a new one for the 2.15 version, this way, when I merge it to 2.16 I can resolve the merge conflict easy because we already have a good 2.16 version, thanks!

ruudk:
Thanks for helping me on this. I created a new PR that targets 2.15. https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4746

SpacePossum:
merged through https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4746  , thanks for keeping the PR it helped we dealing with the merge conflicts :+1:  :)

PR #4714, opened 2019-12-28
julienfalque:deprecate-final-static-access

bluedasher13:
Should be deprecated in v2.16 & v3.0 & master?
https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4673

> **julienfalque** commented on 28 Nov
> 
> In v2.16, we added rules ``final_static_access`` and ``self_static_accessor`` but they almost do the same thing so we can actually drop one. 
> 
> I chose to deprecate ``final_static_access`` in favor of ``self_static_accessor`` because the former does not support anonymous classes while the latter does.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -686,6 +686,7 @@ Choose from the list of available rules:
 * **final_static_access**
 
   Converts ``static`` access to ``self`` access in ``final`` classes.
+  DEPRECATED: use ``self_static_accessor`` instead.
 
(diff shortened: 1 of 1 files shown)


SpacePossum:
As @kubawerlos pointed out Julien also provided a PR which looks more complete ATM, as such I'm closing this PR in favor of his. Thanks for trying to deprecate the fixer, it is indeed needed :)

PR #4715, opened 2019-12-28
Enhancement: Install ergebnis/composer-normalize via Phive

localheinz:
This PR

* [x] installs `ergebnis/composer-normalize` via Phive

Follows #4536.


kubawerlos:
So each time a new version - even patch one - is released, `phive.xml` must be updated, right?

How is that better than having `"ergebnis/composer-normalize":"^2.1"` in `dev-tools/composer.json`?

keradus:
@kubawerlos , it's a valid concern, but it affects other tools installed via phive already

keradus:
Thank you @localheinz.

localheinz:
Thank you, @keradus and @kubawerlos!

PR #4720, opened 2020-01-03
Ignore VCS ignored files by default

julienfalque:
Replaces #3947 (ref https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3947#issuecomment-475628380).

### DIFF
--- a/src/Finder.php
+++ b/src/Finder.php
@@ -32,5 +32,9 @@ class Finder extends BaseFinder
             ->ignoreVCS(true)
             ->exclude('vendor')
         ;
+
+        if (method_exists($this, 'ignoreVCSIgnored')) {
+            $this->ignoreVCSIgnored(true);
+        }
     }
 }


kubawerlos:
It's not that easy, I've tried this some time ago -> see the AppVeyor tests.

IMHO we should drop this idea - if one wants to add `ignoreVCSIgnored` then should do it in the local config.

julienfalque:
Indeed, not worth the hassle IMO. WDYT @keradus @SpacePossum?

SpacePossum:
happy to skip/leave out, might be worth adding a note to the docs maybe?

julienfalque:
We already have this:

> See [Symfony\Finder](https://symfony.com/doc/current/components/finder.html) online documentation for other Finder methods.

which is enough IMO.

SpacePossum:
closing after discussion, @julienfalque feel free to reopen if I missed an action for us here :+1:

PR #4721, opened 2020-01-04
Braces - do while loop gets broken

GrahamCampbell:
The braces fixer is fixing:

```php
<?php
function f()
{
    do {
        foo();
    } //
    while (false);
}
```

to:

```php
<?php
function f()
{
    do {
        foo();
    } // while (false);
}
```

which is invalid syntax.

---

By comparison, the braces fixer leaves the following unchanged:

```php
<?php

do {
    foo();
} //
while (false);
```


julienfalque:
I added this test case into #4884 and it passes so I'm closing here in favor of #4884.

PR #4722, opened 2020-01-06
Fix Circle CI build

julienfalque:


kubawerlos:
Can we bump PHP version?

julienfalque:
Sure thing :)

kubawerlos:
One more thing: can we bump [xcode](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions), so maybe Python 3 will be installed with it, so we don't have to remove Python 2 and have Python 3 installed with PHP?

julienfalque:
Done. Version `11.3.0` still includes Python 2. It also includes PHP 7.2 and 7.3 but not PHP 7.4 so we still need to install it manually.

kubawerlos:
Thank you @julienfalque.

Issue #4674, opened 2019-11-29
Deprecate not_operator_with_space and not_operator_with_successor_space


julienfalque:
I noticed rules `not_operator_with_space` and `not_operator_with_successor_space` are quite similar so they could be merged into a single configurable rule. Then I remembered we also have `unary_operator_spaces`.

IMO we could:
* add an `operators` option to `unary_operator_spaces` (similar to the option for `binary_operator_spaces`)
* deprecate `not_operator_with_space` and `not_operator_with_successor_space` as their scope would be covered by the new option of `unary_operator_spaces`

WDYT?

SpacePossum:
sounds like a plan :)

PR #4735, opened 2020-01-13
NotOperatorWithSpaceFixer - NotOperatorWithSuccessorSpaceFixer - depr…

SpacePossum:
closes 

https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4674


README.rst
```diff
@@ -1851,6 +1852,12 @@ Choose from the list of available rules:
 
   Unary operators should be placed adjacent to their operands.
 
+  Configuration options:
+
+  - ``not_operator_space`` (``'leading_and_trailing'``, ``'no_trailing'``,
```
kubawerlos: This got me confused as the new user of PHP CS Fixer - why only those 3 options? Where are `no_leading_and_trailing`, `no_leading` and `one_leading` options? Are they "worse" then added 3?

Wouldn't be more clear to have 2 boolean options `leading` and `trailing` with value `true|false|null` to cover all combination?

SpacePossum: I moved the current logic of the fixers into one and made the difference in behavior available through these configuration options. I didn't spend time on investigating new or better options, so thanks for the suggestion!

kubawerlos: Ok, and `no_trailing` is the default for new fixer, I missed that, now I get it 👍

src/Fixer/Operator/NotOperatorWithSuccessorSpaceFixer.php
```diff
@@ -48,7 +50,15 @@ public function getDefinition()
      */
     public function getPriority()
```
julienfalque: This method can be dropped now.

src/Fixer/Operator/NotOperatorWithSpaceFixer.php
```diff
@@ -48,7 +50,15 @@ public function getDefinition()
      */
     public function getPriority()
```
julienfalque: This method can be dropped now.

src/Fixer/Operator/UnaryOperatorSpacesFixer.php
```diff
@@ -75,4 +131,19 @@ protected function applyFix(\SplFileInfo $file, Tokens $tokens)
...
     }
+
+    /**
+     * {@inheritdoc}
+     */
+    protected function createConfigurationDefinition()
+    {
+        return new FixerConfigurationResolver(
+            [
+                (new FixerOptionBuilder('not_operator_space', 'Space around logical `!` (`not`) operators.'))
```
julienfalque: Wouldn't it be better to have an `operators` option supporting all possible operators like `binary_operator_spaces` has?

julienfalque:
@SpacePossum deprecations cannot be added to a patch version, this must be merged into master.

SpacePossum:
Im pushing this back to the discussion about the request so we can agree on what to do there, thanks all for the reviewing!

Issue #4717, opened 2019-12-29
Fix more alias-like language construct, eg, die() -> exit


bluedasher13:
For configuration or updating questions please read the README and UPGRADE documentation,
or visit: https://gitter.im/PHP-CS-Fixer

When reporting an issue (bug) please provide the following information:

#### If applicable, please provide minimum samples of PHP code (as plain text, not screenshots):
 * before running PHP CS Fixer (no changes):
```php
echo doubleval('123.0') . "\n";

echo imap_create(null, 0) . "\n";

echo mbsplit('/./', 'str') . "\n";

die();

die('message');

exit;
```

 * with unexpected changes applied when running PHP CS Fixer:
```php
echo floatval('123.0') . "\n";

echo imap_createmailbox(null, 0) . "\n";

echo mb_split('/./', 'str') . "\n";

die();

die('message');

exit;
```

 * with the changes you expected instead:
```php
echo floatval('123.0') . "\n";

echo imap_createmailbox(null, 0) . "\n";

echo mb_split('/./', 'str') . "\n";

exit;

echo 'message';
exit;

exit;
```

I know there's a option called `no_alias_functions`, it could replace aliases of ``imap_*``, ``mb*``, and internal (``chop``, ``close``, ``doubleval``, ``fputs``, ``get_required_files``, ``ini_alter``, ``is_double``, ``is_integer``, ``is_long``, ``is_real``, ``is_writeable``, ``join``, ``key_exists``, ``magic_quotes_runtime``, ``pos``, ``show_source``, ``sizeof``, ``strchr``, ``user_error``) functions.

But I wish that there's an another option which called ``no_mixed_die_exit``, `no_alias_like_language_construct`, or anything we could understand. And it can fix alias-like language construct, for example, replace ``die();`` and ``die('str');`` with ``exit;`` and ``echo 'str'; exit;`` (working with ``no_mixed_echo_print``).

kubawerlos:
Can you give another example? :)

Because this is the only one missing - already discussed [here](https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/1692).

SpacePossum:
Would be a nice fixer :)
I think `no_mixed_die_exit` sounds alright, making it configurable. I don't think there are other language constructs with aliases out there?

kubawerlos:
I'm not sure about the configurable part - isn't `die` alias of `exit`?

SpacePossum:
[...]

PR #4740, opened 2020-01-14
NoAliasLanguageConstructCallFixer - introduction

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4717

### DIFF
--- a/tests/Fixer/Alias/NoAliasLanguageConstructCallFixerTest.php
+++ b/tests/Fixer/Alias/NoAliasLanguageConstructCallFixerTest.php
@@ -1,0 +1,64 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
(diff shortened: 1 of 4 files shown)


tests/Fixer/Alias/NoAliasLanguageConstructCallFixerTest.php
```diff
@@ -0,0 +1,57 @@
...
+            [
+                '<?php exit ("foo");',
+                '<?php die ("foo");',
+            ],
+            [
+                '<?php exit (1); EXIT(1);',
+                '<?php DIE (1); EXIT(1);',
+            ],
+            [
+                '<?php echo "die"; // die; /* die(1); */',
```
dmvdbrugge: Nitpicky comment: I would swap the `// die;` and `/* die(1) */` because the latter is now part of the former comment, instead of its own.

SpacePossum: :+1:  done

Issue #4653, opened 2019-11-20
Risky rule options


julienfalque:
Currently we are able to mark a rule as risky (meaning applying it might change code's behavior). Though this means every possible change that can be applied is considered risky.

Being able to mark only a single option of a non-risky rule would allow making rules that can apply non-risky changes by default and optionally apply risky changes. Current setup requires two separate rules.

Maybe a good use case would be #4652.

SpacePossum:
:+1:  , this would also allow for a risky-option for `no_trailing_whitespace` to just do it everywhere :)

dmvdbrugge:
As long as the enable/disable of all individual options are still under the user's control, I'm all for this! (So not magically enabling all risky options because "allow risky is true")

SpacePossum:
maybe not close but move to discussions, I run into this a bunch of times and would be really great to have this. (not sure about the product roadmap/prio's etc. of course)

keradus:
I believe idea is valid, yet not planned as a focus nowadays.
Any implementation contribution will be great ;)

(idea for moving to discussions is valid too)

PR #4741, opened 2020-01-15
TernaryToElvisOperatorFixer - introduction

SpacePossum:
closes https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/3485

### DIFF
--- a/src/Fixer/Operator/TernaryToElvisOperatorFixer.php
+++ b/src/Fixer/Operator/TernaryToElvisOperatorFixer.php
@@ -1,0 +1,269 @@
+<?php
+
+/*
+ * This file is part of PHP CS Fixer.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *     Dariusz Rumiński <dariusz.ruminski@gmail.com>
+ *
+ * This source file is subject to the MIT license that is bundled
+ * with this source code in the file LICENSE.
+ */
+
+namespace PhpCsFixer\Fixer\Operator;
+
+use PhpCsFixer\AbstractFixer;
+use PhpCsFixer\FixerDefinition\CodeSample;
+use PhpCsFixer\FixerDefinition\FixerDefinition;
+use PhpCsFixer\Tokenizer\CT;
+use PhpCsFixer\Tokenizer\Tokens;
+
+/**
+ * @author SpacePossum
+ */
+final class TernaryToElvisOperatorFixer extends AbstractFixer
+{
+    /**
+     * @internal
+     *
+     * Lower precedence and other valid preceding tokens.
+     *
+     * Ordered by most common types first.
+     *
+     * @var array
+     */
+    const VALID_BEFORE_ENDTYPES = [
+        '=',
+        [T_OPEN_TAG],
+        '(',
+        ',',
+        ';',
+        '[',
+        '{',
+        '}',
+        [CT::T_ARRAY_INDEX_CURLY_BRACE_OPEN],
+        [T_AND_EQUAL],    // &=
+        [T_CONCAT_EQUAL], // .=
+        [T_DIV_EQUAL],    // /=
+        [T_MINUS_EQUAL],  // -=
+        [T_MOD_EQUAL],    // %=
+        [T_MUL_EQUAL],    // *=
+        [T_OR_EQUAL],     // |=
+        [T_PLUS_EQUAL],   // +=
+        [T_POW_EQUAL],    // **=
+        [T_SL_EQUAL],     // <<=
+        [T_SR_EQUAL],     // >>=
+        [T_XOR_EQUAL],    // ^=
+    ];
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getDefinition()
+    {
+        return new FixerDefinition(
+            'Use the Elvis operator `?:` where possible.',
+            [
+                new CodeSample(
+                    "<?php\n\$foo = \$foo ? \$foo : 1;\n"
+                ),
+                new CodeSample(
+                    "<?php \$foo = \$bar[a()] ? \$bar[a()] : 1; # \"risky\" sample, \"a()\" only gets called once after fixing\n"
+                ),
+            ],
+            null,
+            'Risky when relying on functions called on both sides of the `?` operator.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     *
+     * Must run before NoTrailingWhitespaceFixer, TernaryOperatorSpacesFixer.
+     */
+    public function getPriority()
+    {
+        return 1;
+    }
+
+    /**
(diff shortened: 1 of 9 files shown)


src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,231 @@
...
+
+            $prevTokenIndex = $tokens->getPrevMeaningfulToken($index);
+
+            if ($tokens[$prevTokenIndex]->isGivenKind(T_VARIABLE)) {
+                $this->fixIfVariableMatches($tokens, $prevTokenIndex, $index);
+
+                continue;
+            }
+
+            if ($tokens[$prevTokenIndex]->equals(']')) {
```
SpacePossum: note to self: could be array curly index as well...

SpacePossum:
thanks for the check @gharlan , these cases are not supported indeed... I'll start adding support for these

SpacePossum:
@gharlan I basically rewrote the fixer and added a bunch of more tests. Could you review again? I would like to rule out the fixer is "fixing" to much now

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,267 @@
...
+
+        for ($index = \count($tokens) - 5; $index > 1; --$index) {
+            if (!$tokens[$index]->equals('?')) {
+                continue;
+            }
+
+            $nextIndex = $tokens->getNextMeaningfulToken($index);
+
+            if ($tokens[$nextIndex]->equals(':')) {
+                continue; // already Elvis!
```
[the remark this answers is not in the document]
SpacePossum: done :D

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,267 @@
...
+            if ($this->rangeEqualsRange($tokens, $beforeOperator, $afterOperator)) {
+                $this->clearMeaningfulFromRange($tokens, $afterOperator);
+            }
+        }
+    }
+
+    /**
+     * @param int $index
+     *
+     * @return null|array null if contains unary operator
```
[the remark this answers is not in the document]
SpacePossum: would checking for `++` and `--` be enough you think?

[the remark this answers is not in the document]
gharlan: I think so, yes.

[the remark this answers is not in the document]
SpacePossum: nice, thanks :)  updated!

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,265 @@
...
+ * @author SpacePossum
+ */
+final class TernaryToElvisOperatorFixer extends AbstractFixer
+{
+    /**
+     * Lower precedence and other valid preceding tokens.
+     *
+     * @var array
+     */
+    private static $validBeforeEndTypes = [
```
kubawerlos: This can be a constant, right? Seeing variable always suggests me that it can be updated in the runtime and this one is not.

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,265 @@
...
+        '=',
+        [T_AND_EQUAL],    // &=
+        [T_CONCAT_EQUAL], // .=
+        [T_DIV_EQUAL],    // /=
+        [T_DIV_EQUAL],    // =',
```
kubawerlos: delete once again?

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,265 @@
...
+        [T_SR_EQUAL],     // >>=
+        [T_XOR_EQUAL],    // ^=
+        '(',
+        '[',
+        '{',
+        '}',
+        ',',
+        ';',
+        [T_OPEN_TAG],
+        [CT::T_ARRAY_INDEX_CURLY_BRACE_OPEN],
```
kubawerlos: I know you can sort this array better ;)

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
```diff
@@ -0,0 +1,265 @@
...
+            'Risky when relying on functions called on both sides of the `?` operator.'
+        );
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getPriority()
+    {
+        // should be run before TernaryOperatorSpacesFixer and NoTrailingWhitespaceFixer
```
kubawerlos: I guess the new priority comments auto review test will fail on this.

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
SpacePossum: haha :man_facepalming:

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
SpacePossum: I hope so! lets see what a rebase does

src/Fixer/Operator/TernaryToElvisOperatorFixer.php
SpacePossum: thanks, done!

dmvdbrugge:
I think it's a shame that this has to be made risky only because of `foo() ? foo() : bar();`

Can we re-evaluate this choice? Perhaps put it behind an option? I'd love to use this fixer, but I cannot guarantee this case will work 😞

julienfalque:
@dmvdbrugge related to your reasoning: #4653.

dmvdbrugge:
@julienfalque that would be nice to have as well (I 👍'd it), but (for this fixer) that's a step that can happen _after_ the risky behaviour has been placed behind an option.

SpacePossum:
won't be that easy, as it is not limited to direct function calls only, for example;

```php
$a= new Foo(); // implements \ArrayAccess

$b = $a['one'] ? $a['one'] : 2; // Foo::offsetGet is called 2x
$b = $a['one'] ?  : 2;  // Foo::offsetGet is called 1x
```

I don't think it is worth leaving the risky cases out because it is complex and I don't think a lot of code bases are depended on the "called 2x"-behavior in a ternary statement (I think most people don't even know this happens and when and why)

dmvdbrugge:
Oof, that's true. And it doesn't stop there, `__call` and `__callStatic` could also trigger.

Never mind my initial comment then, it's fully risky all right.

